Issue #1745, opened 2018-05-17
sct_check_dependencies crashed after new installation


jcohenadad:
Email from a user (captured by sentry: SCT-28):

~~~
Last login: Thu May 17 14:29:48 on ttys001
-bash: /Users/makis/sct_3.0.1/bin/sct_env: No such file or directory
SatoshinoMacBook-Pro:~ makis$ git clone --depth=1 --branch=master https://github.com/neuropoly/spinalcordtoolbox.git sct
Cloning into 'sct'...
remote: Counting objects: 837, done.
remote: Compressing objects: 100% (534/534), done.
remote: Total 837 (delta 318), reused 609 (delta 288), pack-reused 0
Receiving objects: 100% (837/837), 5.30 MiB | 975.00 KiB/s, done.
Resolving deltas: 100% (318/318), done.
SatoshinoMacBook-Pro:~ makis$ cd sct
SatoshinoMacBook-Pro:sct makis$ ./install_sct

Welcome to the SCT installer!

SCT version ......... dev
Installation type ... in-place
Operating system .... osx
sudo mode ........... 0

To improve user experience and fix bugs, the SCT development team is using a
report system to automatically receive crash reports and errors from users.
These reports are anonymous.
Do you agree to help us improve SCT? [y]es/[n]o: y
--> Crash reports will be sent to the SCT development team. Thank you!

SCT will be installed here: [/Users/makis/sct]
Do you agree? [y]es/[n]o: y

Skipping copy of source files (source and destination folders are the same)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    91  100    91    0     0     36      0  0:00:02  0:00:02 --:--:--    36
exit status is 0

Create launchers for Python scripts
>> mkdir -p /Users/makis/sct/bin
>> cd /Users/makis/sct/install
>> ./update_bin.sh
#!/bin/bash
$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/sct_launcher $(basename $0).py $@
yes isct_check_detection
yes isct_convert_binary_to_trilinear
yes isct_minc2volume-viewer
yes isct_test_ants
yes isct_warpmovie_generator
yes msct_multiatlas_seg
yes sct_analyze_lesion
yes sct_analyze_texture
yes sct_apply_transfo
yes sct_average_data_within_mask
yes sct_check_atlas_integrity
[...]

Issue #1708, opened 2018-05-07
Sentry: display message to indicate how to opt out from Sentry 


jcohenadad:
### Description

It is likely that some users who have initially opted in for sentry (crash report) will be annoyed. They need to be able to opt out at anytime, hence this should be documented. One suggestion is to display a message like:
~~~
If you don't want to send crash report anymore, edit the file ${SCT_DIR}/bin/sct_launcher and delete the line that starts with "export SENTRY_DSN"
~~~

Ideally, this message should be displayed everytime an error is sent. I don't know if there is a method in `raven` that enables to add custom message upon connection.

zougloub:
Fixed by #1753

PR #1753, opened 2018-05-19
Random issue fixes (Sentry, sct_pipeline, install)

zougloub:
This PR brings:

- Some `sct_pipeline` minor fixes
- Fix for #1745: Mac installation on system where "ps" could print non-ASCII
- Fix for #1708: print Sentry opt-out message after sending a report
- Fix for one Sentry false positive https://sentry.io/neuropoly-k0/sct/issues/557250868 due to capture of a control flow exception
- Fix for many Sentry false positives: events due to users misusing command line arguments are not reported anymore
- Change `sct.__version__` format to be more human-readable (installed package will simply be the package version)

### DIFF
--- a/scripts/sct_check_dependencies.py
+++ b/scripts/sct_check_dependencies.py
@@ -96,6 +96,10 @@ def module_import(module_name, suppress_stderr=False):
 # MAIN
 # ==========================================================================================
 def main():
+    path_sct = os.environ.get("SCT_DIR", os.path.dirname(os.path.dirname(__file__)))
+    print("SCT info:")
+    print("- version: {}".format(sct.__version__))
+    print("- path: {0}".format(path_sct))
 
     # initialization
     fsl_is_working = 1
@@ -144,6 +148,7 @@ def main():
         os_running = 'osx'
     elif (platform_running.find('linux') != -1):
         os_running = 'linux'
+
     print('OS: ' + os_running + ' (' + platform.platform() + ')')
 
     # Check number of CPU cores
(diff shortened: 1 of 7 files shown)


scripts/sct_check_dependencies.py
```diff
@@ -154,15 +159,6 @@ def main():
     # check RAM
     sct.checkRAM(os_running, 0)
 
-    path_sct = os.environ.get("SCT_DIR", os.path.dirname(os.path.dirname(__file__)))
-    print('SCT path: {0}'.format(path_sct))
-
-    # fetch SCT version
-    install_type, sct_commit, sct_branch, version_sct = sct._git_info()
-    print('Installation type: %s' % install_type)
```
jcohenadad: don't we want that info anymore?

scripts/sct_check_dependencies.py
```diff
@@ -154,15 +159,6 @@ def main():
...
-    print('  version: ' + version_sct)
-    print('  commit: ' + sct_commit)
```
jcohenadad: i guess this is displayed for all SCT commands (including `sct_check_dependencies`) so we can indeed get rid of it

scripts/sct_deepseg_sc.py
```diff
@@ -224,7 +224,7 @@ def main():
         output_folder = os.getcwd()
 
     if '-r' in arguments:
-        remove_temp_files = arguments['-r']
+        remove_temp_files = int(arguments['-r'])
```
jcohenadad: oops, nice catch!

scripts/sct_pipeline.py
```diff
@@ -349,11 +351,11 @@ def run_function(function, folder_dataset, list_subj, list_args=[], nb_cpu=None,
             arguments = future_dirs[future][2]
             try:
                 result = future.result()
+                sct.log.info('{}/{}: {} done'.format(count, len(list_func_subj_args), subject))
```
jcohenadad: i've modified these lines in PR #1754 so there will be conflicts

scripts/sct_utils.py
```diff
@@ -118,21 +137,22 @@ def _git_info(commit_env='SCT_COMMIT',branch_env='SCT_BRANCH'):
...
+        version_sct = f.read().rstrip()
 
     return install_type, sct_commit, sct_branch, version_sct
 
+def _version_string():
+    install_type, sct_commit, sct_branch, version_sct = _git_info()
+    if install_type == "package":
+        return version_sct
+    else:
+        return "{install_type}-{sct_branch}-{sct_commit}".format(**locals())
```
jcohenadad: ah! ok i see, so disregard my [previous comment](https://github.com/neuropoly/spinalcordtoolbox/pull/1753/files#r189449085)

Issue #1740, opened 2018-05-17
Output cord segmentation is empty on image with reasonable contrast


stephaniealley:
### Description

The function runs successfully, but the resulting segmentation image is empty.

### Steps to Reproduce

Data: sct_testing/issues/20180517_issue1740

Run:
```
sct_deepseg_sc -i 9611-0_0-images_017_jvgre10TE2mm1001_trace15Cav.nii.gz -c t2s
```

### State of spinalcordtoolbox
```
master/9e294bb9dc06f980b12cf80f3fde6ab8416ac860
```

stephaniealley:
Note that the segmentation also fails for propseg.

jcohenadad:
I experience the same problem with `sct_deepseg_gm`, i.e., output segmentation is empty.

jcohenadad:
Investigating further, the issue with `sct_deepseg_sc` happens at the centerline detection: for each slice, the centerline is located at the R-P corner.
@charleygros Maybe a problem with the image intensity (<1), and type conversion done before running Optic (float32 --> uint)?

charleygros:
@jcohenadad : Yes it is certainly caused by this line:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/centerline/optic.py#L95

perone:
So, this volume doesn't seem to be unnormalized (the intensities seems to be zero centered, so they probably did a zscore normalization). Our models assume that the input isn't already normalized because we have to do normalization on the input, so if we feed this normalized data, there will be another normalization done by the algorithm, shifting the distributions depending on the normalization employed in the algorithm. Are they able to send the original data without pre-processing ? We had the same issue with the marmoset dataset, where they sent the data with bias corrections, etc.

As a side note, I'm working actively to find a better normalization (both for input of the model and also inside the model) as part of the domain adaptation work, so after finishing this we can change the way we do normalization, but right now it's risky to change this just for this subject because it will involve retraining and re-validating the entire pipeline again. I spoke with @charleygros and on his case he is doing a different normalization so he will not have the same problems that I have on `sct_deepseg_gm`, but there is always a risk of things going wrong when two stages of normalization is applied on the input data, so it's always good to make sure that the data labs send, are not pre-processed because every lab can do a different pre-processing or corrections that will break our assumptions.

jcohenadad:
@perone two things
[...]

PR #1757, opened 2018-05-19
Adding volume-wise standardization normalization for the sct_deepseg_gm

perone:
This PR will add:

- Two new classes for normalization: `StandardizationTransform` and `VolumeStandardizationTransform`. One is intended to use if we want to go back to the previous normalization mechanism and the other is the new one for doing volume-wise normalization;
- New test cases to test the added classes;
- Fixed some tab/spacing mix;

For more information, please see the discussion in #1740.


spinalcordtoolbox/deepseg_gm/deepseg_gm.py
```diff
@@ -183,8 +227,8 @@ def segment_volume(ninput_volume, model_name):
     axial_slices = np.asarray(axial_slices, dtype=np.float32)
     axial_slices = np.expand_dims(axial_slices, axis=3)
 
-    axial_slices -= metadata['mean_train']
-    axial_slices /= metadata['std_train']
+    normalization = VolumeStandardizationTransform()
+    axial_slices = normalization(axial_slices)
```
zougloub: Use the more accurate term standardization (vs. normalization)?

spinalcordtoolbox/deepseg_gm/deepseg_gm.py
```diff
@@ -113,6 +113,50 @@ def pad(self, image):
...
+        """
+        volume_mean = volume.mean()
+        volume_std = volume.std()
+
+        volume -= volume_mean
+        volume /= volume_std
+
+        return volume
+
+
```
zougloub: A functor seems overkill.

jcohenadad:
@perone let me know when you're happy with the state of this PR, and I'll launch the extensive test with `sct_pipeline` and add violin plots for master and this branch to compare.

spinalcordtoolbox/deepseg_gm/deepseg_gm.py
perone: Standardization is a type of normalization.

spinalcordtoolbox/deepseg_gm/deepseg_gm.py
perone: This is the usual pattern for transformations [in PyTorch for instance](https://github.com/pytorch/vision/blob/master/torchvision/transforms/transforms.py#L120), even for the same purpose, so I don't think it's overkill, but if you want me to change, please tell me what you prefer so I can change it.

perone:
@jcohenadad please go ahead, the functionality is this one. If something changes, it will be just minor cosmetic changes.

perone:
@jcohenadad I'll take a look at the logs, thanks for the testing.

PR #1759, opened 2018-05-21
New function to create violin plots from sct_pipeline results

jcohenadad:
### Description of the Change

Implements a new function to generate visual results from `sct_pipeline`. 

Example usage:
```python
sct_pipeline_makefig -i branch-to-test.pickle master.pickle -label branch-to-test master
```
Generated figure:
violin_plot

Future improvements:
- Implement metadata inside `sct_pipeline` results and use them in this function (notably to get rid of the flag `label`)


scripts/sct_pipeline_makefig.py
```diff
@@ -0,0 +1,82 @@
...
+
+# TODO: generalize this code for other things than Dice
+
+import argparse
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(
+        description='Generate figure from the output of sct_pipeline.')
+    parser.add_argument("-i",
```
zougloub: When using argparse for new code, suggest to use short/long arguments:
```
 "-i", "--input",
"-l", "--label",
"-v", "--verbose",
```

scripts/sct_pipeline_makefig.py
```diff
@@ -0,0 +1,82 @@
...
+def main(args):
+    import sct_utils as sct
+    import pickle
+    import numpy as np
+    import matplotlib.pyplot as plt
+
+    sct.start_stream_logger()
+
+    # fs = 10  # font size
+    nb_plots = args.i.__len__()
```
zougloub: This way, the long argument name will be used as default attribute name (`args.input` if above was used).

scripts/sct_pipeline_makefig.py
```diff
@@ -0,0 +1,82 @@
...
+
+    list_data = []
+    text_results = []  # numerical results to display inside the figure
+    for fname_pickle in args.i:
+        df = pickle.load(open(fname_pickle, "rb"))
```
zougloub: Could be changed to `io.open()` (more explanatory IMO, but it's a question of taste/habits)

scripts/sct_pipeline_makefig.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+#
+# Generate figure from the output of sct_pipeline.
+#
+# Copyright (c) 2013 Polytechnique Montreal <www.neuro.polymtl.ca>
```
zougloub: 2018

scripts/sct_pipeline_makefig.py
```diff
@@ -0,0 +1,82 @@
...
+    ax.spines['left'].set_visible(False)
+    ax.spines['top'].set_visible(False)
+    for i in range(nb_plots):
+        plt.text(i + 0.01, 0.01, text_results[i], horizontalalignment='left', verticalalignment='bottom')
+    plt.savefig('violin_plot.png')
+
+
+if __name__ == '__main__':
+    parser = get_parser()
+    arguments = parser.parse_args()
```
zougloub: Although you can invoke the program with `-i file_1 -label A -i file_2 -label B` you will invariably have users who end up doing `-i file_1 file_2 -label A B` and end up with `-i file_1 file_2 -label A`.
```
if len(arguments.i) != len(arguments.label):
    raise RuntimeError("Mismatch between # of files and labels")
```

Perhaps the labels should be the file names, if not provided.

scripts/sct_pipeline_makefig.py
jcohenadad: fixed in 9d77c82139345e4600a708702d8ba06729c7a1f6

scripts/sct_pipeline_makefig.py
jcohenadad: fixed in 9d77c82139345e4600a708702d8ba06729c7a1f6

scripts/sct_pipeline_makefig.py
jcohenadad: fixed in 9d77c82139345e4600a708702d8ba06729c7a1f6

scripts/sct_pipeline_makefig.py
jcohenadad: fixed in 9d77c82139345e4600a708702d8ba06729c7a1f6

scripts/sct_pipeline_makefig.py
jcohenadad: fixed in 9d77c82139345e4600a708702d8ba06729c7a1f6

PR #1798, opened 2018-05-31
Now possible to specify a vertebral labeling file when using -vert

### LINKED ISSUE #1793: UnboundLocalError: local variable 'suffix_vertebral_labeling' referenced before assignment
Here's what happens when trying to extract metric using a single file (not the typical `label/` folder passed by `-f` flag) combined with the `-vert` flag. Ideally, we should be able to specify a vertebral labeling file, as with the `-vertfile` flag in `sct_process_segmentation`.

~~~
sct_extract_metric -i dti_FA.nii.gz -f label/template/PAM50_wm.nii.gz -vert 2 -o fa.xls

--
Spinal Cord Toolbox (master/fc72f34d4251de09d3e92071c66b40ac7fe4c92a)
Running /Users/julien/code/sct/scripts/sct_extract_metric.py -i dti_FA.nii.gz -f label/template/PAM50_wm.nii.gz -vert 2 -o fa.xls
  File "/Users/julien/code/sct/scripts/sct_extract_metric.py", line 1222, in <module>
    main(fname_data, path_label, method, slices_of_interest, vertebral_levels, fname_output, labels_user, overwrite, fname_normalizing_label, normalization_method, label_to_fix, adv_param_user, fname_output_metric_map, fname_mask_weight)
  File "/Users/julien/code/sct/scripts/sct_extract_metric.py", line 233, in main
    fname_vertebral_labeling_list = sct.find_file_within_folder(suffix_vertebral_labeling, os.path.dirname(path_label))
UnboundLocalError: local variable 'suffix_vertebral_labeling' referenced before assignment
~~~

jcohenadad:
Fixes #1793

### DIFF
--- a/scripts/sct_extract_metric.py
+++ b/scripts/sct_extract_metric.py
@@ -390,25 +390,6 @@ def main(fname_data, path_label, method, slices_of_interest, vertebral_levels, f
         fixed_label = [label_to_fix[0], label_to_fix_name, label_to_fix[1]]
         sct.printv('\n*' + fixed_label[0] + ', ' + fixed_label[1] + ': ' + fixed_label[2] + ' (value fixed by user)', 1, 'info')
 
-    # section = ''
-    # if labels_id_user[0] <= max(indiv_labels_ids):
-    #     section = '\nWhite matter atlas:'
-    # elif labels_id_user[0] > max(indiv_labels_ids):
-    #     section = '\nCombined labels:'
-    # sct.printv(section, 1, 'info')
-    # for i_label_user in labels_id_user:
-    #     # change section if not individual label anymore
-    #     if i_label_user > max(indiv_labels_ids) and section == '\nWhite matter atlas:':
-    #         section = '\nCombined labels:'
-    #         sct.printv(section, 1, 'info')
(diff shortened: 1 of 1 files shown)


scripts/sct_extract_metric.py
```diff
@@ -112,6 +112,11 @@ def get_parser():
                       mandatory=False,
                       example='2:5',
                       default_value=param_default.vertebral_levels)
+    parser.add_option(name='-vertfile',
+                      type_value='str',  # note: even though it's a file, we cannot put the type='file' otherwise the full path will be added in sct_testing and it will crash
```
zougloub: I'm not sure this is the right solution to the problem. We lose error checking in the process.

jcohenadad: as mentioned in my comment, if we put "file", the full path will be added to sct_testing and then it will crash because it does not find the file. I know this is a current fundamental problem in the way sct_testing works, but that would require several hours to fix that. I don't have those few hours for the next few weeks, and having -vertfile is more important than having the file check. So this is just a question of priority. We can open an issue to address the file checking.

PR #1800, opened 2018-05-31
Forcing output type to be float32

### LINKED ISSUE #1790: Image conversion issue (again)
Data: `sct_testing/large/glen_spine-generic_20171128/t1`

~~~
sct_flatten_sagittal -i t1.nii.gz -s t1_seg_manual.nii.gz

--
Spinal Cord Toolbox (master/fc72f34d4251de09d3e92071c66b40ac7fe4c92a)
Running /Users/julien/code/sct/scripts/sct_flatten_sagittal.py -i t1.nii.gz -s t1_seg_manual.nii.gz

Smooth centerline/segmentation...
.. Get center of mass of the centerline/segmentation...
.. Smoothing algo = hanning
.. Windows length = 50
WARNING: To avoid intensity overflow due to convertion to uint16, intensity will be rescaled to the maximum quantization scale.
  File "/Users/julien/code/sct/scripts/sct_flatten_sagittal.py", line 160, in <module>
    main(fname_anat, fname_centerline, degree_poly, centerline_fitting, interp, remove_temp_files, verbose)
  File "/Users/julien/code/sct/scripts/sct_flatten_sagittal.py", line 77, in main
    im_anat_flattened.data[:, :, iz] = img_as_uint(img_reg)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/skimage/util/dtype.py", line 325, in img_as_uint
    return convert(image, np.uint16, force_copy)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/skimage/util/dtype.py", line 205, in convert
    raise ValueError("Images of type float must be between -1 and 1.")
ValueError: Images of type float must be between -1 and 1.
~~~

Possibly related to #1449, #1740, #1691

jcohenadad:
### Description of the Change

Output type is now float32, regardless of input type. This prevents loss of precision when converting back to input type (e.g. if user inputs int16, converting to float32 and then back to int16 will yield loss of precision). The drawback is a potentially larger file. 

Fixes #1790

### DIFF
--- a/scripts/sct_flatten_sagittal.py
+++ b/scripts/sct_flatten_sagittal.py
@@ -64,7 +64,8 @@ def main(fname_anat, fname_centerline, degree_poly, centerline_fitting, interp, 
 
     # loop across slices and apply translation
     im_anat_flattened = im_anat.copy()
-    im_anat_flattened.changeType('uint16')  # force uint16 because outputs are converted using img_as_uint()
+    # change type to float32 because of subsequent conversion (img_as_float). See #1790
+    im_anat_flattened.changeType('float32')
     for iz in range(nz):
         # compute translation along x (R-L)
         translation_x = x_centerline_extended[iz] - round(nx/2.0)
@@ -74,7 +75,7 @@ def main(fname_anat, fname_centerline, degree_poly, centerline_fitting, interp, 
         # important to force input in float to skikit image, because it will output float values
         img = img_as_float(im_anat.data[:, :, iz])
         img_reg = transform.warp(img, tform)
-        im_anat_flattened.data[:, :, iz] = img_as_uint(img_reg)
+        im_anat_flattened.data[:, :, iz] = img_reg  # img_as_uint(img_reg)
 
     # change back to native orientation
     im_anat_flattened.change_orientation(orientation_native)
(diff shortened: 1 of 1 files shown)


scripts/sct_flatten_sagittal.py
```diff
@@ -64,7 +64,8 @@ def main(fname_anat, fname_centerline, degree_poly, centerline_fitting, interp,
 
     # loop across slices and apply translation
     im_anat_flattened = im_anat.copy()
-    im_anat_flattened.changeType('uint16')  # force uint16 because outputs are converted using img_as_uint()
+    # change type to float32 because of subsequent conversion (img_as_float). See #1790
+    im_anat_flattened.changeType('float32')
```
abtahizadeh:
```
try: 
    im_anat_flattened.changeType('uint16')
except ValueError:
    im_anat_flattened.changeType('float32')

```

zougloub:
I wouldn't worry about the bigger size...

scripts/sct_flatten_sagittal.py
jcohenadad: that would not work because `im_anat_flattened.changeType('uint16')` does not produce an error.

abtahizadeh: Do you mean that the subsequent conversion is always *img_as_float* ?

jcohenadad: yes, we always do img_as_float, therefore the dtype need to be prepared, otherwise if an [0,1] float array is input, we end up with a matrix full of zero.

PR #1814, opened 2018-06-04
Fixed wrong disc labeling and other minor improvements

### LINKED ISSUE #1810: Change colormap for fsleyes syntax
### Description

The current colormap **Random** makes it difficult to distinguish the three first levels because it uses three very similar green colors. I suggest we use the **MGH sub-cortical** colormap instead:

**Random** (current)
screen shot 2018-06-04 at 10 24 00

**MGH sub-cortical** (suggested)
screen shot 2018-06-04 at 10 23 50

### LINKED ISSUE #1812: The generated labeled_disc does not follow the convention
### Description

Output label values are C2/C3-->2 instead of C2/C3-->3.

jcohenadad:
Fixes #1812, Fixes #1810

### DIFF
--- a/scripts/sct_label_vertebrae.py
+++ b/scripts/sct_label_vertebrae.py
@@ -907,7 +907,7 @@ def label_segmentation(fname_seg, list_disc_z, list_disc_value, verbose=1):
 
 def label_discs(fname_seg_labeled, verbose=1):
     """
-    Label discs from labaled_segmentation
+    Label discs from labeled_segmentation. The convention is C2/C3-->3, C3/C4-->4, etc.
     :param fname_seg_labeld: fname of the labeled segmentation
     :param verbose:
     :return:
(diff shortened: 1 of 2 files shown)


scripts/sct_label_vertebrae.py
```diff
@@ -917,15 +917,15 @@ def label_discs(fname_seg_labeled, verbose=1):
...
+    vertebral_level_previous = np.max(im_seg_labeled.data)  # initialize with the max label value
+    # loop across z in the superior direction (i.e. starts with the bottom slice)
     for iz in range(nz):
         # get 2d slice
         slice = im_seg_labeled.data[:, :, iz]
         # check if at least one voxel is non-zero
         if np.any(slice):
             slice_one = np.copy(slice)
-            # set all non-zero values to 1
+            # set all non-zero values to 1 (to compute center of mass)
```
zougloub: Just curious, why not use center of mass with the original slice?

scripts/sct_label_vertebrae.py
```diff
@@ -917,15 +917,15 @@ def label_discs(fname_seg_labeled, verbose=1):
     orientation_native = im_seg_labeled.change_orientation('RPI')
     nx, ny, nz = im_seg_labeled.dim[0], im_seg_labeled.dim[1], im_seg_labeled.dim[2]
     data_disc = np.zeros([nx, ny, nz])
-    vertebral_level_previous = np.max(im_seg_labeled.data)
-    # loop across z
+    vertebral_level_previous = np.max(im_seg_labeled.data)  # initialize with the max label value
+    # loop across z in the superior direction (i.e. starts with the bottom slice)
```
zougloub: Just curious: why?

scripts/sct_label_vertebrae.py
jcohenadad: if one slice includes part of vertebral level i and i+1, the center of mass will be shifted towards the i+1 level. We don't want that here (i.e. the goal is to be at the center of the cord)

scripts/sct_label_vertebrae.py
jcohenadad: comments added: https://github.com/neuropoly/spinalcordtoolbox/pull/1814/commits/c7b4b1eb62fa4a05e25e760a9a2ce31f25dcc54f

scripts/sct_label_vertebrae.py
zougloub: Thanks!

PR #1828, opened 2018-06-06
Fixed registration issue caused by labels far from cord centerline

### LINKED ISSUE #1826: Wrong template registration outside of labels
### Description

The error below has been introduced by the recent use of disc labels (as opposed to labels inside the spinal cord). At step=0, a rigid transformation is estimated between the straightened labels and the template label. The problem is: if the labels were not put in the cord, but rather on the discs, then the rigid transformation will fail to properly align the cord on the template.

A solution would be to project the labels inside the cord centerline, after straightening. 

data: sct_006/t2 (from sct_course)

The segmentation was correct all the way to the bottom, so the following error should not happen:

screen shot 2018-06-06 at 15 45 24

jcohenadad:
### Description of the Change

We added a step, before straightening, that orthogonally projects labels onto the cord centerline. This step ensures that the subsequent affine transformation (label_subject --> label_template) is working properly.

### Applicable Issues

Fixes #1826 

### Example

Before fix:
41021271-c823f34a-69a0-11e8-91e8-4c2baeecf386

After fix:
screen shot 2018-06-06 at 20 55 50

### DIFF
--- a/scripts/sct_register_to_template.py
+++ b/scripts/sct_register_to_template.py
@@ -15,7 +15,7 @@
 # TODO: enable vertebral alignment with -ref subject
 
 import sys, io, os, shutil, time
-
+import numpy as np
 import sct_utils as sct
 import sct_label_utils
 import sct_convert
@@ -24,7 +24,7 @@ from sct_utils import add_suffix
 from sct_register_multimodal import Paramreg, ParamregMultiStep, register
 from msct_parser import Parser
 from msct_image import Image, find_zmin_zmax
-
+from sct_straighten_spinalcord import smooth_centerline
 
 # get path of the toolbox
 path_script = os.path.dirname(__file__)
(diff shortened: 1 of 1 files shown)


scripts/sct_register_to_template.py
```diff
@@ -683,6 +690,52 @@ def generate_qc(fname_data, fname_template2anat, fname_seg, args, path_qc):
...
+    labels = im_label.getCoordinatesAveragedByValue()
+    # initialize image of projected labels
+    im_label_projected = im_label.copy()
+    im_label_projected.data = np.zeros(im_label_projected.data.shape, dtype='uint8')
+    # loop across label values
+    for label in labels:
+        # calculate distance between label and each point of the centerline
+        distance_centerline = [np.linalg.norm([centerline_x[i] - label.x,
+                                               centerline_y[i] - label.y,
+                                               centerline_z[i] - label.z]) for i in range(len(centerline_x))]
```
zougloub: If getCoordinatesAveragedByValue() and smooth_centerline() were returning arrays, the syntax could be prettier.

jcohenadad: agreed, another thing to add to the todo list :-)

PR #1838, opened 2018-06-08
The init parameter should accept only positive values

### LINKED ISSUE #1328: The `init` parameter should accept only positive values
### Actual Behavior

For the command `sct_propseg -i A.nii.gz -c t1 -init -15`, throws an exception. 
```
Exception caught while updating cropFilter

itk::InvalidRequestedRegionError (0x22b9770)
Location: "unknown"
File: /home/brain/Downloads/InsightToolkit-4.6.0/Modules/Core/Common/src/itkDataObject.cxx
Line: 393
Description: Requested region is (at least partially) outside the largest possible region.

Exception caught while computing symmetry
/usr/local/itk/include/ITK-4.6/itkImageConstIteratorWithIndex.hxx:86:
itk::ERROR: Region ImageRegion (0x7ffcf5020c20)
Dimension: 2
Index: [0, 0]
Size: [447, 499]
is outside of buffered region ImageRegion (0x22b3450)
Dimension: 2
Index: [0, 0]
Size: [0, 0]

/home/tpl065/sct_3.0_beta27/scripts/sct_utils.pyNone
```

### Expected Behavior

The command should exit with the error `"-15"` is not a valid value for `"-init"`

abtahizadeh:
### Requirements

* [ *] Is there a issue open describing the motivation for this pull request ?
* [ ] Does the pull request need an update of the documentation ?
* [* ] Does the pull request need additional test(s) ?
**non-positive value should be introduced**

* [ ] Clean up the warnings generated by the [style guide](https://www.jetbrains.com/help/pycharm/2017.1/code-inspection.html)

### Description of the Change

Explain the approach and possible drawback(s). Use labels to add context.

### Steps and Constraints

- Please check if zero is accepted as an input (I'm not aware of the algorithmic side)

### Applicable Issues

Implements or Fixes #1328

### DIFF
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -169,9 +169,15 @@ def download_data(urls, verbose):
             session.mount('https://', HTTPAdapter(max_retries=retry))
             response = session.get(url, stream=True)
 
-            _, content = cgi.parse_header(response.headers['Content-Disposition'])
-            tmp_path = os.path.join(tempfile.mkdtemp(), content['filename'])
-            sct.printv('Downloading %s...' % content['filename'], verbose)
+            if "Content-Disposition" in response.headers:
+                _, content = cgi.parse_header(response.headers['Content-Disposition'])
+                filename = content["filename"]
+            else:
+                sct.printv("Unexpected: link doesn't provide a filename", type="warning")
+                continue
+
+            tmp_path = os.path.join(tempfile.mkdtemp(), filename)
+            sct.printv('Downloading %s...' % filename, verbose)
 
             with open(tmp_path, 'wb') as tmp_file:
                 total = int(response.headers.get('content-length', 1))
@@ -188,8 +194,8 @@ def download_data(urls, verbose):
                 tqdm_bar.close()
             return tmp_path
 
-        except requests.RequestException as err:
-            sct.printv(err.message, type='warning')
+        except Exception as e:
+            sct.printv("Link download error, trying next mirror (error was: %s)" % e, type='warning')
     else:
         sct.printv('\nDownload error', type='error')
 


abtahizadeh:
@jcohenadad ayyyyyyyyyyy I am sooo sorry! using an automated script in my laptop causes this ... :( I forgot to fix it :-(

abtahizadeh:
Ignore various format changes (autopep8 applied)
@jcohenadad @zougloub aren't you using pep8 or pylint auto-format? every time I save a code, this will happen

jcohenadad:
@abtahizadeh i removed "bug" pls see: https://github.com/neuropoly/spinalcordtoolbox/wiki/Contribution-Guidelines#labels

abtahizadeh:
```
if the change fixes a error in the code.
```
not related?

jcohenadad:
about [this](https://github.com/neuropoly/spinalcordtoolbox/pull/1838#issuecomment-395618221): i let @zougloub follow-up-- there was already some discussion on this

jcohenadad:
@abtahizadeh: 
> changes [...] can only have one category

if you find the guidelines unclear/inappropriate pls suggest improvements-- thx

abtahizadeh:
@jcohenadad guidelines are specific to project owner, it's really your conventions. I just have to follow, will pay attention next time.

abtahizadeh:
> The job exceeded the maximum time limit for jobs, and has been terminated.

jcohenadad:
ah ok, i didn't know you were referring to travis-- i just relaunched the build on travis-- it happens sometimes (it's a free service after all :-) )

abtahizadeh:
Relevant discussion: https://github.com/travis-ci/travis-ci/issues/6232

zougloub:
On https://github.com/neuropoly/spinalcordtoolbox/pull/1838#issuecomment-395618221:

- The discussions we had before were about enforcing PEP8 compliance using Travis, or not.  
So far we said no to that, based on the fact that unless we all use the same linter as Travis, a lot of time would be spent to fix everything. Not to comment on the fact that automatic indentation sometimes breaks "creative" alignment that is often used in algorithmic code.

- It's OK if you do fixes for formatting issues, better if you could put them in separate commits (using git's stage lines feature) that would be better to separate the actual changes from style changes.

- As for personal stances:
  - PEP8 is not 100% dogmatic, and you will see me not respecting the "one import per line" recommendation when it comes to base stdlib modules (`import sys, io, os, re, datetime, collections`).
  - The only thing that I will systematically fix when touching a file, is trailing whitespace.
  - In this particular commit, the change that adds whitespace to do alignment is not fixing a PEP8 issue though, I believe the previous code was compliant to PEP8. If it were me I wouldn't have "fixed" it. And I find that aligning continuation lines with the same indentation as a code block is less legible; to the opening parenthesis is better, but breaks down in case of function rename (realign everything...), I tend to use a single whitespace for continuation lines, which works in all languages... And I'm in the "indent with tabs, align with spaces" camp of aliens.

scripts/sct_propseg.py
```diff
@@ -10,16 +10,16 @@
...
-from scipy import ndimage as ndi
+import sys
+
 import numpy as np
-from sct_image import orientation
 import sct_image
-
+import sct_utils as sct
+from msct_parser import Parser
+from scipy import ndimage as ndi
```
zougloub: Preferred order is stdlib, 3rd party, then internal; scipy is 3rd party and should go after numpy.

scripts/sct_propseg.py
```diff
@@ -10,16 +10,16 @@
 #
 # About the license: see the file LICENSE.TXT
 #########################################################################################
-from msct_parser import Parser
-import sys
-import sct_utils as sct
 import os
 import shutil
```
zougloub: Is shutil still used? I believe it could be removed.

scripts/sct_propseg.py
```diff
@@ -416,6 +415,9 @@ def generate_qc(fn_in, fn_seg, args, path_qc):
             use_optic = False
     if "-init" in arguments:
         init_option = float(arguments["-init"])
+        if (init_option < 0):
+            sct.log.error('Error! Negative number asigned! Please provide a valid value for "-init"')
```
zougloub: There's a certain command-line error pattern in msct_parter which makes such errors not caught by Sentry. You should use that format (sorry, it's not documented).

scripts/sct_propseg.py
abtahizadeh: I am using "isort", auto-sort option generates this. I belive it might be because of importing from "class-containing module" convention.

scripts/sct_propseg.py
abtahizadeh: "msct_parser" you mean? can you document the format please?

scripts/sct_propseg.py
abtahizadeh: done

scripts/sct_propseg.py
zougloub: Look at `server_log_handler()` in sct_utils. The Sentry logging handler is monkey-patched to discard certain (user PEBKAC) error messages.

Issue #1844, opened 2018-06-10
No need to call Sentry if user quits with CTRL+C


jcohenadad:
~~~
sct_download_data -d deepseg_sc_models

--
Spinal Cord Toolbox (cJ-misc/5fe166c4f750a32192ee110a3460a00e9f5ec1cf)
Running /Users/julien/code/sct/scripts/sct_download_data.py -d deepseg_sc_models

Trying URL: https://osf.io/mg8wr/?action=download
Downloading 20180211_deepseg_sc_models.zip...
Status:   0%|2                                                   | 172K/37.2M [00:00<02:58, 207KB/s]^C
ERROR: User canceled process.


Traceback (most recent call last):
  File "/Users/julien/code/sct/scripts/sct_download_data.py", line 205, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_download_data.py", line 111, in main
    sct.printv('\nERROR: User canceled process.\n', 1, 'error')
  File "/Users/julien/code/sct/scripts/sct_utils.py", line 1093, in printv
    raise RuntimeError("printv(..., type=\"error\")")
RuntimeError: printv(..., type="error")
Sentry is attempting to send 1 pending error messages
Waiting up to 10 seconds
Press Ctrl-C to quit
Note: you can opt out of Sentry reporting by editing the file ${SCT_DIR}/bin/sct_launcher and delete the line starting with "export SENTRY_DSN"
~~~

zougloub:
Fixed by #1872

Issue #1620, opened 2018-02-24
sct_testing TLS memory usage grows beyond limit


zougloub:
It seems that TensorFlow is hungry on TLS memory and this causes sct_testing to fail (on py3k for now) in Travis.

```
ImportError: dlopen: cannot load any more object with static TLS
```

Since `sct_testing`is importing all the code and it's not yet possible to un-import modules, we will end up having a problem.
Suggest to change `sct_testing` to use multiprocessing or subprocesses, to avoid this issue.

jcohenadad:
Fixed by #1872

Issue #1539, opened 2017-11-24
Performance improvements for the tests


perone:
By using Python multiprocessing capabilities, we can launch each test from the `sct_testing` in parallel instead of executing them serially. I did a small test and it improved ~3.4x the execution time of the tests on a machine with 4 cores, however, more work is required to accommodate the logging messages during the parallel execution.

zougloub:
Not only for the resulting performance improvement, but because of #1620 we could also end up using multiprocessing.

joshuacwnewton:
This has been addressed by #1872. Closing.

Issue #1658, opened 2018-03-23
sct_utils create_folder and exception handling/throwing (built-in exceptions)


perone:
I opened this issue to discuss the problem of the exception handling and throwing built-in exceptions.

So, there is an important issue and I think that we should start addressing these kinds of issues.
The `create_folder` from `sct_utils` looks more like a C-API than a Python API, it is handling an exception, silencing it and returning an error code (that is already present in the exception itself). Now what happens is that **everyone calling this function will have to handle this error code and throw an exception again later**, as we can see in the PR #1592.

However, it would be much more appropriate in my opinion, to have the `create_folder()` function to create the directory or throw an exception if it wasn't able to create it, that's what is the purpose of the exception being silenced in the `create_folder` function.

Now, every platform has their own errors and messages, the `errno` attribute in the `OSError` exception is an especial code, and according to this code, Python will throw another kind of exception (such as `PermissionError` for errno = 1 or `FileNotFoundError` for errno = 2). That is also another reason to avoid throwing built-in exceptions as I mentioned before because an `OSError` is only thrown by system errors and not incidental errors (as described in the [official docs](https://docs.python.org/2/library/exceptions.html#exceptions.OSError)).

Besides that, what can happen if we start to catch this error, return a code and then raise it again without a proper `errno` defined is that it is very difficult to handle it correctly and get the Sentry errors with meaningful information. That's why I think that we shouldn't use built-in system exceptions that aren't meant for incidental errors and instead use the traditional `RuntimeError` or another custom exception that inherits from the `Exception` class when we want to propagate structured data together with the exception.

Drulex:
Noticed this pattern in a couple of utility functions during the refactor (#2949). None of the callers checks the return codes so I removed the C style error handling and refactored the callers to handle the exceptions.

PR #1872, opened 2018-06-15
Misc. development fixes & improvements

zougloub:
This PR brings:
- `sct_testing` usage of multiprocessing (#1539)
  - Fix for #1620 (sct_testing TLS memory usage grows beyond limit) (unless `-j 1` is called)
- `sct_testing` usage of argparse (no problem since it's a script used for development only)
- `sct_download_data` to not send Sentry messages (#1844)
- docs
- since coveralls was complaining about coverage... generate coverage info from `sct_testing`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,11 +63,14 @@ script:
 # test functions & integrity
   - bash -c "sudo apt-get install -qq libalglib-dev libinsighttoolkit4-dev; apt-cache show libvtk6-dev && sudo apt-get install -qq libvtk6-dev; apt-cache show libvtk7-dev && sudo apt-get install -qq libvtk7-dev; cd $HOME/build/neuropoly/spinalcordtoolbox/dev/isct_propseg; mkdir -p build; cd build; cmake ..; make; echo"
   - pip install coverage
-  - SCT_PYTHON_WRAPPER="coverage run --parallel-mode --concurrency=multiprocessing" sct_testing -d 1
+  - bash -c "echo 'import coverage; cov = coverage.process_startup()' > sitecustomize.py"
+  - bash -c 'echo -ne "[run]\nconcurrency = multiprocessing\nparallel = True\n" > .coveragerc'
+  - mkdir -p tmp
+  - TMPDIR="$PWD/tmp" COVERAGE_PROCESS_START="$PWD/.coveragerc" PYTHONPATH="$PWD" sct_testing -d 1
 #  - python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
 #  - cd ../spinalcordtoolbox_v*
 #  - yes | ./install_sct  # test installation of package
-  - coverage combine
+  - coverage combine tmp/*/.coverage*
   - ./unit_testing.sh
   - bash -c "pip install sphinx sphinxcontrib.programoutput; cd $HOME/build/neuropoly/spinalcordtoolbox/documentation/sphinx; make html; echo"
   - echo "This is the end, thanks for reading up to here."
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,14 +62,16 @@ script:
 #  - sct_check_dependencies
 # test functions & integrity
   - bash -c "sudo apt-get install -qq libalglib-dev libinsighttoolkit4-dev; apt-cache show libvtk6-dev && sudo apt-get install -qq libvtk6-dev; apt-cache show libvtk7-dev && sudo apt-get install -qq libvtk7-dev; cd $HOME/build/neuropoly/spinalcordtoolbox/dev/isct_propseg; mkdir -p build; cd build; cmake ..; make; echo"
-  - sct_testing -d 1
+  - pip install coverage
+  - SCT_PYTHON_WRAPPER="coverage run --parallel-mode --concurrency=multiprocessing" sct_testing -d 1
 #  - python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
 #  - cd ../spinalcordtoolbox_v*
 #  - yes | ./install_sct  # test installation of package
+  - coverage combine
   - ./unit_testing.sh
   - bash -c "pip install sphinx sphinxcontrib.programoutput; cd $HOME/build/neuropoly/spinalcordtoolbox/documentation/sphinx; make html; echo"
   - echo "This is the end, thanks for reading up to here."
-  - pip install coveralls coverage
+  - pip install coveralls
 
 after_success:
   - CI=true TRAVIS=true coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,17 +63,17 @@ script:
 # test functions & integrity
   - bash -c "sudo apt-get install -qq libalglib-dev libinsighttoolkit4-dev; apt-cache show libvtk6-dev && sudo apt-get install -qq libvtk6-dev; apt-cache show libvtk7-dev && sudo apt-get install -qq libvtk7-dev; cd $HOME/build/neuropoly/spinalcordtoolbox/dev/isct_propseg; mkdir -p build; cd build; cmake ..; make; echo"
   - pip install coverage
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -62,14 +62,18 @@ script:
 #  - sct_check_dependencies
 # test functions & integrity
   - bash -c "sudo apt-get install -qq libalglib-dev libinsighttoolkit4-dev; apt-cache show libvtk6-dev && sudo apt-get install -qq libvtk6-dev; apt-cache show libvtk7-dev && sudo apt-get install -qq libvtk7-dev; cd $HOME/build/neuropoly/spinalcordtoolbox/dev/isct_propseg; mkdir -p build; cd build; cmake ..; make; echo"
-  - sct_testing -d 1
+  - pip install coverage
+  - bash -c 'echo -ne "import coverage\ncov = coverage.process_startup()\n" > sitecustomize.py'
+  - bash -c 'echo -ne "[run]\nconcurrency = multiprocessing\nparallel = True\n" > .coveragerc'
+  - COVERAGE_PROCESS_START="$PWD/.coveragerc" COVERAGE_FILE="$PWD/.coverage" PYTHONPATH="$PWD" sct_testing -d 1
+  - coverage combine
+  - ./unit_testing.sh
```
perone: I wonder if we shouldn't separate these two different tests. Because otherwise, we'll never know what lacks unit-testing or integration tests. However, I'm not sure if coveralls accepts two kinds of distinct testing report.

scripts/sct_download_data.py
```diff
@@ -214,4 +215,6 @@ def download_data(urls, verbose):
 
 if __name__ == "__main__":
     sct.init_sct()
-    main()
+    res = main()
+    raise SystemExit(res)
```
perone: This will change the exception `KeyboardInterrupt` for a `SystemExit`, it might confuse error reporting, exception handling, etc. Please see the #1658. Let's say someone uses `sct_download_data` later (as an API), it will silence the `KeyboardInterrupt` exception in favor of a C-style error reporting of returning a code.
If there is any other way of doing that it would be nice.

scripts/sct_testing.py
```diff
@@ -57,37 +57,86 @@ class bcolors:
...
+    )
+
+    def arg_jobs(s):
+        jobs = int(s)
+        if jobs > 0:
+            pass
+        elif jobs == -1:
+            jobs = None
+        else:
+            raise ValueError()
```
perone: Missing error message in the `ValueError` exception.

scripts/sct_testing.py
```diff
@@ -57,37 +57,86 @@ class bcolors:
...
+def process_function(fname, param):
+    """
+    """
+    param.function_to_test = fname
+    # display script name
+    # load modules of function to test
+    module_testing = importlib.import_module('test_' + fname)
+    # initialize default parameters of function to test
+    param.args = []
+    # param.list_fname_gt = []
```
perone: Commented code.

scripts/sct_testing.py
```diff
@@ -57,37 +57,86 @@ class bcolors:
...
+    param = module_testing.init(param)
+    # loop over parameters to test
+    list_status_function = []
+    list_output = []
+    for i in range(0, len(param.args)):
+        param_test = copy.deepcopy(param)
+        param_test.default_args = param.args
+        param_test.args = param.args[i]
+        param_test.test_integrity = True
+        # if list_fname_gt is not empty, assign it
```
perone: Commented code.

scripts/sct_testing.py
```diff
@@ -57,37 +57,86 @@ class bcolors:
...
+        # if param_test.list_fname_gt:
+        #     param_test.fname_gt = param_test.list_fname_gt[i]
+        # test function
+        try:
+            param_test = test_function(param_test)
+        except Exception as e:
+            list_status_function.append(1)
+            list_output.append("TODO exception: %s" % e)
```
perone: TODO message.

scripts/sct_testing.py
```diff
@@ -134,45 +179,29 @@ def main(args=None):
...
+    if jobs != 1:
+        pool = multiprocessing.Pool(processes=jobs)
+
+        results = list()
+        # loop across functions and run tests
+        for f in list_functions:
+            res = pool.apply_async(process_function, (f, param,))
+            results.append(res)
+
+    for idx_function, f in enumerate(list_functions):
```
perone: My only concern here is if a message is printed from inside the testing, that will appear here out-of-order, causing confusion on which testing the error message came from because it is executing all of them in parallel.
One solution to that would be capturing and returning the `stdout` and `stderrr` inside the `process_function()`, so you can print these messages here later in order.

.travis.yml
zougloub: IMHO:
- We'll know that we lack coverage, and improving it can be achieved either by unit or integration tests...
- Coveralls is nice and sexy but not as flexible as the bare ``coverage``; we can also get coverage reports locally.

scripts/sct_download_data.py
zougloub: The ``main()`` function was not exactly behaving as an API-callable function (should have one named ``download_data()`` or similar) but rather like a ``main()`` function... 
As the user could ^C anytime, I just removed the try/catch and the user will get a standard traceback message, as in every other SCT script.

scripts/sct_testing.py
zougloub: It doesn't matter as the message doesn't get brought to the user's attention (argparse is eating it).

scripts/sct_testing.py
zougloub: I didn't add any commented code, it was just moved from another location.

scripts/sct_testing.py
zougloub: Was here before... but yeah, one day.

scripts/sct_testing.py
zougloub: Not sure what you mean, the messages are indeed collected and (in case of error) printed in order.

.travis.yml
perone: I mentioned having separate coverages because the integration tests are very different than the unit tests. In integration tests, we test with real data, while on unit testing we test with mocked data. But as you mentioned, we can have separate reports locally, so that's not an issue.

scripts/sct_testing.py
perone: As far as I understood, it is capturing only exceptions but not any other message such as prints, logging, etc. That's my concern.

jcohenadad:
for comparison, i tried with different jobs on my 8-cors OSX laptop:
- -j 4: 3 min 25 s

scripts/sct_testing.py
```diff
@@ -57,37 +57,86 @@ class bcolors:
...
+     default=param_default.path_data,
+    )
+    parser.add_argument("--remove-temps", "-r",
+     choices=("0", "1"),
+     help='Remove temporary files.',
+     default=param_default.remove_tmp_file,
+    )
+    parser.add_argument("--jobs", "-j",
+     type=arg_jobs,
+     help="# of simultaneous tests to run (jobs). -1 means # of cores",
```
jcohenadad: i would replace with `[...] -1 means # of available cores`

abtahizadeh:
- Ubuntu image (2G RAM) `-j -1` -> 6m 23s

scripts/sct_testing.py
zougloub: Fixed (plus, it's not -1 but 0 now... I can't remember in what state of mind i was when I decided to use -1 instead of zero).

zougloub:
^C should be handled now.
Note that ^Z + ``kill %1`` can be used in such cases.

abtahizadeh:
@zougloub 
- Tested, `total time = 5m 17s` (somehow it reduced!, don't know why)
- `KeyboardInterrupt` raised, but it does not terminate the process (requires ^Z) -> I'm sure it's because of multiprocessing....

jcohenadad:
i confirm keyboard interrupt works now

PR #1875, opened 2018-06-17
sct_extract_metric: Now possible to calculate metrics slice-wise or level-wise

### LINKED ISSUE #1213: Refactor to enable computation of metrics per slice
related to: 
- https://sourceforge.net/p/spinalcordtoolbox/discussion/help/thread/4587c642/
- https://github.com/neuropoly/spinalcordtoolbox/issues/1351

Features needed:
- normalize metric with sum of values across tracts for use with lesion volume calculation (i.e. percentage) --> flag for it
- compute metric per slice without having to loop from external function
- making it possible to call from other function (i.e., separate the opening of label files from subfunction)
- look for ways to make opening of labels faster

jcohenadad:
### Description of the Change

The new flags `-perslice` and `-perlevel` allow to compute metrics slice-wise or vertebral-level wise. In the case of `-perlevel`, a vertebral label file is required (typically located under `label/template/PAM50_levels.nii.gz`). 

Moreover, this PR brings some refactoring to `sct_extract_metric`, including the outsourcing of a few functions which are now located in the `spinalcordtoolbox.template` and `spinalcordtoolbox.utils` modules. Future refactoring of other functions could make use of these modules. The next in line could be `sct_process_segmentation`.

Fixes #1213


scripts/sct_utils.py
```diff
@@ -401,19 +401,19 @@ def my_emit(self, record):
...
     :param suffix: suffix. Example: _mean
     :return: file name with suffix. Example: t2_mean.nii
     """
-    # get index of extension. Here, we search from the end to avoid issue with folders that have ".nii" in their name.
-    ind_nii = fname.rfind('.nii')
-    # in case no extension was found (i.e. only prefix was specified by user)
-    if ind_nii == -1:
-        return fname[:len(fname)] + suffix
+    # check if extension is .nii.gz
+    if fname[-7:] == '.nii.gz':
```
zougloub: With ``.endswith(".nii.gz")`` you save yourself the magic 7.

scripts/sct_utils.py
```diff
@@ -401,19 +401,19 @@ def my_emit(self, record):
...
+        # get index of extension
+        ind_ext = fname.find('.nii.gz')
```
zougloub: Doesn't work for ``pouetpouet.nii.gz-pouet.nii.gz``, the previous ``.rfind()`` did.
Also, you could have done ``fname[:-7]``.
But maybe you could just piggy-back on ``sct.extract_fname()`` which makes it even cleaner.
Which made me realized that its code is messy.
I imagine that you have better things to do, I can clean that up later.

spinalcordtoolbox/template.py
```diff
@@ -0,0 +1,52 @@
...
+        indx, indy = np.where(data_vertlevel[:, :, iz])
+        # average non-null values and round to closest
+        try:
+            average_value = int(round(np.mean(data_vertlevel[indx, indy, iz])))
+            # if that matches the desired level, append it to slice list
+            if average_value == level:
+                slices.append(iz)
+        except ValueError:
+            # slice is empty (no indx found). Do nothing.
+            log.debug('Empty slice: z=%s'.format(iz))
```
zougloub: That won't fly (mixing ``%`` flag characters and ``str.format()``.
For logging, do ```log.debug("something %s", arg)``` which is better (since the string is constructed lazily, it gets built or not depending on the logging filters/levels).

spinalcordtoolbox/template.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    data_vertlevel = im_vertlevel.data
+    # average non-null values and round to closest
+    try:
+        # find indices of non-null values
+        indx, indy = np.where(data_vertlevel[:, :, slice])
+        vert_level = int(round(np.mean(data_vertlevel[indx, indy, slice])))
+    except ValueError:
+        # slice is empty (no indx found). Do nothing.
+        log.debug('Empty slice: z=%s'.format(slice))
```
zougloub: same here

spinalcordtoolbox/template.py
```diff
@@ -0,0 +1,52 @@
...
+            # if that matches the desired level, append it to slice list
+            if average_value == level:
+                slices.append(iz)
+        except ValueError:
+            # slice is empty (no indx found). Do nothing.
+            log.debug('Empty slice: z=%s'.format(iz))
+    return slices
+
+
+def get_vertebral_level_from_slice(im_vertlevel, slice):
```
zougloub: You can use ``idx_slice`` which is clearer and avoids shadowing a built-in.

testing/test_sct_segment_graymatter.py
```diff
@@ -69,34 +56,44 @@ def test_integrity(param_test):
...
-
-        if dc == '0' or dc == 'nan':
-            null_slices.append(n_slice)
-        else:
-            try:
-                gm_dice.append(float(dc))
-            except ValueError:
-                gm_dice.append(float(dc[:-4]))
+        try:
+            n_slice, dc = line.split(' ')
```
zougloub: You'd better avoid the try/catch and instead use a match/continue while migrating this code.

```
line = line.rstrip()
m = re.match(r"^(?P<term_color>\x1B[31;1m)?(?P<n_slice>[0-9]+) (?P<dc>[0-9.]+)(?P<term_clear>\x1B[0m)?$", line)
if m is None:
    continue
n_slice = int(m.group("n_slice"))
dc = float(m.group("dc"))
```
There, the code is much more straightforward to understand.
I can do it if you'd like.

scripts/sct_utils.py
jcohenadad: cool! didn't know the trick

testing/test_sct_segment_graymatter.py
jcohenadad: thanks a lot for those suggestions! agreed with all (incl. the "I can do it if you'd like" ;-) )

zougloub: I won't do that last one now as it's in the testing code and it takes more than a few minutes to figure out what it's expected to parse...

zougloub:
wow sorry @charleygros / @jcohenadad I was trigger-happy and did the merge.

Issue #1702, opened 2018-04-25
Testing branches


jcohenadad:
I have a git installation of SCT. I wanted to test a branch, did a checkout, fetch, rebase, and then i ran:
~~~
SCT_INSTALL_TYPE=in-place add_to_path=no change_default_path=yes ./install_sct -d -b -m
~~~

as per the instruction here: https://github.com/neuropoly/spinalcordtoolbox/wiki/development-hints#testing-environments

however, the spinalcordtoolbox/gui was not updated.

i had to do the following to update it:
~~~
pip install -e .
~~~

what is the correct procedure for testing a branch so that we can be sure 100% of the time that we are testing the right thing?  Once clarified we can update this link:
https://github.com/neuropoly/spinalcordtoolbox/wiki/development-hints#testing-environments

zougloub:
The installer always performs that `pip install` of SCT (with Conda's pip).
How are you launching the GUI? Any chance the pip you are using isn't the one from the Conda environment?

jcohenadad:
> The installer always performs that pip install of SCT (with Conda's pip).
yes, i'm also puzzled. The GUI was launched from the terminal (in that case i think it was `sct_label_utils -i XX -create-viewer`). I am pretty sure i was using SCT Conda's pip (after doing a `source python/bin/activate` within `$SCT_DIR`). Next time it happens to me i will document better (i.e. with exact commit, etc.).

jcohenadad:
Another fail in updating modules after running ./install_sct, resulting in wasted time: https://github.com/neuropoly/spinalcordtoolbox/pull/1746#issuecomment-390824936

jcohenadad:
OK, I figured out the culprit:
- when i do `pip install .` (default mode in `./install_sct`) --> it does NOT work, meaning, I cannot access the modified functions under `spinalcordtoolbox` modules
- when i do `pip install -e .` (dev mode in `./install_sct`) --> it does work.

looking at the pip help, it says that -e "Installs a project in editable mode (i.e. setuptools "develop mode")", but i don't get it. If someone can explain to me why a fresh SCT install (including conda install) makes `pip install .` works, while running `pip install .` without reinstalling conda is problematic.

zougloub:
What ``-e`` does is (mostly) writing to an `easy-install.pth` file (import path addition) the location of the spinalcordtoolbox parent folder.
[...]

PR #1892, opened 2018-06-28
install: use pip install -e spinalcordtoolbox to gain flexibility

zougloub:
This PR brings a "solution" for #1702, whereby users can tinker with the running code in all cases:

- Users installing sct as a package or in ~/sct_dev will not see anything change, but they may edit scripts as well as spinalcordtoolbox files in their installation.
- Users installing sct in-place (from git) will save some installation time (because of no copy needed) and can stop going crazy (previously ``pip install --upgrade`` needed to be ran).
- Users installing sct in development mode will not see anything change (the development mode actually becomes moot as it basically just enables to use the *source* spinalcordtoolbox folder even when performing package installs... it could be removed to save some installer space).

Bonus:

- hot fix for #1895

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -188,10 +188,6 @@ while getopts ":dmhbpz" opt; do
       echo " Multiprocessing using MPI activated"
       export SCT_MPI_MODE=yes
       ;;
-    z)
-      echo " Development mode"
-      export SCT_DEV_MODE=yes
-      ;;
     h)
       usage
       exit 0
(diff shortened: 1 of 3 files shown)


abtahizadeh:
Ran tests on this, the installation worked, check_dependencies seem to be fine.

install_sct
```diff
@@ -528,7 +520,7 @@ if [ ${SCT_DEV_MODE} ]; then
   pip install -e ${SCT_SOURCE}
   e_status=$?
 else
-  pip install ${SCT_SOURCE}
+  pip install -e ${SCT_DIR}
```
jcohenadad: do we still need to distinguish dev vs. non-dev mode? i.e., maybe we don't need the `-z` flag anymore?

zougloub: If nobody uses it, it could be removed. The only justification for its existence is that with it, we can perform out-of-place installs and edit files that are not in the installation folder (but I have never done that).

jcohenadad: I think we can remove it.

zougloub: OK done.

PR #1903, opened 2018-07-04
sct_propseg: Now possible to rescale data header to be able to segment non-human spinal cord (mice, rats, etc.)

### LINKED ISSUE #1406: Add flag to change voxel dimension before running PropSeg
### Description

In cases of cord compression, propseg often fails because it initialize using a cylindrical mesh. One solution is to stretch the image in the AP direction, run propseg and compress back the seg. Example here:
https://www.dropbox.com/s/okst58h8r9sx2do/20170713_propseg_stretched.m4v?dl=0

### LINKED ISSUE #1874: Add a flag to disable check_and_correct_segmentation()
### Description

In a few scans now, I've noticed that the function `check_and_correct_segmentation()` outputs a blank segmentation, while directly after running `isct_propseg`, the segmentation looks fine.

We should thus give our users the option to disable `check_and_correct_segmentation()`.

jcohenadad:
### Description of the Change

- Fixes #1406: New flag `-rescale`, which rescales the image (only the header, not the data) in order to enable segmentation on spinal cords with dimensions different than that of humans (e.g., mice, rats, elephants, etc.) 
- Fixes #1874: New flag `-correct-seg`, which enables/disables a few corrections on the output segmentation. This was introduced because sometimes the correction makes things worse.

### DIFF
--- a/scripts/sct_image.py
+++ b/scripts/sct_image.py
@@ -42,7 +42,7 @@ def get_parser():
                       mandatory=False,
                       example='data_pad.nii.gz')
 
-    parser.usage.addSection('\nBasic image operations:')
+    parser.usage.addSection('\nImage operations:')
     parser.add_option(name="-pad",
                       type_value="str",
                       description='Pad 3D image. Specify padding as: "x,y,z" (in voxel)',
--- a/scripts/sct_image.py
+++ b/scripts/sct_image.py
@@ -53,11 +53,6 @@ def get_parser():
                       description='Pad 3D image with asymmetric padding. Specify padding as: "x_i,x_f,y_i,y_f,z_i,z_f" (in voxel)',
                       mandatory=False,
                       example='0,0,5,10,1,1')
-    parser.add_option(name="-copy-header",
-                      type_value="file",
-                      description='Copy the header of the input image (specified in -i) to the destination image (specified here)',
-                      mandatory=False,
-                      example='data_dest.nii.gz')
     parser.add_option(name="-split",
                       type_value="multiple_choice",
                       description='Split data along the specified dimension. The suffix _DIM+NUMBER will be added to the intput file name.',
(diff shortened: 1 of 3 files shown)


scripts/sct_image.py
```diff
@@ -84,18 +79,25 @@ def get_parser():
...
     parser.add_option(name="-setorient",
                       type_value="multiple_choice",
-                      description='Set orientation of the input image',
+                      description='Set orientation of the input image (only modifies the header).',
                       mandatory=False,
                       example='RIP LIP RSP LSP RIA LIA RSA LSA IRP ILP SRP SLP IRA ILA SRA SLA RPI LPI RAI LAI RPS LPS RAS LAS PRI PLI ARI ALI PRS PLS ARS ALS IPR SPR IAR SAR IPL SPL IAL SAL PIR PSR AIR ASR PIL PSL AIL ASL'.split())
     parser.add_option(name="-setorient-data",
                       type_value="multiple_choice",
-                      description='Set orientation of the input image\'s data. Use with care !ro',
+                      description='Set orientation of the input image\'s data (does NOT modify the header, but the data). Use with care !ro',
```
zougloub: Sorry, what's the trailing "ro" in that string?

scripts/sct_propseg.py
```diff
@@ -509,6 +517,28 @@ def generate_qc(fn_in, fn_seg, args, path_qc):
...
+        # get qform
+        qform = img.header.get_qform()
+        # multiply by scaling factor
+        qform[0:3, 0:3] *= rescale_header
+        # generate a new nifti file
+        header_rescaled = img.header.copy()
+        header_rescaled.set_qform(qform)
+        # the data are the same-- only the header changes
+        img_rescaled = nib.nifti1.Nifti1Image(img.get_data(), None, header=header_rescaled)
+        path_tmp = sct.tmp_create(basename="propseg", verbose=verbose)
```
zougloub: Looks like the temp. folder is not removed at the end.

jcohenadad:
@zougloub not ready for review (WIP)

scripts/sct_propseg.py
jcohenadad: indeed-- i've added a todo (not urgent)

Issue #1548, opened 2017-11-30
Replace usage of custom command-line parser (`msct_parser`) with argparse


zougloub:
<details>
<summary>Original issue description hidden, click to show</summary>

### Title: Some griefs with the command-line arguments parser

SCT implements its own command-line parser (vs. using the built-in `argparse` or 3rd party alternatives) and it's not very practical.

- Is there a good reason for this? I see that the parser code can do things such as doing basic checks on parameters (eg. making sure an image is a NIFTI, making sure a folder exists) but this can be done with standard ones too.

- There's a good practice of supplying both long options (`--image-contrast`) and short options (`-c`), with documentation favoring the long ones. The parser doesn't do that.

- Inconsistencies in usage of parameters (eg. `-rtmp` vs. `-r`). Not related to the parser per se.

- Parser output doesn't adapt to terminal width.

- Shell completion can't be implemented easily. In the same way that Python users could be using IDE completion for function parameters, having shell completion could be helpful. It could be implemented in 4 lines of code using the optional `argcomplete` if using a standard parser.

I'd suggest to perform an evaluation of replacement of the parser (ensuring compatibility with current uses, or not), and to go ahead and do it.
</details>

-----

Replacement of SCT's custom `msct_parser` with Python's `argparse` is currently in-progress. The remaining uses of `msct_parser` that are exposed to users are:

 - `/scripts/isct_warpmovie_generator.py` (non-functional script, see #2828)
 - [x] `/scripts/sct_dmri_display_bvecs.py`
 - [x] `/scripts/sct_dmri_moco.py`
 - [x] `/scripts/sct_dmri_separate_b0_and_dwi.py`
 - [x] `/scripts/sct_dmri_transpose_bvecs.py`
 - [x] `/scripts/sct_download_data.py`
 - [x] `/scripts/sct_extract_metric.py`
 - [x] `/scripts/sct_flatten_sagittal.py`
 - [x] `/scripts/sct_fmri_compute_tsnr.py`
 - [x] `/scripts/sct_fmri_moco.py`
 - [x] `/scripts/sct_get_centerline.py`
 - [x] `/scripts/sct_label_utils.py`
 - [x] `/scripts/sct_label_vertebrae.py`
 - [x] `/scripts/sct_process_segmentation.py`
 - [x] `/scripts/sct_propseg.py`
 - [x] `/scripts/sct_register_multimodal.py`
 - [x] `/scripts/sct_register_to_template.py`
[...]

PR #1908, opened 2018-07-06
Lazy loading for the deepseg_gm module

perone:
This is a minor change to delay the load of the `deepseg_gm` module for only when required, it reduced the command usage help from 3.2s to 0.3s due to Keras/TensorFlow loading.

## Profile before
image

## Profile after
image


zougloub:
Note that i mentioned doing lazy loading of modules for script entry points in #1548 ( and even on this particular script in https://github.com/neuropoly/spinalcordtoolbox/issues/1548#issuecomment-376345734 ).

PR #1926, opened 2018-07-12
Labels and centerline are now output with correct header

### LINKED ISSUE #1925: if -rescale is used, the output centerline should be scaled back to the original data header

jcohenadad:
### Description of the Change

This PR fixes two issues that are associated with the recent introduction of the `-rescale` flag:
- The output centerline is now scaled back to the original data header
- The manual labels (if used) are now output and rescaled appropriately

Fixes #1925

### DIFF
--- a/scripts/sct_propseg.py
+++ b/scripts/sct_propseg.py
@@ -13,9 +13,7 @@
 
 # TODO: remove temp files in case rescaled is not "1"
 
-import os
-import sys
-
+import os, shutil, sys
 import numpy as np
 from scipy import ndimage as ndi
 import sct_image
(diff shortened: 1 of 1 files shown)


scripts/sct_propseg.py
```diff
@@ -581,18 +587,30 @@ def func_rescale_header(fname_data, rescale_factor):
...
                   fname_seg)
-        fname_centerline = os.path.join(folder_output, sct.add_suffix(os.path.basename(fname_data_propseg),
-                                                                      "_centerline"))
+        os.rename(os.path.join(folder_output, sct.add_suffix(os.path.basename(fname_data_propseg), "_centerline")),
+                  fname_centerline)
+        # if user was used, copy the labelled points to the output folder (they will then be scaled back)
+        if use_viewer:
+            fname_labels_viewer_new = os.path.join(folder_output, os.path.basename(sct.add_suffix(fname_data,
+                                                                                                  "_labels_viewer")))
+            shutil.copy(fname_labels_viewer, fname_labels_viewer_new)
```
zougloub: Actually we added a `sct.copy()`, you could use it.

zougloub: And we also have `sct.mv()` which can replace the `os.rename()` above.

PR #1939, opened 2018-07-17
Remove useless stuff

zougloub:
- PCA is used in the code, and not this one (I'd say for good reasons).
- Image has useless members


zougloub:
argh sorry. this is WIP

zougloub:
Keeping the PR short... I'll add more removals in a subsequent one.

PR #1943, opened 2018-07-18
First pass at msct_image/sct_image consolidations

zougloub:
Use image's `orientation` instead of calling isct_orientation3d, remove some dead code.
Next step will be to deal with set_orientation (#1944).


jcohenadad:
to be confident with the merge, i suggest we run `sct_pipeline` (and make sure it has integrity tests when appropriate, and if not, maybe consider adding tests before merging) on sct_testing/large database, on most of the high-level functions that could be affected by the refactoring. Namely (please update the list):
- [] `sct_analyze_lesion`
- [] `sct_analyze_texture`
- [] `sct_compute_hausdorff_distance`
- [] `sct_compute_snr`
- [] `sct_create_mask`
- [] `sct_detect_pmj`
- [] `sct_image -getorient`
- [] `sct_image -setorient`
- [] ``
- [] ``
- [] ``

zougloub:
I will use the review form to comment on each impacted function and explain how I am confident there is no impact.

scripts/sct_compute_snr.py
```diff
@@ -105,7 +105,7 @@ def main():
 
     # Check if data are in RPI
     input_im = Image(fname_data)
-    input_orient = get_orientation(input_im)
+    input_orient = input_im.orientation
```
zougloub: here, orientation has been set because loadFromPath has set it -> no problem

scripts/sct_create_mask.py
```diff
@@ -109,7 +108,7 @@ def create_mask(param):
     path_tmp = sct.tmp_create(basename="create_mask", verbose=param.verbose)
 
     sct.printv('\nCheck orientation...', param.verbose)
-    orientation_input = get_orientation(Image(param.fname_data))
+    orientation_input = Image(param.fname_data).orientation
```
zougloub: here, orientation has been set because loadFromPath has set it -> no problem

scripts/sct_create_mask.py
```diff
@@ -134,7 +133,7 @@ def create_mask(param):
         sct.run(['sct_image', '-i', 'point.nii.gz', '-o', 'point_RPI.nii.gz', '-setorient', 'RPI', '-v', '0'], verbose=False)
     #
     # if method_type == 'centerline':
-    #     orientation_centerline = get_orientation_3d(method_val, filename=True)
+    #     orientation_centerline = Image(method_val).orientation
```
zougloub: that's commented-out code, so obviously no impact, but if it were uncommented...
orientation has been set because loadFromPath has set it -> no problem

scripts/sct_detect_pmj.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, fname_im, contrast, fname_seg, path_out, verbose):
 
         self.tmp_dir = sct.tmp_create(verbose=self.verbose)  # path to tmp directory
 
-        self.orientation_im = get_orientation(Image(self.fname_im))  # to re-orient the data at the end
+        self.orientation_im = Image(self.fname_im).orientation  # to re-orient the data at the end
```
zougloub: here, orientation has been set because loadFromPath has set it -> no problem

scripts/sct_image.py
```diff
@@ -584,7 +584,7 @@ def orientation(im, ori=None, set=False, get=False, set_data=False, verbose=1, f
             try:
                 printv('Get orientation...', verbose)
                 im_out = None
-                ori = get_orientation(im)
+                ori = im.orientation
```
zougloub: Here the orientation is accessed vs. computed before.

Before, we would be evaluating the orientation from the header.

Cases I see:

1. Image was copied, in which case the original header wasn't touched -> same behaviour as before, as now orientation which is used, has also been copied over

2. Image was loaded, in which case orientation has been set by loadFromPath.

Is there another case? If not, then no problem.

scripts/sct_image.py
```diff
@@ -631,7 +631,7 @@ def orientation(im, ori=None, set=False, get=False, set_data=False, verbose=1, f
             # get orientation
             printv('Get orientation...', verbose)
             im_out = None
-            ori = get_orientation(im_split_list[0])
+            ori = im_split_list[0].orientation
```
zougloub: same as above, but multicomponent_split is the one copying images.

scripts/sct_process_segmentation.py
```diff
@@ -488,7 +488,7 @@ def extract_centerline(fname_segmentation, remove_temp_files, verbose = 0, algo_
     # nx, ny, nz, nt, px, py, pz, pt = im_seg.dim
     #
     # # Extract orientation of the input segmentation
-    # orientation = get_orientation(im_seg)
+    # orientation = im_seg.orientation
```
zougloub: commented code.

Issue #1938, opened 2018-07-17
extract_centerline() should output the real centerline coordinate, not the discrete version


jcohenadad:
### Description

Currently, the text output of extract_centerline() is simply the voxel coordinates of the output image centerline. There is a lack of precision when doing that. We should instead output the floating point coordinates.

zougloub:
TL;DR: I'm wondering what is the convention for real vs. integer coordinates in these images (**pixel-center convention**) and if it's worthwhile to document it.

I'm assuming a coordinate index `idx` corresponds to a real coordinate span of `[idx,idx+1)`, ie. "(z.0,y.0,x.0)" coordinate indicates the corner of a voxel, and a ("z.5", "y.5", "x.5) one indicates the center (**half-integer pixel-center convention**), so that (0,0,0) and (NZ, NY, NX) are corners of a volume (so `int_to_float(x) = x+0.5` and `float_to_int(x) = int(round(x-.5)) = int(x)`).

The alternative I would be less comfortable with would be to have voxel centers on integer coordinates, where the volume boundaries are not as "meaningful".

And I was wondering if this is obvious / common knowledge / or documented somewhere (eg. the convention is documented [in opengl](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_fragment_coord_conventions.txt)).

zougloub:
Note: [looks like](http://nipy.org/nibabel/coordinate_systems.html#voxel-coordinates-are-in-voxel-space) nibabel's physical <-> pixel position transform is using the integer pixel center convention.

zougloub:
The used convention (integer pixel center) is documented, so closing this parenthesis.

zougloub:
Yep.

PR #1948, opened 2018-07-19
Second pass at image refactoring

### LINKED ISSUE #1944: Figuring out orientation so `set_orientation()` can be deprecated in favor of an updated `change_orientation()`.
Context: trying to get rid of `isct_*` binaries.

Throughout the SCT code we're changing image orientations, in 2 different ways:

- change_orientation() which alters im.data but doesn't touch the NIFTI header
- set_orientation() which says it only alters the NIFTI header (but somehow when I ran it and performed a diff, I saw that the extent of modification wasn't just on the header...)

I understand that the idea of the first one is to quickly change the data orientation without going through nibabel's physical to pixel transforms, so that can be done:
- iterating through image slices
- operating in a frame where an algorithm was designed to work in
It's not updating an image header, so if we were to save a resulting file, consistency between the voxel indices and the corresponding physical positions is lost.

The second one is calling an external binary which we'd want to get rid of.

I would like to update change_orientation() so that it also updates the header, so we can get rid of the second thing (+ fix #605, #842), and the files could be saved without fear.

First, as the code is only considering the [`sform`](http://nipy.org/nibabel/reference/nibabel.nifti1.html#nibabel.nifti1.Nifti1Pair.get_qform) transform, I'm thinking of updating the affine and doing `hdr.set_sform()`.
Am I missing something?

Secondly, is it safe to only consider the `sform` ? I don't have much real-life usage history of NIFTI .
nibabel uses [an algorithm](http://nipy.org/nibabel/nifti_images.html#choosing-image-affine) to figure out the affine depending on the available header fields.

zougloub:
A set of change which:
- Makes the image API cleaner and leaner, as the SLOC in the calling code is reduced.
- Brings some consistency and docs (fixes #1944)
- Has some collateral changes (mostly code prettification).

Done:
- Trim msct_image and sct_image
 - make use of properties to reduce amount of members (eg. `.dim` is computed)
- Add generic free_functions to `msct_image`:
  - `change_{type,orientation}()`, which can operate in-place or make copies
  - `to_dtype()`, which convets type names or dtype to `np.dtype`
  - `change_orientation()`, does what it's supposed to do
  - `zeros_like()` which is often used for creating label images or other images with same shape/header but zero data
  - `spatial_crop()` which crops images and replaces some `sct_crop_image` internal calls
- Improve up msct_image's Image API
 - `.change_{type,orientation}()` can generate new file names but by default they just void the original one (to prevent overwriting of original data)
 - `.save()` can modify or not the image contents
 - `.save()` can save with the current basename to a new directory (typically used in the scenario of copying to tmp.dir, then changing orientation, then saving)
 - members that modify an image return `self` so calls can be chained
- add `msct_image` unit tests
- Documentation is added, aiming to make the concepts limpid.
   - #1944 (undocumented reference spaces)
   - #1938 (introduction of floating-point voxel coordinates)
- Functions that write temporary data are the ones responsible for setting up a temporary directory, not their callers


zougloub:
@jcohenadad 's points on https://github.com/neuropoly/spinalcordtoolbox/pull/1943 are valid for this:
> to be confident with the merge, i suggest we run `sct_pipeline` (and make sure it has integrity tests when appropriate, and if not, maybe consider adding tests before merging) on sct_testing/large database, on most of the high-level functions that could be affected by the refactoring. Namely (please verify the list below):
> Functions that are explicitly modified in this PR:
> - [ ] `sct_analyze_lesion`
> - [ ] `sct_analyze_texture`
> - [ ] `sct_compute_hausdorff_distance`
> - [ ] `sct_compute_snr`
> - [ ] `sct_create_mask`
> - [ ] `sct_detect_pmj`
> - [ ] `sct_image -getorient`
> - [ ] `sct_image -setorient`
> - [ ] `sct_process_segmentation -p extract_centerline`
> 
> Functions that do re-orientation and that could be affected:
> - [ ] `sct_propseg`
> - [ ] `sct_deepseg_sc`
> - [ ] `sct_extract_metric`
> - [ ] `sct_register_to_template`
> 
> Now I realize there is a lot to test, and most of these functions don't have proper ground truth on sct_testing. So we need to find a good solution.

scripts/msct_gmseg_utils.py
```diff
@@ -446,17 +440,13 @@ def register_data(im_src, im_dest, param_reg, path_copy_warp=None, rm_tmp=True):
     os.chdir(tmp_dir)
     # save image and seg
     fname_src = 'src.nii.gz'
-    im_src.setFileName(fname_src)
-    im_src.save()
+    im_src.save(fname_src)
```
jcohenadad: i really like this feature. i've been dreaming about it for months

scripts/msct_shape.py
```diff
@@ -23,10 +23,10 @@
```
jcohenadad: maybe don't spend too much time cleaning this function because i've slaughtered it quite a lot already: https://github.com/neuropoly/spinalcordtoolbox/blob/jca/1920-process_seg-csa-perslice/scripts/msct_shape.py

scripts/msct_types.py
```diff
@@ -639,8 +639,7 @@ def save_centerline(self, image=None, fname_output='centerline.sct'):
                     coord_pix = image.transfo_phys2pix([current_coord])[0]
                     image_output.data[int(coord_pix[0]), int(coord_pix[1]), int(coord_pix[2])] = float(self.labels_regions[current_label]) + current_dist_rel
 
-            image_output.setFileName(fname_output)
-            image_output.save(type='float32')
+            image_output.save(fname_output, dtype='float32')
```
jcohenadad: also a very cool feature

scripts/sct_image.py
```diff
@@ -557,198 +543,56 @@ def multicomponent_merge(fname_list):
     im_out = im_0.copy()
     im_out.data = data_out.astype('float32')
     im_out.hdr.set_intent('vector', (), '')
-    im_out.setFileName(im_out.file_name + '_multicomponent' + im_out.ext)
+    im_out.absolutepath = sct.add_suffix(im_out.absolutepath, '_multicomponent')
     return im_out
 
 
-def orientation(im, ori=None, set=False, get=False, set_data=False, verbose=1, fname_out=None):
+def change_orientation_nd(src, orientation, dst=None, inverse=False):
```
jcohenadad: i would suggest to use [our conventions](https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#variables) for input name, to explicit if this is a file name, an Image object or a nibabel object. So in that case i would replace `src` with `im_src` for example.

scripts/sct_image.py
```diff
@@ -557,198 +543,56 @@ def multicomponent_merge(fname_list):
...
-            curdir = os.getcwd()
-            os.chdir(tmp_folder)
-            im_out = set_orientation(im, ori, True)
-            os.chdir(curdir)
-            sct.rmtree(tmp_folder)
-        else:
-            im_out = None
+    nx, ny, nz, nt, px, py, pz, pt = src.dim
+
+    if dst is None:
```
jcohenadad: same comment, suggestion: `im_dst`

scripts/sct_convert.py
```diff
@@ -54,21 +57,19 @@ def get_parser():
...
-    # Save file
-    im.setFileName(fname_out)
-    if type is not None:
-        im.changeType(type=type)
-    im.save(squeeze_data=squeeze_data)
+    import msct_image
+    sct.printv('sct_convert -i ' + fname_in + ' -o ' + fname_out, verbose, 'code')
+    im = msct_image.Image(fname_in)
+    if squeeze_data:
+        im.data = np.squeeze(im.data)
```
jcohenadad: i'm still puzzled as to why in this function `squeeze` is on by default. We probably had some issues in the past but i don't remember what. Maybe looking at old closed issues.

scripts/sct_create_mask.py
```diff
@@ -104,37 +103,29 @@ def create_mask(param):
...
+    orientation_input = Image(param.fname_data).orientation
     sct.printv('.. ' + orientation_input, param.verbose)
     reorient_coordinates = False
 
     # copy input data to tmp folder
-    convert(param.fname_data, os.path.join(path_tmp, "data.nii"))
+    Image(param.fname_data).change_orientation("RPI").save(os.path.join(path_tmp, "data_RPI.nii"))
     if method_type == 'centerline':
-        convert(method_val, os.path.join(path_tmp, "centerline.nii.gz"))
+        Image(method_val).change_orientation("RPI").save(os.path.join(path_tmp, "centerline_RPI.nii"))
```
jcohenadad: any reason to choose nii over nii.gz? i usually prefer nii.gz for sparse data (centerline, seg) because it compresses amazingly well and for extremely large matrix size it speeds up i/o quite a bit

scripts/sct_create_mask.py
```diff
@@ -104,37 +103,29 @@ def create_mask(param):
...
     if method_type == 'point':
-        convert(method_val, os.path.join(path_tmp, "point.nii.gz"))
+        Image(method_val).change_orientation("RPI").save(os.path.join(path_tmp, "point_RPI.nii"))
```
jcohenadad: same comment about nii

scripts/sct_create_mask.py
```diff
@@ -174,7 +165,7 @@ def create_mask(param):
 
     if method_type == 'centerline':
         # get name of centerline from user argument
-        fname_centerline = 'centerline_RPI.nii.gz'
+        fname_centerline = 'centerline_RPI.nii'
```
jcohenadad: same comment about nii

scripts/sct_detect_pmj.py
```diff
@@ -217,11 +211,10 @@ def extract_sagital_slice(self):
     def orient2pir(self):
         """Orient input data to PIR orientation."""
```
jcohenadad: not sure why this "PIR" suddenly comes in, but it has nothing to do with this PR anyways...

scripts/msct_image.py
```diff
@@ -1048,6 +1009,152 @@ def change_data_orientation(data, old_orientation='RPI', orientation="RPI"):
...
+        return np.dtype(dtype)
+
+    raise TypeError("data type {}: {} not understood".format(dtype.__class__, dtype))
+
+def zeros_like(img, dtype=None):
+    dst = change_type(img, dtype)
+    dst.data[:] = 0
+    return dst
+
+def empty_like(img, dtype=None):
```
jcohenadad: not sure i understand the origin of the name `empty_like`. I find it might get confused with `zeros_like`. It just does a copy of an Image object, right? so why not calling it `copy()`? Or maybe i am missing something

scripts/sct_process_segmentation.py
```diff
@@ -22,8 +22,8 @@
 import pandas as pd
```
jcohenadad: i'm also doing quite a lot of modifications on this function currently, so many of these modifications might become obsolete.

scripts/sct_image.py
zougloub: wilco

scripts/sct_convert.py
zougloub: We don't remember because no reference to issues or useful comment was left in the code... anyway, we'll see what happens when running tests.

scripts/sct_create_mask.py
zougloub: Only for temporary objects.

Here's the rationale:

Today a typical computer
- performs gzip compression at 100 MB/s
- writes to mass storage at > 300 MB/s (ssd) or > 150 MB/s (hdd)
- reads from mass storage at > 500 MB/s (ssd) or > 150 MB/s (hdd)

So using .nii makes the most sense from a speed perspective.

Also, arrays from nii are memory-mapped (whereas those from nii.gz are obviously not).
```python
>>> img1 = nibabel.load("pouet.nii.gz")
>>> type(img1.get_data())
<class 'numpy.ndarray'>
>>> img0 = nibabel.load("pouet.nii")
>>> type(img0.get_data())
<class 'numpy.core.memmap.memmap'>
```

scripts/msct_image.py
zougloub: The implementation of `empty_like()` is indeed performing a copy from now, but empty_like is a standard function of numpy which just allocates but doesn't initialize storage. Using this over zeros_like() or copy() makes the intent clearer.

scripts/sct_create_mask.py
jcohenadad: cool, i'm convinced. Only for temporary objects though. Storage is still $$$

scripts/msct_image.py
jcohenadad: i see. so, what would be example cases where one would prefer to use empty_like() vs. copy()?

zougloub: I used it for things like:
```python
scripts/sct_compute_hausdorff_distance.py:            self.thinned_image = msct_image.empty_like(self.image)
scripts/sct_compute_hausdorff_distance.py-            self.thinned_image.data = self.zhang_suen(self.image.data)
scripts/sct_compute_hausdorff_distance.py-            self.thinned_image.absolutepath = sct.add_suffix(self.image.absolutepath, "_thinned")
```

zougloub:
As I'm reviewing a lot of code and I'm currently facing issues:
1. `sct_apply_transfo` can't work (no displacement observed) in images with certain orientations, which is strange:  
Orientations OK: RAI RAS RPI RPS LAI LAS LPI LPS RIA RIP LIA LIP PLI PLS PIR PIL ILP SLP IAL IPL  
Orientations NG: RSA RSP LSA LSP ARI ARS ALI ALS PRI PRS AIR AIL ASR ASL PSR PSL IRA IRP ILA SRA SRP SLA IAR IPR SAR SAL SPR SPL  
2. Some integration tests not OK for now, I'm working on it: sct_register_to_template sct_segment_graymatter sct_straighten_spinalcord sct_label_vertebrae

zougloub:
I was comparing testing results of the new `change_orientation()` vs. the old one based on `isct_orientation3d` and see that while new code passes the unit tests, replacing the code to make use of the old method makes the tests fail.
@jcohenadad you can see for yourself and run `pytest unit_testing/test_image.py -k test_change_orientation` with `#msct_image.change_orientation = old_change_orientation`  commented or not.
Because of the unit tests checking various properties of the reoriented image:
- The physical position of points before/after hasn't changed
- The new orientation is as desired
I am convinced that there is something that's not working right with `isct_orientation3d`; I don't think that it's worth it to troubleshoot what's not working in there.

jcohenadad:
in response to https://github.com/neuropoly/spinalcordtoolbox/pull/1948#issuecomment-408215849, i agree, let's not debug `isct_orientation3d`. What we could do instead is run `sct_pipeline` on our large database, using a function that relies on `change_orientation`, and make sure that results are the same between this branch and master

zougloub:
Apart from the weird stuff above, some progress...```
status: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 99, 99, 0, 0, 0]
Failures: sct_straighten_spinalcord sct_register_to_template
```

jcohenadad:
tested 38cd048f5f86db70ad12e67ebd5e5156f1ea1bf2 on `sct_testing_data/t2`. Below are the center of FOV (in physical coordinate) for each situation. It should be the same for all situations:

`t2`
~~~
-6.719334
2.286655
7.67828
~~~

`sct_image -i t2.nii.gz -setorient RPI (version: master)`
~~~
-5.219334
6.286655
10.17828
~~~
**NOT THE SAME --> FAIL 👎 **

`sct_image -i t2.nii.gz -setorient RPI (version: 38cd048f5f86db70ad12e67ebd5e5156f1ea1bf2)`
~~~
-6.719334
2.286655
7.67828
~~~
**THE SAME --> PASS 👍  **

`fslswapdim t2.nii.gz RL PA IS (sanity check: FSL)`
~~~
-6.719334
2.286655
7.67828
~~~
**THE SAME --> EXPECTED**

At least with this image, this PR seems to fix [this issue](https://github.com/neuropoly/spinalcordtoolbox/pull/1978#issuecomment-410755873), which is pretty awesome. I'll do more testing on this PR.

zougloub:
I've left the function around for the next refactoring pass.

jcohenadad:
bug on `batch_processing` on b8d12e82de8c05127ba29e2cc7a2744c1c104c2b
~~~
--
Spinal Cord Toolbox (cJ/more-image/b8d12e82de8c05127ba29e2cc7a2744c1c104c2b)
Running /home/jcohen/code/sct/scripts/sct_compute_mtsat.py -mt mt1_crop.nii.gz -pd mt0_reg.nii.gz -t1 t1w_reg.nii.gz -trmt 30 -trpd 30 -trt1 15 -famt 9 -fapd 9 -fat1 15
Load data...
Loaded mt1_crop.nii.gz (/home/jcohen/sct_example_data/mt/mt1_crop.nii.gz) orientation LPI shape (61, 57, 22)
Loaded mt0_reg.nii.gz (/home/jcohen/sct_example_data/mt/mt0_reg.nii.gz) orientation LPI shape (61, 57, 22)
Loaded t1w_reg.nii.gz (/home/jcohen/sct_example_data/mt/t1w_reg.nii.gz) orientation LPI shape (61, 57, 22)
Compute T1 map...
Compute A...
Compute MTsat...
Generate output files...
  File "/home/jcohen/code/sct/scripts/sct_compute_mtsat.py", line 98, in <module>
    run_main(arguments)
  File "/home/jcohen/code/sct/scripts/sct_compute_mtsat.py", line 86, in run_main
    fname_b1map=args.b1map, fname_mtsat=args.omtsat, fname_t1map=args.ot1map, verbose=1)
  File "/home/jcohen/code/sct/spinalcordtoolbox/mtsat/mtsat.py", line 133, in compute_mtsat_from_file
    fname_mtsat = os.path.join(nii_mt.path, "mtsat.nii.gz")
AttributeError: 'Image' object has no attribute 'path'
~~~

jcohenadad:
output of batch_processing on 374e8d0bcb7b772ccce47bc34d0e11564d252a55 (`cJ/more-image`)
~~~
t2/CSA:   78.703221, 1.760750
mt/MTR:   2018/08/08 - 15:52:32,/home/jcohen/sct_example_data/mt/mtr.nii.gz,map,"2,3,4,5","15,16,17,11,12,13,14,8,9,10,4,5,6,7",51,white matter,980.883843843,54.1147920284,0
t2s/CSA_GM:   13.240696, 3.370950
t2s/CSA_WM:   65.202414, 3.112070
dmri/FA:  2018/08/08 - 15:57:04,/home/jcohen/sct_example_data/dmri/dti_FA.nii.gz,wa,Unknown,"2,3,4,5,6,7,8,9,10,11,12,13,14",4,WM left lateral corticospinal tract,80.5077761052,0.758846851538,0.167810495669
dmri/FA:  2018/08/08 - 15:57:04,/home/jcohen/sct_example_data/dmri/dti_FA.nii.gz,wa,Unknown,"2,3,4,5,6,7,8,9,10,11,12,13,14",5,WM right lateral corticospinal tract,81.3368825917,0.785060184288,0.143668976472
~~~

v3.2.3:
~~~
t2/CSA:   78.182308, 1.762403
mt/MTR:   51, white matter, 1547.984915, 55.426165, 0.000000
t2s/CSA_GM:   13.282941, 3.440752
t2s/CSA_WM:   65.160860, 3.175766
dmri/FA:  4, WM left lateral corticospinal tract, 81.115248, 0.755613, 0.174256
dmri/FA:  5, WM right lateral corticospinal tract, 81.458362, 0.783391, 0.145285
~~~

zougloub:
Orientation changes and physical coordinate computation is done differently (correctly?) now and that could explain slightly different results. I don't see how they would be less accurate than before, so if numbers change, I'd trust the new ones. The locus of these changes is the phys2pix, pix2phys and change_orientation functions, whose diff you could review (and I bet that you'll favor the new versions).
I have also observed, in conditions which make the old isct_orientation binary and the new code produce same results, that the results were slightly different, and the numerical accuracy of the new code was better (I haven't kept data to back this statement).

jcohenadad:
yeah, i also suspect that results could be more trustable now. I'd still be curious to compare results of `test_sct_register_to_template`. Currently running it-- comparative results should be out in ~2h.

jcohenadad:
added tags/milestone because since the produced results are slightly different, we need to justify where those changes come from (hence listing this PR in the changelog).

jcohenadad:
results of `sct_register_to_template -i t2/t2.nii.gz -s t2/t2_seg_manual.nii.gz -l t2/labels.nii.gz -igt /home/jcohen/code/sct/data/PAM50/template/PAM50_cord.nii.gz` using `sct_testing/large`

**master-4208eff65079fe7ee3626e0099c5fc54b1c01abe**:
~~~
Duration: 332s
Passed: 3/3
Crashed: 0/3
Mean: {'duration': 307.18860363960266, 'dice_template2anat': 0.9381992266982407, 'dice_anat2template': 0.9235895443994163}
STD: {'duration': 9.257684490051743, 'dice_template2anat': 0.017398009712787457, 'dice_anat2template': 0.005100014147897889}
~~~

**cJ/more-image-9b6cdfc0d2967905411c9f1e05987fa4e8fb7aa9**:
~~~
Duration: 1808s
Passed: 152/432
Crashed: 225/432
Mean: {'dice_anat2template': 0.910986766710356, 'dice_template2anat': 0.9215147868289546, 'duration': 364.76865527751266}
STD: {'duration': 389.7401216208057, 'dice_template2anat': 0.02895368689489166, 'dice_anat2template': 0.022843669615022065}
~~~

PR #1956, opened 2018-07-21
Totally pimp the Image Slicer (to act like a sequence, to slice many images), and add unit tests for the slicer

zougloub:
This new code allows to write:
```python
[ im_2d for im_2d in Slicer(im_3d) ]
```
as well as:
```python
[ dice(x, y) for (x,y) in Slicer((im3d_x, im3d_y)) ]
```

illustrating a suggestion in https://github.com/neuropoly/spinalcordtoolbox/issues/1621#issuecomment-406765731

### DIFF
--- a/scripts/msct_image.py
+++ b/scripts/msct_image.py
@@ -198,15 +198,21 @@ def decompose_affine_transform(A44):
 
 class Slicer(object):
     """
-    Image slicer utility class, that can help getting ranges and slice indices
+    Image slicer utility class.
+
+    Can help getting ranges and slice indices.
+    Can provide slices (being an *iterator*).
     """
     def __init__(self, im, axis="IS"):
+        opposite_character = {'L': 'R', 'R': 'L', 'A': 'P', 'P': 'A', 'I': 'S', 'S': 'I'}
         axis_labels = "LRPAIS"
         if len(axis) != 2:
             raise ValueError()
         if axis[0] not in axis_labels:
             raise ValueError()
         if axis[1] not in axis_labels:
+            raise ValueError()
+        if axis[0] != opposite_character[axis[1]]:
             raise ValueError()
 
         for idx_axis in range(2):
@@ -216,6 +222,9 @@ class Slicer(object):
         if dim_nr == -1:
             raise ValueError()
 
+        # SCT convention
+        from_dir = im.orientation[dim_nr]
+        self.direction = +1 if axis[0] == from_dir else -1
         self.nb_slices = im.dim[dim_nr]
         self.im = im
         self.axis = axis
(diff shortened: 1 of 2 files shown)


scripts/msct_image.py
```diff
@@ -237,13 +264,79 @@ def range(self, axis):
...
+
+        Example: [slice for slice in Slicer(im)()]
+        """
+        for idx_slice in self.range(self.axis):
+            yield self.im[self.slice(idx_slice)]
+
+
+class SlicerMany(object):
+    """
+    Images slicer utility class.
```
jcohenadad: this is the same header as for `SlicerSingle`. Would it be possible to explicit a bit more what is the specificity of this function vs. `SlicerSingle` (maybe add an example under each header?).

scripts/msct_image.py
```diff
@@ -198,16 +198,40 @@ def decompose_affine_transform(A44):
 
 class Slicer(object):
     """
-    Image slicer utility class, that can help getting ranges and slice indices
+    Image(s) slicer utility class.
+
+    Can help getting ranges and slice indices.
```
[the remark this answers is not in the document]
zougloub: Yeah that was the mind-blowing part.

But note that the orientation of the remaining slice dimensions is not managed yet.
This will be done when the `Image` refactorings are in.

scripts/msct_image.py
zougloub: There is an 's' at Images, but you're right. I'll add more docs later. For now this was just added for the cool factor. There's a lot of pre-conditions to document (caveats) for using that `SlicerMany` object.

scripts/msct_image.py
```diff
@@ -216,6 +240,9 @@ def __init__(self, im, axis="IS"):
         if dim_nr == -1:
             raise ValueError()
 
+        # SCT convention
```
jcohenadad: (sorry i can't add comment on a line that was not modified).
comment for 253: i would replace "a multi-dimensional slice" by "a 2D slice", because this isn't it what it does? otherwise i would not call it a "slice" (which by definition is 2D).

zougloub: A **multi-dimensional slice** is the conventional term for objects that are an N-tuple or `None` or `slice` objects. 
When you say `arr[1,:,2]` this is multi-dimensional slicing of `arr`.

zougloub:
Except that a range is not a list in Python 3, so *range* is the proper term ...

zougloub:
... but still I can rephrase later, saying that it provides “indices for all the slices along the desired axis”.

zougloub:
(rephrased)

PR #1978, opened 2018-08-02
sct_register_to_template: Fixed wrong projection in case labels not in same space

### LINKED ISSUE #1977: sct_register_to_template: Wrong projection of labels onto centerline is labels are not in the same space
### Description

In `sct_register_to_template.project_labels_on_spinalcord()`, the labels are re-oriented to RPI, which can cause bugs in case labels (i) are not in the same voxel space as the input segmentation and (ii) are not RPI. In fact, labels should *not* be re-oriented, because their coordinates are transformed into physical coordinates.

A proposed fix is to remove the re-orientation of labels before projecting them onto the cord centerline.

jcohenadad:
### Description
In `sct_register_to_template.project_labels_on_spinalcord()`, the labels are re-oriented to RPI, which can cause bugs in case labels (i) are not in the same voxel space as the input segmentation and (ii) are not RPI. In fact, labels should not be re-oriented, because their coordinates are transformed into physical coordinates.

The implemented fix consisted in removing the re-orientation of labels before projecting them onto the cord centerline.

Fixes #1977


jcohenadad:
currently breaking because of `project_labels_on_spinalcord()`. Working on it...

PR #1992, opened 2018-08-12
Misc: quicker Travis, random code tidy-ups and fixups

zougloub:
This PR brings:
- Linting code for py3k issues (and possibly more, but it careful review needs to be done because there are a ton of false positives), instead of running a py3k Travis build; for now as anyway the Travis build would fail, this helps preventing regressions in code, until we can really run the Travis build.
- Tiny code fixes or easy changes that greatly reduce the number of warnings spewed by linters
- Tiny code fixes that fix actual issues that CI or other testing hasn't caught yet
- Some dead code elimination


spinalcordtoolbox/gui/widgets.py
```diff
@@ -63,7 +65,7 @@ def refresh(self, labels=None):
             self._checked = labels
             self._unchecked = set(self._check_boxes.keys()) - set(labels)
 
-        for checkbox in self._check_boxes.itervalues():
+        for checkbox in self._check_boxes.values():
```
jcohenadad: produces the same result? what's the difference with itervalues?

zougloub: In python3, `dict.values()` is always returning an iterator, whereas in Python2 `dict.values()` is creating a list. There's no practical difference in this case.

PR #2012, opened 2018-08-27
sct_deepseg_sc, sct_deepseg_lesion: Allow to input image with one axial slice

charleygros:
**Goal:** Allow to input image with only one axial slice.

**Issue:** `isct_spine_detect` requires ```nz>1```.

**Solution:** If ```nz==1```, then ask the user to use `-centerline cnn` instead of `-centerline svm` (default).

To test:
```
cd sct_example_data/t2s
sct_crop_image -i t2s.nii.gz -dim 2 -start 2 -end 2 -o t2s_Z010.nii.gz
sct_deepseg_sc -i t2s_Z010.nii.gz -c t2s
```
Raises an error message:
```
Your image contains only one axial slice, please re-run the function using -centerline cnn.
```
Then,
```
sct_deepseg_sc -i t2s_Z010.nii.gz -c t2s -centerline cnn
```

Same for `sct_deepseg_lesion`.

Note that I have remove some duplicated code: The centerline detection is now done by a function in `sct_deepseg_sc` named `find_centerline`, which is called by `sct_deepseg_lesion`.

### DIFF
--- a/scripts/sct_deepseg_sc.py
+++ b/scripts/sct_deepseg_sc.py
@@ -393,22 +393,15 @@ def heatmap2optic(fname_heatmap, lambda_value, fname_out, z_max, algo='dpdt'):
     import nibabel as nib
     os.environ["FSLOUTPUTTYPE"] = "NIFTI_PAIR"
 
-    if Image(fname_heatmap).dim[2] > 1:  # isct_spine_detect requires nz>1
-        optic_input = fname_heatmap.split('.nii')[0]
+    optic_input = fname_heatmap.split('.nii')[0]
 
-        cmd_optic = 'isct_spine_detect -ctype="%s" -lambda="%s" "%s" "%s" "%s"' % \
-                    (algo, str(lambda_value), "NONE", optic_input, optic_input)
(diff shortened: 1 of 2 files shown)


scripts/sct_deepseg_sc.py
```diff
@@ -410,6 +416,68 @@ def heatmap2optic(fname_heatmap, lambda_value, fname_out, z_max, algo='dpdt'):
         ctr_nii.save()
 
 
+def find_centerline(algo, image_fname, path_sct, contrast_type, brain_bool, folder_output, remove_temp_files):
+
+    if algo == 'svm' and Image(image_fname).dim[2] > 1:  # isct_spine_detect requires nz > 1
```
jcohenadad: why not checking beforehand if the image is 2D, and if it is, duplicate the slice to create a 3D volume, then run the centerline detection, then crop back to 2D on the output centerline.
This will prevent from having an error with the default parameters if 2D image is input.

charleygros: Excellent idea, thank you.
It is ready to be reviewed (commit #5d2f492).

PR #2018, opened 2018-09-05
(WIP) sct_image: Does not reshape on split

### LINKED ISSUE #2017: When splitting along Z a 4D file, each output becomes 3D
### Description

When splitting along Z (or X or Y) a 4D file (with time: dim=4), each output (e.g. `data_Z0002.nii.gz`) is a 3D file, with the time dimension becoming dim=3 instead of dim=4, and the Z dimension being removed instead of keeping it as a singleton.

### Steps to Reproduce

~~~
cd sct_example_data/dmri
sct_image -i dmri.nii.gz -split z
~~~

jcohenadad:
When splitting along Z (or X or Y) a 4D file (with time: dim=4), each output (e.g. data_Z0002.nii.gz) is a 3D file, with the time dimension becoming dim=3 instead of dim=4, and the Z dimension being removed instead of keeping it as a singleton. This PR fixes is.

Fixes #2017

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ scripts/.idea/modules.xml
 scripts/.idea/scripts.iml
 scripts/.idea/vcs.xml
 scripts/.idea/workspace.xml
+
+.pytest_cache/
+unit_testing/*.csv
+unit_testing/*.nii.gz
+unit_testing/*.roi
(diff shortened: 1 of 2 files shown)


jcohenadad:
i'll close this PR and open another one (which also solves other problems, but has a better fix for the issue specific to this PR)

jcohenadad:
new working branch is `jca/1984-moco-sagittal`

Issue #1985, opened 2018-08-06
Add init-centerline option into sct_deepseg_sc


valosekj:
Dear SCT experts,
_sct_deepseg_sc_ works generally  better than _sct_propseg_ on our data (mainly in patients with cervical myelopathy). But in some cases when patient has greater stenosis and contrast is lower in these levels segmentation fails. I found out that reason of it is bad centerline detection by _Optic_ alg. (I also tried to change -centerline detection method from svm to cnn, but result of segmentation  is empty image)
There is -init-centerline option in _propseg_ alg which I used when I had these problems to initialize the SC segmentation by manually created centerline, but this option is not available in _deepseg_. 
Haven't you thought about adding it?

![sct_deepseg_sc-segmentation_fails](https://user-images.githubusercontent.com/39456460/43714637-e9749d6c-997d-11e8-9de0-6eb94243ed51.png)

jcohenadad:
hi @valosekj, thank you for the feedback. Indeed, we agree and have already opened an issue: https://github.com/neuropoly/spinalcordtoolbox/issues/1952. I have just increased the priority level. We can close this issue (duplicate of #1952).

valosekj:
Hi @jcohenadad, thank you for response. It's nice to see you are working on it already. Sorry for duplication.

charleygros:
Hi @valosekj,

Apologies for the delay.

To add a manual centerline as input of `sct_deepseg_sc` is now available in the current version of the `master` branch of SCT and will be included in the next release.

To use it, please run:
```
sct_deepseg_sc -i t2.nii.gz -c t2 -centerline manual -file_centerline t2_centerline_manual.nii.gz
```
or you could also use the viewer to generate the centerline:
```
sct_deepseg_sc -i t2.nii.gz -c t2 -centerline viewer
```

Thank you for raising this point, really appreciated.

Issue #1952, opened 2018-07-19
Add a flag to sct_deepseg_sc and sct_deepseg_lesion to input centerline


jcohenadad:
**Feature justification:**
In some cases, the centerline detection (svm- or cnn-based) fails, and it would be nice to be able to bypass this step by inputing our own centerline (e.g., obtained manually).

**Describe the solution you'd like**
A new option in the flag "centerline", where a file name would be input. Example of usage:
~~~
sct_deepseg_lesion -i my_image.nii.gz -c t2 -centerline file=my_centerline.nii.gz
~~~

@charleygros what do you think?

jcohenadad:
also see https://github.com/neuropoly/spinalcordtoolbox/issues/1985

charleygros:
@jcohenadad : I agree, this would be a very relevant feature.

Should we also allow the user to call the viewer to create the centerline, as in PropSeg (and suggested here #1985 ) ?
```
sct_deepseg_lesion -i my_image.nii.gz -c t2 -centerline viewer
```

jcohenadad:
yes, we could indeed call the centerline-viewer module directly from `sct_deepseg_sc`.

Issue #1855, opened 2018-06-12
Enable to input a centerline in case the automatic one fails


jcohenadad:
I'm having trouble segmenting fMRI data. The problem is partly caused by a wrong centerline detection. Inputting a manual centerline might help. This option would be equivalent to `-init-centerline` in `sct_propseg`. 

@charleygros what do you think?

PR #2020, opened 2018-09-10
sct_deepseg_sc/lesion: allow to input manual or semi manual centerline

charleygros:
**Rational:**
In some cases, the centerline detection (svm- or cnn-based) fails, and it would be nice to be able to bypass this step by inputing our own centerline (e.g., obtained semi- or fully-manually).

**New features:**
A new option in the flag "centerline", where a file name or "viewer" can be input.
Example of usage with the viewer:
```
sct_deepseg_lesion -i t2.nii.gz -c t2 -centerline viewer
```
Example of usage with the manual centerline:
```
sct_deepseg_sc -i t2.nii.gz -c t2 -centerline t2_centerline_manual.nii.gz
```

**Approach:**
- For the viewer: it uses the new version of the viewer and then use NURBS to interpolate the viewer_labels to a continuous centerline, with the following parameters: `algo_fitting='nurbs', nurbs_pts_number=10000`.
- For the manual centerline: we applied the same preprocessing steps as to the input image.

**Related issues:**
- #1952 
- #1855
- #1985


scripts/sct_deepseg_lesion.py
```diff
@@ -51,14 +51,14 @@ def get_parser():
                       mandatory=True,
                       example=['t2', 't2_ax', 't2s'])
     parser.add_option(name="-centerline",
-                      type_value="multiple_choice",
```
jcohenadad: you want to keep "multiple_choice", otherwise the values "cnn" and "svm" will be considered as a nifti_input and the program will fail after parser checking.

scripts/sct_deepseg_lesion.py
```diff
@@ -51,14 +51,14 @@ def get_parser():
...
-                      description="choice of spinal cord centerline algorithm.",
+                      type_value="image_nifti",
+                      description="Method used for extracting the centerline.\nsvm: automatic centerline detection, based on Support Vector Machine algorithm.\ncnn: automatic centerline detection, based on Convolutional Neural Network.\nviewer: semi-automatic centerline generation, based on manual selection of a few points using an interactive viewer, then approximation with NURBS.\nprovide the filename of a manual centerline (e.g. t2_centerline_manual.nii.gz).\n",
```
jcohenadad: This part of the doc is a bit confusing wrt. input file name (it is not clear what one should write after `-centerline` for inputing a file name. I would rewrite as:
~~~
                      type_value="multiple_choice",
                      description="Method used for extracting the centerline.
"\nsvm: automatic centerline detection, based on Support Vector Machine algorithm."
"\ncnn: automatic centerline detection, based on Convolutional Neural Network."
"\nviewer: semi-automatic centerline generation, based on manual selection of a few points using an interactive viewer, then approximation with NURBS."
"\nFILENAME: filename of an existing centerline (e.g. t2_centerline_manual.nii.gz).\n",
~~~
Now, the problem is that my suggestion above will NOT work, because `multiple_choice` will look for the string `FILENAME` instead of an actual file name :-(

I see two alternatives:
- replace `FILENAME` by `manual` and have another flag dedicated to inputing a centerline file, e.g.: `sct_deepseg_sc -i t2.nii.gz -c t2 -centerline manual -filecenterline t2_centerline_manual.nii.gz`. Cons: heavy.
- modify the Parser to have the possibility to not check the input string when using `multiple_choice`. Cons: defeat the purpose of a parsing checker...

@zougloub any idea?

scripts/sct_deepseg_sc.py
```diff
@@ -411,6 +411,65 @@ def heatmap2optic(fname_heatmap, lambda_value, fname_out, z_max, algo='dpdt'):
...
+
+    image_labels = Image(fname_labels)
+    # fit centerline, smooth it and return the first derivative (in physical space)
+    x_centerline_fit, y_centerline_fit, z_centerline, x_centerline_deriv, y_centerline_deriv, z_centerline_deriv = smooth_centerline(image_labels, algo_fitting='nurbs', nurbs_pts_number=10000, phys_coordinates=True, verbose=False, all_slices=False)
+    centerline = Centerline(x_centerline_fit, y_centerline_fit, z_centerline, x_centerline_deriv, y_centerline_deriv, z_centerline_deriv)
+
+    # average centerline coordinates over slices of the image
+    x_centerline_fit_rescorr, y_centerline_fit_rescorr, z_centerline_rescorr, x_centerline_deriv_rescorr, y_centerline_deriv_rescorr, z_centerline_deriv_rescorr = centerline.average_coordinates_over_slices(image_labels)
+
+    # compute z_centerline in image coordinates for usage in vertebrae mapping
```
jcohenadad: not sure i see the link with vertebral mapping here?

scripts/sct_deepseg_lesion.py
jcohenadad: related to https://github.com/neuropoly/spinalcordtoolbox/pull/2020#discussion_r216363748

scripts/sct_deepseg_lesion.py
charleygros: @jcohenadad: As you said, I used `type:nifti_input` to allow the user to input a manul centerline. While the other options (i.e. `svm`, `cnn`, `viewer`) are not considered as `Image`, from my understanding, thanks to the parameter ` list_no_image=['svm', 'cnn', 'viewer']`.
--> This approach is inspired from PropSeg:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_propseg.py#L242

scripts/sct_deepseg_lesion.py
charleygros: My answer here: https://github.com/neuropoly/spinalcordtoolbox/pull/2020#discussion_r216497052

scripts/sct_deepseg_sc.py
charleygros: Corrected in commit #86fa0db

scripts/sct_deepseg_lesion.py
charleygros: Part of the doc corrected as suggested by @jcohenadad, in commit #f34bf6b.

jcohenadad: the problem is that the way this is currently done in PropSeg is not good either (for the same reasons as here). It would be nice to find a proper solution and then deploy it to the rest of SCT. Waiting for @zougloub's feedback there.

scripts/sct_deepseg_lesion.py
```diff
@@ -190,6 +190,9 @@ def deep_segmentation_MSlesion(fname_image, contrast_type, output_folder, ctr_al
     tmp_folder = sct.TempFolder()
     tmp_folder_path = tmp_folder.get_path()
     fname_image_tmp = tmp_folder.copy_from(fname_image)
+    if os.path.isfile(ctr_algo):  # if the ctr_algo is a manual centerline file
+        ctr_algo = os.path.basename(ctr_algo)
```
zougloub: why take the base name?

scripts/sct_deepseg_sc.py
```diff
@@ -136,7 +136,7 @@ def crop_image_around_centerline(filename_in, filename_ctr, filename_out, crop_s
     x_lst, y_lst = [], []
     data_im_new = np.zeros((crop_size, crop_size, im_in.dim[2]))
     for zz in range(im_in.dim[2]):
-        if 1 in np.array(data_ctr[:, :, zz]):
+        if np.sum(np.array(data_ctr[:, :, zz])):
```
zougloub: could also use `np.any` for more semantic

scripts/sct_deepseg_lesion.py
charleygros: For the same reason as:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg_lesion.py#L189
--> If we pass the if condition, then the `ctr_algo` is the filename of a nifti mask.

Do you recommend to change the name of the variable (initially this variable reffered to algorithm name)? to change the if condition (ie check if it is a nifti image)?

scripts/sct_deepseg_sc.py
charleygros: done

scripts/sct_deepseg_lesion.py
zougloub: I just want to know why you are trimming the path to its basename, why not keep a full path or absolute path. What if the file is in another folder?
Note: I just took like 1 minute to do a very quick review your code, I didn't look at the implications, just asking.

charleygros: hehe! very good point thank you.
I swap the two lines (see commit ccd068a) : now I copy the file in the tmp folder then I extract the basename --> which will be usefull in the remaining of the code when processing in the tmp_folder.

scripts/sct_deepseg_lesion.py
charleygros: @jcohenadad : @zougloub approved the PR, should I merge the branch or wait for further discussions about the point you raised?

jcohenadad: i suggest the following:
~~~
"svm: automatic centerline detection, based on Support Vector Machine algorithm."
"cnn: automatic centerline detection, based on Convolutional Neural Network."
"viewer: semi-automatic centerline generation, based on manual selection of a few points using an interactive viewer, then approximation with NURBS."
"manual: filename of an existing centerline (e.g. t2_centerline_manual.nii.gz).\n",
~~~
and another flag:
~~~
-file-centerline; type:type:image_nifti; description: Input centerline file (to use with flag -centerline manual)
~~~

charleygros: @jcohenadad : I have made the modifications we discussed yesterday.
Ready to be reviewed.

scripts/sct_deepseg_lesion.py
```diff
@@ -339,10 +339,13 @@ def main():
         sct.log.error('Please use the flag -file_centerline to indicate the centerline filename.')
         sys.exit(1)
     
-    manual_centerline_fname = arguments["-file_centerline"]
-    if "-file_centerline" in args and ctr_algo != 'manual':
-        sct.log.error('Please add to the command "-centerline manual".')
-        sys.exit(1)
+    if "-file_centerline" in args:
+        manual_centerline_fname = arguments["-file_centerline"]
+        if ctr_algo != 'manual':
```
jcohenadad: alternatively we can automatically set `ctr_algo='manual'` (or ignore that field) if user enters a manual centerline. It will make lighter syntax.

charleygros: Good suggestion, thank you. Done in the last commit.

Issue #1983, opened 2018-08-03
When input image has 3rd axis which is not I-S, it fails


jcohenadad:
### Description

Those motion correction functions assume that the input image has the 3rd axis along S-I direction, which is usually the case if acquisition is axial. However, when it is not the case, it fails miserably, e.g. [here](https://sourceforge.net/p/spinalcordtoolbox/discussion/help/thread/d2f130c2/).

Two possible solutions. Check if input image is in axial orientation, and if not:
- Exit with ERROR mentioning that user has to re-orient data using `sct_image -setorient`.
- Reorient the image and continue processing.

PR #2022, opened 2018-09-10
sct_fmri_moco: Generalize motion correction for sagittal acquisitions and other improvements

### LINKED ISSUE #820: when using -f box, if box is larger than field of view, it creates wrong result
data: **sct_example_data**

```
sct_create_mask -i dmri_mean.nii.gz -p centerline,t2_seg_reg.nii.gz -size 21 -f box
```

screen shot 2016-04-15 at 20 22 19
--> OK

```
sct_create_mask -i dmri_mean.nii.gz -p centerline,t2_seg_reg.nii.gz -size 51 -f box
```

screen shot 2016-04-15 at 20 22 33
--> weird result

```
sct_create_mask -i dmri_mean.nii.gz -p centerline,t2_seg_reg.nii.gz -size 51 -f cylinder
```

screen shot 2016-04-15 at 20 22 48
--> OK

### LINKED ISSUE #1878: -copy-header usage is misleading
it says it copies the header of the input image (specified in -i) to the destination image, although it does the opposite.

also, when trying this function, it copied the data, not only the header (!)

### LINKED ISSUE #1984: Generalize motion correction for sagittal acquisitions
### Feature justification
Currently, `sct_dmri_moco` and `sct_fmri_moco` are designed such that the input image is assumed to be in axial orientation (i.e. 3rd dimension is S-I), and the algorithm estimates a set of Tx,Ty for each slice, regularized using polynom functions along the S-I direction.

However, for images acquired sagitally (or coronally), this approach does not work anymore. Beyond the simple orientation issue described in https://github.com/neuropoly/spinalcordtoolbox/issues/1983, the type of motion-related corruption is not addressed by the current algorithm. More specifically, if motion occurs between sagittal slices, then the cord is non-linearly distorted in the axial plane and a simple translation within the axial plane is not sufficient. This is shown in the animation below. Without motion correction (left), a few slices are translated in the AP direction, leaving ugly distortion of the cord in the axial plane. With motion correction (right), the algorithm tries to find an average translation, but this is of course not sufficient for this type of motion. 

anim

### Possible solution
A possible solution would be to implement an algorithm for sagittal acquisitions, that would estimate a Ty translation (along AP) for each row (i.e. each sagittal slice), and this Ty translation would be regularized along the SI direction, as currently done by `isct_antsSliceRegularizedRegistration`.

### To reproduce

data: **duke:sct_testing/issues/20180905_issue1984**

Related issue: https://sourceforge.net/p/spinalcordtoolbox/discussion/help/thread/d2f130c2/

### LINKED ISSUE #2017: When splitting along Z a 4D file, each output becomes 3D

jcohenadad:
Currently, sct_fmri_moco is designed such that the input image is assumed to be in axial orientation (i.e. 3rd dimension is S-I), and the algorithm estimates a set of Tx,Ty for each slice, regularized using polynom functions along the S-I direction.

However, for images acquired sagitally (or coronally), this approach does not work anymore. Beyond the simple orientation issue described in #1983, the type of motion-related corruption is not addressed by the current algorithm. More specifically, if motion occurs between sagittal slices, then the cord is non-linearly distorted in the axial plane and a simple translation within the axial plane is not sufficient. This is shown in the animation below. Without motion correction (left), a few slices are translated in the AP direction, leaving ugly distortion of the cord in the axial plane. With motion correction (right), the algorithm tries to find an average translation, but this is of course not sufficient for this type of motion.

anim

The implemented solution is to split the data along Z (R-L dimension) and use `antsRegistration` in 2D mode for each slice independently. For now we use affine transfo, which is fast and robust, but later we could investigate the use of other transfo. What helps in particular is to use a mask to focus on a region of interest for registration (esp. because motion is non-affine). See below a comparison without (left) and with the new moco algorithm (right):

anim

This PR also provides a few other fixes, refactoring and improvements (runs faster).

Fixes #1984, Fixes #820, Fixes #2017, Fixes #1878

### DIFF
--- a/scripts/msct_moco.py
+++ b/scripts/msct_moco.py
@@ -140,7 +140,7 @@ def moco(param):
 
         # Motion correction: Loop across T
         for indice_index in tqdm(range(nt), unit='iter', unit_scale=False,
-                                 desc="Z=" + str(iz) + "/" + str(len(file_data_splitZ)-1), ascii=False, ncols=80):
+                                 desc="Z=" + str(iz) + "/" + str(len(file_data_splitZ)-1), ascii=True, ncols=80):
 
             # create indices and display stuff
             it = index[indice_index]
--- a/scripts/msct_moco.py
+++ b/scripts/msct_moco.py
@@ -128,7 +128,7 @@ def moco(param):
         list_im_zt = split_data(im_z, dim=3)
         file_data_splitZ_splitT = []
         for im_zt in list_im_zt:
-            im_zt.save()
+            im_zt.save(verbose=0)
             file_data_splitZ_splitT.append(im_zt.absolutepath)
         # file_data_splitT = file_data + '_T'
 
@@ -139,14 +139,12 @@ def moco(param):
         failed_transfo = [0 for i in range(nt)]
 
         # Motion correction: Loop across T
-        tqdm_bar = tqdm(total=nt, unit='iter', unit_scale=False, desc="Z=" + str(iz), ascii=True, ncols=80)
+        tqdm_bar = tqdm(total=nt, unit='iter', unit_scale=False,
+                        desc="Z=" + str(iz) + "/" + str(len(file_data_splitZ)-1), ascii=True, ncols=80)
         for indice_index in range(nt):
 
             # create indices and display stuff
             it = index[indice_index]
-            # file_data_splitT_num.append(file_data_splitT + str(it).zfill(4))
-            # file_data_splitZ_splitT_moco.append(file_data + suffix + '_T' + str(it).zfill(4))
-            # sct.printv(('\nVolume ' + str((it)) + '/' + str(nt - 1) + ':'), verbose)
             file_mat[iz][it] = os.path.join(folder_mat, "mat.Z") + str(iz).zfill(4) + 'T' + str(it).zfill(4)
             file_data_splitZ_splitT_moco.append(sct.add_suffix(file_data_splitZ_splitT[it], '_moco'))
             # deal with masking
@@ -167,14 +165,10 @@ def moco(param):
                 data_targetz = (data_targetz * (indice_index + 1) + data_mocoz) / (indice_index + 2)
                 im_targetz.data = data_targetz
                 im_targetz.save(verbose=0)
-                # sct.run(["sct_maths", "-i", file_target_splitZ[iz], "-mul", str(indice_index + 1), "-o", file_target_splitZ[iz]])
-                # sct.run(["sct_maths", "-i", file_target_splitZ[iz], "-add", file_data_splitZ_splitT_moco[it], "-o", file_target_splitZ[iz]])
-                # sct.run(["sct_maths", "-i", file_target_splitZ[iz], "-div", str(indice_index + 2), "-o", file_target_splitZ[iz]])
             tqdm_bar.update()
         tqdm_bar.close()
 
         # Replace failed transformation with the closest good one
-        # sct.printv(('\nReplace failed transformations...'), verbose)
         fT = [i for i, j in enumerate(failed_transfo) if j == 1]
         gT = [i for i, j in enumerate(failed_transfo) if j == 0]
         for it in range(len(fT)):
@@ -199,18 +193,11 @@ def moco(param):
         # Merge data along T
         file_data_splitZ_moco.append(sct.add_suffix(file, suffix))
         if todo != 'estimate':
-            # sct.printv('\nMerge data back along T...', verbose)
-            # im_list = []
-            # fname_list = []
-            # for indice_index in range(len(index)):
-                # im_list.append(Image(file_data_splitZ_splitT_moco[indice_index] + ext))
-                # fname_list.append(file_data_splitZ_splitT_moco[indice_index] + ext)
             im_out = concat_data(file_data_splitZ_splitT_moco, 3)
             im_out.save(file_data_splitZ_moco[iz])
 
     # If sagittal, merge along Z
     if param.is_sagittal:
-        # sct.printv('\nMerge data back along Z...', verbose)
         im_out = concat_data(file_data_splitZ_moco, 2)
         im_out.save(file_data + suffix + ext)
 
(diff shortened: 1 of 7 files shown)


scripts/sct_create_mask.py
```diff
@@ -110,12 +101,11 @@ def create_mask(param):
 
     path_tmp = sct.tmp_create(basename="create_mask", verbose=param.verbose)
 
-    sct.printv('\nCheck orientation...', param.verbose)
+    sct.printv('\nOrientation:', param.verbose)
```
zougloub: Hmmm, which PR do you want to merge first?

jcohenadad: i knew there would be a problem there... i started #2021, then found a bug which is solved by #2022, but then i needed the changes from #2022 to be able to continue working on #2021, so i did a rebase 2021-->2022 and here we are.

scripts/sct_deepseg_sc.py
```diff
@@ -411,7 +416,66 @@ def heatmap2optic(fname_heatmap, lambda_value, fname_out, z_max, algo='dpdt'):
...
+
+    # # compute z_centerline in image coordinates for usage in vertebrae mapping
+    # voxel_coordinates = image_labels.transfo_phys2pix([[x_centerline_fit_rescorr[i], y_centerline_fit_rescorr[i], z_centerline_rescorr[i]] for i in range(len(z_centerline_rescorr))])
+    # x_centerline_voxel = [coord[0] for coord in voxel_coordinates]
+    # y_centerline_voxel = [coord[1] for coord in voxel_coordinates]
+    # z_centerline_voxel = [coord[2] for coord in voxel_coordinates]
+
+    # compute z_centerline in image coordinates with continuous precision
+    voxel_coordinates = image_labels.transfo_phys2pix([[x_centerline_fit_rescorr[i], y_centerline_fit_rescorr[i], z_centerline_rescorr[i]] for i in range(len(z_centerline_rescorr))], real=False)
+    x_centerline_voxel_cont = [coord[0] for coord in voxel_coordinates]
```
zougloub: Note that since transfo_phys2pix now returns a numpy array, you could have used something like `x, y, z = voxel_coordinates.T`

scripts/sct_dmri_moco.py
```diff
@@ -369,36 +372,30 @@ def dmri_moco(param):
 
     # DWI groups
     file_dwi_mean = []
+    tqdm_bar = tqdm(total=nb_groups, unit='iter', unit_scale=False, desc="Merge within groups", ascii=True, ncols=80)
     for iGroup in range(nb_groups):
```
zougloub: You could have used a tqdm iterator object instead of the range() here (avoids to `update()` and `close()`)

scripts/sct_fmri_moco.py
```diff
@@ -6,29 +6,26 @@
...
 
 from __future__ import absolute_import, division
 
 import sys
 import os
-import getopt
 import time
 import math
-
+import shutil
```
zougloub: There is a `copy()` function in `sctutils`

scripts/sct_dmri_moco.py
jcohenadad: i was not aware of that-- i'll look into it

scripts/msct_moco.py
```diff
@@ -66,160 +68,242 @@ def moco(param):
...
+            file_target_splitZ.append(im_targetz.absolutepath)
+        # z-split mask (if exists)
+        if not param.fname_mask == '':
+            im_maskz_list = split_data(Image(file_mask), dim=dim_sag, squeeze_data=False)
+            file_mask_splitZ = []
+            for im_maskz in im_maskz_list:
+                im_maskz.save()
+                file_mask_splitZ.append(im_maskz.absolutepath)
+        # initialize file list for output matrices
+        file_mat = np.chararray([nz, nt],
```
zougloub: `chararray` is deprecated and also not a good idea
https://docs.scipy.org/doc/numpy/reference/generated/numpy.chararray.html

scripts/msct_moco.py
```diff
@@ -66,160 +68,242 @@ def moco(param):
...
+                data_targetz = (data_targetz * (indice_index + 1) + data_mocoz) / (indice_index + 2)
+                im_targetz.data = data_targetz
+                im_targetz.save(verbose=0)
+            tqdm_bar.update()
+        tqdm_bar.close()
+
+        # Replace failed transformation with the closest good one
+        fT = [i for i, j in enumerate(failed_transfo) if j == 1]
+        gT = [i for i, j in enumerate(failed_transfo) if j == 0]
+        for it in range(len(fT)):
```
zougloub: `range(len(x))` -> `enumerate(fT)`

scripts/msct_moco.py
```diff
@@ -66,160 +68,242 @@ def moco(param):
...
+            abs_dist = [abs(gT[i] - fT[it]) for i in range(len(gT))]
```
zougloub: Use numpy?

scripts/sct_fmri_moco.py
```diff
@@ -313,27 +344,40 @@ def fmri_moco(param):
...
+        # if flag g=1, it means that all images have already been corrected, so we just need to rename the file
+        sct.mv('fmri_averaged_groups_moco.nii', 'fmri_moco.nii')
+    else:
+        # create final mat folder
+        sct.create_folder(mat_final)
+
+        # Copy registration matrices
+        sct.printv('\nCopy transformations...', param.verbose)
+        for iGroup in range(nb_groups):
+            for data in range(len(group_indexes[iGroup])):
```
zougloub: enumerate

scripts/sct_image.py
```diff
@@ -347,31 +351,44 @@ def split_data(im_in, dim):
...
+        else:
+            im_out.data = dat
         im_out.absolutepath = sct.add_suffix(im_in.absolutepath, "_{}{}".format(dim_list[dim].upper(), str(idx_img).zfill(4)))
         im_out_list.append(im_out)
 
     return im_out_list
 
 
-def concat_data(fname_in_list, dim, pixdim=None):
+def concat_data(fname_in_list, dim, pixdim=None, squeeze_data=False):
```
zougloub: Why do you absolutely want to have squeezing performed within this function and not elsewhere (caller or otherwise)?

scripts/sct_dmri_moco.py
jcohenadad: fixed in c68600ddc6cc801ebe7ebf52aa7373f1d651050b

scripts/sct_fmri_moco.py
jcohenadad: Fixed in 44e02e4e44c74da51cfb2cc9a95eb025155cd87a

scripts/msct_moco.py
jcohenadad: after spending 10min on this i need help-- i've tried 
~~~
file_mat = np.array([nz, nt], dtype="str")
~~~
but pb because string is immutable. solution?

scripts/msct_moco.py
jcohenadad: Fixed in 14f6794f03adc8a6a14b23ad575a1d6c2161b3bd

scripts/sct_image.py
jcohenadad: squeeze if False, so I guess it is not performed-- or am i missing something?

scripts/sct_fmri_moco.py
jcohenadad: fixed in ef52dd32eb9b816f95e392767a9695d0d2c0cc02

scripts/msct_moco.py
jcohenadad: fixed in 7b3647c678dc874a9146d6cdc34cf950d1596dbd

scripts/sct_image.py
zougloub: just a question of design. why would a function named concat_data perform a squeeze.

Issue #1837, opened 2018-06-08
sct_get_centerline -method viewer: should use the new viewer


jcohenadad:
It would be nice to use the new viewer for `sct_get_centerline` script. The new viewer is for example used by `sct_propseg` when calling e.g.: `sct_propseg -i IMAGE -c CONTRAST -init-centeline viewer` (the viewer is called here: https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_propseg.py#L540)

charleygros:
I think we can close this issue, is it correct?

PR #2035, opened 2018-10-05
sct_get_centerline: use the new viewer

charleygros:
`sct_get_centerline` currently uses `sct_viewer.ClickViewerPropseg()` instead of the new `spinalcordtoolbox.gui.*` viewer.

This Pull Request aims at integrating the new viewer into `sct_get_centerline`.

To test it, please run:
```
cd example_data/t2
sct_get_centerline -i t2.nii.gz -method viewer
sct_image -i t2.nii.gz -setorient SAL -o t2SAL.nii.gz
sct_get_centerline -i t2SAL.nii.gz -method viewer
```

This code was in `sct_deepseg_sc` --> I moved it in `sct_get_centerline` --> that s why both `sct_deepseg_sc` and `sct_deepseg_lesion` now called functions from `sct_get_centerline`.

To test it, please run:
```
cd example_data/t2
sct_deepseg_sc -i t2.nii.gz -c t2 -centerline viewer
```

This PR fixes the issue #1837.

### DIFF
--- a/scripts/sct_get_centerline.py
+++ b/scripts/sct_get_centerline.py
@@ -14,16 +14,22 @@ import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
 
 
-def _call_viewer_centerline(fname_in):
+def _call_viewer_centerline(fname_in, interslice_gap):
     from spinalcordtoolbox.gui.base import AnatomicalParams
     from spinalcordtoolbox.gui.centerline import launch_centerline_dialog
 
+    im_data = Image(fname_in)
+
+    # Get the number of slice along the (IS) axis
+    im_tmp = msct_image.change_orientation(im_data, 'RPI')
+    _, _, nz, _, _, _, pz, _ = im_tmp.dim
+    del im_tmp
+
     params = AnatomicalParams()
     # setting maximum number of points to a reasonable value
-    params.num_points = 20
-    params.interval_in_mm = 30
+    params.num_points = np.ceil(nz * pz / interslice_gap) + 1
+    params.interval_in_mm = interslice_gap
     params.starting_slice = 'top'
-    im_data = Image(fname_in)
 
     im_mask_viewer = msct_image.zeros_like(im_data)
     im_mask_viewer.absolutepath = sct.add_suffix(fname_in, '_labels_viewer')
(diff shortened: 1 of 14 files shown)


jcohenadad:
i still find issues if the number of points is too low (switches to hanning, but we don't know because verbose is set to False). For this PR, i would suggest to only introduce the new viewer, without perturbating too much the behaviour (which requires extensive testing on larger set of images). So i would suggest the following modifs:
- set verbose to "verbose" (instead of 0) when calling smooth_centerline
- set nurbs_pts_number=3000 (as was the case before)
- use the same number of points & gap as before (was calculated based on max number of slices)
- if not too much trouble, replace `_from_viewerLabels_to_centerline` by the already-existing `sct_process_segmentation.extract_centerline` (which does the same thing)

happy to help :-)

charleygros:
@jcohenadad: Thank you for your help.
The last commits do:
- replace `_from_viewerLabels_to_centerline` by the already-existing `sct_process_segmentation.extract_centerline`
- set verbose to "verbose" (instead of 0) when calling `smooth_centerline` in `sct_process_segmentation.extract_centerline`
- use the same number of points & gap as before (calculated based on max number of slices)

I tested it on the example_data with the default param (gap=10) and it seems to work well.
Please let me know your thoughts

scripts/sct_get_centerline.py
```diff
@@ -12,39 +12,39 @@
...
 import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
-from msct_types import Centerline
-from sct_viewer import ClickViewerPropseg
-from sct_straighten_spinalcord import smooth_centerline
+from sct_process_segmentation import extract_centerline
 
 
-def viewer_centerline(image_input_reoriented, interslice_gap, verbose):
+def _call_viewer_centerline(fname_in, interslice_gap=10):
```
jcohenadad: i find it a bit too dense now, how about setting a compromise with `interslice_gap=20`?

charleygros: I agree!

charleygros:
@jcohenadad: By increasing the default gap to 20.0mm, I observed 'not smooth centerline' again.
In the last commits, I proposed to increase the number of points used by the nurbs (3000 to 8000) --> looks much better. However it takes more time.

Another approach would be to leave gap=20 and `nurbs_pts_number=3000` but to send a message to the user if: https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_straighten_spinalcord.py#L160
--> asking the user to decrease the gap used by the viewer.

jcohenadad:
@charleygros let's go with your latest commit. We definitely don't want to have an approach where a message is sent, asking the user to make a decision (99.99% of our users don't read the terminal output).

PR #2073, opened 2018-11-22
sct_smooth_spinalcord: Enable to set smoothing parameters in all axes

### LINKED ISSUE #2072: Enable anisotropic smoothing and sigma in float
Currently, `sct_smooth_spinalcord` only performs smoothing along the Z direction (after straightening). Some users might want to perform an anisotropic smoothing, so we should enable this option.

Also, the flag `-smooth` only accepts "int", whereas it should accept ¨float¨(as is the case for the function `sct_maths.smooth()` which is called by `sct_smooth_spinalcord`).

Also see: https://sourceforge.net/p/spinalcordtoolbox/help/4/

jcohenadad:
Currently, `sct_smooth_spinalcord` only performs smoothing along the Z direction (after straightening). Users [requested](https://sourceforge.net/p/spinalcordtoolbox/help/4/) the possibility to perform an anisotropic smoothing. This PR enables this option. Changes are:
- enable to set smoothing parameters in all axes
- enable to specify kernel Sigma in "float" (previously "int")
- now enabled caching of straightening (if function is run more than one time with the same image, straightening does not need to be re-done)

WARNING: This refactoring breaks cross-compatibility: the default behaviour of flag "-smooth X" will now produce a kernel of "X,X,X" (as opposed to 0,0,X before). This decision was chosen to be consistent with usage of `sct_maths`. This mean that this PR will be integrated in the next X.Y version of SCT (likely v3.3). We will need to update the documentation/course material to accommodate this change --> batch_processing and latest SCT course (Paris) are using the default parameter, so it's all good 👍 

Fixes #2072

### DIFF
--- a/scripts/sct_smooth_spinalcord.py
+++ b/scripts/sct_smooth_spinalcord.py
@@ -198,6 +198,7 @@ def main(args=None):
 
     # Smooth the straightened image along z
     sct.printv('\nSmooth the straightened image...')
+    sigma_smooth = ",".join([str(i) for i in sigma])
     sct_maths.main(args=['-i', 'anat_rpi_straight.nii',
                          '-smooth', ",".join([str(i) for i in sigma]),
                          '-o', 'anat_rpi_straight_smooth.nii',
--- a/scripts/sct_smooth_spinalcord.py
+++ b/scripts/sct_smooth_spinalcord.py
@@ -21,6 +21,7 @@ import sys, io, os, getopt, shutil, time
 import numpy as np
 
 import sct_utils as sct
+import sct_maths
 import spinalcordtoolbox.image as msct_image
 from sct_convert import convert
 from msct_parser import Parser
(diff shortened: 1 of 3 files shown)


scripts/sct_smooth_spinalcord.py
```diff
@@ -195,11 +192,16 @@ def main(args=None):
...
         sct.cache_save(cachefile, cache_sig)
+        # move warping fields locally (to use caching next time)
+        sct.copy('warp_curve2straight.nii.gz', os.path.join(curdir, 'warp_curve2straight.nii.gz'))
+        sct.copy('warp_straight2curve.nii.gz', os.path.join(curdir, 'warp_straight2curve.nii.gz'))
 
     # Smooth the straightened image along z
-    sct.printv('\nSmooth the straightened image along z...')
-    sct.run(['sct_maths', '-i', 'anat_rpi_straight.nii', '-smooth', '0,0,' + str(sigma), '-o', 'anat_rpi_straight_smooth.nii'], verbose)
-
+    sct.printv('\nSmooth the straightened image...')
```
perone:
```suggestion
    sct.printv('\nSmooth the straightened image...')
    sigma_smooth = ",".join([str(i) for i in sigma])
```

scripts/sct_smooth_spinalcord.py
```diff
@@ -195,11 +192,16 @@ def main(args=None):
...
+    sct_maths.main(args=['-i', 'anat_rpi_straight.nii',
+                         '-smooth', ",".join([str(i) for i in sigma]),
```
perone:
```suggestion
                         '-smooth', sigma_smooth,
```

PR #2080, opened 2018-11-23
Now possible to output SNR map, removed requirement for inputing mask, and few other improvements

### LINKED ISSUE #2078: Option to remove volumes for mult mode, output SNR map, make -m optional and other refactoring
Proposed features:
- In case users need to discard the 3 first measurements it would be good to be able to use ":" for selecting a range of volumes with flag "-vol", e.g., `-vol 2:100`
- It would be nice to output an SNR map when the mult method is used (not possible with diff)
- With the previous point, the flag `-m` can be made optional

Proposed refactoring:
- Remove -z and -vert flags. It makes the code heavy. If users want to compute within particular ROIs, then ROI generation should be wrapped outside of this function.

jcohenadad:
The following features/fixes were implemented:
- In case users need to discard the 3 first measurements, it is not possible to use ":" for selecting a range of volumes with flag "-vol", e.g., -vol 2:100
- Now possible to output an SNR map with `-method mult`
- With the previous point, the flag -m has been made optional
- Removed -z and -vert flags. It made the code heavy. If users want to compute within particular ROIs, then ROI generation should be wrapped outside of this function.
- Fixed Rayleigh correction for SNR_diff

Fixes #2078

### DIFF
--- a/scripts/sct_compute_snr.py
+++ b/scripts/sct_compute_snr.py
@@ -12,17 +12,12 @@
 # About the license: see the file LICENSE.TXT
 ########################################################################################
 
-# TODO: compute in float64
-
 from __future__ import division, absolute_import
 
 import sys
-import operator
 import numpy as np
 from msct_parser import Parser
 from spinalcordtoolbox.image import Image, empty_like
-from spinalcordtoolbox.utils import parse_num_list
-from spinalcordtoolbox.template import get_slices_from_vertebral_levels
 import sct_utils as sct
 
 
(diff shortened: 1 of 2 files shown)


scripts/sct_compute_snr.py
```diff
@@ -10,28 +10,31 @@
...
 from msct_parser import Parser
-from spinalcordtoolbox.image import Image
+from spinalcordtoolbox.image import Image, empty_like
 from spinalcordtoolbox.utils import parse_num_list
-from spinalcordtoolbox.template import get_slices_from_vertebral_levels
 import sct_utils as sct
 
 
+# PARAMETERS
+class Param:
```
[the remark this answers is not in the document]
perone: Since we're still using Python 2.x, it's better to use the new-style classes.

scripts/sct_compute_snr.py
```diff
@@ -10,28 +10,31 @@
...
+    # The constructor
+    def __init__(self):
+        self.almost_zero = 0.000000001
```
perone: There is also a numpy machine epsilon for something similar:

    import numpy as np
    print(np.finfo(float).eps)

jcohenadad: thanks for the tip! will do the modif

PR #2084, opened 2018-11-28
spinalcordtoolbox/vertebrae/detect_c2c3 -- New module

### LINKED ISSUE #2066: New module to detect C2-C3 disc
In its current implementation (v3.2.7), `sct_label_vertebrae` is poorly robust for detecting C2-C3 disc. It is an important first step for the algorithm, if it fails the rest of the processing fails.

It would be nice to develop a separate, more robust, module that detects C2-C3 disc.

charleygros:
In its current implementation (v3.2.7), `sct_label_vertebrae` is poorly robust for detecting C2-C3 disc. It is an important first step for the algorithm, if it fails the rest of the processing fails.

This PR implements a separate, more robust, module that detects C2-C3 disc. This approach is based on a method by [Gros et al.](doi.org/10.1016/j.media.2017.12.001).

Fixes #2066.

### To test:
1. Re-install the toolbox: `./install_sct -m -b`
2. Test it on a example data:
```
cd example_data/t2
source sct_launcher
python
```
then:
```
from spinalcordtoolbox.vertebrae.detect_c2c3 import detect_c2c3
i=Image('t2.nii.gz')
s=Image('t2_seg_manual.nii.gz')
label_c2c3=detect_c2c3(i,s,'t2')
label_c2c3.save("t2_c2c3.nii.gz")
```

### Subjects used for training:
- [x] T2 [here](https://drive.google.com/file/d/1yaXVyY0VoX5IR47awaqCWgXHwYmuAswG/view?usp=sharing)
- [x] T1 [here](https://drive.google.com/file/d/11kCYCVcsqsknFJ4McQqfz2eMW7pOm5EQ/view?usp=sharing)


jcohenadad:
hum, with [these data](https://www.dropbox.com/s/zb1lfuut0dwkr21/20181129_sct_c2c3.zip?dl=0)

and this version: 58a33ee3e5df97c87ed637919ccccee5a390d6ab

i'm getting this error:
~~~
>>> import os
>>> os.chdir("/Users/julien/sct_example_data/t2")
>>> i=Image('t2.nii.gz')
>>> s=Image('t2_seg.nii.gz')
>>> label_c2c3=detect_c2c3(i,s,'t2')
detection real   time in sec : 0.0599999
detection user   time in sec : 0.06
detection system time in sec : 0
Assertion failed: (nz>1), function centerline_dpdt, file /Users/creatis/sdika/spine/src/HOGDetect.cpp, line 1324.
~~~

spinalcordtoolbox/vertebrae/detect_c2c3.py
```diff
@@ -0,0 +1,116 @@
...
+import sct_utils as sct
+from sct_flatten_sagittal import flatten_sagittal
+import numpy as np
+import nibabel as nib
+from scipy.ndimage.measurements import center_of_mass
+
+import spinalcordtoolbox.image as msct_image
+from spinalcordtoolbox.image import Image, zeros_like
+
+PATH_MODEL = '/home/charley/Téléchargements/20181128_model_c2c3_t1.yml'
```
jcohenadad: non-ascii characters + hard-coded path

scripts/sct_download_data.py
```diff
@@ -110,7 +111,8 @@ def main(args=None):
         'deepseg_sc_models': ['https://osf.io/avf97/?action=download',
                               'https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip'],
         'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip']
+                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip'],
+        'c2c3_disc_models': ['https://www.neuro.polymtl.ca/_media/downloads/sct/20181129_c2c3_disc_models.zip']
```
jcohenadad: it would be important to add a README in the zip that lists the subjects used to generate the model (for replicability)

spinalcordtoolbox/vertebrae/detect_c2c3.py
```diff
@@ -0,0 +1,116 @@
+# coding: utf-8
```
jcohenadad: i suggest to add here a documentation that explains how to train a model (for replicability/testing purpose)

spinalcordtoolbox/vertebrae/detect_c2c3.py
```diff
@@ -0,0 +1,116 @@
...
+    nii_c2c3 = detect_c2c3(nii_im, nii_seg, contrast, verbose=verbose)
+
+    # Output C2-C3 disc label
+    # by default, output in the same directory as the input images
+    sct.printv('Generate output file...', verbose)
+    if fname_c2c3 is None:
+        fname_c2c3 = os.path.join(os.path.dirname(nii_im.absolutepath), "label_c2c3.nii.gz")
+    nii_c2c3.save(fname_c2c3)
+
+    return fname_c2c3
```
jcohenadad: strange return carriage character at the end of the file

spinalcordtoolbox/vertebrae/detect_c2c3.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Extract mid-slice
+    nii_im.change_orientation('PIR')
+    nii_seg_flat.change_orientation('PIR')
+    mid_RL = int(np.rint(nii_im.dim[2] * 1.0 / 2))
+    midSlice = nii_im.data[:, :, mid_RL]
+    midSlice_seg = nii_seg_flat.data[:, :, mid_RL]
+    nii_midSlice = msct_image.zeros_like(nii_im)
+    nii_midSlice.data = midSlice
+    nii_midSlice.save('data_midSlice.nii')
```
jcohenadad: we should do all the i/o in a temporary folder (see `sct.tmp_create()`) to avoid situations where:
- user has a file called data_midSlice.nii in the local folder
- user has writing permission on local folder (e.g. mounted network disk)

jcohenadad:
@charleygros i've done [some evaluations](https://www.dropbox.com/s/7w16e7cqnkiu8rw/20181129_evaluations_c2c3.zip?dl=0) on t1 and t2, from the "largest differently looking" data from the multi-vendor single subject dataset from the spine_generic study, and only 1 out of 7 images fail (it is a t2). So it's a great preliminary result 👍

spinalcordtoolbox/vertebrae/detect_c2c3.py
charleygros: Not sure to understand, could you help me?

charleygros:
> Update: I think the soft is ready to have an "sct" API so we can insert it into a processing pipeline (to easily test more subjects)
> 
> Update: sorry, i forgot that the idea was to call this function directly within `sct_label_vertebrae` (right?)

Yes (said [here](https://github.com/neuropoly/spinalcordtoolbox/issues/2066#issuecomment-440450921))

spinalcordtoolbox/vertebrae/detect_c2c3.py
jcohenadad: actually what i meant is that the file should end with a blank line, [according to most coding guidelines](https://www.reddit.com/r/Python/comments/1zjugg/question_why_does_pep8_recommend_leaving_a_blank/)

jcohenadad:
great, so i think this PR should already implement the call into `sct_label_vertebrae` (to avoid opening another PR). @charleygros do you mind doing it or do you prefer me to do it (i'm more familiar with the code of label vertebrae)?

spinalcordtoolbox/vertebrae/detect_c2c3.py
charleygros: Note: To re-discuss on Monday.

charleygros:
Remains todo:

- [ ] Add binary `isct_train_svm`

jcohenadad:
Not sure why Travis is failing [here](https://travis-ci.org/neuropoly/spinalcordtoolbox/jobs/464078863#L1061).
@charleygros: The binaries are supposed to be the same as before, right?

Issue #2083, opened 2018-11-26
Long centerline detection on not-so high res T1w scan


jcohenadad:
When running deepseg_sc using the default parameters, the centerline detection seems to take a fairly large amount of time to compute (about 6 min), while it is usually fast, especially with the SVM algo. 

The data are 0.6x0.6x1mm resolution (188x512x512 matrix)

## To reproduce

Data: 

Syntax:
~~~
Spinal Cord Toolbox (master/b16de3a5156459fdf13911690a9c636c1ecfef70)
Running /Users/julien/code/sct/scripts/sct_deepseg_sc.py -i sub-01_T1w.nii.gz -c t1 -qc /Users/julien/results_spine_generic
Folder /Users/julien/results_spine_generic has been created.

Method:
	Centerline algorithm: svm
	Assumes brain section included in the image: True
	Dimension of the segmentation kernel convolutions: 2d

Creating temporary folder...
cp sub-01_T1w.nii.gz /var/folders/6f/wy6ljmx9453cgth2qwv5l1s80000gn/T/sct-20181126173226.800103-0CL4_d
Reorient the image to RPI, if necessary...
Resample the image to 0.5 mm isotropic resolution...
/Users/julien/code/sct/python/lib/python2.7/site-packages/nipy/io/files.py:145: FutureWarning: Default `strict` currently False; this will change to True in a future version of nipy
  ni_img = nipy2nifti(img, data_dtype = io_dtype)
Finding the spinal cord centerline...
cp sub-01_T1w_RPI_resampled.nii.gz /var/folders/6f/wy6ljmx9453cgth2qwv5l1s80000gn/T/sct-20181126173441.543832-W9PXES
cp sub-01_T1w_RPI_resampled_centerline_optic.nii.gz /var/folders/6f/wy6ljmx9453cgth2qwv5l1s80000gn/T/sct-20181126173226.800103-0CL4_d
rm -rf /var/folders/6f/wy6ljmx9453cgth2qwv5l1s80000gn/T/sct-20181126173441.543832-W9PXES
Cropping the image around the spinal cord...
Normalizing the intensity...
Segmenting the spinal cord using deep learning on 2D patches...
Reassembling the image...
WARNING: To avoid intensity overflow due to convertion to uint8, intensity will be rescaled to the maximum quantization scale.
Resampling the segmentation to the original image resolution...
Binarizing the segmentation to avoid interpolation effects...
[...]

PR #2093, opened 2018-12-04
deepseg_sc: speed processing up

charleygros:
This PR aims at accelerating the processing of `sct_deepseg_sc`and `sct_deepseg_lesion`.

The developments are:
- [x] When using `-centerline svm` or `viewer`: perform the resampling to inplane 0.5x0.5 mm2 after the centerline detection (instead of before) --> Reduced the processing time from 6min44s to 5min23s on #2083
- [ ] When using `-kernel 2d`: perform the segmentation in one-go (with specific batch_size) instead of z_slice by z_slice.


charleygros:
Closed because of inappropriate commits.

Issue #2099, opened 2018-12-08
Detect c2c3 failing in one subject from spine_generic project


jcohenadad:
It would be nice to have the automatic c2c3 detection work for all subjects in the [Multi-center/Multi-subjects](https://osf.io/jkxzp/) database. Here, we will list all failing subjects to try to find a solution that works for all subjects:
- unf_sub-04/t1

@charleygros maybe playing with the number of slices to average in the medial plane would increase robustness at inference time?

charleygros:
Will have a look at this subject!
& investigate the number of slices to average, as you suggested. However, based on my first explorations, I do not think that it will improve it significantly: if the input segmentation is correct, then the flatten_sagittal should provide a sufficient view of the target structure.

Will keep you updated

jcohenadad:
fixed by https://github.com/neuropoly/spinalcordtoolbox/pull/2103

PR #2103, opened 2018-12-17
detect_c2c3: improve postprocessing

charleygros:
This PR aims at improving the post-processing of the module `detect_c2c3`.

As reported in the issue #2099, the algorithm fails in some cases because several discs are detected as C2/C3.

In this PR, we propose two simple post-processing operations to fix the here-above mentioned issue:
- mask the heatmap above the spinal cord segmentation mask (in the I-S direction)
- if several discs were detected, we consider that the targeted C2/C3 disc is the one located at the top of the heatmap (in the I-S direction)

To test the PR, please re-install the toolbox:
```
./install_sct -d -m -b
```

### DIFF
--- a/spinalcordtoolbox/vertebrae/detect_c2c3.py
+++ b/spinalcordtoolbox/vertebrae/detect_c2c3.py
@@ -24,6 +24,7 @@ from sct_flatten_sagittal import flatten_sagittal
 import numpy as np
 import nibabel as nib
 from scipy.ndimage.measurements import center_of_mass
+from skimage.measure import label as label_regions
 
 import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image, zeros_like
@@ -84,11 +85,26 @@ def detect_c2c3(nii_im, nii_seg, contrast, verbose=1):
 
     # mask prediction
     pred[midSlice_mask == 0] = 0
+    z_seg_max = np.max(np.where(midSlice_seg)[1])
+    pred[:, z_seg_max:] = 0  # Mask above SC segmentation
 
     # assign label to voxel
     nii_c2c3 = zeros_like(nii_seg_flat)
     if np.any(pred > 0):
         sct.printv('C2-C3 detected...', verbose)
+
+        pred_bin = (pred > 0).astype(np.int_)
+        labeled_pred, nb_regions = label_regions(pred_bin, return_num=True)
+        if nb_regions > 1:  # if there is several detected clusters of voxels
+            region_idx_top, region_z_top = 0, 0
+            for region_idx in range(1, nb_regions+1):
+                pred_idx = (labeled_pred == region_idx).astype(np.int_)
+                pa_com, is_com = center_of_mass(pred_idx)
+                if is_com >= region_z_top:
+                    region_idx_top = region_idx
+                    region_z_top = is_com
+            pred[labeled_pred != region_idx_top] = 0  # then keep the one located at the top (IS direction)
+
         coord_max = np.where(pred == np.max(pred))
         pa_c2c3, is_c2c3 = coord_max[0][0], coord_max[1][0]
         nii_seg.change_orientation('PIR')


jcohenadad:
@charleygros I tried with the latest version (74926f42fd5448e815288f43c5057d7fd4e3c79f), and still have a problem with `ucl_sub-02`. You mentioned in Slack that this subject now works fine with the latest commit, so I don't understand what could cause different behaviour between your site and my site?

Here is the [QC index](https://www.dropbox.com/s/920as08jcm1o093/201901081409_qc.zip?dl=0).

Many thanks!

charleygros:
@jcohenadad: My mistake! I was working on the T2 of `ucl_subj_02`.
Will be right back

capture d ecran 2019-01-10 01 04 17

charleygros:
@jcohenadad: The detection now seems to work on the subject/images your pointed out.
Let me know how now goes your testing

jcohenadad:
Unfortunately `unf_sub-04` is now failing :-(

Here is the [QC_index](https://www.dropbox.com/s/bsmfepfrbzqpjt7/201901091250_spineGeneric_multiSubjects_qc.zip?dl=0)

Maybe it would be more efficient if you could test the labeling for all (current) subjects of the spineGeneric database? It is accessible [here](https://osf.io/76jkx/).

To make things faster, you can create a temporary .sh file and simply remove all lines below [this one](https://github.com/sct-pipeline/spine_generic/edit/master/process_data.sh#L49).

charleygros:
I am sharing, herebelow, some thoughts about the current processing post-detection.

The model produces a heat-map (positive and negative values) on the mid-sagittal slices. The model assigns a positive value to a voxel if the targeted partern has been detected in the surroundings. The value is a metric of confidence, without unit, ie the higher --> the more confident.
A naive approach would be to keep the voxel with the maximum value. However, the model is not perfect, then we should "help it" with some anatomical prior info.

1. In a close surrounding of the spinal cord segmentation --> I create a mask around the provided segmentation and reject all the "positive voxel" which are located out of the mask.

2. A common error made by the model is to detect other disc than C2-C3. If that s the case, then I consider that the cluster of voxels of interest is the one located at the top position in the I-S direction.

However, it was not sufficient for some cases, see issue #2099. I first started to play with some thresholds on the heat-map, but, as pointed out by @jcohenadad, this is not the good approach, since highly dependent on the data, the threshold will need to be changed in the future...etc.

The issue with (2) is that the model can assign some positive values at the top of C1 (.. no idea why..).
To overcome this issue: I noticed that the segmentation of `sct_deepseg_sc` ends quite often around the bottom of the medulla on T1 images, which is quite consistent across the subjects of my mini testing dataset. Consequently, I use this anatomical prior to only consider the voxels below (30mm) the top of the segmentation.
Some limitations:
- do not take into account the curvature of the cord. TODO
- is it reproducible with an input from `sct_propseg`?
- ...

charleygros:
Commit `269b469 ` tested on the `spineGeneric_multiSubject` dataset (ucl, unf, barcelona) --> seems to work on these subjects.

jcohenadad:
> I noticed that the segmentation of sct_deepseg_sc ends quite often around the bottom of the medulla on T1 images, which is quite consistent across the subjects of my mini testing dataset. Consequently, I use this anatomical prior to only consider the voxels below (30mm) the top of the segmentation.

@charleygros i would try to avoid those kind of prior (or use them as "last resort") because as you pointed out it is too sensitive to the input segmentation. I suggest we focus on a solution that can generalize better, e.g., working on the image itself, adding more subjects (or removing non-representative subjects), doing other kind of preprocessing, etc. 
For example, looking at the [train.py](https://github.com/neuropoly/spinalcordtoolbox/blob/master/dev/detect_c2c3/train.py#L37) file, it seems like the training is done in the mid-sag slice, while for the inference, i think we discussed using the average of x sagittal slices centered in the mid-plane, hence the noise property between the trained model and the inference data is different (or maybe the inference is also done in single sag image --> if that's the case, then averaging across 5 slices could help by reducing the noise?).

charleygros:
@jcohenadad: I was wondering: Do you think that a solution would be to re-do the training using only the subjects of spineGeneric dataset? Is it relevant?

Concerning the average of sagittal slices: since the `sct_flatten_sagittal` is quite efficient (ie make the discs visible), then I did not averaged slices. For the training as well as for the inference.
It can be something that we could try. However, on my preliminary tests, I observed that it did not really help (I did with 3 slices).

jcohenadad:
> re-do the training using only the subjects of spineGeneric dataset? Is it relevant?

Hum... only the spineGeneric sounds a bit too extreme, but do you think you could easily identify some subjects that clearly stand out? Is that something that we could output from the training procedure?

Regarding the average of the slices: I think that reducing the noise would definitely help. I would definitely implement the average of 5 (or even 7) slices, both on the training and inference.

charleygros:
I could try what you proposed @jcohenadad:
- retrain the models by removing non-representative subjects
- use the average of 5 (or 7, will investigate) sagittal slices for both training and inference
--> we could then do a comparison between the 2 implementations

charleygros:
Tested strategy:
- remove non representative subjects
- use 7mm around mid sagital slice
- do not use hard coded post processing
--> retrain the models

Results on the `spineGeneric` database (5 centers): https://drive.google.com/drive/folders/1hekFXbDzZMgtN983iVVnPPg9oZm4FtFU?usp=sharing
The detection succeed on all tested subjects, except on 1 image (`sherbrooke/sub-01_T2w.nii.gz`) where nothing is detected (ie not a FP).

PR #2174, opened 2019-02-25
Faster execution and other minor improvements

### LINKED ISSUE #172: sct_warp_template
change display message at the end, in case user does not ask for atlas

### LINKED ISSUE #602: WM atlas: make it a single file for faster warping?
currently the WM atlas has 37 nifti separated files in it-- when apply sct_warp_template, each file is warped-- there is a lot of i/o which could be avoided if the atlas would be a single file.

however this is tricky because currently isct_apply_transfo (from ANTs) only accepts 3d (not 4d) files...  so, discussion is open...

### LINKED ISSUE #1528: improve help about output folder
mention in the usage the content of the output, and where it is located by default.

### LINKED ISSUE #2173: Use isct_antsApplyTransfo instead of SCT system call for faster execution
That would result in much faster execution and is an easy fix.

jcohenadad:
The main purpose of this PR was to use `isct_antsApplyTransfo` instead of SCT system call for warping each volume, resulting in faster execution. In a typical dataset, execution time goes from 4.5min to 2min. Fixes #2173. This change makes #602 less relevant, there we can consider that this PR Fixes #602.

Other improvements include:
- QC now supporting `sct_warp_template`
- Change display message at the end, in case user does not ask for atlas. Fixes #172 
- Cleanup usage. Fixes #1528

TODO:
- Add QC in batch_processing


jcohenadad:
Closing because history was messed up due to parallel merging. A cleaner PR is available here https://github.com/neuropoly/spinalcordtoolbox/pull/2175

PR #2206, opened 2019-04-01
Lock qc report during generation

### LINKED ISSUE #2205: ValueError: Extra data when generating qc report
### Description

When running a pipeline on an ssh-connected station (no tunnelling), I experienced a bug during QC report generation:

<details>

~~~
Mask type float64
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/scripts/sct_register_multimodal.py", line 812, in <module>
    main()
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/scripts/sct_register_multimodal.py", line 553, in main
    path_qc=os.path.abspath(path_qc), process='sct_register_multimodal')
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/spinalcordtoolbox/reports/qc.py", line 569, in generate_qc
    stretch_contrast_method='equalized',
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/spinalcordtoolbox/reports/qc.py", line 481, in add_entry
    layout(qcslice)
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/spinalcordtoolbox/reports/qc.py", line 275, in wrapped_f
    self.qc_report.update_description_file(img.shape)
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/spinalcordtoolbox/reports/qc.py", line 423, in update_description_file
    results = json.load(open(self.qc_params.qc_results, 'r'))
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/python/lib/python2.7/json/__init__.py", line 291, in load
    **kw)
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/python/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/home/GRAMES.POLYMTL.CA/p101317/code/sct/python/lib/python2.7/json/decoder.py", line 367, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 3346 column 2 - line 3348 column 2 (char 208880 - 208890)
~~~

</details>

### Investigations 1

- `sct_testing` --> pass
- Same dataset, running locally --> pass
- output in new qc folder --> pass
- output in old qc folder --> FAIL

After investigations (see Investigations 1) The source of this bug seems related to a corruption of the already-existing json file, which has the following text towards the end:

~~~
  {
    "orientation": "Axial", 
    "python": "/home/GRAMES.POLYMTL.CA/p101317/code/sct/python/bin/python", 
    "moddate": "2019-04-01 12:51:46", 
    "dataset": "mgh_spineGeneric", 
    "cmdline": "sct_deepseg_sc -i sub-06_acq-T1w_MTS_crop_r.nii.gz -c t1 -ofolder /home/GRAMES.POLYMTL.CA/p101317/spineGeneric_result/mgh_spineGeneric/derivatives/labels/sub-06/anat -qc /home/GRAMES.POLYMTL.CA/p101317/spineGeneric_qc", 
    "fname_in": "sub-06_acq-T1w_MTS_crop_r.nii.gz", 
    "subject": "sub-06", 
    "sct_version": "git-master-0b3d3c5a4739f909fea94c2bd4745def35648ba0", 
    "dimension": "30x600", 
    "command": "sct_deepseg_sc", 
    "background_img": "mgh_spineGeneric/sub-06/anat/sct_deepseg_sc/2019_04_01_125145.001709/bkg_img.png", 
    "overlay_img": "mgh_spineGeneric/sub-06/anat/sct_deepseg_sc/2019_04_01_125145.001709/overlay_img.png", 
    "cwd": "/home/GRAMES.POLYMTL.CA/p101317/spineGeneric_result/mgh_spineGeneric/sub-06/anat", 
    "contrast": "anat"
  }
]nat"
  }
]
~~~

Note the "]nat"" which are the cause of the problem. How was this corruption introduced needs further investigations.

It is likely that the corruption appeared while two parallel processes tried to access the same json file at the same time, given that the json was generated by a pipeline working with `parallel` framework. Interesting read:
- https://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/

### Potential solution

The culprit is [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/reports/qc.py#L425).

One possible solution would be to check if the file is being used by another process, and if it is, wait until it is closed before opening and writing in it.

npgoodhue:
Fix qc report where extra conflicting data is generated using locking mechanism with portalocker allowing only one process to use the file at once.

Fixes #2205


spinalcordtoolbox/reports/qc.py
```diff
@@ -28,6 +28,8 @@
 from spinalcordtoolbox.image import Image
 import spinalcordtoolbox.reports.slice as qcslice
 
+import portalocker
```
zougloub: This 3rd party module should be imported before the modules from this library and after importing built-in modules.

spinalcordtoolbox/reports/qc.py
```diff
@@ -422,7 +424,8 @@ def update_description_file(self, dimension):
         if os.path.isfile(self.qc_params.qc_results):
             results = json.load(open(self.qc_params.qc_results, 'r'))
```
jcohenadad: we should probably lock the file and **then** read it (instead of locking it after), otherwise content might change between this line and the line below where you write

zougloub: @npgoodhue you can open with 'r+', load, then `.seek(0)` and dump.

zougloub: My fault, I tried to give hints but didn't think this through.

npgoodhue:
I have one issue concerning the initial creation of the qc report: `portalocker` cannot work in `r+` if the file doesnt exist, but if the code checks for the existence of the file, I probably have to create it outside of the `portalocker` environment and multiple processes can access the section to create the file.

npgoodhue:
This might be useful
https://stackoverflow.com/questions/10349781/how-to-open-read-write-or-create-a-file-with-truncation-allowed

spinalcordtoolbox/reports/qc.py
```diff
@@ -419,12 +424,17 @@ def update_description_file(self, dimension):
         }
         logger.debug('Description file: %s', self.qc_params.qc_results)
         results = []
-        if os.path.isfile(self.qc_params.qc_results):
-            results = json.load(open(self.qc_params.qc_results, 'r'))
-        results.append(output)
-        json.dump(results, open(self.qc_params.qc_results, "w"), indent=2)
+        with portalocker.Lock(self.qc_params.qc_results) as lck_qc_file:
+            with touchopen(lck_qc_file.name, 'r+') as edit_lck_qc:
```
zougloub: It may work but I'm not a huge fan of this manipulation as it's opening the same file twice and makes the code complicated.
Can you encapsulate this in a prettier-looking class?

Portalocker provides portable low-level functions so we could use those.

I believe you could do something like:
```python
class open_with_exclusive_lock(object):
    """
    """
    def __init__(self, filename, mode):
        ...
    def __enter__(self):
        fd = os.open(...)
        self._f = os.fdopen(...)
        portlalocker.lock(self._f)
        return self._f
  
    def __exit__(self, exc_type, exc_val, exc_tb):
        portalocker.unlock(self._f)
```

zougloub: (and this can go in sct_utils.py)

spinalcordtoolbox/reports/qc.py
```diff
@@ -419,12 +424,17 @@ def update_description_file(self, dimension):
...
+                if os.stat(lck_qc_file.name).st_size is not 0:
```
zougloub: Use `os.path.getsize(path)` which is more usual and reads better

spinalcordtoolbox/reports/qc.py
```diff
@@ -24,13 +24,14 @@
 from matplotlib.figure import Figure
 import matplotlib.colors as color
 
+import portalocker
```
zougloub: This import is not needed anymore

spinalcordtoolbox/reports/qc.py
```diff
@@ -419,12 +420,16 @@ def update_description_file(self, dimension):
         }
         logger.debug('Description file: %s', self.qc_params.qc_results)
         results = []
-        if os.path.isfile(self.qc_params.qc_results):
-            results = json.load(open(self.qc_params.qc_results, 'r'))
-        results.append(output)
-        json.dump(results, open(self.qc_params.qc_results, "w"), indent=2)
+        with sct.open_with_exclusive_lock(self.qc_params.qc_results, 'r+') as lck_qc_file:
+            if os.path.getsize(self.qc_params.qc_results) is not 0:
```
zougloub: It turns out that cpython is interning small integers, but it's not a good idea to use `is` for anything but `None, True, False, Ellipsis`.

scripts/sct_utils.py
```diff
@@ -1006,7 +1007,7 @@ def check_if_same_space(fname_1, fname_2):
 
 
 def printv(string, verbose=1, type='normal'):
-    """enables to print (color coded messages, depending on verbose status) 
+    """enables to print (color coded messages, depending on verbose status)
```
zougloub: Not a big deal, but you can see if you check the changes here, that you modified lines unrelated to the scope of the issue. You should have only committed what you intended to modify.

npgoodhue: Apologies for that mistake. I will keep that in mind for future commits.

npgoodhue:
From what I understand, the `portalocker.lock()` does not exactly lock the file. It still allows more than on process to access the file at the same time, but the last process that finishes overwrites the content written by the other processes (I did some basic testing and noticed that). The previous implementation where the file was opened twice with `portalocker.Lock()` followed by `touchopen()` (even though it's probably not the most efficient way) allowed to keep information from both processes.
I am not sure what is the most important, but wanted to point that out.

jcohenadad:
thanks for the clarification @npgoodhue, what is most important is that the information from all processes need to be saved. Maybe it would be a good idea to write a short unit testing to make sure the function does what it is expected to do?

npgoodhue:
My mistake, the file is overwritten only if it has not been created yet and more than one process access it at the same time with `portalocker.lock()`. I am not an expert, but unit testing would probably be useful. Is that the creation of another issue?

jcohenadad:
> Is that the creation of another issue?

Yes, I don't want to drag this PR too long

PR #2216, opened 2019-04-08
New API to generate QC reports

### LINKED ISSUE #2187: Make it possible to build the qc folder given existing output
It would be quite useful in some situations. 

Also requested by a user: http://forum.spinalcordmri.org/t/build-qc-folder-from-existing-data/25

### LINKED ISSUE #2197: Make it possible to specify "Dataset" and "Subject" as input to qc
In case dataset does not respect BIDS convention.

jcohenadad:
The main purpose of this PR is to provide an API (`sct_qc`) for generating QC reports a posteriori (i.e., not *only* after running SCT). Specific changes include:
- Made it possible to build the QC folder given existing output, via new API `sct_qc`. Fixes #2187.
- Removed QC report from `sct_straighten_spinalcord` (not relevant, can use verbose=2 instead)
- Made it possible to specify "Dataset" and "Subject" as input to QC. Fixes #2197

To test this PR:
SCT needs to link new binaries, hence run the following before testing:
~~~
./install_sct -d -b -m
~~~


spinalcordtoolbox/reports/qc.py
```diff
@@ -520,50 +527,55 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, args=None, path_qc=No
...
     elif process in ['sct_detect_pmj']:
-        # sagittal orientation, display PMJ box
         plane = 'Sagittal'
         qcslice_type = qcslice.Sagittal([Image(fname_in1), Image(fname_seg)], p_resample=None)
         qcslice_operations = [QcImage.highlight_pmj]
         qcslice_layout = lambda x: x.single()
     else:
-        sct.log.error('Unrecognised process.')
+        sct.log.error('Unrecognized process.')
+        raise ValueError
```
zougloub: It's nicer to provide something more precise than just an exception type:
```
raise ValueError("Unrecognized process {}".format(process))
```

scripts/sct_deepseg_gm.py
```diff
@@ -44,6 +44,14 @@ def get_parser():
                       description="The path where the quality control generated "
                                   "content will be saved",
                       default_value=None)
+    parser.add_option(name='-qc-dataset',
+                      type_value='str',
+                      description='Name of the dataset to display in the QC report. By default, assumes BIDS.',
```
zougloub: I think there could be a clearer and more concise way to express this, some questions can arise from this wording:

- Are we displaying the dataset in the QC report?
- Are we "displaying" the name of the dataset?
- Is the string "BIDS" the default name of the dataset?

I'd use the word "mention" instead of "display". 
Since the parameter is optional, I'd use "If provided, this string will be mentioned in the QC report as the dataset the process was run on", or something like that.

spinalcordtoolbox/reports/qc.py
jcohenadad: fixed in e127bc062b64366aa26b2b8b57699a5fce9d8fb8

scripts/sct_deepseg_gm.py
jcohenadad: fixed in 1b7663634d88bbf00461c1b580a0c9b47b544c4e

PR #2250, opened 2019-05-03
Fixed corruption of QC json file when running parallel jobs

jcohenadad:
When running several subjects in parallel (using GNU parallel), during QC generation, a `JSONDecodeError` is observed. This error was caused by parallel writing on the json file of the QC report. It was supposed to be fixed by the use of `portalocker` (see https://github.com/neuropoly/spinalcordtoolbox/issues/2205), but in some file systems (OSX, Ubuntu), it is not. 

Another (unsuccessful) locking strategy was tried: create a temporary locked file, and then copy the content in the json file. However, it also failed in some OSs (OSX, Ubuntu).

The fix was to create an individual json file for each instance, instead of appending all the QC metadata within the same json file. 

Fixes https://github.com/neuropoly/spinalcordtoolbox/issues/2248

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,6 @@ Keras
 matplotlib
 nibabel
 pandas
-portalocker
 psutil
 pyqt5
 raven
--- a/scripts/sct_utils.py
+++ b/scripts/sct_utils.py
@@ -1315,22 +1315,3 @@ def cache_save(cachefile, sig):
     with io.open(cachefile, "wb") as f:
         f.write(sig)
 
-class open_with_exclusive_lock(object):
-    """
-    Utility class to prevent the writing of a file by multiple processes.
-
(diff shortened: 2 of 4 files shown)


jcohenadad:
Bad news: when testing this PR (2ade14c4489fbcb0d841297346e2612d856677b1) in the same dataset that produced the error described in this PR, I observe the same error (not systematic, and appears randomly across datasets).

jcohenadad:
bad news: when running 2ade14c4489fbcb0d841297346e2612d856677b1 on the same station that produced the error described in this PR (joplin), the same error is observed.

jcohenadad:
re-bad news: with the very-dirty locking (cd2f95301af9c01bd2240d4fdec7ddabc78aa564), error is still there

PR #2264, opened 2019-05-15
Added output file and/or folder flag

### LINKED ISSUE #2178: sct_compute_mtr missing output file or folder flag
**Feature justification: Is your feature request related to a problem? Please describe.**
The MTR script ([sct_compute_mtr.py](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_compute_mtr.py)) doesn't have a flag to either set the output filename and location ([it defaults to mtr.nii.gz in the active directory](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_compute_mtr.py#L85 )) like MTsat does ([see here](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_compute_mtsat.py#L64-L66)), or a flag to change the output directory of the file mtr.nii.gz (other functions have such a flag, like [here](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_deepseg_sc.py#L47-L52))

Either of the two CLI flags solutions that I've pointed to above, maybe the first?

**Describe alternatives you've considered**
N/A

YangHee-Min:
At the moment, the script [sct_compute_mtr.py](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_compute_mtr.py) does not have a flag to change the output filename or its location ([it would default to mtr.nii.gz in the active directory](https://github.com/neuropoly/spinalcordtoolbox/blob/53b81c4b73fa65ad55ac391649ecb44744c077ad/scripts/sct_compute_mtr.py#L85)).

This PR makes it possible to change the output file name and/or location with the flag `-o`. Fixes #2178 

To do so, the user simply needs to follow the following format when adding the `-o` flag to the rest of the [sct_compute_mtr](https://sourceforge.net/p/spinalcordtoolbox/wiki/sct_compute_mtr/) command:
`-o /path_Leading_To_Output_File/output_File`

Example use with output file name changed to "custom_name" but location (default) unchanged:
```
sct_compute_mtr -mt0 mt0.nii.gz -mt1 mt1.nii.gz -r 0 -v 0 -o /Users/heyan_admin/sct_course-london19/single_subject/data/mt/custom_name
```
Example use with output file location changed to the user's "Desktop"  folder but name ("mtr") unchanged: 

```
sct_compute_mtr -mt0 mt0.nii.gz -mt1 mt1.nii.gz -r 1 -v 0 -o /Users/heyan_admin/Desktop/mtr
```
Example use with output file location changed to the user's "Downloads" folder and name changed to "testing":
````
sct_compute_mtr -mt0 mt0.nii.gz -mt1 mt1.nii.gz -r 0 -v 2 -o /Users/heyan_admin/Downloads/testing

### DIFF
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -41,7 +41,7 @@ def main(args=None):
     # Initialization
     fname_mt0 = ''
     fname_mt1 = ''
-    file_out = param.file_out
+    fname_mtr = ''
     # register = param.register
     # remove_temp_files = param.remove_temp_files
     # verbose = param.verbose
@@ -57,8 +57,6 @@ def main(args=None):
     fname_mt0 = arguments['-mt0']
     fname_mt1 = arguments['-mt1']
     fname_mtr = arguments['-o']
-    file_out = os.path.basename(fname_mtr)[:os.path.basename(fname_mtr).rfind('.nii')]
-    ext_out = os.path.basename(fname_mtr)[os.path.basename(fname_mtr).rfind('.nii'):]
     verbose = int(arguments.get('-v'))
     sct.init_sct(log_level=verbose, update=True)  # Update log level
 
@@ -72,7 +70,7 @@ def main(args=None):
     nii_mtr = nii_mt1
     nii_mtr.data = data_mtr
     nii_mtr.save(fname_mtr)
-    # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 file_out.nii', verbose)
+    # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 fname_mtr', verbose)
 
     # if changing output file name or location, create folder with mt0.nii.gz and mt1.nii.gz files at precised location
     if os.path.dirname(fname_mtr) != '.':
@@ -84,9 +82,9 @@ def main(args=None):
 
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
-    sct.generate_output_file(os.path.join('.', file_out + ext_out), '.')
+    sct.generate_output_file(os.path.join('.', os.path.basename(fname_mtr)), '.')
 
-    sct.display_viewer_syntax([fname_mt0, fname_mt1, file_out + ext_out])
+    sct.display_viewer_syntax([fname_mt0, fname_mt1, os.path.basename(fname_mtr)])
 
 
 # ==========================================================================================
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -63,22 +63,6 @@ def main(args=None):
     verbose = int(arguments.get('-v'))
     sct.init_sct(log_level=verbose, update=True)  # Update log level
 
-    # Copying input data to output folder and convert to nii
-    sct.printv('\nCopying input data to output folder and convert to nii...', verbose)
-    from sct_convert import convert
-    convert(fname_mt0, os.path.join(os.path.dirname(fname_mtr), "mt0.nii"), dtype=np.float32)
-    convert(fname_mt1, os.path.join(os.path.dirname(fname_mtr), "mt1.nii"), dtype=np.float32)
-
-    # if changing output file name or location, create folder with mt0 and mt1 files at precised location
-    if os.path.dirname(fname_mtr) != '.':
-        startdir = os.getcwd()
-        os.chdir(os.path.dirname(fname_mtr))
-        shutil.copy(os.path.join(startdir, fname_mt0), os.path.dirname(fname_mtr))
-        shutil.copy(os.path.join(startdir, fname_mt1), os.path.dirname(fname_mtr))
-
-    # go to output file directory
-    os.chdir(os.path.dirname(fname_mtr))
-
     # compute MTR
     sct.printv('\nCompute MTR...', verbose)
     nii_mt1 = msct_image.Image('mt1.nii')
@@ -88,8 +72,16 @@ def main(args=None):
     # save MTR file
     nii_mtr = nii_mt1
     nii_mtr.data = data_mtr
-    nii_mtr.save(file_out + ext_out)
+    nii_mtr.save(fname_mtr)
     # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 file_out.nii', verbose)
+
+    # if changing output file name or location, create folder with mt0.nii.gz and mt1.nii.gz files at precised location
+    if os.path.dirname(fname_mtr) != '.':
+        shutil.copy(os.path.join('.', fname_mt0), os.path.dirname(fname_mtr))
+        shutil.copy(os.path.join('.', fname_mt1), os.path.dirname(fname_mtr))
+
+    # go to output file directory
+    os.chdir(os.path.dirname(fname_mtr))
 
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -72,9 +72,6 @@ def main(args=None):
     nii_mtr.save(fname_mtr)
     # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 fname_mtr', verbose)
 
-    # go to output file directory
-    os.chdir(os.path.dirname(fname_mtr))
-
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
     sct.generate_output_file(os.path.join('.', fname_mtr), '.')
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -15,7 +15,6 @@ from __future__ import absolute_import, division
 
 import sys
 import os
-import shutil
 
 import sct_utils as sct
 from msct_parser import Parser
@@ -65,9 +64,9 @@ def main(args=None):
 
     # compute MTR
     sct.printv('\nCompute MTR...', verbose)
-    nii_mt1 = msct_image.Image('mt1.nii')
+    nii_mt1 = msct_image.Image(fname_mt1)
     data_mt1 = nii_mt1.data
-    data_mt0 = msct_image.Image('mt0.nii').data
+    data_mt0 = msct_image.Image(fname_mt0).data
     data_mtr = 100 * (data_mt0 - data_mt1) / data_mt0
     # save MTR file
     nii_mtr = nii_mt1
@@ -77,8 +76,8 @@ def main(args=None):
 
     # if changing output file name or location, create folder with mt0.nii.gz and mt1.nii.gz files at precised location
     if os.path.dirname(fname_mtr) != '.':
-        shutil.copy(os.path.join('.', fname_mt0), os.path.dirname(fname_mtr))
-        shutil.copy(os.path.join('.', fname_mt1), os.path.dirname(fname_mtr))
+        sct.copy(os.path.join('.', fname_mt0), os.path.dirname(fname_mtr))
+        sct.copy(os.path.join('.', fname_mt1), os.path.dirname(fname_mtr))
 
     # go to output file directory
     os.chdir(os.path.dirname(fname_mtr))
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -157,7 +157,7 @@ def get_parser():
                       default_value='1')
     parser.add_option(name="-o",
                       type_value="str",
-                      description="Creates output file with the specified path.",
+                      description="Creates output file with the specified path. Add extension '.nii.gz' to output file name.",
                       mandatory=False,
                       example='My_File_Folder/My_New_File',
                       default_value=os.path.join(os.getcwd(),'mtr'))
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -104,7 +104,7 @@ def main(args=None):
 
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
-    sct.generate_output_file(os.path.join(path_tmp, file_out + ".nii"), os.path.join(path_out, file_out + ext_out))
+    sct.generate_output_file(os.path.join(path_tmp, file_out + ".nii"), os.path.join(path_out, file_out))
 
     # Remove temporary files
     if remove_temp_files == 1:
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -100,12 +100,6 @@ def get_parser():
                       description="Image with MT pulse (MT1)",
                       mandatory=False,
                       deprecated_by="-mt1")
-    parser.add_option(name="-r",
-                      type_value="multiple_choice",
-                      description='Remove temporary files.',
-                      mandatory=False,
-                      default_value='1',
-                      example=['0', '1'])
     parser.add_option(name="-v",
                       type_value='multiple_choice',
                       description="verbose: 0 = nothing, 1 = classic, 2 = expended",
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -15,6 +15,7 @@ from __future__ import absolute_import, division
 
 import sys
 import os
+import shutil
 
 import sct_utils as sct
 from msct_parser import Parser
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/sct_compute_mtr.py [hunk 1]
```diff
@@ -141,6 +155,12 @@ def get_parser():
                       mandatory=False,
                       example=['0', '1', '2'],
                       default_value='1')
+    parser.add_option(name="-omtr",
```
jcohenadad: i would make it `-o` as per [our convention](https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#flags)

scripts/sct_compute_mtr.py
```diff
@@ -141,6 +155,12 @@ def get_parser():
...
+                      type_value="str",
+                      description="Creates output file with the specified path.",
```
jcohenadad: i would specify to add the extension

scripts/sct_compute_mtr.py
```diff
@@ -141,6 +155,12 @@ def get_parser():
...
+                      mandatory=False,
+                      example='My_File_Folder/My_New_File',
```
jcohenadad: give more concrete example, e.g.: `bob/data/mtr.nii.gz`

scripts/sct_compute_mtr.py
```diff
@@ -141,6 +155,12 @@ def get_parser():
...
+                      default_value=os.path.join(os.getcwd(),'mtr'))
```
jcohenadad: add extension `.nii.gz`

scripts/sct_compute_mtr.py
```diff
@@ -86,21 +96,25 @@ def main(args=None):
...
+    # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 file_out.nii', verbose)
 
     # come back
-    os.chdir(curdir)
+    os.chdir(os.path.split(fname_mtr)[0])
 
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
-    sct.generate_output_file(os.path.join(path_tmp, "mtr.nii"), os.path.join(path_out, file_out + ext_out))
+    sct.generate_output_file(os.path.join(path_tmp, file_out + ".nii"), os.path.join(path_out, file_out + ext_out))
```
jcohenadad: extension should be specified by user (what if user wants nii.gz)

scripts/sct_compute_mtr.py
```diff
@@ -86,21 +96,25 @@ def main(args=None):
...
 
     # Remove temporary files
     if remove_temp_files == 1:
         sct.printv('\nRemove temporary files...')
         sct.rmtree(path_tmp)
 
+    # if output file location changed, notify user to move to file location
```
jcohenadad: is that really necessary?

scripts/sct_compute_mtr.py
```diff
@@ -73,8 +75,16 @@ def main(args=None):
     convert(fname_mt0, os.path.join(path_tmp, "mt0.nii"), dtype=np.float32)
     convert(fname_mt1, os.path.join(path_tmp, "mt1.nii"), dtype=np.float32)
 
-    # go to tmp folder
+    # if changing output file name or location, create folder with mt0 and mt1 files at precised location
     curdir = os.getcwd()
+    if os.path.split(fname_mtr)[0] != curdir:
+        import shutil
```
jcohenadad: move imports to the top

scripts/sct_compute_mtr.py
```diff
@@ -73,8 +75,16 @@ def main(args=None):
     convert(fname_mt0, os.path.join(path_tmp, "mt0.nii"), dtype=np.float32)
     convert(fname_mt1, os.path.join(path_tmp, "mt1.nii"), dtype=np.float32)
 
-    # go to tmp folder
```
jcohenadad: actually, you can remove the whole part where we copy files to temp folder. We can simply open the data in the RAM, do the computation of mtr and then save to the specified output.

jcohenadad:
@YangHee-Min pls answer each comment and refer to the commit in the comment directly (e.g.: "agreed, fixed in COMMIT_NUMBER", or "I disagree because blablabla)
otherwise it is difficult to understand which comment you addressed, and in which commit

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [f37a7f1](https://github.com/YangHee-Min/spinalcordtoolbox/commit/f37a7f16c3093a2fdfb7b313da3396aacb4683d9)

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [d3d4c22](https://github.com/YangHee-Min/spinalcordtoolbox/commit/d3d4c2292bee56e8d76e3fb201c3ec3ea770c0f5). 
However, is the extension ".nii.gz" the only one possible? If not, the description will be slightly modified.

jcohenadad: it is not. It could also be `.nii`

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [5052b25](https://github.com/YangHee-Min/spinalcordtoolbox/commit/5052b25c7044ccd53eb796f4af44dbafeff86f85)

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [3b64fb9](https://github.com/YangHee-Min/spinalcordtoolbox/commit/3b64fb92ce0377e63dca16e522411af0ec71ab38)

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [b07b98d](https://github.com/YangHee-Min/spinalcordtoolbox/commit/b07b98da34920eea9d8d44ad0e8f74552da21aa5)

scripts/sct_compute_mtr.py
YangHee-Min: Understood, fixed in commit [bc49853](https://github.com/YangHee-Min/spinalcordtoolbox/commit/bc49853d811f41636125c67d4e8624438724d0e5)

scripts/sct_compute_mtr.py
YangHee-Min: agreed, fixed in commit [16baeaf](https://github.com/YangHee-Min/spinalcordtoolbox/commit/16baeaf5c88a63412b5adca430da010a74fbd07a)

scripts/sct_compute_mtr.py
YangHee-Min: Understood, fixed in commit [10da873](https://github.com/YangHee-Min/spinalcordtoolbox/commit/10da873575e77cd1af81520931e3769ccb57c5fa)

scripts/sct_compute_mtr.py
YangHee-Min: if we do this, should we also remove the 'r' argument [(argument shown here)](https://github.com/YangHee-Min/spinalcordtoolbox/blob/10da873575e77cd1af81520931e3769ccb57c5fa/scripts/sct_compute_mtr.py#L142-L147)? According to my understanding, this argument is used to decide whether or not to delete the temporary files and would therefore be unnecessary if the temp folder does not exist to begin with (since we are removing that process).

jcohenadad:
@YangHee-Min no need to add " (fixes #2178)" to all of your commits, because not all your commits actually fix 2178.

scripts/sct_compute_mtr.py
YangHee-Min: Changes to temp folder steps were made in commit [15f6de8](https://github.com/YangHee-Min/spinalcordtoolbox/commit/15f6de8d9e4cf1505687ead6e3208490f62dbff4)

YangHee-Min: When we open the computation, we copy mt0.nii and mt1.nii in the wanted file, but these files are not necessary to open the mtr file. Since these files are created (or overwritten), should we delete these files at the end of the script to save space?

jcohenadad: you don't need to copy the mt1/mt0 files. Just open them, compute mtr, and save nifti file, that's all

YangHee-Min: Ok understood! Steps removed in commit [e380997](https://github.com/YangHee-Min/spinalcordtoolbox/commit/e380997f44c0fd31c193a6c92250f7ed528b8e07)

scripts/sct_compute_mtr.py
jcohenadad: update description: https://github.com/neuropoly/spinalcordtoolbox/pull/2264#issue-279247920

scripts/sct_compute_mtr.py
YangHee-Min: Correct me if I am wrong, but since opening .nii.gz files creates the corresponding .nii file, wouldn't calling sct_compute_mtr multiple times cause the creation of multiple copies of the .nii file? (thus needing a way to delete the .nii file)

jcohenadad: To open an image use spinalcordtoolbox.image.Image, not convert.

YangHee-Min: Ok makes sense! Changes applied in commit [6d30511](https://github.com/YangHee-Min/spinalcordtoolbox/commit/6d30511f684581e32fa533079316134638a7bc24)

scripts/sct_compute_mtr.py
```diff
@@ -56,52 +56,37 @@ def main(args=None):
...
 
-    # come back
-    os.chdir(curdir)
+    # if changing output file name or location, create folder with mt0.nii.gz and mt1.nii.gz files at precised location
+    if os.path.dirname(fname_mtr) != '.':
+        sct.copy(os.path.join('.', fname_mt0), os.path.dirname(fname_mtr))
+        sct.copy(os.path.join('.', fname_mt1), os.path.dirname(fname_mtr))
+
+    # go to output file directory
+    os.chdir(os.path.dirname(fname_mtr))
```
jcohenadad: you need to remove because you don't use temp folder anymore

scripts/sct_compute_mtr.py
```diff
@@ -56,52 +56,37 @@ def main(args=None):
 
     fname_mt0 = arguments['-mt0']
     fname_mt1 = arguments['-mt1']
-    remove_temp_files = int(arguments['-r'])
+    fname_mtr = arguments['-o']
+    file_out = os.path.basename(fname_mtr)[:os.path.basename(fname_mtr).rfind('.nii')]
```
jcohenadad: why do you need file_out and ext_out? just use fname_mtr to write your output file

scripts/sct_compute_mtr.py
```diff
@@ -56,52 +56,37 @@ def main(args=None):
...
     nii_mtr = nii_mt1
     nii_mtr.data = data_mtr
-    nii_mtr.save("mtr.nii")
-    # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 mtr.nii', verbose)
+    nii_mtr.save(fname_mtr)
+    # sct.run(fsloutput+'fslmaths -dt double mt0.nii -sub mt1.nii -mul 100 -div mt0.nii -thr 0 -uthr 100 file_out.nii', verbose)
 
-    # come back
-    os.chdir(curdir)
+    # if changing output file name or location, create folder with mt0.nii.gz and mt1.nii.gz files at precised location
```
jcohenadad: why do you need that?

YangHee-Min: if we don't copy the mt0 and mt1 files, the command to visualize the output file with fsleyes does not work

scripts/sct_compute_mtr.py
```diff
@@ -56,52 +56,37 @@ def main(args=None):
...
-    sct.generate_output_file(os.path.join(path_tmp, "mtr.nii"), os.path.join(path_out, file_out + ext_out))
-
-    # Remove temporary files
-    if remove_temp_files == 1:
-        sct.printv('\nRemove temporary files...')
-        sct.rmtree(path_tmp)
+    sct.generate_output_file(os.path.join('.', file_out + ext_out), '.')
 
-    sct.display_viewer_syntax([fname_mt0, fname_mt1, file_out])
+    sct.display_viewer_syntax([fname_mt0, fname_mt1, file_out + ext_out])
```
jcohenadad: just use fname_mtr

scripts/sct_compute_mtr.py
jcohenadad: > the command to visualize the output file with fsleyes does not work

why doesn't it work? the files are physically there

YangHee-Min: The command works when outputting the file into the default folder (the same one that physically holds the mt1 and mt0 files) but if changing the output file location (i.e. Desktop) and the mt1 and mt0 files are not there, you get the following terminal log:
```
$ fsleyes mt0.nii.gz mt1.nii.gz mtr.nii.gz &
[1] 3362
$ unrecognized arguments: mt0.nii.gz mt1.nii.gz
FSLeyes version 0.26.1
Usage: fsleyes [options] file [displayOpts] file [displayOpts] ...
```

jcohenadad: when the user launches `sct_compute_mtr`, the output location of `fname_mtr` is relative to where the user is located when launching the command.

scripts/sct_compute_mtr.py
YangHee-Min: ok got it! Fixed in commit [f49fe06](https://github.com/YangHee-Min/spinalcordtoolbox/commit/f49fe066999249401f2ca4e9692792f8472b72c3)

scripts/sct_compute_mtr.py
YangHee-Min: I revised the script and the problem was in the way I wrote the code. With the way I wrote it, the command could not be issued from the folder the user is located. However, with your latest suggestion (changing file_out + ext_out to fname_mtr), the command can now be issued from the current folder.

Changes can be viewed in commit [25defd6](https://github.com/YangHee-Min/spinalcordtoolbox/commit/25defd618e92c464100eee81949aafddf4f8cbfd)

scripts/sct_compute_mtr.py
YangHee-Min: Understood, changes in commit [25defd6](https://github.com/YangHee-Min/spinalcordtoolbox/commit/25defd618e92c464100eee81949aafddf4f8cbfd)

scripts/sct_compute_mtr.py
```diff
@@ -72,19 +72,14 @@ def main(args=None):
...
-        sct.copy(os.path.join('.', fname_mt0), os.path.dirname(fname_mtr))
-        sct.copy(os.path.join('.', fname_mt1), os.path.dirname(fname_mtr))
-
     # go to output file directory
     os.chdir(os.path.dirname(fname_mtr))
 
     # Generate output files
     sct.printv('\nGenerate output files...', verbose)
-    sct.generate_output_file(os.path.join('.', os.path.basename(fname_mtr)), '.')
+    sct.generate_output_file(os.path.join('.', fname_mtr), '.')
```
jcohenadad: why not simply removing `generate_output_file`? (it has no use because the file is already where it should be)

scripts/sct_compute_mtr.py
YangHee-Min: Understood, changes in commit [92bfe7c](https://github.com/YangHee-Min/spinalcordtoolbox/commit/92bfe7c4d49babffbcb5a0a47b7921f603213d62)

scripts/sct_compute_mtr.py
YangHee-Min: That would make sense. Changes applied in commit [e6ecad8](https://github.com/YangHee-Min/spinalcordtoolbox/commit/e6ecad88dc7961055d37b44479d1e598aea5b4ef)

scripts/sct_compute_mtr.py
```diff
@@ -56,52 +56,23 @@ def main(args=None):
 
     fname_mt0 = arguments['-mt0']
     fname_mt1 = arguments['-mt1']
-    remove_temp_files = int(arguments['-r'])
```
jcohenadad: you also need to remove the entry in `get_parser()`

YangHee-Min: Agreed, fixed in commit [4b41af9](https://github.com/YangHee-Min/spinalcordtoolbox/commit/4b41af9fae9256d27c4ad7ae7d31b11a99291d9b)

jcohenadad:
@YangHee-Min update your description: https://github.com/neuropoly/spinalcordtoolbox/pull/2264#issue-279247920
(no more -r flag)

PR #2276, opened 2019-05-23
Fixed Keras/Tensorflow compatibility with CentOS 7 by downgrading to Python 3.6

jcohenadad:
As of today, Tensorflow is not compatible with Python 3.7 on Centos7 (see: tensorflow/tensorflow#26826).

In this PR, we downgrade SCT's Python from 3.7 to 3.6 to ensure cross-platform compatibility. In order to do this, we modified `install_sct` as follows: after miniconda install, a virtual environment is created with py3.6, and all SCT external and internal dependencies are installed within this venv. 

This approach offers more flexibility in the future, where similar issue could happen. That way, we will have the possibility to specify a Python version for SCT.

As of now, in order to activate SCT's python, the following should be done:
```bash
conda activate venv_sct
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ before_script:
     ls -lA bin  # Make sure all binaries and aliases are there
     sct_download_data -d sct_testing_data  # for tests
     source python/etc/profile.d/conda.sh  # to be able to call conda
-    conda activate py36  # reactivate conda for the pip install below
+    conda activate venv_sct  # reactivate conda for the pip install below
 
 script:
   - |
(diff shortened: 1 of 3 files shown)


jcohenadad:
batch_processing results (Ubuntu 16.04):
~~~
t2/CSA:          78.54000458355117
mt/MTR(WM):      54.04598798619651
t2s/CSA_GM:      12.57431836855896
t2s/CSA_WM:      65.41108245601296
dmri/FA(CST_r):  0.7810107054015337
dmri/FA(CST_l):  0.764457983509693
~~~

PR #2283, opened 2019-05-28
Added possibility to remove or keep labels by specifying values

### LINKED ISSUE #2247: Add possibility to keep or remove labels by specifying their values
Example:

This would only keep labels 2 and 5:
~~~
sct_label_utils -i IMAGE -keep 2,5
~~~

This would remove labels 2 and 5:
~~~
sct_label_utils -i IMAGE -remove 2,5
~~~

YangHee-Min:
At the moment, the function `sct_label_utils` did not have a way to remove or keep specific labels based on their values. 

This PR introduces a feature that allows one to remove one or many specific labels with the `-remove-specific` flag and to keep labels from a reference image with the `-keep-specific` flag when using the sct_label_utils command. (fixes #2247)

Example use of removing labels of value 1,2,3 from file `t2_label.nii.gz`:
```
sct_label_utils -i t2_label.nii.gz -remove-specific 1,2,3
```
Example use of keeping labels of value 7,11 from file `labeled_disc.nii`:
```
sct_label_utils -i labeled_disc.nii -keep-specific 7,11 
```

### DIFF
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -604,13 +604,9 @@ class ProcessLabels(object):
 
     def remove_or_keep_labels(self, labels, action):
         """
-        Remove or keep specific labels from a reference image by specifying their values.
-        :param labels: List of all the labels to be kept or removed
-        :type labels: list[int]
-        :param str action: Action to be performed by function, can be 'remove' or 'keep'
-        :return: image_output: Image with labels.
-        :raises ValueError: if certain elements within labels are not in reference image
-        :raises TypeError: if labels are not of type int
+        Create or remove labels from self.image_input
+        :param list(int): Labels to add or remove
+        :param str: 'remove': remove labels (i.e. set to zero), 'keep': keep labels
         """
         image_output = msct_image.zeros_like(self.image_input) if action == 'keep' else  self.image_input.copy()
         coordinates_input = self.image_input.getNonZeroCoordinates()
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -125,10 +125,8 @@ class ProcessLabels(object):
             self.output_image = self.continuous_vertebral_levels()
         if type_process == 'create-viewer':
             self.output_image = self.launch_sagittal_viewer(self.value)
-        if type_process == 'remove':
-            self.output_image = self.remove_or_keep_labels(self.value, action='remove')
-        if type_process == 'keep':
-            self.output_image = self.remove_or_keep_labels(self.value, action='keep')
+        if type_process in ['remove', 'keep']:
+            self.output_image = self.remove_or_keep_labels(self.value, action=type_process)
 
         if self.fname_output is not None:
             if change_orientation:
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -98,7 +98,7 @@ class ProcessLabels(object):
         if type_process == 'MSE':
             self.MSE()
             self.fname_output = None
-        if type_process == 'remove':
+        if type_process == 'remove-reference':
             self.output_image = self.remove_label()
         if type_process == 'remove-symm':
             self.output_image = self.remove_label(symmetry=True)
@@ -125,7 +125,7 @@ class ProcessLabels(object):
             self.output_image = self.continuous_vertebral_levels()
         if type_process == 'create-viewer':
             self.output_image = self.launch_sagittal_viewer(self.value)
-        if type_process == 'remove-reference':
+        if type_process == 'remove':
             self.output_image = self.remove_labels(self.value)
         if type_process == 'keep':
             self.output_image = self.remove_labels(self.value,isKeep = True)
@@ -685,7 +685,7 @@ def get_parser():
                       type_value='file',
                       description='Compute Mean Square Error between labels from input and reference image. Specify reference image here.',
                       mandatory=False)
-    parser.add_option(name='-remove',
+    parser.add_option(name='-remove-reference',
                       type_value='file',
                       description='Remove labels from input image (-i) that are not in reference image (specified here).',
                       mandatory=False)
@@ -693,7 +693,7 @@ def get_parser():
                       type_value='file',
                       description='Remove labels from input image (-i) and reference image (specified here) that don\'t match. You must provide two output names separated by ",".',
                       mandatory=False)
-    parser.add_option(name='-remove-reference',
+    parser.add_option(name='-remove',
                       type_value=[[','], 'int'],
                       description='Remove labels of specific value (specified here) from reference image',
                       mandatory=False)
@@ -773,18 +773,18 @@ def main(args=None):
     elif '-MSE' in arguments:
         process_type = 'MSE'
         input_fname_ref = arguments['-r']
-    elif '-remove' in arguments:
-        process_type = 'remove'
-        input_fname_ref = arguments['-remove']
+    elif '-remove-reference' in arguments:
+        process_type = 'remove-reference'
+        input_fname_ref = arguments['-remove-reference']
     elif '-remove-symm' in arguments:
         process_type = 'remove-symm'
         input_fname_ref = arguments['-r']
     elif '-create-viewer' in arguments:
         process_type = 'create-viewer'
         value = arguments['-create-viewer']
-    elif '-remove-reference' in arguments:
-        process_type = 'remove-reference'
-        value = arguments['-remove-reference']
+    elif '-remove' in arguments:
+        process_type = 'remove'
+        value = arguments['-remove']
     elif '-keep' in arguments:
         process_type = 'keep'
         value = arguments['-keep']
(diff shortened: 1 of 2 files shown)


scripts/sct_label_utils.py
```diff
@@ -125,6 +125,10 @@ def process(self, type_process):
             self.output_image = self.continuous_vertebral_levels()
         if type_process == 'create-viewer':
             self.output_image = self.launch_sagittal_viewer(self.value)
+        if type_process == 'remove-specific':
```
jcohenadad: why not simply `remove` and `keep` for flags?

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,29 @@ def launch_sagittal_viewer(self, labels):
 
         return output
 
+    def remove_specific_labels(self, labels, isKeep = False):
```
jcohenadad: same here, why not `remove_labels` for function name?

scripts/sct_label_utils.py
YangHee-Min: The `remove` flag is already used as shown [here](https://github.com/neuropoly/spinalcordtoolbox/blob/4f22bd4176428763397840ce407fc0df74342e6d/scripts/sct_label_utils.py#L742) so the `-specific` part is simply for consistency between the two similar yet different flags.

YangHee-Min: Would something such as `-delete` instead be preferable?

jcohenadad: ah! i see. I would suggest to rename this flag `remove-reference`, which I find more intuitive (although it will break retro-compatibility, but this feature is rarely used anyways, so no big deal)

jcohenadad: > Would something such as -delete instead be preferable?

no, let's keep the same terminology, but as suggested above, let's make it a `remove` and a `remove-reference`

scripts/sct_label_utils.py
YangHee-Min: Sounds good! Changed in commit [59747a7](https://github.com/neuropoly/spinalcordtoolbox/commit/59747a751482a08228783fec1cbc9623d988a3d9)

scripts/sct_label_utils.py
YangHee-Min: If we use `remove-reference`, should we keep `keep-specific` or change it to `keep-reference`?

YangHee-Min: changes for `remove-reference` applied in commit [c158f65](https://github.com/neuropoly/spinalcordtoolbox/commit/c158f655d45706726d0e3173caf81ced17b1ac84)

jcohenadad: as mentioned here: https://github.com/neuropoly/spinalcordtoolbox/pull/2283#discussion_r288259468, the suggestion is to:
- rename remove --> remove-reference
- use remove for removing specified labels
- add keep for keeping specified labels

YangHee-Min: Oops sorry I misunderstood! Newest changes in commit [bce701a](https://github.com/neuropoly/spinalcordtoolbox/commit/bce701a2843c85bf6686287dcdf5025df771929a)

scripts/sct_label_utils.py
```diff
@@ -125,6 +125,10 @@ def process(self, type_process):
             self.output_image = self.continuous_vertebral_levels()
         if type_process == 'create-viewer':
             self.output_image = self.launch_sagittal_viewer(self.value)
+        if type_process == 'remove':
+            self.output_image = self.remove_labels(self.value)
+        if type_process == 'keep':
+            self.output_image = self.remove_labels(self.value,isKeep = True)
```
jcohenadad: the name of the function is not intuitive (we expect it to remove labels, not to keep them). I suggest to rename it `remove_or_keep_labels(self.value, action={'keep', 'remove'})`

YangHee-Min: Got it. Changes applied in commit [ec624a5](https://github.com/neuropoly/spinalcordtoolbox/commit/ec624a5d324d90bcddf03897b78c861cb55abc82)

scripts/sct_label_utils.py
```diff
@@ -125,6 +125,10 @@ def process(self, type_process):
             self.output_image = self.continuous_vertebral_levels()
         if type_process == 'create-viewer':
             self.output_image = self.launch_sagittal_viewer(self.value)
+        if type_process == 'remove':
```
jcohenadad: you can save two lines of code:
```python
if type_process == 'remove' or 'keep':
   self.output_image = self.remove_or_keep_labels(self.value, action=type_process)
```

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,30 @@ def launch_sagittal_viewer(self, labels):
 
         return output
 
+    def remove_or_keep_labels(self, labels, action):
+        """
+        Create or remove labels from self.image_input
+        :param list(int): Labels to add or remove
+        :param str: 'remove': remove labels (i.e. set to zero), 'keep': keep labels
```
jcohenadad: i would be a bit more explicit: 
~~~
'remove': remove specified labels, 'keep': keel specified labels and remove the others
~~~

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,30 @@ def launch_sagittal_viewer(self, labels):
 
         return output
 
+    def remove_or_keep_labels(self, labels, action):
+        """
+        Create or remove labels from self.image_input
+        :param list(int): Labels to add or remove
```
jcohenadad: i would replace "add" by "keep"

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,30 @@ def launch_sagittal_viewer(self, labels):
...
+                if labelNumber == coord.value:
+                    new_coord = coord
+                    isInLabels = True
+            if isInLabels:
+                if action == 'keep':
+                    image_output.data[int(new_coord.x), int(new_coord.y), int(new_coord.z)] = new_coord.value
+                else:
+                    image_output.data[int(new_coord.x), int(new_coord.y), int(new_coord.z)] = 0.0
+            else:
+                sct.printv("ERROR: Label " + str(float(labelNumber)) + " not found in input image.", type='error')
```
jcohenadad: i would not throw an error. Maybe just a warning

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,30 @@ def launch_sagittal_viewer(self, labels):
...
+        for labelNumber in labels:
+            isInLabels = False
+            for coord in coordinates_input:
+                if labelNumber == coord.value:
+                    new_coord = coord
+                    isInLabels = True
+            if isInLabels:
+                if action == 'keep':
+                    image_output.data[int(new_coord.x), int(new_coord.y), int(new_coord.z)] = new_coord.value
+                else:
```
jcohenadad: replace "else" by "elif action == 'remove'" for more clarity and code expandability

scripts/sct_label_utils.py
```diff
@@ -598,6 +602,30 @@ def launch_sagittal_viewer(self, labels):
 
         return output
 
+    def remove_or_keep_labels(self, labels, action):
+        """
+        Create or remove labels from self.image_input
+        :param list(int): Labels to add or remove
+        :param str: 'remove': remove labels (i.e. set to zero), 'keep': keep labels
+        """
+        image_output = msct_image.zeros_like(self.image_input) if action == 'keep' else  self.image_input.copy()
```
jcohenadad: for clarity, always use:
```python
if action == 'keep':
  image_output = msct_image.zeros_like(self.image_input)
elif action == 'remove':
  image_output = self.image_input.copy()
```

scripts/sct_label_utils.py
YangHee-Min: Got it. Changes made in commit [0fe767d](https://github.com/neuropoly/spinalcordtoolbox/commit/0fe767df0718e84e21a26f1ca21a444364431075)

scripts/sct_label_utils.py
YangHee-Min: Oops my mistake. Changes made in commit [0fe767d](https://github.com/neuropoly/spinalcordtoolbox/commit/0fe767df0718e84e21a26f1ca21a444364431075)

scripts/sct_label_utils.py
YangHee-Min: Understood. Changed in commit [cd29609](https://github.com/neuropoly/spinalcordtoolbox/commit/cd2960958c1f2a4215d14b7b53d21f5db8d9d096)

jcohenadad:
Make sure that all functions in SCT don't break after the change of `remove` --> `remove-reference`, see example of breakage in `sct_register_to_template`: https://travis-ci.org/neuropoly/spinalcordtoolbox/jobs/538759171#L1661

scripts/sct_label_utils.py
YangHee-Min: Understood, changed in commit [70d3f2e](https://github.com/neuropoly/spinalcordtoolbox/commit/70d3f2e11c5153c5cd3a576a930e66dd48291ca7)

PR #2292, opened 2019-06-03
Fixing straightening when using NURBS 

### LINKED ISSUE #2258: NURBS is broken
Recent changes broke the integrity of nurbs output, see example below:

to reproduce:
~~~
cd sct_example_data/t1
sct_straighten_spinalcord -i t1.nii.gz -s t1_seg.nii.gz -param algo_fitting=nurbs -o t1_straight_algo-nurbs.nii.gz
~~~

nurbs:
Screen Shot 2019-05-09 at 11 07 03

polyfit:
Screen Shot 2019-05-09 at 11 07 00

benjamindeleener:
This Pull Request fixes issue #2258, which is related to calculating the straightening transformation based on the centerline fitting.

The PR also adds some QC capabilities for fitting, including the derivatives of the fitting. The QC is only available in the temporary folder when using the option `-v 2`.

Fixes #2258

### DIFF
--- a/spinalcordtoolbox/centerline/core.py
+++ b/spinalcordtoolbox/centerline/core.py
@@ -74,10 +74,12 @@ def get_centerline(im_seg, algo_fitting='polyfit', minmax=True, contrast=None, d
     if algo_fitting == 'polyfit':
         x_centerline_fit, x_centerline_deriv = curve_fitting.polyfit_1d(z_mean, x_mean, z_ref, deg=degree)
         y_centerline_fit, y_centerline_deriv = curve_fitting.polyfit_1d(z_mean, y_mean, z_ref, deg=degree)
+        z_centerline_deriv = np.ones_like(z_ref)
 
     elif algo_fitting == 'bspline':
         x_centerline_fit, x_centerline_deriv = curve_fitting.bspline(z_mean, x_mean, z_ref, smooth=smooth)
         y_centerline_fit, y_centerline_deriv = curve_fitting.bspline(z_mean, y_mean, z_ref, smooth=smooth)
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/centerline/core.py
```diff
@@ -86,6 +88,7 @@ def get_centerline(im_seg, algo_fitting='polyfit', minmax=True, contrast=None, d
         # Compute derivatives using polynomial fit due to undefined derivatives using linear interpolation
         _, x_centerline_deriv = curve_fitting.polyfit_1d(z_mean, x_mean, z_ref, deg=degree)
         _, y_centerline_deriv = curve_fitting.polyfit_1d(z_mean, y_mean, z_ref, deg=degree)
+        z_centerline_deriv = np.ones_like(z_ref)
 
     elif algo_fitting == 'nurbs':
         from spinalcordtoolbox.centerline.nurbs import b_spline_nurbs
```
jcohenadad: @benjamindeleener I'm trying to wrap my head around the fact that `z_centerline_deriv` from nurbs is not a vector of ones, given that the space between each element of `z_centerline_fit` is one.

jcohenadad: ok i figured this out. This is because NURBS alters the density of sampling, and then outputs the derivatives (which are scaled by the density). In this commit (978a9c24b3878ec9a7de2a7dab2d793a2ae7616a) I have normalized the x and y derivatives by the z-derivatives, such that all algorithms in `get_centerline()` output comparable derivatives which can readily be used (e.g. for straightening, correcting for CSA, etc.).

Issue #2262, opened 2019-05-11
Allow the possibility to discard "Dataset" column on the output html QC report


jcohenadad:
In some cases, for regular BIDS dataset, the column "Dataset" is not necessary because redundant. It would be nice to allow the possibility to discard it, so that there is more room to display the other columns.

PR #2301, opened 2019-06-06
Allow the possibility to discard column of choice on the output html QC report

YangHee-Min:
At the moment, when generating the QC report html file, it is not possible to hide columns that can become repetitive. 

This PR introduces a feature that allows the user to discard any column by introducing a button labeled "Hide Columns" that is found within the html file to the left of the search bar. When the button "Hide Columns" is pressed, a dropdown menu appears and lets the user pick which columns to remove. (enhances #2262)

Note: It was requested to simply remove the Dataset column in issue #2262, but I figured there could be some use in being able to remove any column. Feature may be subject to change. 

Example of new index.html file when opened (button "Hide Columns" closed):
2262_index

Example of new index.html file when opened (button "Hide Columns" closed):
Screen Shot 2019-06-06 at 4 03 06 PM


spinalcordtoolbox/reports/assets/index.html
```diff
@@ -100,4 +112,70 @@
...
+        open even if the inside contents or the body of the html file are clicked.
+        */
+        e.stopPropagation();
+    }
+
+    body.onclick = stopDropdownClose;
+    btn.onclick = manualOpenClose;
+
+</script>
+</html>
```
jcohenadad: add carriage return

YangHee-Min: Ok! Added in commit [156df39](https://github.com/neuropoly/spinalcordtoolbox/commit/156df39d228c7d127a52204e1ff903b3b43a1750)

YangHee-Min:
"Hide Columns"  label replaced in commit [a6c2be2](https://github.com/neuropoly/spinalcordtoolbox/pull/2301/commits/a6c2be2cd9a4295706483df07e1f749bfcf21cbf)

YangHee-Min:
Would the additional columns vary based on the function called? Not quite sure which ones would be added and which ones to add at the moment since there were only these 4 by default.

jcohenadad:
> Would the additional columns vary based on the function called?

No

YangHee-Min:
> Along with the previous statement, we could leave the possibility to add even more columns (switched off by default). For example: `contrast`, `fname_in`

So would it be a feature where the user would manually input more columns? Sorry I'm not quite sure if I understood your statement.

jcohenadad:
> So would it be a feature where the user would manually input more columns? Sorry I'm not quite sure if I understood your statement.

no, you could simply fetch these variables (already produced by reports/qc, see [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/reports/qc.py#L346)) and add them in the index.html.

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -68,10 +68,20 @@
               <div id="dropdown" class="dropdown">
                   <button id="dropdownButton" class="btn btn-default dropdown-toggle" type="button" data-toggle="dropdown" aria-haspopup="true">Display Columns <span class="caret"></span></button>
                   <ul class="dropdown-menu">
-                      <li><a class="dropdown-item" id="Date" onclick="toggleColumn(id)">Display Date &#10004</button></a></li>
-                      <li><a class="dropdown-item" id="Dataset" onclick="toggleColumn(id)">Display Dataset &#10004</a></li>
-                      <li><a class="dropdown-item" id="Subject" onclick="toggleColumn(id)">Display Subject &#10004</a></li>
-                      <li><a class="dropdown-item" id="Function" onclick="toggleColumn(id)">Display Function &#10004</a></li>
+                      <li><a class="dropdown-item" id="Date" onclick="toggleColumn(id)">Display Date</button></a></li>
```
jcohenadad: a lot of these fields are used internally. Just stick to the following (x means default):
- Date: x
- Dataset: 
- Subject: x
- File (rename from Input_file_name): x
- Contrast
- Function: x

YangHee-Min:
All variables from qc.py output can now be found within the dropdown button. Changes in commit [5435cc5](https://github.com/neuropoly/spinalcordtoolbox/pull/2301/commits/5435cc5d13fb3edcf69847821037cf4ab58d2d89). 

In this commit, I used a time based method[ (as seen here)](https://github.com/neuropoly/spinalcordtoolbox/blob/5435cc5d13fb3edcf69847821037cf4ab58d2d89/spinalcordtoolbox/reports/assets/index.html#L196-L208), because if I tried to hide the unwanted columns when the page loaded using the 'onload' method, the columns would not go away. It seemed that after a few tests, the problem seemed to be that even though the page was loaded, the table was not actually completely loaded (probably since the table is still following the script). I don't like the fact I rely on a certain time delay to hide the unwanted columns. Do you have any suggestions on how to proceed or is the time based method alright for now?

spinalcordtoolbox/reports/assets/index.html
YangHee-Min: > UPDATE 2019-06-07 11.01:
You can remove the "Display" in the dropdown menu (people get it)

Remove the "display" from the button, the internal elements of the menu or both?

jcohenadad:
> In this commit, I used a time based method (as seen here), because if I tried to hide the unwanted columns when the page loaded using the 'onload' method, the columns would not go away. It seemed that after a few tests, the problem seemed to be that even though the page was loaded, the table was not actually completely loaded (probably since the table is still following the script). I don't like the fact I rely on a certain time delay to hide the unwanted columns. Do you have any suggestions on how to proceed or is the time based method alright for now?

I tested and didn't find any UI problem. What are the disadvantages of your approach? Maybe @TommyBoshkovski has some tips for you.

A minor request: currently, when we want to unselect the dropdown menu, I intuitively click outside of the menu, but it does not disappear (in many other web-based application, this behaviour is expected). Is there an easy way to make this happen?

spinalcordtoolbox/reports/assets/index.html
jcohenadad: just from the internal elements

YangHee-Min:
I am aware of that issue. I believe I have to stop the default behaviour of one of the tags but I still have to test which one will work. I tried using `e.stopPropagation();` on the submenu but it had no effect. I will look more into it but if you find something let me know!

YangHee-Min:
> I tested and didn't find any UI problem. What are the disadvantages of your approach? Maybe @TommyBoshkovski has some tips for you.
Only disadvantage (but it's really a minor detail) is that the code simply isn't fully optimized time-wise (since it doesn't execute the function right away). If it isn't that relevant and this method is fine, I am okay with keeping it this way.

jcohenadad:
> Only disadvantage (but it's really a minor detail) is that the code simply isn't fully optimized time-wise (since it doesn't execute the function right away).

What delay are you talking about? microseconds? ms? seconds? I really didn't notice any time lag.

jcohenadad:
@YangHee-Min could you also take this opportunity to update the "About" hyperlink (top right of the html QC report), with the following URL: https://github.com/neuropoly/spinalcordtoolbox

YangHee-Min:
> What delay are you talking about? microseconds? ms? seconds? I really didn't notice any time lag.

I was able to reduce it to a 2ms delay. It's very short so if we don't need to worry about how optimized the table load time is then I don't mind moving on.

YangHee-Min:
Changed "About" link in commit [0267460](https://github.com/neuropoly/spinalcordtoolbox/commit/0267460eaa5e95a935ffc51d6037ef8771a48d91)

jcohenadad:
> > What delay are you talking about? microseconds? ms? seconds? I really didn't notice any time lag.
> 
> I was able to reduce it to a 2ms delay. It's very short so if we don't need to worry about how optimized the table load time is then I don't mind moving on.

I think people can wait 2ms :sweat_smile:

YangHee-Min:
> A minor request: currently, when we want to unselect the dropdown menu, I intuitively click outside of the menu, but it does not disappear (in many other web-based application, this behaviour is expected). I have to click on the "Display Column" to make it disappear. Is there an easy way to make this happen?

Solution found! Changes made in commit [18cb9ef](https://github.com/neuropoly/spinalcordtoolbox/commit/18cb9ef988e0a4c77464a435ade8c15ee223690e)

YangHee-Min:
> does it also close if users clicks on the "Display Column" button?

Yes

jcohenadad:
> > does it also close if users clicks on the "Display Column" button?
> 
> Yes

in that case, you should correct the mis-leading comment. Suggestion: "... closes the dropdown menu if user clicks on the button or outside of the dropdown-menu."

YangHee-Min:
>>> does it also close if users clicks on the "Display Column" button?
>>
>>Yes
>
> in that case, you should correct the mis-leading comment. Suggestion: "... closes the dropdown menu if user clicks on the button or outside of the dropdown-menu."

Commit message now updated [2f73396](https://github.com/neuropoly/spinalcordtoolbox/commit/2f73396cd62c704e8b06186c87c006e566ae0913)

jcohenadad:
> > > > does it also close if users clicks on the "Display Column" button?
> > > 
> > > 
> > > Yes
> > 
> > 
> > in that case, you should correct the mis-leading comment. Suggestion: "... closes the dropdown menu if user clicks on the button or outside of the dropdown-menu."
> 
> Commit message now updated [2f73396](https://github.com/neuropoly/spinalcordtoolbox/commit/2f73396cd62c704e8b06186c87c006e566ae0913)

i didn't mean to change the commit message but the actual function header ;-)

YangHee-Min:
> >>>does it also close if users clicks on the "Display Column" button?
> >>
> >>Yes
> >>
> >>in that case, you should correct the mis-leading comment. Suggestion: "... closes the dropdown menu if user clicks on the button or outside of the dropdown-menu."
> >
> >Commit message now updated 2f73396
> 
> i didn't mean to change the commit message but the actual function header ;-)

Woops I misunderstood! Changed in commit [4c2ae78](https://github.com/neuropoly/spinalcordtoolbox/commit/4c2ae78838ed8496c88afeb6dfd78a9f29090979)

jcohenadad:
@YangHee-Min is this PR ready for re-review?

YangHee-Min:
> @YangHee-Min is this PR ready for re-review?
@jcohenadad yes it's ready!

PR #2306, opened 2019-06-10
Add CSA results on QC report 

jcohenadad:
This PR features a new entry to the QC report: CSA results. 

This report would also facilitate the optimization of default parameter for `get_centerline` (used for angle correction in CSA computation), see https://github.com/neuropoly/spinalcordtoolbox/pull/2299

To test:
```bash
cd sct_testing_data/t2
sct_process_segmentation -i t2_seg.nii.gz -qc qc-csa
```

### DIFF
--- a/scripts/sct_process_segmentation.py
+++ b/scripts/sct_process_segmentation.py
@@ -173,8 +173,8 @@ def _make_figure(metric):
     ax.plot(z, angle_rl, 'r')
     ax.plot(z, angle_rl, 'r.')
     ax.grid(True)
-    ax.set_xlabel('Slice (in superior-inferior direction)')
-    ax.set_ylabel('Angle [deg]')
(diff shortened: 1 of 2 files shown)


jcohenadad:
hum, strange. Can you go back to the master version and run `sct_testing`?

YangHee-Min:
After running `sct_testing` on the master branch, all scripts get result `[OK]` except for `sct_process_segmentation` that fails. [Here is the full log](https://pastebin.com/WpmwYsT9)

jcohenadad:
ah! interesting (only on Sierra build):
https://travis-ci.org/neuropoly/spinalcordtoolbox/jobs/542929536#L903

YangHee-Min:
bug on Sierra likely related to: https://stackoverflow.com/questions/49414841/process-finished-with-exit-code-139-interrupted-by-signal-11-sigsegv

Issue #2246, opened 2019-05-01
Fix minor issues caused by passage to py3k 


jcohenadad:
After merging #2238 the following issues remain:

- Travis:
  - test install with CSHRC
  - Remove libopenmpi --> addressed in https://github.com/neuropoly/spinalcordtoolbox/pull/2310
  - ~~put back sphinx building (removed because was failing on OSX)~~ --> will be addressed by https://github.com/neuropoly/spinalcordtoolbox/issues/1656
  - Exit travis on error if pb during install (currently failing)
  - Test Travis on OSX without xcode installed (to see how dipy compilation goes)
- ~~Remove `$SCT_DIR` variable everywhere in the code and in installer~~
- ~~Fix tensorflow install in Trusty (and remove "allow_failures" on Travis)~~
- ~~Fix PyQt5 changes in QtWidget (cosmetic stuff)~~ --> will be addressed in https://github.com/neuropoly/spinalcordtoolbox/issues/2357
- ~~Fix broken `sct_segment_graymatter` (put back in sct_testing).~~ --> will be addressed in https://github.com/neuropoly/spinalcordtoolbox/issues/2244
- ~~Remove code and files related to Conda requirements (no more needed :-) )~~ --> to be addressed in https://github.com/neuropoly/spinalcordtoolbox/issues/2358
- ~~Remove MPI stuff (not needed anymore, difficult to maintain)~~ --> to be addressed in https://github.com/neuropoly/spinalcordtoolbox/issues/2359
- Fix `scikit-image [...] Could not open font file /Library/Fonts/NISC18030.ttf` that happens when running `sct_check_dependencies` on some stations (e.g., `monk` @mojave)
- ~~Display SCT version and platform info for each call, for easier debugging.~~

## Deprecations

The following lists the deprecation to fix in a subsequent PR:

<details>
  <summary>DeprecationWarning: get_header method is deprecated</summary>

~~~
/Users/julien/code/sct-py3k/scripts/sct_create_mask.py:163: DeprecationWarning: get_header method is deprecated.
Please use the ``img.header`` property instead.

* deprecated from version: 2.1
* Will raise <class 'nibabel.deprecator.ExpiredDeprecationError'> as of version: 4.0
  hdr = centerline.get_header()  # get header
  OK: sub-03_acq-T1w_MTS_crop_r_seg.nii.gz
~~~

</details>

<details>
[...]

PR #2310, opened 2019-06-11
Fixed minor issues caused by the move to py3k

jcohenadad:
After merging #2238, issues listed in #2246 remain. This PR fixes some of those issues (also checked in the original issue):
- Travis: Remove libopenmpi


.travis.yml
```diff
@@ -37,7 +37,7 @@ matrix:
 before_install:
   - echo "HOME="
   - echo $HOME
-  - sudo apt-get update -qq; sudo apt-get install -qq libopenmpi-dev; echo
+  - sudo apt-get update -qq; echo
```
gmotzespina: I'm not sure what the `echo` at the end is used for. Is it necessary? If that echo is needed, it looks good to me.

jcohenadad: indeed, not needed! thx for the review

PR #2315, opened 2019-06-17
Conversion from msct_parser to argparse

### LINKED ISSUE #1548: Replace usage of custom command-line parser (`msct_parser`) with argparse

YangHee-Min:
At the moment, SCT uses its own command-line parser msct_parser. However, as mentioned in issue #1548, it is not very practical and some griefs with the command-line arguments parser would make the switch from msct_parser to argparse worthwhile.

In this PR, we look to convert the scripts seen in the list below (more functions to be added) from msct_parser to argparse. (Related to #1548 ):

DONE:
- sct_analyze_lesion
- sct_analyze_texture
- sct_apply_transfo
- sct_check_dependencies
- sct_compute_ernst_angle
- sct_compute_hausdorff_distance
- sct_compute_mscc
- sct_compute_mtr
- sct_compute_mtsat
- sct_compute_snr
- sct_concat_transfo
- sct_convert
- sct_create_mask
- sct_crop_image
- Refactored `sct_testing` to be compatible with this new parser. Re-added all the tests.
- Now using space instead of comma to separate multiple input warping fields in `sct_apply_transfo`.

TODO: 
- Removed flag `-igt` in functions, because ground truth data are now input directly into the testing framework (no need to expose this to the user).

Example of code: https://github.com/neuropoly/spinalcordtoolbox/blob/7378068d23892d33b82f3dbb489a204fcb9611cb/scripts/sct_compute_mtsat.py#L63-L82
Example of usage output: https://pastebin.com/Y6KTz1Ly

### DIFF
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -17,13 +17,12 @@ import numpy as np
 import pandas as pd
 from skimage.measure import label
 
-import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
-from msct_types import Centerline
-import sct_utils as sct
-from sct_utils import extract_fname, printv, tmp_create
 from spinalcordtoolbox.centerline.core import get_centerline
 from spinalcordtoolbox.utils import Metavar
+
+import sct_utils as sct
+from sct_utils import extract_fname, printv, tmp_create
 
 
 def get_parser():
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -510,7 +510,7 @@ class AnalyzeLeion:
         os.chdir(self.tmp_dir)  # go to tmp directory
 
 
-def main(args):
+def main(args=None):
     """
     Main function
     :param args:
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -60,6 +60,7 @@ def get_parser():
         help="Spinal cord centerline or segmentation file, which will be used to correct morphometric measures with "
              "cord angle with respect to slice. (e.g.'t2_seg.nii.gz')",
         metavar=Metavar.file,
+        default=None,
         required=False)
     optional.add_argument(
         "-i",
@@ -517,8 +518,6 @@ def main():
         if not os.path.isfile(fname_sc):
             fname_sc = None
             printv('WARNING: -s input file: "' + arguments.s + '" does not exist.\n', 1, 'warning')
-    else:
-        fname_sc = None
 
     # Reference image
     if arguments.i is not None:
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -43,12 +43,14 @@ def get_parser():
         formatter_class=argparse.RawDescriptionHelpFormatter,
         prog=os.path.basename(__file__).strip(".py")
     )
+
     mandatory_arguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
     mandatory_arguments.add_argument(
         "-m",
         help='Binary mask of lesions (lesions are labeled as "1").',
         metavar=Metavar.file,
         required=False)
+
     optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
     optional.add_argument(
         "-h",
@@ -67,6 +69,7 @@ def get_parser():
         help='Image from which to extract average values within lesions (e.g. "t2.nii.gz"). If provided, the function '
              'computes the mean and standard deviation values of this image within each lesion.',
         metavar=Metavar.file,
+        default=None,
         required=False)
     optional.add_argument(
         "-f",
@@ -75,11 +78,13 @@ def get_parser():
              "region of the template (e.g. GM, WM, WM tracts) and (ii) the proportion of ROI (e.g. vertebral level, "
              "GM, WM) occupied by lesion.",
         metavar=Metavar.str,
+        default=None,
         required=False)
     optional.add_argument(
         "-ofolder",
         help='Output folder (e.g. "./")',
         metavar=Metavar.folder,
+        default='./',
         required=False)
     optional.add_argument(
         "-r",
@@ -509,43 +514,23 @@ def main():
     parser = get_parser()
     arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
 
-    # set param arguments ad inputted by user
     fname_mask = arguments.m
-
-    # SC segmentation
-    if arguments.s is not None:
-        fname_sc = arguments.s
-        if not os.path.isfile(fname_sc):
-            fname_sc = None
-            printv('WARNING: -s input file: "' + arguments.s + '" does not exist.\n', 1, 'warning')
-
-    # Reference image
-    if arguments.i is not None:
-        fname_ref = arguments.i
-        if not os.path.isfile(fname_sc):
-            fname_ref = None
-            printv('WARNING: -i input file: "' + arguments.i + '" does not exist.\n', 1, 'warning')
-    else:
-        fname_ref = None
+    fname_sc = arguments.s
+    fname_ref = arguments.i
 
     # Path to template
-    if arguments.f is not None:
-        path_template = arguments.f
-        if not os.path.isdir(path_template) and os.path.exists(path_template):
-            path_template = None
-            printv("ERROR output directory %s is not a valid directory" % path_template, 1, 'error')
-    else:
-        path_template = None
+    path_template = arguments.f
+    # TODO: check this in the parser
+    # if not os.path.isdir(path_template) and os.path.exists(path_template):
+    #     path_template = None
+    #     printv("ERROR output directory %s is not a valid directory" % path_template, 1, 'error')
 
     # Output Folder
-    if arguments.ofolder is not None:
-        path_results = arguments.ofolder
-        if not os.path.isdir(path_results) and os.path.exists(path_results):
-            printv("ERROR output directory %s is not a valid directory" % path_results, 1, 'error')
-        if not os.path.exists(path_results):
-            os.makedirs(path_results)
-    else:
-        path_results = './'
+    path_results = arguments.ofolder
+    # if not os.path.isdir(path_results) and os.path.exists(path_results):
+    #     printv("ERROR output directory %s is not a valid directory" % path_results, 1, 'error')
+    if not os.path.exists(path_results):
+        os.makedirs(path_results)
 
     # Remove temp folder
     if arguments.r is not None:
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -510,9 +510,17 @@ class AnalyzeLeion:
         os.chdir(self.tmp_dir)  # go to tmp directory
 
 
-def main():
+def main(args):
+    """
+    Main function
+    :param args:
+    :return:
+    """
+    # get parser args
+    if args is None:
+        args = None if sys.argv[1:] else ['--help']
     parser = get_parser()
-    arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
+    arguments = parser.parse_args(args=args)
 
     fname_mask = arguments.m
     fname_sc = arguments.s
--- a/scripts/sct_analyze_texture.py
+++ b/scripts/sct_analyze_texture.py
@@ -306,7 +306,7 @@ class ParamGLCM(object):
         self.angle = '0,45,90,135'  # Rotation angles for co-occurrence matrix
 
 
-def main(args):
+def main(args=None):
     """
     Main function
     :param args:
--- a/scripts/sct_analyze_texture.py
+++ b/scripts/sct_analyze_texture.py
@@ -306,9 +306,17 @@ class ParamGLCM(object):
         self.angle = '0,45,90,135'  # Rotation angles for co-occurrence matrix
 
 
-def main():
+def main(args):
+    """
+    Main function
+    :param args:
+    :return:
+    """
+    # get parser args
+    if args is None:
+        args = None if sys.argv[1:] else ['--help']
     parser = get_parser()
-    arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
+    arguments = parser.parse_args(args=args)
 
     # create param object
     param = Param()
--- a/scripts/sct_analyze_texture.py
+++ b/scripts/sct_analyze_texture.py
@@ -95,12 +95,14 @@ def get_parser():
         "-r",
         help="Remove temporary files.",
         required=False,
+        type=int,
         choices=(0, 1),
         default=int(Param().rm_tmp))
     optional.add_argument(
         "-v",
         help="Verbose: 0 = nothing, 1 = classic, 2 = expended.",
         required=False,
+        type=int,
         choices=(0, 1, 2),
         default=Param().verbose)
 
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -68,7 +68,7 @@ def get_parser():
              'field, then directly input the inverse warping field in flag -w.',
         nargs='+',
         metavar=Metavar.file,
-        default=None)
+        default=[])
     optional.add_argument(
         "-h",
         "--help",
@@ -112,7 +112,7 @@ def get_parser():
 
 
 class Transform:
-    def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=None, output_filename='', verbose=0, crop=0,
+    def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=[], output_filename='', verbose=0, crop=0,
                  interp='spline', remove_temp_files=1, debug=0):
         self.input_filename = input_filename
         self.list_warp = list_warp
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -288,6 +288,12 @@ def main(args=None):
     # get parser args
     if args is None:
         args = None if sys.argv[1:] else ['--help']
+    else:
+        # flatten the list of input arguments because -w and -winv carry a nested list
+        lst = []
+        for line in args:
+            lst.append(line) if isinstance(line, str) else lst.extend(line)
+        args = lst
     parser = get_parser()
     arguments = parser.parse_args(args=args)
 
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -55,11 +55,20 @@ def get_parser():
         metavar=Metavar.file)
     mandatoryArguments.add_argument(
         "-w",
-        help='Transformation, which can be a warping field (nifti image) or an affine transformation matrix (text '
-             'file). (e.g. "warp1.nii.gz, warp2.nii.gz")',
+        help='Transformation(s), which can be warping fields (nifti image) or affine transformation matrix (text '
+             'file). Separate with space. Example: warp1.nii.gz warp2.nii.gz',
         nargs='+',
-        metavar='')
+        metavar=Metavar.file)
+
     optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        "-winv",
+        help='Affine transformation(s) listed in flag -w which should be inverted before being used. Note that this'
+             'only concerns affine transformation (not warping fields). If you would like to use an inverse warping'
+             'field, then directly input the inverse warping field in flag -w.',
+        nargs='+',
+        metavar=Metavar.file,
+        default=None)
     optional.add_argument(
         "-h",
         "--help",
@@ -103,12 +112,11 @@ def get_parser():
 
 
 class Transform:
-    def __init__(self, input_filename, warp, fname_dest, output_filename='', verbose=0, crop=0, interp='spline', remove_temp_files=1, debug=0):
+    def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=None, output_filename='', verbose=0, crop=0,
+                 interp='spline', remove_temp_files=1, debug=0):
         self.input_filename = input_filename
-        if isinstance(warp, str):
-            self.warp_input = list([warp])
-        else:
-            self.warp_input = warp
+        self.list_warp = list_warp
+        self.list_warpinv = list_warpinv
         self.fname_dest = fname_dest
         self.output_filename = output_filename
         self.interp = interp
@@ -120,7 +128,7 @@ class Transform:
     def apply(self):
         # Initialization
         fname_src = self.input_filename  # source image (moving)
-        fname_warp_list = self.warp_input  # list of warping fields
+        list_warp = self.list_warp  # list of warping fields
         fname_out = self.output_filename  # output
         fname_dest = self.fname_dest  # destination image (fix)
         verbose = self.verbose
@@ -133,20 +141,20 @@ class Transform:
         sct.printv('\nParse list of warping fields...', verbose)
         use_inverse = []
         fname_warp_list_invert = []
-        # fname_warp_list = fname_warp_list.replace(' ', '')  # remove spaces
-        # fname_warp_list = fname_warp_list.split(",")  # parse with comma
-        for idx_warp, path_warp in enumerate(fname_warp_list):
-            # Check if inverse matrix is specified with '-' at the beginning of file name
-            if path_warp.startswith("-"):
+        # list_warp = list_warp.replace(' ', '')  # remove spaces
+        # list_warp = list_warp.split(",")  # parse with comma
+        for idx_warp, path_warp in enumerate(self.list_warp):
+            # Check if this transformation should be inverted
+            if path_warp in self.list_warpinv:
                 use_inverse.append('-i')
-                fname_warp_list[idx_warp] = path_warp[1:]  # remove '-'
-                fname_warp_list_invert += [[use_inverse[idx_warp], fname_warp_list[idx_warp]]]
+                # list_warp[idx_warp] = path_warp[1:]  # remove '-'
+                fname_warp_list_invert += [[use_inverse[idx_warp], list_warp[idx_warp]]]
             else:
                 use_inverse.append('')
                 fname_warp_list_invert += [[path_warp]]
-            path_warp = fname_warp_list[idx_warp]
+            path_warp = list_warp[idx_warp]
             if path_warp.endswith((".nii", ".nii.gz")) \
-             and msct_image.Image(fname_warp_list[idx_warp]).header.get_intent()[0] != 'vector':
+             and msct_image.Image(list_warp[idx_warp]).header.get_intent()[0] != 'vector':
                 raise ValueError("Displacement field in {} is invalid: should be encoded" \
                  " in a 5D file with vector intent code" \
                  " (see https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h" \
@@ -192,13 +200,11 @@ class Transform:
             else:
                 dim = '3'
             sct.run(['isct_antsApplyTransforms',
-              '-d', dim,
-              '-i', fname_src,
-              '-o', fname_out,
-              '-t',
-             ] + fname_warp_list_invert + [
-             '-r', fname_dest,
-             ] + interp, verbose=verbose, is_sct_binary=True)
+                     '-d', dim,
+                     '-i', fname_src,
+                     '-o', fname_out,
+                     '-t'] + fname_warp_list_invert + ['-r', fname_dest] + interp,
+                    verbose=verbose, is_sct_binary=True)
 
         # if 4d, loop across the T dimension
         else:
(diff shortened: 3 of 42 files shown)


jcohenadad:
@YangHee-Min I changed "fixes #1548" to "Related to #1548" because this PR only partially fixes the issue (i.e. we want to keep it open after this PR)

jcohenadad:
> > you can fetch that automatically (use os.path.basename(**file**))
> 
> But this returns `sct_compute_mtsat.py`. Would you like me to use a string method to remove the `.py` at the end?
> 
> Edit: For example:
> `prog= os.path.basename(__file__).replace(".py", "")`

actually, i find this a bit more elegant:
```python
prog= os.path.basename(__file__).strip('.py')
```

YangHee-Min:
Got it, changed in commit [a783529](https://github.com/neuropoly/spinalcordtoolbox/commit/a783529b3685a08d10e02908937e7543e5b99a10)

YangHee-Min:
While reviewing the functions, I realized that we do not have a standard way to show examples for arguments when -help is called.

Here are 2 methods I recall using:

Method 1 [(example here)](https://github.com/neuropoly/spinalcordtoolbox/blob/e27c56e8840b2a312c10177629b669c4730aaccd/scripts/sct_analyze_lesion.py#L52):
`-flag argumentDescription (e.g. insertExampleHere)`

Method 2 [(example here)](https://github.com/neuropoly/spinalcordtoolbox/blob/e27c56e8840b2a312c10177629b669c4730aaccd/scripts/sct_create_mask.py#L300):
`-flag argumentDescription E.g.: insertExampleHere`

Would you like us to use one of these methods (or a different one) as a standard method to present examples for all arguments or should we not bother which one is used?

jcohenadad:
> While reviewing the functions, I realized that we do not have a standard way to show examples for arguments when -help is called.
> 
> Here are 2 methods I recall using:
> 
> Method 1 [(example here)](https://github.com/neuropoly/spinalcordtoolbox/blob/e27c56e8840b2a312c10177629b669c4730aaccd/scripts/sct_analyze_lesion.py#L52):
> `-flag argumentDescription (e.g. insertExampleHere)`
> 
> Method 2 [(example here)](https://github.com/neuropoly/spinalcordtoolbox/blob/e27c56e8840b2a312c10177629b669c4730aaccd/scripts/sct_create_mask.py#L300):
> `-flag argumentDescription E.g.: insertExampleHere`
> 
> Would you like us to use one of these methods (or a different one) as a standard method to present examples for all arguments or should we not bother which one is used?

Instead of writing a custom example syntax, could we automatically generate example syntaxes (i.e. following the field `example=` that existed in `msct_parser`)?

If not, then you can simply go with Method 1

YangHee-Min:
> Instead of writing a custom example syntax, could we automatically generate example syntaxes (i.e. following the field example= that existed in msct_parser)?
> 
> If not, then you can simply go with Method 1

I did not find a way to automatically generate example syntax so I will proceed with method 1

jcohenadad:
@YangHee-Min i am currently reviewing this PR (this this take a while because I would like to make sure that argument values are properly handled later in the script). One thing I noticed: when no argument is input, it shows an error (mandatory argument missing). It would be more elegant to display the help instead. Is that possible with the current feature that checks for mandatory arguments?

scripts/sct_analyze_lesion.py
```diff
@@ -27,53 +27,65 @@
...
-                      example=['0', '1'])
-    parser.add_option(name="-v",
-                      type_value='multiple_choice',
-                      description="Verbose: 0 = nothing, 1 = classic, 2 = expended",
-                      mandatory=False,
-                      example=['0', '1', '2'],
-                      default_value='1')
+
+    parser = argparse.ArgumentParser(
+        description='Compute statistics on lesions of the input binary file (1 for lesion, 0 for background). The function assigns an ID value to each lesion (1, 2, 3, etc.) and outputs morphometric measures for each lesion:'
```
jcohenadad: @YangHee-Min the visual output can be improved. I am working on it (on your branch, for this file)

scripts/sct_analyze_lesion.py
```diff
@@ -27,53 +27,65 @@
...
+                    '\n\nN.B. If the proportion of lesion in each region (e.g. WM and GM) does not sum '
+                    '\nup to 100%, it means that the registered template does not fully cover the '
+                    '\nlesion, in that case you might want to check the registration results.',
+        add_help=None,
+        formatter_class =argparse.RawDescriptionHelpFormatter,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+    mandatoryArguments = parser.add_argument_group("\nMandatory arguments")
+    mandatoryArguments.add_argument("-m",
+                        help="Lesion mask to analyze (e.g. 't2_lesion.nii.gz')",
```
jcohenadad: @YangHee-Min check your indentation (i'm fixing it for this file, please correct for the other files)

scripts/sct_analyze_lesion.py
```diff
@@ -27,53 +27,65 @@
...
+    optional.add_argument("-i",
+                        help="Image from which to extract average values within lesions (e.g. 't2.nii.gz')",
+                        required=False
+                        )
+    optional.add_argument("-f",
+                        help="Path to folder containing the atlas/template registered to the anatomical image.",
+                        required=False
+                        )
+    optional.add_argument("-ofolder",
+                        help="Output folder (e.g. './')",
```
jcohenadad: better to display " instead of ' if you want to quote something. Which means that you should switch " for ' when defining the string (I'll take care of this file)

jcohenadad:
> One thing I noticed: when no argument is input, it shows an error (mandatory argument missing). It would be more elegant to display the help instead. Is that possible with the current feature that checks for mandatory arguments?

See this solution: 713574925f0a4fd7f11b522fd025a629085ed498

scripts/sct_analyze_lesion.py
```diff
@@ -27,53 +27,72 @@
...
+                    'lesion. In that case you might want to check the registration results.',
+        add_help=None,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+    mandatory_arguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatory_arguments.add_argument(
+        "-m",
+        help='Binary mask of lesions (lesions are labeled as "1").',
+        metavar='',
```
jcohenadad: @YangHee-Min please add this field to change this:
~~~
  -m M         Binary mask of lesions (lesions are labeled as "1").
~~~

into this:
~~~
  -m          Binary mask of lesions (lesions are labeled as "1").
~~~

jcohenadad:
@YangHee-Min I've made some modification on `sct_analyze_lesion`. Could you please review my commits and update the other files of this PR accordingly?

charleygros:
@YangHee-Min : I got an error when running the following:
```
charley@charley:~/sct$ sct_compute_ernst_angle 

--
Spinal Cord Toolbox (git-HMY/1548-argparse-substitution-3a94e2c7b7f2dc694f905cb86d610baa7d4bac73)

usage: sct_compute_ernst_angle -tr TR [-h] [-t1 T1] [-b [B [B ...]]] [-o O]
                               [-ofig OFIG] [-v {0,1,2}]
sct_compute_ernst_angle: error: the following arguments are required: -tr
```

Note that everything is fine when running `sct_compute_ernst_angle --help`.

scripts/sct_compute_hausdorff_distance.py
```diff
@@ -431,56 +431,62 @@ def non_zero_coord(data):
...
-                      type_value="file",
-                      description="Second Image on which you want to find the skeleton",
-                      mandatory=False,
-                      default_value=None,
-                      example='t2star_manual_gmseg.nii.gz')
-    parser.add_option(name="-r",
-                      type_value=None,
-                      description="Second Image on which you want to find the skeleton",
-                      mandatory=False,
-                      deprecated_by='-d')
```
charleygros: @YangHee-Min : I was wondering: Is there a way to have this option (ie `deprecated_by`) in the new version?

scripts/sct_create_mask.py
```diff
@@ -284,70 +276,67 @@ def get_parser():
...
+    mandatoryArguments = parser.add_argument_group("\nMandatory arguments")
+    mandatoryArguments.add_argument('-i',
+                        help='Image to create mask on. Only used to get header. Must be 3D, e.g. \'data.nii.gz\')',
+                        required = True
+                        )
+    mandatoryArguments.add_argument('-p',
+                        help='Process to generate mask.\n'
+                        'coord: X,Y coordinate of center of mask. e.g. \'coord,20x15\'\n'
+                        'point: volume that contains a single point. e.g. \'point,label.nii.gz\'\n'
+                        'center: mask is created at center of FOV.\n'
```
charleygros: @YangHee-Min : (I remember that it has been mentioned during the last SCT meeting, but can't remember the answer) --> Would it be possible to add a "return" here?
Note that I can see the `\n` in the code but herebelow my terminal output:
```
  -p P                  Process to generate mask. coord: X,Y coordinate of
                        center of mask. e.g. 'coord,20x15' point: volume that
                        contains a single point. e.g. 'point,label.nii.gz'
                        center: mask is created at center of FOV. centerline:
                        volume that contains centerline or segmentation. (e.g.
                        'centerline,t2_seg.nii.gz')
```

YangHee-Min: It can be done using `formatter_class = argparse.RawDescriptionHelpFormatter`, but be aware that it might change the way argument help text is displayed

scripts/sct_compute_hausdorff_distance.py
jcohenadad: i think we can remove all the "deprecated_by" in 4.0.0

scripts/sct_create_mask.py
jcohenadad: @YangHee-Min can you show us an example when using this formatter class? Other function help (e.g. register_multimodal) are much more complex than this one so we will definitely end up needing proper formatting for some parameters.

jcohenadad:
I am wondering wether we should create a class with generic object names to display as "metavar" to make it more explicit to the user what input is expected. For example:
```python
        "-m",
        help='Binary mask of lesions (lesions are labeled as "1").',
        metavar='',
```
which outputs this:
~~~
 -m                         Lesion mask to analyze
~~~

would become:
```python
        "-m",
        help='Binary mask of lesions (lesions are labeled as "1").',
        metavar=ParserMetavar().file,
```

and in **spinalcordtoolbox.utils**:
```python
class ParserMetavar:
  def __init__(self):
    file = "<file>"
```

which would output this:
~~~
 -m <file>                    Lesion mask to analyze
~~~

YangHee-Min:
> I am wondering wether we should create a class with generic object names to display as "metavar" to make it more explicit to the user what input is expected. For example:

I think this is a great idea to improve the clarity of the help display!

scripts/sct_create_mask.py
YangHee-Min: [(Here)](https://github.com/neuropoly/spinalcordtoolbox/blob/48514eee05c3b32dca3394bf19995461c6a7082a/scripts/sct_analyze_lesion.py#L42) is an example with `sct_analyze_lesion` and [(here) ](https://pastebin.com/U6bcpRWf) is the output

scripts/sct_compute_hausdorff_distance.py
YangHee-Min: @charleygros I'm still looking for documentation but did not find any up till now. Will keep you posted

scripts/sct_create_mask.py
jcohenadad: @YangHee-Min this example (already implemented) is for the description, not the help specific to a flag (which is what we are discussing here)

YangHee-Min: Oh sorry about that! I gave a bad example for it but as the link [(here)](https://stackoverflow.com/questions/3853722/python-argparse-how-to-insert-newline-in-the-help-text/22157136) suggests, the formatter class is applied onto the the arguments as well if I understood the proposed solution in the link correctly

jcohenadad: ok, cool. Can we see how the formatting would look like when applied to a flag descrption (not the general description)? thanks

YangHee-Min: Sure! [(Here)](https://pastebin.com/72krH0pL) is how the the output looks with `formatter_class= argparse.RawTextHelpFormatter` added (`-p` argument is on line 14). If you want the new code without the other changes to be added (just formatter_class added), you can find it [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/455a431f8e1b217e1d0e31ad65936b601c28a82e/scripts/sct_create_mask.py#L283)

jcohenadad: ok that looks great! i would just indent (and use same convention for e.g.):
~~~
  -p P                  Process to generate mask:
                          <coord,XxY>: Center mask at the X,Y coordinates. Example: "coord,20x15"
                          <point,FILE>: Center mask at the X,Y coordinates of the label defined in input volume FILE. Example: "point,label.nii.gz"
                          <center>: Center mask in the middle of the FOV (nx/2, ny/2).
                          <centerline,FILE>: At each slice, the mask is centered at the spinal cord centerline, defined by the input segmentation FILE. Example: "centerline,t2_seg.nii.gz"
~~~

jcohenadad:
> > I am wondering wether we should create a class with generic object names to display as "metavar" to make it more explicit to the user what input is expected. For example:
> 
> I think this is a great idea to improve the clarity of the help display!

Actually, we should use [enum](https://docs.python.org/3/library/enum.html) instead of creating an empty class

scripts/sct_create_mask.py
YangHee-Min: Is [this (commit 64269d9)](https://github.com/neuropoly/spinalcordtoolbox/commit/64269d9979757dcbd2a354e8c87babb58be4b07e) what you wanted? With this you get the following output: https://pastebin.com/ns4vvpB3 
(note that in terminal the lines with the example that aren't indented are supposed to continue)

YangHee-Min:
> >>I am wondering wether we should create a class with generic object names to display as "metavar" to make it more explicit to the user what input is expected. For example:
> >
> >I think this is a great idea to improve the clarity of the help display!
> 
> Actually, we should use enum instead of creating an empty class

Here are a set of commit where I have applied the use of enum along with the help output:

sct_analyze_lesion: Commit [5cc9624](https://github.com/neuropoly/spinalcordtoolbox/commit/5cc962482fddb25d1022ff4762f4055fa4a28dde)
Output: https://pastebin.com/zmp1Tde3

sct_analyze_texture: Commit [52eb47e](https://github.com/neuropoly/spinalcordtoolbox/commit/52eb47efeabc64a2e9d1c998a8c3d49f71d12a84)
Output: https://pastebin.com/5gS8uzfZ

sct_apply_transfo: Commit [10234b8](https://github.com/neuropoly/spinalcordtoolbox/commit/10234b8f26121599c9065b9caa439271a3cde8b7)
Output: https://pastebin.com/BnzgRFfD

sct_compute_ernst_angle: Commit [7bd5e81](https://github.com/neuropoly/spinalcordtoolbox/commit/7bd5e814e5f0bbbc6ca4175920ac5c8ac287b451)
Output: https://pastebin.com/MnGSN4DY

sct_compute_hausdorff_distance: Commit [580d2bd](https://github.com/neuropoly/spinalcordtoolbox/commit/580d2bd4d023ecf823568798cab10c49f52a891a)
Output: https://pastebin.com/c6E7qGEy

sct_compute_mscc: Commit [be73246](https://github.com/neuropoly/spinalcordtoolbox/commit/be73246cc984b79fc29107b8ac80493a0ebec994)
Output: https://pastebin.com/itDRt4Fi

sct_compute_mtr: Commit [210998c](https://github.com/neuropoly/spinalcordtoolbox/commit/210998c8f035b42beb8ea616b13d12b1e5ef58ab)
Output: https://pastebin.com/QFpnzR4K

sct_compute_mtsat: Commit [e650ec7](https://github.com/neuropoly/spinalcordtoolbox/commit/e650ec795d29d0082b79f43b6a41fdb565a4b088)
Output: https://pastebin.com/2wxSvKVy

sct_compute_snr: Commit [d0db99d](https://github.com/neuropoly/spinalcordtoolbox/commit/d0db99d4f06858650f69646faa42ef8335159cee)
Output: https://pastebin.com/QpNLdiHy

sct_concat_transfo: Commit [439ad97](https://github.com/neuropoly/spinalcordtoolbox/commit/439ad979d584ca08798557d55810b50a8e88d519)
Output: https://pastebin.com/xF8H9Q52

sct_convert: Commit [b27ad86](https://github.com/neuropoly/spinalcordtoolbox/commit/b27ad86fd71950e6fa90df8b8d482d267d41649b)
Output: https://pastebin.com/CX3nfyQk

sct_create_mask: Commit [574bdf7](https://github.com/neuropoly/spinalcordtoolbox/commit/574bdf7b4c19bd7eec34c5157efef7b7aaecde05)
Output: http://pasted.co/f6a69878

sct_crop_image: Commit [9107139](https://github.com/neuropoly/spinalcordtoolbox/commit/9107139f58e22a07868f42c1b640129bac77bfb8)
Output: http://pasted.co/34c8f1ed

Please feel free to suggest changes (for example I am not sure if `<str>` is a good argument label for `-o` type arguments).

zougloub:
@YangHee-Min I think the code can get a bit more "elegant" by using something like:

```py
#!/usr/bin/env python

from enum import Enum

class Metavar(Enum):
	File = "<some file>"
	def __str__(self):
		return self.value

if __name__ == "__main__":
	import argparse

	parser = argparse.ArgumentParser(
	 description="Test",
	)

	parser.add_argument("--file",
	 metavar=Metavar.File,
	)

	try:
		import argcomplete
		argcomplete.autocomplete(parser)
	except:
		pass

	args = parser.parse_args()
```

Adding `__str__()` to the enum class allows you to not have to use `.value` everywhere.

scripts/sct_analyze_lesion.py
```diff
@@ -485,47 +504,41 @@ def ifolder2tmp(self):
...
+def main(arguments):
 
     # set param arguments ad inputted by user
-    fname_mask = arguments["-m"]
+    fname_mask = arguments.m
 
     # SC segmentation
-    if '-s' in arguments:
-        fname_sc = arguments["-s"]
+    if vars(arguments)["s"] != None:
```
zougloub: You can use `arguments.s is not None`. Specifying an argument has it default to None.

scripts/sct_analyze_texture.py
```diff
@@ -22,68 +22,81 @@
 import sct_utils as sct
 import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
-from msct_parser import Parser
+import argparse
```
zougloub: argparse is in the Python standard lib -> import it in the first block of imports. Order is stdlib / 3rd party / internal.

scripts/msct_types.py
```diff
@@ -852,3 +853,11 @@ def compare_centerline(self, other, reference_image=None):
         std = np.std(distances)
         max = np.max(distances)
         return mse, mean, std, max, distances
+
+class flagTypes(Enum):
```
jcohenadad: we are trying to get away from having importable libraries inside `scripts/`, so i would suggest to move this inside `spinalcordtoolbox/utils.py`

scripts/sct_apply_transfo.py
```diff
@@ -267,40 +271,34 @@ def apply(self):
 
 # MAIN
 # ==========================================================================================
-def main(args=None):
-
-    # check user arguments
-    if not args:
-        args = sys.argv[1:]
+def main(arguments):
```
zougloub: Usually a `main()` function receives command-line arguments but not digested argparse arguments. So I would recommend to do the parsing inside main. You can then call another function that receives the args.

YangHee-Min:
> @YangHee-Min I think the code can get a bit more "elegant" by using something like:

Oh wow looks much better! Changes applied in the following commits: 

Definition of Enum class `Metavar`: Commit [16a7a8f](https://github.com/neuropoly/spinalcordtoolbox/commit/16a7a8f58cb839af347fb932715cdf31ca1041c0)

scripts/msct_types.py
YangHee-Min: Ok sounds good! Changes applied in commit [16a7a8f](https://github.com/neuropoly/spinalcordtoolbox/commit/16a7a8f58cb839af347fb932715cdf31ca1041c0) (addition of Enum class `Metavar` to `spinalcordtoolbox/utils.py`) and removed it from `msct_types` in commit [96d4f3f](https://github.com/neuropoly/spinalcordtoolbox/commit/96d4f3febe741022197f44f5c83c976a4e3d417e).

scripts/sct_analyze_texture.py
YangHee-Min: Ok sounds good! Changed in commit [54d0476](https://github.com/neuropoly/spinalcordtoolbox/commit/54d04763befd566da8b96f2ee6e80071a4de08d9)

YangHee-Min:
> Please comment with docstrings. E.g.:
"""This class is used to display intuitive input types via the metavar field of argparse"""

Ok sounds good! Added in commit [d347f47](https://github.com/neuropoly/spinalcordtoolbox/commit/d347f47dbc004d9315865bc2537a65b6ff638301)

scripts/sct_apply_transfo.py
YangHee-Min: Ok sounds good. Does something along [these lines](https://github.com/neuropoly/spinalcordtoolbox/blob/d19a0813a896230d601a4f9eaec5dca07e43ad5b/scripts/sct_analyze_lesion.py#L507-L509) follow what you had in mind?

scripts/sct_analyze_lesion.py
```diff
@@ -485,47 +504,43 @@ def ifolder2tmp(self):
...
+    arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
 
     # set param arguments ad inputted by user
-    fname_mask = arguments["-m"]
+    fname_mask = arguments.m
 
     # SC segmentation
-    if '-s' in arguments:
-        fname_sc = arguments["-s"]
+    if arguments.s is not None:
```
jcohenadad: @YangHee-Min do you need to add this line, given that is always a default value when we declare the parser fields?

YangHee-Min: @jcohenadad If there is a default I believe it is not necessary. However, in the case for this argument as seen [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/ee58d326fcda797a947dff823d0e106542680e05/scripts/sct_analyze_lesion.py#L58-L63), there is no default. Therefore I believe this was necessary due to there being no default value. 

I will look into it and do some tests related to default values and whatnot just to see how the commands behave.

jcohenadad: > @jcohenadad If there is a default I believe it is not necessary. However, in the case for this argument as seen [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/ee58d326fcda797a947dff823d0e106542680e05/scripts/sct_analyze_lesion.py#L58-L63), there is no default. Therefore I believe this was necessary due to there being no default value.
> 
> I will look into it and do some tests related to default values and whatnot just to see how the commands behave.

for the example you gave, you can simply move [this declaration](https://github.com/neuropoly/spinalcordtoolbox/blob/ee58d326fcda797a947dff823d0e106542680e05/scripts/sct_analyze_lesion.py#L521) in the parser definition (as opposed to in the main)-- it will save you a lot of lines of code.

YangHee-Min: Does something along the lines of this commit (Commit [fe41bbe](https://github.com/neuropoly/spinalcordtoolbox/commit/fe41bbeed8aafce11c83176d296af2bd857c3985)) do the trick?

jcohenadad:
@YangHee-Min to follow-up on our conversation this morning, the missing argument in the `main()` of all SCT function is breaking when the main is called with arguments. See this commit a8a0c44.

I will now fix this in all the modified functions of this PR

jcohenadad:
@YangHee-Min I identified a major issue related to the way we used to treat multiple inputs. If you look at the [previous and current](https://github.com/neuropoly/spinalcordtoolbox/pull/2315/files#diff-150e29f70efce4a35502abbf266f5e66) versions of `sct_apply_transfo`, the flag -w used to accept multiple inputs separated with comma (","). In the current implementation, this is no more possible, make for example `sct_register_to_template` to break.

One approach I started to investigate is to use `nargs='+'`, which enables multiple input files after a flag. However, two issues:
- It breaks retro-compatibility (not a big deal since we move to major release)
- It makes it more difficult to declare inverse transformation: currently we use a prefix "-" to the file name, to declare this is an inverse transformation. Now, if we separate with space (not comma), the "-" will be interpreted as a flag.

Suggestions welcome...

YangHee-Min:
> One approach I started to investigate is to use nargs='+', which enables multiple input files after a flag. However, two issues:
> 
> It breaks retro-compatibility (not a big deal since we move to major release)
> It makes it more difficult to declare inverse transformation: currently we use a prefix "-" to the file name, to declare this is an inverse transformation. Now, if we separate with space (not comma), the "-" will be interpreted as a flag.
> Suggestions welcome...

@jcohenadad Maybe we could approach this problem using custom actions as seen [in the documentation for argparse under action classes ( "__ call __" method)](https://docs.python.org/3/library/argparse.html#name-or-flags). With this approach, maybe we could make an action that stores all arguments while ignoring all "-" until an existing flag is detected? (This might not work depending on the way argparse reads the input).

This is a mere suggestion but maybe something along these lines could work.

jcohenadad:
I like the idea of the "action" flag but I don't think this is going to work because the parser will immediately interpret the "-" as a potential flag (before entering the "action" routine). 

Other possibilities: 
- instead of calling the file "-file.ext", we call it "file.ext_" so there is no interaction with the parser. The disadvantage is that we could not elegantly parse this file using Python's internal file checking routines.
- we introduce the flag `-winv` to specify that a transformation should be inverted, and we use positional arguments with `-w/winv`. Example: 
~~~
sct_apply_transfo -i IMAGE -s SEG -w TRANSFO1 -w TRANSFO2 -winv TRANSFO3(INV) -w TRANSFO 4
~~~

jcohenadad:
@YangHee-Min We are getting there! Old tests have been re-implemented and they all pass :-) I'm doing a final pass across all functions and we should be able to merge soon

jcohenadad:
`batch_processing.sh` results for commit 931b77a6a54f3f7a4c75aacc89292d5a7b04d4fe:
~~~
t2/CSA:          78.47504823863535
mt/MTR(WM):      54.16241055860758
t2s/CSA_GM:      12.57431836855896
t2s/CSA_WM:      65.41108245601296
dmri/FA(CST_r):  0.789334198512836
dmri/FA(CST_l):  0.7565620337219434
~~~

jcohenadad:
@YangHee-Min I noticed that `sct_concat_transfo` field `-w` was not updated appropriately, as done in `sct_apply_transfo`, and will break as we remove the system call (i.e. call `sct_concat_transfo.main(args)` vs. `sct.run('sct_concat_transfo')`. Could you please open a new PR, to continue the refactoring with argparse, and make sure to include this function as well (mention my comment so we don't forget).

YangHee-Min:
> @YangHee-Min I noticed that sct_concat_transfo field -w was not updated appropriately, as done in sct_apply_transfo, and will break as we remove the system call (i.e. call sct_concat_transfo.main(args) vs. sct.run('sct_concat_transfo'). Could you please open a new PR, to continue the refactoring with argparse, and make sure to include this function as well (mention my comment so we don't forget).

@jcohenadad Do we want to convert the rest of the functions? In other words, which functions do we want to convert from msct_parser to argparse next?

jcohenadad:
> > @YangHee-Min I noticed that sct_concat_transfo field -w was not updated appropriately, as done in sct_apply_transfo, and will break as we remove the system call (i.e. call sct_concat_transfo.main(args) vs. sct.run('sct_concat_transfo'). Could you please open a new PR, to continue the refactoring with argparse, and make sure to include this function as well (mention my comment so we don't forget).
> 
> @jcohenadad Do we want to convert the rest of the functions? In other words, which functions do we want to convert from msct_parser to argparse next?

how about the next 10 ones?

PR #2328, opened 2019-06-27
(WIP) Add automatic modality detection

bsauty:
### 2020-12-05 update
The old working branch (`bs/modality_detection`) is not in a usable state because of how greatly it has diverged from `master`. It's been left alone, but the base branch has been switched to a temporary base branch (instead of `master`) so that GitHub's UI will [display only the new commits](https://github.com/neuropoly/spinalcordtoolbox/pull/2328/commits). This is only for show, so that people in the future can more easily refer to the past work. 

<details>
<summary> Original issue description </summary>

---

> ⚠️ Do not merge yet ⚠️ 
> 
> 
> This folders contains the implementation of the automatic modality classifier. To repo for the training of the model can be found [here](https://github.com/bsauty/sc_mri_modality_detection).
> 
> # Motivations
> - Automatic contrast detection for subsequent segmentation or labeling task, which could be useful in big datasets without information about the contrast.
> 
> # TODO
> - Upload the model on [OSF](https://osf.io/jyp2z/) 
> - Implement unit tests
> 
> 
> # DONE
> - Document code and cleanup according to PEP8 standard, including the training repo
> - Implement loading the acquisition using the image module
> - Add documentation about the specificities of dwi contrast
> - Upload the model on [wiki](https://www.neuro.polymtl.ca)
> - Add resampling and padding to avoid issues with resolution 
> 
> And add automatic contrast detection for :
> - sct_label_vertebrae
> - sct_detect_pmj
> - sct_deepseg_sc
> - sct_get_centerline (with optic)
> - sct_propseg
> - sct_deepseg_lesion
> - sct_registration_to_template

</details>


Nicolas-Pinon:
Writing down a discussion we had with @bsauty about DWI : 

Why is it necessary to have DWI "detection", the DWI file unlike other contrast should be 4D instead of 3D right ? Which makes it easy to "detect". (Not 100% sure about that, that's why I'm asking)

jcohenadad:
The input is typically dwi_mean, which is the average of all DW data across time (hence 3D)

bsauty:
Okay so I implemented most of the functions but I still have a few pending questions and observations : 

1. `sct_deepseg_lesion` allows to choose from `t2ax` and `t2` for the T2w, should I include a specific check in the header to detect if it is indead axial. If so what slice thickness makes a good discriminator ?
2. `sct_propseg`, `sct_deepseg_sc` and `sct_get_centerline` allow to chose `dwi`. It should normally be detected as `t1` by the network.
3. For `sct_label_vertebrae` and `sct_detect_pmj` I let the network output `t1`, `t2` or `t2s` and then I map `t2s` to `t2` since there are only two available contrasts.
4. Maybe I should add a test to check in the filename to make sure that the contrast is not specified in the name of the acquisition. That sounds a bit silly but maybe it would be wise to add a "safeguard" like this.

Do not bother reviewing the PR for now, there are still a lot of things to cleanup and to test so it would be wasting time !

jcohenadad:
> 1. `sct_deepseg_lesion` allows to choose from `t2ax` and `t2` for the T2w, should I include a specific check in the header to detect if it is indead axial. If so what slice thickness makes a good discriminator ?

this function is not as utilized as the others, so i would suggest to leave it as it is and not develop an additional layer on top of this function (will make the code heavier, more maintenance load, etc.).

> 2. `sct_propseg`, `sct_deepseg_sc` and `sct_get_centerline` allow to chose `dwi`. It should normally be detected as `t1` by the network.

which is a problem (should be detected as dwi). But it's fine since it represent <5% of the use case. If people have dwi to segment, they will simply use the `-c dwi` mode (we will document + add in batch_processing).

> 3. For `sct_label_vertebrae` and `sct_detect_pmj` I let the network output `t1`, `t2` or `t2s` and then I map `t2s` to `t2` since there are only two available contrasts.

good.

> 4. Maybe I should add a test to check in the filename to make sure that the contrast is not specified in the name of the acquisition. That sounds a bit silly but maybe it would be wise to add a "safeguard" like this.

hum... not a big fan of adding such a layer in SCT code (user-specific, heavier code, more maintenance, etc.)

> Do not bother reviewing the PR for now, it would be wasting time since it is not ready at all ! 🔜

👍

scripts/sct_detect_pmj.py
```diff
@@ -40,9 +41,9 @@ def get_parser():
                         example="t2.nii.gz")
     parser.add_option(name="-c",
                         type_value="multiple_choice",
-                        description="type of image contrast, if your contrast is not in the available options (t1, t2), use t1 (cord bright / CSF dark) or t2 (cord dark / CSF bright)",
-                        mandatory=True,
-                        example=["t1", "t2"])
+                        description="type of image contrast, automatically detected but can be specified manually (t2: cord dark / CSF bright ; t1: cord bright / CSF dark)",
+                        mandatory=False,
+                        example=["t1", "t2", "t2s"])
```
charleygros: We do not have a model for detecting the pmj on T2star images (which is almost impossible on these images and probably never requested by the user). I would suggest to remove `t2s`. But I am maybe missing something :)

scripts/sct_detect_pmj.py
```diff
@@ -242,9 +243,16 @@ def main(args=None):
...
     fname_in = arguments["-i"]
-    contrast = arguments["-c"]
+
+    # Automatic detection of contrast
+    if not "-c" in arguments:
+        contrast_type = modality_detection.classify_from_path(fname_in)
+    else:
+        contrast_type = arguments["-c"]
+    contrast_type_conversion = {'t1': 't1', 't2': 't2', 't2s': 't2'}
+    contrast = contrast_type_conversion[contrast_type]
```
charleygros: Great!!!

scripts/sct_download_data.py
```diff
@@ -115,7 +116,8 @@ def main(args=None):
         'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
                               'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip'],
         'c2c3_disc_models': ['https://osf.io/t97ap/?action=download',
-                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip']
+                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip'],
+        'modality_pred_model': ['https://www.neuro.polymtl.ca/_media/downloads/sct/20190706_modality_pred_model.pt.zip']
```
charleygros: I have not checked on the wiki yet, but: Did you put your model into a folder? or did you directly zip the model?

scripts/sct_label_vertebrae.py
```diff
@@ -75,9 +76,9 @@ def get_parser():
                       example="t2_seg.nii.gz")
     parser.add_option(name="-c",
                       type_value="multiple_choice",
-                      description="type of image contrast, t2: cord dark / CSF bright ; t1: cord bright / CSF dark",
-                      mandatory=True,
-                      example=['t1', 't2'])
+                      description="type of image contrast, automatically detected but can be specified manually (t2: cord dark / CSF bright ; t1: cord bright / CSF dark)",
+                      mandatory=False,
+                      example=['t1', 't2', 't2s'])
```
charleygros: Same comment as for `sct_detect_pmj`.

spinalcordtoolbox/modality_prediction/README.md
```diff
@@ -0,0 +1,2 @@
+This folders contains the implementation of the automatic modality classifier. To repo for the training of the model
```
charleygros: Great!!

spinalcordtoolbox/modality_prediction/core.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # We load the model
+    model = M.Classifier()
+    model_path = os.path.join(sct.__sct_dir__, 'data', 'modality_pred_model.pt')
+    model.load_state_dict(torch.load(model_path, map_location='cpu'))
+    model.eval()
+
+    modality = classify_acquisition(input_image, model)
+    
+    return(modality)
```
charleygros: You did really great Benoît!

scripts/sct_download_data.py
charleygros: (I would recommend to add the model into a folder and to zip the folder)

bsauty: Ok I can do that! Why though ?

scripts/sct_detect_pmj.py
jcohenadad: > I would suggest to remove t2s.

agreed

scripts/sct_download_data.py
jcohenadad: just to follow the convention of other models, which are enclosed into function-specific folder under **data/** (e.g., deepseg_gm_models/, deepseg_lesion_models/, gm_model/, etc.)

scripts/sct_label_vertebrae.py
jcohenadad: https://github.com/neuropoly/spinalcordtoolbox/pull/2328#issuecomment-508770873

scripts/sct_detect_pmj.py
charleygros: I agree with "I let the network output t1, t2 or t2s and then I map t2s to t2 since there are only two available contrasts." (quotation from the discussion). But, please correct me if I am missing something: we are here talking about the parser, we do not want the user to input `t2s` since we do not have a model for these images. Moreover, if the user specifies the argument `-c`, then the network will be bypassed.

spinalcordtoolbox/modality_prediction/core.py
```diff
@@ -0,0 +1,118 @@
...
+    input_image = Image(input_path)
+    input_image.change_orientation('RPI')
+
+    # We load the model
+    model = M.Classifier()
+    model_path = os.path.join(sct.__sct_dir__, 'data', 'modality_pred_model.pt')
+    model.load_state_dict(torch.load(model_path, map_location='cpu'))
+    model.eval()
+
+    modality = classify_acquisition(input_image, model)
```
charleygros: I would recommend to print the result of the automatic modality detection at this point.
Could be very helpful for the debugging etc.
Something like:
```
logger.info('Modality detected: {}'.format(modality))
```

scripts/sct_detect_pmj.py
bsauty: Exact I will delete the option t2s !

charleygros:
Quite sorry to play the annoying guy:
I tried it on `sct_testing/large/milan_2017-filippi_004/t2_sag`.

Herebelow the output:
```
sct_detect_pmj -i t2_sag.nii.gz 

--
Spinal Cord Toolbox (git-master-4a1795fb75b8367a3cdf5c7ea5c1896c6d408802*)

  File "/home/charley/sct/scripts/sct_detect_pmj.py", line 309, in <module>
    main()
  File "/home/charley/sct/scripts/sct_detect_pmj.py", line 251, in main
    contrast_type = modality_detection.classify_from_path(fname_in)
  File "/home/charley/sct/spinalcordtoolbox/modality_prediction/core.py", line 115, in classify_from_path
    modality = classify_acquisition(input_image, model)
  File "/home/charley/sct/spinalcordtoolbox/modality_prediction/core.py", line 76, in classify_acquisition
    acq.CenterCropTransform()
  File "/home/charley/sct/spinalcordtoolbox/modality_prediction/core.py", line 47, in CenterCropTransform
    raise RuntimeError("Negative crop.")
RuntimeError: Negative crop.
```

scripts/sct_detect_pmj.py
jcohenadad: good point!

jcohenadad:
@bsauty to resample any 3D matrix to a fixed matrix size (with zero-padding): https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_sc/core.py#L86

charleygros:
It can also be useful:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_gm/deepseg_gm.py#L81

bsauty:
@charleygros  any idea as to why the build fails for ubuntu ? I am not very familiar with this kind of error. For instance I have no idea how to fix the issue for the ubuntu build "job exceeds maximum log length"..

charleygros:
> @charleygros any idea as to why the build fails for ubuntu ? I am not very familiar with this kind of error. For instance I have no idea how to fix the issue for the ubuntu build "job exceeds maximum log length"..

It looks like it is due to a bug of the progress bar for the `pip` installation:
- https://travis-ci.community/t/large-pip-installs-exceed-maximum-log-length/1599/4
- https://github.com/pypa/pip/issues/6101

My understanding is that using `--progress-bar off` with `pip` could help while installing `torch`.

bsauty:
> > @charleygros any idea as to why the build fails for ubuntu ? I am not very familiar with this kind of error. For instance I have no idea how to fix the issue for the ubuntu build "job exceeds maximum log length"..
> 
> It looks like it is due to a bug of the progress bar of the `pip` installation:
> 
> * https://travis-ci.community/t/large-pip-installs-exceed-maximum-log-length/1599/4
> * [pypa/pip#6101](https://github.com/pypa/pip/issues/6101)
> 
> My understanding is that using `--progress-bar off` with `pip` could help while installing `torch`.

Ok thank you (also sorry I could have investigated it myself..) I will correct it but I guess it will disable progress bar for all requirements installation, is it a concern ?

bsauty:
To solve building issues in Travis, I have to add `import torch` in the begging of `sct_deepseg_lesion` and `sct_deepseg_sc` or it fails because of an intrinsic problem of PyTorch. Besides I still have an issue with ubuntu 14.01 that can be solved according to this [issue](https://github.com/r9y9/gantts/issues/14). Do you reckon there is an easy way to implement this only for the appropriate os ?

charleygros:
> Besides I still have an issue with ubuntu 14.01 that can be solved according to this [issue](https://github.com/r9y9/gantts/issues/14). Do you reckon there is an easy way to implement this only for the appropriate os ?

Same issue / solution mentioned [here](https://github.com/pytorch/pytorch/issues/10567) for a project with `pytorch`.

bsauty:
Okay so to sum up what I had to do to add PyTorch in the requirements to solve building issues: 
- Delete the progress bar because it breaks ubuntu installation with unknown caracters : see [issue](https://github.com/pypa/pip/issues/6101)
- Add `import torch` in the beginning of `sct_deepseg_lesion` and `sct_deepseg_sc` because if not imported first, it breaks the installation : see [issue](https://github.com/r9y9/deepvoice3_pytorch/issues/28)
- Add a specific export for Ubuntu14.04 because memory allocation is faulty and leads to a seg fault with PyTorch : see [issue](https://github.com/r9y9/gantts/issues/14)
 `sudo apt-get install libtcmalloc-minimal4`
 `export LD_PRELOAD="/usr/lib/libtcmalloc_minimal.so.4"`
I am currently trying to add it in the bash profile to see if it solves the issue.

install_sct
```diff
@@ -537,6 +537,13 @@ if [[ $UID != 0 ]]; then
   if [[ ${OS} == "osx" ]]; then
     sed -ie '/sct_env/ s/^#*/#/' ${THE_RC}
   else
+    # For Ubuntu-14.04 we need to export a path to make sure the build does not fail with pytorch
+    if uname -a | grep 14.04; then
+        sudo apt-get install libtcmalloc-minimal4
+        echo "" >> ${THE_RC}
+        echo "export LD_PRELOAD='/usr/lib/libtcmalloc_minimal.so.4'" >> ${THE_RC}
```
bsauty: This seems to have failed, maybe I am not exporting it into the right bash profile ? I noticed there are other adresses like `${THE_BASHRC}` but I am not sure if I placed it correctly. Any thoughts about this @jcohenadad or @charleygros ? Besides it is weird because the built fails right at the end after having successfully tested all functions...

jcohenadad: @bsauty can you update the bash profile [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/install_sct#L91) (to cover both bash and t/csh)

bsauty:
I am back to civilisation ! However I still do not know how to solve the issue with pytorch installation for ubuntu 14 because the solutions I found on the internet do not work here... Any idea ?

jcohenadad:
> I am back to civilisation ! However I still do not know how to solve the issue with pytorch installation for ubuntu 14 because the solutions I found on the internet do not work here... Any idea ?

quite tricky... i would need to do some tests (unfortunately my time is limited for the next few weeks). @gmotzespina any chance you could look into this?

gmotzespina:
> > I am back to civilisation ! However I still do not know how to solve the issue with pytorch installation for ubuntu 14 because the solutions I found on the internet do not work here... Any idea ?
> 
> quite tricky... i would need to do some tests (unfortunately my time is limited for the next few weeks). @gmotzespina any chance you could look into this?

Sure I can start checking it out.

joshuacwnewton:
Following https://github.com/neuropoly/spinalcordtoolbox/issues/2984#issuecomment-738435486, the base branch of this PR has been changed, and the PR has been closed. This was done to preserve the history of the work in this PR.

(Before, GitHub's UI showed that the PR wanted to merge ~7000 commits, making the work unreadable. This happened because the history of `master` had been rewritten at some point.)

jcohenadad:
@joshuacwnewton i would keep this one open-- an excellent work has been done by @bsauty on this, and it was almost ready to be merged (too bad the branch diverged so much...).

by doing a quick search on the modified files with the keyword "modality" i am able to find the original contribution specific to this branch. It would probably be ~30min of work to figure out the specific code that needs to be recovered. I would rather have these 30min spent rather that closing this for good (which would be ~2months of work for nothing)

oops... but for some reasons i cannot reopen that branch...?  so maybe we can open an issue that would deal with this code recovery specifically...?

joshuacwnewton:
@jcohenadad I very much agree that the work should not be thrown out. That's why, even though this PR has been closed, I've left the branch `bs/modality_detection` open. So, the work itself should still be available. (e.g. by running `git checkout bs/modality_detection` locally).

joshuacwnewton:
I've clarified the issue description a bit to better explain what I did to this PR. (I haven't made any changes to the actual work -- only cosmetic changes to the GitHub side of things.)

jcohenadad:
Closing (for now), until it becomes a relevant issue again at some point

PR #2218, opened 2019-04-09
Refactoring to allow use of im AND seg in the registration process

Nicolas-Pinon:
For now, in the registration process (namely in `msct_register`) it is not possible to implement a new method that uses both the image and the segmentation of the source and destination MRI. The contributor is forced to either develop a method that register based on segmentations or on images but cannot use both. This can be easily understood when looking at the parameter `type` which must be equal to `im` or `seg` (or `label`).

In this PR we introduce the possibility to use both `im` and `seg`, without breaking compatibility, which will allow further development of registration methods, in particular a method based on the symmetry of the SC, discussed in #2204 and #836.

PR #2331, opened 2019-07-02
New method for detecting rotation in centermassrot

### LINKED ISSUE #2204: Account for rotation in default settings
We should re-introduce centermassrot (or alternative algorithm) in `sct_register_to_template`, because in some cases, e.g. here for the London course, rotation of the cord is obvious and should be accounted for:

anim

@Nicolas-Pinon could you open a PR on your current investigations so we can start cross-referencing all related issues

Nicolas-Pinon:
The main purpose of this PR is to introduce a new method to detect the rotation of the spinal cord. Previous work in #2218 has already been done to prepare the ground to integrate this new method that uses the image to detect the rotation and not the segmentation. 

**Main changes:**
- New param `rot_method` for` sct_register_multimodal` and `sct_register_to_template`, which is active when using `algo=centermassrot`.
- Implemented `rot_method='hog'` to estimate rotation
- Implemented `rot_method='pcahog'`, which combines the PCA method, followed by the HOG method is PCA fails.
- Implemented a QC method for rotation detection by prompting the image, segmentation and the list of angles. See: `spinalcordtoolbox.reports.qc.QcImage.line_angle()`
- Refactored `qc.py` and `slice.py` to accommodate new `line_angle()` QC.
- Regularization of angle along z in centermassrot now uses Gaussian filtering instead of polynomial fitting (more robust)
- Angle searching in all methods is now restrained to a certain range: parameter `register2d_centermassrot(th_max_angle=)`
- Massive refactoring: moved multi-step code into a new function `msct_register.register_wrapper()` and call this function inside `sct_register_multimodal` and `sct_register_to_template` (lots of duplicated code was removed).
- Added `filter_size` parameter in Paramreg, this parameter influence the regularization for centermassrot (renaming of `poly` param but did not deleted because was used elsewhere)
- Added the possibility to fix the angle of the destination image to 0 deg, in case we **know** this is the template. Implemented that feature by default in `sct_register_to_template`.
- Use 'pcahog' method as the new default.

**Points to discuss :**

1. There is now an a priori range of angle to search for, what value do we set it for ? Do we allow the user to set it ? Is this a good idea ?
--> let's set it to 40. Yes, we could let the user change it.
2. Do we integrate the rotation QC somewhere in the pipeline or do we let it as a development tool for now ?
--> we leave it for dev
3. I have added functions in `msct_register`, maybe they are duplicates of others that are already present in SCT (circular convolution and circular filtering for instance)
--> i've reworked some of the refactoring. 
4. Strictly personal opinion : I would suggest to remove the parameter `pca_eigenratio_th`, I don’t think it is often used and it adds to the already many parameters of the registration
--> the same could be said for `th_max_angle`. I think we should keep it.
5. Putting the auto method as default instead of PCA
--> agreed.

Fixes #2204

Further changes are needed that can be addressed in a subsequent PR, see https://github.com/neuropoly/spinalcordtoolbox/issues/2547


Nicolas-Pinon:
This PR is ready (I think) to be integrated to the master branch. (Sorry for the delay !)

jcohenadad:
for the records, results of batch_processing are:

**nipin/issue836-detect-rotation *223a43d2c2addc657ffd67d0cd5bbf0401fb8100)**
~~~
mt/MTR(WM):      54.364850159765645
dmri/FA(CST_r):  0.794761331925753
dmri/FA(CST_l):  0.7726672458856273
~~~

**v4.1.1**
~~~
mt/MTR(WM):      54.364850159765645
dmri/FA(CST_r):  0.7947429202929487
dmri/FA(CST_l):  0.7726948674392992
~~~

jcohenadad:
@Nicolas-Pinon as you can see I made quite a few changes to your original PR. Let me know if you agree with these changes before merging

Nicolas-Pinon:
@jcohenadad after looking at the changes, I'm completely ok with them :+1: 
Just one small question though, why removing the figure generation with hog only (commit 2d4b2a1ae6f5556581968dac8a908df840426f21) ? Was there a problem with it ?
I'll let you merge ?
Thanks again

jcohenadad:
> Just one small question though, why removing the figure generation with hog only (commit [2d4b2a1](https://github.com/neuropoly/spinalcordtoolbox/commit/2d4b2a1ae6f5556581968dac8a908df840426f21)) ? Was there a problem with it ?

this figure shows the angles of the src and dest images calculated from the PCA only. So, by including the hog method in this figure generation, users/developers would be misled as to what this figure actually represents.

PR #2366, opened 2019-07-27
Support zsh configuration.

gmotzespina:
Adds support for `zsh`. In case it is installed, it adds the required env variables to the `.zshrc` file.

Addresses: #2367


gmotzespina:
> looks good to me, although i have not tested.
> 
> minor comment: the variable `SHELL_PATH` could maybe be renamed? it suggests to point to the path of the shell, although it points to the "run commands" (or "rc") file.

Good point, I'll rename it to something more intuitive.

gmotzespina:
@jcohenadad I've updated the name to `RC_FILE_PATH` how does it sounds to you?

PR #2371, opened 2019-07-31
Argparse refactoring of scripts that break retro-compatibility

### LINKED ISSUE #2342: argparse refactoring that will break retro-compatibility
While parser refactoring is underway (see #1548), we need to urgently perform the long overdue release of 4.0.0 (major release). Once 4.0.0 is released, only patches and retro-compatible changes will be allowed. Hence, changes in CLIs caused by the parser refactoring will break retro-compatibility and should therefore be performed **before** this major release.

The purpose of this issue is to identify and refactor the functions where this refactoring will change the CLI syntax.

The following files will cause problem:
- `sct_maths` flag `-add` and `-mul`: the function `get_data_or_scalar()` uses the old parser. I suggest that instead of listing multiple files separated with "," we separate them with space and use argparse to fetch individual file names as a list.
- `sct_merge_images`: like before: files should be separated by space, not comma (flags -i, -w)
- `sct_straighten_spinalcord`: replace _ with - for some flags (ldisc_input, ldisc_dest, speed_factor, xy_size
- `sct_image`: has -i flag that requires files to be seperated by spaces rather than commas

YangHee-Min:
At the moment, certain functions break retro-compatibility with the conventions of the other scripts that will be coming with the new release.

In this PR, we refactor the following functions to use Python's argparse:
- `sct_maths` flag -add and -mul: the function get_data_or_scalar() uses the old parser. I suggest that instead of listing multiple files separated with "," we separate them with space and use argparse to fetch individual file names as a list.
- ` sct_image`: has -i flag that requires files to be seperated by spaces rather than commas
- `sct_merge_images`: like before: files should be separated by space, not comma (flags -i, -w)
- `sct_straighten_spinalcord`: replace _ with - for some flags (ldisc_input, ldisc_dest, speed_factor, xy_size

Fixes #2342


scripts/sct_image.py
```diff
@@ -31,169 +30,191 @@ def get_parser():
...
+                    'Inputs can be a number, a 4d image, or several 3d images separated with ","',
+        formatter_class=SmartFormatter,
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py"))
+    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatory.add_argument(
+        "-i",
+        nargs="+",
+        metavar=Metavar.file,
+        help='Input file(s). If several inputs: separate them by a coma without white space. Example: "data.nii.gz"',
```
jcohenadad: separation should be with space, not comma (see refactoring or previous functions)

jcohenadad: note: this is exactly the reason we need to refactor this before 4.0.0 (this is where retrocompat breaks)

scripts/sct_image.py
```diff
@@ -525,10 +546,10 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
 
 def multicomponent_split(im):
     """
-    Convert composite image (e.g., ITK warping field, 5dim) into several 3d volumes.
-    Replaces "c3d -mcs warp_comp.nii -oo warp_vecx.nii warp_vecy.nii warp_vecz.nii"
-    :param im:
-    :return:
+        Convert composite image (e.g., ITK warping field, 5dim) into several 3d volumes.
```
jcohenadad: any reason for adding this indentation?

YangHee-Min: Nope! Changed in commit [4af3ca2](https://github.com/neuropoly/spinalcordtoolbox/commit/4af3ca27dbf6fb625c94f2750e3ac8e8297c2530)

scripts/sct_image.py
YangHee-Min: Oops forgot to update the descirption! Changed in commit [831b857](https://github.com/neuropoly/spinalcordtoolbox/commit/831b857b434e67eb948b8cdc82d78fead6c070fe)

scripts/sct_straighten_spinalcord.py
```diff
@@ -15,235 +15,242 @@
...
-    parser.add_option(name='-centerline-smooth',
-                      type_value='int',
-                      description='Degree of smoothing for centerline fitting. Only use with -centerline-algo {bspline, linear}.',
-                      mandatory=False,
-                      default_value=10)
+    optional.add_argument(
+        "-param",
+        metavar=Metavar.list,
+        help="R|Parameters for spinal cord straightening. Separate arguments with ','."
+             "\nprecision: [1.0,inf[. Precision factor of straightening, related to the number of slices. Increasing this parameter increases the precision along with increased computational time. Not taken into account with hanning fitting method. Default=2"
```
jcohenadad: i would add a space for visual purpose and to be consistent with previous refactoring

note: the same applies for every other similar case of course

scripts/sct_straighten_spinalcord.py
```diff
@@ -15,235 +15,242 @@
...
-                      description="Parameters for spinal cord straightening. Separate arguments with ','."
-                                  "\nprecision: [1.0,inf[. Precision factor of straightening, related to the number of slices. Increasing this parameter increases the precision along with increased computational time. Not taken into account with hanning fitting method. Default=2"
-                                  "\nthreshold_distance: [0.0,inf[. Threshold at which voxels are not considered into displacement. Increase this threshold if the image is blackout around the spinal cord too much. Default=10"
-                                  "\naccuracy_results: {0, 1} Disable/Enable computation of accuracy results after straightening. Default=0"
-                                  "\ntemplate_orientation: {0, 1} Disable/Enable orientation of the straight image to be the same as the template. Default=0",
-                      mandatory=False)
+    optional.add_argument(
+        "-x",
+        help="Final interpolation.",
+        choices = ("nn", "linear", "spline"),
```
jcohenadad: as mentioned before, no space around "=" (PEP8)

scripts/sct_straighten_spinalcord.py
```diff
@@ -15,235 +15,242 @@
...
-        sc_straight.interpolation_warp = str(arguments["-x"])
-    if "-o" in arguments:
-        sc_straight.output_filename = str(arguments["-o"])
-    if '-ofolder' in arguments:
-        sc_straight.path_output = arguments['-ofolder']
-    else:
-        sc_straight.path_output = './'
-    path_qc = arguments.get("-qc", None)
-    verbose = int(arguments.get('-v'))
+    if arguments.r is not None:
```
jcohenadad: it's never None (see how you define it).

the same applies at many other places in your code

scripts/sct_straighten_spinalcord.py
YangHee-Min: So every time there is an instance of separating numbers (example: 2,3,4) we would now apply space?

scripts/sct_straighten_spinalcord.py
YangHee-Min: Oh my mistake. Changed in commit [d151dab](https://github.com/neuropoly/spinalcordtoolbox/commit/d151daba8828921ce84006da053bfbfb6694016c)

scripts/sct_straighten_spinalcord.py
YangHee-Min: Is there by chance a preferred default value for arguments.o? Other than that other changes applied in commit [343e697](https://github.com/neuropoly/spinalcordtoolbox/commit/343e697e733bed1064430c73fcc8d77e73d2fc41)

YangHee-Min:
@jcohenadad made a small change in `sct_straighten_spinalcord` in commit [af7211f](https://github.com/neuropoly/spinalcordtoolbox/pull/2371/commits/af7211f92da695cbbd9b4c0f3ca448935ced7916) to fix Travis fails. Can you give it a quick look and confirm your approval for the changes?

jcohenadad:
no, that doesn't look right. First of all, there is a discrepancy between your example and the default. Secondly: the script should not fail if the output name change

YangHee-Min:
> no, that doesn't look right. First of all, there is a discrepancy between your example and the default. Secondly: the script should not fail if the output name change

But in `sct_label_vertebrae`, as shown [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/af7211f92da695cbbd9b4c0f3ca448935ced7916/scripts/sct_label_vertebrae.py#L266), it is supposed to get its input file from either [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/af7211f92da695cbbd9b4c0f3ca448935ced7916/scripts/sct_label_vertebrae.py#L254) or [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/af7211f92da695cbbd9b4c0f3ca448935ced7916/scripts/sct_label_vertebrae.py#L256). In `sct_testing`, it seems to use the line: 
`sct_straighten_spinalcord.main(args=[
            '-i', 'data.nii',
            '-s', 'segmentation.nii',
            '-r', str(remove_temp_files),
            '-v', str(verbose),
        ])`
which had (before I pushed my changes with commit [af7211f](https://github.com/neuropoly/spinalcordtoolbox/pull/2371/commits/af7211f92da695cbbd9b4c0f3ca448935ced7916)) `data_straight.nii.gz` as default rather than `data_straight.nii`. However, as shown [(this line)](https://github.com/neuropoly/spinalcordtoolbox/blob/af7211f92da695cbbd9b4c0f3ca448935ced7916/scripts/sct_label_vertebrae.py#L266), we need input `data_straight.nii` rather than `data_straight.nii.gz`.  That's what I believe is causing the error message: 

`Command-line usage error: Option -i file doesn't exist: data_straight.nii
Aborted...`

When we do `sct_testing`. Wouldn't this be caused by a default output name change?

jcohenadad:
@YangHee-Min as you can see [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2371/files#diff-06511e50890a7fbffb44629b754dc3f7L116), the default value for output file was `''`, which you changed for something else [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2371/files#diff-06511e50890a7fbffb44629b754dc3f7R127). Now, the problem is that later, in the [SpinalCord Straightener class](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/straightening.py#L599), the variable `fname_output` gets filled automatically by adding a suffix "_straight" to the input file name. You thought that this name was generic, but in fact it was not. So even thought it would make the test work (in that particular situation), that would have cause breakage at other places. I will take care of this fix. If you could please make sure that such a change did not happen in your other functions that would be great.

YangHee-Min:
> @YangHee-Min as you can see here, the default value for output file was '', which you changed for something else here. Now, the problem is that later, in the SpinalCord Straightener class, the variable fname_output gets filled automatically by adding a suffix "_straight" to the input file name. You thought that this name was generic, but in fact it was not. So even thought it would make the test work (in that particular situation), that would have cause breakage at other places. I will take care of this fix. If you could please make sure that such a change did not happen in your other functions that would be great.

@jcohenadad  Oooh I see. Thanks for the explanation! I will give it a look

scripts/sct_image.py
```diff
@@ -12,188 +12,209 @@
...
+        help='Input file(s). If several inputs: separate them by white space. Example: "data.nii.gz"',
+        required = True)
+    optional = parser.add_argument_group("OPTIONAL ARGUMENTS")
+    optional.add_argument(
+        "-h",
+        "--help",
+        action="help",
+        help="Show this help message and exit")
+    optional.add_argument(
+        "-o",
```
YangHee-Min: @jcohenadad Is it normal that "-o" has  no default value even though it's optional?

YangHee-Min:
>> @YangHee-Min as you can see here, the default value for output file was '', which you changed for something else here. Now, the problem is that later, in the SpinalCord Straightener class, the variable fname_output gets filled automatically by adding a suffix "_straight" to the input file name. You thought that this name was generic, but in fact it was not. So even thought it would make the test work (in that particular situation), that would have cause breakage at other places. I will take care of this fix. If you could please make sure that such a change did not happen in your other functions that would be great.
> >
> @jcohenadad Oooh I see. Thanks for the explanation! I will give it a look

Everything looks good to me

jcohenadad:
tested `batch_processing` on commit 36d50478dacfc29ab12c797c914e6b5d46804369, it works fine:
~~~
t2/CSA:          78.57328415897263
mt/MTR(WM):      54.08675584008292
t2s/CSA_WM:      65.08261276857283
dmri/FA(CST_r):  0.7879034599204916
dmri/FA(CST_l):  0.7628544780305481
~~~

PR #2396, opened 2019-08-07
Fixed wrong morphometric measures with anisotropic in-plane resolution

jcohenadad:
If axial in-plane resolution is anisotropic, some shape morphometric measures, such as RL/AP diameter and orientation are wrong. 

This PR solves it by resampling the data to isotropic resolution (chooses the smallest pixel size between RL and AP dimension for target resolution) before computing shape morphometrics.

Also see: http://forum.spinalcordmri.org/t/sct-process-segmentation-is-mixing-up-diameter-rl-and-diameter-ap/135/3

### DIFF
--- a/spinalcordtoolbox/process_seg.py
+++ b/spinalcordtoolbox/process_seg.py
@@ -10,10 +10,13 @@ import numpy as np
 from skimage import measure, transform
 from tqdm import tqdm
 import logging
+import nibabel
+from nipy.io.nifti_ref import nifti2nipy, nipy2nifti
 
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.aggregate_slicewise import Metric
 from spinalcordtoolbox.centerline.core import ParamCenterline, get_centerline
+from spinalcordtoolbox.resampling import resample_nipy
 
 
 def compute_shape(segmentation, angle_correction=True, param_centerline=None, verbose=1):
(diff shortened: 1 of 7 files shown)


spinalcordtoolbox/process_seg.py
```diff
@@ -39,12 +42,30 @@ def compute_shape(segmentation, angle_correction=True, param_centerline=None, ve
...
+    # Note: need to convert to nibabel, then nipy (input type of resample_nipy)
+    im_seg_nibabel = nibabel.nifti1.Nifti1Image(im_seg.data, im_seg.hdr.get_best_affine())
+    im_segr_nipy = resample_nipy(
+        nifti2nipy(im_seg_nibabel), new_size=[pr, pr, pz], new_size_type='mm', interpolation='linear')
+    # Convert back to Image type
+    im_segr_nibabel = nipy2nifti(im_segr_nipy)
+    im_segr = Image(im_segr_nibabel.get_data(), hdr=im_segr_nibabel.header, orientation='RPI', dim=im_segr_nibabel.header.get_data_shape())
+
+
+    # img = msct_image.Image(i.get_data(), hdr=i.header,
```
gmotzespina: Do we need to leave this code commented out here?

spinalcordtoolbox/process_seg.py
```diff
@@ -54,15 +75,15 @@ def compute_shape(segmentation, angle_correction=True, param_centerline=None, ve
     if angle_correction:
```
gmotzespina: Not a problem at all but maybe the else statement could be avoided writing the code like this: 

```python
fit_results = None
if angle_correction:
     # compute the spinal cord centerline based on the spinal cord segmentation
     # here, param_centerline.minmax needs to be False because we need to retrieve the total number of input slices
     _, arr_ctl, arr_ctl_der, fit_results = get_centerline(im_segr, param=param_centerline, verbose=verbose)
```

Not a big change. If you consider it is better to leave it as it is, it's fine too!

spinalcordtoolbox/process_seg.py
jcohenadad: no, we can remove

spinalcordtoolbox/process_seg.py
jcohenadad: we can go your way, can you do those changes? thanks!

gmotzespina: Sure, I'll do it!

gmotzespina:
@jcohenadad should we get this merged?

jcohenadad:
@gmotzespina I just want to copy/paste results of batch_processing to assess potential retro-compatibility change

jcohenadad:
results of `batch_processing`:

process-seg 2ecd4bc558e8e6e3f9e807c2e001ab303a16c9be
~~~
t2/CSA:          74.05758272483698
mt/MTR(WM):      54.29899900160193
t2s/CSA_GM:      12.774964384249389
t2s/CSA_WM:      64.64670129025727
dmri/FA(CST_r):  0.7723247785662822
dmri/FA(CST_l):  0.7469896944701713
~~~

PR #2397, opened 2019-08-07
Proper handling of int arguments contained in list type input

### LINKED ISSUE #2395: New argparse in sct_image.py doesn't handle integer arguments
@jcohenadad @YangHee-Min Sorry, just realized I had commented on a closed issue. There's a problem with the recent update to sct_image.py introduced in commit 7abacb0

-keep-vol (and other) command line arguments were formerly specified in the
argument parser as type_value=[[','], 'int']

The typespec is no longer present and now I get this error:

sct_image -keep-vol 0 -i fmri.nii.gz -o fmri_0.nii.gz

--
Spinal Cord Toolbox (git-HEAD-7fad17a0db089cbf7cdcb635062f8fac65f174c1*)

File "/opt/sct/scripts/sct_image.py", line 644, in 
main()
File "/opt/sct/scripts/sct_image.py", line 207, in main
im_out = [remove_vol(im_in, index_vol, todo='keep')]
File "/opt/sct/scripts/sct_image.py", line 493, in remove_vol
data_out = data[:, :, :, index_vol]
IndexError: only integers, slices (:), ellipsis (...), numpy.newaxis (None) and integer or boolean arrays are valid indices

YangHee-Min:
When int type arguments were inputted as `-removed-vol` or `-keep-vol`, those respective arguments were not parsed as the proper variable type.

This PR parses the arguments correctly by changing the variable type of each element of the inputted list from a character to an int (fixes #2395 ).

To test the PR, one may do the following within `sct_testing_data/fmri`: 
For `-keep-vol`: 
`sct_image -keep-vol 0 -i fmri.nii.gz -o fmri_0.nii.gz`

For `-remove-vol`:
`sct_image -remove-vol 0 -i fmri.nii.gz -o fmri_0.nii.gz`

### DIFF
--- a/scripts/sct_image.py
+++ b/scripts/sct_image.py
@@ -31,45 +31,45 @@ def get_parser():
                     'Inputs can be a number, a 4d image, or several 3d images separated with ","',
         formatter_class=SmartFormatter,
         add_help=None,
-        prog=os.path.basename(__file__).strip(".py"))
-    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+        prog=os.path.basename(__file__).strip('.py'))
+    mandatory = parser.add_argument_group('MANDATORY ARGUMENTS')
     mandatory.add_argument(
-        "-i",
-        nargs="+",
+        '-i',
+        nargs='+',
         metavar=Metavar.file,
         help='Input file(s). If several inputs: separate them by white space. Example: "data.nii.gz"',
         required = True)
-    optional = parser.add_argument_group("OPTIONAL ARGUMENTS")
+    optional = parser.add_argument_group('OPTIONAL ARGUMENTS')
     optional.add_argument(
-        "-h",
-        "--help",
-        action="help",
-        help="Show this help message and exit")
+        '-h',
+        '--help',
+        action='help',
+        help='Show this help message and exit')
     optional.add_argument(
-        "-o",
+        '-o',
         metavar=Metavar.file,
         help='Output file. Example: data_pad.nii.gz',
         required = False)
 
     image = parser.add_argument_group('IMAGE OPERATIONS')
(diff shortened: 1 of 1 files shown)


scripts/sct_image.py
```diff
@@ -202,7 +202,9 @@ def main(args=None):
         im_out = None
 
     elif arguments.keep_vol is not None:
-        index_vol = arguments.keep_vol
+        index_vol = (arguments.keep_vol).split(",")
```
gmotzespina: I see that we use sometimes single quotes and sometimes double quotes.

For instance 
```python
    dim_list = ['x', 'y', 'z', 't']
```
But here,
```python
        index_vol = (arguments.keep_vol).split(",")
```
Would it be worth it to try to keep the same convention across the code?

YangHee-Min: If that's the case, I think sticking to single quotes might be a good idea since when we use `argparse`, when defining our strings, we stick to having single quotes for strings and use double quotes within strings when we want to add examples. 

If ever it is worth  to assign a convention when single/double quotes in examples, I think it would be better to stick with single. However, I don't know if we want to assign a convention for something like this.

Thoughts?

gmotzespina: I think we should stick to single quotes. In my opinion standardizing the code is a good practice to keep it clean. Any thoughts @jcohenadad ?

jcohenadad: yes for convention and yes for single quote 👍

YangHee-Min: Sounds good! Changed in commit [8bb3201](https://github.com/neuropoly/spinalcordtoolbox/commit/8bb320172a2485ce8815f257ee1c1f55bab3c0b4)

PR #2409, opened 2019-08-14
Modify bashrc on sudo

### LINKED ISSUE #2405: Modify bashrc for root installation
Currently, if user installs SCT as root, the installer does not modify the .bashrc (or equivalent if using other shell) and display a message: "Please modify your .bashrc". 

Of course, nobody reads those message... so as a result, users try to launch SCT functions and it fails. 

So: i suggest we simply modify the .bashrc (or equivalent), wether users are root or not. For experienced users, they will know what to do if they do *not* want their bashrc touched. For unexperienced users, this process will be a black box anyway and they won't care.

gmotzespina:
Fixes #2405 

## Changes

- Removes sudo check and modifies `.bashrc` or respective terminal file for every type of user.


install_sct
```diff
@@ -562,45 +562,42 @@ fi
...
 Please copy the historic of this Terminal (starting with the command install_sct) and paste it in the SCT Help forum (create a new discussion):\n
 http://forum.spinalcordmri.org/c/sct\n"
-  fi
-# if sudo=1
-else
-  echo -e "\nWARNING: sudo mode: Profile file was not edited. To complete installation, please add the following line to $THE_RC:"
-  echo -e $DISPLAY_UPDATE_PATH
-  echo -e "Then, please open a new Terminal and run sct_check_dependencies to make sure SCT is properly installed.\n"
-fi
+fi
```
jcohenadad: i think it's a good habit to add new line at the end of the file

gmotzespina: Right, I've added the new line.

Issue #2404, opened 2019-08-12
SCT cannot install if gcc not installed


jcohenadad:
The compilation is needed for nipy (and maybe other modules).

A solution is to check wether gcc is installed. If it is not, send error message telling user what to do. On OSX, installing brew, followed by `brew install gcc` is a simple solution.

PR #2410, opened 2019-08-14
Check if gcc is installed

gmotzespina:
Fixes: #2404 

## Change list

- Fetching the OS was refactored to be a function and is now called at the beginning of the script. Nice to check from the beginning on the requirements.

- Checks if `gcc` installed and ask the `osx` users if they want to install it right away.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -97,7 +97,7 @@ function check_requirements() {
         if [[ ! $(which brew) ]]; then
           yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
         fi
-        yes | sudo brew install gcc
+        yes | brew install gcc
         # check if gcc install ran properly
         gcc --version > /dev/null 2>&1  # run silently, then check output status
         if [[ $? ]]; then
--- a/install_sct
+++ b/install_sct
@@ -81,7 +81,7 @@ function fetch_os_type() {
 # Checks if the necessary tools for SCT are installed on the machine
 function check_requirements() {
 
-  echo "Checking requirements..."
+  echo; echo "Checking requirements..."
   # check curl
   if [[ ! $(which curl) && ! $(which wget) ]]; then
     die "ERROR: neither \"curl\" nor \"wget\" is installed. Please install either of them and restart SCT installation."
@@ -91,8 +91,8 @@ function check_requirements() {
   gcc --version > /dev/null 2>&1  # run silently, then check output status
   if [[ $? ]]; then
     if [[ $OS == "osx" ]]; then
-      echo -e "\nWARNING: \"gcc\" is not installed."
-      read -p 'Do you want to install \"gcc\" now? (accepting to install gcc will also install \"brew\" in case it is not installed already)? [y]es/[n]o: ' -r GCC_INSTALL
+      echo -e "WARNING: \"gcc\" is not installed."
+      read -p 'Do you want to install "gcc" now? (accepting to install "gcc" will also install "brew" in case it is not installed already)? [y]es/[n]o: ' -r GCC_INSTALL
       if [[ $GCC_INSTALL =~ [Yy](es)? ]]; then
         if [[ ! $(which brew) ]]; then
           yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
--- a/install_sct
+++ b/install_sct
@@ -89,7 +89,7 @@ function check_requirements() {
 
   # check gcc
   gcc --version > /dev/null 2>&1  # run silently, then check output status
-  if [[ ! $? ]]; then
+  if [[ $? ]]; then
     if [[ $OS == "osx" ]]; then
       echo -e "\nWARNING: \"gcc\" is not installed."
       read -p 'Do you want to install \"gcc\" now? (accepting to install gcc will also install \"brew\" in case it is not installed already)? [y]es/[n]o: ' -r GCC_INSTALL
--- a/install_sct
+++ b/install_sct
@@ -45,6 +45,38 @@ function die() {
   exit 1
 }
 
+function fetch_os_type() {
+  # Fetch OS type
+  if uname -a | grep -i darwin >/dev/null 2>&1; then
+    # OSX
+
+    # Fix for non-English Unicode systems on MAC
+    if [[ -z $LC_ALL ]]; then
+      export LC_ALL=en_US.UTF-8
+    fi
+
+    if [[ -z $LANG ]]; then
+      export LANG=en_US.UTF-8
+    fi
+
+    OS=osx
+    force_bashrc_loading
+  elif uname -a | grep -i linux >/dev/null 2>&1; then
+    if cat /etc/issue | grep -i centos | grep 6. 2>&1; then
+      # CentOS 6.X
+      OS=linux_centos6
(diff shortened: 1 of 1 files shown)


install_sct
```diff
@@ -57,6 +89,19 @@ for installation instructions check: https://git-scm.com/book/en/v2/Getting-Star
 "
   fi
 
+  if [[ ! $(which gcc) ]]; then
+    if [[ $OS == "osx" ]]; then
+      read -p 'You need to install gcc before you can use the SPINALCORDTOOLBOX. Do you want to install gcc now? [y]es/[n]o: ' -r GCC_INSTALL
+      if [[ $GCC_INSTALL =~ [Yy](es)? ]]; then
+        brew install gcc
```
jcohenadad: `brew` is not installed natively on OSX (unfortunately), so this could break

gmotzespina: Dang! 😢 

I'll remove that part.

jcohenadad: one possibility would be to add another condition that would check if brew is installed, and if not, [installs it](https://brew.sh/)

gmotzespina: Or I could add a check to see if the user has brew installed.

install_sct
```diff
@@ -45,6 +45,38 @@ function die() {
   exit 1
 }
 
+function fetch_os_type() {
+  # Fetch OS type
```
jcohenadad: it is a good habit to add a header in functions that indicates what are the outputs or updated variables (in this case `OS`).

install_sct
```diff
@@ -91,8 +93,12 @@ for installation instructions check: https://git-scm.com/book/en/v2/Getting-Star
...
   if [[ ! $(which gcc) ]]; then
     if [[ $OS == "osx" ]]; then
-      read -p 'You need to install gcc before you can use the SPINALCORDTOOLBOX. Do you want to install gcc now? [y]es/[n]o: ' -r GCC_INSTALL
+      echo -e "\nYou need to install gcc before you can use the SPINALCORDTOOLBOX."
+      read -p 'Do you want to install gcc now (accepting to install gcc will also install brew in case it is not installed already)? [y]es/[n]o: ' -r GCC_INSTALL
       if [[ $GCC_INSTALL =~ [Yy](es)? ]]; then
+        if [[ ! $(which brew) ]]; then
+          /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+        fi
         brew install gcc
```
jcohenadad: i would install gcc and brew it in silence mode (in case installer asks for confirmation, etc. that the user might not be confortable with)

gmotzespina: Sure. Changed :+1

jcohenadad:
Let's wait to understand better the cause of this potentially-related issue https://github.com/neuropoly/spinalcordtoolbox/issues/2399. If the issue is indeed coming from the xcode-gcc, then we should also test for the type of gcc that is installed (and thus update this PR accordingly).

Also, let's wait for this PR https://github.com/neuropoly/spinalcordtoolbox/pull/2414, because this might remove the need to compile wheel (and thus this PR will not be needed anymore).

jcohenadad:
@gmotzespina This PR https://github.com/neuropoly/spinalcordtoolbox/pull/2414 drops the dependency to gcc (unless I am mistaken), therefore you can remove the gcc check in this PR. But keep the OS check at the beginning since it is a nice refactoring.

gmotzespina:
@jcohenadad nice, I'll change the PR accordingly then.

jcohenadad:
@gmotzespina oops! scrap that. I just tested on Ubuntu without gcc and it failed 😞 
so i guess we keep the check!

jcohenadad:
that's something i don't understand though... i temporarily removed gcc on my OSX and on an Ubuntu. Install passed on OSX but not on Ubuntu: got the error when trying to build psutil. Is psutil installed by default on OSX?

install_sct
```diff
@@ -57,8 +91,26 @@ for installation instructions check: https://git-scm.com/book/en/v2/Getting-Star
 "
   fi
 
+  if [[ ! $(which gcc) ]]; then
```
jcohenadad: this is not a sensitive test. For example, on an OSX where gcc is not installed (but still "recognized"), I get this:
~~~
$ which gcc
/usr/bin/gcc
$ gcc
** INTERNAL ERROR: Uncaught exception **
Uncaught Exception: Error getting value for key 'initializationClass' of extension 'Xcode.DVTKitDFRSupport.Initializer' in plug-in 'com.apple.dt.IDE.DVTKitDFRSupport'
Stack:
  0   __exceptionPreprocess (in CoreFoundation)
  1   objc_exception_throw (in libobjc.A.dylib)
  2   -[DVTExtension valueForKey:error:] (in DVTFoundation)
  3   _IDEInitializeOnePlugInAndPrerequisites (in IDEFoundation)
  4   _IDEInitializePlugIns (in IDEFoundation)
  5   IDEInitialize (in IDEFoundation)
  6   -[Xcode3CommandLineBuildTool run] (in Xcode3Core)
  7  0x00000001083b6202 (in xcodebuild)
  8   start (in libdyld.dylib)

sh: line 1:  3211 Abort trap: 6           /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -sdk macosx -find gcc 2> /dev/null
gcc: error: unable to find utility "gcc", not a developer tool or in PATH
~~~

jcohenadad: nevermind-- your code works (because output status is not 0)

jcohenadad: actually, when running the actual installer, it does **not** work ;-)

on an OSX that has missing gcc (see message above), the installer fails, giving:
~~~
...
    
    sh: line 1:  4420 Abort trap: 6           /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -sdk macosx -find gcc 2> /dev/null
    gcc: error: unable to find utility "gcc", not a developer tool or in PATH
    error: command 'gcc' failed with exit status 72
    ----------------------------------------
ERROR: Command "/Users/jcohen/sct-dev2/python/envs/venv_sct/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/private/var/folders/r3/9skf4b5x4y14t1jh0zs0g0l80000gn/T/pip-install-757cu9wr/psutil/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/var/folders/r3/9skf4b5x4y14t1jh0zs0g0l80000gn/T/pip-record-uxgqf44l/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /private/var/folders/r3/9skf4b5x4y14t1jh0zs0g0l80000gn/T/pip-install-757cu9wr/psutil/

Failed running pip install -r requirements.txt: 0
~~~

jcohenadad: which means that the `if [[ ! $(which gcc) ]]; then` clause was not effective

jcohenadad: ok, i get it. To be effective, we would have to do something like:
~~~
$gcc --version
$ echo $?
72  # --> GCC NOT WORKING
~~~

jcohenadad:
ok, i've added some layers of checking...

now i get this...
~~~
Error: Your Xcode (8.2.1) is too outdated.
Please update to Xcode 10.2 (or delete it).
Xcode can be updated from the App Store.
~~~

looks like it would be useful to check minimum OS requirements before running installer.

PR #2417, opened 2019-08-16
Fixed flag '-evecs' not detecting input as of type int

### LINKED ISSUE #2416: Flag -evecs not working after switching to argparse
After the recent move from msct_parser to argparse (see: https://github.com/neuropoly/spinalcordtoolbox/pull/2337), a bug has been introduced in the way -evecs param is interpreted.

User report: http://forum.spinalcordmri.org/t/error-sct-4-0-0-sct-dmri-compute-dti/142

YangHee-Min:
At the moment, the `-evecs` argument is not detected as an `int` and therefore causes an error message when inputs are either `0` or `1` (even though the help displays these two as being options).

In this PR, we added the type expected from this argument, therefore fixing the issue related to argparse not recognizing the supposedly valid inputs (fixes #2416 )

To test this PR, I ran the following commands: 

1. cd sct_testing_data
2. sct_dmri_compute_dti -i dmri/dmri.nii.gz -bvec dmri/bvecs.txt -bval dmri/bvals.txt -method restore -evecs 1


YangHee-Min:
@jcohenadad Changes for sct_dmri_compute_dti done in commit [e85a840](https://github.com/neuropoly/spinalcordtoolbox/commit/e85a84065d59d871244f358f1d279f970d9414f7).

However, there was something odd in `sct_pipeline_makefig`. Indeed, as seen [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/551627a62fe96e2ef9956a1d64969c9ff211c4e2/scripts/sct_pipeline_makefig.py#L18-L34), this script has not been converted to the new argparse. Along with this, there doesn't seem to be anywhere within this script where argument `-v` is actually used. I have two questions:
1. Can we simply remove '-v' from this part of code 
2. Should we convert this script to the new argparse?

jcohenadad:
> @jcohenadad Changes for sct_dmri_compute_dti done in commit [e85a840](https://github.com/neuropoly/spinalcordtoolbox/commit/e85a84065d59d871244f358f1d279f970d9414f7).
> 
> However, there was something odd in `sct_pipeline_makefig`. Indeed, as seen [(here)](https://github.com/neuropoly/spinalcordtoolbox/blob/551627a62fe96e2ef9956a1d64969c9ff211c4e2/scripts/sct_pipeline_makefig.py#L18-L34), this script has not been converted to the new argparse. Along with this, there doesn't seem to be anywhere within this script where argument `-v` is actually used. I have two questions:
> 
> 1. Can we simply remove '-v' from this part of code

yes

> 2. Should we convert this script to the new argparse?

no (script isn't used and will be removed at some point)

YangHee-Min:
@jcohenadad All scripts have been reviewed. All necessary instances where `type=int` is required due to `choices` being of that type seem to be there

PR #2425, opened 2019-09-03
Check if folder already exists by checking its actual name, not the name of the entry to -d flag

gmotzespina:
Removes folder names based on it's name not on `data_name`

Fixes: #2425

### DIFF
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -132,7 +132,6 @@ def main(args=None):
     sct.init_sct(log_level=verbose, update=True)  # Update log level
     dest_folder = arguments.get('-o', os.path.abspath(os.curdir))
     dest_tmp_folder = sct.tmp_create()
-    print('Temp folder: ' + dest_tmp_folder)
 
     # Download data
     url = dict_url[data_name]
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -19,7 +19,7 @@ import tarfile
 import tempfile
 import zipfile
 
-# importing shutil module  
+# importing shutil module
 import shutil
 
 import requests
@@ -43,9 +43,9 @@ def get_parser():
             'sct_example_data',
             'sct_testing_data',
             'course_hawaii17',
-            'course_paris18', 
-            'course_london19', 
-            'sct_course-beijing19',
+            'course_paris18',
+            'course_london19',
+            'course_beijing19',
             'PAM50',
             'MNI-Poly-AMU',
             'gm_model',
@@ -112,8 +112,8 @@ def main(args=None):
                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20180612_sct_course-paris18.zip'],
         'course_london19': ['https://osf.io/4q3u7/?action=download',
                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20190121_sct_course-london19.zip'],
-        'sct_course-beijing19': ['https://osf.io/ef4xz/?action=download',
-                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip'],
+        'course_beijing19': ['https://osf.io/ef4xz/?action=download',
+                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip'],
         'deepseg_gm_models': ['https://osf.io/b9y4x/?action=download',
                               'https://www.neuro.polymtl.ca/_media/downloads/sct/20180205_deepseg_gm_models.zip'],
         'deepseg_sc_models': ['https://osf.io/avf97/?action=download',
@@ -160,12 +160,12 @@ def main(args=None):
                 'warning',
             )
             sct.rmtree(data_extracted_name)
-    
+
     # Destination path
     for source_path in extracted_files_paths:
         # Move the content of source to destination
         shutil.move(source_path, dest_folder, copy_function=shutil.copytree)
-        
+
     sct.printv('\nRemove temporary file...', verbose)
     os.remove(tmp_file)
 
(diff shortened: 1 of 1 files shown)


scripts/sct_download_data.py
```diff
@@ -127,24 +128,46 @@ def main(args=None):
     data_name = arguments['-d']
     verbose = int(arguments.get('-v'))
     sct.init_sct(log_level=verbose, update=True)  # Update log level
-    dest_folder = arguments.get('-o', os.path.abspath(os.curdir))
+    dest_tmp_folder = arguments.get('-o', os.path.abspath(os.curdir)+'/tmp')
```
jcohenadad: "/" doesn't work in Windows, and what if user already has a local folder called "./tmp/"?
there are better ways to generate temp folders, see class `sct_utils.TempFolder`

scripts/sct_download_data.py
```diff
@@ -127,24 +128,46 @@ def main(args=None):
...
 
     # Download data
     url = dict_url[data_name]
     tmp_file = download_data(url, verbose)
 
+    # unzip
+    unzip(tmp_file, dest_tmp_folder, verbose)
+    data_extracted_name = data_name
+    if dest_tmp_folder != os.path.abspath(os.curdir)+'/bin':
```
jcohenadad: why "bin"?

scripts/sct_download_data.py
gmotzespina: Oh ok, nice one, looking at it.

scripts/sct_download_data.py
gmotzespina: Because it is the path where the installation requirements are downloaded.

jcohenadad: `sct_download_data` is not only used for installing binaries (look at the help)

scripts/sct_download_data.py
```diff
@@ -84,92 +86,174 @@ def main(args=None):
...
-                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip'],
-        'deepseg_gm_models': ['https://osf.io/b9y4x/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180205_deepseg_gm_models.zip'],
-        'deepseg_sc_models': ['https://osf.io/avf97/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip'],
-        'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip'],
-        'c2c3_disc_models': ['https://osf.io/t97ap/?action=download',
-                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip']
+        "sct_example_data": [
```
jcohenadad: @gmotzespina please see the convention: https://github.com/neuropoly/spinalcordtoolbox/wiki/Single-or-double-quotes

unless you have a good reason for going with double quote?

scripts/sct_download_data.py
```diff
@@ -84,92 +86,174 @@ def main(args=None):
...
     tmp_file = download_data(url, verbose)
 
-    # Check if folder already exists
-    sct.printv('\nCheck if folder already exists...', verbose)
-    if os.path.isdir(data_name):
-        sct.printv('WARNING: Folder ' + data_name + ' already exists. Removing it...', 1, 'warning')
-        sct.rmtree(data_name)
+    # Extracting the binaries doesn't require removing the previously created directories
+    # since binaries are files, not directories
+    if "binaries" in data_name:
```
jcohenadad: will break if compressed files don't have the term "binaries" in it. Why not simply:
- check if the uncompressed zip gives you a folder
- check if the folder already exists
- if it does, remove it

scripts/sct_download_data.py
```diff
@@ -84,92 +86,174 @@ def main(args=None):
...
     # unzip
-    unzip(tmp_file, dest_folder, verbose)
+    unzip(tmp_file, dest_tmp_folder, verbose)
+    folders = []
+    # Get the name of the extracted data
+    for root, dirs, files in os.walk(dest_tmp_folder):
+        for folder in dirs:
+            folders.append(os.path.join(root, folder))
+    data_extracted_path = folders[0]
+    directories = data_extracted_path.split("/")
```
jcohenadad: python has built-in modules to do such manipulations which will break with windows (when, one day, SCT will be compatible with windows...)

scripts/sct_download_data.py
gmotzespina: Yeah, I figured out that this would happen, I'm currently working on a fix. to have this done in a cleaner way.

scripts/sct_download_data.py
gmotzespina: I started testing out Black, and it changed my file to double quotes, since it is kind of the black standard. I'll revert all of this changes before sending the request again.

While checking out black I also figured out that using double quotes instead of single might be a better option since we can write stuff like this: `"it's a test"` otherwise, while using single quotes we would need to write the same text this way: `'it\'s a test'`.

What do you think about it?

jcohenadad: the double quote argument is indeed a good argument, but we could also make the argument that with single quote, we can do that: 'When using this option, enter parameter "n"'. Also, single quotes makes the code lighter.

Anyway, regardless the decision, this should be mixed with this PR, which is about fixing a specific problem, not changing code cosmetic. Otherwise, it makes it difficult for me to review your code.

so: open a separate issue regarding code cosmetic, and we will address it specifically

gmotzespina: Right, I didn't thought about the options. 

Yeah, I'll revert those changes once the PR is ready, they just got mixed because of Black, but it was not intended to have those changes in this PR.

PR #2431, opened 2019-09-06
Compute cord length

jcohenadad:
With the recent removal of `-p` in `sct_process_segmentation`, it was no more possible to measure cord length. This feature was removed because we (wrongly) thought it was not used, but [some users expressed interest](http://forum.spinalcordmri.org/t/computing-cord-length-in-v4-0-0/105) in having this feature back. 

This PR implements cord length computation as part of `sct_process_segmentation`.

In brief, the algorithm for computing cord length is to sum slice thickness (corrected for cord angle) across a specified number of slices (by default, all slices). A new column with label `SUM(length)` is now present in the output .csv file.

To test it:
~~~
cd sct_testing_data/t2
sct_process_segmentation -i t2_seg_manual.nii.gz -perslice 0
~~~

### DIFF
--- a/spinalcordtoolbox/aggregate_slicewise.py
+++ b/spinalcordtoolbox/aggregate_slicewise.py
@@ -44,8 +44,173 @@ class LabelStruc:
         self.map_cluster = map_cluster
 
 
+def func_bin(data, mask, map_clusters=None):
+    """
+    Get the average of data after binarizing the input mask
+    :param data: nd-array: input data
+    :param mask: (n+1)d-array: input mask
+    :param map_clusters: not used
+    :return:
+    """
+    # Binarize mask
+    mask_bin = np.where(mask >= 0.5, 1, 0)
+    # run weighted average
+    return func_wa(data, mask_bin)
+
+
+def func_max(data, mask=None, map_clusters=None):
+    """
+    Get the max of an array
+    :param data: nd-array: input data
+    :param mask: not used
+    :param map_clusters: not used
+    :return:
+    """
+    return np.max(data), None
+
+
+def func_map(data, mask, map_clusters):
+    """
+    Compute maximum a posteriori (MAP) by aggregating the last dimension of mask according to a clustering method
+    defined by map_clusters
+    :param data: nd-array: input data
+    :param mask: (n+1)d-array: input mask. Note: this mask should include ALL labels to satisfy the necessary condition for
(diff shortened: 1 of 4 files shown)


spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -44,8 +44,173 @@ def __init__(self, id, name, filename=None, map_cluster=None):
...
+    """
+    # Check if mask has an additional dimension (in case it is a label). If so, select the first label
+    if mask.ndim == data.ndim + 1:
+        mask = mask[..., 0]
+    return np.average(data, weights=mask), None
+
+
 def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], perslice=None, perlevel=False,
-                                 vert_level=None, group_funcs=(('MEAN', np.mean),), map_clusters=None):
+                                 vert_level=None, group_funcs=(('MEAN', func_wa),), map_clusters=None):
```
charleygros: Just curious: Why did you change this (`np.mean` --> `func_wa`)?

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -55,7 +220,8 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
     :param Bool perslice: Aggregate per slice (True) or across slices (False)
     :param Bool perlevel: Aggregate per level (True) or across levels (False). Has priority over "perslice".
     :param vert_level: Vertebral level. Could be either an Image or a file name.
-    :param tuple group_funcs: Functions to apply on metric. Example: (('mean', np.mean),))
+    :param tuple group_funcs: Name and function to apply on metric. Example: (('MEAN', func_wa),)). Note, the function
```
charleygros: Nice!

spinalcordtoolbox/aggregate_slicewise.py
jcohenadad: good question. because the default `np.mean` was failing, due to recent changes in the function which expects two outputs, as can be seen [here](https://github.com/neuropoly/spinalcordtoolbox/blob/0469ed131fb580f01185d54fbda3ee86e9d37025/spinalcordtoolbox/aggregate_slicewise.py#L279). It never crashed before because we always give input to this function, but the default value would crash.

Issue #1074, opened 2016-12-22
weird output re-scaling if anisotropic input image


jcohenadad:
data: 

<img width="142" alt="screen shot 2016-12-21 at 19 14 33" src="https://cloud.githubusercontent.com/assets/2482071/21410863/f6353ec0-c7b1-11e6-8eba-f2fe68635383.png">

<img width="57" alt="screen shot 2016-12-21 at 19 13 51" src="https://cloud.githubusercontent.com/assets/2482071/21410865/f90be374-c7b1-11e6-99b9-d8eaddeb5247.png">

benjamindeleener:
NURBS seem to badly approximate the segmentation centerline on this one (see Figure below).
<img width="752" alt="screen shot 2016-12-21 at 22 29 51" src="https://cloud.githubusercontent.com/assets/7757117/21414404/1f2d3b28-c7cd-11e6-87ed-7b53a54ef439.png">

And it is due to edge effects on the segmentation (not anisotropic resolution):
<img width="517" alt="screen shot 2016-12-21 at 22 31 52" src="https://cloud.githubusercontent.com/assets/7757117/21414426/52b8d196-c7cd-11e6-9e14-d3877e4c12a9.png">

Much better when correcting the segmentation:
<img width="752" alt="screen shot 2016-12-21 at 22 33 50" src="https://cloud.githubusercontent.com/assets/7757117/21414464/9c96bec2-c7cd-11e6-8f6c-b52c8b9491ed.png">

Here are the results of straightening:
<img width="287" alt="screen shot 2016-12-21 at 22 42 03" src="https://cloud.githubusercontent.com/assets/7757117/21414594/bcbbc020-c7ce-11e6-82a8-3f431c74220a.png">

benjamindeleener:
We need to find solutions for these edge effects in spinal cord segmentations.

This one can be closed right?

jcohenadad:
yes, we can close this one if we open another one specific to edge effect, with the same priority level (this is a very important issue).

jcohenadad:
@benjamindeleener: if propseg can output a "true" centerline in physical coordinates, would it be possible to compare this centerline with that generated by computing the slicewise center of mass of the generated segmentation, and if it deviates by more than XX mm, then we assume this is caused by a truncated segmentation and we remove the slice.

benjamindeleener:
[...]

PR #2437, opened 2019-09-12
Remove isolated voxels at the edge of the output segmentation

### LINKED ISSUE #2435: Wrong center of mass detection caused by spurious blob related to NURBS instabilities caused by isolated voxels from segmentation
Based on a [user report](http://forum.spinalcordmri.org/t/sct-label-vertebrae-fails-v-4-0-0/162?u=jcohenadad), the center of mass of the c2-c3 label blob is improperly calculated. This only happens in the image below (but maybe more...).

To reproduce, download this label:
[labelz_straight.nii.gz](https://github.com/neuropoly/spinalcordtoolbox/files/3602043/labelz_straight.nii.gz)

Then run:
~~~
conda activate venv_sct
python
nii = Image('labelz_straight.nii.gz')
import scipy.ndimage.measurements
cx, cy, cz = scipy.ndimage.measurements.center_of_mass(nii.data)
print(cx, cy, cz)
76.58216783216783 91.0367132867133 257.25174825174827
~~~

When interrogating the voxel value of this center of mass (rounded), result is 0.0 whereas it should be 2.0:
~~~
nii.data[int(cx), int(cy), int(cz)]
0.0
~~~

This can be verified, when opening the image with any viewer (e.g., fsleyes), it is clear that the center mass estimate is shifted by ~10 voxels.

There might be an issue with the matrix definition of the image (something went wrong during the loading of the image???), although I doubt it (the shape is consistent: 139x139x338). 

So maybe there is a bug in `scipy.ndimage.measurements.center_of_mass`?  And if this is the case, the function could be replaced by a [simple implementation](https://stackoverflow.com/a/29357037).

jcohenadad:
When using `sct_deepseg_sc`, sometimes the top and/or bottom segmented slice shows incomplete segmentation: only few isolated voxels have the value "1". This could create some issues later on, see for example #2435. 

In this PR, we implement a fix, which consists in computing the CSA across all slices, and checking if the extreme slices have a CSA that is 50% smaller than the median of the CSA on the previous slides. If it is, it will zero the slide.

Specific changes to this PR include:
- Now checking extreme segmented slices and remove isolated voxels (see description above). Fixes #2435
- Refactored `deepseg_sc.core` by creating a new file `deepseg_sc.postprocessing` that includes post-processing code. For clarity.

Note: this issue is similar to #1074 when using `sct_propseg`, for which a fix was also created. In the future, we could merge the two fixes to save lines of code.

### DIFF
--- a/scripts/sct_deepseg_sc.py
+++ b/scripts/sct_deepseg_sc.py
@@ -167,9 +167,10 @@ def main():
 
     im_image = Image(fname_image)
     # note: below we pass im_image.copy() otherwise the field absolutepath becomes None after execution of this function
-    im_seg, im_image_RPI_upsamp, im_seg_RPI_upsamp, im_labels_viewer, im_ctr = deep_segmentation_spinalcord(
-        im_image.copy(), contrast_type, ctr_algo=ctr_algo, ctr_file=manual_centerline_fname,
-        brain_bool=brain_bool, kernel_size=kernel_size, remove_temp_files=remove_temp_files, verbose=verbose)
+    im_seg, im_image_RPI_upsamp, im_seg_RPI_upsamp, im_labels_viewer, im_ctr = \
+        deep_segmentation_spinalcord(im_image.copy(), contrast_type, ctr_algo=ctr_algo,
+                                     ctr_file=manual_centerline_fname, brain_bool=brain_bool, kernel_size=kernel_size,
+                                     remove_temp_files=remove_temp_files, verbose=verbose)
 
     # Save segmentation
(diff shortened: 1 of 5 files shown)


jcohenadad:
ran `batch_processing` for sanity check (core functions are affected by this PR), all good!: 
~~~
t2/CSA:          73.98379456979737
mt/MTR(WM):      54.31259385601459
t2s/CSA_GM:      12.530353265289374
t2s/CSA_WM:      65.07686820036822
dmri/FA(CST_r):  0.7940335588596036
dmri/FA(CST_l):  0.7704675850543123
~~~

Issue #2454, opened 2019-09-25
Cropping is setting a wrong qform factor


jcohenadad:
### Description

When cropping an image using `sct_crop_image`, the output qform code is wrongly set to 2 (Aligned Anat), which goes against the [Nifti convention](https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform_brief_usage), specifying that the qform_code should be either 0 or 1.

Moreover, the output sform_name and sform_code are inconsistent: it gives 
~~~
sform_name	Aligned Anat
sform_code	1  # ERROR! This code should correspond to Scanner Anat
~~~

I have not investigated if this issue is systematic.

Useful links:
- https://github.com/ANTsX/ANTs/wiki/How-does-ANTs-handle-qform-and-sform-in-NIFTI-1-images%3F
- https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html
- https://nipy.org/nibabel/nifti_images.html#the-qform-affine

Related to #2429.

### Steps to Reproduce

SCT commit: 

~~~
cd sct_testing_data/mt
sct_crop_image -i mt1.nii.gz -ymin 2 -ymax 1 -o mt1_crop.nii.gz
fslhd mt1
# --> qform_name	Scanner Anat
# --> qform_code	1
fslhd mt1_crop
# --> qform_name	Aligned Anat
# --> qform_code	2  # ERROR! Forbidden value
~~~

joshuacwnewton:
Adding this to the `5.2.1` milestone to be discussed, because it is listed as high-priority and might be worth investigating prior to a patch release.

joshuacwnewton:
> Moreover, the output sform_name and sform_code are inconsistent: it gives
> 
> ```
> sform_name	Aligned Anat
> sform_code	1  # ERROR! This code should correspond to Scanner Anat
> ```

This is strange, because `fslhd` infers the name directly from the code. From [`fslhd.cc`](http://ftp.nmr.mgh.harvard.edu/pub/dist/freesurfer/tutorial_packages/OSX/fsl_501/src/avwutils/fslhd.cc):

```cpp
  cout << "sform_name     " << nifti_xform_string(fslio->niftiptr->sform_code) << endl;
  cout << "sform_code     " << fslio->niftiptr->sform_code << endl;
```

[...]

PR #2455, opened 2019-09-25
QC report: Fixed shift along slice direction between image and overlay

### LINKED ISSUE #2453: Shift along slice direction between image and overlay
Sometimes, the QC report shows an overlay which is shifter by one slice compared to the background image, as illustrated below:
Screen Shot 2019-09-23 at 5 23 47 PM

To reproduce:
SCT commit 6cd471ae1c153c71d51ad0c8c808082cd4e1e756
Data:  [data_issue2453.zip](https://github.com/neuropoly/spinalcordtoolbox/files/3653640/data_issue2453.zip)

Syntax:
~~~
sct_qc -i PAM50_t1_reg.nii.gz -p sct_register_multimodal -s sub-amu01_acq-T1w_MTS_seg.nii.gz -d sub-amu01_acq-T1w_MTS_crop.nii.gz
~~~

jcohenadad:
### Description

Sometimes, the QC report shows an overlay which is shifter by one slice compared to the background image, as illustrated below:
Screen Shot 2019-09-23 at 5 23 47 PM

To reproduce:
SCT commit 6cd471ae1c153c71d51ad0c8c808082cd4e1e756
Data:  [data_issue2453.zip](https://github.com/neuropoly/spinalcordtoolbox/files/3653640/data_issue2453.zip)

Syntax:
~~~
sct_qc -i PAM50_t1_reg.nii.gz -p sct_register_multimodal -s sub-amu01_acq-T1w_MTS_seg.nii.gz -d sub-amu01_acq-T1w_MTS_crop.nii.gz
~~~

The problem is rooted in the resampling method, which works by applying an affine transformation matrix to an input image (from nibabel's code). What has likely happened in this issue (and similar issues), is that sometimes there are imprecision in the qform matrix, yielding a slight rotation of the image upon interpolation, even though in this case, the slice resolution was not changed (i.e. S-I direction). If that happens, the edge slices are replaced with "zero", as specified by the default mode=constant of the interpolation. This can be fixed by choosing another interpolation mode.

### Proposed solution

The issue was fixed by changing resampling mode from `constant` to `nearest`.

Fixes #2453, Related to #2454

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -255,6 +255,12 @@ class Image(object):
         else:
             raise TypeError('Image constructor takes at least one argument.')
 
+        # TODO: In the future, we might want to check qform_code and enforce its value. Related to #2454
+        # Check qform_code
+        # if not self.hdr['qform_code'] in [0, 1]:
+        #     # Set to 0 (unknown)
+        #     self.hdr.set_qform(self.hdr.get_qform(), code=0)
+        #     self.header.set_qform(self.hdr.get_qform(), code=0)
 
     @property
     def dim(self):
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -55,7 +55,7 @@ class Slice(object):
                 else:
                     # Otherwise it's an image: use spline interpolation
                     type_img = 'im'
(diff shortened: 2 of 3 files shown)


jcohenadad:
> Tested it, it works well!
> (However... I am not sure to understand why it works! Just curious :) )

thanks for the feedback! i've added an explanation in the PR description.

PR #2467, opened 2019-09-30
Fixed segmentation issue that mostly appeared on DWI data

### LINKED ISSUE #2464: ValueError: min() arg is an empty sequence caused by empty segmentation
For some reasons, a DWI dataset yields an empty segmentation, which causes the postprocessing module to crash. 

We should:
- figure out why the segmentation is failing
- raise a more elegant error during postprocessing if inputing an empty segmentation

To reproduce:

Data: 
[sub-brno02_dwi_crop_moco_dwi_mean.nii.gz](https://github.com/neuropoly/spinalcordtoolbox/files/3672918/sub-brno02_dwi_crop_moco_dwi_mean.nii.gz)
[sub-perform02_dwi_dwi_mean.nii.gz](https://github.com/neuropoly/spinalcordtoolbox/files/3673498/sub-perform02_dwi_dwi_mean.nii.gz)

SCT commit: 2b85953c2f22b0d9225a148173631313b8c3566d
Syntax (with error):
~~~
sct_deepseg_sc -i sub-brno02_dwi_crop_moco_dwi_mean.nii.gz -c dwi -r 0

--
Spinal Cord Toolbox (git-jca/2458-crop-2b85953c2f22b0d9225a148173631313b8c3566d)


Method:
	Centerline algorithm: svm
	Assumes brain section included in the image: False
	Dimension of the segmentation kernel convolutions: 2d


Create temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20190930152748.787127-p3a3u7kb)...
Remove temporary files...
rm -rf /var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20190930152748.955878-z_k4czod
Resample the image to 0.5x0.5 mm in-plane resolution...
load data...
load data...
Cropping the image around the spinal cord...
Normalizing the intensity...
Segmenting the spinal cord using deep learning on 2D patches...
Reassembling the image...
WARNING: To avoid intensity overflow due to convertion to uint8, intensity will be rescaled to the maximum quantization scale.
Resampling the segmentation to the native image resolution using linear interpolation...
Binarizing the resampled segmentation...
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 208, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 178, in main
    remove_temp_files=remove_temp_files, verbose=verbose)
  File "/Users/julien/code/sct/spinalcordtoolbox/deepseg_sc/core.py", line 578, in deep_segmentation_spinalcord
    im_seg_r_postproc = post_processing_volume_wise(im_seg_r)
  File "/Users/julien/code/sct/spinalcordtoolbox/deepseg_sc/postprocessing.py", line 139, in post_processing_volume_wise
    im_seg = _remove_isolated_voxels_on_the_edge(im_seg)
  File "/Users/julien/code/sct/spinalcordtoolbox/deepseg_sc/postprocessing.py", line 110, in _remove_isolated_voxels_on_the_edge
    metrics, _ = compute_shape(im_seg, angle_correction=False)
  File "/Users/julien/code/sct/spinalcordtoolbox/process_seg.py", line 61, in compute_shape
    min_z_index, max_z_index = min(Z), max(Z)
ValueError: min() arg is an empty sequence
~~~

jcohenadad:
For some reasons, some DWI datasets yield empty segmentations (or highly sub-segmented), causing the postprocessing module to crash later on.

This PR fixes:
- The segmentation issue, which was related to a resampling problem.
- Raises a more elegant error during postprocessing if inputing an empty segmentation.

Fixes #2464 

**WARNING: Output segmentation is overall larger (especially on dwi scans), therefore CSA studies should be based on this SCT version (or previous version), but not a mix of both versions.**

### DIFF
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -344,7 +344,7 @@ def segment_2d(model_fname, contrast_type, input_size, im_in):
     for zz in range(im_in.dim[2]):
         pred_seg = seg_model.predict(np.expand_dims(np.expand_dims(data_norm[:, :, zz], -1), 0),
                                      batch_size=BATCH_SIZE)[0, :, :, 0]
-        pred_seg_th = (pred_seg > 0.5).astype(int)
+        pred_seg_th = (pred_seg > 0).astype(int)
         pred_seg_pp = post_processing_slice_wise(pred_seg_th, x_cOm, y_cOm)
         seg_crop.data[:, :, zz] = pred_seg_pp
 
@@ -504,20 +504,12 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     logger.info("Resampling the segmentation to the native image resolution using linear interpolation...")
     im_seg_r = resampling.resample_nib(im_seg, image_dest=im_image, interpolation='linear')
 
-    if ctr_algo == 'viewer':  # resample and reorient the viewer labels
-        im_viewer_r = resampling.resample_nib(im_labels_viewer, new_size=input_resolution, new_size_type='mm',
-                                                    interpolation='linear')
-        # TODO: binarize it?
-    else:
-        im_viewer_r = None
+    if ctr_algo == 'viewer':  # for debugging
+        im_labels_viewer.save(sct.add_suffix(fname_orient, '_labels-viewer'))
 
+    # for debugging
     # TODO: Deal with that later-- ideally this file should be written when debugging, not with verbose=2
-    if verbose == 2:
-        im_ctl_r = resampling.resample_nib(im_ctl, new_size=input_resolution, new_size_type='mm',
-                                                    interpolation='linear')
-        # TODO: binarize it?
-    else:
-        im_ctl_r = None
+    im_ctl.save(sct.add_suffix(fname_orient, '_centerline'))
 
     # Binarize the resampled image to remove interpolation effects
     logger.info("Binarizing the resampled segmentation...")
@@ -542,6 +534,4 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     # reorient to initial orientation
     return im_seg_r_postproc.change_orientation(original_orientation), \
            im_image_res, \
-           im_seg.change_orientation('RPI'), \
-           im_viewer_r, \
-           im_ctl_r
+           im_seg.change_orientation('RPI')
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -464,7 +464,6 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     X_CROP_LST, Y_CROP_LST, Z_CROP_LST, im_crop_nii = crop_image_around_centerline(im_in=im_image_res,
                                                                                    ctr_in=im_ctl,
                                                                                    crop_size=crop_size)
-    del im_ctl
 
     # normalize the intensity of the images
     logger.info("Normalizing the intensity...")
@@ -515,12 +514,10 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
         im_viewer = None
 
     # TODO: Deal with that later-- ideally this file should be written when debugging, not with verbose=2
-    # if verbose == 2:
-    #     fname_res_ctr = sct.add_suffix(fname_orient, '_ctr')
-    #     resampling.resample_file(fname_res_ctr, fname_res_ctr, initial_resolution, 'mm', 'linear', verbose=0)
-    #     im_image_res_ctr_downsamp = Image(fname_res_ctr).change_orientation(original_orientation)
-    # else:
-    im_image_res_ctr_downsamp = None
+    if verbose == 2:
+        im_ctl_r = resampling.resample_nib(im_seg, image_dest=im_image, interpolation='linear')
+    else:
+        im_ctl_r = None
 
     # Binarize the resampled image to remove interpolation effects
     logger.info("Binarizing the resampled segmentation...")
@@ -547,4 +544,4 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
            im_image_res, \
            im_seg.change_orientation('RPI'), \
            im_viewer, \
-           im_image_res_ctr_downsamp
+           im_ctl_r
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -497,17 +497,19 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     # seg_uncrop_nii.save(sct.add_suffix(fname_res, '_seg'))  # for debugging
     del seg_crop
 
+    # Change type uint8 --> float32 otherwise resampling will produce binary output (even with linear interpolation)
+    im_seg.change_type(np.float32)
     # resample to initial resolution
     logger.info("Resampling the segmentation to the native image resolution using linear interpolation...")
     im_seg_r = resampling.resample_nib(im_seg, image_dest=im_image, interpolation='linear')
-    # TODO: the output is alreay binary! should be linear interp. To check!!
 
     if ctr_algo == 'viewer':  # for debugging
         im_labels_viewer.save(sct.add_suffix(fname_orient, '_labels-viewer'))
 
     # Binarize the resampled image to remove interpolation effects
     logger.info("Binarizing the resampled segmentation...")
-    thr = 0.0001 if contrast_type in ['t1', 'dwi'] else 0.5
+    # thr = 0.0001 if contrast_type in ['t1', 'dwi'] else 0.5
+    thr = 0.5
     # TODO: optimize speed --> np.where is slow
     im_seg_r.data[np.where(im_seg_r.data > thr)] = 1
     im_seg_r.data[np.where(im_seg_r.data <= thr)] = 0
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -84,18 +84,18 @@ def find_centerline(algo, image_fname, contrast_type, brain_bool, folder_output,
                                         dilation_layers=dct_params_ctr[contrast_type]['dilation_layers'])
         ctr_model.load_weights(ctr_model_fname)
 
-        logger.info("Resample the image to 0.5x0.5 mm in-plane resolution...")
-        fname_res = sct.add_suffix(image_fname, '_resampled')
-        input_resolution = Image(image_fname).dim[4:7]
-        new_resolution = 'x'.join(['0.5', '0.5', str(input_resolution[2])])
-
-        resampling.resample_file(image_fname, fname_res, new_resolution, 'mm', 'linear', verbose=0)
-
+        # logger.info("Resample the image to 0.5x0.5 mm in-plane resolution...")
+        # fname_res = sct.add_suffix(image_fname, '_resampled')
+        # input_resolution = Image(image_fname).dim[4:7]
+        # new_resolution = 'x'.join(['0.5', '0.5', str(input_resolution[2])])
+        #
+        # resampling.resample_file(image_fname, fname_res, new_resolution, 'mm', 'linear', verbose=0)
+        #
         # compute the heatmap
         fname_heatmap = sct.add_suffix(image_fname, "_heatmap")
         img_filename = ''.join(sct.extract_fname(fname_heatmap)[:2])
         fname_heatmap_nii = img_filename + '.nii'
-        z_max = heatmap(filename_in=fname_res,
+        z_max = heatmap(filename_in=image_fname,
                         filename_out=fname_heatmap_nii,
                         model=ctr_model,
                         patch_shape=dct_patch_ctr[contrast_type]['size'],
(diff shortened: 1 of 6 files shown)


spinalcordtoolbox/deepseg_sc/core.py
```diff
@@ -111,38 +111,28 @@ def find_centerline(algo, image_fname, contrast_type, brain_bool, folder_output,
...
-        logger.info("Resample the image to 0.5x0.5 mm in-plane resolution...")
-        fname_res = sct.add_suffix(image_fname, '_resampled')
-        input_resolution = Image(image_fname).dim[4:7]
-        new_resolution = 'x'.join(['0.5', '0.5', str(input_resolution[2])])
-
-        resampling.resample_file(image_fname, fname_res, new_resolution, 'mm', 'linear', verbose=0)
-
-        resampling.resample_file(centerline_filename, centerline_filename, new_resolution, 'mm', 'linear', verbose=0)
+    # TODO: for some reason, when algo == 'file', the absolutepath is changed to None out of the method find_centerline
+    im_ctl.absolutepath = ctl_absolute_path
```
charleygros: For some reason, when algo == 'file', I could not set the absolute path of `im_ctl`, while it works with `svm` and `viewer`.
@jcohenadad : Any ideas?

jcohenadad: no quick idea unfortunately. I will have to run the debugger to understand the cause of the problem.

charleygros:
Centerline on `dmri` data is now correct, but segmentation fails.
Seems all good for the other contrasts.

Try to compare between `master` and current branch:
- `im_crop_nii`: similar.
- `im_norm_in`: slightly different, but really small changes: I would be surprised if that's the reason why the segmentation fails.

spinalcordtoolbox/deepseg_sc/core.py
jcohenadad: btw: an alternative would be to specify file name upon saving:
~~~
im_ctl.save(ctl_absolute_path)
~~~

jcohenadad:
> Centerline on `dmri` data is now correct, but segmentation fails.
> Seems all good for the other contrasts.
> 
> Try to compare between `master` and current branch:
> 
> * `im_crop_nii`: similar.
> * `im_norm_in`: slightly different, but really small changes: not sure if that's the reason why the segmentation fails.

Sorry I'm not sure I fully understand what was done here. Did you compare the output segmentation from the master branch, with the segmentation run on this branch using either `im_crop_nii` (similar than master branch?) or `im_norm_in` (slightly different than master branch?) as an input of `segment_2d()`?

charleygros:
> Sorry I'm not sure I fully understand what was done here. Did you compare the output segmentation from the master branch, with the segmentation run on this branch using either `im_crop_nii` (similar than master branch?) or `im_norm_in` (slightly different than master branch?) as an input of `segment_2d()`?

Sorry, it was not very clear indeed.
Update: segmentation fails on dmri images on both `master` and current branch.
As the temp files (eg `im_norm_nii`) are similar between the two branches: I assume that the segmentation issue does not come from the recent changes (ie current branch).

charleygros:
Segmentation failure on `dmri` data observed with SCT 3.2.7.

charleygros:
> Update: segmentation fails on dmri images on both `master` and current branch.

Nevermind --> Was testing on an inappropriate data. Did some more extensive testing: it seems to be fixed on my side. Sorry about that.

@jcohenadad: Could you please test on your data as well? :)

jcohenadad:
getting this error (syntax + commit number included):
~~~
sct_deepseg_sc -i /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop.nii.gz -c dwi -r 0 -v 2
...
Saving image to /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop_seg.nii.gz orientation RPI shape (25, 29, 15)
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 208, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 198, in main
    im_ctr.save(fname_ctr)
AttributeError: 'NoneType' object has no attribute 'save'
~~~

jcohenadad:
> > Update: segmentation fails on dmri images on both `master` and current branch.
> 
> Nevermind --> Was testing on an inappropriate data. Did some more extensive testing: it seems to be fixed on my side. Sorry about that.
> 
> @jcohenadad: Could you please test on your data as well? :)

hum... doesn't seem to be fixed. I get the results below using commit f4019b2ccfeda946eaad5118f90878a8a581db4e on [this image](https://github.com/neuropoly/spinalcordtoolbox/files/3672918/sub-brno02_dwi_crop_moco_dwi_mean.nii.gz):

anim

I am wondering if the lack of performance comes from the normalization, because the background/cord contrast is already pretty low, and the edge of the image contains null voxels, meaning that the normalization procedure, which take all range of values in the image, flattens the cord/background contrast. 

Doing some investigations by cropping the image and running the segmentation again produces very different results, suggesting there might be something going on with the normalization

jcohenadad:
after removing the "null" edges, the segmentation is still suboptimal:
~~~
sct_crop_image -i sub-brno02_dwi_crop_moco_dwi_mean.nii.gz -xmin 5 -xmax -5 -ymin 5 -ymax -5
sct_deepseg_sc -i sub-brno02_dwi_crop_moco_dwi_mean_crop.nii.gz -c dwi
~~~
anim

charleygros:
The normalisation step is quite critical indeed.
Could you please point me to where your image is saved? Or copy it under `duke/sct_testing/issues/`?
Thanks

jcohenadad:
> The normalisation step is quite critical indeed.
> Could you please point me to where your image is saved? Or copy it under `duke/sct_testing/issues/`?
> Thanks

The image is in the issue associated with this PR

charleygros:
> getting this error (syntax + commit number included):
> 
> ```
> sct_deepseg_sc -i /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop.nii.gz -c dwi -r 0 -v 2
> ...
> Saving image to /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop_seg.nii.gz orientation RPI shape (25, 29, 15)
>   File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 208, in <module>
>     main()
>   File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 198, in main
>     im_ctr.save(fname_ctr)
> AttributeError: 'NoneType' object has no attribute 'save'
> ```

Regarding this error: This is due to `-v 2` --> This commented lines: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_sc/core.py#L562).

Happy to fix this. Quick question: What is the best way to setup a debug mode?
What about creating a `DEBUG` variable at the top of the script such as: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/fc520c3ccd5cf0cd629ea70d134221b57eb9db68/install/create_package.py#L11).

jcohenadad:
> > getting this error (syntax + commit number included):
> > ```
> > sct_deepseg_sc -i /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop.nii.gz -c dwi -r 0 -v 2
> > ...
> > Saving image to /Users/julien/data/spineGeneric_20190925/results/data/sub-perform02/dwi/sub-perform02_dwi_crop_moco_dwi_mean_crop_seg.nii.gz orientation RPI shape (25, 29, 15)
> > Traceback (most recent call last):
> >   File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 208, in <module>
> >     main()
> >   File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 198, in main
> >     im_ctr.save(fname_ctr)
> > AttributeError: 'NoneType' object has no attribute 'save'
> > ```
> 
> Regarding this error: This is due to `-v 2` --> This commented lines: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_sc/core.py#L562).
> 
> Happy to fix this. Quick question: What is the best appropriate way to setup a debug mode in the code?
> What about creating a `DEBUG` variable at the top of the script such as: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/fc520c3ccd5cf0cd629ea70d134221b57eb9db68/install/create_package.py#L11).

Hum. I would uncomment the line (but not the todo). I would like to minimize the use of a global DEBUG. It happened in the past that we would forget to set these back to 0. Ideally we should have a config file, external from SCT repos, where we would set this variables.

charleygros:
> Hum. I would uncomment the line (but not the todo). I would like to minimize the use of a global DEBUG. It happened in the past that we would forget to set these back to 0. Ideally we should have a config file, external from SCT repos, where we would set this variable.

Done! And same for the viewer labels.

However, after the resampling (back to initial resolution), we lose the center point on few slices + not binarised (since we use interpolation = linear).
We could dilate/resample/center_of_mass. But was wondering if we have something already implemented under centerline or labels_utils that deals with this (ie resampling of centerline / labels)?

charleygros:
@jcohenadad: This is not the best solution but... I got a good segmentation on both images by setting this [threshold](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_sc/core.py#L395) to 0.

Extensive testing on `dmri` data could be of interest.

On a side note, we could also think of adding this threshold to the sct_function parameters.

jcohenadad:
> However, after the resampling (back to initial resolution), we lose the center point on few slices + not binarised (since we use interpolation = linear).

@ you are only referring to the "-v 2" output anyway, right? so this has no impact on regular use of the function. If that's indeed the case, then do we really need to resample to the native resolution? I mean: do we need this output for debugging? (we could assess the quality of the centerline in the 0.5mm resolution, right?)

jcohenadad:
> @jcohenadad: This is not the best solution but... I got a good segmentation on both images by setting this [threshold](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg_sc/core.py#L395) to 0.
> 
> Extensive testing on `dmri` data could be of interest.
> 
> On a side note, we could also think of adding this threshold to the sct_function parameters.

@charleygros great catch! I will run deepseg_sc with threshold=0 on all contrasts. Based on the results, we will decide what to do (set a specific threshold for dwi? same for all contrasts? what threshold value?)

charleygros:
> > However, after the resampling (back to initial resolution), we lose the center point on few slices + not binarised (since we use interpolation = linear).
> 
> @charleygros you are only referring to the "-v 2" output anyway, right? so this has no impact on regular use of the function. If that's indeed the case, then do we really need to resample to the native resolution? I mean: do we need this output for debugging? (we could assess the quality of the centerline in the 0.5mm resolution, right?)

Yes that's right: I was referring to the "-v 2", no impact for the segmentation, makes more sense to assess the centerline in the 0.5x0.5mm2 resolution --> we could simply save this centerline into the temp folder and use this one for debugging.
--> Will do!

jcohenadad:
OK, after running `sct_deepseg_sc` on a few subjects and contrasts, I confirm that the segmentation is generally better for threshold=0, compared to threshold=0.5.

In some instances however, the segmentation was missing with threshold=0 (for reasons that need further investigations), like for the example below:

threshold=0
Screen Shot 2019-10-03 at 2 21 07 PM

threshold=0.5
Screen Shot 2019-10-03 at 2 21 16 PM

charleygros:
> OK, after running `sct_deepseg_sc` on a few subjects and contrasts, I confirm that the segmentation is generally better for threshold=0, compared to threshold=0.5.
When you said "few ... contrasts," --> did you mean that it was better on T2w, T1w ...etc. as well?


> In some instances however, the segmentation was missing with threshold=0 (for reasons that need further investigations), like for the example below:
Interesting! Happy to look into it. Is the data on `sct_testing`?

jcohenadad:
Yes, better for t2, T1w, etc

Yes it would be nice if you could look into it :-)

I used data  from spine generic: amu02, ucl01, beijingGE01

lrouhier:
Hi, it seems to be better for my test image. I have something similar to what Julien have in his previous comment (last slice missing). I will run it on all images and will let you know.
image

jcohenadad:
> Hi, it seems to be better for my test image. I have something similar to what Julien have in his previous comment (last slice missing). I will run it on all images and will let you know.
> image

thanks for the feedback @lrouhier -- in the future can you please indicate SCT commit + data used. Also when doing a comparison you need to show the before/after (otherwise no comparison possible from our perspective). Thx

lrouhier:
*sct_version* : git-jca/2464-deepseg_sc-empty-ac5357e3a73253ce6877d738818e2751f248839e
*path_to_image*:/duke/projects/ml_sci_prognosis/issue/issue_deepseg_#2470
*command run*: sct_deepseg_sc -i axial_t2_ANU-004.nii -brain 0 -qc

jcohenadad:
> _command run_: sct_deepseg_sc -i axial_t2_ANU-004.nii -brain 0 -qc

this is the wrong syntax (-c parameter is missing). If you copy/paste the output of the qc like i did [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2467#issuecomment-538067061), it (i) prevents you from making copy/paste errors and (ii) enables easier before/after comparison (same FOV and cropping-- easier to compare images).

lrouhier:
It seems to work better on the whole dataset. 
*Master version*
image

*modified version*
image

*image in*:/Volumes/projects/ml_sci_prognosis/issue/issue_deepseg_#2470

baxpr:
For what it's worth - with the 4.0.2 release I was getting an empty segmentation from sct_deepseg_sc on a more standard T2W image (not DWI) with 0.3x0.3x5.0 mm voxels that used to segment fine under an earlier version 4 (didn't record which). This commit seems to have fixed that and it's working again.

jcohenadad:
Awesome, thanks for the feedback @lrouhier and @baxpr. We will make a few additional investigations before merging this PR, notably fix the problem of the edge slices, and look at the stability of the algorithm across resolutions and contrasts. This might take a few more days.

jcohenadad:
The empty slice at z=0 is caused by `postprocessing._remove_isolated_voxels_on_the_edge()`, where the output of `metrics, _ = compute_shape(im_seg, angle_correction=False)` includes a nan:
```python
metrics['area'].data
array([      nan, 66.609375, 69.671875, 59.71875 , 67.375   , 68.140625,
       61.25    , 64.3125  , 65.84375 , 60.484375, 71.203125, 65.078125,
       62.78125 , 65.84375 , 65.84375 ])
```

jcohenadad:
OK the missing segmentation at the edge has been fixed in https://github.com/neuropoly/spinalcordtoolbox/pull/2467/commits/f6c71d1560f9988ed3794130f42f2e9a102ac6f2 (stupid mistake: forgot to divide the median by 2 to find outlier CSA at the edge)

jcohenadad:
Results across multiple contrasts, subjects, vendors and resolution:
https://www.dropbox.com/s/60aawmnzii3l4gv/20191005_SCT_issue2464_qc.zip?dl=0

Conclusion: Works great, except for the MToff contrast, but I don't think this is related to the issue related to this PR.

jcohenadad:
Results of batch_processing on OSX:

branch jca/2464-deepseg_sc-empty (commit c801fcd85368fdc5d27cdda7059429d7a2dbfc78):
~~~
t2/CSA:          74.49902509135494
mt/MTR(WM):      54.22596560112912
t2s/CSA_WM:      66.44240519216386
dmri/FA(CST_r):  0.7890679073313994
dmri/FA(CST_l):  0.7609108835133579
~~~

master 
~~~
t2/CSA:          73.98363895856627
mt/MTR(WM):      54.29594350190386
t2s/CSA_WM:      64.97073246634238
dmri/FA(CST_r):  0.7934020965216604
dmri/FA(CST_l):  0.7673452412911614
~~~

PR #2473, opened 2019-10-03
Fixed issue that appeared when trying to remove temporary folder from different file systems

### LINKED ISSUE #2472: Crashes when trying to remove folder under Singularity container
[this change](https://github.com/neuropoly/spinalcordtoolbox/commit/79145306f83157557ddb2a3fc5a13727318f4483#diff-096b5eba0646d82b4e379018d50deaa6R164) causes an issue running the sct install while building a singularity container. More details [here](https://github.com/neuropoly/spinalcordtoolbox/commit/79145306f83157557ddb2a3fc5a13727318f4483#commitcomment-35336200).

Issue likely related to:
- https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link/42400063
- https://github.com/pypa/pip/issues/103
- https://github.com/pypa/pip/issues/6943

A possible solution would be to copy the folder/file instead of moving it.

jcohenadad:
This PR accommodates fixes an issue that appeared when installing SCT under Singulary container, where a folder located in another file system could not be moved.

The implemented solution is to copy the files (instead of moving them). This PR also includes a few minor improvements, notably by replacing shutil by distutils for copying files to existing directories (not possible with shutil.copytree).

Fixes #2472

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -485,9 +485,9 @@ fi
 
 ## Create launchers for Python scripts
 echo -e "\nCreating launchers for Python scripts..."
-mkdir -p $SCT_DIR/bin
+mkdir -p $SCT_DIR/$BIN_DIR
 for file in $SCT_DIR/python/envs/venv_sct/bin/*sct*; do
-  cp "$file" "$SCT_DIR/bin/"
+  cp "$file" "$SCT_DIR/$BIN_DIR/"
   res=$?
   if [[ $res != 0 ]]; then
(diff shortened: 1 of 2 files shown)


jcohenadad:
@baxpr could you please test this branch and let me know how it goes?

baxpr:
@jcohenadad Thanks for this! I tested, and it does fix my singularity build problem.

PR #2479, opened 2019-10-07
Fine-adjustment of threshold for binarization of soft segmentation

jcohenadad:
After merging this PR: https://github.com/neuropoly/spinalcordtoolbox/pull/2467, the main issue is the slightly larger segmentation on T2w data. The problem is that some users might update SCT in the middle of processing a study, which would bring inevitable biases pre/post SCT upgrade. 

If we are to modify the output segmentation (as done in this recent PR), I suggest to spend a bit more time adjusting the threshold, in such a way that the output T1w and T2w segmentation are similar (i.e. produce the same CSA). That would notably help for registration to the PAM50 template, where a single segmentation is associated with the T1w, T2w and T2sw contrasts.

To achieve this investigation, the current PR aims at running the segmentation on the spineGeneric dataset, with a variable threshold level, and compare pairwise CSA across threshold.

More specifically, this PR includes the following changes:
- Refactoring to clarify thresholding method (and dtype) for 2D and 3D CNN inferences
- Added flag to specify threshold for binarization. If `-1`, output soft segmentation (i.e. no binarization).
- Optimized threshold for each contrast using [this method](https://github.com/sct-pipeline/deepseg-threshold), and set them as default value.


charleygros:
Note that there is also a binarisation threshold set after resampling to initial resolution --> will need to modify this section too in order to be compatible with `-thr -1`.

charleygros:
Few changes:
- Slice wise post processing --> compatible with soft segmentation
- No binarisation after resampling if thr >= 0
- `thr` added as argument of `sct_deepseg_sc`

TODO: Refactor `post_processing_volume_wise` to make it compatible with soft segmentation.

jcohenadad:
@charleygros very cool refactoring of postprocessing functions to allow soft seg input! 😃

jcohenadad:
I think we should restrict `keep_largest_object()` to int only. When binarizing soft segmentation before entering this algo, numerous objects are created, which (i) might create errors and (ii) might not be overall relevant for soft segs (what does "object" means when dealing with soft seg). See below results of the binarization before entering `keep_largest_object()` algo:

Before binarization (soft seg):
z_slice

After binarization:
z_slice_bin

The restriction to int() was implemented in 82e8ae2 (with `assert` for checking input type)

jcohenadad:
The script to test various threshold across multiple contrasts is available [here](https://github.com/sct-pipeline/spine-generic/blob/master/processing/test_deepseg_threshold.sh).

I am currently running it across the ~230 subjects of the spineGeneric DB.

jcohenadad:
The algorithm that finds the best threshold per contrast has been compiled into a [Jupyter notebook](https://github.com/sct-pipeline/deepseg-threshold/blob/master/deepseg_find_threshold.ipynb). Still work in progress, but the algorithm is there and preliminary results make sense.

jcohenadad:
Results of optimal thresholds (commit: 6e3bf3606fe60af2efe580359e071ef4b5a0ff95):
~~~
['T1', 'T2', 'T2s', 'MTon', 'T1w', 'DWI']
[0.74353448 0.34353448 0.89008621 0.33060345 0.1125     0.01422414]
~~~

jcohenadad:
The main problem with these results is for the "T1" contrast: T1 (MPRAGE) gives a 0.74353448 threshold, while T1w (GRE-T1w) gives a 0.1125 threshold... not sure what we should do.

charleygros:
> The main problem with these results is for the "T1" contrast: T1 (MPRAGE) gives a 0.74353448 threshold, while T1w (GRE-T1w) gives a 0.1125 threshold... not sure what we should do.

Is there a more predominant contrast in the nature (MPRAGE or GRE?)?
Not ideal but: If so we could assign the associated threshold to the default value for "T1 contrast" and put this information (ie Warning, if XX we recommand the use of thr=XX) in the function help or even in the terminal output.

jcohenadad:
> > The main problem with these results is for the "T1" contrast: T1 (MPRAGE) gives a 0.74353448 threshold, while T1w (GRE-T1w) gives a 0.1125 threshold... not sure what we should do.
> 
> Is there a more predominant contrast in the nature (MPRAGE or GRE?)?
> Not ideal but: If so we could assign the associated threshold to the default value for "T1 contrast" and put this information (ie Warning, if XX we recommand the use of thr=XX) in the function help or even in the terminal output.

MPRAGE is definitely more used, so let's go with the T1 threshold. I would output a warning though, because we cannot trust file names. In the future, we should continue our investigations for a unified framework that accounts for image type/contrast for generating cord segmentations. This is a challenging but much needed feature!

PR #2490, opened 2019-10-15
Fix deepseg threshold (again)

### LINKED ISSUE #2488: Adjust deepseg_sc (again)
The recently-merged PR (https://github.com/neuropoly/spinalcordtoolbox/pull/2479), which consisted in finding the optimal threshold per contrast, by minimizing the STD of CSA across contrasts, might not have been the best decision. 

A few issues I noticed after merging this PR:
- Qualitatively, the T2 and MT1_on look under-segmented (see the [QC outputs](https://github.com/neuropoly/spinalcordtoolbox/pull/2479#issue-325498952))
- The sum of the GM and WM CSA compared to the full-cord CSA on the T2 is more consistent with a zero-threshold (78 vs. 74) vs. optimal threshold (61 vs. 76).

I suggest we revert the threshold to zero, until we find a better way to optimize the threshold.

jcohenadad:
The recently-merged PR (https://github.com/neuropoly/spinalcordtoolbox/pull/2479), which consisted in finding the optimal threshold per contrast, by minimizing the STD of CSA across contrasts, was implemented with a flaw: the binarization after resampling was not done properly. This PR fixes it, and updated the optimal threshold per contrast using [this repository](https://github.com/sct-pipeline/deepseg-threshold) (commit a482d205120d76893c4f21c76e4535cc45bbce11).

Fixes #2488

### DIFF
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -25,7 +25,7 @@ BATCH_SIZE = 4
 # Thresholds to apply to binarize segmentations from the output of the 2D CNN. These thresholds were obtained by
 # minimizing the standard deviation of cross-sectional area across contrasts. For more details, see:
 # https://github.com/sct-pipeline/deepseg-threshold
-THR_DEEPSEG = {'t1': 0.74353448, 't2': 0.34353448, 't2s': 0.89008621, 'dwi': 0.01422414}
+THR_DEEPSEG = {'t1': 0.15, 't2': 0.7, 't2s': 0.89, 'dwi': 0.01}
 
 logger = logging.getLogger(__name__)
 
--- a/spinalcordtoolbox/deepseg_sc/core.py
+++ b/spinalcordtoolbox/deepseg_sc/core.py
@@ -548,7 +548,7 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     # Binarize the resampled image (except for soft segmentation, defined by threshold_seg=-1)
     if threshold_seg >= 0:
         logger.info("Binarizing the resampled segmentation...")
-        im_seg_r.data = im_seg_r.data.astype(np.uint8)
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/deepseg_sc/core.py
```diff
@@ -548,7 +548,7 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
     # Binarize the resampled image (except for soft segmentation, defined by threshold_seg=-1)
     if threshold_seg >= 0:
         logger.info("Binarizing the resampled segmentation...")
-        im_seg_r.data = im_seg_r.data.astype(np.uint8)
+        im_seg_r.data = (im_seg_r.data > 0.5).astype(np.uint8)
```
charleygros: Sorry to change my mind but: was thinking of setting 0 instead of 0.5 so that "no voxel are threshold" during this step and everything is controlled by the "optimised threshold" just after the prediction (ie "everything is controlled by only one threshold"). Does that make sense? or it does not really matter?

jcohenadad: actually, to have a proper (i.e. unbiased) thresholding, after resampling, it should be set to 0.5 (equivalent to nearest neighbor). Setting to 0 will introduce a bias.

charleygros: Right! Thanks for clarifying, makes sense!

jcohenadad:
And here are the results of the optimal threshold calculation:
~~~
  Cumulative average optimal threshold: [0.15862069 0.70732759 0.89181034 0.61206897 0.14827586 0.01336207]
~~~

charleygros:
Make much more sense! :+1:

jcohenadad:
~~~
mt/MTR(WM):      54.36934620641078
dmri/FA(CST_r):  0.7947085328716705
dmri/FA(CST_l):  0.7725978925271186
~~~

PR #2503, opened 2019-10-23
Fixed aberrant mtr values

### LINKED ISSUE #2501: Aberrant mtr calculation caused by RuntimeWarning: invalid value encountered in true_divide
In commit d5d903a7cf845bd805191907046e47c94aff5241, some data yield aberrant mtr calculation: value=178398311743488.0 everywhere.

Data: `duke:sct_testing/issues/20191022_issue2500`

Syntax:
~~~
sct_compute_mtr -mt1 mt1.nii -mt0 mt0_reg.nii
~~~

User report: http://forum.spinalcordmri.org/t/4-0-2-mtr-value-error/180

A minor other thing to fix: SCT version does not appear when running the function.

jcohenadad:
A combination of issues yielded aberrant MTR values as the output of `sct_compute_mtr`:
- The output dtype was based on the input mt1 image. If it was int, it created ugly precision.
- Division by very small numbers yielded aberrant values (highly dependent on output dtype).

This PR fixes it by:
- Forcing float32 as the output of mtr
- Create a "smart" divider, where zeros are removed before division and output values can be clipped given application-specific priors (e.g., MTR should not be higher than 100). The advantage of aggressive clipping is that the output NIFTI file has a "human-friendly" range (e.g. -100/+100), so people don't have to adjust the dynamics when opening the image in a viewer, as [can happen sometimes](http://forum.spinalcordmri.org/t/computed-mtr-resulting-in-dark-image/46).
- Added CLI param to change default threshold: `-thr`
- SCT version now appears when running the function.

Fixes #2501

### DIFF
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -77,7 +77,7 @@ def main():
     sct.printv('\nCompute MTR...', verbose)
     nii_mtr = compute_mtr(nii_mt1=Image(args.mt1), nii_mt0=Image(args.mt0))
     # save MTR file
-    nii_mtr.save(fname_mtr)
+    nii_mtr.save(fname_mtr, dtype='float64')
 
     sct.display_viewer_syntax([args.mt0, args.mt1, fname_mtr])
 
--- a/scripts/sct_compute_mtr.py
+++ b/scripts/sct_compute_mtr.py
@@ -80,12 +80,11 @@ def main():
     args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
     fname_mtr = args.o
     verbose = args.v
-    # Update log level
-    sct.init_sct(log_level=verbose, update=True)
 
     # compute MTR
     sct.printv('\nCompute MTR...', verbose)
     nii_mtr = compute_mtr(nii_mt1=Image(args.mt1), nii_mt0=Image(args.mt0), threshold_mtr=args.thr)
+
     # save MTR file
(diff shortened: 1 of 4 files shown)


spinalcordtoolbox/qmri/mt.py
```diff
@@ -13,16 +13,41 @@
 logger = logging.getLogger(__name__)
 
 
-def compute_mtr(nii_mt1, nii_mt0):
+def divide_after_removing_zero(dividend, divisor, threshold, replacement=np.nan):
```
charleygros: GReat!

scripts/sct_compute_mtr.py
```diff
@@ -45,22 +45,32 @@ def get_parser():
         )
 
     optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        "-thr",
+        type=float,
+        help="Threshold to clip MTR output values in case of division by small number. This threshold is applied on the"
+             "absolute value. Default: 100.",
+        default=100
```
charleygros: So values will be between -100 and 100, is it right? or am I misunderstanding?
If so, could it be helpful to mention this in the help?

jcohenadad: agreed-- clarified in 509c99b2ee3d4659d61156a34912f6fd4497f993

charleygros: great!

Issue #2509, opened 2019-10-25
Enable spinal-level-based registration to the PAM50


jcohenadad:
In projects where spinal (not vertebral) levels are known (example [here](https://github.com/sct-pipeline/exvivo-template/issues/1)), it would be desirable to have the option to register these data to the PAM50, which already has a spinal level atlas. 

A suggestion for implementation includes (based on @charleygros's ideas):
- For each PAM50/spinal_levels/spinal_levels_*.nii.gz, compute the center of mass of the distribution.
- Generate a single nifti file (could be called `PAM50_spinal_levels.nii.gz`) and create a voxel for each of the identified spinal level center of mass. The value of the level will be: 2 for spinal level 2, etc.
- Update the `PAM50` package to include this new file, upload to OSF+neuropoly website and update `sct_download_data` with the new URLs
- Refactor `sct_register_to_template` to add a new flag `-lspinal`, which will point to the file `PAM50_spinal_levels.nii.gz` for doing a z-based non-linear registration of spinal levels between the source and the dest (PAM50) images (step=0 of the algo).

charleygros:
Fixed with #2513 .

PR #2513, opened 2019-10-27
Spinal-level-based registration to the PAM50

charleygros:
This PR aims at enabling spinal-level-based registration to the PAM50, as described in #2509.

### Done
- `dev/spinal_level/create_PAM50_label_spinal_level.py`.
- Update the PAM50 package to include this new file + update `info_label.txt`, upload to OSF+neuropoly website and update sct_download_data with the new URLs 
- Refactor sct_register_to_template to add a new flag -lspinal, which will point to the file PAM50_spinal_levels.nii.gz for doing a z-based non-linear registration of spinal levels between the source and the dest (PAM50) images (step=0 of the algo).
- refactoring of `get_label_file` function from `spinalcordtoolbox/metadata.py`.

TODO:
- symmetrize the registered ex vivo template 

**WARNING:** SCT needs to be reinstalled after merging this PR because of the updated data/ package. Due to the potentially large impact, let's merge this PR into a minor (not patch) release.


charleygros:
### New script:
- `dev/spinal_level/create_PAM50_label_spinal_level.py`
- For each PAM50/spinal_levels/spinal_levels_*.nii.gz, compute the center of mass of the distribution.
- Generate a single nifti file (`PAM50_label_spinal_levels.nii.gz`) and create a voxel for each of the identified spinal level center of mass. The value of the level will be: 2 for spinal level 2, etc.

```
(venv_sct) charley@charley:~/sct$ sct_label_utils -i data/PAM50/template/PAM50_label_spinal_levels.nii.gz -display

--
Spinal Cord Toolbox (git-test-0581070b981d553b34ffbe4ec7f88a33811b4c8f)

Position=(70,70,945) -- Value= 2
Position=(70,70,927) -- Value= 3
Position=(70,70,891) -- Value= 4
Position=(70,70,857) -- Value= 5
Position=(70,70,826) -- Value= 6
Position=(70,71,794) -- Value= 7
Position=(70,71,757) -- Value= 8
Position=(70,70,713) -- Value= 9
Position=(70,71,668) -- Value= 10
Position=(70,71,622) -- Value= 11
Position=(70,71,572) -- Value= 12
Position=(70,71,522) -- Value= 13
Position=(70,71,471) -- Value= 14
Position=(70,71,417) -- Value= 15
Position=(70,71,365) -- Value= 16
Position=(70,71,310) -- Value= 17
Position=(70,71,251) -- Value= 18
Position=(70,70,194) -- Value= 19
Position=(70,70,169) -- Value= 20
All labels (useful syntax):
70,70,945,2:70,70,927,3:70,70,891,4:70,70,857,5:70,70,826,6:70,71,794,7:70,71,757,8:70,70,713,9:70,71,668,10:70,71,622,11:70,71,572,12:70,71,522,13:70,71,471,14:70,71,417,15:70,71,365,16:70,71,310,17:70,71,251,18:70,70,194,19:70,70,169,20
```

charleygros:
@jcohenadad: Not sure about the variable `vertebral_alignement` ([here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_register_to_template.py#L276)): from what I understand, it does a cropping with an offset in the (IS) direction when set at `True` (eg for `label_type == True`.
My intuition is that we also want this for the spinal levels, is it correct?

charleygros:
@jcohenadad: Do you agree with:
- filename: `PAM50_label_spinal_levels.nii.gz`
- new line in `info_label.txt`: `14, spinal body labeling, PAM50_label_spinal_levels.nii.gz`

Just checking before uploading the new PAM50 package on OSF/neuropoly wiki.

jcohenadad:
> @jcohenadad: Do you agree with:
> 
> * filename: `PAM50_label_spinal_levels.nii.gz`
> * new line in `info_label.txt`: `14, spinal body labeling, PAM50_label_spinal_levels.nii.gz`
> 
> Just checking before uploading the new PAM50 package on OSF/neuropoly wiki.

i would replace the file `info_label.txt` with this one. It has a couple of improvements, clarifications, etc.

[info_label.txt](https://github.com/neuropoly/spinalcordtoolbox/files/3779873/info_label.txt)

scripts/sct_register_to_template.py
```diff
@@ -110,6 +110,15 @@ def get_parser():
                       mandatory=False,
                       default_value='',
                       example="anat_labels.nii.gz")
+    parser.add_option(name="-lspinal",
+                      type_value="file",
+                      description="Labels located at the center of the spinal cord, on the mid-spinal level slice. If you are using "
```
jcohenadad: i would replace the description with:
Labels located in the center of the spinal cord, at the superior-inferior level corresponding to the mid-point of the spinal level. Each label is a single voxel, which value corresponds to the spinal level (e.g.: 2 for spinal level 2). If you are using more than 2 labels, all spinal levels covering the region of interest should be provided (e.g., if you are interested in levels C2 to C7, then you should provide spinal level labels 2,3,4,5,6,7).

scripts/sct_register_to_template.py
```diff
@@ -241,7 +253,10 @@ def main(args=None):
     zsubsample = param.zsubsample
 
     # retrieve template file names
-    file_template_vertebral_labeling = get_file_label(os.path.join(path_template, 'template'), 'vertebral labeling')
+    if label_type == 'spinal':
+        file_template_labeling = get_file_label(os.path.join(path_template, 'template'), 'spinal body labeling')
```
jcohenadad: ah! my refactoring of the field "name" in the file `info_label.txt` will break this part of the code. Instead of fetching file name by field "name", we should use the field "ID", so we have more flexibility to alter the file `info_label.txt`. What do you think @charleygros? That would imply a minor refactoring of the function `get_file_label()` to input ID instead of name (then we need to check where the function is used).

scripts/sct_register_to_template.py
```diff
@@ -250,15 +265,15 @@ def main(args=None):
 
     # get fname of the template + template objects
     fname_template = os.path.join(path_template, 'template', file_template)
-    fname_template_vertebral_labeling = os.path.join(path_template, 'template', file_template_vertebral_labeling)
+    fname_template_labeling = os.path.join(path_template, 'template', file_template_labeling)
```
jcohenadad: good call!

scripts/sct_register_to_template.py
charleygros: Done, thanks!

scripts/sct_register_to_template.py
charleygros: Seems a good idea to me!
Was wondering why we prefered the names instead of the IDs. It was maybe for this case: [here](https://github.com/neuropoly/spinalcordtoolbox/blob/4c2aca38d2536e202d58b9be646652a0f4280b6a/spinalcordtoolbox/vertebrae/core.py#L67). Was it?

charleygros: Happy to work on this!

charleygros: After more thoughts, I think that fetching by name was easier than ID (not self explanatory).
Alternatively, we could go with the IDs, and add a comment on those modified lines, something like:
```python
          fname_src,
          spinalcordtoolbox.metadata.get_file_label(path_template, '1', output="filewithpath"),  # label = 'T2-weighted template'
          spinalcordtoolbox.metadata.get_file_label(path_template, '5', output="filewithpath"),  # label = 'gray matter mask (probabilistic)'
          spinalcordtoolbox.metadata.get_file_label(path_template, '4', output="filewithpath")  # label = 'white matter mask (probabilistic)'
```
What do you think @jcohenadad ?

jcohenadad: Yes, commenting the code is exactly what I had in mind 👍

charleygros: Super!
Done

charleygros:
Now that we agreed on `info_label.txt` and `PAM50_label_spinal_levels.nii.gz` --> Do I get your green light for uploading PAM50 package on OSF & neuropoly wiki?

spinalcordtoolbox/metadata.py
```diff
@@ -224,3 +224,17 @@ def get_indiv_label_names(directory):
 
     return tuple([_name for (_id, _name, _file) in il._indiv_labels])
 
+def get_indiv_label_ids(directory):
```
charleygros: Not sure if we want to create a new function or to refactor the function `get_indiv_label_names`, which could return two tuples (one for names, one for IDs).

unit_testing/test_metadata.py
```diff
@@ -56,6 +56,6 @@ def test_read_label_file(info_labels):
 
 def test_get_file_label(info_labels):
     for info_label in info_labels:
-        _in = spinalcordtoolbox.metadata.get_indiv_label_names(os.path.dirname(info_label))
-        spinalcordtoolbox.metadata.get_file_label(os.path.dirname(info_label), _in[0], output="file")
-        spinalcordtoolbox.metadata.get_file_label(os.path.dirname(info_label), _in[0], output="filewithpath")
+        _in = spinalcordtoolbox.metadata.get_indiv_label_ids(os.path.dirname(info_label))
```
charleygros: Here the reason why I created the function `get_indiv_label_ids`.

spinalcordtoolbox/metadata.py
jcohenadad: Ok for the new function, but ideally we should output an object with the id, name and file name. That would minimize the code

charleygros: Agree! Done :-)

spinalcordtoolbox/metadata.py
```diff
@@ -210,31 +211,26 @@ def get_file_label(path_label='', id_label=0, output='file'):
...
-    :return: the ids (int)
-    """
-
-    file_info_label = 'info_label.txt'
-    il = InfoLabel()
-    fname_label = os.path.join(directory, file_info_label)
-    il.load(fname_label)
-
-    return tuple([id_ for (_id, _name, _file) in il._indiv_labels])
+    return {'id_label': tuple(id_lst),
```
jcohenadad: i would use the same field names as in the info_label file, i.e.: 'id', 'name' and 'file' (although strictly speaking id should be capitalized...)

charleygros: Done :+1:

charleygros:
PAM50 package updated on both OSF and neuropoly wiki.
PR ready for review :-)

charleygros:
Travis fails because of
```
sct_label_vertebrae -i t2/t2.nii.gz -s t2/t2_seg.nii.gz -c t2 -initfile t2/init_label_vertebrae.txt -t template -qc testing-qc
```
-->
```
RuntimeError: Label ID 1 not found in /home/travis/build/neuropoly/spinalcordtoolbox/sct_testing_data/template/template/info_label.txt
```

This is due to a mismatch of `ID` between `data/PAM50/template/info_label.txt` and `sct_testing/template/template/info_label.txt` for `PAM50_t2.nii.gz`: 1 vs 0.
This bug was not detected before because we were using the label names (vs the IDs) which are similar in both files.

@jcohenadad: Would you agree to replace the current `sct_testing/template/template/info_label.txt`:
```
# Template labels
# Keyword=IndivLabels (Please DO NOT change this line)
# ID, name, file
0, T2-weighted template, PAM50_small_t2.nii.gz
3, spinal cord, PAM50_small_cord.nii.gz
7, vertebral labeling, PAM50_small_levels.nii.gz
```

by:
```
# Template labels
# Keyword=IndivLabels (Please DO NOT change this line)
# ID, name, file
1, T2-weighted template, PAM50_small_t2.nii.gz
3, spinal cord, PAM50_small_cord.nii.gz
7, vertebral labeling, PAM50_small_levels.nii.gz
```

on both OSF & neuropoly wiki + update `sct_download_data`.

scripts/sct_register_to_template.py
```diff
@@ -274,7 +292,7 @@ def main(args=None):
     labels = check_labels(fname_landmarks, label_type=label_type)
 
     vertebral_alignment = False
```
charleygros: Would it make sense to rename this variable `label_alignment`?

jcohenadad: yup! it would definitely would. I would call it call it `level_alignment` though

jcohenadad:
Results of `batch_processing.sh` (ran on Abbey/Ubuntu 16.04):

**4.1.1**
~~~
mt/MTR(WM):      54.364850159765645
dmri/FA(CST_r):  0.7947429202929487
dmri/FA(CST_l):  0.7726948674392992
~~~

**cg/reg_spinal_level (ab7dd83418877b87f59e643e3ae1a6fbd8e12f06)**
~~~
TBD
~~~

PR #2518, opened 2019-11-05
Fixed issue when passing absolute path to script

### LINKED ISSUE #2517: sct_run_batch: crashes when pointing to non-local parameters or script
When calling for example this:
~~~
sct_run_batch /users/bob/parameters.sh myscript.sh
~~~
the function would crash (error message saying that the file parameters cannot be found), whereas copying the file locally and running this:
~~~
sct_run_batch parameters.sh myscript.sh
~~~
works fine.

So there is something to fix in the way input path are handled in this script.

valosekj:
I fixed building of path for input arguments/files for `sct_run_batch.sh ` script. Script failed when input files were not in same directory as the script or when input files were entered with absolute path (script expected only file names).

Fixes #2517

### DIFF
--- a/shell/sct_run_batch.sh
+++ b/shell/sct_run_batch.sh
@@ -24,11 +24,20 @@ On_Black='\033[40m'  # Black
 
 # Functions
 # =============================================================================
+
+check_file_exist() {
+  local path="$1"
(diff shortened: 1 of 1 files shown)


jcohenadad:
Thank you @valosekj !

PR #2520, opened 2019-11-11
Now possible to add existing label from an external file in the create-viewer option

lrouhier:
(WIP do not merge) 
The goal of this PR is to create an option to see previously existing labels when using the `-create-viewer ` option in `sct_label_utils`. The user could then add some missing labels or modify the existing one within the GUI. 
DONE : 
* Create a `-ilabel` option allowing to enter the existing label files and extract the existing label coordinates and value 
* Modify the existing `SagittalController` in gui/sagittal.py to create the existing points on the GUI when showing the image. 

*  When the user fails to enter enough label. example: `sct_label_utils -i t2.nii.gz  -ilabel labels.nii.gz -create-viewer 1,2` will create the GUI with all pre-existing labels as discussed in this PR.  

* Modify the code to project the existing point onto the mid_saggital plane.

### DIFF
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -69,7 +69,6 @@ class ProcessLabels(object):
                 self.fname_output = fname_output
         else:
             self.fname_output = fname_output
-        
         self.fname_previous = fname_previous
         self.cross_radius = cross_radius
         self.vertebral_levels = vertebral_levels
@@ -855,7 +854,6 @@ def main(args=None):
         input_fname_previous = arguments['-ilabel']
     else:
         input_fname_previous = None
-    
     verbose = int(arguments.get('-v'))
     sct.init_sct(log_level=verbose, update=True)  # Update log level
 
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -640,6 +640,8 @@ class ProcessLabels(object):
         params.input_file_name = self.image_input.absolutepath
         params.output_file_name = self.fname_output
         params.subtitle = self.msg
+        if previous_points is not None: 
+            params.message_warn = 'Please select the label you want to add \nor correct in the list below before clicking \non the image'
         output = msct_image.zeros_like(self.image_input)
         output.absolutepath = self.fname_output
         launch_sagittal_dialog(self.image_input, output, params, previous_points)
@@ -708,6 +710,11 @@ def get_parser():
                       type_value=[[','], 'int'],
                       description='Manually label from a GUI a list of labels IDs, separated with ",". Example: 2,3,4,5',
                       mandatory=False)
+    parser.add_option(name="-ilabel",
+                      type_value="file",
+                      description="File that contain labels that you want to correct. It is possible to add new points with this option. Use with -create-viewer",
+                      mandatory=False,
+                      example="t2_labels_auto.nii.gz",)
     parser.add_option(name='-cubic-to-point',
                       type_value=None,
                       description='Compute the center-of-mass for each label value.',
@@ -761,11 +768,6 @@ def get_parser():
                       mandatory=False,
                       example="t2_labels_cross.nii.gz",
                       default_value="labels.nii.gz")
-    parser.add_option(name="-ilabel",
-                      type_value="file",
-                      description="File that contain label that you want to correct. It is possible to add new points with this option. Use with -create-viwer",
-                      mandatory=False,
-                      example="t2_labels_auto.nii.gz",)
     parser.add_option(name="-v",
                       type_value="multiple_choice",
                       description='Verbose. 0: nothing. 1: basic. 2: extended.',
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -138,7 +138,7 @@ class ProcessLabels(object):
                 previous_lab.change_orientation('SAL')
                 mid=int(np.round(previous_lab.data.shape[2]/2))
                 previous_points=np.transpose(previous_lab.data.nonzero())
-               #boolean used to mark first element t initiate the list. 
+               #boolean used to mark first element to initiate the list. 
                 first=True
 
                 for i in range (len(previous_points)):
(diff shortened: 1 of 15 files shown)


spinalcordtoolbox/gui/sagittal.py
```diff
@@ -20,8 +20,12 @@
 
 
 class SagittalController(base.BaseController):
-    def __init__(self, image, params, init_values=None):
+    def __init__(self, image, params, init_values=None,previous_point=None):
```
jcohenadad: could you please add a space after "," (PEP8)

lrouhier: Yes will do.

scripts/sct_label_utils.py
```diff
@@ -69,6 +69,14 @@ def __init__(self, fname_label, fname_output=None, fname_ref=None, cross_radius=
                 self.fname_output = fname_output
         else:
             self.fname_output = fname_output
+
+        if isinstance(fname_previous, list):
```
jcohenadad: why the need for this if? how about simply enforcing `fname_previous` to be a file name (could be verified by the parser)

lrouhier: I think you are right it will be enough. I just modeled it after the fname_output  to begin with.

scripts/sct_label_utils.py
```diff
@@ -138,14 +138,32 @@ def process(self, type_process):
...
+                        if first: 
+                            points=np.append(previous_points[i], np.array([previous_lab.data[previous_points[i][0], previous_points[i][1], previous_points[i][2]]]), axis=-1)
+                            points=np.reshape(points, (1, 4))
+                            previous_label=points
+                            first=False
+                        else:
+                            points=np.append(previous_points[i], np.array([previous_lab.data[previous_points[i][0], previous_points[i][1], previous_points[i][2]]]), axis=-1)
+                            points=np.reshape(points, (1, 4))
+                            previous_label=np.append(previous_label, points, axis=0)
+
```
jcohenadad: no need for these extra lines

lrouhier:
update on the decision: when the user fails to enter all the labels already existing in the -ilabel argument files, it will just show the ones he asked for. 
Existing points are projected onto the mid-sagittal plane as this is what shows up in the GUI, it is easier for the user to correct the label.

jcohenadad:
> update on the decision: when the user fails to enter all the labels already existing in the -ilabel argument files, it will just show the ones he asked for.

sounds good (if i understand correctly)

> Existing points are projected onto the mid-sagittal plane as this is what shows up in the GUI, it is easier for the user to correct the label.

that's the part i don't understand: if the gui only show the labels he asked for, how can he correct for the one he didn't ask for?

lrouhier:
the -ilabel flag launch an algorithm that retrieves the existing coordinates. The retrieved coordinates are then passed as input for the GUI which considers them the same way it does selected points. The checklist on the right allows the modification of the coordinates of each label (retrieved from another file or clicked on).

The drawback is that the points that are not asked for won’t be part of the new label image (for now). I was thinking that it was more coherent to just save what the user sees

jcohenadad:
> the -ilabel flag launch an algorithm that retrieves the existing coordinates. The retrieved coordinates are then passed as input for the GUI which considers them the same way it does selected points. The checklist on the right allows the modification of the coordinates of each label (retrieved from another file or clicked on). The user can correct the labels he did ask for if one is misplaced.
> 
> The drawback is that the points that are not asked for won’t be part of the new label image (for now). I was thinking that it was more coherent to just save what the user sees

sorry i still don't understand-- let's me write a scenario to clarify:

- users runs: `sct_label_utils -i bla -create-viewer 3,4,5,6 -ilabel FILE_WITH_LABELS_2-3-4`
- GUI shows:
  - on the left (not on the right): the list: 2,3,4,5,6
  - on the image: points: 2,3,4
- User can select/unselect whatever he wants. He can correct labels 2,3,4 if he wants
- When clicking on save, labels 2,3,4,5,6 are saved

is that what you suggested?

if not, please explain with a step-by-step scenario for clarity-- thanks

lrouhier:
Not entirely. the scenario is the following: 
* user run `sct_label_utils -i bla  -create-viewer 3,4,5,6 -ilabel FILE_WITH_LABELS_2-3-4`
* GUI shows : 
    * on the left : 3,4,5,6
    * on the image : points 3,4 (common points between what is already defined and what he asked for)
* he can select what he wants and correct 3,4 if he wants to. 
* when clicking, on save labels 3,4,5,6 are saved

jcohenadad:
> Not entirely. the scenario is the following:
> 
> * user run `sct_label_utils -i bla  -create-viewer 3,4,5,6 -ilabel FILE_WITH_LABELS_2-3-4`
> * GUI shows :
>   
>   * on the left : 3,4,5,6
>   * on the image : points 3,4 (common points between what is already defined and what he asked for)
> * he can select what he wants and correct 3,4 if he wants to.
> * when clicking, on save labels 3,4,5,6 are saved

ok got it!  how about adding 5,6 on the right, to give the user the possibility to correct for them?  what are the cons?

lrouhier:
> ok got it! how about adding 5,6 on the right, to give the user the possibility to correct for them? what are the cons?

Do you mean 2? ( I assumed `FILE_WITH_LABELS_2-3-4` contained only label 2, 3 and 4  ). 
The user can already add 5,6 on the image (they are not defined yet).

jcohenadad:
> > ok got it! how about adding 5,6 on the right, to give the user the possibility to correct for them? what are the cons?
> 
> Do you mean 2? ( I assumed `FILE_WITH_LABELS_2-3-4` contained only label 2, 3 and 4 ).
> The user can already add 5,6 on the image (they are not defined yet).

yes i meant 2, sorry

lrouhier:
I don't see any cons to it right now. I'll implement it and report if I find any problem with it.

lrouhier:
I will modify the TO DO right now. Both of these are now implemented.

scripts/sct_label_utils.py
```diff
@@ -725,6 +762,11 @@ def get_parser():
                       mandatory=False,
                       example="t2_labels_cross.nii.gz",
                       default_value="labels.nii.gz")
+    parser.add_option(name="-ilabel",
```
jcohenadad: move this option right below -create-viewer. It will be more intuitive

PR #2529, opened 2019-11-15
Fixed compatibility between SCT/ANTs and FSL warping fields

### LINKED ISSUE #2525: Transforming ITK-style warping field to FSL/SPM-style produces error when used
When trying to convert the ITK-based warping field (produced by SCT) to an FSL/SPM-style, as instructed in the [SCT course (page 49), version:5](https://osf.io/arfv7/?version=5) it produces an error when trying to use it. Below the code to reproduce the error (based on `batch_processing.sh` output):

Tested on OSX Mojave with SCT 4.1.0.

Syntax:
~~~
cd sct_example_data/t2
sct_image -i warp_template2anat.nii.gz -mcs -o warp_vec.nii.gz
sct_image -i warp_vec_*.nii.gz -concat t -o warp_template2anat_fsl.nii.gz 
applywarp -i /Users/julien/code/sct/data/PAM50/template/PAM50_t2.nii.gz -r t2.nii.gz -w warp_template2anat_fsl.nii.gz -o PAM50_t2_reg_fsl.nii.gz
~~~

Error:
~~~
error: inv(): matrix appears to be singular

An error occured while reading file: warp_template2anat_fsl.nii.gz
~~~

Related to: http://forum.spinalcordmri.org/t/errors-converting-sct-generated-warps-for-use-with-fsl/233

jcohenadad:
The main purpose of this PR is to make it possible to convert the ITK-based warping field (produced by SCT) to an FSL/SPM-style. When following instructions of the [SCT course (page 49) version:5](https://osf.io/arfv7/?version=5), it produces an error or wrong results. This is due to the fact that the definition of warps between ITK and FSL is largely different: beyond the shaping of the warping field, the concatenation of affine/nonlinear warps and the coordinate systems (world vs. voxel) differ. 

Fixes #2525

Related to: http://forum.spinalcordmri.org/t/errors-converting-sct-generated-warps-for-use-with-fsl/233


bcdarwin:
I have `git cherry-pick`ed the last few commits here (`git rebase` wasn't immediately working) and pushed to bcdarwin/cherry-pick-2529.  If we want to keep discussion in this PR, I guess the most sensible thing is for me to force-push to jca/fsl-warp?  @jcohenadad and others, let me know if you're OK with this.

(Similarly for the other 10 or so open PRs. I haven't fixed those up yet but it would only take a few minutes.)

jcohenadad:
> the most sensible thing is for me to force-push to jca/fsl-warp? @jcohenadad and others, let me know if you're OK with this.

OK 👍 

> Similarly for the other 10 or so open PRs. I haven't fixed those up yet but it would only take a few minutes.

that would be awesome! thanks

bcdarwin:
Pushed.

cfhammill:
Ok so I have a work in progress conversion. It works for both `sct_straighten_spinalcord` and `sct_register_multimodal` derived deformation fields.

The trick was noticing in [https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1509&L=FSL&D=0&P=178218] that the FSL deformation fields are neither in voxel nor world coordinates but an FSL specific space that is the voxel coordinates (after mirroring in x if the affine space has positive determinant) multiplied by the voxel spacings. The proposed addition gives an extra flag (`-to-fsl`) to sct_image to generate FSL style absolute warps. E.g

```
sct_image -i warp_curv2straight.nii.gz \
  -o warp_curv2straight_absfsl.nii.gz \
  -to-fsl t2.nii.gz

applywarp -i t2.nii.gz -r t2_straight.nii.gz -w warp_curv2straight_absfsl.nii.gz \
  --abs \
  -o t2_straight_fsl.nii.gz
```

The absolute transform can be easily converted to relative using convertwarp

```
convertwarp --abs --relout -i warp_curv2straight_absfsl.nii.gz \
  -o warp_curv2straight_relfsl.nii.gz
```

@jcohenadad can you have a look and let me know what you think?

jcohenadad:
@cfhammill FANTASTIC!!! We are indeed getting there. Thank you so much!🥇  

I tried it for curve2straight and it indeed worked perfectly. However, when I tried it the other way around (straight2curve), I end up with a slight shift in the IS direction. See code below to reproduce:
~~~
sct_image -i warp_straight2curve.nii.gz -o warp_straight2curve_absfsl.nii.gz -to-fsl t2_straight.nii.gz
applywarp -i t2_straight.nii.gz -r t2.nii.gz -w warp_straight2curve_absfsl.nii.gz --abs -o t2_fsl.nii.gz
sct_apply_transfo -i t2_straight.nii.gz -w warp_straight2curve.nii.gz -d t2.nii.gz -o t2_itk.nii.gz
~~~

Output (t2, t2_fsl, t2_itk):
anim

cfhammill:
I think this has to do with the difference between the affine space of the straight2curve warp and the source image I mentioned in https://github.com/neuropoly/spinalcordtoolbox/issues/2525#issuecomment-583921286. I tried the basic thing of replacing the affine for the warp with the proper one from the source image, but that mostly fixed the shift but added a slight rotation. 

I'll work on solving this, I suspect this is limited to `sct_straighten_spinalcord`. I'm surprised that ANTs can apply the transform, the affine and shape are different (64, 320, 328) for the warp and (64, 320, 320) for t2.nii.gz

jcohenadad:
> I'll work on solving this, I suspect this is limited to sct_straighten_spinalcord. I'm a bit surprised that ANTs can apply the transform, the affine and shape are different (64, 320, 328) for the warp and (64, 320, 320) for t2.nii.gz

I'm sure you've looked into it, but could it be a slight difference between the qform (used by ANTs) and the sform (used by FSL)?

cfhammill:
So for both `sct_straighten_spinalcord` and `sct_register_multimodal` the maximum elementwise difference between the qform and the sform is < 10^-6. So I don't think that's the issue in this case. I will need to check to see if we should be worried about this difference generally however. Up to now I've been trusting nibabel's get_best_affine.

`-to-fsl` works properly on the inverse transform derived from `sct_register_multimodal`:

```sh
sct_register_multimodal \
  -i t2.nii.gz \
  -d t2_straight.nii.gz \
  -o t2_straight_ants.nii.gz \
  -owarp ants_curve2straight.nii.gz

mv warp_t2_straight2t2.nii.gz ants_straight2curve.nii.gz

sct_apply_transfo \
  -i t2_straight.nii.gz \
  -d ants_straight2curve.nii.gz \
  -d t2.nii.gz \
  -w ants_straight2curve.nii.gz \
  -o t2_ants.nii.gz

sct_image \
  -i ants_straight2curve.nii.gz \
  -o warp_s2c_fsl.nii.gz \
  -to-fsl t2_straight.nii.gz

applywarp \
  -i t2_straight.nii.gz \
  -r t2.nii.gz \
  -w warp_s2c_fsl.nii.gz \
  --abs \
  -o t2_fsl.nii.gz

fsleyes t2_fsl.nii.gz t2_ants.nii.gz
```

There are some minor infidelities along the boundary of the image, and the background voxels are slightly different, but otherwise identical. 

For `sct_straighten_spinalcord` the problem is that the target for applying the transform (t2.nii.gz) is different then the target used to generate the deformation (a padded version of the t2, with ~3 voxel z-translation).  

This may be fixable by generating a matrix for `--postmat`. I will experiment and report back. Ideally I'd like to fix it inside sct_image to reduce the number of moving parts for users.

cfhammill:
Taking the transformation from the target space to the deformation field space and applying it as a `--postmat` mostly works, provided the target is converted to the same orientation as the deformation field first. You lose some voxels along the boundary that get rotated out of field prior to the `--postmat` application. I'll see if I can fix that and try to migrate the logic in to `sct_image`

jcohenadad:
> For sct_straighten_spinalcord the problem is that the target for applying the transform (t2.nii.gz) is different then the target used to generate the deformation (a padded version of the t2, with ~3 voxel z-translation).

Maybe we could also reconsider the way this deformation field is generated, and not pad it? AFAIR we did it to fix extrapolation issues at the edge (i.e. the "zeros" voxels produced by FSL when extrapolating towards NaN values), but in the meantime the codebase of the straightening algo changed quite a lot so we might be able to address that issue differently.

cfhammill:
I don't think changing `sct_straighten_spinalcord` is necessary. I think I have working solution. We can, inside `sct_image`, resample the deformation field to have the same affine as the target image (preserving most of the pad).

Downside is the resampling with nibabel is pretty slow. Processing time goes from ~ 1 second to ~ 30 seconds. I don't think that's a deal breaker however.

jcohenadad:
> I don't think changing sct_straighten_spinalcord is necessary. I think I have working solution. We can, inside sct_image, resample the deformation field to have the same affine as the target image (preserving most of the pad).
Downside is the resampling with nibabel is pretty slow. Processing time goes from ~ 1 second to ~ 30 seconds. I don't think that's a deal breaker however.

I guess that resampling would only be applied when running the straightening, right?  Once we validate this solution is working, maybe we can set up the correct affine at the first place (i.e. when the deformation field is being created, as opposed to patching it after)?  That could be left for another PR though.

cfhammill:
It would make sense to do the resampling in `sct_straighten_spinalcord`. I will work on doing some validation. Do you have any thoughts on how to validate/test this?

jcohenadad:
Proposal for testing: @lrouhier @AnBucquet 
- Run the London'20 SCT course batch (single subject), then transform the warping fields to be FSL-compatible, then run `applywarp` with both forward and backward transfo, and verify that images match the output of `sct_apply_transfo` with the native warping fields.

@cfhammill : could you please find ways to fetch `applywarp` from Travis (check license issue). E.g. host the binary on OSF and curl it from Travis.

jcohenadad:
@AnBucquet @lrouhier do you have any update on testing this PR? If not, could you please do it Friday so we can merge this PR before the week-end. Thx!

lrouhier:
It works for me. There is no shift when i use `sct_image -i warp_straight2curve.nii.gz -o warp_straight2curve_absfsl.nii.gz -to-fsl t2_straight.nii.gz,t2.nii.gz` to generate the fsl warping field, However, I needed Chris input to do so. We need to add in the doc that `to_fsl` can take a list of file as input. 
Also, there is a slight intensity variation between the `sct_apply_transfo` and `applywarp` method. I don't think it is a problem but I just want to document it. You can see it by the value of the pixel at the bottom right of the image below.
image

jcohenadad:
@lrouhier according to [this post](https://github.com/neuropoly/spinalcordtoolbox/pull/2529#issuecomment-589997662), `-to-fsl` only takes the source image (not a list of source,dest). That's also the way I tried it (and it worked).

lrouhier:
@jcohenadad  I tried it on t2 with 
~~~
sct_image -i warp_straight2curve.nii.gz -o warp_straight2curve_absfsl.nii.gz -to-fsl t2_straight.nii.gz
applywarp -i t2_straight.nii.gz -r t2.nii.gz -w warp_straight2curve_absfsl.nii.gz --abs -o t2_fsl.nii.gz
sct_apply_transfo -i t2_straight.nii.gz -w warp_straight2curve.nii.gz -d t2.nii.gz -o t2_itk.nii.gz
~~~
And I still had the shift that you reported previously ( I pulled this morning so after the commit that fixed it). I discussed then with Chris who told me to use the list of files which fixed it. I reproduce this error on my station (marsalis) and on my private laptop with T1 and T2

jcohenadad:
> I discussed then with Chris who told me to use the list of files which fixed it.

discussion should always be on team slack channel (not private) or github-- to prevent these kind of confusions and share knowledge within the team at large. thanks!

cfhammill:
This was my fault, I should have documented. I cleaned up the help so that `-to-fsl` indicates it takes an optional target file. I also used argparse properly instead of manually splitting an input tuple. Use is now:

`-to-fsl SRC [TGT]` instead of `-to-fsl SRC[,TGT]`

cfhammill:
Exporting some notes from my conversation with @lrouhier:

- There appear to be minor intensity differences between `sct_apply_transfo` and `applywarp`, this I consider expected due to interpolation differences
- There was some worry this differences would interfere with label resampling. I suspect that using nearest neighbour interpolation will resolve any intensity differences.

cfhammill:
Following up on https://github.com/neuropoly/spinalcordtoolbox/pull/2529#issuecomment-591073147

FSL's license permits redistribution provided it is non-commercial use and we include the license. So if we create a tarball with applywarp binaries and include the license we should be fine to host it on OSF. We'll need to acquire/build binaries for each of our test architectures.

cfhammill:
I've now tracked down applywarp binaries for bionic, xenial, trusty, and mac (conveniently all the builds for our mac platforms are the same). Next chance I get I'll package them up and put them on OSF with the FSL license.

Still waiting on confirmation that this works on more examples than just the `sct_straighten_spinalcord` example.

cfhammill:
I'm now of the opinion that we can't easily supply the applywarp binaries without carting along the C/C++ shared objects it depends on for at least 3 platforms. @jcohenadad do you still think it's worth integrating tests into `unit_testing` and travis, or should I add it to `testing` with test skips if applywarp can't be found on your PATH?

jcohenadad:
awesome @cfhammill! if that doesn't work don't worry, we can skip those tests. SCT currently has much more important tests lacking so we need to priorities our invested time.

cfhammill:
Understood. I'll see what I can do tonight, if the automated tests don't make the cut so be it.

cfhammill:
- added a feature to generate random deformations. 
- added the applywarp bundle to OSF and integrated it in to `sct_download_data`
- run simple experiments on the random deformations and `-to-fsl` appears to be working great.

Remaining: 
- figure out a decent MI threshold for the test, will be very high
- try different orientations and affine spaces

@jcohenadad do you think I should spend a few more hours and finish this up?

scripts/sct_image.py
```diff
@@ -111,7 +114,7 @@ def get_parser():
     orientation.add_argument(
         '-setorient-data',
         help='Set orientation of the input image\'s data (does NOT modify the header, but the data). Use with care !',
-        choices='RIP LIP RSP LSP RIA LIA RSA LSA IRP ILP SRP SLP IRA ILA SRA SLA RPI LPI RAI LAI RPS LPS RAS LAS PRI PLI ARI ALI PRS PLS ARS ALS IPR SPR IAR SAR IPL SPL IAL SAL PIR PSR AIR ASR PIL PSL AIL ASL'.split(),
+        choices='RIP LIP RSP LSP RIA LIA RSA LSA IRP ILP SRP SLP IRA ILA SRA SLA RPI LPI RAI LAI RPS LPS RAS LAS PRI PLI ARI ALI PRS PLS ARS ALS IPR SPR IAR SAR IPL SPL IAL SAL PIR PSR AIR ASR PIL PSL AIL00 ASL'.split(),
```
jcohenadad:
```suggestion
        choices='RIP LIP RSP LSP RIA LIA RSA LSA IRP ILP SRP SLP IRA ILA SRA SLA RPI LPI RAI LAI RPS LPS RAS LAS PRI PLI ARI ALI PRS PLS ARS ALS IPR SPR IAR SAR IPL SPL IAL SAL PIR PSR AIR ASR PIL PSL AIL ASL'.split(),
```

spinalcordtoolbox/testing/create_test_data.py
```diff
@@ -157,3 +157,31 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
...
+    data = np.random.random((nx, ny, nz, 1, 3)) * 3
+
+    # Create nibabel object
+    nii = nib.nifti1.Nifti1Image(data.astype(dtype), affine)
+
+    img = Image(nii.get_data(), hdr=nii.header, dim=nii.header.get_data_shape())
+    img.header.set_intent("vector")
+    # Update orientation
+    img.change_orientation(orientation)
+    if debug:
```
jcohenadad: We [recently](https://github.com/neuropoly/spinalcordtoolbox/pull/2616/commits/23cffa7f124b80ca9fcfe58397341758dc22e9ba) moved image saving by the calling function. It makes more sense and avoid passing the debug argument.

jcohenadad:
thank you so much @cfhammill this is fantastic!

> @jcohenadad do you think I should spend a few more hours and finish this up?

i would say, no need for exhaustive testing with many orientations. Maybe just a few tests just to track down possible breakage in the future.

cfhammill:
I've been unable to get the tests to work on travis. There were some simple problems of missing libraries on linux that I was able to fix by fetching those from neurodebian. Now the test is failing with no informative output, just that applywarp doesn't appear to be generating a file that we can read with sct `Image`.  Debugging from here I suspect would be relatively time intensive, feedback cycles are on the order of 45 minutes per change attempt. My memories of debugging travis builds are that there is no easy way to run it locally for quick feedback, although this experience is from a few years ago, so it's possible this has changed.

My suggestion is that we separate the test part of this to a separate PR that we can leave on the back burner. We then merge this feature but indicate that it hasn't been extensively validated, so users should spot check the generated warps against `sct_apply_transfo` for their own source and targets before feeding to an FSL pipeline. What do you think @jcohenadad? 

If I were to continue trying to get this to work I would probably try:
- abandoning the applywarp binaries I bundled, instead using the neurodebian builds.
- hardcode the deb package path to work around trusty failing to read the neurodebian apt-keys
- skip the test on mac (or use my bundle)
- investigate getting a travis VM or docker image.

jcohenadad:
@cfhammill i agree with all [your suggestions](https://github.com/neuropoly/spinalcordtoolbox/pull/2529#issuecomment-605680038). Yes, please update the branch and let's merge

cfhammill:
Ok, PR is now clean and ready for review

scripts/sct_image.py
jcohenadad: @cfhammill did you see my comment?

cfhammill: ahh sorry, I thought this was in relation to part of the test code. Will fix.

scripts/sct_image.py
```diff
@@ -127,12 +130,29 @@ def get_parser():
...
+        required=False)
+    multi.add_argument(
+        '-to-fsl',
+        metavar=Metavar.file,
+        help='Transform displacement field values to absolute FSL warps. To be used with FSL\'s applywarp function with the '
+        '`--abs` flag. Input the file that will be used as the input (source) for applywarp and optionally the target '
+        '(ref). The target file is necessary for the case where the warp is in a different space than the target. For '
+        'example, the inverse warps generated by `sct_straighten_spinalcord`. This feature has not been extensively '
+        'validated so consider checking the results of `applywarp` against `sct_apply_transfo` before using in FSL '
+        'pipelines',
```
jcohenadad, line 153: It took me some time to understand what to input and where, and then I forgot to set the flag -o so my input warp was overwritten ;-). so I suggest joining an example syntax:

```suggestion
        'validated so consider checking the results of `applywarp` against `sct_apply_transfo` before using in FSL '
        'pipelines. Example syntax: "sct_image -i WARP_SRC2DEST -to-fsl IM_SRC (IM_DEST) -o WARP_FSL", '
        'followed by FSL: "applywarp -i IM_SRC -r IM_DEST -w WARP_FSL --abs -o IM_SRC2DEST" ',
```

jcohenadad:
@cfhammill I have a somewhat bad news. I've tried the following example, and although the registration with `applywarp` worked fine (in terms of alignment), part of the destination image is cropped. I believe this is caused by the intermediate straightening warping field, which is spatially limited (in the R-L direction). If my diagnosis is correct, one way to mitigate this would be to increase the dimension of the straightening warping field. 

To reproduce (with 840dcfcf98ee7a8298ff0ec32efae138c0ac1816, FSL 6.0.1, on OSX):
~~~
# Get data, various preprocessing
cd sct_example_data/t2
sct_propseg -i t2.nii.gz -c t2
sct_label_vertebrae -i t2.nii.gz -s t2_seg.nii.gz -c t2
sct_label_utils -i t2_seg_labeled.nii.gz -vert-body 3,7 -o labels_vert.nii.gz
# Register to template
sct_register_to_template -i t2.nii.gz -s t2_seg.nii.gz -l labels_vert.nii.gz 
# Get FSL-compatible warping field
sct_image -i warp_anat2template.nii.gz -to-fsl t2.nii.gz anat2template.nii.gz -o warp_anat2template-fsl.nii.gz
# apply with FSL's applywarp
applywarp -i t2.nii.gz -r anat2template.nii.gz -w warp_anat2template-fsl.nii.gz --abs -o t2-template-fsl.nii.gz
~~~

See cropped animation below: flip between t2 registered to template using (i) SCT and (ii) applywarp (the cropped is with applywarp):
anim

cfhammill:
thanks for having a look @jcohenadad. I can easily reproduce. In this case there is a trivial fix, in

```
# Get FSL-compatible warping field
sct_image -i warp_anat2template.nii.gz -to-fsl t2.nii.gz anat2template.nii.gz -o warp_anat2template-fsl.nii.gz
```

the second argument to `-to-fsl` can be dropped (because warp and tgt are in the same space)

```
# Get FSL-compatible warping field
sct_image -i warp_anat2template.nii.gz -to-fsl t2.nii.gz -o warp_anat2template-fsl.nii.gz
```

this yields the same output as `sct_apply_transfo` for me. 

So the interesting thing is what happens when both arguments are passed to `-to-fsl`. The dimensionalities of the relevant images are:

| type | file | x | y | z |
| --- | --- | --- | --- | --- |
| src | t2.nii.gz |  64 | 320 | 320 |
| original tgt (I assume) | straight_ref.nii.gz | 51 | 141 | 197 |
| final tgt | anat2template.nii.gz | 141 | 141 | 991 | 
|  warp | warp_anat2template.nii.gz | 141 | 141 | 991 |  

The code in principle should resample the warp to match the space of the target, but, the space of the warp and target are already the same. So if the problem is on our end, the relevant code:

```python
    if tgt is not None:
        hdr = disp_im.header.copy()
        shp = disp_im.data.shape
        hdr.set_data_shape(shp)
        disp_nib = Nifti1Image(disp_im.data.copy(), aff(disp_im), hdr)
        disp_resampled = resample_from_to(disp_nib, (shp, aff(tgt)))
        disp_im.data = disp_resampled.dataobj.copy()
        disp_im.header = disp_resampled.header.copy()
``` 
https://github.com/neuropoly/spinalcordtoolbox/blob/jca/fsl-warp/scripts/sct_image.py#L398 somehow isn't doing nothing as we'd expect. I can't immediately see what's wrong with it. This is the only part of the code that is different if the second argument is passed to `-to-fsl`.

Any thoughts?

cfhammill:
So the problem is coming from nibabel's `resample_from_to` which in turn comes fro `scipy.ndimage.affine_transform` combined with how ITK fills deformations outside of the image. ITK codes out of image deformations as e.g. 3.4e+38. Nibabel does spline interpolation of order 3 which blends voxels along the image boundary with the out of image values causing them to be pushed out of the image (creating the appearance of a crop). If the spline order is dropped to 0 or 1 the problem disappears.

cfhammill:
Should be fixed now (via ed050ec )

jcohenadad:
> Should be fixed now (via ed050ec )

Indeed! Good job @cfhammill 👏 
I let you update the branch and merge whenever you want

cfhammill:
Thanks @jcohenadad! Currently this branch still has the `-world2vox` flag you added, do you want to keep it? If so I think we should remove the reference to FSL in the docstring to avoid users getting confused. What do you think?

Otherwise feel free to merge when ready! (I'm not authorized to merge I don't think.)

jcohenadad:
> Currently this branch still has the -world2vox flag you added, do you want to keep it?

ah! i completely forgot about it. Well, that flag was supposed to do what you've implemented so you can remove it. Your choice of flag name is much more explicit (i.e. better).

cfhammill:
Oh ok, I kept it around just in case you thought it might be useful for something else. Removing it now.

PR #2533, opened 2019-11-21
Added an option for keypoints transformation

### LINKED ISSUE #2519: Creation of extra points when using sct_straighten_spinalcord on disc label files
### Description
When using sct_straighten_spinal cord on disc label files, the resulting image has a non-zero background as well as some points appearing around the correct label.
Screen Shot 2019-11-11 at 9 58 04 AM
### Steps to Reproduce

0. Install SCT Spinal Cord Toolbox (git-master-8e790d825d9124cc80fd819c669e904f3c1d70b8)-MAC OSX Mojave

1. run sct_straighten_spinalcord -i labels_disc_all_posterior_order_r.nii.gz -s t2_seg_manual.nii.gz

**Expected behavior:** I expected the label to be cleanly translated. 

**Actual behavior:**  Some point appears next to what seems to be the correct label. The value of the label are changed as well. 

(data is located in Duke: /projects/ivado-medical-imaging/lucas_vertebral_labeling/20191111-issue/)

lrouhier:
This PR aims to add the -label flag to the sct_apply_transfo script. This flag can be used when applying a warping field to images containing point-like labels (e.g., discs label).
It performs the following :

* Change interpolation to 'nearest neighbor'
* Dilate labels on the image 
* Apply the warping field to the dilated image
* Apply cubic_to_point function to retrieve point-like labels on the transformed image.

Fixes #2519

### DIFF
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -18,10 +18,7 @@ from __future__ import division, absolute_import
 
 import sys, io, os, time, functools
 import argparse
-import scipy.ndimage.filters as scf
-from skimage.feature import peak_local_max
-import numpy as np
-import math
+
 
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
 from spinalcordtoolbox.image import Image
@@ -140,7 +137,7 @@ class Transform:
         self.list_warpinv = list_warpinv
         self.fname_dest = fname_dest
         self.output_filename = output_filename
-        self.interp = insterp
+        self.interp = interp
         self.crop = crop
         self.verbose = verbose
         self.remove_temp_files = remove_temp_files
@@ -234,9 +231,16 @@ class Transform:
                      '-i', fname_src,
                      '-o', output,
                      '-dilate','2'])
-                fname_src=os.path.join(path_tmp, "dilated_data.nii")
-
-            sct.run(['isct_antsApplyTransforms',
+                tmp_src=os.path.join(path_tmp, "dilated_data.nii")
+                tmp_out=os.path.join(path_tmp, "dilated_data_reg.nii")
+                sct.run(['isct_antsApplyTransforms',
+                     '-d', dim,
+                     '-i', tmp_src,
+                     '-o', tmp_out,
+                     '-t'] + fname_warp_list_invert + ['-r', fname_dest] + interp,
+                    verbose=verbose, is_sct_binary=True)
+            else:
+                sct.run(['isct_antsApplyTransforms',
                      '-d', dim,
                      '-i', fname_src,
                      '-o', fname_out,
@@ -336,7 +340,7 @@ class Transform:
 
         if label==1:
             sct.run(['sct_label_utils',
-                     '-i', fname_out,
+                     '-i', tmp_out,
                      '-o', fname_out,
                      '-cubic-to-point'])
        
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -24,14 +24,7 @@ from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.cropping import ImageCropper
 
 import sct_utils as sct
-import sct_convert
 import sct_image
-
-
-class Param:
-    def __init__(self):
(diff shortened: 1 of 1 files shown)


scripts/sct_apply_transfo.py
```diff
@@ -19,10 +19,12 @@
 import sys, io, os, time, functools
 import argparse
 
+
```
jcohenadad: only single line separation
see: https://github.com/neuropoly/spinalcordtoolbox/wiki/example_script

scripts/sct_apply_transfo.py
```diff
@@ -19,10 +19,12 @@
...
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.cropping import ImageCropper
 
+
```
jcohenadad: single line

scripts/sct_apply_transfo.py
```diff
@@ -120,7 +128,7 @@ def get_parser():
 
 
 class Transform:
-    def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=[], output_filename='', verbose=0, crop=0,
+    def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=[], output_filename='', verbose=0, crop=0,label=0,
```
jcohenadad: space after comma-- PEP8

scripts/sct_apply_transfo.py
```diff
@@ -132,7 +140,8 @@ def __init__(self, input_filename, fname_dest, list_warp, list_warpinv=[], outpu
         self.verbose = verbose
         self.remove_temp_files = remove_temp_files
         self.debug = debug
-
+        self.label=label
```
jcohenadad: space between =.
PEP8

scripts/sct_apply_transfo.py
```diff
@@ -142,6 +151,9 @@ def apply(self):
         verbose = self.verbose
         remove_temp_files = self.remove_temp_files
         crop_reference = self.crop  # if = 1, put 0 everywhere around warping field, if = 2, real crop
+        label=self.label
```
jcohenadad: space. Please apply PEP8 for all your modifs (there are other formatting issues below)

scripts/sct_apply_transfo.py
```diff
@@ -207,7 +219,26 @@ def apply(self):
...
+                output=os.path.join(path_tmp, "dilated_data.nii")
+                #dilate points
+                sct.run(['sct_maths',
+                     '-i', fname_src,
+                     '-o', output,
+                     '-dilate', '2'])
+                tmp_src=os.path.join(path_tmp, "dilated_data.nii")
+                tmp_out=os.path.join(path_tmp, "dilated_data_reg.nii")
+
+                sct.run(['isct_antsApplyTransforms',
```
jcohenadad: duplicated code (see below)

jcohenadad:
@lrouhier after giving it some thoughts, how about adding a possible value `label` to the flag `-x`, instead of creating a new boolean flag `-label`? It would save some lines of code, and make it explicit that "label" is very much related to the interpolation (and cannot be used with any other interpolation method)

lrouhier:
Maybe it will be clearer! It does not change much in the code for me and i think it's easier for the user. Any thoughts on how to explain it in the help? 

for now I wrote :  `Interpolation method. the 'label' interpolation method is designed to improve results on keypoints label file (e.g., disc label files) by dilating the original label and retrieving key points after the transformation.` (line 102-103)

jcohenadad:
i would go with: 
~~~
Interpolation method. the 'label' method is to be used if you would like to apply a transformation on a file that has single-voxel labels (classical interpolation methods won't work, as resampled labels might disappear or their values be altered). The function will dilate each label, apply the transformation using nearest neighbour interpolation, and then take the center-of-mass of each "blob" and output a single voxel per blob.
~~~

jcohenadad:
Let's merge https://github.com/neuropoly/spinalcordtoolbox/pull/2522 before this one

PR #2559, opened 2019-12-19
Jca/4.2.0 bis

jcohenadad:
new release


alexfoias:
tested . the installation worked well.

Issue #2587, opened 2020-01-18
Unresponsive command in FSLeyes integration within Docker


jcohenadad:
When using FSLeyes integration via Docker (version 4.2.1), the commands don't run. The terminal output shows the following WARNING:
~~~
 WARNING       annotations.py  254: draw            - GLError(
	err = 1283,
	description = b'stack overflow',
	baseOperation = glPushMatrix,
	cArguments = ()
)
Traceback (most recent call last):
  File "/home/sct/.local/lib/python3.6/site-packages/fsleyes/gl/annotations.py", line 251, in draw
    obj.draw2D(zpos, axes)
  File "/home/sct/.local/lib/python3.6/site-packages/fsleyes/gl/annotations.py", line 838, in draw2D
    calcSize=True)
  File "/home/sct/.local/lib/python3.6/site-packages/fsleyes/gl/routines.py", line 403, in text2D
    gl.glPushMatrix()
  File "/home/sct/.local/lib/python3.6/site-packages/OpenGL/logs.py", line 67, in __call__
    return function( *args, **named )
  File "src/errorchecker.pyx", line 58, in OpenGL_accelerate.errorchecker._ErrorChecker.glCheckError
OpenGL.error.GLError: GLError(
	err = 1283,
	description = b'stack overflow',
	baseOperation = glPushMatrix,
	cArguments = ()
)
 WARNING              logs.py   70: __call__        - Failure on glPushMatrix: Traceback (most recent call last):
  File "/home/sct/.local/lib/python3.6/site-packages/OpenGL/logs.py", line 67, in __call__
    return function( *args, **named )
  File "src/errorchecker.pyx", line 58, in OpenGL_accelerate.errorchecker._ErrorChecker.glCheckError
OpenGL.error.GLError: GLError(
	err = 1283,
	description = b'stack overflow',
	baseOperation = glPushMatrix,
	cArguments = ()
)

 WARNING       annotations.py  254: draw            - GLError(
	err = 1283,
	description = b'stack overflow',
	baseOperation = glPushMatrix,
	cArguments = ()
)
Traceback (most recent call last):
  File "/home/sct/.local/lib/python3.6/site-packages/fsleyes/gl/annotations.py", line 251, in draw
    obj.draw2D(zpos, axes)
  File "/home/sct/.local/lib/python3.6/site-packages/fsleyes/gl/annotations.py", line 838, in draw2D
    calcSize=True)
[...]

PR #2581, opened 2020-01-13
Added more functions to the FSLeyes plugin

jcohenadad:
The purpose of this PR is to add more SCT functions to the FSLeyes plugin. For more info, see the original PR: https://github.com/neuropoly/spinalcordtoolbox/pull/2339

Useful doc for Python's wx module (used by FSLeyes): https://wxpython.org/presentations/OSCON2003/wxPython-OSCON2003-v5.pdf

Fixes #1914

This PR also fixes another issue related to the non-existence of PYTHONHOME variable in our Docker container. Fixes #2587

PR #2589, opened 2020-01-23
Fixed QC display syntax for Docker users

### LINKED ISSUE #2585: QC open command should be hidden if users don't have access to local browser
For users running SCT under Docker in Windows via Xming, when launching a process with the flag `-qc`, the command will output a syntax to open this html report, using the default Linux program: `xdg-open`. The problem is that, even though the program is installed, it cannot find a default browser because there is no browser installed in the SCT Docker container, yielding the following error: 

~~~
$open_cmd index.html
Couldn't find a suitable web browser!
Set the BROWSER environment variable to your desired browser.
Warning: program returned non-zero exit code #1
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: www-browser: not found
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: links2: not found
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: elinks: not found
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: links: not found
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: lynx: not found
/usr/bin/xdg-open: 851: /usr/bin/xdg-open: w3m: not found
xdg-open: no method available for opening 'index.html'
~~~

One solution would be to look interrogate the variable `BROWSER`, and if not defined, change the message for "Double click on the file: PATH/index.html"

lrouhier:
Two issues appeared regarding the use of SCT in our released docker container : 

#2585 : For docker user, they can't copy-paste the `xdg-open` command because no browser is set. If the variable `BROWSER` is not set for the Linux user, this command won't work as well. For now, we just modified the command given to the user if he uses Linux without defining `$BROWSER`. 

#2587: Fsleyes plugin is not working for docker users. After investigations, the crash is caused by the deletion  [here](https://github.com/neuropoly/spinalcordtoolbox/blob/0ab51b7d97eaecd539ac11b6fdb07a6ea6fb7548/contrib/fsl_integration/sct_plugin.py#L35) of 'PYTHONHOME' and 'PYTHONPATH' that are not defined. I added a test to avoid this situation. The plugin seems to be responsive now.  Fixed by #2581 

However, it does not solve the warning problem reported also in the issue. It might come from the installation of fsleyes in the docker container which is currently investigated (WIP)

intent to fix #2585, #2587


spinalcordtoolbox/reports/qc.py
```diff
@@ -561,7 +561,12 @@ def layout(qslice):
     try:
         from sys import platform as _platform
         if _platform == "linux" or _platform == "linux2":
-            sct.printv('xdg-open "{}/index.html"'.format(path_qc), type='info')
+            try:  
+                os.environ["BROWSER"]
+                sct.printv('xdg-open "{}/index.html"'.format(path_qc), type='info')
+            except KeyError:
```
jcohenadad: could you please comment in the code when this KeyError is supposed to happen?

jcohenadad: also, are you 100% sure that in ALL linux platforms, this BROWSER variable exists if a browser is accessible?

lrouhier: Xdg-open was asking for it, but now I am not sure that it is automatically set. 
The error is thrown if the variable is not previously declared or empty. 
I see two option : 
* there is webbrowser python package that can do the test without going through the $BROWSER variable.
* We can set a $DOCKER env variable for the docker container and use the same Key error.

jcohenadad: > We can set a $DOCKER env variable for the docker container and use the same Key error.

I like this approach.

jcohenadad: it seems a bit cleaner somehow...

jcohenadad:
@lrouhier before we merge: is the DOCKER variable now set in the new Dockerfile? if so can you please point to the modified code?

lrouhier:
> @lrouhier before we merge: is the DOCKER variable now set in the new Dockerfile? if so can you please point to the modified code?

It is on the PR https://github.com/neuropoly/sct_docker/pull/51, here is the [link](https://github.com/neuropoly/sct_docker/blob/1a2de7a8f0602c2087ba27d5bb5ad91b8150accb/sct_docker.py#L84) to the exact line.  Do you want me to add something as a comment inside the QC file ?

jcohenadad:
>  inside the QC file

which file are you talking about?

lrouhier:
Sorry I meant in the `qc.py` file. For futur reference if it is needed.

jcohenadad:
ah! yes, it would be nice. Thanks!

PR #2604, opened 2020-02-19
Fixed problem with parameters xmax, ymax and zmax

### LINKED ISSUE #2603: Parameters xmax, ymax and zmax are not effective
### Description
It seems that  there is an issue with the params **xmax**/ **ymax** / **zmax** of the cmd `sct_crop_image`. Indeed the higher bound for cropping along the axis is bigger than expected. 

For example:
 
- `sct_crop_image -i T2w.nii.gz -zmax -4 -o T2w_z_4.nii.gz`

should create an image with the same dimension as `T2w.nii.gz` minus 3 along the axis **z**.
However `T2w_z_4.nii.gz` has actually only 2 slices less than `T2w.nii.gz`.

AnBucquet:
I just removed the `+1` in that [line](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/cropping.py#L43) to get the correct dimension.

Fixes #2603

### DIFF
--- a/scripts/sct_crop_image.py
+++ b/scripts/sct_crop_image.py
@@ -109,14 +109,14 @@ def get_parser():
         '-zmin',
         type=int,
         default=0,
-        help="Lower bound for cropping along Z. Follows the same rules as xmax.",
+        help="Lower bound for cropping along Z.",
         metavar=Metavar.int,
         )
     optional.add_argument(
         '-zmax',
         type=int,
         default=-1,
-        help="Higher bound for cropping along Z. Inputting -1 will set it to the maximum dimension.",
+        help="Higher bound for cropping along Z. Follows the same rules as xmax.",
         metavar=Metavar.int,
         )
     optional.add_argument(
--- a/scripts/sct_crop_image.py
+++ b/scripts/sct_crop_image.py
@@ -87,8 +87,8 @@ def get_parser():
         '-xmax',
         type=int,
         default=-1,
-        help="Higher bound for cropping along X. Setting '-1' will crop to the maximum dimension, '-2' will crop to "
-             "the maximum dimension minus 1 slice, etc.",
+        help="Higher bound for cropping along X. Setting '-1' will crop to the maximum dimension (i.e. no change), "
+             "'-2' will crop to the maximum dimension minus 1 slice, etc.",
         metavar=Metavar.int,
         )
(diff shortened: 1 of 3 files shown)


spinalcordtoolbox/cropping.py
```diff
@@ -38,9 +38,9 @@ def _get_max_value(input, dim):
             # If empty, return dim+1 (corresponds to the maximum of the given dimension, e.g. nx)
             if input is None:
                 return dim + 1
```
jcohenadad: to be consistent, this should be dim (not dim+1), right?

spinalcordtoolbox/cropping.py
```diff
@@ -38,9 +38,9 @@ def _get_max_value(input, dim):
...
-            # If negative sign, return dim+1 if -1, dim if -2, dim-1 if -3, etc.
+            # If negative sign, return dim if -1, dim-1 if -2, dim-2 if -3, etc.
```
jcohenadad: To clarify (the comment is in contradiction with the code), i would write:
~~~
# If input is "-1", return maximum dimension (i.e. no change). If input is "-2", returns maximum dimension minus one, etc.
~~~
~~~

spinalcordtoolbox/cropping.py
AnBucquet: Yes, you're right! Thx

AnBucquet:
It should be good now. Thx for the review! Let me know if you find anything else.

AnBucquet:
Need to change `test_sct_crop_image.py` as well.

scripts/sct_crop_image.py
```diff
@@ -87,8 +87,8 @@ def get_parser():
         '-xmax',
         type=int,
         default=-1,
-        help="Higher bound for cropping along X. Setting '-1' will crop to the maximum dimension, '-2' will crop to "
-             "the maximum dimension minus 1 slice, etc.",
+        help="Higher bound for cropping along X. Setting '-1' will crop to the maximum dimension (i.e. no change), "
+             "'-2' will crop to the maximum dimension minus 1 slice, etc.",
```
jcohenadad: zmin should be: "Lower bound for cropping along Z.",
zmax should be: "Higher bound for cropping along Z. Follows the same rules as xmax."

PR #2616, opened 2020-02-28
Enable 2D kernel for morpho math operations and various improvements

### LINKED ISSUE #2611: Check data type before running morphological operation
To output a clearer message to the user. 

See: http://forum.spinalcordmri.org/t/problem-with-latest-version-installation/311/25?u=jcohenadad

### LINKED ISSUE #2615: Enable 2D kernel for morpho math operations
Things to check/fix:
- redefinition of selem variable in erosion/dilation kernel.
- use the iteration feature from scikit-image to get multi-step erosion (make it a feature)
- start porting some of the functions of sct_maths into a dedicated python module, which we can call `math`

Needed for two projects:
- https://github.com/neuropoly/spinalcordtoolbox/issues/2615
- https://github.com/sct-pipeline/cord-erosion/issues/1

jcohenadad:
This PR brings some improvements to the dilate/erode methods in `sct_maths`:
- ported some of the functions of sct_maths into a new dedicated python module called `math`
- enabled 2D kernel for morpho math operations. Fixes #2615 
- redefinition of selem variable in erosion/dilation kernel.
- added unit tests for morpho math operations
- Updated usage to mention grayscale input. Fixes #2611

### DIFF
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -352,12 +352,10 @@ def main(args=None):
         data_out = smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = sctmath.morphomath(data, filter='dilation', size=arguments.dilate, shape=arguments.shape,
-                                      dim=arguments.dim)
+        data_out = sctmath.dilate(data, size=arguments.dilate, shape=arguments.shape, dim=arguments.dim)
 
     elif arguments.erode is not None:
-        data_out = sctmath.morphomath(data, filter='erosion', size=arguments.erode, shape=arguments.shape,
-                                      dim=arguments.dim)
+        data_out = sctmath.erode(data, size=arguments.erode, shape=arguments.shape, dim=arguments.dim)
 
     elif arguments.denoise is not None:
         # parse denoising arguments
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -19,6 +19,7 @@ import argparse
 
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
+import spinalcordtoolbox.math as sctmath
 
 from sct_utils import printv, extract_fname
 import sct_utils as sct
@@ -336,10 +337,10 @@ def main(args=None):
         data_out = smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = dilate(data, convert_list_str(arguments.dilate, "int"))
+        data_out = sctmath.dilate(data, convert_list_str(arguments.dilate, "int"))
 
     elif arguments.erode is not None:
-        data_out = erode(data, convert_list_str(arguments.erode, "int"))
+        data_out = sctmath.erode(data, convert_list_str(arguments.erode, "int"))
 
     elif arguments.denoise is not None:
         # parse denoising arguments
@@ -469,40 +470,6 @@ def perc(data, perc_value):
 
 def binarise(data, bin_thr=0):
     return data > bin_thr
-
-
-def dilate(data, radius):
-    """
-    Dilate data using ball structuring element
-    :param data: 2d or 3d array
-    :param radius: radius of structuring element OR comma-separated int.
-    :return: data dilated
-    """
-    from skimage.morphology import dilation, ball
-    if len(radius) == 1:
-        # define structured element as a ball
-        selem = ball(radius[0])
-    else:
-        # define structured element as a box with input dimensions
-        selem = np.ones((radius[0], radius[1], radius[2]), dtype=np.dtype)
-    return dilation(data, selem=selem, out=None)
-
-
-def erode(data, radius):
-    """
-    Erode data using ball structuring element
-    :param data: 2d or 3d array
-    :param radius: radius of structuring element
-    :return: data eroded
-    """
-    from skimage.morphology import erosion, ball
-    if len(radius) == 1:
-        # define structured element as a ball
-        selem = ball(radius[0])
-    else:
-        # define structured element as a box with input dimensions
-        selem = np.ones((radius[0], radius[1], radius[2]), dtype=np.dtype)
-    return erosion(data, selem=selem, out=None)
 
 
 def get_data(list_fname):
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -140,18 +140,33 @@ def get_parser():
     mathematical = parser.add_argument_group("MATHEMATICAL MORPHOLOGY")
     mathematical.add_argument(
         '-dilate',
-        metavar='',
-        help='Dilate binary image. If only one input is given, structured element is a ball with input radius (in '
-             'voxel). If comma-separated inputs are given (Example: 2,4,5 ), structured element is a box with input '
-             'dimensions.',
+        type=int,
+        metavar=Metavar.int,
+        help="Dilate binary image with specified size. If shape={'square', 'cube'}: size corresponds to the length of "
+             "an edge (size=1 has no effect). If shape={'disk', 'ball'}: size corresponds to the radius, not including "
+             "the center element (size=0 has no effect).",
         required=False)
     mathematical.add_argument(
         '-erode',
-        metavar='',
-        help='Erode binary image. If only one input is given, structured element is a ball with input radius (in '
-             'voxel). If comma-separated inputs are given (Example: 2,4,5), structured element is a box with input '
-             'dimensions.',
+        type=int,
+        metavar=Metavar.int,
+        help="Erode binary image with specified size. If shape={'square', 'cube'}: size corresponds to the length of "
+             "an edge (size=1 has no effect). If shape={'disk', 'ball'}: size corresponds to the radius, not including "
+             "the center element (size=0 has no effect).",
         required=False)
+    mathematical.add_argument(
+        '-shape',
+        help="Shape of the structuring element for the mathematical morphology operation.",
+        required=False,
+        choices=('square', 'cube', 'disk', 'ball'))
+    mathematical.add_argument(
+        '-dim',
+        type=int,
+        metavar=Metavar.int,
+        help="Dimension of the array which 2D structural element will be orthogonal to. For example, if you wish to "
+             "apply a 2D disk kernel in the X-Y plane, leaving Z unaffected, parameters will be: shape=disk, dim=2.",
+        required=False,
+        choices=(0, 1, 2))
 
     filtering = parser.add_argument_group("FILTERING METHODS")
     filtering.add_argument(
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -356,7 +356,7 @@ def main(args=None):
                                       dim=arguments.dim)
 
     elif arguments.erode is not None:
-        data_out = sctmath.morphomath(data, filter='erosion', size=arguments.dilate, shape=arguments.shape,
+        data_out = sctmath.morphomath(data, filter='erosion', size=arguments.erode, shape=arguments.shape,
                                       dim=arguments.dim)
 
     elif arguments.denoise is not None:
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -17,12 +17,12 @@ import sys
 import numpy as np
 import argparse
 
+import spinalcordtoolbox as sct
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
-import spinalcordtoolbox.math as sctmath
+import spinalcordtoolbox.math
 
-from sct_utils import printv, extract_fname
-import sct_utils as sct
+from sct_utils import printv, extract_fname, display_viewer_syntax, init_sct
 
 
 ALMOST_ZERO = 0.000000001
@@ -251,7 +251,7 @@ def main(args=None):
     fname_in = arguments.i
     fname_out = arguments.o
     verbose = arguments.v
-    sct.init_sct(log_level=verbose, update=True)  # Update log level
+    init_sct(log_level=verbose, update=True)  # Update log level
     if '-type' in arguments:
         output_type = arguments.type
     else:
@@ -352,10 +352,10 @@ def main(args=None):
         data_out = smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = sctmath.dilate(data, size=arguments.dilate, shape=arguments.shape, dim=arguments.dim)
+        data_out = sct.math.dilate(data, size=arguments.dilate, shape=arguments.shape, dim=arguments.dim)
 
     elif arguments.erode is not None:
-        data_out = sctmath.erode(data, size=arguments.erode, shape=arguments.shape, dim=arguments.dim)
+        data_out = sct.math.erode(data, size=arguments.erode, shape=arguments.shape, dim=arguments.dim)
 
     elif arguments.denoise is not None:
         # parse denoising arguments
@@ -428,7 +428,7 @@ def main(args=None):
 
     # display message
     if data_out is not None:
-        sct.display_viewer_syntax([fname_out], verbose=verbose)
+        display_viewer_syntax([fname_out], verbose=verbose)
     else:
         printv('\nDone! File created: ' + fname_out, verbose, 'info')
 
@@ -496,7 +496,7 @@ def get_data(list_fname):
     try:
         nii = [Image(f_in) for f_in in list_fname]
     except Exception as e:
-        sct.printv(str(e), 1, 'error')  # file does not exist, exit program
+        printv(str(e), 1, 'error')  # file does not exist, exit program
     data0 = nii[0].data
     data = nii[0].data
     # check that every images have same shape
@@ -756,5 +756,5 @@ def correlation(x, y, type='pearson'):
 
 
 if __name__ == "__main__":
-    sct.init_sct()
+    init_sct()
     main()
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -337,10 +337,12 @@ def main(args=None):
         data_out = smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = sctmath.dilate(data, convert_list_str(arguments.dilate, "int"))
+        data_out = sctmath.morphomath(data, filter='dilation', size=arguments.dilate, shape=arguments.shape,
+                                      dim=arguments.dim)
 
     elif arguments.erode is not None:
-        data_out = sctmath.erode(data, convert_list_str(arguments.erode, "int"))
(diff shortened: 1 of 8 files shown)


spinalcordtoolbox/math.py
```diff
@@ -13,10 +13,11 @@
 logger = logging.getLogger(__name__)
 
 
-def dilate(data, size, shape, dim=None):
+def morphomath(data, filter, size, shape, dim=None):
```
kousu: I think this was better as two separate functions for two separate filters. Now the operation is implied by a string which makes it 2nd-class code.

The common parts can still be factored to share code (maybe `selem = _morpomath(...)`) while retaining a straightforward API.

spinalcordtoolbox/math.py
```diff
@@ -7,26 +7,31 @@
...
-        selem = ball(radius[0])
-    else:
-        # define structured element as a box with input dimensions
-        selem = np.ones((radius[0], radius[1], radius[2]), dtype=np.dtype)
+    # TODO: make a build_selem(radius, shape) function called here
+    # TODO: enable custom selem
+    shape = 'ball'
+    # Create structuring element of desired shape and radius
+    # Note: the trick here is to use the variable shape as the skimage.morphology function itself
+    selem = globals()[shape](radius)
```
kousu: This is clever but it would be safer to just do

```suggestion
    selem = {'square': square, 'cube': cube, 'disk': disk, 'ball': ball}[shape](radius)
```

You could retain some of the cleverness by

```suggestion
    selem = {k: globals()[k] for k in ['square', 'cube', 'disk', 'ball']}[shape](radius)
```

but I am wary of using `globals()` because anything could end up in there and this might allow someone to do some kind of injection attack on this code perhaps a la [prototype pollution](https://snyk.io/vuln/npm:lodash:20180130). But maybe that's an irrelevant concern for scientific computing.

Also, I would split the call from the lookup to make it more readable (I repurpose the variable here but you equally well make up a second name instead)

```suggestion
    shape = {'square': square, 'cube': cube, 'disk': disk, 'ball': ball}[shape]
    selem = shape(radius)
```

spinalcordtoolbox/math.py
```diff
@@ -42,23 +43,5 @@ def dilate(data, size, shape, dim=None):
...
-    if len(radius) == 1:
-        # define structured element as a ball
-        selem = ball(radius[0])
-    else:
-        # define structured element as a box with input dimensions
-        selem = np.ones((radius[0], radius[1], radius[2]), dtype=np.dtype)
-    return erosion(data, selem=selem, out=None)
\ No newline at end of file
+    # Applies the specified filter by running skimage.morphology command
+    return globals()[filter](data, selem=selem, out=None)
```
kousu: See my other comment about `globals()`.

unit_testing/create_test_data.py
```diff
@@ -16,7 +16,7 @@
 
 def dummy_blob(size_arr=(9, 9, 9), pixdim=(1, 1, 1), coordvox=None, debug=False):
     # nx, ny, nz = size_arr
-    data = np.zeros(size_arr)
+    data = np.zeros(size_arr, dtype=np.uint8)
```
kousu: This might hide bugs. Is real data necessarily uint8? I would make this a parameter.

spinalcordtoolbox/math.py
```diff
@@ -29,6 +29,19 @@ def dilate(data, size, shape, dim=None):
     # Create structuring element of desired shape and radius
     # Note: the trick here is to use the variable shape as the skimage.morphology function itself
     selem = globals()[shape](size)
+    # If 2d kernel, replicate it along the specified dimension
+    if len(selem.shape) == 2:
+        selem3d = np.zeros([selem.shape[0]]*3)  # Note: selem.shape[0] and selem.shape[1] are supposed to be the same
```
kousu: This comment could be an `assert` or a `ValueError`.

spinalcordtoolbox/math.py
```diff
@@ -29,6 +29,19 @@ def dilate(data, size, shape, dim=None):
...
+        imid = np.floor(selem.shape[0] / 2).astype(int)
+        if dim == 0:
+            selem3d[:, imid, imid] = selem
+        elif dim == 1:
+            selem3d[imid, :, imid] = selem
+        elif dim == 2:
+            selem3d[:, :, imid] = selem
+        else:
+            raise ValueError("dim can only take values: {0, 1, 2}")
```
kousu: I love an exhaustive switch-case!

unit_testing/test_math.py
```diff
@@ -24,11 +25,29 @@
...
-    a=1
+    # cube (only asserting along one dimension for convenience)
+    data_dil = math.dilate(im.data, size=1, shape='cube')
+    assert np.array_equal(data_dil[4, 2:7, 4], np.array([0, 0, 1, 0, 0]))
+    data_dil = math.dilate(im.data, size=2, shape='cube')
+    assert np.array_equal(data_dil[4, 2:7, 4], np.array([0, 1, 1, 0, 0]))
+    data_dil = math.dilate(im.data, size=3, shape='cube')
+    assert np.array_equal(data_dil[4, 2:7, 4], np.array([0, 1, 1, 1, 0]))
+
+    # cube (only asserting along one dimension for convenience)
```
kousu: Testing matrix math reminds me a lot of [testing HTTP results](https://vcrpy.readthedocs.io/en/latest/) or [web UIs](https://jestjs.io/docs/en/snapshot-testing) -- lots of detail that you don't want to look over but need to know is reliable. It would be helpful if scientific python had snapshot testing.

In the meantime, I think this is fine.

unit_testing/test_math.py
```diff
@@ -0,0 +1,137 @@
...
+    {'input': -math.pi / 8, 'expected': 22.5},
+    {'input': 3 * math.pi / 8, 'expected': 67.5},
+    {'input': -3 * math.pi / 8, 'expected': 67.5},
+    ]
+
+
+# noinspection 801,PyShadowingNames
+@pytest.mark.parametrize('test_orient', dict_test_orientation)
+def test_fix_orientation(test_orient):
+    assert process_seg.fix_orientation(test_orient['input']) == pytest.approx(test_orient['expected'], rel=0.0001)
```
kousu: It confuses me that this function takes in rads and outputs degrees. I would have made it take in rads, output rads, and called it "normalize_orientation".

unit_testing/test_math.py
```diff
@@ -0,0 +1,56 @@
...
+# pytest unit tests for spinalcordtoolbox.math
+
+
+from __future__ import absolute_import
+import sys
+import os
+import pytest
+import math
+import numpy as np
+from skimage.morphology import erosion, dilation, disk, ball, square, cube
```
kousu: > +./unit_testing/test_math.py:12: 'skimage.morphology.erosion' imported but unused
> +./unit_testing/test_math.py:12: 'skimage.morphology.dilation' imported but unused
> +./unit_testing/test_math.py:12: 'skimage.morphology.disk' imported but unused
> +./unit_testing/test_math.py:12: 'skimage.morphology.ball' imported but unused
> +./unit_testing/test_math.py:12: 'skimage.morphology.square' imported but unused
> +./unit_testing/test_math.py:12: 'skimage.morphology.cube' imported but unused

unit_testing/test_math.py
```diff
@@ -0,0 +1,56 @@
...
+
+from spinalcordtoolbox.utils import __sct_dir__
+sys.path.append(os.path.join(__sct_dir__, 'scripts'))
+from spinalcordtoolbox import math
```
kousu: > +./unit_testing/test_math.py:16: redefinition of unused 'math' from line 10

spinalcordtoolbox/math.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+# Functions that perform mathematical operations on an image.
+
+from __future__ import absolute_import
+
+import logging
+import numpy as np
+
+from skimage.morphology import erosion, dilation, disk, ball, square, cube
```
kousu: pyflakes doesn't understand the clever `globals()` trick:

> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.erosion' imported but unused
> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.dilation' imported but unused
> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.disk' imported but unused
> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.ball' imported but unused
> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.square' imported but unused
> +./spinalcordtoolbox/math.py:10: 'skimage.morphology.cube' imported but unused

unit_testing/test_math.py
```diff
@@ -0,0 +1,137 @@
...
+sys.path.append(os.path.join(__sct_dir__, 'scripts'))
+from spinalcordtoolbox import process_seg
+from spinalcordtoolbox.centerline.core import ParamCenterline
+
+from create_test_data import dummy_segmentation
+
+
+# Define global variables
+VERBOSE = 0  # set to 2 to save files
+DEBUG = False  # Set to True to save images
```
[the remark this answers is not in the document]
jcohenadad: smart 👍

unit_testing/test_math.py
```diff
@@ -0,0 +1,137 @@
...
+
+from __future__ import absolute_import
+import sys
+import os
+import pytest
+import math
+import numpy as np
+
+from spinalcordtoolbox.utils import __sct_dir__
+sys.path.append(os.path.join(__sct_dir__, 'scripts'))
```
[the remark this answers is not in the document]
jcohenadad: totally agree with you-- i've been meaning to do that but of course never had the time. happy for you to take the lead!

unit_testing/create_test_data.py
```diff
@@ -14,6 +14,24 @@
...
+        coordvox = tuple([round(i / 2) for i in size_arr])
+    # Add non-null voxel at specified coordinates
+    data[coordvox] = 1
+    # Create image with default orientation LPI
+    affine = np.eye(4)
+    affine[0:3, 0:3] = affine[0:3, 0:3] * pixdim
+    nii = nib.nifti1.Nifti1Image(data, affine)
+    img = Image(data, hdr=nii.header, dim=nii.header.get_data_shape())
+    if debug:
+        img.save('tmp_dummy_im_'+datetime.now().strftime("%Y%m%d%H%M%S%f")+'.nii.gz')
```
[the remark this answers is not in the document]
jcohenadad: agreed-- would you suggest to save the file within `test_math.py` then?

spinalcordtoolbox/math.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+# Functions that perform mathematical operations on an image.
+
+from __future__ import absolute_import
```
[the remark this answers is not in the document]
jcohenadad: fixed in 1fbcea6e8ebb46e1bf963b00c56dce774b9e6caa

unit_testing/test_math.py
jcohenadad: fixed in 5a8e88d29a694198f4e46b7279ef1d625e9bb50f

jcohenadad: oops scrap that-- good one c69ec1e14f6b88c487823b9ff2f28c8e10dec776

spinalcordtoolbox/math.py
jcohenadad: yes i hesitated between the current option and your suggestion, but your suggestion makes more sense-- will do! thanks

unit_testing/create_test_data.py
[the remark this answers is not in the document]
jcohenadad: fixed in 63993177290f1f345798b22c9bd24bd5800dfa94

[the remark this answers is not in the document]
kousu: Yes, that's exactly what I had in mind.

spinalcordtoolbox/math.py
jcohenadad: done in 2886a99ef9a3acbe82d963ad1a3113449b7c60bf

spinalcordtoolbox/math.py
jcohenadad: now using a less elegant (but clearer) switch-case

spinalcordtoolbox/math.py
jcohenadad: if false, an `assert` will raise an `AssertionError`, but in this case, `selem.shape` could be 2 or 3 (both are fine). So, does assert make sense here? (i.e. 2 or 3 are not "wrong" per se)

unit_testing/test_math.py
jcohenadad: i love that idea. especially because right now, my test is not exhaustive (i.e. i am only testing a small portion of the array, to avoid dumping the entire array in the code). if you have suggestions in that direction that would be awesome

unit_testing/test_math.py
jcohenadad: oops! big mistake on my part. we now have python's `math` and `spinalcordtoolbox.math` packages which are both different. i changed my import for:
```python
import spinalcordtoolbox.math as sctmath
```
here: 2886a99ef9a3acbe82d963ad1a3113449b7c60bf

is that good practice?

spinalcordtoolbox/math.py
jcohenadad: the switch-case now fixes it

spinalcordtoolbox/math.py
kousu: You could write

```
if not (len(selem.shape) in [2,3] and selem.shape[0] == selem.shape[1]):
    raise ValueError("Invalid shape") 
```

unit_testing/test_math.py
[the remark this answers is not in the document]
kousu: I'll start taking notes as I do more code reviews.

unit_testing/test_math.py
kousu: I think it doesn't matter so long as the naming is consistent across modules. This is one of the places Python's flexibility bites back. I saw elsewhere you used

```
import spinalcoretoolbox as sct

...
sct.math
```

I think that's the best option on the table.

spinalcordtoolbox/math.py
kousu: And I've come around to the idea that ValueErrors are better than asserts because the latter can be turned off.

scripts/sct_maths.py
```diff
@@ -336,10 +352,12 @@ def main(args=None):
         data_out = smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = dilate(data, convert_list_str(arguments.dilate, "int"))
+        data_out = sctmath.morphomath(data, filter='dilation', size=arguments.dilate, shape=arguments.shape,
```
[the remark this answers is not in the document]
jcohenadad: i tried this way of importing sct to prevent confusion with python's math package, but ran into this issue:
```python
sct.math
  File "/Applications/PyCharm CE.app/Contents/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
AttributeError: module 'spinalcordtoolbox' has no attribute 'math'
```
any idea how we could do it?

jcohenadad:
@kousu while doing the refactoring i noticed a limitation. `dilate()` is currently accepting a numpy array. It implies that when called by other functions, we need to do something like this:
```python
im_in = Image(filein)
im_out = im_in.copy()
im_out.data = dilate(im.data)
im_out.save(fileout)
```
by having `dilate()` accept images instead, we could instead do something like this:
```python
im_out = dilate(im_in).save(fileout)
```
 So, I see two options:
- have `dilate()` accept Image objects instead of numpy arrays? 
- have `dilate()` accept both Image objects and numpy arrays, and we deal with the input type inside the function (with appropriate object to return)

scripts/sct_maths.py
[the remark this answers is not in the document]
kousu: Oops. You need to do:

```
import spinalcordtoolbox.math

sct.math
```

Python submodules don't get loaded until they're imported explicitly.

The usual way to handle this is to pre-import all the important modules in `__init__.py`; so put into `spinalcordtoolbox/__init__.py`:

```
import spinalcordtoolbox.math
```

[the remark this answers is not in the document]
jcohenadad: ok fixed here d0c942320a2ce786c015ad2746689f95269cab90

unit_testing/create_test_data.py
jcohenadad: fixed 0fa3111388a23fee28ac82fc55e223cd105dc17e

spinalcordtoolbox/math.py
```diff
@@ -0,0 +1,92 @@
...
+    :param dim: {0, 1, 2}: Dimension of the array which 2D structural element will be orthogonal to. For example, if
+    you wish to apply a 2D disk kernel in the X-Y plane, leaving Z unaffected, parameters will be: shape=disk, dim=2.
+    :return: numpy array: data dilated
+    """
+    if isinstance(data, Image):
+        im_out = data.copy()
+        im_out.data = dilation(data.data, selem=_get_selem(shape, size, dim), out=None)
+        return im_out
+    else:
+        return dilation(data, selem=_get_selem(shape, size, dim), out=None)
```
kousu: You can make this more cleanly polymorphic with recursion:

```suggestion
    if isinstance(data, Image):
        im_out = data.copy()
        im_out.data = dilate(data.data, size, shape, dim)
        return im_out
    else:
        return dilation(data, selem=_get_selem(shape, size, dim), out=None)
```

spinalcordtoolbox/math.py
```diff
@@ -0,0 +1,92 @@
...
+    :param dim: {0, 1, 2}: Dimension of the array which 2D structural element will be orthogonal to. For example, if
+    you wish to apply a 2D disk kernel in the X-Y plane, leaving Z unaffected, parameters will be: shape=disk, dim=2.
+    :return: numpy array: data dilated
+    """
+    if isinstance(data, Image):
+        im_out = data.copy()
+        im_out.data = erosion(data.data, selem=_get_selem(shape, size, dim), out=None)
+        return im_out
+    else:
+        return erosion(data, selem=_get_selem(shape, size, dim), out=None)
```
kousu:
```suggestion
    if isinstance(data, Image):
        im_out = data.copy()
        im_out.data = erode(data.data, size, shape, dim)
        return im_out
    else:
        return erosion(data, selem=_get_selem(shape, size, dim), out=None)
```

spinalcordtoolbox/math.py
kousu: This is somewhere where Julia/Java-style overloading would be really handy.

Issue #2330, opened 2019-06-29
Crash upon folder creation if already exists


jcohenadad:
During generation of QC report, a folder under _asset/img is created. If it already exists, the program crashes. This happens when processing several subjects in parallel (e.g. with GNU parallel), and some permission issues are encountered upon folder creation: I suspect two jobs are trying to create the folder at the same time.

<details><summary>Example of crash</summary>

~~~
Traceback (most recent call last):
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 185, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_deepseg_sc.py", line 180, in main
    dataset=qc_dataset, subject=qc_subject, process='sct_deepseg_sc')
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 593, in generate_qc
    stretch_contrast_method='equalized',
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 491, in add_entry
    layout(qcslice)
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 281, in wrapped_f
    self.qc_report.update_description_file(img.shape)
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 432, in update_description_file
    self._update_html_assets(get_json_data_from_path(path_json))
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 448, in _update_html_assets
    os.makedirs(dest_full_path)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/Users/julien/data/spineGeneric_sub/qc/_assets/imgs'
~~~

</details>

A similar (although different) issue was encountered in the past, when trying to write into the same json file, see https://github.com/neuropoly/spinalcordtoolbox/issues/2248

jcohenadad:
Cannot reproduce

jcohenadad:
can (and did) reproduce, on SCT course London 2020, for multi-subject processing:
~~~
*** Generate Quality Control (QC) html report ***
Resample images to 0.6x0.6 mm
QcImage: layout with Axial slice
Compute center of mass at each slice

[...]

PR #2628, opened 2020-03-07
Test for parallel qc crash

cfhammill:
Presently unable to reproduce the failure in #2330 using multiprocess but I figured it would be useful to get some extra eyes on this.

Using `multiprocess` with 6 processes generating 300 reports fails to trip this bug. 

@jcohenadad any thoughts on how to increase our chances of encountering the bug? It is possible I'm generating the reports wrong, or perhaps two different reporters need to be used.


jcohenadad:
@cfhammill I am able to get it crashed using your test, see [here](https://gist.github.com/jcohenadad/d4d0eda3afa64d25269588720bc50206). Note that it does not fail all the time (see below). I am running 64f99eaa8ec7c3e1b223cc84bf1f9410e1463462.

Further info:
- Regardless the number of iterations and jobs, the test always crashes. E.g., I used that config and the test crashed:
```python
    with multiprocessing.Pool(2) as p:
        p.map(gen_qc, range(5))
```
However, when running the test another time, it always passes. It is only if I manually remove the `/tmp/qc` folder that the test crashes again.

I am running on an OSX Mojave. Are you an Linux?  Maybe it is an Apple file system specific issue?

cfhammill:
@jcohenadad yes I was running on ubuntu 17.10, so maybe it is file system specific. I started at two cores parallelism and 100 trials and worked my way up, periodically clearing `/tmp/qc`. Also passes on my 18.04 running under windows subsystem linux.

Looking at the travis results, there's a failure on Ubuntu 14.04 that seems independent of the problem (I think it's mad that I pool mapped a function with no return value), and failing on mac high sierra (although passing on mojave). Seems suggestive that this is a mac specific problem.

I don't have a mac to test on unfortunately, but I can implement the fix suggested in #2330. We can see if the problem resolves on travis.

jcohenadad:
@cfhammill [works on 18.04](https://travis-ci.org/neuropoly/spinalcordtoolbox/jobs/659871092?utm_medium=notification&utm_source=github_status), no?

if it is a stupid Mojave-specific problem, then maybe we could try to use another folder creation function, or play with the "ignore error" flags when creating folders...

cfhammill:
It fails on 18.04 in the next push: https://travis-ci.org/neuropoly/spinalcordtoolbox/jobs/659871081?utm_medium=notification&utm_source=github_status

I don't know if I'm satisfied this test triggers the bug reliably enough, @jcohenadad what do you think? Should I invest more time making the test robust or just go ahead and try the fix?

jcohenadad:
hum... i would say, try the fix. Even if the test fails with 20% probability, we do so many pushes that i'm sure within days it will fail again (if the fix does not work).

PR #2635, opened 2020-03-15
GNU parallel flag -p added. 

### LINKED ISSUE #2610: sct_run_batch: add flag to decide to use GNU parallel (or not)

PaulBautin:
DONE: 
- flag to decide to use GNU parallel (or not) in sct_run_batch. The flag does not have to be used for the moment.

TODO:
- I will better document the use of the flag for the user if this is the fix to be made

Fixes #2610

### DIFF
--- a/shell/sct_run_batch.sh
+++ b/shell/sct_run_batch.sh
@@ -90,6 +90,7 @@ if [ "$#" -ne 2 -a "$#" -ne 3 ]; then
     usage
  fi
 
+# getopts processes flag input parmaeters
 par='false'
 while getopts "ph" opt "$3"; do
   case $opt in
@@ -134,7 +135,7 @@ else
 fi
 
 
-# Run processing with or without "GNU parallel", depending if it is installed or not
+# Run processing with or without "GNU parallel", depending if it is flagged and installed or not
 if [[ $par == 'true' ]]; then
   for path_subject in ${list_path_subject[@]}; do
     subject=`basename $path_subject`
--- a/shell/sct_run_batch.sh
+++ b/shell/sct_run_batch.sh
@@ -55,8 +55,13 @@ command_open() {
 
 # Usage of this script
 usage() {
+  echo -e "This function runs a batch script across multiple subjects that are
+  present in a source folder (as defined in the file parameters.sh). If
+  GNU parallel is installed, it will used by distributing subjects across the
+  available CPU cores. For more information about the file parameters.sh,
+  see the example file: $SCT_DIR/shell/parameters_example.sh."
   echo -e "Correct usage ./sct_run_batch.sh <parameters.sh> <process_data.sh>"
-  echo -e "-s  processing subjects sequentially"
+  echo -e "-s  processing subjects sequentially, without parallel processing."
   echo -e "-h  Help"
   exit 99
 }
@@ -73,10 +78,10 @@ if [ "$#" -ne 2 -a "$#" -ne 3 ]; then
  fi
 
 # getopts processes flag input parmaeters
-par=true
-while getopts "psh" opt "$3"; do
+use_parallel=true
+while getopts "sh" opt "$3"; do
   case $opt in
-    s)  par=false;;
+    s)  use_parallel=false;;
     h)  echo "help instruction:"; usage;;
     *)  echo "incorrect flag entered:"; usage;;
   esac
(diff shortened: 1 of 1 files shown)


shell/sct_run_batch.sh
```diff
@@ -58,7 +58,7 @@ command_open() {
 # =============================================================================
 
 # Check number of input params
-if [ "$#" -ne 2 ]; then
+if [ "$#" -ne 2 -a "$#" -ne 3 ]; then
```
jcohenadad: as we are adding more parameters, including optional ones, i think it would make sense to move to `getopts`. @PaulBautin would you be OK refactoring using this feature? you can get some inspiration from the shell script `install_sct`. Thanks a lot for your contribution!

PaulBautin: I've added getopts.

PaulBautin:
Using GNU parallel by default if installed + flag -s if user doesn't want to use GNU parallel

shell/sct_run_batch.sh
```diff
@@ -91,11 +92,12 @@ if [ "$#" -ne 2 -a "$#" -ne 3 ]; then
  fi
 
 # getopts processes flag input parmaeters
-par='false'
-while getopts "ph" opt "$3"; do
+par='true'
```
jcohenadad: in shell language, Bool variables exist so i recommend you use it instead of using string type.

shell/sct_run_batch.sh
```diff
@@ -52,27 +52,69 @@ command_open() {
...
+        esac
+    done
+  fi
+}
+
+# Usage of this script
+usage() {
+  echo -e "Correct usage ./sct_run_batch.sh <parameters.sh> <process_data.sh> [-p]"
+  echo -e "-p  Uses GNU-parallel if installed"
+  echo -e "-s  processing subjects sequentially"
```
jcohenadad: it's getting confusing: why two flags p and s? if parallel is used, it necessarily means that subjects are **not** processed sequentially. Why not simply keeping one flag: `-p` which can take values 0 or 1 (1: default)?

jcohenadad: alternatively, you can get rid of -p and only keep -s

PaulBautin:
Thanks for the fast feedback, I've modified for simplicity:
-flag -p removed
-function check_GNU_parallel removed
-use boolean variable

shell/sct_run_batch.sh
```diff
@@ -94,16 +116,15 @@ else
   list_path_subject=( "${ONLY_PROCESS_THESE_SUBJECTS[@]/#/${PATH_DATA}/}" )
 fi
 
-# Run processing with or without "GNU parallel", depending if it is installed or not
-if [ -x "$(command -v parallel)" ]; then
-  echo 'GNU parallel is installed! Processing subjects in parallel using multiple cores.' >&2
+
+# Run processing with or without "GNU parallel", depending if it is flagged and installed or not
```
jcohenadad: i would do the logic differently:
1. test if -s is specified: `echo Processing subjects sequentially.`
2. if not specified, check if parallel is installed
3. if yes: `echo GNU parallel is installed! Processing subjects in parallel using multiple cores.`
4. if no: `echo GNU parallel is not installed! Processing subjects sequentially.`

shell/sct_run_batch.sh
```diff
@@ -53,26 +53,48 @@ command_open() {
...
-    exit 1
-fi
+# check number of input parameters
+if [ "$#" -ne 2 -a "$#" -ne 3 ]; then
+    echo "Wrong number of input parameters."
+    usage
+ fi
+
+# getopts processes flag input parmaeters
+par=true
```
jcohenadad: i would use a more explicit variable name, e.g. `use_parallel`

shell/sct_run_batch.sh
```diff
@@ -53,26 +53,48 @@ command_open() {
   fi
 }
 
+# Usage of this script
+usage() {
+  echo -e "Correct usage ./sct_run_batch.sh <parameters.sh> <process_data.sh>"
+  echo -e "-s  processing subjects sequentially"
```
jcohenadad: i would make the usage more explicit:
~~~
  echo -e "-s  Processing subjects sequentially, without parallel processing."
~~~
Also, I would ass a description of the function. Something like:
~~~
This function runs a batch script across multiple subjects that are present in a source folder (as defined in the file parameters.sh). If GNU parallel is installed, it will use it by distributing subjects across the available CPU cores. For more information about the file parameters.sh, see the example file: $SCT_DIR/shell/parameters_example.sh.
~~~

PaulBautin:
I've added your modifications. Unfortunately I do not fully understand your comment on using a different logic. It seems to me that the logic is roughly the same.

jcohenadad:
> I've added your modifications. Unfortunately I do not fully understand your comment on using a different logic. It seems to me that the logic is roughly the same.

your recent commit 1e7fc37d6c3ca9bf03a6e0fe6e73ba3b67043b41 addresses my comment-- so it's all good now 👍

PR #2641, opened 2020-03-22
-initz flag and label value correction

### LINKED ISSUE #2608: Label value specified by -initz flag is shifted by one value
### Description

In the current implementation (b5d01bdad5248be280d53c3786e19ba2ba9f64c9), when specifying `-initz`, the value corresponds to the upper vertebral level whereas it should be the lower levels. E.g., Label 4 is interpreted as C4/C5 disc, whereas it should correspond to the C3/C4 disc.

The reason for this confusion is that in the past, we used to define disc labels that way, but since then we modified the convention. The convention for disc labeling is made clear notably in the recent [London'20 SCT course](https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%5FCourse%5F20200121) (page 47), but apparently this function has not been updated appropriately. The usage, though is correct. But what the actual function does is incorrect. 

### Proposed solution

Shift the label value by one (in the code) to make it correct.

PaulBautin:
> In the current implementation (b5d01bd), when specifying -initz, the value corresponds to the upper vertebral level whereas it should be the lower levels. E.g., Label 4 is interpreted as C4/C5 disc, whereas it should correspond to the C3/C4 disc.

The bug occurred when user used -initz flag but not when automatically finding c2c3 disc. This is because automatic c2c3 used the precedent label value convention (label value 2).

I've updated sct_label_vertebrae to use the new label value convention (label value 3). 

However I've left vertebral_detection (core.py) use the precedent convention and shiftted it's init_disc value by one (should it be done in sct_label_vertebrae?). This could have for consequence to crash other programs still using the precedent convention. 

DONE:
-No change in the usage
-modification in core.py: label value is shifted by one  
-flag -initz takes correct label value
-C2_C3 automatic detection now takes label value 3
-Could the modification in core.py crash other scripts? 

Fixes #2608


scripts/sct_label_vertebrae.py
```diff
@@ -315,9 +315,6 @@ def main(args=None):
             im_label.save(fname_labelz)
         elif fname_initlabel:
             import sct_label_utils
-            # subtract "1" to label value because due to legacy, in this code the disc C2-C3 has value "2", whereas in the
-            # recent version of SCT it is defined as "3". Therefore, when asking the user to define a label, we point to the
-            # new definition of labels (i.e., C2-C3 = 3).
             sct_label_utils.main(['-i', fname_initlabel, '-add', '-1', '-o', fname_labelz])
```
jcohenadad: to be consistent, if you remove the explanation you also need to remove the code that actually does the "minus one"

jcohenadad:
> Could the modification in core.py crash other scripts?

good question-- i guess we (incl. you) need to find out :-)

scripts/sct_label_vertebrae.py
PaulBautin: -I've changed `sct_label_utils` for `Image(fname_initlabel).save(fname_labelz)`. It seems to work fine. 
-Responding to second comment. vertebral_detection in `core.py` uses the precedent convention. Should I shift label value in `core.py`  and see what happens or shift label value in `sct_label_vertebrae`

jcohenadad: > -I've changed sct_label_utils for Image(fname_initlabel).save(fname_labelz). It seems to work fine.

I don't see your change. Which commit are you referring to? But regardless, my point was that you should not do this "minus one" operation anymore (for the same reason you removed the comment)

jcohenadad: > -Responding to second comment. vertebral_detection in core.py uses the precedent convention. Should I shift label value in core.py and see what happens or shift label value in sct_label_vertebrae

I don't know which "second comment." you are referring to (there is only one comment in this thread). If you are referring to another post in this PR, a good habit within Github is to add a hyperlink to the post you are referring to.

scripts/sct_label_vertebrae.py
```diff
@@ -315,8 +315,7 @@ def main(args=None):
             # which does not apply a convolution across all voxels (highly inneficient)
             im_label.save(fname_labelz)
         elif fname_initlabel:
-            import sct_label_utils
-            sct_label_utils.main(['-i', fname_initlabel, '-add', '-1', '-o', fname_labelz])
+            Image(fname_initlabel).save(fname_labelz)
```
jcohenadad: no need to read and save the same file: 
```suggestion
            fname_labelz = fname_initlabel
```

jcohenadad:
I've tested the following options and the script works as expected (i.e. no bug introduced by this PR):
~~~
 -initz <list of: int>
 -initcenter <int>
 -initlabel <file>
~~~

PR #2646, opened 2020-03-29
Fit centerline across all slices with input segmentation

### LINKED ISSUE #2629: Fit centerline across all slices with input segmentation
When running `sct_create_mask -p CENTERLINE`, if the centerline has some slices that are missing (i.e., empty slice), then the mask will also be missing for those slices, which could be problematic when the mask is subsequently used for `sct_dmri_moco`, where all slices need to have a non-zero mask.

To prevent that, we could have an additional option in `sct_get_centerline -m` where users could input the segmentation, and the output would be the segmentation fitted (e.g. bspline, etc.) for all slices.

PaulBautin:
> When running sct_create_mask -p CENTERLINE, if the centerline has some slices that are missing (i.e., empty slice), then the mask will also be missing for those slices, which could be problematic when the mask is subsequently used for sct_dmri_moco, where all slices need to have a non-zero mask.

The new feature:  

> An additional option has been added in sct_get_centerline -m where users could input the segmentation, and the output would be the segmentation fitted (e.g. bspline, etc.) for all slices.

My understanding is: if the image has been segmented then the get_centerline function will not skip slices. This new option would be used for images where the spinal cord isn't visible on all slices.

DONE:
-I've added flag -m that takes for argument the centerline fitting method. 
-When the flag -m is called a new method "seg" is used to parameter the `get_centerline` function. 
-With method "seg" the function `get_centerline` has: input a segmented image and output the centerline fitted to the segmentation.

Fixes #2629

### DIFF
--- a/scripts/sct_get_centerline.py
+++ b/scripts/sct_get_centerline.py
@@ -36,9 +36,10 @@ def get_parser():
                       type_value="multiple_choice",
                       description="Method used for extracting the centerline.\n"
                                   "optic: automatic spinal cord detection method\n"
-                                  "viewer: manual selection a few points followed by interpolation.",
+                                  "viewer: manual selection a few points followed by interpolation\n"
+                                  "fitseg: automatic spinal cord detection method on allready segmented image.",
                       mandatory=False,
-                      example=['optic', 'viewer'],
+                      example=['optic', 'viewer', 'fitseg'],
                       default_value='optic')
 
     parser.add_option(name='-centerline-algo',
@@ -52,12 +53,6 @@ def get_parser():
                       description='Degree of smoothing for centerline fitting. Only for -centerline-algo {bspline, linear}.',
                       mandatory=False,
                       default_value=30)
-    parser.add_option(name='-m',
-                      type_value='multiple_choice',
-                      description='Algorithm for centerline fitting. Fits centerline across all slices with input segmentation',
-                      mandatory=False,
-                      example=['polyfit', 'bspline', 'linear', 'nurbs'],
-                      default_value='bspline')
     parser.add_option(name="-o",
                       type_value='file_output',
                       description='File name (without extension) for the centerline output files. By default, output'
@@ -97,10 +92,6 @@ def run_main():
     if "-method" in arguments:
         method = arguments["-method"]
 
-    if "-m" in arguments:
-        method = 'seg'
-        algo_fitting = arguments["-m"]
-
     # Contrast type
     contrast_type = ''
     if "-c" in arguments:
@@ -134,9 +125,8 @@ def run_main():
     if method == 'viewer':
         # Manual labeling of cord centerline
         im_labels = _call_viewer_centerline(Image(fname_data), interslice_gap=interslice_gap)
-    if method == 'seg':
+    if method == 'fitseg':
         im_labels = Image(fname_data)
-        param_centerline.algo_fitting = algo_fitting
     else:
         # Automatic detection of cord centerline
         im_labels = Image(fname_data)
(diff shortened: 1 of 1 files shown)


PaulBautin:
-I've removed option -m and added method "fitseg". 
-Users are able to select their fitting algo with "-centerline-algo".
-help for method "fitseg" has been updated.
If the user  selects a non segmented image no warning is displayed.

scripts/sct_get_centerline.py
```diff
@@ -16,10 +16,10 @@
 def get_parser():
     # Initialize the parser
     parser = Parser(__file__)
-    parser.usage.set_description("""This function extracts the spinal cord centerline. Two methods are 
+    parser.usage.set_description("""This function extracts the spinal cord centerline. Two methods are
```
jcohenadad:
```suggestion
    parser.usage.set_description("""This function extracts the spinal cord centerline. Three methods are
```

scripts/sct_get_centerline.py
```diff
@@ -16,10 +16,10 @@
...
     available: OptiC (automatic) and Viewer (manual). This function outputs (i) a NIFTI file with labels corresponding
```
jcohenadad: to update with new method

scripts/sct_get_centerline.py
```diff
@@ -36,9 +36,10 @@ def get_parser():
                       type_value="multiple_choice",
                       description="Method used for extracting the centerline.\n"
                                   "optic: automatic spinal cord detection method\n"
-                                  "viewer: manual selection a few points followed by interpolation.",
+                                  "viewer: manual selection a few points followed by interpolation\n"
+                                  "fitseg: automatic spinal cord detection method on allready segmented image.",
```
jcohenadad:
```suggestion
                                  "fitseg: fit a regularized centerline on an already-existing cord segmentation. It will interpolate if slices are missing and extrapolate beyond the segmentation boundaries (i.e., every axial slice will exhibit a centerline pixel).",
```

scripts/sct_get_centerline.py
```diff
@@ -125,6 +125,8 @@ def run_main():
     if method == 'viewer':
         # Manual labeling of cord centerline
         im_labels = _call_viewer_centerline(Image(fname_data), interslice_gap=interslice_gap)
+    if method == 'fitseg':
```
jcohenadad: should be changed for `elif` now that you have 3 cases

scripts/sct_get_centerline.py
```diff
@@ -125,6 +125,8 @@ def run_main():
...
+        im_labels = Image(fname_data)
     else:
```
jcohenadad: change for elif method == 'optic' and add an else to catch error (better programming habit)

PaulBautin:
I've updated the parser (description and -method fitseg) and added condition error after 'if method'

scripts/sct_get_centerline.py
```diff
@@ -125,11 +125,18 @@ def run_main():
...
+    elif method == 'fitseg':
+        im_labels = Image(fname_data)
+    elif method == 'optic':
         # Automatic detection of cord centerline
         im_labels = Image(fname_data)
         param_centerline.algo_fitting = 'optic'
         param_centerline.contrast = contrast_type
+    else:
+        error = 'ERROR: please refer to help'
+        sct.printv(error, type='error')
```
jcohenadad: could save one line of code-- and make message more explicit
```suggestion
        sct.printv("ERROR: The selected method is not available: {}".format(method), type='error')
```

PaulBautin:
I modified error message according to last review.

PR #2658, opened 2020-04-05
Migrate sct_run_batch to python

cfhammill:
Start of a PR to migrate the `sct_run_batch` functionality from bash to python. Presently just some stock imports and a basic parser.

Some open design questions:

Do we keep the params file, or let argparse handle it? Argparse is capable of reading arguments from files if that behaviour is desired (https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars)

Which of the parameters do we need. right now there are:
- [ ] PATH_PARENT
- [ ] PATH_DATA
- [ ] SUBJECT_PREFIX
- [ ] PATH_RESULTS
- [ ] PATH_QC
- [ ] PATH_LOG
- [ ] PATH_SEGMANUAL
- [ ] ONLY_PROCESS_THESE_SUBJECTS
- [ ] TO_EXCLUDE
- [ ] JOBS
- [ ] ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS

My proposed CLI arguments are:

- [ ] `--data-path`
- [ ] `--subject-prefix`
- [ ] `--out-path` with sub-directories for results/QC/log
- [ ] `--include` regex or glob
- [ ] `--exclude` regex or glob
- [ ] `--segmentation`
- [ ] `--jobs` 
- [ ] `--itk-threads`

then positional a position argument for task

How much of the original semantics do we keep. For example `sct_run_batch` runs things relative to the parameters file. If we keep the parameters file, do we think this is the best behaviour?


jcohenadad:
> Do we keep the params file, or let argparse handle it? Argparse is capable of reading arguments from files if that behaviour is desired (https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars)

i like the param file because it enforces reproducibility (i.e. people can ship a dataset with a parameter file). That being said, there are not that many fields in the parameters file, and many of them (e.g. PATH_RESULTS, PATH_LOG, etc.) are fine when set by default (in fact, we could also simply hard-code these values inside `sct_run_batch`). So I agree we could simply get rid of the param file.

And I agree with your proposed CLI. In terms of mandatory/optional, i suggest the following:
MANDATORY:
- [ ] `--data-path`
OPTIONAL:
- [ ] `--subject-prefix` --> sub-
- [ ] `--out-path` with sub-directories for results/QC/log --> ./
- [ ] `--include` regex or glob --> ''
- [ ] `--exclude` regex or glob --> ''
- [ ] `--segmentation` --> ''
- [ ] `--jobs` --> 1 (or should we grab system's number of cores automatically -- minus one so people can still use their OS without too much disruption)
- [ ] `--itk-threads` --> 1

cfhammill:
Ok, so if you're not opposed I think we should get rid of `parameters.sh`, we can include a feature where a user may have a file say `params.txt`  

```
--data-path
/home/chammill
--out-path 
/tmp/out
--jobs
4
```

then they can run

```
sct_run_batch @params.txt 
```

to use those params. This is supported in argparse. Or alternatively we can have a JSON file.

```
{
  data_path: "/home/chammill"
  out_path: "/tmp/out"
  jobs: 4
}
```
either detected automatically or passed as a command line argument like

```
sct_run_batch --pars params.json
```

For default number of cores I prefer 1 to ncores, on hpc for example you may not have access to all of the visible cores or the memory to use all of them.

cfhammill:
Ok I think this is ready for test. I've experimented with the London20 course data like so:

```
wget -O sct_course_london20.zip https://osf.io/bze7v/?action=download
unzip sct_course_london20.zip
sed -i "s|source|#source|" sct_course_london20/multi_subject/process_data.sh
sct_run_batch \
  --data-path sct_course_london20/multi_subject/data/ \
  --out-path multiproc_test \
  --jobs 3 \
  sct_course_london20/multi_subject/process_data.sh 
```

I've tried:
- different levels of parallelism, 1 and 3 jobs (haven't tried the negation indexing yet)
- tried inclusion and exclusion regexes

@jcohenadad Can you think of any features I missed or that we should consider adding?

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+import subprocess
+import re
+import time
+
+import spinalcordtoolbox.image as msct_image
+from spinalcordtoolbox.image import Image, concat_data
+from spinalcordtoolbox.utils import Metavar, SmartFormatter
+
+parser = argparse.ArgumentParser(
+  description = 'Wrapper to processing scripts, which loops across subjects. Data should be '
```
jcohenadad, line 31: BIDS structure is not mandatory for the script to run. In fact, the processing does not follow BIDS convention (where outputs should be placed under a separate derivatives/ folder). So I would relax this statement by just saying something like: "All subjects' folder should be placed under PATH_DATA and the file naming should be consistent across subjects. We recommend to use the BIDS convention (https://bids.neuroimaging.io/)"

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+  formatter_class = SmartFormatter,
+  prog=os.path.basename(__file__).strip('.py'))
+
+parser.add_argument("--jobs", type = int, default = 1,
+                    help = 'The number of jobs to run in parallel. '
+                           'either an integer greater than or equal to one '
+                           'specifying the number of cores, 0 or a negative integer '
+                           'specifying number of cores minus that number. For example '
+                           '--jobs -1 indicates run ncores - 1 jobs in parallel.')
+parser.add_argument("--data-path", help = 'Path containing subject directories in the BIDS format')
```
jcohenadad, line 51: to be consistent with "--path-segmanual", and also to follow [SCT's variable naming conventions](https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#variables)
```suggestion
parser.add_argument("--path-data", help = 'Path containing subject directories in the BIDS format')
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+import spinalcordtoolbox.image as msct_image
+from spinalcordtoolbox.image import Image, concat_data
+from spinalcordtoolbox.utils import Metavar, SmartFormatter
+
+parser = argparse.ArgumentParser(
+  description = 'Wrapper to processing scripts, which loops across subjects. Data should be '
+                'organized according to the BIDS structure: '
+                'https://github.com/sct-pipeline/spine_generic#file-structure.\n'
+                'The processing script (task) should accept a subject directory as its only argument. '
+                'Additional information is passed via environment variables \n'
```
jcohenadad, line 35:
```suggestion
                'Additional information is either passed via environment variables or using flags \n'
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+                'PATH_SEGMANUAL (--path-segmanual)\n'
+                'PATH_DATA (--data-path)\n'
```
jcohenadad, line 37: I'm wondering if we should not rename it "FOLDER_DATA" (same for LOG and QC), to make it explicit that here we expect a folder (not a full path to a folder).

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+from spinalcordtoolbox.image import Image, concat_data
+from spinalcordtoolbox.utils import Metavar, SmartFormatter
+
+parser = argparse.ArgumentParser(
+  description = 'Wrapper to processing scripts, which loops across subjects. Data should be '
+                'organized according to the BIDS structure: '
+                'https://github.com/sct-pipeline/spine_generic#file-structure.\n'
+                'The processing script (task) should accept a subject directory as its only argument. '
+                'Additional information is passed via environment variables \n'
+                'PATH_SEGMANUAL (--path-segmanual)\n'
```
jcohenadad, line 36: instead of listing all the shell env variables here, why not saying that the name of the env variables to create appears next to each flag (in capital letter)?

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,155 @@
...
+                    help = 'Optional regex used to filter the list of subject directories. Only process '
+                           'a subject if they do not match the regex. Exclusions are processed '
+                           'after inclusions')
+parser.add_argument("--path-segmanual", default = ".",
+                    help = 'A path containing manual segmentations to be used by the task program.')                
+parser.add_argument("--itk-threads", default = 1,
+                    help = 'Number of threads to use for ITK based programs including ANTs. Set to a low '
+                           'number to avoid a large increase in memory. Defaults to 1')
+parser.add_argument("task",
+                    help = 'Shell script used to process the data.'
```
jcohenadad, line 70:
```suggestion
                    help = 'Shell script used to process the data.')
```

scripts/sct_run_batch.py
cfhammill: Make sense, I will fix all the parser arguments to match the conventions in the document above.

scripts/sct_run_batch.py
cfhammill: I prefer PATH_DATA, folder as far as I know is windows terminology for path. On the other hand there is definitely a UX argument that many users on mac/windows would be more familiar with the name folder instead of path. 

I'll leave the decision to you, keeping PATH for now. 

Additionally this flag can handle both relative and absolute paths so I'm not sure I understand your last point.

scripts/sct_run_batch.py
cfhammill: Presently the script runner can't accept additional flags. We could add an additional argument `--script-args` which takes a quoted string to be passed to the script. What do you think?

scripts/sct_run_batch.py
cfhammill: I chose a wording similar to your suggestion but with no explicit reference to `PATH_DATA`, from the perspective of the python program it's `--path-data` but I think the key is that it's all in one directory. We can obviously change this to be closer to your original wording if you want.

scripts/sct_run_batch.py
cfhammill: Should all follow convention now

scripts/sct_run_batch.py
cfhammill: I went ahead and added this feature

scripts/sct_run_batch.py
cfhammill: What do you mean regarding the environment variables? I have coded it so that the subprocess inherits all environment variables from the parent process.

scripts/sct_run_batch.py
cfhammill: This is fixed in my new batch of commits

cfhammill:
I think this should be ready for another review.

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+parser.add_argument("--jobs", type = int, default = 1,
+                    help = 'The number of jobs to run in parallel. '
+                           'either an integer greater than or equal to one '
+                           'specifying the number of cores, 0 or a negative integer '
+                           'specifying number of cores minus that number. For example '
+                           '--jobs -1 indicates run ncores - 1 jobs in parallel.')
+parser.add_argument("--path-data", help = 'Setting for environment variable: PATH_DATA\n'
+                                          'Path containing subject directories in a consistent format')
+parser.add_argument("--subject-prefix", default = "sub-",
+                    help = 'Subject prefix, defaults to "sub-"')
```
jcohenadad, line 49: I anticipate a lot of users will not have their data organized as BIDS, so this won't work by default. How about setting the default to ''? If data *are* BIDS, the default '' will also work.

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                'following the BIDS convention (https://bids.neuroimaging.io/). '
+                'The processing script (task) should accept a subject directory as its only argument. '
+                'Additional information is passed via environment variables and the arguments '
+                'passed via `--task-args`',
+  formatter_class = SmartFormatter,
+  prog=os.path.basename(__file__).strip('.py'))
+
+parser.add_argument("--jobs", type = int, default = 1,
+                    help = 'The number of jobs to run in parallel. '
+                           'either an integer greater than or equal to one '
```
jcohenadad, line 42:
```suggestion
                           'Either an integer greater than or equal to one '
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+  description = 'Wrapper to processing scripts, which loops across subjects. Subjects '
+                'should be organized as folders within a single directory. We recommend '
+                'following the BIDS convention (https://bids.neuroimaging.io/). '
+                'The processing script (task) should accept a subject directory as its only argument. '
+                'Additional information is passed via environment variables and the arguments '
+                'passed via `--task-args`',
+  formatter_class = SmartFormatter,
+  prog=os.path.basename(__file__).strip('.py'))
+
+parser.add_argument("--jobs", type = int, default = 1,
```
jcohenadad, line 40: [according to PEP8](https://www.python.org/dev/peps/pep-0008/#other-recommendations) no whitespace in assignments inside function calls 
```suggestion
parser.add_argument("--jobs", type=int, default=1,
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                    help = 'The number of jobs to run in parallel. '
+                           'either an integer greater than or equal to one '
+                           'specifying the number of cores, 0 or a negative integer '
+                           'specifying number of cores minus that number. For example '
+                           '--jobs -1 indicates run ncores - 1 jobs in parallel.')
+parser.add_argument("--path-data", help = 'Setting for environment variable: PATH_DATA\n'
```
jcohenadad, line 46: "\n" is not interpreted
```suggestion
parser.add_argument("--path-data", help = 'Setting for environment variable: PATH_DATA. '
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                                          'Path containing subject directories in a consistent format')
+parser.add_argument("--subject-prefix", default = "sub-",
+                    help = 'Subject prefix, defaults to "sub-"')
+parser.add_argument("--path-output", default = "./",
+                    help = 'Base directory for environment variables:\n'
+                           'PATH_RESULTS=' + os.path.join('<path-output>', 'results') + '\n'
+                           'PATH_QC=' + os.path.join('<path-output>', 'QC') + '\n'
+                           'PATH_LOG=' + os.path.join('<path-output>', 'results') + '\n'
```
jcohenadad, line 54:
```suggestion
                           'PATH_LOG=' + os.path.join('<path-output>', 'log') + '\n'
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                           'either an integer greater than or equal to one '
+                           'specifying the number of cores, 0 or a negative integer '
+                           'specifying number of cores minus that number. For example '
+                           '--jobs -1 indicates run ncores - 1 jobs in parallel.')
+parser.add_argument("--path-data", help = 'Setting for environment variable: PATH_DATA\n'
+                                          'Path containing subject directories in a consistent format')
+parser.add_argument("--subject-prefix", default = "sub-",
+                    help = 'Subject prefix, defaults to "sub-"')
+parser.add_argument("--path-output", default = "./",
+                    help = 'Base directory for environment variables:\n'
```
jcohenadad, line 51: trick: use "R|" at the beginning to have \n interpreted. See: [example](https://github.com/neuropoly/spinalcordtoolbox/wiki/code-style)
```suggestion
                    help = 'R|Base directory for environment variables:\n'
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                           'PATH_RESULTS=' + os.path.join('<path-output>', 'results') + '\n'
+                           'PATH_QC=' + os.path.join('<path-output>', 'QC') + '\n'
+                           'PATH_LOG=' + os.path.join('<path-output>', 'results') + '\n'
+                           'Which are output paths for results, logs, and QC')
+parser.add_argument("--include",
+                    help = 'Optional regex used to filter the list of subject directories. Only process '
+                           'a subject if they match the regex. Inclusions are processed before exclusions')
```
jcohenadad, line 58:
```suggestion
                           'a subject if they match the regex. Inclusions are processed before exclusions.')
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+parser.add_argument("--exclude",
+                    help = 'Optional regex used to filter the list of subject directories. Only process '
+                           'a subject if they do not match the regex. Exclusions are processed '
+                           'after inclusions')
```
jcohenadad, line 62:
```suggestion
                           'after inclusions.')
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+parser.add_argument("--path-segmanual", default = ".",
+                    help = 'Setting for environment variable: PATH_SEGMANUAL\n'
```
jcohenadad, line 64:
```suggestion
                    help = 'R|Setting for environment variable: PATH_SEGMANUAL\n'
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                           'A path containing manual segmentations to be used by the task program.')                
+parser.add_argument("--itk-threads", default = 1,
+                    help = 'Setting for environment variable: ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS\n'
```
jcohenadad, line 67:
```suggestion
                    help = 'R|Setting for environment variable: ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS\n'
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                    help = 'Optional regex used to filter the list of subject directories. Only process '
+                           'a subject if they match the regex. Inclusions are processed before exclusions')
+parser.add_argument("--exclude",
+                    help = 'Optional regex used to filter the list of subject directories. Only process '
+                           'a subject if they do not match the regex. Exclusions are processed '
+                           'after inclusions')
+parser.add_argument("--path-segmanual", default = ".",
+                    help = 'Setting for environment variable: PATH_SEGMANUAL\n'
+                           'A path containing manual segmentations to be used by the task program.')                
+parser.add_argument("--itk-threads", default = 1,
```
jcohenadad, line 66: Maybe we could force it to be int

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,159 @@
...
+                    help = 'Setting for environment variable: ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS\n'
+                           'Number of threads to use for ITK based programs including ANTs. Set to a low '
+                           'number to avoid a large increase in memory. Defaults to 1')
+parser.add_argument("--task-args", default = "",
+                    help = 'A quoted string with extra flags and arguments to pass to the task script. '
+                           'The constructed call will look like <task> <subject> <task-args>')
```
jcohenadad, line 72: it is not clear what "subject" means here. Also, it would be nice to be explicit and exhaustive about what are the possible values for task-args

scripts/sct_run_batch.py
cfhammill: @jcohenadad can you clarify? Thanks

scripts/sct_run_batch.py
cfhammill: @jcohenadad was there more I needed to do for this?

scripts/sct_run_batch.py
cfhammill: I'll clarify in the help that this default works for BIDS, and that an empty string can be used if the subject directories don't share a prefix.

scripts/sct_run_batch.py
cfhammill: `task-args` is in response to your suggestion above to allow the user to pass extra flags, I suspect this will generally be unused, but maybe we could clarify with a concrete example? An exhaustive list is impossible, given that we don't know in advance what script is being run and what arguments it might require.

scripts/sct_run_batch.py
cfhammill: I've fixed in the code. @jcohenadad, why does `'R|'` do the opposite of its usual behaviour here? I see in the linked document that this is the behaviour of `SmartFormatter`, but my understanding is that usually `'R|'` prevents interpretation of escape characters.

scripts/sct_run_batch.py
cfhammill: @jcohenadad, let me know if you're happy with the new wording here.

scripts/sct_run_batch.py
cfhammill: I started using Flake8 for linting and autopep8 for auto reformatting, shouldn't happen again!

scripts/sct_run_batch.py
cfhammill: @jcohenadad Can you confirm if the new wording is ok?

scripts/sct_run_batch.py
jcohenadad: > Additionally this flag can handle both relative and absolute paths so I'm not sure I understand your last point.

you're right-- i was actually making the point for PATH_RESULTS, PATH_LOG, etc. which are actually single "folder" (not paths with several folder depths). But yes, PATH_DATA should stay PATH_DATA

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,166 @@
...
+parser.add_argument('--subject-prefix', default="sub-",
+                    help='Subject prefix, defaults to "sub-" which is the prefix used for BIDS directories. '
+                    'If the subject directories do not share a common prefix, and empty string can be '
+                    'passed here.')
+parser.add_argument('--path-output', default="./",
+                    help='R|Base directory for environment variables:\n'
+                    'PATH_RESULTS=' + os.path.join('<path-output>', 'results') + '\n'
+                    'PATH_QC=' + os.path.join('<path-output>', 'QC') + '\n'
+                    'PATH_LOG=' + os.path.join('<path-output>', 'log') + '\n'
+                    'Which are output paths for results, logs, and QC')
```
jcohenadad, line 55:
```suggestion
                    'Which are respectively output paths for results, QC and logs')
```

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,166 @@
...
+                    'Either an integer greater than or equal to one '
+                    'specifying the number of cores, 0 or a negative integer '
+                    'specifying number of cores minus that number. For example '
+                    '--jobs -1 indicates run ncores - 1 jobs in parallel.',
+                    metavar=Metavar.int)
+parser.add_argument("--path-data", help='R|Setting for environment variable: PATH_DATA\n'
+                                        'Path containing subject directories in a consistent format')
+parser.add_argument('--subject-prefix', default="sub-",
+                    help='Subject prefix, defaults to "sub-" which is the prefix used for BIDS directories. '
+                    'If the subject directories do not share a common prefix, and empty string can be '
```
jcohenadad, line 48:
```suggestion
                    'If the subject directories do not share a common prefix, an empty string can be '
```

scripts/sct_run_batch.py
jcohenadad: perfect! just a [small typo](https://github.com/neuropoly/spinalcordtoolbox/pull/2658/files#r413449751)

scripts/sct_run_batch.py
jcohenadad: good question! because we are using two formatters (code [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/utils.py#L67)): one that adjust the text break with the terminal width, and another one that can interpret carriage return if "|R" is present. Unfortunately, both formatters cannot be used at the same time.  See example of the formatted usage from your latest commit 65429fd4e339e124dd0cce54d3fe6995f758a495:
~~~
# --> here, "|R" was not used and usage appropriately breaks lines and left-align it
  --exclude EXCLUDE     Optional regex used to filter the list of subject
                        directories. Only process a subject if they do not match the
                        regex. Exclusions are processed after inclusions.
# --> here, "|R" was used, and there is an ugly breaking of the word "task", warping it at the extreme left of the terminal. 
  --path-segmanual PATH_SEGMANUAL
                        Setting for environment variable: PATH_SEGMANUAL
                        A path containing manual segmentations to be used by the t
ask program.
~~~
If you or some else has a workaround for this, it would be awesome!

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,166 @@
...
+                    help='R|Setting for environment variable: ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS\n'
+                    'Number of threads to use for ITK based programs including ANTs. Set to a low '
+                    'number to avoid a large increase in memory. Defaults to 1',
+                    metavar=Metavar.int)
+parser.add_argument('--task-args', default='',
+                    help='A quoted string with extra flags and arguments to pass to the task script. '
+                    'The constructed call will look like <task> <subject> <task-args>. Where <subject> is the '
+                    'base name of the subject directory. For example, if the subject path is '
+                    '/home/path-data/sub-01 then <subject> will be sub-01. `--task-args` can include any number '
+                    'of additional arguments and flags')
```
jcohenadad, line 76: I find it confusing to have the description of constructed call written " <task> <subject> <task-args>" because from a UX, people might think this is how they should call sct_run_batch (whereas you are describing how arguments will be passed on to the <task> script). So i suggest this modification:
```suggestion
                    help='A quoted string with extra environment variables to pass to the <task> script. '
                    'For example: sct_run_batch process_data.sh --task-args "VAR1=23 VAR2=/usr/bob"')
```

scripts/sct_run_batch.py
jcohenadad: ah! ok i didn't understand at first what this flag was for. I've suggested a rewording.

scripts/sct_run_batch.py
```diff
@@ -0,0 +1,166 @@
...
+    description='Wrapper to processing scripts, which loops across subjects. Subjects '
+    'should be organized as folders within a single directory. We recommend '
+                'following the BIDS convention (https://bids.neuroimaging.io/). '
+                'The processing script (task) should accept a subject directory as its only argument. '
+                'Additional information is passed via environment variables and the arguments '
+                'passed via `--task-args`',
+    formatter_class=SmartFormatter,
+    prog=os.path.basename(__file__).strip('.py'))
+
+parser.add_argument("--jobs", type=int, default=1,
```
jcohenadad, line 37: So far the (unofficial) convention in SCT for long flags is to only use single dash. An increasing number of recent functions have used that convention so it would be great to do it for this function as well

jcohenadad:
Tried the script with commit 65429fd4e339e124dd0cce54d3fe6995f758a495 and got the following error:

<details>

~~~
(venv_sct) julien-macbook:~/sct_course_london20/multi_subject $ sct_run_batch --path-data data/ process_data.sh 
Running sub-01. See log file /Users/julien/sct_course_london20/multi_subject/log/process_data_sub-01.log
Running sub-05. See log file /Users/julien/sct_course_london20/multi_subject/log/process_data_sub-05.log
Running sub-03. See log file /Users/julien/sct_course_london20/multi_subject/log/process_data_sub-03.log
"""
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/Users/julien/code/sct/scripts/sct_run_batch.py", line 133, in run_single
    stderr=subprocess.STDOUT)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'process_data.sh': 'process_data.sh'
"""

The above exception was the direct cause of the following exception:

  File "/Users/julien/code/sct/scripts/sct_run_batch.py", line 153, in <module>
    p.map(run_single, subject_dirs)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 266, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
FileNotFoundError: [Errno 2] No such file or directory: 'process_data.sh': 'process_data.sh'
~~~

</details>

Then, I tried it from outside the folder, and got another error:

<details>

~~~
(venv_sct) julien-macbook:~ $ sct_run_batch --path-data sct_course_london20/multi_subject/data/ sct_course_london20/multi_subject/process_data.sh 
Running sub-01. See log file /Users/julien/log/process_data_sub-01.log
Running sub-05. See log file /Users/julien/log/process_data_sub-05.log
Running sub-03. See log file /Users/julien/log/process_data_sub-03.log
"""
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/Users/julien/code/sct/scripts/sct_run_batch.py", line 141, in run_single
    assert res.returncode == 0, 'Processing of subject {} failed'.format(subject)
AssertionError: Processing of subject sub-01 failed
"""

The above exception was the direct cause of the following exception:

  File "/Users/julien/code/sct/scripts/sct_run_batch.py", line 153, in <module>
    p.map(run_single, subject_dirs)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 266, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
AssertionError: Processing of subject sub-01 failed
~~~

</details>

scripts/sct_run_batch.py
cfhammill: Happy to change over, although I do like the argparse auto short flags, so `-j` can be used instead of `--jobs`

scripts/sct_run_batch.py
cfhammill: I like the change, although `--task-args` isn't compatible with environment variable, so I'll use something like `--task-args "-foo bar -baz /qux"`. It is possible to use environment variables, so if you want we could do 

```bash
VAR=23 VAR2=/usr/bob sct_run_batch process_data.sh --task-args "-foo bar -baz /qux"
```

but I think that overcomplicates it a bit

cfhammill:
@jcohenadad I'm surprised you're getting a file not found error. I'll see if I can reproduce, I don't think I ever ran from within the task directory but it's hard to see how that would cause a problem.

For the second test, could you post one of the logs either here or on slack?

cfhammill:
Somewhat interestingly python follows the POSIX behaviour of executables need to be at the end of a path. So:

```bash
 sct_run_batch --path-data data/ process_data.sh
```

fails, but

```bash
 sct_run_batch --path-data data/ ./process_data.sh
```

succeeds. I'll need to think about how to fix this. One simple solution is to absolute path the script file, we then say task cannot be a program on the users `$PATH`, which I think is ok? @jcohenadad do you agree?

As for the subjects failing, did you run the `sed` command on the task script?

Also I see that the program spawned 3 jobs before failing, on my machine if I don't specify `--jobs` it would only spawn a single job before dying. Did you include the `--jobs` flag, or is it computing the number of jobs incorrectly on mac.

scripts/sct_run_batch.py
jcohenadad: that sounds like a good discussion for our next SCT meeting ;-) (added to the agenda)

cfhammill: Converted in my last commit.

scripts/sct_run_batch.py
cfhammill: Followed you suggestion up to the suggestion above without the environment variables

scripts/sct_run_batch.py
cfhammill: Ahh this seems to be a tricky one. I'll give it a look and see if I can figure out a workaround

scripts/sct_run_batch.py
jcohenadad: ahhh! i still did not understand the purpose of "--task-args", now with your example implemented in 5aff84ba1ad229960e01b8ec66e9f4b38d80aa59 i finally get it :-)

jcohenadad:
@cfhammill you probably want to remove this entire section from the installer: https://github.com/neuropoly/spinalcordtoolbox/blob/cfh/migrate-sct-run-batch_2655/install_sct#L575

Also, you can remove the shell scripts (since they will be replaced with your function). Keeping them will introduce confusion (I think).

jcohenadad:
> @jcohenadad I'm surprised you're getting a file not found error. I'll see if I can reproduce, I don't think I ever ran from within the task directory but it's hard to see how that would cause a problem.
> As for the subjects failing, did you run the sed command on the task script?

Ah! I'm so sorry I didn't run the "sed" function 😬, which led to the observed issue (due to the source command inside the shell script). For the records, `sed` command had to be modified with "-e" added to work on OSX (see [here](https://stackoverflow.com/questions/2320564/sed-i-command-for-in-place-editing-to-work-with-both-gnu-sed-and-bsd-osx)):
~~~
sed -i -e "s|source|#source|" sct_course_london20/multi_subject/process_data.sh
~~~ 

> For the second test, could you post one of the logs either here or on slack?

i guess the issue has now been solved (with both calls)

> Also I see that the program spawned 3 jobs before failing, on my machine if I don't specify --jobs it would only spawn a single job before dying. Did you include the --jobs flag, or is it computing the number of jobs incorrectly on mac.

Not sure what was happening before, but -jobs is now behaving correctly on my machine. Default: only one job at a time.

cfhammill:
Great news, working on removing the shell scripts now, and fixing the help formatting.

scripts/sct_run_batch.py
cfhammill: I don't feel strongly about single vs. double tick, but happy to talk about it at the next meeting

scripts/sct_run_batch.py
cfhammill: As far as I can tell, the `_fill_text` method of `SmartFormatter` is never getting used. I tried instrumenting with a print statement with no luck. Instrumenting `_split_lines` does however get used. So it looks like `argparse.HelpFormatter` is doing most of the work. Should this be a separate issue?

jcohenadad: Thanks for digging! Yes if you could open an issue that would be awesome 🙏

cfhammill: Conversation moved to #2685

cfhammill:
@jcohenadad, I think I'm almost ready to take draft status off. Anything you want to see before I do? I haven't incorporated config files either via the argparse native line reader or a manual JSON. Do you want either of those before I change status?

jcohenadad:
> @jcohenadad, I think I'm almost ready to take draft status off. Anything you want to see before I do? 

nothing on top of my head :-)

> I haven't incorporated config files either via the argparse native line reader or a manual JSON. Do you want either of those before I change status?

No need for now. Let's give this implementation a test run across different scenario. If we notice this feature would be useful we can always add it in the future

jcohenadad:
@cfhammill i've added this to make sure we update the doc: https://github.com/neuropoly/spinalcordtoolbox/issues/2697

is it ready for merging?

cfhammill:
@jcohenadad yes I think it should be ready to go!

PR #2660, opened 2020-04-16
Fixed forgot to reorient mask

jcohenadad:
If input image is not RPI and mask is input, all images were reoriented to RPI except the mask, which made some registration algorithms relying on a mask to fail (e.g. slicereg). This PR fixes it.

### DIFF
--- a/scripts/msct_register.py
+++ b/scripts/msct_register.py
@@ -211,6 +211,9 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
         Image('dest_seg.nii').change_orientation("RPI").save('dest_seg_RPI.nii')
     if fname_dest_label:
         Image('dest_label.nii').change_orientation("RPI").save('dest_label_RPI.nii')
+    if fname_mask:
+        # TODO: change output name
+        Image('mask.nii.gz').change_orientation("RPI").save('mask.nii.gz')
 
     if identity:
         # overwrite paramregmulti and only do one identity transformation


scripts/msct_register.py
```diff
@@ -211,6 +211,9 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
         Image('dest_seg.nii').change_orientation("RPI").save('dest_seg_RPI.nii')
     if fname_dest_label:
         Image('dest_label.nii').change_orientation("RPI").save('dest_label_RPI.nii')
+    if fname_mask:
+        # TODO: change output name
+        Image('mask.nii.gz').change_orientation("RPI").save('mask.nii.gz')
```
kousu, line 216: It seems like you want this

```suggestion
        Image('mask.nii.gz').change_orientation("RPI").save('mask_RPI.nii.gz')
```

but I assume this needs more fixes elsewhere.

kousu:
This seems reasonable but I don't know how to trigger the bug. Can you provide an command line using `sct_testing_data/` that demonstrates the problem?

scripts/msct_register.py
kousu: I think you would have to change

a.

https://github.com/neuropoly/spinalcordtoolbox/blob/c8ffd6d4d9d4d6c03afd4c26fc6707d1dc104e48/scripts/msct_register.py#L408L-L414

to read `param.fname_mask` instead of hardcoding the value

b. Remove fname_mask from `param` or from the kwargs; it's duplicated here

https://github.com/neuropoly/spinalcordtoolbox/blob/c8ffd6d4d9d4d6c03afd4c26fc6707d1dc104e48/scripts/sct_register_multimodal.py#L319

which means you can't make the edit you would need to `fname_mask` without getting into a tricky inconsistent state.

kousu: I think it's okay as a quickfix. Thanks for tagging me and giving me the chance to read through some of the other parts of the code. I'm gathering a clearer idea of some cleanups to target.

jcohenadad: yes this is a quick-and-dirty fix-- i  have too many other priorities so i cannot dedicate any more time to this

jcohenadad:
> This seems reasonable but I don't know how to trigger the bug. Can you provide an command line using `sct_testing_data/` that demonstrates the problem?

sorry i dont have the time--

PR #2639, opened 2020-03-19
Centralized all deep learning segmentation tasks with new function "sct_deepseg" and refactored sct_download_data

jcohenadad:
With the need to more custom deep learning segmentation models and architectures, as well as new tasks (e.g. multi-class segmentation, CSF segmentation, etc.), it becomes more convenient to have a single CLI function, called `sct_deepseg`, that calls a specific model. 

One notable feature of this new function is that it will rely on the actively-maintained [ivadomed](https://github.com/neuropoly/ivado-medical-imaging) package. Another notable change is that we will now rely on the Pytorch framework (vs. Tensorflow).

For now, this new function will not replace `sct_deepseg_X`, but will be complementary to the existing functions, so we can gather feedback from users before making these functions deprecated. The "big" change will likely be associated with a major release (e.g. 5.0.0).

Changes include:
- Introduced new function `sct_deepseg` that can point to a specific model. Fixes #2626 
- Centralize hard-coded access to path for models and data. Fixes #2160
- Deep-learning models are installed under `$SCT_DIR/models`
  - New SCT variable: `sct.__models_dir__`
- Added useful functions under `sct.utils`: `add_suffix()`, `extract_fname()`, `tmp_create()`
- Implemented feature to set threshold for output segmentation
- Installing default models during SCT installation by adding: `sct_deepseg -download-default-models`
- Added ivadomed/master in `requirements.txt`
- refactored `sct_download_data` and move some functions under `spinalcordtoolbox.download`

Todo in subsequent PRs:
- Add a few default models in OSF with the appropriate json fields: `threshold`: #2695
- Document how to create new models: #2696
- Add testing for online models: #2714

### Test this branch on a development ivadomed version

```bash
# Checkout
cd $SCT_DIR
git fetch origin
git checkout -b jca/2626-deepseg origin/jca/2626-deepseg
# Activate SCT's venv
source ${SCT_DIR}/python/etc/profile.d/conda.sh
conda activate venv_sct
# Create alias to new sct_deepseg executable
pip install -e .
cp python/envs/venv_sct/bin/sct_deepseg bin/
# Clone ivadomed repos and pip install it inside SCT venv
cd ~  # or wherever you'd like to download that
git clone https://github.com/neuropoly/ivado-medical-imaging.git
cd ivado-medical-imaging
pip install -e .
```

PR #2662, opened 2020-04-18
rm`sct_check_dependencies`

kousu:
This script was useful, I'm guessing, for debugging peoples' failing installs. But the install log includes this information in the `pip install` part of it, and maintaining this script is a lot of effort for little gain. In particular, it fails on the dependency "ivadomed@git+https://github.com/neuropoly/ivado-medical-imaging@master" introduced by #2639. Even if we hacked around that, not all pypi packages have their import names match their package names, so the currently logic of trying to import names will bite us eventually.

Let's just trust `pip install` to tell us if it failed.

This is a fairly radical solution. It should be quick and effective, though.

Addresses https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/676705662 in #2639.


kousu:
This is failing, but with a test failure that's part of the new code introduced in #2639.

jcohenadad:
wow this is indeed a pretty radical PR ;-) 
`sct_check_dependencies` was put together over the past few years and we've added a few checks that have helped us tremendously with debugging on users' side, notably:
- make sure they are running the proper python's venv (sometimes users use their own python and don't have the proper dependencies)
- make sure they are using the intended SCT install (sometimes people have several SCT installs and don't know about their .bashrc)
- make sure some binaries are running properly (ANTs, propseg)
- gives us full picture about their OS, RAM, CPUs
- check if DISPLAY variable is set to debug PyQt stuff

So all in all, it was pretty useful. As you rightfully mentioned, some python's packages don't have the proper `__version__` field (or are imported with another name), but it is not a big pain to add special cases (you probably noticed some of these cases in the code).

As for the issue with the ivadomed@git, ultimately we expect this repository to become a pipy package, so that will solve it.

kousu:
I was being overeager to try to get it to build. Maybe this should just remove the `for dep_pkg, dep_ver_spec in get_dependencies():` loop, but keep the other checks.

PR #2642, opened 2020-03-24
Updated ANTs binaries to solve slow processing in some systems

jcohenadad:
When running it in some systems, motion correction is very slow. The solution was to re-build ANTs binaries on Ubuntu 16.04 and OSX. Fixes #947 

In this PR, we also take the opportunity to review the various distros that cause problems with some of the Python's libraries, causing problems such as https://github.com/neuropoly/spinalcordtoolbox/issues/2523.

TODO:
- Update binaries for Linux CentOS and OSX
- Check if Linux binaries work on CentOS
- Update call to ants function with SamplingStrategy=None, for reproducibility

## Compatibility of new Linux binaries (compiled on Ubuntu 16.04) with CentOS:

CentOS8 --> OK
<details><summary>CentOS6 --> FAIL </summary>

~~~
./isct_antsRegistration: /lib64/libc.so.6: version `GLIBC_2.15' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.15' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.14' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /usr/lib64/libstdc++.so.6: version `CXXABI_1.3.11' not found (required by ./isct_antsRegistration)
./isct_antsRegistration: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by ./isct_antsRegistration)
~~~

</details>

## Compatibility of Python's libraries
CentOS8 --> OK
<details><summary>CentOS6 --> FAIL</summary>

~~~
SCT info:
[...]

PR #2664, opened 2020-04-20
Improve moco for interleaved data

valosekj:
Added `-interleaved` option to `sct_dmri_moco.py`, if it is selected, newly created function in `moco.py` for splitting input data to even and odd datasets is called. This function splits data and calls `moco_wrapper` function two times, then merge datasets back together.
This current solution is created to influence original code as least as possible and needs improvements.

I tried current workaround on testing data `sub-issue2637` with commands from http://forum.spinalcordmri.org/t/sct-dmri-moco-doesnt-seem-to-correct-motion/358/6 (average DWI, segment SC, create mask, run moco).
Compare to results from original moco, I can see improvements in sagittal plane in slice-by-slice misalignment.

Fixes https://github.com/neuropoly/spinalcordtoolbox/issues/2637


scripts/sct_dmri_moco.py
```diff
@@ -141,13 +147,19 @@ def main():
...
     param.verbose = int(arguments.get('-v'))
 
     # Update log level
     sct.init_sct(log_level=param.verbose, update=True)
 
     # run moco
-    moco_wrapper(param)
+    if param.interleaved == 1:
+        # split input data to two datasets (even and odd slices), run moco in each sub-dataset and merge back the data
+        split_to_odd_and_even(param)
```
jcohenadad, line 160: i would call it `moco_wrapper_interleaved`, because `split_to_odd_and_even` is doing more than just splitting: it splits, applies moco, and then merges back.

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+
+    # Load data and get its dimensions
+    sct.printv('\nGet dimensions of data...', param.verbose)
+    im_data = Image(file_data)
+    nx, ny, nz, nt, px, py, pz, pt = im_data.dim
+    sct.printv('  ' + str(nx) + ' x ' + str(ny) + ' x ' + str(nz) + ' x ' + str(nt), param.verbose)
+
+    # Split 4D data along z dimension
+    #sct.printv('\nSplit along z dimension...', param.verbose)
+    im_data_split_list = split_data(im_data, 2)  # 0 - x, 1 - y, 2 - z, 3 - t
```
jcohenadad, line 903: you want to split along the SI direction. Dimension 2 does not always correspond to SI. E.g., if your orientation is AIL, this code will split along the right-left direction.

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+
+    sct.printv('\nWARNING: Data were acquired in interleaved mode.', param.verbose, 'warning')
```
jcohenadad, line 905: i wouldn't set it as a warning considering that the user asked for it

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+    # loop across odd slices
+    for index_odd in range(1, len(im_data_split_list), 2):
+        data_odd.append(im_data_split_list[index_odd])
+    concat_data(data_odd, dim=2).save(file_data_odd, verbose=0)  # Concatenate in z-axis and save
+
+    # Split mask if was passed
+    if param.fname_mask != '':
+        mask_data = Image(file_mask)
+        # Split 4D data along z dimension
+        im_mask_split_list = split_data(mask_data, 2)  # 0 - x, 1 - y, 2 - z, 3 - t
```
jcohenadad, line 926: same comment as before about assumed orientation

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+
+    # Run moco only on odd slices
+    sct.printv('\nStarting moco on odd slices...', param.verbose)
+    param.fname_data = 'data_odd.nii'
+    if param.fname_mask != '':
+        param.fname_mask = 'mask_odd.nii'
+    im_data_odd = moco_wrapper(param)
+
+    # TODO: files moco_params_x, moco_params_y.nii.gz and moco_params.tsv are overwritten now (even by odd)
+    #       - Do we want to preserve these files for both subdatasets or merge them?
```
jcohenadad, line 955: ideally we would like to save them into two separate folders (with explicit suffix _odd and _even)

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+        data_even.append(im_data_split_list[index_even])
+    concat_data(data_even, dim=2).save(file_data_even, verbose=0)  # Concatenate in z-axis and save
+
+    # Get only odd slices across all volumes
+    data_odd = []
+    # loop across odd slices
+    for index_odd in range(1, len(im_data_split_list), 2):
+        data_odd.append(im_data_split_list[index_odd])
+    concat_data(data_odd, dim=2).save(file_data_odd, verbose=0)  # Concatenate in z-axis and save
+
```
jcohenadad, line 921: there is probably a way to minimize code duplication here

spinalcordtoolbox/moco.py
```diff
@@ -853,3 +856,128 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+        # loop across even slices
+        for index_even in range(0, len(im_mask_split_list), 2):
+            mask_even.append(im_mask_split_list[index_even])
+        concat_data(mask_even, dim=2).save(file_mask_even, verbose=0)  # Concatenate in z-axis and save
+
+        mask_odd = []
+        # loop across odd slices
+        for index_odd in range(1, len(im_mask_split_list), 2):
+            mask_odd.append(im_mask_split_list[index_odd])
+        concat_data(mask_odd, dim=2).save(file_mask_odd, verbose=0)  # Concatenate in z-axis and save
```
jcohenadad, line 938: code duplication could be minimized here as well

spinalcordtoolbox/moco.py
```diff
@@ -966,18 +931,49 @@ def split_to_odd_and_even(param):
...
+
+    # Get only odd slices across all volumes
+    data_odd = []
+    # loop across odd slices
+    for index_odd in range(1, len(im_data_split_list), 2):
+        data_odd.append(im_data_split_list[index_odd])
+    # Concatenate in SI and save
+    concat_data(data_odd, dim=2).save(sct.add_suffix(file_name, '_odd'), verbose=0)
+
+    # TODO - reorient data back if they were not RPI
```
jcohenadad, line 978: right, this needs to be done (because down the line, the `moco_wrapper` treats axially- and sagittally-acquired data differently

spinalcordtoolbox/moco.py
```diff
@@ -867,6 +867,9 @@ def moco_wrapper_interleaved(param):
     orig_name = param.fname_data    # store orig name passed by user
     file_data = 'data.nii'          # corresponds to the full input data (e.g. dmri or fmri)
     file_mask = 'mask.nii'
+    file_moco_params_x = 'moco_params_x.nii.gz'
+    file_moco_params_y = 'moco_params_y.nii.gz'
+    file_moco_params_csv = 'moco_params.tsv'
```
jcohenadad, line 872: weird to call variable with _csv suffix if extension is tsv

spinalcordtoolbox/moco.py
```diff
@@ -915,29 +922,55 @@ def moco_wrapper_interleaved(param):
...
+        for name in file_moco_params_x, file_moco_params_y, file_moco_params_csv:
+            sct.mv(name, sct.add_suffix(name, '_odd'))
 
-    # TODO: files moco_params_x, moco_params_y.nii.gz and moco_params.tsv are overwritten now (even by odd),
-    #  merge them for posiible usage in GLM
+    # TODO: merge .tsv motion params file
 
     # Merge even and odd datasets after moco back together
     im_data_merged = im_data.copy()
+    moco_params_x_odd = Image('moco_params_x_odd.nii.gz')
```
jcohenadad, line 934: shouldn't this go under the "    if param.output_motion_param:" clause?

spinalcordtoolbox/moco.py
```diff
@@ -915,29 +922,55 @@ def moco_wrapper_interleaved(param):
...
+    moco_params_x_merged = np.zeros((1, 1, nz, nt), dtype=np.float32)     # create empty array for x motion params
+    moco_params_y_odd = Image('moco_params_y_odd.nii.gz')
+    moco_params_y_even = Image('moco_params_y_even.nii.gz')
+    moco_params_y_merged = np.zeros((1, 1, nz, nt), dtype=np.float32)     # create empty array for y motion params
     counter_even = 0
     counter_odd = 0
     for index in range(0, im_data_merged.dim[2]):
         if index % 2 == 0:
             im_data_merged.data[:, :, index, :] = im_data_even.data[:, :, counter_even, :]
+            moco_params_x_merged[:, :, index, :] = moco_params_x_even.data[:,:, counter_even, :]
```
jcohenadad, line 945: same comment here about the "    if param.output_motion_param:" clause

spinalcordtoolbox/moco.py
valosekj: You are right. I chose this variable name based on  https://github.com/neuropoly/spinalcordtoolbox/blob/be4f8e9890bfc0ba141a8961d21bc8a12c5c979b/spinalcordtoolbox/moco.py#L131
When you prefer csv over tsv, should I switch to csv over whole moco script?

jcohenadad: yes, csv would be nice.

jcohenadad: Actually no, let’s keep tsv. Let’s not extend the purpose of this PR too much.

valosekj: okay, I preserved tsv and added future TODO to replace tsv by csv

kousu:
Heads up, this might run into #2642. It doesn't edit `sct_dmri_moco` directly but it does change the underlying image registration algorithm; I had to update the snapshot tests to get CI to pass -- not by much, but by enough. When that gets merged you should rebase on top of it ((including reinstalling `sct_download_data -d binaries_linux -o bin/`)) early to catch any issues.

valosekj:
Can be reviewed/tested.
To recap; this PR instroduces interleaved option for `sct_dmri_moco.py` for dMRI data acquired in interleaved mode. 
When `-interleaved`  is set to `1`, `moco_wrapper_interleaved` inside `moco.py` is called. Input data are split into even and odd subdata and moco is run separately, then corrected subdata are merged back.
Tested on [sub-glen](https://github.com/spine-generic/data-single-subject/tree/master/sub-glen).

(One integrity test failed during `pip install` - https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/707626877 - is it caused by commits in this PR or is it any independent problem?)

jcohenadad:
> (One integrity test failed during pip install

was just a connection problem on the travis end

jcohenadad:
@valosekj wasn't the point of  https://github.com/neuropoly/spinalcordtoolbox/pull/2684 to implement a test for this current PR?

spinalcordtoolbox/moco.py
```diff
@@ -865,3 +868,209 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+    # Save to tmp dir
+    im_data_merged.save(sct.add_suffix(file_data, '_merged'), verbose=0)
+
+    # Generate b0_moco_mean and dwi_moco_mean
+    args = ['-i', os.path.join(path_tmp,sct.add_suffix(file_data, '_merged')),
+            '-bvec', param.fname_bvecs, '-a', '1', '-v', '0']
+    if not param.fname_bvals == '':
+        # if bvals file is provided
+        args += ['-bval', param.fname_bvals]
+    fname_b0, fname_b0_mean, fname_dwi, fname_dwi_mean = sct_dmri_separate_b0_and_dwi.main(args=args)
```
jcohenadad, line 975: would be nice to make the code compatible with `sct_fmri_moco`

scripts/sct_dmri_moco.py
```diff
@@ -99,6 +99,12 @@ def get_parser():
                       mandatory=False,
                       default_value=param_default.path_out,
                       example='dmri_moco_results/')
+    parser.add_option(name='-interleaved',
+                      type_value='multiple_choice',
+                      description='Interleaved acquisition: 0 = NOT-interleaved, 1 = interleaved',
+                      mandatory=False,
+                      default_value='0',
+                      example=['0','1'])
```
jcohenadad, line 107: i would use int here (i know the rest of the code uses str, but that was a mistake).

spinalcordtoolbox/moco.py
```diff
@@ -865,3 +868,209 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+    :param param: ParamMoco class
+    :return: None
+    """
+
+    orig_name = param.fname_data    # store orig name passed by user
+    file_data = 'data.nii'          # corresponds to the full input data (e.g. dmri or fmri)
+    file_mask = 'mask.nii'
+    file_moco_params_x = 'moco_params_x.nii.gz'
+    file_moco_params_y = 'moco_params_y.nii.gz'
+    file_moco_params_csv = 'moco_params.tsv'
```
jcohenadad, line 884: `file_moco_params_csv` should be renamed to `file_moco_params_tsv` otherwise it's confusing

spinalcordtoolbox/moco.py
```diff
@@ -865,3 +868,209 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+    # Update field in param (because used later in another function, and param class will be passed)
+    param.fname_data = 'data_odd.nii'
+    param.path_out = ''
+    if param.fname_mask != '':
+        param.fname_mask = 'mask_odd.nii'
+    im_data_odd = moco_wrapper(param)
+    if param.output_motion_param:
+        # Rename .nii and tsv moco parameters after moco (add _odd suffix)
+        for name in file_moco_params_x, file_moco_params_y, file_moco_params_csv:
+            sct.mv(name, sct.add_suffix(name, '_odd'))
```
jcohenadad, line 947: this duplicated code could easily be called by a private function (and not be duplicated)

spinalcordtoolbox/moco.py
```diff
@@ -865,3 +868,209 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+            if index % 2 == 0:
+                moco_params_x_merged[:, :, index, :] = moco_params_x_even.data[:, :, counter_even, :]
+                moco_params_y_merged[:, :, index, :] = moco_params_y_even.data[:, :, counter_even, :]
+                counter_even += 1
+            elif index % 2 != 0:
+                moco_params_x_merged[:, :, index, :] = moco_params_x_odd.data[:, :, counter_odd, :]
+                moco_params_y_merged[:, :, index, :] = moco_params_y_odd.data[:, :, counter_odd, :]
+                counter_odd += 1
+        # Save nii to tmp dir
+        import nibabel as nib
```
jcohenadad, line 998: import should be moved to the beginning of the file

spinalcordtoolbox/moco.py
```diff
@@ -865,3 +868,209 @@ def spline(folder_mat, nt, nz, verbose, index_b0 = [], graph=0):
...
+
+        # tsv motion parameters (slice-wise average)
+        moco_param = np.concatenate((np.squeeze(np.mean(moco_params_x_merged, 2)).reshape(-1, 1),
+                                     np.squeeze(np.mean(moco_params_y_merged, 2)).reshape(-1, 1)), axis=1)
+        with open(sct.add_suffix(file_moco_params_csv, '_merged'), 'wt') as out_file:
+            tsv_writer = csv.writer(out_file, delimiter='\t')
+            tsv_writer.writerow(['X', 'Y'])
+            for mocop in moco_param:
+                tsv_writer.writerow([mocop[0], mocop[1]])
+
```
jcohenadad, line 1013: you should add a unit test for this

spinalcordtoolbox/moco.py
valosekj: 👍 Thanks, I renamed variable `file_moco_params_csv` to `file_moco_params` inside `moco_wrapper_interleaved`. However, there is also similar discrepancy in original `moco_wrapper`:

https://github.com/neuropoly/spinalcordtoolbox/blob/048c0a5c4703c48ad1aab52838bf4200f3666101/spinalcordtoolbox/moco.py#L132

I do not want modify code of original (working and tested) `moco_wrapper` inside this PR. I will fix it when this PR will be finished.

scripts/sct_dmri_moco.py
valosekj: It looks that currently used parser (`from msct_parser import Parser`) does not support int. When I tried int, I got error. When I briefly checked other sct scripts using this parser (`sct_dmri_transpose_bvecs.py` or `sct_dmri_create_noisemask.py`), all of them use str. For the future, we can consider to switch to `argparse`.

spinalcordtoolbox/moco.py
valosekj: Okay. It means to extend existing [test_sct_dmri_moco.py](https://github.com/neuropoly/spinalcordtoolbox/blob/master/testing/test_sct_dmri_moco.py) script by interleaved data, right?

spinalcordtoolbox/moco.py
jcohenadad: > I do not want modify code of original moco_wrapper (which works and is tested) inside this PR. I will fix it when this PR will be finished (I opened separated issue for it #2792).

thanks, this is a sensible approach

scripts/sct_dmri_moco.py
jcohenadad: indeed, we eventually need to get rid of the custom `msct_parser`.

valosekj:
> @valosekj wasn't the point of #2684 to be able to develop a specific test to quantify the improvements of your moco-interleaved feature?

You are right! I meant by "_Tested on sub-glen._" that all parts of code work and script finish without errors. I'll test/validate performance of this PR on some systemic data asap.

jcohenadad:
> > @valosekj wasn't the point of #2684 to be able to develop a specific test to quantify the improvements of your moco-interleaved feature?
> 
> You are right! I meant by "_Tested on sub-glen._" that all parts of code work and script finish without errors. I'll test/validate performance of this PR on some systemic data asap.

ok let me know when the PR is at its final stage for me to review

valosekj:
I have tested this PR on a synthetic 4D dummy interleaved data generated by [create_test_data.py](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/testing/create_test_data.py).
I run both moco (`sct_dmri_moco -interleaved 0`) and moco_interleaved (`sct_dmri_moco -interleaved 1`) on this data. I can see that generated moco_params.tsv files differ between original moco and moco_interleaved. But visually in fsleyes, I can see only very slight (or mostly no) difference between raw image and images after moco or moco_interleaved.
Therefore, I am now a little bit unsure how exactly validate efficiency of the interleaved approach. 

<details><summary>Code how to generate synthetic interleaved 4D data:</summary>

```
from spinalcordtoolbox.testing.create_test_data import dummy_segmentation_4d
DEBUG = True  # Set to True to save images
seg_4D = dummy_segmentation_4d(vol_num=10, create_bvecs=True, size_arr=(64, 64, 49), pixdim=(1,1,1),radius_RL=13.0, radius_AP=5.0, shape='ellipse', interleaved=True, debug=DEBUG)
```

</details>

[dummy_4D_interleaved.zip](https://github.com/neuropoly/spinalcordtoolbox/files/4938121/dummy_4D_interleaved.zip)

jcohenadad:
> But visually in fsleyes, I can see only very slight (or mostly no) difference between raw image and images after moco or moco_interleaved.
Therefore, I am now a little bit unsure how exactly validate efficiency of the interleaved approach.

what if you try to increase the amount of motion in your simulation? how much motion is there now?

jcohenadad:
i've looked at the simulated data uploaded  in https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-660115334, motion is very small (sub-mm) in most of the middle slices:
anim

so i suggest you increase motion to be able to better validate the performance of your algorithm @valosekj 

also: i suggest you upload gif anim-- more convenient that downloading zip, opening fsleyes, etc.

valosekj:
> > But visually in fsleyes, I can see only very slight (or mostly no) difference between raw image and images after moco or moco_interleaved.
> > Therefore, I am now a little bit unsure how exactly validate efficiency of the interleaved approach.
> 
> what if you try to increase the amount of motion in your simulation? how much motion is there now?

Generated image has 1mm isotropic resolution (or should have, because, to be precise - `pixdim` flag is set to (1,1,1):

https://github.com/neuropoly/spinalcordtoolbox/blob/67c679deadd8af0c2eb7445041b0c61628d8bbfa/spinalcordtoolbox/testing/create_test_data.py#L217
but fsleyes states at pixdim "[unknown units]"):
image

Slice-wise shift is in range between 0.5-3mm (or again - unknown units?). Unfortunately now set as fixed value (I should change it to be input arg):
https://github.com/neuropoly/spinalcordtoolbox/blob/67c679deadd8af0c2eb7445041b0c61628d8bbfa/spinalcordtoolbox/testing/create_test_data.py#L232

(for context; middle slices have lower shift that outer ones due to polynomial function - see [this image](https://user-images.githubusercontent.com/39456460/81703240-98410c00-946c-11ea-99ad-e824faa58b95.png))

I'll increase motion and will update gif, thanks 👍

jcohenadad:
> middle slices have lower shift that outer ones due to polynomial function

that's not necessarily the case, e.g. with 2nd-order you could have largest shift in the middle (depending on how you define you constant 0-term offset)

valosekj:
> > middle slices have lower shift that outer ones due to polynomial function
> 
> that's not necessarily the case, e.g. with 2nd-order you could have largest shift in the middle (depending on how you define you constant 0-term offset)

hm, currently, the order of polynomial is equal to number of slices (`nz`):
https://github.com/neuropoly/spinalcordtoolbox/blob/67c679deadd8af0c2eb7445041b0c61628d8bbfa/spinalcordtoolbox/testing/create_test_data.py#L160

Lower degree results into worse fit (i.e. slicewise regularized fit (orange) does not properly fit/simulate interleaved acquisition mode (blue dots)):
2d_polyfit_various_degree

jcohenadad:
@valosekj what exactly are the graphs https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-660344400 supposed to represent?

if the blue dots are the pixel shift in the Y direction across slices, for both odd and even slices, then why are you trying to fit both odd and even slices at the same time? shouldn't the simulated data be constructed by fitting one polynomial function to the odd slices, and another one to the even slices?

valosekj:
> if the blue dots are the pixel shift in the Y direction across slices, for both odd and even slices

Yes, blue dots represent the pixel shift in A-P (Y) direction across slices for one certain volume from 4D dataset. I added numbers to individual dots for clarity in figure below.

> shouldn't the simulated data be constructed by fitting one polynomial function to the odd slices, and another one to the even slices?

You mean something like is plotted in figure below? Then, however, individual pixel shift (blue dots) have to vary for odd and even slices. Otherwise, the fit would be linear (a.). I plotted two scenarios of variation for odd and even slices: constant variation (b.) or random variation (c.).

5th_polynomial

valosekj:
I generated some dummy 4D interleaved data using two polynomials (one fitting even slices and another one fitting odd slices) based on the [previous comment](https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-660689483). Voxel shift/motion varies randomly in interval from 0.5 to 3.5 for all examples below.

<details><summary>Example of 3rd order polynomial:</summary>

tmp_dummy_4d_20200721184427834248_degree_3

</details>

<details><summary>Example of 3rd order polynomial after moco_interleaved (in red are changes caused by moco):</summary>

tmp_dummy_4d_20200721184427834248_degree_3_moco

</details>

<details><summary>Example of fitting 3rd order polynomial for one volume:</summary>

debug_plot_0_volume_3_degree

</details >


<details><summary>Example of 10th order polynomial:</summary>

tmp_dummy_4d_20200721184449566868_degree_10

</details>

<details><summary>Example of 10th order polynomial after moco_interleaved (in red are changes caused by moco):</summary>

tmp_dummy_4d_20200721184449566868_degree_10_moco

</details>

<details><summary>Example of fitting 10th order polynomial for one volume:</summary>

debug_plot_0_volume_10_degree

</details >


Higher order of polynomial causes larger variation in S-I direction. So, I think that slowly-varying changes in the B0 field are better captured in polynomial of lower order (e.g. 3rd). @jcohenadad what do you think?

jcohenadad:
@valosekj i'm having hard time understanding the overall workflow. To me, the synthetic data could be made by:
- create two synthetic segmentations (e.g. the one produced by dummy_segmentation()), by choosing random scalar for each term of the polynomial function (i.e. `y=sum{a_i*x^i}`, with `a_i ∈ |-1, 1[`)
- interleave these two segmentations (one for odd, the other for even)
- loop across time points.

So, i don't really see the need for "fitting" polynomial functions here.

valosekj:
> * create two synthetic segmentations (e.g. the one produced by dummy_segmentation()), by choosing random scalar for each term of the polynomial function (i.e. `y=sum{a_i*x^i}`, with `a_i ∈ |-1, 1[`)

The current implementation of `dummy_segmentation` function does not allow polynomial function:

https://github.com/neuropoly/spinalcordtoolbox/blob/cbc18e92806138bd612957ab71b6e9cd394fa00d/spinalcordtoolbox/testing/create_test_data.py#L140-L152

I implemented polynomial function only for [interleaved option](https://github.com/neuropoly/spinalcordtoolbox/blob/cbc18e92806138bd612957ab71b6e9cd394fa00d/spinalcordtoolbox/testing/create_test_data.py#L153), but not globally for `dummy_segmentation`. However, I implemented it in bad way (discussion above). 
So, I'll start with reimplementation from scratch. The first step will be update of `dummy_segmentation` function to allow polynomial function, right? In one of your [previous comment](https://github.com/neuropoly/spinalcordtoolbox/issues/2678#issuecomment-620824747) you suggested to use code from `dummy_centrerline`. But I can see now, that `dummy_centrerline` switched from polynomial function to regularized curve meanwhile in this PR https://github.com/neuropoly/spinalcordtoolbox/pull/2754.

Instead of this, I can probably use `polyfit_1d` function to produce polynomial function, right?: https://github.com/neuropoly/spinalcordtoolbox/blob/cbc18e92806138bd612957ab71b6e9cd394fa00d/spinalcordtoolbox/centerline/curve_fitting.py#L15

jcohenadad:
> The first step will be update of dummy_segmentation function to allow polynomial function, right? 

👍

> In one of your previous comment you suggested to use code from dummy_centrerline. But I can see now, that dummy_centrerline switched from polynomial function to regularized curve meanwhile in this PR #2754. Instead of this, I can probably use polyfit_1d function to produce polynomial function, right?

👍

jcohenadad:
@valosekj are you still working on that PR?

valosekj:
> @valosekj are you still working on that PR?

@jcohenadad very sorry, when I worked on https://github.com/neuropoly/spinalcordtoolbox/pull/2835 I somehow forgot to finish and merge this PR.
I can see now that two script modified by this PR (`sct_dmri_moco.py ` and `moco.py`) have been moved meanwhile:

- scripts/sct_dmri_moco.py --> spinalcordtoolbox/spinalcordtoolbox/scripts/sct_dmri_moco.py 

- spinalcordtoolbox/moco.py --> spinalcordtoolbox/spinalcordtoolbox/moco.py

I am not sure now, if I should try to rebase this PR or start from scratch with moved scripts.

jcohenadad:
> I am not sure now, if I should try to rebase this PR or start from scratch with moved scripts.

the scripts only moved, so you should be able to resolve conflicts easily. What I would do (i know that's not the _proper_ git way), would be to 
- copy the files you worked on in a temp folder outside of SCT dir,
- git merge master (to update your branch with master)
- copy back the files you worked on in the proper directory

@kousu @joshuacwnewton might have other/better advices

valosekj:
> What I would do (i know that's not the _proper_ git way), would be to
> 
> * copy the files you worked on in a temp folder outside of SCT dir,
> * git merge master (to update your branch with master)
> * copy back the files you worked on in the proper directory

Thanks, I have tried it but I have had troubles with unresolved conflicts and unsuccessful rebase. 
I have created locally new branch derived from current master and implemented interleaved option for both `sct_dmri_moco.py` and `moco.py` again. If it is not problem, I would close this PR and discard [this branch](https://github.com/neuropoly/spinalcordtoolbox/tree/jv/2637-interleaved-dmri-moco) and open new PR from my fresh new branch (which is local so far).

valosekj:
Closed due to out-of-date with master and unresolved conflicts. Replaced by https://github.com/neuropoly/spinalcordtoolbox/pull/3172

PR #2672, opened 2020-04-24
sct_testing: Use boolean flags

kousu:
Switch to using standard boolean exist/nonexisting flags instead of passing an integer value. It's weird to say `--download 0` just to disable downloads, and also the types were broken, possibly because argparse changed?

These were weird:

```
(sct_venv) $ sct_testing -v

--
Spinal Cord Toolbox (git-master-be4f8e9890bfc0ba141a8961d21bc8a12c5c979b)

usage: sct_testing.py [-h] [--function FUNCTION [FUNCTION ...]]
                      [--download {0,1}] [--path PATH] [--remove-temps {0,1}]
                      [--jobs JOBS] [--verbose {0,1}] [--abort-on-failure]
                      [--continue-from CONTINUE_FROM] [--check-filesystem]
                      [--execution-folder EXECUTION_FOLDER]
sct_testing.py: error: argument --verbose/-v: expected one argument
```

```
(sct_venv) $ sct_testing -v 0

--
Spinal Cord Toolbox (git-master-be4f8e9890bfc0ba141a8961d21bc8a12c5c979b)

Traceback (most recent call last):
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_testing.py", line 619, in <module>
    main()
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_testing.py", line 227, in main
    sct.init_sct(log_level=param.verbose, update=True)  # Update log level
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_utils.py", line 73, in init_sct
    logger.setLevel(getattr(logging, dict_log_levels[log_level]))
KeyError: '0'
```

```
(sct_venv) $ sct_testing -v 1

--
Spinal Cord Toolbox (git-master-be4f8e9890bfc0ba141a8961d21bc8a12c5c979b)

Traceback (most recent call last):
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_testing.py", line 619, in <module>
    main()
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_testing.py", line 227, in main
    sct.init_sct(log_level=param.verbose, update=True)  # Update log level
  File "/home/kousu/src/neuropoly/spinalcordtoolbox/scripts/sct_utils.py", line 73, in init_sct
    logger.setLevel(getattr(logging, dict_log_levels[log_level]))
KeyError: '1'
```

Now this makes sense:

```
[...]

PR #2673, opened 2020-04-24
sct_testing: "allow" download

kousu:
I want to propose that `--download` means "allow" not "force" download.

It is more intuitive to me that `--download` would control whether `sct_testing` is allowed to download a dataset behind my back, than to say that it *must* download the dataset.
    
If you want to force redownloading you can use `git clean -df` or just `rm -r sct_testing_data; sct_testing`. There's no need for the default behaviour to be `rm -r sct_testing_data; sct_download_data -d sct_testing_data`

This depends on #2672.

### DIFF
--- a/scripts/sct_testing.py
+++ b/scripts/sct_testing.py
@@ -232,7 +232,7 @@ def main(args=None):
     param.path_data = os.path.abspath(param.path_data)
 
     # check existence of testing data folder
-    if not os.path.isdir(param.path_data) or param.download:
+    if not os.path.isdir(param.path_data) and param.download:
         downloaddata(param)
 
     # display path to data
--- a/scripts/sct_testing.py
+++ b/scripts/sct_testing.py
@@ -61,12 +61,12 @@ def fs_ok(sig_a, sig_b, exclude=()):
 # Parameters
 class Param:
     def __init__(self):
-        self.download = 0
+        self.download = False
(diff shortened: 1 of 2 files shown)


scripts/sct_testing.py
```diff
@@ -232,7 +232,7 @@ def main(args=None):
     param.path_data = os.path.abspath(param.path_data)
 
     # check existence of testing data folder
-    if not os.path.isdir(param.path_data) or param.download:
```
jcohenadad, line 235: the reason we forced download if folder exists is because sometimes the data are altered (new version, processed inputs), which could affect the integrity of testing. I remember having spent 1 hour on a bug, after realizing that the bug was caused by corrupted testing data (which were not re-downloaded). So if user asks for download, i would just download, regardless if the folder already exists

kousu: Ah okay so there is a good reason. Thanks. I'll close this.

kousu:
https://github.com/neuropoly/spinalcordtoolbox/pull/2673#discussion_r414769950

scripts/sct_testing.py
jcohenadad: maybe no need to close this? the changes in CLI flags are relevant, no?

kousu: I split it up, that's over in #/2672.

Issue #2677, opened 2020-04-26
Output lesion statistics per slice


jcohenadad:
> **Note**: This issue is part of a set of issues focused on the overlap in scope between `sct_process_segmentation` and `sct_analyze_lesion`.  Both of these scripts compute statistics on binary segments, but both scripts have certain features that the other lacks.
>
> - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2677
> - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378
> - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2707
> - https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2682

As currently implemented, `sct_analyse_lesion` can only provide level-wise statistics (with flag `-f`). It would be nice to have "ROI Occupied by lesion" and "Lesion distribution" calculated per-slice.

We could maybe consider using the same code that is used by `sct_process_segmentation` for dealing with aggregation between slices and/or vertebral levels. 

Related to [Forum post](http://forum.spinalcordmri.org/t/lesion-analysis-outputs/390)

Nilser3:
Hi Julien,
I have tried to solve the problem by introducing **-perslice** function from _sct_process_segmentation_ in _sct_analyze_lesion_ command:
![sct_analyze_lesion](https://user-images.githubusercontent.com/77469192/204771439-293ffb55-054f-4fcd-8ed6-dc10a57532ff.png)

In this way we would be obtaining the area of the lesions for each slice, so I have a problem. What would happen if I have two or more lesions in a slice?
![t2s_7t_ms](https://user-images.githubusercontent.com/77469192/204770898-ea06979e-93be-4b50-89f5-9576797b8016.png)

Maybe we could calculate the area of each lesion per label and per slice.
[it's a work in progress]

jcohenadad:
Good question. Given that the code already creates one label per lesion, I _think_ the approach would be to run the statistics per label, and for each label, have the option to do it per slice or per vertebral level. However, I would be careful with the implementation, because there are current discussions to try to leverage the existing APIs instead of duplicating them. See notably:
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2682

Nilser3:
[...]

PR #2682, opened 2020-05-01
Refactor sct_analyze_lesion

charleygros:
This PR aims at refactoring `sct_analyze_lesion`.

Main objectives:
- avoid duplication with existing code in `process_seg` and `aggregate_slicewise`.
- allow the call of metric computation modules from external scripts.

While doing it, might be worth considering these related issues:
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3681#discussion_r804822552
- #3691
- #2677


scripts/sct_analyze_lesion.py
```diff
@@ -71,13 +66,13 @@ def get_parser():
         default=None,
         required=False)
     optional.add_argument(
-        "-f",
-        help="Path to folder containing the atlas/template registered to the anatomical image. If provided, the "
-             "function computes: (i) the distribution of each lesion depending on each vertebral level and on each"
+        "-t",
```
charleygros, line 69: Old version: template analysis was not mandatory --> in that case the output was simply "lesion count, volume, length".

The template allows to compute statistics per tracts / levels.

@jcohenadad: Do we want to make this template analysis mandatory? It does make the processing longer. If not, do you have a suggestion of parser_parameter to prevent the template analysis?

scripts/sct_analyze_lesion.py
```diff
@@ -550,6 +538,11 @@ def main(args=None):
     verbose = arguments.v
     sct.init_sct(log_level=verbose, update=True)  # Update log level
 
+    # TODO: CG: I would like to replace AnalyzeLesion by two classes:
+    #               - LesionVolume: which contains all the lesions, an ID assigned to each individual lesion, and
+    #                               and computing / gathering overall (ie across all lesions) metrics
+    #               - Lesion: which is dedicated to one individual lesion and computes / gathers metrics specific to
+    #                           this lesion.
```
charleygros, line 545: @jcohenadad: What do you think? :-)

scripts/sct_analyze_lesion.py
jcohenadad: how about a flag "-no-atlas"?
p.s. we are slowly moving towards more verbose flags (discussion for next SCT meeting)

jcohenadad: actually, after thinking about it, we could also make "-t" optional (which is already the case). When not declared, there is no atlas analysis. When declared, there is. I find it more intuitive.

scripts/sct_analyze_lesion.py
jcohenadad: it's hard to tell without seeing how these classes will interact with the `aggregate_metrics` module. I am also wondering if a class wouldn't be an overkill, given that its scope would be limited (only used here), and a dict could potentially achieve the same. It would be great to have @kousu input for overall design of the function

scripts/sct_analyze_lesion.py
charleygros: Sounds good!

scripts/sct_analyze_lesion.py
```diff
@@ -1,21 +1,17 @@
 #!/usr/bin/env python
-
-# Analyze lesions
+# -*- coding: utf-8
 #
-# Copyright (c) 2014 Polytechnique Montreal <www.neuro.polymtl.ca>
-# Author: Charley
-# Modified: 2017-08-19
+# Function to analyze lesions or tumours by computing statistics on binary masks.
```
jcohenadad, line 4:
```suggestion
# Function to analyze morphometrics of binary objects. These binary objects could represent lesions, tumours or any other pathology. 
```

scripts/sct_analyze_lesion.py
```diff
@@ -71,12 +68,12 @@ def get_parser():
         default=None,
         required=False)
     optional.add_argument(
-        "-f",
+        "-t",
```
jcohenadad, line 71: red flag: broken cross-compatibility. I've added the label on the PR.

scripts/sct_analyze_lesion.py
```diff
@@ -104,410 +101,123 @@ def get_parser():
...
-                self._orient(self.path_levels, 'RPI')
-                for fname_atlas in self.atlas_roi_lst:
-                    self._orient(fname_atlas, 'RPI')
-
-    def ifolder2tmp(self):
-        # copy input image
-        if self.fname_mask is not None:
-            sct.copy(self.fname_mask, self.tmp_dir)
-            self.fname_mask = ''.join(extract_fname(self.fname_mask)[1:])
+def check_binary(arr, fname, verbose=1):
```
jcohenadad, line 104: this one could potentially go under `spinalcordtoolbox.utils` or `spinalcordtoolbox.image` (or else?)
@kousu any suggestion?

scripts/sct_analyze_lesion.py
```diff
@@ -104,410 +101,123 @@ def get_parser():
...
+    :return: XX
+    """
+    im_mask, im_voi = Image(fname_mask), Image(fname_voi)
+
+    # Check if input data is binary and not empty
+    check_binary(im_mask.data, fname=fname_mask, verbose=verbose)
+    check_binary(im_voi.data, fname=fname_voi, verbose=verbose)
+
+    # re-orient image to RPI
+    logger.info("Reorient the image to RPI, if necessary...")
```
jcohenadad, line 170: as per [convention](https://github.com/neuropoly/spinalcordtoolbox/wiki/logging) we use printv in CLI and logger in APIs. However, i'm thinking maybe we should move `analyze_lesion` under `spinalcordtoolbox`, and maybe make it more "general", i.e., `analyze_binary_objects`. I see a lot or resemblance in terms of thematics with `spinalcordtoolbox.process_seg`, so maybe we should move it there?

charleygros:
@jcohenadad --> In order to mimic `sct_extract_metric`, what about including the following parameters:
- `-f` (instead of `-t`) --> like [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_extract_metric.py#L67) --> if specified then we compute statistics per tracts
- `-l` --> like [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_extract_metric.py#L74)
- `-vertfile` (instead of `-t`) --> like [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_extract_metric.py#L131) --> if specified then we compute statistics per levels

spinalcordtoolbox/process_seg.py
```diff
@@ -231,3 +233,127 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
     diameter_AP *= dim[0]
     diameter_RL *= dim[1]
     return diameter_AP, diameter_RL
+
+
+def get_angle_correction(im_seg):
```
jcohenadad, line 238: a very similar code is already in process_seg-- could you please try to see if we can replace the current code with your function here?

charleygros: Done :)

jcohenadad:
@charleygros while doing the refactoring it would be a good opportunity to remove this dependency: https://github.com/neuropoly/spinalcordtoolbox/pull/2739/commits/f8faaab439815ede36526b914e3d1512e71b8a84

joshuacwnewton:
For cleanup sake, I'm going to close this PR (but preserve the underlying branch).

The reason for this is:

- This PR has drifted too far from the `master` branch, and would require quite a bit of work to resolve the conflicts.
- This PR is not currently a priority for 6.3.
- There is already an open issue for refactoring the lesion/process_seg functionality, and so we can simply link to this PR from that issue, and say "use this branch as a starting point for the rewrite", if and when we eventually get to this PR:
    - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378

PR #2685, opened 2020-05-02
Fix argparse linewrap for R| strings

cfhammill:
This prevents lines being broken within words when using R| at the beginning of an argparse help string

Useful for #2658

### DIFF
--- a/spinalcordtoolbox/utils.py
+++ b/spinalcordtoolbox/utils.py
@@ -98,7 +98,7 @@ class SmartFormatter(argparse.HelpFormatter):
     # this is the RawTextHelpFormatter._split_lines
     def _split_lines(self, text, width):
         if text.startswith('R|'):
-            return text[2:].splitlines()
+            return [argparse._textwrap.fill(l, width) for l in text[2:].splitlines()]
         return argparse.HelpFormatter._split_lines(self, text, width)
 
 
--- a/spinalcordtoolbox/utils.py
+++ b/spinalcordtoolbox/utils.py
@@ -98,7 +98,20 @@ class SmartFormatter(argparse.HelpFormatter):
     # this is the RawTextHelpFormatter._split_lines
     def _split_lines(self, text, width):
         if text.startswith('R|'):
-            return [argparse._textwrap.fill(l, width) for l in text[2:].splitlines()]
+            lines = text[2:].splitlines()
+            offsets = [re.match("^[ \t]*", l).group(0) for l in lines]
+            wrapped = []
+            for i in range(len(lines)):
+                li = lines[i]
+                o = offsets[i]
+                ol = len(o)
+                init_wrap = argparse._textwrap.fill(li, width).splitlines()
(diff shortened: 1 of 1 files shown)


jcohenadad:
i tried de580da3e12bd3b9b85f71511f1f2329be03fe33 with `sct_create_mask` on a Terminal with 100 col, here is the output (notice the broken word):
~~~
  -p <str>              Process to generate mask.
                          <coord,XxY>: Center mask at the X,Y coordinates. (e.g. "coord,20x15")
                          <point,FILE>: Center mask at the X,Y coordinates of the label defined in i
nput volume FILE. (e.g. "point,label.nii.gz")
~~~

cfhammill:
hmm that's the output I get before the PR, but it's fixed in de580da3e for me. Could it be a mac thing?

jcohenadad:
>  Could it be a mac thing?

Yeah that's what i thought... let me try from within docker

jcohenadad:
i confirm this is likely a mac thing... in ubuntu (via docker), using the same terminal size, i get the proper word break:
~~~
  -p <str>              Process to generate mask.
                          <coord,XxY>: Center mask at the X,Y coordinates. (e.g. "coord,20x15")
                          <point,FILE>: Center mask at the X,Y coordinates of the label defined in
input volume FILE. (e.g. "point,label.nii.gz")
~~~

jcohenadad:
but regardless, it would be nice to have the text aligned with "<point,FILE>" (which is the case when we do not use "R|"). Do you think that would be possible without too much pain?

cfhammill:
I can think of a pretty rough solution. Just match any number of tabs and spaces in the previous line and insert it in all following lines. I'll see if it works.

cfhammill:
I think I have a working solution. Not sure what to do about the mac issue. Do we think it's using `_split_lines` on linux and `_fill_text` on linux? It could explain why `_fill_text` is more developed than `_split_lines` despite apparently not being used on Ubuntu 18.04

jcohenadad:
> Do we think it's using _split_lines on linux and _fill_text on mac? It could explain why _fill_text is more developed than _split_lines despite apparently not being used on Ubuntu 18.04

no idea...

alexfoias:
I tried sct_create_mask with terminal 2.10(433), osx 10.15.4 (19E287). And it seems to work well. 

Here are my screenshots: 
Screen Shot 2020-05-06 at 10 55 09 AM

Screen Shot 2020-05-06 at 10 55 56 AM

jcohenadad:
I just tested the updated 353485916fc7f4c0e024058fbf63019eeb8a56a3 on OSX 10.14 and I get the same output as @alexfoias [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2685#issuecomment-624699756), so it looks like your fix fixed it @cfhammill 🥇

cfhammill:
Great, thanks @alexfoias and @jcohenadad. This is ready for review/merge!

PR #2668, opened 2020-04-21
Use binaries that are packaged by the `spinalcordtoolbox-binaries` repo

kousu:
## Context

Currently, SCT depends on [8 total external C++ tools](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-How-to-compile-binaries-(ANTs,-isct)#c-binary-dependencies) that must be pre-built and packaged:

* 4/8: `ANTs` binaries we borrow from the external ANTs package.
* 2/8: `ctrDetect` binaries we borrow from an external contributor to SCT.
* 2/8: `isct_` binaries (from the `spinalcordtoolbox/dev` folder).

To make re-building the ANTs binaries easier/more reproducible, PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2821 automated the build process via the [`kousu/ANTs`](https://github.com/kousu/ANTs) repo. 

However, there was still a bit of work left to be done, as:

1. The ANTs-building workflow is stored in a personal repo, rather than a `spinalcordtoolbox`-affiliated repo.
2. Packaging and hosting the ANTs binaries (together with the `ctrDetect`/`isct_` binaries) still had to be done manually.
3. Windows is now natively supported as a platform, and this comes with its own set of ANTs binaries.

## Description

To address issues 1, 2, and 3, this PR updates `binaries_linux`, `binaries_osx`, and `binaries_windows` to use packages built online by:

* [`build_ANTs`](https://github.com/spinalcordtoolbox/build_ANTs) - [`build-ants.yml`](https://github.com/spinalcordtoolbox/build_ANTs/blob/master/.github/workflows/build-ants.yml): A workflow that builds ANTs, which then feeds into:
* [`spinalcordtoolbox-binaries`](https://github.com/spinalcordtoolbox/spinalcordtoolbox-binaries) - [`package-binaries.yml`](https://github.com/spinalcordtoolbox/spinalcordtoolbox-binaries/blob/master/.github/workflows/package-binaries.yml): A workflow that packages up all 8 binaries.

Additionally, the above CI workflows include the licenses for the external binaries (ANTs + ctrDetect), which addresses https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2665.

## Discussion

This makes Github our primary mirror (for these files), which is something we should talk about. On the plus side, it's easily integrated in an **automated** build pipeline, it's accessible in China. On the negative side, it's not-quite vendor lock-in and it breaks the symmetry in `sct_download_data`.

[...]

PR #2667, opened 2020-04-21
License binaries

kousu:
To partially address #2665, I manually gathered the relevant open source licenses ([ANTS](https://github.com/ANTsX/ANTs/blob/master/COPYING.txt), [ctrDetect](http://www.creatis.insa-lyon.fr/~sdika/soft/ctrDetect-v1_x86_64.tar.gz)) and included them in our binary distribution. This is awkward because it just dumps the licenses under `bin/`, which is not a good place for them, but at least they exist, and it helps me clarify what my planned automated packaging solution should look like.

```
$ tree copyright/
copyright/
├── LICENSE_ants.txt
├── LICENSE_ctrDetect.txt
└── LICENSE_opencv.txt

0 directories, 3 files
```

As in https://github.com/neuropoly/spinalcordtoolbox/pull/2642#issuecomment-616783109 and https://github.com/neuropoly/spinalcordtoolbox/pull/2666#issue-406442891, I unpacked the [previous](https://osf.io/admxy/?action=download) [versions](https://osf.io/7vx5g/?action=download), added these files, and repacked:

<details><summary>Build log</summary>

```
$ mkdir binaries_osx; tar -zxf 20200421_sct_binaries_osx.tar.gz -C binaries_osx
$ cp -r copyright/ binaries_osx/
$ tar --sort=name -czvf 20200422_sct_binaries_osx.tar.gz -C binaries_osx/ .
./
./copyright/
./copyright/LICENSE_ants.txt
./copyright/LICENSE_ctrDetect.txt
./copyright/LICENSE_opencv.txt
./isct_ComposeMultiTransform
./isct_antsApplyTransforms
./isct_antsRegistration
./isct_antsSliceRegularizedRegistration
./isct_dice_coefficient
./isct_propseg
./isct_spine_detect
./isct_train_svm
```

``` 
$ mkdir binaries_linux; tar -zxf 20200421_sct_binaries_linux.tar.gz -C binaries_linux
$ cp -r copyright/ binaries_linux/
$ tar --sort=name -czvf 20200422_sct_binaries_linux.tar.gz -C binaries_linux/ .
./
./copyright/
./copyright/LICENSE_ants.txt
./copyright/LICENSE_ctrDetect.txt
./copyright/LICENSE_opencv.txt
./isct_ComposeMultiTransform
./isct_antsApplyTransforms
./isct_antsRegistration
./isct_antsSliceRegularizedRegistration
./isct_dice_coefficient
./isct_propseg
./isct_spine_detect
[...]

PR #2666, opened 2020-04-21
Update ctrDetect Binaries

kousu:
Manually version bump ctrDetect to match [upstream](https://www.creatis.insa-lyon.fr/site7/en/ctrDetect)'s distribution.

This should be a minor change. It is in preparation for better automating the build which will download from M. Sdika's upstream directly.

<details><summary>How The Packages Were Built</summary>

```
$ curl -JLO https://osf.io/mka78/?action=download
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   459  100   459    0     0   1577      0 --:--:-- --:--:-- --:--:--  1577
100 74.6M  100 74.6M    0     0  2870k      0  0:00:26  0:00:26 --:--:-- 3842k
curl: Saved to filename '20200420_sct_binaries_linux.tar.gz'
$ curl -JLO https://osf.io/dn67h/?action=download 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   459  100   459    0     0   1645      0 --:--:-- --:--:-- --:--:--  1645
100 57.8M  100 57.8M    0     0  3473k      0  0:00:17  0:00:17 --:--:-- 3851k
curl: Saved to filename '20200420_sct_binaries_osx.tar.gz'
```

Follow https://github.com/neuropoly/spinalcordtoolbox/wiki/Compile-binaries -> https://www.creatis.insa-lyon.fr/site7/en/ctrDetect ->

```
$ curl -JLO http://www.creatis.insa-lyon.fr/~sdika/soft/ctrDetect-v1_x86_64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   279  100   279    0     0    690      0 --:--:-- --:--:-- --:--:--   692
100 33.9M  100 33.9M    0     0  1092k      0  0:00:31  0:00:31 --:--:-- 1131k
$ curl -JLO http://www.creatis.insa-lyon.fr/~sdika/soft/ctrDetect-v1_macos10.11.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   283  100   283    0     0    686      0 --:--:-- --:--:-- --:--:--   685
[...]

PR #2689, opened 2020-05-05
sct_download_data safety checks

kousu:
Some cleanups I found during #2668, but pulled out separately.

This **depends on** #2700.

### DIFF
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -53,8 +53,13 @@ def download_data(urls):
 
             # protect against directory traversal
             filename = os.path.basename(filename)
+            if not filename:
+                # this handles cases where you're loading something like an index page
+                # instead of a specific file. e.g. https://osf.io/ugscu/?action=view.
+                raise ValueError("Unable to determine target filename for URL: %s" % (url,))
 
             tmp_path = os.path.join(tempfile.mkdtemp(), filename)
+
             logger.info('Downloading: %s' % filename)
 
             with open(tmp_path, 'wb') as tmp_file:
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -8,6 +8,7 @@ import distutils.dir_util
 import logging
 import cgi
 import tempfile
+import urllib.parse
 import tarfile
 import zipfile
 import requests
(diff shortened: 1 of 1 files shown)


scripts/sct_dmri_moco.py
```diff
@@ -84,8 +84,7 @@ def get_parser():
                                   "poly [int]: Degree of polynomial function used for regularization along Z. For no regularization set to 0. Default=" + param_default.poly + ".\n"
                                                 "smooth [mm]: Smoothing kernel. Default=" + param_default.smooth + ".\n"
                                                   "metric {MI, MeanSquares, CC}: Metric used for registration. Default=" + param_default.metric + ".\n"
-                                                  "gradStep [float]: Searching step used by registration algorithm. The higher the more deformation allowed. Default=" + param_default.gradStep + ".\n"
-                                                    "sample [0-1]: Sampling rate used for registration metric. Default=" + param_default.sampling + ".\n",
+                                                  "gradStep [float]: Searching step used by registration algorithm. The higher the more deformation allowed. Default=" + param_default.gradStep + ".\n",
```
jcohenadad, line 87: out of PR scope?

kousu: This is because I did this off #2642. But I've changed my mind about that. I'm going to do it off #2700 instead.

kousu:
> some of the changes seem out of scope (e.g. moco)-- is that on purpose?

I've rebased onto #2700 instead of leaving this towering on my currently open stack of PRs (#2642, #2666 , #2667 , #2668). It's good to go now and can be merged before them.

spinalcordtoolbox/download.py
```diff
@@ -50,6 +50,9 @@ def download_data(urls):
                 _, content = cgi.parse_header(response.headers['Content-Disposition'])
                 filename = content["filename"]
 
+            # protect against directory traversal
+            filename = os.path.basename(filename)
```
kousu, line 55: Does this need to worry about symlinks?

jcohenadad: so far we don't have symlinks to download, so it should be fine?

kousu: This was to protect against [directory traversal](https://owasp.org/www-community/attacks/Path_Traversal).

My thought about symlinks is a related but obscure attack:

1. The mirror -- or someone who manages to impersonate it -- knows that you have a symlink in your download folder, say `keys -> ~/.ssh/authorized_keys`, 
2. They return "Content-Disposition: attachment; filename=keys" with the contents "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKYKsUxEseqswXVJ3LbmMq6eXkluXNdU6EszFUBx8LUN attacker@evil"
3. They can now ssh into your machine.

Issue #2675, opened 2020-04-24
Deprecate CentOS 6


kousu:
CentOS6 [is end-of-life](https://wiki.centos.org/About/Product) in November 2020.

We are currently maintaining ancient CentOS6 compiled dependencies

https://github.com/neuropoly/spinalcordtoolbox/blob/f0a6478d6b0a91209f7a76c1897f4fd1263ad70a/scripts/sct_download_data.py#L105-L106

but we don't test this very well. I also discovered in #2669 that ANTS won't even compile on CentOS 6, so people running on CentOS 6 are using subtly different maths than people on other OSes. We should plan to drop support by November and let everyone know they need to upgrade.

## TODO

* [ ] Remove `binaries_centos(6?)` from `sct_download_data`
* [ ] Update README
* [ ] Make sure this makes it into the changelog for our users.

kousu:
This might have already happened without us! https://api.travis-ci.org/v3/job/683313090/log.txt:

```
Check if tensorflow (1.5.0) is installed............[FAIL]

Traceback (most recent call last):

  File "/repo/python/envs/venv_sct/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow.py", line 58, in <module>

    from tensorflow.python.pywrap_tensorflow_internal import *

  File "/repo/python/envs/venv_sct/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 28, in <module>

    _pywrap_tensorflow_internal = swig_import_helper()

  File "/repo/python/envs/venv_sct/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 24, in swig_import_helper

    _mod = imp.load_module('_pywrap_tensorflow_internal', fp, pathname, description)

  File "/repo/python/envs/venv_sct/lib/python3.6/imp.py", line 243, in load_module

    return load_dynamic(name, filename, file)

  File "/repo/python/envs/venv_sct/lib/python3.6/imp.py", line 343, in load_dynamic

    return _load(spec)

ImportError: /lib64/libc.so.6: version `GLIBC_2.17' not found (required by /repo/python/envs/venv_sct/lib/python3.6/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so)
```

This also happens on old Ubuntu/Debian: #2523

jcohenadad:
Ok let's just drop centos6

jcohenadad:
[...]

Issue #2674, opened 2020-04-24
Speed up CI


kousu:
Locally, a complete test suite run is about 10 minutes. Doing this on Travis is much, much slower.

[![2020-04-24-152034_980x621_scrot](https://user-images.githubusercontent.com/987487/80249077-b4693e80-8660-11ea-9927-e8273afd0430.png)](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds)

Ideas:

1. It seems that running the Ubuntu builds are fast, generally running in parallel and taking about 12 minutes each (so about 12 minutes total), but the OS X builds are slow, taking about 20 minutes each, and running serially since they have to queue for an available OS X VM, including the extra time they spend blocked.
2. Since we're [using rebasing](https://github.com/neuropoly/spinalcordtoolbox/blob/master/CONTRIBUTING.rst#commit-sequences), consider disabling ["Build pushed pull requests"](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/settings), because it causes [doubled building](https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests):
  
  ![2020-04-24-151835_693x270_scrot](https://user-images.githubusercontent.com/987487/80249197-eed2db80-8660-11ea-9c5a-35d684432dd5.png)
  
  The difference between "travis-ci/push" and "travis-ci/pr" is that the latter makes a merge commit with master -- but if we're always rebasing we shouldn't have merges so building the PR's branch is always enough to test that it's properly integrated with `master`. This might not speed up the overall build time since Travis runs the two as isolated build jobs, but I am hoping it will make the VMs more available to us, particularly the OS X VMs.

kousu:
I turned off `travis-ci/pr` and builds have been faster since then.

Drulex:
Hey I noticed that CI still runs on push and PR (example https://github.com/neuropoly/spinalcordtoolbox/pull/2781). Is this desired?

kousu:
@drulex thanks for noticing! I turned that back on to try to get https://github.com/neuropoly/spinalcordtoolbox/pull/2751 merged because it turned out that disabling "PR" means that no branches from forks can get merged. I don't see how to balance these needs. I hope there's something I've missed or misunderstood.

[...]

PR #2690, opened 2020-05-05
Support Travis testing arbitrary distros via Docker

### LINKED ISSUE #2313: Add CentOS7 build on Travis
Useful links:
- https://travis-ci.com/saltstack-formulas/postgres-formula/jobs/197417111/config
- https://github.com/brianmario/mysql2/blob/master/.travis.yml
  - Dockerfile: https://github.com/brianmario/mysql2/blob/master/.travis_Dockerfile_centos

Also see NeuroPoly Slack #random (internal)

kousu:
This adds Debian, CentOS, and ArchLinux to Travis, and it is easy to add other Linux distros thanks to Docker.

I could not decide on a better way than to [take Travis' advice](https://docs.travis-ci.com/user/job-lifecycle/#complex-build-commands) and move the entire test script to a separate shell script because we depend deeply on being able to pass environment variables from one step to the next in order to set up our venv (for `bin/` and for `conda`), and that just doesn't work with a series of `docker exec`s. This means losing the nice collapsing of sections that Travis does, but I don't think that's a big deal; I usually look at the raw plaintext logs anyway.

This adds only minimal time to CI because each build runs in mostly parallel. I've seen Travis running 6 at a time. I predict on average it adds 10 minutes to the time.

~CentOS6 is failing because of [Tensorflow](https://github.com/neuropoly/spinalcordtoolbox/issues/2675#issuecomment-623978583) and because of [antsSliceRegularizedRegistration](https://github.com/neuropoly/spinalcordtoolbox/issues/2600). I did not fix the latter in #2642, it was too hard for me, and the former is out of our purview.~ **Depends on** #2699, which removes CentOS6.

## TODO

* [x] Decide if we will deprecate CentOS6 ( #2675 ) now, or try to fix the issues to keep it alive, or just comment out the CentOS6 case and leave it untested. => yes, drop it
* [x] Drop centos6
* [x] Test script cleanups:
  * [x] There's no need to edit PATH manually; it's only for accessing `sct_*` and those are available by `conda activate`.
  * [x] The splitbrain `if [ "${TRAVIS_OS_NAME}" = "osx" ]; then ... -j 1 ...` can be simplified
* ~~Mount the docker volume at the *same* `pwd` as the outer OS, to improve consistency between the environments~~ this isn't going to work: 15fbb00
  * ~~done in https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-2 but not yet merged because I want to see what Travis says about individual steps~~

### Optional/future TODOs

* [x] Pull the deep docker magic into its own wrapper shell script: https://github.com/neuropoly/spinalcordtoolbox/pull/2698#discussion_r422447848
  * done in https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-3 but not yet merged back here because I want to see what Travis says about it
* [x] Test if we can drop the `-j 1` restriction on OS X entirely
  * testing in https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-testosx
* [x] Add more distros
  * [x] Debian `sid`
  * [x] Debian `testing`
  * [x] Debian 10
  * [x] Debian 9
  * [x] Debian 8
    * done in https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-4; waiting on CI to see if it's worth merging here
  * ~~NixOS~~
* [ ] Detect and pass relevant environment variables into Docker
  - Docker is usually meant to be an isolated environment so it inherits no envs by default; but we probably want to at least pass through all the TRAVIS_ envs.
*  ~~Move pylint to an entirely separate travis job~~
  - ~~this should speed up the build; linting takes about 30s to a minute each run; no need to duplicate the work~~
  - I'm planning to do this on the same PR that adds PEP8 (https://github.com/neuropoly/spinalcordtoolbox/issues/1538)
* ~~Verify that `coverage` is actually producing useful information.~~ @ #2702.

Fixes #2313.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+# CI testing script
+# This is meant to be called from .travis.yml -- it works around limitations
+
+set -e # Error build immediately if install script exits with non-zero
+
+# if this is a docker job, set up and then recurse into the container,
+# instead of continuing.
+# TODO: figure out and pass needed environment variables individually with -e
+
+if ! [ -z "$IMAGE" ]; then
+   docker run \
+     --name container \
+     --init \
+     -it -d \
+     --rm \
+     -v `pwd`:/repo -w /repo \
+     "$IMAGE"
+   trap "docker stop container" EXIT
+   # set up groups to match the owner of /repo, so the installer can't tell
+   docker exec container groupadd -g "`id -g`" `id -g -n`
+   docker exec container useradd -m -u "`id -u`" -g "`id -g`" `id -u -n`
+   docker exec container $DEPS    # install platform-specific dependencies
+   # recurse to run the real test script
+   # --user `id -u` makes sure the build script is the owner of the files at /repo
+   docker exec --user `id -u`:`id -g` container "$0"
+   exit 0
+fi
+
+
+echo Installing SCT
+yes | ASK_REPORT_QUESTION=false ./install_sct
+echo $?
+echo "... STATUS"
+PATH="$PATH:$PWD/bin"
+
+echo *** CHECK PATH ***
+echo $PATH  # Make sure PATH includes sct/bin folder
+ls -lA bin  # Make sure all binaries and aliases are there
+sct_download_data -d sct_testing_data  # for tests
+source python/etc/profile.d/conda.sh  # to be able to call conda
+conda activate venv_sct  # reactivate conda for the pip install below
+
+echo *** UNIT TESTS ***
+pytest
+
+echo *** INTEGRATION TESTS ***
+pip install coverage
+echo -ne "import coverage\ncov = coverage.process_startup()\n" > sitecustomize.py
+echo -ne "[run]\nconcurrency = multiprocessing\nparallel = True\n" > .coveragerc
+if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
+  COVERAGE_PROCESS_START="$PWD/.coveragerc" COVERAGE_FILE="$PWD/.coverage" sct_testing -j 1 --abort-on-failure
+  coverage combine
+else
+  COVERAGE_PROCESS_START="$PWD/.coveragerc" COVERAGE_FILE="$PWD/.coverage" sct_testing --abort-on-failure
+  coverage combine
+fi
+
+# TODO: move this part to a separate travis job; there's no need for each platform to lint the code
+echo *** ANALYZE CODE ***
+pip install pylint
+bash -c 'PYTHONPATH="$PWD/scripts:$PWD" pylint -j3 --py3k --output-format=parseable --errors-only $(git ls-tree --name-only -r HEAD | sort | grep -E "(spinalcordtoolbox|scripts|testing).*\.py" | xargs); exit $(((($?&3))!=0))'
+
+#
+# echo *** BUILD DOCUMENTATION ***
+# pip install sphinx sphinxcontrib.programoutput sphinx_rtd_theme
+# cd documentation/sphinx
+# make html
+# cd -
+
+# python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
+# cd ../spinalcordtoolbox_v*
+# yes | ./install_sct  # test installation of package
+
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,76 +1,18 @@
 #!/bin/bash
-# CI testing script
-# This is meant to be called from .travis.yml -- it works around limitations
+# TravisCI testing harness.
+#  Supports running locally (i.e. on whatever platform Travis has loaded us in)
+#  or in a docker container specified by $DOCKER_IMAGE.
+#
+# usage: .travis.sh
+#
+# e.g. DOCKER_IMAGE="centos:8" .travis.sh
 
 set -e # Error build immediately if install script exits with non-zero
 
-# if this is a docker job, set up and recurse into the container,
-# instead of continuing.
+# if this is a docker job, run in the container instead; but if not just run it here.
 if [ -n "$DOCKER_IMAGE" ]; then
-   docker run \
-     --name container \
-     --init \
-     -it -d \
-     --rm \
-     -v "`pwd`":/repo -w /repo \
-     "$DOCKER_IMAGE"
-   trap "docker stop container" EXIT
-   # set up a user:group matching that of the volume mount /repo, so the installer isn't confused
-   #
-   # TODO: it would be nice if the volume was mounted at `pwd`/, to further reduce the distinction
-   # between docker/nondocker runs, but docker gets the permissions wrong:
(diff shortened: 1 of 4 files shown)


kousu:
[Screenshot_2020-05-05_07-35-26](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/683324425)

.travis.sh
```diff
@@ -0,0 +1,74 @@
...
+conda activate venv_sct  # reactivate conda for the pip install below
+
+echo *** UNIT TESTS ***
+pytest
+
+echo *** INTEGRATION TESTS ***
+pip install coverage
+echo -ne "import coverage\ncov = coverage.process_startup()\n" > sitecustomize.py
+echo -ne "[run]\nconcurrency = multiprocessing\nparallel = True\n" > .coveragerc
+if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
```
zougloub, line 51: Added benefit of the `.sh` is that we could run it locally.
Instead of passing the variable name is it possible to use `$(uname -n) = Darwin` here?

.travis.sh
```diff
@@ -0,0 +1,74 @@
...
+# CI testing script
+# This is meant to be called from .travis.yml -- it works around limitations
+
+set -e # Error build immediately if install script exits with non-zero
+
+# if this is a docker job, set up and then recurse into the container,
+# instead of continuing.
+# TODO: figure out and pass needed environment variables individually with -e
+
+if ! [ -z "$IMAGE" ]; then
```
zougloub, line 11: For a bit more clarity maybe use better names than IMAGE and DEPS?
eg. CONTAINER_IMAGE / CONTAINER_INSTALL_DEPS_CMDLINE

Also `! [ -z ]` could be `[ -n ]` but that's nothing.

.travis.sh
zougloub: (`-s` sorry)

kousu: Genius! Yes that sidesteps the problem of passing in the travis envs to Docker -- the less dependent on specific platform features the better IMO.

.travis.sh
kousu: Yes and yes. I like both of those. I'll fix them up.

kousu: Yep, I'll do this!

kousu:
Getting closer:

Screenshot_2020-05-09 Build #13080 - neuropoly spinalcordtoolbox - Travis CI

I'm still waiting on 
* https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-testosx
* https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-2
* https://github.com/neuropoly/spinalcordtoolbox/tree/ko/CI-centos7-3

to see if I can merge those into this PR, but if they pass this is ready for review (pending #2699 being reviewed first).

kousu:
The `if OSX -> sct_testing -j 1` It looks like it was first added in 9f6d2d2c817fde3a692a94d8403dd9c2ecf0ba2e; @zougloub do you remember why? Maybe the OS X build was unstable then?

It looks like it's unnecessary now: https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/685165536

Screenshot_2020-05-09 Job #13081 7 - neuropoly spinalcordtoolbox - Travis CI

zougloub:
I don't remember (sorry for not leaving a breadcrumb).

kousu:
This tests a lot of cases now! It's perhaps a bit excessive.

kousu:
> > This tests a lot of cases now! It's perhaps a bit excessive.
> 
> waiting ~1h after each push might indeed be a bit excessive and slow down our quick PRs. Can we keep builds short for regular maintenance (e.g. the latest Ubuntu and OSX), and have this exhaustive build for when we plan new releases? I am wondering how this could be automatized in our current release pipeline.

The build times are already over an hour -- and sometimes double that before I did some steps in #2674.

<details><summary>(some examples)</summary>

[TODO]

</details>

Even with all the extra linux VMs, my run only ran for 1h in total -- *faster* than the 1h30 that several of the examples from this week too.

The slowest part of Travis is not waiting for Linux VMs to boot up; it's waiting for OS X VMs to become available. The best thing we could do for build times is dropping Sierra and High Sierra, which I think is more reasonable because Sierra is end-of-life already and High Sierra is end-of-life in November.

But yes, we can set options in `.travis.yml` - https://docs.travis-ci.com/user/conditional-builds-stages-jobs/, e.g.

> `if: branch = master`

but if we do that then we don't know what PR introduced a platform-specific bug until it's already on `master` and potentially already filtered out into other branches.

Maybe we could do:

> `if: branch = release-candidate`

?

jcohenadad:
Thanks for the clarifications. I like the `if: branch = release-candidate` approach. As described in the [wiki](https://github.com/neuropoly/spinalcordtoolbox/wiki/versioning), we do have a single release branch that could be tested with all builds. But if we prefer to catch the issue *before* pushing to the release branch (to keep it clean), we could maybe keep your suggestion and update the documentation so that the intermediate release branch (the one called `jca/X.Y.Z` in the doc) should always be `release-candidate`. Does that make sense?

kousu:
What's

> Thanks for the clarifications. I like the `if: branch = release-candidate` approach. As described in the [wiki](https://github.com/neuropoly/spinalcordtoolbox/wiki/versioning), we do have a single release branch that could be tested with all builds. But if we prefer to catch the issue _before_ pushing to the release branch (to keep it clean), we could maybe keep your suggestion and update the documentation so that the intermediate release branch (the one called `jca/X.Y.Z` in the doc) should always be `release-candidate`. Does that make sense?

So, the plan is:

- quick tests on every branch
- slow tests on every release

In that case, do we even need the `release-candidate`/`jca/X.Y.Z` branch? I would just use `release` as both the the release candidate and final release (it fits into the [release line from this model](https://nvie.com/posts/a-successful-git-branching-model/)) -- it graduates from release-candidate to release at the point you ask github to make a release out of it.

jcohenadad:
> In that case, do we even need the release-candidate/jca/X.Y.Z branch? I would just use release as both the the release candidate and final release (it fits into the release line from this model) -- it graduates from release-candidate to release at the point you ask github to make a release out of it.

Indeed, that makes sense. Pls feel free to update the wiki.

kousu:
Here's what I mean about Travis' OS X machine: even reducing to just three builds, the two linuxes ran in parallel and finished before OS X even started:

Screenshot_2020-05-11 Build #13118 - neuropoly sspinalcordtoolbox - Travis CI

kousu:
5 minutes later: still waiting for an OS X slot

Screenshot_2020-05-11 Build #13118 - neuropoly spinalcordtoolbox - Travis CI

kousu:
30 minutes in all:

Screenshot_2020-05-11 Build #13118 - neuropoly spinalcordtoolbox - Travis CI

PR #2694, opened 2020-05-06
sagittal dialog: fixes error when trying to access out of bound slice

### LINKED ISSUE #2576: Manual C2-C3 identification : out of bound error
p.56 -> 

When navigating the slices (left/right arrows) to identify manually C2-C3, the viewer crashes with an out of bound error :

```
sct_label_utils -i t2.nii.gz -create-viewer 3 -o label_c2c3.nii.gz -msg "Click at the posterior tip of C2/C3 inter-vertebral disc"

--
Spinal Cord Toolbox (git-master-846d5d1bf50fe37d602b99b142a0c3032492e566)

  File "/home/olix86/sct/spinalcordtoolbox/gui/sagittal.py", line 89, in increment_vertical_nav
    self.sagittal.refresh()
  File "/home/olix86/sct/spinalcordtoolbox/gui/widgets.py", line 296, in refresh
    data = self._image.data[:, :, self._z]
IndexError: index 64 is out of bounds for axis 2 with size 64
Aborted (core dumped)
```

saulo-p:
Avoid application error trying to access of out of bounds image slice.
In the event of a slice increment/decrement from the UI, check boundaries before trying to access the new slice. If the index change goes outside the boundary, loop around to the valid slice index at the other end.

Fixes #2576


jcohenadad:
@saulo-p if you could label according to convention it would be great: https://github.com/neuropoly/spinalcordtoolbox/blob/master/CONTRIBUTING.rst#pr-labels
i've added "bug" already

PR #2698, opened 2020-05-09
CI for WSL (2691)

### LINKED ISSUE #2691: CI for Windows Subsystem for Linux
Following #2313 and given that @cfhammill has been working on getting us fully working on Windows 10 -- via WSL -- we should add WSL to Travis.

cfhammill:
WIP PR for testing SCT on WSL via travis.

So far failing, got to the point of having and being able to run the ubuntu binary, however failing to attach the tty to the main travis git-bash session. 

When done, will fix #2691

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -14,6 +14,14 @@ if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
 elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
     choco install wsl-ubuntu-1804
+
+    # disable apt's helpful (and build-breaking) interactive mode
+    # https://linuxhint.com/debian_frontend_noninteractive/
+    export DEBIAN_FRONTEND="noninteractive"
+    # Use WSLENV to actually pass it into WSL
+    # https://devblogs.microsoft.com/commandline/share-environment-vars-between-wsl-and-windows/
+    export WSLENV=DEBIAN_FRONTEND
+
     wsl apt-get update
     #wsl apt-get -y upgrade  # this step is probably important, but it's also sooo slow
     wsl apt-get install -y gcc git curl
--- a/.travis.sh
+++ b/.travis.sh
@@ -12,7 +12,9 @@ set -e # Error build immediately if install script exits with non-zero
 # if this is a docker job, run in the container instead; but if not just run it here.
 if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
+elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
+    choco install wsl-ubuntu-1804
+    /c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe run ./.ci.sh
 else
     ./.ci.sh
 fi
-
--- a/.travis.sh
+++ b/.travis.sh
@@ -14,7 +14,10 @@ if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
 elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
     choco install wsl-ubuntu-1804
-    /c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe run ./.ci.sh
+    wsl apt-get update
+    #wsl apt-get -y upgrade  # this step is probably important, but it's also sooo slow
+    wsl apt-get install -y gcc git curl
+    wsl ./.ci.sh
 else
     ./.ci.sh
 fi
--- a/.travis.sh
+++ b/.travis.sh
@@ -14,6 +14,13 @@ if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
 elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
     choco install wsl-ubuntu-1804
+     # or, instead of choco, use curl + powershell:
+     # https://docs.microsoft.com/en-us/windows/wsl/install-manual#downloading-distros-via-the-command-line
+    # wsl --setdefault "Ubuntu-18.04"
+     # TODO: Travis's version of wsl is too old for --setdefault.
+     # Instead we trust that wsl will default to the installed
+     # Ubuntu because it is the only option, but it would be
+     # better to be explicit when it becomes possible.
 
     # disable apt's helpful (and build-breaking) interactive mode
     # https://linuxhint.com/debian_frontend_noninteractive/
--- a/.travis.sh
+++ b/.travis.sh
@@ -13,7 +13,7 @@ set -e # Error build immediately if install script exits with non-zero
 if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
 elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
-    choco install wsl-ubuntu-1804
+    choco install wsl-ubuntu-1804 -y
      # or, instead of choco, use curl + powershell:
      # https://docs.microsoft.com/en-us/windows/wsl/install-manual#downloading-distros-via-the-command-line
     # wsl --setdefault "Ubuntu-18.04"
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,7 +116,7 @@ matrix:
       if: branch = release or type = cron
       os: osx
       osx_image: xcode9.2
-    - name: "Windows 10"
+    - name: "Windows Server, 1809"
       language: bash
       os: windows
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ notifications:
     on_success:change
     on_failure:always
 
+git:
+  autocrlf: false
+
 # this enables to avoid recompilation of dipy if it was already compiled previously
 #cache:
 #  directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,5 +113,8 @@ matrix:
       if: branch = release or type = cron
       os: osx
       osx_image: xcode9.2
+    - name: "Windows 10"
+      language: bash
+      os: windows
 
 script: ./.travis.sh
--- a/install_sct
+++ b/install_sct
@@ -457,7 +457,6 @@ else
 fi
 
 # Update PATH variables based on Shell type
-UPDATE_PATH="export PATH=$SCT_DIR/$BIN_DIR:$PATH"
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -15,28 +15,34 @@ notifications:
...
+    - os: windows
+      language: bash
 
 install:
   - |
     echo Installing SCT
+    if [ "${TRAVIS_OS_NAME}" = "windows" ]; then
+      choco install wsl-ubuntu-1804
+      sudo /c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe
+    fi
```
kousu, line 45: This is much like the issue I had in #2690: you need to launch a sub-environment and continue the script from there.

I solved it by moving the entire Travis script into a separate file and then running that instead: https://github.com/neuropoly/spinalcordtoolbox/blob/4225bd6652707c077e4af2c44678c53d862c6fbf/.travis.sh

It does the important reboot-into-a-different-OS step with:

https://github.com/neuropoly/spinalcordtoolbox/blob/4225bd6652707c077e4af2c44678c53d862c6fbf/.travis.sh#L26-L27

kousu: I suppose for integrating both WSL-containers and Docker-containers we need the test script to not have that header that assumes docker. Maybe we want this:

```
     if [ "${TRAVIS_OS_NAME}" = "windows" ]; then 
         ./wsl.sh .travis.sh
     elif [ -n "$DOCKER_IMAGE" ]; then
        ./docker.sh .travis.sh
     else
        .travis.sh
      fi
```

With 

```
   # docker.sh
   docker run \
     --name container \
     --init \
     -it -d \
     --rm \
     -v `pwd`:/repo -w /repo \
     "$DOCKER_IMAGE"
   trap "docker stop container" EXIT
   # set up groups to match the owner of /repo, so the installer can't tell
   docker exec container groupadd -g "`id -g`" `id -g -n`
   docker exec container useradd -m -u "`id -u`" -g "`id -g`" `id -u -n`
   docker exec container $DOCKER_DEPS    # install platform-specific dependencies
   # recurse to run the real test script
   # --user `id -u` makes sure the build script is the owner of the files at /repo
   docker exec --user `id -u`:`id -g` container "$1"
   exit 0
```

and

```
      # wsl.sh
      choco install wsl-ubuntu-1804
      sudo /c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe "$1"
```

cfhammill:
Awesome, thanks @kousu, I'll give your suggestion a shot when I next get a chance.

cfhammill:
After some fussing, this is building now, test report should be done in a few minutes, but at least the goal of getting CI to build SCT on WSL has been accomplished.

@jcohenadad and @kousu, should WSL enter the main set that gets tested every PR or be in the release only suite? I (not very strongly) think it should be in the main set depending on build time. I suspect that at least a reasonable number of users are on windows.

jcohenadad:
Fantastic! Indeed, if the build time is reasonable (<15min) i would leave this travis test in the main set. It seems to be broken now, thoughts?

.travis.sh
```diff
@@ -12,7 +12,12 @@ set -e # Error build immediately if install script exits with non-zero
 # if this is a docker job, run in the container instead; but if not just run it here.
 if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
+elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
+    choco install wsl-ubuntu-1804
+    brun="/c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe run"
+    $brun sudo apt update
+    $brun sudo apt install -y gcc git curl
+    $brun ./.ci.sh
```
kousu, line 20: This is great.

.travis.sh
```diff
@@ -12,7 +12,12 @@ set -e # Error build immediately if install script exits with non-zero
 # if this is a docker job, run in the container instead; but if not just run it here.
 if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
+elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
+    choco install wsl-ubuntu-1804
+    brun="/c/ProgramData/chocolatey/lib/wsl-ubuntu-1804/tools/unzipped/ubuntu1804.exe run"
+    $brun sudo apt update
+    $brun sh -c 'DEBIAN_FRONTEND="noninteractive" sudo apt install -y gcc git curl'
```
kousu, line 19: Because of

> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

```suggestion
    $brun sudo apt-get update
    $brun sh -c 'DEBIAN_FRONTEND="noninteractive" sudo apt-get install -y gcc git curl'
```

kousu:
I think this is using WSL v1, but Microsoft is [pushing out v2 already](https://devblogs.microsoft.com/commandline/wsl2-will-be-generally-available-in-windows-10-version-2004/).

> After some fussing, this is building now, test report should be done in a few minutes, but at least the goal of getting CI to build SCT on WSL has been accomplished.
> 
> @jcohenadad and @kousu, should WSL enter the main set that gets tested every PR or be in the release only suite? I (not very strongly) think it should be in the main set depending on build time. I suspect that at least a reasonable number of users are on windows.

I think it should be in the main set. Windows is still the dominant desktop OS and what most of our users are probably using on a daily basis. And few of us are working on it on a daily basis. Making sure it's tested thoroughly is a good idea.

kousu:
I think this is using WSL v1, but Microsoft is [pushing out v2 already](https://devblogs.microsoft.com/commandline/wsl2-will-be-generally-available-in-windows-10-version-2004/). Something to watch out for.

.travis.yml
```diff
@@ -24,94 +27,97 @@ matrix:
...
+    # - name: "OSX 10.13 (High Sierra)"
+    #   if: branch = release or type = cron
+    #   os: osx
+    #   osx_image: xcode9.4
+    # - name: "OSX 10.12 (Sierra)"
+    #   if: branch = release or type = cron
+    #   os: osx
+    #   osx_image: xcode9.2
+    - name: "Windows 10"
+      language: bash
```
kousu, line 120: This line isn't necessary.

https://docs.travis-ci.com/user/reference/windows/:

> Git BASH is the shell that’s used to run your build. See Git for Windows for more details.

```suggestion
```

.travis.yml
```diff
@@ -24,94 +27,97 @@ matrix:
...
+    #   osx_image: xcode11.2
+    # - name: "OSX 10.13 (High Sierra)"
+    #   if: branch = release or type = cron
+    #   os: osx
+    #   osx_image: xcode9.4
+    # - name: "OSX 10.12 (Sierra)"
+    #   if: branch = release or type = cron
+    #   os: osx
+    #   osx_image: xcode9.2
+    - name: "Windows 10"
```
kousu, line 119: Is this true? The docs [say](https://docs.travis-ci.com/user/reference/windows/#windows-version)

>  Only Windows Server, version 1809 is currently supported.

Windows is currently at 1909, and will soon be at 2004.

I guess it's true.

.travis.sh
```diff
@@ -12,7 +12,12 @@ set -e # Error build immediately if install script exits with non-zero
 # if this is a docker job, run in the container instead; but if not just run it here.
 if [ -n "$DOCKER_IMAGE" ]; then
     ./util/dockerize.sh ./.ci.sh
+elif [ "${TRAVIS_OS_NAME}" = "windows" ]; then
+    choco install wsl-ubuntu-1804
```
kousu, line 16: This seems to install Ubuntu in a way that can only be accessed by Administrators. Presumably none of our users will be using WSL this way. This subtle distinction might bite us.

kousu, line 16: This probably needs

```suggestion
    choco install -y wsl-ubuntu-1804
```

to be reliable. otherwise it might hang waiting for user input.

.travis.yml
```diff
@@ -6,6 +6,9 @@ notifications:
     on_success:change
     on_failure:always
 
+git:
+  autocrlf: false
+
```
kousu, line 11: Ah!  Important catch. :rocket:

.travis.sh
kousu, line 20: We need to demonstrate that this actually errors out if there are bugs. When I was working on #2718 I had at least one case where there were failing tests but CI passed. Someone (me or you) needs to demonstrate with screenshots:

- a failed pytest test
- a `exit 9` in an obscure place in .ci.sh
- a failed integration test

kousu:
@cfhammill I sorted out the main problem with CI in #2718, which is a PR into this branch. Please take a look when you have a moment :balloon:

.travis.yml
cfhammill: It is not, the windows builds only support bash and it defaults to ruby. I suppose language could be set to python but bash seemed easiest because our build script is bash.

.travis.yml
cfhammill: Good catch!

.travis.sh
cfhammill: There's an answer down below on how to do it, I suspect it's not critical, but if you'd like to investigate, go for it.

cfhammill: Good idea!

.travis.sh
cfhammill: There's some evidence that at least using the ubuntu binary errors properly on a test fail. https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/688649596

cfhammill:
> 
> 
> Fantastic! Indeed, if the build time is reasonable (<15min) i would leave this travis test in the main set. It seems to be broken now, thoughts?

First successful build came in at around 17 minutes https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/689000830. Not sure if there is a way we can trim that down.

jcohenadad:
let's say 17min is reasonable 😅

kousu:
The version I ran (on #2718 ) also was 18 minutes: https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/688763385.

Linux and macOS have been taking 10-12 minutes lately. There's probably some overhead in using WSL, since, like Wine, it has to translate the Linux API to the Windows API. Installing Ubuntu on top of Windows is going to take a bit of time -- it's 200MB. And maybe the Travis Windows machines are just slower -- it is a new feature for them after all, they probably haven't allocated their fastest machines to it.

It think it's reasonable.

.travis.yml
kousu: Alright!

cfhammill:
Ok, main set it is. Reactivating the entire test suite, bringing in master, and removing draft status.

PR #2700, opened 2020-05-10
Moved sct_download_data functions to new download module

jcohenadad:
While working on https://github.com/neuropoly/spinalcordtoolbox/pull/2639, I had to make `sct_download_data`'s subfunctions callable by other Python processes, so the purpose of this PR was to move those functions (previously located in the CLI script) into a new `spinalcordtoolbox.download` module. 

This PR also introduces the following changes to `sct.utils`:
- added: `add_suffix` and `splitext` (formerly `extract_fname`)
- created unit tests for these

### DIFF
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -12,19 +12,10 @@
 
 from __future__ import absolute_import
 
-import cgi
 import os
 import sys
-import tarfile
-import tempfile
-import zipfile
-from shutil import rmtree
-from distutils.dir_util import copy_tree
 
-import requests
-from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util import Retry
-from tqdm import tqdm
+from spinalcordtoolbox.download import DICT_URL, install_data
 
 from msct_parser import Parser
 import sct_utils as sct
@@ -38,6 +29,7 @@ def get_parser():
         type_value="multiple_choice",
         description="Name of the dataset.",
         mandatory=True,
+        # TODO: replace with key dict items
         example=[
             'sct_example_data',
             'sct_testing_data',
@@ -51,6 +43,7 @@ def get_parser():
             'optic_models',
             'pmj_models',
             'binaries_debian',
+            'binaries_centos',
             'binaries_osx',
             'deepseg_gm_models',
             'deepseg_sc_models',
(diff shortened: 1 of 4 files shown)


scripts/sct_download_data.py
```diff
@@ -51,6 +43,7 @@ def get_parser():
             'optic_models',
             'pmj_models',
             'binaries_debian',
+            'binaries_centos',
```
kousu, line 46: This was gone in #2699.

scripts/sct_download_data.py
```diff
@@ -38,6 +29,7 @@ def get_parser():
         type_value="multiple_choice",
         description="Name of the dataset.",
         mandatory=True,
+        # TODO: replace with key dict items
```
kousu, line 32: Are you thinking of loading the docstring directly from `DICT_URL`? `example=sorted(DICT_URL)`?

spinalcordtoolbox/utils.py
```diff
@@ -138,83 +156,19 @@ def is_exe(fpath):
...
-    sct_branch = os.getenv(branch_env, "unknown")
-    if check_exe("git") and os.path.isdir(os.path.join(__sct_dir__, ".git")):
-        sct_commit = __get_commit() or sct_commit
-        sct_branch = __get_branch() or sct_branch
-
-    if sct_commit is not 'unknown':
-        install_type = 'git'
+    parent, filename = os.path.split(fpath)
+    if filename.endswith(".nii.gz"):
+        stem, ext = filename[:-7], ".nii.gz"
```
kousu, line 168: You could add the same special-case for `.tar.gz`. The number of special cases in here might fast get unwieldy.

spinalcordtoolbox/utils.py
```diff
@@ -138,83 +156,19 @@ def is_exe(fpath):
     return None
 
 
-def __get_branch():
+def extract_fname(fpath):
```
kousu, line 159: I would just make this

```suggestion
def splitext(filename):
```

If users need the folder path, they already have `os.path.split()`. `sct.utils.splitext()` can be a drop in neuroimaging-aware replacement for `os.path.splitext()`.

spinalcordtoolbox/utils.py
```diff
@@ -115,6 +116,23 @@ def _split_lines(self, text, width):
...
+    :param suffix: suffix. Example: _mean
+    :return: file name with suffix. Example: t2_mean.nii
+
+    Examples:
+
+    - add_suffix(t2.nii, _mean) -> t2_mean.nii
+    - add_suffix(t2.nii.gz, a) -> t2a.nii.gz
+    """
+    parent, stem, ext = extract_fname(fname)
+    return os.path.join(parent, stem + suffix + ext)
```
kousu, line 133: You don't need to separate the parent:

```
>>> base, ext = os.path.splitext("path/to/dataset/t2.nii.gz")
>>> base + '_mean' + ext
'path/to/dataset/t2.nii_mean.gz'
```

If that was `sct.utils.splitext()` instead that was `.nii.gz`-aware that would have produced the right result.

spinalcordtoolbox/utils.py
kousu: It almost looks like this exists in the stdlib @ https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffixes but it only does half the job. :abacus:

spinalcordtoolbox/utils.py
kousu: It handles `.`s in the parent dirs properly:

```
>>> base, ext = os.path.splitext("path/to.tar.gz/dataset/t2.tar.gz")
>>> base + "_mean" + ext
'path/to.tar.gz/dataset/t2.tar_mean.gz'
>>> 
```

spinalcordtoolbox/utils.py
kousu: What about

```suggestion
    dir, filename = os.path.split(filename)
    stem, ext = filename.split(".",1)
    ext = "." + ext
    return os.path.join(dir, stem), ext
```

scripts/sct_download_data.py
jcohenadad: fixed in 190acb6bb392c179a7e40e2ea7e5e6f58cfd10b1

scripts/sct_download_data.py
jcohenadad: yup! fixed in 9d531465638a9b7299f52526eb7b2ba36b0cc5b3

spinalcordtoolbox/utils.py
jcohenadad: the problem with that implementation is that it fails the following test:
```python
    assert utils.splitext('nice.image.nii.gz') == ('nice.image', '.nii.gz')
```
I've seen such file names in my life 🙄...

jcohenadad: now accommodating tar.gz (and we can add to the list of extensions if we encounter more): f537788664a3fb81876235d67f92ef910f1bd6f6

spinalcordtoolbox/utils.py
jcohenadad: great suggestion. fixed in ea0c10616049b8a433c2e7ebba5aedf3e498568b

spinalcordtoolbox/utils.py
kousu: > the problem with that implementation is that it fails the following test:
> 
> ```python
>     assert utils.splitext('nice.image.nii.gz') == ('nice.image', '.nii.gz')
> ```
> 
> I've seen such file names in my life roll_eyes...

I was hoping they wouldn't be an issue. Alright, special-case away!

scripts/sct_download_data.py
```diff
@@ -12,23 +12,51 @@
...
+                         'https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip'],
+    'deepseg_gm_models': ['https://osf.io/b9y4x/?action=download',
+                          'https://www.neuro.polymtl.ca/_media/downloads/sct/20180205_deepseg_gm_models.zip'],
+    'deepseg_sc_models': ['https://osf.io/avf97/?action=download',
+                          'https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip'],
+    'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
+                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip'],
+    'c2c3_disc_models': ['https://osf.io/t97ap/?action=download',
+                         'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip']
+    }
```
kousu, line 58: DICT_URL is still getting deleted/added, so it's going to end up make two copies of it when I rebase my PRs. Can you put it back down below `get_parser`?

Changing the name to uppercase should be okay.

unit_testing/test_utils.py
```diff
@@ -7,6 +7,20 @@
 from spinalcordtoolbox import utils
 
 
+def test_add_suffix():
+    assert utils.add_suffix('t2.nii', '_mean') == 't2_mean.nii'
+    assert utils.add_suffix('t2.nii.gz', 'a') == 't2a.nii.gz'
```
kousu, line 12:
```suggestion
    assert utils.add_suffix('t2.nii.gz', 'a') == 't2a.nii.gz'
    assert utils.add_suffiix('var/lib/usr/t2.nii.gz', 'sfx') == 'var/lib/usr/t2sfx'.nii.gz'
    assert utils.add_suffiix('var/lib.version.3/usr/t2.nii.gz', 'sfx') == 'var/lib.version.3/usr/t2sfx'.nii.gz'
```

scripts/sct_download_data.py
jcohenadad: i can, but in that case i don't know how to implement [this](https://github.com/neuropoly/spinalcordtoolbox/pull/2700#pullrequestreview-408886355)

kousu: That's inside of `get_parser()` so it should be fine -- it doesn't actually run until `main()` calls it.

kousu: I know, that list really should be more prominent in this file. But we can do it in a small PR next week.

jcohenadad: hum, i am still getting the error (in 9840ff76f611518c288e3db8c5fdd2dce779e636):
```bash
 sct_download_data -h

--
Spinal Cord Toolbox (git-jca/download-data-9840ff76f611518c288e3db8c5fdd2dce779e636)

  File "/Users/julien/code/sct/scripts/sct_download_data.py", line 113, in <module>
    res = main()
  File "/Users/julien/code/sct/scripts/sct_download_data.py", line 97, in main
    parser = get_parser()
  File "/Users/julien/code/sct/scripts/sct_download_data.py", line 32, in get_parser
    example=sorted(dict_url))
NameError: name 'dict_url' is not defined
```

kousu: Oh it was a local variable. I should have tried it myself, sorry.

Two fixes:

1. Very quickly: `parser = get_parser()` -> `parser = get_parser(dict_url)`
2. You could probably move ` def main(args=None): ` *after* `dict_url`; then there will be a small patch on the `dict_url` line but the rest of its contents should stay put.

We'll clean this up better once we've merged our work fully!

unit_testing/test_utils.py
jcohenadad: done in 58e1d86e8e4c7cb1961a6e68b643947f7390854a

scripts/sct_download_data.py
jcohenadad: 👍 i did the first suggestion 8ad4aee81c0a03b42f87cd5a03fb0eade07bf51c

PR #2704, opened 2020-05-11
warp_template: replaces global Param instance with class parameter

### LINKED ISSUE #2320: Calling WarpTemplate fails
Hello, I was trying to use the class `WarpTemplate` (from `sct_warp_template`), but I couldn't call it directly in my python environment. Here is the output I get:

```
  File "sct_crop_template.py", line 137, in <module>
    folder_out=folder_out, path_template=path_template, verbose=cropper.verbose)
  File "/cerebro/cerebro1/dataset/bmpd/derivatives/sc_preproc/code/sct/scripts/sct_warp_template.py", line 53, in __init__
    self.folder_template = param.folder_template
NameError: name 'param' is not defined
```

The reason is that the variable "param" is defined at the end of the script, in the scope of `if __name__ == "__main__":`, which is not run when calling `WarpTemplate` directly.

There might be other scripts that have the same issue.

saulo-p:
Replaces the global instance of Param with a parameter in `WarpTemplate` class,
allowing `WarpTemplate` class to be instantiated outside this script.

Fixes #2320

### DIFF
--- a/scripts/sct_warp_template.py
+++ b/scripts/sct_warp_template.py
@@ -44,6 +44,7 @@ class WarpTemplate:
     def __init__(self, fname_src, fname_transfo, warp_atlas, warp_spinal_levels, folder_out, path_template, verbose):
 
         # Initialization
+        param = _param if '_param' in globals() else Param()
         self.fname_src = fname_src
         self.fname_transfo = fname_transfo
         self.warp_atlas = warp_atlas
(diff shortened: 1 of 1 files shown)


scripts/sct_warp_template.py
```diff
@@ -248,5 +252,4 @@ def main(args=None):
 
 if __name__ == "__main__":
     sct.init_sct()
-    param = Param()
```
zougloub, line 251: You can keep the param_default in the module scope and avoid passing a useless/cryptic argument.

saulo-p: I accepted your suggestion.
My only concern is that when instantiating the class `WarpTemplate`, the parameters below will not be accessible through the constructor and therefore can't be changed from what's in the `Param` class definition. 
```
        self.folder_template = param.folder_template
        self.folder_atlas = param.folder_atlas
        self.folder_spinal_levels = param.folder_spinal_levels
```

But that seems to be the desired behavior anyway.

joshuacwnewton:
Closing, as this exact change just so happened to be introduced in #3091. From that PR description:

> This PR refactors scripts to all have the same consistent structure, as described in the [CLI script conventions](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-CLI-script-conventions-(structure,-argparse)) dev Wiki page.
>
> But, for some scripts, a more involved refactor was needed:
> * `sct_warp_template`: `Param()` was used as a sort of global default. I refactored this to use explicitly passed variables (`verbose`, `folder_template`, `folder_atlas`, `folder_spinal_levels`, `file_info_label`, `list_labels_nn`)

PR #2706, opened 2020-05-12
Use Travis Caching

kousu:
This should speed up the tests by a couple of minutes per run.

This caches:
* the user-wide pip packages (which strangely we appear to have despite only using conda?) (which are 350M)
* the conda install, and `venv_sct` (which is 1.1G)

To achieve this, I removed the forcible `rm -rf python/` in the installer, so check that over and tell me if you agree with this strategy.

We might think about also caching:
* the trained models
* `sct_testing_data/`
If we cached those, then we would need to remember to clear the cache manually when we update them. But they change slowly, at the speed of our other dependencies, so it could make a lot of sense to treat them the same.

ref: https://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -7,12 +7,14 @@
 #
 # e.g. DOCKER_IMAGE="centos:8" .travis.sh
 
+set -x
 ls -l $HOME/.cache
 ls -l $HOME/.cache/pip
 du -hs $HOME/.cache/pip
 ls -l ~/.cache/pip/wheels/*/*/*/*/*.whl
 ls -l python
 du -hs python
+set +x
 
 set -e # Error build immediately if install script exits with non-zero
 
--- a/.travis.sh
+++ b/.travis.sh
@@ -7,6 +7,11 @@
 #
 # e.g. DOCKER_IMAGE="centos:8" .travis.sh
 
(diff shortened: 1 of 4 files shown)


kousu:
Actually this doesn't seem to be improving build times at all. Maybe it's because of ?

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source:

- https://docs.travis-ci.com/user/caching/#things-not-to-cache

jcohenadad:
> If we cached those, then we would need to remember to clear the cache manually when we update them. But they change slowly, at the speed of our other dependencies, so it could make a lot of sense to treat them the same.

this was the reason i explicitly removed the caching option a few years ago, when fighting against a bug for several days, after realizing that this was because the data were not udpated on travis because of the caching. I have learned it the hard way, and i suspect this might reproduce in the future. Given that the download is pretty fast (travis has a huge BW, i think around 50+MB/s) so i would not worry about a few seconds wasted.

kousu:
I'm going to hang onto the branch because I might want to revive this -- particularly, the part that makes installing conda idempotent, that drops the `rm -rf python/` line -- but I think a more useful step would be to split the unit and integration tests to separate jobs first.

PR #2710, opened 2020-05-13
label_utils: remove unecessary indexing permutation

### LINKED ISSUE #2424: sct_label_utils: IndexError: index 118 is out of bounds for axis 2 with size 118
When using a segmentation file as input.  

syntax:
`sct_label_utils -i nifti_seg.nii.gz -create-seg 0,14:117,8`

```
> Spinal Cord Toolbox (git-master-ff7553098ea27d2b6a51a90965a526f53368805e*)
> 
>   File "/home/felix/Repos/sct/scripts/sct_label_utils.py", line 828, in <module>
>     main()
>   File "/home/felix/Repos/sct/scripts/sct_label_utils.py", line 818, in main
>     processor.process(process_type)
>   File "/home/felix/Repos/sct/scripts/sct_label_utils.py", line 112, in process
>     self.output_image = self.create_label_along_segmentation()
>   File "/home/felix/Repos/sct/scripts/sct_label_utils.py", line 215, in create_label_along_segmentation
>     x, y = ndimage.measurements.center_of_mass(np.array(im_rpi.data[:, :, z_rpi]))
> IndexError: index 118 is out of bounds for axis 2 with size 118 
```

saulo-p:
The input image is reoriented at the function start, and later on, indexed according to its original orientation instead of the new one.

Please see comment on commit for further explanation.

Fixes #2424

### DIFF
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -239,13 +239,9 @@ class ProcessLabels(object):
             # split coord string
             list_coord = coord.split(',')
             # convert to int() and assign to variable
-            z, value = [int(i) for i in list_coord]
-            # update z based on native image orientation (z should represent superior-inferior axis)
-            coord = Coordinate([z, z, z])  # since we don't know which dimension corresponds to the superior-inferior
-            # axis, we put z in all dimensions (we don't care about x and y here)
-            _, _, z_rpi = coord.permute(self.image_input, 'RPI')
+            z_rpi, value = [int(i) for i in list_coord]
             # if z=-1, replace with nz/2
-            if z == -1:
+            if z_rpi == -1:
                 z_rpi = int(np.round(im_output_rpi.dim[2] / 2.0))
             # get center of mass of segmentation at given z
             x, y = ndimage.measurements.center_of_mass(np.array(im_rpi.data[:, :, z_rpi]))


saulo-p:
> thank you @saulo-p, i feel it would be a great opportunity to start introducing unit tests for this function, so we can reproduce the bug and make sure the proposed change fixes it, what do you think?

Agreed. I will add the test as part of the PR.

saulo-p:
After taking a second look at what I've proposed before, I noticed that my initial reading of the issue was wrong.
The coordinate change (z coordinate) is indeed necessary because the user will pass the input Z coordinate in the same orientation of the input image `-i`.

This being said, the _docstring_ of the function `create_label_along_segmentation()` states that the input image orientation is not constrained to be RPI.
```
Create an image with labels defined along the spinal cord segmentation (or centerline).
Input image does **not** need to be RPI (re-orientation is done within this function).
```

If the input image is not RPI, the input Z coordinates are going to be permuted to follow the image reorientation:
```
coord = Coordinate([z, z, z])  # since we don't know which dimension corresponds to the superior-inferior
_, _, z_rpi = coord.permute(self.image_input, 'RPI')
```

But checking the dimensions of the input image x reoriented image and the Z coordinates and with `Z=0` I got the following:
```
input image (LSA): (150, 107, 120) -> reoriented image (RPI): (150, 120, 107)
coord: 0,0,0,0 -> coord permuted: [150 120 107] -> z_rpi = 107
```

The problem here is that even with a valid(?) input coordinate `z=0`, the transformed `z_rpi` is out of bounds of the reoriented image causing the crash.


@jcohenadad  Could this be an edge case of permute()?
If not, should I just enforce the input image to be RPI (as you suggested in the issue) and change the documentation/help accordingly?

joshuacwnewton:
`sct_label_utils` has undergone significant refactoring since this PR. It would be nontrivial to resolve merge conflicts here, and since there is only a single commit, I think I am going to start fresh as I continue from @saulo-p's work. 

I'm going to close this, but thank you very much for your detailed analysis, @saulo-p.

Issue #2683, opened 2020-05-02
test_many_qc: json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)


kousu:
Found [by travis](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/682023340), this is an intermittent bug. It happened on one of the scheduled trial [rebuilds](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/682023338) even though all the previous scheduled builds on the same branch/commit and even the other jobs in the same build passed.

```
=================================== FAILURES ===================================

_________________________________ test_many_qc _________________________________

multiprocessing.pool.RemoteTraceback: 

"""

Traceback (most recent call last):

  File "/home/travis/build/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 119, in worker

    result = (True, func(*args, **kwds))

  File "/home/travis/build/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar

    return list(map(*args))

  File "/home/travis/build/neuropoly/spinalcordtoolbox/unit_testing/test_qc_parallel.py", line 21, in gen_qc

    process = "sct_deepseg_gm")

  File "/home/travis/build/neuropoly/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 673, in generate_qc

    angle_line=angle_line

  File "/home/travis/build/neuropoly/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 545, in add_entry

    layout(qcslice)

  File "/home/travis/build/neuropoly/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 332, in wrapped_f

    self.qc_report.update_description_file(img.shape)

  File "/home/travis/build/neuropoly/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 483, in update_description_file

    self._update_html_assets(get_json_data_from_path(path_json))

  File "/home/travis/build/neuropoly/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 683, in get_json_data_from_path

    results.append(json.load(fjson))

  File "/home/travis/build/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/json/__init__.py", line 299, in load

[...]

PR #2711, opened 2020-05-13
Add retrying to QC JSON reading

cfhammill:
Should solve the flaky test bug in #2683

Adds retrying as a dependency, which I think is lightweight enough, but happy to refactor to avoid the added dependency. As noted in the issue thread, I was unable to reproduce the failures locally, but I'm reasonably confident this will fix the error we've observed in CI.


cfhammill:
@kousu would you mind giving this a quick review?

kousu:
I have one idea for a more robust but still pretty basic solution: the reader uses glob to decide what files to load, so we can make sure the output file isn't marked ready for reading until it's fully written, the same way Firefox and Chrome do, with this:

```
        # Create json file
        with open(self.qc_params.qc_results+".tmp", 'w+') as qc_file:
            json.dump(output, qc_file, indent=1)
        os.rename(self.qc_params.qc_results+".tmp", self.qc_params.qc_results)
```
(assuming `rename()` is atomic, which it is on most systems).

This would stop the crash but it wouldn't deal with the basic problem that the reader isn't blocking on the writers like it should be -- so it would make the reader lose some data.

My next thought would be to set up a queue of results from the writers to the readers -- which should be easy; in fact that's pretty much exactly what `multiprocessing.Pool.map()`. But I am going to keep reading the code to see why/why not that may be possible here.

jcohenadad:
for the records, we did use file locking in the past (https://github.com/neuropoly/spinalcordtoolbox/pull/2206), introduced the function `open_with_exclusive_lock()`, but then it was dropped [here](https://github.com/neuropoly/spinalcordtoolbox/commit/7d7e0d868653f1be400f74f8f23a6796d503e9e2#diff-cd0ed16d3a599345c58ea8bbd2fe675d) because it didn't solve the issue-- the solution (which, as we now today was not a solution), was to split the single `qc_results.json` file into separate files per process.

kousu:
I do like `retrying` though; it's a very elegant little decorator.

kousu:
> for the records, we did use file locking in the past (#2206), introduced the function `open_with_exclusive_lock()`, but then it was dropped [here](https://github.com/neuropoly/spinalcordtoolbox/commit/7d7e0d868653f1be400f74f8f23a6796d503e9e2#diff-cd0ed16d3a599345c58ea8bbd2fe675d) because it didn't solve the issue-- the solution (which, as we know today was not a solution), was to split the single `qc_results.json` file into separate files per process.

Ah that's good context! That solution is halfway there. The problem is that after writing the per-process output, each process immediately goes and re-reads *all* the output files.

If I was actually doing this I would do something like:

```
cat jobs.csv | parallel -n 2 generate_qc | generate_qc_summary
```

In this phrasing, the bug is because `generate_qc_summary` is embedded inside of `generate_qc`.

PR #2715, opened 2020-05-17
Minimal Torch install.

### LINKED ISSUE #2712: Torch-cpu installation in Spinal cord Toolbox
Following issue [208](https://github.com/neuropoly/ivado-medical-imaging/issues/208) from Ivadomed. 
The pytorch-cpu installation is OS-specific due to a lack of availability of CUDA package for OSx.

One solution could be to add : 
~~~
case $OS in
linux*)
        pip install torchvision==0.6.0+cpu torch==1.5.0+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html;;
osx)
        pip install torch==1.5.0 torchvision==0.6.0;;
esac
~~~
[here](https://github.com/neuropoly/spinalcordtoolbox/blob/f80551b61639f46e93760dc8329dba74aad24559/install_sct#L533)

But I am not sure if this is the best solution.

kousu:
Fixes #2712 .


requirements.txt
```diff
@@ -13,10 +13,18 @@ pytest
...
 scikit-image
 scikit-learn
 tensorflow==1.5.0
+# PyTorch's Linux/Windows distribution is very large due to its GPU support,
+# but we only need that for training models. For users, use the CPU-only version
+# (only available directly from the PyTorch project).
+# The OS X version has never had GPU support, so doesn't need the workaround.
+-f https://download.pytorch.org/whl/cpu/torch_stable.html
+torch==1.5.0+cpu; sys_platform != "darwin"
+torch==1.5.0; sys_platform == "darwin"
```
cfhammill, line 27: Is it possible to make this conditional? we may want to install to get the gpu version by default when we're not on travis

cfhammill:
I tried this PR on my machine and got:

```
sct_check_dependencies

--
Spinal Cord Toolbox (git-ko/minimal-torch-7528512c58628f2fb8b288fed8bc39277885f255)

SCT info:
- version: git-ko/minimal-torch-7528512c58628f2fb8b288fed8bc39277885f255
- path: /home/chammill/documents/spinalcordtoolbox
OS: linux (Linux-4.4.0-18362-Microsoft-x86_64-with-debian-buster-sid)
CPU cores: Available: 6, Used by SCT: 6
RAM: MemTotal:        8244400 kB
total        used        free      shared  buff/cache   available
Mem:           8051        5790        2036          17         223        2129
Swap:         15105         596       14509
  Using bundled python 3.6.10 |Anaconda, Inc.| (default, Jan  7 2020, 21:14:29)
[GCC 7.3.0] at /home/chammill/documents/spinalcordtoolbox/python/envs/venv_sct/bin/python
Check if numpy is installed.........................[OK] (1.17.4)
Check if cryptography is installed..................[OK] (2.8)
Check if dipy is installed..........................[OK] (1.0.0)
Check if matplotlib is installed....................[OK] (3.1.2)
Check if nibabel is installed.......................[OK] (3.0.0)
Check if pandas is installed........................[OK] (0.25.3)
Check if psutil is installed........................[OK] (5.6.7)
Check if pytest is installed........................[OK] (5.3.2)
Check if pytest-cov is installed....................[OK] (2.8.1)
Check if requests is installed......................[OK] (2.22.0)
Check if scipy is installed.........................[OK] (1.4.0)
Check if scikit-image is installed..................[OK] (0.16.2)
Check if scikit-learn is installed..................[OK] (0.22)
Check if torch (1.5.0+cpu) is installed.............Segmentation fault (core dumped)
```

requirements.txt
bcdarwin: It's a bit hard to depend on Pytorch having CUDA without using a system package manager like Nix since conda/pip don't know what cuda version your system is running (see https://pytorch.org/get-started/locally/) and trying to guess/autodetect also seems like a mess.

kousu:
To handle training I think we should write a setup.py that sets `extras_requires = {'training': ["torch=1.5.0; sys_platform == 'linux'}`, and leave this default install, the one most users will use, to CPU-only mode. For us, we would do `pip install -e .[training]`, and the training code would just have to detect if CUDA is available and warn if not.

Before doing that, we can leave in our dev docs "if you want to do training, pip install torch==1.5.0". We could even keep a training branch around with that patched in requirements.txt (and a note about keeping it up to date with, but not merged into, master).
-- 
kousu

kousu:
On May 17, 2020 2:15:04 PM EDT, Chris Hammill <notifications@github.com> wrote:
>I tried this PR on my machine and got:
>
>Check if tensorflow (1.5.0) is installed............[OK] (1.5.0)
>Check if torch (1.5.0+cpu) is installed.............Segmentation fault
>(core dumped)
>```

Drat. I'm glad we caught this early whatever it is. Do you have CUDA installed? What platform are you on? etc?

Do you know how to run Python under gdb to get a C-level backtrace?

bcdarwin:
> To handle training I think we should write a setup.py that sets `extras_requires = {'training': ["torch=1.5.0; sys_platform == 'linux'}`, and leave this default install, the one most users will use, to CPU-only mode. For us, we would do `pip install -e .[training]`, and the training code would just have to detect if CUDA is available and warn if not. Before doing that, we can leave in our dev docs "if you want to do training, pip install torch==1.5.0". We could even keep a training branch around with that patched in requirements.txt (and a note about keeping it up to date with, but not merged into, master). -- kousu

Agree but -1 for keeping a separate branch as I think it's not the most convenient way to have this separation in terms of user or developer workflow.

requirements.txt
jcohenadad: @cfhammill inference time is so fast on modern CPUs (a few seconds for an entire volume) that i don't think we need to bother maintaining GPU version of Pytorch.

jcohenadad:
> To handle training I think we should write a setup.py that sets `extras_requires = {'training': ["torch=1.5.0; sys_platform == 'linux'}`, and leave this default install, the one most users will use, to CPU-only mode. For us, we would do `pip install -e .[training]`, and the training code would just have to detect if CUDA is available and warn if not. Before doing that, we can leave in our dev docs "if you want to do training, pip install torch==1.5.0". We could even keep a training branch around with that patched in requirements.txt (and a note about keeping it up to date with, but not merged into, master). -- kousu

no need to support training at this point. Training of new models is done under the separate repos: https://github.com/neuropoly/ivado-medical-imaging

requirements.txt
kousu: Oops github-email didn't put my comment in the right thread: I think we can handle different use cases nicely if we write up setup.py and use `extras_requires`: https://github.com/neuropoly/spinalcordtoolbox/pull/2715#issuecomment-629878087. Another approach I've seen is to maintain two requirements.txts: the main one, and a requirements-dev.txt, but I think `extras` is nicer.

kousu:
> On May 17, 2020 2:15:04 PM EDT, Chris Hammill ***@***.***> wrote: I tried this PR on my machine and got: Check if tensorflow (1.5.0) is installed............[OK] (1.5.0) Check if torch (1.5.0+cpu) is installed.............Segmentation fault (core dumped) ```
> Drat. I'm glad we caught this early whatever it is. Do you have CUDA installed? What platform are you on? etc? Do you know how to run Python under gdb to get a C-level backtrace?

I read the log closer: it's Debian-on-WSL-on-Windows. I don't have a Windows VM set up yet to test this but I'm not surprised PyTorch has problems there, given that PyTorch seems to target a very limited set of platforms. It looks like [CUDA is a hard no](https://github.com/Microsoft/WSL/issues/3789), but I don't see why that would cause problems since it _should_ have installed the +cpu version.

Does `import torch` work in any shell @cfhammill? What does `pip freeze | grep torch` say about it?

kousu:
Hold off merging this until we get this working on WSL.

requirements.txt
lrouhier, line 27: You should add torchvision==0.6.0(+cpu for linux) which is needed in all cases.

kousu: Thank you for catching this. I've patched it and it's running CI now.

lrouhier:
> Check if tensorflow (1.5.0) is installed............[OK] (1.5.0)
> Check if torch (1.5.0+cpu) is installed.............Segmentation fault (core dumped)
> ```

Reminded me of: https://github.com/scipy/scipy/issues/11237. (sorry for being so slow I've already seen this I just didn't know where )

cfhammill:
Seems to be the scipy issue @lrouhier suggested.

cfhammill:
Upgrading scipy to 1.4.1 fixed my issue, suggested fix, lower bound scipy to 1.4.1 in `requirements.txt`

PR #2731, opened 2020-05-24
Drop support for 2014-era Linux.

kousu:
Merging pytorch (6aa1f4e) caused

> OSError: dlopen: cannot load any more object with static TLS

on [Debian 8](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/690411071) and [Ubuntu Trusty](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/690411076) ( caught by #2690  :) ).

Screenshot_2020-05-24 Build #13281 - neuropoly spinalcordtoolbox - Travis CI

This is a known issue with pytorch, see e.g.

https://github.com/pytorch/pytorch/issues/2083
https://github.com/scikit-learn/scikit-learn/issues/14485
https://github.com/pytorch/pytorch/issues/2575
https://stackoverflow.com/questions/14892101/cannot-load-any-more-object-with-static-tls

Apparently there might be a workaround involving reordering import statements, but I think the root issue is that pytorch doesn't get testing on "obscure" platforms. Including pytorch means we can't reliably support them.

---

An alternate fix would be to try to reorder the import statements until this works for the old Linux. I'm not sure what we want to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,14 +67,6 @@ matrix:
         - docker
       env:
         - DOCKER_IMAGE="debian:9"       DOCKER_DEPS_CMD="apt update && apt install -y libglib2.0-0 procps gcc git curl"
-    - name: "Debian 8"
-      if: branch = release or type = cron
-      os: linux
-      dist: xenial # kernel ~= Debian:8
-      services:
-        - docker
-      env:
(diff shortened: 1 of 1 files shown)


kousu:
Note: Travis passing on this PR isn't going to mean it's actually fixed, because it's specifically failing on the extra platforms that aren't tested on every PR. I'm pretty confident I didn't make any typos, however.

Issue #2415, opened 2019-08-14
Use CDN for downloading data, binaries and miniconda


jcohenadad:
Currently, data and binaries are hosted on OSF + neuropoly website (as a mirror where OSF is not accessible for some countries). 

This approach is suboptimal as:
- it requires more maintenance (two URLs to manage)
- download could be faster if server is closer to user

A suggestion would be to use a CDN for data, binaries and miniconda.

Related to #2403

kousu:
The advantage to the current approach is that we have some redundancy.

We could also make a Github project for each binary/data package we need and use Github Releases as our CDN. Github is less blocked around the world than what we're using now.

I think we can probably eventually package everything -- the data, the binaries -- as python packages, and then we can trust in PyPI's CDN and there's minimal maintenance on our end -- all we need is to run `twine` somewhere, and as we saw this week in https://github.com/neuropoly/ivado-medical-imaging/pull/217 we can fully automate that.

jcohenadad:
i'm not sure i understand where physically the data/binary would be, and how we would maintain them.

kousu:
https://pypi.org/sponsors/
> PyPI uses Fastly's CDN to quickly serve content to end-users, allowing us to minimize our hosting infrastructure and obscure possible downtime.

So, files get uploaded to PyPI's core mirrors and are stored there -- probably in the US? When a request comes in for a user via Fastly it hits Fastly's servers, which are [scattered around the world](https://www.fastly.com/network-map), and if those servers have that file cached they return it, if not they go talk to PyPI's core mirror first, download it, then relay it to the user.

(if we're worried about slow/blocked downloads in certain countries then it's important to note that Fastly only has one mirror in China and two in Africa and none in Russia/north Asia.)

Other CDNs work the same way. They're basically just automated mirrors, doing what we're currently doing by hand but on-demand, and usually they're paired with high bandwidth datacenters.

kousu:
I just did some experiments and found that Github has a CDN too. Here in Canada, Github serves from `lb-*-iad.github.com` (eg `lb-140-82-114-4-iad.github.com`); from Germany, it serves from `lb-*-ams.github`.com` (eg `lb-140-82-118-4-ams.github.com`). "iad" is somewhere in the US according to [MaxMind](https://www.maxmind.com/en/geoip-demo), and "ams" is short for Amsterdam -- Amsterdam has become something of a big data hub lately.

[...]

PR #2733, opened 2020-05-25
Retrieve data bundles from their new location

zougloub:
For #2415.

Using Github releases as CDN, but another CDN fetching data from the releases can also be used.

### History Recreation

Data in sct_download_data is converted to be in their own repository, using a conversion script (https://github.com/sct-data/sct-data-archeology), which:

- [x] Looks in spinalcordtoolbox history for different versions of the various bundles
  - [x] in sct_download_data
  - [x] in [speenalcordtoolbox.deepseg.models](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L20)
- [x] One repo for a set of bundles (one or more)
- [x] Recreate a (linear) git DAG
- [x] Commits have date, corresponding SCT commit ref
- [x] Releases: create releases
- [x] Releases: script adds release artifact with the original bundle archive

Data is stored in the new https://github.com/sct-data/

### Bundle Release Process

Introduce a tool used to automate deployment.

### DIFF
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -52,43 +52,85 @@ def get_parser(dict_url):
 
 def main(args=None):
 
-    # Dictionary containing list of URLs for data names. Mirror servers are listed in order of priority.
+    # Dictionary containing list of URLs for data names.
+    # Mirror servers are listed in order of decreasing priority.
+    # If exists, favour release artifact straight from github
     dict_url = {
-        'sct_example_data': ['https://osf.io/kjcgs/?action=download',
-                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20180525_sct_example_data.zip'],
-        'sct_testing_data': ['https://osf.io/yutrj/?action=download',
-                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20191031_sct_testing_data.zip'],
-        'PAM50': ['https://osf.io/u79sr/download',
-                  'https://www.neuro.polymtl.ca/_media/downloads/sct/20191029_PAM50.zip'],
-        'MNI-Poly-AMU': ['https://osf.io/sh6h4/?action=download',
-                         'https://www.neuro.polymtl.ca/_media/downloads/sct/20170310_MNI-Poly-AMU.zip'],
-        'gm_model': ['https://osf.io/ugscu/?action=download',
-                     'https://www.neuro.polymtl.ca/_media/downloads/sct/20160922_gm_model.zip'],
-        'optic_models': ['https://osf.io/g4fwn/?action=download',
-                         'https://www.neuro.polymtl.ca/_media/downloads/sct/20170413_optic_models.zip'],
-        'pmj_models': ['https://osf.io/4gufr/?action=download',
-                       'https://www.neuro.polymtl.ca/_media/downloads/sct/20170922_pmj_models.zip'],
-        'binaries_linux': ['https://osf.io/mka78/?action=download',
-                           'https://www.neuro.polymtl.ca/_media/downloads/sct/20200420_sct_binaries_linux.tar.gz'],
-        'binaries_osx': ['https://osf.io/dn67h/?action=download',
-                         'https://www.neuro.polymtl.ca/_media/downloads/sct/20200420_sct_binaries_osx.tar.gz'],
-        'course_hawaii17': 'https://osf.io/6exht/?action=download',
-        'course_paris18': ['https://osf.io/9bmn5/?action=download',
-                           'https://www.neuro.polymtl.ca/_media/downloads/sct/20180612_sct_course-paris18.zip'],
-        'course_london19': ['https://osf.io/4q3u7/?action=download',
-                            'https://www.neuro.polymtl.ca/_media/downloads/sct/20190121_sct_course-london19.zip'],
-        'course_beijing19': ['https://osf.io/ef4xz/?action=download',
-                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip'],
-        'deepseg_gm_models': ['https://osf.io/b9y4x/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180205_deepseg_gm_models.zip'],
-        'deepseg_sc_models': ['https://osf.io/avf97/?action=download',
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip'],
-        'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
-                                  'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models'
-                                  '.zip'],
-        'c2c3_disc_models': ['https://osf.io/t97ap/?action=download',
-                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip']
-        }
+        "sct_example_data": [
+            "https://github.com/sct-data/sct_example_data/releases/download/r20180525/20180525_sct_example_data.zip",
+            "https://osf.io/kjcgs/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20180525_sct_example_data.zip",
+        ],
+        "sct_testing_data": [
+            "https://github.com/sct-data/sct_testing_data/releases/download/r20200504/20200504_sct_testing_data.zip",
+            "https://osf.io/6x5a2/download",
+        ],
+        "PAM50": [
+            "https://github.com/sct-data/PAM50/releases/download/r20191029/20191029_pam50.zip",
+            "https://osf.io/u79sr/download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20191029_PAM50.zip",
+        ],
+        "MNI-Poly-AMU": [
+            "https://github.com/sct-data/MNI-Poly-AMU/releases/download/r20170310/20170310_MNI-Poly-AMU.zip",
+            "https://osf.io/sh6h4/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20170310_MNI-Poly-AMU.zip",
+        ],
+        "gm_model": [
+            "https://osf.io/ugscu/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20160922_gm_model.zip",
+        ],
+        "optic_models": [
+            "https://github.com/sct-data/optic_models/releases/download/r20170413/20170413_optic_models.zip",
+            "https://osf.io/g4fwn/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20170413_optic_models.zip",
+        ],
+        "pmj_models": [
+            "https://github.com/sct-data/pmj_models/releases/download/r20170922/20170922_pmj_models.zip",
+            "https://osf.io/4gufr/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20170922_pmj_models.zip",
+        ],
+        "binaries_linux": [
+            "https://osf.io/mka78/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20200420_sct_binaries_linux.tar.gz",
+        ],
+        "binaries_osx": [
+            "https://osf.io/dn67h/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20200420_sct_binaries_osx.tar.gz",
+        ],
+        "course_hawaii17": "https://osf.io/6exht/?action=download",
+        "course_paris18": [
+            "https://osf.io/9bmn5/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20180612_sct_course-paris18.zip",
+        ],
+        "course_london19": [
+            "https://osf.io/4q3u7/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20190121_sct_course-london19.zip",
+        ],
+        "course_beijing19": [
+            "https://osf.io/ef4xz/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip",
+        ],
+        "deepseg_gm_models": [
+            "https://github.com/sct-data/deepseg_gm_models/releases/download/r20180205/20180205_deepseg_gm_models.zip",
+            "https://osf.io/b9y4x/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20180205_deepseg_gm_models.zip",
+        ],
+        "deepseg_sc_models": [
+            "https://github.com/sct-data/deepseg_sc_models/releases/download/r20180610/20180610_deepseg_sc_models.zip",
+            "https://osf.io/avf97/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip",
+        ],
+        "deepseg_lesion_models": [
+            "https://github.com/sct-data/deepseg_lesion_models/releases/download/r20180613/20180613_deepseg_lesion_models.zip",
+            "https://osf.io/eg7v9/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models.zip",
+        ],
+        "c2c3_disc_models": [
+            "https://github.com/sct-data/c2c3_disc_models/releases/download/r20190117/20190117_c2c3_disc_models.zip",
+            "https://osf.io/t97ap/?action=download",
+            "https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip",
+        ],
+    }
 
     if args is None:
         args = sys.argv[1:]
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -40,7 +40,10 @@ def get_parser(dict_url):
     parser.add_option(
         name="-o",
         type_value="folder_creation",
-        description="Path where to save the downloaded data (defaults to dataset name in current working directory)",
+        description=("Path where to save the downloaded data"
+         " (defaults to ./${dataset-name})."
+         " Directory will be created."
+         " Warning: existing data in the directory will be erased."),
         mandatory=False)
     parser.add_option(
         name="-h",
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -43,12 +43,17 @@ def get_parser(dict_url):
         description=("Path where to save the downloaded data"
          " (defaults to ./${dataset-name})."
          " Directory will be created."
-         " Warning: existing data in the directory will be erased."),
+         " Warning: existing data in the directory will be erased unless -k is provided."),
         mandatory=False)
     parser.add_option(
         name="-h",
         type_value=None,
         description="Display this help",
+        mandatory=False)
+    parser.add_option(
+        name="-k",
+        type_value=None,
+        description="Keep existing data in destination directory",
         mandatory=False)
     return parser
 
@@ -147,7 +152,7 @@ def main(args=None):
     dest_folder = arguments.get('-o', os.path.join(os.path.abspath(os.curdir), data_name))
 
     url = dict_url[data_name]
-    install_data(url, dest_folder)
+    install_data(url, dest_folder, keep=arguments.get("-k", False))
 
     sct.printv('Done!\n', verbose)
     return 0
--- a/scripts/sct_download_data.py
+++ b/scripts/sct_download_data.py
@@ -40,7 +40,7 @@ def get_parser(dict_url):
     parser.add_option(
         name="-o",
         type_value="folder_creation",
-        description="path to save the downloaded data",
+        description="Path where to save the downloaded data (defaults to dataset name in current working directory)",
         mandatory=False)
     parser.add_option(
         name="-h",
@@ -141,7 +141,7 @@ def main(args=None):
     data_name = arguments['-d']
     verbose = int(arguments.get('-v'))
     sct.init_sct(log_level=verbose, update=True)  # Update log level
-    dest_folder = arguments.get('-o', os.path.abspath(os.curdir))
+    dest_folder = arguments.get('-o', os.path.join(os.path.abspath(os.curdir), data_name))
 
     url = dict_url[data_name]
     install_data(url, dest_folder)
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -106,42 +106,75 @@ def unzip(compressed, dest_folder):
 
 def install_data(url, dest_folder):
     """
-    Download data from a URL and install in the appropriate folder. Deals with multiple mirrors, retry download,
-    check if data already exist.
-    :param url: string or list or strings. URL or list of URLs (if mirrors).
-    :param dest_folder:
-    :return:
+    Download a data bundle from a URL and install in the destination folder.
+
+    :param url: URL or sequence thereof (if mirrors).
+    :param dest_folder: destination folder
+    :return: None
+
+    Notes:
+
+    - Uses `download_data()` to retrieve the data.
+    - Uses `unzip()` to extract the bundle.
+    - If destination already exists, it is cleared first,
+      from files that would be in the bundle (not the rest).
     """
 
-    # Download data
     tmp_file = download_data(url)
 
-    # unzip
-    dest_tmp_folder = sct.utils.tmp_create()
-    unzip(tmp_file, dest_tmp_folder)
-    extracted_files_paths = []
-    # Get the name of the extracted files and directories
-    extracted_files = os.listdir(dest_tmp_folder)
-    for extracted_file in extracted_files:
-        extracted_files_paths.append(os.path.join(os.path.abspath(dest_tmp_folder), extracted_file))
+    extraction_folder = sct.utils.tmp_create()
 
-    # Check if files and folder already exists
+    unzip(tmp_file, extraction_folder)
+
+    # Identify whether we have a proper archive or a tarbomb
+    with os.scandir(extraction_folder) as it:
+        has_dir = False
+        nb_entries = 0
+        for entry in it:
+            nb_entries += 1
+            if entry.is_dir():
+                has_dir = True
+
+    if nb_entries == 1 and has_dir:
+        # tarball with single-directory -> go under
+        with os.scandir(extraction_folder) as it:
+            for entry in it:
+                bundle_folder = entry.path
+    else:
+        # bomb scenario -> stay here
+        bundle_folder = extraction_folder
+
     logger.info("Destination folder: {}".format(dest_folder))
-    logger.info("Checking if folder already exists...")
-    for data_extracted_name in extracted_files:
-        fullpath_dest = os.path.join(dest_folder, data_extracted_name)
-        if os.path.isdir(fullpath_dest):
-            logger.warning("Folder {} already exists. Removing it...".format(data_extracted_name))
-            shutil.rmtree(fullpath_dest)
 
-    # Destination path
-    # for source_path in extracted_files_paths:
-        # Copy the content of source to destination (and create destination folder)
-    distutils.dir_util.copy_tree(dest_tmp_folder, dest_folder)
+    os.makedirs(dest_folder, exist_ok=True)
+
+    for cwd, ds, fs in os.walk(bundle_folder):
+        ds.sort()
+        fs.sort()
+        for d in ds:
+            srcpath = os.path.join(cwd, d)
+            relpath = os.path.relpath(srcpath, bundle_folder)
+            dstpath = os.path.join(dest_folder, relpath)
+            if os.path.exists(dstpath):
+                # lazy -- we assume it's a directory, otherwise it will crash safely
+                logger.debug("- d- %s", relpath)
+            else:
+                logger.debug("- d+ %s", relpath)
+                os.makedirs(dstpath)
+
+        for f in fs:
+            srcpath = os.path.join(cwd, f)
+            relpath = os.path.relpath(srcpath, bundle_folder)
+            dstpath = os.path.join(dest_folder, relpath)
+            if os.path.exists(dstpath):
+                logger.debug("- f! %s", relpath)
+                logger.warning("Updating existing %s", dstpath)
+                os.unlink(dstpath)
+            else:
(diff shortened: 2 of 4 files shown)


zougloub:
I opened a pull request (dummy for now) so we may discuss on the WIP.

saulo-p:
@zougloub, I currently have something like this: https://github.com/saulo-p/SCT-data/releases

zougloub:
So what we had discussed:

- Make a conversion script which looks in spinalcordtoolbox history for different versions of the various bundles in sct_download_data
- Script makes one repo for a set of bundles (one or more)
- Script recreates a (linear) git DAG
  - commits have date, corresponding SCT commit ref
- Script adds release artifact with the original bundle archive

zougloub:
@saulo-p you can remove dot-files and dot-folders and other cruft? From a cursory look I see `__MACOSX` and `._*` stuff.

saulo-p:
@zougloub, I've managed to script the release creation and asset uploading ([reference](https://developer.github.com/v3/repos/releases/)). The result is shown in [auto-release-latest](https://github.com/saulo-p/SCT-data/releases/tag/4.1.0).

My script creates one release for each **tagged** commit, not for all commits.
Therefore, if multiple commits changing the data (in the SCT repo) were done between releases, the script will create a release only for the latest (enforcing coherence with the SCT repo regarding tags/releases).

There is a message saying "[...] 0 commits to master since this release".
The reason for this message is I'm running my tests on unmerged branches, so the master contains only the initial commit. Once I merge the data snapshots commits, this will be fixed.

saulo-p:
Hello @jcohenadad  ,
@zougloub opened this PR so we could discuss and come up with a solution for issue #2415. 

As discussed in the issue's comments and through slack, the chosen alternative for data provision was Github itself.
@zougloub suggested me to create GitHub repositories for each dataset, containing the history of changes to the data throughout the commits of sct.
This [repository](https://github.com/saulo-p/SCT-data) contains a demo (using the PAM50 dataset as example) of what the solution would look like.

The linked repository contains a commit for each change in the download link of the particular dataset.
The commit title contains the date of the original commit followed by its original title.
The commit body contains a link to GitHub's view of the original commit, showing the URL changes in the diff.

The repository contents consist of the extracted dataset itself, so the commit history in this repo shows the incremental changes in the dataset (particularly the CHANGES.md)

The commits in this repo are tagged matching the tags in sct. So a tag in the dataset repository allows to download the dataset contents of a particular tag/release of sct (might be helpful to replicate older issues (?)).
Also, for each tag, a release was created containing the dataset compressed file from the URL at that point in time.

I would like some feedback/suggestions on the presented solution.

jcohenadad:
@saulo-p that looks cool! i have a few questions/concerns:
- if i understand correctly, the PAM50 binaries in [this repos](https://github.com/saulo-p/SCT-data) are version-tracked (i.e. are part of the git master tree) instead of being "packaged" within a release. I thought we wanted to avoid this approach as it is not nice to pollute GH's server with binaries?  I know there is little difference with uploading the data directly to each release, but at least that would minimise the amount of data stored (there is currently 16 commits vs. 9 releases). I might misunderstand something here..
- What is the "human" procedure for going with that solution? More specifically: when we create a new SCT release, what should be done?

scripts/sct_download_data.py
```diff
@@ -54,6 +54,7 @@ def main(args=None):
 
     # Dictionary containing list of URLs for data names. Mirror servers are listed in order of priority.
     dict_url = {
+
```
jcohenadad, line 57: why this change?

zougloub: Hi, this was a dummy change in order to be able to create a PR in github ;)

jcohenadad: ah! i should learn to read commit messages...

zougloub:
> @saulo-p that looks cool! i have a few questions/concerns:
> 
>     * if i understand correctly, the PAM50 binaries in [this repos](https://github.com/saulo-p/SCT-data) are version-tracked (i.e. are part of the git master tree) instead of being "packaged" within a release.

Yes, that is correct, each curated "bundle" comes from a repo and has its release artifacts.

> I thought we wanted to avoid this approach as it is not nice to pollute GH's server with binaries?

What I said was not nice was to pollute pipy's package namespace and CDN. No problem with GH.

>  I know there is little difference with uploading the data directly to each release, but at least that would minimise the amount of data stored (there is currently 16 commits vs. 9 releases). I might misunderstand something here..

Before there were just zip files, now there's also a git repo on top of that, so yeah there is an overhead, however you must have had the sources of these zip somewhere, and now they would be tracked in a repo. I don't really see the problem, the end user won't see a difference.

>     * What is the "human" procedure for going with that solution? More specifically: when we create a new SCT release, what should be done?

Nothing should be done.
But when a bundle needs a new release, it would be updated & tagged in its own repo, then optionally if you care about it, do your own zipping of the data so it's immutable, then the sct_download_data would have a new URL.

scripts/sct_download_data.py
saulo-p: > if i understand correctly, the PAM50 binaries in [this repos](https://github.com/saulo-p/SCT-data) are version-tracked (i.e. are part of the git master tree) instead of being "packaged" within a release. 

In the dataset repo, the binaries are _both_ tracked and packaged. 
The packaging is done using the release assets, you can see for each release the original dataset `.zip` file.

> I thought we wanted to avoid this approach as it is not nice to pollute GH's server with binaries?  

The only problem I can see with the "scattered" binaries is that the dataset repo maintenance (cloning, pushing...) might be a little slow, but with the advantage that now the dataset files will be tracked which is usually good.

>   What is the "human" procedure for going with that solution? More specifically: when we create a new SCT release, what should be done?

The way I envision this, each dataset will have its own repository, so whenever there are changes to a particular dataset those would be committed to the respective repository.

As for the SCT releases, we could add some sort of script to SCT (replacing the functionality of `sct_download_data`) that points to the corresponding commits/releases in each dataset repository.

jcohenadad:
minor detail: why keeping the "./" in the URL (e.g: this commit message: https://github.com/saulo-p/SCT-data/commit/48704943df7b2b448a5627b5ef23e376cca415b3)

saulo-p:
> minor detail: why keeping the "./" in the URL (e.g: this commit message: [saulo-p/SCT-data@4870494](https://github.com/saulo-p/SCT-data/commit/48704943df7b2b448a5627b5ef23e376cca415b3))

Those should be removed in the final version, thanks.

jcohenadad:
> As for the SCT releases, we could add some sort of script to SCT (replacing the functionality of sct_download_data) that points to the corresponding commits/releases in each dataset repository.

what kind of scripts would that be and how would they be run (e.g. when calling sct_deepseg, it would download the necessary model if not already present? would it be called during SCT install?)

zougloub:
> As for the SCT releases, we could add some sort of script to SCT (replacing the functionality of sct_download_data) that points to the corresponding commits/releases in each dataset repository.

@saulo-p one step at a time. We're still relying on sct_download_data for now.

> what kind of scripts would that be and how would they be run (e.g. when calling sct_deepseg, it would download the necessary model if not already present? would it be called during SCT install?)

Sorry @jcohenadad, we can revisit later, let's keep this PR on topic.

jcohenadad:
i guess the repos would sit there: https://github.com/spinalcordtoolbox,right?

we should probably agree on naming for these repos. Personally i’m not a fan of capital letters for repos (i could be convinced otherwise)-- how about `sct-pam50`; `sct-testing-data`; `sct-example-data`; `sct-deepseg-t2-tumor`, etc.

zougloub:
> we should probably agree on naming for these repos. Personally i’m not a fan of capital letters for repos (i could be convinced otherwise)-- how about sct-pam50; sct-testing-data; sct-example-data; sct-deepseg-t2-tumor, etc.

Since they are into spinalcordtoolbox they could even afford to be called `testing-data`, `example-data`, but I'm OK with your proposal.

> EDIT: or, can we have a "parent" repos that encloses all the data bundles as "sub-repos" (i vaguely remember this was a feature in github)? just to avoid having 45 repos at the organization root (which might become irritable for the eyes)

IIRC Github doesn't have that (Gitlab does).

jcohenadad:
> > we should probably agree on naming for these repos. Personally i’m not a fan of capital letters for repos (i could be convinced otherwise)-- how about sct-pam50; sct-testing-data; sct-example-data; sct-deepseg-t2-tumor, etc.
> 
> Since they are inside spinalcordtoolbox they could even afford to be called `testing-data`, `example-data`, but I'm OK with your proposal.

maybe one thing that could help for maintenance/newbie in the team, would be to use the same name as the dataset, i.e., `PAM50`, `sct_testing_data`, etc.

> 
> > EDIT: or, can we have a "parent" repos that encloses all the data bundles as "sub-repos" (i vaguely remember this was a feature in github)? just to avoid having 45 repos at the organization root (which might become irritable for the eyes)
> 
> IIRC Github doesn't have that (Gitlab does).

indeed-- i got confused with "sub-modules" (which is a different thing...). Is it crazy to consider having a "sct-data" organization, with all the repos for the bundles?  My previous comment in this post would still apply (i.e. match name of dataset and repos)

jcohenadad:
@saulo-p do you have any comment about my post above re: sct-data organization? do you foresee any pros/cons?

saulo-p:
> @saulo-p do you have any comment about my post above re: sct-data organization? do you foresee any pros/cons?

To be honest I don't know much about Github organizations, I'll do some reading and come back with an opinion.

zougloub:
Creating a new organization is fine by me.

jcohenadad:
https://github.com/sct-data is now created-- i,ve added @zougloub @saulo-p and @kousu and you should be able to create new repos

alexfoias:
@jcohenadad do you think it's a good idea to add me on the org in case we need to add more people/repos in the future ? I can take care of that task.

saulo-p:
Transferring repositories between accounts/organizations seems to be straightforward as well, so I can't foresee issues.
[https://help.github.com/en/github/administering-a-repository/transferring-a-repository](https://help.github.com/en/github/administering-a-repository/transferring-a-repository).

It might be hard to keep track of all organizations, though:
https://github.com/spinalcordtoolbox
https://github.com/sct-pipeline
https://github.com/sct-data
...

zougloub:
Summary on the bundles destiny (WIP):

sct_download_data:
- images: sct_example_data, sct_testing_data
  -> they have their own repos already so they'll be moved and the history will be fixed if needed
- templates & "geometric" models: PAM50, MNI-Poly-AMU, pmj_models, optic_models, gm_model(?), c2c3_disc_models
  -> reconstituted repos
- "data-crunching" models: deepscseg_models, deepseg_gm_models, deepseg_lesion_models, deepseg_sc_models
  -> reconstituted repos
- binaries: binaries_centos, binaries_debian, binaries_linux, binaries_osx
  -> won't touch for now
- courses data: course_hawaii17, course_paris18, course_london19, course_beijing19 / sct_course-beijing19
  -> single revision for each bundle, propose to postpone their handling

Others: WIP

jcohenadad:
> * images: sct_example_data, sct_testing_data
>   -> they have their own repos already so they'll be moved and the history will be fixed if needed

sct_testing_data: the GH repos was not up to date: https://github.com/sct-data/sct_testing_data/issues/1

sct_example_data: same here. This repos (https://github.com/neuropoly/sct_example_data) has not been updated for the past ~5y. The up to date data are in OSF: https://osf.io/6zbyf/files/

zougloub:
We are working on the assumption that all this data was not revision-controlled, but if you happen to know of data that was revision-controlled in a better fashion, let us know.

jcohenadad:
> binaries: binaries_centos, binaries_debian, binaries_linux, binaries_osx
-> won't touch for now

change of decision? https://github.com/sct-data/binaries_linux

if so, we need to make sure to liaise with @kousu's devs: https://github.com/neuropoly/spinalcordtoolbox/pull/2668

zougloub:
@jcohenadad no, that was just a test. I have some time tonight to work on this.

You can check PAM50.
Conversion code is at: https://github.com/sct-data/sct-data-archeology

zougloub:
https://github.com/sct-data/sct_testing_data & https://github.com/sct-data/sct_example_data were updated with addition of the zip history on top of the original git history.

I opted to name the auto-generated tags & releases `r${date}`, which would separate from `v${rev}` if next releases change versioning scheme to something more standard.

zougloub:
@saulo-p can you write a script to help automate deployment of further sct-data work?

It would take a repo key name (eg. sct_testing_data) and tag name and does:
- Create in the GH repo a release from the tag
- Download default release artifact (.zip)
- Upload to OSF, get OSF download link (check with @jcohenadad / @alexfoias for creds, get from os.environ or getpass)
- Upload to github as manually added asset
- Update release description to include OSF mirror link
- Print GH and OSF URLs, so they can be integrated in sct_download_data.py

zougloub:
@saulo-p review of  https://github.com/sct-data/sct-data-archeology & the sct-data repos is welcome.

zougloub:
@kousu idem

saulo-p:
> @saulo-p review of https://github.com/sct-data/sct-data-archeology & the sct-data repos is welcome.

Since I'm using a good chunk of your code to write the script you requested above, I'll leave a comment if I find something odd.

zougloub:
Status:
- [x] Convert SCT data into github repos (under sct-data and ivadomed organizations)
- [x] Update URIs (in this PR's code) to link to github releases
- [ ] Document release flow

zougloub:
Release tooling applicable to https://github.com/ivadomed/ivadomed/issues/269

scripts/sct_download_data.py
```diff
@@ -52,43 +52,85 @@ def get_parser(dict_url):
...
-                              'https://www.neuro.polymtl.ca/_media/downloads/sct/20180610_deepseg_sc_models.zip'],
-        'deepseg_lesion_models': ['https://osf.io/eg7v9/?action=download',
-                                  'https://www.neuro.polymtl.ca/_media/downloads/sct/20180613_deepseg_lesion_models'
-                                  '.zip'],
-        'c2c3_disc_models': ['https://osf.io/t97ap/?action=download',
-                             'https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip']
-        }
+        "sct_example_data": [
+            "https://github.com/sct-data/sct_example_data/releases/download/r20180525/20180525_sct_example_data.zip",
+            "https://osf.io/kjcgs/?action=download",
```
jcohenadad, line 61: do you recommend we keep maintaining mirrors? it's additional workload (usually on my shoulders)

zougloub: The idea would be to have some code that can do at least the OSF uploads automatically, and provide the list of primary + mirror URLs so they can be put in the code.

zougloub:
The build fails because `spinalcordtoolbox.download` extracts bundles as is, which is causing a problem: it is standard for zipballs / tarballs to have a root directory named like `${product}-${version}`, in which case we want to step into it to get the data, and then put it in a destination folder named with the unversioned bundle name.

-> Will update the downloading code.

The ensuing test failure lead to opening https://github.com/ivadomed/ivadomed/issues/330.

zougloub:
To release a repo it becomes:

```sh
python -m distriploy 
```
```json
["https://github.com/exmakhina/distriploy/releases/download/v0.14-1-g5bda869/distriploy-v0.14-1-g5bda869.zip", "https://osf.io/download/5efc1414ea59af0023d36d11/", "https://exmakhina.com/distriploy/distriploy-v0.14-1-g5bda869.zip", "https://academictorrents.com/download/53d3f1a237581ce2f8b6ccd3aac92ade84f1603c.torrent", "magnet:?xt=urn:btih:53d3f1a237581ce2f8b6ccd3aac92ade84f1603c&tr=http%3A%2F%2Facademictorrents.com%2Fannounce.php&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337%2Fannounce&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969"]
```

jcohenadad:
@zougloub ok to move https://github.com/exmakhina/distriploy under https://github.com/neuropoly/distriploy?

zougloub:
> @zougloub ok to move https://github.com/exmakhina/distriploy under https://github.com/neuropoly/distriploy?

Done.

zougloub:
Currently struggling with the current Travis failure, which is not happening on my end.

spinalcordtoolbox/download.py
```diff
@@ -106,42 +106,80 @@ def unzip(compressed, dest_folder):
...
-        if os.path.isdir(fullpath_dest):
-            logger.warning("Folder {} already exists. Removing it...".format(data_extracted_name))
-            shutil.rmtree(fullpath_dest)
-
-    # Destination path
-    # for source_path in extracted_files_paths:
-        # Copy the content of source to destination (and create destination folder)
-    distutils.dir_util.copy_tree(dest_tmp_folder, dest_folder)
+
+    os.makedirs(dest_folder, exist_ok=True)
```
jcohenadad, line 153: this change of behaviour might become a problem. I do remember that in the past, we did not remove the folder entirely, and remaining legacy files were conflicting with SCT (e.g. functions doing a glob.glob()). I know this is not a proper way of coding, but this is what it is (at least for now), so if we want to make sure to not break anything, why not simply removing the folder if it already exists?

zougloub:
> i just have one comment about not removing existing folder (i feel quite strongly about this-- curious to hear your arguments)-- otherwise it looks great +1

Before, the code was only removing stuff that would be present in the archive.
Somehow I kept this logic.
But now that we provide the full path (not only the parent path) there is no reason to keep dangling files around. We can totally clean the destination folder.
I will add a warning in the command-line tool.

jcohenadad:
> Before, the code was only removing stuff that would be present in the archive.
> Somehow I kept this logic.

i don't think this is true. See this experiment below, where i manually added the file `PAM50_atlas_37.nii.gz`, which is erased after downloading PAM50 (current behaviour, from master 7d757ae0e99bf6a1452900ae75b4e34e30b849b1):

<details><summary></summary>

~~~
julien-macbook:~/temp/sct-2733/data/PAM50/atlas $ ll
total 11704
drwxr-xr-x  41 julien  staff    1312  7 Jul 10:36 .
drwxr-xr-x   7 julien  staff     224  7 Jul 10:36 ..
-rw-r--r--   1 julien  staff  158043  7 Jul 10:36 PAM50_atlas_00.nii.gz
-rw-r--r--   1 julien  staff  153015  7 Jul 10:36 PAM50_atlas_01.nii.gz
-rw-r--r--   1 julien  staff  186671  7 Jul 10:36 PAM50_atlas_02.nii.gz
-rw-r--r--   1 julien  staff  186673  7 Jul 10:36 PAM50_atlas_03.nii.gz
-rw-r--r--   1 julien  staff  179939  7 Jul 10:36 PAM50_atlas_04.nii.gz
-rw-r--r--   1 julien  staff  179647  7 Jul 10:36 PAM50_atlas_05.nii.gz
-rw-r--r--   1 julien  staff  173488  7 Jul 10:36 PAM50_atlas_06.nii.gz
-rw-r--r--   1 julien  staff  173442  7 Jul 10:36 PAM50_atlas_07.nii.gz
-rw-r--r--   1 julien  staff  145783  7 Jul 10:36 PAM50_atlas_08.nii.gz
-rw-r--r--   1 julien  staff  145979  7 Jul 10:36 PAM50_atlas_09.nii.gz
-rw-r--r--   1 julien  staff   92449  7 Jul 10:36 PAM50_atlas_10.nii.gz
-rw-r--r--   1 julien  staff   92467  7 Jul 10:36 PAM50_atlas_11.nii.gz
-rw-r--r--   1 julien  staff  171333  7 Jul 10:36 PAM50_atlas_12.nii.gz
-rw-r--r--   1 julien  staff  171519  7 Jul 10:36 PAM50_atlas_13.nii.gz
-rw-r--r--   1 julien  staff  130847  7 Jul 10:36 PAM50_atlas_14.nii.gz
-rw-r--r--   1 julien  staff  130929  7 Jul 10:36 PAM50_atlas_15.nii.gz
-rw-r--r--   1 julien  staff  109757  7 Jul 10:36 PAM50_atlas_16.nii.gz
-rw-r--r--   1 julien  staff  109851  7 Jul 10:36 PAM50_atlas_17.nii.gz
-rw-r--r--   1 julien  staff  117504  7 Jul 10:36 PAM50_atlas_18.nii.gz
-rw-r--r--   1 julien  staff  117491  7 Jul 10:36 PAM50_atlas_19.nii.gz
-rw-r--r--   1 julien  staff  114442  7 Jul 10:36 PAM50_atlas_20.nii.gz
-rw-r--r--   1 julien  staff  114494  7 Jul 10:36 PAM50_atlas_21.nii.gz
-rw-r--r--   1 julien  staff  116855  7 Jul 10:36 PAM50_atlas_22.nii.gz
-rw-r--r--   1 julien  staff  117706  7 Jul 10:36 PAM50_atlas_23.nii.gz
-rw-r--r--   1 julien  staff  118155  7 Jul 10:36 PAM50_atlas_24.nii.gz
-rw-r--r--   1 julien  staff  118294  7 Jul 10:36 PAM50_atlas_25.nii.gz
-rw-r--r--   1 julien  staff  105106  7 Jul 10:36 PAM50_atlas_26.nii.gz
-rw-r--r--   1 julien  staff  105027  7 Jul 10:36 PAM50_atlas_27.nii.gz
-rw-r--r--   1 julien  staff  103840  7 Jul 10:36 PAM50_atlas_28.nii.gz
-rw-r--r--   1 julien  staff  103466  7 Jul 10:36 PAM50_atlas_29.nii.gz
-rw-r--r--   1 julien  staff  151696  7 Jul 10:36 PAM50_atlas_30.nii.gz
-rw-r--r--   1 julien  staff  151387  7 Jul 10:36 PAM50_atlas_31.nii.gz
-rw-r--r--   1 julien  staff  166632  7 Jul 10:36 PAM50_atlas_32.nii.gz
-rw-r--r--   1 julien  staff  167459  7 Jul 10:36 PAM50_atlas_33.nii.gz
-rw-r--r--   1 julien  staff  125958  7 Jul 10:36 PAM50_atlas_34.nii.gz
-rw-r--r--   1 julien  staff  125912  7 Jul 10:36 PAM50_atlas_35.nii.gz
-rw-r--r--   1 julien  staff  499817  7 Jul 10:36 PAM50_atlas_36.nii.gz
-rw-r--r--   1 julien  staff  499817  7 Jul 10:30 PAM50_atlas_37.nii.gz
-rw-r--r--   1 julien  staff    2804  7 Jul 10:36 info_label.txt
julien-macbook:~/temp/sct-2733/data/PAM50/atlas $ cd ..
julien-macbook:~/temp/sct-2733/data/PAM50 $ cd ..
julien-macbook:~/temp/sct-2733/data $ sct_download_data -d PAM50

--
Spinal Cord Toolbox (git-master-7d757ae0e99bf6a1452900ae75b4e34e30b849b1)

Trying URL: https://osf.io/u79sr/download
Downloading: 20191029_PAM50.zip
Status: 100%|██████████████████████████████████████████████████████████████████████████████████████████████| 70.8M/70.8M [00:38<00:00, 1.85MB/s]
Creating temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20200707120323.350167-wy0hdwl8)
Unzip data to: /var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20200707120323.350167-wy0hdwl8
Destination folder: /Users/julien/temp/sct-2733/data
Checking if folder already exists...
Folder PAM50 already exists. Removing it...
Removing temporary folders...
Done!

julien-macbook:~/temp/sct-2733/data $ ll PAM50/atlas/
total 10720
drwxr-xr-x  40 julien  staff    1280  7 Jul 12:03 .
drwxr-xr-x   6 julien  staff     192  7 Jul 12:03 ..
-rw-r--r--   1 julien  staff  158043  7 Jul 12:03 PAM50_atlas_00.nii.gz
-rw-r--r--   1 julien  staff  153015  7 Jul 12:03 PAM50_atlas_01.nii.gz
-rw-r--r--   1 julien  staff  186671  7 Jul 12:03 PAM50_atlas_02.nii.gz
-rw-r--r--   1 julien  staff  186673  7 Jul 12:03 PAM50_atlas_03.nii.gz
-rw-r--r--   1 julien  staff  179939  7 Jul 12:03 PAM50_atlas_04.nii.gz
-rw-r--r--   1 julien  staff  179647  7 Jul 12:03 PAM50_atlas_05.nii.gz
-rw-r--r--   1 julien  staff  173488  7 Jul 12:03 PAM50_atlas_06.nii.gz
-rw-r--r--   1 julien  staff  173442  7 Jul 12:03 PAM50_atlas_07.nii.gz
-rw-r--r--   1 julien  staff  145783  7 Jul 12:03 PAM50_atlas_08.nii.gz
-rw-r--r--   1 julien  staff  145979  7 Jul 12:03 PAM50_atlas_09.nii.gz
-rw-r--r--   1 julien  staff   92449  7 Jul 12:03 PAM50_atlas_10.nii.gz
-rw-r--r--   1 julien  staff   92467  7 Jul 12:03 PAM50_atlas_11.nii.gz
-rw-r--r--   1 julien  staff  171333  7 Jul 12:03 PAM50_atlas_12.nii.gz
-rw-r--r--   1 julien  staff  171519  7 Jul 12:03 PAM50_atlas_13.nii.gz
-rw-r--r--   1 julien  staff  130847  7 Jul 12:03 PAM50_atlas_14.nii.gz
-rw-r--r--   1 julien  staff  130929  7 Jul 12:03 PAM50_atlas_15.nii.gz
-rw-r--r--   1 julien  staff  109757  7 Jul 12:03 PAM50_atlas_16.nii.gz
-rw-r--r--   1 julien  staff  109851  7 Jul 12:03 PAM50_atlas_17.nii.gz
-rw-r--r--   1 julien  staff  117504  7 Jul 12:03 PAM50_atlas_18.nii.gz
-rw-r--r--   1 julien  staff  117491  7 Jul 12:03 PAM50_atlas_19.nii.gz
-rw-r--r--   1 julien  staff  114442  7 Jul 12:03 PAM50_atlas_20.nii.gz
-rw-r--r--   1 julien  staff  114494  7 Jul 12:03 PAM50_atlas_21.nii.gz
-rw-r--r--   1 julien  staff  116855  7 Jul 12:03 PAM50_atlas_22.nii.gz
-rw-r--r--   1 julien  staff  117706  7 Jul 12:03 PAM50_atlas_23.nii.gz
-rw-r--r--   1 julien  staff  118155  7 Jul 12:03 PAM50_atlas_24.nii.gz
-rw-r--r--   1 julien  staff  118294  7 Jul 12:03 PAM50_atlas_25.nii.gz
-rw-r--r--   1 julien  staff  105106  7 Jul 12:03 PAM50_atlas_26.nii.gz
-rw-r--r--   1 julien  staff  105027  7 Jul 12:03 PAM50_atlas_27.nii.gz
-rw-r--r--   1 julien  staff  103840  7 Jul 12:03 PAM50_atlas_28.nii.gz
-rw-r--r--   1 julien  staff  103466  7 Jul 12:03 PAM50_atlas_29.nii.gz
-rw-r--r--   1 julien  staff  151696  7 Jul 12:03 PAM50_atlas_30.nii.gz
-rw-r--r--   1 julien  staff  151387  7 Jul 12:03 PAM50_atlas_31.nii.gz
-rw-r--r--   1 julien  staff  166632  7 Jul 12:03 PAM50_atlas_32.nii.gz
-rw-r--r--   1 julien  staff  167459  7 Jul 12:03 PAM50_atlas_33.nii.gz
-rw-r--r--   1 julien  staff  125958  7 Jul 12:03 PAM50_atlas_34.nii.gz
-rw-r--r--   1 julien  staff  125912  7 Jul 12:03 PAM50_atlas_35.nii.gz
-rw-r--r--   1 julien  staff  499817  7 Jul 12:03 PAM50_atlas_36.nii.gz
-rw-r--r--   1 julien  staff    2804  7 Jul 12:03 info_label.txt
julien-macbook:~/temp/sct-2733/data $ 
~~~

</details>

spinalcordtoolbox/download.py
zougloub: Code was updated to remove the destination directory.

zougloub:
It was partly true... proof is that we are now wiping data from bin/ if I remove the destination directory ;)

zougloub:
Should I add an argument to the script / function, or revert to a conservative behaviour, or a hack?

jcohenadad:
> Should I add an argument to the script / function, or revert to a conservative behaviour, or a hack?

How about setting the default -o to ./<dataname> ? (Current behaviour)

zougloub:
> How about setting the default -o to `./<dataname>`? (Current behaviour)

The behavior in master is:

- Unpack whatever is in the archive to ./
- From there, remove directories that would be in the newly extracted archive

The new proposed behavior was:
- Unpack archive to `./<dataname>`
- Remove things that would be overwritten by the archive

Now it seems there might be a concern with files that would be in the old archive, but not in the new archive. So you're proposing to wipe everything in the destination directory. However this doesn't work for `binaries_linux` which is *adding* files to `${SCT_DIR}/bin`, which is now wiped.

I would propose, to use your "remove everything" behaviour but to have in the SCT installer, the binaries download to install files to a temporary directory then move everything to `${SCT_DIR}/bin`... does that work for you?

spinalcordtoolbox/download.py
```diff
@@ -104,12 +104,13 @@ def unzip(compressed, dest_folder):
...
 
 
-def install_data(url, dest_folder):
+def install_data(url, dest_folder, keep=False):
     """
     Download a data bundle from a URL and install in the destination folder.
 
     :param url: URL or sequence thereof (if mirrors).
-    :param dest_folder: destination directory. Will be cleaned if exists.
+    :param dest_folder: destination directory. Will be cleaned if exists,
```
jcohenadad, line 112: great solution!

zougloub:
Updated code to give option to keep existing data, so bin/ special case works.

spinalcordtoolbox/download.py
```diff
@@ -104,44 +104,101 @@ def unzip(compressed, dest_folder):
...
+         The right thing to do in this case is to take the files from there and install them
+         in `${dst}`.
+
+    - Uses `download_data()` to retrieve the data.
+    - Uses `unzip()` to extract the bundle.
     """
 
-    # Download data
+    if not keep and os.path.exists(dest_folder):
+        shutil.rmtree(dest_folder)
```
jcohenadad, line 143: how about keeping this warning? 
https://github.com/neuropoly/spinalcordtoolbox/pull/2733/files#diff-af0f1afd831de3f57ce82bbbca494094L134

zougloub: > how about keeping this warning?
> https://github.com/neuropoly/spinalcordtoolbox/pull/2733/files#diff-af0f1afd831de3f57ce82bbbca494094L134

Added back

PR #2734, opened 2020-05-25
Implemented QC report for sct_label_utils

### LINKED ISSUE #2703: Generate QC feature for disc labeling
The idea is to implement QC feature for disc labeling based on [sct_qc.py](https://github.com/neuropoly/spinalcordtoolbox/blob/4c2aca38d2536e202d58b9be646652a0f4280b6a/scripts/sct_qc.py) and [qc.py](https://github.com/neuropoly/spinalcordtoolbox/blob/637d0f2a0ff476389257309a69adc64cad36926c/spinalcordtoolbox/reports/qc.py).

alexfoias:
Fixes #2703


alexfoias:
The PR is ready for review. 

The QC looks something like this:
Screen Shot 2020-06-01 at 2 03 05 PM

If you want to try it you can do:
```
sct_qc -i /Volumes/sct_testing/large/sub-amuPAM50002/anat/sub-amuPAM50002_T1w.nii.gz -s /Volumes/sct_testing/large/derivatives/labels/sub-amuPAM50002/anat/sub-amuPAM50002_T1w_labels-disc-manual.nii.gz -p sct_label_utils
```

scripts/sct_qc.py
```diff
@@ -31,7 +31,7 @@ def get_parser():
                         help='SCT function associated with the QC report to generate',
                         choices=('sct_propseg', 'sct_deepseg_sc', 'sct_deepseg_gm', 'sct_register_multimodal',
                                  'sct_register_to_template', 'sct_warp_template', 'sct_label_vertebrae',
-                                 'sct_detect_pmj'),
+                                 'sct_detect_pmj', 'sct_label_utils'),
                         required=True)
     parser.add_argument('-s',
                         metavar='SEG',
```
jcohenadad, line 37: we probably want to change "help='Input segmentation'," for "help='Input segmentation or label',"

alexfoias:
> * It would be nice to have the value of the label appear next to it
Do you mean pixel intensity for "the value" ?
> * Why didn't you use a vectorial matplotlib object (e.g. a dot), instead of using overlaid pixels?
I could implement the dot/circle considering the coordinates of the pixel location of the mask.

jcohenadad:
> > * It would be nice to have the value of the label appear next to it
> 
> Do you mean pixel intensity for "the value" ?

yes

> > * Why didn't you use a vectorial matplotlib object (e.g. a dot), instead of using overlaid pixels?
> 
> I could implement the dot/circle considering the coordinates of the pixel location of the mask.

no, what i meant was: instead of overlaying an actual pixelated image, use the vectorized display (if that's even possible?)

alexfoias:
Something funky it happens to some labels for example **sub-amuPAM50002_T2w_labels-disc-manual.nii.gz**

Screen Shot 2020-06-02 at 3 09 58 PM
While it renders well for **sub-amuPAM50002_T1w_labels-disc-manual.nii.gz**. 
Screen Shot 2020-06-02 at 3 09 51 PM

My understanding is that current implementation of QC looks into the mask (2d image) on the middle slice, I think the labels are in a different slice.

@jcohenadad @lrouhier is it a problem with the labels ? How should I approach this problem ?

jcohenadad:
> @jcohenadad @lrouhier is it a problem with the labels ? How should I approach this problem ?

how do you retrieve the label coordinates to display it? if you only grab the AP and SI coordinates (sagittal orientation), then you don't care about where the label is defined in the RL coordinates, right? Or am I missing something?

alexfoias:
I do `results_mask_pixels = np.where(mask > 0)` to find the coordinates, but this give me a 2D result. Maybe I should do a recursive np.where or for the complete 3D volume.

jcohenadad:
> I do `results_mask_pixels = np.where(mask > 0)` to find the coordinates, but this give me a 2D result. Maybe I should do a recursive np.where or for the complete 3D volume.

can you point us to the code?

alexfoias:
Here is the line : https://github.com/neuropoly/spinalcordtoolbox/pull/2734/files#diff-cd0ed16d3a599345c58ea8bbd2fe675dR183

jcohenadad:
> Here is the line : https://github.com/neuropoly/spinalcordtoolbox/pull/2734/files#diff-cd0ed16d3a599345c58ea8bbd2fe675dR183
> 
> I even added a line `print(mask.ndim)` and I get 2, that means it's not passing the complete image.

`mask` is already processed by the `qcslice.Sagittal()` method, and it tries to find a center of mass [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/reports/slice.py#L377), which in your case is not relevant. You probably want to consider adding a method under the `Sagittal()` class that outputs the XY coordinates of interest. Does that make sense?

alexfoias:
@jcohenadad thank you so much for the tip. I'll look into it.

alexfoias:
@jcohenadad I tried to understand and implement the retrieve of the mask coordinates using the `Sagittal()` method, but I'm not familiar enough with the complete implementation of the class & method.

@lrouhier did you use these methods  ? are you more familiar with them ? do you think you can help me ?

lrouhier:
I didn't use it previously but I might be able to help you. I will look into it!

alexfoias:
@lrouhier you can use:

`sct_qc -i /Volumes/sct_testing/large/sub-amuPAM50002/anat/sub-amuPAM50002_T1w.nii.gz -s /Volumes/sct_testing/large/derivatives/labels/sub-amuPAM50002/anat/sub-amuPAM50002_T1w_labels-disc-manual.nii.gz -p sct_label_utils`

But the problem is that the current implementation fails for ` sub-amuPAM50002_T2w_labels-disc-manual.nii.gz` as it showed [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2734#issuecomment-637754548).

lrouhier:
One of the solution is to figure out a condition to put [here](https://github.com/neuropoly/spinalcordtoolbox/blob/09a47b2bed7f13d04bd01a962948883f6ccb7f38/spinalcordtoolbox/reports/slice.py#L389)

I am currently trying with: 
~~~
elif np.argwhere(image.data).shape[0] < 25:
            print(np.argwhere(image.data))
            return [np.argwhere(image.data)[0][2]] * image.data.shape[2]
~~~
which seems to work but poses two problems:

* if all label are not on the same slice we will have a problem --> we will need to create a function to project all label in one slice (I can do that) 
* If you do a spinal segmentation of a small image (let's say 2 or 3 slices you might have less than 25 pixels in the mask) --> we might need to check if the label are "discrete" in the IS direction.

spinalcordtoolbox/reports/qc.py
```diff
@@ -641,12 +660,19 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
...
         plane = 'Sagittal'
         dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
         qcslice_type = qcslice.Sagittal([Image(fname_in1), Image(fname_seg)], p_resample=None)
         qcslice_operations = [QcImage.label_vertebrae]
         qcslice_layout = lambda x: x.single()
+    #  Sagittal orientation, display posterior labels
+    elif process in ['sct_label_utils']:
+        plane = 'Sagittal'
+        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
+        qcslice_type = qcslice.Sagittal([Image(fname_in1), Image(fname_seg)], p_resample=None)
```
lrouhier, line 673: We could perform the projection here (into the relevant slice) And transforms this. This could look like:
~~~
elif process in ['sct_label_utils']:
        plane = 'Sagittal'
        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
        projected_image = projection (Image(fname_seg))
        qcslice_type = qcslice.Sagittal([Image(fname_in1), projected_image], p_resample=None)
        qcslice_operations = [QcImage.label_utils]
        qcslice_layout = lambda x: x.single()

jcohenadad:
IIUC the problem described in https://github.com/neuropoly/spinalcordtoolbox/pull/2734#issuecomment-637754548 is when the label is on an other slice than the medial slice (which is showed). Why not simply grab the I-S and A-P coordinate, and display the dot with these coordinates?

lrouhier:
I think I misunderstood your [previous comment](https://github.com/neuropoly/spinalcordtoolbox/pull/2734#issuecomment-637787455). I thought that Qcimage() methods (used [here] (https://github.com/neuropoly/spinalcordtoolbox/blob/bc4c0e99e557cc4f654d2d919dd47ee86f9fa0f4/spinalcordtoolbox/reports/qc.py#L674) ) was expecting a 2D image produced by the slice method previously. Therefore we can create this image in slice() (by adding a projection method) or generate it with another function.

jcohenadad:
> Therefore we can create this image in slice()

which image, and what is "slice()"?

> (by adding a projection method)

why do you want to do a projection?

lrouhier:
>which image, and what is "slice()"?
Sorry if that was not clear I meant Sagittal method, which is used as a slicer (if I understand the code correctly). The created image would be a 3D image with all the label on the same slice . 

>why do you want to do a projection?
Because this will allow us to have the same workflow as before, we just modify the fname_seg image.

jcohenadad:
> > which image, and what is "slice()"?
> > Sorry if that was not clear I meant Sagittal method, which is used as a slicer (if I understand the code correctly). The created image would be a 3D image with all the label on the same slice .
> 
> > why do you want to do a projection?
> > Because this will allow us to have the same workflow as before, we just modify the fname_seg image.

ok got it. yes, it makes sense (i think)

jcohenadad:
ok, this implementation is needed urgently (see https://github.com/spine-generic/spine-generic/pull/146)

who's on it? @lrouhier @alexfoias?

lrouhier:
I Added the projection in qc not sure if it is the best option (WIP)

jcohenadad:
@alexfoias @lrouhier i'm onto something. i'll find another way to support sct_label_utils with the QC, without the need to do the projection. stay tuned

jcohenadad:
Example of QC output (with e7f1809fe13aa8174464d04f2531ee2588a80a69):

Screen Shot 2020-07-23 at 10 32 53 PM

alexfoias:
Tested  os `sct_testing/large` and it works well.

PR #2748, opened 2020-06-19
Accommodate a cascade of deep learning models

### LINKED ISSUE #2744: sct_deepseg: implement compatibility with 2-steps pipeline
### Current workflow:
```
sct_deepseg -i image.nii.gz -m modelZ
```
1. Get model's path (already installed or download it): [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg.py#L143). Models being listed [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L20).
2. Call `segment_nifti` [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/core.py#L88) which returns the name of the produced segmentation.

### Issue:
The current implementation only allows to run **one** model. In some cases, we would like to cascade more than one model. For example: 
- model 1: input: image, output: cord segmentation
- model 2: input: image + cord segmentation, output: lesion segmentation. 

How can we refactor the code to enable this feature?

### Proposition
1. Convert the URL of the model location to a list, [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L22)
2. When installing the model [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg.py#L151), follow: "pipeline/model_stepX", compatible with 1-step or N-step approaches. Example:
```
t2_tumor/
    model_detection/  # step 1, first URL in the list
        model_detection.onnx
        model_detection.json
    model_segmentation/  # step 2, second URL in the list
        model_segmentation.onnx
        model_segmentation.json
```
3. So we will have a list of [path_model](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg.py#L154) eg `["t2_tumor/model_detection", "t2_tumor/model_segmentation"]`. Warning: the order of the models in this list matters. It allows, for each iteration (ie each model), to load the model and metadata (json file) that are specific for this step.
4. We loop across this list of `path_model` and call `segment_nifti`, [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg.py#L160). After running the first model, we add the first_model_output to `segment_nifti` inputs (in [params](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/core.py#L94) or as an additional parameter to be defined, eg fname_roi).
5. The additional input (ie the output of the first step) is used by `segment_volume` ([here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/core.py#L97)) to perform the second step of the pipeline: eg via `fname_roi` parameter [here](https://github.com/neuropoly/ivado-medical-imaging/blob/master/ivadomed/utils.py#L343).

charleygros:
### Context:
In its current implementation, `sct_deepseg` can only run one deep learning model for the specified input. The purpose of this PR is to make it possible to run a cascade of models, e.g., the output of model 1 would go through model 2, etc. This would enable, for example, the identification of the cord (first step), and then a precise segmentation of the lesions (second step). 

### Done:
- Add TASKS to `models.py` following [this model](https://github.com/neuropoly/spinalcordtoolbox/issues/2744#issuecomment-645018192)
- Add tasks related parameters to parser:
    - `-task`
    - `-list-tasks`
    - `-install-task`
- List tasks
- Install task

### Todo:
XX

Fixes #2744.


charleygros:
Illustration of `-list-tasks` and `-install-task`:

1. Default:
Screenshot 2020-06-19 19:10:47

2. `sct_deepseg -install-model t2_tumor`
Screenshot 2020-06-19 19:30:45

3. `sct_deepseg -install-task segment_mice_sc`
Screenshot 2020-06-19 19:34:33

charleygros:
In order to be robust to these scenarii:
```
# existing models: the task "segment-cord" will use models: [findcord, segmentcord]
sct_deepseg -task segment-cord
# to replace first model by another official model:
sct_deepseg -model findcord_another -task segmentcord 
# to replace first model by a custom local model:
sct_deepseg -model path_to_my_model/ -task segmentcord 
# to use model instead of task
sct_deepseg -model path_to_my_model/
```

```python
    # Get pipeline model names
    if 'task' in args:
        if 'model' in args:
            # Replace the first model of task by the -model argument
            name_models = [args['model']] + sct.deepseg.models.TASKS[args['install_task']]['models'].pop(0)
        else:
            name_models = sct.deepseg.models.TASKS[args['install_task']]['models']
    else:
        name_models = [args['model']]

    # Run pipeline by iterating through the models
    fname_prior = None
    for name_model in name_models:
        ...
```

charleygros:
Add `fname_prior` to `segment_nifti`:
```python
def segment_nifti(fname_image, folder_model, fname_prior=None, param={}):
    """
    Segment a nifti file.

    :param fname_image: str: Filename of the image to segment.
    :param folder_model: str: Folder that encloses the deep learning model.
    :param fname_prior: str: Filename of a previous segmentation that is used here as a prior.
    :param param: dict: Dictionary of user's parameter
    :return: fname_out: str: Output filename. If directory does not exist, it will be created.
    """
    nii_seg = imed.utils.segment_volume(folder_model, fname_image, fname_prior)
```

charleygros:
@jcohenadad & @andreanne-lemay : I am not    # Save output seg
    if 'o' in options:
        fname_out = options['o']
    else:
        fname_out = ''.join([sct.utils.splitext(fname_image)[0], '_seg.nii.gz']) two sure of two things:

1. File resulting from prior steps (ie non final):
Do we want to save them or in a temp folder that we delete after? I would say to save them along with the final segmentation. Or at least make it possible via a new parameter.
How to name these files? ie which suffix? using the model name? or `_seg-tmp01`?

2. Some post processing parameters are specified in the model metadata, while some are specified via the parser: how are we using them for the tmp files?
I would suggest:
- for the temp segmentation: apply only the postprocessing that is specified in the model metadata
- for the final segmentaiton: apply the postprocessing that is specified in the parser, ie update the model metadata with the parser values, like currently done here:
```python
options = {**DEFAULTS, **metadata, **param}
```

andreanne-lemay:
> ```python
>     # Get pipeline model names
>     if 'task' in args:
>         if 'model' in args:
>             # Replace the first model of task by the -model argument
>             name_models = [args['model']] + sct.deepseg.models.TASKS[args['install_task']]['models'].pop(0)
>         else:
>             name_models = sct.deepseg.models.TASKS[args['install_task']]['models']
>     else:
>         name_models = [args['model']]
> 
>     # Run pipeline by iterating through the models
>     for name_model in name_models:
>         ...
> ```
> 
> In order to be robust to these scenarii:
> 
> ```
> # existing models: the task "segment-cord" will use models: [findcord, segmentcord]
> sct_deepseg -task segment-cord
> # to replace first model by another official model:
> sct_deepseg -model findcord_another -task segmentcord 
> # to replace first model by a custom local model:
> sct_deepseg -model path_to_my_model/ -task segmentcord 
> # to use model instead of task
> sct_deepseg -model path_to_my_model/
> ```

I find it a bit confusing to use -model parameter and -task to use custom models. Also, if we want to change second model instead of first? According to the [issue](https://github.com/neuropoly/spinalcordtoolbox/issues/2744#issuecomment-645018192), we agreed on this form:

# existing models: the task "segment-cord" will use models: [findcord, segmentcord]
sct_deepseg -task segment-cord
# to replace first model by another official model:
sct_deepseg -model findcord_another segmentcord 
# to replace first model by a custom local model:
sct_deepseg -model path_to_my_model/ segmentcord 

Do you agree to stick to this form?

andreanne-lemay:
> @jcohenadad & @andreanne-lemay : I am not too sure of two things:
> 
> 1. File resulting from prior steps (ie non final):
>    Do we want to save them or in a temp folder that we delete after? I would say to save them along with the final segmentation. Or at least make it possible via a new parameter.
>    How to name these files? ie which suffix? using the model name? or `_seg-tmp01`?
>    Currently we have:
> 
> ```python
> # Save output seg
>     if 'o' in options:
>         fname_out = options['o']
>     else:
>         fname_out = ''.join([sct.utils.splitext(fname_image)[0], '_seg.nii.gz']) 
> ```
> 
> 1. Some post processing parameters are specified in the model metadata, while some are specified via the parser: how are we using them for the tmp files?
>    I would suggest:
> 
> * for the temp segmentation: apply only the postprocessing that is specified in the model metadata
> * for the final segmentaiton: apply the postprocessing that is specified in the parser, ie update the model metadata with the parser values, like currently done here:
> 
> ```python
> options = {**DEFAULTS, **metadata, **param}
> ```

1. I vote for the option to add `-r`parameter to remove the temporary files like it is done on other functions set by default to 1. As for the suffix, good question. `seg` is too specific I think since it might not always be a segmentation (ex: classification). The model name seems to be a good option to me. In case of custom models, we can use the folder name ("path/to/model.split("/")[-1]). 

2. I agree with your suggestion

scripts/sct_deepseg.py
```diff
@@ -46,6 +46,21 @@ def get_parser():
              "suffix '_seg' will be added and output extension will be .nii.gz.",
         metavar=sct.utils.Metavar.str)
 
+    seg = parser.add_argument_group('\nTASKS')
+    seg.add_argument(
+        "-task",
+        help="Task to perform. It could either be an official SCT task (in that case, simply enter the name of the "
+             "task, example: -task segment_t2star_sc). To list official tasks, run: sct_deepseg -list-tasks.",
```
jcohenadad, line 53: sorry i didn't deal with PEP8 carriage return in this comment
```suggestion
        help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task. To list available tasks, run: sct_deepseg -list-tasks",
```

scripts/sct_deepseg.py
```diff
@@ -130,34 +145,82 @@ def main():
                 default[True]))
         exit(0)
 
+    # Deal with task
+    if args['list_tasks']:
```
jcohenadad, line 149: looks a lot like the code for list-models-- we should probably make it a function

jcohenadad:
> @jcohenadad & @andreanne-lemay : I am not too sure of two things:
> 
> 1. File resulting from prior steps (ie non final):
>    Do we want to save them or in a temp folder that we delete after? I would say to save them along with the final segmentation. Or at least make it possible via a new parameter.
>    How to name these files? ie which suffix? using the model name? or `_seg-tmp01`?

i would start simple, and as the need is expressed (by users or by us), we will implement the flag. Meaning: i suggest we only output the final seg for now.

> 1. Some post processing parameters are specified in the model metadata, while some are specified via the parser: how are we using them for the tmp files?
>    I would suggest:
> 
> * for the temp segmentation: apply only the postprocessing that is specified in the model metadata
> * for the final segmentaiton: apply the postprocessing that is specified in the parser, ie update the model metadata with the parser values, like currently done here:
> 
> ```python
> options = {**DEFAULTS, **metadata, **param}
> ```

that sounds like a reasonable approach to me.

jcohenadad:
> I find it a bit confusing to use -model parameter and -task to use custom models. Also, if we want to change second model instead of first? According to the issue, we agreed on this form:

# existing models: the task "segment-cord" will use models: [findcord, segmentcord]
sct_deepseg -task segment-cord
# to replace first model by another official model:
sct_deepseg -model findcord_another segmentcord 
# to replace first model by a custom local model:
sct_deepseg -model path_to_my_model/ segmentcord 
Do you agree to stick to this form?

i agree with @andreanne-lemay

charleygros:
> > I find it a bit confusing to use -model parameter and -task to use custom models. Also, if we want to change second model instead of first? According to the issue, we agreed on this form
> 
> i agree with @andreanne-lemay

I agree with that :-) I think we should change this. (I thought that using `-task` would make it more explicit but .. nah)
I am not too sure how to deal with " Also, if we want to change second model instead of first?" except by running the command twice?

jcohenadad:
> I am not too sure how to deal with " Also, if we want to change second model instead of first?" except by running the command twice?

Not sure if I understand correctly, but to me, using -model enables any possible combination in a one-liner:

-model model1 model2
-model model2 model1

charleygros:
> > I am not too sure how to deal with " Also, if we want to change second model instead of first?" except by running the command twice?
> 
> Not sure if I understand correctly, but to me, using -model enables any possible combination in a one-liner:
> 
> -model model1 model2
> -model model2 model1
> -model model1custom model2

Oooo ! I got it now, my bad!

andreanne-lemay:
I modified the `task` vs `model` option as discussed. The user have the choice to either select a task or a model (or list of models). If a user uses `-task` and `-model` at the same time, only the model arguments are considered and `-task` and its argument is ignored.

I moved the code of -list-models and -list-tasks in functions.

I updated t2_tumor model to the new ivadomed refactoring and added the model findcord_tumor. We should also update the config files of the other models (t2star_sc, mice_uqueensland_sc and mice_uqueensland_gm). Should we move the models on github in this PR or in a subsequent one?

ivadomed sister PR: https://github.com/ivadomed/ivadomed/pull/312

spinalcordtoolbox/deepseg/core.py
```diff
@@ -85,16 +85,20 @@ def fill_holes(nii_seg):
     return nii_seg
 
 
-def segment_nifti(fname_image, folder_model, param={}):
+def segment_nifti(fname_image, folder_model, fname_prior=None, param=None):
```
andreanne-lemay, line 88: Changed `param={}` to `param=None` because setting default argument value to a mutable variable is not a [good practice](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)

scripts/sct_deepseg.py
```diff
@@ -102,62 +116,115 @@ def get_parser():
     return parser
 
 
+def display_list_models():
```
charleygros, line 119: :+1: 
Not sure: should we move `display_list_models` and `display_list_tasks` to `deepseg/models.py`?

charleygros:
Models updated on OSF!

scripts/sct_deepseg.py
andreanne-lemay: I asked myself the same thing, I was unsure between scripts/sct_deepseg.py, deeseg/models.py and deepseg/core.py. I don't have a strong preference, glad to move it to wherever makes more sense.

andreanne-lemay: Moved it to deepseg/models.py :)

jcohenadad:
@charleygros i'd like to review this PR before it is merged (i just need to deal with a few priorities...)

spinalcordtoolbox/deepseg/models.py
```diff
@@ -19,23 +20,44 @@
...
+    'findcord_tumor':
+        {'url': 'https://osf.io/qj6d5/download?version=1',
+         'description': 'Cord localisation model, trained on T2-weighted images with tumor.',
          'default': False}
     }
 
+# List of task. The convention for task names is: action_(animal)_(contrast)_region
+# Regions could be: sc, gm, lesion, tumor
+TASKS = {
+    'segment_t2star_sc':
```
jcohenadad, line 47: i find it a bit too verbose. I suggest to replace segment by seg.

also, I find it more intuitive to talk about the task, then object, then modality. So, something like: `seg_sc_t2star`. It could be the purpose of another PR, but since we are introducing a new TASKS section, it does not hurt to do it now.

Now, it also raises this question: if we implement the suggestion above, should we also propagate this logic to model names? this is more of a contentious point due to the large changes that it implies (i apologize in advance 😬). On the other hand, I don't find it shocking to have tasks called "seg_sc_t2" and models called "findcord_t2", "t2_sc", etc.

now, i am also realizing that we don't _really_ have a convention (yet) for our ivadomed models. For example:
- findcord_tumor → no modality is listed, cord instead of sc, and the task (find) is not separated with "_"
- t2star_sc → modality comes first, no task listed (seg). 

so, to minimize the amount of changes for this PR, maybe we can go with the above suggestion and only touch the TASK names, and for model names we open an issue on ivadomed and settle on a convention.

jcohenadad:
While playing with the script, i noticed that the output binary seg is INT64 (probably an overkill?). Since this is out-of-scope i've opened this https://github.com/neuropoly/spinalcordtoolbox/issues/2758

spinalcordtoolbox/deepseg/models.py
charleygros: Naming convention --> indeed, important points.
Change TASK names: I agree, happy to do this

charleygros: Note: when I named the TASK, I tried to match the[ convention that was stated for MODELS](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L17), ie region last.

charleygros: Done in [7b86bb9](https://github.com/neuropoly/spinalcordtoolbox/pull/2748/commits/7b86bb96e228e1ac926cbdcba78cbe2d3151e2df)

spinalcordtoolbox/deepseg/models.py
```diff
@@ -19,23 +20,44 @@
...
+    'seg_sc_t2star':
+        {'description': 'Cord segmentation on T2*-weighted contrast.',
+         'models': ['t2star_sc']},
+    'seg_mice_sc':
+        {'description': 'Cord segmentation on mouse MRI.',
+         'models': ['mice_uqueensland_sc']},
+    'seg_mice_gm':
+        {'description': 'Gray matter segmentation on mouse MRI.',
+         'models': ['mice_uqueensland_gm']},
+    'seg_tumor_t2w':
```
jcohenadad, line 56: To be consistent with the « t2star » case
```suggestion
    'seg_tumor_t2':
```

jcohenadad:
Re: [Travis failing](https://github.com/neuropoly/spinalcordtoolbox/pull/2748/commits/e7887c6b9d40e17a789f2f48faad9057cf086951). I've created another branch and PR to narrow down the cause of this fail https://github.com/neuropoly/spinalcordtoolbox/pull/2761. It is now clear (based on [this](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/702495027)) that the problem comes from the bumb in ivadomed 1.2.1 → 2.0.2. I will continue my investigations (and document) in https://github.com/neuropoly/spinalcordtoolbox/pull/2761

jcohenadad:
OK, the only change that caused Travis to fail was to put back pytest. I suspect a test is modifying the sct_testing_data, which are then used by sct_testing, and somehow causing problem.

jcohenadad:
i found the bug 🎉!

it was caused by the `pytest:test_deepseg:test_segment_nifti` which overwrite the `t2s_seg.nii.gz` file. Then, the other integrity test `test_sct_get_centerline` uses the (now modified) `t2s_seg.nii.gz` file as a ground truth, which generates this failure. The reason it did not fail with ivadomed==1.2.1 is because `test_sct_get_centerline` was "tuned" for this particular segmentation, but since the output is slightly different in 2.0.2, it now fails.

I will fix this ASAP in another PR.

jcohenadad:
https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/702564991#L1544

ok that's weird. @charleygros or @andreanne-lemay Did you change the behaviour of the output?

jcohenadad:
> https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/702564991#L1544
> 
> ok that's weird. @charleygros or @andreanne-lemay Did you change the behaviour of the output?

ha! 😉 
https://github.com/neuropoly/spinalcordtoolbox/pull/2748/files#diff-581105324591b1ce7fd7dc6d0012a386R88

jcohenadad:
@charleygros @andreanne-lemay good to
Merge?

andreanne-lemay:
> @charleygros @andreanne-lemay good to
> Merge?

Good on my end!

PR #2761, opened 2020-06-26
Dummy PR to investigate Travis fail

jcohenadad:
[This Travis fail](https://github.com/neuropoly/spinalcordtoolbox/pull/2748/commits/e7887c6b9d40e17a789f2f48faad9057cf086951) from #2748 cannot be reproduced locally, so the purpose of this temporary PR is to investigate this mystery.

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -16,7 +16,7 @@ conda activate venv_sct  # reactivate conda for the pip install below
 
 echo *** UNIT TESTS ***
 sct_download_data -d sct_testing_data  # for tests
-pytest
+# pytest
 
 echo *** INTEGRATION TESTS ***
 pip install coverage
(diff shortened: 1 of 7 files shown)


jcohenadad:
It is now clear (based on [this](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/702495027)) that the problem comes from the bumb in ivadomed 1.2.1 → 2.0.2. 

What's concerning is that this issue cannot be reproduced on local stations (nor via ssh). So far, it only happens on Travis. @kousu @Drulex @zougloub do you have any idea what could cause that? (ie something "Travis-specific").

jcohenadad:
even weirder, [this travis build passes](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/702499867), even though the only change to `test_sct_get_centerline` is a few print() functions added to the code???

jcohenadad:
mystery continues: i removed all the print(), the repos is at the same state as e7887c6, and yet all test are now passing: https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/702504323

jcohenadad:
the issue is now fixed, so we can close this. More details in #2748

PR #2798, opened 2020-07-16
Add a safelist to avoid double-building on PRs

kousu:
https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches
https://docs.travis-ci.com/user/pull-requests#double-builds-on-pull-requests

part of #2674

PR #2800, opened 2020-07-17
Testing #2798

kousu:
This will let us test how many build travis does with the new configuration in https://github.com/neuropoly/spinalcordtoolbox/pull/2798.
This isn't to be merged.


kousu:
test complete

PR #2812, opened 2020-07-23
Refactor msct_register, msct_register_landmarks to spinalcordtoolbox.registration

### LINKED ISSUE #2547: Refactoring of msct_register and add more testing
Following-up on the initial refactoring of `sct_register_multimodal` and `sct_register_to_template` in #2331, a deeper refactoring is needed. Notably:
- Move `msct_register` into a proper Python module: spinalcordtoolbox/register/register
- Move `msct_register_landmarks` into spinalcordtoolbox/register/landmarks
- Add unit tests for the rotation estimation. To do so, generate an image, rotate it by a known angle and run each method.

Drulex:
- In the scope of https://github.com/neuropoly/spinalcordtoolbox/projects/7
- Partly fixes #2547 

In a nutshell:
- `msct_register.py` high-level logic refactored into `sct_register_to_template.py`
- `msct_register.py` low-level/reusable logic refactored into `spinalcordtoolbox.registration.register`
- due to the massive amount of dependencies living in `scripts/` more issues will be opened to refactor those in proper API modules. Some were refactored into their corresponding modules (e.g `Image`), but the rest remain for as-is for now.
- placeholder unit tests added in `test_register.py`

This PR is purely cosmetic at this point (no functional changes), `msct_register` no longer exists and the core logic has moved to the API. More localized PRs will be opened to avoid merge-related troubles (WIP list here: https://github.com/neuropoly/spinalcordtoolbox/projects/7#column-3061677)


Drulex:
I am stopping this PR here since the scope is too broad an may cause potential merge issues. The refactor work will continue in other PRs. At this point `msct_register` and `msct_register_landmarks` are removed and code as been re-organized in preparation for a full "APIfication".

The functional tests are still passing and placeholder unit tests have been created.

jcohenadad:
@Drulex given that this PR has a lot of content to review, could you please run `batch_processing.sh` and compare output results with those in the [latest release](https://github.com/neuropoly/spinalcordtoolbox/releases/tag/4.3).

In parallel, I will review as much as I can.

spinalcordtoolbox/image.py
```diff
@@ -516,19 +504,16 @@ def getNonZeroCoordinates(self, sorting=None, reverse_coord=False, coordValue=Fa
         if self.dim[2] == 1:
             n_dim = 2
 
-        try:
-            if n_dim == 3:
+        if n_dim == 3:
```
jcohenadad, line 507: hum, i'm not sure i follow the logic here (we lost the try/except for the n_dim=3 case), but the original code probably had some flaws... so maybe addressing it in this PR is out of scope. Just mentioning in case there is a problem that needs to be addressed at some point

spinalcordtoolbox/utils.py
```diff
@@ -3,9 +3,8 @@
 # Collection of useful functions
 
 
-import io
-import os
-import re
+import io, sys, os
```
jcohenadad, line 6: doesn't it go against [PEP8](https://www.python.org/dev/peps/pep-0008/#imports)?

jcohenadad (approved):
Looks good to me! If the `batch_processing` test is positive, we can probably merge safely

spinalcordtoolbox/image.py
Drulex: I did this wherever I saw the following pattern:

```
try:
    something()
except Exception as e:
   sct.printv("some error message)
```

The reason being that an error is detected, but the exception is ignored without an explicit `pass` or message saying that it's normal for the exception to be ignored. This leads me to believe that it's possibly a mistake and that the code should raise the exception to the parent.

jcohenadad: ok got it

Drulex:
>  could you please run `batch_processing.sh` and compare output results with those in the [latest release](https://github.com/neuropoly/spinalcordtoolbox/releases/tag/4.3)

```
Version:         git-aj/2547_refactor_msct_register-9f7ae226e607c142c73faf946d0f9d7d3807311d
Ran on:          Linux drupad-p1 5.4.58-1-MANJARO
Duration:        0hrs 13min 29sec
---
mt/MTR(WM):      54.42162290395471
```

spinalcordtoolbox/utils.py
Drulex: Yes indeed. I actually disabled the PEP8 on my editor due to too many warnings :sweat_smile: . 

If we want to be really consistent we could run the PEP8 formatter on the code (https://pypi.org/project/autopep8/). Developers could run it as a pre-commit hook (or manually) prior to pushing code.

PR #2814, opened 2020-07-24
Cleanup documentation on readthedocs

jcohenadad:
This PR brings will move the documentation from the Github's README to readthedocs. 

➤ **Built documentation for this PR: https://spinalcordtoolbox.readthedocs.io/en/jca-1656-rtd**

This PR includes the following changes:

- [x] Display SCT logo on top left of RTD
- [ ] ~Automatically fetch the latest version (currently hard-coded in the conf.py)~ --> not urgent
- [x] Made the page secure - use SSL for forum #2830
- [x] Create spinalcordtoolbox.com domain on NameCheap and create ALIAS for readthedocs.io. For the records, the working config on NameCheap is: `Type=ALIAS Record | Host=@ | Value=readthedocs.io`

Remaining issues not addressed here:
- Display Changelog in the RTD documentation #2822
- Transfer the remaining documentation from sourceforge #2829
- Fetching version.txt from site-package does not work for RTD #2832


jcohenadad:
@Drulex I will now go through the API section. I was wondering: do we need to keep the term "API" for each of the sub-headings? It makes it "heavy to look" on the left tab to have all these API terms, which I think is not necessary because it's pretty clear we are in the API section:

image

Drulex:
> do we need to keep the term "API" for each of the sub-headings?

Not at all, the previous doc used this nomenclature so I just followed. I agree with you.

Drulex:
My browser is showing that "part of this page are note secure" on the index page.

image

I suspect it has to do with the forum badge which points to a domain (spinalcordmri.org) that doesn't use SSL.

Drulex:
It looks like the `automodule` did not generate the docs for the API code: https://spinalcordtoolbox.com/en/jca-1656-rtd/dev_section/api.html

Drulex:
> It looks like the `automodule` did not generate the docs for the API code: https://spinalcordtoolbox.com/en/jca-1656-rtd/dev_section/api.html

It probably has to do with this setting on `readthedocs` website: 

image

which doesn't match the new path in `.readthedocs.yaml`.

Will see what happens if I just remove it.

jcohenadad:
> It looks like the `automodule` did not generate the docs for the API code: https://spinalcordtoolbox.com/en/jca-1656-rtd/dev_section/api.html

weird. it was working 2-3 commits ago on RTD. i suspect it is linked to the added recommark extension which i added to render changes.md

jcohenadad:
> My browser is showing that "parts of this page are note secure" on the index page.
> 
> image
> 
> I suspect it has to do with the forum badge which points to a domain (spinalcordmri.org) that doesn't use SSL.

good catch-- issue opened https://github.com/neuropoly/spinalcordtoolbox/issues/2830

jcohenadad:
Following on https://github.com/neuropoly/spinalcordtoolbox/pull/2814#issuecomment-669168870:
- Tried to add .md in `source_suffix` in conf.py 16af99d --> didn't work

documentation/source/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinxcontrib.programoutput
+recommonmark
```
joshuacwnewton, line 2:
```suggestion
sphinx
sphinx_rtd_theme
sphinxcontrib.programoutput
recommonmark
```
As per setup.py, are `sphinx` and `sphinx_rtd_theme` also needed here? Or are the installation of those packages handled elsewhere?

I ask because when building the docs locally using `make html`, I had to manually install them  before it would build.

documentation/source/overview/concepts.rst
```diff
@@ -236,6 +155,8 @@ look at:
 https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir
 
 
+
+
 Matlab Integration on Mac
```
joshuacwnewton, line 160: I am wondering if `Matlab Integration for Mac` might be better suited for the `Installation` section. To me, it doesn't seem to fit with the other "SCT Concepts".

joshuacwnewton: Side question: should I be providing suggestions on content like this? Or is it preferred to simply comment only on the changes made by you?

documentation/source/user_section/command-line.rst
```diff
@@ -135,176 +192,181 @@ sct_convert
...
 sct_crop_image
-=================
+==============
 
 .. program-output:: sct_crop_image -h
 
 
-.. _sct_deepseg_gm:
+sct_deepseg
+==============
```
joshuacwnewton, line 207: Nitpick, but it seems this header also should also be updated.
```suggestion
===========
```

setup.py
```diff
@@ -64,11 +65,9 @@
      console_scripts=[
         '{}=spinalcordtoolbox.compat.launcher:main'.format(x) for x in \
         [
-            'isct_check_detection',
             'isct_convert_binary_to_trilinear',
             'isct_minc2volume-viewer',
             'isct_test_ants',
-            'isct_warpmovie_generator',
```
joshuacwnewton, line 71: Is this because of #2828? I see the "Internal Commands" documentation section was changed, and so if I understand, this was updated to match.  Just a nitpick, but could this change to `setup.py` instead be part of a larger PR addressing #2828? Or, does the presence of these two lines affect how the documentation is built?

documentation/source/user_section/command-line.rst
```diff
@@ -13,117 +13,174 @@ Command-Line Tools
...
+
+System tools
+============
+
+- sct_check_dependencies_ - Check installation and compatibility of SCT.
+- sct_testing_ - Runs complete testing to make sure SCT is working properly.
+- sct_version_ - Display SCT version.
 
 
 Main Tools
```
joshuacwnewton, line 124: I'm noticing that the inline commands show an error message when rendered on this page, e.g. https://spinalcordtoolbox.com/en/jca-1656-rtd/user_section/command-line.html#sct-analyze-lesion. 

```
System Message: ERROR/6 (/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/checkouts/jca-1656-rtd/documentation/source/user_section/command-line.rst, line 131)

Command 'sct_analyze_lesion -h' failed: [Errno 2] No such file or directory: 'sct_analyze_lesion': 'sct_analyze_lesion'
```

Is this expected behavior?

documentation/source/overview/concepts.rst
jcohenadad: > Side question: should I be providing suggestions on content like this? Or is it preferred to simply comment only on the changes made by you?

you are more than welcome to provide suggestions!

documentation/source/requirements.txt
jcohenadad: > As per setup.py, are sphinx and sphinx_rtd_theme also needed here? Or are the installation of those packages handled elsewhere?

needed by readthedocs, to avoid https://github.com/neuropoly/spinalcordtoolbox/pull/2814#issuecomment-669168870

jcohenadad: full disclosure-- however i haven't tested if sphinx is needed specifically here-- it doens't hurt and i have no time to test-- so i leave it here for now (feel free to experiment)

documentation/source/overview/concepts.rst
jcohenadad: > I am wondering if `Matlab Integration for Mac` might be better suited for the `Installation` section. To me, it doesn't seem to fit with the other "SCT Concepts".

fixed in 3f095f8c82c0198528b876287a5b57eb3d37c7f7

documentation/source/user_section/command-line.rst
jcohenadad: fixed in bde6e598ede15e70bfd84e133228fef71d00d3f0

setup.py
jcohenadad: > Is this because of #2828? 

yes

> I see the "Internal Commands" documentation section was changed, and so if I understand, this was updated to match. Just a nitpick, but could this change to `setup.py` instead be part of a larger PR addressing #2828? Or, does the presence of these two lines affect how the documentation is built?

it would ideally be part of a larger PR, but as you suspected, the presence of these two lines affect how the doc is built: it display an ugly message under these functions in the doc:
~~~
isct_warpmovie_generator
  File "/Users/julien/code/sct/scripts/isct_warpmovie_generator.py", line 18, in <module>
    from scipy.misc import toimage
ImportError: cannot import name 'toimage'
~~~

joshuacwnewton: Thank you for the clarification -- that makes sense. :slightly_smiling_face:

documentation/source/user_section/command-line.rst
joshuacwnewton: Ah, it seems an extra "g" character was added in that commit!

jcohenadad: oopsi dc773aab5fceff541117dd5a851f6c9535ee1a09

documentation/source/user_section/command-line.rst
Drulex: No, we are trying to use a sphinx plugin to run a command (in this case a `sct_command -h`) to show the help message of the specific command, but this won't run on the RTD build system since sct hasn't been installed.

The install of sct is not a simple `pip install`, therefore we need a strategy on how to deal with this.

jcohenadad: hum... this is strange. It also happens on the master branch (which was working before): 

image

so i'm wondering if this is a temporary glitch in RTD...?

jcohenadad: > No, we are trying to use a sphinx plugin to run a command (in this case a `sct_command -h`) to show the help message of the specific command, but this won't run on the RTD build system since sct hasn't been installed.
> 
> The install of sct is not a simple `pip install`, therefore we need a strategy on how to deal with this.

ah! so i was wrong @Drulex : this has never worked so far on the master branch (from RTD)?

jcohenadad: now that we have the scripts listed in the setup entry points, this _should_ work with a pip install by RTD. I'll do some testing...

Drulex: From what I recall no. It was first brought up here: https://github.com/neuropoly/spinalcordtoolbox/pull/2781#issuecomment-658818129 then I guess it fell in the cracks.

jcohenadad: ok, i figured the proper config for .readthedocs.yml, which installs the scripts, however now i'm bumping in the following issue:
~~~
  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/bin/sct_analyze_lesion", line 5, in <module>
    from spinalcordtoolbox.compat.launcher import main
  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/lib/python3.6/site-packages/spinalcordtoolbox/__init__.py", line 3, in <module>
    from .utils import __version__, __sct_dir__, __data_dir__, __deepseg_dir__
  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/lib/python3.6/site-packages/spinalcordtoolbox/utils.py", line 462, in <module>
    __version__ = _version_string()
  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/lib/python3.6/site-packages/spinalcordtoolbox/utils.py", line 444, in _version_string
    install_type, sct_commit, sct_branch, version_sct = _git_info()
  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/lib/python3.6/site-packages/spinalcordtoolbox/utils.py", line 437, in _git_info
    with io.open(os.path.join(__sct_dir__, 'version.txt'), 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/jca-1656-rtd/lib/python3.6/site-packages/version.txt'
~~~

which i will solve by adding a copy of the version.txt via the setup.py

jcohenadad: ok, it seems the problem is more fundamental: in `spinalcordtoolbox.utils`, the version is fetched using `os.path.join(__sct_dir__, 'version.txt')`, which, for a pip-installed package, is pointing to `envs/sct-env/lib/python3.6/site-packages/version.txt`. 

By adding this line:
https://github.com/neuropoly/spinalcordtoolbox/blob/734602113cdea1ad1d687ec82c24b1c4e8ff21c2/setup.py#L44

the file `version.txt` is copied under `envs/sct-env/version.txt`, therefore it does not solve the problem.

A possibly more elegant solution would be to edit `spinalcordtoolbox.utils` so it points to the `version.txt` file in site-package.

@Drulex @joshuacwnewton does that sounds like a reasonable approach? if so i'll open an issue

jcohenadad: ok i went ahead: https://github.com/neuropoly/spinalcordtoolbox/issues/2832

Issue #2818, opened 2020-08-01
ImportError: cannot import name 'toimage' when executing isct_warpmovie_generator


joshuacwnewton:
### Description

`isct_warpmovie_generator.py` uses `scipy.misc.toimage`. This function was [removed in SciPy 1.2.0](https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.misc.toimage.html#scipy.misc.toimage). The latest version of SciPy is 1.5.2, which is used in new installs because no version is specified in `requirements.txt` . As such, `toimage` is unavailable, causing the to script fail when run.

### Steps to Reproduce

0. Install SCT 
```
SCT info:
- version: git-master-7dfb515728f5ba702df1dabdd9a33f00c085e224
- path: /home/joshua/repos/spinalcordtoolbox
OS: linux (Linux-5.4.0-7634-generic-x86_64-with-debian-bullseye-sid)

[...]

Check if scipy is installed.........................[OK] (1.5.2)
```

1. Run `isct_warpmovie_generator`

**Expected behavior:** <What you expect to happen>

Display help without error.

**Actual behavior:** <What actually happens>

```bash
Traceback (most recent call last):
  File "/home/joshua/repos/spinalcordtoolbox/scripts/isct_warpmovie_generator.py", line 18, in <module>
    from scipy.misc import toimage
ImportError: cannot import name 'toimage'
```

### Additional comments

https://github.com/neuropoly/spinalcordtoolbox/blob/7dfb515728f5ba702df1dabdd9a33f00c085e224/scripts/isct_warpmovie_generator.py#L108-L110

* This is the only usage of `toimage` I could find in SCT.
* [SciPy's deprecation notes](https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.misc.toimage.html#scipy.misc.toimage) recommend replacing with `PIL.fromarray`.
    * We use `cmin=0.0` in our call to `toimage`, which [changes how intensity values are scaled](https://github.com/scipy/scipy/blob/v0.19.1/scipy/misc/pilutil.py#L89-L102), but I believe this is just for visual clarity. `PIL.fromarray` doesn't provide this scaling option, so it might be worth checking the difference on sample data.
* `isct_warpmovie_generator.py` is not called in `batch_processing.sh` or `sct_testing.py`, which may explain how this was missed. This script is listed in `setup.py` as a user-facing script -- I would guess that it should be tested in some way.
[...]

PR #2337, opened 2019-07-08
Continuation conversion from msct_parser to argparse

YangHee-Min:
As mentioned in PR #2315, the conversion of all scripts within spinalcordtoolbox from msct_parser to argparse is necessary due to the general impracticability of the internal parser. 

In this PR, we look to do the following:
**TODO**
- Look into bug related to running `batch_processing` (works on Hee-Min's side but not Julien's)

**DONE**
Conversion from msct_parser to argparse of the scripts and their associated tests:

 -  sct_deepseg_gm
 - sct_deepseg_lesion
 - sct_deepseg_sc
 - sct_denoising_onlm
 - sct_detect_pmj
 - sct_dice_coefficient
 - sct_dmri_compute_bvalue
 - sct_dmri_compute_dti
 - sct_dmri_concat_bvals
 - sct_dmri_concat_bvecs
- sct_concat_transfo [(details mentioned here)](https://github.com/neuropoly/spinalcordtoolbox/pull/2315#issuecomment-508865758)
- Issue related to sct_crop_image (Issue #2336 )

Update functions to SmartFormatter:
- sct_analyze_lesion
- sct_analyze_texture
- sct_apply_transfo
- sct_check_dependencies
- sct_compute_ernst_angle
- sct_compute_hausdorff_distance
- sct_compute_mscc
- sct_compute_mtr
- sct_compute_mtsat
- sct_compute_snr
- sct_concat_transfo
- sct_convert
- sct_create_mask
- sct_crop_image

Other changes:
- set the lesion mask as mandatory, since this function aims at analysing the lesions of a scan
- set the SC mask as mandatory in order to take into account the curvature of the cord. 
Fixes #2343.

PR #2819, opened 2020-08-01
Convert 1-10 out of 20 remaining scripts from msct_parser to argparse

joshuacwnewton:
### Reference issues/PRs

Follows #2315, #2337. Partially addresses #1548.

### What does this implement/fix? Explain your changes.

This PR converts the first 10 of 20 remaining scripts that still use `msct_parser`. See below for progress on task list.

 - `/scripts/isct_warpmovie_generator.py` (possibly non-functional script, see #2818)


scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,106 @@ def get_parser():
...
+        nargs='+',
+        help=f"R|Advanced parameters. Assign value with \"=\", and separate arguments with a space (\" \")\n"
+             f"    -poly [int]: Degree of polynomial function used for regularization along Z. For no regularization set to "
+             f"0. Default={param_default.poly}.\n"
+             f"    -smooth [mm]: Smoothing kernel. Default={param_default.smooth}.\n"
+             f"    -metric {{MI, MeanSquares, CC}}: Metric used for registration. Default={param_default.metric}.\n"
+             f"    -gradStep [float]: Searching step used by registration algorithm. The higher the more deformation "
+             f"allowed. Default={param_default.gradStep}.\n"
+             f"    -sample [None or 0-1]: Sampling rate used for registration metric. Default={param_default.sampling}.\n"
+    )
```
joshuacwnewton, line 115: Grouping "advanced parameters" into a single parameter feels non-intuitive and easy to mess up from a user perspective. Grouping in this way also denies us the usual typing and validation that argparse provides. I was thinking that this could instead be an argument group (see "OPTIONAL ARGUMENTS" and "MANDATORY ARGUMENTS" also used here.) 

I left it for now, but this could be changed for all the scripts once the argparse conversion is done.

scripts/sct_dmri_transpose_bvecs.py
```diff
@@ -24,66 +24,71 @@
...
-                      type_value='file',
-                      description='Input bvecs file.',
-                      mandatory=True,
-                      example='bvecs.txt')
-    parser.add_option(name='-i',
-                      type_value='file',
-                      description='Input bvecs file.',
-                      mandatory=False,
-                      example='bvecs.txt',
-                      deprecated_by='-bvec')
```
joshuacwnewton, line 56: A couple of parameters had two names (e.g. `-i`/`-bvec` here). Argparse doesn't have a deprecation mechanism like this. Do we want to still try to allow usage of the old parameter names? Or, is now a good time to make the switch?

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,213 @@ def get_parser():
...
-                      type_value='image_nifti',
-                      description='Nifti mask to weight each voxel during ML or MAP estimation.',
-                      example='PAM50_wm.nii.gz',
-                      mandatory=False)
-    parser.add_option(name='-discard-neg-val',
-                      type_value='multiple_choice',
-                      mandatory=False,
-                      description='Discard voxels with negative value when computing metrics statistics.',
-                      example=["0", "1"],
-                      default_value="0")
```
joshuacwnewton, line 191: These old argument names have hyphens (`-`) which isn't allowed in argparse. [A possible workaround](https://stackoverflow.com/a/12834986) is possible, but it might be better to just switch to underscores.

The trouble comes with any existing users who are using the hyphenated names. It's similar to the deprecated names, in that this could be a breaking change.

scripts/sct_extract_metric.py
```diff
@@ -352,70 +398,29 @@ def main(fname_data, path_label, method, slices, levels, fname_output, labels_us
...
+    vertebral_levels = arguments.vert
+    fname_vertebral_labeling = arguments.vertfile
+    perslice = arguments.perslice
+    perlevel = arguments.perlevel
+    fname_normalizing_label = arguments.norm_file  # TODO: Not used. Why?
+    normalization_method = arguments.norm_method  # TODO: Not used. Why?
+    label_to_fix = arguments.fix_label  # TODO: Not used. Why?
+    fname_output_metric_map = arguments.output_map  # TODO: Not used. Why?
+    fname_mask_weight = arguments.mask_weighted  # TODO: Not used. Why?
+    discard_negative_values = int(arguments.discard_neg_val)  # TODO: Not used. Why?
```
joshuacwnewton, line 422: Several of these arguments are just not used at all. Were they intended to be used? Abandoned? Should we keep them? 

I've marked them with comments for now.

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,106 @@ def get_parser():
...
+        help="Output folder. Example: dmri_moco_results/"
+    )
+    # TODO: Convert -r to flag using action=store_true
+    optional.add_argument(
+        "-r",
+        metavar=Metavar.str,
+        choices=('0', '1'),
+        default='1',
+        help="Remove temporary files. 0 = no, 1 = yes"
+    )
```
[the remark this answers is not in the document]
jcohenadad: agreed about the store_true. Also agreed to deal with it in another PR. However, given that this change in CLI will provoke cross-compatibility issues, it would be nice to have it implemented before our planned 4.4.0 release.

scripts/sct_dmri_moco.py
jcohenadad: 💯 agreed

same comment as above about the cross-compatibility issue and the strategy re: new release

scripts/sct_dmri_transpose_bvecs.py
jcohenadad: definitely a good time to make the switch!

scripts/sct_extract_metric.py
jcohenadad: i'm not sure i understand the problem. For example, `-discard-neg-val` would be interpreted in the code as `arguments.discard_neg_val`, wouldn't it?

joshuacwnewton: Aha, you're right. My apologies.

scripts/sct_dmri_moco.py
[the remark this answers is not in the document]
joshuacwnewton: It may be better to leave argparse conversion (this PR and also scripts #11-20) for 4.5.0, then. I would have to complete the remaining scripts plus the two cross-compatibility changes in 1 week, which I think might be a little tight, time-wise. Some of those remaining scripts are quite a bit longer than the ones in this PR.

Would that be alright?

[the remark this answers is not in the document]
jcohenadad: we can also wait a bit more for 4.4.0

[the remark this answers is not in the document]
joshuacwnewton: Sounds good! I'll make this resolved for now and address this separately once all of the scripts have been converted. :)

scripts/sct_dmri_moco.py
joshuacwnewton: Same as above :)

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,106 @@ def get_parser():
     param_default = ParamMoco(is_diffusion=True, group_size=3, metric='MI', smooth='1')
```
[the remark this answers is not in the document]
joshuacwnewton: In hindsight, I think this is inherent to the usage of a class for `Param`. I can still clean this up a little, but other than that fixing this requires moving away from `Param` objects, which would be a much larger change.

joshuacwnewton:
@jcohenadad  I've implemented the requested changes. (Preserving hyphenated arg names, and properly parsing list values.) I'll continue updating the remaining scripts (11-20) which should take a few days, so no rush on reviewing/merging.

Also, here's a summary of things to be taken care of by other PRs. Verbosity already has an issue, but I'll make issues for the others for further discussion if needed.
* Make verbosity consistent across all scripts
* Convert binary flags (0/1, on/off) to use `action="store_true"` across all scripts
* Potentially split `-param` into an `"ADVANCED ARGUMENTS"` group.
* Update any args that use `type_value=[['delim'], 'type']` or `nargs="+"` to using the new `type=list_type()` approach. (To preserve conventions in SCT documentation/workshops/etc.)

jcohenadad:
> * Make verbosity consistent across all scripts

👍 

> * Convert binary flags (0/1, on/off) to use `action="store_true"` across all scripts

not necessarily. I know we had a discussion thread about this (cannot find it...), but the pros of keeping 0/1 is that in some cases, we would like the default to be 1. In that case, having a flag called `-do-not-do-this` might be less intuitive than `-do-this 0/[1]`. So, it's a case by case decision.

> * Potentially split `-param` into an `"ADVANCED ARGUMENTS"` group.

👍

> * Update any args that use `type_value=[['delim'], 'type']` or `nargs="+"` to using the new `type=list_type()` approach. (To preserve conventions in SCT documentation/workshops/etc.)

👍

joshuacwnewton:
> not necessarily. I know we had a discussion thread about this (cannot find it...), but the pros of keeping 0/1 is that in some cases, we would like the default to be 1. In that case, having a flag called `-do-not-do-this` might be less intuitive than `-do-this 0/[1]`. So, it's a case by case decision.

Ah, my apologies! I was unclear. There is also `action='store_false'` for the case you describe (which defaults to `True`). I was just using shorthand there but you can do either for flags. :)

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,109 @@ def get_parser():
...
-                      type_value='multiple_choice',
-                      description="verbose: 0 = nothing, 1 = classic, 2 = expended",
-                      mandatory=False,
-                      example=['0', '1', '2'],
-                      default_value='1')
+    parser = argparse.ArgumentParser(
+        description=(
+            '  Motion correction of dMRI data. Some of the features to improve robustness were proposed in Xu et al. '
+            '(http://dx.doi.org/10.1016/j.neuroimage.2012.11.014) and include:\n'
+            '    -group-wise (-g)\n'
```
jcohenadad, line 48: i think the 2-space indent is sufficient. Also, i think it is important to keep a space between "-" and the item, otherwise people might think this is a flag (whereas this is a list of items). Same comment applies to the other items below
```suggestion
            '  - group-wise (-g)\n'
```

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,109 @@ def get_parser():
...
-                      example=['0', '1'])
-    parser.add_option(name="-v",
-                      type_value='multiple_choice',
-                      description="verbose: 0 = nothing, 1 = classic, 2 = expended",
-                      mandatory=False,
-                      example=['0', '1', '2'],
-                      default_value='1')
+    parser = argparse.ArgumentParser(
+        description=(
+            '  Motion correction of dMRI data. Some of the features to improve robustness were proposed in Xu et al. '
```
jcohenadad, line 46:
```suggestion
            'Motion correction of dMRI data. Some of the features to improve robustness were proposed in Xu et al. '
```

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,109 @@ def get_parser():
...
+        metavar=Metavar.file,
+        default=param_default.fname_mask,
+        help='Binary mask to limit voxels considered by the registration metric. Example: dmri_mask.nii.gz',
+    )
+    optional.add_argument(
+        '-param',
+        metavar=Metavar.list,
+        type=list_type(',', str),
+        help=f"R|Advanced parameters. Assign value with \"=\", and separate arguments with \",\".\n"
+             f"    -poly [int]: Degree of polynomial function used for regularization along Z. For no regularization set to "
```
jcohenadad, line 111: (same for the other items)
```suggestion
             f"  - poly [int]: Degree of polynomial function used for regularization along Z. For no regularization set to "
```

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,109 @@ def get_parser():
...
+             f"    -smooth [mm]: Smoothing kernel. Default={param_default.smooth}.\n"
+             f"    -metric {{MI, MeanSquares, CC}}: Metric used for registration. Default={param_default.metric}.\n"
+             f"    -gradStep [float]: Searching step used by registration algorithm. The higher the more deformation "
+             f"allowed. Default={param_default.gradStep}.\n"
+             f"    -sample [None or 0-1]: Sampling rate used for registration metric. Default={param_default.sampling}.\n"
+    )
+    optional.add_argument(
+        '-x',
+        metavar=Metavar.str,
+        choices=['nn', 'linear', 'spline'],
```
jcohenadad, line 122: the choices are presently not visible from the usage. I _think_ this is because you set the metavar variable. See e.g. [sct_apply_transfo](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_apply_transfo.py#L101), where this variable is not set (and the different choices are visible when launching the CLI's help)

scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -40,84 +42,102 @@ def __init__(self):
...
+        "-h",
+        "--help",
+        action="help",
+        help="Show this help message and exit."
+    )
+    optional.add_argument(
+        '-a',
+        type=int,
+        metavar=Metavar.int,
+        choices=(0, 1),
```
jcohenadad, line 78: i think everytime the variable "choices" is set, we should not display "metavar" because it will overwrite the display of the available choices to the user

scripts/sct_download_data.py
```diff
@@ -14,47 +14,57 @@
...
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+    mandatory = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatory.add_argument(
+        '-d',
+        metavar=Metavar.str,
+        required=True,
+        choices=list(dict_url.keys()),
+        help=f"Name of the dataset. Choices are: {sorted(dict_url)}"
```
jcohenadad, line 37: same comment as above-- you don't need to add "Choices are...", if you remove metavar.

scripts/sct_download_data.py
```diff
@@ -14,47 +14,57 @@
...
+        action="help",
+        help="Show this help message and exit."
+    )
+    optional.add_argument(
+        '-o',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="Path to directory to save the downloaded data to."
+         " (Defaults to ./${dataset-name})."
+         " Directory will be created if it does not exist.."
```
jcohenadad, line 52:
```suggestion
         " Directory will be created if it does not exist."
```

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,212 @@ def get_parser():
...
+              "of consecutive labels use ':'. Example: 1:3 is equivalent to 1,2,3. Maximum Likelihood (or MAP) is "
+              "computed using all tracts, but only values of the selected tracts are reported.")
+    )
+    optional.add_argument(
+        '-method',
+        metavar=Metavar.str,
+        choices=['ml', 'map', 'wa', 'bin', 'max'],
+        default=param_default.method,
+        help=("R|Method to extract metrics.\n"
+              "    - ml: maximum likelihood (only use with well-defined regions and low noise)\n"
```
jcohenadad, line 130: (same comment applies to other lists)
```suggestion
              "  - ml: maximum likelihood (only use with well-defined regions and low noise)\n"
```

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,212 @@ def get_parser():
...
+              "          N.B. ONLY USE THIS METHOD WITH THE WHITE MATTER ATLAS! The sum of all tracts should be 1 in "
```
jcohenadad, line 131: the space before the "N.B." appears weirdly depending on the number of columns on the user's terminal. Eg: with 100 columns:
~~~
                            - ml: maximum likelihood (only use with well-defined regions and low
                            noise)
                                  N.B. ONLY USE THIS METHOD WITH THE WHITE MATTER ATLAS! The sum of
                                  all tracts should be 1 in           all voxels (the algorithm
                                  doesn't normalize the atlas).

~~~

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,212 @@ def get_parser():
...
+        default='',
+        help=("Label IDs to extract the metric from. Default = all labels. Separate labels with ','. To select a group " 
+              "of consecutive labels use ':'. Example: 1:3 is equivalent to 1,2,3. Maximum Likelihood (or MAP) is "
+              "computed using all tracts, but only values of the selected tracts are reported.")
+    )
+    optional.add_argument(
+        '-method',
+        metavar=Metavar.str,
+        choices=['ml', 'map', 'wa', 'bin', 'max'],
+        default=param_default.method,
```
jcohenadad, line 128: at some point it would be really cool to automatically display what is the default value https://github.com/neuropoly/spinalcordtoolbox/issues/2756

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,212 @@ def get_parser():
...
+        default=0,
+        help="Whether to combine multiple labels into a single estimation. 0 = no, 1 = yes"
+    )
+    optional.add_argument(
+        '-o',
+        metavar=Metavar.file,
+        default=param_default.fname_output,
+        help=("File name (including the file extension) of the output result file collecting the metric estimation "
+              "results.\n"
+              "Three file types are available: a CSV text file (extension .txt), a MS Excel file (extension .xls) and "
```
jcohenadad, line 163: we might want to take the opportunity to update this-- i don't think we cover xls and pickle anymore

scripts/sct_fmri_moco.py
```diff
@@ -12,81 +12,112 @@
...
+        '-m',
+        metavar=Metavar.file,
+        help="Binary mask to limit voxels considered by the registration metric."
+    )
+    optional.add_argument(
+        '-param',
+        metavar=Metavar.list,
+        type=list_type(',', str),
+        help=(f"R|Advanced parameters. Assign value with \"=\"; Separate arguments with \",\".\n"
+              f"    -poly [int]: Degree of polynomial function used for regularization along Z. For no regularization "
```
jcohenadad, line 80: same cosmetic suggestion wrt. displaying lists
```suggestion
              f"  - poly [int]: Degree of polynomial function used for regularization along Z. For no regularization "
```

joshuacwnewton:
Thank you for your thorough review, @jcohenadad. I imagine a PR of this size is slightly tedious to review, so I very much appreciate the time and care. :slightly_smiling_face:

scripts/sct_extract_metric.py
joshuacwnewton: Thank you for pointing this out! I've put that issue on the argparse Project Board so I can keep it in mind. There are lots of argparse-related tasks on that board, so perhaps we could have a chat about which ones are worth getting done before the next release.

scripts/sct_extract_metric.py
joshuacwnewton: Yes, good eye. It looks like we use `save_as_csv` only in this script.

https://github.com/neuropoly/spinalcordtoolbox/blob/49cb4a16ade7c91533e669c8d1d2a777f5bbfe66/scripts/sct_extract_metric.py#L343

joshuacwnewton:
I believe I've made all of the requested changes, @jcohenadad. Some of your comments also apply to the other scripts in #2842 as well, so I will make similar changes there as well. :slightly_smiling_face:

joshuacwnewton:
> not necessarily. I know we had a discussion thread about this (cannot find it...), but the pros of keeping 0/1 is that in some cases, we would like the default to be 1. In that case, having a flag called -do-not-do-this might be less intuitive than -do-this 0/[1]. So, it's a case by case decision.

My apologies! When I first replied a couple of days ago, I misunderstood. Now that I've looked more closely, I very much agree with what you're saying here... it does feel less intuitive to supply an option to turn _off_ behavior.

scripts/sct_dmri_display_bvecs.py
```diff
@@ -29,13 +31,27 @@
...
-                      description='bvecs file.',
-                      mandatory=True,
-                      example='bvecs.txt')
+    parser = argparse.ArgumentParser(
+        description='Display scatter plot of gradient directions from bvecs file.',
+        formatter_class=SmartFormatter,
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+    mandatory = parser.add_argument_group("\nMANDATORY ARGUMENTS")
```
cfhammill, line 40: Do we want to keep the mandatory and optional argument discipline everywhere, or allow `argparse` to indicate what is required and what isn't?

scripts/sct_dmri_moco.py
```diff
@@ -40,78 +41,107 @@ def get_parser():
...
+            "(http://dx.doi.org/10.1016/j.neuroimage.2012.11.014) and include:\n"
+            "  - group-wise (-g)\n"
+            "  - slice-wise regularized along z using polynomial function (-param). For more info about the method, "
+            "type: isct_antsSliceRegularizedRegistration\n"
+            "  - masking (-m)\n"
+            "  - iterative averaging of target volume\n"
+        ),
+        formatter_class=SmartFormatter,
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
```
cfhammill, line 56: Is this snippet required? In e.g. `sct_run_batch` the `prog` argument isn't set explicitly and the usage comes out correct.

scripts/sct_dmri_moco.py
```diff
@@ -122,26 +152,19 @@ def main():
...
-    param.verbose = int(arguments.get('-v'))
+    arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
+    param.fname_data = arguments.i
+    param.fname_bvecs = os.path.abspath(arguments.bvec)
+    param.fname_bvals = arguments.bval
+    param.bval_min = arguments.bvalmin
+    param.group_size = arguments.g
+    param.fname_mask = arguments.m
+    param.interp = arguments.x
+    param.path_out = arguments.ofolder
```
cfhammill, line 163: so much cleaner!

scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -40,84 +42,99 @@ def __init__(self):
...
+    optional.add_argument(
+        '-bvalmin',
+        type=float,
+        metavar=Metavar.float,
+        help='B-value threshold (in s/mm2) below which data is considered as b=0. Example: 50.0',
+    )
+    optional.add_argument(
+        '-ofolder',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
```
cfhammill, line 96: I didn't know about this, very cool

scripts/sct_dmri_transpose_bvecs.py
```diff
@@ -24,66 +24,69 @@
...
+    mandatory = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatory.add_argument(
+        '-bvec',
+        metavar=Metavar.file,
+        required=True,
+        help="Input bvecs file. Example: bvecs.txt"
+    )
+    optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        "-h",
```
cfhammill, line 56: Is adding help as an optional argument here to make sure that it's added to the optional group? Because it's usually added to whole parser automatically I think.

scripts/sct_download_data.py
```diff
@@ -14,47 +14,54 @@
...
+        "-h",
+        "--help",
+        action="help",
+        help="Show this help message and exit."
+    )
+    optional.add_argument(
+        '-o',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help=("R|Path to directory to save the downloaded data to.\n"
```
cfhammill, line 49:
```suggestion
        help=("R|Path to a directory to save the downloaded data.\n"
```

scripts/sct_fmri_moco.py
```diff
@@ -12,81 +12,109 @@
...
-                      example=['0', '1'])
-    parser.add_option(name="-v",
-                      type_value="multiple_choice",
-                      description="""Verbose.""",
-                      mandatory=False,
-                      default_value='1',
-                      example=['0', '1', '2'])
+    # parser initialisation
+    parser = argparse.ArgumentParser(
+        description=("Motion correction of fMRI data. Some robust features include:\n"
```
cfhammill, line 31: Are the brackets necessary here? Also out of curiousity, why switch from `"""` to iterated `"` strings? Did the `SmartFormatter` not do the right thing?

scripts/sct_dmri_moco.py
jcohenadad: you're right-- maybe that's a collateral benefit of having added the scripts as entry point in the setup.py (in the past, that snippet was required)

scripts/sct_dmri_display_bvecs.py
jcohenadad: keeping these fields: I see pros (clarity from user standpoint) and cons (redundancy of logic, heavier code, in some functions what is optional/mandatory depends on some flags).

that's a good discussion topic for tmr's SCT meeting

scripts/sct_extract_metric.py
```diff
@@ -55,169 +57,203 @@ def get_parser():
...
+    optional.add_argument(
+        '-method',
+        choices=['ml', 'map', 'wa', 'bin', 'max'],
+        default=param_default.method,
+        help=("R|Method to extract metrics.\n"
+              "  - ml: maximum likelihood (only use with well-defined regions and low noise)\n"
+              "    N.B. ONLY USE THIS METHOD WITH THE WHITE MATTER ATLAS! The sum of all tracts should be 1 in " 
+              "all voxels (the algorithm doesn't normalize the atlas).\n"
+              "  - map: maximum a posteriori. Mean priors are estimated by maximum likelihood within three clusters "
+              "         (white matter, gray matter and CSF). Tract and  noise variance are set with flag -p.\n"
```
jcohenadad, line 133:
```suggestion
              "         (white matter, gray matter and CSF). Tract and noise variance are set with flag -p.\n"
```

jcohenadad, line 133: this line should be displayed with an indent (the indent added in the code does not appear in the terminal):
~~~
                          - map: maximum a posteriori. Mean priors are estimated by maximum likelihood within three clusters
                          (white matter, gray matter and CSF). Tract and  noise variance are set with flag -p.
                            N.B. ONLY USE THIS METHOD WITH THE WHITE MATTER ATLAS! The sum of all tracts should be 1 in all voxels (the
                            algorithm doesn't normalize the atlas).

~~~

scripts/sct_extract_metric.py
jcohenadad: thanks for marking them. Yes, it's a long story. In short: some features were tried and then abandoned. But: some should be revived (perhaps)

scripts/sct_download_data.py
```diff
@@ -14,47 +14,54 @@
...
+              "  (Defaults to ./${dataset-name}. Directory will be created if it does not exist. Warning: existing "
```
jcohenadad, line 50: this indent is unusual (compared to other functions)

scripts/sct_dmri_moco.py
joshuacwnewton: As far as I understand, setting `prog` here is done to avoid the extra `.py` in the help message, right? Unsetting it shows `sct_dmri_moco.py` instead of `sct_dmri_moco`.

> Is this snippet required? In e.g. sct_run_batch the prog argument isn't set explicitly and the usage comes out correct.

Hmm, where are you seeing it not being set explicitly?

https://github.com/neuropoly/spinalcordtoolbox/blob/bf3de687c20ac86717de3584296af1931e247784/scripts/sct_run_batch.py#L55

scripts/sct_dmri_display_bvecs.py
joshuacwnewton: If we remove the argument groups but keep our convention of having everything be keyword arguments (e.g. `-bvecs`), argparse will try to group everything under optional arguments:

```bash
usage: sct_dmri_display_bvecs [-h] [-bvec <file>]

Display scatter plot of gradient directions from bvecs file.

optional arguments:
  -h, --help    show this help message and exit
  -bvec <file>  Input bvecs file. Example: bvecs.txt
```

This is even with `required=True`! I'm not sure there's a native way in argparse to have a "mandatory" keyword argument and have it be displayed properly. 

I think this is because the natural argparse-y way is `mandatory -> positional` and `optional -> keyword`. But, I'm not sure we want positional arguments... _(explicitly named keyword arguments seem important from a user perspective, considering the diversity of arguments we use across our many scripts.)_

joshuacwnewton: Also, I think because of this limitation, that's why we're adding `--help` manually, too.

joshuacwnewton: I could add comments explaining all of this. It definitely isn't standard usage, so it reads a bit funny when doing reviews like this. But, I think it might be an unavoidable consequence of using kwargs for both mandatory and optional arguments.

scripts/sct_dmri_transpose_bvecs.py
joshuacwnewton: See https://github.com/neuropoly/spinalcordtoolbox/pull/2819#discussion_r476119649 for why we're doing this, I think.

scripts/sct_fmri_moco.py
joshuacwnewton: I had the same thought! The trouble is, multiline strings capture all whitespace, so to get it to come out properly, it would need to have some funky indentation in-code:

```python
    parser = argparse.ArgumentParser(
        description="""Motion correction of fMRI data. Some robust features include:
    - group-wise (-g)
    - slice-wise regularized along z using polynomial function (-p)
     (For more info about the method, type: isct_antsSliceRegularizedRegistration)
    - masking (-m)
    - iterative averaging of target volume

The outputs of the motion correction process are:
    - the motion-corrected fMRI volumes
    - the time average of the corrected fMRI volumes
    - a time-series with 1 voxel in the XY plane, for the X and Y motion direction (two separate 
    files), as required for FSL analysis.
    - a TSV file with the slice-wise average of the motion correction for XY (one file), that 
    can be used for Quality Control.""",
        formatter_class=SmartFormatter,
        add_help=None,
        prog=os.path.basename(__file__).strip(".py")
    )
```

I tried finding a "best of both worlds" way to use multiline strings, and arrived at this answer when I was digging into this: https://stackoverflow.com/a/2504457 (Hence why I used `(` and `"` -- mostly for readability.) 

There could definitely be a better way out there, though.

joshuacwnewton:
Thanks much for your attention to detail, @jcohenadad and @cfhammill. I appreciate catching those small spelling/whitespace issues. :)

> https://github.com/neuropoly/spinalcordtoolbox/pull/2819/files#diff-806b4e56303d9d586a963802737d1e7dR180
> 
> this could probably be changed for an input

Apologies, but what is this referencing? The link isn't highlighting a specific LOC for me.

jcohenadad:
> Thanks much for your attention to detail, @jcohenadad and @cfhammill. I appreciate catching those small spelling/whitespace issues. :)
> 
> > https://github.com/neuropoly/spinalcordtoolbox/pull/2819/files#diff-806b4e56303d9d586a963802737d1e7dR180
> > this could probably be changed for an input
> 
> Apologies, but what is this referencing? The link isn't highlighting a specific LOC for me.

oops! sorry, that's because this was pointing to a chunk that was not highlighted as diff by github. I was referring to this line:
https://github.com/neuropoly/spinalcordtoolbox/blob/a67bd5fd74f1aa2eb01fab5ba5feed5fd157ef1f/scripts/sct_maths.py#L180

joshuacwnewton:
Ahh, thanks for pointing that out, @jcohenadad. There's a related issue (#2844) that involves the same lines and is on the 5.0.0 milestone, so I'll include your suggestion with those changes.

scripts/sct_fmri_moco.py
cfhammill: I don't think the `(` is necessary because it is enclosed by the argument bracket - doesn't really matter. I was under the impression that the `SmartFormatter` did the right thing with triple quoted strings. Otherwise there is `textwrap.dedent`, not that anything is wrong with this style.

joshuacwnewton: > I don't think the ( is necessary because it is enclosed by the argument bracket

I didn't realize this was the case, but you're absolutely right. Good to know!

> I don't think it's this PR's job to remove multilabel support from OvR.

I've just checked now, and I couldn't get `SmartFormatter` to respect newlines (with or without `\n`) when using `"""`. Hrm. Well, nbd I think.

joshuacwnewton:
Thanks @jcohenadad! I've made one last commit (just a small code style tweak, shouldn't affect any of the output) but I think this should be good to go. 

Once this and #2842 are merged, I can open PRs for the pending argparse changes I have locally. :smile:

jcohenadad:
@joshuacwnewton can we merge this?

joshuacwnewton:
I was going to ask the same thing! Thanks for the nudge. :)

jcohenadad:
> Does SCT typically use squash and merge, @jcohenadad?

yup!

Issue #2817, opened 2020-07-31
Early termination not working in sct_run_batch


cfhammill:
Setting `-continue-on-error` to 0 doesn't stop the batch from running every subject. Contrary to earlier experiments this appears to be known behaviour of multiprocessing and can be fixed by using `imap` instead of `map`.

This issue will become irrelevant if we switch to `concurrent.futures` and `mpi4py.futures`

cfhammill:
I get partial early stopping behaviour using `concurrent.futures` by wrapping the `map` result iterator in a list comprehension, but strangely it still queues up jobs equal to the number of cpus I have. It's a bit of a strange behaviour (potentially bug), but not too catastrophic I think given that early stopping is not the default behaviour.

kousu:
Do you have a minimal test case we can look at, Chris? I think I ran into a related issue and posted about it here: https://bugs.python.org/issue40110: imap runs infinitely ahead of the consumer. It would fit if conversely error don't percolate at the obvious time either.

kousu:
I think we need to be careful with `multiprocessing.map()`. Unlike plain unix pipes (e.g. in bash) or using temp files, it buffers results in memory. It would be very easy to overwhelm a system on a large dataset.

I haven't looked into `concurrent.futures`' `map`.

jcohenadad:
> Unlike plain unix pipes (e.g. in bash) or using temp files, it buffers results in memory. It would be very easy to overwhelm a system on a large dataset.

ah! so maybe that's related to what i was struggling with: https://github.com/spine-generic/spine-generic/wiki/compute-canada

cfhammill:
@kousu I can work up a small example tonight, I unfortunately deleted my example code.

@kousu and @jcohenadad  no problems regarding memory though it only returns the status codes I think and maybe a few kb of log output. This would be important if we were multiprocessing pipelines in pure python as opposed to external scripts. In relation to memory I have a feeling `-itk_threads` isn't being honoured, in one of my tests one python process was using 600% cpu. But this is another issue

kousu:
bash is a difficult language but it exposes most of the low-level unix concepts very thinly, and the unix concepts are very efficient and powerful. Like thin pipes and redirections and all that.

kousu:
> I have a feeling `-itk_threads` isn't being honoured, in one of my tests one python process was using 600% cpu. But this is another issue

It's not:

[...]

PR #2827, opened 2020-08-03
Fix thread reporting, early termination, and indentation error in sct_run_batch

### LINKED ISSUE #2806: Improve sct_run_batch cluster awareness
Currently, `sct_run_batch` works with OpenMP, not MPI.  In clusters that can accommodate MPI, it would be nice to have that capabilities. In case we will implement it, we could re-use some of the code from an old (now deprecated https://github.com/neuropoly/spinalcordtoolbox/pull/2759) SCT function [`sct_pipeline`](https://github.com/neuropoly/spinalcordtoolbox/blob/09a47b2bed7f13d04bd01a962948883f6ccb7f38/scripts/sct_pipeline.py#L55).

cfhammill:
Converts `sct_run_batch` from using `multiprocessing` to either `concurrent.futures` or `mpi4py.futures` for managing the worker pool. This resolves #2806 and #2817

### DIFF
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -35,6 +35,16 @@ import psutil
 from textwrap import dedent
 from types import SimpleNamespace
 
+if "SCT_MPI_MODE" in os.environ:
+    from mpi4py.futures import MPIPoolExecutor as PoolExecutor
+    __MPI__ = True
+    sys.path.insert(0, path_script)
+else:
+    import concurrent.futures
+    from concurrent.futures import ProcessPoolExecutor as PoolExecutor
+    __MPI__ = False
+
+
 from spinalcordtoolbox.utils import Metavar, SmartFormatter, Tee, send_email, __get_commit
 from spinalcordtoolbox import __version__
 
@@ -403,7 +413,7 @@ def main(argv):
 
     # Trap errors to send an email if a script fails.
     try:
-        with multiprocessing.Pool(jobs) as p:
+        with PoolExecutor(jobs) as p:
             run_single_dir = functools.partial(run_single,
                                                script=script,
                                                script_args=args.script_args,
@@ -415,7 +425,7 @@ def main(argv):
                                                path_qc=path_qc,
                                                itk_threads=args.itk_threads,
                                                continue_on_error=args.continue_on_error)
-            results = p.map(run_single_dir, subject_dirs)
+            results = list(p.map(run_single_dir, subject_dirs))
     except Exception as e:
         if do_email is not None:
             message = ('Oh no there has been the following error in your pipeline:\n\n'
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -342,8 +342,7 @@ def main(argv):
     print('OS: ' + os_running + ' (' + platform.platform() + ')')
 
     # Display number of CPU cores
-    output = int(os.getenv('ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS', 0))
-    print('CPU cores: Available: {} | Used by SCT: {}'.format(multiprocessing.cpu_count(), output))
+    print('CPU cores: Available: {} | Threads used by SCT: {}'.format(multiprocessing.cpu_count(), args.itk_threads))
 
     # Display RAM available
     print("RAM: Total {} MB | Available {} MB | Used {} MB".format(
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -397,6 +397,10 @@ def main(argv):
     else:
         jobs = args.jobs
 
+    print("RUNNING")
+    print("-------")
+    print("Running {} jobs in parallel.\n".format(jobs))
+
     # Run the jobs, recording start and end times
     start = datetime.datetime.now()
 
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -332,7 +332,7 @@ def main(argv):
     print('OS: ' + os_running + ' (' + platform.platform() + ')')
 
     # Display number of CPU cores
-    print('CPU cores: Available: {} | Threads used by SCT: {}'.format(multiprocessing.cpu_count(), args.itk_threads))
+    print('CPU cores: Available: {} | Threads used by ITK Programs: {}'.format(multiprocessing.cpu_count(), args.itk_threads))
 
     # Display RAM available
     print("RAM: Total {} MB | Available {} MB | Used {} MB".format(
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -35,31 +35,6 @@ import psutil
 from textwrap import dedent
 from types import SimpleNamespace
 
-# mpiexec -n 3 python -c "
-# import mpi4py.rc
-# mpi4py.rc.thread_level = 'serialized'
-# from mpi4py import MPI
-# from mpi4py.futures import MPICommExecutor as E
-# with E() as e:
-#   if e is not None:
-#     e.map(print, range(10))
-# "
-
-if "SCT_MPI_MODE" in os.environ:
-    import mpi4py.rc
-    mpi4py.rc.thread_level = 'serialized'
-    from mpi4py.futures import MPICommExecutor
-
-    def PoolExecutor(worker):
-        return MPICommExecutor()
-    
-    __MPI__ = True
-else:
-    import concurrent.futures
-    from concurrent.futures import ProcessPoolExecutor as PoolExecutor
-    __MPI__ = False
-
-
 from spinalcordtoolbox.utils import Metavar, SmartFormatter, Tee, send_email, __get_commit
 from spinalcordtoolbox import __version__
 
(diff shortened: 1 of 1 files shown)


jcohenadad:
@cfhammill is it ready for review? were you able to test it on graham or niagara? if so, what was your test config? (slurm file and script launched)

cfhammill:
It depends on what you think. I think the local futures mode is a drop in replacement for the current functionality, so having someone give that a try would be helpful. 

I haven't been able to get time on niagara yet for testing the MPI mode. I'm planning on trying with an interactive session so that I can experiment quickly and monitor with top, so I don't have a slurm script yet. Once I've validated that the MPI futures are working, I'll try to get a slurm script configured correctly so that we can take advantage of multiple nodes without running out of memory.

I'm planning on doing a mix of testing with the london course and spine-generic.

cfhammill:
After fighting with MPI on niagara for a bit today it looks like the gcc compatible openmpi they have installed has a bug (intel compilers are more completely supported by Compute Canada as far as I can tell). I will report it and wait for an update.

cfhammill:
I think there are useful fixes in this PR that would be worth merging even with the MPI code in an in progress state. Alternatively I can separate the MPI specific bits to a separate PR.

jcohenadad:
> I think there are useful fixes in this PR that would be worth merging even with the MPI code in an in progress state.

Happy with that suggestion @cfhammill . Do you want to open an follow-up issue, turn off draft mode and tag ppl for reviewing it?

cfhammill:
@jcohenadad I think now that getting MPI mode to work would require a substantial rewrite. Niagara doesn't support `MPIPoolExecutor` which is needed for MPI mode to work in the same way as `ProcessPoolExecutor`. 

The alternative is `MPICommExecutor` which has a different semantics. Instead of spawning jobs directly from within python you run `sct_run_batch` prefixed with `mpiexec -n <number of cores>`. This creates n copies of the script, of the n created copies one takes the role of coordinator and the others take the role of workers. 

The problem first problem is this duplicates all output, so the code would need to be refactored in such a way that only the coordinator process produces the main logging output. Second, the executables produced by the python install seem to cause problems for `mpiexec` so you can't run `sct_run_batch` you need to run `scripts/sct_run_batch.py`. This all might be ok but something in the actual script file is triggering a recursive use of mpiexec (somehow), possibly from ANTs, which causes the whole thing to fail. So all told, think adding MPI capabilities would be a substantial undertaking and I think would provide less benefit than refactoring to use some kind of pipeline engine.

Depending on what you want to do with MPI idea, I will either: delete all the MPI related bits from this PR and we can review it as an assortment of small `sct_run_batch` fixes, or I can extract the small fixes to a separate PR and we can leave this one until someone wants to take a crack at it.

jcohenadad:
@cfhammill thank you for looking deeper into this!  We can definitely live without MPI. The spine-generic dataset (250 subjects) took about 4h to run on Graham, which is very reasonable. 

> Depending on what you want to do with MPI idea, I will either: delete all the MPI related bits from this PR and we can review it as an assortment of small sct_run_batch fixes, or I can extract the small fixes to a separate PR and we can leave this one until someone wants to take a crack at it.

to avoid having a stale branch that will stay there for several years, i would go with the former approach: delete all the MPI related bits from this PR and we can review it as an assortment of small sct_run_batch fixes.

cfhammill:
How was spine-generic run on graham? In chunks of a certain number of subjects?

Sounds good re cleaning up the PR, will handle before the end of the week.

jcohenadad:
> How was spine-generic run on graham? In chunks of a certain number of subjects?

yup! 20 subjects

scripts/sct_run_batch.py
```diff
@@ -332,8 +332,7 @@ def reset_streams():
     print('OS: ' + os_running + ' (' + platform.platform() + ')')
 
     # Display number of CPU cores
-    output = int(os.getenv('ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS', 0))
-    print('CPU cores: Available: {} | Used by SCT: {}'.format(multiprocessing.cpu_count(), output))
+    print('CPU cores: Available: {} | Threads used by SCT: {}'.format(multiprocessing.cpu_count(), args.itk_threads))
```
valosekj, line 335: I tried this PR with:

`sct_run_batch -jobs -1 -path-data ~/ds002902-download -path-output ~/ds002902-download_testing -continue-on-error 1 -script /usr/local/lib/spine-generic/process_data.sh`
 
It started computation on 15 cores (16 - 1), but in initial _INFO SYSTEM_ dialog, there is stated "Threads used by SCT: 1". Shouldn´t it be 15?

```
--
Spinal Cord Toolbox (git-cfh/run-batch-mpi_2806-aec57a3015cc3da4b00baa018b44cc95fb73f554)

INFO SYSTEM
-----------
OS: linux (Linux-5.4.0-0.bpo.2-amd64-x86_64-with-debian-10.5)
CPU cores: Available: 16 | Threads used by SCT: 1
RAM: Total 128606 MB | Available 124624 MB | Used 2661 MB
```

cfhammill: Nope, passing `--itk-threads` should override `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS` (which it does slightly later in the code) which should limit the number of threads used by SCT to that number. This is the desired number right @jcohenadad?

jcohenadad: actually, `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS` only concerns the number of threads used for ITK-based functions (eg. ANTs), so I find it counter-intuitive to report "Threads used by SCT" with the variable `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS`. I would rather report the number of cores allocated for `sct_run_batch`, which in this case would be `--jobs`, right?

cfhammill: Ahh, I took that as the intent from the original code which took this number as `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS` but before this env var was modified to match the `--itk-threads` argument. 

Given that the number of jobs is reported in the config, and that the number of threads used by non-ITK programs is unknowable, maybe we should just remove this from the output? 

Although to my knowledge we don't use threading anywhere in the python code, so threading would need to be coming from one of the other binaries.

jcohenadad: > Given that the number of jobs is reported in the config

yes, but i find it nice/reassuring to have it reported when the software runs. Esp. when ppl enter "-1" for jobs, they know for sure the number of cores that is allocated for parallel computing by `sct_run_batch` (does that make sense?).

cfhammill: Makes sense, but we also don't do anything to stop the script from using more cores, so I think this should be reworded as `Running <n> parallel jobs` or something to that effect.

jcohenadad: ah, you're right. 

I was under the (wrong) impression we were, because when looking at the log file of one subject (ie. corresponds to one job), the output of `sct_check_dependencies` (which is ran within the script) gives:
~~~
Spinal Cord Toolbox (git-cfh/add-git-origin-in-sct-run-batch_2808-dd5fc48ad70a30df475e5fb6dc57b6f33e14a538)

SCT info:
- version: git-cfh/add-git-origin-in-sct-run-batch_2808-dd5fc48ad70a30df475e5fb6dc57b6f33e14a538
- path: /Users/julien/code/sct
OS: osx (Darwin-18.7.0-x86_64-i386-64bit)
CPU cores: Available: 16, Used by SCT: 1
~~~

So i thought that the multiprocessing "forced" to run the job with one core. **But**, after looking into `sct_check_dependencies`, the number of cores "Used by SCT" is actually the output of the variable `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS`, which is misleading (and often wrong!). So we should open an issue to address this, and report that the number of cores used by SCT is *not* limited, and specifically report the itk-thread variable (if useful, but it could also be confusing from the user perspective).

jcohenadad: here --> https://github.com/neuropoly/spinalcordtoolbox/issues/2849

jcohenadad: so, similarly for the suggestion in #2849, maybe we could also change the message in `sct_run_batch` for:
~~~
CPU cores: Available: 16 | Used by ITK functions: X
~~~

cfhammill: I think, but I'm not sure, that multiprocessing doesn't implement core pinning, so `multiprocessing.Pool(4)` can launch four subprocesses, which can each go one to spawn jobs that use additional cores.

I'll open an issue for `sct_check_dependencies`

For `sct_run_batch` specifically, do we want to report both `--itk-threads` (`ITK_GLOBAL...` is set to this value for each job) and the computed number of jobs after resolving `0` and negative `-job` arguments?

cfhammill: One additional point, I don't know if I trust all ITK programs to honour this environment variable, so even the `Threads used by ITK programs` may not be correct. Probably generally correct though.

cfhammill: > yes, but i find it nice/reassuring to have it reported when the software runs. Esp. when ppl enter "-1" for jobs, they know for sure the number of cores that is allocated for parallel computing by sct_run_batch 

Sorry for spamming. I've added the ITK threads report back. Do we also want `Running n parallel` jobs in the output, for peace of mind that e.g. `-2` resolved to `14` parallel jobs?

jcohenadad: yes, but maybe not under the “system config” as the job number is more related to user’s input (although the same applies to itk thread which would be reported here). Not sure what’s best here. On the other hand job number is somewhat related to CPU cores...

cfhammill: I added a new header after data called `RUNNING` where it prints the number of jobs running in parallel. Although I can almost see a reason to resolve everything (number of jobs, absolute paths, etc.) before printing the config. So even if you said `-2` jobs, the config will say `14` jobs. This reduces the dependency of the config on the system its run on, but ultimately I don't think it matters much.

scripts/sct_run_batch.py
```diff
@@ -398,6 +397,10 @@ def reset_streams():
     else:
         jobs = args.jobs
```
jcohenadad, line 399: should there be a \n before the RUNNING?

cfhammill: There's one after the origin report above, so it looks right to me.

cfhammill:
Ready to merge I think

PR #2754, opened 2020-06-24
Fixed test_centerline crashing due to bump in Python library

jcohenadad:
Since recently, `test_get_centerline` is [failing on Travis](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/700617618). This is probably related to an upgrade of a Python library and/or platform-dependent (as noted in https://github.com/neuropoly/spinalcordtoolbox/pull/2686)

This PR: 
- Replaced `numpy.polyfit` by `numpy.polynomial.Polynomial` as per [numpy's doc](https://numpy.org/doc/stable/reference/generated/numpy.polyfit.html#numpy-polyfit), in `centerline.curve_fitting.polyfit_1d()`
- Removed: test_get_centerline_nurbs[img_ctl5-expected5-params5] due to a ReconstructionError caused by the change in the dummy centerline generation. 

Fixes #2753

PR #2835, opened 2020-08-07
Allow polynomial function to dummy_segmentation function

valosekj:
This PR addresses creation of synthetic 4D interleaved data discusses here - https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-662041997.

So far, I focused on this comment (https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-664729303) and allowed polynomial function to `dummy_segmentation` function in this commit - https://github.com/neuropoly/spinalcordtoolbox/commit/d5618071f4685eb1de55c39911bb4888aa434fbd. I took inspiration from `dummy_centerline` function.

Is this implementation correct? I can produce following data with this implementation

image

(code used for creation of this data:)
```
from spinalcordtoolbox.testing.create_test_data import dummy_segmentation_4d
DEBUG = True  # Set to True to save images
seg_4D = dummy_segmentation_4d(vol_num=10,create_bvecs=True, size_arr=(20, 20, 30), pixdim=(1,1,1),radius_RL=5.0, radius_AP=3.0, shape='ellipse', interleaved=True, debug=DEBUG)
```


spinalcordtoolbox/testing/create_test_data.py
```diff
@@ -151,19 +151,20 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
...
                 data[:, :, iz] = (((xx - nx / 2) / radius_RL) ** 2 + ((yy - ny / 2) / radius_AP) ** 2 <= 1) * 1
     elif interleaved:
-        # define array based on a polynomial function, within Y-Z plane to simulate slicewise motion in A-P
-        y = np.matlib.repmat([round(nx / 2.) + pixdim[0]*factor, round(nx / 2.) - pixdim[0]*factor], 1, round(nz / 2))
-        if nz % 2 != 0:         # if z-dimension is odd, add one more element to fit size
-            y = numpy.append(y,round(nx / 2.) + pixdim[0]*factor)
-        y = y.reshape(nz)       # reshape to vector (1,R) -> (R,)
-        z = np.arange(0, nz)
-        p = np.poly1d(np.polyfit(z, y, deg=nz))
+        # create regularized curve, within Y-Z plane (A-P), located at x=nx/2, passing through the following points:
```
jcohenadad, line 154: I find it counterintuitive to have the polynomial shape only under interleaved. A clearer approach, would be to allow polynomial shape regardless what interleaved is, and in order to keep the same default behaviour, polynomial order could be set to one by default. 

also note that i am starting to see a lot of similarity with `dummy_centerline()`. At the end, a segmentation is just a centerline with a shape around it (ellipse or rectangle). Would it make sense to merge the two functions? Maybe this is a bad idea…

valosekj: >A clearer approach, would be to allow polynomial shape regardless what interleaved is, and in order to keep the same default behavior, polynomial order could be set to one by default.

Of course, it sounds better. Do we want to move to solely polynomial function (i.e., delete current code and let only code for polynomial) or let both options (polynomial and "_non-polynomial_")? So far, I modified code to choose either from original _non-polynomial_ behavior or polynomial one in this commit: https://github.com/neuropoly/spinalcordtoolbox/pull/2835/commits/6f139927e55d60765aefbd8456dec8144356cc06


> also note that i am starting to see a lot of similarity with `dummy_centerline()`. At the end, a segmentation is just a centerline with a shape around it (ellipse or rectangle). Would it make sense to merge the two functions? Maybe this is a bad idea…

Yeah, there are a lot of similarities between `dummy_centerline()` and `dummy_segmentation()`, so we could probably merge these two functions. But I am not sure, if to start with this merging within this PR, because purpose of this PR should be to create interleaved 4D data to test `sct_dmri_moco -interleaved `.

jcohenadad: > Do we want to move to solely polynomial function (i.e., delete current code and let only code for polynomial) or let both options (polynomial and "non-polynomial")

simplest approach is always the best. if we can achieve the same functionality (and more) with less code, we should always go with less code

jcohenadad: > (current non-polynomial segmentations are used in this unit testing, so probably the second option would be better)

as mentioned above ^ if we can reach the same functionality with more general code, then the unit tests should pass (if they don't pass, it means there was a change in behaviour, which we should avoid-- expect if it affects input arguments, which is manageable)

jcohenadad: > But I am not sure, if to start with this merging within this PR, because purpose of this PR should be to create interleaved 4D data to test sct_dmri_moco -interleaved .

right-- let's do this merge in another PR

valosekj:
After switch to purely polynomial curve in `dummy_segmentation` in last commits, [test_process_seg.py](https://github.com/neuropoly/spinalcordtoolbox/blob/81ebad505180ab18270eb926cca4a134996f8c45/unit_testing/test_process_seg.py) fails for:

https://github.com/neuropoly/spinalcordtoolbox/blob/81ebad505180ab18270eb926cca4a134996f8c45/unit_testing/test_process_seg.py#L82

This fail is caused by slightly different segmentation when using polynomial function (red-yellow) compared to segmentation created without polynomial function (white): 

dummy_seg_angle_AP_20

The second fail is for:

https://github.com/neuropoly/spinalcordtoolbox/blob/81ebad505180ab18270eb926cca4a134996f8c45/unit_testing/test_process_seg.py#L77

(again, polynomially created segmentation is slightly different compared to original one)

spinalcordtoolbox/testing/create_test_data.py
valosekj: > as mentioned above ^ if we can reach the same functionality with more general code, then the unit tests should pass (if they don't pass, it means there was a change in behavior, which we should avoid-- expect if it affects input arguments, which is manageable)

Okay, I tried to reduce and simplify code to let only polynomial function. Unfortunately, unit tests fail (see comment below).

jcohenadad:
@valosekj no worries, you can edit the "expected" values of the unit tests-- it is justified because the produced image is slightly different.

valosekj:
Hm, I found out why segmentations were sometimes slightly different. Problem was with conversion of fitted polynomial curve from float to int. I added rounding before float -> int conversion in last commit https://github.com/neuropoly/spinalcordtoolbox/pull/2835/commits/00af6f51b460ec4a2aa0a2c60e73c86de3711ffe and now unit tests finished successfully.

valosekj:
Now, after allowing regularized curve for `dummy_segmentation()`, I would continue with implementation of suggestions from [this comment](https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-662041997). Specifically, you suggested @jcohenadad:  
> create two synthetic segmentations (e.g. the one produced by dummy_segmentation()), by choosing random scalar for each term of the polynomial function (i.e. y=sum{a_i*x^i}, with a_i ∈ |-1, 1[)

But, bspline which I am using now does return _[B-spline coefficients](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.splrep.html)_ instead of polynomial terms which are returned by polynomial fitting. How to deal with it? When I try to choose random scalar for individual _B-spline coefficients_, it returns to me completely different segmentation.

jcohenadad:
@valosekj maybe you can replace bspline fit by polynomial fit, and https://github.com/neuropoly/spinalcordtoolbox/pull/2754 could be addressed using low order values (e.g. max 2-3). We don't need more than that to synthesize a centerline.

valosekj:
> @valosekj maybe you can replace bspline fit by polynomial fit, and #2754 could be addressed using low order values (e.g. max 2-3). We don't need more than that to synthesize a centerline.

👍 I tried this in the last commit https://github.com/neuropoly/spinalcordtoolbox/pull/2835/commits/b1b52010848955c9c733909161bdecf10a1010c6 and it looks fine (all unit tests passed successfully).

Now, if I want to choose random scalar for each term of the polynomial function, it is probably necessary to do this directly in `curve_fitting.py` script, where I can directly achieve the poly coefficients by `p.coef`:

https://github.com/neuropoly/spinalcordtoolbox/blob/ba63a6212f118024b0aaa2a8a09971239e00da6b/spinalcordtoolbox/centerline/curve_fitting.py#L25-L26

by for example code like this:

```
p = np.polynomial.Polynomial.fit(x, y, deg)
if random_scalar:
        # choose random scalar for each term
        p.coef = [element * random.uniform(-1, 1) for element in p.coef]
return p(xref), p.deriv(1)(xref)
```

Because `polyfit_1d` function then returns fitted curve and coefficients are no longer achievable.

jcohenadad:
> Now, if I want to choose random scalar for each term of the polynomial function, it is probably necessary to do this directly in curve_fitting.py script, where I can directly achieve the poly coefficients by p.coef:

well in that case you can simply bypass `curve_fitting` and reimplement the call to Polynomial (it's literally 2 lines of code)

valosekj:
I tried to implement remaining steps from [this comment](https://github.com/neuropoly/spinalcordtoolbox/pull/2664#issuecomment-662041997) in the last commit https://github.com/neuropoly/spinalcordtoolbox/pull/2835/commits/b7cef137e51c50b2b6d74f54f4156a6879d9ee10. 
If `interleaved=True`, then for each 3D volume, two individuals polynomial fits are created, each with choosing random scalar for each polynomial term. Then, these two polynomials are interleaved (odd and even slices). Finally, segmentation is created based on this interleaved polynomial fit. This is repeated for each 3D volume in 4D data.

Example output:

dummy_4d_interleaved

jcohenadad:
Fantastic work @valosekj ! I let you rebase the branch with master and switch off Draft mode if ready for review

PR #2836, opened 2020-08-07
Fix docs API section rendering.

Drulex:
This fixes the rendering of the `API` section in the rst docs: https://spinalcordtoolbox.com/en/aj-fix-docs-api-render/dev_section/api.html

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -18,12 +18,11 @@
 #
 import os
 import sys
-# sys.path.insert(0, os.path.abspath('.'))
 
 sct_root = os.path.abspath('../../')
 print("sct_root: {}".format(sct_root))
 
-sys.path.insert(0, sct_root)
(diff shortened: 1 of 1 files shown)


jcohenadad:
it doesn't seem to render:

image

Drulex:
same link:

image

Drulex:
ahhh I broke it.  2m

Drulex:
Should be ok now @jcohenadad

jcohenadad:
@Drulex should we merge this PR, and then move to this other PR (because we need to do something with this version.txt file): https://github.com/neuropoly/spinalcordtoolbox/pull/2833

Drulex:
> @Drulex should we merge this PR, and then move to this other PR (because we need to do something with this version.txt file): #2833

Sure, but I suspect that the version problem is part of a bigger effort to properly package the project.

Issue #2750, opened 2020-06-21
"Uninstalling numpy-1.19.0"


kousu:
On [the most recent release](https://github.com/neuropoly/spinalcordtoolbox/releases/tag/4.3), `install_sct` gives this funny message:

```
Installing Python dependencies...

Collecting numpy
  Downloading numpy-1.19.0-cp36-cp36m-manylinux2010_x86_64.whl (14.6 MB)
     |████████████████████████████████| 14.6 MB 67 kB/s 
Installing collected packages: numpy
Successfully installed numpy-1.19.0

Using requirements-freeze.txt (release installation)

Ignoring torch: markers 'sys_platform == "darwin"' don't match your environment
Ignoring torchvision: markers 'sys_platform == "darwin"' don't match your environment
Looking in links: https://download.pytorch.org/whl/cpu/torch_stable.html
[...]
Installing collected packages: six, absl-py, attrs, pytz, python-dateutil, numpy, pandas, pyparsing, packaging, nibabel, docopt, num2words, bids-validator, scipy, patsy, SQLAlchemy, pybids, bids-neuropoly, html5lib, bleach, cachetools, pycparser, cffi, chardet, colored, coverage, cryptography, cycler, decorator, h5py, dipy, future, futures, pyasn1, pyasn1-modules, rsa, google-auth, idna, urllib3, requests, oauthlib, requests-oauthlib, google-auth-oauthlib, grpcio, Pillow, imageio, zipp, importlib-metadata, tifffile, networkx, PyWavelets, kiwisolver, matplotlib, scikit-image, seaborn, typing-extensions, protobuf, onnx, onnxruntime, torch, torchvision, tqdm, pluggy, py, more-itertools, wcwidth, pytest, Werkzeug, Markdown, tensorboard-plugin-wit, tensorboard, joblib, threadpoolctl, scikit-learn, ivadomed, PyYAML, Keras, pyOpenSSL, PyQt5-sip, PyQt5, pytest-cov, raven, requirements-parser, tensorflow-tensorboard, tensorflow, transforms3d, xlwt
  Attempting uninstall: numpy
    Found existing installation: numpy 1.19.0
    Uninstalling numpy-1.19.0:
      Successfully uninstalled numpy-1.19.0
```

It doesn't hurt the functioning, but it makes the installer undo its own work.

kousu:
This is because

[...]

PR #2840, opened 2020-08-14
Pre-install numpy to build.

### LINKED ISSUE #2841: Travis build failing on macOS Sierra (10.12)
https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/717990889

should we drop support?

kousu:
Travis started failing on macOS 10.12 a few weeks ago, e.g. https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/717668875. Due to changes I made earlier this summer (#2690) this only nagged me once a day when the cron build kicked in and I ignored it. But now I'm here to quiet it.

Here's the error:

> ```
> Collecting requirements-parser
>  Downloading requirements-parser-0.2.0.tar.gz (6.3 kB)
> Collecting scipy
>  Downloading scipy-1.5.2-cp36-cp36m-macosx_10_9_x86_64.whl (28.8 MB)
> Collecting scikit-image
>  Downloading scikit-image-0.17.2.tar.gz (29.8 MB)
>
>    ERROR: Command errored out with exit status 1:
>     command: /Users/travis/build/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-install-d6qkfes3/scikit-image/setup.py'"'"'; __file__='"'"'/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-install-d6qkfes3/scikit-image/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-pip-egg-info-ekhwucq2
>         cwd: /private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-install-d6qkfes3/scikit-image/
>    Complete output (7 lines):
>    Traceback (most recent call last):
>      File "<string>", line 1, in <module>
>      File "/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-install-d6qkfes3/scikit-image/setup.py", line 234, in <module>
>        'build_ext': openmp_build_ext(),
>      File "/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/pip-install-d6qkfes3/scikit-image/setup.py", line 58, in openmp_build_ext
>        from numpy.distutils.command.build_ext import build_ext
>    ModuleNotFoundError: No module named 'numpy'
>    ----------------------------------------
> ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
> ```

The clue that this only happens on 10.12 and nowhere else is important. Notice how just before, it says "Downloading scipy .... something.whl" and then "Downloading scikit-learn ... something.tar.gz"? `.whl` is python's pre-compiled binary package format, which means that .tar.gz is probably the source dist. Looking at https://pypi.org/project/scikit-image/#files we can see that scikit-learn builds for Windows32, Windows64, macOS 10.13, and "Linux" (which [as I have yet to properly document](https://github.com/neuropoly/spinalcordtoolbox/issues/2669#issuecomment-665230020), means CentOS6). So because scikit-image isn't providing Sierra binaries, we can't install ourselves there in any reliable way.

And looking at [that source](https://github.com/scikit-image/scikit-image/blob/59fd9682aad41a1f69a35f05d2a7e15ebdfb8b29/setup.py#L57) we see that their setup.py tries to import numpy, which is because it wants to make use of [`numpy.distutils`](https://numpy.org/doc/stable/f2py/distutils.html), some extensions to python's build system to support linking in Fortran code.

Okay, fair enough, but why can't numpy be declared as a build dependency for it? Lots of other packaging systems make this work. But it looks like that can't work here because setup() expects an [instantiated builder object](https://github.com/scikit-image/scikit-image/blob/59fd9682aad41a1f69a35f05d2a7e15ebdfb8b29/setup.py#L234) and the most obvious way for `numpy.distutils` to provide that is to let itself [be subclassed](https://github.com/scikit-image/scikit-image/blob/59fd9682aad41a1f69a35f05d2a7e15ebdfb8b29/setup.py#L66). So there's a chicken-egg problem here.

PyPA is aware that numpy.distutils is pretty fragile w.r.t. being itself a build-time dependency: https://www.python.org/dev/peps/pep-0518/#rationale
> [..] which means that projects such as numpy.distutils are largely incapable of utilizing it [..]

but lots of packaging systems manage to have separate build/runtime dependencies. So maybe this is `numpy.distutils`'s fault. Or maybe design choices setuptools made forced numpy.distutils to be like this?

Anyway the upshot of all this is that it turns out there was a good reason for installing numpy explicitly, the line that we removed in 2bb19c5c9ad9dfc5919c4431613faa54f806ee2d, and we can't fix any of it from here. But I can readd the build dep and this time at least mark it as one (except, that, according to pep518, it's already old news to do it this way :? 🗑️🔥)

Fixes #2841.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,6 +112,7 @@ matrix:
       os: osx
       osx_image: xcode9.2
     - name: "Windows Server, 1809"
+      # runs on all branches
       language: bash
       os: windows
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
-numpy # need to install first otherwise dependency issues
 colored
 dipy
 futures
@@ -7,6 +6,7 @@ ivadomed==2.0.2
 Keras==2.1.5
 matplotlib
 nibabel
+numpy
 pandas
 psutil
 pyqt5==5.11.3
--- a/install_sct
+++ b/install_sct
@@ -529,6 +529,11 @@ conda activate venv_sct
 
 # Install Python dependencies
 print info "Installing Python dependencies..."
+# numpy is an undeclared build dependency: https://github.com/scikit-image/scikit-image/issues/4919
+# so explicitly install it as a workaround; note that in releases with requirements-freeze.txt
+# this can cause a strange double-install: https://github.com/neuropoly/spinalcordtoolbox/issues/2750
+# but we can live with that until https://github.com/scikit-image/scikit-image/pull/4920 is merged.
+pip install numpy
 # Check if a frozen version of the requirements exist (for release only)
 if [[ -f "requirements-freeze.txt" ]]; then
   print info "Using requirements-freeze.txt (release installation)"


kousu:
**EDIT** : my proposed solution did not work. I guess `setup_requires` only applies when building or installing the package it's applied to -- I think pip makes a virtualenv for each build, now.

I don't know how to fix this. @Drulex ? @zougloub ?

kousu:
Maybe we *can* get scikit-image patched somehow? It seems silly that to support OpenMP -- an entirely optional feature -- they are forced to load numpy at build time, and not just at build time but before they themselves can even invoke `setup_requires`. There's gotta be a way to rejig it. What if instead of inheritence, it was a proxy to `numpy.distutils.command.build_ext`, and it lazily-loaded that object into a `@functools.lru_cache`'d `@property`?

Drulex:
The `numpy` dependency needs to be installed prior to building `scikit-image`, because it is a build time dependency. Adding the dependency to `setup_requires` has no effect, because `pip install -r requirements.txt` is ran before the `setup.py` and we are trying to build `scikit-image` from source before that.

I think that if we are trying to build stuff from source it's on us to make sure the required dependencies are installed. Although [this](https://github.com/neuropoly/spinalcordtoolbox/commit/2bb19c5c9ad9dfc5919c4431613faa54f806ee2d) wasn't very elegant, it did satisfy the build time dependency.

Drulex:
Related to https://github.com/scikit-learn/scikit-learn/issues/4164. As you can see, the temporary fix proposed was basically running `pip install numpy` prior to the scikit-image setup: https://github.com/scikit-learn/scikit-learn/issues/4164#issuecomment-243259447.

kousu:
> Related to [scikit-learn/scikit-learn#4164](https://github.com/scikit-learn/scikit-learn/issues/4164). As you can see, the temporary fix proposed was basically running `pip install numpy` prior to the scikit setup: [scikit-learn/scikit-learn#4164 (comment)](https://github.com/scikit-learn/scikit-learn/issues/4164#issuecomment-243259447).

Thanks for finding those so quickly.

D: and this is what we used to do. But there was no comment explaining it so I didn't know and we got rid of it because it was causing a double-install (https://github.com/neuropoly/spinalcordtoolbox/issues/2750).

kousu:
I took a look at how scipy handles this. It turns out they're using the new, unstable?, [pep517](https://pypi.org/project/pep517/). That's where `setup_requires` is migrating to:

https://github.com/scipy/scipy/blob/2ed8169c5c4c9181cd38a52db614c692eb802c15/pyproject.toml#L6-L8

I copied this into this branch to see if it will work. And I think I know how to patch scikit-image now, at least, so that it works for everyone.

kousu:
I tried using

```
setup(
  # ...
   setup_requires=[
       'numpy',  # needed to build scikit-learn on platforms scikit-learn isn't packaged as a .whl
   ],
  # ...
```

But it didn't work.

I tried adding the New Way To Do Build Requirements

```
# pyproject.toml
[build-system]
requires = [
    "wheel",
    "setuptools",
    "numpy==1.14.5; python_version=='3.6' and platform_system!='AIX'",
    "numpy==1.14.5; python_version=='3.7' and platform_system!='AIX'",
    "numpy==1.17.3; python_version>='3.8' and platform_system!='AIX'",
    "numpy==1.16.0; python_version=='3.6' and platform_system=='AIX'",
    "numpy==1.16.0; python_version=='3.7' and platform_system=='AIX'",
    "numpy==1.17.3; python_version>='3.8' and platform_system=='AIX'",
]
```

but it didn't work either.

In both cases pip does isolated builds for each dependency in turn, so the build environment for spinalcordtoolbox can't control that for scikit-image. Which makes sense, but I was hoping.

This needs to be fixed over in scikit-image.

kousu:
I've opened an issue with sklearn about it (https://github.com/scikit-image/scikit-image/issues/4919) and this branch now reapplies the old workaround, this time with a better explanation.

.travis.yml
```diff
@@ -112,6 +112,7 @@ matrix:
       os: osx
       osx_image: xcode9.2
     - name: "Windows Server, 1809"
+      # runs on all branches
```
joshuacwnewton, line 115: Is this relevant to this PR? Or is this one of those comments that's too small to create another PR for, so it's best to just bundle it into this one?

requirements.txt
```diff
@@ -1,4 +1,3 @@
-numpy # need to install first otherwise dependency issues
```
joshuacwnewton, line 1: Based on the commit message, I believe this change was to address #2750, right? 

But, I'm not sure this change to `requirements.txt` actually helps us here... I believe #2750 happens because `pip install numpy` installs the newest version, but `requirements-freeze.txt` pins an older version for reproducibility, so the newer version gets uninstalled to make room for the pinned version. 

Here, with simple `requirements.txt`, there ends up being two `pip install numpy` (no version pinning), so I imagine there won't be a double-install. Hence, no need to remove this line.

.travis.yml
kousu: Yeah I'm slipping it in. Look it's even got its own commit message, but it didn't seem worth another PR.

requirements.txt
kousu: Ah that's a good point. I didn't test how this interacts with `requirements-freeze.txt` because I was too deep in pip specs.

Another fix would be to version-pin numpy in requirements.txt to prevent freeze from being able to disagree and overriding it.

---

But my main thinking here was that we should only write down numpy in one place.

---

I'm also working on fixing this upstream in https://github.com/scikit-image/scikit-image/issues/4919

kousu:
> Two general questions/comments:
> 
>     * Out of curiosity, how do we know this solution works? I can't tell because the commit history was rewritten... was there a time where the Sierra build in `.travis.yml` was enabled (then changed back)?

Because it only shows up on our non-mainstream platforms, I hacked up .travis.yml to make Sierra the only platform that ran.

>     * On that note, would it have been better not to force-push to rewrite history? Although those past commits didn't end up working, it could have maybe been good to retain the history (incl. the links to the Travis error logs.) It might have made this PR a little easier to follow, and it provides good context if we ever return to this issue in the future.

But they were soooo messssyyy. Though I thought github did hang on to previous versions, at least while a PR was open. Sorry. I would have linked the Travis logs if I'd double-checked. 

It's hard to debug these packaging things because they're hard to test properly without doing a real merge and a real release.

requirements.txt
kousu: Okay I was swayed. I've replaced removing this with a comment with all the backstory in install_sct. What do you think?

kousu:
If you want to test this, look at https://github.com/neuropoly/spinalcordtoolbox/tree/ng/macOS-sierra-build-test. I hacked up .travis.yml to cause https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/718240514;

requirements.txt
joshuacwnewton: That seems reasonable to me. :slightly_smiling_face:

kousu:
It did! And as you might expect, Sierra now takes 10 minutes longer because it has to compile scikit-image:

Screenshot_2020-08-15_16-37-54

Drulex:
@kousu note that this PR caused an obscure side-effect for a user who tried to install `sct` on an NFS drive (https://forum.spinalcordmri.org/t/installing-sct-4-3-fails-pycache-not-empty/493).

It may be an isolated/rare occurrence, thus probably not worth re-opening this can of worms right now. We should definitely monitor this in the future in case we need to revisit the topic though.

PR #2842, opened 2020-08-14
Convert 11-20 out of 20 remaining scripts from msct_parser to argparse

### LINKED ISSUE #1548: Replace usage of custom command-line parser (`msct_parser`) with argparse

joshuacwnewton:
### Reference issues/PRs

Follows #2315, #2337, and #2819. 

Partially addresses #1548. (When combined with #2819. But, there may be some additional clean-up needed considering the scope of that issue.)

### What does this implement/fix? Explain your changes.

This PR converts the last 10 of 20 remaining scripts that still use `msct_parser`. See below for progress on task list.

 - [x] `/scripts/sct_get_centerline.py`
 - [x] `/scripts/sct_label_utils.py`
 - [x] `/scripts/sct_label_vertebrae.py`
 - [x] `/scripts/sct_process_segmentation.py`
 - [x] `/scripts/sct_propseg.py`
 - [x] `/scripts/sct_register_multimodal.py`
 - [x] `/scripts/sct_register_to_template.py`
 - [x] `/scripts/sct_resample.py`
 - [x] `/scripts/sct_smooth_spinalcord.py`
 - [x] `/scripts/sct_warp_template.py`

### Additional notes

This PR depends on changes in #2819. That PR adds a `list_type` function which can be used to parse arg strings formatted as delimited lists.


joshuacwnewton:
The build is currently failing because this PR depends on changes in #2819 which haven't been merged yet.

scripts/sct_create_mask.py
```diff
@@ -299,13 +299,14 @@ def create_line(param, fname, coord, nz):
...
+    create_add_str = ""
     for iz in range(nz):
         if iz == nz - 1:
-            cmd += [str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1']
+            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1'
         else:
-            cmd += [str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1:']
+            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1:'
 
+    cmd = ['sct_label_utils', '-i', 'line.nii', '-o', 'line.nii', '-create-add', create_add_str]
```
cfhammill, line 309: Is this a performance optimization? Or does it have to do with `argparse` interpreting the list version as too many arguments?

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
-                      description='If provided, this string will be mentioned in the QC report as the dataset the process was run on',
-                      )
-    parser.add_option(name='-qc-subject',
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the subject the process was run on',
-                      )
+    parser = argparse.ArgumentParser(
+        description=(
+            "This function extracts the spinal cord centerline. Three methods are available: OptiC (automatic), "
+            "Viewer (manual) and Fitseg (applied on segmented image). These functions output (i) a NIFTI file with "
```
cfhammill, line 23:
```suggestion
            "Viewer (manual), and Fitseg (applied on segmented image). These functions output (i) a NIFTI file with "
```

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the dataset the process was run on',
-                      )
-    parser.add_option(name='-qc-subject',
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the subject the process was run on',
-                      )
+    parser = argparse.ArgumentParser(
+        description=(
+            "This function extracts the spinal cord centerline. Three methods are available: OptiC (automatic), "
```
cfhammill, line 22: @jcohenadad here it says `OptiC` but in the argument it documents and wants it as `optic`. Should we keep the methods in both places?

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+              "  - viewer: manual selection a few points followed by interpolation\n"
+              "  - fitseg: fit a regularized centerline on an already-existing cord segmentation. It will "
+              "interpolate if slices are missing and extrapolate beyond the segmentation boundaries (i.e., every "
+              "axial slice will exhibit a centerline pixel).")
+    )
+    optional.add_argument(
+        "-centerline-algo",
+        choices=['polyfit', 'bspline', 'linear', 'nurbs'],
+        default='bspline',
+        help="Show this help message and exit."
```
cfhammill, line 70:
```suggestion
        help='Algorithm for centerline fitting. Only relevant with -method fitseg'
```

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+    )
+    optional.add_argument(
+        "-qc",
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        "-qc-dataset",
+        metavar=Metavar.str,
```
cfhammill, line 111: `Metavar.str` is the default isn't it?

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        "-qc-subject",
+        metavar=Metavar.str,
```
cfhammill, line 116:
```suggestion
   
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
-    parser.add_option(name='-qc-dataset',
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the dataset the process was run on',
-                      )
-    parser.add_option(name='-qc-subject',
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the subject the process was run on',
-                      )
+    parser = argparse.ArgumentParser(
+        description="Utility function for label image.",
```
cfhammill, line 687:
```suggestion
        description="Utility function for label images.",
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+        formatter_class=SmartFormatter,
```
cfhammill, line 688: We have this everywhere, maybe not worth the added mental overhead but we could have and `sct_parser` function that wraps `argparse.ArgumentParser` and sets this at the default

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+            "Viewer (manual) and Fitseg (applied on segmented image). These functions output (i) a NIFTI file with "
+            "labels corresponding to the discrete centerline, and (ii) a csv file containing the float (more precise) "
+            "coordinates of the centerline in the RPI orientation. \n"
+            "\n"
+            "Reference: C Gros, B De Leener, et al. Automatic spinal cord localization, robust to MRI contrast using "
+            "global curve optimization (2017). doi.org/10.1016/j.media.2017.12.001"
+        ),
+        formatter_class=SmartFormatter,
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
```
cfhammill, line 32: is the verdict in on whether this is needed?

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+
+    mandatory = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatory.add_argument(
+        '-i',
+        metavar=Metavar.file,
+        required=True,
+        help="Input image. Example: tw_labels.nii.gz"
```
cfhammill, line 698:
```suggestion
        help="Input image. Example: t2_labels.nii.gz"
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+    )
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 832:
```suggestion
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 837:
```suggestion
```

scripts/sct_label_vertebrae.py
```diff
@@ -59,116 +61,162 @@ def get_parser():
...
+        '-denoise',
+        choices=['0', '1'],
+        default='0',
+        help="Apply denoising filter to the data. Sometimes denoising is too aggressive, so use with care."
+    )
+    optional.add_argument(
+        '-laplacian',
+        choices=['0', '1'],
+        default='0',
+        help="Apply Laplacian filtering. More accuracy but could mistake disc depending on anatomy."
```
cfhammill, line 165:
```suggestion
        help="Apply Laplacian filtering. More accurate but could mistake disc depending on anatomy."
```

scripts/sct_label_vertebrae.py
```diff
@@ -59,116 +61,162 @@ def get_parser():
...
+        metavar=Metavar.file,
+        action=ActionCreateFolder,
+        default='',
+        help=("Output folder.")
+    )
+    optional.add_argument(
+        '-denoise',
+        choices=['0', '1'],
+        default='0',
+        help="Apply denoising filter to the data. Sometimes denoising is too aggressive, so use with care."
```
cfhammill, line 159: What denoising filter does this use @jcohenadad?

scripts/sct_label_vertebrae.py
```diff
@@ -59,116 +61,162 @@ def get_parser():
...
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        default=param_default.path_qc,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 211:
```suggestion
```

scripts/sct_label_vertebrae.py
```diff
@@ -59,116 +61,162 @@ def get_parser():
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 216:
```suggestion
```

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+              "same as outputting a single metric at once across all slices.")
+    )
+    optional.add_argument(
+        '-vert',
+        metavar=Metavar.str,
+        help="Vertebral levels to compute the metrics across. Example: 2:9 for C2 to T2."
+    )
+    optional.add_argument(
+        '-vertfile',
+        metavar=Metavar.str,
```
cfhammill, line 114:
```suggestion
```

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+        '-perslice',
+        choices=['0', '1'],
+        default='0',
+        help=("Set to 1 to output one metric per slice instead of a single output metric. Please note that when "
+              "methods ml or map is used, outputing a single metric per slice and then averaging them all is not the "
+              "same as outputting a single metric at once across all slices.")
+    )
+    optional.add_argument(
+        '-vert',
+        metavar=Metavar.str,
```
cfhammill, line 109:
```suggestion
```

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+    )
+    optional.add_argument(
+        '-append',
+        choices=[0, 1],
+        default=0,
+        help="Append results as a new line in the output csv file instead of overwriting it."
+    )
+    optional.add_argument(
+        '-z',
+        metavar=Metavar.str,
```
cfhammill, line 95:
```suggestion
```

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 166:
```suggestion
```

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+    )
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 161:
```suggestion
```

scripts/sct_propseg.py
```diff
@@ -146,205 +147,240 @@ def check_and_correct_segmentation(fname_segmentation, fname_centerline, folder_
...
-                                  "consistent with the centerline provided by isct_propseg.",
-                      mandatory=False,
-                      example=["0", "1"],
-                      default_value="1")
-    parser.add_option(name='-igt',
-                      type_value='image_nifti',
-                      description='File name of ground-truth segmentation.',
-                      mandatory=False)
-
-    # DEPRECATED OPTIONS
```
cfhammill, line 322: @jcohenadad are these deprecated options clear for deletion?

scripts/sct_propseg.py
```diff
@@ -146,205 +147,240 @@ def check_and_correct_segmentation(fname_segmentation, fname_centerline, folder_
...
+    )
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 363:
```suggestion
```

scripts/sct_propseg.py
```diff
@@ -146,205 +147,240 @@ def check_and_correct_segmentation(fname_segmentation, fname_centerline, folder_
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 368:
```suggestion
```

scripts/sct_register_multimodal.py
```diff
@@ -52,162 +53,194 @@ def get_parser(paramregmulti=None):
...
+    )
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 224:
```suggestion
```

scripts/sct_register_multimodal.py
```diff
@@ -52,162 +53,194 @@ def get_parser(paramregmulti=None):
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 229:
```suggestion
```

scripts/sct_register_to_template.py
```diff
@@ -65,133 +68,210 @@ def __init__(self):
...
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        default=param.path_qc,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 249:
```suggestion
```

scripts/sct_register_to_template.py
```diff
@@ -65,133 +68,210 @@ def __init__(self):
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 254:
```suggestion
```

scripts/sct_warp_template.py
```diff
@@ -139,79 +140,103 @@ def get_parser():
...
+    optional.add_argument(
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        default=param_default.path_qc,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-dataset',
+        metavar=Metavar.str,
```
cfhammill, line 205:
```suggestion
```

scripts/sct_warp_template.py
```diff
@@ -139,79 +140,103 @@ def get_parser():
...
+        help="If provided, this string will be mentioned in the QC report as the dataset the process was run on."
+    )
+    optional.add_argument(
+        '-qc-subject',
+        metavar=Metavar.str,
```
cfhammill, line 210:
```suggestion
```

joshuacwnewton:
Thank you so much for taking the time to do a thorough review, @cfhammill! You're absolutely right about it being exhausting. It's quite a tedious thing to review, so I'm especially thankful for your input. :)

scripts/sct_create_mask.py
joshuacwnewton: It's closer to the latter, but not quite.

With the previous way, each coordinate would be added as a separate list element, so there ended up being an erroneous space between each coordinate. (i.e. You'd get `x1,y1,z1,val1: x2,y2,z2,val2` instead of the expected `x1,y1,z1,val1:x2,y2,z2,val2`.) This caused some parsing issues IIRC.

scripts/sct_get_centerline.py
joshuacwnewton: This is embarrassing! Nice catch. :stuck_out_tongue_closed_eyes:

scripts/sct_get_centerline.py
joshuacwnewton: If `metavar=Metavar.str` isn't provided, then the example value will instead be:

```
  -qc-dataset QC_DATASET
                        If provided, this string will be mentioned in the QC report as the dataset the
                        process was run on.
```

compared to

```
  -qc-dataset <str>     If provided, this string will be mentioned in the QC report as the dataset the
                        process was run on.
```

I'm happy with either, depending on preference!

scripts/sct_get_centerline.py
joshuacwnewton: I got a :+1: in https://github.com/neuropoly/spinalcordtoolbox/pull/2819#discussion_r476097347 but feel free to let me know what you think about the reasoning, too. :slightly_smiling_face:

scripts/sct_label_vertebrae.py
joshuacwnewton: Looks like this flag is used in a call to `sct_maths`. The description there provides some extra contex, I thinkt:

https://github.com/neuropoly/spinalcordtoolbox/blob/004b3262d6cde2209d6af8c96f7e03370ada0928/scripts/sct_maths.py#L185-L192

scripts/sct_label_utils.py
joshuacwnewton: You're right, that same pattern is used everywhere, and it is a bit repetitive.

On the other hand, this way of doing it nice because what exactly is being set is clear at the location the parser is defined. If we used `sct_parser`, it wouldn't be immediately obvious what's different about it from the normal parser, so you'd have to go to the source to find out: "Ah! `sct_parser` just sets `formatter_class`."

scripts/sct_get_centerline.py
jcohenadad: i like the idea of showing `<str>`, it aligns with the convention for the other flags (and QC_DATASET is not more informative)

scripts/sct_label_utils.py
joshuacwnewton: In a theoretical `sct_parser`, there are other defaults that could be set, too. e.g. we could have common arguments like `verbose` and `remove_temp` (`-v` and `-r`) added there too.

scripts/sct_get_centerline.py
jcohenadad: i haven't tested-- if you test it and see that the ".py" is present without this line of code, then it should be kept

scripts/sct_label_vertebrae.py
jcohenadad: good point-- we should probably mention it:
~~~
help="Apply denoising filter (non-local means adaptative denoising) to the data. Sometimes denoising is too aggressive, so use with care."
~~~

scripts/sct_get_centerline.py
joshuacwnewton: Just to make sure I understand, is the concern that using the exact name of the method [(`OptiC`)](https://link.springer.com/chapter/10.1007/978-3-319-66185-8_80) in the description might cause some case sensitivity issues from a user perspective, e.g. they might think to type in `-method OptiC`?

jcohenadad: i would keep this option the way it is (ie: optic) for backward compatibility (also, it is faster to type)

scripts/sct_get_centerline.py
cfhammill: I find the default behaviour preferable, because it requires less discipline and is more similar to output you would get from other python programs using argparse. I also doubt that most string type arguments are given an explicit metavar across all sct scripts, so this convention would only be partially followed.

cfhammill: I don't feel strongly however

scripts/sct_label_utils.py
cfhammill: > so you'd have to go to the source to find out: "Ah! sct_parser just sets formatter_class but nothing else."

This is what I meant by overhead. I think the repeating this argument is probably preferable, but you are right there are more common arguments we could be setting. If we chose to go with a custom class we could have `optional` and `required` as part of the parser instead of separate groups, so we could do e.g. `parser.optional.add_argument`

cfhammill: Resolving, because this is good for now (possibly forever)

scripts/sct_get_centerline.py
jcohenadad: > I find the default behaviour preferable, because it requires less discipline and is more similar to output you would get from other python programs using argparse. 

This is a fair argument. I was putting myself in the shoes of a typical user of SCT (coming from the neuroimaging community), as opposed to a typical user of other Python software. E.g., I like to often refer to FSL CLI, because SCT users are used to FSL. For example, flirt output shows the type (or the explicit meaning of an input by "paraphrasing" it), example for `flirt`:
~~~
        -in  <inputvol>                    (no default)
        -ref <refvol>                      (no default)
        -init <matrix-filname>             (input 4x4 affine matrix)
        -omat <matrix-filename>            (output in 4x4 ascii format)
        -out, -o <outputvol>               (default is none)
        -datatype {char,short,int,float,double}                    (force output data type)
~~~

> I also doubt that most string type arguments are given an explicit metavar across all sct scripts, so this convention would only be partially followed.

Indeed, most of SCT's conventions are only partially applied because some are fairly recent and we didn't have the chance to fully implement them. Given that the dev team is relatively small, I see the application of convention in a more prospective manner rather than retrospectively (and exhaustively).

Drulex:
How close are we to merge this? I am playing in some scripts (currently `sct_label_utils.py` and `sct_register_to_template.py`). 

I noticed that some changes I was about to implement are already implemented here so I guess I could rebase off this branch to make my life easier and avoid potential conflicts for @joshuacwnewton's later?

joshuacwnewton:
@Drulex I believe @cfhammill has given this an informal approval (on Slack), but I've also put it on today's meeting agenda to discuss. It affects a few other in-progress PRs, too, so the sooner it gets merged, the better, I think.

jcohenadad:
@joshuacwnewton have you tried to run `batch_processing` with this branch? it sometimes detects glitches in the CLI (things that are poorly tested with our testing suite)

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+    optional.add_argument(
+        "-c",
+        choices=['t1', 't2', 't2s', 'dwi'],
+        help="Type of image contrast. Only with method=optic."
+    )
+    optional.add_argument(
+        "-method",
+        choices=['optic', 'viewer', 'fitseg'],
+        default='optic',
+        help=("Method used for extracting the centerline.\n"
```
jcohenadad, line 59: in my terminal, this displays as:
~~~
  -method {optic,viewer,fitseg}
                        Method used for extracting the centerline. - optic: automatic spinal cord
                        detection method - viewer: manual selection a few points followed by
                        interpolation - fitseg: fit a regularized centerline on an already-existing
                        cord segmentation. It will interpolate if slices are missing and extrapolate
                        beyond the segmentation boundaries (i.e., every axial slice will exhibit a
                        centerline pixel).
~~~

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
+        add_help=None,
+        prog=os.path.basename(__file__).strip(".py")
+    )
+
+    mandatory = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatory.add_argument(
+        '-i',
+        metavar=Metavar.file,
+        required=True,
+        help="Input image. Example: ti.nii.gz"
```
jcohenadad, line 40:
```suggestion
        help="Input image. Example: t1.nii.gz"
```

jcohenadad:
It might be out of scope, but sct_create_mask shows options like this:
~~~
  -p <str>              Process to generate mask.
                          <coord,XxY>: Center mask at the X,Y coordinates. (e.g. "coord,20x15")
                          <point,FILE>: Center mask at the X,Y coordinates of the label defined in
                          input volume FILE. (e.g. "point,label.nii.gz")
                          <center>: Center mask in the middle of the FOV (nx/2, ny/2).
                          <centerline,FILE>: At each slice, the mask is centered at the spinal cord
                          centerline, defined by the input segmentation FILE. (e.g.
                          "centerline,t2_seg.nii.gz")
~~~

whereas it would be nice to have an additional indent for line return:
~~~
  -p <str>              Process to generate mask.
                          <coord,XxY>: Center mask at the X,Y coordinates. (e.g. "coord,20x15")
                          <point,FILE>: Center mask at the X,Y coordinates of the label defined in
                            input volume FILE. (e.g. "point,label.nii.gz")
                          <center>: Center mask in the middle of the FOV (nx/2, ny/2).
                          <centerline,FILE>: At each slice, the mask is centered at the spinal cord
                            centerline, defined by the input segmentation FILE. (e.g.
                            "centerline,t2_seg.nii.gz")
~~~

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+        '-add',
+        metavar=Metavar.int,
+        type=int,
+        help="Add value to all labels. Value can be negative."
+    )
+    optional.add_argument(
+        '-create',
+        metavar=Metavar.list,
+        type=list_type(':', Coordinate),
+        help=("Create labels in a new image. List labels as: x1,y1,z1,value1:x2,y2,z2,value2."
```
jcohenadad, line 718:
```suggestion
        help=("Create labels in a new image. List labels as: x1,y1,z1,value1:x2,y2,z2,value2. "
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+              "Example: 12,34,32,1:12,35,33,2")
+    )
+    optional.add_argument(
+        '-create-add',
+        metavar=Metavar.list,
+        type=list_type(':', Coordinate),
+        help=("Same as '-create', but add labels to the input image instead of creating a new image."
```
jcohenadad, line 725:
```suggestion
        help=("Same as '-create', but add labels to the input image instead of creating a new image. "
```

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
+        help=("Takes all non-zero values, sort them along the inverse z direction, and attributes the values "
+              "1, 2, 3, etc.")
+    )
+    optional.add_argument(
+        '-vert-body',
+        metavar=Metavar.list,
+        type=list_type(',', int),
+        help=("R|From vertebral labeling, create points that are centered at the mid-vertebral levels. Separate "
+              "desired levels with ','. Example: 3,8\n"
+              "To get all levels, enter '0'.")
```
jcohenadad, line 771:
```suggestion
              "To get all levels, enter 0.")
```
in case people write '0'

scripts/sct_register_multimodal.py
```diff
@@ -53,162 +54,194 @@ def get_parser(paramregmulti=None):
...
+    )
+    optional.add_argument(
+        '-iseg',
+        metavar=Metavar.file,
+        help="Segmentation source. Example: src_seg.nii.gz"
+    )
+    optional.add_argument(
+        '-dseg',
+        metavar=Metavar.file,
+        help="Segmentation destination. dest_seg.nii.gz"
```
jcohenadad, line 99:
```suggestion
        help="Segmentation destination. Example: dest_seg.nii.gz"
```

scripts/sct_register_multimodal.py
```diff
@@ -53,162 +54,194 @@ def get_parser(paramregmulti=None):
...
+    )
+    optional.add_argument(
+        '-param',
+        metavar=Metavar.list,
+        type=list_type(':', str),
+        help=(f"R|Parameters for registration. Separate arguments with \",\". Separate steps with \":\".\n"
+              f"Example: step=1,type=seg,algo=slicereg,metric=MeanSquares:step=2,type=im,algo=syn,metric=MI,iter=5,"
+              f"shrink=2\n"
+              f"  - step: <int> Step number (starts at 1, except for type=label).\n"
+              f"  - type: {{im, seg, imseg, label}} type of data used for registration. Use type=label only at "
```
jcohenadad, line 146: it's a fairly minor issue, but i was wondering if we could go from there:
~~~
                          - step: <int> Step number (starts at 1, except for type=label).
                          - type: {im, seg, imseg, label} type of data used for registration. Use
                          type=label only at step=0.
                          - algo: Default=syn
~~~
to there:
~~~
                          - step: <int> Step number (starts at 1, except for type=label).
                          - type: {im, seg, imseg, label} type of data used for registration. Use
                            type=label only at step=0.
                          - algo: Default=syn
~~~

scripts/sct_register_to_template.py
```diff
@@ -72,133 +75,210 @@ def __init__(self):
...
-                      mandatory=False,
-                      default_value=param.verbose,
-                      example=['0', '1', '2'])
+    parser = argparse.ArgumentParser(
+        description=(
+            "Register an anatomical image to the spinal cord MRI template (default: PAM50).\n"
+            "\n"
+            "The registration process includes three main registration steps:\n"
+            "  1. straightening of the image using the spinal cord segmentation (see sct_straighten_spinalcord for "
+            "details);\n"
```
jcohenadad, line 84: again a comment about indentation for list. Currently doing that:
~~~
The registration process includes three main registration steps:
  1. straightening of the image using the spinal cord segmentation (see sct_straighten_spinalcord fo
r details);
  2. vertebral alignment between the image and the template, using labels along the spine;
  3. iterative slice-wise non-linear registration (see sct_register_multimodal for details)
~~~
where this would be much nicer:
~~~
The registration process includes three main registration steps:
  1. straightening of the image using the spinal cord segmentation (see sct_straighten_spinalcord fo
     r details);
  2. vertebral alignment between the image and the template, using labels along the spine;
  3. iterative slice-wise non-linear registration (see sct_register_multimodal for details)
~~~

scripts/sct_get_centerline.py
cfhammill: Ok, I think matching FSL is a reasonable justification

scripts/sct_register_multimodal.py
cfhammill: I bet this is tricky. It would add a special case to the indentation calculator I wrote for `R|` strings. Best to save for a separate small PR.

scripts/sct_register_to_template.py
```diff
@@ -72,133 +75,210 @@ def __init__(self):
...
+            "mis-alignment of discs because a subject''s intervertebral discs distance might differ from that of the "
+            "template.\n"
+            "\n"
+            "If more than two labels (only with the parameter '-disc') are used, a non-linear registration will be "
+            "applied to align the each intervertebral disc between the subject and the template, as described in "
+            "sct_straighten_spinalcord. This the most accurate and preferred method. This feature does not work with "
+            "the parameter '-ref subject'.\n"
+            "\n"
+            "More information about label creation can be found at "
+            "https://www.slideshare.net/neuropoly/sct-course-20190121/42"
```
jcohenadad, line 123: The link is dead. I know this alternative is not great (icloud not accessible everywhere), but until we have a proper documentation on spinalcordmri.com, that's probably the best we can do (?)
```suggestion
            "https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%5FCourse%5F20200121 (p47)"
```

scripts/sct_register_to_template.py
```diff
@@ -72,133 +75,210 @@ def __init__(self):
...
+        action="help",
+        help="Show this help message and exit."
+    )
+    optional.add_argument(
+        '-l',
+        metavar=Metavar.file,
+        help=("R|One or two labels (preferred) located at the center of the spinal cord, on the mid-vertebral slice. "
+              "Example: anat_labels.nii.gz\n"
+              "For more information about label creation, please see: "
+              "https://www.slideshare.net/neuropoly/sct-course-20190121/42")
```
jcohenadad, line 157:
```suggestion
              "https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%5FCourse%5F20200121 (p47)")
```

scripts/sct_register_to_template.py
```diff
@@ -72,133 +75,210 @@ def __init__(self):
...
+    )
+    optional.add_argument(
+        '-ldisc',
+        metavar=Metavar.file,
+        help=("R|Labels located at the posterior edge of the intervertebral discs. Example: anat_labels.nii.gz\n"
+              "If you are using more than 2 labels, all disc covering the region of interest should be provided. "
+              "(E.g., if you are interested in levels C2 to C7, then you should provide disc labels 2,3,4,5,6,7.) "
+              "For more information about label creation, please refer to "
+              "https://www.slideshare.net/neuropoly/sct-course-20190121/42")  # TODO: update URL
```
jcohenadad, line 166:
```suggestion
              "https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%5FCourse%5F20200121 (p47)")  # TODO: update URL
```

scripts/sct_smooth_spinalcord.py
```diff
@@ -45,58 +47,72 @@ def update(self, param_user):
...
+        help=("Sigma (standard deviation) of the smoothing Gaussian kernel (in mm). For isotropic smoothing you only "
+              "need to specify a value (e.g. 2). For anisotropic smoothing specify a value for each axis, separated "
+              "with a comma. The order should follow axes Right-Left, Antero-Posterior, Superior-Inferior "
+              "(e.g.: 1,1,3). For no smoothing, set value to 0.")
+    )
+    optional.add_argument(
+        '-param',
+        metavar=Metavar.list,
+        type=list_type(',', str),
+        help=(f"R|Advanced parameters. Assign value with '='; Separate params with ','\n"
```
jcohenadad, line 99: since this is the only "param", maybe we should just create a flag `-algo-fitting`? I know this will break compatibility, but now is a good time to break it 😉

scripts/sct_get_centerline.py
joshuacwnewton: Good eyes!!

scripts/sct_label_utils.py
joshuacwnewton: I'm not sure I understand this suggestion. Could you clarify?

scripts/sct_register_multimodal.py
joshuacwnewton: A fix for this could also cover a fix for https://github.com/neuropoly/spinalcordtoolbox/pull/2842#issuecomment-689129919 if we use list formatting there, too, e.g.:

```
  -p <str>              Process to generate mask.
                          - <coord,XxY>: Center mask at the X,Y coordinates. (e.g. "coord,20x15")
                          - <point,FILE>: Center mask at the X,Y coordinates of the label defined in
                            input volume FILE. (e.g. "point,label.nii.gz")
                          - <center>: Center mask in the middle of the FOV (nx/2, ny/2).
                          - <centerline,FILE>: At each slice, the mask is centered at the spinal cord
                            centerline, defined by the input segmentation FILE. (e.g.
                            "centerline,t2_seg.nii.gz")
```

scripts/sct_label_utils.py
cfhammill: Just to add a space between the period and the continuation on the next line so it's `value2. Example` not `value2.Example`

scripts/sct_register_to_template.py
joshuacwnewton: Ahh, it looks like this is related to #2134. Good catch.

scripts/sct_label_utils.py
joshuacwnewton: Ah! Of course. :stuck_out_tongue_closed_eyes:

joshuacwnewton:
> @joshuacwnewton have you tried to run batch_processing with this branch? it sometimes detects glitches in the CLI (things that are poorly tested with our testing suite)

I've just tried now, but it fails due to the recently-introduced bug described by #2883. I'll take a look at that ASAP.

joshuacwnewton:
@jcohenadad just re-ran `batch_processing.sh` now that the fix has been applied. Results below:

```
set +v
~~~
Version:         git-jn/1548-msct_parser-conversion-2-a2c409b8aa94e38aa9123e1e81ced827ff5ad335*
Ran on:          Linux pop-os 5.4.0-7642-generic
Duration:        0hrs 18min 34sec
---
t2/CSA:          74.69254399438869
mt/MTR(WM):      54.42052837384227
dmri/FA(CST_r):  0.7756557354302693
dmri/FA(CST_l):  0.7801552118509679
~~~
To open Quality Control (QC) report on a web-browser, run the following:
xdg-open /home/joshua/Repos/spinalcordtoolbox/qc_example_data/index.html
```

scripts/sct_register_multimodal.py
joshuacwnewton: Since @cfhammill created a separate issue for this (#2889), I'm going to mark the identation comments as resolved.

jcohenadad:
> Is this cause for concern?

yup! it definitely is. I'm wondering where that comes from, considering that T2 CSA is independent from the MTsat module.

scripts/sct_get_centerline.py
```diff
@@ -4,135 +4,157 @@
...
-                      description='If provided, this string will be mentioned in the QC report as the dataset the process was run on',
-                      )
-    parser.add_option(name='-qc-subject',
-                      type_value='str',
-                      description='If provided, this string will be mentioned in the QC report as the subject the process was run on',
-                      )
+    parser = argparse.ArgumentParser(
+        description=(
+            "This function extracts the spinal cord centerline. Three methods are available: OptiC (automatic), "
+            "Viewer (manual), and Fitseg (applied on segmented image). These functions output (i) a NIFTI file with "
```
jcohenadad, line 23: please keep the same case as before (ie optic, fitseg, etc.), otherwise it will break cross-compatibility.

also: could you please "force" the parser to select those specific choices?

scripts/sct_process_segmentation.py
```diff
@@ -37,109 +40,138 @@ def get_parser():
...
+    )
+    optional.add_argument(
+        '-vertfile',
+        metavar=Metavar.str,
+        default='./label/template/PAM50_levels.nii.gz',
+        help="Vertebral labeling file. Only use with flag -vert"
+    )
+    optional.add_argument(
+        '-perlevel',
+        choices=['0', '1'],
```
jcohenadad, line 120: the previous implementation output the variable as an `int`:
https://github.com/neuropoly/spinalcordtoolbox/pull/2842/files#diff-8a6cdb0c20af3a3924608a180063db4eL90

Now, the variable is a str. The problem is that it is then passed in  `aggregate_per_slice_or_level()`, which expects an int (0 or 1), not a str. If str, then perlevel will be considered to be always True, causing the discrepancy observed in https://github.com/neuropoly/spinalcordtoolbox/pull/2842#issuecomment-691325349.

Conclusion:
- could you please review all the modified code that has the parameter "int" (or float) in the previous parser, to make sure that the behaviour is kept the same
- batch_processing is _very_ useful, as it helped catching that bug, otherwise difficult to catch. Tagging @Drulex here because he is currently working on adding this batch script in the CI

joshuacwnewton: Thank you for catching this. I'll start looking at this immediately.

scripts/sct_label_utils.py
```diff
@@ -681,195 +683,243 @@ def get_parser():
...
-                      mandatory=False)
-    parser.add_option(name="-ilabel",
-                      type_value="file",
-                      description="File that contain labels that you want to correct. It is possible to add new points with this option. Use with -create-viewer",
-                      mandatory=False,
-                      example="t2_labels_auto.nii.gz",)
-    parser.add_option(name='-cubic-to-point',
-                      type_value=None,
-                      description='Compute the center-of-mass for each label value.',
-                      mandatory=False)
```
joshuacwnewton, line 722: type_val=None

scripts/sct_process_segmentation.py
joshuacwnewton: Excellent catch. You nailed it, @jcohenadad. Getting the expected results after fixing the two flags in that file. 

<details>
  <summary>Results</summary>

```bash
set +v
~~~
Version:         git-jn/1548-msct_parser-conversion-2-398c7d0aea5dd2e1a17e0cc48a5d4bdc9fff45a9*
Ran on:          Linux pop-os 5.4.0-7642-generic
Duration:        0hrs 17min 21sec
---
mt/MTR(WM):      54.42026850022487
dmri/FA(CST_r):  0.7756557354302693
dmri/FA(CST_l):  0.7801552118509679
~~~

```
</details>

I checked and couldn't see this issue anywhere else.

Related: #2843 was created specifically to help prevent issues like this.

joshuacwnewton: Excellent catch, @jcohenadad. Fixing the two options in this file produces the expected results.

> The same issue applies to other flags (-perslice, etc.), and likely to other scripts.

This wasn't an issue in other scripts, because the original authors of these scripts were either casting the string to an int, e.g.

https://github.com/neuropoly/spinalcordtoolbox/blob/b0154f493365f893885f7b2098c21f0a097f6524/scripts/sct_warp_template.py#L154-L159

https://github.com/neuropoly/spinalcordtoolbox/blob/b0154f493365f893885f7b2098c21f0a097f6524/scripts/sct_warp_template.py#L206

Or, the authors would check for `"1"` specifically:

https://github.com/neuropoly/spinalcordtoolbox/blob/b0154f493365f893885f7b2098c21f0a097f6524/scripts/sct_propseg.py#L309-L316

https://github.com/neuropoly/spinalcordtoolbox/blob/b0154f493365f893885f7b2098c21f0a097f6524/scripts/sct_propseg.py#L590

These '0'/'1' options are handled inconsistently across all of our scripts (and not just the 10 in this PR), so I was planning on addressing that fully in #2843.

jcohenadad: > This is quite inconsistent, and the scope of this inconsistency goes beyond the 10 scripts in this PR, which is why I had created #2843, and was going to follow through on that after this PR was merged.

awesome! thanks a lot

jcohenadad (approved):
Thank you so much for taking on and finalizing this daunting task! 🙏

joshuacwnewton:
Thank you so much for your patience and attention to detail when reviewing it, @jcohenadad!

Issue #1526, opened 2017-11-16
Allow SCT library installation "the standard way", ie. using setuptools


zougloub:
In the rest of the world, python libraries are installed on the system using setup instructions contained in setup.py, and installation is first performed in a staging directory, where files are catalogued, then merged on the system hierarchy (/usr/lib/python2.7/..).

Today if we try to install SCT this way, we get a very partial installation:
- not much in site-packages/spinalcordtoolbox
- something in site-packages/scripts

zougloub:
Pending issues:
1. `version.txt` issue (something should be under `spinalcordtoolbox/`)
2. Launcher (which helps mostly for MPI) should be optional by default
3. frozen requirements (prevents installation along other libraries that would require newer versions)
4. requirements are split between pip and conda, the default should be all-pip with the installer doing its custom thing

kousu:
I wonder what's the status on this. `pip install -e` does most of the work in `./install_sct` now. We still need to use `sct_download_data` to get the `bin/` folder and the pre-trained models.

I am leaning towards trying to package the contents of `bin/` and the models folders with pip. Then we could version them in one place (`requirements.txt`), would be integrated in the python system, and would have access to pypi's mirrors, and we wouldn't have to maintain our own package manager. See #2669.

kousu:
Just a note on some research I stumbled into:

Right now `pip install .` does not install the package, because all the dependencies are declared in requirements.txt. A common hack (used by e.g. https://github.com/neuropoly/axondeepseg) to glue the two together ((in contravention of python's supposed "there should be one obvious way to do it") is to read the requirements.txt file in to setup.py. I would just put the requirements in setup.py directly and forget about requirements.txt. I believe we *need* to do that if we're going to be packagable.

However there is a snag. We're depending on `torch+cpu` now (#2715 -- because it's sooo much smaller) but that is **not on pypi**; instead it's over at https://download.pytorch.org/whl/cpu/torch_stable.html, which is a `pip install --find-links` URL. To make the install reproducible, we put this URL into our requirements.txt:

https://github.com/neuropoly/spinalcordtoolbox/blob/b64cad3c846fd6bd7a557688b67b80fe0b2c6dc2/requirements.txt#L26-L30

[...]

PR #2846, opened 2020-08-15
Put dependencies in setup.py

kousu:
This is needed for a buildable package. Users installing off pypi won't have access to our requirements.txt. See #1526.

There's a snag though (https://github.com/neuropoly/spinalcordtoolbox/issues/1526#issuecomment-674330466) which this addresses by keeping the `requirements.txt` around just to hold the pointer to `torch+cpu`. Eventually (if https://github.com/pytorch/pytorch/issues/26340) we should be able to do away with it entirely.

So for now, install instructions have to remain run `./install_sct`, whose meat is in `pip install -r requirements.txt`.


kousu:
A small part of #1526.

kousu:
The requirements.txt -> install.txt change is kind of radical. I *think* it's a good, clever, idea, but you should do a test release to see how it goes.

joshuacwnewton (changes requested):
I think this change might conflict with how SCT freezes requirements for releases.

install_sct
```diff
@@ -538,22 +538,13 @@ pip install numpy
 if [[ -f "requirements-freeze.txt" ]]; then
   print info "Using requirements-freeze.txt (release installation)"
   pip install -r requirements-freeze.txt
-else
-  # Not a package
-  print info "Using requirements.txt (git installation)"
-  pip install -r requirements.txt
 fi
+
+pip install -r install.txt
```
joshuacwnewton, line 543: Now that `pip install -r install.txt` is outside the if/else block, it'll get run in addition to `pip install -r requirements-freeze.txt`, which I think could undo the installations of the pinned versions.

Before, when we had two steps (`pip install -r requirements.txt` and `pip install -e $SCT_DIR`), this if/else block provided two separate installation paths, which helped prevent that.

PR #2848, opened 2020-08-16
Add git origin checking for sct_run_batch

cfhammill:
Adds a small function to `utils` for checking the url associated with the `origin` remote. Then adds this to the `sct_run_batch` output.

Also modifies the commit check to use the absolute path for both the script and data directory instead of the raw version in the `args` namespace.


joshuacwnewton:
With two approvals, I'm going to merge this. Thanks for the contribution, @cfhammill! :slightly_smiling_face:

PR #2858, opened 2020-08-22
Check qform and sforms match first

### LINKED ISSUE #2429: Wrong straightening if qform/sform differ
In [this dataset](https://www.dropbox.com/s/f32pa1iamovm6zj/20190904_data.zip?dl=0), the qform and sform differ:

~~~
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	0.033277 -0.029163 0.999021 -83.944572 
qto_xyz:2	-0.999446 -0.000971 0.033263 111.507751 
qto_xyz:3	0.000000 0.999574 0.029179 -134.296982 
qform_xorient	Anterior-to-Posterior
qform_yorient	Inferior-to-Superior
qform_zorient	Left-to-Right
sform_name	Scanner Anat
sform_code	2
sto_xyz:1	-0.022228 0.010291 0.999700 -87.573250 
sto_xyz:2	-0.972715 0.230757 -0.024004 72.559143 
sto_xyz:3	0.230935 0.972957 -0.004881 -186.981384 
~~~

Which is causing `sct_straighten_spinalcord` to output wrong straightening results:

Screen Shot 2019-09-04 at 1 44 12 PM

To reproduce, download the data and run:
~~~
sct_straighten_spinalcord -i sub-02_ses-01_echo-MEAN_T1w.nii -s sub-02_ses-01_echo-MEAN_T1w_seg.nii -v 2 -r 0
~~~

SCT version: 4.0.2

We need to figure out where it breaks (maybe the resampling?), and solve the problem or at least inform the user about a potential issue during processing.

cfhammill:
Solves issue with `sct_straighten_spinalcord` #2429. Errors in case of mismatch, pointing users to a new utility in `sct_image` to fix a mismatch.

Fixes #2429


scripts/sct_image.py
```diff
@@ -204,6 +211,13 @@ def main(args=None):
         im_out = [im_dest_new]
         fname_out = arguments.copy_header
 
+    elif arguments.set_sform_to_qform:
```
jcohenadad, line 214: should it be an elif or an isolated if? ie: the copy of the sform could potentially be done _in addition to_ other processes, couldn't it?

cfhammill: Definitely. So the idea would be to set the sform to the qform for the input image before other manipulations?

jcohenadad: hum... i would say so. I don't think it matters given that the current functions in sct_image only use the qform, but it is probably wiser to do it at the beginning. And hopefully there is no process that alters _only_ the qform 😬

cfhammill: This refactor is slightly more involved than I thought. The input image is read in multiple times throughout the the script, so unless we do an `elif` then we're going to need to alter the code in multiple places.

cfhammill: I thought of a reasonably simple fix, just replace `Image` with the sform fixer code if the user wants.

joshuacwnewton:
@cfhammill Is there anything you'd like me to do to help with this PR (e.g. give a review and/or assist with the CI error)?

jcohenadad (changes requested):
it would maybe make sense to move `check_affines_match()` inside image.py

cfhammill:
I moved the logic from `check_affines_match` into `Image.__init__`, with an optional argument `check_sform` which can be disabled to allow loading of files where the two don't match (this is needed in at least the case where we want to fix the qform and sform not matching issue).

The problem now is that some tests have begun to fail. It turns out that within the tolerance of `np.allclose` the `qform` and `sform` do not match after `change_rotation("ALS")`, this is a bit perplexing because the `change_orientation` method calls the `change_orientation` function, which in standard usage `set_qform`s and `set_sform`s to the same affine matrix. My first guess was the two were stored at different precisions, but `get_qform().dtype` and `get_sform().dtype` both return `float64`, so if it's getting cast to different precisions then it's happening internally in the nibabel header. I could potentially fix this by increasing the tolerance for `np.allclose` but I think it would be worth knowing for sure what's happening.

cfhammill:
After using the comparison approach described above I'm now getting the following 5 test failures.

```
FAILED unit_testing/test_straightening.py::test_straighten - numpy.linalg.Lin...
FAILED unit_testing/test_transfo.py::test_transfo_null - numpy.linalg.LinAlgE...
FAILED unit_testing/test_transfo.py::test_transfo_figure_out_ants_frame_exhaustive
FAILED unit_testing/test_transfo.py::test_transfo_exhaustive_wrt_orientations
FAILED unit_testing/test_transfo.py::test_transfo_skip_pix2phys - numpy.linal...
===== 5 failed, 134 passed, 12 skipped, 4382 warnings in 73.62s (0:01:13) ======
```

The failure in `test_straightening` is due to attempting to set the `qform` to the `sform`, which involves using `numpy.linalg.svd` to convert from matrix to quaternion representations. Somewhere inside test_straighten there is an image generated with sform =

```
[[0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 1.]]
```

When you attempt to `set_qform` to this matrix it fails with the linear algebra error.

cfhammill:
Adding an explicit check for this situation still leads to an error, but it is now the correct `qform` doesn't match `sform`

@jcohenadad it looks like there are ~5 places in the tests where images are created internally where the sform doesn't match the qform, and the sform is a nonsensical affine. Do you think it is worth tracking these down and fixing them? These presently stands in the way of have the sform check on every image read.

jcohenadad:
> @jcohenadad it looks like there are ~5 places in the tests where images are created internally where the sform doesn't match the qform, and the sform is a nonsensical affine. Do you think it is worth tracking these down and fixing them? These presently stands in the way of have the sform check on every image read.

If I understand correctly, the problem comes from sforms that are "irrelevant", e.g.
````
[[0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 1.]]
````
In this case, i would suggest to replace them with a more relevant sform (e.g. 1 in the diagonal). But of course it depends what the tests are intending to do. I could spend some time digging, but have you looked into this by any chance?

**Comment by cfhammill** [comment]
@jcohenadad I'm not sure what the nonsense affines are for - is it a feature of ANTs perhaps that doesn't initialize the sform? I can do a bit of digging. I suspect they are all the same error.

**Comment by cfhammill** [comment]
My proposed compromise is that we trap linear algebra errors of this kind. If the `Image` has an `absolutepath` then we report that a the file has an unset sform and may cause unexpected behaviour. If the `absolutepath` attribute hasn't been set, some of our internal code has produced a nameless file with a broken sform and we should treat it as a bug.

Does that sound reasonable? 

I think it is preferable to silently fixing the sform to match the qform, I can't confidently say that there is no good reason to have an uninitialized sform.

**Comment by jcohenadad** [comment]
Taking over this PR. I've decided to reduce its scope and only fix #2429 by checking sform/qform match in `straightening` API. In the future, we should fix all the sform/qform inconsistencies introduced by SCT code and in the unit tests, which produced the various [CI errors](https://github.com/neuropoly/spinalcordtoolbox/pull/2858/checks?check_run_id=1710120289). After fixing it, the field `check_sform` in `spinalcordtoolbox.image.Image()` can be set to `True`.

**Review by jcohenadad — APPROVED** [review]

File: spinalcordtoolbox/scripts/sct_image.py [hunk 2]
```diff
@@ -488,6 +506,14 @@ def multicomponent_merge(fname_list):
     return im_out
 
 
+def read_and_fix_sform(fname):
+    im_in = Image(fname, check_sform=False)
+    im_reaff = im_in.copy()
+    im_reaff.header = im_in.header.copy()
+    im_reaff.header.set_sform(im_reaff.header.get_qform())
+    return im_reaff
```
joshuacwnewton, line 514: To follow [API code vs application code](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-API-code-vs-application-code), it could be better to put this functionality into the API, via `spinalcordtoolbox/image.py`.

```python
# Inside spinalcordtoolbox/image.py
class Image(object):
    # [...]
    def set_sform_to_qform(self):
        """Use this (or set_qform_to_sform) when matching matrices are required."""
        self.header.set_sform(self.header.get_qform())
 
    def set_qform_to_sform(self):
        """Use this or (set_sform_to_qform) when matching matrices are required."""
        self.header.set_qform(self.header.get_sform())
```

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -302,6 +315,11 @@ def main(argv=None):
             spaces.append(None)
         im_out = [ displacement_to_abs_fsl(Image(fname_in[0]), spaces[0], spaces[1]) ]
 
+    elif arguments.set_sform_to_qform is not None:
+        # This argument can be used stand alone or with other operations so
+        # it should go last before the else
+        im_out = [Image(fname_in[0])]
```
joshuacwnewton, line 321: Then, we could call the API method from `sct_image`.

```suggestion
    elif arguments.set_sform_to_qform is not None:
        # This argument can be used stand alone or with other operations so
        # it should go last before the else
        im_out = [Image(fname_in[0]).set_sform_to_qform()]
```

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -175,6 +182,12 @@ def main(argv=None):
     fname_in = arguments.i
     n_in = len(fname_in)
 
+    # If the user wants to fix sforms replace `Image` with the sform fixer
+    if arguments.set_sform_to_qform:
+        Image = read_and_fix_sform
+    else:
+        Image = spinalcordtoolbox.image.Image
```
joshuacwnewton, line 189: That way we wouldn't need this...

spinalcordtoolbox/image.py
```diff
@@ -259,12 +292,21 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
...
-        #     self.hdr.set_qform(self.hdr.get_qform(), code=0)
-        #     self.header.set_qform(self.hdr.get_qform(), code=0)
+        # Make sure sform and qform are the same.
+        # Context: https://github.com/neuropoly/spinalcordtoolbox/issues/2429
+        if check_sform and not check_affines_match(self):
+            if self.absolutepath is None:
+                logger.error("Internal code has produced an image with inconsistent qform and sform "
+                             "please report this on github at https://github.com/neuropoly/spinalcordtoolbox/issues "
+                             " or on the SCT forum https://forum.spinalcordmri.org/.")
+            else:
```
joshuacwnewton, line 302: The logging error here looks possibly redundant with the `try`/`except` block inside `check_affines_match`. (See below)

spinalcordtoolbox/image.py
```diff
@@ -220,12 +221,44 @@ def __getitem__(self, idx):
...
+        hdr2.set_qform(hdr.get_sform())
+    except np.linalg.LinAlgError:
+        if im.absolutepath is None:
+            logger.error("Internal code has produced an image with inconsistent qform and sform "
+                         "please report this on github at https://github.com/neuropoly/spinalcordtoolbox/issues "
+                         "or on the SCT forums https://forum.spinalcordmri.org/.")
+
+        logger.warning("The sform for {} is uninitialized and may cause unexpected behaviour."
+                       ''.format(im.absolutepath))
+        return(True)
```
joshuacwnewton, line 238: Redundant section of code -- are both logger errors necessary?

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -93,6 +94,12 @@ def get_parser():
         help='Copy the header of the source image (specified in -i) to the destination image (specified here) '
              'and save it into a new image (specified in -o)',
         required = False)
+    header.add_argument(
+        '-set-sform-to-qform',
+        help='Set the sform matrix of all input images to their respective qform matrix. '
+             'This operation is applied anytime an image is read in by `sct_image`.',
```
joshuacwnewton, line 100: This description says "of all input images", but is that accurate? 

I notice we only apply `sct_image -set-sform-to-qform` to only the first image, `fname_in[0]`:

https://github.com/neuropoly/spinalcordtoolbox/blob/ff54a929cfe447d40958364ee1a35c77b04189e4/spinalcordtoolbox/scripts/sct_image.py#L318-L321

joshuacwnewton (changes requested):
Thanks for taking this on, @jcohenadad!

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: Ah, I understand now why this was done. It was even explicitly mentioned in https://github.com/neuropoly/spinalcordtoolbox/pull/2858#discussion_r485985597.

I'm going to write a `TODO` to limit the scope of this PR.

joshuacwnewton: While patching `Image` like this does do a good job at limiting the scope of the PR, I think the benefit isn't worth leaving the script like this.

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: Ah, I see now. Only `-concat` uses multiple images. And, in that case, setting sform to qform doesn't make sense?

I will rewrite some of the help descriptions to make single/multi image behavior more clear.

joshuacwnewton: Ah, I see now. Most functions use only the first image, but some use a list of images. 

This is... weird, and not communicated well to the user. I think that work could be done here, but it's outside the scope of this PR, so I'll just make some notes about it.

joshuacwnewton: Done in db03173.

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: Done in f9c6c08.

spinalcordtoolbox/image.py
joshuacwnewton: Ah I see now. It's not redundant -- it was just copied and pasted but not edited.

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -174,6 +184,12 @@ def main(argv=None):
 
     fname_in = arguments.i
     n_in = len(fname_in)
+    # TODO: The functions for '-concat', '-omc', and '-display-warp' take in filenames, so they ignore 'im_in'.
+    #  Instead, we should harmonize this functionality so that 'im_in' is used everywhere.
+    im_in = Image(fname_in[0])
```
joshuacwnewton, line 189: While finishing up this PR, I made this observation.

This is a relatively simple fix, but I don't want the scope of this PR to grow any further, so I'll follow up in a separate PR.

spinalcordtoolbox/scripts/sct_image.py
jcohenadad: > This is... weird, and not communicated well to the user. I think that work could be done here, but it might be outside the scope of this PR, so I'll just make some notes about it.

100% agreed! i'm glad you caught (and fixed) it!

jcohenadad (approved):

joshuacwnewton (approved):
This is another one of those pull requests that I have to self-approve because I had requested changes previously. :sweat_smile:

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: I made an issue for this here: https://github.com/neuropoly/spinalcordtoolbox/issues/3205

PR #2875, opened 2020-09-03
Implement json sidecar parser to fetch TR and flip angles for MTsat calculation

### LINKED ISSUE #1673: Implement test for sct_compute_mtsat
### Prerequisites

* [x] Did you check past issues for similar issue(s) ?
* [x] If you are using a development environment, is your environment updated ?

### Description

Implement test for `sct_compute_mtsat`. A new testing dataset that includes MTw, PDw and T1w will be required.

### LINKED ISSUE #2874: Implement json sidecar parser to fetch TR and flip angles for MTsat calculation
Some dataset yields corrupted MTsat maps. There might be something wrong with the numerical computation (e.g. rounding? overflow? …).

To reproduce:
- SCT version: git-master-a3672108090a213c93a9bc5a46a2c0a51fac29b6
- Data: [issue_mtsat.zip](https://github.com/neuropoly/spinalcordtoolbox/files/5171079/issue_mtsat.zip)
- syntax:
~~~
sct_compute_mtsat -mt sub-sherbrooke03_acq-MTon_MTS_reg.nii.gz -pd sub-sherbrooke03_acq-MToff_MTS_reg.nii.gz -t1 sub-sherbrooke03_acq-T1w_MTS_crop.nii.gz -trmt 00645023 -trpd 00645023 -trt1 0015 -famt 9 -fapd 9 -fat1 15
~~~

Full context with example of corrupted map: https://github.com/spine-generic/data-multi-subject/issues/44

jcohenadad:
Instead of asking the user to input TRs and FAs into the function `sct_compute_mtsat`, a more elegant solution would be to read the json sidecar of the nifti file, if it exists.

Fixes #2874, Fixes #1673


unit_testing/test_sct_compute_mtsat.py
```diff
@@ -0,0 +1,49 @@
...
+    assert sct_compute_mtsat.get_tr_and_flipangle(InputParam1()) == (0.03, 0.03, 0.015, 9, 9, 15)
+    assert sct_compute_mtsat.get_tr_and_flipangle(InputParam2()) == (1, 2, 3, 4, 5, 6)
+
+
+def test_with_json_sidecar():
+    sct_compute_mtsat.main(['-mt', sct_test_path('mt', 'mt1.nii.gz'),
+                            '-pd', sct_test_path('mt', 'mt0.nii.gz'),
+                            '-t1', sct_test_path('mt', 't1w.nii.gz')])
+    # Check if output file exists
+    assert os.path.isfile(sct_test_path('mt', 'mtsat.nii.gz'))
```
joshuacwnewton, line 49: Would this be better as an integrity test in `testing/`, considering that's where the other SCT script tests are located?

scripts/sct_download_data.py
```diff
@@ -70,9 +70,8 @@ def main(args=None):
             "https://www.neuro.polymtl.ca/_media/downloads/sct/20180525_sct_example_data.zip",
         ],
         "sct_testing_data": [
-            "https://github.com/sct-data/sct_testing_data/releases/download/r20200707232231/sct_testing_data-r20200707232231.zip",
-            "https://osf.io/download/5f053c176cbca300dad363d3/"
-        ],
+            "https://github.com/sct-data/sct_testing_data/releases/download/r20200504-13-g713fcb6/sct_testing_data-r20200504-13-g713fcb6.zip",
+            "https://osf.io/download/5f516b634f1e5e00226f0599/"],
```
joshuacwnewton, line 74: I'm a bit confused about the naming convention here. Why have we gone from July 7th to May 4th ([`r20200504`](https://github.com/sct-data/sct_testing_data/releases/tag/r20200504-13-g713fcb6)) in this newest dataset release?

scripts/sct_compute_mtsat.py
```diff
@@ -132,9 +127,49 @@ def get_parser():
...
+        fat1 = fetch_metadata(fname_json, 'FlipAngle')
+    else:
+        fat1 = args.fat1
+    return trmt, trpd, trt1, famt, fapd, fat1
+
+
+def main(argv):
     parser = get_parser()
-    args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
+    args = parser.parse_args(argv if argv else ['--help'])
```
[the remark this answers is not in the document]
joshuacwnewton: Although, perhaps making `main(args)` consistent across all scripts could be handled in another PR, considering this is inconsistent in many SCT scripts.

joshuacwnewton (commented):
The core logic looks good, I think. I just had a few questions regarding coding conventions compared to other scripts. (Code organization, argparse, etc.)

scripts/sct_compute_mtsat.py
```diff
@@ -132,9 +127,49 @@ def get_parser():
...
 
 
-def main():
+def get_tr_and_flipangle(args):
+    """
+    Retrieve metadata from the json sidecar or from the input arguments.
+    :param args:
+    :return: trmt, trpd, trt1, famt, fapd, fat1
+    """
+    # TODO: deal with the code duplication below
```
Drulex, line 136: To simplify this code you can leverage the `default` param of `add_argument()` and avoid checking the optional arguments.

For example:
```
mandatoryArguments.add_argument(
        "-trmt",
        help="TR [in ms] for mt image. By default, will be fetch from the json sidecar (if it exists).",
        type=float,
        default=fetch_metadata(get_json_file_name(args.mt, check_exist=True), 'RepetitionTime')
        metavar=Metavar.float,
        )
```

Since all args checked in `get_tr_and_flipangle()` are optional, you can do this and get rid of the `get_tr_and_flipangle()` function all together.

unit_testing/test_sct_compute_mtsat.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+
+from __future__ import print_function, absolute_import
```
Drulex, line 3: Needed?

unit_testing/test_sct_compute_mtsat.py
```diff
@@ -0,0 +1,49 @@
...
+
+import os
+import sys
+
+from spinalcordtoolbox.utils import sct_test_path
+from spinalcordtoolbox import __sct_dir__
+sys.path.append(os.path.join(__sct_dir__, 'scripts'))
```
Drulex, line 10: Is this needed? If these tests are meant to be run by `pytest` only I don't think this is necessary.

scripts/sct_compute_mtsat.py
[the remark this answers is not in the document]
jcohenadad: > I see you're passing argv to main so that test_sct_compute_mtsat.py can call main with arguments.

yes, exactly!

> Did you have a preferred method in mind for this? I vaguely remember seeing one method endorsed in the Wiki, but I can't find it after a quick search.

I defer to the judgement and expertise for the software engineers in the SCT dev team 😉 
there is this [recent wiki page](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Logging#how-logging-appears-to-users-of-sct-scripts), which does call main(arguments), but what happens _inside_  the main is not documented there (it should).

> Although, perhaps making main(args) consistent across all scripts could be handled in another PR, considering this is inconsistent in many SCT scripts.

sounds good!

spinalcordtoolbox/bids.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+# Deals with Brain Imaging Data Structure (BIDS)
```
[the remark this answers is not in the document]
jcohenadad: i thought about it... i'm afraid `utils.py` will quickly become a dump for many functions, and in this case, `bids` is a fairly specific use-case for everything related to BIDS. Also: we could anticipate adding more bids-related functions in the future. Happy to discuss further.

unit_testing/test_sct_compute_mtsat.py
jcohenadad: indeed, but i think we decided (I can't find the discussion thread-- maybe it was in SCT meeting?) to slowly get rid of the `testing/` folder, and instead rely on the `unit_testing` for the long term, as it is more aligned with Python conventions. Integrity testing could also be included in unit_testing, with prefix test_sct_*** (instead of test_module***).

scripts/sct_download_data.py
jcohenadad: wow! good catch, i didn't notice the date. however, i was not able to specify the folder name and i had no time to dig. i opened an issue [here](https://github.com/neuropoly/distriploy/issues/9); if it is fixed quickly (ie within the next hours), we can re-create a tag with proper naming before merging this PR.

scripts/sct_compute_mtsat.py
jcohenadad: that sounds like a genius idea indeed, but i'm having an issue implementing it because `args.mt` is not known yet. Before i spend too much time finding a solution, maybe you know what to do?

unit_testing/test_sct_compute_mtsat.py
jcohenadad: no, fixed in e496a39d09d17fd9dd8816fd62e5cd85d4199fa6

unit_testing/test_sct_compute_mtsat.py
jcohenadad: indeed, i thought it was needed for the `import sct_compute_mtsat` below, but i removed these lines and it worked fine.  Fixed in 312a6bdecefd88a0901715450f2da23672a08ad4

scripts/sct_compute_mtsat.py
Drulex: Maybe if the order of the arguments is changed so that `args.mt` is before other arguments that try to use it? If it makes sense for the usage interface of course.

jcohenadad: it is currently placed before, but within the `get_parser()` function, which does not know about the `args` variable (this variable is generated by `parser.parse_args` inside the main()

Drulex: Ok we can optimize later, not critical.

jcohenadad: i think i found a workaround...

jcohenadad: here: d953ad37eb7470d43d81f123e134729902c5724e
the trick was to use `parse_known_args()` (ignore arguments that are not defined in the parser yet)

Drulex: Good to know!

jcohenadad: trickier than that... now the unit test is failing, because it cannot get the arguments passed to main(arguments):
~~~
usage: sct_compute_mtsat -mt <file> -pd <file> -t1 <file>
sct_compute_mtsat: error: the following arguments are required: -mt, -pd, -t1
  File "/Applications/PyCharm CE.app/Contents/helpers/pydev/_pydevd_bundle/pydevd_comm.py", line 1466, in do_it
    result = pydevd_console_integration.console_exec(self.thread_id, self.frame_id, self.expression, dbg)
  File "/Applications/PyCharm CE.app/Contents/helpers/pydev/_pydevd_bundle/pydevd_console_integration.py", line 203, in console_exec
    Exec(code, updated_globals, frame.f_locals)
  File "/Applications/PyCharm CE.app/Contents/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1766, in parse_known_args
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 2001, in _parse_known_args
    ', '.join(required_actions))
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 2393, in error
    self.exit(2, _('%(prog)s: error: %(message)s\n') % args)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 2380, in exit
    _sys.exit(status)
SystemExit: 2
~~~
investigating...

jcohenadad: fixed in 19f18a69ed32f79d3dcdda2931c58af91d61a7cf

spinalcordtoolbox/bids.py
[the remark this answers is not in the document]
Drulex: I'm going to side with @joshuacwnewton on this one. Both of these functions are generic and should instead live in the `sct_compute_mstat` script. In general anything dealing with user input (args parsing, file paths, simple format conversions, input sanitization) should live at a higher level. Only core logic (processing, analysis, pure business logic) should live in the API.

scripts/sct_download_data.py
jcohenadad: ok, i manually edited the zip file on GH and OSF and updated the URL: 4864a95ecdcdcbdb753dcf73216205f21f7e9c92

spinalcordtoolbox/bids.py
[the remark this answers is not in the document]
jcohenadad: ok, i will move these functions under `sct_compute_mtsat`. In the future, if we need the same functionality in other scripts, we can rediscuss it i guess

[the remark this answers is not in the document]
jcohenadad: Here: fe11966046ba278fb6813e00be0dbf090546553d

Drulex (approved):

spinalcordtoolbox/bids.py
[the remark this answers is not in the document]
Drulex: Each script is an application built on top of the API. If there is a need to share functionality between scripts it's totally ok to the `sct_utils.py`, while respecting the following:
- Should not implement core/business logic
- Should only be used for convenience, high-level/generic stuff close to the user (files, extensions, conversion, sanitization)
- Should not be called via subprocess or `sct_utils.main`, instead it should be imported cleanly

I am currently updating the wiki to centralize this info. Will make a PR soon.

unit_testing/test_sct_compute_mtsat.py
joshuacwnewton: Ahh, thank you for the clarification!

> i think we decided (I can't find the discussion thread-- maybe it was in SCT meeting?) 

In the future, when a decision like this is made during a meeting, perhaps it could be an action item for someone to update the GitHub Wiki? @Drulex has already taken some great initiative in doing something similar (https://github.com/neuropoly/spinalcordtoolbox/pull/2875#discussion_r483724457) and it would be lovely if we could make this a habit. There's a chance there are other nuggets of information regarding SCT's design that are hidden away in issues and meetings, but that could also be on the Wiki. :stuck_out_tongue:

Issue #2873, opened 2020-09-03
Refactoring of sct_register_to_template():


Drulex:
- move core logic to api `sct.registration`
- remove syscalls to other scripts and replace with proper API calls

PR #2877, opened 2020-09-04
sct_maths refactor to spinalcordtoolbox.math

### LINKED ISSUE #2876: sct_maths script refactor
Motivation:
- Some helper scripts call this script via subprocess call to run a simple mathematical operation. This introduces unnecessary overhead and is not efficient.
- Some API modules use functionality from this script -> there is a use case to move the logic at the API level and avoid the layering violation.

Drulex:
Closes #2876
Needed for #2873

This PR does not include refactoring of callers using subprocess calls or `sct_maths.main()`, compatibility is maintained but we will want to refactor those to use `spinalcordtoobox.math` at some point (tracked via #2878)


scripts/sct_maths.py
```diff
@@ -505,254 +459,67 @@ def get_data(list_fname):
...
+            f.write(metric_full + ': \n' + str(res))
+    elif ext_out == '.pklz':
+        pickle.dump(res, gzip.open(fname_out, 'wb'), protocol=2)
+    else:
+        pickle.dump(res, open(fname_out, 'w'), protocol=2)
 
+def compute_similarity_qc_func(data1_1d, data2_1d, res, metric):
+    import matplotlib
+    matplotlib.use('Agg')
+    import matplotlib.pyplot as plt
```
joshuacwnewton, line 509: Nit, but could these imports be done at the top of the script?

joshuacwnewton (changes requested):
+1 on these efforts! Feel free to ping me on future refactors. I'd love to help out with reviews. :slightly_smiling_face:

scripts/sct_maths.py
```diff
@@ -505,254 +459,67 @@ def get_data(list_fname):
...
-    block_radius = block_radius_max if block_radius > block_radius_max else block_radius
-
-    sigma = estimate_sigma(data_in)
-    denoised = nlmeans(data_in, sigma, patch_radius=patch_radius, block_radius=block_radius)
-
-    return denoised
+    if verbose > 1:
+        res = math.compute_similarity(img1.data, img2.data, metric=metric, qc_func=compute_similarity_qc_func)
+    else:
+        res = math.compute_similarity(img1.data, img2.data, metric=metric)
```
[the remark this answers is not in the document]
Drulex: You're right I should have though about it a few seconds more.

scripts/sct_maths.py
```diff
@@ -349,13 +342,13 @@ def main(args=None):
...
         # smooth data
-        data_out = smooth(data, sigmas)
+        data_out = math.smooth(data, sigmas)
 
     elif arguments.dilate is not None:
-        data_out = sct.math.dilate(data, size=arguments.dilate, shape=arguments.shape, dim=arguments.dim)
+        data_out = math.dilate(data, size=arguments.dilate, shape=arguments.shape, dim=arguments.dim)
 
     elif arguments.erode is not None:
-        data_out = sct.math.erode(data, size=arguments.erode, shape=arguments.shape, dim=arguments.dim)
```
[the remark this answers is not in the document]
Drulex: Good catch

scripts/sct_maths.py
Drulex: > +1 on these efforts! Feel free to ping me on future refactors. I'd love to help out with reviews. slightly_smiling_face

Thanks for the excellent feedback. I will involve you in all refactor PRs

scripts/msct_gmseg_utils.py
```diff
@@ -151,7 +151,7 @@ def pre_processing(fname_target, fname_sc_seg, fname_level=None, fname_manual_gm
         cropper.get_bbox_from_mask(Image(fname_mask))
         im_sc_seg_rpi_crop = cropper.crop()
         # denoising
-        from sct_maths import denoise_nlmeans
+        from spinalcordtoolbox.math import denoise_nlmeans
```
jcohenadad, line 154: out of scope, but i'm just realizing that this function is probably not used anymore (was used by sct_segment_graymatter, which is now deprecated for sct_deepseg_gm).

scripts/sct_maths.py
```diff
@@ -14,20 +14,21 @@
...
+import pickle
+import gzip
+import matplotlib
+import matplotlib.pyplot as plt
 
-import spinalcordtoolbox as sct
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
-import spinalcordtoolbox.math
+import spinalcordtoolbox.math as math
```
jcohenadad, line 27: the reason i was specifically calling `sct.math` is to avoid confusion with [Python's math libary](https://docs.python.org/3/library/math.html). I was not sure what was good practice here…?

scripts/sct_maths.py
```diff
@@ -526,233 +467,52 @@ def get_data_or_scalar(argument, data_in):
...
+            new_type_list[inew_type_list] = int(ele)
+        elif type is "float":
+            new_type_list[inew_type_list] = float(ele)
 
-    return denoised
+    return new_type_list
 
 
-def smooth(data, sigmas):
+def compute_similarity(img1: Image, img2: Image, fname_out: str, metric: str, metric_full: str, verbose):
```
jcohenadad, line 488: that's cool, i didn't know about the type checking in input args. I'm wondering if this could be leveraged for sphinx doc (ie instead of specifying the type inside the docstrings)

Drulex: The're typehints, unfortunatlely they don't provide type checking, but they are useful to have. They can definitely be used by sphinx (see https://pypi.org/project/sphinx-autodoc-typehints/), I wanted to add them everywhere when updating the doc, but it seemed like a massive effort.

scripts/sct_maths.py
Drulex: Good point. We could import as `sct_math` to avoid potential clash.

jcohenadad: out of curiosity, why would `sct_math` be preferred over `sct.math`? (same number of chars)

Drulex: Do you mean `import spinalcordtoolbox.math as sct.math`? That is invalid syntax. 

It is a good practice to import the narrowest scope we need, therefore avoid importing the whole `spinalcordtoolbox` namespace.

jcohenadad: i actually meant 
~~~
import spinalcordtoolbox.math

…
# when used:
sct.math
~~~

but our 2nd sentence (narrow scope) answers it

joshuacwnewton: > Good point. We could import as `sct_math` to avoid potential clash.

+1 to this suggestion. Even if we don't use Python's `math` module, I think it would still help with readability, so that it's clearer to someone skimming through the code that the functions come from SCT's `math.py`.

> It's even better to define a `__all__` in `math.py`. 

Could you clarify how `__all__` is relevant here? My understanding was that it's only for when `from <module> import *` is used, but we're not doing that sort of importing here, I don't think.

Drulex: > Could you clarify how __all__ is relevant here?

On the topic of narrowing the scope of imported variables. Using `__all__` allows to further narrow down what is imported.

> My understanding was that it's only for when from <module> import * is used, but we're not doing that sort of importing here, I don't think.

Yes you are right. I am proposing to add the `__all__` in API modules, which makes it cleaner for API consumers who use `from spinalcordtoolbox.math import *`. In this specific case it would have no effect, but it could be used by someones else who wants to build on top of the `math` API.

For the case of `math` it's not extremely relevant, because I only see 2 or 3 private variables in the module, but some other modules have a lot more. It does come with the cost of maintaining the `__all__` list.

I don't have a very strong opinion on this, but I think it would be a good practice to adopt.

joshuacwnewton: Ah, thank you, I understand better now.

Because this idea could apply to other modules too (and not just `sct.math`), would this discussion be better for a separate issue?

scripts/msct_gmseg_utils.py
Drulex: Is it useful to keep in the `math` API or should I remove it?

jcohenadad: i think it could be useful

scripts/sct_maths.py
Drulex: > would this discussion be better for a separate issue?

Yes. I'm thinking of proposing changes in https://github.com/neuropoly/spinalcordtoolbox/blob/master/CONTRIBUTING.rst  and opening a PR to discuss/collaborate (wiki doesn't support PR which makes it not practical).

Drulex:
Can you give it a second look when you get a change @joshuacwnewton ?

joshuacwnewton (approved):
LGTM!

scripts/sct_maths.py
joshuacwnewton: Ah, sorry, I missed this question until just now. Hm... I wonder if this is something we could discuss on Slack? e.g. proposing a certain style or convention in #sct_dev, and if there's agreement, then just adding it directly to the Wiki.

joshuacwnewton: When it comes to simpler things like this (`__all__` seems simple to me, at least), I imagine we don't need a big process for it, just some informal chats between me/you/Chris/Nick/Julien instead.

Drulex:
@jcohenadad ok to merge on your side?

scripts/sct_maths.py
Drulex: +1 for the chat

scripts/sct_maths.py
```diff
@@ -375,19 +370,19 @@ def main(args=None):
         # input 1 = from flag -i --> im
         # input 2 = from flag -mi
         im_2 = Image(arguments.mi)
-        compute_similarity(im.data, im_2.data, fname_out, metric='mi', verbose=verbose)
+        compute_similarity(im, im_2, fname_out, metric='mi', metric_full='Mutual information', verbose=verbose)
```
jcohenadad, line 373: why changing the object from np.array to Image if not changing the current input?
https://github.com/neuropoly/spinalcordtoolbox/blob/149afde200700707324fca68b1d0d248d5ff0afe/spinalcordtoolbox/math.py#L190

jcohenadad: ah sorry-- i'm just seeing that there is another `compute_similarity` defined here-- nevermind.

jcohenadad (approved):
I haven't tested by it looks good to me

scripts/sct_maths.py
Drulex: Yes the API was doing too much, only kept the computation in the API and moved the rest in the script.

PR #2882, opened 2020-09-05
Change default value for average to maintain cross-compatibility

### LINKED ISSUE #2881: Broken compatibility due to previous bug in a script
Previously, `sct_dmri_separate_b0_and_dwi` had a bug:

The default value was supposedly average=0:
https://github.com/neuropoly/spinalcordtoolbox/blob/2d29de6918f7b85bda96e8c2b6d58f720442ac77/scripts/sct_dmri_separate_b0_and_dwi.py#L34

However the input flag was converting to a string:
https://github.com/neuropoly/spinalcordtoolbox/blob/2d29de6918f7b85bda96e8c2b6d58f720442ac77/scripts/sct_dmri_separate_b0_and_dwi.py#L63

And the variable was then evaluated as a Bool, so the result was always "True":
https://github.com/neuropoly/spinalcordtoolbox/blob/2d29de6918f7b85bda96e8c2b6d58f720442ac77/scripts/sct_dmri_separate_b0_and_dwi.py#L212

The recent PR #2819 fixed that bug, but also changed the behaviour: the average scan is not more produced, which breaks all the current processing pipelines already in place, e.g. 
https://github.com/spine-generic/spine-generic/blob/4143cbfd7a206f06e3fd9f42485d5087129899d7/process_data.sh#L261

I suggest we change the default to be average=1 to ensure cross-compatibility

jcohenadad:
The recent PR #2819 fixed that bug, but also changed the behaviour: the average scan is not more produced, which breaks all the current processing pipelines already in place, e.g.
https://github.com/spine-generic/spine-generic/blob/4143cbfd7a206f06e3fd9f42485d5087129899d7/process_data.sh#L261

This PR changes the default to be average=1 to ensure cross-compatibility.

Fixes #2881

### DIFF
--- a/scripts/sct_dmri_separate_b0_and_dwi.py
+++ b/scripts/sct_dmri_separate_b0_and_dwi.py
@@ -33,7 +33,7 @@ import argparse
 class Param:
     def __init__(self):
         self.debug = 0
-        self.average = 0
+        self.average = 1
         self.remove_temp_files = 1
         self.verbose = 1
         self.bval_min = 100  # in case user does not have min bvalues at 0, set threshold.


joshuacwnewton (approved):
Good catch!

PR #2886, opened 2020-09-07
Automatically show defaults in argparse help descriptions

### LINKED ISSUE #2756: Parser: automatically display the default value in the help
**Justification:**
We are currently adding manually the default value for each argument in the "description" entry, [see example](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_straighten_spinalcord.py#L111).

To automatically display this value in the help would reduce the maintenance load and reduce the developer errors.
Also, it would fill the gaps where default values are not displayed in the help, [see example](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_crop_image.py#L83).

[ArgumentDefaultHelpFormatte](https://docs.python.org/3/library/argparse.html#formatter-class)r could be a good solution. As recommended [here](https://stackoverflow.com/a/12151325).

joshuacwnewton:
### Related issues/PRs

Fixes #2756.

### Description

This PR modifies SCT's `SmartFormatter` so that default values are shown if they are not `None`. 

Two questions:

* Are there any cases where we **want** to show `(default: None)`? (See https://github.com/neuropoly/spinalcordtoolbox/issues/2756#issuecomment-650613090.)
* Should this PR go through all of the descriptions to edit redundant info out? If so, this PR is dependent on #2842. An example:
```
  -ofolder <folder>  Output folder (e.g. "./") (default: ./)
```

### DIFF
--- a/spinalcordtoolbox/utils.py
+++ b/spinalcordtoolbox/utils.py
@@ -79,7 +79,7 @@ class Metavar(Enum):
         return self.value
 
 
-class SmartFormatter(argparse.HelpFormatter):
+class SmartFormatter(argparse.ArgumentDefaultsHelpFormatter):
     """
     Custom formatter that inherits from HelpFormatter, which adjusts the default width to the current Terminal size,
     and that gives the possibility to bypass argparse's default formatting by adding "R|" at the beginning of the text.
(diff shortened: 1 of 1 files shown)


jcohenadad:
> Are there any cases where we want to show (default: None)? (See #2756 (comment).)

If not too cumbersome, I would not show the "(default: )" string in this case. When testing this PR, i get this (with `sct_dmri_moco`):
~~~
  -bval <file>          Bvals file. Example: bvals.nii.gz (default: )
~~~

joshuacwnewton:
> If not too cumbersome, I would not show the "(default: )" string in this case.

Ah, good point! You're right, some arguments use an empty string `''` as the default instead of `None`.

jcohenadad (approved):
super useful 🥇

joshuacwnewton:
I've created #2892 for the redundant info in the description, as there could be some larger discussion about how to handle this "Example:" text.

Otherwise, merging now. :smile:

PR #2887, opened 2020-09-08
Refactor sct_label_utils into proper API

### LINKED ISSUE #2870: Refactor sct_label_utils core logic in API
Seems like this script contains lots of useful/re-use-able core logic that could be moved to its own API module.

```
$ grep 'sct_label_utils' scripts/*.py 
scripts/sct_apply_transfo.py:            sct.run(['sct_label_utils',
scripts/sct_create_mask.py:        status, output = sct.run(['sct_label_utils', '-i', 'point_RPI.nii', '-display'], verbose=param.verbose)
scripts/sct_create_mask.py:    cmd = ['sct_label_utils', '-i', 'line.nii', '-o', 'line.nii', '-create-add']
scripts/sct_label_utils.py:                    path_qc=os.path.abspath(path_qc), dataset=qc_dataset, subject=qc_subject, process='sct_label_utils')
scripts/sct_label_vertebrae.py:from sct_label_utils import ProcessLabels
scripts/sct_label_vertebrae.py:                      description='Initialize vertebral labeling by providing a nifti file that has a single disc label. An example of such file is a single voxel with value "3", which would be located at the posterior tip of C2-C3 disc. Such label file can be created using: sct_label_utils -i IMAGE_REF -create-viewer 3 ; or by using the Python module "detect_c2c3" implemented in "spinalcordtoolbox/vertebrae/detect_c2c3.py".',
scripts/sct_label_vertebrae.py:                sct.printv('Automatic C2-C3 detection failed. Please provide manual label with sct_label_utils', 1, 'error')
scripts/sct_qc.py:                                 'sct_detect_pmj', 'sct_label_utils', 'sct_get_centerline'),
scripts/sct_register_to_template.py:import sct_label_utils
scripts/sct_register_to_template.py:        sct_label_utils.main(args=['-i', ftmp_template_label_, '-vert-body', '0', '-o', ftmp_template_label])
scripts/sct_register_to_template.py:            sct.run(['sct_label_utils', '-i', ftmp_template_label, '-o', ftmp_template_label, '-remove-reference', ftmp_label])
scripts/sct_register_to_template.py:        sct.run(['sct_label_utils', '-i', ftmp_template_label, '-o', ftmp_template_label, '-remove-reference', ftmp_label])
scripts/sct_register_to_template.py:    processor = sct_label_utils.ProcessLabels(fname_labels)
scripts/sct_register_to_template.py:    sct.run(['sct_label_utils', '-i', fname_dest, '-create', label_new_list, '-v', '1', '-o', fname_output])
scripts/sct_testing.py:        'sct_label_utils',
```

Motivated by current refactor of `sct_register_to_template()` which uses it a lot. Also subprocess calls are useless (in this case) and add overhead + violate layering principles.

### LINKED ISSUE #2905: Argument error in sct_label_utils
### Description of the issue
The script `sct_label_utils` fails when using it. As shown below, the script raises an Argument error on one of the parameter of the function, even if it is not used.

### Steps to Reproduce

0. Install SCT <release or git revision and platform, you can use the
   first lines of output of sct_check_dependencies>

1. On any image, try the following command:
```
sct_label_utils -i input_image -o output_image -create-viewer 1,2,3
```

A GUI opens to perform manual labeling.

An Argument error on `remove_symm` occurs.
```
--
Spinal Cord Toolbox (git-master-432bbc681d8f7fdb27d007f0f782eaf0285582be)

  File "/Users/benjamindeleener/code/spinalcordtoolbox/scripts/sct_label_utils.py", line 945, in <module>
    main()
  File "/Users/benjamindeleener/code/spinalcordtoolbox/scripts/sct_label_utils.py", line 897, in main
    elif arguments.remove_symm is not None:
AttributeError: 'Namespace' object has no attribute 'remove_symm'
```

Easy fix: it's a spelling mistake between the parser and the main function.

Drulex:
Closes #2870 
Closes #2905 
Needed by #2873 
After #2842 
After #2903 

batch_processing results
```
Version:         git-aj/2870-refactor_sct_label_utils-144f6950224434a98a474f719c63950064f55ec4*
Ran on:          Linux drupad-p1 5.4.64-1-MANJARO                                                                                                              
Duration:        0hrs 13min 53sec                                 
---
mt/MTR(WM):      54.420871701809666
```

### DIFF
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -92,10 +92,10 @@ def get_parser():
         default='')
     optional.add_argument(
         "-x",
-        help=""" Interpolation method. The 'label' method is to be used if you would like to apply a transformation 
-        on a file that has single-voxel labels (classical interpolation methods won't work, as resampled labels might 
-        disappear or their values be altered). The function will dilate each label, apply the transformation using 
-        nearest neighbour interpolation, and then take the center-of-mass of each "blob" and output a single voxel per 
+        help=""" Interpolation method. The 'label' method is to be used if you would like to apply a transformation
+        on a file that has single-voxel labels (classical interpolation methods won't work, as resampled labels might
+        disappear or their values be altered). The function will dilate each label, apply the transformation using
+        nearest neighbour interpolation, and then take the center-of-mass of each "blob" and output a single voxel per
         blob.""",
         required=False,
         default='spline',
@@ -299,8 +299,7 @@ class Transform:
 
         if islabel:
             sct.printv("\nTake the center of mass of each registered dilated labels...")
-            out = cubic_to_point(im_src_reg)
-            out.save(path=fname_out)
+            cubic_to_point(im_src_reg).save(path=fname_out)
             if remove_temp_files:
                 sct.printv('\nRemove temporary files...', verbose)
                 sct.rmtree(path_tmp, verbose=verbose)
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -23,6 +23,7 @@ from spinalcordtoolbox.utils import Metavar, SmartFormatter
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.cropping import ImageCropper
 from spinalcordtoolbox.math import dilate
+from spinalcordtoolbox.labels import cubic_to_point
 
 import sct_utils as sct
 import sct_image
@@ -298,10 +299,8 @@ class Transform:
 
         if islabel:
             sct.printv("\nTake the center of mass of each registered dilated labels...")
-            sct.run(['sct_label_utils',
-                     '-i', fname_out,
-                     '-o', fname_out,
-                     '-cubic-to-point'])
+            out = cubic_to_point(im_src_reg)
+            out.save(path=fname_out)
             if remove_temp_files:
                 sct.printv('\nRemove temporary files...', verbose)
                 sct.rmtree(path_tmp, verbose=verbose)
--- a/scripts/sct_apply_transfo.py
+++ b/scripts/sct_apply_transfo.py
@@ -299,7 +299,8 @@ class Transform:
 
         if islabel:
             sct.printv("\nTake the center of mass of each registered dilated labels...")
-            cubic_to_point(im_src_reg).save(path=fname_out)
+            labeled_img = cubic_to_point(im_src_reg)
+            labeled_img.save(path=fname_out)
             if remove_temp_files:
                 sct.printv('\nRemove temporary files...', verbose)
                 sct.rmtree(path_tmp, verbose=verbose)
--- a/scripts/sct_create_mask.py
+++ b/scripts/sct_create_mask.py
@@ -30,7 +30,6 @@ import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
 from sct_image import concat_data
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
-from sct_label_utils import display_voxel
 from spinalcordtoolbox.labels import create_labels
 from spinalcordtoolbox.types import Coordinate
 
@@ -216,7 +215,7 @@ def create_mask(param):
     if method_type == 'point':
         # extract coordinate of point
         sct.printv('\nExtract coordinate of point...', param.verbose)
-        coord = display_voxel(Image("point_RPI.nii"), verbose=param.verbose)
+        coord = Image("point_RPI.nii").getNonZeroCoordinates()
 
     if method_type == 'center':
         # set coordinate at center of FOV
--- a/scripts/sct_create_mask.py
+++ b/scripts/sct_create_mask.py
@@ -214,13 +214,9 @@ def create_mask(param):
         coord = [x for x in map(int, method_val.split('x'))]
 
     if method_type == 'point':
-        # get file name
         # extract coordinate of point
         sct.printv('\nExtract coordinate of point...', param.verbose)
         coord = display_voxel(Image("point_RPI.nii"), verbose=param.verbose)
-        # parse to get coordinate
-        # TODO fixup... this is quite magic
-        # coord = output[output.find('Position=') + 10:-17].split(',')
 
     if method_type == 'center':
         # set coordinate at center of FOV
--- a/scripts/sct_create_mask.py
+++ b/scripts/sct_create_mask.py
@@ -307,8 +307,7 @@ def create_line(param, fname, coord, nz):
         # compat
         labels = [ Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz) ]
 
-    out = create_labels(Image("line.nii"), labels)
-    out.save(path="line.nii")
+    create_labels(Image("line.nii"), labels).save(path="line.nii")
 
     return 'line.nii'
 
--- a/scripts/sct_create_mask.py
+++ b/scripts/sct_create_mask.py
@@ -30,6 +30,9 @@ import spinalcordtoolbox.image as msct_image
 from spinalcordtoolbox.image import Image
 from sct_image import concat_data
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
+from sct_label_utils import display_voxel
+from spinalcordtoolbox.labels import create_labels
+from spinalcordtoolbox.types import Coordinate
 
 
 # DEFAULT PARAMETERS
@@ -214,11 +217,10 @@ def create_mask(param):
         # get file name
         # extract coordinate of point
         sct.printv('\nExtract coordinate of point...', param.verbose)
-        # TODO: change this way to remove dependence to sct.run. ProcessLabels.display_voxel returns list of coordinates
-        status, output = sct.run(['sct_label_utils', '-i', 'point_RPI.nii', '-display'], verbose=param.verbose)
+        coord = display_voxel(Image("point_RPI.nii"), verbose=param.verbose)
         # parse to get coordinate
         # TODO fixup... this is quite magic
-        coord = output[output.find('Position=') + 10:-17].split(',')
+        # coord = output[output.find('Position=') + 10:-17].split(',')
 
     if method_type == 'center':
         # set coordinate at center of FOV
@@ -299,15 +301,14 @@ def create_line(param, fname, coord, nz):
     # set all voxels to zero
     sct.run(['sct_maths', '-i', 'line.nii', '-mul', '0', '-o', 'line.nii'], param.verbose)
 
-    create_add_str = ""
-    for iz in range(nz):
-        if iz == nz - 1:
-            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1'
-        else:
-            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1:'
+    if isinstance(coord[0], Coordinate):
+        labels = coord
+    else:
+        # compat
+        labels = [ Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz) ]
 
-    cmd = ['sct_label_utils', '-i', 'line.nii', '-o', 'line.nii', '-create-add', create_add_str]
-    sct.run(cmd, param.verbose)
+    out = create_labels(Image("line.nii"), labels)
+    out.save(path="line.nii")
 
     return 'line.nii'
 
--- a/scripts/sct_create_mask.py
+++ b/scripts/sct_create_mask.py
@@ -296,11 +296,14 @@ def create_line(param, fname, coord, nz):
     # set all voxels to zero
     sct.run(['sct_maths', '-i', 'line.nii', '-mul', '0', '-o', 'line.nii'], param.verbose)
 
+    labels = []
+
     if isinstance(coord[0], Coordinate):
-        labels = coord
+        for x, y, _, _ in coord:
+            labels.extend([ Coordinate([x, y, iz, 1]) for iz in range(nz) ])
     else:
-        # compat
-        labels = [ Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz) ]
+        # backwards compat
+        labels.extend([ Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz) ])
 
     create_labels(Image("line.nii"), labels).save(path="line.nii")
 
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -259,6 +259,8 @@ def main(args=None):
         out = sct_labels.increment_z_inverse(img)
     elif arguments.vert_body is not None:
         levels = arguments.vert_body
+        if len(levels) == 1 and levels[0] == 0:
+            levels = None # all levels
         out = sct_labels.label_vertebrae(img, levels)
     elif arguments.vert_continuous:
         out = sct_labels.continuous_vertebral_levels(img)
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -249,7 +249,10 @@ def main(args=None):
         labels = arguments.create_add
         out = sct_labels.create_labels(img, labels)
     elif arguments.create_seg is not None:
-        labels = arguments.create_seg
+        labels = []
+        for str_pair in arguments.create_seg:
+            z, z_val = str_pair.split(',')
+            labels.append(tuple([int(z), int(z_val)]))
         out = sct_labels.create_labels_along_segmentation(img, labels)
     elif arguments.cubic_to_point:
         out = sct_labels.cubic_to_point(img)
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -244,10 +244,7 @@ def main(args=None):
         labels = arguments.create_add
         out = sct_labels.create_labels(img, labels)
     elif arguments.create_seg is not None:
-        labels = []
-        for str_pair in arguments.create_seg:
-            z, z_val = str_pair.split(',')
-            labels.append(tuple([int(z), int(z_val)]))
+        labels = arguments.create_seg
         out = sct_labels.create_labels_along_segmentation(img, labels)
     elif arguments.cubic_to_point:
         out = sct_labels.cubic_to_point(img)
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -300,7 +300,7 @@ def main(args=None):
                     subject=arguments.qc_subject, process='sct_label_utils')
 
 
-def display_voxel(img: Image, verbose: int) -> Sequence[Coordinate]:
+def display_voxel(img: Image, verbose: int = 1) -> Sequence[Coordinate]:
     """
     Display all the labels that are contained in the input image.
     :param img: source image
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -320,8 +320,6 @@ def display_voxel(img: Image, verbose: int = 1) -> Sequence[Coordinate]:
     sct.printv('All labels (useful syntax):', verbose=verbose)
     sct.printv(useful_notation, verbose=verbose)
 
-    return coordinates_input
-
 
 # TODO [AJ] -> just print warnings or log or raise?
 def check_distance_between_labels(img: Image, max_dist_mm: float):
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -62,8 +62,8 @@ def get_parser():
     io_group.add_argument(
         '-o',
         metavar=Metavar.file,
-        help=("Output image. Example: t2_labels.nii.gz"
-              "Note: If no output image is provided, input image will be overwritten!")
+        default='labels.nii.gz',
+        help=("Output image. Example: t2_labels.nii.gz")
     )
 
     io_group.add_argument(
@@ -231,13 +231,10 @@ def main(args=None):
     sct.init_sct(log_level=verbosity, update=True)  # Update log level
 
     input_filename = arguments.i
+    output_fname = arguments.o
+
     img = Image(input_filename)
     dtype = None
-
-    if arguments.o is not None:
-        output_fname = arguments.o
-    else:
-        output_fname = input_filename
 
     if arguments.add is not None:
         value = arguments.add
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -250,6 +250,7 @@ def main(args=None):
         out = sct_labels.cubic_to_point(img)
     elif arguments.display:
         display_voxel(img, verbosity)
+        return
     elif arguments.increment:
         out = sct_labels.increment_z_inverse(img)
     elif arguments.vert_body is not None:
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -277,8 +277,7 @@ def main(args=None):
 
         # second pass use previous pass result as reference
         ref_out = sct_labels.remove_missing_labels(ref, out)
-        ref_out.absolutepath = ref.absolutepath
-        ref_out.save()
+        ref_out.save(path=ref.absolutepath)
     elif arguments.remove is not None:
         labels = arguments.remove
         out = sct_labels.remove_labels_from_image(img, labels)
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -45,7 +45,7 @@ def get_parser():
         prog=os.path.basename(__file__).strip(".py")
     )
 
-    req_group = parser.add_argument_group("\nREQUIRED")
+    req_group = parser.add_argument_group("\nREQUIRED I/O")
     req_group.add_argument(
         '-i',
         metavar=Metavar.file,
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -187,8 +187,10 @@ def get_parser():
 
     optional.add_argument(
         '-v',
-        choices=['0', '1', '2'],
+        choices=[0, 1, 2],
         default=1,
+        metavar=Metavar.int,
+        type=int,
         help="Verbose. 0: nothing. 1: basic. 2: extended."
     )
 
--- a/scripts/sct_label_utils.py
+++ b/scripts/sct_label_utils.py
@@ -265,7 +265,7 @@ def main(args=None):
         out = sct_labels.label_vertebrae(img, levels)
     elif arguments.vert_continuous:
         out = sct_labels.continuous_vertebral_levels(img)
-        dtype='float32'
+        dtype = 'float32'
     elif arguments.MSE is not None:
         ref = Image(arguments.MSE)
         mse = sct_labels.compute_mean_squared_error(img, ref)
@@ -324,6 +324,7 @@ def display_voxel(img: Image, verbose: int):
     sct.printv('All labels (useful syntax):', verbose=verbose)
     sct.printv(useful_notation, verbose=verbose)
 
+
 # TODO [AJ] -> just print warnings or log or raise?
 def check_distance_between_labels(img: Image, max_dist_mm: float):
     """
@@ -345,7 +346,7 @@ def check_distance_between_labels(img: Image, max_dist_mm: float):
                 coordinates_input[i + 1].value) + ' is larger than ' + str(max_dist_mm) + '. Distance=' + str(dist))
 
 
-def launch_sagittal_viewer(img: Image, labels: Sequence[int], previous_points: Sequence[Coordinate] = None, output_img: Image = None, msg: str = "") -> Image:
+def launch_sagittal_viewer(img: Image, labels: Sequence[int], msg: str, previous_points: Sequence[Coordinate] = None, output_img: Image = None) -> Image:
     params = base.AnatomicalParams()
     params.vertebraes = labels
     params.input_file_name = img.absolutepath
@@ -367,7 +368,7 @@ def launch_sagittal_viewer(img: Image, labels: Sequence[int], previous_points: S
     return out
 
 
-def launch_manual_label_gui(img: Image, input_labels_img: Image, labels: Sequence[int], msg: str = None):
+def launch_manual_label_gui(img: Image, input_labels_img: Image, labels: Sequence[int], msg):
     # the input image is reoriented to 'SAL' when open by the GUI
     input_labels_img.change_orientation('SAL')
     mid = int(np.round(input_labels_img.data.shape[2] / 2))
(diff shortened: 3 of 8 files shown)


joshuacwnewton:
@Drulex #2842 has now been merged (my apologies for the delay), so you should be able to base your changes off `master`, now.

Drulex:
@joshuacwnewton @jcohenadad , there are still a few little things WIP, but it would be good to start reviewing this PR.

conftest.py
```diff
@@ -19,12 +20,8 @@
...
-    # session scope at the moment.
-    # https://github.com/pytest-dev/pytest/issues/2704#issuecomment-603387680
-    capmanager = request.config.pluginmanager.getplugin("capturemanager")
-    with capmanager.global_and_fixture_disabled():
+def pytest_collectstart():
+    # TODO [AJ] check integrity eventually
+    if not os.path.exists(sct_test_path()):
         print('\nDownloading sct testing data.')
-        downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
+        downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
```
joshuacwnewton, line 27: There is overlap with #2903, right? i.e. if that PR gets merged first, these changes won't be needed?

joshuacwnewton (commented):
I started to look at this and got caught up exploring the other testing data-related issues/PRs. I plan to continue a review later today. :slightly_smiling_face:

conftest.py
Drulex: Yes sorry for the confusion, I will rebase cleanly.

joshuacwnewton: No worries! Thanks much. :smile:

lrouhier:
I tested this version with multiple images for disc labeling and it worked fine with the GUI 👍  though the test seems to fail ...

Drulex:
> I tested this version with multiple images for disc labeling and it worked fine with the GUI +1

Excellent thanks a lot! The GUI is what I was most worried about. 

> though the test seems to fail ...

It's ok, the unit tests will fail until #2903 is merged

spinalcordtoolbox/labels.py
```diff
@@ -0,0 +1,473 @@
...
+            out.data[int(list_coordinates[i].x), int(list_coordinates[i].y), int(list_coordinates[i].z)] = 0
+
+    if out.orientation != og_orientation:
+        out.change_orientation(og_orientation)
+        img.change_orientation(og_orientation)
+
+    return out
+
+
+# FIXME [AJ]: this is slow on large images
```
joshuacwnewton, line 270: This looks fun! I love NumPy optimizations like this. LMK if you want help (or if you want me to do it.)

scripts/sct_label_utils.py
```diff
@@ -21,711 +21,81 @@
...
-                c_p = im_dest.transfo_phys2pix([[c.x, c.y, c.y]], real=False)[0]
-            else:
-                raise ValueError("The value of 'type' should either be 'discrete' or 'continuous'.")
-            dest_coord.append(CoordinateValue([c_p[0], c_p[1], c_p[2], c.value]))
-        return dest_coord
-
-    def diff(self):
-        """
-        Detect any label mismatch between input image and reference image
-        """
```
joshuacwnewton, line 536: I can't find this in the new `labels.py` api file. Was this not used anywhere?

scripts/sct_label_utils.py
```diff
@@ -850,93 +225,185 @@ def main(args=None):
...
+
+
+# TODO [AJ] -> just print warnings or log or raise?
+def check_distance_between_labels(img: Image, max_dist_mm: float):
+    """
+    Calculate the distances between each label in the input image.
+    If a distance is larger than max_dist_mm, a warning message is displayed.
+    :param img: input image
+    :param max_dist: maximum distance (in mm) allowed between labels
+    """
```
joshuacwnewton, line 333: What was the rationale for keeping some functions (like this one) in `sct_label_utils` instead of moving them into `labels.py`?

spinalcordtoolbox/labels.py
```diff
@@ -0,0 +1,473 @@
...
+    return out
+
+
+def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
+    """
+    Remove labels other than specified from an image
+    :param img: source image
+    :param labels: list of specified labels to keep
+    :returns: image with labels specified kept only
+    """
```
joshuacwnewton, line 462: Could this function just be a parameter in `remove_labels_from_image`?

For example `inverse` with description `If inverse=True, this removes all labels *except* the ones provided`.

scripts/sct_label_utils.py
```diff
@@ -850,93 +225,185 @@ def main(args=None):
...
+        ref = Image(arguments.remove_reference)
+        out = sct_labels.remove_missing_labels(img, ref)
+    elif arguments.remove_sym is not None:
+        # first pass use img as source
+        ref = Image(arguments.remove_reference)
+        out = sct_labels.remove_missing_labels(img, ref)
+
+        # second pass use previous pass result as reference
+        ref_out = sct_labels.remove_missing_labels(ref, out)
+        ref_out.save(path=ref.absolutepath)
```
joshuacwnewton, line 281: Wow, you did a great job at simplifying the "remove labels" functions. This replaces a _lot_ of old code.

scripts/sct_label_utils.py
```diff
@@ -21,711 +21,81 @@
...
-            elif len(im_output_rpi.data.shape) == 2:
-                assert str(z) == '0', "ERROR: 2D coordinates should have a Z value of 0. Z coordinate is :" + str(z)
-                im_output_rpi.data[x, y] = value
-        # change orientation back to native
-        return im_output_rpi.change_orientation(self.image_input.orientation)
-
-    def plan(self, width, offset=0, gap=1):
-        """
-        Create a plane of thickness="width" and changes its value with an offset and a gap between labels.
-        """
```
joshuacwnewton, line 270: I can't find this in the new `labels.py` api file. Was this not used anywhere?

scripts/sct_label_utils.py
```diff
@@ -21,711 +21,81 @@
...
-        # for all points with non-zeros neighbors, force the neighbors to 0
-        for coord in coordinates_input:
-            image_output.data[:, :, int(coord.z) - width:int(coord.z) + width] = offset + gap * coord.value
-
-        return image_output
-
-    def plan_ref(self):
-        """
-        Generate a plane in the reference space for each label present in the input image
-        """
```
joshuacwnewton, line 284: I can't find this in the new `labels.py` api file. Was this not used anywhere?

joshuacwnewton (changes requested):
Phew! Gigantic effort here, so thank you @Drulex.

I've done a first pass over everything except the new tests. Stopping now because it's late, and I've said enough already! :stuck_out_tongue:

scripts/sct_apply_transfo.py
```diff
@@ -91,10 +92,10 @@ def get_parser():
...
         "-x",
-        help=""" Interpolation method. The 'label' method is to be used if you would like to apply a transformation 
-        on a file that has single-voxel labels (classical interpolation methods won't work, as resampled labels might 
-        disappear or their values be altered). The function will dilate each label, apply the transformation using 
-        nearest neighbour interpolation, and then take the center-of-mass of each "blob" and output a single voxel per 
+        help=""" Interpolation method. The 'label' method is to be used if you would like to apply a transformation
+        on a file that has single-voxel labels (classical interpolation methods won't work, as resampled labels might
+        disappear or their values be altered). The function will dilate each label, apply the transformation using
+        nearest neighbour interpolation, and then take the center-of-mass of each "blob" and output a single voxel per
         blob.""",
```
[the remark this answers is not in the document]
Drulex: No it's PEP8 formatter.

scripts/sct_label_utils.py
Drulex: Initially there was a `ProcessLabels` class with a lot of labels processing logic. All of that was moved in the API module.

Then there were some free functions laying around in `sct_label_utils.py`. By default I kept all of them at the application level. I'll do another pass.

scripts/sct_label_utils.py
Drulex: Not used anywhere

scripts/sct_label_utils.py
Drulex: Not used anywhere

scripts/sct_label_utils.py
Drulex: :smile:  Yes it was a bit convoluted before.

spinalcordtoolbox/labels.py
Drulex: It could, and this may be just a question of style, but I like to have functions as specific as possible (SRP) in order to simplify tests and maintenance. 

I though that since we are iterating over potentially _a lot_ of labels (depending on size on input image) the check for `inverse` would unnecessarily slow down the code.

scripts/sct_create_mask.py
```diff
@@ -30,6 +30,9 @@
 from spinalcordtoolbox.image import Image
 from sct_image import concat_data
 from spinalcordtoolbox.utils import Metavar, SmartFormatter
+from sct_label_utils import display_voxel
```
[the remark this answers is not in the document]
Drulex: Yes you're right, I'll clean this up

scripts/sct_create_mask.py
```diff
@@ -299,15 +297,13 @@ def create_line(param, fname, coord, nz):
...
-            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1:'
+    if isinstance(coord[0], Coordinate):
+        labels = coord
+    else:
+        # compat
+        labels = [ Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz) ]
 
-    cmd = ['sct_label_utils', '-i', 'line.nii', '-o', 'line.nii', '-create-add', create_add_str]
-    sct.run(cmd, param.verbose)
+    create_labels(Image("line.nii"), labels).save(path="line.nii")
```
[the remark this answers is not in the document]
Drulex: :smile:

spinalcordtoolbox/types.py
```diff
@@ -89,6 +89,9 @@ def __init__(self, coord=None, mode='continuous'):
         except ValueError:
             raise TypeError("All coordinates must be int and the value can be a float or a int. x=" + str(self.x) + ", y=" + str(self.y) + ", z=" + str(self.z) + ", value=" + str(self.value))
 
+    def __iter__(self):
+        return iter((self.x, self.y, self.z, self.value))
+
```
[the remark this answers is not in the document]
Drulex: See commit comment: https://github.com/neuropoly/spinalcordtoolbox/pull/2887/commits/9d1b29a23fb9567a9bff222fa32f08106b579d7e

image

Arguably cleaner than the pattern below:
```python
for coord in img.getNonZeroCoordinates():
    do_something(coord.x)
    do_something(coord.y)
    do_something(coord.z)
    do_something(coord.value)
```
Although probably not faster.

spinalcordtoolbox/labels.py
joshuacwnewton: That seems very fair to me. And you're right, it is mostly a nitpicky style thing, so I don't think it's a big deal. :)

spinalcordtoolbox/types.py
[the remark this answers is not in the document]
joshuacwnewton: Aha! I understand now. That is very clean, although I had a little trouble understanding it at first glance. Maybe a comment would help here.

```suggestion
    def __iter__(self):
        # Allows for this usage: "for x, y, z, v in [list of Coordinate]"
        return iter((self.x, self.y, self.z, self.value))

```

scripts/sct_label_utils.py
Drulex: Not used anywhere.

scripts/sct_label_utils.py
joshuacwnewton: @Drulex Instead of removing them, is possible we still want access to the `diff`, `plan`, and `plan_ref` functions, e.g. for future development? Might be worth checking in with someone who might have used them in the past, e.g. @jcohenadad.

scripts/sct_label_utils.py
```diff
@@ -21,711 +21,81 @@
...
-        "--help",
-        action="help",
-        help="Show this help message and exit."
+    io_group = parser.add_argument_group("\nOPTIONAL I/O")
+
+    io_group.add_argument(
+        '-o',
+        metavar=Metavar.file,
+        default='labels.nii.gz',
+        help=("Output image. Example: t2_labels.nii.gz")
```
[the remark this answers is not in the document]
Drulex: I agree. I kept the same CLI interface for backwards compatibility, but I have to admit that it isn't the most user friendly.

scripts/sct_label_utils.py
```diff
@@ -21,711 +21,81 @@
...
         type=list_type(':', Coordinate),
         help="Same as '-create', but add labels to the input image instead of creating a new image. "
              "Example: 12,34,32,1:12,35,33,2"
     )
-    optional.add_argument(
+
+    func_group.add_argument(
         '-create-seg',
         metavar=Metavar.list,
         type=list_type(':', str),
```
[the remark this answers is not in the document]
Drulex: Good catch, I must have missed during the rebase merge conflicts resolutions.

scripts/sct_apply_transfo.py
```diff
@@ -299,7 +299,8 @@ def apply(self):
 
         if islabel:
             sct.printv("\nTake the center of mass of each registered dilated labels...")
-            cubic_to_point(im_src_reg).save(path=fname_out)
+            labeled_img = cubic_to_point(im_src_reg)
+            labeled_img.save(path=fname_out)
```
joshuacwnewton: Ah! This is present in more places than just `cubic_to_point`. My apologies, I didn't point all of them out in my initial review. 

But, it's easy enough to spot, e.g. `").save"` can be searched for.

joshuacwnewton (changes requested):

scripts/sct_label_utils.py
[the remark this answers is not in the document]
Drulex: Ah no wait I parse it myself later down, I need a list of tuples.

scripts/sct_label_utils.py
Drulex: A wise man once said:
> if it's not usable nor unit-tested it's just a liability...

credit @zougloub

scripts/sct_create_mask.py
```diff
@@ -211,14 +214,9 @@ def create_mask(param):
...
     if method_type == 'point':
-        # get file name
         # extract coordinate of point
         sct.printv('\nExtract coordinate of point...', param.verbose)
-        # TODO: change this way to remove dependence to sct.run. ProcessLabels.display_voxel returns list of coordinates
-        status, output = sct.run(['sct_label_utils', '-i', 'point_RPI.nii', '-display'], verbose=param.verbose)
-        # parse to get coordinate
-        # TODO fixup... this is quite magic
-        coord = output[output.find('Position=') + 10:-17].split(',')
+        coord = display_voxel(Image("point_RPI.nii"), verbose=param.verbose)
```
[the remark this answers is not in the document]
Drulex: Yea this one was a head scratcher..

spinalcordtoolbox/labels.py
```diff
@@ -0,0 +1,473 @@
...
+    Create an image with labels defined along the spinal cord segmentation (or centerline).
+    Input image does **not** need to be RPI (re-orientation is done within this function).
+    :param img: source segmentation
+    :param labels: list of label tuples as (z_value, label_value)
+    :returns: labeled segmentation (Image)
+    """
+    og_orientation = img.orientation
+
+    if og_orientation != "RPI":
+        img.change_orientation("RPI")
```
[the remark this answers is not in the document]
Drulex: The low-level wasn't doing the check. Perhaps the check is better done at the low-level, but I don't think it's bad to have it here. 

In any case the check is less expensive than changing an already properly oriented image.

spinalcordtoolbox/labels.py
```diff
@@ -0,0 +1,473 @@
...
+
+        if len(out.data.shape) == 3:
+            out.data[x, y, z_rpi] = value
+        elif len(out.data.shape) == 2:
+            if z != 0:
+                raise ValueError(f"2D coordinates should have a Z value of 0! Current value: {coord.z}")
+
+            out.data[x, y] = value
+
+    if out.orientation != og_orientation:
```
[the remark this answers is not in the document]
Drulex: Yea I guess it would make sense to move the check at the low level (see https://github.com/neuropoly/spinalcordtoolbox/pull/2887/files#r496073836)

spinalcordtoolbox/labels.py
```diff
@@ -0,0 +1,473 @@
...
+        out.change_orientation(og_orientation)
+        img.change_orientation(og_orientation)
```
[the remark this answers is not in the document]
Drulex: `img` is used in the computation of `out`. If the orientation is not `RPI` it is re-oriented, therefore I am restoring the original orientation. If I didn't that would cause a side-effect for the caller.

The alternative is to make a copy of `img` and use that in the processing. In this specific case I opted to avoid the image copy for performance reasons. I would not want to sacrifice readability for (a probably small) RAM performance improvement however. 

Do you think this negatively affects the readability of the functions?

scripts/sct_label_utils.py
joshuacwnewton: Fair point! :)

scripts/sct_label_utils.py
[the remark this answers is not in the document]
joshuacwnewton: Is a list of tuples meaningfully different than a list of lists, in this context? 

Although, I wonder if I can update my `list_type` factory function to also allow for tuples. Hrm.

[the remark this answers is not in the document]
Drulex: Yes sorry I glanced too quickly. It should be compatible, I'll adjust.

scripts/sct_apply_transfo.py
Drulex: Is this bad though? If the returned image won't be used it's fair to no store it in a variable right?

spinalcordtoolbox/labels.py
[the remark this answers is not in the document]
joshuacwnewton: > Do you think this negatively affects the readability of the functions?

To me it does, but only a little. It's a small trade-off either way, so I trust your judgement here. :)

spinalcordtoolbox/labels.py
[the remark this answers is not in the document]
joshuacwnewton: No worries, I don't think it's a huge deal either way. I think I agree with your rationale for both checks. :)

scripts/sct_apply_transfo.py
joshuacwnewton: Ah, right. This is the same tradeoff as the "make a copy for RPI conversion", isn't it? (readability vs. memory usage)

I tend to err on the side of readability, but you raise a really good point. Honestly, I'm not knowledgable enough about Python's garbage collection, so it's hard for me to say how much of a cost it really is.

For now I'm happy to go with whatever your preference is. :)

Drulex: > Ah, right. This is the same tradeoff as the "make a copy for RPI conversion", isn't it? (readability vs. memory usage)

In this specific case no, it is simply [method chaining](https://en.wikipedia.org/wiki/Method_chaining) for the sake of readability. `cubic_to_point()` returns an `Image,` but we only save it and do nothing else with it so why store it in a variable.

```python
cubic_to_point(img).save()
```
is equivalent to 
```python
out = cubic_to_point(img)
out.save()
```
But `out` is never used anywhere so why not use the first form

joshuacwnewton: > I have to admit that cubic_to_point() doesn't really hint that it would return an Image, that's probably the confusing part.

Yeah! By splitting it up, it's clearer that:
* `cubic_to_point()` returns an `Image`
* That `save()` is a method of an `Image` object.

Mostly just subjective / personal preference, though, considering how much back and forth goes on in threads like this: https://stackoverflow.com/questions/1103985/method-chaining-why-is-it-a-good-practice-or-not (Lots of interesting discussion, though!)

scripts/sct_create_mask.py
```diff
@@ -299,15 +296,13 @@ def create_line(param, fname, coord, nz):
     # set all voxels to zero
     sct.run(['sct_maths', '-i', 'line.nii', '-mul', '0', '-o', 'line.nii'], param.verbose)
 
-    create_add_str = ""
-    for iz in range(nz):
-        if iz == nz - 1:
-            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1'
-        else:
-            create_add_str += str(int(coord[0])) + ',' + str(int(coord[1])) + ',' + str(iz) + ',1:'
+    if isinstance(coord[0], Coordinate):
```
jcohenadad, line 299: i've noticed a change of behaviour, and i _think_ it comes from there. Below is a result from the following syntax:
~~~
sct_create_mask -i t2.nii.gz -p point,label.nii -o mask_point.nii
~~~
The gif anim alternates between master (c3054eb25ab9befc920446be8282bac05ccfb2a4) and aj/2870-refactor_sct_label_utils (50daa2e24575e8026982de14488d500e5ba73c71):
anim

As can be seen, the mask generated from the code in this branch only spans a single slice, while it should extend across all the slices in the superior-inferior direction. The "create_line" function is supposed to create a vertical line out of the input single coord, and i think that when inputing a single coord (Coordinate) object, it does not iterate across z (while it should).

scripts/sct_label_utils.py
```diff
@@ -850,93 +225,159 @@ def main(args=None):
...
-    #     process_type = 'cross'
-    #     input_cross_radius = arguments.cross
+        labels = arguments.create_seg
+        out = sct_labels.create_labels_along_segmentation(img, labels)
     elif arguments.cubic_to_point:
-        process_type = 'cubic-to-point'
+        out = sct_labels.cubic_to_point(img)
     elif arguments.display:
-        process_type = 'display-voxel'
+        display_voxel(img, verbosity)
```
jcohenadad, line 252: i am getting this error:
````
sct_label_utils -i label.nii -display

--
Spinal Cord Toolbox (git-aj/2870-refactor_sct_label_utils-50daa2e24575e8026982de14488d500e5ba73c71)

Position=(25,31,26) -- Value= 3
All labels (useful syntax):
25,31,26,3
  File "/Users/julien/code/sct/scripts/sct_label_utils.py", line 386, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_label_utils.py", line 293, in main
    out.save(path=output_fname, dtype=dtype)
UnboundLocalError: local variable 'out' referenced before assignment
````

Comment (Drulex) [reply scripts/sct_label_utils.py:252]: Fixed in 1927188

File: scripts/sct_create_mask.py (hunk 20 above)
Comment (Drulex) [reply scripts/sct_create_mask.py:299]: wow good catch. Investigating.

Comment (Drulex) [reply scripts/sct_create_mask.py:299]: @jcohenadad could you share the exact input files you used to reproduce the issue? That would make my life easier.

Comment (jcohenadad) [reply scripts/sct_create_mask.py:299]: github is buggy... cannot upload. So here you go: https://www.dropbox.com/s/klgfekh16k4bry2/Archive.zip?dl=0

Comment (Drulex) [reply scripts/sct_create_mask.py]: Fixed in 0a6ca12

Comment (jcohenadad) [reply scripts/sct_create_mask.py:301]: tested with:
- <coord,XxY>
- <point,FILE>
- <center>
- <centerline,FILE>

all good!

**Review by jcohenadad — APPROVED** [review]

**Comment by Drulex** [comment]
@jcohenadad thanks for taking the time to look. If you have any quick ideas for the unit tests marked `TODO [AJ] implement test` I would appreciate, currently the unit test just ensures that it runs without crashing.

File: unit_testing/test_labels.py [hunk 22]
```diff
@@ -0,0 +1,208 @@
...
+
+@pytest.mark.parametrize("test_seg", [seg_img])
+def test_create_labels_along_segmentation(test_seg):
+    a = test_seg.copy()
+    labels = [(5, 1), (14, 2), (23, 3)]
+
+    b = sct_labels.create_labels_along_segmentation(a, labels)
+
+    assert b.orientation == a.orientation
+    # TODO [AJ] implement test
```
jcohenadad, line 89: simple way (assumes test data don't change):
- run the test and see where the output labels are, then hard-code the labels coordinates and verify that they match the output of this test (e.g. image difference) 

complex way
- compute center of mass at the indicated slices, and verify that those correspond to the output labels

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_cubic_to_point(test_image):
+    a = test_image.copy()
+    sct_labels.cubic_to_point(a)
+    # TODO [AJ] implement test
```
jcohenadad, line 96: same as in https://github.com/neuropoly/spinalcordtoolbox/pull/2887/files#r496933625

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_increment_z_inverse(test_image):
+    a = test_image.copy()
+    sct_labels.increment_z_inverse(a)
+    # TODO [AJ] implement test
```
jcohenadad, line 103: - create label image:
  - im = Image(t2.nii.gz)
  - im.data = 0 everywhere
  - im.data[5, 5, 1] = 1
  - im.data[5, 5, 3] = 5
  - im.data[5, 5, 6] = 2
- create ground truth:
  - im_test = im.copy
  - im_test_data[5, 5, 1] = 1
  - im_test_data[5, 5, 3] = 2
  - im_test_data[5, 5, 6] = 3
- assert a == im_test

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("test_seg,test_labels", [(seg_img, labels_img)])
+def test_labelize_from_disks(test_seg, test_labels):
```
jcohenadad, line 107: `labelize_from_disks` corresponded to the flag `-disks` (https://github.com/neuropoly/spinalcordtoolbox/pull/2887/files#diff-f5b39fbae8b0d2fda75fb8532e60c386L103), which I don't see anymore on the refactored version. Note: i don't think it was there before either...

so, i'm wondering if this API is used at all?

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+    sct_labels.labelize_from_disks(seg, ref)
+    # TODO [AJ] implement test
+
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_label_vertebrae(test_image):
+    a = test_image.copy()
+    sct_labels.label_vertebrae(a)
+    sct_labels.label_vertebrae(a, [1, 2, 3, 4])
+    # TODO [AJ] implement test
```
jcohenadad, line 120: same as in https://github.com/neuropoly/spinalcordtoolbox/pull/2887/files#r496933625

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+    src = test_image.copy()
+    ref = test_image.copy()
+
+    for z in range(3):
+        for y in range(2):
+            for x in range(1):
+                ref.data[x, y, z] *= 10
+
+    sct_labels.compute_mean_squared_error(src, ref)
+    # TODO [AJ] implement test
```
jcohenadad, line 134: i would compute it from the test image, get the number and hard-code it here

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.skip(reason="Too long to run on large image!")
```
jcohenadad, line 137: indeed, i've also noticed that. The algorithm is sub-optimal and should be changed (not in this PR of course)

unit_testing/test_labels.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_continuous_vertebral_levels(test_image):
+    a = test_image.copy()
+    b = sct_labels.continuous_vertebral_levels(a)
+
+    # check that orientation is maintained
+    assert b.orientation == a.orientation
+    # TODO [AJ] implement test
```
jcohenadad, line 169: i would run it and hard-code the output (eg: test a portion of the image that corresponds to the continuously-labeled segmentation, and assert if the vectors are the same)

unit_testing/test_labels.py
Drulex: `-disk` was in missing from the parser in the previous version, probably why I forgot it. I can add it since the API function is here.

jcohenadad: ok, if you add it could you please call it `-disc` instead, since we decided to go with that convention for spelling intervertebral discs

unit_testing/test_labels.py
Drulex: Is it possible that ground truth should be:
```
im_test_data[5, 5, 1] = 3
im_test_data[5, 5, 3] = 2
im_test_data[5, 5, 6] = 1
```
instead of 
```
im_test_data[5, 5, 1] = 1
im_test_data[5, 5, 3] = 2
im_test_data[5, 5, 6] = 3
```
?

jcohenadad: yup! i think you're right (probably why the function name includes "inverse")

Drulex:
I added a few more unit tests based on feedback from @jcohenadad. There are still a few missing, but since this is already significantly better than before (no unit tests before) I'm thinking of leaving it as is and focusing on other efforts. Want to add something @joshuacwnewton?

jcohenadad (approved):
fantastic contribution, thank you!

joshuacwnewton:
Sorry for the delay, @Drulex. Looking at this now. :)

unit_testing/test_labels.py
```diff
@@ -0,0 +1,218 @@
...
+    expected = zeros_like(a)
+
+    labels = [Coordinate(l) for l in [[0, 0, 0, 7], [0, 1, 2, 5]]]
+    expected.data[0, 0, 0] = 7
+    expected.data[0, 1, 2] = 5
+
+    b = sct_labels.create_labels_empty(a, labels)
+
+    diff = b.data == expected.data
+    assert diff.all()
```
joshuacwnewton, line 67: Just curious, why this form instead of using [`numpy.testing`](https://numpy.org/doc/stable/reference/routines.testing.html)?

joshuacwnewton (approved):
LGTM! I just had one question about test style but it's not critical. Just thinking about what to add to [this section of the Wiki's testing page](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests#unit-tests).

unit_testing/test_labels.py
joshuacwnewton: Ah, sorry, my apologies. Disregard this, I misremembered something.

PR #2891, opened 2020-09-10
sct_compute_mtsat: Check json sidecars only if Repetition Time/Flip Angle arguments aren't first provided

### LINKED ISSUE #2883: batch_processing.sh: FileNotFoundError: No such file or directory: 'mt1_crop.json'
### Description

`./batch_processing.sh` now fails as a result of #2875.

I believe the cause is that `sct_example_data` was not updated to include the relevant `json` sidecar files. (Only `sct_testing_data` was updated, in https://github.com/neuropoly/spinalcordtoolbox/pull/2875/commits/b6d99e015e80aa9f497ee256a64b98b12d5a43c8.)

### Steps to Reproduce

* Check out a recent revision on `master`.
* Run `./batch_processing.sh`. Result:

```bash
# Compute MTsat
# Tips: Check your TR and Flip Angle from the Dicom data
sct_compute_mtsat -mt mt1_crop.nii.gz -pd mt0_reg.nii.gz -t1 t1w_reg.nii.gz -trmt 30 -trpd 30 -trt1 15 -famt 9 -fapd 9 -fat1 15
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/sct_compute_mtsat.py", line 221, in <module>
    main(sys.argv[1:])
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/sct_compute_mtsat.py", line 179, in main
    parser = get_parser(argv)
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/sct_compute_mtsat.py", line 74, in get_parser
    get_json_file_name(parser.parse_known_args(argv)[0].mt, check_exist=True), 'RepetitionTime')
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/sct_compute_mtsat.py", line 170, in fetch_metadata
    with open(fname_json) as f:
FileNotFoundError: [Errno 2] No such file or directory: 'mt1_crop.json'
```

joshuacwnewton:
### Related issues/PRs

Fixes #2883 

### Description

As we read `RepetitionTime`/`FlipAngle` values from `json` sidecars, we raise an error if the sidecar file doesn't exist. This happens even if a value is provided for the argument. But, if an argument is being provided, then we don't need to check the sidecar in the first place.

So, to fix this, we check sidecar files only if an argument hasn't been provided.

This PR also fixes the `FileNotFoundError` so that it's properly raised. (This wasn't mentioned in #2833, but was a bug nonetheless.)


jcohenadad (commented):
i guess there is no more need to pass `argv` here?
https://github.com/neuropoly/spinalcordtoolbox/blob/beb824ab495febcfc542f456cb084b0f71ea73d7/scripts/sct_compute_mtsat.py#L168

joshuacwnewton:
> i guess there is no more need to pass argv here?

Right now it's still needed for the test, because it was written using a `main()` call. 

https://github.com/neuropoly/spinalcordtoolbox/blob/c7ec03543192a110e3e3177ae7c4ecefb97600cf/unit_testing/test_sct_compute_mtsat.py#L24-L28

But, I think this test can be refactored. e.g. as a regular `sct_testing` test. That would indeed make the `argv` passing unnecessary.

jcohenadad:
> I believe right now it's still needed for the test, because it was written using a `main()` call.

ah! you're right

joshuacwnewton:
@jcohenadad is something like d777173 alright? Or is it preferred for these types of tests (i.e. `test_integrity`) to check the output more thoroughly (e.g specific voxel values)? 

Asking as I'm still a little shaky on the conventions SCT uses for writing its tests.

jcohenadad:
> @jcohenadad is something like [d777173](https://github.com/neuropoly/spinalcordtoolbox/commit/d777173258d0e1149d5198f1954a320ba8c41b0b) alright? Or is it preferred for these types of tests (i.e. `test_integrity`) to check the output more thoroughly (e.g specific voxel values)?

I think the idea is to progressively get rid of `sct_testing`, to be replaced exclusively by the more traditional pytest. @cfhammill started an example of using pytest for CLI with his test_sct_run_batch, so I used the same approach. I cannot find the discussion about this though... ( i haven't searched much though-- in the middle of something else)

joshuacwnewton:
> I think the idea is to progressively get rid of sct_testing, to be replaced exclusively by the more traditional pytest. @cfhammill started an example of using pytest for CLI with his test_sct_run_batch, so I used the same approach.

Ahh, I see! I also found the [**Testing - Make your SCT function testing-compatible**](https://github.com/neuropoly/spinalcordtoolbox/wiki/Misc%3A-Testing#make-your-sct-function-testing-compatible) section of the internal dev Wiki, but that page appears to be out of date compared to what you describe.

>  I cannot find the discussion about this though... ( i haven't searched much though-- in the middle of something else)

No worries! It is understandable, considering the amount of discussions! :stuck_out_tongue: 

As an aside (and I sound like a broken record here, hehe), but hopefully if we update the internal developer Wiki when decisions like this are made, this will be less of an issue in the future. :smile:

joshuacwnewton:
I think we can leave the `argv`/testing stuff as-is for now, then? 

For the core changes in this PR, they let `batch_processing.sh` run without error:

```
set +v
~~~
Version:         git-jn/2883-fix-json-sidecar-loading-1dd2cd8201a6870bbaa0b08fc890b7cc6d0272a7
Ran on:          Linux pop-os 5.4.0-7642-generic
Duration:        0hrs 18min 11sec
---
mt/MTR(WM):      54.40129136937407
dmri/FA(CST_r):  0.7756557354302693
dmri/FA(CST_l):  0.7801552118509679
~~~
To open Quality Control (QC) report on a web-browser, run the following:
xdg-open /home/joshua/Repos/spinalcordtoolbox/qc_example_data/index.html
```

If someone else could verify, though, that would be lovely.

jcohenadad:
> I think we can leave the argv/testing stuff as-is for now, then?

absolutely!

jcohenadad (approved):
thanks for the fix!

PR #2896, opened 2020-09-18
Update verbose arguments and usage to be integer 0/1 instead of string 0/1/2

### LINKED ISSUE #2676: Change verbosity flag `-v 1` -> `-v`
It's more conventional in unix to pass `--verbose` or `-v` as a flag, not an option, i.e. it's existence determines whether it applies, not an argument passed after it.

Our wiki [used to suggest](https://github.com/neuropoly/spinalcordtoolbox/wiki/variables/_compare/97006a885dde276c9e8469f3f69b0089639eb609...0de1bc657993c94740da8db96a21cf162b8a49c5) an argument of `-v 0` through `-v 2` to control log level. I've changed this in `sct_testing` to be just `` for "off" or `-v` for "on". We should go through the rest of our CLIs and make them conform.

We can use [`action="count"`](https://docs.python.org/3/library/argparse.html#action) so you can do `$script -v -v -v` or `$script --verbose -vv` or `$script -vvv`

Ref: https://github.com/neuropoly/spinalcordtoolbox/pull/2672#issuecomment-619124796_

joshuacwnewton:
### Related PRs/issues

Fixes #2676.

### Description

* Standardizes `-v` options to use integer 0/1 which switches between INFO/DEBUG (instead of string 0/1/2 as was before)
* Corrects usage of verbose in sct.run/main() calls
* Combines numerical checks for verbose=1/2, e.g. `if verbose == "2"`, `if verbose > 1`, etc. become `if verbose:`

### DIFF
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -94,12 +94,13 @@ def get_parser():
         default=1,
         choices=(0, 1))
     optional.add_argument(
-        "-v",
+        '-v',
+        metavar=Metavar.int,
         type=int,
-        help="Verbose: 0 = nothing, 1 = classic, 2 = expended",
-        required=False,
-        choices=(0, 1, 2),
-        default=1)
+        choices=(0, 1),
+        default=1,
+        help="Enable verbose output. 0 = off, 1 = on.",
+    )
 
     return parser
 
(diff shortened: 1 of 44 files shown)


joshuacwnewton:
My apologies. The work in this PR is not useful. I misunderstood a key detail and need to start over. Please disregard this PR.

PR #2897, opened 2020-09-19
Update CLI argparse verbosity to use `-v` instead of `-v 0/1/2`

### LINKED ISSUE #2626: Centralize deep-learning segmentation tools into a single entry point and module
Currently, deep learning tools include `sct_deepseg_sc`, `sct_deepseg_lesion` and `sct_deepseg_gm`. Each have their entry point (i.e. CLI function) and some of the modules are not shared (e.g., deepseg_gm and deepseg_sc use different modules for preprocessing and running the segmentation). 

Given the increasing need to (i) train models specific to other structures (e.g., CSF #2277, [mouse spinal cord](http://forum.spinalcordmri.org/t/segmetation-centerline-errors/352)), (ii) switch to Pytorch #2305 and (iii) centralize the code for easier maintenance, it would be nice to start implementing a prototype.

A suggestion would be to create a CLI called `sct_deepseg`, which would take an argument to point to the model, e.g., `-m {csf, cord, graymatter, mouse-t1}`.

A possibility would be to get rid of `-c`, if users specify the contrast directly in the model name, e.g. `-m cord-t2`

### LINKED ISSUE #2676: Change verbosity flag `-v 1` -> `-v`

joshuacwnewton:
### Related PRs/Issues

Fixes #2676 .

### Description

The `-v / -verbose` argument is currently handled inconsistently:
* In some scripts, the options are `0/1/2` (`WARNING/INFO/DEBUG`) with `default=1` (or `INFO`).
* In other scripts, the options are `0/1` (`WARNING/INFO`) with `default=1` (or `INFO`).

This PR changes all CLI scripts to use `action="store_true"` instead. This effect of this is:
* Verbosity is `False` (`INFO`) by default
* Verbosity is `True` (`DEBUG`) when `-v` is passed. 

The justification for this is:
* Using `-v` (without a value) is more conventional usage for UNIX CLI scripts.
* The lesser verbosity level (`WARNING`) might not be necessary for users, considering we use `INFO` by default.

### Remaining work

This PR preserves existing behavior wherever `sct.init_sct` is used to set the logging level:

https://github.com/neuropoly/spinalcordtoolbox/blob/ec2d01c1b0e31c63e8ffaad792b66ffbcf2139f0/scripts/sct_qc.py#L87

However, verbosity is _also_ used as-is in some places. (e.g. `param.verbose = arguments.v` then later `if param.verbose:`) This results in some nuanced behavior because the old way (0/1/2) doesn't map cleanly to the new way (False/True).

**For example:** Because the default was `1`, `if verbose:` would always be True. Now, the default is `False`, which causes issues in `sct.printv`, because `printv` doesn't use `sct.init_sct` and logging levels like everything else.

https://github.com/neuropoly/spinalcordtoolbox/blob/ec2d01c1b0e31c63e8ffaad792b66ffbcf2139f0/scripts/sct_utils.py#L798-L817

To fix this, there would need to be a more involved change.

### DIFF
--- a/dev/atlas/validate_atlas/validate_atlas.py
+++ b/dev/atlas/validate_atlas/validate_atlas.py
@@ -200,7 +200,7 @@ def validate_atlas(folder_cropped_atlas, nb_bootstraps, std_noise, range_tract, 
                             index_manual = int(list_methods[i_method][list_methods[i_method].find('man')+3])
                             fname_mask = mask_folder[index_manual] + mask_prefix + list_tracts_txt[i_tract] + mask_ext
                             # manual extraction
-                            status, output = sct.run('sct_average_data_within_mask -i ' + fname_phantom_noise + ' -m ' + fname_mask + ' -v 0')
+                            status, output = sct.run('sct_average_data_within_mask -i ' + fname_phantom_noise + ' -m ' + fname_mask)
                             x_estim_i = float(output)
                         else:
                             # automatic extraction
--- a/scripts/sct_analyze_lesion.py
+++ b/scripts/sct_analyze_lesion.py
@@ -94,12 +94,10 @@ def get_parser():
         default=1,
         choices=(0, 1))
     optional.add_argument(
-        "-v",
-        type=int,
-        help="Verbose: 0 = nothing, 1 = classic, 2 = expended",
-        required=False,
-        choices=(0, 1, 2),
-        default=1)
+        '-v',
+        action="store_true",
+        help="Increase verbosity. Setting this option will enable showing DEBUG messages.",
+    )
 
     return parser
 
@@ -548,7 +546,7 @@ def main(args=None):
 
     # Verbosity
     verbose = arguments.v
-    sct.init_sct(log_level=verbose, update=True)  # Update log level
+    sct.init_sct(log_level=2 if verbose else 1, update=True)
 
     # create the Lesion constructor
     lesion_obj = AnalyzeLeion(fname_mask=fname_mask,
--- a/scripts/sct_analyze_texture.py
+++ b/scripts/sct_analyze_texture.py
@@ -104,12 +104,10 @@ def get_parser():
         choices=(0, 1),
         default=int(Param().rm_tmp))
     optional.add_argument(
-        "-v",
-        help="Verbose: 0 = nothing, 1 = classic, 2 = expended.",
-        required=False,
-        type=int,
-        choices=(0, 1, 2),
-        default=Param().verbose)
+        '-v',
+        action="store_true",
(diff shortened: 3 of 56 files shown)


jcohenadad:
@joshuacwnewton thank you very much for the thorough analysis of the current situation. Some thoughts:
- About the debate for using `store_true` vs. input values: I understand from the dev team that `store_true` is preferred, and i'm happy to align with this, finally closing the debate. I promise I won't reopen the debate 😅 
- Complications with the custom `sct.printv()`. We need to consider a few things here. 
  - First, there is an ongoing refactoring, to separate API code, currently embedded into scripts, to their proper sct python modules. This comes with the convention of changing the printv calls to logging calls. But because this large refactoring is still ongoing, we don't know what will stay "printv" and what will become "logging". 
  - Related to that previous point, and in an attempt to find a solution to minimize the surgery required for this PR, how about we keep the printv calls as they are (ie: they will _always_ be printed), and only touch the parts that ask for the "new" verbose mode (ie: the old 1 or 2 or "1" or "2"). I'm mostly thinking of parts of the code where the variable 'verbose' is passed across scripts and APIs, which could then be replaced by interrogating the logging.level. Does that make sense?

joshuacwnewton:
Thank you for your thoughts, @jcohenadad! I think you're raising some really helpful points. 

Quickly though, before I dive deeper into this: the scope of this is looking to be a little larger than previously thought, and might hold up the release of 5.0.0. For the sake of priotization, I was thinking of putting this on pause to focus on the remaining (smaller) tasks for 5.0.0. Does that sound fair? :slightly_smiling_face:

joshuacwnewton:
I'm closing this PR as I feel a more comprehensive plan is needed for handling verbosity.

While `-v 0/1/2` is indeed nonstandard, it works for our purposes, and I would rather leave things in working order than to try and rush through a half-baked approach for switching to `-v` (i.e. `action=store_true`).

PR #2900, opened 2020-09-20
Convert argparse '0'/'1' options to be typed as ints rather than strings

### LINKED ISSUE #2843: Convert argparse '0'/'1' options to be typed as ints rather than strings
> **Note:** This issue was previously about converting 0/1 options to `store_true`/`store_false`. The description has now been updated to instead standardize how we do 0/1 options.

### Proposal

Our argument parsing uses integer 0/1 for "on/off" options. (This often makes more sense from a user prespective than argparse's `action="store_true"`.) However, the usage for this is often inconsistent:

* Argument is parsed as `0/1 (int)`, then checked using `if argument:`
* Argument is parsed as `"0"/"1" (str)`, then cast to `int` and checked using `if argument:`
* Argument is parsed as `"0"/"1" (str)`, then checked using `if argument == "1":`

We should update all instances of this behavior to the first method, and also update the dev Wiki with relevant info.

EDIT: This PR excludes `verbose` as our usage is slightly more nuanced (despite having similar `"0"/"1"/"2"` values).

### Related issues/PRs

* https://github.com/neuropoly/spinalcordtoolbox/pull/2819#discussion_r467516660
* #2672
* #2676 (Another project-wide argparse update for a more canonical parsing solution.)

joshuacwnewton:
### Related Issues/PRs

Fixes #2843.

### Description

Updates remaining binary options to be typed as integers (rather than using string values). Also updates usage so that `arg == "1"` or `int(arg)` are no longer used.

I've also created a new dev Wiki page called [Argparse in CLI Scripts](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming:-Argparse-in-CLI-scripts) that includes a section for this change called [Integer (0/1) options vs. action="store_true"](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming:-Argparse-in-CLI-scripts#integer-01-options-vs-actionstore_true). (It also combines information from a few other recent discussions.)

### DIFF
--- a/scripts/sct_fmri_moco.py
+++ b/scripts/sct_fmri_moco.py
@@ -105,8 +105,10 @@ def get_parser():
     )
     optional.add_argument(
         '-r',
-        choices=['0', '1'],
-        default='1',
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=1,
         help="Remove temporary files. O = no, 1 = yes"
     )
     optional.add_argument(
@@ -129,7 +131,7 @@ def main():
     arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
     param.fname_data = arguments.i
     param.path_out = arguments.ofolder
-    param.remove_temp_files = int(arguments.r)
+    param.remove_temp_files = arguments.r
     param.interp = arguments.x
     if arguments.g is not None:
         param.group_size = arguments.g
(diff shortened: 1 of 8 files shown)


kousu (commented):
Nice and simple. Good job. Ship it!

kousu (approved):

PR #2901, opened 2020-09-20
sct_maths: Convert usage of convert_list_str to use list_type

### LINKED ISSUE #2844: sct_maths: Convert usage of convert_list_str to use list_type
### Proposal

In previous PRs for converting msct_parser to argparse (#2315, #2337), some list arguments (e.g. in the format `v1,v2,v3`) in `sct_maths` were converted to use the following function: 

https://github.com/neuropoly/spinalcordtoolbox/blob/a67bd5fd74f1aa2eb01fab5ba5feed5fd157ef1f/scripts/sct_maths.py#L436-L451

However, this implementation is rigid in the types and delimiters it allows. In a recent PR (#2819), a factory function was written to allow for more flexible list parsing to handle the same issue in other scripts. So, this issue is to update the usage of this function.

joshuacwnewton:
### Related Issues/PRs

Fixes #2844.

### Description

A factory function (`list_type`) was written as part of #2819 to parse typed lists in arguments. 

`list_type` is more general than the `convert_list_str` function in sct_maths, which was created for a similar purpose. So, this PR replaces the 4 instances of `convert_list_str`.

### DIFF
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -183,7 +183,7 @@ def get_parser():
         type=list_type(',', float),
         help='Gaussian smoothing filtering. Supply values for standard deviations in mm. If a single value is provided, '
              'it will be applied to each axis of the image. If multiple values are provided, there must be one value '
-             'per image axis. (Examples: "-smooth 2,3,2" (3D image), "-smooth 2" (any-D image)).',
+             'per image axis. (Examples: "-smooth 2.0,3.0,2.0" (3D image), "-smooth 2.0" (any-D image)).',
         required=False)
     filtering.add_argument(
         '-laplacian',
@@ -191,7 +191,7 @@ def get_parser():
         type=list_type(',', float),
         help='Laplacian filtering. Supply values for standard deviations in mm. If a single value is provided, it will '
              'be applied to each axis of the image. If multiple values are provided, there must be one value per '
-             'image axis. (Examples: "-laplacian 2,3,2" (3D image), "-laplacian 2" (any-D image)).',
+             'image axis. (Examples: "-laplacian 2.0,3.0,2.0" (3D image), "-laplacian 2.0" (any-D image)).',
         required=False)
     filtering.add_argument(
         '-denoise',
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -179,14 +179,17 @@ def get_parser():
         "-smooth",
         metavar=Metavar.list,
         type=list_type(',', float),
-        help='Gaussian smoothing filter with specified standard deviations in mm for each axis (Example: 2,2,1) or '
-             'single value for all axis (Example: 2).',
+        help='Gaussian smoothing filtering. Supply values for standard deviations in mm. If a single value is provided, '
+             'it will be applied to each axis of the image. If multiple values are provided, there must be one value '
+             'per image axis. (Examples: "-smooth 2,3,2" (3D image), "-smooth 2" (any-D image)).',
         required=False)
     filtering.add_argument(
         '-laplacian',
         metavar=Metavar.list,
         type=list_type(',', float),
-        help='Laplacian filtering with specified standard deviations in mm for all axes (Example: 2).',
+        help='Laplacian filtering. Supply values for standard deviations in mm. If a single value is provided, it will '
+             'be applied to each axis of the image. If multiple values are provided, there must be one value per '
+             'image axis. (Examples: "-laplacian 2,3,2" (3D image), "-laplacian 2" (any-D image)).',
         required=False)
     filtering.add_argument(
         '-denoise',
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -115,9 +115,12 @@ def get_parser():
         "-adap",
         metavar=Metavar.list,
         type=list_type(',', int),
-        help="R|Threshold image using Adaptive algorithm (from skimage). Separate following arguments with ',':"
-             "\n Block size: Odd size of pixel neighborhood which is used to calculate the threshold value (e.g. 3, 7, 21, ...)"
-             "\n Offset: Constant subtracted from weighted mean of neighborhood to calculate the local threshold value. Suggested offset is 0.",
+        help="R|Threshold image using Adaptive algorithm (from skimage). Provide 2 values separated by ',' that "
+             "correspond to the parameters below. For example, '-adap 7,0' corresponds to a block size of 7 and an "
+             "offset of 0.\n"
+             "  - Block size: Odd size of pixel neighborhood which is used to calculate the threshold value. \n"
+             "  - Offset: Constant subtracted from weighted mean of neighborhood to calculate the local threshold "
+             "value. Suggested offset is 0.",
         required=False)
     thresholding.add_argument(
         "-otsu-median",
--- a/scripts/sct_maths.py
+++ b/scripts/sct_maths.py
@@ -23,7 +23,7 @@ import matplotlib
 import matplotlib.pyplot as plt
 
 from spinalcordtoolbox.image import Image
-from spinalcordtoolbox.utils import Metavar, SmartFormatter
+from spinalcordtoolbox.utils import Metavar, SmartFormatter, list_type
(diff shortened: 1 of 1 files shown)


scripts/sct_maths.py
```diff
@@ -114,16 +114,18 @@ def get_parser():
     thresholding.add_argument(
         "-adap",
         metavar=Metavar.list,
+        type=list_type(',', int),
         help="R|Threshold image using Adaptive algorithm (from skimage). Separate following arguments with ',':"
              "\n Block size: Odd size of pixel neighborhood which is used to calculate the threshold value (e.g. 3, 7, 21, ...)"
```
jcohenadad, line 119: just curious: is that OK to have a space after the comma? if not, we might want to update the examples

joshuacwnewton: In this specific case, it's okay, as `int(" 7")` for example will evaluate to `7`. See below for a quick test I did:

```
>>> from spinalcordtoolbox.utils import list_type
>>> str_with_spaces = "3, 7, 21"
>>> type_func = list_type(',', int)
>>> type_func(str_with_spaces)
[3, 7, 21]
```

In other cases (e.g. a list of strings) it could be an issue, though. So, we might want to make the change anyway, just to encourage a consistent input format.

joshuacwnewton: @jcohenadad I've just realized that we both misintepreted! The text `(e.g. 3, 7, 21, ...)` above isn't a list, but actually 3 individual examples of Block size, which only needs a single value. 

I have rewritten the usage to be clearer in d443363b1028fe5fc653abd8c3b06b8af9913d9d

jcohenadad: oh geeze! you're right.  Wow, that was really not a good usage example! (i'm authorized to say that because i was the one writing the usage at the first place 😬 )

joshuacwnewton:
Ah, I've just noticed another instance of a custom function for parsing typed lists.

https://github.com/neuropoly/spinalcordtoolbox/blob/69287b0953fad48bbabc4722d74ccf484ebfbeee/spinalcordtoolbox/utils.py#L227-L240

This parsing could possibly also be handled by argparse and replaced by `list_type`, but I'll have to look more closely into this.

scripts/sct_maths.py
```diff
@@ -269,11 +275,11 @@ def main(args=None):
...
 
     elif arguments.adap is not None:
-        param = convert_list_str(arguments.adap, "int")
+        param = arguments.adap
         data_out = sct_math.adap(data, param[0], param[1])
 
     elif arguments.otsu_median is not None:
-        param = convert_list_str(arguments.otsu_median, "int")
+        param = arguments.otsu_median
         data_out = sct_math.otsu_median(data, param[0], param[1])
```
kousu, line 283: Splay is good for this:

```suggestion
        data_out = sct_math.otsu_median(data, *arguments.otsu_median)
```

But there should be a way to specify that `len(arguments.otsu_median) == 2` or bust. Maybe `list_type()` could be extended to take an size? There's details, like how to distinguish no size from zero size from upper size from specific size.

scripts/sct_maths.py
```diff
@@ -172,14 +177,15 @@ def get_parser():
...
         help='Gaussian smoothing filter with specified standard deviations in mm for each axis (Example: 2,2,1) or '
              'single value for all axis (Example: 2).',
         required=False)
     filtering.add_argument(
         '-laplacian',
-        nargs="+",
-        metavar='',
+        metavar=Metavar.list,
+        type=list_type(',', float),
         help='Laplacian filtering with specified standard deviations in mm for all axes (Example: 2).',
```
kousu, line 189: This example should show the syntax you need to use the list type. And while we're here, does it mean you need to give as many values as axes? So 3 or 4 depending on the image? Except there's a special case if you only give one value?

It's worded difficultly.

```suggestion
        help='Laplacian filtering with specified standard deviations in mm for all axes (Example: 2,3,2).',
```

??

scripts/sct_maths.py
```diff
@@ -114,16 +114,21 @@ def get_parser():
...
+             "  - Offset: Constant subtracted from weighted mean of neighborhood to calculate the local threshold "
+             "value. Suggested offset is 0.",
         required=False)
     thresholding.add_argument(
         "-otsu-median",
+        metavar=Metavar.list,
+        type=list_type(',', int),
         help="R|Threshold image using Median Otsu algorithm. Separate following arguments with ',':"
              "\n Size of the median filter (e.g. 2, 3)"
              "\n Number of iterations (e.g. 3, 4, 5)\n",
```
kousu, line 131: The examples here don't make sense anymore. They should, at least, read "(e.g. 2,3)" or "(e.g. 3, 4, 5)" but would be better written "(e.g. -otsu-median 3,4,5)". I also don't understand how there can be three numbers here when otsu_median() only looks at two.

kousu: I know this derails a bit from the focus of the PR but we should make sure it makes sense.

scripts/sct_maths.py
```diff
@@ -172,14 +177,15 @@ def get_parser():
     filtering = parser.add_argument_group("FILTERING METHODS")
     filtering.add_argument(
         "-smooth",
-        metavar='',
+        metavar=Metavar.list,
+        type=list_type(',', float),
         help='Gaussian smoothing filter with specified standard deviations in mm for each axis (Example: 2,2,1) or '
              'single value for all axis (Example: 2).',
```
kousu, line 183: If this is `float` the examples should reflect that.

scripts/sct_maths.py
```diff
@@ -172,14 +177,15 @@ def get_parser():
...
         required=False)
     filtering.add_argument(
         '-laplacian',
-        nargs="+",
```
kousu, line 181: We can't use `nargs="+"` because that only works once in a CLI?

kousu (commented):
Keeping the cli consistent is a cranky job.

scripts/sct_maths.py
joshuacwnewton: The reason I switched here is because `-laplacian` used to be comma-delimited, but `nargs="+"` is space-delimited. I wanted to put things back to what they used to be so that they would stay consistent.

e.g. see this usage which is still buried in the API:

https://github.com/neuropoly/spinalcordtoolbox/blob/69287b0953fad48bbabc4722d74ccf484ebfbeee/spinalcordtoolbox/registration/register.py#L192-L198

In fact... now that I think about it, this call would crash with `nargs="+"` right now.

scripts/sct_maths.py
joshuacwnewton: I very much agree. This is a similar concern to what happened with Julien's comments earlier, i.e. https://github.com/neuropoly/spinalcordtoolbox/pull/2901#discussion_r493109814

scripts/sct_maths.py
joshuacwnewton: It's possible the other `nargs="+"` calls could be changed too (e.g. `-add`, `-mul`, etc.) to keep syntax consistent.

scripts/sct_maths.py
joshuacwnewton: I agree that this isn't ideal...

But, at that point, we're reimplementing built-in argparse functionality all over again (i.e. the combination of `nargs=2` and `type=int`) just to continue to preserve our custom delimiting. :stuck_out_tongue:

scripts/sct_maths.py
joshuacwnewton: I agree this is confusing. I think the error message later in the file is much clearer about what this means:

https://github.com/neuropoly/spinalcordtoolbox/blob/d443363b1028fe5fc653abd8c3b06b8af9913d9d/scripts/sct_maths.py#L311

I'll update the help description.

kousu:
So do we want to put this in as is or try to redo it with `nargs`?

kousu:
I think it could go in as-is.

joshuacwnewton:
> I think it could go in as-is.

I think I'm fine with keeping the `,` convention, so that sounds good to me too.

jcohenadad (approved):
awesome

PR #2903, opened 2020-09-22
pytest: download test data before test collection

### LINKED ISSUE #2902: pytest: unable to parametrize unit tests with data from sct_testing_data
In https://github.com/neuropoly/spinalcordtoolbox/pull/2887 I added parametrized unit tests that use both generated images and "real" images from disk. When `pytest` collects the unit tests, it is not able to resolve the images located in `sct_testing_data`, because the sample data hasn't yet been downloaded.

Error example: https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/728992346#L1039

### LINKED ISSUE #2907: Add unit test for checking testing data integrity
Related to https://github.com/neuropoly/spinalcordtoolbox/pull/2903#issuecomment-696977499

Running unit tests or functional tests should not change the input testing data. There should be a unit testing checking the integrity of the testing data after the unit tests have run.

This will be partially useful until all functional tests have been ported to pytest.

Drulex:
closes #2902
closes #2907
after #2909

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -11,6 +11,7 @@
 ###############################################################################
 
 import sys
+import os
 
 from spinalcordtoolbox.utils import sct_dir_local_path, sct_test_path
 sys.path.append(sct_dir_local_path('scripts'))
@@ -19,12 +20,8 @@ import pytest
 import sct_download_data as downloader
 
 
-@pytest.fixture(scope='session', autouse=True)
-def download_data(request):
-    # This is a hack because the capsys fixture can't be used with
-    # session scope at the moment.
-    # https://github.com/pytest-dev/pytest/issues/2704#issuecomment-603387680
-    capmanager = request.config.pluginmanager.getplugin("capturemanager")
-    with capmanager.global_and_fixture_disabled():
+def pytest_collectstart():
+    # TODO [AJ] check integrity eventually
+    if not os.path.exists(sct_test_path()):
         print('\nDownloading sct testing data.')
         downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
--- a/conftest.py
+++ b/conftest.py
@@ -65,7 +65,7 @@ def check_testing_data_integrity(files_checksums: Mapping[os.PathLike, str]):
             after.append(fname)
 
             if fname not in files_checksums:
-                logger.error(f"Discovered new file in sct_testing_data that didn't exist before: {(fname, chksum)}")
+                logger.warning(f"Discovered new file in sct_testing_data that didn't exist before: {(fname, chksum)}")
                 new.append((fname, chksum))
 
             elif files_checksums[fname] != chksum:
@@ -78,5 +78,5 @@ def check_testing_data_integrity(files_checksums: Mapping[os.PathLike, str]):
             missing.append((fname, chksum))
 
     assert not changed
-    assert not new
+    # assert not new
     assert not missing
--- a/conftest.py
+++ b/conftest.py
@@ -12,16 +12,72 @@
 
 import sys
 import os
+import logging
+
+import pytest
+from typing import Mapping
+from hashlib import md5
 
 from spinalcordtoolbox.utils import sct_dir_local_path, sct_test_path
 sys.path.append(sct_dir_local_path('scripts'))
 
-import pytest
 import sct_download_data as downloader
 
 
+logger = logging.getLogger(__name__)
+
+
 def pytest_collectstart():
-    # TODO [AJ] check integrity eventually
+    """ Download sct_testing_data prior to test collection. """
     if not os.path.exists(sct_test_path()):
         print('\nDownloading sct testing data.')
         downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
+
+
+@pytest.fixture(scope="session", autouse=True)
+def test_data_integrity(request):
+    files_checksums = dict()
+    for root, _, files in os.walk(sct_test_path()):
+        for f in files:
+            fname = os.path.join(root, f)
+            chksum = checksum(fname)
+            files_checksums[fname] = chksum
+
+    request.addfinalizer(lambda: check_testing_data_integrity(files_checksums))
+
+
+def checksum(fname: os.PathLike) -> str:
+    with open(fname, 'rb') as f:
+        data = f.read()
+    return md5(data).hexdigest()
+
+
+def check_testing_data_integrity(files_checksums: Mapping[os.PathLike, str]):
+    changed = []
+    new = []
+    missing = []
+
+    after = []
+
+    for root, _, files in os.walk(sct_test_path()):
+        for f in files:
+            fname = os.path.join(root, f)
+            chksum = checksum(fname)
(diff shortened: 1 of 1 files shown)


Drulex:
I removed the `download_data()` fixture and replaced it with a hook that is called prior to test collection to work around #2902 . I am not sure why the fixture was written with `capmanager`, @cfhammill are you aware if this is still necessary?

Also I don't re-download the test data on every test run, do you think it is reasonable to assume that if the `sct_testing_data` folder exists we don't need to re-download it? Ideally we should add some caching/integrity checking in the download function, but in the meantime I am assuming that people who really care about this (developers) can simply delete the folder if they want to test with "fresh" data.

jcohenadad:
> Also I don't re-download the test data on every test run, do you think it is reasonable to assume that if the sct_testing_data folder exists we don't need to re-download it? Ideally we should add some caching/integrity checking in the download function, but in the meantime I am assuming that people who really care about this (developers) can simply delete the folder if they want to test with "fresh" data.

we've had problems in the past, where testing data were corrupted across time, and testing would not pass unless we would re-download the testing data, but before realizing that, it took a lot of wasting debugging time. So, in case this happens again in the future, i'd rather force a fresh download, but maybe i'm being extra paranoid, and maybe our testing suite is cleaner now...

Drulex:
> we've had problems in the past, where testing data were corrupted across time, and testing would not pass unless we would re-download the testing data

This is definitely an issue. As a minimum I would propose to add a unit test to validate that testing data hasn't been changed by the unit tests.

Drulex:
I added the test for integrity checking. Right away I can see one failure (https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/729621964#L1229) where 2 files were added to the `sct_testing_data` folder by the unit tests.

For now I will leave the the download of testing data on each run until the downloader is updated to check check integrity properly.

conftest.py
```diff
@@ -11,20 +11,72 @@
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def test_data_integrity(request):
+    files_checksums = dict()
+    for root, _, files in os.walk(sct_test_path()):
+        for f in files:
+            fname = os.path.join(root, f)
+            chksum = checksum(fname)
+            files_checksums[fname] = chksum
```
joshuacwnewton, line 43: From https://github.com/neuropoly/spinalcordtoolbox/pull/2903#issuecomment-696974432

> we've had problems in the past, where testing data were corrupted across time, and testing would not pass unless we would re-download the testing data

Since the issue described here is about maintaining the pure state of `sct_testing_data`, would it be useful to provide these checksums with the dataset itself instead?

If we did that, we could do something like  _"Only redownload `sct_testing_data` if the files in the existing `sct_testing_data` directory don't match the checksum."_ This could possibly allow us to not have to redownload `sct_testing_data` each time. (Although, I suppose it only takes a few seconds each time, so perhaps the savings are not worth the effort.)

Drulex: Yes definitely. I think there are 2 improvements to be made here:
1. Make sure that tests don't change the testing data (this PR).
2. Improve the downloader to support data caching, integrity checking, etc (another issue, to be discussed/designed).

For now, downloading the data on each run + checking integrity at the end ensures that the tests ran with the correct input and that they didn't cause side-effects on the input data.

Once the downloader is improved, I would expect `download_test_data()` to internally check if data is cached on disk and verify the integrity against the official checksums probably included in some kind of manifest, something that is not yet done to my knowledge.

conftest.py
```diff
@@ -11,20 +11,72 @@
...
+
+    request.addfinalizer(lambda: check_testing_data_integrity(files_checksums))
+
+
+def checksum(fname: os.PathLike) -> str:
+    with open(fname, 'rb') as f:
+        data = f.read()
+    return md5(data).hexdigest()
```
joshuacwnewton, line 51: Just out of curiosity, for my own learning, why `md5.hexdigest`? (as opposed to other options provided by `hashlib`)

conftest.py
joshuacwnewton, line 43: I have one idea for a way to refactor this, but I'm not sure it's too critical.

Could this be its own function, e.g. `generate_checksums()` that returns `files_checksums`?

That way, `check_testing_data_integrity` could just compare `files_checksums_before` and `files_checksums_after` e.g. https://stackoverflow.com/a/18860653. Then the directory walking code wouldn't be repeated, too.

joshuacwnewton (commented):
LGTM, apart from one idea I had to refactor this in a simpler way.

conftest.py
Drulex: No particular reason. `md5` is a fast (128bits) and popular hashing function. It shouldn't be used for security applications anymore, but for what we're doing here it's ok. Honestly I didn't really evaluate other options.

joshuacwnewton: That makes sense to me! For something like this, I agree, "perfect is the enemy of the good", etc. :slightly_smiling_face:

conftest.py
Drulex: You're technically right. My instinct would be to leave it for now since it's a minor cosmetic change and revisit it later if there is a need to compute checksums somewhere else in the project (for the enhanced `downloader` perhaps?). We'll see if/when this happens.

You're welcome to refactor it though.

joshuacwnewton: That makes sense to me. :)

joshuacwnewton (dismissed):

cfhammill:
Hey @Drulex the capmanager code was used to transiently disable the output capture used in pytest so that the download progress bar was visible to the user but didn't end up in the pytest output. It is probably still necessary.

Drulex:
> Hey @Drulex the capmanager code was used to transiently disable the output capture used in pytest so that the download progress bar was visible to the user but didn't end up in the pytest output. It is probably still necessary.

It looks like the hook I am using (`pytest_sessionstart`) runs too early to be captured in the `pytest` output anyways so no change of behavior here.

But even if it did, why a special case for `sct_download` only? Why not use the `pytest` CLI to control the capture behaviour from `stdout/stderr/stdin`? (see `--capture`, `--show-capture` flags). 

That piece of code seem to suggest that what is outputted by `sct_download()` deserves special treatment, whereas it looks like in reality we don't care about anything captured from `stdout/stderr` at all.

cfhammill:
> It looks like the hook I am using (pytest_sessionstart) runs too early to be captured in the pytest output anyways so no change of behavior here.

cool

> But even if it did, why a special case for sct_download only? Why not use the pytest CLI to control the capture behaviour from stdout/stderr/stdin? (see --capture, --show-capture flags).

because the default capture behaviour is a better global default I think. `capsys` is supposed to allow controlling this behaviour more locally but as noted in the original comment it didn't work for this case. I think we generally do want to consume output, the progress bar for downloading the test data is potentially useful for the person running pytest, otherwise the pytest output gives progress and shouldn't be polluted with generic test output.

Drulex:
If no one else has comments can we approve this? Thanks.

jcohenadad (approved):

PR #2909, opened 2020-09-23
sct_compute_mtsat.py: more sane defaults

### LINKED ISSUE #2908: test_sct_compute_mtsat.py creates files in sct_testing_data
- It's best to avoid changing stuff in `sct_testing_data`: we want the testing data to remain pristine.
- Tests should (ideally) not leave artifacts in the filesystem; they should clean up after themselves.

https://github.com/neuropoly/spinalcordtoolbox/blob/432bbc681d8f7fdb27d007f0f782eaf0285582be/unit_testing/test_sct_compute_mtsat.py#L13-L28

This causes a new unit test that checks test data integrity to fail: https://github.com/neuropoly/spinalcordtoolbox/pull/2903

Drulex:
Closes #2908 

- When a help entry says `-omtsat <str>  Output file for MTsat. Default is mtsat.nii.gz` -> it is misleading to save the output file in the same folder as the input file (when `-omtsat` is not provided). Instead use the current working directory which better reflects the default.

- Unit test should not save in `sct_testing_data`. For lack of a better place use the current working directory instead.

### DIFF
--- a/scripts/sct_compute_mtsat.py
+++ b/scripts/sct_compute_mtsat.py
@@ -109,13 +109,13 @@ def get_parser(argv):
     optional.add_argument(
         "-omtsat",
         metavar=Metavar.str,
-        help="Output file for MTsat. Default is mtsat.nii.gz",
-        default=None)
+        help="Output file for MTsat",
+        default="mtsat.nii.gz")
     optional.add_argument(
         "-ot1map",
         metavar=Metavar.str,
-        help="Output file for T1map. Default is t1map.nii.gz",
-        default=None)
+        help="Output file for T1map",
+        default="t1map.nii.gz")
     optional.add_argument(
         "-v",
         help="Verbose: 0 = no verbosity, 1 = verbose (default).",
@@ -199,20 +199,11 @@ def main(argv):
                                          nii_b1map=nii_b1map)
 
     # Output MTsat and T1 maps
-    # by default, output in the same directory as the input images
     sct.printv('Generate output files...', verbose)
-    if args.omtsat is None:
-        fname_mtsat = os.path.join(os.path.dirname(nii_mt.absolutepath), "mtsat.nii.gz")
-    else:
-        fname_mtsat = args.omtsat
-    nii_mtsat.save(fname_mtsat)
-    if args.ot1map is None:
-        fname_t1map = os.path.join(os.path.dirname(nii_mt.absolutepath), "t1map.nii.gz")
-    else:
-        fname_t1map = args.ot1map
-    nii_t1map.save(fname_t1map)
+    nii_mtsat.save(args.omtsat)
+    nii_t1map.save(args.ot1map)
 
-    sct.display_viewer_syntax([fname_mtsat, fname_t1map],
+    sct.display_viewer_syntax([args.omtsat, args.ot1map],
                               colormaps=['gray', 'gray'],
                               minmax=['-10,10', '0, 3'],
                               opacities=['1', '1'],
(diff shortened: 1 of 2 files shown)


scripts/sct_compute_mtsat.py
```diff
@@ -199,20 +199,11 @@ def main(argv):
...
-    else:
-        fname_mtsat = args.omtsat
-    nii_mtsat.save(fname_mtsat)
-    if args.ot1map is None:
-        fname_t1map = os.path.join(os.path.dirname(nii_mt.absolutepath), "t1map.nii.gz")
-    else:
-        fname_t1map = args.ot1map
-    nii_t1map.save(fname_t1map)
+    nii_mtsat.save(args.omtsat)
+    nii_t1map.save(args.ot1map)
```
joshuacwnewton, line 204: A consideration: ~15 of our scripts are currently using the  `-ofolder` argument to accompish a similar thing, e.g.

https://github.com/neuropoly/spinalcordtoolbox/blob/432bbc681d8f7fdb27d007f0f782eaf0285582be/scripts/sct_fmri_moco.py#L93-L99

https://github.com/neuropoly/spinalcordtoolbox/blob/432bbc681d8f7fdb27d007f0f782eaf0285582be/scripts/sct_propseg.py#L199-L204

File output locations in CLI scripts could be a larger discussion in a different issue, though. (The locations are not always consistent, for example if you compare `sct_testing_data/` before/after running `sct_testing`.)

unit_testing/test_sct_compute_mtsat.py
```diff
@@ -9,20 +9,28 @@
...
     ]
 
 
 @pytest.mark.parametrize('input_params', INPUT_PARAMS)
 def test_with_json_sidecar(input_params):
     sct_compute_mtsat.main(input_params)
-    # Check if output file exists
-    assert os.path.isfile(sct_test_path('mt', 'mtsat.nii.gz'))
+    # Check if output files exist
+    for f in [out_mstat, out_t1map]:
```
joshuacwnewton, line 34: Nitpick: Since we're already using `pytest.mark.parameterize`, could these paths be accessed using `input_params['-omtsat']` and `input_params['-ot1map']`?

That way, we wouldn't need to create and access these variables:
```
out_mstat = "out_mtsat.nii.gz"
out_t1map = "out_t1map.nii.gz"
```

joshuacwnewton (commented):
Thanks so much for doing this! :)

unit_testing/test_sct_compute_mtsat.py
Drulex: I agree that it would be nice to parametrize, but `input_params` is a list at the moment. I guess it's possible to make it a dict and transform it prior to passing to the script, but I don't think its worth it.

What I like about explicitly defining the output files is that it makes the code more readable as opposed to not defining them and relying on the default values (don't need to guess what those are).

scripts/sct_compute_mtsat.py
Drulex: > if you compare sct_testing_data/ before/after running sct_testing, files are output in a few different places.

Yes I noticed this yesterday and created #2911. In my opinion:
- Unit tests should not leave artifacts on the filesystem. The output should be a PASS/FAIL + traceback, etc only.
- If a script generates a lot of output data, it's perfectly reasonable to have an `ofolder` parameter. If it only generates a file or two, specifying the output file paths is preferable.

unit_testing/test_sct_compute_mtsat.py
joshuacwnewton: Sorry, you're absolutely right. I misread! It functions so much like a dict that I read it as if it were one.

scripts/sct_compute_mtsat.py
joshuacwnewton: I think that's a very reasonable take. I'm going to update the Wiki with this information... I think the Testing page needs a bit of an update. :)

joshuacwnewton (approved):

Issue #2888, opened 2020-09-08
Investigate running batch_processing during CI


Drulex:
Objective: to run `batch_processing.sh` in the CI, preferably only on builds for the `master`branch in order to increase confidence in the software.

We would want:
- To run concurrently as much as possible to avoid increasing the time of the build
- Validate the results (other than exit code) of the script somehow.

TODO:
- [x] How to run? Logic in `.ci.sh` vs build stages?
- [x] When to run? `master` branch, `release` branch?
- [ ] How to validate results of batch processing?

Drulex:
Started to look at this. Initially I added a background job inside the script right after the installation and collected the results at the end:

<details><summary>View diff</summary>
<p>

``` patch
diff --git a/.ci.sh b/.ci.sh
index 883b2b5b..7fd1fa6a 100755
--- a/.ci.sh
+++ b/.ci.sh
@@ -14,6 +14,10 @@ ls -lA bin  # Make sure all binaries and aliases are there
 source python/etc/profile.d/conda.sh  # to be able to call conda
 conda activate venv_sct  # reactivate conda for the pip install below
 
+echo "*** Running batch_processing.sh in the background ***"
+./batch_processing.sh > batch_proc.log 2>&1 &
+bproc_pid=$!
+
 echo *** UNIT TESTS ***
 pytest
 
@@ -30,6 +34,20 @@ echo *** ANALYZE CODE ***
 pip install pylint
 bash -c 'PYTHONPATH="$PWD/scripts:$PWD" pylint -j3 --py3k --output-format=parseable --errors-only $(git ls-tree --name-only -r HEAD | sort | grep -E "(spinalcordtoolbox|scripts|testing).*\.py" | xargs); exit $(((($?&3))!=0))'
 
+echo "Waiting for batch_processing.sh to finish"
+wait $bproc_pid
+bproc_ret=$?
+
+echo "batch_processing.sh done. Results:"
+echo `tail -14 batch_proc.log`
+echo "Return code of batch_processing.sh: ${bproc_ret}"
+
+if [ $bproc_ret -ne 0 ]; then
+  echo "batch_processing.sh exited with non-zero! Full log:"
+  echo `cat batch_proc.log`
+  exit $bproc_ret
+fi
+
 #
 # echo *** BUILD DOCUMENTATION ***
 # pip install sphinx sphinxcontrib.programoutput sphinx_rtd_theme

```

</p>
</details>

[...]

PR #2910, opened 2020-09-23
Run batch_processing if building master

Drulex:
Partially addresses #2888.

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -30,8 +30,7 @@ echo *** ANALYZE CODE ***
 pip install pylint
 bash -c 'PYTHONPATH="$PWD/scripts:$PWD" pylint -j3 --py3k --output-format=parseable --errors-only $(git ls-tree --name-only -r HEAD | sort | grep -E "(spinalcordtoolbox|scripts|testing).*\.py" | xargs); exit $(((($?&3))!=0))'
 
-branch=`git symbolic-ref --short -q HEAD`
-if [[ "$branch" == "master" ]]; then
+if [[ "$BRANCH" == "master" ]]; then
   echo "*** Running batch_processing.sh ***"
   ./batch_processing.sh
 fi
--- a/.ci.sh
+++ b/.ci.sh
@@ -35,15 +35,3 @@ if [[ "$branch" == "master" ]]; then
   echo "*** Running batch_processing.sh ***"
   ./batch_processing.sh
 fi
-
-#
-# echo *** BUILD DOCUMENTATION ***
-# pip install sphinx sphinxcontrib.programoutput sphinx_rtd_theme
-# cd documentation/sphinx
-# make html
-# cd -
-
-# python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
-# cd ../spinalcordtoolbox_v*
-# yes | ./install_sct  # test installation of package
-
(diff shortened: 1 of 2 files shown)


Drulex:
I guess we'll need to merge to see make sure it happens. Assigning @kousu for sanity check.

Drulex:
Hmm looks like we can't ask `git` for the branch.. https://unhashable.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/

.ci.sh
```diff
@@ -30,14 +30,7 @@ echo *** ANALYZE CODE ***
...
-# pip install sphinx sphinxcontrib.programoutput sphinx_rtd_theme
-# cd documentation/sphinx
-# make html
-# cd -
-
-# python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
-# cd ../spinalcordtoolbox_v*
-# yes | ./install_sct  # test installation of package
-
+if [[ "$BRANCH" == "master" ]]; then
```
kousu, line 33: Maybe `$TRAVIS_BRANCH` to be more explicit? :shrug:

.ci.sh
```diff
@@ -30,14 +30,7 @@ echo *** ANALYZE CODE ***
...
+  echo "*** Running batch_processing.sh ***"
+  ./batch_processing.sh
+fi
```
kousu, line 36: I was hoping that ./ci.sh could be more agnostic than this.

In https://github.com/neuropoly/spinalcordtoolbox/issues/2674 I mused about adopting the pattern in https://docs.travis-ci.com/user/speeding-up-the-build/#parallelizing-your-builds-across-virtual-machines of

```
env:
  - SUITE=integration_tests
  - SUITE=unit_tests
  - SUITE=lint
  - SUITE=super_integration_tests # i.e. batch_processing.sh

...

script: "./travis.sh $SUITE"
```

Looking now I'm not confident I understand how `env` interacts with `matrix: include: ...`. We probably want batch_processing.sh to run on the same cases as `if: branch = release or type = cron`, but would want `SUITE=lint` to only run on one VM (say, Ubuntu). Is there a good way to build that kind of cross-product in Travis on top of what we already have in there?

kousu (approved):
Take a look at my thoughts, but this seems fine and if you want to merge as is go for it.

.ci.sh
Drulex: `TRAVIS_BRANCH` is actually something else. See (https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables). Relevant section:
image

So based on this logic, when testing PRs, `TRAVIS_BRANCH` == target branch == `master`. This is the reason for the little gymnastics in `.ci.sh` to figure out if we are in a PR or not and use the proper env var.

.ci.sh
Drulex: I agree, I also investigated parallelization, though I investigated the "distribute on multiple cores" approach (use of stages) vs the "distribute on multiple vm" approach that you are proposing (see https://github.com/neuropoly/spinalcordtoolbox/issues/2888#issuecomment-689585541).

Your approach would be more environment agnostic which I think is a good idea. I propose that we continue that investigation in a separate issue (created #2923 ).

PR #2919, opened 2020-09-26
test skimage compilation fix

kousu:
skimage finally merged https://github.com/scikit-image/scikit-image/pull/4920, so remove the hack of pre-installing numpy.

I'm not sure this has been published yet on their end.

I hacked travis.yml to test the case that was failing (OS X sierra); don't merge this without undoing that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,11 +104,11 @@ matrix:
       os: osx  # list of OSX env: https://docs.travis-ci.com/user/reference/osx/
       osx_image: xcode11.2
     - name: "OSX 10.13 (High Sierra)"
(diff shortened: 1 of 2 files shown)


kousu:
If this works, then https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/730467853 should take ~10 minutes longer than its peers (the time it takes to compile skimage) and there should be a compilation log in the build log, but the tests should till pass.

joshuacwnewton:
Ah, I believe we can close this now. This fix is no longer needed as we dropped support for `10.12` (and thus the numpy pre-install hack) in #2946.

PR #2924, opened 2020-09-28
Revert "Run batch_processing if building master"

Drulex:
Reverts neuropoly/spinalcordtoolbox#2910.

I appears that this breaks the build on certain platforms (see https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/730936670)

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -30,7 +30,14 @@ echo *** ANALYZE CODE ***
 pip install pylint
 bash -c 'PYTHONPATH="$PWD/scripts:$PWD" pylint -j3 --py3k --output-format=parseable --errors-only $(git ls-tree --name-only -r HEAD | sort | grep -E "(spinalcordtoolbox|scripts|testing).*\.py" | xargs); exit $(((($?&3))!=0))'
 
-if [[ "$BRANCH" == "master" ]]; then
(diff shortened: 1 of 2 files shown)


Drulex:
Premature guess: `$SCT_DIR` not defined for some obscure reason? (https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/730936673#L2210)

jcohenadad:
> Premature guess: `$SCT_DIR` not defined for some obscure reason? (https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/730936673#L2210)

indeed, SCT's batch install requires to start a new terminal or run: `source /Users/travis/.bashrc` (from [here)](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/730936673#L826)
How about we add this line of code in Travis config after SCT install?

Drulex:
> How about we add this line of code in Travis config after SCT install?

We would need to check which shell is used and what is the corresponding `.rc` file? Also I don't understand why some platforms built correctly while others didn't. 

Perhaps we can approve this revert and I'll investigate a clean way to do this.

jcohenadad (approved):

PR #2931, opened 2020-09-30
Pin onnxruntime version to fix "ImportError: cannot import name 'get_all_providers'" for macOS in CI

### LINKED ISSUE #2930: CI build failing for OSX due to ImportError: cannot import name 'get_all_providers'
https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/731373490#L922

joshuacwnewton:
### Related PRs/Issues

Fixes #2930.

### Description

SCT installs `onnxruntime` as it's an IVADOMED dependency. Recently, it's pypi version was bumped from 1.4.0 to 1.5.1. This change breaks our TravisCI job for macOS Mojave. The underlying reason is unclear, but in the meantime, this PR applies a temporary fix to restore the CI job.

This should be followed up on after conferring with the `onnxruntime` project devs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ Keras==2.1.5
 matplotlib
 nibabel
 numpy
+onnxruntime==1.4.0
 pandas
 psutil
 pyqt5==5.11.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,6 @@ nibabel
 numpy
 # Note: If a newer version (1.5.1) is installed, Mojave CI jobs will fail
 # This is temporary until a better upstream fix can be determined. (See #2930)
-onnxruntime; sys_platform != "darwin"
 onnxruntime==1.4.0; sys_platform == "darwin"
 pandas
 psutil
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,10 @@ Keras==2.1.5
 matplotlib
 nibabel
 numpy
-onnxruntime==1.4.0
+# Note: If a newer version (1.5.1) is installed, Mojave CI jobs will fail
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -7,6 +7,7 @@ Keras==2.1.5
 matplotlib
 nibabel
 numpy
+onnxruntime==1.4.0
```
Drulex, line 10: Not a huge fan of adding this, but at the same time I don't know what else is better than having the issue fixed upstream.

Anyways, is it worth adding the platform check too? `sys_platform == "darwin"`?

lrouhier: Maybe something like what is done for torch ? (line 28)

joshuacwnewton: > Not a huge fan of adding this, but at the same time I don't know what else is better

I agree. I'll add a comment making it clear that this should be a temporary measure, rather than a long-term fix, and to refer back to the issue to check if it's safe to revert.

> having the issue fixed upstream.

By upstream here, do you mean pinning it in IVADOMED instead? Or do you mean even further upstream (i.e. opening an issue with `onnxruntime` and troubleshooting with them)? 

I'm curious about the opening an issue, because the error we're getting has come up on `onnxruntime` issues before, just in different contexts. So perhaps they might have some insight as to why this is happening in the first place.

> Maybe something like what is done for torch ? (line 28)

I think that's a good idea. :)

Drulex (approved):

Drulex:
@joshuacwnewton I went ahead and merged since the build failure blocks other PR's from being mergeable, I hope you don't mind. Thanks for your investigation!

joshuacwnewton:
I agree with moving quickly here. Thanks so much for being on this! And, thanks for catching the redundant line, you're absolutely correct. :)

PR #2932, opened 2020-10-01
Replace onnxruntime==1.4.0 version pinning with libomp install for macOS CI jobs

joshuacwnewton:
Potentially a better fix for #2930. See https://github.com/microsoft/onnxruntime/issues/5344#issuecomment-702124166.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,7 +106,7 @@ matrix:
       addons:
         homebrew:
           packages:
-            - libomp
+            - libomp  # Needed for onnxruntime Python package, see issue #2930
     - name: "OSX 10.13 (High Sierra)"
       if: branch = release or type = cron
       os: osx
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
Libomp installed here: https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/731934950#L93

No error! Huzzah.

jcohenadad:
> No error! Huzzah.

hopefully it will still the case for our extensive testing (ie across more platforms) when merging to master

Drulex:
> No error! Huzzah.

Nice! We should add the fix for the other 2 osx builds as well, they all seemed to fail with the same issue (https://travis-ci.org/github/neuropoly/spinalcordtoolbox/builds/731682995)

Drulex (approved):

PR #2940, opened 2020-10-07
sct_image: -copy-header should use -o as output file

### LINKED ISSUE #2505: sct_image -copy-header does not consider flag -o
This is confusing behaviour and should be clarified. Maybe force new output if -o is used.

Drulex:
Closes #2505

### DIFF
--- a/scripts/sct_image.py
+++ b/scripts/sct_image.py
@@ -95,7 +95,9 @@ def get_parser():
     header = parser.add_argument_group('HEADER OPERATIONS')
     header.add_argument(
         '-copy-header',
-        help='Copy the header of the source image (specified in -i) to the destination image (specified in -o). ',
+        metavar=Metavar.file,
+        help='Copy the header of the source image (specified in -i) to the destination image (specified here) '
+             'and save it into a new image (specified in -o)',
         required = False)
 
     orientation = parser.add_argument_group('ORIENTATION OPERATIONS')
@@ -196,7 +198,7 @@ def main(args=None):
         if fname_out is None:
             raise ValueError("Need to specify output image with -o!")
         im_in = Image(fname_in[0])
-        im_dest = Image(fname_out)
+        im_dest = Image(arguments.copy_header)
         im_dest_new = im_in.copy()
         im_dest_new.data = im_dest.data.copy()
         # im_dest.header = im_in.header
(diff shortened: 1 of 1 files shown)


Drulex:
Oups I broke the interface without changing the tests. On it.

scripts/sct_image.py
```diff
@@ -195,14 +193,16 @@ def main(args=None):
...
         im_in = Image(fname_in[0])
-        im_dest = Image(arguments.copy_header)
+        im_dest = Image(fname_out)
         im_dest_new = im_in.copy()
         im_dest_new.data = im_dest.data.copy()
         # im_dest.header = im_in.header
         im_dest_new.absolutepath = im_dest.absolutepath
         im_out = [im_dest_new]
-        fname_out = arguments.copy_header
+        fname_out = fname_out
```
jcohenadad, line 205: i'm not a big fan of this approach, because this will overwrite the destination image (the one you targeted as `-o`). In a typical scenario, we have src.nii, dest.nii, and we would like to create dest_with_src_header.nii (not always overwrite dest.nii).

Drulex: Woops this is a mistake from when I amended my previous commit. This line shouldn't exist. Fixed in bf2b00a

Drulex: So basically the behavior is that if the user wants to use `copy_header`, he must provide the output image (like before), but via `-o`.

Drulex: > So basically the behavior is that if the user wants to use `-copy-header`, he must provide the output image (like before), but via `-o` instead of via a value to `-copy-header`.

jcohenadad: i have a small problem with that approach. In some instances, users wouldn't like to have their images being overwritten: this is what would happen here (ie: the file specified by `-o` would be overwritten). 

Alternatively, it would be great to give the option to duplicate the destination file and output it as another file. So if we go with that approach, `-copy-header` would keep the previous behaviour (ie: point to the destination file), and if `-o` exists, the destination file will be duplicated, named as `-o` and the header will be changed in that duplicated file. Does that make sense?

Drulex: Ah ok I get it. Sorry seems I misunderstood what was required. Will adjust.

Drulex: fixed 4cb0708

jcohenadad: I was thinking of using -copy-header if -o is not specified, but your enforcement approach (ie -o mandatory) is probably the safest

jcohenadad (approved):

PR #2949, opened 2020-10-14
sct_utils/sct.utils further refactor

### LINKED ISSUE #1658: sct_utils create_folder and exception handling/throwing (built-in exceptions)

### LINKED ISSUE #2627: Can't import spinalcordtoolbox.reports.qc outside of the scripts directory
On current master 4878c3b7 spinalcordtoolbox.reports.qc can't be imported from outside the scripts directory:

```python
import spinalcordtoolbox.reports.qc
```

```
  File "/home/chammill/Documents/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 29, in <module>
    import sct_utils as sct
ModuleNotFoundError: No module named 'sct_utils'
```

This works from in the scripts directory because `sct_utils.py` is in there.

### LINKED ISSUE #2898: Refactor functions from scripts/sct_utils.py into spinalcordtoolbox/utils.py
### Description

> **Note**: Continuation of discussion from https://github.com/neuropoly/spinalcordtoolbox/issues/2627#issuecomment-596239336 and https://github.com/neuropoly/spinalcordtoolbox/issues/2890#issuecomment-692213618. 
> I believe this is already an ongoing effort -- this issue simply provides a centralized place for design discussion.

Currently we have two places we put assorted utility functions: `scripts/sct_utils.py` and `spinalcordtoolbox/utils.py`. In many cases, this functionality overlaps in theme (for example, file/directory-related functions are present in both). 

The `scripts/` directory is meant for CLI scripts, but `sct_utils.py` isn't actually a script. So, `spinalcordtoolbox/utils.py` seems like the more appropriate place for these functions. This would also match the organization encouraged by the ongoing Python API refactor, and potentially prevent issues like #2627.

### Alternatives

Some food for thought: It's possible that we should avoid having any sort of `utils.py` file at all. Instead, it may be better to be more specific about the role of each of these functions and group them thematically. See also this opinionated article: [Stop Naming Your Python Modules Utils](https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/) and its comments section.

__edit__: hijacked description by @Drulex for visibility
### Objectives
- [x]  Decouple logic in `sct_utils.py` into proper layers:
   - [x] API code should __not__ import `sct_utils.py`, instead it should import from other API modules. This avoids circular dependencies and other hidden problems such as #2922. 
   - [x] `sct_utils.py`, if it still exists, should only contain high-level utilities for other scripts (not for API modules) and only if those utilities don't belong in other scripts.
- [x] Remove unused/dead code
- [x] Cleanup imports as per PEP8
- [x] Utilities that belong in other modules should be moved from `sct.utils` to the appropriate module.
- [x] `spinalcordtoolbox.utils` can be further modularized if necessary (e.g `sct.utils.fs`, `sct.utils.shell`, `sct.utils.sys`, etc).

### LINKED ISSUE #2922: SpinalCordStraightener depends on sct_utils without importing it
`SpinalCordStraightener` cannot be imported without first loading `sct_utils.py`

```python
>>> from spinalcordtoolbox.straightening import SpinalCordStraightener
  File "/home/chammill/documents/spinalcordtoolbox/unit_testing/test_straightening.py", line 50, in <module>
  File "/home/chammill/documents/spinalcordtoolbox/spinalcordtoolbox/straightening.py", line 22, in <module>
    import sct_utils as sct
ModuleNotFoundError: No module named 'sct_utils'
```

Drulex:
Closes #2898 
Closes #2627 
Closes #2922
Closes #1658

Another pass at refactoring code in `sct_utils.py`.

- Refactor functions that belong in other modules
- Code that is shared between scripts and API will move to the API (for the moment at least)
- Remove imports of `sct_utils.py`

Since I touched practically every file I took the opportunity to:
- Cleanup imports order as per PEP8
- Remove unused imports
- PEP8 format

```
~~~
Version:         git-aj/2898-refactor_sct_utils_apify_all_the_things-8f292395939c498ce1624bedc8fbdfa1ce808d3d
Ran on:          Linux drupad-p1 5.4.64-1-MANJARO
Duration:        0hrs 14min 17sec
---
mt/MTR(WM):      54.42128965730348
~~~
```

Still missing (for another PR):
- Update docs
- Update unit tests
- Further refactor of some utils functions that do nasty stuff (`chdir()`, `printv()` and friends)


Drulex:
I don't expect anyone to realistically review all of this. Given the amount of changes I prefer to implement the unit tests and documentation update in different PRs and merge this soon to avoid conflicts.

jcohenadad (approved):
i trust you 😅

Issue #2957, opened 2020-10-18
Test data downloaded too many times in parallel pytest


cfhammill:
### Description

Parallel pytest downloads tests numprocesses + 1 times, instead of once.

### Steps to Reproduce

I set numprocesses to 1 in setup.cfg to avoid swamping my machine then run

```
source python/etc/profile.d/conda.sh
conda activate venv_sct
pytest
```

and observe

```
Status: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6.82M/6.82M [00:02<00:00, 3.03MB/s]
Done!

=========================================================================================================== test session starts ===========================================================================================================
platform linux -- Python 3.6.12, pytest-6.1.1, py-1.9.0, pluggy-0.13.1 -- /home/chammill/documents/spinalcordtoolbox/python/envs/venv_sct/bin/python
cachedir: .pytest_cache
rootdir: /home/chammill/documents/spinalcordtoolbox, configfile: setup.cfg
plugins: console-scripts-1.0.0, cov-2.10.1, forked-1.3.0, xdist-2.1.0
[gw0] linux Python 3.6.12 cwd: /home/chammill/documents/spinalcordtoolbox
[gw0] Python 3.6.12 |Anaconda, Inc.| (default, Sep  8 2020, 23:10:56)  -- [GCC 7.3.0]
Status: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6.82M/6.82M [00:02<00:00, 3.05MB/s]
```

The two status bars for 6.82M downloads are the 2 downloads. The number of downloads increases with the number of workers. This could be fixed by either getting download caching working or by fiddling with the pytest fixtures.

A duplicate download can also be seen in the most recent travis run for 18.04 (although interestingly it didn't seem to be there for Arch linux) https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/736864114#L1040

Drulex:
Yes this is a side-effect of using `x-dist` for parallel test distribution, each worker has its own session scope thus the initial download (session hook) is run multiple times. It appears that it is also run one additional time prior to distribution, during the initial test collection step.

[...]

PR #2958, opened 2020-10-19
conftest.py: don't download test data if user specifies --no-sct-download

Drulex:
Relates to #2957.

- This is a convenience option for developers, not a fix.
- Default behavior is to download test data no matter what. If `--no-sct-download` is provided the download step is skipped.


Drulex:
Closing since it doesn't seem to be useful to anyone.

PR #2679, opened 2020-04-29
Add new DL-based method (Countception) for C2-C3 disc detection and template matching 

jcohenadad:
This PR aims at improving the sct_label_vertebrae function with deep learning detection using a fully convolutional redundant counting (Countception) model implemented in Pytorch.  It also aimed at adding a module to label the posterior tip of each disc. 

Changes include:
- Download models for inference (added in  `spinalcordtoolbox/deepseg/models`
- Slight modification to `deepseg/core.py` API to avoid output binarization 
- Modification of `spinalcordtoolbox/vertebrae/core.py`: The template matching is now done between 'PAM50_label_disc_posterior' convolved with a gaussian and the heatmap output from the model.
- Changes of correlation metrics in the sliding windows: Mutual_information(data, template) has been replaced by np.correlate(data,template)/np.correlate(template,template)
- Changement of the size of the sliding windows (needed due to template change) 
- Addition of 'PAM50_label_disc_posterior_small' to the sct_testing_data  for integrity testing
- Creation of a `label_disc_posterior`: output an image with disc labels at the detected posterior tip of each disc. 
- Improves robustness of C2-C3 disc detection, fixes #2237 

To evaluate the performance of this new approach, we have designed a [framework](https://github.com/sct-pipeline/vertebral-labeling-validation).

TO DO: 
- Decide between HOG-SVM and Deep learning for C2/C3 detection (Other PR outside of the scope of this one)

PR #2962, opened 2020-10-22
Use logger in API code, not printv.

### LINKED ISSUE #2809: Use proper logging instead of printing to terminal
While working on #2547 , I noticed that a lot of what should be low-level code is printing to the terminal. I then looked into already modularization code to notice that API code is importing `sct_utils` (that lives in `scripts/`)  to use `printv()`. This is bad because:

- Layering violation: low-level api  code should not be aware/depend on higher-level application code. A developer should be able to take the API and develop their own application on top of it. Also it can lead to circular dependencies.
- API code should not print to the terminal: for the same reason that you don't want an imported package/library to unexpectedly start printing stuff to the terminal. 

Alternatives are:
- Use the `logging` library. This is more powerful, flexible and can be called from anywhere. We could create a `console` handler with a formatting identical to the `printv()` function and redirect all messages of a certain level (e.g. `INFO`) to the terminal.
- Inject a logging dependency into low-level code. This dependency would take care of the printing/logging and is a more flexible way for the API consumer to control how stuff is logged. For example, a first (side-effect free) step we could pass the `printv()` function to all the low-level API functions that need it, thus removing the `import sct_utils` dependency.

Drulex:
API should use logging, not `printv`.

At the same time cleanup dead code and imports here and there.

Closes #2809

### DIFF
--- a/spinalcordtoolbox/vertebrae/core.py
+++ b/spinalcordtoolbox/vertebrae/core.py
@@ -5,16 +5,18 @@
 # TODO: remove i/o as much as possible
 
 import os
+import logging
 
 import numpy as np
 import scipy.ndimage.measurements
 from scipy.ndimage.filters import gaussian_filter
+from scipy.signal import gaussian
 
 from spinalcordtoolbox.image import Image, add_suffix
 from spinalcordtoolbox.metadata import get_file_label
 from spinalcordtoolbox.math import dilate, mutual_information
-from spinalcordtoolbox.utils import printv
 
+logger = logging.getLogger(__name__)
 
 def label_vert(fname_seg, fname_label, verbose=1):
     """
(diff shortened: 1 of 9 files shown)


spinalcordtoolbox/vertebrae/core.py
```diff
@@ -5,16 +5,18 @@
 # TODO: remove i/o as much as possible
```
jcohenadad, line 5: heads up @lrouhier : this will create conflicts with #2679

jcohenadad (approved):
I haven't done a thorough review, but i'm planning to test different pipelines (batch_processing, spine-generic), before creating v5.0.0. So if something breaks we will find it (hopefully).

joshuacwnewton:
Out of curiosity, after these changes, what does the future of `printv` look like long-term? i.e. is the end-goal to use `logging` everywhere?

I ask because `printv` is currently a bit of a headache for #2676 and #2897. If `printv` was no longer used, then it would make updates to verbosity much easier.

PR #2969, opened 2020-10-23
Remove deepseg postprocessing

andreanne-lemay:
Postprocessing is dealt by ivadomed, hence is removed from SCT to avoid code duplication.
Changes include:
- New postprocessing option: remove small objects
- Remove postprocessing code from SCT
- Change default params of '-fill-holes' and '-largest' (now set to 0 by default)

Before merging, a release of ivadomed will be required (sister PR: https://github.com/ivadomed/ivadomed/pull/486)

### DIFF
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -95,6 +95,11 @@ def get_parser():
         help="Fill small holes in the segmentation.",
         choices=(0, 1),
         default=deepseg.core.DEFAULTS['fill_holes'])
+    misc.add_argument(
+        "-remove-small",
+        type=str,
+        help="Minimal object size to keep with unit (mm or vox). Example: 3mm, 5vox. (default: 0mm)",
+        default=deepseg.core.DEFAULTS['remove_small'])
 
     misc = parser.add_argument_group('\nMISC')
     misc.add_argument(
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -98,7 +98,7 @@ def get_parser():
     misc.add_argument(
         "-remove-small",
         type=str,
-        help="Minimal object size to keep with unit (mm or vox). Example: 3mm, 5vox. (default: 0mm)",
+        help="Minimal object size to keep with unit (mm3 or vox). Example: 1mm3, 5vox. (default: 0vox)",
         default=deepseg.core.DEFAULTS['remove_small'])
 
     misc = parser.add_argument_group('\nMISC')
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -170,7 +170,7 @@ def main():
                 parser.error("The input model is invalid: {}".format(path_model))
 
         # Call segment_nifti
-        fname_seg = deepseg.core.segment_nifti(args.i, path_model, fname_prior, args)
+        fname_seg = deepseg.core.segment_nifti(args.i, path_model, fname_prior, vars(args))
         # Use the result of the current model as additional input of the next model
         fname_prior = fname_seg
 
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -117,11 +117,11 @@ def main():
     args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
 
     # Deal with model
-    if args.list_models is not None:
+    if args.list_models:
         deepseg.models.display_list_models()
 
     # Deal with task
-    if args.list_tasks is not None:
+    if args.list_tasks:
         deepseg.models.display_list_tasks()
 
     if args.install_model is not None:
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -16,6 +16,8 @@ import os
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
+import spinalcordtoolbox.deepseg.models
+import spinalcordtoolbox.deepseg.core
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
 from spinalcordtoolbox.utils.sys import init_sct, printv
 
@@ -30,7 +32,6 @@ def get_parser():
     input_output = parser.add_argument_group("\nINPUT/OUTPUT")
     input_output.add_argument(
         "-i",
-        required=True,
         help="Image to segment.",
         metavar=Metavar.file)
     input_output.add_argument(
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -16,8 +16,6 @@ import os
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
-import spinalcordtoolbox.deepseg.models
-import spinalcordtoolbox.deepseg.core
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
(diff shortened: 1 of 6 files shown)


spinalcordtoolbox/deepseg/core.py
```diff
@@ -22,71 +22,12 @@
...
-    if options['thr']:
-        nii_seg = threshold(nii_seg, options['thr'])
-    if options['largest']:
-        nii_seg = keep_largest_objects(nii_seg, options['largest'])
-    if options['fill_holes']:
-        nii_seg = fill_holes(nii_seg)
-    return nii_seg
-
-
 def segment_nifti(fname_image, folder_model, fname_prior=None, param=None):
```
jcohenadad, line 31: given that there is almost nothing left in this `core.py` file, and that the only thing `segment_nifti()` does is call ivadomed, maybe we can just get rid of `core.py` and move the ivadomed call to the CLI? @Drulex @joshuacwnewton

joshuacwnewton: I very much agree that there's not much reason in having this function at all, especially since the file I/O should be handled at the CLI script level anyway. 

We would need to replace `test_segment_nifti()` in `unit_testing/test_deepseg.py` as well, I believe.

jcohenadad: @andreanne-lemay: if you want we can take care of this part (since we are more familiar with SCT's codebase). @joshuacwnewton do you mind helping with this?

joshuacwnewton: Certainly! I would love to. :slightly_smiling_face:

scripts/sct_deepseg.py
```diff
@@ -94,6 +95,11 @@ def get_parser():
         help="Fill small holes in the segmentation.",
         choices=(0, 1),
         default=deepseg.core.DEFAULTS['fill_holes'])
+    misc.add_argument(
+        "-remove-small",
+        type=str,
+        help="Minimal object size to keep with unit (mm3 or vox). Example: 1mm3, 5vox.",
```
charleygros, line 101: Great! Parsed [here](https://github.com/ivadomed/ivadomed/pull/486/files).

spinalcordtoolbox/deepseg/core.py
```diff
@@ -22,71 +22,12 @@
 # Default values if not asked during CLI call and if not present in json metadata.
 DEFAULTS = {
     'thr': 0.9,
-    'largest': 1,
-    'fill_holes': 1,
+    'largest': 0,
+    'fill_holes': 0,
+    'remove_small': '0vox'
```
charleygros, line 27: I was wondering if we should resort to  `None` instead of 0 as default values. Both ways work with this [sister PR](https://github.com/ivadomed/ivadomed/pull/486/files).

charleygros (approved):

spinalcordtoolbox/deepseg/core.py
andreanne-lemay: None triggers an error during parsing: `TypeError: argument of type 'NoneType' is not iterable`, so I left '0vox' since it gives the same result :)

I also included 'fname_prior' in options here: 9c936c3 (change made in sister PR)

andreanne-lemay:
The test `test_segment_nifti` is currently failing because the ground truth file and the output file from segment_nifti are not the same (this assertion is failing:[ assert np.all(nibabel.load(output).get_fdata() == nibabel.load(params['fname_seg_manual']).get_fdata())](https://github.com/neuropoly/spinalcordtoolbox/blob/master/unit_testing/test_deepseg.py#L66)).

Here is a visual comparison of the two files (red: ground truth, blue: output from segment_nifti). 
image

The issue seems to come from different thresholds used to binarize the segmentations between the "ground truth" and the model's predictions. By default, the threshold for the prediction is set to 0.9. The ground truth file is generated prior to the test, @Drulex I was wondering how this GT was generated?

jcohenadad:
> The issue seems to come from different thresholds used to binarize the segmentations between the "ground truth" and the model's predictions. By default, the threshold for the prediction is set to 0.9. The ground truth file is generated prior to the test, @Drulex I was wondering how this GT was generated?

I generated it by running `sct_deepseg` with the default values at that time, and saving it under `sct_testing_data/t2s/t2s_seg-deepseg.nii.gz`, so it makes sense that if the prediction changes, the test fails. If we are confident these new results are *better*, then a solution would be to update the GT on the test data.

andreanne-lemay:
> generated it by running `sct_deepseg` with the default values at that time, and saving it under `sct_testing_data/t2s/t2s_seg-deepseg.nii.gz`, so it makes sense that if the prediction changes, the test fails. If we are confident these new results are _better_, then a solution would be to update the GT on the test data.

I am not sure why the output would be different, I'll investigate.

scripts/sct_deepseg.py
```diff
@@ -16,6 +16,8 @@
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
+import spinalcordtoolbox.deepseg.models
+import spinalcordtoolbox.deepseg.core
```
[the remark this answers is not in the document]
andreanne-lemay: On my station, I had import issues without these lines which is why I re-added them. Do you have an idea why so?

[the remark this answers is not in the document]
Drulex: Recently? This was merged 2 days ago so if if you rebased off master since Monday it should run fine without those lines.

andreanne-lemay:
After investigation, the difference in the output file comes from the fact that binarization used to be done right after the predictions, but with the current changes, the binarization is done as a postprocessing step. On the "GT" image, the threshold used was 0.5 which is indicated in the configuration file of the model. When changing the threshold to 0.5 instead of 0.9, we get this:
image

There are still a few voxels of difference. ivadomed changed a lot since 2.0.2 (now at 2.4.0), the small difference must come from one of the changes made from 2.0.2 to 2.4.0. I think it would be safe to change the output in sct_testing_data.

jcohenadad:
> I think it would be safe to change the output in sct_testing_data.

perfect! do you want to submit a PR on https://github.com/sct-data/sct_testing_data?

scripts/sct_deepseg.py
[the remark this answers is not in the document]
andreanne-lemay: After merging master into my branch with the new code, everything works perfectly :)! 

Updated code: b495abe

andreanne-lemay:
Do you have a specific procedure to release data from [sct_testing_data](https://github.com/sct-data/sct_testing_data)? I see here the testing data is also located on OSF (https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_download_data.py#L79).
1- Release on sct_testing_data named rYYYYMMDD
2-Upload zip on OSF

Is there anything else? @jcohenadad

jcohenadad:
> Do you have a specific procedure to release data from [sct_testing_data](https://github.com/sct-data/sct_testing_data)? I see here the testing data is also located on OSF (https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_download_data.py#L79).
> 1- Release on sct_testing_data named rYYYYMMDD
> 2-Upload zip on OSF
> 
> Is there anything else? @jcohenadad

@andreanne-lemay you can replace the URLs with these: 
~~~
["https://github.com/sct-data/sct_testing_data/releases/download/r20201030/sct_testing_data-r20201030.zip", "https://osf.io/download/5f9c271187b7df04593b03e0/"]
~~~

PR #2975, opened 2020-10-27
Refactor sct_maths callers to remove subprocess and use API

### LINKED ISSUE #2878: Refactor callers of sct_math to use spinalcordtoolbox.math instead
See #2876 and #2877 for context.

Drulex:
Continuation of #2876 
After #2966 
Closes #2878 

Remove subprocess calls of `sct_maths` and use `spinalcordtoolbox.math` API.

~~~
Version:         git-aj/2878-refactor_sct_math_callers-9ed9a41ed88c9d04ae3c56e5e03a64301fdb6feb*
Ran on:          Linux gentoo-p53 5.10.5-gentoo-x86_64
Duration:        0hrs 11min 42sec
---
mt/MTR(WM):      54.38001546261108
t2s/CSA_GM:      12.487834828856176
t2s/CSA_WM:      64.93830702088249
dmri/FA(CST_r):  0.7756265311035226
dmri/FA(CST_l):  0.7800112266489608
~~~

### DIFF
--- a/spinalcordtoolbox/registration/register.py
+++ b/spinalcordtoolbox/registration/register.py
@@ -205,16 +205,19 @@ def register_step_ants_registration(src, dest, step, masking, ants_registration_
 
         src_img = image.Image(src)
         src_out = src_img.copy()
+
+        src = image.add_suffix(src, '_laplacian')
+        dest = image.add_suffix(dest, '_laplacian')
+
+        sigmas = [sigmas[i] / src_img.dim[i + 4] for i in range(3)]
+
         src_out.data = laplacian(src_out.data, sigmas)
-        src_out.save(path=image.add_suffix(src, '_laplacian'))
+        src_out.save(path=src)
 
         dest_img = image.Image(dest)
         dest_out = dest_img.copy()
         dest_out.data = laplacian(dest_out.data, sigmas)
-        dest_out.save(path=image.add_suffix(dest, '_laplacian'))
-
-        src = image.add_suffix(src, '_laplacian')
-        dest = image.add_suffix(dest, '_laplacian')
+        dest_out.save(path=dest)
 
     # Estimate transformation
     logger.info(f"\nEstimate transformation")
--- a/spinalcordtoolbox/registration/register.py
+++ b/spinalcordtoolbox/registration/register.py
@@ -21,6 +21,7 @@ from sklearn.decomposition import PCA
 from scipy.io import loadmat
 
 import spinalcordtoolbox.image as image
+from spinalcordtoolbox.math import laplacian
 from spinalcordtoolbox.registration.landmarks import register_landmarks
 from spinalcordtoolbox.utils import sct_progress_bar, copy_helper, run_proc, tmp_create
 
@@ -199,10 +200,19 @@ def register_step_ants_registration(src, dest, step, masking, ants_registration_
     # apply Laplacian filter
     if not step.laplacian == '0':
         logger.info(f"\nApply Laplacian filter")
-        run_proc(['sct_maths', '-i', src, '-laplacian', step.laplacian + ','
-                 + step.laplacian + ',0', '-o', image.add_suffix(src, '_laplacian')])
-        run_proc(['sct_maths', '-i', dest, '-laplacian', step.laplacian + ','
-                 + step.laplacian + ',0', '-o', image.add_suffix(dest, '_laplacian')])
+
+        sigmas = [step.laplacian, step.laplacian, 0]
+
+        src_img = image.Image(src)
+        src_out = src_img.copy()
+        src_out.data = laplacian(src_out.data, sigmas)
+        src_out.save(path=image.add_suffix(src, '_laplacian'))
+
+        dest_img = image.Image(dest)
+        dest_out = dest_img.copy()
+        dest_out.data = laplacian(dest_out.data, sigmas)
+        dest_out.save(path=image.add_suffix(dest, '_laplacian'))
+
         src = image.add_suffix(src, '_laplacian')
         dest = image.add_suffix(dest, '_laplacian')
 
--- a/spinalcordtoolbox/scripts/sct_analyze_texture.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_texture.py
@@ -201,7 +201,12 @@ class ExtractGLCM:
             # Average across angles and save it as wrk_folder/fnameIn_feature_distance_mean.extension
             fname_out = im_m + str(self.param_glcm.distance) + '_mean' + extension
             run_proc('sct_image -i ' + ' '.join(im2mean_lst) + ' -concat t -o ' + fname_out)
-            run_proc('sct_maths -i ' + fname_out + ' -mean t -o ' + fname_out)
+
+            img = Image(fname_out)
+            dim_idx = 3 # 3 dimensions + time
+            img.data = np.mean(img.data, dim_idx)
+            img.save()
+
             self.fname_metric_lst[im_m + str(self.param_glcm.distance) + '_mean'] = fname_out
 
     def extract_slices(self):
--- a/spinalcordtoolbox/scripts/sct_analyze_texture.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_texture.py
@@ -15,7 +15,7 @@ import itertools
 import numpy as np
 from skimage.feature import greycomatrix, greycoprops
 
-from spinalcordtoolbox.image import Image, add_suffix, zeros_like
+from spinalcordtoolbox.image import Image, add_suffix, zeros_like, concat_data
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, ActionCreateFolder
 from spinalcordtoolbox.utils.sys import init_sct, printv, sct_progress_bar, run_proc, set_global_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, extract_fname, copy, rmtree
@@ -200,10 +200,11 @@ class ExtractGLCM:
 
             # Average across angles and save it as wrk_folder/fnameIn_feature_distance_mean.extension
             fname_out = im_m + str(self.param_glcm.distance) + '_mean' + extension
-            run_proc('sct_image -i ' + ' '.join(im2mean_lst) + ' -concat t -o ' + fname_out)
 
+            dim_idx = 3 # img is [x, y, z, angle] so specify 4th dimension (angle)
+
+            img = concat_data(im2mean_lst, dim_idx).save(fname_out)
             img = Image(fname_out)
-            dim_idx = 3 # 3 dimensions + time
             img.data = np.mean(img.data, dim_idx)
             img.save()
 
--- a/spinalcordtoolbox/scripts/sct_analyze_texture.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_texture.py
@@ -219,18 +219,6 @@ class ExtractGLCM:
         # extract axial slices in self.dct_im_seg
         self.dct_im_seg['im'], self.dct_im_seg['seg'] = [im.data[:, :, z] for z in range(im.dim[2])], [seg.data[:, :, z] for z in range(im.dim[2])]
 
-    # def init_metric_im(self):
-    #     # open image and re-orient it to RPI if needed
-    #     im_tmp = Image(self.param.fname_im)
-    #     if self.orientation_im != self.orientation_extraction:
-    #         im_tmp = msct_image.change_orientation(im_tmp, self.orientation_extraction)
-
-    #     # create Image objects with zeros values for each output image needed
-    #     for m in self.metric_lst:
-    #         im_2save = msct_image.zeros_like(im_tmp, dtype=np.float64)
-    #         fname_out = add_suffix(''.join(extract_fname(self.param.fname_im)[1:]), '_' + m)
-    #         im_2save.save(fname_out)
-    #         self.fname_metric_lst[m] = fname_out
 
     def compute_texture(self):
 
--- a/spinalcordtoolbox/scripts/sct_analyze_texture.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_texture.py
@@ -203,8 +203,10 @@ class ExtractGLCM:
 
             dim_idx = 3 # img is [x, y, z, angle] so specify 4th dimension (angle)
 
-            img = concat_data(im2mean_lst, dim_idx).save(fname_out)
-            img = Image(fname_out)
+            img = concat_data(im2mean_lst, dim_idx).save(fname_out, mutable=True)
+
+            if len(np.shape(img.data)) < 4: # in case input volume is 3d and dim=t
+                img.data = img.data[..., np.newaxis]
             img.data = np.mean(img.data, dim_idx)
             img.save()
 
--- a/spinalcordtoolbox/scripts/sct_create_mask.py
+++ b/spinalcordtoolbox/scripts/sct_create_mask.py
@@ -28,7 +28,9 @@ from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv, set_global_l
 from spinalcordtoolbox.utils.fs import tmp_create, check_file_exist, extract_fname, rmtree, copy
 from spinalcordtoolbox.labels import create_labels
 from spinalcordtoolbox.types import Coordinate
+from spinalcordtoolbox.math import concatenate_along_4th_dimension
 
+from spinalcordtoolbox.scripts.sct_maths import get_data_or_scalar
 from spinalcordtoolbox.scripts.sct_image import concat_data
 
 
@@ -286,7 +288,11 @@ def create_line(param, fname, coord, nz):
     copy(fname, 'line.nii', verbose=param.verbose)
 
     # set all voxels to zero
-    run_proc(['sct_maths', '-i', 'line.nii', '-mul', '0', '-o', 'line.nii'], param.verbose)
+    img = Image('line.nii')
+    data = get_data_or_scalar('0', img.data)
+    data_concat = concatenate_along_4th_dimension(img.data, data)
+    img.data = np.prod(data_concat, axis=3)
+    img.save()
 
     labels = []
 
@@ -297,7 +303,7 @@ def create_line(param, fname, coord, nz):
         # backwards compat
         labels.extend([Coordinate([coord[0], coord[1], iz, 1]) for iz in range(nz)])
 
-    create_labels(Image("line.nii"), labels).save(path="line.nii")
+    create_labels(img, labels).save()
 
     return 'line.nii'
 
--- a/spinalcordtoolbox/scripts/sct_dice_coefficient.py
+++ b/spinalcordtoolbox/scripts/sct_dice_coefficient.py
@@ -130,19 +130,16 @@ def main(argv=None):
         fname_input1_bin = add_suffix(fname_input1, '_bin')
         im_1_bin = im_1.copy()
         im_1_bin.data = binarize(im_1.data, 0)
-        im_1_bin.save(fname_input1_bin)
-        fname_input1 = fname_input1_bin
+        im_1_bin.save(fname_input1_bin, mutable=True)
+        im_1 = im_1_bin
 
         fname_input2_bin = add_suffix(fname_input2, '_bin')
         im_2_bin = Image(fname_input2)
         im_2_bin.data = binarize(im_2.data, 0)
-        im_2_bin.save(fname_input2_bin)
-        fname_input2 = fname_input2_bin
+        im_2_bin.save(fname_input2_bin, mutable=True)
+        im_2 = im_2_bin
 
     # copy header of im_1 to im_2
-    im_1 = Image(fname_input1)
-    im_2 = Image(fname_input2)
-
     im_2.header = im_1.header
     im_2.save()
 
--- a/spinalcordtoolbox/scripts/sct_dice_coefficient.py
+++ b/spinalcordtoolbox/scripts/sct_dice_coefficient.py
@@ -16,7 +16,8 @@ import os
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar
 from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv, set_global_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, copy, extract_fname, rmtree
-from spinalcordtoolbox.image import add_suffix
+from spinalcordtoolbox.image import Image, add_suffix
+from spinalcordtoolbox.math import binarize
 
 
 def get_parser():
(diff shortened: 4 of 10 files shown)


spinalcordtoolbox/scripts/sct_analyze_texture.py
```diff
@@ -206,7 +206,12 @@ def mean_angle(self):
             # Average across angles and save it as wrk_folder/fnameIn_feature_distance_mean.extension
             fname_out = im_m + str(self.param_glcm.distance) + '_mean' + extension
             run_proc('sct_image -i ' + ' '.join(im2mean_lst) + ' -concat t -o ' + fname_out)
-            run_proc('sct_maths -i ' + fname_out + ' -mean t -o ' + fname_out)
+
+            img = Image(fname_out)
+            dim_idx = 3 # 3 dimensions + time
```
joshuacwnewton, line 211: Nit: I don't think the comment here is quite accurate...

```suggestion
            dim_idx = 3  # img is [x, y, z, angle] so specify 4th dimension (angle)
```

spinalcordtoolbox/scripts/sct_analyze_texture.py
```diff
@@ -206,7 +206,12 @@ def mean_angle(self):
             # Average across angles and save it as wrk_folder/fnameIn_feature_distance_mean.extension
             fname_out = im_m + str(self.param_glcm.distance) + '_mean' + extension
             run_proc('sct_image -i ' + ' '.join(im2mean_lst) + ' -concat t -o ' + fname_out)
```
joshuacwnewton, line 208: Just to make things even clearer, you could add a comment here saying something like:

```
            # We specify '-concat t' here just to concatenate along the 4th dimension 
            # But, to be clear, the 4th dimension in this case is "angle" and not "time"
```

Admittedly It's a bit weird that we have to specify `t` for these functions instead of just specifying a dimension (0/1/2/3). :stuck_out_tongue:

spinalcordtoolbox/scripts/sct_create_mask.py
```diff
@@ -293,7 +295,11 @@ def create_line(param, fname, coord, nz):
     copy(fname, 'line.nii', verbose=param.verbose)
 
     # set all voxels to zero
-    run_proc(['sct_maths', '-i', 'line.nii', '-mul', '0', '-o', 'line.nii'], param.verbose)
+    img = Image('line.nii')
+    data = get_data_or_scalar('0', img.data)
+    data_concat = concatenate_along_4th_dimension(img.data, data)
+    img.data = np.prod(data_concat, axis=3)
```
joshuacwnewton, line 301: Would it be better to merge these 3 lines into a function called, say, `multiply()` in `math.py`? Since we're repeating ourselves here and also in:

https://github.com/neuropoly/spinalcordtoolbox/blob/6129ec7ccb16b777462d57276e1f63498a1f9ef2/spinalcordtoolbox/scripts/sct_maths.py#L318-L321

(just to make things less error prone if these lines ever have to change in the future)

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -200,7 +200,13 @@ def main(args=None):
     # Average b=0 images
     if average:
         printv('\nAverage b=0...', verbose)
-        run_proc(['sct_maths', '-i', b0_name + ext, '-o', b0_mean_name + ext, '-mean', 't'], verbose)
+        img = Image(b0_name + ext)
+        out = img.copy()
+        dim_idx = 3 # 3 dimensions + time
+        if dim_idx + 1 > len(np.shape(img.data)):
+            out.data = out.data[..., np.newaxis]
```
joshuacwnewton, line 207: Do we need this check here? 

While yes, it was present in `sct_maths`, I think we can guarantee that the `b=0` images should always be of shape [x, y, z, t] because of the line immediately above:

https://github.com/neuropoly/spinalcordtoolbox/blob/6129ec7ccb16b777462d57276e1f63498a1f9ef2/spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py#L198


If the condition `if dim_idx + 1 > len(np.shape(img.data)):` ever isn't true, then there's a bigger problem... (so just appending a new axis would be hiding that).

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -211,7 +217,13 @@ def main(args=None):
     # Average DWI images
     if average:
         printv('\nAverage DWI...', verbose)
-        run_proc(['sct_maths', '-i', dwi_name + ext, '-o', dwi_mean_name + ext, '-mean', 't'], verbose)
+        img = Image(dwi_name + ext)
+        out = img.copy()
+        dim_idx = 3 # 3 dimensions + time
+        if dim_idx + 1 > len(np.shape(img.data)):
+            out.data = out.data[..., np.newaxis]
```
joshuacwnewton, line 224: Same comment here.

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -200,7 +200,13 @@ def main(args=None):
     # Average b=0 images
     if average:
         printv('\nAverage b=0...', verbose)
-        run_proc(['sct_maths', '-i', b0_name + ext, '-o', b0_mean_name + ext, '-mean', 't'], verbose)
+        img = Image(b0_name + ext)
+        out = img.copy()
+        dim_idx = 3 # 3 dimensions + time
```
joshuacwnewton, line 205: Nit: I'm not sure we need to be this explicit about what `3` means here... I think it might be clear from context.

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -320,8 +321,13 @@ def main(args=None):
              verbose=verbose,
              is_sct_binary=True,
              )
+
     # Threshold segmentation at 0.5
-    run_proc(['sct_maths', '-i', 'segmentation_straight.nii', '-thr', '0.5', '-o', 'segmentation_straight.nii'], verbose)
+    img = Image('segmentation_straight.nii')
+    data_thr = threshold(img.data, 0.5)
+    img.data = data_thr
```
joshuacwnewton, line 328: Nit: One-line?

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -396,7 +402,18 @@ def main(args=None):
...
         if laplacian:
             printv('\nApply Laplacian filter...', verbose)
-            run_proc(['sct_maths', '-i', 'data_straightr.nii', '-laplacian', '1', '-o', 'data_straightr.nii'], verbose)
+            img = Image("data_straightr.nii")
+
+            # apply std dev to each axis of the image
+            sigmas = [1 for i in range(len(img.data.shape))]
+
+            # adjust sigma based on voxel size
+            sigmas = [sigmas[i] / img.dim[i + 4] for i in range(3)]
```
joshuacwnewton, line 411: See my earlier comment about this adjustment. ("Should this be a part of the `laplacian` function, since we do it every time?")

spinalcordtoolbox/scripts/sct_analyze_texture.py
joshuacwnewton: Actually, can we replace this call to `sct_image` entirely, while we're at it?

I mean, we have `image.concat_data`, right? See also this call, which accomplishes the same thing, I think (concatenating across the 4th dimension):

https://github.com/neuropoly/spinalcordtoolbox/blob/6129ec7ccb16b777462d57276e1f63498a1f9ef2/spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py#L198

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -210,11 +209,19 @@ def main(args=None):
...
-                         '-o', 'anat_rpi_straight_smooth.nii',
-                         '-v', '0'])
+
+    img = Image("anat_rpi_straight.nii")
+    out = img.copy()
+
+    if len(sigmas) == 1:
+        sigmas = [sigmas[0] for i in range(len(img.data.shape))]
+    elif len(sigmas) != len(img.data.shape):
+            raise ValueError("-smooth need the same number of inputs as the number of image dimension OR only one input")
```
joshuacwnewton, line 219: I wonder if this argument validation logic would be better if it were placed earlier in `sct_smooth_spinalcord` (see next review comment)

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -130,7 +129,7 @@ def main(args=None):
     fname_centerline = arguments.s
     param.algo_fitting = arguments.algo_fitting
     if arguments.smooth is not None:
-        sigma = arguments.smooth
+        sigmas = arguments.smooth
```
joshuacwnewton, line 132: Argument validation logic go here?

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -210,11 +209,19 @@ def main(args=None):
...
+
+    img = Image("anat_rpi_straight.nii")
+    out = img.copy()
+
+    if len(sigmas) == 1:
+        sigmas = [sigmas[0] for i in range(len(img.data.shape))]
+    elif len(sigmas) != len(img.data.shape):
+            raise ValueError("-smooth need the same number of inputs as the number of image dimension OR only one input")
+
+    sigmas = [sigmas[i] / img.dim[i + 4] for i in range(3)]
```
joshuacwnewton, line 221: Same comment -- maybe this sigma adjustment should be placed within `math.smooth` instead.

joshuacwnewton (changes requested):
Sorry for the delay on the review... thanks for doing this! :slightly_smiling_face:

spinalcordtoolbox/registration/register.py
```diff
@@ -192,10 +193,19 @@ def register_step_ants_registration(src, dest, step, masking, ants_registration_
     # apply Laplacian filter
     if not step.laplacian == '0':
         logger.info(f"\nApply Laplacian filter")
-        run_proc(['sct_maths', '-i', src, '-laplacian', step.laplacian + ','
-                 + step.laplacian + ',0', '-o', image.add_suffix(src, '_laplacian')])
-        run_proc(['sct_maths', '-i', dest, '-laplacian', step.laplacian + ','
-                 + step.laplacian + ',0', '-o', image.add_suffix(dest, '_laplacian')])
+
+        sigmas = [step.laplacian, step.laplacian, 0]
```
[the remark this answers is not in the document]
joshuacwnewton: Actually, should this step be automatically included in the `maths.laplacian` function so that it's applied by default?

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
joshuacwnewton: There's also an identical line for `dwi` images lower down.

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
joshuacwnewton: Sorry, this isn't actually a very useful review comment. Please disregard.

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
joshuacwnewton: Sorry, this isn't actually a very useful review comment. Please disregard.

jcohenadad:
@Drulex what is the status of this PR?

Drulex:
> @Drulex what is the status of this PR?

It's in my backlog I should get to it in the next few days. It's background refactoring work, nothing potentially blocking for the release.

spinalcordtoolbox/registration/register.py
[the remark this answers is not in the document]
Drulex: Good catch.

> Actually, should this step be automatically included in the `math.laplacian` function so that it's applied by default?

I agree that it would remove some duplication, but from a design point of view I think that the current interface for `def laplacian(data, sigmas)` is correct. Just imagine someone using this API in the future and wonder why all of sudden the user provided standard deviation is modified internally by `math.laplacian()`..

spinalcordtoolbox/registration/register.py
```diff
@@ -192,10 +193,19 @@ def register_step_ants_registration(src, dest, step, masking, ants_registration_
...
+
+        src_img = image.Image(src)
+        src_out = src_img.copy()
+        src_out.data = laplacian(src_out.data, sigmas)
+        src_out.save(path=image.add_suffix(src, '_laplacian'))
+
+        dest_img = image.Image(dest)
+        dest_out = dest_img.copy()
+        dest_out.data = laplacian(dest_out.data, sigmas)
+        dest_out.save(path=image.add_suffix(dest, '_laplacian'))
```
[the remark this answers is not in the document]
Drulex: Good point.

spinalcordtoolbox/scripts/sct_analyze_texture.py
Drulex: Yes absolutely.

spinalcordtoolbox/registration/register.py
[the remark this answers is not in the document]
joshuacwnewton: > Just imagine someone using this in the future and wonder why all of sudden the user provided standard deviation is modified internally

Good point. Although, now that I think about it, what you describe is kind of already happening on the CLI level, isn't it?

Say the user calls `sct_maths -i t2.nii.gz -laplacian 2.0,3.0,2.0`... those sigma values get modified internally as you describe. :p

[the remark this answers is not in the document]
joshuacwnewton: Thinking about it a little more... I think the step might actually be necessary, and should be included `sct_math.laplacian` like I first suggested. But, this time I have a different reason than code de-duplication.

Since we're specifically a toolbox for spinal cord images, I think `sct_maths -laplacian` is trying to be more than just an invisible wrapper around `ndimage.gaussian_laplace` -- it's specifically tailoring the function to NIFTI images, which have voxel size as part of their spec. So, to ever exclude this adjustment would be applying the wrong laplacian smoothing, I think? It seems mandatory.

My new suggestion would be to have something like this:

```python
def laplacian(img: Image, sigmas: Sequence[float]):
    """
    Apply Laplacian filter to a NIFTI image. 

    :param img: Image object
    :param sigmas: List of floats, one standard deviation value per image dimension. 
    These values are adjusted internally to account for the image voxel size.
    """
    assert len(img.data.shape) == len(sigmas)
    
    # adjust sigma based on voxel size 
    sigmas = [sigmas[i] / img.dim[i + 4] for i in range(len(sigmas))]
    img.data = gaussian_laplace(img.data.astype(float), sigmas)
    return img
```

Otherwise, why should `sct_math.laplacian` exist? Wouldn't the user of the API just use `ndimage.gaussian_laplace` directly?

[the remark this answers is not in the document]
Drulex: > Otherwise, what's the use of sct_math.laplacian? As it currently is, it's no more useful than just using ndimage.gaussian_laplace directly. :stuck_out_tongue_closed_eyes: 

Yes I agree, my initial reflex was to remove it entirely.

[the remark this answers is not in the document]
Drulex: After a quick look it seems there are a few more similar cases of seemingly useless wrappers in `math.py` (and probably other easy optimizations). I think I'll save that for a different PR and focus on the subprocess calls for this PR.

spinalcordtoolbox/scripts/sct_create_mask.py
Drulex: https://github.com/neuropoly/spinalcordtoolbox/issues/3189

spinalcordtoolbox/registration/register.py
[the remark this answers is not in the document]
Drulex: https://github.com/neuropoly/spinalcordtoolbox/issues/3189

Drulex:
Thanks for the useful feedback @joshuacwnewton! I think it's ready for another pass.

spinalcordtoolbox/scripts/sct_analyze_texture.py
```diff
@@ -200,8 +200,14 @@ def mean_angle(self):
...
             fname_out = im_m + str(self.param_glcm.distance) + '_mean' + extension
-            run_proc('sct_image -i ' + ' '.join(im2mean_lst) + ' -concat t -o ' + fname_out)
-            run_proc('sct_maths -i ' + fname_out + ' -mean t -o ' + fname_out)
+
+            dim_idx = 3 # img is [x, y, z, angle] so specify 4th dimension (angle)
+
+            img = concat_data(im2mean_lst, dim_idx).save(fname_out)
+            img = Image(fname_out)
+            img.data = np.mean(img.data, dim_idx)
+            img.save()
```
joshuacwnewton, line 209: `sct_maths -mean` has a check that adds a new axis:

https://github.com/neuropoly/spinalcordtoolbox/blob/073bac3b0307fb90ebc49c98c09058044fa20bea/spinalcordtoolbox/scripts/sct_maths.py#L322-L326

In  case `sct_analyze_texture` is used with 3D data, would it safer to repeat the check before the `np.mean` step here?

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -194,7 +194,13 @@ def main(argv=None):
     # Average b=0 images
     if average:
         printv('\nAverage b=0...', verbose)
-        run_proc(['sct_maths', '-i', b0_name + ext, '-o', b0_mean_name + ext, '-mean', 't'], verbose)
+        img = Image(b0_name + ext)
+        out = img.copy()
+        dim_idx = 3
+        if len(np.shape(img.data)) < dim_idx + 1:
+            raise ValueError("Expecting image with 4 dimensions!")
```
joshuacwnewton, line 201: Is this the right check for `-mean t`? See the previous review comment, which has a different check for `-mean t`.

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
```diff
@@ -205,7 +211,13 @@ def main(argv=None):
     # Average DWI images
     if average:
         printv('\nAverage DWI...', verbose)
-        run_proc(['sct_maths', '-i', dwi_name + ext, '-o', dwi_mean_name + ext, '-mean', 't'], verbose)
+        img = Image(dwi_name + ext)
+        out = img.copy()
+        dim_idx = 3
+        if len(np.shape(img.data)) < dim_idx + 1:
+            raise ValueError("Expecting image with 4 dimensions!")
```
joshuacwnewton, line 218: Is this the right check for `-mean t`? See the previous review comment, which has a different check for `-mean t`.

spinalcordtoolbox/scripts/sct_dice_coefficient.py
```diff
@@ -122,18 +123,26 @@ def main(argv=None):
...
-        run_proc(['sct_maths', '-i', fname_input2, '-bin', '0', '-o', fname_input2_bin])
+        im_2_bin = Image(fname_input2)
+        im_2_bin.data = binarize(im_2.data, 0)
+        im_2_bin.save(fname_input2_bin)
         fname_input2 = fname_input2_bin
 
     # copy header of im_1 to im_2
-    from spinalcordtoolbox.image import Image
     im_1 = Image(fname_input1)
     im_2 = Image(fname_input2)
```
joshuacwnewton, line 144: Nit: It feels a bit weird to see the `im_1 = Image(fname_input1)` and `im_2 = Image(fname_input2)` lines repeated twice. We're saving the images to disk, only to immediately reload them to the same variables.

If we perform the binarize steps in-place (i.e. modifying `im_1` and `im_2` directly), then I think we wouldn't have reload the images in order to copy the header over.

joshuacwnewton (commented):
Thanks for addressing my feedback! Had a few small comments on second pass, but mostly looks good. :)

spinalcordtoolbox/scripts/sct_dice_coefficient.py
jcohenadad: agreed-- that also made me realize that we should probably merge the DICE calculation 👉 https://github.com/neuropoly/spinalcordtoolbox/issues/3201

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
Drulex: Comment: https://github.com/neuropoly/spinalcordtoolbox/pull/2975#discussion_r535385495

> If the condition if dim_idx + 1 > len(np.shape(img.data)): ever isn't true, then there's a bigger problem... (so just appending a new axis would be hiding that).

Is that not what you were proposing?

spinalcordtoolbox/scripts/sct_dice_coefficient.py
Drulex: Yes definitely! Didn't pay attention :sweat_smile:

spinalcordtoolbox/scripts/sct_dmri_separate_b0_and_dwi.py
joshuacwnewton: I apparently have the memory of a goldfish. Thank you for reminding me. :sweat_smile:

spinalcordtoolbox/scripts/sct_dice_coefficient.py
joshuacwnewton: Hmm, I'm not sure the fix in d4fed51 works how we want it to...

Right now `im_1` and `im_2` refer to the original image files, while  `im_1_bin` and `im_2_bin` refer to the binarized ones. So, right now, we save the header to the wrong image file if if `arguments.bin` is passed:

https://github.com/neuropoly/spinalcordtoolbox/blob/d4fed51aabe3f51a8c2f83a991cbd0fb4069f6ee/spinalcordtoolbox/scripts/sct_dice_coefficient.py#L142-L145

Drulex: Oups sorry about that, I'm a bit distracted. Do you agree with https://github.com/neuropoly/spinalcordtoolbox/commit/4652767faa1fbe09c52bcac6acc57a07fc5a19dc ?

joshuacwnewton: It looks good! Although, I just noticed another detail in the above snippet. :sweat_smile: 

There is both `im_1_bin = im_1.copy()` and `im_2_bin = Image(fname_input2)`...

joshuacwnewton: What do you think of this?

```python
    im_1 = Image(fname_input1)
    im_2 = Image(fname_input2)

    if arguments.bin is not None:
        fname_input1 = add_suffix(fname_input1, '_bin')
        im_1.data = binarize(im_1.data, 0)
        im_1.save(fname_input1, mutable=True)

        fname_input2 = add_suffix(fname_input2, '_bin')
        im_2.data = binarize(im_2.data, 0)
        im_2.save(fname_input2, mutable=True)
```

joshuacwnewton: Side note, but it is a bit weird that we have to update the fname in two places... 

1. In the image variables, using `mutable=True`
2. In the filename variables

Drulex: Sure. Save some RAM at the same time :+1:

spinalcordtoolbox/scripts/sct_analyze_texture.py
joshuacwnewton: Small nudge about this @Drulex, and then I can approve. :slightly_smiling_face:

Drulex: Yes good observation

spinalcordtoolbox/scripts/sct_analyze_texture.py
```diff
@@ -203,8 +203,10 @@ def mean_angle(self):
 
             dim_idx = 3 # img is [x, y, z, angle] so specify 4th dimension (angle)
 
-            img = concat_data(im2mean_lst, dim_idx).save(fname_out)
-            img = Image(fname_out)
+            img = concat_data(im2mean_lst, dim_idx).save(fname_out, mutable=True)
```
joshuacwnewton: Nice catch with `mutable=True`, as well. :)

joshuacwnewton (approved):
Thanks, @Drulex!

PR #2978, opened 2020-10-28
Add TravisCI code folding and timing

joshuacwnewton:
This is a test PR to see if exported travis commands work.


joshuacwnewton:
The timing observations from this PR don't really provide any new insight compared with what @Drulex had already found here: https://github.com/neuropoly/spinalcordtoolbox/issues/2923#issuecomment-718708497

PR #2980, opened 2020-10-29
Enable fail on warning in RTD

joshuacwnewton:
### Related issues/PRs

Fixes #2973.
After #2934.
After #2966.
After #2982.

### Description

Once all of the existing Sphinx warnings are taken care of (see "After" linked PRs above), enabling `sphinx.fail_on_warning` should help us catch any new warnings introduced by PRs.

This PR also cleans up a bit of dead code found during #2973.

PR #2982, opened 2020-10-29
Update sys.exit(2) calls in isct scripts so that a nonzero error code isn't given when '-h' is passed

### LINKED ISSUE #2981: sys.exit(2) from isct scripts causing warnings in Sphinx build
The `usage()` functions in `isct_convert_binary_to_trilinear` and `isct_test_ants` both end using `sys.exit(2)` in all cases, even when the usage is requested using `-h`, which is totally valid. 

Sphinx interprets this non-zero exit code as a failure of some sort and throws a warning.

```
WARNING: Unexpected return code 2 from command 'isct_convert_binary_to_trilinear -h'
WARNING: Unexpected return code 2 from command 'isct_test_ants -h'
```

These will need to be dealt with if we want to use `sphinx.fail_on_warning` to help catch documentation issues.

joshuacwnewton:
### Related issues/PRs

Fixes #2981.
Needed for #2980.

### Description

When usage (i.e. `'-h'`) is displayed for `isct_convert_binary_to_trilinear` and `isct_test_ants`, it _always_ exits with error code 2, even when explicitly requested. This throws a warning on Sphinx, which uses `'-h'` for documentation purposes.

This PR tweaks the behavior of those scripts to avoid throwing an error with valid usage.


joshuacwnewton:
Ah, the fix wasn't that straightforward. See https://readthedocs.org/projects/spinalcordtoolbox/builds/12216635/

joshuacwnewton:
These two scripts are quite old, and do not use `argparse` _or_ `msct_parser` even! So, the way these scripts invoke the `-h` usage is nonstandard.

Since these are internal scripts, I'm using a simple fix here (returning a successful error code) rather than refactoring, as that would be a much bigger effort for potentially little gain.

kousu:
The exit(2)s that you've edited so far are in there legitimately. You *do* want `usage()` getting called to be considered a failure; usage only happens when the input was bad, unless you pass `-h` on purpose, in which case it should probably be `exit(0)`; but we can probably skip handling that.

sys.exit() is a really heavy hammer. It is an immediate termination, skipping all cleanup and exception handling ((unless you use `_atexit` but that's another rabbit hole)). We should replace it with simple returns, by raising exceptions, or, where we need to signal an explicit exit code -- **like** the paths that call `usage()` -- use `raise SystemExit(2)` which is a special case in python, like `KeyboardInterrupt`, that connects python's exception system to the unix signalling/communication system.

joshuacwnewton:
Thanks much for your input, @kousu... yes, this was hasty of me. :sweat_smile: 

I think your reasoning is very sound here.

joshuacwnewton:
Closing as this is superseded by #2983.

kousu:
I think it's probably faster to show what I mean than to explain so look over #2983 to see! Feel free to ignore that and write it again your own way, I did it really quickly.

kousu:
hooray! helping :)

jcohenadad (approved):

jcohenadad:
@joshuacwnewton can we merge?

PR #2983, opened 2020-10-29
Use SystemExit not sys.exit & only on error paths.

kousu:
Suggestion for https://github.com/neuropoly/spinalcordtoolbox/pull/2982

### DIFF
--- a/scripts/isct_convert_binary_to_trilinear.py
+++ b/scripts/isct_convert_binary_to_trilinear.py
@@ -61,8 +61,10 @@ def main():
             opts, args = getopt.getopt(sys.argv[1:], 'hi:v:r:s:')
         except getopt.GetoptError:
             usage()
+            raise SystemExit(2)
         if not opts:
             usage()
+            raise SystemExit(2)
         for opt, arg in opts:
             if opt == '-h':
                 usage()
@@ -180,9 +182,6 @@ def usage():
           '\n'
           'EXAMPLE\n'
           '  ' + os.path.basename(__file__) + ' -i segmentation.nii \n')
-
-    # exit program
-    sys.exit(0)
 
 
 # =======================================================================================================================
--- a/scripts/isct_convert_binary_to_trilinear.py
+++ b/scripts/isct_convert_binary_to_trilinear.py
@@ -68,6 +68,7 @@ def main():
         for opt, arg in opts:
             if opt == '-h':
                 usage()
+                return
             elif opt in ('-i'):
                 fname_data = arg
             elif opt in ('-r'):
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
Actually, huh. https://stackoverflow.com/a/13992714

scripts/isct_convert_binary_to_trilinear.py
```diff
@@ -61,8 +61,10 @@ def main():
             opts, args = getopt.getopt(sys.argv[1:], 'hi:v:r:s:')
         except getopt.GetoptError:
             usage()
+            raise SystemExit(2)
         if not opts:
             usage()
+            raise SystemExit(2)
         for opt, arg in opts:
             if opt == '-h':
                 usage()
```
joshuacwnewton, line 70: I think we need a `raise SystemExit(0)` here or else `usage()` gets called twice.

https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/scripts/isct_convert_binary_to_trilinear.py#L78-L80

(I also think another `raise SystemExit(2)` is needed after this `fname_data` `usage()` call.)

joshuacwnewton (changes requested):
It looks like `isct_convert_binary_to_trilinear -h` fails because we didn't exit after the other `usage()` calls.

```
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py", line 196, in <module>
    main()
  File "/home/joshua/Repos/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py", line 92, in main
    check_file_exist(fname_data, verbose)
  File "/home/joshua/Repos/spinalcordtoolbox/spinalcordtoolbox/utils/fs.py", line 109, in check_file_exist
    if fname[0] == '-':
IndexError: string index out of range
```

scripts/isct_convert_binary_to_trilinear.py
kousu: What about just using `return` instead of `raise SystemExit(0)`?

kousu: Oh no of course you were right.

kousu:
@joshuacwnewton i think i got it this time. 
but btw i made this a PR into your PR so either you have to reenable yours and we need to do the two merges or I need to close this and resubmit it pointed at `master`. Can we do the former?

joshuacwnewton:
> but btw i made this a PR into your PR so either you have to reenable yours and we need to do the two merges or I need to close this and resubmit it pointed at `master`. Can we do the former?

Terribly sorry for not noticing that! Done and done. :)

PR #2990, opened 2020-11-02
Fixed missing top disc label with using -discfile 

### LINKED ISSUE #2967: sct_label_vertebrae -discfile: top disc missing
### Context

I have a file called `sub-XXX_labels-disc-manual` which includes disc labels, located posteriorly. I would like to project each disc label inside the spinal cord (orthogonally to the SC centerline). In order to do so, I run the function:
```
sct_label_vertebrae -i sub-XXX -s sub-XXX_seg -discfile sub-XXX_labels-disc-manual
```

### Problem

A disc label which was present in `sub-XXX_labels-disc-manual` is missing in the output file `sub-XXX_seg_labeled-disc`. The top label is missing. See below for a dilated example on sub-amu02 from [spine generic/multi-subject](https://github.com/spine-generic/data-multi-subject/releases/tag/r20201001) ).
image

The file produced by the command includes sub-XXX_seg_labeled-disc, this file contains the label for each disc in the middle of the spinal cord.  However, it is missing the first disc from "sub-XXX_labels-disc-manual". The image below shows the `label_disc_manual` file((red/yellow) and the `seg_labeled-disc`  file ((blue/green) overlaid on the same image. Labels have been dilated for better visual. Subject is sub-amu02 from [spine generic/multi-subject](https://github.com/spine-generic/data-multi-subject/releases/tag/r20201001).
image

As seen on the image the first disc (in dark red in `labeled_disc_manual`) is not present in `seg_labeled-disc` (no blue point near this position).  

I think that this is due to the fact that the labeled segmentation is "cut" at the IS coordinates of the "top" disc. [`label_disc`
](https://github.com/neuropoly/spinalcordtoolbox/blob/d9a8a88fb2e0a6462a46db572ed1e222fe1a4681/spinalcordtoolbox/vertebrae/core.py#L518) is using the transition between level on the `sub-XXX_seg_labeled` file (shown below) to place the label for the disc.
image

The transition at the top (3-->0) does not seem to be detected which is what causes the issue.   

### To reproduce

Data: 47a3bdcddeef8da0e5f19d15e9b1464bccf67b12
SCT version: 9a928bf254ade9a4933605e74a66225262b32169
Code:
~~~
# segment cord
sct_deepseg_sc -i ~/code/spine-generic/data-multi-subject/sub-amu/anat/sub-amu02_T2w.nii.gz -o sub-amu02_T2w_seg.nii.gz -c t2
# label vertebrae, with discfile as input
sct_label_vertebrae -i ~/code/spine-generic/data-multi-subject/sub-amu02/anat/sub-amu02_T2w.nii.gz -s sub-amu02_T2w_seg.nii.gz -discfile ~/code/spine-generic/data-multi-subject/derivatives/labels/sub-amu02/anat/sub-amu02_T2w_labels-disc-manual.nii.gz -c t2
~~~

**Expected result:**
the file `sub-amu02_T2w_seg_labeled_discs.nii.gz` shows all the labels present in `sub-amu02_T2w_labels-disc-manual.nii.gz`.

**Actual result:**
The file `sub-amu02_T2w_seg_labeled_discs.nii.gz` is missing the top label (value: 3)

lrouhier:
Fix #2967 

This PR aims at fixing the problem reported in #2967. 
This modifies the code from function `label_disc` in `spinalcordtoolbox/vertebrae/core.py`. This function now adds a label for a disc at the top of the labeled segmentation.
There is a new workflow to produce `sub-xxx_seg_labeled-disc.nii.gz` which is: 
* from the straight space, fetch the z corresponding to each disc (you have it from the variable `list_disc_z`)
* generate a single voxel label at this z location, centered in the cord
* apply the un-straightening to these blobs using `label` interpolation with sct_apply_transfo (which dilate the label to be a 2x2x2 blob and uses `nn` interpolation afterward, and retrieve the center of mass of the blob)
this could create some changes in  `sub-xxx_seg_labeled-disc.nii.gz`. Some labels might be a voxel below their previous position.  

This PR also introduce a new flag `-clean-labels` which takes a boolean argument.  When `-clean-labels 1` is used, the output `sub-xxx_seg_labeled` will go through the[ `clean_labeled_segmentation`](https://github.com/neuropoly/spinalcordtoolbox/blob/8da39b13b11c15fea4fd5fa121aa41b1d6320b9c/spinalcordtoolbox/vertebrae/core.py#L301-L330) function. 
this will provide a solution for #3059


jcohenadad (changes requested):
@lrouhier I've looked in the code, and realized that this approach to generate a disc level file is not very accurate, because of the curvature of the cord: it is not possible to precisely find the right z-slice at the interface between two levels. This does not only apply to your changes, but to the previous implementation as well.

A much more elegant approach would be to:
- from the straight space, fetch the z corresponding to each disc (you have it from the variable `list_disc_z`)
- generate a blob (e.g. 3x3x3) at this z location, centered in the cord
- apply the un-straightening to these blobs using trilinear interpolation
- find the discrete center of mass of each of these blob

lrouhier:
@jcohenadad I slightly changed the workflow to this. 

* from the straight space, fetch the z corresponding to each disc (you have it from the variable `list_disc_z`)
* generate a single voxel label at this z location, centered in the cord
* apply the un-straightening to these blobs using `label` interpolation with sct_apply_transfo (which dilate the label to be a 2x2x2 blob and uses `nn` interpolation afterward, and retrieve the center of mass of the blob)
Do you think using trilinear interpolation instead of `nn` would be better? if so maybe we could change the `label` interpolation?

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -497,45 +500,26 @@ def label_segmentation(fname_seg, list_disc_z, list_disc_value, verbose=1):
     seg.change_orientation(init_orientation).save(add_suffix(fname_seg, '_labeled'))
 
 
-def label_discs(fname_seg_labeled, verbose=1):
+def label_disc(fname_seg, list_disc_z, list_disc_value, verbose=1):
```
jcohenadad, line 503: why removing the "s" if you label more than 1 disc?

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -497,45 +500,26 @@ def label_segmentation(fname_seg, list_disc_z, list_disc_value, verbose=1):
...
-    # save disc labeled file
-    im_seg_labeled.data = data_disc
-    im_seg_labeled.change_orientation(orientation_native).save(add_suffix(fname_seg_labeled, '_disc'))
+    seg = Image(fname_seg)
+    init_orientation = seg.orientation
+    seg.change_orientation("RPI")
+    disc_data = np.zeros_like(seg.data)
+
+    for i in range(len(list_disc_z)):
+        # detected point is just above the disc
```
jcohenadad, line 519: why is that?

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -497,45 +500,26 @@ def label_segmentation(fname_seg, list_disc_z, list_disc_value, verbose=1):
...
+        slices = seg.data[:, :, list_disc_z[i]-1]
+        cx, cy = [int(x) for x in np.round(center_of_mass(slices)).tolist()]
+        disc_data[cx, cy, list_disc_z[i]-1] = list_disc_value[i] + 1
```
jcohenadad, line 522: could you add a comment for the "+1" in "list_disc_value[i] + 1"

scripts/sct_label_vertebrae.py
```diff
@@ -435,7 +435,17 @@ def main(args=None):
 
     # label discs
     printv('\nLabel discs...', verbose)
-    label_discs('segmentation_labeled.nii', verbose=verbose)
+    printv('\nUn-straighten labeling...', verbose)
```
jcohenadad, line 438: i would remove the line that says `    printv('\nLabel discs...', verbose)` and replace this one:
```suggestion
    printv('\nUn-straighten labeled discs...', verbose)
```

jcohenadad:
Looks like the new behaviour (fa3648dc1468d3587fe4ba87c07391b80dcd4f34) places the discs one slice below, compared to the old behaviour (5.0.0):

image

@lrouhier any reason for this new behaviour?

jcohenadad:
On the sct_example_data/t1, the first disc label (C1) appears 3 slices below the top part of the labeled segmentation:

image

Could be due to unreliable straightening warping field at the edge of the image (or it could be something else...)

jcohenadad (changes requested):
@lrouhier I finished reviewing your implementation. Overall, I really like your proposed changes: code is much lighter, and it makes more sense to generate disc labeling from the list of z-coordinates.

I just have a few comments, which are probably easy to address.

Also please: make sure to always add labels to PRs

lrouhier:
> On the sct_example_data/t1, the first disc label (C1) appears 3 slices below the top part of the labeled segmentation: 
> 
> Could be due to unreliable straightening warping field at the edge of the image (or it could be something else...)
> 
> **UPDATE 2020-11-11 17:55:02:** I notice the same behaviour on the t2:

I am looking into it, but in the "straight" space we have the correct label 

image
In the straight space. 

image
In the curve space.

jcohenadad:
> I am looking into it, but in the "straight" space we have the correct label

what i find strange, then, is that un-straightened labeled segmentation does not seem to be affected in the same way as the un-straightened labels. I think there is something fishy there...

lrouhier:
> what i find strange, then, is that un-straightened labeled segmentation does not seem to be affected in the same way as the un-straightened labels. I think there is something fishy there...

Update:  What seems to happen is that there are two extra slices at the top of the segmentation.
Example with this experiment (to reproduce files are [here](https://drive.google.com/drive/folders/1m9wEdsn-hqYkYrw3-LZ2IreVGAUP6ZEy?usp=sharing) with my results and the modified label): 
* I used `sct_label_vertebrae -i sub-amu01_T2w.nii.gz -s ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_seg.nii.gz -c t2 -discfile ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_labels-disc-manual.nii.gz -r 0 -ofolder keep_files_original`
* Then open fsleyes and erase label number 3 (Position=(34,145,177)), save this file as `sub-amu01_T2w_labels-disc-manual-modified.nii.gz`
* run  `sct_label_vertebrae -i sub-amu01_T2w.nii.gz -s ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_seg.nii.gz -c t2 -discfile ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_labels-disc-manual_modified.nii.gz -r 0 -ofolder keep_files_modified`

you should obtain the result presented in this gif which show the seg_labeled obtain with "original" discfile and then the seg_labeled obtained with "modified" discfile. 
ezgif com-gif-maker


You will see that on label 4 the segmentation is different which seems weird. The position of the discs (`seg_labeled_disc` file) are the same in both instances. So This actually comes from a weird behavior from the segmentation, as you may notice on the gif there seem to also have a difference of two slices between the original and the modified. 
I guess something is adding slices to the segmentation at the top. I think it may come from "clean_labeled_segmentation". I am looking. I'll update my comment later.

lrouhier:
UPDATE: I definetly think `clean_labeled_segmentation` is the culprit (see [here](https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/spinalcordtoolbox/vertebrae/core.py#L318-L327) ). 
The first line actually creates a dilation with factor 2, which would explain the two added slices at the top of the segmentation.
This gif is here to explain: 
ezgif com-gif-maker (1)

here you see what we obtain with the previous  `run sct_label_vertebrae -i sub-amu01_T2w.nii.gz -s ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_seg.nii.gz -c t2 -discfile ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_labels-disc-manual_modified.nii.gz -r 0 -ofolder keep_files_modified` 
* in pale green is the segmentation obtained with the current code. 
* In yellow you can see what we get when we comment [this line](https://github.com/neuropoly/spinalcordtoolbox/blob/99ec9229dbbecff6f9435c9a30adf4676a7408e3/scripts/sct_label_vertebrae.py#L434) so we don't use clean_labeled_segmentation.

lrouhier:
I think the code change in 6fc58eb fix the 'extra slice' issue

spinalcordtoolbox/vertebrae/core.py
lrouhier: Sorry I actually took that from the segmentation labeling, which is due to the algorithm they use and not the fact that the detected point is just above the disc. I removed that part. This error also explains https://github.com/neuropoly/spinalcordtoolbox/pull/2990#issuecomment-725704923 this behavior

spinalcordtoolbox/vertebrae/core.py
lrouhier: Done in f599d97

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -312,21 +315,30 @@ def clean_labeled_segmentation(fname_labeled_seg, fname_seg, fname_labeled_seg_n
...
     img_labeled_seg = Image(fname_labeled_seg)
+    img_labeled_seg.change_orientation('RPI')
+    native_or = img_labeled_seg.orientation
     img_seg = Image(fname_seg)
+    img_seg.change_orientation('RPI')
     data_labeled_seg_mul = img_labeled_seg.data * img_seg.data
-    # dilate to add voxels in segmentation that are not in segmentation_labeled
-    data_labeled_seg_dil = dilate(img_labeled_seg.data, 2, 'ball')
+
+    # locate 'holes' in the labeled spinal cord segmentation
```
jcohenadad, line 324:
```suggestion
    # Locate voxels that are in the native cord segmentation but not in the output labeled segmentation.
    # We obtain the index of these voxels, and they will later be filled with the appropriate label value.
```

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -312,21 +315,30 @@ def clean_labeled_segmentation(fname_labeled_seg, fname_seg, fname_labeled_seg_n
...
     data_labeled_seg_mul_bin = data_labeled_seg_mul > 0
     data_diff = img_seg.data - data_labeled_seg_mul_bin
     ind_nonzero = np.where(data_diff)
     img_labeled_seg_corr = img_labeled_seg.copy()
     img_labeled_seg_corr.data = data_labeled_seg_mul
+
+    # We dilate the labeled seg to wind the value of of the voxel in the 'holes'
```
jcohenadad, line 331:
```suggestion
    # We dilate the labeled segmentation to fill the missing voxels identified above, and obtain the closest
    # label value to assign to these voxels. 
```

lrouhier:
The goal of clean_labeled_segmentation is: 
(i) to remove voxels that are in the in `sub-XXX_seg_labeled` but not in `sub-XXX_seg`
(ii) to "fill" missing voxels (voxels that are in  `sub-XXX_seg`, `sub-XXX_seg_labeled`)
To that effect the strategy would be: 
* To multiply `sub-XXX_seg` with `sub-XXX_seg_labeled`. This would accomplish  (i)
* To find the difference between `sub-XXX_seg` and `sub-XXX_seg_labeled`. This will give us the voxels that are missing. We then find the value of these voxels by dilating `sub-XXX_seg_labeled`.
The problem with this method is that it adds slices on top of the labeled segmentation, because of the dilation. 

After discussion, the best solution would be to deactivate `clean_labeled_segmentation`. To do that and keep the code (which can be useful in certain cases), we'll create a flag `--clean-labeled`. If the flag is present the labeled segmentation will go through `clean_labeled_segmentation`, if not we will skip this step.

jcohenadad:
> we'll create a flag --clean-labeled.

to follow our convention, it should be single dash. Also, i suggest the following wording: `-clean-labels`.

also, to be consistent with the concept of "cleaning", we should only do the following multiplication if this flag is on.
https://github.com/neuropoly/spinalcordtoolbox/blob/afc5be3fe0c9da431789f41a3988739883c230c8/spinalcordtoolbox/vertebrae/core.py#L322

lrouhier:
I think this is fairly ready!

spinalcordtoolbox/vertebrae/core.py
lrouhier: I changed to code of `clean_labeled_segmentation` back to the original since we are creating the flag.

spinalcordtoolbox/vertebrae/core.py
lrouhier: done

jcohenadad:
@lrouhier pls create a unit test

lrouhier:
@jcohenadad I added a test to check if the function run with the new flag, is this what you had in mind?

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -418,13 +427,25 @@ def main(args=None):
...
-    clean_labeled_segmentation('segmentation_labeled.nii', 'segmentation.nii', 'segmentation_labeled.nii')
+
+    if clean_labels:
+        # Clean labeled segmentation
+        printv('\nClean labeled segmentation (correct interpolation errors)...', verbose)
+        clean_labeled_segmentation('segmentation_labeled.nii', 'segmentation.nii', 'segmentation_labeled.nii')
 
     # label discs
     printv('\nLabel discs...', verbose)
-    label_discs('segmentation_labeled.nii', verbose=verbose)
```
jcohenadad, line 427: isn't `label_discs` not used at all anymore?

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -20,9 +20,18 @@ def test_sct_label_vertebrae_initz_error():
...
-    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19'
+def test_sct_label_vertebrae_high_value_warning(caplog, tmp_path):
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19 -ofolder ' + tmp_path
     sct_label_vertebrae.main(command.split())
     assert 'Disc value not included in template.' in caplog.text
 
 
+def test_sct_label_vertebrae_clean_labels(tmp_path):
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19 -clean-labels 1 -ofolder tmp_path'
+    sct_label_vertebrae.main(command.split())
```
jcohenadad, line 31: maybe you could verify it does what it is expected to do? (ie clean labels)

spinalcordtoolbox/scripts/sct_label_vertebrae.py
lrouhier: No, it is used in `spinalcordtoolbox/vertebrae/core.py` to create the image with the label (in the straighten space).

lrouhier:
@jcohenadad I think this is ready! 
The test now computes the labeled_segmentation with and without the `clean-labels` flag. Then we compare the labeled_segmentation with the original segmentation in both cases. The 'cleaned' labeled_segmentation is supposed to be closer to the original one.

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -20,9 +22,37 @@ def test_sct_label_vertebrae_initz_error():
...
+
+
+
+
+
+
+
+
+
+
```
jcohenadad, line 58: needed?

jcohenadad:
Problem:
Discs missing:
~~~
Position=(156,222,7) -- Value= 1.0
Position=(156,208,7) -- Value= 2.0
Position=(156,191,7) -- Value= 3.0
Position=(162,119,6) -- Value= 7.0
Position=(166,102,6) -- Value= 8.0
Position=(184,52,7) -- Value= 10.0
Position=(192,27,7) -- Value= 11.0
~~~

To reproduce:
Version: 5bb113c5a7f17b10eadc9a1ebb01f50b94514f53
Data: @lrouhier TO DOCUMENT --> ASK ALEX

joshuacwnewton:
Just adding a note that I have an in-progress review (I didn't forget!), but I'll finish the review after https://github.com/neuropoly/spinalcordtoolbox/pull/2990#issuecomment-743398182 is addressed, as things might change. :stuck_out_tongue:

unit_testing/cli/test_cli_sct_label_vertebrae.py
lrouhier: removed !

lrouhier:
I think that this come from the fact that the `label` interpolation is not aggressive enough" What I mean is that a dilation with a ball of radius 2 might not be enough, if we increase the radius to 4 (see [here](https://github.com/neuropoly/spinalcordtoolbox/pull/2990/files#diff-f2f8fa39b49238a5b59d1739b036f8fcfa1ffbf60ac30b97a23e716eca771254R220)) it seems to work on the data used for this https://github.com/neuropoly/spinalcordtoolbox/pull/2990#issuecomment-743398182 . 
More test needs to be done to be certain that this is a good solution (or that a radius of 4 is enough)

lrouhier:
@jcohenadad I Think this is ready. 
There were two problems:
- The first one is that the dilation in `label` interpolation was a bit too small for spines that were really 'curved'. 
- The second one was when using disc file. The disc file was straightened with `nn` interpolation which leads to some error.

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -331,14 +340,13 @@ def main(args=None):
     if fname_disc:
         # Apply straightening to disc-label
         printv('\nApply straightening to disc labels...', verbose)
-        run_proc('isct_antsApplyTransforms -d 3 -i %s -r %s -t %s -o %s -n %s' %
+        run_proc('sct_apply_transfo -i %s -d %s -w %s -o %s -x %s' %
```
jcohenadad, line 343: what is the reason for replacing the call to the binary by the call to SCT's script?  historically, we were calling the script, and then it was changed to call the binary directly because execution was faster:
https://github.com/neuropoly/spinalcordtoolbox/commit/32a4decd2291be75b9c22b1caf84d130f4ac1f14#diff-f35ac866e7c1dd2ca3d9bfadff4715823e178618892decb6d03bcb9dcdf4d109

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -418,13 +426,25 @@ def main(args=None):
...
+    if clean_labels:
+        # Clean labeled segmentation
+        printv('\nClean labeled segmentation (correct interpolation errors)...', verbose)
+        clean_labeled_segmentation('segmentation_labeled.nii', 'segmentation.nii', 'segmentation_labeled.nii')
 
     # label discs
     printv('\nLabel discs...', verbose)
-    label_discs('segmentation_labeled.nii', verbose=verbose)
+    printv('\nUn-straighten labeled discs...', verbose)
+    run_proc('sct_apply_transfo -i %s -d %s -w %s -o %s -x %s' %
```
jcohenadad, line 438: same comment as in https://github.com/neuropoly/spinalcordtoolbox/pull/2990/files#r543841827, why not using the binary directly?

spinalcordtoolbox/scripts/sct_label_vertebrae.py
jcohenadad: nevermind-- the reason is because you wanted to use the feature 'label'

spinalcordtoolbox/scripts/sct_label_vertebrae.py
jcohenadad: nevermind again 😅 -- same reasoning: to be able to use the 'label' feature

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -20,9 +22,24 @@ def test_sct_label_vertebrae_initz_error():
...
 
-def test_sct_label_vertebrae_high_value_warning(caplog):
-    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19'
+def test_sct_label_vertebrae_high_value_warning(caplog, tmp_path):
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19 -ofolder ' + str(tmp_path)
     sct_label_vertebrae.main(command.split())
     assert 'Disc value not included in template.' in caplog.text
 
 
+def test_sct_label_vertebrae_clean_labels(tmp_path):
```
jcohenadad, line 31: could you also verify that the output disc labels all exist as expected? (you remember the issue with the dilation of labels, resulting in the discs disappearing)

lrouhier: I guess so. It is now done!

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -20,9 +23,34 @@ def test_sct_label_vertebrae_initz_error():
...
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19 -clean-labels 1 -ofolder ' + str(os.path.join(str(tmp_path), 'clean'))
+    sct_label_vertebrae.main(command.split())
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40,19 -ofolder ' + str(os.path.join(str(tmp_path), 'no_clean'))
```
jcohenadad, line 35: how come this one doesn't get a high value warning (see test_sct_label_vertebrae_high_value_warning)

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -20,9 +23,34 @@ def test_sct_label_vertebrae_initz_error():
...
+    image_clean.data = image_clean.data > 0.5
+    image_no_clean.data = image_no_clean.data > 0.5
+    dice_clean = compute_dice(image_clean, image_seg)
+    dice_no_clean = compute_dice(image_no_clean, image_seg)
+    # The cleaned version should be closer to the segmentation
+    assert dice_clean >= dice_no_clean
+
+
+def test_sct_label_vertebrae_consistent_disc(tmp_path):
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -discfile sct_testing_data/t2/labels.nii.gz -ofolder ' + str(tmp_path)
```
jcohenadad, line 50:
```suggestion
    """Check that all expected output labeled discs exist"""
    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -discfile sct_testing_data/t2/labels.nii.gz -ofolder ' + str(tmp_path)
```

unit_testing/cli/test_cli_sct_label_vertebrae.py
lrouhier: There is a warning in the log, but pytest does not print it ( I guess because it is in the API).

jcohenadad: my point was more: what is the reasoning for inputting "-initz 40,19", if this value creates a warning (see previous test)-- the point here is not to test the presence of a warning, but to verify the integrity of the output in "normal" (not with missing label) behaviour

lrouhier: Ah yes right, I forgot to remove it.. It's been corrected.

jcohenadad (approved):
Thank you @lrouhier !

PR #2992, opened 2020-11-03
Add function to detect missing label

### LINKED ISSUE #2986: sct_label_utils: compute_mean_square_error should gives missing label.
## Context

I am using `sct_label_utils -i <image> -MSE <REF>` .
My image was missing a label compared to the ref. Therefore the function was exiting early (see [here] )(https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/spinalcordtoolbox/labels.py#L283)

## Issue

While digging to try and see what was happening I stumbled upon these [these lines](https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/spinalcordtoolbox/labels.py#L291-L297) which seems strange. 

for example [here](https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/spinalcordtoolbox/labels.py#L293) we check if `np.round(coord.value)`  is this [list](https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/spinalcordtoolbox/labels.py#L287) which is defined from `coordinates_input`. However, `coord` is also an element from coordinates input. see [here](https://github.com/neuropoly/spinalcordtoolbox/blob/9a928bf254ade9a4933605e74a66225262b32169/spinalcordtoolbox/labels.py#L291)
It seems to perform the wrong comparison to find if there is a label mismatch.

It could also be of interest to output the MSE even if there is a missing (while warning the user about the missing label).

lrouhier:
* This PR introduces a new function named `find_missing_label` that compares a label image with a reference and return the false positive value (labels that are in the image and not in the reference) and false negative values (Labels that are present in the reference and not in the image).  

The `compute_mean_squared_error` function was also changed. 
* The MSE is now computed even if the image and the reference do not have the same number of labels.
* As mentioned in #2986, the warning for label discrepancy between the reference and the image was not working and was therefore removed.  This has been replaced with a call to `find_missing_label` which will display false negative and false positive values in the console (if there are any). 

fix #2986.


jcohenadad (changes requested):
could you please add a unit test? we try to change habits in SCT dev, and everytime a new functionality/API is added, we require to add a unit test for it. thanks!

joshuacwnewton:
@lrouhier The failing tests are not because of your PR. They are due to #3009, and will be fixed by #3010.

lrouhier:
the unit test has been implemented.
I think that it is fairly ready.

unit_testing/test_labels.py
```diff
@@ -216,3 +216,17 @@ def test_remove_other_labels_from_image(test_image):
...
+    img = test_image.copy()
+    false_positive = test_image.copy()
+
+    # modifying the data to create one false negative and one false positive
+    # label 1500 is not in test_image originally, and label 111 is.
+    false_positive.data = np.where(false_positive.data == 111, 1500, false_positive.data)
+    FP, FN = sct_labels.find_missing_label(false_positive, img)
+
+    assert int(FP[0]) == 1500
+    assert int(FN[0]) == 111
```
joshuacwnewton, line 232: I think it might be better to make this an test for `compute_mean_square_error` instead. The context for this bug is because of `sct_label_utils -i <image> -MSE <REF>` didn't work right (from #2986). So, I think the test should match the actual real-world usage as closely as possible to verify that things are working as expected. e.g.

* Running MSE with images that have mismatched labels -> asserting that a warning is thrown
* Running MSE with images that have matching labels -> asserting that no warning is thrown

spinalcordtoolbox/labels.py
```diff
@@ -267,6 +267,39 @@ def label_vertebrae(img: Image, vertebral_levels: Sequence[int] = None) -> Image
     return out
 
 
+def find_missing_label(img, ref):
```
joshuacwnewton, line 270: I'm not sure an entirely new function is necessary here, because it's not needed by anything else, is it?

 I think it might be better to focus on fixing the bug inside MSE, just so that things don't get overly complicated.

spinalcordtoolbox/labels.py
```diff
@@ -278,24 +311,13 @@ def compute_mean_squared_error(img: Image, ref: Image) -> float:
...
-    # check if all the labels in both the images match
-    if len(coordinates_input) != len(coordinates_ref):
-        raise ValueError(f"Input and reference image don't have the same number of labels! {len(coordinates_input)} vs {len(coordinates_ref)}")
-
-    rounded_coord_ref_values = [np.round(c.value) for c in coordinates_ref]
-    rounded_coord_in_values = [np.round(c.value) for c in coordinates_input]
     result = 0.0
 
+    # This line will add warning in the log if there are missing label.
+    find_missing_label(img, ref)
```
joshuacwnewton, line 317: The operation itself is pretty simple, so to fix the buggy behavior, it might be safer to just use something from [this StackOverflow question](https://stackoverflow.com/questions/35713093/how-can-i-compare-two-lists-in-python-and-return-not-matches) rather than making an entirely new function.

joshuacwnewton (changes requested):

spinalcordtoolbox/labels.py
lrouhier: I think that it could be useful. It could be of interest for some user, to know which labels are missing (check if it is systematic or not). It would be easier to retrieve the information than to use the 'Warning' in the MSE

spinalcordtoolbox/labels.py
lrouhier: You are right, I can modify find_missing_label, but I do think this is useful to have a function that returns the missing label. 
For example, to evaluate the sct_label_vertebrae, it would be easier to use this function than to go through each log file to check if there is a warning or not, as we would use `sct_run_batch` and perform prediction on 200 subjects (twice).

unit_testing/test_labels.py
lrouhier: I will add a test for that.

lrouhier: in 08fc3e4, I added a test to check if the warning is here when needed.

spinalcordtoolbox/labels.py
joshuacwnewton: > it would be easier to use this function than to go through each log file to check if there is a warning or not, as we would use `sct_run_batch` and perform prediction on 200 subjects (twice).

So, if I'm understanding correctly, we would expose this function in `sct_label_utils`, then people could include it in their `sct_run_batch` scripts as a processing step?

spinalcordtoolbox/labels.py
joshuacwnewton: Thanks for the clarification! Your understanding of typical SCT usage is helpful here -- I'm glad to know there's a use case outside of MSE. :)

spinalcordtoolbox/labels.py
lrouhier: I though abou it as more as an api, calls, but i guess it would make more sens

joshuacwnewton: Designing new API-specific features is tricky as I think [the API is not yet available](https://forum.spinalcordmri.org/t/using-sct-via-python/575/2?u=joshuacwnewton) to users... it's hard to anticipate what future usage could look like!

I was sort of thinking it could be good to leave this as a function and come back to this feature idea if it is high enough priority. (There are so many feature ideas in the issue backlog, and only so much time. :sweat_smile:)

spinalcordtoolbox/labels.py
```diff
@@ -267,6 +267,33 @@ def label_vertebrae(img: Image, vertebral_levels: Sequence[int] = None) -> Image
...
+    rounded_coord_in_values = [np.round(c.value) for c in coordinates_input]
+
+    FP = [x for x in rounded_coord_in_values if x not in rounded_coord_ref_values]
+    FN = [x for x in rounded_coord_ref_values if x not in rounded_coord_in_values]
+
+    if len(FP) > 0:
+        logger.warning("False positive value for label {}".format(' '.join(map(str, FP))))
+
+    if len(FN) > 0:
+        logger.warning("False negative value for label {}".format(' '.join(map(str, FN))))
```
joshuacwnewton, line 292: I'm having some second thoughts about the terms `false positive` and `false negative` here. They have a very specific meaning (detection tasks), but this function is more general (any labeled images, not necessarily tied to detection). So the warning messages might be a bit unclear for users. (They might think "False postiive? What does that mean here, I was just trying to calculate MSE!")
 
The old messages used more general language. Maybe it would be good to preserve that. e.g. you could rename the variables `FP` -> `missing_labels_ref`, `FN` -> `missing_labels_inp`. Then you could reword the messages something like:

```suggestion
    if missing_labels_ref:
        logger.warning(f"Label mismatch: Labels {missing_labels_ref} present in input image but missing from reference image.")

    if missing_labels_inp:
        logger.warning(f"Label mismatch: Labels {missing_labels_inp} present in reference image but missing from input image.)
```

spinalcordtoolbox/labels.py
```diff
@@ -278,24 +305,13 @@ def compute_mean_squared_error(img: Image, ref: Image) -> float:
...
-    # check if all the labels in both the images match
-    if len(coordinates_input) != len(coordinates_ref):
-        raise ValueError(f"Input and reference image don't have the same number of labels! {len(coordinates_input)} vs {len(coordinates_ref)}")
-
-    rounded_coord_ref_values = [np.round(c.value) for c in coordinates_ref]
-    rounded_coord_in_values = [np.round(c.value) for c in coordinates_input]
     result = 0.0
 
+    # This line will add warning in the log if there are missing label.
+    find_missing_label(img, ref)
```
joshuacwnewton, line 311: It might be good to add this to indicate the function returns something, but we're just not using it. 

```suggestion
    _, _ = find_missing_label(img, ref)
```

spinalcordtoolbox/labels.py
```diff
@@ -267,6 +267,33 @@ def label_vertebrae(img: Image, vertebral_levels: Sequence[int] = None) -> Image
     return out
 
 
+def find_missing_label(img, ref):
```
joshuacwnewton, line 270:
```suggestion
def check_missing_label(img, ref):
```

What do you think of this function name instead? In the usage for `MSE`, we don't do anything with the missing labels... we only use this function to warn the user. So, `check` might be a more descriptive word to use.

spinalcordtoolbox/labels.py
```diff
@@ -267,6 +267,33 @@ def label_vertebrae(img: Image, vertebral_levels: Sequence[int] = None) -> Image
...
+    """
+    Function that return the list of label that are present in ref and not in img.
+    This is useful to compute false negative (label that are in ref and not img) and false positive
+    (labels that are present in img and not ref)
+    :param img: source image
+    :param ref: reference image
+    :return: list of missing label
```
joshuacwnewton, line 277: We return two lists, don't we?

joshuacwnewton (changes requested):

unit_testing/test_labels.py
```diff
@@ -139,11 +139,28 @@ def test_compute_mean_squared_error():
...
 
+def test_compute_mean_squared_error_warning(caplog):
+    src = fake_3dimage_sct()
+    ref = src.copy()
+
+    src.data[0, 0, 0] = 95
+
+    sct_labels.compute_mean_squared_error(src, ref)
+    assert 'False positive value for label 95.0' in caplog.text
+    assert 'False negative value for label 111.0' in caplog.text
```
joshuacwnewton, line 154: It might be good to try to fetch the value `111.0` programmatically. That way, if `fake_3dimage_sct` changes in the future, the test won't break.

unit_testing/test_labels.py
```diff
@@ -216,3 +233,17 @@ def test_remove_other_labels_from_image(test_image):
...
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_find_missing_label(test_image):
+    img = test_image.copy()
+    false_positive = test_image.copy()
+
+    # modifying the data to create one false negative and one false positive
+    # label 1500 is not in test_image originally, and label 111 is.
+    false_positive.data = np.where(false_positive.data == 111, 1500, false_positive.data)
+    FP, FN = sct_labels.find_missing_label(false_positive, img)
```
joshuacwnewton, line 246: I notice that this looks a little different than the `test_compute_mean_squared_error_warning` test. Is there a reason why 2 different ways were used? Might be good to be consistent here.

unit_testing/test_labels.py
```diff
@@ -216,3 +233,17 @@ def test_remove_other_labels_from_image(test_image):
     diff = res.data == expected.data
 
     assert diff.all()
+
+
+@pytest.mark.parametrize("test_image", test_images)
+def test_find_missing_label(test_image):
+    img = test_image.copy()
```
joshuacwnewton, line 240: Ditto about consistency for this. In one test, `test_images` is used, and for the other test, only `fake_3dimage_sct` is used.

lrouhier: You are right, I think `test_image` might more complete (and I used a few tests in this file)  but the MSE is only tested on `fake_3d_image_sct` (like other tests in this file). Do you have an opinion on which one should be used? Should I correct the rest of the test as well for the sake of consistency?

unit_testing/test_labels.py
lrouhier: The reason was: this way is needed for `test_image` but not for `fake_3dimage_sct`. I will make it more consistant.

unit_testing/test_labels.py
lrouhier: Agreed.

unit_testing/test_labels.py
joshuacwnewton: I think in general it's good to parametrize only when the differences in inputs are relevant to the code being tested.  

In this case, the only thing the `find_missing_label` does is compare lists of coordinate values, so I don't think the difference between 3d/2d is very relevant. Using only 1 image should be sufficient, I think.

spinalcordtoolbox/labels.py
lrouhier: corrected in 
8158ad4

unit_testing/test_labels.py
lrouhier: It has been corrected (and the message has changed per your suggestion)

unit_testing/test_labels.py
lrouhier: I made it more consistent across test

unit_testing/test_labels.py
lrouhier: Right, I used `fake_3d_image_sct`

spinalcordtoolbox/labels.py
lrouhier: Yes I see what you mean. we can leave it here for now and we think it could be useful for the user we could add it pretty easily to sct_label_utils.

unit_testing/test_labels.py
```diff
@@ -139,11 +139,30 @@ def test_compute_mean_squared_error():
...
+def test_compute_mean_squared_error_warning(caplog):
+    src = fake_3dimage_sct()
+    ref = src.copy()
+
+    src.data = np.where(src.data == ref.data[0, 0, 0], 1500, ref.data)
+
+    sct_labels.compute_mean_squared_error(src, ref)
+    # Cannot use f-string in assert, I needed to create a variable before
+    string_form = f'Label mismatch: Labels [{ref.data[0,0,0]}] present in reference image but missing from input image.'
+    assert 'Label mismatch: Labels [1500.0] present in input image but missing from reference image.' in caplog.text
```
joshuacwnewton, line 155: I think this assertion might need to be updated to use `string_form`.

spinalcordtoolbox/labels.py
```diff
@@ -267,6 +267,34 @@ def label_vertebrae(img: Image, vertebral_levels: Sequence[int] = None) -> Image
     return out
 
 
+def check_missing_label(img, ref):
+    """
+    Function that return the list of label that are present in ref and not in img.
+    This is useful to compute false negative (label that are in ref and not img) and false positive
+    (labels that are present in img and not ref)
```
joshuacwnewton, line 274: Nit: This part of the docstring might need to be updated too. (Re: comments about 2 lists, FP/FN)

unit_testing/test_labels.py
```diff
@@ -139,11 +139,30 @@ def test_compute_mean_squared_error():
...
+    assert string_form in caplog.text
+
+
+def test_compute_mean_squared_error_no_warning(caplog):
+    src = fake_3dimage_sct()
+    ref = src.copy()
+    sct_labels.compute_mean_squared_error(src, ref)
+    assert 'WARNING' not in caplog.text
```
joshuacwnewton, line 163: This test will fail for any warning, even unrelated warnings (e.g. `DeprecationWarnings` from dependencies). I think that might make the test too broad to be useful, when we're interested in specifically the `Label mismatch` warning. 

(Similarly, the test names are a little too broad... it might be good to give them slightly more specific names, like `test_compute_mse_no_label_warning` and `test_compute_mse_label_warning`)

joshuacwnewton (changes requested):
Thanks for all your edits, @lrouhier! My apologies for all my nitpicky comments, heh. I hope this isn't too bothersome. :sweat_smile:

unit_testing/test_labels.py
lrouhier: Like use inp.data[0,0,0] ?

unit_testing/test_labels.py
lrouhier: Got it I will make it more specific :)

lrouhier: done :)

spinalcordtoolbox/labels.py
lrouhier: Done :)

unit_testing/test_labels.py
lrouhier: Changed it, tell me if that is what you had in mind !

joshuacwnewton: Ah sorry, I more meant the variable `string_form` was created but not used. The two lines above are near-duplicates.  :sweat_smile:

joshuacwnewton: Oh, I am just blind. I missed that `string_form` was indeed being used at the time. My bad. :sweat_smile: 

Well, either way, it's all good now!

unit_testing/test_labels.py
```diff
@@ -139,11 +139,31 @@ def test_compute_mean_squared_error():
...
-
     mse = sct_labels.compute_mean_squared_error(src, ref)
     assert mse == 1.1547005383792515
 
 
+def test_compute_mse_label_warning(caplog):
+    src = fake_3dimage_sct()
+    ref = src.copy()
+
+    src.data = np.where(src.data == ref.data[0, 0, 0], 1500, ref.data)
```
joshuacwnewton, line 150: It might be good to add the same comment you wrote for the other test ("modifying the data to create label mismatch", or "false positive/false negative") just to clarify what this operation does.

joshuacwnewton (approved):
I have one tiny nitpick but I'm giving my pre-approval as I think this looks good. :smile: 

Thanks much!

unit_testing/test_labels.py
lrouhier: Done

joshuacwnewton:
I think @jcohenadad may need to approve this PR as well, because GitHub sees that he requested changes.

jcohenadad (approved):

PR #2996, opened 2020-11-03
remove models for sct_deepseg 

### LINKED ISSUE #2993: Logging is doubled in sct_deepseg
currently for tasks with 2+ models, the output message is a bit confusing as it duplicates the syntax for viewing results, example below for a task with 2 models:

~~~

Loaded 1 sagittal volumes of shape [256, 128, 16].
Done! To view results, type:
Done! To view results, type:
fsleyes /Volumes/sct_testing/large/sub-montpellierLesion001/anat/sub-montpellierLesion001_acq-sagthor_T2w.nii.gz -cm greyscale /Volumes/sct_testing/large/sub-montpellierLesion001/anat/sub-montpellierLesion001_acq-sagthor_T2w_seg.nii.gz -cm red -a 70.0 &
fsleyes /Volumes/sct_testing/large/sub-montpellierLesion001/anat/sub-montpellierLesion001_acq-sagthor_T2w.nii.gz -cm greyscale /Volumes/sct_testing/large/sub-montpellierLesion001/anat/sub-montpellierLesion001_acq-sagthor_T2w_seg.nii.gz -cm red -a 70.0 &
~~~

The problem is this loop:
https://github.com/neuropoly/spinalcordtoolbox/blob/8e611ef205c512012b1b3fef413174514b411541/scripts/sct_deepseg.py#L156

We should display the final message only once.

Also: the "fsleyes" message is usually written in green

### LINKED ISSUE #2994: Some models in sct_deepseg cannot be run in standalone
There are currently tasks and models available on sct_deepseg. One task can require multiple models (e.g., `seg_tumor_t2` task will use two models: `findcord_tumor` and `t2_tumor`). Some models are not standalone and raise an error when used alone (e.g. `t2_tumor`: see image).

image

Possible solutions:
- To avoid confusion, models that can't be used alone could be excluded from the list of models and it would be impossible to call them. 
- Another option would be to call the task `seg_tumor_t2` when the model `t2_tumor` is called. In this case, we might want to get rid of tasks and only keep models (since models and tasks can get confusing for the user).

andreanne-lemay:
This PR introduces the following changes:
- To avoid confusion and errors, sct_deepseg will only have task (and no models). Fixes #2994
- Update legend to remove the "default [*]" in `-list-tasks`. Fixes #2996

To address in subsequent PRs:
- Remove the field "default" in models: https://github.com/neuropoly/spinalcordtoolbox/issues/3007
- Duplication of logs output. Fixes #2993

### DIFF
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -17,7 +17,7 @@ import sys
 
 import spinalcordtoolbox.deepseg as deepseg
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
-from spinalcordtoolbox.utils.sys import printv
+from spinalcordtoolbox.utils.sys import printv, init_sct
 
 
 def get_parser():
@@ -145,4 +145,5 @@ def main():
 
 
 if __name__ == '__main__':
+    init_sct()
     main()
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -145,5 +145,4 @@ def main():
 
 
 if __name__ == '__main__':
-    init_sct()
     main()
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -17,7 +17,7 @@ import sys
 
 import spinalcordtoolbox.deepseg as deepseg
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
-from spinalcordtoolbox.utils.sys import init_sct, printv
+from spinalcordtoolbox.utils.sys import printv
 
 
 def get_parser():
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -100,6 +100,9 @@ def main():
     parser = get_parser()
     args = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
 
+    # Update log_level
+    init_sct(log_level=args.v, update=True)
+
     # Deal with task
     if args.list_tasks:
         deepseg.models.display_list_tasks()
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -16,8 +16,6 @@ import os
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
-import spinalcordtoolbox.deepseg.models
-import spinalcordtoolbox.deepseg.core
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
 from spinalcordtoolbox.utils.sys import init_sct, printv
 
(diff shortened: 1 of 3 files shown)


scripts/sct_deepseg.py
```diff
@@ -141,15 +114,11 @@ def main():
...
+    if args.task is None:
+        parser.error("You need to specify a task.")
 
     # Get pipeline model names
-    if args.task is not None:
-        name_models = deepseg.models.TASKS[args.task]['models']
-
-    if args.model is not None:
-        name_models = args.model
+    name_models = deepseg.models.TASKS[args.task]['models']
```
charleygros, line 121: Agree with all this, great!

unit_testing/test_deepseg.py
```diff
@@ -12,7 +12,6 @@
 import spinalcordtoolbox as sct
 from spinalcordtoolbox.utils import sct_test_path
 import spinalcordtoolbox.deepseg.core
```
charleygros, line 14:
```suggestion
```

charleygros:
> #2993 : The duplication in logs is due to this line: [init_sct()](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg.py#L179) I removed it in this PR to solve the issue. Is this line important?

I think this line is important for the logger and displaying SCT version etc. See difference herebelow:
```
charley@charley:~/sct-dev$ sct_deepseg
usage: sct_deepseg [-i <file>] [-o <str>] [-task <str>] [-list-tasks] [-install-task {seg_sc_t2star,seg_mice_sc,seg_mice_gm,seg_tumor_t2}]
                   [-thr <float>] [-largest LARGEST] [-fill-holes {0,1}] [-remove-small REMOVE_SMALL] [-v {0,1}] [-h]

Segment an anatomical structure or pathologies according to the specified deep learning model.
```

Whereas on `master`:
```
charley@charley:~/sct-dev$ sct_deepseg

--
Spinal Cord Toolbox (git-master-5a23e7df5d9b99a661d8a74b684330fdcfe9b8d0)

usage: sct_deepseg [-i <file>] [-o <str>] [-task <str>] [-list-tasks] [-install-task {seg_sc_t2star,seg_mice_sc,seg_mice_gm,seg_tumor_t2}]
                   [-model <str> [<str> ...]] [-list-models]
                   [-install-model {t2star_sc,mice_uqueensland_sc,mice_uqueensland_gm,t2_tumor,findcord_tumor}] [-thr <float>]
                   [-largest LARGEST] [-fill-holes {0,1}] [-remove-small REMOVE_SMALL] [-v {0,1}] [-h]

Segment an anatomical structure or pathologies according to the specified deep learning model.
```

scripts/sct_deepseg.py
```diff
@@ -145,4 +145,5 @@ def main():
 
 
 if __name__ == '__main__':
+    init_sct()
```
charleygros: Observation: when doing:
```
init_sct(update=True)
```
--> it solves #2993 , ie no duplication of the output / fsleyes command --> the issue may come from [here](https://github.com/neuropoly/spinalcordtoolbox/blob/5a23e7df5d9b99a661d8a74b684330fdcfe9b8d0/spinalcordtoolbox/utils/sys.py#L67).

charleygros:
[de3c97c](https://github.com/neuropoly/spinalcordtoolbox/pull/2996/commits/de3c97c71691c5b97f6abf75ef1727aa5b421a8f) does not solve #2993, sadly.
I was trying to reproduce what is done in `sct_deepseg_sc`, [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg_sc.py#L175), since this function display the output only once.

andreanne-lemay:
> [de3c97c](https://github.com/neuropoly/spinalcordtoolbox/pull/2996/commits/de3c97c71691c5b97f6abf75ef1727aa5b421a8f) does not solve #2993, sadly.
> I was trying to reproduce what is done in `sct_deepseg_sc`, [here](https://github.com/neuropoly/spinalcordtoolbox/blob/master/scripts/sct_deepseg_sc.py#L175), since this function display the output only once.

Not all prints appear twice, only the logger.info, not sure why only with sct_deepseg functions :woman_shrugging: , I quickly looked into it and couldn't find the source of the problem. Maybe we can merge this PR and create another one for the print duplicate problem.

jcohenadad:
The double logging is caused by this import:
https://github.com/neuropoly/spinalcordtoolbox/blob/66f2246ffd904ee0c7345c7d309c92fc6edba850/scripts/sct_deepseg.py#L18

investigating the exact cause...

jcohenadad (approved):
Thank you @andreanne-lemay !

PR #2997, opened 2020-11-04
Revert "More sane package imports"

Drulex:
This reverts commit f65df321b078b939e5de66755ca0561fe773ed82.

Based on discussion: https://github.com/neuropoly/spinalcordtoolbox/issues/2989

Implicit imports causing undesirable increase in load time.

### DIFF
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -16,6 +16,9 @@ import os
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
+import spinalcordtoolbox.deepseg.models
+import spinalcordtoolbox.deepseg.core
+
 from spinalcordtoolbox.utils.shell import SmartFormatter, Metavar, display_viewer_syntax
 from spinalcordtoolbox.utils.sys import init_sct, printv
 
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -21,8 +21,8 @@ from matplotlib.figure import Figure
 import matplotlib.colors as color
 
 from spinalcordtoolbox.image import Image
-from . import slice as qcslice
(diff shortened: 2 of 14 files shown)


jcohenadad:
@joshuacwnewton could you please review this? I think you would do a better job than me

scripts/sct_deepseg.py
```diff
@@ -16,6 +16,9 @@
 import sys
 
 import spinalcordtoolbox.deepseg as deepseg
+import spinalcordtoolbox.deepseg.models
+import spinalcordtoolbox.deepseg.core
+
```
joshuacwnewton, line 21: Just to be clear about the PR description, this change is new, and wasn't a part of f65df321b078b939e5de66755ca0561fe773ed82, right?

From what I understand, though, this is needed to fix the usage of `deepseg.models.TASKS` and `deepseg.core.DEFAULTS`. So, it looks OK to me.

spinalcordtoolbox/reports/qc.py
```diff
@@ -21,8 +21,8 @@
 import matplotlib.colors as color
 
 from spinalcordtoolbox.image import Image
-from . import slice as qcslice
-from spinalcordtoolbox.utils import sct_dir_local_path, list2cmdline, __version__, copy, extract_fname
+import spinalcordtoolbox.reports.slice as qcslice
+from spinalcordtoolbox.utils import sct_dir_local_path, list2cmdline, __version__, printv, copy, extract_fname
```
joshuacwnewton, line 25: I don't think `printv` should be re-added here.

joshuacwnewton (changes requested):

scripts/sct_deepseg.py
Drulex: Yes this was from another commit which I didn't bother to track and revert.

joshuacwnewton (approved):

PR #3012, opened 2020-11-06
Introduce flag to list labels

### LINKED ISSUE #2894: update usage for map
usage says: "N.B. ONLY USE THIS METHOD WITH THE WHITE MATTER ATLAS! The sum of all tracts should be 1 in all voxels (the algorithm doesn't normalize the atlas)." 

should be updated now that atlas includes GM as well

### LINKED ISSUE #3003: Introduce flag to list labels
With the proposed changed in https://github.com/neuropoly/spinalcordtoolbox/pull/3000, labels ID will not be listed automatically when running the usage of `sct_extract_metric`. 

A solution would be to introduce a flag, e.g. `list-labels`, which would list labels in the file "info_label.txt", which located in the folder specified by `-f`.

jcohenadad:
With the proposed changed in https://github.com/neuropoly/spinalcordtoolbox/pull/3000, labels ID will not be listed automatically when running the usage of `sct_extract_metric`. 

This PR introduces a flag `-list-labels`, which lists labels in the file "info_label.txt", which located in the folder specified by `-f`. Fixes #3003 

Additional minor changes:
- Improve documentation usage for map. Fixes #2894


scripts/sct_extract_metric.py
```diff
@@ -46,6 +45,35 @@ def __init__(self):
...
+                 option_strings,
+                 dest=argparse.SUPPRESS,
+                 default=argparse.SUPPRESS,
+                 help=None):
+        super(_ListLabelsAction, self).__init__(
+            option_strings=option_strings,
+            dest=dest,
+            default=default,
+            nargs=0,
+            help=help)
```
joshuacwnewton, line 60: Is this `__init__` necessary? I was able to get rid of it by adding `nargs=0,` to the `-list-label` definition.

scripts/sct_extract_metric.py
```diff
@@ -103,18 +131,26 @@ def get_parser():
              "of consecutive labels use ':'. Example: 1:3 is equivalent to 1,2,3. Maximum Likelihood (or MAP) is "
              "computed using all tracts, but only values of the selected tracts are reported."
     )
+    optional.add_argument(
+        '-list-labels',
+        action=_ListLabelsAction,
```
joshuacwnewton, line 136: For the above suggestion.

```suggestion
        '-list-labels',
        nargs=0,
        action=_ListLabelsAction,
```

scripts/sct_extract_metric.py
```diff
@@ -46,6 +45,35 @@ def __init__(self):
...
+            default=default,
+            nargs=0,
+            help=help)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        file_label = os.path.join(namespace.f, param_default.file_info_label)
+        check_file_exist(file_label, 0)
+        default_info_label = open(file_label, 'r')
+        label_references = default_info_label.read()
+        default_info_label.close()
```
joshuacwnewton, line 67: Small Python style nit, but this can be simplified using `with`:

```suggestion
        with open(file_label, 'r') as default_info_label:
            label_references = default_info_label.read()
```

joshuacwnewton (commented):
Clever thinking! One quirk of using this approach, however, is that it's dependent on the order of arguments:

```sh
# This works. It looks for `test/info_label.txt`
sct_extract_metric -f test -list-labels 
# This doesn't work. It uses the default value for `-f`
sct_extract_metric -list-labels -f test
```

I'm having a hard time thinking of a way around this...

jcohenadad:
> Clever thinking! One quirk of using this approach, however, is that it's dependent on the order of arguments:

wow 😲  good catch! i didn't test that. Hum... not sure how to fix that either, but it will likely represent <5% of our user base, so we could leave it for later (releasing v5.0 ASAP is more important)

scripts/sct_extract_metric.py
jcohenadad: so is the idea to accept this commit, and remove the `__init__` from above?

joshuacwnewton: Yes, exactly. (Unless there was a separate reason for using `__init__` that I missed. I had never seen that usage before so it was unclear to me.)

joshuacwnewton (approved):

scripts/sct_extract_metric.py
jcohenadad: Implemented in 7cbe901, thank you for the suggestion!

jcohenadad: And regarding the motivation for putting this `__init__` in the first place: I applied the solution proposed in this [stackoverflow thread](https://stackoverflow.com/questions/34352405/python-argparse-help-like-option) without taking the time to properly analyze it (it worked so I just went ahead with it due to lack of time)

PR #3020, opened 2020-11-08
Display command when scripts are called from the command-line

### LINKED ISSUE #2388: Display command with input arguments upon launching SCT
Having a verbose of the input arguments would be useful for:
- helping debugging on the user end
- knowing what was launched when running batch processing pipelines, esp. with the new `sct_run_batch`

joshuacwnewton:
### Related issues/PRs

Fixes #2388.

### Description

This PR causes the command to be output when scripts are called from the command line. It does not add this output when scripts are called internally using `<sct_script>.main()`.

#### Sample output (`sct_create_mask -i mt/mt1.nii.gz -p coord,15x17 -size 10 -r 0`)

Before 

```sh
--
Spinal Cord Toolbox (git-master-c29d6c332662ae8fad0954f70991f35078d5c399)

Creating temporary folder (/tmp/sct-20201108124345.693440-create_mask-gcnlyd6q)

# Rest of program output goes here

Process finished with exit code 0
```

After

```sh
--
Spinal Cord Toolbox (git-jn/2388-display-script-command-245baf340bb9d07ade29aa2f708cbedeb0b06502)

Command run: sct_create_mask -i mt/mt1.nii.gz -p coord,15x17 -size 10 -r 0
--

Creating temporary folder (/tmp/sct-20201108124257.376323-create_mask-radd3pxn)

# Rest of program output goes here

Process finished with exit code 0
```

### DIFF
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -80,6 +80,14 @@ def init_sct(log_level=1, update=False):
         # Display SCT version
         logger.info('\n--\nSpinal Cord Toolbox ({})\n'.format(__version__))
 
+        # Display command (Only if called from CLI: check for .py in first arg)
+        # Use next(iter()) to not fail on empty list (vs. sys.argv[0])
+        if '.py' in next(iter(sys.argv), None):
+            script = os.path.basename(sys.argv[0]).strip(".py")
+            arguments = ' '.join(sys.argv[1:])
+            logger.info(f"Command run: {script} {arguments}\n"
+                        f"--\n")
+
 
 def add_elapsed_time_counter():
     class Timer():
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/utils/sys.py
```diff
@@ -80,6 +80,14 @@ def _format(record):
         # Display SCT version
         logger.info('\n--\nSpinal Cord Toolbox ({})\n'.format(__version__))
 
+        # Display command (Only if called from CLI: check for .py in first arg)
+        # Use next(iter()) to not fail on empty list (vs. sys.argv[0])
+        if '.py' in next(iter(sys.argv), None):
+            script = os.path.basename(sys.argv[0]).strip(".py")
+            arguments = ' '.join(sys.argv[1:])
+            logger.info(f"Command run: {script} {arguments}\n"
```
jcohenadad, line 88: I would personally remove the "Command run:". People will quickly get it and it adds superfluous text.
```suggestion
            logger.info(f"{script} {arguments}\n"
```

joshuacwnewton: Good point. Also, it makes it harder to copy and paste the command by double-clicking. :stuck_out_tongue:

jcohenadad (approved):

PR #3021, opened 2020-11-09
RTD: Re-enable showing version text underneath logo 

### LINKED ISSUE #3015: Add tag-based release versions to RTD
I notice our documentation only has a single version: `v: latest`. That means that changes to the documentation appear immediately once merged (independent of our release cycle). 

I think it would be good to activate versions based on our release tags, so that we have a stable snapshot of what the docs looked like exactly on release. That way, future changes to the docs don't end up being misleading for people who are sticking with an older release. (Without this, development of the docs will become out of sync with the development of SCT.)

More info here: https://docs.readthedocs.io/en/stable/versions.html#states

joshuacwnewton:
### Related Issues/PRs

Fixes #3015. (https://github.com/neuropoly/spinalcordtoolbox/issues/3015#issuecomment-723709323)

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -106,7 +106,7 @@ html_logo = '../imgs/logo_sct.png'
 
 html_theme_options = {
     "collapse_navigation": True,
-    "display_version": False,
+    "display_version": True,
     "sticky_navigation": True,  # Set to False to disable the sticky nav while scrolling.
     "logo_only": True,  # if we have a html_logo below, this shows /only/ the logo with no title text
     "style_nav_header_background": "#FFFFFF",  # background of the logo (top left)


jcohenadad (changes requested):
Currently it is showing the PR number, we need to fetch the version.txt from the release branch, I think.

joshuacwnewton:
> Currently it is showing the PR number, we need to fetch the version.txt from the release branch, I think.

That's because RTD is generating the docs from the PR, so in this case, the version is the PR number.

If this were merged, then the version would display as `latest`, `stable`, `5.0.0`, etc. depending on the version selected.

jcohenadad:
@joshuacwnewton thank you for the clarification, i thought that "stable" was not displayed, but instead would show the latest stable version. It is clearer now.

Just to be sure: how are "latest" and "stable" defined? Is it a feature in RTD that would automatically link "latest" with master (or the new "main"), and "stable" with the latest release?

joshuacwnewton:
> Just to be sure: how are "latest" and "stable" defined? Is it a feature in RTD that would automatically link "latest" with master (or the new "main"), and "stable" with the latest release?

Yes, precisely. `latest` is set using a drop-down menu:

Screenshot from 2020-11-11 10-05-55

For `stable`, From [**How we envision versions working**](https://docs.readthedocs.io/en/stable/versions.html#how-we-envision-versions-working):

> You should push a tag for each version of your project. These tags should be numbered in a way that is consistent with semantic versioning. This will map to your stable branch by default.

> We in fact are parsing your tag names against the rules given by PEP 440. This spec allows “normal” version numbers like 1.4.2 as well as pre-releases. An alpha version or a release candidate are examples of pre-releases and they look like this: 2.0a1.

> We only consider non pre-releases for the stable version of your documentation.

jcohenadad (approved):

PR #3022, opened 2020-11-09
Refactor deepseg/core.py into sct_deepseg and update relevant test

### LINKED ISSUE #2971: Refactor deepseg/core.py after postprocessing PR
After #2969, the only function left in `spinalcordtoolbox/deepseg/core.py` will be `segment_nifti`. This function is a very light wrapper around `imed.utils.segment_volume` with some file output afterwards.

Since we want to be handling file I/O at the CLI level anyway, there isn't much reason to have this API function (compared to calling the `imed` function directly). Once this is removed, there is no reason for `core.py` to exist either. So, the following steps are needed:

* [ ] Move the `imed` call and file I/O to `sct_deepseg`
* [ ] Remove the now empty `core.py` file
* [ ] Refactor `unit_testing/test_deepseg` to into a functional test of `sct_deepseg`

joshuacwnewton:
### Related Issues/PRs

Fixes #2971.


jcohenadad (approved):
awesome! could you pls add labels + milestone
thanks

PR #3028, opened 2020-11-09
Update references

alexfoias:
The reference is added without a link cause it's not available yet. The submission was complete and accepted.


jcohenadad (changes requested):
we should wait until we get an URL to add

jcohenadad:
> we should wait until we get an URL to add

link to Burke et al will be added later https://github.com/neuropoly/spinalcordtoolbox/issues/3058

jcohenadad (approved):

PR #3031, opened 2020-11-11
sct_run_batch: handle the case of unexecutable script

### LINKED ISSUE #3030: Empty log files produced by `sct_run_batch` due to non-executable script
When trying to run `sct_run_batch` with a script that includes trivial command (e.g., `echo`), it produces errored and empty log files.

More context in the [SCT forum issue](https://forum.spinalcordmri.org/t/minimal-correct-script-for-sct-run-batch/565).

cfhammill:
Either sets the script itself to executable if the output directory is the current working directory, or sets the output directory copy to executable and uses that.

This should resolve #3030.

Potential improvement:
Right now, `sct_run_batch` copies the user's script to the output directory, in the event the current working directory is the output directory there is no copy made. In this case I opted to modify the original script, which is perhaps distasteful, but I feel it is relatively harmless. If others disagree, we could make a copy with a suffix and modify the permissions on that.

### DIFF
--- a/scripts/sct_run_batch.py
+++ b/scripts/sct_run_batch.py
@@ -38,6 +38,7 @@ from spinalcordtoolbox.utils.shell import Metavar, SmartFormatter
 from spinalcordtoolbox.utils.sys import send_email, init_sct, __get_commit, __get_git_origin, __version__
 from spinalcordtoolbox.utils.fs import Tee
 
+from stat import S_IEXEC
 
 def get_parser():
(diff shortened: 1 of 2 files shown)


leandervaneekelen:
I can confirm this solves the issue described in #3030 and on [the forum post ](https://forum.spinalcordmri.org/t/minimal-correct-script-for-sct-run-batch/565). Thanks!

jcohenadad (approved):

PR #3034, opened 2020-11-11
Bump ivadomed version to 2.5.0

### LINKED ISSUE #3033: SCT will not install correctly on macOS due to libomp dependency of onnxruntime v1.5.1
### Context

https://forum.spinalcordmri.org/t/installation-validation-failed-installation-failed/570

### Description

SCT depends on `ivadomed`, which depends on `onnxruntime`, which (as of version `1.5.1`) depends on `libomp`. We encountered this in #2930 and fixed it in our CI by installing `libomp` as part of the config.

https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/.travis.yml#L106-L109

The problem with this solution is that it requires users to install `libomp` on their end, too.

However, separately, the `ivadomed` team fixed this in `v2.5.0` using a different solution: pinning `onnxruntime==1.4.0`. But, SCT currently uses 2.4.0:

https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/requirements.txt#L7

If we upgrade `ivadomed` (or even just pin `onnxruntime==1.4.0` ourselves), then the `libomp` fixes won't be required, and users can install SCT without hassle.

lrouhier:
This PR aims at bumping ivadomed version to 2.5.0. 
This should fix this [reported issue](https://forum.spinalcordmri.org/t/installation-validation-failed-installation-failed/570) by pinning `onnxruntime` to version 1.4.0 
Changes made: 
* update requirements.txt
* Change function call in `scripts/sct_deepseg.py`

fix #3033

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -537,6 +537,10 @@ else
   # Not a package
   print info "Using requirements.txt (git installation)"
   pip install -r requirements.txt
+  yes | pip uninstall tensorflow-tensorboard
+  yes | pip uninstall tensorboard
+  pip install tensorboard
+
 fi
 if [[ $? != 0 ]]; then
   die "Failed running pip install: $?"
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,6 @@ scipy
 scikit-image
 scikit-learn
 tensorflow==1.5.0
-tensorboard==2.3.0
 # PyTorch's Linux/Windows distribution is very large due to its GPU support,
 # but we only need that for training models. For users, use the CPU-only version
 # (only available directly from the PyTorch project).
@@ -37,3 +36,4 @@ transforms3d
 urllib3[secure]
 pytest_console_scripts
 pytest-xdist
+tensorboard==2.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ futures
 # h5py is pinned to minor than 3 due to issues with Keras/TF
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
-ivadomed==2.4.0
+ivadomed==2.5.0
 Keras==2.1.5
 matplotlib
 nibabel
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,8 @@ futures
 # h5py is pinned to minor than 3 due to issues with Keras/TF
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
-ivadomed==2.5.0
 Keras==2.1.5
+ivadomed==2.5.0
 matplotlib
 nibabel
 numpy
@@ -21,6 +21,7 @@ scipy
 scikit-image
 scikit-learn
 tensorflow==1.5.0
+tensorboard==2.3.0
 # PyTorch's Linux/Windows distribution is very large due to its GPU support,
 # but we only need that for training models. For users, use the CPU-only version
 # (only available directly from the PyTorch project).
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,4 +40,3 @@ transforms3d
 urllib3[secure]
 pytest_console_scripts
 pytest-xdist
-tensorboard==2.3.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ futures
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.1.5
-ivadomed==2.5.0
+ivadomed==2.6.0
 matplotlib
 nibabel
 numpy
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -15,7 +15,7 @@ import argparse
 import os
 import sys
 
-import ivadomed as imed
+from ivadomed import inference as imed_inference
 import nibabel as nib
 
 import spinalcordtoolbox as sct
@@ -145,7 +145,7 @@ def main(argv):
 
         # Call segment_nifti
         options = {**vars(args), "fname_prior": fname_prior}
-        nii_seg = imed.inference.segment_volume(path_model, args.i, options=options)
+        nii_seg = imed_inference.segment_volume(path_model, args.i, options=options)
 
         # Save output seg
         if 'o' in options and options['o'] is not None:
--- a/scripts/sct_deepseg.py
+++ b/scripts/sct_deepseg.py
@@ -145,7 +145,7 @@ def main(argv):
 
         # Call segment_nifti
         options = {**vars(args), "fname_prior": fname_prior}
-        nii_seg = imed.utils.segment_volume(path_model, args.i, options=options)
+        nii_seg = imed.inference.segment_volume(path_model, args.i, options=options)
 
         # Save output seg
         if 'o' in options and options['o'] is not None:


jcohenadad:
@lrouhier please cross-ref https://github.com/neuropoly/spinalcordtoolbox/issues/3033

joshuacwnewton:
With this fix, `libomp` is no longer required in our CI, as well. So, I believe we can drop all instances of this:

https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/.travis.yml#L106-L109

lrouhier:
> With this fix, `libomp` will no longer be required in our CI, as well. So, I believe we can drop all instances of this:
> 
> https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/.travis.yml#L106-L109

I'll do it once the first travis complete (I want to see if I forgot anything through the test)

lrouhier:
There seems to be a mismatch between two packages on tensorboard.. I am looking into it. (see [this log](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/743004134)

joshuacwnewton:
### Cause

I believe the issue is due to a conflict between `tensorboard==2.3.0` and `tensorflow-tensorboard==1.5.1`.

 _(`tensorflow-tensorboard` is a [deprecated version of tensorboard](https://pypi.org/project/tensorflow-tensorboard/), but it gets installed alongside `tensorflow==1.5.0`, the version that SCT uses.)_

I don't think this is a new issue... I think it was always lurking. But, when the `ivadomed` call was changed, the underlying issue was uncovered.

### Steps to reproduce

```sh
# Open a fresh py3.6 venv/conda environment
$ python3 -m venv ./venv
$ source venv/bin/activate

# Install ivadomed by itself (which installs just tensorboard==2.3.0)
(venv) $ pip install ivadomed
# Try the import from this PR (Works OK!)
(venv) $ ipython3
In [1]: from ivadomed import inference as imed_inference

# Install outdated tensorflow version (which installs tensorflow-tensorboard==1.5.1)
(venv) $ pip install tensorflow==1.5.0
# Try the import from this PR again (fails)
(venv) $ ipython3
In [1]: from ivadomed import inference as imed_inference

<Long output>

File "/home/joshua/Repos/spinalcordtoolbox/python/envs/spinalcordtoolbox/lib/python3.6/site-packages/tensorboard/compat/__init__.py", line 28, in <module>
    import tensorboard.lazy as _lazy
AttributeError: module 'tensorboard' has no attribute 'lazy'
```

See also https://github.com/tensorflow/tensorboard/issues/1862#issuecomment-521874990

joshuacwnewton:
I fear this new issue will roadblock is from quickly solving the core issue, #3033. So I think it might be quicker to simply pin `onnxruntime==1.4.0` and leave the `ivadomed` calls as they are for now.

That would let us get the patch out quickly, which would then give us more time to more thoroughly explore this new  `tensorboard` conflict issue.

joshuacwnewton:
@lrouhier I think the `tensorflow-tensorboard` uninstall idea went OK. But, the CI is showing [a new error](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/744522455#L3746). I was curious so I started investigating. 

If I understand correctly, the relevant model (`t2star_sc.pt`) was trained using an older version of ivadomed (`<2.5.0`). But the model is loaded using `ivadomed==2.5.0`. And, the `is_2d` attribute was added in https://github.com/ivadomed/ivadomed/pull/491, between `2.4.0` and `2.5.0`.

lrouhier:
yes indeed. I was checking a bit and it is a known retro compatibility issue that we will fix in ivadomed actually (maybe v2.5.1 will integrate this) See here https://github.com/ivadomed/ivadomed/issues/512. Otherwise do you think this fix would be ok between now and the time we remove keras?

install_sct
```diff
@@ -537,6 +537,10 @@ else
   # Not a package
   print info "Using requirements.txt (git installation)"
   pip install -r requirements.txt
+  yes | pip uninstall tensorflow-tensorboard
```
joshuacwnewton, line 540:
```suggestion
  # `tensorflow-tensorboard` is installed by `tensorflow==1.5.0` but is not needed, 
  # and conflicts with the other installation of `tensorboard`. So, we uninstall here.
  # This is hacky, and should be removed as soon as we stop using `tensorflow==1.5.0`.
  # See https://github.com/neuropoly/spinalcordtoolbox/issues/3035 for more info.
  yes | pip uninstall tensorflow-tensorboard
```

install_sct
```diff
@@ -537,6 +537,10 @@ else
...
+  yes | pip uninstall tensorboard
+  pip install tensorboard
```
joshuacwnewton, line 542: Also, do you know if reinstalling `tensorboard` is absolutely necessary?

joshuacwnewton (commented):
> Otherwise do you think this fix would be ok between now and the time we remove keras?

The fix is a bit hacky, so I'm still a little hesitant. I think at the very least we should add a comment to clearly explain the changes. Something like what I've suggested below.

But, I would also want to ask @Drulex or @kousu, as I value their opinions on packaging.

install_sct
lrouhier: Yes actually, I think uninstalling `tensorflow-tensorboard` leaves some dependencies for tensorboard, but they are outdated.. Maybe an upgrade would be enough?

install_sct
Drulex: I agree that this is hacky. Are we doing this to avoid telling the user to install `libomp`?

lrouhier: No we are doing this to use ivadomed 2.5 which is needed in #3041 for example.

Drulex: Definitely cleaner to remove the `tensorflow==1.5.0` dependency and update code to use the latest version. Is there a reason we need `1.5.0`?

Alternatively I have seen somewhere that it is also possible to do:
```python
# in SCT before tensorflow is ever imported
import pkg_resources
pkg_resources.require("tensorflow==1.5.0")
import tensorflow as tf15 # import tensorflow could also probably work? TBD

tf15.do_something()
```
Which should avoid the namespace clash with `ivadomed`.

To be honest I think it's better to update the code unless we have a strong incentive to maintain compatibility with `tensorflow==1.5.0`. It just makes our lives harder.

joshuacwnewton: We pin `tensorflow==1.5.0` to support older CPUs that do not support AVX, see #2618. More info about the `tensorboard` clash in #3035.

The trouble with waiting until we move away from `tensorflow` is that it's quite a large effort (I think), see #3046. So it may take some time, and cause us to fall out of sync with `ivadomed`'s development.

lrouhier: Yes there is a reason explained in #2618. We will eventually get rid of it but it will take some time and might prevent current development.

Drulex: Ah ok I understand. This is because `pip` fetches a pre-built version of tensorflow that is compiled with the AVX instruction set. Computers that don't have this instruction available should use a version that was built without the `AVX` compilation flag. 

For reference: https://github.com/tensorflow/tensorflow/issues/24548

Looking at the wiki for AVX it appears to have been introduced a while ago (2011): https://en.wikipedia.org/wiki/Advanced_Vector_Extensions which may explain why `tensorflow` don't even bother with the non-AVX binary.

In any case it is definitely possible to compile a recent version of `tensorflow` without the AVX CPU flag. Question is do we want to bother with that (or put the burden on the user, which I already know @jcohenadad is opposed to :smile: )

Maybe we can find some pre-compiled binaries online for all platforms.

lrouhier:
Errors in the test are corrected in #3041, which the first PR that needs the ivadomed upgrade. Maybe the discussion will need to be moved there.

Issue #3035, opened 2020-11-11
2 versions of tensorboard get installed with SCT


joshuacwnewton:
When SCT is installed, 2 versions of tensorboard get installed:

* ~~`tensorboard==2.3.0`~~ `tensorboard==2.4.0` (i.e. the latest version), which gets installed [because of `ivadomed`](https://github.com/ivadomed/ivadomed/blob/master/requirements.txt#L13).
* `tensorflow-tensorboard==1.5.1`, which is ["a relic of a bygone era; there’s no need to install it anymore."](https://github.com/tensorflow/tensorboard/issues/1862#issuecomment-521876133)
    * The reason the older, differently-named `tensorflow-tensorboard` gets installed is because it's a dependency for `tensorflow==1.5.0`. (The newer-named `tensorboard` was only [introduced in Tensorflow 1.6.0+](https://pypi.org/project/tensorflow-tensorboard/).)

**This has the potential to create conflicts.** For example, it is currently roadblocking #3034. An unrelated import from `ivadomed` triggers a new exception [in this CI build](https://travis-ci.org/github/neuropoly/spinalcordtoolbox/jobs/743004132#L2003):

```
unit_testing/test_deepseg.py:16: in <module>
    import sct_deepseg
scripts/sct_deepseg.py:18: in <module>
    from ivadomed import inference as imed_inference
python/envs/venv_sct/lib/python3.6/site-packages/ivadomed/inference.py:16: in <module>
    from ivadomed import training as imed_training
python/envs/venv_sct/lib/python3.6/site-packages/ivadomed/training.py:13: in <module>
    from torch.utils.tensorboard import SummaryWriter
python/envs/venv_sct/lib/python3.6/site-packages/torch/utils/tensorboard/__init__.py:2: in <module>
    from tensorboard.summary.writer.record_writer import RecordWriter  # noqa F401
python/envs/venv_sct/lib/python3.6/site-packages/tensorboard/summary/__init__.py:31: in <module>
    from tensorboard.summary import v2
python/envs/venv_sct/lib/python3.6/site-packages/tensorboard/summary/v2.py:24: in <module>
    from tensorboard.plugins.audio.summary_v2 import audio
python/envs/venv_sct/lib/python3.6/site-packages/tensorboard/plugins/audio/summary_v2.py:30: in <module>
    from tensorboard.compat import tf2 as tf
[...]

PR #3037, opened 2020-11-12
Pin onnxruntime==1.4.0 to avoid libomp issue on macOS

### LINKED ISSUE #3033: SCT will not install correctly on macOS due to libomp dependency of onnxruntime v1.5.1

joshuacwnewton:
### Related Issues/PRs

Fixes #3033.
Alternate solution compared to #3034.

### Description

I believe #3034 is the better long-term solution. But, that PR is blocked by #3035. 

Because of the urgent need for a patch, this PR applies a quicker solution that is easier to test/verify.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,27 +103,14 @@ matrix:
       # runs on all branches
       os: osx  # list of OSX env: https://docs.travis-ci.com/user/reference/osx/
       osx_image: xcode12.2
-      addons:
-        homebrew:
-          packages:
-            - libomp  # Needed for onnxruntime Python package, see issue #2930
     - name: "OSX 10.14 (Mojave)"
       if: branch = release or type = cron
       os: osx
       osx_image: xcode11.3
-      addons:
-        homebrew:
-          packages:
-            - libomp  # Needed for onnxruntime Python package, see issue #2930
     - name: "OSX 10.13 (High Sierra)"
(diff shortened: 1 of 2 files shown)


jcohenadad (approved):
thanks for the quick fix 👍

PR #3041, opened 2020-11-12
Support of multichannel and multiclass models

andreanne-lemay:
This PR aims to support multichannel models (e.i., more than one image as input) and multiclass models (e.i., 
generate as many images as there are classes).

New argument `-r` to remove (or keep) temporary files for cascaded architecture has been added.

There is one issue we need to consider for multichannel models. Multichannel models are trained with the modalities presented in a certain order (e.g., T1w, T2w vs T2w, T1w). Changing this order impacts performance. Each multichannel model will have its own modalities and order. There are multiple options to deal with this:
- The proper order is indicated in the task description and we trust the user to read the description and put the input images in the right order.
- The input images can be put in any order, but the name of the modality needs to be in the filename (e.g., `_T1w` or `_T2w`). This way we can order them with the filename. Once again, we trust the user to read this instruction and follow it.
- Add parameters to identify each modality. Instead of using `-i` the user would use `-t2 img_a.nii.gz  -t1 img_b.nii.gz`. We might end up with a lot of parameters if we have more models with more modalities (would require one parameter per modality).
- Add a parameter to identify the order of the modalities. `-i img_a.nii.gz img_b.nii.gz -modality-order t2 t1`.

Personally, I would go towards the last option. If the user forgets to put the modality order and there is more than one input file, we can throw an error. This solution seems to be less prone to errors on the user's side.

Sister PR: https://github.com/ivadomed/ivadomed/pull/524


charleygros:
(On my TODO list today, sorry for delay)

scripts/sct_deepseg.py
```diff
@@ -69,6 +70,11 @@ def get_parser():
              "(more info at https://github.com/sct-pipeline/deepseg-threshold).",
         metavar=Metavar.float,
         default=0.9)
+    misc.add_argument(
+        "-r",
+        type=int,
+        help="Whether to remove temporary files. 0 = no, 1 = yes (default: 1)",
+        default=1)
```
charleygros, line 77:
```suggestion
        help="Remove temporary files.",
        choices=(0, 1),
        default=1)
```

scripts/sct_deepseg.py
```diff
@@ -37,6 +37,7 @@ def get_parser():
     input_output = parser.add_argument_group("\nINPUT/OUTPUT")
     input_output.add_argument(
         "-i",
+        nargs="+",
         help="Image to segment.",
```
charleygros, line 41:
```suggestion
        help='Image to segment. Can be multiple images (separated with space).',
```

scripts/sct_deepseg.py
```diff
@@ -145,19 +152,25 @@ def main(argv):
...
+        nii_lst, target_lst = imed_inference.segment_volume(path_model, args.i, options=options)
 
-        # Save output seg
-        if 'o' in options and options['o'] is not None:
-            fname_seg = options['o']
-        else:
-            fname_seg = ''.join([sct.image.splitext(args.i)[0], '_seg.nii.gz'])
+        # Delete intermediate outputs
+        if fname_prior and os.path.isfile(fname_prior) and args.r:
+            os.remove(fname_prior)
```
charleygros, line 159:
```suggestion
            logger.info("Remove temporary files...")
            os.remove(fname_prior)
```

charleygros:
In favor of option 4 as well! At least for now / this PR.
Maybe using `-c` instead of `-modalities` to be more consistent with other SCT functions?

The sub-parser thing could def be a really nice option in the (near?) future when multi input models will be more common in SCT.

scripts/sct_deepseg.py
```diff
@@ -37,7 +40,15 @@ def get_parser():
...
     input_output.add_argument(
         "-i",
-        help="Image to segment.",
+        nargs="+",
+        help="Image to segment. Can be multiple images (separated with space).",
+        metavar=Metavar.file)
+    input_output.add_argument(
+        "-c",
+        nargs="+",
+        help="Type of image contrast. Indicates the order in which the images have been presented with -i. "
```
jcohenadad, line 49: i understand the need for this argument to be mandatory in case of 2+ input files, but why making it mandatory for single input? As of now, the contrast information is already included in `-task`, e.g. `-task seg_sc_t2star`.

Also, in the future, we might have models that accommodate several contrasts, in which case inputting the contrast will be irrelevant.

andreanne-lemay: It is not mandatory if there is only one input (see [here](https://github.com/neuropoly/spinalcordtoolbox/pull/3041/files#diff-3cb3eafa0ea1f4fbd635239caf0ecf415e67062f0e35a638b72cecb8cb1dc892R152)). I can explicit this in the help of this flag.

Currently, if there are more than 1 contrast the flag `-c` is required, so for now it doesn't accommodate the fact that we might have a HeMIS model (optional multichannel model). It is not too late to choose option 1 (no flag `-c`) and trust that the user will read the models' description.

jcohenadad: OK that makes sense. I got confused by why you would [modify our current models](https://github.com/neuropoly/spinalcordtoolbox/blob/25f60b28f032e98872c27880e8ba64b9b900507e/spinalcordtoolbox/deepseg/models.py#L20) which only require a single input, but you're probably doing it to accommodate potential future cascading of these existing models. So i agree with these proposed changes 👍 

And yes: mentioning that `-c` is only required for tasks with 2+ models would be useful.

andreanne-lemay: > OK that makes sense. I got confused by why you would [modify our current models](https://github.com/neuropoly/spinalcordtoolbox/blob/25f60b28f032e98872c27880e8ba64b9b900507e/spinalcordtoolbox/deepseg/models.py#L20) which only require a single input, but you're probably doing it to accommodate potential future cascading of these existing models.

Yes exactly and also to display the input contrast(s) needed with `-list-tasks`.


> And yes: mentioning that `-c` is only required for tasks with 2+ models would be useful.

I added the info here: 25f60b2

scripts/sct_deepseg.py
```diff
@@ -117,18 +135,30 @@ def main(argv):
...
 
+    # Check if all input images are provided
+    required_contrasts = deepseg.models.get_required_contrasts(args.task)
+    if len(args.i) != len(required_contrasts):
+        parser.error("You need to provide all required input files for the task {}. Contrasts {} are required."
+                     .format(args.task, ', '.join(required_contrasts)))
+
+    # Check modality order
+    if len(args.i) > 1 and args.c is None:
+        parser.error("You need to specify the order in which you put the contrasts in the input images with flag -c.")
```
charleygros, line 154:
```suggestion
        parser.error("Please specify the order in which you put the contrasts in the input images (-i) with flag -c, e.g., -c t1 t2")
```

scripts/sct_deepseg.py
```diff
@@ -143,26 +173,46 @@ def main(argv):
             if not deepseg.models.is_valid(path_model):
                 parser.error("The input model is invalid: {}".format(path_model))
 
+        # Order input images
+        if args.c is not None:
+            input_filenames = []
+            for required_contrasts in deepseg.models.MODELS[name_model]['contrasts']:
+                for provided_contrasts, input_filename in zip(args.c, args.i):
+                    if required_contrasts == provided_contrasts:
```
charleygros, line 181:
```suggestion
            for required_contrast in deepseg.models.MODELS[name_model]['contrasts']:
                for provided_contrast, input_filename in zip(args.c, args.i):
                    if required_contrast == provided_contrast:
```
Am I being picky? simply because `required_contrasts` variable already exists.. :3

charleygros (approved):
Looks great to me! Minor comments

unit_testing/test_deepseg.py
```diff
@@ -58,4 +58,4 @@ def test_segment_nifti(fname_image, fname_seg_manual, fname_out, task,
     assert os.path.isfile(fname_out)
     # Compare with ground-truth segmentation
     assert np.all(nibabel.load(fname_out).get_fdata() ==
-                  nibabel.load(fname_seg_manual).get_fdata())
+                  nibabel.load(fname_seg_manual).get_fdata()[..., 0])
```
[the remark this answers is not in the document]
andreanne-lemay: This PR changed the format of saved predictions. All classes used to be saved in a single file (4D), but this PR saves as many files as prediction classes. This explains why the "GT" (which is a copy of the output of the model made with a previous version of SCT/ivadomed) is 4D and why we need this extra code to make the test pass.

install_sct
```diff
@@ -538,6 +538,10 @@ else
   # Not a package
   print info "Using requirements.txt (git installation)"
   pip install -r requirements.txt
+  yes | pip uninstall tensorflow-tensorboard
```
andreanne-lemay, line 541: See #3034 for details

PR #3060, opened 2020-11-24
Check that there is two inputs for initz

### LINKED ISSUE #626: should check that there is two inputs for initz
```
sct_label_vertebrae -i t2.nii.gz -seg t2_seg_manual.nii.gz -initz 120 -v 2

Create temporary folder...
main
mkdir tmp.151024191808/

Copying input data to tmp folder...
sct_convert -i t2.nii.gz -o tmp.151024191808/data.nii
sct_convert -i t2_seg_manual.nii.gz -o tmp.151024191808/segmentation.nii.gz

Create label to identify disc...
  File "/Users/julien/code/spinalcordtoolbox/bin/sct_label_vertebrae", line 649, in <module>
    main()
  File "/Users/julien/code/spinalcordtoolbox/bin/sct_label_vertebrae", line 130, in main
    create_label_z('segmentation.nii.gz', initz[0], initz[1])  # create label located at z_center
IndexError: list index out of range
```

lrouhier:
#### GitHub

#### PR contents

## Description

This PR modifies sct_label_vertebrae. the `-initz` argument takes 2 inputs (position on IS axis, label value). The current version of SCT does not check if the users gave the 2 inputs before starting the program. 
Done here: 
* Raise an error if argument `-initz` does not contain 2 inputs

## Linked issues

fix #626


joshuacwnewton (changes requested):
It's too bad that sct_label_vertebrae uses comma-delimited values here. (`-initz 1,2` instead of `-initz 1 2`). If it were space-delimited, then we could fix this more cleanly using `nargs=2`.

But, moving away from comma-delimiting would be a breaking change, and it is used across many scripts, so... alas!

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -12,3 +13,13 @@ def test_sct_label_vertebrae_backwards_compat(script_runner):
...
+
+
+def test_sct_label_vertebrae_initz_error():
+    a = 'sct_label_vertebrae -i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -initz 40'
+    proc = subprocess.Popen(a.split(' '),
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            )
+    out, err = proc.communicate()
+    assert 'ValueError: --initz takes two arguments: position in superior-inferior direction, label value' in str(err)
```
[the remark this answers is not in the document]
joshuacwnewton: I've updated the [Programming: Tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming:-Tests) page a bit. Let me know if anything's unclear. :)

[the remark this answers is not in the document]
lrouhier: I do think it looks better now, Thanks for the help. I think the instruction are fairly clear, but an example (like 'look a test_sct_run_batch') could help understand better :)

jcohenadad:
> But, moving away from comma-delimiting would be a breaking change, and it is used across many scripts, so... alas!

yes, although this particular feature is not, so we could also consider breaking it (if we bump SCT to a minor release)

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -1,7 +1,8 @@
 from pytest_console_scripts import script_runner
 import pytest
 import logging
-
+import subprocess
```
joshuacwnewton, line 4: I don't think `subprocess` is needed anymore. :slightly_smiling_face:

joshuacwnewton (changes requested):

unit_testing/cli/test_cli_sct_label_vertebrae.py
lrouhier: Yes right!

lrouhier: Done :)

lrouhier:
> yes, although this particular feature is not, so we could also consider breaking it (if we bump SCT to a minor release)

I would need to change the code/test a bit and test but it might be cleaner... I don't have a strong opinion on that, I guess it more that it would be inconsistent with the rest of sct (for example `sct_label_utils -create-viewer` takes a comma-delimited list). I guess this would start a refactoring.

joshuacwnewton:
> my only problem is that it would be inconsistent with the rest of sct (for example sct_label_utils -create-viewer takes a comma-delimited list).

I was just about to type the same thing. I think if we had `,`/`:` for some arguments and spaces for others, it'd be a little harder to keep track of the proper usage from both a user perspective and a developer perspective. 

[I've edited the  "Documentation" dev wiki page to mention this, at the very least.](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming:-Documentation#delimiting-for-list-arguments) I think future refactor could be a nice-to-have, but I think it would require some discussion, maybe in a separate issue.

joshuacwnewton (approved):

PR #3069, opened 2020-12-01
Documentation fixes

### LINKED ISSUE #3056: Python API documentation page needs to be updated following recent refactors
We use [automodule + manually listed module entries](https://github.com/neuropoly/spinalcordtoolbox/blob/master/documentation/source/dev_section/api.rst) for our Sphinx-generated API documentation. These need to be updated whenever there are changes to module structure. 

Right now, the documentation isn't quite correct (e.g. [`spinalcordtoolbox.utils`](https://spinalcordtoolbox.com/en/stable/dev_section/api.html#module-spinalcordtoolbox.utils)), so I think this might have been missed with recent refactors (`labels.py` as well as, `utils.py` becoming `utils/fs.py`, `utils/shell.py`, `utils/sys.py`).

Drulex:
Closes #3056

### DIFF
--- a/documentation/source/dev_section/api.rst
+++ b/documentation/source/dev_section/api.rst
@@ -174,6 +174,15 @@ spinalcordtoolbox.labels
 .. automodule:: spinalcordtoolbox.labels
    :members:
 
+Spinal Cord Flattening
+**********************
+
+spinalcordtoolbox.flattening
+============================
+
+.. automodule:: spinalcordtoolbox.flattening
+   :members:
+
 
 Motion Correction
 *****************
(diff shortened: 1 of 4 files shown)


Drulex:
Haven't performed an exhaustive check of a API documentation since I don't think a lot of people actually use it, but I went over the recently added modules that were missing.

joshuacwnewton (changes requested):
I notice the spacing issues that you fixed is also present in `math`. (e.g. https://spinalcordtoolbox--3069.org.readthedocs.build/en/3069/dev_section/api.html#spinalcordtoolbox.math.compute_similarity) Now might be a good time to fix this, too.

I'm surprised Sphinx doesn't emit a warning for this!

joshuacwnewton (approved):

PR #3070, opened 2020-12-01
Set PYTHONNOUSERSITE=True to prevent user site packages from interfering

### LINKED ISSUE #3067: sys.path will use user site-packages over SCT conda environment's site-packages
### Context

This concerns the post **[AttributeError: ‘str’ object has no attribute ‘decode’ - v5.0.0](https://forum.spinalcordmri.org/t/attributeerror-str-object-has-no-attribute-decode-v5-0-0/567/)** from the forum.

The user in the post above has installed packages to their [user site directory](https://www.python.org/dev/peps/pep-0370/#specification). Despite SCT using a conda environment, the user site packages packages take precedent over the packages SCT installs. The specific error in the forum post happens because SCT's `h5py==2.10.0` is ignored in favor of the user's `h5py==3.1.0`. There are other package mismatches (like incorrect `pytorch` versions) which may have further consequences.

This is caused by https://github.com/conda-forge/python-feedstock/issues/171. (See also: the linked issues in the timeline for that issue.)

The best description of the issue I found was in https://github.com/mozilla/OpenWPM/issues/689.

### Discussion

1. **How can we detect this earlier?**

The reason this wasn't caught in the installation script is because we set the [`PYTHONNOUSERSITE` environment variable](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONNOUSERSITE) here:

https://github.com/neuropoly/spinalcordtoolbox/blob/d9c2d4df9c271f68c454294d576dd3ba03e9627c/install_sct#L502-L504

This change only takes effect within the scope of `./install_sct`. So, when we run `sct_check_dependencies` as part of the installation, everything _looks_ okay. But, as soon as the user finishes the installation and tries to use an SCT script, the user site packages are used, and the problem reveals itself.

The fact that `PYTHONNOUSERSITE` is set just hides the underlying problem. So, I think it might be good to unset this right before `sct_check_dependencies` is run to better reflect the user's actual environment and allow us to detect this issue.

2. **How can we prevent the user site directory from interfering long-term?**

The most prominent idea from https://github.com/conda-forge/python-feedstock/issues/171 is [to `export PYTHONNOUSERSITE=True`](https://github.com/conda-forge/python-feedstock/issues/171#issuecomment-374261534). This could be done, for example, by adding this to the user's `.bashrc`. But, this would affect more than just SCT, however, so perhaps there's a cleaner solution. 

This also ties into the ongoing discussions re: how SCT installs itself. (Bringing its own `conda` environment vs. pip installing, etc.) So, #1526 is definitely relevant here.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR sets an environment variable to prevent user site packages from being used over SCT's installed package versions.

Re: the two unchecked checklist items, I don't think there's any relevant tests/documentation that could be added regarding this change. This is a difficult thing to test! 

## Linked issues

Fixes #3067.

### DIFF
--- a/spinalcordtoolbox/compat/launcher.py
+++ b/spinalcordtoolbox/compat/launcher.py
@@ -24,6 +24,9 @@ def main():
     if "ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS" not in os.environ:
         env["ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS"] = str(multiprocessing.cpu_count())
 
+    # Prevent user site packages from interfering with SCT dependencies (See issue #3067)
+    env["PYTHONNOUSERSITE"] = "True"
+
     command = os.path.basename(sys.argv[0])
     pkg_dir = os.path.dirname(sct.__file__)
 


jcohenadad (approved):
Thank you for this tricky investigation and for finding a fix @joshuacwnewton !!

Drulex (approved):
Looks ok to me, but @joshuacwnewton can you confirm that the change doesn't leak into the system global environment variables (i.e if you open another shell to use `python` outside `SCT` the variable `PYTHONNOUSERSITE` isn't set).

joshuacwnewton:
> Looks ok to me, but @joshuacwnewton can you confirm that the change doesn't leak into the system global environment variables (i.e if you open another shell to use `python` outside `SCT` the variable `PYTHONNOUSERSITE` isn't set).

Good thinking. :slightly_smiling_face: 

Luckily, you don't even have to open another shell. Even in the same shell, `PYTHONNOUSERSITE` won't be set, because this is only in the scope of the Python dict `env`, which only gets used in this call to [`os.execvpe`](https://docs.python.org/3/library/os.html#os.execvpe):

 https://github.com/neuropoly/spinalcordtoolbox/blob/9c3dcf6e4b59f8a19578510fb9314b2cd93bc573/spinalcordtoolbox/compat/launcher.py#L41

To double-check, though:

<details>
<summary> Console output from checking PYTHONNOUSERSITE </summary>

```
(venv_sct) joshua@XPS-15-9560:~/repos/spinalcordtoolbox$ sct_check_dependencies

--
Spinal Cord Toolbox (git-jn/3067-set-pythonnousersite-b6c6646404031bd73fe701abb35bd96afbfb6b88)

sct_check_dependencies 
--

SCT info:
- version: git-jn/3067-set-pythonnousersite-b6c6646404031bd73fe701abb35bd96afbfb6b88
- path: /home/joshua/repos/spinalcordtoolbox
OS: linux (Linux-5.4.0-54-generic-x86_64-with-debian-bullseye-sid)
CPU cores: Available: 8, Used by ITK functions: 8
RAM: Total: 7812MB, Used: 5086MB, Available: 2222MB
  Using bundled python 3.6.11 | packaged by conda-forge | (default, Nov 27 2020, 18:57:37) 
[GCC 9.3.0] at /home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/bin/python3.6
Check if dipy is installed..........................[OK] (1.3.0)
Check if ivadomed (2.4.0) is installed..............[OK] (2.4.0)
Check if matplotlib is installed....................[OK] (3.3.3)
Check if numpy is installed.........................[OK] (1.19.4)
Check if pandas is installed........................[OK] (1.1.4)
Check if psutil is installed........................[OK] (5.7.3)
Check if pytest is installed........................[OK] (6.1.2)
Check if pytest-cov is installed....................[OK] (__version__ = '2.10.1')
Check if requests is installed......................[OK] (2.25.0)
Check if scikit-learn is installed..................[OK] (0.23.2)
Check if tqdm is installed..........................[OK] (4.54.0)
Check if urllib3 is installed.......................[OK] (1.26.2)
Check if pytest-xdist is installed..................[OK] (2.1.0)

(venv_sct) joshua@XPS-15-9560:~/repos/spinalcordtoolbox$ echo $PYTHONNOUSERSITE

(venv_sct) joshua@XPS-15-9560:~/repos/spinalcordtoolbox$ 
```
</details>

PR #3072, opened 2020-12-03
Crop mask used by slicereg

charleygros:
When looking at [register_step_ants_slice_regularized_registration](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/registration/register.py#L127), `src` and `dest` are cropped but not the `mask`... which, I think, could lead to some errors.


jcohenadad:
I believe the cause of the issue described in https://github.com/neuropoly/spinalcordtoolbox/issues/3075 is different.

ANTs works in the physical space coordinate so i don't think that matrix (ie vox based coordinates) mismatch is a problem.

PR #3073, opened 2020-12-03
Change Error type so program doesn't quit when labels are too high

### LINKED ISSUE #2607: ValueError thrown if label value is too high
### Description

When trying to run the function in lumbar FOV, if specifying a starting point at label =19 (corresponding to T11-T12 disc), the code crashes with a ValueError. The reason being that the labels on the template are not going high enough.

### Steps to Reproduce

~~~
# here we use sct_testing_data/t2.nii.gz image, even if it is centered in the cervical cord, but the crash is independent from the image content anyways.
cd sct_testing_data/t2
sct_label_vertebrae -i t2.nii.gz -s t2_seg.nii.gz -c t2 -initz 25,19
...
Current disc: 19 (z=50). Direction: superior
.. Peak found: z=0 (correlation = 0.24951264940910967)
.. Switching to inferior direction.
  File "/Users/julien/code/sct/scripts/sct_label_vertebrae.py", line 427, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_label_vertebrae.py", line 372, in main
    verbose=verbose, path_template=path_template, path_output=path_output, scale_dist=scale_dist)
  File "/Users/julien/code/sct/spinalcordtoolbox/vertebrae/core.py", line 222, in vertebral_detection
    current_z = init_disc[0] - list_distance[list_disc_value_template.index(current_disc)]
ValueError: 20 is not in list
~~~

### Possible solution

Increase the number of labels from the template.

lrouhier:
#### GitHub

#### PR contents

## Description
In `sct_label_vertebrae`, there is an exception designed for very high-value labels to provide the user the opportunity to label the lumbar part of the vertebrae. This is needed because our PAM50 template only has labels from 1 to 20. 
This PR changes the error type for the exception. It is needed because the `try` statement throws a `ValueError`.

## Linked issues
fix #2607


joshuacwnewton (dismissed):

joshuacwnewton (commented):
I'm wondering if changes like this should be tested. It would be a somewhat time-consuming test (because `t2_seg.nii.gz` must be generated first) just to check a small error. 

But, maybe that means we should include a `t2_seg.nii.gz` file inside `sct_testing_data`. That way, we could more easily write a quick test for the `vertebral_detection` function to ensure that the warning is raised instead of an error. 

What do you think, @lrouhier?

joshuacwnewton:
Ah, my apologies, there already are segmented spinal cord images in `sct_testing_data`. They're just named differently than `t2_seg.nii.gz`.

In that case, I think it would be good to write a test for this.

lrouhier:
I'll get on it.

lrouhier:
@joshuacwnewton done in 84e9ece

joshuacwnewton (approved):
Thanks much @lrouhier!

PR #3081, opened 2020-12-04
Implement -o flag for a few functions

### LINKED ISSUE #897: add -o flag in sct_smooth_spinalcord

### LINKED ISSUE #1802: Implement -o flag for a few functions
### Description

In addition to the already-existing `-ofolder` for putting the outputs there, it would be useful to also have a `-o` flag to specify output file (not folder). Functions where we could implement that are:
- sct_propseg
- sct_deepseg_sc
- sct_deepseg_gm

lrouhier:
#### GitHub

#### PR contents

## Description
This PR aims at adding the `-o` flag on a few function in sct:
- [x] sct_deepseg_sc
- [x] sct_detect_pmj
- [x] sct_propseg
- [x] sct_smooth_spinalcord

## Linked issues
fix #1802
fix #897

### DIFF
--- a/spinalcordtoolbox/scripts/sct_deepseg_sc.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg_sc.py
@@ -96,7 +96,7 @@ def get_parser():
         default=os.getcwd())
     optional.add_argument(
         "-o",
-        metavar=Metavar.str,
+        metavar=Metavar.file,
         help='Output filename. Example: spinal_seg.nii.gz '),
     optional.add_argument(
         "-r",
--- a/spinalcordtoolbox/scripts/sct_deepseg_sc.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg_sc.py
@@ -95,6 +95,10 @@ def get_parser():
         action=ActionCreateFolder,
         default=os.getcwd())
     optional.add_argument(
+        "-o",
+        metavar=Metavar.str,
+        help='Output filename. Example: spinal_seg.nii.gz '),
+    optional.add_argument(
         "-r",
         type=int,
         help="Remove temporary files.",
@@ -165,6 +169,11 @@ def main():
     else:
         manual_centerline_fname = None
 
+    if args.o is not None:
+        fname_out = args.o
+    else:
+        fname_out = extract_fname(fname_image)[1] + '_seg' + extract_fname(fname_image)[2]
+
     threshold = args.thr
     if threshold is not None:
         if threshold > 1.0 or (threshold < 0.0 and threshold != -1.0):
@@ -192,8 +201,7 @@ def main():
                                      threshold_seg=threshold, remove_temp_files=remove_temp_files, verbose=verbose)
 
     # Save segmentation
-    fname_seg = os.path.abspath(os.path.join(output_folder, extract_fname(fname_image)[1] + '_seg' +
-                                             extract_fname(fname_image)[2]))
+    fname_seg = os.path.abspath(os.path.join(output_folder, fname_out))
     im_seg.save(fname_seg)
 
     # Generate QC report
--- a/spinalcordtoolbox/scripts/sct_deepseg_sc.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg_sc.py
@@ -172,7 +172,8 @@ def main():
     if args.o is not None:
         fname_out = args.o
     else:
-        fname_out = extract_fname(fname_image)[1] + '_seg' + extract_fname(fname_image)[2]
+        path, file_name, ext = extract_fname(fname_image)
+        fname_out = file_name + '_seg' + ext
 
     threshold = args.thr
     if threshold is not None:
--- a/spinalcordtoolbox/scripts/sct_deepseg_sc.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg_sc.py
@@ -89,11 +89,11 @@ def get_parser():
         action=ActionCreateFolder,
         default=os.getcwd())
     optional.add_argument(
-        "-o",
+        '-o',
         metavar=Metavar.file,
         help='Output filename. Example: spinal_seg.nii.gz '),
     optional.add_argument(
-        "-r",
+        '-r',
         type=int,
         help="Remove temporary files.",
         choices=(0, 1),
--- a/spinalcordtoolbox/scripts/sct_detect_pmj.py
+++ b/spinalcordtoolbox/scripts/sct_detect_pmj.py
@@ -75,7 +75,7 @@ def get_parser():
     optional.add_argument(
         "-o",
         metavar=Metavar.file,
-        help='Output filename. Example: spinal_seg.nii.gz '),
+        help='Output filename. Example: pmj.nii.gz '),
     optional.add_argument(
         '-qc',
         metavar=Metavar.str,
--- a/spinalcordtoolbox/scripts/sct_detect_pmj.py
+++ b/spinalcordtoolbox/scripts/sct_detect_pmj.py
@@ -73,6 +73,10 @@ def get_parser():
         action=ActionCreateFolder,
         required=False)
     optional.add_argument(
+        "-o",
+        metavar=Metavar.file,
+        help='Output filename. Example: spinal_seg.nii.gz '),
+    optional.add_argument(
         '-qc',
         metavar=Metavar.str,
         help='The path where the quality control generated content will be saved.',
@@ -101,7 +105,7 @@ def get_parser():
 
 
 class DetectPMJ:
-    def __init__(self, fname_im, contrast, fname_seg, path_out, verbose):
+    def __init__(self, fname_im, contrast, fname_seg, path_out, verbose, fname_out):
 
         self.fname_im = fname_im
         self.contrast = contrast
@@ -124,7 +128,7 @@ class DetectPMJ:
 
         self.threshold = -0.75 if self.contrast == 't1' else 0.8  # detection map threshold, depends on the contrast
 
-        self.fname_out = extract_fname(self.fname_im)[1] + '_pmj.nii.gz'
+        self.fname_out = fname_out
 
         self.fname_qc = 'qc_pmj.png'
 
@@ -283,6 +287,10 @@ def main():
             os.makedirs(path_results)
     else:
         path_results = '.'
+    if arguments.o is not None:
+        fname_o = arguments.o
+    else:
+        fname_o = extract_fname(fname_in)[1] + '_pmj.nii.gz'
 
     path_qc = arguments.qc
 
@@ -297,7 +305,8 @@ def main():
                          contrast=contrast,
                          fname_seg=fname_seg,
                          path_out=path_results,
-                         verbose=verbose)
+                         verbose=verbose,
+                         fname_out=fname_o)
 
     # run the extraction
     fname_out, tmp_dir = detector.apply()
--- a/spinalcordtoolbox/scripts/sct_detect_pmj.py
+++ b/spinalcordtoolbox/scripts/sct_detect_pmj.py
@@ -68,7 +68,7 @@ def get_parser():
         action=ActionCreateFolder,
         required=False)
     optional.add_argument(
-        "-o",
+        '-o',
         metavar=Metavar.file,
         help='Output filename. Example: pmj.nii.gz '),
     optional.add_argument(
@@ -322,4 +322,3 @@ def main(argv=None):
 if __name__ == "__main__":
     init_sct()
     main(sys.argv[1:])
-
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -211,6 +211,10 @@ def get_parser():
         help="Output folder."
     )
     optional.add_argument(
+        "-o",
+        metavar=Metavar.file,
+        help='Output filename. Example: spinal_seg.nii.gz '),
+    optional.add_argument(
         '-down',
         metavar=Metavar.int,
         type=int,
@@ -447,6 +451,11 @@ def propseg(img_input, options_dict):
     if not os.path.exists(folder_output):
         os.makedirs(folder_output)
 
+    if arguments.o is not None:
+        fname_out = arguments.o
+    else:
+        fname_out = os.path.basename(add_suffix(fname_data, "_seg"))
+
     if arguments.down is not None:
         cmd += ["-down", str(arguments.down)]
     if arguments.up is not None:
@@ -620,7 +629,7 @@ def propseg(img_input, options_dict):
         sys.exit(1)
 
     # build output filename
-    fname_seg = os.path.join(folder_output, os.path.basename(add_suffix(fname_data, "_seg")))
+    fname_seg = os.path.join(folder_output,fname_out)
     fname_centerline = os.path.join(folder_output, os.path.basename(add_suffix(fname_data, "_centerline")))
     # in case header was rescaled, we need to update the output file names by removing the "_rescaled"
     if rescale_header is not 1:
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -214,6 +214,7 @@ def get_parser():
         "-o",
         metavar=Metavar.file,
         help='Output filename. Example: spinal_seg.nii.gz '),
+
     optional.add_argument(
         '-down',
         metavar=Metavar.int,
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -434,18 +434,18 @@ def propseg(img_input, options_dict):
 
     # Starting building the command
     cmd = ['isct_propseg', '-t', contrast_type_propseg]
-    
-    folder_output = './'
-    cmd += ['-o', folder_output]
-    if not os.path.isdir(folder_output) and os.path.exists(folder_output):
-        logger.error("output directory %s is not a valid directory" % folder_output)
-    if not os.path.exists(folder_output):
-        os.makedirs(folder_output)
 
     if arguments.o is not None:
         fname_out = arguments.o
     else:
         fname_out = os.path.basename(add_suffix(fname_data, "_seg"))
+    
+    folder_output = os.path.dirname(fname_out)
+    cmd += ['-o', folder_output]
+    if not os.path.isdir(folder_output) and os.path.exists(folder_output):
+        logger.error("output directory %s is not a valid directory" % folder_output)
+    if not os.path.exists(folder_output):
+        os.makedirs(folder_output)
 
     if arguments.down is not None:
         cmd += ["-down", str(arguments.down)]
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -203,8 +203,8 @@ def get_parser():
     optional.add_argument(
         '-o',
         metavar=Metavar.file,
-        help='Output filename. Example: spinal_seg.nii.gz '),
-
+        help='Output filename. Example: spinal_seg.nii.gz '
+        )
     optional.add_argument(
         '-down',
         metavar=Metavar.int,
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -201,12 +201,6 @@ def get_parser():
         help="Show this help message and exit."
     )
     optional.add_argument(
-        '-ofolder',
-        metavar=Metavar.folder,
-        action=ActionCreateFolder,
-        help="Output folder."
-    )
-    optional.add_argument(
         '-o',
         metavar=Metavar.file,
         help='Output filename. Example: spinal_seg.nii.gz '),
@@ -440,11 +434,8 @@ def propseg(img_input, options_dict):
 
     # Starting building the command
     cmd = ['isct_propseg', '-t', contrast_type_propseg]
-
-    if arguments.ofolder is not None:
-        folder_output = arguments.ofolder
-    else:
-        folder_output = './'
+    
+    folder_output = './'
     cmd += ['-o', folder_output]
     if not os.path.isdir(folder_output) and os.path.exists(folder_output):
         logger.error("output directory %s is not a valid directory" % folder_output)
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -207,7 +207,7 @@ def get_parser():
         help="Output folder."
     )
     optional.add_argument(
-        "-o",
+        '-o',
         metavar=Metavar.file,
         help='Output filename. Example: spinal_seg.nii.gz '),
 
@@ -684,4 +684,3 @@ def main(argv=None):
 if __name__ == "__main__":
     init_sct()
     main(sys.argv[1:])
-
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -630,7 +630,7 @@ def propseg(img_input, options_dict):
         sys.exit(1)
 
     # build output filename
-    fname_seg = os.path.join(folder_output,fname_out)
+    fname_seg = os.path.join(folder_output, fname_out)
     fname_centerline = os.path.join(folder_output, os.path.basename(add_suffix(fname_data, "_centerline")))
     # in case header was rescaled, we need to update the output file names by removing the "_rescaled"
     if rescale_header is not 1:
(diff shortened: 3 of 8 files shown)


spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -133,6 +137,10 @@ def main(args=None):
         sigma = arguments.smooth
     remove_temp_files = arguments.r
     verbose = int(arguments.v)
+    if arguments.o is not None:
+        fname_out = arguments.o
+    else:
+        fname_out = extract_fname(fname_anat)[1] + '_smooth.nii.gz'
```
jcohenadad, line 143: please make sure to not change the default behaviour-- in this case if input is NII the output will be NIIGZ (whereas the current default is NII)

spinalcordtoolbox/scripts/sct_deepseg_sc.py
```diff
@@ -165,6 +169,11 @@ def main():
     else:
         manual_centerline_fname = None
 
+    if args.o is not None:
+        fname_out = args.o
+    else:
+        fname_out = extract_fname(fname_image)[1] + '_seg' + extract_fname(fname_image)[2]
```
jcohenadad, line 175: there is probably a more elegant way than launching the same API twice (extract_fname), maybe @joshuacwnewton you have some ideas?

lrouhier: We can run it only once and to get the all the output and then use that.

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
lrouhier: Ah ! Sorry I'll change it.

jcohenadad:
it might be good to create tests for these, no?  e.g. test if -o works as expected-- and check maybe the default behaviour?

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -12,3 +14,9 @@ def test_sct_detect_pmj_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+def test_sct_detect_pmj_o_flag():
```
joshuacwnewton, line 18: Same comments re: subprocess and `tmp_path`.

unit_testing/cli/test_cli_sct_propseg.py
```diff
@@ -12,3 +14,10 @@ def test_sct_propseg_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+def test_sct_propseg_o_flag():
```
joshuacwnewton, line 19: Same comments re: subprocess and `tmp_path`.

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
```diff
@@ -12,3 +14,10 @@ def test_sct_smooth_spinalcord_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+def test_sct_smooth_spinalcord_o_flag():
```
joshuacwnewton, line 19: Same comments re: subprocess and `tmp_path`.

joshuacwnewton (changes requested):
Thanks for your ongoing efforts to tackle all of these older issues, @lrouhier!

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -246,7 +254,7 @@ def main(args=None):
     elapsed_time = time.time() - start_time
     printv('\nFinished! Elapsed time: ' + str(int(np.round(elapsed_time))) + 's\n')
 
-    display_viewer_syntax([file_anat, file_anat + '_smooth'], verbose=verbose)
+    display_viewer_syntax([fname_anat, fname_out], verbose=verbose)
```
joshuacwnewton, line 257: Here's the relevant change for one of my review comments.

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -12,3 +14,11 @@ def test_sct_deepseg_sc_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+def test_sct_deepseg_sc_o_flag():
+    command = 'sct_deepseg_sc -i sct_testing_data/t2/t2.nii.gz -c t2 -o test_seg.nii.gz'
+    subprocess.check_output(command, shell=True)
+    assert os.path.isfile('test_seg.nii.gz')
+    os.remove('test_seg.nii.gz')
```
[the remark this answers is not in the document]
lrouhier: I will look into the tmp_path fixture. 
However for the sct_function.main(args), I do have an issue. As you can see [here](https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_deepseg_sc.py#L130), the sct_deepseg.main() function does not take any argument. Should I change that for the sake of the test ? (this is not the only one)

unit_testing/cli/test_cli_sct_propseg.py
lrouhier: About subprocess: [Here](https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_propseg.py#L657-L676) sct_propseg needs the output of the parser as it is not called inside the 'main' function.

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
lrouhier: [Here](https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_smooth_spinalcord.py#L120-L127), it does not take into consideration if args are passed to main (only if they are passed through the sys)

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
```diff
@@ -235,7 +243,7 @@ def main(args=None):
     # Generate output file
     printv('\nGenerate output file...')
     generate_output_file(os.path.join(path_tmp, "anat_rpi_straight_smooth_curved_nonzero.nii"),
-                         file_anat + '_smooth' + ext_anat)
+                         fname_out)
```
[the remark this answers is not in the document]
lrouhier: They are not equal, actually, the change that you showed in your [comment](https://github.com/neuropoly/spinalcordtoolbox/pull/3081#discussion_r537536763) was necessary due to a mistake. Indeed it gave you the command to open the file but there was no extension. So now it displays the full name.

unit_testing/cli/test_cli_sct_deepseg_sc.py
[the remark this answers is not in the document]
joshuacwnewton: Ahh, good point. It looks like some `main()` functions have been edited to take args but not all. I'll make an issue for this. :)

unit_testing/cli/test_cli_sct_propseg.py
joshuacwnewton: Interesting... thanks for pointing that out! Yes, this is definitely inconsistent between scripts.

unit_testing/cli/test_cli_sct_detect_pmj.py
lrouhier: Tmp_path fixed. [main ](https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_detect_pmj.py#L260) does not take args.

spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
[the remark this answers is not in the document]
joshuacwnewton: Ahhh, I see. Thank you for clarifying.

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
lrouhier: tmp_path fixed

spinalcordtoolbox/scripts/sct_detect_pmj.py
```diff
@@ -283,6 +287,10 @@ def main():
             os.makedirs(path_results)
     else:
         path_results = '.'
+    if arguments.o is not None:
+        fname_o = arguments.o
```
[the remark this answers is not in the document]
lrouhier: Right. I guess we could directly go with fname_o = path_result+arguments.o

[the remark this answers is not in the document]
joshuacwnewton: Hmm... the `DetectPMJ` class is written in a way where that wouldn't work:

https://github.com/neuropoly/spinalcordtoolbox/blob/1b220017c7f48613f3168a33a9dbd7d72eab4bdf/spinalcordtoolbox/scripts/sct_detect_pmj.py#L159-L160

So, `DetectPMJ`'s `fname_out` attribute has to be in the form filename+ext without the path... which makes things a little weird.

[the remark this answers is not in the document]
joshuacwnewton: Maybe this is too much effort for too little gain. :sweat_smile:

unit_testing/cli/test_cli_sct_deepseg_sc.py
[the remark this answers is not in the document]
joshuacwnewton: I've opened #3085. We could either wait for discussion there, or just leave the tests as subprocess for now and add a note. Something like:

```python
# TODO: Change to main(args) call once Issue #3085 has been resolved
subprocess.check_output(command, shell=True)
```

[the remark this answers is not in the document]
joshuacwnewton: #3091 has been opened to fix #3085. 

Once #3091 gets merged, then you will be able to use `sct_deepseg_sc.main` (or the `main` of any SCT script!)

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -12,3 +14,11 @@ def test_sct_deepseg_sc_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+def test_sct_deepseg_sc_o_flag():
```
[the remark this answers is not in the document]
jcohenadad: > Perhaps in a new issue?

sounds good to me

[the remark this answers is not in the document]
joshuacwnewton: Done in #3093.

jcohenadad:
@lrouhier @joshuacwnewton what is the status of this PR, can it be updated and merged?

joshuacwnewton:
> @lrouhier @joshuacwnewton what is the status of this PR, can it be updated and merged?

I think it's very close to done, and was just waiting on #3091 to allow relevant sct scripts to be tested using `main()` (https://github.com/neuropoly/spinalcordtoolbox/pull/3081#discussion_r541060418). Since #3091 now has approval, I can merge it, and this one can be finished up.

lrouhier:
@joshuacwnewton, I think this is fairly ready now. I'm just waiting on the test but everything should run smoothly. If you could review the PR it would be great!

lrouhier:
The sct_propseg test is not using the tmp_path fixture because there was an error with the `mv` function from  spinalcordtoolbox.utils.fs see [this log](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495).

joshuacwnewton:
> The sct_propseg test is not using the tmp_path fixture because there was an error with the `mv` function from spinalcordtoolbox.utils.fs see [this log](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495).

Hmm... I don't think this is necessarily a problem with `tmp_path`. Instead, I think this means we [should be using `shutil.move`](https://stackoverflow.com/a/43967659) instead of `os.rename`.

I'm going to try it out in a quick PR and if all goes well then `tmp_path` should be usable here too.

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -12,3 +15,10 @@ def test_sct_deepseg_sc_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+def test_sct_deepseg_sc_o_flag(tmp_path):
+    command = """-i sct_testing_data/t2/t2.nii.gz -c t2 -o""" + str(os.path.join(str(tmp_path), 'test_seg.nii.gz'))
+    sct_deepseg_sc.main(command.split())
+    assert os.path.isfile(os.path.join(str(tmp_path), 'test_seg.nii.gz'))
```
joshuacwnewton, line 23: `tmp_path` returns a `pathlib.Path` object, which [has some useful properties](https://realpython.com/python-pathlib/). You can use `/` to combine paths instead of `os.path.join`! And, when combined with f-strings (`f""`), you don't need any of the `str()` conversion, so it can be written in a short way:

```suggestion
    tmp_filepath = tmp_path/'test_seg.nii.gz'
    command = f"-i sct_testing_data/t2/t2.nii.gz -c t2 -o {tmp_filepath}"
    sct_deepseg_sc.main(command.split())
    assert os.path.isfile(tmp_filepath)
```

joshuacwnewton: This same refactor could apply to the other tests, too. :)

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -1,6 +1,9 @@
 from pytest_console_scripts import script_runner
 import pytest
 import logging
+import os
+import subprocess
+from spinalcordtoolbox.scripts import sct_deepseg_sc as sct_deepseg_sc
```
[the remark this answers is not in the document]
joshuacwnewton: Ditto for the other tests!

joshuacwnewton (changes requested):
I have a few small style nitpicks.

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -12,3 +15,8 @@ def test_sct_detect_pmj_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+def test_sct_detect_pmj_o_flag(tmp_path):
+    command = '-i sct_testing_data/template/template//PAM50_small_t2.nii.gz -igt sct_testing_data/template/template/PAM50_small_t2_pmj_manual.nii.gz  -c t2 -o' + os.path.join(str(tmp_path), 'test_pmj.nii.gz')
+    sct_detect_pmj.main(command.split())
```
kousu, line 21: Heyo, how's it gong? I'm parachuting in here from https://github.com/neuropoly/spinalcordtoolbox/pull/3146 to understand context. I noticed here that this split() isn't really safe. It's not quite as prone to problems as `os.system()` but it's still vulnerable to mistakes. If some system sometime happens to have spaces in its `$TMPDIR` then this code is going to have a bad time.

The thing that people usually do with `subprocess.run()` is to pass ARGV explicitly, and I think that's a better option here:

```suggestion
    argv = ['-i', 'sct_testing_data/template/template//PAM50_small_t2.nii.gz', '-igt','sct_testing_data/template/template/PAM50_small_t2_pmj_manual.nii.gz', '-c', 't2', '-o', os.path.join(str(tmp_path), 'test_pmj.nii.gz')]
    sct_detect_pmj.main(argv)
```

I see you already removed this tmp_path from the other test, and that sidesteps the issue: the string is static so it can't have extra spaces slip in to mess it up. But it would be more reliable to just make it a list from the start.

spinalcordtoolbox/scripts/sct_deepseg_sc.py
```diff
@@ -94,6 +94,10 @@ def get_parser():
         help='Output folder. Example: My_Output_Folder/ ',
         action=ActionCreateFolder,
         default=os.getcwd())
+    optional.add_argument(
+        "-o",
```
kousu, line 98: All the other options use single quotes

```suggestion
        '-o',
```

spinalcordtoolbox/scripts/sct_detect_pmj.py
```diff
@@ -72,6 +72,10 @@ def get_parser():
         help='Output folder. Example: My_Output_Folder/',
         action=ActionCreateFolder,
         required=False)
+    optional.add_argument(
+        "-o",
```
kousu, line 76:
```suggestion
        '-o',
```

spinalcordtoolbox/scripts/sct_deepseg_sc.py
```diff
@@ -94,6 +94,10 @@ def get_parser():
...
+        metavar=Metavar.file,
+        help='Output filename. Example: spinal_seg.nii.gz '),
     optional.add_argument(
         "-r",
```
kousu, line 102: oh, except not this one line? but the other scripts all do use single quotes.

```suggestion
        '-r',
```

spinalcordtoolbox/scripts/sct_propseg.py
```diff
@@ -210,6 +210,11 @@ def get_parser():
         action=ActionCreateFolder,
         help="Output folder."
     )
+    optional.add_argument(
+        "-o",
```
kousu, line 214:
```suggestion
        '-o',
```

spinalcordtoolbox/scripts/sct_propseg.py
```diff
@@ -450,6 +455,11 @@ def propseg(img_input, options_dict):
     if not os.path.exists(folder_output):
         os.makedirs(folder_output)
 
+    if arguments.o is not None:
+        fname_out = arguments.o
+    else:
+        fname_out = os.path.basename(add_suffix(fname_data, "_seg"))
+
```
kousu, line 462: This is the cause of the error in https://github.com/neuropoly/spinalcordtoolbox/pull/3081#issuecomment-752034463. It needs to be aware of `-ofolder` too.

**Probably** this PR should **remove** `-ofolder` entirely since `-o` covers the same purpose. You should replace the above `folder_out` section with:

```
folder_out = os.path.dirname(arguments.o)
```

If you need to keep both arguments -- which I really recommend against -- then you need to make sure they are consistent with each other. That means doing something like:

```
# quick draft, don't commit as is, please test it out
folder_out, fname_out = os.path.split(arguments.o)
if arguments.ofolder is not None and arguments.ofolder != folder_out:
    raise ValueError(f"Output file {arguments.o} must be in output folder {arguments.ofolder}") 
```

What's `-ofolder` for anyway? Does `propseg` output extra metadata to that folder?

kousu: By the time it gets to

https://github.com/neuropoly/spinalcordtoolbox/blob/552b908c63234c23cf65e9e520ec3c88a55bcec8/spinalcordtoolbox/scripts/sct_propseg.py#L635-L640

you must have

```
assert not os.path.dirname(fname_out)
```

That's what that section of code is implicitly assuming. You can either fix it there, or fix it above. Adding that assertion is probably a good idea, at least for the tests.

kousu: Ou.... j'ai pas encore tout lire. Il y ..trois? fichiers? `f"{F}_rescaled.nii.gz"` et `f"{F}_rescaled_centerline.nii.gz"` et ça veux redire vers `f"{F}.nii.gz"` et `f"{F}_centerline.nii.gz"?

Est-ce qu'on c'est possible de juste demander de `isct_propseg` de faire le bon chose en prèmiere?

jcohenadad: > What's -ofolder for anyway? Does propseg output extra metadata to that folder?

propseg can output more than one stuff, e.g. `-CSF`, ` -centerline-coord`. Historically, the `-ofolder` was created when we were using `sct_testing`: this script included a function that used the flag `-ofolder`, when it existed in some scripts. Since then, we completely refactored the testing suite and we don't use -ofolder anymore, so I agree we could get rid of -ofolder as it is confusing.

But we need to keep in mind the multiple (possible) outputs. However, since these are not widely used, we can keep -o to specify the output of the cord segmentation (primary interest for sct_propseg) and have the other outputs with fix naming (very few people use the other outputs anyway).

unit_testing/cli/test_cli_sct_deepseg_sc.py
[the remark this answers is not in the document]
lrouhier: Done

spinalcordtoolbox/scripts/sct_propseg.py
lrouhier: I tried something to remove this -ofolder issue, tell me what you think! (bab7bff)

unit_testing/cli/test_cli_sct_deepseg_sc.py
lrouhier: I went with argv so no need for formatted string ;)

spinalcordtoolbox/scripts/sct_propseg.py
joshuacwnewton: If the -ofolder issue is fixed, `tmp_path` should be usable again, so could you add `tmp_path` back to the `sct_propseg` test?  Would help to test this change. :slightly_smiling_face:

joshuacwnewton: To clarify what @kousu was saying, the error occurs with this line:

https://github.com/neuropoly/spinalcordtoolbox/blob/fd00c800467aa194fcbae073eb10bc589d772920/spinalcordtoolbox/scripts/sct_propseg.py#L639-L640

When called in CI, the two values are:

joshuacwnewton: To clarify what @kousu was saying, the error occurs with this line:

https://github.com/neuropoly/spinalcordtoolbox/blob/fd00c800467aa194fcbae073eb10bc589d772920/spinalcordtoolbox/scripts/sct_propseg.py#L639-L640

When this command is called in the test, the two values are this: [(CI link)](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495#L2096-L2105)

```
src = './t2_rescaled_seg.nii.gz'
dst = '/tmp/pytest-of-travis/pytest-0/popen-gw0/test_sct_propseg_o_flag0/test_seg.nii.gz'
```

The problem here is that `src` is not using `tmp_path`. (Every output file should be created in the `tmp_path` directory.)

To fix this, you need to make sure that `folder_output` uses `tmp_path`.

https://github.com/neuropoly/spinalcordtoolbox/blob/3df9aeffcade01e30befd9b6074fc77663d051e5/spinalcordtoolbox/scripts/sct_propseg.py#L438

So, you might use [`os.path.dirname`](https://docs.python.org/3/library/os.path.html#os.path.dirname) get the parent directory from the `-o` argument, so that everything is output to that directory.

```diff
- folder_output = './' 
+ folder_output = os.path.dirname(fname_out)  # Get base directory from output path
```

lrouhier: @joshuacwnewton what do you think about f81733d ?

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
```diff
@@ -1,6 +1,9 @@
 from pytest_console_scripts import script_runner
 import pytest
 import logging
+import subprocess
```
joshuacwnewton, line 4:
```suggestion
```

unit_testing/cli/test_cli_sct_propseg.py
```diff
@@ -1,6 +1,9 @@
 from pytest_console_scripts import script_runner
 import pytest
 import logging
+import subprocess
```
joshuacwnewton, line 4:
```suggestion
```

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -1,6 +1,9 @@
 from pytest_console_scripts import script_runner
 import pytest
 import logging
+import os
+import subprocess
```
joshuacwnewton, line 5:
```suggestion
```

joshuacwnewton (approved):
I have some small nits, but they're very straightforward to fix, so I'm pre-approving.

spinalcordtoolbox/scripts/sct_propseg.py
joshuacwnewton: Looks good to me! :)

PR #3090, opened 2020-12-09
Change default option values to None

andreanne-lemay:
Change default values to None for postprocessing steps (`-remove-small`, `-fill-holes`, `-largest`, `-thr`). When no values are set, the postprocessing values in the configuration file of each model will be applied. The user has the option to overwrite the postprocessing steps by using the corresponding flag. The old behavior with a default values were overwriting the values set in the configuration file.

Other change:
- `-remove-small` is now compatible with multiclass models. One or more threshold values can be set for this postprocessing. If more than one value is used, each threshold is associated with its corresponding output class. The order of the output class will be specified in the task description. 

- Destination parameters now match the postprocessing name in ivadomed (`thr` became `binarize_prediction` and `largest` became `keep_largest`)

ivadomed sister PRs: https://github.com/ivadomed/ivadomed/pull/549 (change in postprocessing behavior) and https://github.com/ivadomed/ivadomed/pull/570 (-remove-small compatible with multiclass).

### DIFF
--- a/spinalcordtoolbox/scripts/sct_deepseg.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg.py
@@ -76,6 +76,7 @@ def get_parser():
     misc.add_argument(
         "-thr",
         type=float,
+        dest='binarize_prediction',
         help="Binarize segmentation with specified threshold. Set to 0 for no thresholding (i.e., soft segmentation). "
              "Default value is model-specific and was set during optimization "
              "(more info at https://github.com/sct-pipeline/deepseg-threshold).",
@@ -89,21 +90,22 @@ def get_parser():
         default=1)
     misc.add_argument(
         "-largest",
+        dest='keep_largest',
         type=int,
         help="Keep the largest connected-objects from the output segmentation. Specify the number of objects to keep."
              "To keep all objects, set to 0",
-        default=0)
+        default=None)
     misc.add_argument(
         "-fill-holes",
         type=int,
         help="Fill small holes in the segmentation.",
         choices=(0, 1),
-        default=0)
+        default=None)
     misc.add_argument(
         "-remove-small",
         type=str,
         help="Minimal object size to keep with unit (mm3 or vox). Example: 1mm3, 5vox.",
-        default='0vox')
+        default=None)
 
     misc = parser.add_argument_group('\nMISC')
     misc.add_argument(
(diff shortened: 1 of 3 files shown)


spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -76,11 +76,12 @@ def get_parser():
     misc.add_argument(
         "-thr",
         type=float,
+        dest='binarize_prediction',
         help="Binarize segmentation with specified threshold. Set to 0 for no thresholding (i.e., soft segmentation). "
              "Default value is model-specific and was set during optimization "
              "(more info at https://github.com/sct-pipeline/deepseg-threshold).",
         metavar=Metavar.float,
-        default=0.9)
+        default=None)
```
charleygros, line 84: Great!!!

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -89,21 +90,25 @@ def get_parser():
...
     misc.add_argument(
         "-largest",
+        dest='keep_largest',
         type=int,
         help="Keep the largest connected-objects from the output segmentation. Specify the number of objects to keep."
              "To keep all objects, set to 0",
-        default=0)
+        default=None)
     misc.add_argument(
         "-fill-holes",
```
charleygros, line 99:
```suggestion
        "-fill-holes",
        dest="fill_holes",
```

Maybe it is not needed?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -89,21 +90,25 @@ def get_parser():
...
         type=int,
         help="Fill small holes in the segmentation.",
         choices=(0, 1),
-        default=0)
+        default=None)
     misc.add_argument(
         "-remove-small",
```
charleygros, line 105:
```suggestion
        "-remove-small",
        dest="remove_small",
```

charleygros (approved):
💯

spinalcordtoolbox/scripts/sct_deepseg.py
andreanne-lemay: By default, the destination value is `fill_holes` (since the flag name is `-fill-holes`), so it is not required

PR #3091, opened 2020-12-10
Refactor CLI init steps to be consistent across scripts

### LINKED ISSUE #3085: Initialization steps (arg passing, init_sct, main) not consistent between CLI scripts
### `main()` vs. `main(args)`

Historically, CLI scripts have parsed args implicitly using `sys.argv`. However, we've started to write tests that involve calls to a CLI script's `main` function. So, scripts have been tweaked on a case-by-case basis to accept args (`main()` -> `main(args)`).

Right now, some of our scripts can be called using `main(args)`, while others can't and still require `subprocess`. This has presented a roadblock in new PRs where it is unclear what changes need to be done, e.g. https://github.com/neuropoly/spinalcordtoolbox/pull/3081#discussion_r537582769.

### Other initialization steps

Beyond this, there are little differences in the initialization steps across our CLI scripts as a whole, resulting in inconsistent behavior between scripts. For example, compare `sct_propseg` to `sct_deepseg`:

https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_propseg.py#L671-L676

https://github.com/neuropoly/spinalcordtoolbox/blob/dc97d42730d7e603f83e54c68f242ad2bef7189b/spinalcordtoolbox/scripts/sct_deepseg.py#L224-L226

Here, `sct_propseg` accepts parsed args (i.e. a namespace/dict), while `sct_deepseg` accepts unparsed args (i.e. a list of arg-value pairs). The exit behavior is different, too. And, this is just one example.

I think it might be useful to settle on a single convention, document it in the dev wiki, then apply it to each script. That way, we can expect predictable behavior from each script, e.g. when we're writing tests.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Motivation: Currently, we can call some, but not all scripts using `sct_function.main()`. This is because script init steps are inconsistent between scripts.

So, this PR refactors scripts to all have the same consistent structure, as described in the [CLI script conventions](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-CLI-script-conventions-(structure,-argparse)) dev Wiki page. This includes some format tidying. For most of the scripts, the necessary changes were very straightforward:

* Adding arguments to main.
* Renaming variables (`args` -> `argv`, `args` -> `arguments`, `verbosity` -> `verbose`).
* Grouping statements at the top of main.
* Removing some `Param()` objects that were unnecessary.
* Adding missing verbose arguments (`-v`).
* Creating a main function when it didn't exist.

But, for some scripts, a more involved refactor was needed. So, extra care should be given to the following scripts when reviewing:

* `sct_apply_transfo`: Had some unique parsing behavior (flatting lists for `-w` and `-winv` before parsing). But, this was only ever used in one call to `sct_apply_transfo`, so I figured it wasn't worth the extra complexity. I removed the parsing behavior, and left it to be handled at the site of the call. (In this case, a single call within `sct_register_to_template` was modified.)
* `sct_check_dependencies`: Doesn't have `-v` and instead uses `-complete` for verbose. A bit strange, but I cleaned this up a bit.
* `sct_denoising_olm`: Did parsing within `if __name__ == "__main__"`, then main() accepted parsed arguments. I refactored so that main accepted the unparsed argument list, then did parsing inside main.
* `sct_dmri_compute_dti`: `Param().verbose` was used as a sort of global default. I refactored this to use an explicitly passed `verbose` variable.
* `sct_download_data`: The function `get_parser` took a `dict_url` parameter, when usually this function doesn't take parameters. Instead, I made `dict_url` a global variable. 
* `sct_extract_metric`: Like `sct_denoising_olm`, it did parsing within `if __name__ == "__main__"`, then main() accepted parsed arguments. I refactored so that main accepted the unparsed argument list, then did parsing inside main.
* `sct_flatten_sagittal`: Like `sct_denoising_olm`, it did parsing within `if __name__ == "__main__"`, then main() accepted parsed arguments. I refactored so that main accepted the unparsed argument list, then did parsing inside main.
* `sct_get_centerline`: Add verbose typing.
* `sct_register_multimodal`: The function `get_parser` took a `paramregmulti` parameter, when normally this function doesn't take parameters. Instead, I made `paramregmulti` a global variable, which also removed a duplicate definition in the process.
* `sct_warp_template`: `Param()` was used as a sort of global default. I refactored this to use an explicitly passed variables (`verbose`, `folder_template`, `folder_atlas`, `folder_spinal_levels`, `file_info_label`, `list_labels_nn`)
* `unit_testing/test_centerline` and `unit_testing/test_qmri`: I made a quick tweak because they were calling `init_sct` to set logging levels.

## Linked issues

Fixes #3085.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -24,7 +24,7 @@ from spinalcordtoolbox.cropping import ImageCropper
 from spinalcordtoolbox.math import dilate
 from spinalcordtoolbox.labels import cubic_to_point
 from spinalcordtoolbox.utils.shell import Metavar, SmartFormatter, get_interpolation, display_viewer_syntax
-from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv
+from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv, set_global_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, rmtree, extract_fname, copy
 
 from spinalcordtoolbox.scripts import sct_image
@@ -336,25 +336,16 @@ class Transform:
 
 # MAIN
 # ==========================================================================================
-def main(args=None):
+def main(argv=None):
     """
     Entry point for sct_apply_transfo
-    :param args: list of input arguments. For parameters -w and -winv, args list should include a nested list for every
-    item. Example: args=['-i', 'file.nii', '-w', ['warp1.nii', 'warp2.nii']]
+    :param argv: list of input arguments.
     :return:
     """
-
-    # get parser args
-    if args is None:
-        args = None if sys.argv[1:] else ['--help']
-    else:
-        # flatten the list of input arguments because -w and -winv carry a nested list
-        lst = []
-        for line in args:
-            lst.append(line) if isinstance(line, str) else lst.extend(line)
-        args = lst
     parser = get_parser()
-    arguments = parser.parse_args(args=args)
+    arguments = parser.parse_args(argv if argv else ['--help'])
+    verbose = arguments.v
+    set_global_loglevel(verbose=verbose)
 
     input_filename = arguments.i
     fname_dest = arguments.d
@@ -368,15 +359,12 @@ def main(args=None):
     transform.output_filename = arguments.o
     transform.interp = arguments.x
     transform.remove_temp_files = arguments.r
-    transform.verbose = arguments.v
-    init_sct(log_level=transform.verbose, update=True)  # Update log level
+    transform.verbose = verbose
 
     transform.apply()
 
 
-# START PROGRAM
-# ==========================================================================================
 if __name__ == "__main__":
     init_sct()
-    # call main function
-    main()
+    main(sys.argv[1:])
+
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -114,8 +114,8 @@ def get_parser():
         type=int,
         choices=[0, 1, 2],
         default=1,
-        # Values [0, 1, 2] map to log levels [WARNING, INFO, DEBUG]
-        help="Verbosity. 0: Minimal, 1: Default, 2: Expanded (Display figures)")
+        # Values [0, 1, 2] map to logging levels [WARNING, INFO, DEBUG], but are also used as "if verbose == #" in API
+        help="Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode")
 
     return parser
 
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -109,12 +109,13 @@ def get_parser():
         default=1,
         choices=(0, 1))
     optional.add_argument(
-        "-v",
-        help="Verbose: 0: nothing, 1: classic, 2: expended.",
-        required=False,
+        '-v',
+        metavar=Metavar.int,
         type=int,
+        choices=[0, 1, 2],
         default=1,
-        choices=(0, 1, 2))
+        # Values [0, 1, 2] map to log levels [WARNING, INFO, DEBUG]
+        help="Verbosity. 0: Minimal, 1: Default, 2: Expanded (Display figures)")
 
     return parser
 
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -269,9 +269,12 @@ def get_parser():
     )
     optional.add_argument(
         '-v',
-        choices=['0', '1', '2'],
-        default=param.verbose,
-        help="Verbose. 0: nothing. 1: basic. 2: extended."
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1, 2],
+        default=1,
+        # Values [0, 1, 2] map to log levels [WARNING, INFO, DEBUG]
+        help="Verbosity. 0: Minimal, 1: Default, 2: Expanded (Display figures)"
     )
 
     return parser
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -279,17 +279,14 @@ def get_parser():
 
 # MAIN
 # ==========================================================================================
-def main(args=None):
+def main(argv=None):
+    parser = get_parser()
+    arguments = parser.parse_args(argv if argv else ['--help'])
+    verbose = arguments.v
+    set_global_loglevel(verbose=verbose)
 
     # initializations
     param = Param()
-
-    # check user arguments
-    parser = get_parser()
-    if args:
-        arguments = parser.parse_args(args)
-    else:
-        arguments = parser.parse_args(args=None if sys.argv[1:] else ['--help'])
 
     fname_data = arguments.i
     fname_seg = arguments.s
@@ -316,8 +313,6 @@ def main(args=None):
     contrast_template = arguments.c
     ref = arguments.ref
     param.remove_temp_files = arguments.r
-    verbose = int(arguments.v)
-    init_sct(log_level=verbose, update=True)  # Update log level
     param.verbose = verbose  # TODO: not clean, unify verbose or param.verbose in code, but not both
     param_centerline = ParamCenterline(
         algo_fitting=arguments.centerline_algo,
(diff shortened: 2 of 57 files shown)


spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -336,25 +336,16 @@ def apply(self):
 
 # MAIN
 # ==========================================================================================
-def main(args=None):
+def main(argv=None):
     """
     Entry point for sct_apply_transfo
-    :param args: list of input arguments. For parameters -w and -winv, args list should include a nested list for every
-    item. Example: args=['-i', 'file.nii', '-w', ['warp1.nii', 'warp2.nii']]
+    :param argv: list of input arguments.
```
lrouhier, line 342: Why remove the example? is the nested list no longer needed ?

lrouhier:
See https://github.com/neuropoly/spinalcordtoolbox/pull/3091/files#diff-67e76316116b9f8c81ccb907f018ca6710adf11a6f8f261a06fa5823ab48d0a8R119 --> should it display 'all' the verbose possibilities ? (0,1 and 2)

lrouhier:
Thanks for this amazing work. This is huge! I tried to be as exhaustive as possible in my list regarding the inconsistent verbose help. Otherwise it looks pretty good !

spinalcordtoolbox/scripts/sct_apply_transfo.py
joshuacwnewton: Yes, exactly. To make nested lists work, there was some extra parsing logic outside of `argparse`.

https://github.com/neuropoly/spinalcordtoolbox/blob/e07cd93cdd9f856b326f81efdd470d10ea642af2/spinalcordtoolbox/scripts/sct_apply_transfo.py#L347-L355

I removed this extra parsing logic (and the example). I did this for two reasons:

* I felt that all parsing behavior should be handled only inside `argparse`.
* I felt the usage would be clearer if the input formatting was exactly same for main as it is for the CLI parsing. 
    * CLI: `-w warp1.nii warp2.nii` 
    * main: `argv=['-w', 'warp1.nii', 'warp2.nii']`

joshuacwnewton:
> See https://github.com/neuropoly/spinalcordtoolbox/pull/3091/files#diff-67e76316116b9f8c81ccb907f018ca6710adf11a6f8f261a06fa5823ab48d0a8R119 --> should it display 'all' the verbose possibilities ? (0,1 and 2)
> 
> same for sct_crop_image, sct_deepseg, sct_deepseg_gm, sct_dice_coefficient, sct_dmri_separate_b0_and_dwi, sct_extract_metrics, sct_fmri_compute_tsnr, sct_propseg
> 
> this will also unify the verbose documentation across scripts. (sct_dmri_moco ,sct_dmri_transpose_bvecs, sct_flatten_sagittal, sct_fmri_moco, sct_process_segmentation, sct_register_to_template are not consistent)

I agree, verbosity is indeed inconsistent. I was kind of avoiding fixing verbosity in this PR, just because there is another issue for handling verbosity (#2676), and I thought it would be better to split the work across 2 PRs. :sweat_smile:

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -1027,12 +1022,13 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
...
             for ifile in range(len(src)):
-                sct_apply_transfo.main(args=[
+                sct_apply_transfo.main(argv=[
                     '-i', src[ifile],
                     '-d', dest[ifile],
-                    '-w', warp_forward,
                     '-o', add_suffix(src[ifile], '_reg'),
-                    '-x', interp_step[ifile]])
+                    '-x', interp_step[ifile],
+                    '-w'] + warp_forward
```
joshuacwnewton, line 1030: Replacing `['-w', warp_forward]` with `['-w'] + warp_forward` to deal with nested list.

(This was the only call to `sct_apply_transfo.main()` that used nested lists.)

spinalcordtoolbox/scripts/sct_apply_transfo.py
joshuacwnewton: Besides, the nested list usage was only used in 1 place, and that is quite easy to fix. (See https://github.com/neuropoly/spinalcordtoolbox/pull/3091#discussion_r541946975 for more info.)

lrouhier:
> I agree, verbosity is indeed inconsistent. I was kind of avoiding fixing verbosity in this PR, just because there is another issue for handling verbosity (#2676), and I thought it would be better to split the work across 2 PRs. 😅

Oh ok, I get it! No problem, you are right it might be a bit out of the scope of this PR.

jcohenadad (approved):
LGTM! 🏅

joshuacwnewton:
On second look, @lrouhier, I think you were right about verbosity. I think it needs to be handled in this PR.

In many places, `verbose = int(arguments.v)` was used. (i.e. the type conversion was done outside of argparse). Thus, I also needed to fix every instance of string verbosity in argparse, which I haven't done yet.

joshuacwnewton:
Apologies for the expanded scope of the PR, but could you re-review the PR (specifically a89c09b), @jcohenadad?

joshuacwnewton:
Ah, I didn't realize how strange comment threads look on GitHub when a review is given on a specific commit. Oops. :sweat_smile:

jcohenadad (commented):
didn't have time to dig deep, but i'm helping a group in Zurich with an SCT pipeline, i'm on your branch, and i ran into this:
~~~
sct_process_segmentation -i t2_tra_seg.nii.gz -perslice 1 -vertfile t2_tra_vertlevels.nii.gz -o /Users/julien/temp/20201217_inspired/results4/results/csa-SC_T2.csv -append 1 -qc /Users/julien/temp/20201217_inspired/results4/qc -qc-subject HC02
--

  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 352, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 343, in main
    generate_qc(fname_segmentation, args=args, path_qc=os.path.abspath(path_qc), dataset=qc_dataset,
NameError: name 'args' is not defined
~~~

jcohenadad:
> didn't have time to dig deep, but i'm helping a group in Zurich with an SCT pipeline, i'm on your branch, and i ran into this:
> 
> ```
> sct_process_segmentation -i t2_tra_seg.nii.gz -perslice 1 -vertfile t2_tra_vertlevels.nii.gz -o /Users/julien/temp/20201217_inspired/results4/results/csa-SC_T2.csv -append 1 -qc /Users/julien/temp/20201217_inspired/results4/qc -qc-subject HC02
> --
> 
>   File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 352, in <module>
>     main(sys.argv[1:])
>   File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 343, in main
>     generate_qc(fname_segmentation, args=args, path_qc=os.path.abspath(path_qc), dataset=qc_dataset,
> NameError: name 'args' is not defined
> ```

meaning: we might also want to make our tests more sensitive (this is the kind of stuff that should be caught by our CI)

joshuacwnewton:
> meaning: we might also want to make our tests more sensitive (this is the kind of stuff that should be caught by our CI)

Yes, agreed. `sct_process_segmentation` is only tested using the command `sct_process_segmentation -i t2/t2_seg-manual.nii.gz`. None of its other arguments are tested (e.g. `-qc` in this case), and no results are checked.

I would be happy to spend time writing additional tests to increase coverage. It's just a matter of prioritizing. :sweat_smile:

jcohenadad:
> I would be happy to spend time writing additional tests to increase coverage. It's just a matter of prioritizing. 😅

that would be amazing, and yes, i am fully aware of the massive amount of work ahead of us 😅

kousu:
Hiyo, this seems to have broken `install_sct`? https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465498727#L1036

```
Validate installation...
--
Spinal Cord Toolbox (git-HEAD-ade09ab7f5b22aa3c9028d3a719a591125968a85)
sct_check_dependencies 
--
usage: sct_check_dependencies
       [-h]
       [-complete]
       [-short]
Check the installation and environment variables of the toolbox and its dependencies.
OPTIONAL ARGUMENTS:
  -h, --help  Show this
              help
              message and
              exit
  -complete   Complete
              test.
              (default:
              False)
  -short      Short test.
              Only shows
              SCT
              version,
              CPU cores
              and RAM
              available.
              (default:
              False)
Open a new Terminal window to load environment variables, or run:
source /Users/travis/.bashrc
```

`-h` should return an error code *unless the user asked for it*. I think the subtle mistake is this line:

https://github.com/neuropoly/spinalcordtoolbox/blob/d925ab5da582f52c8d8f495876a13adceea01523/spinalcordtoolbox/scripts/sct_check_dependencies.py#L207

argparse should already be smart enough to handle no args -- and to return an error code -- in that case. What happens if you just leave that off?

kousu:
Answering my own question: removing `else ['--help']` repairs `sct_check_dependencies`. It's usual call has always just been `sct_check_dependencies`.
Are there other scripts that might have the same glitch?

PR #3098, opened 2020-12-16
Update tumor segmentation models

andreanne-lemay:
The following changes were made:
- New model for multiclass (tumor, edema, cavity) tumor segmentation is added.
- The following models are updated with new patch size and stride parameters: findcord_tumor, t2_tumor.
- Only github URLs are kept for new models


charleygros (approved):
:100:

Issue #3071, opened 2020-12-02
Use Github Actions for continuous integration


Drulex:
Motivated by this https://github.com/neuropoly/spinalcordtoolbox/issues/3057#issuecomment-735281988 and as discussed during the weekly meeting we will migrate the continuous integration infrastructure to Github Actions.

I am creating this issue to document the process and centralize discussions on the topic.

kousu:
Working on this!

So far https://github.com/neuropoly/spinalcordtoolbox/runs/1602878596 has gotten to running the tests in 3 minutes. Travis always took at least 6 to get to the same place.

I hate to say it but vive la GithubCI.

kousu:
I've ported the platform matrix over (almost); I sidestepped my `util/dockerize.sh` because Actions supports docker via its API: just say: `container: debian:10`.

However I've hit a stupid growing-pains snag: in travis we blocked off platforms into always/nightly, e.g. 

always:

https://github.com/neuropoly/spinalcordtoolbox/blob/42f5a2ec1857a145f2de05e2b7fb6da627b5c7e1/.travis.yml#L94-L97

nightly:

https://github.com/neuropoly/spinalcordtoolbox/blob/42f5a2ec1857a145f2de05e2b7fb6da627b5c7e1/.travis.yml#L37-L40

but the equivalent on Actions:

https://github.com/neuropoly/spinalcordtoolbox/blob/514179be769410e1366cfc6a535344bc194ffd86/.github/workflows/tests.yml#L57

[fails with](https://github.com/neuropoly/spinalcordtoolbox/actions/runs/441382509)

> The workflow is not valid. .github/workflows/tests.yml (Line: 57, Col: 9): Unrecognized named-value: 'matrix'. Located at position 3 within expression: !(matrix.nightly) \|\| (matrix.nightly && (github.event_name == 'schedule' \|\| (github.event_name == 'push' && github.ref == 'refs/heads/release'))) |  
-- | --

This is a discovered bug, though Github hasn't addressed it so maybe it's not a known one? https://github.community/t/how-to-conditionally-include-exclude-items-in-matrix-eg-based-on-branch/16853/5

It works if I move the `if:` a level down into `step:`. That'll work but leave us with a bunch of vacuous test runs and I don't want that.

joshuacwnewton:
[...]

Issue #3029, opened 2020-11-10
Activating conda env doesn't override user's pyenv during install_sct


Drulex:
A user is experiencing issues installling `sct`: https://forum.spinalcordmri.org/t/installation-failed-problem-creating-launchers/562

Upon further inspection it appears that the installation is not happening within the bundled `venv`, but instead in the environment provided by the system interpreter (in this case using a `pyenv` shim).

jcohenadad:
Could be related to more restrictive permissions in Catalina. Possibly relevant:
- https://stackoverflow.com/questions/58293755/how-to-get-conda-and-virtualenv-to-work-on-mac-os-catalina
- https://github.com/ContinuumIO/anaconda-issues/issues/11387

joshuacwnewton:
> Could be related to more restrictive permissions in Catalina. Possibly relevant:
> 
> * https://stackoverflow.com/questions/58293755/how-to-get-conda-and-virtualenv-to-work-on-mac-os-catalina
> * [ContinuumIO/anaconda-issues#11387](https://github.com/ContinuumIO/anaconda-issues/issues/11387)

Regarding the first link, I do think we're already using a Catalina-approved installation method. See [Anaconda | How to Restore Anaconda after Update to MacOS Catalina](https://www.anaconda.com/blog/how-to-restore-anaconda-after-macos-catalina-update). It says:

> Start fresh with a new Anaconda installation. Pay attention during installation to make sure that your install path is a subfolder of your home folder, such as /Users/me/anaconda3

Which is indeed what we're doing with our Miniconda installation.

jcohenadad:
i'm trying a few things to reproduce this error on my macOs, starting by removing disk access by terminal:
<img width="780" alt="Screen Shot 2020-11-10 at 9 56 22 AM" src="https://user-images.githubusercontent.com/2482071/98690663-2c682f00-233b-11eb-9dd8-bcd9a8883423.png">

Drulex:
Maybe related to:
- https://github.com/conda/conda/issues/8211
- https://stackoverflow.com/questions/54429210/how-do-i-prevent-conda-from-activating-the-base-environment-by-default/57974390#57974390

jcohenadad:
I tried launching SCT installation from within a venv, but I cannot reproduce the error.

<details><summary>Full output</summary>

```
julien-macbook:~/code/sct $ conda activate
(base) julien-macbook:~/code/sct $ 
[...]

PR #3102, opened 2020-12-16
Turn on stricter shell rules + maintainence

### LINKED ISSUE #3105: install_sct: homebrew's installer is now in bash
Us:

https://github.com/neuropoly/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/install_sct#L173

but that link (currently pointed at https://github.com/Homebrew/install/blob/e141d766960553367441362490e42a2fead644c6/install) suggests

> Warning: The Ruby Homebrew installer is now deprecated and has been rewritten in
> Bash. Please migrate to the following command:
>  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"

Why do we even have this again? In order to be able to install gcc? Which is because some the dependencies of our dependencies don't have macOS .whls? Seems like these are the historical context? #2404, #2410, #2577.

Anyway I'm going to do this migration but then we should talk about getting rid of it entirely if we can.

### LINKED ISSUE #3112: Unsuccessful installation of SCT on GH actions of shimming-toolbox
### Description

We want to install SCT on the GH actions of the shimming-toolbox in shimming-toolbox/shimming-toolbox#184 but it's unsuccessful for MacOs while it is for Linux. However, the installation steps are exactly the same for both of them.

### Steps to Reproduce

0. Install last release of SCT in `.github/workflows/tests.yml` file of the repository with the following commands:

1. `git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox`

2. `cd spinalcordtoolbox`

3. `yes | ASK_REPORT_QUESTION=false PIP_PROGRESS_BAR=off ./install_sct` , `yes | ./install_sct` or other commands according to this principle

**Expected behavior:** 
A successful installation on MacOs and Linux

**Actual behavior:** 
The installation is successful for Linux but not for MacOs. Most of the time the error comes from [this step](https://github.com/shimming-toolbox/shimming-toolbox/pull/184/checks?check_run_id=1514698276#step:8:1388) of the installation. This one is cancelled after running for a few hours.
After checking `install_sct`, we are wondering if the error could be from this [line](https://github.com/neuropoly/spinalcordtoolbox/blob/e07cd93cdd9f856b326f81efdd470d10ea642af2/install_sct#L525).

kousu:
#### GitHub

#### PR contents

## Description

Our installer is fragile around this edges and part of that is that shell itself is fragile. This sets the recommended shell strict-mode settings that make it a bit less of a fragile language to work in.

I did these safety improvements:

- setting [recommended shell safeties](https://sipb.mit.edu/doc/safe-shell/): `set -euo pipefail; shopt -s failglob`
- to appease `set -o pipefail`: switch to `#!/usr/bin/env bash`
    - every time we're not polyglottal is unfortunate, but this option is so useful that I think it's worth depending on bash
-  to appease `set -u`: initializing all variables; in the cases of checking external envs, using the cryptic (but standard) `${var:-}` syntax
- to appease `set -o pipefail`: replacing all our uses of `yes |` with the proper non-interactive flag for each script (mostly: `-y`, but for the brew installer: `NONINTERACTIVE=1`, and for using brew: `-f`)
- inspired by that, adding our own `-y` instead of recommending `yes | install_sct`. This is much less fragile. This fixes #3112.
    - merge `ASK_REPORT_QUESTION` into `-y`; **this is an API change**
        - but is anyone really using this flag anyway?
- to appease `shopt -s failglob`: fixing the pre-clean `find` step
    - in that case, we actually do want to pass a literal `sct_*` into `find(1)` for it to parse.
- set appease `set -e`: check for command results as `if ! cmd;` instead of "cmd; if [[ $? -ne 0 ]];` or sometimes `cmd || die ".."` or `cmd && ....`
    - that's nicer to read anyway
- quoting *all* the variables everywhere
- drop unused variables (`$SCRIPT_DIR` and `$line`) (found by [shellcheck](https://github.com/koalaman/shellcheck))
- use `read -r` (found by [shellcheck](https://github.com/koalaman/shellcheck))

These maintenances:

- update deprecated brew installer; fixes #3105 .
- update deprecated "source conda/bin/deactivate" -> "conda deactivate"
    - (this doesn't actually undo everything conda loads :(; it leaves the `conda` function and `CONDA_*` env vars in memory)

I also did some maintainability improvements:
- removing unnecessary `which` dependency
    - found by @joshuacwnewton in #3103 (https://github.com/neuropoly/spinalcordtoolbox/pull/3103#discussion_r545956795)
- making `run()` much safer; `"$@"` is designed for this kind of thing
    - this is some shell wizardry so ask me to explain it if you need to
- merge the macOS and linux calls to `sed`
    - it looks like the linux call was there first, that was discovered to be incompatible, and the macOS call was added; but it turns out the macOS version is the more compatible one (which sort of makes sense: it's old BSD styles)
- simplify `download()` 
- shortened `edit_shellrc` to only use a single `>>` (the power of subshells!)

I also **tweaked `install_sct`'s UI**:

- moving *all* the prompts (in interactive mode) up to the start of the script instead of the user needing to wait around to answer when the script finishes
- simplifying `usage()`
- *change the wording of the install error message* to ask users to upload instead of paste their logs
- moving "Checking requirements!" after `getopts` so that it doesn't run for just `./install_sct -h`
- `./install_sct -h` returns 0, not 99; it's not an error.

I don't know how to test this well. We don't have a good way to do a test matrix of all possible install cases. We test a bunch of platforms, and that's good, but we don't mix and match the installer's options anywhere.

I've run it as both `./install_sct` and `./install_sct -y` and both work (except for #3107, which I ignored by commenting torch out of my local requirements.txt for now), so I think it's good enough.

Aside: It would be good to test what happens if we try installing on a disk that runs out of space. This branch *should* terminate early in that case, whereas `master` should run through to the end giving many many errors.

## Linked issues

Part of #3029. 
Fixes #3105.  
Fixes #3112.
Unblocks #3071 (both https://github.com/actions/runner/issues/884, discovered @Drulex first discovered (and tested at https://github.com/Drulex/conda-test-actions, https://github.com/Drulex/spinalcordtoolbox/tree/aj-basic-github-actions) about a month ago)

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -2,7 +2,7 @@
 # CI testing script
 #  Installs SCT from scratch and runs all the tests we've ever written for it.
 
-set -e # Error build immediately if install script exits with non-zero
+set -ueo pipefail # stricter shell rules
 
 echo Installing SCT
 PIP_PROGRESS_BAR=off ./install_sct -y
--- a/.ci.sh
+++ b/.ci.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # CI testing script
 #  Installs SCT from scratch and runs all the tests we've ever written for it.
 
--- a/.ci.sh
+++ b/.ci.sh
@@ -5,7 +5,7 @@
 set -e # Error build immediately if install script exits with non-zero
 
 echo Installing SCT
-yes | ASK_REPORT_QUESTION=false PIP_PROGRESS_BAR=off ./install_sct
+PIP_PROGRESS_BAR=off ./install_sct -y
 echo $?
 echo "... STATUS"
 
--- a/.ci.sh
+++ b/.ci.sh
@@ -2,7 +2,11 @@
 # CI testing script
 #  Installs SCT from scratch and runs all the tests we've ever written for it.
 
-set -ueo pipefail # stricter shell rules
+# stricter shell mode
+# https://sipb.mit.edu/doc/safe-shell/
+set -eo pipefail  # exit if non-zero error is encountered (even in a pipeline)
+set -u            # exit if unset variables used
+shopt -s failglob # error if a glob doesn't find any files, instead of remaining unexpanded
 
 echo Installing SCT
 PIP_PROGRESS_BAR=off ./install_sct -y
--- a/.ci.sh
+++ b/.ci.sh
@@ -10,8 +10,6 @@ shopt -s failglob # error if a glob doesn't find any files, instead of remaining
 
 echo Installing SCT
 PIP_PROGRESS_BAR=off ./install_sct -y
-echo $?
-echo "... STATUS"
 
 echo *** CHECK PATH ***
 ls -lA bin  # Make sure all binaries and aliases are there
@@ -34,4 +32,3 @@ bash -c 'PYTHONPATH="$PWD/scripts:$PWD" pylint -j3 --py3k --output-format=parsea
 # python create_package.py -s ${TRAVIS_OS_NAME}  # test package creation
 # cd ../spinalcordtoolbox_v*
 # yes | ./install_sct  # test installation of package
-
--- a/install_sct
+++ b/install_sct
@@ -532,7 +532,7 @@ fi
 # Copy files to destination directory
 if [[ "$SCT_DIR" != "$SCT_SOURCE" ]]; then
   print info "Copying source files from $SCT_SOURCE to $SCT_DIR"
-  cp -vR "$SCT_SOURCE/"* "$SCT_DIR/" | while read line; do echo -n "."; done
+  cp -vR "$SCT_SOURCE/"* "$SCT_DIR/" | while read; do echo -n "."; done
 else
   print info "Skipping copy of source files (source and destination folders are the same)"
 fi
--- a/install_sct
+++ b/install_sct
@@ -421,8 +421,8 @@ else
 The install_sct script must be executed from the source directory"
 fi
 
-# Get installation type (from git or from package)
-if [[ -z "${SCT_INSTALL_TYPE:-}" ]]; then
+# Get installation type (from git or from package) if not already specified
+if [[ -z "$SCT_INSTALL_TYPE" ]]; then
   if [[ -d ".git" ]]; then
     # folder .git exist, therefore it is a git installation
     SCT_INSTALL_TYPE="in-place"
@@ -658,7 +658,7 @@ export PATH="$SCT_DIR/$BIN_DIR:$PATH"
 # ----------------------------------------------------------------------------------------------------------------------
 
 # Install binaries
-if [[ -n "${NO_SCT_BIN_INSTALL:-}" ]]; then
+if [[ -n "$NO_SCT_BIN_INSTALL" ]]; then
   print warning "WARNING: SCT binaries will not be (re)-installed"
 else
   print info "Installing binaries..."
@@ -667,7 +667,7 @@ fi
 print info "All requirements installed!"
 
 # Install data
-if [[ -n "${NO_DATA_INSTALL:-}" ]]; then
+if [[ -n "$NO_DATA_INSTALL" ]]; then
   print warning "WARNING: data/ will not be (re)-install"
 else
   # Download data
--- a/install_sct
+++ b/install_sct
@@ -180,7 +180,7 @@ function check_requirements() {
         GCC_INSTALL="no"
         if [ -z "$NONINTERACTIVE" ]; then
           print question "Do you want to install it now? (accepting to install \"gcc\" will also install \"brew\" in case it is not installed already)? [y]es/[n]o: "
-          read GCC_INSTALL
+          read -r GCC_INSTALL
         fi
       done
       if [[ "$GCC_INSTALL" =~ [Yy](es)? ]]; then
@@ -429,7 +429,7 @@ report system to automatically receive crash reports and errors from users.
 These reports are anonymous.
 
 Do you agree to help us improve SCT? [y]es/[n]o: "
-  read REPORT_STATS
+  read -r REPORT_STATS
 fi
 
 if [[ "$REPORT_STATS" =~ [Yy](es)? ]]; then
@@ -455,7 +455,7 @@ while true; do
     if [ -z "$NONINTERACTIVE" ]; then
       print question "
 Do you agree? [y]es/[n]o: "
-      read keep_default_path
+      read -r keep_default_path
     fi
   done
   if [[ "$keep_default_path" =~ ^[Yy] ]]; then
@@ -467,7 +467,7 @@ Do you agree? [y]es/[n]o: "
 
   print question "Choose install directory. Warning! Give full path (e.g. /usr/django/sct_v3.0): \n"
   # user enters new path
-  read new_install
+  read -r new_install
 
   # Expand ~/
   new_install="${new_install/#\~\//$HOME\/}"
@@ -499,7 +499,7 @@ while [[ ! "$add_to_path" =~ ^([Yy](es)?|[Nn]o?)$ ]]; do
   add_to_path="yes"
   if [ -z "$NONINTERACTIVE" ]; then
     print question "Do you want to add the sct_* scripts to your PATH environment? [y]es/[n]o: "
-    read add_to_path
+    read -r add_to_path
   fi
 done
 
@@ -532,7 +532,7 @@ fi
 # Copy files to destination directory
 if [[ "$SCT_DIR" != "$SCT_SOURCE" ]]; then
   print info "Copying source files from $SCT_SOURCE to $SCT_DIR"
-  cp -vR "$SCT_SOURCE/"* "$SCT_DIR/" | while read; do echo -n "."; done
+  cp -vR "$SCT_SOURCE/"* "$SCT_DIR/" | while read -r; do echo -n "."; done
 else
   print info "Skipping copy of source files (source and destination folders are the same)"
 fi
--- a/install_sct
+++ b/install_sct
@@ -41,7 +41,6 @@ shopt -s failglob # error if a glob doesn't find any files, instead of remaining
 TMP_DIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'TMP_DIR')"
 # Start Directory So we go back there at the end of the Script
 SCT_SOURCE="$PWD"
-SCRIPT_DIR="scripts"
 DATA_DIR="data"
 PYTHON_DIR="python"
 BIN_DIR="bin"
--- a/install_sct
+++ b/install_sct
@@ -30,14 +30,14 @@ shopt -s failglob # error if a glob doesn't find any files, instead of remaining
 
 
 # Where tmp file are stored
-TMP_DIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'TMP_DIR')
+TMP_DIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'TMP_DIR')"
 # Start Directory So we go back there at the end of the Script
-SCT_SOURCE=$PWD
+SCT_SOURCE="$PWD"
 SCRIPT_DIR="scripts"
 DATA_DIR="data"
 PYTHON_DIR="python"
 BIN_DIR="bin"
-MACOSSUPPORTED=13  # Minimum version of macOS 10 supported
+MACOSSUPPORTED="13"  # Minimum version of macOS 10 supported
 
 
 # ======================================================================================================================
@@ -50,7 +50,7 @@ MACOSSUPPORTED=13  # Minimum version of macOS 10 supported
 function print() {
   type=$1
   txt=$2
-  case $type in
+  case "$type" in
   # Display useful info (green)
   info)
     echo -e "\n\033[0;32m${txt}\033[0m\n"
@@ -83,7 +83,7 @@ function die() {
 # Run a command and display it in color. Exit if error.
 # @input: string: command to run
 function run() {
-  cmd=$1
+  cmd="$1"
   print code "$cmd"
   $cmd
   if [[ $? != 0 ]]; then
@@ -94,12 +94,12 @@ function run() {
 # Force a clean exit
 function finish() {
   # Catch the last return code
-  value=$?
+  value="$?"
   # Get back to starting point
-  cd $SCT_SOURCE
-  if [[ $value -eq 0 ]]; then
+  cd "$SCT_SOURCE"
+  if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
-  elif [[ $value -eq 99 ]]; then
+  elif [[ "$value" -eq 99 ]]; then
     # Showing usage with -h
     echo ""
   else
@@ -108,8 +108,8 @@ Please copy the historic of this Terminal (starting with the command install_sct
 --> http://forum.spinalcordmri.org/c/sct"
   fi
   # clean tmp_dir
-  rm -rf $TMP_DIR
-  exit $value
+  rm -rf "$TMP_DIR"
+  exit "$value"
 }
 
 # reenable tty echo when user presses keyboard interrupt and output non-zero status for finish() function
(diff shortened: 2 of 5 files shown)


kousu:
There's a lot of ways this could go wrong. Our installer wasn't written with this in mind so I probably, and more than that we should probably try to induce at least the common sorts of errors that might happen (like, a read-only disk or a disk out of space, or the network failing halfway through downloading conda) and make sure the installer correctly terminates on them.

joshuacwnewton:
I'm not sure I understand how these changes related to #3029. For that issue, I don't think there even was an error to catch, because the weird `python`/`pyenv` shenanigans were happening silently.

Perhaps a new issue should be opened about shell safety? Because, #3103 seems sufficient for handling #3029.

kousu:
This was inspired by 3029, and I stand by it being "part of" that, though it's not wholly subset of it. `conda activate something` throws an error if it can't find "something", so we should be prepared to catch that; that `conda activate venv_sct` wasn't throwing an error is a bug in conda, IMO.

kousu:
also: not quite done but almost there

kousu:
Okay, ready for a look-see :light_rail:

kousu:
Boo. CI failed because of https://github.com/pytorch/pytorch/issues/49559.

kousu:
At least they fixed it quick! https://github.com/pytorch/pytorch/issues/49559#issuecomment-747764224

> Sorry about that, there was an issue with the refactor to the HTML index update script.
> 
> This should be mostly resolved now.
> 
> Let me know if you run into anything else.

CI is running again, it should pass this time.

install_sct
```diff
@@ -165,14 +171,18 @@ function check_requirements() {
     print warning "WARNING: \"gcc\" is not installed."
     if [[ $OS == "osx" ]]; then
       while [[ ! $GCC_INSTALL =~ ^([Yy](es)?|[Nn]o?)$ ]]; do
-        print question "Do you want to install it now? (accepting to install \"gcc\" will also install \"brew\" in case it is not installed already)? [y]es/[n]o: "
-        read GCC_INSTALL
+        GCC_INSTALL=yes # rude?
+        if [ -z "$NONINTERACTIVE" ]; then
+          print question "Do you want to install it now? (accepting to install \"gcc\" will also install \"brew\" in case it is not installed already)? [y]es/[n]o: "
+          read GCC_INSTALL
+        fi
```
joshuacwnewton, line 178: I think I agree with the `# rude?` note here. I think I want to keep this prompt. The intent of `-y` is usually just to silence the 3 most common questions. Any issues with `gcc` would be a rare occurrence, so the interruption won't be too bothersome, even with `-y`. 

Plus, installing `gcc`+`homebrew` is a bit heavier, so I think the user should have the option to decline regardless of `-y`.

joshuacwnewton (commented):
Mostly looks good to me! :slightly_smiling_face:

install_sct
kousu: I meant `-y` to run the whole thing interactively, which means either the script should complete successfully or it should error; it shouldn't hang in the middle.

So in that case I would set the default to `no` I think. In the rare case that someone is installing non-interactively (for CI or deploying to a computing cluster or something) and also doesn't have gcc installed already I think we should just let it fail. We should give it a shot to see what that looks like on a fresh macOS VM; maybe we can print a warning like "Warning: gcc is not installed. The installation might fail." 

Julien's attempt last night to actually use this in #3108 makes me think it's -very very rare-.

kousu: I really think this entire code block should go away so I don't want to maintain it too hard.

We should be using better packaging tools instead of trying to futz around with this install script. We can package SCT *in* brew itself. If we're going to potentially make people install brew anyway let's just do that.

joshuacwnewton: > I meant -y to run the whole thing interactively, which means either the script should complete successfully or it should error; it shouldn't hang in the middle.

Ah, that makes a lot more sense. My usage was similar to Julien's... sometimes I am too lazy to want to answer all of the questions when I just want the whole thing to run in the background. :sweat_smile: 

> maybe we can print a warning like "Warning: gcc is not installed. The installation might fail."

I think this is a good idea! Good thing we already have that warning. :stuck_out_tongue_winking_eye: 

https://github.com/neuropoly/spinalcordtoolbox/blob/3761f16569bf44df819e39127665e7dab2c5a432/install_sct#L171

joshuacwnewton (approved):
LGTM! But, huh! The CentOS build failed both here _and_ #3103. ~~Maybe there's a deeper problem here...~~

Ah, NVM, restarting #3103 worked fine. :)

joshuacwnewton (dismissed):
Re-adding my approval on this as it looks like it does indeed fix #3112.

install_sct
kousu: I changed the default to "no"!

kousu:
Passing now; I did a run with the complete suite: https://github.com/neuropoly/spinalcordtoolbox/pull/3102/checks?check_run_id=1609732229

It failed on that one old macOS, but that's that libomp thing again (https://github.com/scikit-learn/scikit-learn/issues/19063, #3124). I think I'm ready to call this done.

install_sct
```diff
@@ -636,16 +656,20 @@ python -c 'import spinalcordtoolbox.deepseg.models; spinalcordtoolbox.deepseg.mo
...
 # In case of previous SCT installation (4.0.0-beta.1 or before), remove sct_env declaration in bashrc
 print info "In case an old version SCT is already installed (4.0.0-beta.1 or before), remove 'sct_env' declaration in RC file"
-if [[ $OS == "osx" ]]; then
-  sed -ie '/sct_env/ s/^#*/#/' $RC_FILE_PATH
+if [[ -e "$RC_FILE_PATH" && $OS == "osx" ]]; then
+  sed -ie '/sct_env/ s/^#*/#/' "$RC_FILE_PATH"
 else
-  sed -e '/sct_env/ s/^#*/#/' -i $RC_FILE_PATH
+  sed -e '/sct_env/ s/^#*/#/' -i "$RC_FILE_PATH"
 fi
```
joshuacwnewton, line 663: The `-e` existence check was added to the `osx` case but not the else. Would nested ifs be better? e.g.

```bash
if [[ -e "$RC_FILE_PATH" ]]; then
  if [[ $OS == "osx" ]]; then
    sed -ie '/sct_env/ s/^#*/#/' "$RC_FILE_PATH"
  else
    sed -e '/sct_env/ s/^#*/#/' -i "$RC_FILE_PATH"
  fi
fi
```

install_sct
```diff
@@ -335,7 +345,8 @@ for arg in "$@"; do
   esac
 done
 
-while getopts ":dhbpv" opt; do
+NONINTERACTIVE=""
+while getopts ":dhbpvy" opt; do
```
joshuacwnewton, line 349: `-p` isn't used in the case statement, is it?

```suggestion
while getopts ":dhbvy" opt; do
```

joshuacwnewton (commented):
Some minor observations. Everything else looked good to me. :)

install_sct
kousu: oh good catch!

joshuacwnewton (changes requested):
Ah, looks like #3128 has cropped up here as well.

https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/464972811#L1354-L1366

kousu:
😝 Thanks! I'll get it. Sun Dec 27 16:06:27 EST 2020 neuropoly/spinalcordtoolbox <reply@reply.github.com>: @joshuacwnewton requested changes on this pull request. Ah, looks like #3128 has cropped up here as well. https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/464972811#L1354-L1366 In install_sct: > @@ -636,16 +657,22 @@ python -c 'import spinalcordtoolbox.deepseg.models; spinalcordtoolbox.deepseg.mo # In case of previous SCT installation (4.0.0-beta.1 or before), remove sct_env declaration in bashrc print info "In case an old version SCT is already installed (4.0.0-beta.1 or before), remove 'sct_env' declaration in RC file" -if [[ $OS == "osx" ]]; then - sed -ie '/sct_env/ s/^#*/#/' $RC_FILE_PATH -else - sed -e '/sct_env/ s/^#*/#/' -i $RC_FILE_PATH -fi +if [[ -e "$RC_FILE_PATH" ]]; then + if [[ "$OS" == "osx" ]]; then + sed -ie '/sct_env/ s/^#*/#/' "$RC_FILE_PATH" + else + sed -e '/sct_env/ s/^#*/#/' -i "$RC_FILE_PATH" + fi +fo Fee fi fo fum 😉 ⬇️ Suggested change -fo +fi — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

kousu:
This has grown a bit beyond my original intentions. Sorry about that! I was annoyed at the hour it takes for Travis to check each commit that I started throwing everything into a giant PR instead of separating them.

kousu:
I can pull this apart if I need to to get it re-reviewed. It'll just take some time fiddling with git.

install_sct
```diff
@@ -365,16 +382,19 @@ done
...
-  SCT_VERSION=$(cat spinalcordtoolbox/version.txt)
+  SCT_VERSION="$(cat spinalcordtoolbox/version.txt)"
 else
   die "ERROR: version.txt not found. \n
 The install_sct script must be executed from the source directory"
 fi
 
 # Get installation type (from git or from package)
-if [[ "x$SCT_INSTALL_TYPE" == "x" ]]; then
+if [[ -z "${SCT_INSTALL_TYPE:-}" ]]; then
```
kousu, line 397: For consistency this one should NOT use `:-`, because it's one of our internal vars and we should be able to initialize it ourselves.

install_sct
```diff
@@ -552,45 +588,38 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
...
-  yes | pip uninstall tensorflow-tensorboard &&
-  yes | pip uninstall tensorboard &&
-  pip install tensorboard
-if [[ $? != 0 ]]; then
+  pip uninstall -y tensorflow-tensorboard &&
+  pip uninstall -y tensorboard &&
+  pip install tensorboard &&
+  # install sct itself
+  print info "Installing spinalcordtoolbox..." &&
+  pip install -e .
```
kousu, line 613: It's redundant to run `pip install $SCT_DIR`  and also `pip install -r requirements.txt`, but that's how it was when I found it.

Also, `pip install -e`? That's really more meant for developers, not for installations. But we can deal with that in #1526. pip-only install! right @Drulex ?

install_sct
```diff
@@ -601,25 +630,24 @@ export PATH="$SCT_DIR/$BIN_DIR:$PATH"
 # ----------------------------------------------------------------------------------------------------------------------
 
 # Install binaries
-if [[ $NO_SCT_BIN_INSTALL ]]; then
+if [[ -n "${NO_SCT_BIN_INSTALL:-}" ]]; then
```
kousu, line 633: ditto, this shouldn't have to have `:-}` on it since it's an internal var.

install_sct
```diff
@@ -601,25 +630,24 @@ export PATH="$SCT_DIR/$BIN_DIR:$PATH"
...
   print info "Installing binaries..."
-  run "sct_download_data -d binaries_${OS} -o ${SCT_DIR}/${BIN_DIR} -k" || \
-      die "Unsupported OS $OS: can't install binaries."
+  run sct_download_data -d binaries_${OS} -o ${SCT_DIR}/${BIN_DIR} -k
 fi
 print info "All requirements installed!"
 
 # Install data
-if [[ $NO_DATA_INSTALL ]]; then
+if [[ -n "${NO_DATA_INSTALL:-}" ]]; then
```
kousu, line 642: ditto

install_sct
```diff
@@ -77,33 +90,34 @@ function die() {
...
+  if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
-  elif [[ $value -eq 99 ]]; then
+  elif [[ "$value" -eq 99 ]]; then
     # Showing usage with -h
     echo ""
   else
     print error "Installation failed!\n
-Please copy the historic of this Terminal (starting with the command install_sct) and paste it in a new created topic on SCT's forum:\n
+Please copy the output of this Terminal (starting with the command install_sct) and upload it as a .txt attachment in a new topic on SCT's forum:\n
```
kousu, line 115: @jcohenadad please vet this wording change. I recommend it, but it doesn't have to go in.

jcohenadad: good idea, i was getting increasingly annoyed by the large copy/paste without respecting markdown "code" format

install_sct
```diff
@@ -601,25 +634,24 @@ export PATH="$SCT_DIR/$BIN_DIR:$PATH"
...
 
 # Install binaries
-if [[ $NO_SCT_BIN_INSTALL ]]; then
+if [[ -n "${NO_SCT_BIN_INSTALL:-}" ]]; then
   print warning "WARNING: SCT binaries will not be (re)-installed"
 else
   print info "Installing binaries..."
-  run "sct_download_data -d binaries_${OS} -o ${SCT_DIR}/${BIN_DIR} -k" || \
-      die "Unsupported OS $OS: can't install binaries."
+  run sct_download_data -d binaries_${OS} -o ${SCT_DIR}/${BIN_DIR} -k
```
kousu, line 641: ack, forgot to quote this
```suggestion
  run sct_download_data -d "binaries_${OS}" -o "${SCT_DIR}/${BIN_DIR}" -k
```

install_sct
kousu:
```suggestion
if [[ -n "$NO_DATA_INSTALL" ]]; then
```

install_sct
kousu:
```suggestion
if [[ -n "$NO_SCT_BIN_INSTALL" ]]; then
```

.ci.sh
```diff
@@ -1,11 +1,15 @@
...
+# https://sipb.mit.edu/doc/safe-shell/
+set -eo pipefail  # exit if non-zero error is encountered (even in a pipeline)
+set -u            # exit if unset variables used
+shopt -s failglob # error if a glob doesn't find any files, instead of remaining unexpanded
 
 echo Installing SCT
-yes | ASK_REPORT_QUESTION=false PIP_PROGRESS_BAR=off ./install_sct
+PIP_PROGRESS_BAR=off ./install_sct -y
 echo $?
 echo "... STATUS"
```
kousu, line 14: These should go; set -e supersedes them.
```suggestion
```

install_sct
```diff
@@ -335,25 +351,30 @@ for arg in "$@"; do
...
 done
 
-while getopts ":dhbpv" opt; do
+NONINTERACTIVE=""
+while getopts ":hydbv" opt; do
   case $opt in
+  y)
+    echo " non-interactive mode"
+    NONINTERACTIVE="yes"
+  ;;
```
kousu, line 360:
```suggestion
    ;;
```

install_sct
kousu: Also this should be exposed as another option flag.

joshuacwnewton:
> update deprecated brew installer; fixes #3108 .

Tiny note: This is pointing to the PR, not the issue. I've edited this so that the proper issue is linked. 

Other tiny note: "Fixes <issue>" all have to be on their own line for GH to autolink. Silly GH!

kousu:
Whoopsies. Thanks!

install_sct
```diff
@@ -218,74 +234,77 @@ if [[ -n \"\$BASH_VERSION\" ]]; then
...
 
 # Usage of this script
 function usage() {
-  echo -e "\nUsage: $0 [-d] [-b] [-v]" 1>&2
-  echo -e "\nOPTION"
-  echo -e "\t-d \v Prevent the (re)-installation of the \"data/\" directory "
-  echo -e "\n\t-b \v Prevent the (re)-installation of the SCT binaries files "
-  echo -e "\n\t-v \v Full verbose"
+  # extract the usage block from our own header
+  awk '
```
joshuacwnewton, line 285: This is really nifty. Slight nitpick:

<details>
<summary> Output of <code>./install_sct -h</code> </summary>

```
(venv_sct) joshua@XPS-15-9560:~/repos/spinalcordtoolbox$ ./install_sct -h


USAGE
  ./install_sct [-h] [-i] [-y] [-d] [-b] [-v]

OPTIONS
 -h   Show this help
 -i   Install in-place; this is the default when working from git.
 -y   Install without interruption with 'yes' as default answer
 -d   Prevent the (re)-installation of the data/ directory
 -b   Prevent the (re)-installation of the SCT binaries files
 -v   Full verbose



```
</details>

How much do we care about it printing `Installation finished successfully!`?

install_sct
```diff
@@ -335,25 +351,36 @@ for arg in "$@"; do
   esac
 done
 
-while getopts ":dhbpv" opt; do
+SCT_INSTALL_TYPE=""
+NONINTERACTIVE=""
+NO_DATA_INSTALL=""
+NO_SCT_BIN_INSTALL=""
```
joshuacwnewton, line 355: We set a bunch of other variables higher up. Would it be worth initializing these there, too?

https://github.com/neuropoly/spinalcordtoolbox/blob/281e85da979e5973653584da7370d74d39ab7470/install_sct#L40-L48

joshuacwnewton (commented):
This is a massive effort, and a wonderful bit of cleaning up. It's very satisfying to see shellcheck returning no errors or warnings. (I tried doing it on `master` and... there are a _lot_ of extra warnings you fixed!)

Apart from my minor comments, I think it looks good to go. :slightly_smiling_face:

install_sct
kousu: Hmmm you're probably right.

install_sct
kousu: oh, that's annoying. drat. let's fix that.

Maybe I should move the `trap finish EXIT` line to after the CLI parsing?

kousu: If this was better structured there would be a `install()` function, and that would run a subshell, and that subshell would define `finish` and run `trap finish EXIT`, not the overall script.

But I think just moving where the trapping happens will fix it.

kousu: I am super proud of this awk script! awk is underappreciated.

kousu: FIxed in dd187ba5f3308eda73fe95d40157dd9739e089b8

```
$ ./install_sct -h


USAGE
  ./install_sct [-h] [-i] [-y] [-d] [-b] [-v]

OPTIONS
 -h   Show this help
 -i   Install in-place; this is the default when working from git.
 -y   Install without interruption with 'yes' as default answer
 -d   Prevent the (re)-installation of the data/ directory
 -b   Prevent the (re)-installation of the SCT binaries files
 -v   Full verbose



```

vs

```
$ ./install_sct -h
USAGE
  ./install_sct [-h] [-i] [-y] [-d] [-b] [-v]

OPTIONS
 -h   Show this help
 -i   Install in-place; this is the default when working from git.
 -y   Install without interruption with 'yes' as default answer
 -d   Prevent the (re)-installation of the data/ directory
 -b   Prevent the (re)-installation of the SCT binaries files
 -v   Full verbose

```

install_sct
kousu: Fixed in a618ee30b3484fb0da2cbc95df701f858140d4ad

kousu:
@joshuacwnewton it's done! If you want, double-check the commits I linked in your comments and then resolve them, and let's put it in :)

joshuacwnewton (approved):
Thank you again for all your hard work! This is a lovely collection of improvements.

jcohenadad:
> Thank you again for all your hard work! This is a lovely collection of improvements.

Ditto! Thank you so much @kousu

PR #3103, opened 2020-12-16
Check that conda actually activated during install.

kousu:
This glues over a rare glitch in conda (#3029, https://github.com/conda/conda/issues/10401).

Reviewers: please run this under the bug we reproduced in #3029 to make sure it handles that well.

We might also check for: `pip`, or `python3`? Open to feedback.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -288,6 +288,22 @@ function usage() {
   echo -e "\n\t-v \v Full verbose"
 }
 
+if [ "$(uname)" = "Darwin" ]; then
+  # macOS polyfills
+
+  # Linux has `realpath` and `readlink -f`.
+  # BSD has `readlink -f`.
+  # macOS has neither: https://izziswift.com/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac/
+  # even though it *has* the function in its libc: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/realpath.3.html
+  # Someone even wrote a whole C program just for this: https://github.com/harto/realpath-osx/.
+  # https://stackoverflow.com/a/3572105 suggests some bash trickery but it
+  # seems pretty fragile and that it probably doesn't actually expand symlinks.
+  # So here, solve it with python. It's not a great either, but since much of
+  # this script already is just calling python it's fairly safe.
+  realpath() {
+    python -c 'import sys, path; for f in sys.argv[1:]: print(os.path.realpath(f))' "$@"
+  }
+fi
 
 # ======================================================================================================================
 # SCRIPT STARTS HERE
--- a/install_sct
+++ b/install_sct
@@ -527,7 +527,7 @@ yes | python/bin/conda create -n venv_sct python=3.6
 # activate miniconda
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
-if [ "$(which python)" != "$(which python/envs/venv_sct/bin/python)" ]; then
+if [ "$(command -v python)" != "$(command -v python/envs/venv_sct/bin/python)" ]; then
   echo "error: venv_sct failed to activate." >&2
   exit 1
 fi
--- a/install_sct
+++ b/install_sct
@@ -300,7 +300,7 @@ if [ "$(uname)" = "Darwin" ]; then
   # seems pretty fragile and that it probably doesn't actually expand symlinks.
   # So here, solve it with python. It's not great either, but since much of
   # this script already is just calling python at least it's reliable.
-  realpath() {
+  (command -v realpath >/dev/null) || realpath() {
     python -c 'import sys, path; for f in sys.argv[1:]: print(os.path.realpath(f))' "$@"
   }
 fi
--- a/install_sct
+++ b/install_sct
@@ -527,8 +527,9 @@ yes | python/bin/conda create -n venv_sct python=3.6
 # activate miniconda
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
+# double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
 if [ "$(command -v python)" != "$(command -v python/envs/venv_sct/bin/python)" ]; then
-  echo "error: venv_sct failed to activate." >&2
+  echo "error: venv_sct failed to activate. Do you have a conflicting python installation?" >&2
   exit 1
 fi
 
--- a/install_sct
+++ b/install_sct
@@ -528,7 +528,7 @@ yes | python/bin/conda create -n venv_sct python=3.6
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
 # double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
-if [ "$(command -v python)" != "$(command -v python/envs/venv_sct/bin/python)" ]; then
+if [ "$(realpath "$(command -v python)")" != "$(realpath "$(command -v python/envs/venv_sct/bin/python)")" ]; then
   echo "error: venv_sct failed to activate. Do you have a conflicting python installation?" >&2
   exit 1
 fi
--- a/install_sct
+++ b/install_sct
@@ -298,8 +298,8 @@ if [ "$(uname)" = "Darwin" ]; then
   # Someone even wrote a whole C program just for this: https://github.com/harto/realpath-osx/.
   # https://stackoverflow.com/a/3572105 suggests some bash trickery but it
   # seems pretty fragile and that it probably doesn't actually expand symlinks.
-  # So here, solve it with python. It's not a great either, but since much of
-  # this script already is just calling python it's fairly safe.
+  # So here, solve it with python. It's not great either, but since much of
+  # this script already is just calling python at least it's reliable.
   realpath() {
     python -c 'import sys, path; for f in sys.argv[1:]: print(os.path.realpath(f))' "$@"
   }
@@ -544,8 +544,10 @@ yes | python/bin/conda create -n venv_sct python=3.6
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
 # double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
-if [ "$(realpath "$(command -v python)")" != "$(realpath "$(command -v python/envs/venv_sct/bin/python)")" ]; then
-  echo "error: venv_sct failed to activate. Do you have a conflicting python installation?" >&2
+EXPECTED_PYTHON="$(realpath "$(command -v python/envs/venv_sct/bin/python)")"
+ACTUAL_PYTHON="$(realpath "$(command -v python)")"
+if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
+  echo "Error: Activating venv_sct failed to switch Python interpreter to Miniconda. (Incorrect interpreter loaded: $ACTUAL_PYTHON)" >&2
   exit 1
 fi
 
--- a/install_sct
+++ b/install_sct
@@ -527,6 +527,11 @@ yes | python/bin/conda create -n venv_sct python=3.6
 # activate miniconda
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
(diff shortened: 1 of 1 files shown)


install_sct
```diff
@@ -527,6 +527,11 @@ yes | python/bin/conda create -n venv_sct python=3.6
 # activate miniconda
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
+if [ "$(which python)" != "$(which python/envs/venv_sct/bin/python)" ]; then
```
joshuacwnewton, line 530: On my Ubuntu machine, `which python/envs/venv_sct/bin/python` still only yields the relative path (hence the CI errors). 

Might be better just to use the explicit path we expect.

```suggestion
if [ "$(which python)" != "$SCT_DIR/python/envs/venv_sct/bin/python" ]; then
```

joshuacwnewton, line 530: Alternatively, a different check could be:

```suggestion
if [ "$(echo $PATH | cut -f1 -d:)" != "$SCT_DIR/python/envs/venv_sct/bin" ]; then
```

Which covers all the related binaries.

joshuacwnewton: Alternatively, a different check could be:

```suggestion
if [ "$(echo $PATH | cut -f1 -d:)" != "$SCT_DIR/python/envs/venv_sct/bin" ]; then
```

Which covers all the related binaries.

joshuacwnewton: Also, I came across an interesting discussion thread on `which`: https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then

kousu: Drat. On my Arch machine which gives me absolute paths:

```
[kousu@requiem spinalcordtoolbox]$ which python/envs/venv_sct/bin/python
/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python
```

Okay so they suggest `command -v`?

What about readlink or realpath?

```
[kousu@requiem spinalcordtoolbox]$ readlink -f python/envs/venv_sct/bin/python
/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python3.6
```

```
[kousu@requiem spinalcordtoolbox]$ realpath python/envs/venv_sct/bin/python
/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python3.6
```

but that, as in this example, might still not be quite right because "3.6" got appended there.

Oh but running the same on both sides should work, shouldn't it?

```suggestion
[kousu@requiem spinalcordtoolbox]$ if [ "$(command -v /usr/bin/python)" = "$(command -v python3)" ]; then echo Aye; fi
[kousu@requiem spinalcordtoolbox]$ 
```

Hm nope. Why? Even though these are the same program, `command` doesn't resolve symlinks:

```
[kousu@requiem spinalcordtoolbox]$ command -v python
/usr/bin/python
[kousu@requiem spinalcordtoolbox]$ command -v python3
/usr/bin/python3
[kousu@requiem spinalcordtoolbox]$ ls -l $(command -v python)
lrwxrwxrwx 1 root root 7 Dec 13 06:55 /usr/bin/python -> python3
[kousu@requiem spinalcordtoolbox]$ ls -l $(command -v python3)
lrwxrwxrwx 1 root root 9 Dec 13 06:55 /usr/bin/python3 -> python3.9
```

Both, then?

```
[kousu@requiem spinalcordtoolbox]$ if [ "$(readlink -f "$(command -v /usr/bin/python)")" = "$(readlink -f "$(command -v python3)")" ]; then echo Aye; fi
Aye
```

Or, I guess we don't really need to worry about symlinks? We just want to know we're pointed at the right subfolder.

kousu: Tried one of them. Gonna be lazy and wait for CI to see if it worked.

kousu:
Should we give a link in the error message? Right now it's a dead end for a user who does happen to run into this. If we had a troubleshooting page we could point them at that would be ideal.

kousu:
Or maybe we should use your fix of just running `conda deactivate` unilaterally before loading the new conda?

install_sct
kousu: Bleh not yet, okay I'll take another swipe at this.

kousu: Okay the problem with `command -v` is the same as with `which`: it doesn't get the absolute path necessarily:

```
$ command -v python
/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python
$ command -v python/envs/venv_sct/bin/python
python/envs/venv_sct/bin/python
```

kousu: Okay I solved it by stacking `realpath` on `command -v`. Here's what the check looks like on my machine:

```
+ '[' /home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python3.6 '!=' /home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python3.6 ']'
```

Unfortunately, `realpath` doesn't exist on macOS:

- https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
- macOS's libc *has* `realpath()`: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/realpath.3.html; but it doesn't exist expose it; someone wrote a whole C program just to expose it: https://github.com/harto/realpath-osx/blob/master/realpath.c

joshuacwnewton:
> Should we give a link in the error message? Right now it's a dead end for a user who does happen to run into this. If we had a troubleshooting page we could point them at that would be ideal

Hmm... all of the `install_sct` errors I've seen in the past were trapped by this:

https://github.com/neuropoly/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/install_sct#L298-L299

Which leads to this error message:

https://github.com/neuropoly/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/install_sct#L100-L102

I was thinking the user would come to us to troubleshoot on the forum, as future errors could be as obscure as this one was. :sweat_smile:

install_sct
joshuacwnewton: Ah, whoops, I think there might have been a miscommunication on my part. Sorry about that. :sweat_smile: 

1. `command -v python` give the absolute path of `python`.
2. `command -v python/envs/venv_sct/bin/python` is the one that gives the relative path.

So, 1. is fine to leave as-is. For 2., since we already know the full path that we're expect, can't we just use the filepath as-is? (See https://github.com/neuropoly/spinalcordtoolbox/pull/3103#discussion_r545236719) That way, no `realpath` shenanigans are needed, I think.

kousu: No apologies needed. Really!

My thinking is that I want something that reliably gives the same format. Neither `which` nor `command` do that; because their algorithms are basically: try concatenating to each thing in $PATH until something catches and output. Plus symlinks can get involved (they often do: `$path/$to/python` is usually linked to `$path/$to/python3` is usually linked to `$path/$to/python3.7`) and that makes comparisons more unreliable.

joshuacwnewton: Well, even if symlinks are involved, won't `$path/$to/python`, `$path/$to/python3`, and `$path/$to/python2.7` all be `!=` to the Miniconda python if something's amiss? Even if we don't have the perfectly accurate python interpreter on the left side, we're only really checking 1 possible path on the right side. :yum:

kousu: I dunno. Maybe you're right. But in my experience there's too many ways you can get tripped up by thinking you know how things work and that they'll work in all cases and then they don't. [realpath(3)](https://www.man7.org/linux/man-pages/man3/realpath.3.html) is guaranteed to give a canonical answer, so we'll know.

The only way I can think it might fail is if hardlinks are involved too, but then I'd debate even more...

joshuacwnewton: OK, I'm sold. :slightly_smiling_face:

install_sct
```diff
@@ -527,6 +543,12 @@ yes | python/bin/conda create -n venv_sct python=3.6
 # activate miniconda
 source python/etc/profile.d/conda.sh
 conda activate venv_sct
+# double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
+if [ "$(realpath "$(command -v python)")" != "$(realpath "$(command -v python/envs/venv_sct/bin/python)")" ]; then
+  echo "error: venv_sct failed to activate. Do you have a conflicting python installation?" >&2
```
joshuacwnewton, line 548: I think `Do you have a conflicting python installation?` might not be informative enough. Perhaps we could give the value of `$(realpath "$(command -v python)` in the error message?  

```suggestion
# double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
export EXPECTED_PYTHON = "$(realpath "$(command -v python/envs/venv_sct/bin/python)")"
export ACTUAL_PYTHON = "$(realpath "$(command -v python)")"
if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
  echo "Error: Activating venv_sct failed to switch Python interpreter to Miniconda. (Incorrect interpreter used: $ACTUAL_PYTHON)" >&2
```

joshuacwnewton: Side question about shell scripts: Would it be necessary to use `export` here in my suggestion? (See also https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export)

joshuacwnewton (changes requested):
I tried running this on my Anaconda+pyenv Ubuntu setup (that had reproduced the error), and this new error was thrown successfully, so the change looks good to me. See: [venv_sct-failed-to-activate-install-log.txt](https://github.com/neuropoly/spinalcordtoolbox/files/5716962/venv_sct-failed-to-activate-install-log.txt)

Relevant bit here:

```
# To activate this environment, use
#
#     $ conda activate venv_sct
#
# To deactivate an active environment, use
#
#     $ conda deactivate

error: venv_sct failed to activate. Do you have a conflicting python installation?

Installation failed!

Please copy the historic of this Terminal (starting with the command install_sct) and paste it in a new created topic on SCT's forum:

```

I just have one nitpick about the wording.

install_sct
kousu: `export` means that subprocesses receive it in their `os.environ` (or equivalent for whatever lang you're working in). People think `export` means `set` but it doesn't, the `=` means `set`. Also, you can set env for a specific subprocess like this:

```
DEBIAN_FRONTEND=noninteractive SOMEOTHERVAR=somethingelse apt-get install -y
```

That's the same as quoting like:

```

```
(
  export DEBIAN_FRONTEND=noninteractive
  export SOMEOTHERVAR=somethingelse
  apt-get install -y
)
```

In general I'd only use `export` if I really wanted to set something globally, like if I was running `apt` multiple times in a script or needed to change where `TMPDIR` was.

Comment (kousu) [reply install_sct:548]: Whiiiich is all to say that I'll use this instead:

(another shell gotcha: you can't have spaces around `=`!!)

```suggestion
# double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
EXPECTED_PYTHON="$(realpath "$(command -v python/envs/venv_sct/bin/python)")"
ACTUAL_PYTHON="$(realpath "$(command -v python)")"
if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
  echo "Error: Activating venv_sct failed to switch Python interpreter to Miniconda. (Incorrect interpreter used: $ACTUAL_PYTHON)" >&2
```

kousu:
@joshuacwnewton applied your suggestion, good now?
(i..haven't actually tested yet. i'm gonna go catch some sun at the moment)

joshuacwnewton (approved):
> @joshuacwnewton applied your suggestion, good now?
> (i..haven't actually tested yet. i'm gonna go catch some sun at the moment)

LGTM! The only weird thing is the CentOS build stalled... I've restarted it just now. Assuming it goes well, then feel free to merge.

install_sct
kousu: Looks like you were on the nose with this @joshuacwnewton: `which` is being **deleted**: https://lwn.net/SubscriberLink/874049/bf89a969ed3dde87/, https://lwn.net/ml/debian-devel/9e3a2df146f73437fe8cce10272ddf3913e48757.camel@debian.org/

joshuacwnewton: Oh my!

PR #3104, opened 2020-12-16
Forces output label image to be UINT8 with -create-viewer

### LINKED ISSUE #3101: Created labels are not integer
### Description

When creating manual labels, the created labels do not have the value asked for: they are output in float (it whould be integer) and the value is slightly different (precision issue). 

User forum: https://forum.spinalcordmri.org/t/sct-register-to-template-source-and-destination-landmarks-are-not-the-same/590/4

### Steps to Reproduce

Data: [t2.nii.gz](https://github.com/neuropoly/spinalcordtoolbox/files/5704767/t2.nii.gz)

~~~
sct_label_utils -i t2.nii.gz -create-viewer 3,9 -ldisc t2_labels_disc_manual.nii.gz 
sct_label_utils -i t2_labels_disc_manual.nii.gz -display
~~~

output:
~~~
Position=(214,298,7) -- Value= 2.999999999301508
Position=(199,103,7) -- Value= 8.999999997904524
~~~

Expected output:
~~~
Position=(214,298,7) -- Value= 3
Position=(199,103,7) -- Value= 9
~~~

Interestingly, when doing the same experiment on the `sct_testing_data/t2/t2.nii.gz`, it works as expected.

jcohenadad:
When creating manual labels, the created labels do not have the value asked for: they are output in float (it whould be integer) and the value is slightly different (precision issue). This bug was introduced in v5.0.0.

This PR fixes the problem by forcing the output label to be UINT8 type.

Also see user forum discussion: https://forum.spinalcordmri.org/t/sct-register-to-template-source-and-destination-landmarks-are-not-the-same/590/4

Fixes #3101

### DIFF
--- a/spinalcordtoolbox/scripts/sct_label_utils.py
+++ b/spinalcordtoolbox/scripts/sct_label_utils.py
@@ -345,7 +345,7 @@ def launch_sagittal_viewer(img: Image, labels: Sequence[int], msg: str, previous
     if previous_points is not None:
         params.message_warn = 'Please select the label you want to add \nor correct in the list below before clicking \non the image'
 
-    out = zeros_like(img)
+    out = zeros_like(img, dtype='uint8')
     out.absolutepath = params.output_file_name
     launch_sagittal_dialog(img, out, params, previous_points)
 


jcohenadad:
With the proposed change, the following warning is spitted out, which is annoying:
~~~
To avoid intensity overflow due to convertion to +uint8+, intensity will be rescaled to the maximum quantization scale
~~~

This issue should be avoided by bypassing the chunk of code from this function:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/image.py#L1089

We don't need to do all these image scaling, because the input is a zero matrix (!). Any idea how to solve this problem elegantly @joshuacwnewton @kousu ?  Also, we should probably move that to a separate issue...

joshuacwnewton:
> We don't need to do all these image scaling, because the input is a zero matrix (!). Any idea how to solve this problem elegantly @joshuacwnewton @kousu ? Also, we should probably move that to a separate issue...

I think this happens because the steps in `zeros_like` happen in the following order:

1. Change the datatype (with original data intact)
2. Set data to 0

https://github.com/neuropoly/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/spinalcordtoolbox/image.py#L1216-L1229

The reason to do it in this order is to `"avoid doing a copy, which is slower than initialization with a constant."`. But, that doesn't make sense to me, because a copy is already being made inside `change_type`! :sweat_smile: 

https://github.com/neuropoly/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/spinalcordtoolbox/image.py#L1113-L1115

Since a copy is made either way, I think it could be more elegant for `zeros_like` to be ordered like this:

1. Make a copy
2. Set data to 0
3. Change the datatype

I don't think that this will be any slower, and it will also prevent the warning from occurring.

jcohenadad:
@joshuacwnewton agreed! would you mind opening an issue (and take on the issue if you like?) many thanks!

joshuacwnewton (approved):

PR #3108, opened 2020-12-17
Switch to brew's bash-installer.

### LINKED ISSUE #3105: install_sct: homebrew's installer is now in bash

kousu:
This will conflict with #3102 but :shrug: . I don't want to force one or the other to get merged before the other. Just fix it up when you get to that point.

Fixes #3105.

**EDIT**: this now depends on #3102. Julien found that `yes |` wasn't working with their installer anymore (https://github.com/neuropoly/spinalcordtoolbox/pull/3108#discussion_r545541225), and since I fixed that in that PR it makes the most sense to just merge that first and then this.
Marking this draft until that's merged to avoid an accident.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -433,17 +433,17 @@ fi
 
 # Set install dir
 while true; do
-  change_default_path=""
-  while [[ ! $change_default_path =~ ^([Yy](es)?|[Nn]o?)$ ]]; do
+  keep_default_path=""
+  while [[ ! $keep_default_path =~ ^([Yy](es)?|[Nn]o?)$ ]]; do
     print info "SCT will be installed here: [$SCT_DIR]"
-    change_default_path=no
-    if [ -z "$NONINTERACTIVE"]; then
+    keep_default_path=yes
+    if [ -z "$NONINTERACTIVE" ]; then
       print question "
 Do you agree? [y]es/[n]o: "
-      read change_default_path
+      read keep_default_path
     fi
   done
-  if [[ $change_default_path =~ ^[Yy] ]]; then
+  if [[ $keep_default_path =~ ^[Yy] ]]; then
     # user accepts default path --> exit loop
     break
   fi
--- a/install_sct
+++ b/install_sct
@@ -550,7 +550,10 @@ yes | python/bin/conda create -n venv_sct python=3.6
 
 # activate miniconda
 source python/etc/profile.d/conda.sh
+set +u #disable safeties, for conda is not written to their standard.
 conda activate venv_sct
+set -u # reactivate safeties
+
 # double-check conda activated (it can be glitchy): https://github.com/neuropoly/spinalcordtoolbox/issues/3029
 EXPECTED_PYTHON="$(realpath "$(command -v python/envs/venv_sct/bin/python)")"
 ACTUAL_PYTHON="$(realpath "$(command -v python)")"
--- a/install_sct
+++ b/install_sct
@@ -176,9 +176,9 @@ function check_requirements() {
       done
       if [[ $GCC_INSTALL =~ [Yy](es)? ]]; then
         if [[ ! $(which brew) ]]; then
-          yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+          NONINTERACTIVE=1 /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
         fi
-        yes | brew install gcc
+        brew install -f gcc
         # check if gcc install ran properly
         gcc --version > /dev/null 2>&1  # run silently, then check output status
         if [[ $? -ne 0 ]]; then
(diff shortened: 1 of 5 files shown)


install_sct
```diff
@@ -170,7 +170,7 @@ function check_requirements() {
       done
       if [[ $GCC_INSTALL =~ [Yy](es)? ]]; then
         if [[ ! $(which brew) ]]; then
-          yes | /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+          yes | /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
jcohenadad, line 173: hum, some users will don't understand this message:
~~~
julien-macbook:~ $ yes | /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
Need sudo access on macOS (e.g. the user julien to be an Administrator)!
~~~

jcohenadad: ahem... and it is failing when calling it with sudo (for another reason):
~~~
julien-macbook:~ $ sudo yes | /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
Password:
==> This script will install:
/usr/local/bin/brew
/usr/local/share/doc/homebrew
/usr/local/share/man/man1/brew.1
/usr/local/share/zsh/site-functions/_brew
/usr/local/etc/bash_completion.d/brew
/usr/local/Homebrew
==> Downloading and installing Homebrew...
remote: Enumerating objects: 3156, done.
remote: Counting objects: 100% (3156/3156), done.
remote: Compressing objects: 100% (21/21), done.
remote: Total 10136 (delta 3141), reused 3143 (delta 3135), pack-reused 6980
Receiving objects: 100% (10136/10136), 2.98 MiB | 7.98 MiB/s, done.
Resolving deltas: 100% (7815/7815), completed with 1254 local objects.
From https://github.com/Homebrew/brew
   e30ed4243..9db324ab7  master          -> origin/master
 * [new branch]          zsh-fpath-fixes -> origin/zsh-fpath-fixes
 * [new tag]             2.5.10          -> 2.5.10
 * [new tag]             2.5.11          -> 2.5.11
 * [new tag]             2.5.12          -> 2.5.12
 * [new tag]             2.5.6           -> 2.5.6
 * [new tag]             2.5.7           -> 2.5.7
 * [new tag]             2.5.8           -> 2.5.8
 * [new tag]             2.5.9           -> 2.5.9
 * [new tag]             2.6.0           -> 2.6.0
 * [new tag]             2.6.1           -> 2.6.1
 * [new tag]             2.6.2           -> 2.6.2
HEAD is now at 9db324ab7 Merge pull request #10045 from jonchang/remove-basic-autho
Error: 
  homebrew-core is a shallow clone.
  homebrew-cask is a shallow clone.
To `brew update`, first run:
  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core fetch --unshallow
  git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask fetch --unshallow
This restriction has been made on GitHub's request because updating shallow
clones is an extremely expensive operation due to the tree layout and traffic of
Homebrew/homebrew-core and Homebrew/homebrew-cask. We don't do this for you
automatically to avoid repeatedly performing an expensive unshallow operation in
CI systems (which should instead be fixed to not use shallow clones). Sorry for
the inconvenience!
Failed during: /usr/local/bin/brew update --force
~~~

jcohenadad: well... this is a particular case because I _already_ had brew... maybe that wouldn't fail if brew was not installed already

kousu: Just about that first part, that's because you'd need to run it as

```
yes | sudo /bin/bash -c ...
```
or

```
sudo sh -c "yes | bash -c ...."
```

Running `yes` as root doesn't grant brew's installer root.

kousu: But the brew installer calls `sudo` inside of itself.

I think the other problem here is that `yes | ` is *not allowed*; that causes stdin to not be a tty (`isatty()`, or `tty -s`, fails), but `sudo` demands having a tty because it wants to prevent as much as possible people trying to save passwords to files and pasting them into it.

In #3102 I replaced `yes |` with using the noninteractive flags all the parts actually wanted, precisely because it's sketchy for this and similar cases.

So maybe #3102 does need to be merged first afterall.

kousu:
I'm pulling this into #3102 because Travis takes so long to run. I don't want to run an hour long CI just for a one line change here.

Issue #2959, opened 2020-10-19
Improvements to `sct_download_data`


Drulex:
Related to #2957, maybe #2788

This is an open discussion. Dumping ideas and thoughts below:

- Implement some kind of caching mechanism to avoid re-downloading the same data.
   - Should verify that data exists locally.
   - Should verify that local data integrity is ok.
   - This means that the data integrity should be published somewhere. Possibly in a manifest alongside the actual data. Alternatively on official website, github repo.
   - Don't really care about security.
- Some questions about current architecture:
   - Is there a reason for users to run `sct_download_data` manually or is usage only required by software? I am questioning if the user script should exist or if an API-only design is ok. Currently `pytest` would need it and in the future possibly `setup.py` as well.
   - Mirrors are currently in the application level (user script). This effectively prevents a piece of code from calling the API directly since it doesn't have access to the list of mirrors.

Drulex:
Comments/ideas welcome @cfhammill @kousu @jcohenadad @joshuacwnewton.

I think that having some kind of manifest file listing the contents of the data package and associated versions and checksums (or similar) is a good idea.

If the data is already in git we could maybe get away without using a manifest depending on how it's packaged/released.

The manifest would definitely facilitate other processes that require some type of version tracking.

kousu:
I prefer not maintaining an ad-hoc package manager. [This project](https://erpnext.com/) wrote [this package manager](https://github.com/frappe/bench) which is based on git and has some clever ideas, but it *only* integrates with [their web framework](https://github.com/frappe/frappe) meaning it's completely non-reusable, and installing their software is a very picky process -- it basically demands an Ubuntu VM, nothing else will do. I think that approach is annoying, and not very FLOSS-friendly (the key point of FLOSS, to me, is that you can readapt work without too many barriers), but I can live with it if we're upfront about it.

Better than that, I think we should make everything in the list [1] into packages under a more common package manager. I think our options in that direction are:

0. package ourselves as [nix packages](https://github.com/NixOS/nixpkgs/)
1. get packaged in [neurodebian](https://salsa.debian.org/neurodebian-team)
2. get packaged as a [conda feedstock](https://conda-forge.org/feedstocks/)
3. make everything into pip wheels

[...]

PR #3116, opened 2020-12-20
Fix incorrect indexing in get_center_spit to prevent QC report cropping

### LINKED ISSUE #3087: Image is cropped on QC report for sct_label_utils
When generating QC report for function `sct_label_utils`, the image is cropped (while it should not):

Screen Shot 2020-12-08 at 5 48 43 PM

here is the original image, indicating that it has indeed been cropped by the QC report:
Screen Shot 2020-12-08 at 5 48 30 PM

To reproduce:
data (internal): `duke/temp/sebeda/test1_30_sub/files_test_qc`
syntax:
~~~
sct_qc -i sub-1002358_T1w_RPI_r_gradcorr.nii.gz -s sub-1002358_T1w_RPI_r_gradcorr_labels-manual.nii.gz -p sct_label_utils -qc qc
~~~
SCT version: git-jca/3044-ants-sampling-770e37f8b25f3283ede65eb171501f55bf5416e5

joshuacwnewton:
#### GitHub

#### PR contents

## Description

The function `get_center_spit` is supposed to return `index: [int] * n_SI` (indices along in the R-L direction for each S-I slice). But, it was incorrectly returning `index: [int] * n_RL` because it was using the wrong axis shape.

This PR fixes the problem so the correct axis shape is used, and also tweaks comments to make this clearer.

It also fixes a bug where `logger.error` was used when an error should have been raised instead.

## Linked issues

Fixes #3087.

### DIFF
--- a/unit_testing/test_reports.py
+++ b/unit_testing/test_reports.py
@@ -15,6 +15,7 @@ def labeled_data_test_params(path_in='sct_testing_data/t2/t2.nii.gz',
     test_sagittal_slice_get_center_spit."""
     im_in = Image(path_in)            # Base anatomical image
     im_seg_labeled = Image(path_seg)  # Base labeled segmentation
+    assert np.count_nonzero(im_seg_labeled.data) >= 2, "Labeled segmentation image has fewer than 2 labels"
 
     # Create image with all but one label removed
     im_seg_one_label = im_seg_labeled.copy()
@@ -34,9 +35,9 @@ def labeled_data_test_params(path_in='sct_testing_data/t2/t2.nii.gz',
 @pytest.mark.parametrize('im_in,im_seg', labeled_data_test_params())
 def test_sagittal_slice_get_center_spit(im_in, im_seg):
     """Test that get_center_split returns a valid index list."""
-    assert im_in.orientation == im_seg.orientation
+    assert im_in.orientation == im_seg.orientation, "im_in and im_seg aren't in the same orientation"
     qcslice = Sagittal([im_in, im_seg], p_resample=None)
-    
+
     if np.count_nonzero(im_seg.data) == 0:
         # If im_seg contains no labels, get_center_spit should fail
         with pytest.raises(ValueError):
@@ -46,4 +47,4 @@ def test_sagittal_slice_get_center_spit(im_in, im_seg):
         index = qcslice.get_center_spit()
         for i, axis in enumerate(im_in.orientation):
             if axis in ['S', 'I']:
-                assert len(index) == im_in.data.shape[i]
+                assert len(index) == im_in.data.shape[i], "Index list doesn't have expected length (n_SI)"
--- a/unit_testing/test_reports.py
+++ b/unit_testing/test_reports.py
@@ -9,43 +9,41 @@ from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.reports.slice import Sagittal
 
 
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
This PR still needs to add tests for this bugfix, so I've left it as a draft PR for now. I'll mark it ready for review as soon as that's done.

joshuacwnewton:
For now, I've written a test that uses the subject data. I'll need to replace this with testing data that replicates the subject data. But, I still need to figure out how to do that. :sweat_smile:

jcohenadad:
> For now, I've written a test that uses the subject data. I'll need to replace this with testing data that replicates the subject data. But, I still need to figure out how to do that. 😅

Maybe the error can be reproduced if the input data is in RPI orientation? The T2w on the testing data is LPI. If my hypothesis is true, an easy way to reorient the input data is using SCT's API [`image.Image.change_orientation()`](https://spinalcordtoolbox.com/en/stable/dev_section/api.html#spinalcordtoolbox.image.Image.change_orientation)

unit_testing/test_reports.py
```diff
@@ -0,0 +1,51 @@
...
+    im_seg_one_label = im_seg.copy()
+    for x, y, z in np.argwhere(im_seg_one_label.data)[1:]:
+        im_seg_one_label.data[x, y, z] = 0
+
+    # Create image with no labels
+    im_seg_no_labels = im_seg.copy()
+    for x, y, z in np.argwhere(im_seg_no_labels.data):
+        im_seg_no_labels.data[x, y, z] = 0
+
+    return [im_seg, im_seg_one_label, im_seg_no_labels]
```
joshuacwnewton, line 28: I wanted to use this image list with [`pytest.mark.parametrize`](https://docs.pytest.org/en/stable/parametrize.html). 

But, this usage seems to be prevented by this long-standing highly-requested issue, https://github.com/pytest-dev/pytest/issues/349.

joshuacwnewton: https://github.com/pytest-dev/pytest/issues/349#issuecomment-112203543 suggests calling it as a function like this:

```python
@pytest.mark.parametrize('im_seg', generate_labeled_data())
def test_sagittal_slice_get_center_spit(im_seg):
```

But, this creates a weird interaction with `xdist`/`download_data`, when more than 1 worker is used. Sometimes it works, but sometimes redownloading the data causes a FileNotFound error, because the data folder will be wiped for each worker. (See attached [log.txt](https://github.com/neuropoly/spinalcordtoolbox/files/5743237/xdist_download-data_filenotfound-error_log.txt).) This is related to #2957 and #2959.

I can fix this by setting `-n 1` as recommended in https://github.com/neuropoly/spinalcordtoolbox/issues/2957#issuecomment-729699201, but I'm not sure what the behavior will be like in the CI.

joshuacwnewton: I've implemented this in https://github.com/neuropoly/spinalcordtoolbox/pull/3116/commits/78382101598696cb9033ba2eefb2180eec186942. It's probably a bit overkill for such a simple bug in a minor function, though. :sweat_smile:

jcohenadad:
While testing this PR i noticed another bug https://github.com/neuropoly/spinalcordtoolbox/issues/3130

jcohenadad (approved):
Working fine now 👌 
Thank you @joshuacwnewton !!!

PR #3120, opened 2020-12-22
Validate '-vertfile' dimensions in aggregate_per_slice_or_level()

### LINKED ISSUE #3111: Need to specify if Vertebral labeling file (-vertfile) needs to be in the same voxel coordinate system as the input segmentation (-i)
Currently the usage says:
~~~
  -vertfile <str>       Vertebral labeling file. Only use with flag -vert (default: ./label/template/PAM50_levels.nii.gz)
~~~

We need to check (look into the code) if the input file is transformed into the same voxel coordinate system as the input image, or if it is not.

And depending on the answer, we need to specify in the usage what is expected from this input vertfile.

yanchunCoBIUS:
### EDIT (2020-01-29)

<details>
<summary> This PR was taken over by @joshuacwnewton. Original issue description hidden, click to show. </summary>

Do the following fix:

- In sct_process_segmentation.py: First, add the dimensions check between the input and the vertebral labelling file. Second, update the help.
- In sct_extract_metric.py: First, add the dimensions check between the input image and the vertebral labelling file. Second, update the help.

</details>

#### GitHub

#### PR contents

## Description

This PR:

* Adds a check to `aggregate_slicewise.py : aggregate_per_slice_or_level()` to verify that the `-vertfile` has the same number of slices as the calculated metric. 
* Sets the value of `-vertfile` only when the `-vert` argument is also passed. 
    * This way, the default value for `-vertfile` doesn't get passed along if `-vert` isn't explicitly chosen by the user. Otherwise, the above check will be erroneously triggered. 
    * This was only needed in `sct_process_segmentation`.
* Updates the help description for `-vertfile` argument in both `sct_extract_metric` and `sct_process_segmentation` to warn about this requirement.
* Updates some tests that were silently failing this requirement, because they were using dummy data with mismatched dimensions.
* Adds a new test to verify that a ValueError is called with mismatched inputs.

## Linked issues

Partially addresses #3111.


spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -198,7 +198,7 @@ def get_parser():
         '-vertfile',
         metavar=Metavar.file,
         default="./label/template/PAM50_levels.nii.gz",
-        help="Vertebral labeling file. Only use with flag -vert. Example: PAM50_levels.nii.gz"
+        help="Vertebral labeling file. Only use with flag -vert. The input Image and the vertebral labelling file must in the same voxel coordinate system and must match the dimensions between each other. Example: PAM50_levels.nii.gz"
```
jcohenadad, line 201:
```suggestion
        help="Vertebral labeling file. Only use with flag -vert. The input Image and the vertebral labelling file must in the same voxel coordinate system and must match the dimensions between each other. Example: label/template/PAM50_levels.nii.gz"
```

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -358,6 +358,14 @@ def main(fname_data, path_label, method, slices, levels, fname_output, labels_us
...
     if vertebral_levels:
         im_vertebral_labeling = Image(fname_vertebral_labeling).change_orientation("RPI")
+
+        # Get dimensions of input Image and the vertebral labeling file
+        nx, ny, nz = input_im.data.shape
+        nx_vertebral, ny_vertebral, nz_vertebral = im_vertebral_labeling.data.shape
+
+        # Check dimensions consistency between the input Image and the vertebral labeling file
+        if (nx, ny, nz) != (nx_vertebral, ny_vertebral, nz_vertebral):
+            printv('\nERROR: Metric data and vertebral labeling file DO NOT HAVE SAME DIMENSIONS.', 1, type='error')
```
jcohenadad, line 368: shouldn't we be doing this check inside the [API](aggregate_per_slice_or_level) instead? To limit duplication of code in the CLI (here: `sct_extract_metric` and `sct_process_segmentation`). What do you think @joshuacwnewton ?

joshuacwnewton:
```python
# Rewritten to make it more compact
if input_im.data.shape != im_vertebral_labeling.data.shape:
    printv('\nERROR: Metric data and vertebral labeling file DO NOT HAVE SAME DIMENSIONS.', 1, type='error')
```

Because this is such a simple and self-explanatory check, I think duplication is OK. If we did make an API function (to prevent duplication), I think it might just add clutter to the API, becuase I think users/developers would just write out the check (rather than use the function). (In general, though, for more complex logic, I absolutely agree that avoiding duplication is a helpful mindset to have.) 

Besides that, though, I do agree that we should be doing the check inside the API, but for a different reason. :stuck_out_tongue:

Say a future user calls the relevant API functions (`extract_metric()` and `aggregate_per_slice_or_level()`) in one of their own scripts. If we only do the check in the CLI, then the problem from #3111 would still exist at the API level, I think.

(This is something that `scikit-learn` did quite rigorously when I contributed there. You can see in an example such as [`TSNE`](https://github.com/scikit-learn/scikit-learn/blob/6b4f82433dc2f219dbff7fe8fa42c10b72379be6/sklearn/manifold/_t_sne.py#L681-L735) that they devote a lot of space for validation of function parameters, and I like this.)

yanchunCoBIUS: Thanks for the suggestions from both of you, also sorry about the mistake, I will do the check inside the API.

joshuacwnewton: No need to apologize! :slightly_smiling_face:  That's what reviews are for, so we can share different perspectives as a team. :smile:

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -357,7 +357,7 @@ def main(fname_data, path_label, method, slices, levels, fname_output, labels_us
     labels = np.concatenate(labels_tmp[:], 3)  # labels: (x,y,z,label)
     # Load vertebral levels
     if vertebral_levels:
-        im_vertebral_labeling = Image(fname_vertebral_labeling).change_orientation("RPI")
+        im_vertebral_labeling = fname_vertebral_labeling        
     else:
         im_vertebral_labeling = None
```
jcohenadad, line 362: maybe we can simply get rid of that, and pass `fname_vertebral_labeling` into the function `extract_metric()` below. We would need to make sure the argument is "None" if not explicitly set by the user in the CLI parser. 

also, just for your information, we have [conventions for variable names](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Variables#variables), and "im_" corresponds to Image() objects (so in this case, the proposed change would go against the convention).

spinalcordtoolbox/utils/validation.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+# Validation related utilites
+
+
+def check_dimensions_match(input_im, im_vertebral_labeling):
```
jcohenadad: @joshuacwnewton @kousu what do you think about this new utils/validation file?

spinalcordtoolbox/utils/validation.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+# -*- coding: utf-8
+# Validation related utilites
+
+
+def check_dimensions_match(input_metric, im_vertebral_labeling):
```
joshuacwnewton, line 6: Could you also add a test for this function? Say, inside a new `unit_testing/test_validation.py` module? It would be nice to verify that:

* If called with two images that have different dimensions, an error is thrown
* If called with two images that have the same dimensions, no error is thrown

For test images, you can use data from the `sct_testing_data/`. (But, I also think this is a little unclear, so I've opened #3136 so our team can discuss.)

joshuacwnewton (changes requested):

joshuacwnewton:
Also, I think `check_dimensions_match` needs to be added inside the `aggregate_slicewise.py : extract_metric()` function too, since that's the API function called by `sct_extract_metric`:

https://github.com/neuropoly/spinalcordtoolbox/blob/d7ca252603ba44f6942c463b3ae16fdd326424d9/spinalcordtoolbox/scripts/sct_extract_metric.py#L385-L389

spinalcordtoolbox/scripts/sct_extract_metric.py
yanchunCoBIUS: Resolved by e9ddb86b718d7fd5ef5fb04078358539e5e9b930

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -115,7 +115,7 @@ def get_parser():
         '-vertfile',
         metavar=Metavar.str,
         default='./label/template/PAM50_levels.nii.gz',
-        help="Vertebral labeling file. Only use with flag -vert"
+        help="Vertebral labeling file. Only use with flag -vert. The input and the vertebral labelling file must in the same voxel coordinate system and must match the dimensions between each other. "
```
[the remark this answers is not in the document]
yanchunCoBIUS: Resolved by d39e11e5064b0dcb4eaaa3cd0a584e2c08ea9179.

spinalcordtoolbox/utils/validation.py
```diff
@@ -0,0 +1,11 @@
...
+    # Check dimensions consistency between the input and the vertebral labeling file
+    if (input_metric.data.shape) != (im_vertebral_labeling.data.shape):
+        return False
+    else:
+        return True
```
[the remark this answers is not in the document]
yanchunCoBIUS: Thanks for your comments, resolved by bfc109083aec1536dbbac4c7c77cb752bd10ee3f.

yanchunCoBIUS:
> Also, I think `check_dimensions_match` needs to be added inside the `aggregate_slicewise.py : extract_metric()` function too

The reason why I do the `check_dimensions_match` in `aggregate_per_slice_or_level()` is that, in 'extract_metric()' the only place use both 'metric' and 'image'. 
Please see: 
https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/aggregate_slicewise.py#L426-L428

spinalcordtoolbox/utils/validation.py
yanchunCoBIUS: That's a good point, also the test data is unclear to me. After the discussion in #3136 , I will add a new unit test.

joshuacwnewton:
> The reason why I do the `check_dimensions_match` in `aggregate_per_slice_or_level()` is that, in 'extract_metric()' the only place use both 'metric' and 'image' is in `aggregate_per_slice_or_level`.
> Please see:
> 
> https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/aggregate_slicewise.py#L426-L428

Thanks for the clarification! That makes sense. :slightly_smiling_face:

spinalcordtoolbox/utils/validation.py
kousu: Oh uh I have no idea. let me look.

spinalcordtoolbox/utils/validation.py
```diff
@@ -0,0 +1,14 @@
...
+
+def check_dimensions_match(input_im, im_vertebral_labeling):
+    nx, ny, nz = input_im.data.shape
+    nx_vertebral, ny_vertebral, nz_vertebral = im_vertebral_labeling.data.shape
+
+    # Check dimensions consistency between the input and the vertebral labeling file
+    if (nx, ny, nz) != (nx_vertebral, ny_vertebral, nz_vertebral):
+        return False
+    else:
+        return True
```
kousu: This could be

```suggestion
def check_dimensions_match(input_im, im_vertebral_labeling):
    return input_im.data.shape == im_vertebral_labeling.data.shape
```

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -359,12 +360,7 @@ def main(fname_data, path_label, method, slices, levels, fname_output, labels_us
...
         im_vertebral_labeling = Image(fname_vertebral_labeling).change_orientation("RPI")
 
-        # Get dimensions of input Image and the vertebral labeling file
-        nx, ny, nz = input_im.data.shape
-        nx_vertebral, ny_vertebral, nz_vertebral = im_vertebral_labeling.data.shape
-
-        # Check dimensions consistency between the input Image and the vertebral labeling file
-        if (nx, ny, nz) != (nx_vertebral, ny_vertebral, nz_vertebral):
+        if check_dimensions_match(input_im=input_im, im_vertebral_labeling=im_vertebral_labeling):
             printv('\nERROR: Metric data and vertebral labeling file DO NOT HAVE SAME DIMENSIONS.', 1, type='error')
```
kousu: Is there a reason this isn't an exception?

```suggestion
        if input_im.shape != im_vertebral_labeling.shape:
            raise ValueError('\nERROR: Metric data and vertebral labeling file DO NOT HAVE SAME DIMENSIONS.')
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -294,12 +295,7 @@ def main(args=None):
...
         fname_vert_levels = arguments.vertfile
         if arguments.vert is not None:
-            # Get dimensions of input and the vertebral labeling file
-            nx, ny, nz = Image(fname_segmentation).change_orientation('RPI').data.shape
-            nx_vertebral, ny_vertebral, nz_vertebral =  Image(fname_vert_levels).change_orientation('RPI').data.shape
-
-            # Check dimensions consistency between the input and the vertebral labeling file
-            if (nx, ny, nz) != (nx_vertebral, ny_vertebral, nz_vertebral):
+            if check_dimensions_match(input_im=Image(fname_segmentation).change_orientation('RPI'), im_vertebral_labeling=Image(fname_vert_levels).change_orientation('RPI')):
                 print('\nERROR: Input data and vertebral labeling file DO NOT HAVE SAME DIMENSIONS.', 1, type='error')
```
kousu: ditto

spinalcordtoolbox/utils/validation.py
kousu: I think you could do without this since so far its contents can be simplified to a one-liner. Are there other kinds of validation we have in mind?

spinalcordtoolbox/utils/validation.py
yanchunCoBIUS: The file that you are referring to is outdated, please see the latest changes: https://github.com/neuropoly/spinalcordtoolbox/pull/3120/files#diff-745abc1c2e85edaa536f7b2eb0f8a7c18c4cc8e7db38e0e7fe5e74bee517fbd2.

spinalcordtoolbox/scripts/sct_extract_metric.py
yanchunCoBIUS: The file that you are referring to is outdated, I do the `check_dimensions_match` in `aggregate_per_slice_or_level()`. Also, catch the `ValueError` here: https://github.com/neuropoly/spinalcordtoolbox/blob/bfc109083aec1536dbbac4c7c77cb752bd10ee3f/spinalcordtoolbox/aggregate_slicewise.py#L252-L255

spinalcordtoolbox/scripts/sct_process_segmentation.py
yanchunCoBIUS: I do the `check_dimensions_match` in `aggregate_per_slice_or_level()`.
Please see the comment: https://github.com/neuropoly/spinalcordtoolbox/pull/3120#issuecomment-751926217

spinalcordtoolbox/utils/validation.py
yanchunCoBIUS: > I think you could do without this since so far its contents can be simplified to a one-liner. 

I agree with that.

> Are there other kinds of validation we have in mind?

Currently, I don't have other validations in mind, but if we have some in the future, we can add more validations functions here, which could be easier for the management of all validations. Please see the examples from https://github.com/neuropoly/spinalcordtoolbox/pull/3120#discussion_r549174620

spinalcordtoolbox/utils/validation.py
joshuacwnewton: Hi @yanchunCoBIUS. I have written some recommendations in the [Test Data](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Test-data) wiki page. 

Could you take a look and tell me if they're clear enough? If anything is confusing or unclear, feel free to ask, and I will be happy to update things to make more sense. :slightly_smiling_face:

joshuacwnewton:
Introducing the new dimensions check caused 7 existing tests to fail:

* 6 tests in `test_aggregate_slicewise.py`
    * These failures come from dummy data that did not pass the dimension check. This should be an easy fix, I believe -- just edit the dummy data so that the dimensions match (as they should).
* 1 test in `test_cli_process_segmentation`
    * This comes from the fact that the default value for `-vertfile` doesn't exist by default, but we use it anyway. We should only be using it if `-vert` is explicitly set, too.

joshuacwnewton:
@kousu I've also followed your advice to not add `utils/validation.py`. I was getting too fancy with my suggestion, since it's basically a one-liner, as you point out. :sweat_smile:

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -232,6 +232,13 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
     :param map_clusters: list of list of int: See func_map()
     :return: Aggregated metric
     """
+    if vert_level:
+        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[2]
+        metric_slices = metric.data.shape[0]
+        if vert_level_slices != metric_slices:
```
jcohenadad, line 238: i'll approve it (we have other priority stuff to deal with), but the test could be made more sensitive. e.g., if a user inputs a vertebral level file from another region/subject/whatever that happens to have the same number of slices as the metric, processing will go on but will be wrong. Alternatively, the qform and XYZ dimensions of the image header could be compared (ie: to make sure both volumes are in the exact same coordinate system).

jcohenadad (approved):
thank you for the fix!

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Good points. I'll unlink the issue so that it stays open, then add your comments here to the issue description.

joshuacwnewton (approved):
Since this is not my PR, and I had requested changes in a previous review, I have to approve to be able to merge.

PR #3123, opened 2020-12-23
Fixup realpath polyfill

kousu:
Adding in #3103, but https://github.com/neuropoly/spinalcordtoolbox/runs/1601504567?check_suite_focus=true#step:4:1407 shows it failing on Github Actions with

> ```
>   File "<string>", line 1
>    import sys, path; for f in sys.argv[1:]: print(os.path.realpath(f))
>                        ^
> SyntaxError: invalid syntax
> ```

I guess I didn't actually test this.

Strange that it got past CI?? Maybe Travis's macOS images have GNU realpath installed but Actions' don't??

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -301,7 +301,7 @@ if [ "$(uname)" = "Darwin" ]; then
   # So here, solve it with python. It's not great either, but since much of
   # this script already is just calling python at least it's reliable.
   (command -v realpath >/dev/null) || realpath() {
-    python -c 'import sys, path; for f in sys.argv[1:]: print(os.path.realpath(f))' "$@"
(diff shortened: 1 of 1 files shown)


joshuacwnewton (approved):
Sorry about this. I didn't think to test the original command either! 

Testing this using:

```
(command -v realpath >/dev/null) || realpath() {
  python3 -c 'import sys, os; [print(os.path.realpath(f)) for f in sys.argv[1:]]' "$@"
}
  
ACTUAL_PYTHON="$(realpath "$(command -v python)")"
echo "$ACTUAL_PYTHON"
```

Yields:

```
joshua@XPS-15-9560:~$ ./test_script.sh 
/home/joshua/.pyenv/shims/python
```

PR #3139, opened 2020-12-29
Replace "argv if argv else '--help'" behavior with subclassed ArgumentParser

### LINKED ISSUE #3137: Auto-help in sct scripts should return an error
It looks like after #3091 every script has the glitch that being run without arguments returns 0 ("no error"), because they all have this line in them:

```
arguments = parser.parse_args(argv if argv else ['--help'])
```

For some scripts it's truly not an error to be called with no args (like `sct_check_requirements`), but for others it is.

We need this behaviour for correct shell scripting with them:

```
$ ./script -h; echo $?
usage: script [-h]

optional arguments:
  -h, --help  show this help message and exit
0
$ ./script; echo $?
usage: script [-h]

optional arguments:
  -h, --help  show this help message and exit
2
```

_Originally posted by @kousu in https://github.com/neuropoly/spinalcordtoolbox/pull/3091#issuecomment-751946174_

joshuacwnewton:
#### GitHub

#### PR contents

#### Issue

#3091 included the line `arguments = parser.parse_args(argv if argv else ['--help'])`, which shows the help message if no arguments are passed. The problem is, passing `--help` is valid usage, and the error code returned is 0. This is a bug, because passing no arguments should be invalid usage for many scripts!

#### Work done

This PR introduces a subclassed ArgumentParser that overrides the `error` method to print the help instead. This way, a non-zero error code is returned AND the help is shown.

The same changes are made to every script (switching from `ArgumentParser` to `SCTArgumentParser`, removing `argv if argv else --help`) except these two:

* `sct_testing`
* `sct_qc`

These scripts also got an extra `--help` argument, because they never had one in the first place.

#### Aside about centralization

An added benefit of having a subclassed parser is that we get to centralize the default ArgumentParser values for all scripts, which limits duplicated code. 

In theory, we could also centralize the `-h`, `-v`, and `-r` arguments, too. But, that's a more involved effort, and can be done later as part of #2676.

## Linked issues

Fixes #3137.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_testing.py
+++ b/spinalcordtoolbox/scripts/sct_testing.py
@@ -23,7 +23,7 @@ import signal
 import numpy as np
 from pandas import DataFrame
 
-from spinalcordtoolbox.utils import init_sct, run_proc, tmp_create, printv, rmtree, __sct_dir__, set_global_loglevel
+from spinalcordtoolbox.utils import init_sct, run_proc, tmp_create, printv, rmtree, __sct_dir__, set_global_loglevel, SCTArgumentParser
 from spinalcordtoolbox.scripts import sct_download_data
 
 # FIXME
@@ -99,11 +99,9 @@ class bcolors:
 # PARSER
 # ==========================================================================================
 def get_parser():
-    import argparse
-
     param_default = Param()
 
-    parser = argparse.ArgumentParser(
+    parser = SCTArgumentParser(
         description="Crash and integrity testing for functions of the Spinal Cord Toolbox. Internet connection is required for downloading testing data.",
     )
 
@@ -158,6 +156,9 @@ def get_parser():
     parser.add_argument("--execution-folder",
                         help="Folder where to run tests from (default. temporary)",
                         )
+    parser.add_argument('-h', "--help",
+                        help="show this message and exit",
+                        action="help")
 
     return parser
 
@@ -210,7 +211,7 @@ def process_function_multiproc(fname, param):
 # ==========================================================================================
 def main(argv=None):
     parser = get_parser()
-    arguments = parser.parse_args(argv if argv else ['--help'])
+    arguments = parser.parse_args(argv)
     verbose = arguments.verbose
     set_global_loglevel(verbose=verbose)
 
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -100,7 +100,12 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
 
 
 class SCTArgumentParser(argparse.ArgumentParser):
-    """Parser that centralizes initialization steps common across all SCT scripts."""
+    """
+        Parser that centralizes initialization steps common across all SCT scripts.
+
+        TODO: Centralize `-v`, `-r`, and `-h` arguments here too, as they're copied
+              and pasted across all SCT scripts.
+    """
     def __init__(self, *args, **kwargs):
         def update_parent_default(key, value):
             """A polite way of letting a child class have different default values than the parent class."""
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -8,6 +8,7 @@ import re
 import shutil
 import logging
 import argparse
+import inspect
 
 from enum import Enum
 
@@ -96,6 +97,38 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
     if verbose:
         printv('\nDone! To view results, type:')
         printv(cmd + '\n', verbose=1, type='info')
+
+
+class SCTArgumentParser(argparse.ArgumentParser):
+    """Parser that centralizes initialization steps common across all SCT scripts."""
+    def __init__(self, *args, **kwargs):
+        def update_parent_default(key, value):
+            """A polite way of letting a child class have different default values than the parent class."""
+            # Source: https://stackoverflow.com/a/41623488
+            argspec = inspect.getfullargspec(super(SCTArgumentParser, self).__init__)
+            arg_index = argspec.args.index(key)
+            if len(args) < arg_index and key not in kwargs:
+                kwargs[key] = value
+
+        update_parent_default('formatter_class', SmartFormatter)
+
+        # Update "usage:" message to match how SCT scripts are actually called (no '.py')
+        frame = inspect.stack()[1]
+        module = inspect.getmodule(frame[0])
+        update_parent_default('prog', os.path.basename(module.__file__).strip(".py"))
+
+        # Disable "add_help", because it won't properly add '-h' to our custom argument groups
+        # (We use custom argument groups because of https://stackoverflow.com/a/24181138)
+        update_parent_default('add_help', False)
+
+        super(SCTArgumentParser, self).__init__(*args, **kwargs)
+
+    def error(self, message):
+        """Overridden parent method. Ensures that calling script with no args prints the help. See issue #3137."""
+        # Source: https://stackoverflow.com/a/4042861
+        sys.stderr.write(f'{self.prog}: error: {message}\n\n')
+        self.print_help(sys.stderr)
+        sys.exit(2)
 
 
 class ActionCreateFolder(argparse.Action):
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -124,7 +124,11 @@ class SCTArgumentParser(argparse.ArgumentParser):
         super(SCTArgumentParser, self).__init__(*args, **kwargs)
 
     def error(self, message):
-        """Overridden parent method. Ensures that calling script with no args prints the help. See issue #3137."""
+        """
+            Overridden parent method. Ensures that calling script with no args prints the help.
+
+            See https://github.com/neuropoly/spinalcordtoolbox/issues/3137.
+        """
         # Source: https://stackoverflow.com/a/4042861
         sys.stderr.write(f'{self.prog}: error: {message}\n\n')
         self.print_help(sys.stderr)
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -130,7 +130,7 @@ class SCTArgumentParser(argparse.ArgumentParser):
 
     def error(self, message):
         """
-            Overridden parent method. Ensures that calling script with no args prints the help.
+            Overridden parent method. Ensures that help is printed when called with invalid args.
 
             See https://github.com/neuropoly/spinalcordtoolbox/issues/3137.
         """
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -13,7 +13,6 @@ import math
 import sys
 import pickle
 import shutil
-import argparse
 
 import numpy as np
 import pandas as pd
@@ -21,15 +20,13 @@ from skimage.measure import label
 
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.centerline.core import ParamCenterline, get_centerline
-from spinalcordtoolbox.utils.shell import Metavar, SmartFormatter, ActionCreateFolder
+from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, ActionCreateFolder
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_global_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, extract_fname, copy, rmtree
 
 
 def get_parser():
-    # Initialize the parser
-
-    parser = argparse.ArgumentParser(
+    parser = SCTArgumentParser(
         description='R|Compute statistics on segmented lesions. The function assigns an ID value to each lesion (1, 2, '
                     '3, etc.) and then outputs morphometric measures for each lesion:\n'
                     '- volume [mm^3]\n'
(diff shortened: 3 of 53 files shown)


spinalcordtoolbox/scripts/sct_testing.py
```diff
@@ -158,6 +156,9 @@ def arg_jobs(s):
     parser.add_argument("--execution-folder",
                         help="Folder where to run tests from (default. temporary)",
                         )
+    parser.add_argument('-h', "--help",
+                        help="show this message and exit",
+                        action="help")
```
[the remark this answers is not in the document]
kousu: Oh I see. It's these two that were the outliers; everything else has 

```
    optional.add_argument(
        "-h",
        "--help",
        action="help",
        help="Show this help message and exit."
    )
```

in it. Maybe we can do better?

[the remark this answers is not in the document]
kousu: **See below.**

spinalcordtoolbox/utils/shell.py
```diff
@@ -98,6 +99,38 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
...
+        frame = inspect.stack()[1]
+        module = inspect.getmodule(frame[0])
+        update_parent_default('prog', os.path.basename(module.__file__).strip(".py"))
+
+        # Disable "add_help", because it won't properly add '-h' to our custom argument groups
+        # (We use custom argument groups because of https://stackoverflow.com/a/24181138)
+        update_parent_default('add_help', False)
+
+        super(SCTArgumentParser, self).__init__(*args, **kwargs)
+
```
kousu, line 125: It's inconsistency-prone to set `add_help=False` here but then expect every single script to add `--help` itself. I know the argparse API isn't great here -- there's an `add_` but no `get_` and even though there's titles to key on they're not in a dict. Upstream is

https://github.com/python/cpython/blob/056c08211b402b4dbc1530a9de9d00ad5309909f/Lib/argparse.py#L1436-L1439

```
    def add_argument_group(self, *args, **kwargs):
        group = _ArgumentGroup(self, *args, **kwargs)
        self._action_groups.append(group)
        return group
```

What if we turned the insert into an up-sert instead:

```suggestion

    def add_argument_group(self, title, *args, **kwargs):
         g = [g for g in parser._action_groups if g.title == title]
         if g:
             return g
         else:
             return super().add_argument_group(title, *args, **kwargs)
             
```

And added this to `__init__`:

```
    optional = self.add_argument_group("\nOPTIONAL ARGUMENTS")
    optional.add_argument(
        "-h",
        "--help",
        action="help",
        help="Show this help message and exit."
    )
```

then we could get rid of all the duplicate `--help`s, and plus the `--help` addition be on the same level as the `add_help=False` that makes room for it and nothing else has to change.

This is mildly buggy in that if a class *did* want to customize the --help group the cached version won't pick them up. I don't know how to handle that. Maybe add

```
if args or kwargs: raise ValueError
```
?

kousu (changes requested):
Your branch successfully fixes the ci glitch!

```
$ ./install_sct
[...]
Validate installation...


--
Spinal Cord Toolbox (git-jn/3137-add-subclassed-parser-10c0e3d1087cf479118ce15c02e7c6c37fc20982)

sct_check_dependencies 
--

SCT info:
- version: git-jn/3137-add-subclassed-parser-10c0e3d1087cf479118ce15c02e7c6c37fc20982
- path: /home/kousu/src/neuropoly/spinalcordtoolbox
OS: linux (Linux-5.9.14-arch1-1-x86_64-with-arch)
CPU cores: Available: 4, Used by ITK functions: 4
RAM: Total: 15710MB, Used: 2255MB, Available: 12601MB
  Using bundled python 3.6.12 |Anaconda, Inc.| (default, Sep  8 2020, 23:10:56) 
[GCC 7.3.0] at /home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin/python
Check if dipy is installed..........................[OK] (1.3.0)
Check if ivadomed (2.6.1) is installed..............[OK] (2.6.1)
Check if matplotlib is installed....................[OK] (3.3.3)
Check if numpy is installed.........................[OK] (1.19.4)
Check if pandas is installed........................[OK] (1.1.5)
Check if psutil is installed........................[OK] (5.8.0)
Check if pytest is installed........................[OK] (6.2.1)
Check if pytest-cov is installed....................[OK] (__version__ = '2.10.1')
Check if requests is installed......................[OK] (2.25.1)
Check if scikit-learn is installed..................[OK] (0.24.0)
Check if tqdm is installed..........................[OK] (4.55.0)
Check if urllib3 is installed.......................[OK] (1.26.2)
Check if pytest-xdist is installed..................[OK] (2.2.0)
Check if tensorboard (1.14.0) is installed..........[WARNING] (2.4.0 != 1.14.0 mandated version))


To use SCT, please update your environment by running:
export PATH="/home/kousu/src/neuropoly/spinalcordtoolbox/bin:$PATH"


```

The right return codes are coming out too:

```
$ ./bin/sct_check_dependencies --help; echo $?

--
Spinal Cord Toolbox (git-jn/3137-add-subclassed-parser-10c0e3d1087cf479118ce15c02e7c6c37fc20982)

sct_check_dependencies --help
--

usage: sct_check_dependencies [-h] [-complete] [-short]

Check the installation and environment variables of the toolbox and its dependencies.

OPTIONAL ARGUMENTS:
  -h, --help  Show this help message and exit
  -complete   Complete test. (default: False)
  -short      Short test. Only shows SCT version, CPU cores and RAM available. (default: False)
0
$ ./bin/sct_check_dependencies -j; echo $?

--
Spinal Cord Toolbox (git-jn/3137-add-subclassed-parser-10c0e3d1087cf479118ce15c02e7c6c37fc20982)

sct_check_dependencies -j
--

sct_check_dependencies: error: unrecognized arguments: -j

usage: sct_check_dependencies [-h] [-complete] [-short]

Check the installation and environment variables of the toolbox and its dependencies.

OPTIONAL ARGUMENTS:
  -h, --help  Show this help message and exit
  -complete   Complete test. (default: False)
  -short      Short test. Only shows SCT version, CPU cores and RAM available. (default: False)
2
```

I haven't checked every single script for the right return codes. But my smoke test has passed so I'm happy. If I've missed something we can fix it later.

---

You got through this really fast. It all looks correct to me! Take a look at my suggestions and yay/nay them and then it can go in.

spinalcordtoolbox/utils/shell.py
kousu: Oh I just noticed you addressed this:

> In theory, we could also centralize the -h, -v, and -r arguments, too. But, that's a more involved effort, and can be done later as part of #2676.

I think that's a good idea too.

spinalcordtoolbox/utils/shell.py
```diff
@@ -98,6 +99,38 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
         printv(cmd + '\n', verbose=1, type='info')
 
 
+class SCTArgumentParser(argparse.ArgumentParser):
```
kousu, line 102: Today in stupid OOP tricks for 100 points: You might consider splitting this into parts:

```
class AutoHelpParser(argparse.ArgumentParser):
   def error(): ...
...

class SCTArgumentParser(AutoHelpParser, argparse.ArgumentParser):
   def __init__(...):
       self.add_argument_group("OPTIONAL ARGUMENTS") ......
...
```

That would make using it for #2676 clearer because you could add:


```
class SCTArgumentParser(VerbosityParser, AutoHelpParser, argparse.ArgumentParser):
   def __init__(...):
       self.add_argument_group("OPTIONAL ARGUMENTS") ......
...
```


But you could also split it when you get to #2676. Or just not? It's easy to go overboard with OOP.

spinalcordtoolbox/utils/shell.py
```diff
@@ -98,6 +99,38 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
...
+        # Update "usage:" message to match how SCT scripts are actually called (no '.py')
+        frame = inspect.stack()[1]
+        module = inspect.getmodule(frame[0])
+        update_parent_default('prog', os.path.basename(module.__file__).strip(".py"))
+
+        # Disable "add_help", because it won't properly add '-h' to our custom argument groups
+        # (We use custom argument groups because of https://stackoverflow.com/a/24181138)
+        update_parent_default('add_help', False)
+
+        super(SCTArgumentParser, self).__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
joshuacwnewton: > Or maybe we should just say "our custom formatter" and assume people will go read its source code if they need to figure out what it can do?

Exactly. `SmartFormatter` and `SCTArgumentParser` are in the same `utils/shell.py` file, so all someone has to do is scroll down a little.

Perhaps `utils/shell.py` could be reordered so that `SCTArgumentParser` is right next to `SmartFormatter`? The ordering is a bit haphazard right now.

spinalcordtoolbox/utils/shell.py
joshuacwnewton: Now that you mention it, `-h` is a lot easier to centralize compared to `-v` and `-r`. 

The reason I put it off was for the complexity of `-v` and `-r`, but I can definitely do `-h`. (Which also addresses https://github.com/neuropoly/spinalcordtoolbox/pull/3139#discussion_r549888651.)

joshuacwnewton: Thanks for pointing this out. :)

spinalcordtoolbox/utils/shell.py
joshuacwnewton: Interesting! Admittedly, I haven't explored multiple inheritance very deeply, so I like that you brought it up. Gets me to think about new things. :D

My only reservation is that it makes things a little more complicated to understand. And, that added complexity might not be worth it, because `AutoHelpParser` and `VerbosityParser` would only be used within `SCTArgumentParser` (since we only need one argparser across all of our scripts.).

spinalcordtoolbox/utils/shell.py
```diff
@@ -98,6 +99,38 @@ def display_viewer_syntax(files, colormaps=[], minmax=[], opacities=[], mode='',
...
+
+    def error(self, message):
+        """Overridden parent method. Ensures that calling script with no args prints the help. See issue #3137."""
```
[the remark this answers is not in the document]
joshuacwnewton: Ahh, good point. I've been doing this in a lot of places, because it's helpful for keeping comments one-liners, but one-liners don't _reeeeally_ matter in this context. So, this is probably a habit I should update. :stuck_out_tongue_closed_eyes:

spinalcordtoolbox/utils/shell.py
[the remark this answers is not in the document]
joshuacwnewton: > Instead of that stackoverflow thing I think you can use the (new-ish) dict-splat syntax: https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python-taking-union-o#26853961

The StackOverflow pattern checks whether the argument was passed not just via `kwargs` but also `args` (which is the trickier one). 

With the dict-splat method, if someone passed an argument via `args`, I think our default will still get added to `kwargs` (i.e. that argument it will be passed twice with possibly conflicting values). I'm honestly not sure what the result is when that happens... hrm.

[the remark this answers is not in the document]
joshuacwnewton: Yeah, I get an error when I try the suggestion. :cry:  This part of the test:

```python
    # Check that new defaults can still be overridden (setting add_help via args AND kwargs)
    parser1 = utils.SCTArgumentParser(None, None, None, None, [], utils.SmartFormatter, '-', None, None, 'error', True)
    assert parser1.add_help is True
```

Results in this error:

```python
>       super(SCTArgumentParser, self).__init__(*args, **{**defaults, **kwargs})
E       TypeError: __init__() got multiple values for argument 'formatter_class'
```

Since it gets passed in both `args` and `defaults`.

spinalcordtoolbox/utils/shell.py
joshuacwnewton: > What if we turned the insert into an up-sert instead:
> 
> ```python
>     def add_argument_group(self, title, *args, **kwargs):
>          g = [g for g in parser._action_groups if g.title == title]
>          if g:
>              return g
>          else:
>              return super().add_argument_group(title, *args, **kwargs)             
> ```

The tricky part about this pattern is how sensitive it is to the specific `\nOPTIONAL ARGUMENTS` string. For example, I've seen some scripts use e.g. `\nMISC` instead, and in that case you would end up with 2 categories.

At the same time, this complexity is sort of inherent to our choice of using custom argument groups in the first place. :sweat_smile: 

TBH, since we're already messing with built-in argparse behavior, I wonder if we could just... _fix_ the whole "need for custom groups" quirk that `argparse` has.

joshuacwnewton: Hmm... these custom argument groups seem like they may a headache, and it seems tricky to quickly include the suggestion in this PR. I'm leaning toward what you said here:

> but it's okay for it to be in flux and a bit inconsistent temporarily, so long as it doesn't go long enough that we forget why it was like that. I'll link this thread over there so my investigation can help when we get to that.

I've added a TODO in 4cb11b6 to help not forget this. :sweat_smile:

spinalcordtoolbox/utils/shell.py
kousu: Totally, always glad to share roadshow programming tricks.

spinalcordtoolbox/utils/shell.py
kousu: Great, je suis d'accord.

kousu (approved):
Great, done! :shipit:

spinalcordtoolbox/utils/shell.py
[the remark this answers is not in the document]
kousu: I didn't think about that at all. Thanks for trying it out. Now I've learned something.

This seems like it should be solvable with shorter syntax but :shrug: it's not a big deal. And you documented it well so it won't be a big mystery in the future.

PR #3146, opened 2021-01-02
Replace os.rename with shutil.move to fix tmp_path issues on CI

### LINKED ISSUE #3147: `mv` function in `utils.py` fails when using `tmp_path` on CI
Error reported in https://github.com/neuropoly/spinalcordtoolbox/pull/3081#issuecomment-752034463. See: https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495#L2100-L2105

```
OSError: [Errno 18] Invalid cross-device link: './t2_rescaled_seg.nii.gz' -> '/tmp/pytest-of-travis/pytest-0/popen-gw0/test_sct_propseg_o_flag0/test_seg.nii.gz'
```

Cause: Our use of `os.rename`, which has issues [when the src and dst files are on different filesystems](https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link/43967659#43967659). 

https://github.com/neuropoly/spinalcordtoolbox/blob/d64af23b69dbc396322cdc963a90a899b73736e0/spinalcordtoolbox/utils/fs.py#L225-L229

The fix is pretty simple: replace `os.rename` with `shutil.move`.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

`pytest.tmp_path` is useful for working in temporary directories (automatic cleanup), but combining `tmp_path` with our use of `os.rename` caused an error in https://github.com/neuropoly/spinalcordtoolbox/pull/3081#issuecomment-752034463. This PR replaces `os.rename` with `shutil.move` to fix the problem. (https://stackoverflow.com/a/43967659)

The solution was first tested by basing this PR off of #3081, starting from the state in #3081 that produced the error, see: https://github.com/neuropoly/spinalcordtoolbox/commit/3de1099252238e2f36aebee5be9e78e46c6eb079. That run passed successfully: https://github.com/neuropoly/spinalcordtoolbox/runs/1637789000

I've just rebased so that this PR contains only the relevant change.

## Linked issues

Fixes #3147.
Before #3081.

### DIFF
--- a/spinalcordtoolbox/utils/fs.py
+++ b/spinalcordtoolbox/utils/fs.py
@@ -226,7 +226,7 @@ def mv(src, dst, verbose=1):
     """Move a file from src to dst, almost like os.rename
     """
     printv("mv %s %s" % (src, dst), verbose=verbose, type="code")
-    os.rename(src, dst)
+    shutil.move(src, dst)
 
 
 def copy(src, dst, verbose=1):


kousu:
<placeholder>

kousu (changes requested):
Thanks for the helpful CI logs! I'm convinced, I'm sure this fixes the immediate issue, and I'll :+1: it for that.

But;

I am skeptical about `spinalcordtoolbox.utils.shell`; it seems like it's trying to be `shutil`, but probably buggier. But that's a bigger issue than one PR can solve.

As for this, this change causes a non-obvious performance hit: `rename()` works atomically, by editing pointers without actually moving bulk data around. This introduces a "surprise it's actually `rsync!`" case that only kicks in when someone is on some systems doing some projects but not others. Why does the test need to do that?

The failed test was

https://github.com/neuropoly/spinalcordtoolbox/blob/fd00c800467aa194fcbae073eb10bc589d772920/unit_testing/cli/test_cli_sct_propseg.py#L20-L22

The exception is in `mv()` but I think the error really [starts here](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495#L2092):

https://github.com/neuropoly/spinalcordtoolbox/blob/fd00c800467aa194fcbae073eb10bc589d772920/spinalcordtoolbox/scripts/sct_propseg.py#L635-L640

This is *trying* to rename a file in `folder_output` to match our naming scheme; it's definitely not trying to trigger moving potentially gigabytes of data from one disk to another. Or in other words, `rename()` is the intended meaning here.

It would be better to look into why renaming is impossible here. Probably what's happening is that `path.join(A, B)` is (correctly) discarding `A` if `B` is an absolute path. So which is it?

It's gotta be the `dst` because the `src` is computed using `basename()`, which discards all folder components. And that makes sense: `dst` is `os.path.join(folder_output, fname_out)`, but `fname_out` [is](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/465807495#L2097) `/tmp/pytest-of-travis/pytest-0/popen-gw0/test_sct_propseg_o_flag0/test_seg.nii.gz`, and apparently `output_folder` is `.`?

So, no, I don't think we should put this in as-is. It's fixing the wrong problem. It *might* help us out in some cases, but maybe there's other parts of the code that used `mv()` trusting it would be `rename()` like its comment says, and we'll have a surprise  performance problem creep up on us in a year.

kousu:
Here's my analysis of the problem, what do you think? https://github.com/neuropoly/spinalcordtoolbox/pull/3081#discussion_r550971021

kousu:
You made this really really easy to review. Thank you so much for that.

joshuacwnewton:
> So in summary: it's mistakenly running isct_propseg with data outputting to the repo. This patch would quietly patch over the mistake by running a bunch of copying from disk to RAM (/tmp is often mounted as a tmpfs).

Thank you for digging deeper! I had a mistaken understanding of the problem.

> So, no, I don't think we should put this in as-is. It's fixing the wrong problem. It _might_ help us out in some cases, but maybe there's other parts of the code that used `mv()` trusting it would be `rename()` like its comment says, and we'll have a surprise performance problem creep up on us in a year.

As an aside, for what it's worth, `shutil.move` [uses `os.rename` under the hood](https://docs.python.org/3/library/shutil.html#shutil.move):

> If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied to dst using copy_function and then removed. 

So, this change shouldn't actually change any behavior wherever we were trusting it for `rename`. But, you're right that it would quietly patch over the mistake, so thanks for this review. :slightly_smiling_face:

joshuacwnewton:
Closing this PR, as it is not needed for now.

(Although, I do agree that these filesystem utilities are a bit suspect still, as @kousu points out.)

> I am skeptical about spinalcordtoolbox.utils.shell; it seems like it's trying to be shutil, but probably buggier. But that's a bigger issue than one PR can solve.

PR #3152, opened 2021-01-07
Fix test download concurrency bug.

kousu:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This replaces using `pytest_sessionstart()` with a [fixture](https://docs.pytest.org/en/stable/fixture.html). This is especially important when using xdist to run tests in parallel, because it [is probably not the right choice](https://stackoverflow.com/questions/12586489/pytest-are-pytest-sessionstart-and-pytest-sessionfinish-valid-hooks/12600154#12600154) while [a fixture is](https://stackoverflow.com/questions/17801300/how-to-run-a-method-before-all-tests-in-all-classes/17844938#17844938).

The order of operations is really complicated here. 

This unfortunately caused a small cascade of changes. At least two files were were generating their own fixtures internally (in a perfectly reasonable way!) that depended on having the test data downloaded, but it wasn't integrated with pytests's fixtures so they would run before the data was downloaded by the main fixture.

There's some ugliness here too: the pseudo-fixtures were being used in `@pytest.mark.parameterize`, but that seems incompatible with fixtures -- pytest doesn't recognize them as fixtures and the test crashes with "function object doesn't have such and such a method". The best workaround I found was https://stackoverflow.com/questions/46368468/run-a-test-with-two-different-pytest-fixtures, so I did that, but it's kind of ugly. I think there's definitely room for improvement here, and I would like some backup on looking into it, please; I just don't understand why I can't write

```
@pytest.fixture
def a():
   return "a"

@pytest.fixture
def b():
   return "b"

@pytest.mark.parameterize("letter", [a,b])
def test_stuff(letter):
    assert letter.upper() == letter.lower().upper()
```

## Linked issues

[...]

PR #3148, opened 2021-01-03
Add new '-create-seg-mid' option to replace bugged '-create-seg -1' behavior

joshuacwnewton:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [x] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

The form "`-create-seg -1,value`" doesn't work with `argparse`, because `argparse` interprets `-1,value` as a separate argument. So, we split this functionality off into a new option, as per https://github.com/neuropoly/spinalcordtoolbox/issues/3119#issuecomment-750388733.

## Linked issues

Fixes #3119.

PR #3156, opened 2021-01-11
Make sure (most) scripts return error code and print usage when no args are passed

### LINKED ISSUE #3154: TypeError: 'NoneType' object is not iterable
After a fresh install of SCT version 3dfd4f81f9e4364dab4d94894ea687a625dbdea5, I run into the following error:
~~~
julien-macbook:~/code/sct $ sct_deepseg

--
Spinal Cord Toolbox (git-master-3dfd4f81f9e4364dab4d94894ea687a625dbdea5)

sct_deepseg 
--

  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_deepseg.py", line 235, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_deepseg.py", line 144, in main
    for file in arguments.i:
TypeError: 'NoneType' object is not iterable
~~~

Probably a regression error related to the recent PR dealing with parser-related settings? @joshuacwnewton @kousu

kousu:
#### GitHub

#### PR contents

## Description

In most cases, calling an SCT script with no args is invalid usage. In that case, a script should return a nonzero error code, then print the usage. But, some scripts weren't doing this. So, this PR fixes those scripts.

The test added to this PR also caught a bug in `sct_label_utils`, so that's fixed here too.

## Linked issues

Fixes #3154.
Related to #3139.


joshuacwnewton (changes requested):
Adapted from https://github.com/neuropoly/spinalcordtoolbox/issues/3154#issuecomment-757619441:

> My understanding was that `-i` was not made mandatory in `sct_deepseg` because it's also proper usage to say `sct_deepseg -list-tasks`. We'd need a way to set it so that [either could be required]( https://stackoverflow.com/questions/11154946/require-either-of-two-arguments-using-argparse); setting just `-i` to be required would muddy things up.

> We faced the same problem in `sct_extract_metric`: We wanted to make `-i` required, but we also wanted to allow an argument to be used without `-i`, so we made an action that exits before `-i` could trigger an error for not being present.

 https://github.com/neuropoly/spinalcordtoolbox/blob/639ca81438763346b9fd4271c3fb696e6febd44a/spinalcordtoolbox/scripts/sct_extract_metric.py#L48-L61

unit_testing/cli/test_cli.py
```diff
@@ -0,0 +1,27 @@
...
+    if not li.startswith("#"):
+        scripts.append(li)
+
+# Exclude scripts where no arguments is valid usage
+no_args_scripts = [s for s in scripts if s not in [
+    'isct_test_ants',          # No args -> tests ants binaries
+    'sct_testing',             # No args -> runs full test suite
+    'sct_check_dependencies',  # No args -> checks dependencies
+    'sct_version'              # No args -> prints version
+]]
```
kousu, line 17: Could these be `@fixture`s?

kousu: Loading test cases from a file dynamically is the case that pytest *doesn't* support natively. So... I guess not really.

joshuacwnewton: Well, it's the exact same problem we're facing in #3152, eh?

Test collection dependent on loading files. :P

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -258,6 +258,9 @@ def main(argv=None):
         # Reparse the arguments
         arguments = parser.parse_args(argv)
 
+    if arguments.script is None:
+        parser.error("The -script argument must be provided, either via command-line or via the -config/-c argument.")
```
kousu, line 262: This is simpler than trying to trick argparse into calling `error()` for us, which is the ultimate goal here: I initially added `required=True` to `-i` to fix https://github.com/neuropoly/spinalcordtoolbox/issues/3154, but that didn't work because these scripts have pseudo-subcommands, and that would cause it to `error()` before we could check for them.

If we were already following the GNU conventions then I would say we should rewrite to use `argparse.add_subparsers()`. But we're not. So yeah, let's do this!

kousu (commented):
Oh I can't approve this because I posted the first commit OH WELL

joshuacwnewton:
I think I might want to ask for a review from @jcohenadad too. Just to double-check my understanding of the proper usage of `sct_run_batch` and `sct_deepseg`. (Maybe I have the required args wrong?)

kousu (commented):
Could we remove the [workaround](https://github.com/neuropoly/spinalcordtoolbox/issues/3154#issuecomment-757619462) in `sct_extract_metric` and replace it with the same simple `.error()` call?

unit_testing/cli/test_cli.py
kousu: I don't really think it's the same. This one would have collection dependent on loading data, that one just has the test execution dependent on loading data. (which is why I am pushing to use fixtures in that one: we *should* be able to without causing pytest any problems)

scripts.txt
```diff
@@ -0,0 +1,58 @@
+# List of tools/scripts that can be called via command-line
```
kousu, line 1: This list is in setup.py already. So long as we *install* the package before testing it (which we *should* be doing:

* https://github.com/pypa/packaging.python.org/issues/320
* https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
* https://hynek.me/articles/testing-packaging/

)

you can use `pkg_resources.get_entry_map('spinalcordtoolbox')['console_scripts'].keys()` below instead of loading this file.

kousu: ps. there's also [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) coming to the stdlib, but it doesn't have a stable API yet.

unit_testing/cli/test_cli.py
kousu, line 17:
```suggestion
scripts = pkg_resources.get_entry_map('spinalcordtoolbox')['console_scripts'].keys()

# Exclude scripts where no arguments is valid usage
no_args_scripts = [s for s in scripts if s not in [
    'isct_test_ants',          # No args -> tests ants binaries
    'sct_testing',             # No args -> runs full test suite
    'sct_check_dependencies',  # No args -> checks dependencies
    'sct_version'              # No args -> prints version
]]
```

jcohenadad (approved):
looks fine on my end 😊

unit_testing/cli/test_cli.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.script_launch_mode('subprocess')
+@pytest.mark.parametrize("script", no_args_scripts)
+def test_argparse_no_arguments(script, script_runner):
+    """Ensure all scripts return error code and show usage descriptions
+    when called with no arguments."""
+    ret = script_runner.run(script)
+    assert ret.returncode is 2
+    assert 'usage' in ret.stdout.lower() or 'usage' in ret.stderr.lower()
```
kousu, line 27: Hm I wonder if this would help: https://docs.pytest.org/en/stable/capture.html#accessing-captured-output-from-a-test-function

```suggestion
@pytest.mark.script_launch_mode('subprocess')
@pytest.mark.parametrize("script", no_args_scripts)
def test_argparse_no_arguments(capfd, script, script_runner):
    """Ensure all scripts return error code and show usage descriptions
    when called with no arguments."""
    ret = script_runner.run(script)
    assert ret.returncode is 2
    out, err = capfd.readouterr
    assert 'usage' in out.lower() or 'usage' in err.lower()
```

If so, then maybe you can avoid subprocess (which is probably slow) with

```suggestion
@pytest.mark.parametrize("script", no_args_scripts)
def test_argparse_no_arguments(capfd, script, script_runner):
    """Ensure all scripts return error code and show usage descriptions
    when called with no arguments."""
    ret = script_runner.run(script)
    assert ret.returncode is 2
    out, err = capfd.readouterr
    assert 'usage' in out.lower() or 'usage' in err.lower()
```

? maybe?

kousu: I dunno, does that help at all?

joshuacwnewton: I implemented a modified version of this using our `main()` function calling in  f94200c. Test went from ~50s to ~5s. :smile:

joshuacwnewton:
When writing the test in f94200c, I discovered a bug that was introduced by me in #3148. :fearful: 

Fixed it in 452f5d1.

unit_testing/cli/test_cli.py
jcohenadad: wow that's crazy! is it because it bypasses all the re-imports

joshuacwnewton:
Oop, just noticed #3160 was just merged. I'll resolve that conflict now. :slightly_smiling_face:

unit_testing/cli/test_cli.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+import pkg_resources
+import importlib
+
+scripts = pkg_resources.get_entry_map('spinalcordtoolbox')['console_scripts'].keys()
```
kousu, line 5: Nice! I'm glad this actually worked.

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -232,10 +232,11 @@ def get_parser():
...
 def main(argv=None):
-    for i, arg in enumerate(argv):
-        if arg == '-create-seg' and len(argv) > i+1 and '-1,' in argv[i+1]:
-            raise DeprecationWarning("The use of '-1' for '-create-seg' has been deprecated. Please use "
-                                     "'-create-seg-mid' instead.")
+    if hasattr(argv, '__iter__'):
+        for i, arg in enumerate(argv):
+            if arg == '-create-seg' and len(argv) > i+1 and '-1,' in argv[i+1]:
+                raise DeprecationWarning("The use of '-1' for '-create-seg' has been deprecated. Please use "
+                                         "'-create-seg-mid' instead.")
```
kousu, line 239: I usually use this for filling in default arguments:

```suggestion
    if argv is None: argv = []
    for i, arg in enumerate(argv):
        if arg == '-create-seg' and len(argv) > i+1 and '-1,' in argv[i+1]:
            raise DeprecationWarning("The use of '-1' for '-create-seg' has been deprecated. Please use "
                                     "'-create-seg-mid' instead.")
```


Ideally you would write `def main(argv=[]):`, but you need the `is None` workaround to avoid [python's mutable default arguments glitch](https://blog.finxter.com/the-mutable-default-argument-in-python/).

kousu (commented):
(Y)

spinalcordtoolbox/scripts/sct_label_utils.py
joshuacwnewton: Hrm... would writing `def main(argv=()):` avoid this, too?

joshuacwnewton (approved):
Whoops, I needed to approve this because it's your PR and I had requested changes. :sweat_smile: 

Up to you regarding the `argv=()`/`if argv is None: argv = []` change. :slightly_smiling_face:

jcohenadad:
@kousu @joshuacwnewton is this one good to go?

spinalcordtoolbox/scripts/sct_label_utils.py
joshuacwnewton: Ahh. After thinking about this more, I don't think we should have `None` as a default value at all. I think main should require `argv` to be passed, specifically as a list using type hinting


```diff
-  def main(argv=None):
+  def main(argv: list):
```

joshuacwnewton:
@jcohenadad I've just had a realization regarding https://github.com/neuropoly/spinalcordtoolbox/pull/3156#discussion_r559028584. I'll make a slight tweak and then it should be good to go.

spinalcordtoolbox/scripts/sct_label_utils.py
joshuacwnewton: Made an issue for this here: #3176.

PR #3160, opened 2021-01-11
Add option to have a custom task

### LINKED ISSUE #3155: Add the possibility to use custom model
### Description

After training a new model with [ivadomed](http://ivadomed.org/), it would be great to be able to use it directly with `sct_deepseg`.

However, it is unclear how to do it. As of SCT version 5.0.1dev (3dfd4f81f9e4364dab4d94894ea687a625dbdea5), the usage specifies the following:
~~~
  -task <str>           Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task. To list
                        available tasks, run: sct_deepseg -list-tasks
  -list-tasks           Display a list of tasks that can be achieved. (default: False)
  -install-task {seg_sc_t2star,seg_mice_sc,seg_mice_gm,seg_tumor_t2,seg_tumor-edema-cavity_t1-t2}
                        Install models that are required for specified task.
~~~

under "-task", it says that we could choose a "custom task", but how to achieve it is unclear: should "-task" point to the .pt model file? Something else?

This issue is somewhat related to #2996, which removed the possibility to point to models. But what is the proposed alternative if someone wants to use a custom model?

@charleygros @andreanne-lemay any thoughts?

andreanne-lemay:
## Description
This PR adds the possibility to use a custom model with sct_deepseg. One or more paths to the folder containing a model can be indicated next to the `-task` flag to use custom model(s). If more than one path is used, the models will be applied one after the other (the output of the previous model will become the input of the next).

`sct_deepseg -i img.nii.gz -task <path(s) to folder containing the packaged model>`

## Linked issue
Fixes #3155

### DIFF
--- a/spinalcordtoolbox/scripts/sct_deepseg.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg.py
@@ -59,7 +59,9 @@ def get_parser():
         "-task",
         nargs="+",
         help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task."
-             " To list available tasks, run: sct_deepseg -list-tasks",
+             " To list available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the "
+             " ivadomed packaged model. More than one path can be indicated for cascaded application of the models. "
+             " See tutorials at https://ivadomed.org/ for more details. ",
         metavar=Metavar.str)
     seg.add_argument(
         "-list-tasks",
--- a/spinalcordtoolbox/scripts/sct_deepseg.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg.py
@@ -60,7 +60,7 @@ def get_parser():
         nargs="+",
         help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task."
              " To list available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the "
-             " ivadomed packaged model (see https://ivadomed.org/en/latest/pretrained_models.html for more details). "
+             " ivadomed packaged model (see https://ivadomed.org/en/latest/pretrained_models.html#packaged-model-format for more details). "
              " More than one path can be indicated for cascaded application of the models.",
         metavar=Metavar.str)
     seg.add_argument(
--- a/spinalcordtoolbox/scripts/sct_deepseg.py
+++ b/spinalcordtoolbox/scripts/sct_deepseg.py
@@ -60,8 +60,8 @@ def get_parser():
         nargs="+",
         help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task."
              " To list available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the "
-             " ivadomed packaged model. More than one path can be indicated for cascaded application of the models. "
-             " See tutorials at https://ivadomed.org/ for more details. ",
+             " ivadomed packaged model (see https://ivadomed.org/en/latest/pretrained_models.html for more details). "
+             " More than one path can be indicated for cascaded application of the models.",
         metavar=Metavar.str)
     seg.add_argument(
         "-list-tasks",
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -149,9 +152,18 @@ def main(argv=None):
     if arguments.task is None:
         parser.error("You need to specify a task.")
 
-    # Check if all input images are provided
-    required_contrasts = deepseg.models.get_required_contrasts(arguments.task)
-    if len(arguments.i) != len(required_contrasts):
+    # Verify if the task is custom
```
jcohenadad, line 155: the primary clause of the if is to check if task is official (the "custom" is the else)
```suggestion
    # Verify if the task is part of the "official" tasks, or if it is pointing to paths containing custom models
```

jcohenadad:
i'll do a full functional test when i have a minute (today i promise)

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -57,8 +57,11 @@ def get_parser():
     seg = parser.add_argument_group('\nTASKS')
     seg.add_argument(
         "-task",
+        nargs="+",
         help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task."
-             " To list available tasks, run: sct_deepseg -list-tasks",
+             " To list available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the "
+             " ivadomed packaged model. More than one path can be indicated for cascaded application of the models. "
```
jcohenadad, line 63: hum... I feel we need to create a specific section on the ivadomed RTD that explains what we mean by "packaged model". It is not clear that we mean a folder, including a .pt and .json having the same basename as the folder name. 

In terms of order, in order to save one PR: we could first clarify on ivadomed, and then add the URL in this function usage.

charleygros: > hum... I feel we need to create a specific section on the ivadomed RTD that explains what we mean by "packaged model". It is not clear that we mean a folder, including a .pt and .json having the same basename as the folder name.

We could move some of t[hese info](https://github.com/ivadomed/ivadomed/wiki/repositories#models) from the wiki to the RTD, eg [here](https://ivadomed.org/en/latest/pretrained_models.html#pre-trained-models). + add info about json and .onnx etc.

andreanne-lemay: Yes I think we could add this information in [Pre-trained models ](https://ivadomed.org/en/latest/pretrained_models.html#pre-trained-models). I think it is fine to leave the conventions where they are since it is more for ivadomed developers.

charleygros: yep we agree about where to move info

jcohenadad:
For cross-reference, the documentation on ivadomed is being updated here: https://github.com/ivadomed/ivadomed/pull/625

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -57,8 +57,11 @@ def get_parser():
     seg = parser.add_argument_group('\nTASKS')
     seg.add_argument(
         "-task",
+        nargs="+",
         help="Task to perform. It could either be a pre-installed task, task that could be installed, or a custom task."
-             " To list available tasks, run: sct_deepseg -list-tasks",
+             " To list available tasks, run: sct_deepseg -list-tasks. To use a custom task, indicate the path to the "
+             " ivadomed packaged model (see https://ivadomed.org/en/latest/pretrained_models.html for more details). "
```
charleygros, line 63:
```suggestion
             " ivadomed packaged model (see https://ivadomed.org/en/latest/pretrained_models.html#packaged-model-format for more details). "
```

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -57,8 +57,11 @@ def get_parser():
...
+             " More than one path can be indicated for cascaded application of the models.",
```
charleygros, line 64:
```suggestion
             " More than one path can be indicated (separated with space) for cascaded application of the models.",
```

charleygros (approved):
Good to go

PR #3196, opened 2021-01-24
(WIP) Test batch_processing.sh in CI

joshuacwnewton:
Just a quick WIP to test:
* Running batch_processing using CI
* Uploading the resulting csv files using the upload-artifact GH action

Don't review, etc. I'll write a proper description when ready. :)

Issue #3194, opened 2021-01-23
Discrepancies in batch_processing results between Linux and macOs


jcohenadad:
Following on https://github.com/neuropoly/spinalcordtoolbox/pull/3050, which ensured intra-OS reproducible results for `batch_processing.sh`, we still observe slight discrepancies across OSs. Example, from https://github.com/neuropoly/spinalcordtoolbox/releases/tag/5.0.1:

**Linux:**
~~~
mt/MTR(WM):      54.38365862512906
~~~

**macOS:**
~~~
mt/MTR(WM):      54.414474990448184
~~~

I _think_ this discrepancy originates from some code in `sct_extract_metric`. There might be some numerical operations that are OS-dependent. By playing with the different scipy/numpy/scikit-learn libraries, we might be able to find a workaround.

joshuacwnewton:
I've edited the description slightly, because this happens with both mt/MTR and dmri/FA calculations.

joshuacwnewton:
I have one new development to report for this issue! 

I spent a little bit of time trying to get `batch_processing.sh` working for our native Windows installation (#3770), and it looks like the values that Windows produces [match the Ubuntu results](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/6032714214?check_suite_focus=true#step:10:42)? 

I was assuming that all 3 OSs would have produced different results, and that the test would have failed for Windows, too. But, it looks like it's only macOS that's the outlier here.

joshuacwnewton:
I'm commenting on this issue to expand its scope a little, because there have been new cross-OS discrepancies uncovered in recent weeks.

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5185#issuecomment-4984009824

> @mguaypaq But I'm not sure why ubuntu-24.04 got different numerical results from ubuntu-22.04 in this test run:
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/29358399139/job/87171817374#step:10:166
> 
> ```
> FAILED testing/batch_single_subject/test_batch_single_subject.py::test_batch_single_subject_results[mt/mtr_in_wm.csv-0-MAP()] - assert 53.89463396000851 == 54.05076215545097 ± 5.4e-05
>   
>   comparison failed
>   Obtained: 53.89463396000851
>   Expected: 54.05076215545097 ± 5.4e-05
[...]

Issue #3140, opened 2020-12-30
Make CI test the `SCT_INSTALL_TYPE=package` path


kousu:
While working on #3125 I came across a quirk: Actions' runners are more bare bones than Travis, and as a result `actions/checkout@v2`, the standard script that downloads our repo to test it, will *get the .zip of the repo from the Github API*.

This fools our install script because

https://github.com/neuropoly/spinalcordtoolbox/blob/6172cd22feeb0f64ac1bd68bce3f89f7109e68b8/install_sct#L376-L384

https://github.com/neuropoly/spinalcordtoolbox/blob/6172cd22feeb0f64ac1bd68bce3f89f7109e68b8/install_sct#L416-L421

so it ends up installing to `/github/home/sct_dev/python`.

Trouble is, we never test this case in CI, even though it's our official release method. We should make sure to cover this case. We at least need a spot-check on one platform that it's behaving. I don't know exactly what that looks like; maybe we `rm -rf .git/` before running the installer?

kousu:
For example

https://github.com/neuropoly/spinalcordtoolbox/blob/6172cd22feeb0f64ac1bd68bce3f89f7109e68b8/spinalcordtoolbox/utils/sys.py#L352-L358

is buggy. It is incompatible with being installed under; But we've never ever run the tests outside of the source repo. It *should* be using

https://github.com/neuropoly/spinalcordtoolbox/blob/6172cd22feeb0f64ac1bd68bce3f89f7109e68b8/spinalcordtoolbox/utils/sys.py#L485

defined even in the same file(!).

The tests should be just as good if run in "release" mode as in "dev" mode -- arguably, you'd want them even more reliable there.

PR #3198, opened 2021-01-25
Run and test batch_processing.sh using GitHub Actions

### LINKED ISSUE #2888: Investigate running batch_processing during CI

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR follows the experimentation done in #3196. It introduces a new GH Actions that does the following:
* Runs `batch_processing.sh`
* Runs a test that compares the results to cached csv files committed to the repo.

Some remaining design questions:
- [X] When should we run this?
     * Right now, this PR adopts the triggers used in #3125. See justification here:
- [X] What tolerance values should be chosen for each metric?
     * I've chosen tolerances as small as can be without the tests failing, so that any deviations will be caught.
- [ ] Should any other new values be tested?

## Linked issues

Fixes #2888.


joshuacwnewton:
The [test is failing](https://github.com/neuropoly/spinalcordtoolbox/runs/1764635577#step:6:41), but that's intentional. I set it to showcase how different values need different tolerances. 

(Also, look at the pretty colour in that test! I found a workaround to re-enable colour in GH Actions.)

https://github.com/neuropoly/spinalcordtoolbox/blob/796d3011b1e7b50104b833a979272b1fae4ab2d4/.github/workflows/tests.yml#L56-L58

jcohenadad:
> (Also, look at the pretty colour in that test! I found a workaround to re-enable colour in GH Actions.)

awesome, i love pretty colors!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      # We use the REST API to fetch the most recently uploaded results from batch_processing.
+      # jq and awk are used to filter the JSON response by artifact ID, which is then used to download the right .zip.
+      # Relevant documentation: https://docs.github.com/en/rest/reference/actions#artifacts
+      - name: Fetch past results
+        run: |
+          export NEWEST_ARTIFACT_ID=$(curl -H "Accept: application/vnd.github.v3+json" \
+                                           https://api.github.com/repos/neuropoly/spinalcordtoolbox/actions/artifacts |
+                                      jq '.artifacts[].id' |
+                                      awk '{for(i=1;i<=NF;i++) if($i>maxval) maxval=$i;}; END { print maxval;}')
```
Drulex, line 50: Not a problem at all, but `jq` has the capability to sort elements so you can use that directly instead of piping into`awk`:

```bash
curl /artifacts | jq '.artifacts|=sort_by(.id)|.artifacts[-1].id'
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+          curl -u admin:${{ secrets.GITHUB_TOKEN }} \
```
Drulex, line 51: Is there a reason for passing username and password to this request? From the docs it looks like anyone with read access can request this endpoint (just like the `list` endpoint above).

.github/workflows/tests.yml
joshuacwnewton: Thank you for pointing this out. Truthfully, I had never used `jq` before, so I appreciate tips like this. :)

.github/workflows/tests.yml
joshuacwnewton: I haven't tested extensively, but locally when I try without `-u`, it gives me this response:

```bash
joshua@XPS-15-9560:~$ curl -H "Accept: application/vnd.github.v3+json" -L "https://api.github.com/repos/neuropoly/spinalcordtoolbox/actions/artifacts/36987543/zip"
{
  "message": "You must have the actions scope to download artifacts.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#download-an-artifact"
}
```

So I think it's necessary to explicitly identify as a user in the request? I tested it locally using a personal access token and it worked OK.

It's hard to find documentation on this though -- the closest I've found for background context is the issue description in: https://github.com/actions/upload-artifact/issues/51#issue-558653519

> Likewise, when the Actions API was released, API requests worked without any authentication. One could simply query the API for an artifact download URL and redirect the guests to it. - All of that worked without GitHub account and could have worked from JavaScript or a small lightweight webservice which redirects the end-user to the latest artifact download through HTTP redirects (I wrote https://github.com/JayFoxRox/GitHub-artifact-URL for this purpose).
>
> However, within the last days these artifact download URLs were suddenly made private - they only work for registered GitHub users now. Everyone clicking the download button on our website (who isn't logged into GitHub) gets a confusing 404 error for direct artifact download URLs now.

joshuacwnewton:
Here's a question that's been on my mind this past day: Should we even use this Artifact API? Or, would it be better to hardcode the results via a commit to the repo?

Pros for Artifact API
* In theory, it would be a more seamless way of updating the cached values.

Cons for Artifact API
* It's a pretty opaque -- there's not really an easy way to check what values are being stored. Commits would force us to be more transparent/explicit about what values are being stored. To update them, we would need to be clear that something has changed.
* It feels brittle. It relies on refreshing the values within a 90 day window, which is fine, but it's inherently temporary. Also, what happens if the `.csv` files ever don't get uploaded/downloaded properly? Commits feel much more reliable -- no chance of "oops, the cached values are gone!"

Using the Artifacts API was a fun experiment, but I'm leaning towards not using it at all.

kousu:
Since you're using this as a cache, I lean towards committing them to the repo too. 

I keep linking this, and I wish I had a better example than a js project, but https://jestjs.io/docs/en/snapshot-testing shows us the way to do this, imo: when values change it fails the test; to update the values on purpose, the process is just `rm sct_example_data/t2/csa_c2c3.csv; pytest; git add -u; git commit -m "update snapshot tests because we changed [...]"`. Uh, assume we have some plugin that can make the snapshots for us.

Drulex:
> Here's a question that's been on my mind this past day: Should we even use this Artifact API? Or, would it be better to hardcode the results via a commit to the repo?

I agree that using the artifacts API introduces unneeded complexity. At the end of the day we just need to compare numbers. Committing the files to the repo sounds reasonable to me.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+               -L "https://api.github.com/repos/neuropoly/spinalcordtoolbox/actions/artifacts/$NEWEST_ARTIFACT_ID/zip" \
+               --output "${{ env.zip_filepath }}"
+
+      - name: Validate new results
+        env:
+          PY_COLORS: "1"
+        run: |
+          source python/etc/profile.d/conda.sh
+          conda activate venv_sct
+          BATCH_PROCESSING_ZIP_FILEPATH="${{ env.zip_filepath }}" \
```
kousu, line 62: You already have this as `$zip_filepath`! Because you set it in `env`.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+    branches:
+      - '*'
+  schedule:
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onschedule
+    # > Scheduled workflows run on the latest commit on the default or base branch
+    # i.e. this can only run on master
+    - cron:  '0 11 * * *'
+
+env:
+  zip_filepath: "batch_processing_results.zip"
```
kousu, line 18: shell convention:

```suggestion
  ZIP_FILEPATH: "batch_processing_results.zip"
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
+name: Tests
```
kousu, line 1:
```suggestion
name: Test batch_processing.sh
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - release
+  pull_request:
+    branches:
+      - '*'
```
kousu, line 10: This will make it run for everyone everywhere! We probably wanna limit this one to just the cronjob.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+  batch_processing:
+    name: Test batch_processing.sh
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout SCT
+        uses: actions/checkout@v2
+
+      # install_sct edits ~/.bashrc, but those environment changes don't get passed to subsequent steps in GH Actions.
+      # So, we filter through the .bashrc and pass the values to $GITHUB_ENV and $GITHUB_PATH.
+      # Relevant documentation: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files
```
kousu, line 30: Another reason we should be using more standard installation tools/paths.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install SCT
+        run: |
+          PIP_PROGRESS_BAR=off ./install_sct -y
+          cat ~/.bashrc | grep "export SCT_DIR" | cut -d " " -f 2 >> $GITHUB_ENV
+          cat ~/.bashrc | grep "export MPLBACKEND" | cut -d " " -f 2 >> $GITHUB_ENV
+          cat ~/.bashrc | grep "export PATH" | grep -o "/.*" | cut -d ':' -f 1 >> $GITHUB_PATH
```
kousu, line 36: What if you did

```suggestion
      - name: Install SCT
        run: |
          PIP_PROGRESS_BAR=off ./install_sct -y
          . ~/.bashrc
```

Does that work I wonder?

This is what I've been assuming is the closest thing we have to a canonical installation instruction. Using other methods non-canonical installations already gave me some trouble in #3125 because

https://github.com/neuropoly/spinalcordtoolbox/blob/073bac3b0307fb90ebc49c98c09058044fa20bea/.ci.sh#L16

doesn't work right in #3140 installs. I think `. ~/.bashrc` is more reliable, though it might be a toss up.

.github/workflows/tests.yml
joshuacwnewton: :facepalm:

.github/workflows/tests.yml
joshuacwnewton: My thinking was that, by running in on PRs, we can catch changes in batch_processing values, then make decisions about  within that PR:

* If the change is undesired, then we can edit the PR to make sure batch_processing stays the same.
* If the change was expected, then we can update the values in the cached .csv files within the same PR.

.github/workflows/tests.yml
joshuacwnewton: Good thinking. I'll give it a shot. :)

.github/workflows/tests.yml
kousu: Okay, cool, so that was intentional.

I think batch_processing takes like 10 minutes to run though, so it might not be a good candidate for an every-day kinda thing. But then people need to know how to figure out what it means when the error kicks them after their PR has been merged.

joshuacwnewton: > I think batch_processing takes like 10 minutes to run though, so it might not be a good candidate for an every-day kinda thing.

Good point... the total time for the full workflow is ~21m, while with #3125, the slowest job takes only 10m. 

The alternative could be making batch_processing a non-required check on PRs. That way it still runs, and we can be aware of any changes, but also intentionally merge PRs without waiting for it to finish on a case-by-case basis.

.github/workflows/tests.yml
joshuacwnewton: No dice: 757830e :(

joshuacwnewton:
`batch_processing.sh` is now failing because of https://github.com/neuropoly/spinalcordtoolbox/issues/3202.

This is actually pretty neat -- the CI run of `batch_processing.sh` caught a bug that the tests didn't catch.

joshuacwnewton:
Huzzah, #3203 fixed the failing CI in this PR. 

I have one more thing I want to try: Running `batch_processing.sh` on both Ubuntu and macOS to more clearly demonstrate #3194.

.github/workflows/tests.yml
joshuacwnewton: Right now it's showing as this:

Screenshot from 2021-02-01 17-22-29

joshuacwnewton:
Well, the macOS CI job did fail, but not for the reason I predicted. Instead, it ran into the concurrency bug from #2957 which should be fixed by #3152.

I'm going to try restarting the build...

.github/workflows/tests.yml
kousu: Oh of course, that's the best of both worlds. Assuming we can afford that.

On Travis, our CPU time was throttled. While working on this I did manage to hit the throttle, but it seems that Actions is a lot more generous: I only hit it after having something like 30 hung processes; on Travis, the throttle is more like... 4 per account. I would be worried about tens of 10-minute checks running in parallel eating into our throttle.

This is why I wanted to give this workflow a different name (https://github.com/neuropoly/spinalcordtoolbox/pull/3198/files#r564848578), because I wasn't imagining we could run it in the same batch; and we probably don't want to run it on all the platforms, either. Platform differences *could* cause numerical differences, but I think `batch_processing.sh` is more of a lumbering large-scale check and it's a lot more likely that changes will come from changed algorithms than from ANTS being miscompiled but only for MacOS. Of course, if that *did* happen and we missed it it would be very bad.

Maybe @jcohenadad can make a ruling on how often we should plan to run `batch_processing.sh`.

.github/workflows/tests.yml
kousu: Oh, right. I see. But I don't really think they should sync up: https://github.com/neuropoly/spinalcordtoolbox/pull/3198/files#r568224050

:shrug: hbu?

.github/workflows/tests.yml
kousu: Drat.

Okay maybe this is where this PR and mine *should* be syncing up: we could split `.ci.sh` into:

* `.ci.sh install`
* `.ci.sh test`
* `.ci.sh batch_processing`

The latter two would (using a subroutine instead of copy-pasting between workflow files) both source `~/.bashrc` the same way before running their tests. And then we can use that as a launching point for #2674 like we were talking about at the meeting this week.

unit_testing/batch_processing/test_batch_processing.py
```diff
@@ -0,0 +1,41 @@
...
+
+    with open(csv_filepath_old, newline='') as csvfile:
+        reader = csv.reader(csvfile, delimiter=',')
+        metric_value_old = float([row for row in reader][row][pos])  # Row/position varies depending on metric
+
+    with open(csv_filepath_new, newline='') as csvfile:
+        reader = csv.reader(csvfile, delimiter=',')
+        metric_value_new = float([row for row in reader][row][pos])  # Row/position varies depending on metric
+
+    assert metric_value_new == pytest.approx(metric_value_old, rel=rel_tolerance)
```
kousu, line 41: Use `np.genfromtxt` here. I discovered it while helping Julien with https://github.com/neuropoly/spinalcordtoolbox/pull/3050#discussion_r557700055:

https://github.com/neuropoly/spinalcordtoolbox/blob/6b66ae770ec85650a74f3b201152831f72d491f0/unit_testing/test_register.py#L202-L206

kousu: Is this only spot-checking a single entry right now?

joshuacwnewton: Nifty! Thanks for the tip. 😄

joshuacwnewton: > Is this only spot-checking a single entry right now?

These 6 values:

https://github.com/neuropoly/spinalcordtoolbox/blob/64315ccdc50eec434a391739f46a0fe177419b4b/unit_testing/batch_processing/test_batch_processing.py#L19-L24

Are mimicking these 6 values:

https://github.com/neuropoly/spinalcordtoolbox/blob/255140ade8632524122e09ea53908bb56cf10ca3/batch_processing.sh#L257-L262

joshuacwnewton:
Huzzah x2! 

#3194 suggests that 3 values (`mt/MTR(WM)`, `dmri/FA(CST_r)`, and `dmri/FA(CST_l)`) aren't consistent between Ubuntu and macOS. And, as expected, [those 3 tests](https://github.com/neuropoly/spinalcordtoolbox/pull/3198/checks?check_run_id=1810856212) are failing.

.github/workflows/tests.yml
joshuacwnewton: Since your PR (#3125) is dependent on #3152, and that may take some time, perhaps we could stick with `steps:` here temporarily, then switch to your suggestion as part of a later PR.

joshuacwnewton: I've added a note referencing this in 42a6520.

unit_testing/batch_processing/test_batch_processing.py
joshuacwnewton: Do you think we should be checking more than 1 value per csv? Or am I misunderstanding?

At the very least, I've added a note about this in fee0953.

Drulex (commented):
Thanks for the update @joshuacwnewton, this looks good.

At what point are new values added to the cached results file?

joshuacwnewton:
My apologies, @Drulex. I thought I wrote a reply, but I must not have submitted it.

> At what point are new values added to the cached results file?

To clarify, do you mean updating the values for existing metrics? Or, adding new (different) metrics to test?

* For updating existing metrics, I was thinking we would decide on a case-by-case basis whenever a PR causes a shift:
    * If we expected a PR to shift the values, we would update the cached values within that PR. 
    * If the shift was unexpected, then we would look more closely at the PR to see why it caused a change.
* For adding new vmetrics, it would depend on which other metrics are valuable to test. I'm still becoming familiar with typical MRI processing pipelines, so instead @jcohenadad (or a student in the lab) might be the best to ask.

Drulex:
> If we expected a PR to shift the values, we would update the cached values within that PR. (This has the benefit of creating a history of changes that are linked with specific PR changesets.)

That answers the question thanks!

.github/workflows/test-batch-processing.yml
```diff
@@ -0,0 +1,54 @@
...
+  # Used in 'pytest.mark.skipif()' for test, which prevents the test from being run during the usual test suite
+  BATCH_PROCESSING_CI_JOB: "1"
+
+jobs:
+  batch_processing:
+    name: Test batch_processing.sh
+    strategy:
+      matrix:
+        os: [ ubuntu-18.04 ]  # TODO: Change to [ ubuntu-18.04, macos-10.15 ]
+                              # macOS currently fails due to https://github.com/neuropoly/spinalcordtoolbox/issues/3194
```
joshuacwnewton, line 29: I've removed `macos-10.15` in b978a6e, because I figure it will only be useful when the discrepancies are fixed.

joshuacwnewton (commented):
I've created a follow-up issue in https://github.com/neuropoly/spinalcordtoolbox/issues/3250 to investigate testing more than just the current 6 values. For now, I think this PR would still be useful as-is, and perhaps more values could be added in another PR.

@kousu @Drulex Is there anything else, then, that needs to be done before this can be approved/merged?

Drulex:
> I've created a follow-up issue in #3250 to investigate testing more than just the current 6 values. For now, I think this PR would still be useful as-is, and perhaps more values could be added in another PR.
> 
> @kousu @Drulex Is there anything else, then, that needs to be done before this can be approved/merged?

Looks ok to me.

Drulex (approved):

PR #3203, opened 2021-01-31
Fix recently-introduced faulty slice index comparison

### LINKED ISSUE #3202: New aggregate_slicewise.py check causes error in batch_processing.sh
### Description

[One of the commands](https://github.com/neuropoly/spinalcordtoolbox/pull/3198/checks?check_run_id=1802242428) in `batch_processing.sh` failed the new check added in #3120. 

```python
--
Spinal Cord Toolbox (git-jn/2888-test-batch_processing.sh-881b504ab2ea7140375186c54e2e9b8bb3f5ab51*)

sct_extract_metric -i mtr.nii.gz -method map -o mtr_in_wm.csv -l 51 -vert 2:5
--

ValueError: Shape mismatch between vertfile [22] and metric [61]). Please verify that your vertfile has the same number of slices as your input image.
```

This is my mistake -- the new check is comparing the wrong dimension:

https://github.com/neuropoly/spinalcordtoolbox/blob/49ed68e029886661ad93546f5c0cd981839a6541/spinalcordtoolbox/aggregate_slicewise.py#L237

`[0]` does not refer to the proper slice index for 3D RPI data. Changing to `[-1]` will properly fetch slices from the last dimension of the `Metric()`. 

### Postmortem

This slipped through because tests in `test_aggregate_slicewise.py` use only 1D `Metric()` data.

https://github.com/neuropoly/spinalcordtoolbox/blob/49ed68e029886661ad93546f5c0cd981839a6541/unit_testing/test_aggregate_slicewise.py#L25-L32

Because of this, I falsely assumed all metrics were 1D, however `mtr.nii.gz` is a 3D metric. So, we should include a dummy 3D metric in there as well.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

In #3120, I mistakenly assumed that all metrics would be 1D, so I indexed the metric using `[0]`. This updates the check to index the right dimension depending on the metric shape.

This PR also adds a new dummy metric that is 3D, so that the tests are more representative of real-world data. 

## Linked issues

Fixes #3202.

### DIFF
--- a/spinalcordtoolbox/aggregate_slicewise.py
+++ b/spinalcordtoolbox/aggregate_slicewise.py
@@ -233,8 +233,10 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
     :return: Aggregated metric
     """
     if vert_level:
-        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[-1]
-        metric_slices = metric.data.shape[-1]
+        # Assumption: vert_level image will only ever be 3D or 4D
+        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[2]
+        # Get slices ('z') from metrics regardless of whether they're 1D [z], 3D [x, y, z], and 4D [x, y, z, t]
+        metric_slices = metric.data.shape[2] if len(metric.data.shape) >= 3 else metric.data.shape[0]
         if vert_level_slices != metric_slices:
             raise ValueError(f"Shape mismatch between vertfile [{vert_level_slices}] and metric [{metric_slices}]). "
                              f"Please verify that your vertfile has the same number of slices as your input image, "
--- a/spinalcordtoolbox/aggregate_slicewise.py
+++ b/spinalcordtoolbox/aggregate_slicewise.py
@@ -233,11 +233,12 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
     :return: Aggregated metric
     """
     if vert_level:
-        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[2]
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -233,11 +233,12 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
     :return: Aggregated metric
     """
     if vert_level:
-        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[2]
-        metric_slices = metric.data.shape[0]
+        vert_level_slices = Image(vert_level).change_orientation('RPI').data.shape[-1]
+        metric_slices = metric.data.shape[-1]
```
joshuacwnewton, line 237: Hrmm, wait a minute. Is `[-1]` what I want? 4D images have `t` as their last dimension, right?

joshuacwnewton:
Note: The build didn't fail from this PR's changes. 

For some reason, the `centos:8` Travis jobs fail intermittently. The job says `Unable to find image 'centos:8' locally`, then it tries to fetch the image, but it only succeeds some of the time. Since the issue happens intermittently, usually restarting the job causes it to pass.

I might open a new issue for this, but considering we'll soon be transitioning away from Travis, I wonder if that's even worth it. For now, I've simply restarted the job.

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: I've changed this to more explicitly index the I-S slices depending on the shape of the data.

Drulex (approved):
+1 for adding the failed test case!

Issue #3219, opened 2021-02-13
Move scattered testing files/folders into a central location


joshuacwnewton:
Currently, testing code is a little disorganized:

```python
# Current
spinalcordtoolbox/
├─ spinalcordtoolbox/
│  ├─ testing/
│  │  ├─ create_test_data.py  # Contains dummy data fixtures 
├─ testing/
│  ├─ sct_crop/  # There are 4 subfolders like this, all containing sh scripts 
│  │  ├─ test_sct_crop.sh
│  ├─ test_sct_analyze_lesion.py  # 'sct_testing' tests
│  ├─ ...  
├─ tests/  # Do these tests get invoked by anything?
│  ├─ test_gui.py      # Uses 'unittest' 
│  ├─ test_reports.py  # Uses 'pytest' 
├─ unit_testing/  # Main directory where new tests are written
│  ├─ cli/
│  │  ├─ test_cli_sct_analyze_lesion.py  # Functional tests (pytest)
│  │  ├─ ...
│  ├─ test_aggregate_slicewise.py  # Unit tests (pytest)
│  ├─ ...
├─ conftest.py  # Data downloading/integrity checking
├─ setup.cfg    # Contains pytest configuration
```

I think if the test structure were simpler, it would make the project easier to navigate and contribute to:

```python
# Proposed
spinalcordtoolbox/
├─ testing/  # Main directory where new tests are written
│  ├─ api/  # Unit tests (pytest)
│  │  ├─ test_aggregate_slicewise.py  
│  │  ├─ test_gui.py
│  │  ├─ test_reports.py 
│  │  ├─ ...
│  ├─ cli/  # Functional tests (pytest)
│  │  ├─ test_sct_analyze_lesion.py 
│  │  ├─ ...
│  ├─ conftest.py  # Data downloading/integrity checking + dummy data fixtures
├─ setup.cfg    # Contains pytest configuration
```

#### Summary of proposed changes

* Cosmetic changes
    - [x] Rename the main testing directory to `testing/` instead of `unit_testing/`
    - [x] Move pytest CLI/functional tests into `tests/cli`
    - [x] Move pytest API/unit tests into `tests/api`
    - [x] Move `conftest.py` up a level
* Likely to require rewriting/refactoring
    - [x] Move 'sct_testing' tests into `tests/cli` (#3093)
[...]

Issue #3093, opened 2020-12-12
Replace older `sct_testing` tests with new `pytest` tests


joshuacwnewton:
In #2893, we decided to move away from `sct_testing` and instead start using `pytest`. As a short-term solution, we minimally wrapped the old tests so that they can be run as-is using `pytest`. As a long-term solution, though, we should start replacing the old `sct_testing` tests with equivalent pure `pytest` tests. 

<details>
<summary> List of sct_testing files to convert </summary>

## Tests already converted to `pytest`
- [X] test_sct_deepseg_sc.py
- [X] test_sct_get_centerline.py

## Tests that only run the CLI command (PR #3244)

Listed as "Not implemented.":
- [x] test_sct_compute_mtr.py
- [x] test_sct_compute_snr.py
- [x] test_sct_concat_transfo.py
- [x] test_sct_create_mask.py
- [x] test_sct_deepseg_lesion.py
- [x] test_sct_dmri_compute_bvalue.py
- [x] test_sct_dmri_compute_dti.py
- [x] test_sct_dmri_concat_b0_and_dwi.py
- [x] test_sct_dmri_concat_bvals.py
- [x] test_sct_dmri_concat_bvecs.py
- [x] test_sct_dmri_transpose_bvecs.py
- [x] test_sct_flatten_sagittal.py
- [x] test_sct_fmri_compute_tsnr.py
- [x] test_sct_fmri_moco.py
- [x] test_sct_label_vertebrae.py
- [x] test_sct_maths.py
- [x] test_sct_merge_images.py
- [x] test_sct_process_segmentation.py
- [x] test_sct_qc.py
- [x] test_sct_smooth_spinalcord.py
- [x] test_sct_straighten_spinalcord.py
- [x] test_sct_warp_template.py

Tests that are currently broken:
- [x] test_sct_denoising_onlm.py

## Tests that run the CLI command + `test_integrity` (PR #3246)

Tests that only perform very basic `test_integrity` checks (e.g. file existence check, value comparison with cached values, etc.):
- [x] test_sct_analyze_lesion.py
- [x] test_sct_compute_ernst_angle.py
- [x] test_sct_compute_hausdorff_distance.py
- [x] test_sct_convert.py
- [x] test_sct_crop_image.py
- [x] test_sct_dice_coefficient.py
- [x] test_sct_extract_metric.py
- [x] test_sct_resample.py

Tests with slightly more complex assertions (e.g. some calculations are involved):
[...]

Issue #2853, opened 2020-08-18
Replace verbose flag in API with environment variable


jcohenadad:
Currently a lot of scripts and API functions in SCT need to carry a flag 'verbose', which sets the logging level at the beginning of the CLI. More details [in the wiki](https://github.com/neuropoly/spinalcordtoolbox/wiki/logging#how-logging-appears-to-users-of-sct-scripts).

Instead of keep maintaining a verbose argument across all functions, which is prone to errors (e.g. https://github.com/neuropoly/spinalcordtoolbox/pull/2852), why not getting rid of it and instead rely on the logging level?

joshuacwnewton:
I very much agree with the spirit of having a unified verbose setting.

Another possible point in favor of a change is, currently, passing around per-function `verbose` arguments leads to unintuitive behavior like this: 
```python
sct.run(['sct_label_utils', '-i', 'point_RPI.nii'], verbose=param.verbose)
```
When I read this for the first time, I assumed `verbose` here would control the verbosity for the script that was being run (i.e.`sct_label_utils`). But, it turns out to get `sct_label_utils` to run in verbose mode, you'd have to set a second verbose parameter like this:
```python
sct.run(['sct_label_utils', '-i', 'point_RPI.nii', '-v'], verbose=param.verbose)
```
I think this sort of behavior could be improved with the suggestion you're making, @jcohenadad.

joshuacwnewton:
This issue possibly overlaps with #1968.

joshuacwnewton:
Duplicate of #1968.

PR #3244, opened 2021-02-23
Migrate (to pytest) the 'sct_testing' tests marked "Not implemented"

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR migrates 23/43 of the `sct_testing` tests (the ones that _did not_ have `test_integrity` implemented). It is split up from #3246 to make reviewing it more manageable.

* If you want to run **just** the new sct_testing tests, you can call `pytest -m sct_testing`.
* Also adds a fixture in `conftest.py` that can be invoked to run a test inside the `sct_testing_data/` directory. This is to preserve the old behaviour of `sct_testing`.

### Future work

I am leaving the old tests for now so that they can still be called via `sct_testing`. But, once both PRs are merged, `sct_testing` should be removed entirely (e.g. as part of cleanup in #3219).

Also, checks should eventually be added to these migrated tests to make them more useful. But, this PR is just about moving away from `sct_testing`, so improving the tests is out of scope.

### Reviewing this PR

Each new test has a corresponding `sct_testing` test. They **should** match (e.g. the new test in `unit_testing/cli/test_cli_sct_analyze_lesion.py` should test the same things as whatever was in `testing/test_sct_analyze_lesion.py` and so on).

So, to review, it would help to verify that everything was translated correctly. This may mean bouncing back and forth between the old tests and the new tests, which is admittedly a bit of a pain. (But, once this effort is done, we won't have to do this ever again, because our full test suite will be in pytest only.)

## Linked issues

Partially addresses #3093.


joshuacwnewton:
As an FYI, the errored Travis CI job seems to be an anomaly, but I'm not able to restart Travis jobs, so it persists.

unit_testing/cli/test_cli_sct_denoising_onlm.py
```diff
@@ -0,0 +1,14 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_denoising_onlm_no_checks():
+    """Run the CLI script without checking results.
+    TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
+    sct_denoising_onlm.main(argv=['-i', 't2/t2.nii.gz', '-v', '2'])
```
joshuacwnewton, line 14: Hm, this test keeps hanging on specifically Catalina. At first I thought it was a Travis glitch, but it's happened multiple times now. (e.g. [Job #15237.3](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/488252879#L1266), [Job #15202.3](https://travis-ci.com/github/neuropoly/spinalcordtoolbox/jobs/486386514#L1266).

So, this may actually be an issue with the underlying script + Catalina.

dyt811:
Anyone with OS Catalina locally able to verify this?

unit_testing/cli/test_cli_sct_denoising_onlm.py
jcohenadad: i'm catalina (still)-- will check in ~15min

jcohenadad: hum, no problem on my end:
https://gist.github.com/jcohenadad/82df099e957005a28b32bb96700f91ed

dyt811: CI issues are... the least fun issues to debug. My suggestion would probably be increasing granularity of the logging temporarily (or add more debug print out/checks for that specific function being pytested) to at least start narrow down the region where the hanging happens on CI and potentially rewrite/reimplement those regions if this is consistently giving issues.

My colleague once had a CI issue on that was not reproducible on any local machines and it turned out most likely to be a low level error with the way we update pointers to items within a list dynamically while trying to be efficient. We were only able to narrow down the cause to a region of code and rewrote it but never fully 100% sure what was the precise cause. It costed us about 3 months of off-and-on part-time trying to identify the cause. I really hope what you see now is much easier to debug...

jcohenadad: also: it is not worth spending hours on this since `sct_denoising_onlm` is probably rarely used

joshuacwnewton: > also: it is not worth spending hours on this since sct_denoising_onlm is probably rarely used

:+1:

> My suggestion would probably be increasing granularity of the logging temporarily (or add more debug print out/checks for that specific function being pytested)

As an aside, this is a helpful observation, and another motivation for #2853. Right now, I don't believe SCT has an easy way to change verbosity on the fly during testing -- the `-v` argument has to be specified individually for each and every `main` call.

(Also, use of verbosity in SCT is a bit complicated right now, and hard to summarize. I tried in https://github.com/neuropoly/spinalcordtoolbox/pull/2897#issue-489598310, but that summary could use cleaning up.)

joshuacwnewton:
I'm going to close both PRs for now. 

I realize I've made these pull requests very challenging to review. So, I would rather discuss with the team and come up with a plan for reviewing, instead of leaving these gigantic PRs open.

(But, I think the converted tests themselves are still good, so I'm going to leave the branches to preserve the work. I'll just need to split things up further, I think.)

PR #3253, opened 2021-02-26
Interactive QC assessment: Add Pass/Fail/Artifact and download YAML file.

### LINKED ISSUE #3187: Enable to interactively assess QC on a per-subject basis and export file listing subjects to correct
Currently, the QC report enables to go through subjects and verify the quality of segmentation, etc.

If segmentations are bad, the user manually note (eg on a piece of paper) what subjects were problematic, and then generate e.g. a YAML file that lists those subjects. This should ideally be automatized as follows:
- During QC, the user could press "space" for ✅ , and "space space" for ❌ 
- At any moment, the user can press a button that would write a file listing the subjects that warrant correction

Drulex:
#### GitHub

#### PR contents

## Description
- [x] Interactively assess QC report entries by toggling a "problematic" flag using keyboard (`p` key)
- [x] Generate a `yml` results file automatically based on entries marked as problematic
- [x] Download `yml` file button
- [x] Download button disabled if no problematic entries, enabled otherwise
- [x] Add additional key bindings (vi style: `hjkl`) since current ones may clash with default keyboard shortcuts
- [x] Table automatically updated (seems that data model is updated, but view remains unchanged until an event like column sorting is applied) -- Fixed by 29d32ed.

TODO:
- [x] Change QC column name to something more intuitive.
- [x] Change QC values from True/False to three or four value scheme (empty, :white_check_mark:, :x:, and perhaps data quality)
- [x] Change keyboard shortcut from "p" to something on the left side to enabled two-handed QC.
- [x] Make keyboard shortcuts (and how they function) more clearly discoverable/understandable. (Currently we have "Keyboard shortcuts:" text, but it points to buttons, and doesn't actually show all of the keyboard shortcuts or how they're supposed to work.)
- [x] Provide a shortcut to reset the statuses of all images.
- [x] Make the QC column the first column (so that it's not necessary to scroll over to see the QC status.)
- [x] Address `seg_type` "FIXME" comment (see https://github.com/neuropoly/spinalcordtoolbox/pull/3253#discussion_r610327389).
- [x] Test on multiple browsers and operating systems
- [x] Change the name "About" (top right) to "SCT Home" and redirect to RTD page (currently it links to GH).

## Linked issues
Resolves #3187


Drulex:
If anyone javascript-savvy can help I would appreciate. I am currently unable to figure out why the view (table) isn't updated when the data model is updated (pressing `p` to set as problematic).

@joshuacwnewton @kousu ?

jcohenadad (commented):
Very cool prototype @Drulex! Few comments:
- I would replace "Problematic" by "QC" so it takes less space (as you probably noticed, when the lateral size of the table is too large, the row split into double row, which is ugly)
- I would display the "Problematic" column by default
- We should write somewhere the keystrokes for pass/fail (and in the future we could have a "shortcuts" button help)
- "j" and "k" don't work for me (macOs Chrome). "l" works though.

Drulex:
>     * I would replace "Problematic" by "QC" so it takes less space (as you probably noticed, when the lateral size of the table is too large, the row split into double row, which is ugly)
>     * We should write somewhere the keystrokes for pass/fail (and in the future we could have a "shortcuts" button help)
:ok_hand: 
 
>     * "j" and "k" don't work for me (macOs Chrome). "l" works though.
Interesting. For me it works on both Firefox and Chrome (Linux).

>     * I would display the "Problematic" column by default
For me Firefox shows all columns by default, but Chrome respects the "hidden" columns. I have to admit that the code handling the show/hiding of columns is quite bizarre. `bootstrap` has native functionality to hide/show columns and much more, it may be worth refactoring the code to leverage the builtin bootstrap table functionalities.

jcohenadad:
Some comments on 
- I would use three values for QC: "empty" (default), ✅  and ❌ . If we choose "false" by default, then the user does not know which image was already QCed or not. 
- I would use another key than "p" for the QC, because the way we do it quickly is with two hands: right hand goes up/down and left hand would assign pass/fail. I would propose "v" for verified or "x". 
- We should write somewhere obvious the keystroke for the pass/fail, otherwise people won't know it. 

@sandrinebedard can you please also give some feedback?

sandrinebedard:
> @sandrinebedard can you please also give some feedback?
yes I'm on it!

sandrinebedard:
The feature is really nice!
Here are a few comments I have:
* To me, having the column named "QC" is not very intuitive on what it displays, even more if the default is "empty". Maybe if we pass over it with the cursor, there could be a description that appears, maybe people will be wondering what is the purpose of the column.
* For manual correction in my pipeline, I seperate FILES_SEG and FILES_LABEL in two different yml list because manual segmentations do not depend on SCT, so I was wondering if there could be the option to download it seperatly, or maybe if we have an option to clear the column that would also do the trick (it clears out anyway when we close it). 
* For images with bad data quality, it would be nice to add a parameter for exclusion and another list to download (it may getting complicated and too much to display, I find so many possibilities for this feature 😄 )

> I would use another key than "p" for the QC, because the way we do it quickly is with two hands: right hand goes up/down and left hand would assign pass/fail. I would propose "v" for verified or "x".

I find that "v" and "x" and it are not very ergonomic, considering that the person has to type the keys for each image and there may be a lot of images in the QC (e.g. I have QC with over 650 subjects).
What about "v" and "f" ? f for fail would still be intuitive.

Other than that, I agree with  @jcohenadad previous suggestions!

jcohenadad:
> find that "v" and "x" and it are not very ergonomic, considering that the person has to type the keys for each image and there may be a lot of images in the QC (e.g. I have QC with over 650 subjects).
What about "v" and "f" ? f for fail would still be intuitive.

ah sorry my comment wasn’t clear: I meant either v or f. The idea being that the user would press once for pass and twice for fail (and the 3rd time would restart the cycle). So there only one keystroke to use.

> For images with bad data quality, it would be nice to add a parameter for exclusion and another list to download (it may getting complicated and too much to display, I find so many possibilities for this feature 😄 )

good idea. How about the 3rd keystroke would bring ☠️, meaning “data not usable”.

kousu:
I'm going to see if I can fix up the final little glitches.

spinalcordtoolbox/reports/assets/_assets/js/filesaver.min.js
```diff
@@ -0,0 +1 @@
...d=g,j.rel="noopener","string"==typeof b?(j.href=b,j.origin===location.origin?e(j):d(j.href)?c(b,g,h):e(j,j.target="_blank")):(j.href=i.createObjectURL(b),setTimeout(function(){i.revokeObjectURL(j.href)},4E4),setTimeout(function(){e(j)},0))}:"msSaveOrOpenBlob"in navigator?function(f,g,h){if(g=g||f.name||"download","string"!=typeof f)navigator.msSaveOrOpenBlob(b(f,h),g);else if(d(f))c(f,g,h);else{var i=document.createElement("a");i.href=f,i.target="_blank",setTimeout(function(){e(i)})}}:function(b,d,e,g){if(g=g||open("","_blank"),g&&(g.document.title=g.document.body.innerText="downloading..."),"string"==typeof b)return c(b,d,e);var h="application/octet-stream"===b.type,i=/constructor/i.test(f.HTMLElement)||f.safari,j=/CriOS\/[\d]+/.test(navigator.userAgent);if((j||h&&i||a)&&"undefined"!=typeof FileReader){var k=new FileReader;k.onloadend=function(){var a=k.result;a=j?a:a.replace(/^data:[^;]*;/,"data:attachment/file;"),g?g.location.href=a:location=a,g=null},k.readAsDataURL(b)}else{var l=f.URL||f.webkitURL,m=l.createObjectURL(b);g?g.location=m:location.href=m,g=null,setTimeout(function(){l.revokeObjectURL(m)},4E4)}});f.saveAs=g.saveAs=g,"undefined"!=typeof module&&(module.exports=g)});
```
kousu, line 1: This file is under MIT: https://github.com/eligrey/FileSaver.js/blob/master/LICENSE.md

so we need to copy that license file into our own LICENSE.md file and annotate it with what files we took.

spinalcordtoolbox/reports/assets/_assets/js/yaml.min.js
```diff
@@ -0,0 +1 @@
...tils.getStringFromFile(path,function(_this){return function(input){var result;result=null;if(input!=null){result=_this.parse(input,exceptionOnInvalidType,objectDecoder)}callback(result)}}(this))}else{input=Utils.getStringFromFile(path);if(input!=null){return this.parse(input,exceptionOnInvalidType,objectDecoder)}return null}};Yaml.dump=function(input,inline,indent,exceptionOnInvalidType,objectEncoder){var yaml;if(inline==null){inline=2}if(indent==null){indent=4}if(exceptionOnInvalidType==null){exceptionOnInvalidType=false}if(objectEncoder==null){objectEncoder=null}yaml=new Dumper;yaml.indentation=indent;return yaml.dump(input,inline,0,exceptionOnInvalidType,objectEncoder)};Yaml.stringify=function(input,inline,indent,exceptionOnInvalidType,objectEncoder){return this.dump(input,inline,indent,exceptionOnInvalidType,objectEncoder)};Yaml.load=function(path,callback,exceptionOnInvalidType,objectDecoder){return this.parseFile(path,callback,exceptionOnInvalidType,objectDecoder)};return Yaml}();if(typeof window!=="undefined"&&window!==null){window.YAML=Yaml}if(typeof window==="undefined"||window===null){this.YAML=Yaml}module.exports=Yaml},{"./Dumper":1,"./Parser":7,"./Utils":10}]},{},[11]);
```
kousu, line 1: Ditto for this file https://github.com/jeremyfa/yaml.js/blob/develop/LICENSE

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -262,7 +268,55 @@
...
+            "sct_straighten_spinalcord": "FILES_SEG",
+            "sct_warp_template": "FILES_SEG",
+        }
+        sct_data.forEach(item => {
+            if (item.qc === true) {
+                let command = item.command;
+                if (command in seg_cmd_to_seg_type) {
+                    segtype = seg_cmd_to_seg_type[command];
+                } else {
+                    segtype = "FIXME";
```
kousu, line 300: TODO

kousu:
Hey y'all, it would really help in the future if you could include the commands you use to test a feature. It took me over a day of procrastination to decide I probably had to do:

```
cd spinalcordtoolbox
./batch_processing.sh
xdg-open qc_example_data/index.html
```


And since batch_processing.sh takes 15 minutes to run, to actually iterate on this I picked out one of the rows in the table and the command, then I reverse-engineered what it takes to actually use that outside of batch_processing to:

```
(venv_sct) $ time sct_deepseg_sc -c t2s -i sct_example_data/t2s/t2s.nii.gz -qc /home/kousu/src/neuropoly/spinalcordtoolbox/qc_example_data
```

I should be able to fix up the keybinds and add those icons now.

jcohenadad:
sorry @kousu, indeed we should have documented that better-- note that you can test it _even_ faster by picking a file from sct_testing_data (smaller files, hence faster processing) instead of sct_example_data.

kousu:
That's a great tip, I'll make use of that for sure.

spinalcordtoolbox/reports/assets/_assets/js/filesaver.min.js
joshuacwnewton: > so we need to copy that license file into our own LICENSE.md file and annotate it with what files we took.

I was wondering if there was a different way to reproduce the license, and I found [this SE answer](https://softwareengineering.stackexchange.com/a/377850) that says:

> The default assumption in copyright law is that a copyright license only applies to whatever is placed explicitly under that license.
> 
>  * A license notice at the top (or bottom) of a source file applies only to the contents of that file
>  * A license notice at a common location in a project applies only to the files of that project

So, I'm wondering if it might be preferable to put the license inside the source file themselves. Especially since we're running QC locally (rather than serving these files on the web), so minimizing file size isn't the biggest concern (going from 2.7 kB to 3.8 kB seems like a nonissue). There's also precedence re: [preserving licenses in minified files](https://softwareengineering.stackexchange.com/a/45170).

joshuacwnewton:
I've updated the TODO list in the PR description to summarize the new suggestions. 

I've also touched base with @kousu and @Drulex on Slack, and after chatting with them I'm going to try and work on this PR. :slightly_smiling_face:

spinalcordtoolbox/reports/assets/_assets/js/filesaver.min.js
joshuacwnewton: Done in 1150567, but let me know if there are any objections.

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: I've updated this to the following:

https://github.com/neuropoly/spinalcordtoolbox/blob/37fe602eaa0eceedee309c1d053cf47bccd49e9b/spinalcordtoolbox/reports/assets/index.html#L301

It's not the prettiest, but we should only hit this case if a new QC command is added and `seg_cmd_to_seg_type` isn't updated, so making it stand out is probably a good thing.

Writing that out, though, `seg_cmd_to_seg_type` feels error-prone; contributors have to remember to update an html doc when writing Python QC code, which I have a feeling we'll miss one day in the future. I wonder if it's possible to store the seg type in a column, and associate it with the Python code instead.

joshuacwnewton:
I've tried giving the functional changes a shot. See the updated TODO checklist in https://github.com/neuropoly/spinalcordtoolbox/pull/3253#issue-580973236 to see what I've done so far, but the main decisions I made were:

* Use 'f' key for marking images. (Because of "home row" typing, 'f' and 'j' often have tactile nubs on keyboards, so alternating between f/j for performing QC felt really smooth.)
* Use :exclamation: for data quality rather than :skull_and_crossbones:. (Considering we're working with medical imaging data of people, using a skull and crossbones felt... a little iffy. :sweat_smile: )
* Use :heavy_check_mark: / :x: for the column title. (Symbols felt a little more intuitive than words, but there's still the issue of making the QC workflow itself discoverable...)

But, where I'm really getting tripped up is with the layout design. HTML/CSS/Bootstrap/JS is really outside my wheelhouse, and I've fumbling around unsuccessfully trying to get elements to behave. :sweat: 

Screenshot from 2021-04-16 12-36-50

Even something as simple as fixing these overlapping elements is eating a lot of my time. I wanted to have a second dropdown button, but I couldn't for the life of me get it to behave (despite puzzling over this [Dropdowns - Bootstrap](https://getbootstrap.com/docs/4.0/components/dropdowns/#disabled-menu-items) page for a while). So, it might be better if someone more skilled in UI/webdev handled the remaining work...

jcohenadad:
@joshuacwnewton i tested the latest version and it is very nice. 

> Even something as simple as fixing these overlapping elements is eating a lot of my time. I wanted to have a second dropdown button, but I couldn't for the life of me get it to behave (despite puzzling over this Dropdowns - Bootstrap page for a while). So, it might be better if someone more skilled in UI/webdev handled the remaining work...

Maybe @TommyBoshkovski could help again?  😁  🙇

I also thought of other cosmetic features:
- move the download buttons at the bottom
- clarify keyboard usage by centralizing them at the same location
- use sexy keyboard buttons
- move the pass/fail to the right (the eyes need to be close to the image)

See below the suggested changes (which I did with an image editor):
Screen Shot 2021-04-16 at 9 02 45 PM

sandrinebedard:
I tested the new version, thanks @joshuacwnewton for the changes! It is getting really nice!

Maybe a few suggestions, if we could increase the size of ✔️ / ❌ in the column it would be great and quicker to see!
I also noticed that when downloading the list for images to exclude  "❗",  the name is also qc.yml, we should change that, it will get confusing between both downloaded lists!

jcohenadad:
@TommyBoshkovski said he will help with the HTML/CSS/Bootstrap/JS stuff 🎉

TommyBoshkovski:
@jcohenadad I tried to address all of your suggestions.  Hope It looks better now.

jcohenadad:
> @jcohenadad I tried to address all of your suggestions. Hope It looks better now.

this is truly awesome 🏅  Thank you so much @TommyBoshkovski. I'll try to find other keys with which background-- it will look nicer

jcohenadad:
Hum... I wanted to change the design of the pass/fail/warning with these: 
```
      const heavy_check_mark = '✅ '
      const heavy_ballot_x = '❌'
      const heavy_excl_mark = '⚠️'
```
But when I do that, the symbols are not updated on the HTML table. I need to click on the arrow that sorts the table to have them appear. Funnily, when I _only_ change the first line with ✅ , then it works as expected. It's only when I put the cross and the warning that this bug occurs. @TommyBoshkovski any idea?

TommyBoshkovski:
It should be fixed now.

jcohenadad:
> It should be fixed now.

Indeed! thank you so much!!!

jcohenadad:
i can fix https://github.com/neuropoly/spinalcordtoolbox/pull/3253#issuecomment-821887235

TommyBoshkovski:
Sure, should be fixed now.

jcohenadad (approved):
Amazing contribution from @Drulex @joshuacwnewton and @TommyBoshkovski 🙏

jcohenadad:
LGTM. @sandrinebedard do you want to give it a last pass before we merge?

sandrinebedard:
Wow, it's truly amazing 🎉

I noticed one thing, when the QC column is empty when opening the QC report, you can download both yml list but they will be empty. However, when you start filling the column, now they are only available to download if there are ❌ or ⚠️, it shouldn't be avaible for downloading at the start I think. It is a minor detail!

Something weird also happended one time, when changing ✅ / ❌/ ⚠️ for an image, it changed for another:


https://user-images.githubusercontent.com/71230552/115148220-9a242f80-a02c-11eb-8c83-1e8bdd40efba.mp4

I refreshed the html and it was fine and I wasn't able to reproduce again, but I still wanted to share!

jcohenadad:
thank you for discovering that bug @sandrinebedard , this is something we should definitely fix before merging. @TommyBoshkovski can we please ask you this favour again?

TommyBoshkovski:
Hey @jcohenadad, I've disabled the buttons at init. @sandrinebedard thanks for testing it and reporting this issue. Regarding the issue that you experienced while changing the QC status of images, I think the behaviour is ok since you've been sorting according to the QC, so the table is automatically sorted while changing the status.

sandrinebedard:
> Regarding the issue that you experienced while changing the QC status of images, I think the behaviour is ok since you've been sorting according to the QC, so the table is automatically sorted while changing the status.

You are absolutely right @TommyBoshkovski! I tested it out again with the last changes and LGTM!

jcohenadad:
Oh boy... we are running into [this distros-specific QC issue](https://github.com/neuropoly/spinalcordtoolbox/pull/3253/checks?check_run_id=2375768380#step:5:1300) now 😢  

Anybody has an idea what could be the cause (and solution)? Is there any centOs package that needs to be installed with the fancy ASCII chars we've introduced? @TommyBoshkovski @joshuacwnewton @kousu

joshuacwnewton:
> Oh boy... we are running into [this distros-specific QC issue](https://github.com/neuropoly/spinalcordtoolbox/pull/3253/checks?check_run_id=2375768380#step:5:1300) now cry
> 
> Anybody has an idea what could be the cause (and solution)? Is there any centOs package that needs to be installed with the fancy ASCII chars we've introduced? @TommyBoshkovski @joshuacwnewton @kousu

I haven't looked too deeply at this yet, but just as a quick surface-level observation: When I first introduced the unicode characters for checks/crosses, the CI didn't have this error (e.g. here's the last passing run: https://github.com/neuropoly/spinalcordtoolbox/runs/2364089577)

The first commit with the CI error was https://github.com/neuropoly/spinalcordtoolbox/pull/3253/commits/d9163cc79135d23dfbbfffa08b23d8d9d9fc3be3, so those changes are worth focusing on.

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -241,9 +246,19 @@
          for the case where the table should be empty (i.e. when search
          results correspond to nothing).
          */
+        b=document.createElement("div")
+        b.className = "fixed-table-toolbar";
+        b.style.paddingTop="10px";
+        b.innerHTML="<button class=\"btn btn-default\" type=\"button\" id=\"download_yaml_btn_274c\" onclick=\"download_yaml('❌')\" disabled>Download \"❌\"</button><button class=\"btn btn-default\" type=\"button\" id=\"download_yaml_btn_26a0\" onclick=\"download_yaml('⚠️')\" disabled>Download \"⚠️\"</button>"
```
joshuacwnewton, line 252: From https://github.com/neuropoly/spinalcordtoolbox/commit/d9163cc79135d23dfbbfffa08b23d8d9d9fc3be3, this line is what stands out to me. 

I'm not sure how `.innerHTML` is parsed, but I suspect it's handled differently (compared to our usage of the unicode characters elsewhere).

joshuacwnewton:
Please disregard my previous comment.

I've found this excellent guide to how Python 3 handles file decoding which helped with my understanding of the problem: http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html. The key point is this one:

> The `locale.getpreferredencoding()` call reports the encoding that Python will use by default for most operations that require an encoding (e.g. reading in a text file without a specified encoding). This is designed to aid interoperability between Python and the host operating system, but can cause problems with interoperability between systems (if encoding issues are not managed consistently).

This behavior is combined with the fact that we don't explicitly specify the encoding type here:

https://github.com/neuropoly/spinalcordtoolbox/blob/0211c4adeb951b3bc4d3e44d9e560b4710ef5e1d/spinalcordtoolbox/reports/qc.py#L515-L517

In our case, though, our situation falls under ["Files in a consistent, known encoding"](http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html#files-in-a-consistent-known-encoding) from the aforementioned guide. So, I think we can go ahead and make this change:

```diff
- with io.open(os.path.join(assets_path, 'index.html')) as template_index: 
+ with io.open(os.path.join(assets_path, 'index.html'), encoding="utf-8") as template_index: 
```

I'm going to give this a shot.

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: Please disregard this comment -- I believe the problem is due to https://github.com/neuropoly/spinalcordtoolbox/pull/3253#issuecomment-822719480.

spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -34,24 +34,45 @@ $(document).ready(function(){
 
   $('html').keydown( function(evt) {
     var obj = $('#table tr.active');
-    // Arrow down: next subject
-    if (evt.which == 40) {
-      if (obj.length == 0 || obj.text() === "DateDatasetSubjectPathFileContrastFunctionFunction+Args") {
+    // Arrow down: next subject (or j)
+    if (evt.which == 40 || evt.which == 74) {
+      if (obj.length == 0 || obj.text() === "QCDateDatasetSubjectPathFileContrastFunctionFunction+Args") {
```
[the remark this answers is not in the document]
joshuacwnewton: I wonder if there's a better way to do this, though...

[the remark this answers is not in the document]
jcohenadad: note: "j" is not really used, in case this is a problem (we use the arrow instead)

jcohenadad:
I've tried the latest commit 55d7d78b6357d6c2b61b6ab991db2a54aae08f7c and the output QC looks good on my end:
Screen Shot 2021-04-19 at 9 31 16 PM

@joshuacwnewton do you still observe the very high columns on your end?

joshuacwnewton:
> @joshuacwnewton do you still observe the very high rows on your end?

I do. I also figured out that `f` not working was actually the refresh problem we had encountered before. (When I re-sort the rows, the changes appear.)

<details>
<summary> Output of sct_check_dependencies </summary>

```console
> sct_check_dependencies

--
Spinal Cord Toolbox (git-aj/3187-interactive-qc-assesment-55d7d78b6357d6c2b61b6ab991db2a54aae08f7c)

sct_check_dependencies 
--

SCT info:
- version: git-aj/3187-interactive-qc-assesment-55d7d78b6357d6c2b61b6ab991db2a54aae08f7c
- path: /home/joshua/repos/spinalcordtoolbox
OS: linux (Linux-5.8.0-50-generic-x86_64-with-debian-bullseye-sid)
CPU cores: Available: 8, Used by ITK functions: 8
RAM: Total: 15863MB, Used: 5246MB, Available: 9449MB
  Using bundled python 3.6.13 |Anaconda, Inc.| (default, Feb 23 2021, 21:15:04) 
[GCC 7.3.0] at /home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/bin/python
Check if dipy is installed..........................[OK] (1.4.0)
Check if ivadomed is installed......................[OK] (2.7.4)
Check if matplotlib is installed....................[OK] (3.3.4)
Check if numpy is installed.........................[OK] (1.19.2)
Check if pandas is installed........................[OK] (1.1.5)
Check if psutil is installed........................[OK] (5.8.0)
Check if pytest is installed........................[OK] (6.2.3)
Check if pytest-cov is installed....................[OK] (__version__ = '2.11.1')
Check if requests is installed......................[OK] (2.25.1)
Check if scikit-learn is installed..................[OK] (0.24.1)
Check if tqdm is installed..........................[OK] (4.60.0)
Check if urllib3 is installed.......................[OK] (1.26.4)
Check if wquantiles is installed....................[OK] (0.4)
```
</details>

joshuacwnewton:
> Let me check something though... this might be a problem on my end, because I was developing at the same time as @TommyBoshkovski -- perhaps when I `git pull` I'm missing an upstream commit somehow.

This doesn't seem to be the case. (I deleted the local branch, checked out the remote again, same issues.)

Screenshot from 2021-04-19 22-36-17

joshuacwnewton:
Ah. I tried using Chrome and everything works as expected. :smile:

jcohenadad:
> Ah. I tried using Chrome and everything works as expected. 😄

hum... should we advise to use chrome? (i can also check with safari).

joshuacwnewton:
> hum... should we advise to use chrome? 

That seems worth avoiding if we can help it! My hunch is that, layout-wise, what we are doing with QC is simple enough that it should work on any browser. So, I think we should first try to reproduce using Firefox on another machine, and then depending on that, try to identify/fix the problem. 

> (i can also check with safari).

Thanks! :+1:

spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -34,24 +34,45 @@ $(document).ready(function(){
...
+      const heavy_excl_mark = '\u26A0\uFE0F'
+      sct_data[index].qc = (
+          sct_data[index].qc === heavy_check_mark
+          ? heavy_ballot_x
+          : sct_data[index].qc === heavy_ballot_x
+            ? heavy_excl_mark
+            : heavy_check_mark
+      );
+      set_download_yml_btn_state(heavy_excl_mark);
+      set_download_yml_btn_state(heavy_ballot_x);
```
joshuacwnewton, line 69: These calls fail whenever I use `f`.

Screenshot from 2021-04-20 09-32-09

This error message occurs when `set_download_yml_btn_state` can't find an element with the requested ID. So, for me, on Firefox, the buttons are indeed not being created.

(This would also explain why the table isn't being refreshed for me.)

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -241,9 +246,37 @@
...
+        btn2.disabled="disabled";
+        btn2.innerText="Download QC Artifacts";
+        b.appendChild(btn1);
+        b.appendChild(btn2);
+        
+        a = document.getElementsByClassName("bootstrap-table");
+        a[0].appendChild(b);
+        c = a[0].children[2];
+        a[0].children[2] = b;
+        a[0].children[3] = c;
```
joshuacwnewton, line 275: This section of code fails for me:

Screenshot from 2021-04-20 09-33-50

I set a breakpoint and took a look at `a`, and it's returning an empty collection:

Screenshot from 2021-04-20 09-35-25

So, the div element `b` never gets added, hence explaining the other failure.

joshuacwnewton (commented):
I'm trying to use my browser's html/js debugger to inspect the behavior during button presses, etc. and detect any errors. I think I've identified the cause of the strange behavior (https://github.com/neuropoly/spinalcordtoolbox/pull/3253#discussion_r616690188), but I don't currently understand the browser differences.

spinalcordtoolbox/reports/assets/_assets/js/main.js
joshuacwnewton: The reason why the buttons don't exist is because of https://github.com/neuropoly/spinalcordtoolbox/pull/3253#discussion_r616690188.

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: Elsewhere, we identify the table using `"table"` rather than `"bootstrap-table"`:

https://github.com/neuropoly/spinalcordtoolbox/blob/55d7d78b6357d6c2b61b6ab991db2a54aae08f7c/spinalcordtoolbox/reports/assets/_assets/js/main.js#L73-L74

So, I'm going to try using that instead and see if it fixes things.

jcohenadad: What is strange is that you didn't have that issue before we were making all these changes in this PR. @TommyBoshkovski do you have any clue?

TommyBoshkovski: Yeah, there is always an issue with Mozilla since they don't always have the same html/css attributes. I will look into the issue and will try to make it Mozilla compatible.

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -20,6 +20,11 @@
       document.getElementById("mosaic-container").style.marginBottom=image_height+"px";
     }
   </script>
+  <style>
+      .fixed-table-container{
+          height: fit-content;
+      }
+  </style>
```
joshuacwnewton, line 27: Aha. I think the issue is because `fixed-table-container:: height` is defined in two separate places. Right here, but also here:

https://github.com/neuropoly/spinalcordtoolbox/blob/0211c4adeb951b3bc4d3e44d9e560b4710ef5e1d/spinalcordtoolbox/reports/assets/_assets/css/style.css#L71-L74

I've modified `style.css` and my page loads almost as expected:

Screenshot from 2021-04-20 10-26-16

The table styling is still a little odd, though. The borders are missing, and when I first highlight the image row, the header row also gets highlighted.

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: @TommyBoshkovski I've been looking into this (sorry!) but could you please check https://github.com/neuropoly/spinalcordtoolbox/pull/3253#discussion_r616737845?

jcohenadad:
OK, so I gave it a try with Safari and it doesn't behave as expected. First thing is that all columns are displayed, regardless of the dropdown menu:

Screen Shot 2021-04-20 at 10 23 30 AM

Screen Shot 2021-04-20 at 10 23 47 AM

My suggestion would be to say upfront that "This page is best viewed with Google Chrome". Maybe we can add a "Help" next to the "About", that will make a pop up window and indicate the thing about browser compat and also have a blurb about what this QC report is about (how to use it, etc.)

joshuacwnewton:
> OK, so I gave it a try with Safari and it doesn't behave as expected. First thing is that all columns are displayed, regardless of the dropdown menu:

I encountered this problem too on Firefox, but it was fixed by increasing the timeout. I can push changes -- if you could wait 1m then test again that would be great.

spinalcordtoolbox/reports/assets/index.html
TommyBoshkovski: Usually, when the page is loaded, the browser takes the last definition of the style for a specific element (since it loads everything sequentially).

joshuacwnewton: That makes sense. I wonder why changing it fixed things for me, then. :thinking:

TommyBoshkovski:
The issue should be fixed now. I tested it on Safari, Firefox, Chrome and it was having expected behaviour.

joshuacwnewton:
Can confirm, the table looks perfect on my end, too. :slightly_smiling_face: 

Screenshot from 2021-04-20 15-05-26

joshuacwnewton:
> Change the name "About" (top right) to "SCT Home" and redirect to RTD page (currently it links to GH).

Just noticed this in the issue description, not sure when it was added but I'll do that now.

jcohenadad:
> Just noticed this in the issue description, not sure when it was added but I'll do that now.

it was me i added it yesterday 😊

jcohenadad (approved):
Everything looks perfect on my end too (chrome, safari). Thank you SOOOOO much @TommyBoshkovski 🙏  You're the best 🏅

spinalcordtoolbox/reports/assets/index.html
```diff
@@ -262,7 +290,63 @@
         }
     }
 
-    <!-- Toggle overlay -->
+    function download_yaml(marker) {
+        let res = {};
+        // TODO: Coordinate with PR #3320 to add entries for sct_dmri_moco/sct_fmri_moco.
```
joshuacwnewton, line 295: Now that #3320 has been merged, I'll quickly address this TODO before this is merged. :slightly_smiling_face:

sandrinebedard:
I tested out on various browsers (chrome, safari, edge and firefox) and everything worked perfectly!

One last minor thing, the QC is not checked in the *Display Columns* even if is is displayed by default:
image

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: Fixed by 86ec6a2. 

I chose to use a new image type marker, `FILES_TIMESERIES`, because the dmri/fmri images are 4D and the QC displays gifs (which is different than any other type of QC report).

joshuacwnewton:
@sandrinebedard 88ed9e8 should add the missing check that you noticed. :slightly_smiling_face:

joshuacwnewton:
@jcohenadad Out of curiosity, I ran `batch_processing.sh` and exported `qc_fail.yml` with every image set to failure (just to see what it would look like):

```yml
FILES_SEG:
    - t2.nii.gz
    - t2.nii.gz
    - t2_seg.nii.gz
    - t2s.nii.gz
    - t1.nii.gz
    - mt1_crop.nii.gz
    - mt0_reg.nii.gz
    - mt1_crop.nii.gz
    - t1w_reg.nii.gz
    - dmri_crop_moco_dwi_mean.nii.gz
    - PAM50_t1_reg.nii.gz
    - dmri_crop_moco_dwi_mean.nii.gz
    - PAM50_t2_reg.nii.gz
FILES_LABEL:
    - t2.nii.gz
FILES_GMSEG:
    - t2s.nii.gz
```
Currently, we're exporting just the filenames. But, this can be ambiguous when files are labeled identically. And, that's not uncommon, since often subjects are identified using the folder name, right?

So, perhaps we want to change each entry to include the full path (which should uniquely identify each subject).

jcohenadad:
About https://github.com/neuropoly/spinalcordtoolbox/pull/3253#issuecomment-824274695: This is an excellent remark @joshuacwnewton . I also thought about this when trying the QC report on the testing datasets. 

Originally, the QC report was created with in mind the processing or large BIDS databases. In this case, the subject ID (and/or site) is encoded in the file name. Example [here](https://spine-generic.readthedocs.io/en/latest/analysis-pipeline.html#segmentation-and-vertebral-labeling). However, I completely agree with you that if the subject ID is not encoded (which is actually the case for _many_ datasets from our users), then this will become a BIG problem, and in that case full path is indeed the right thing to do 😊 . So +1 on this idea.

sandrinebedard:
> Do you think this ambiguity is a relevant concern for typical QC workflows

For BIDS datasets, it won't be a problem, but otherwise it probably will. We proposed that list format because we have a procedure for manual correction using that exact list format, (before this feature, we copy/paste the filename manually in the list). 

I find that for BIDS dataset it will be a bit heavier (it is more a problem on my side because of the size of the dataset that I am using/ other people are generating the lists and getting them back to me/ I manually check and add all images with artefacts 😬, but I can just modify the list or adjust to it afterward).

I defintly understand the purpose and agree wih the idea to add the full path, good catch @joshuacwnewton  😊

joshuacwnewton:
> We proposed that list format because we have a procedure for manual correction using that exact list format, (before this feature, we copy/paste the filename manually in the list).

Ahhh, so this `.yml` file is meant to be used directly with `sg_manual_correction` in `spine-generic`. I did not have that context before, so thank you for the feedback. :slightly_smiling_face: 

Looking at how [`manual_correction` handles `.yml` config files](https://github.com/spine-generic/spine-generic/blob/6d7a6a2ef1695a3994e472aac2981dab77b0ddfd/spinegeneric/cli/manual_correction.py#L183), using the full path wouldn't work [because of how the filenames are parsed](https://github.com/spine-generic/spine-generic/blob/6d7a6a2ef1695a3994e472aac2981dab77b0ddfd/spinegeneric/utils.py#L125). Hrm... this is a bit tricky.

joshuacwnewton:
I've switched from filename to absolute path in ea76c49. 

I've also summarized the YAML compatibility issues with spine-generic in https://github.com/spine-generic/spine-generic/issues/248 (to possibly be addressed in the future).

Does anyone else have any other thoughts / review comments to make before this PR should be merged?

sandrinebedard:
> I've also summarized the YAML compatibility issues with spine-generic in [spine-generic/spine-generic#248](https://github.com/spine-generic/spine-generic/issues/248) (to possibly be addressed in the future).

I can open an issue for [uk-biobank](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa) which I am working on right now, the compatibility issue will also be a problem here, I am using a similar script from the on in spine-generic [manual_correction.py](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/master/pipeline_ukbiobank/cli/manual_correction.py)

>Does anyone else have any other thoughts / review comments to make before this PR should be merged?

@joshuacwnewton Everything is good on my side!

joshuacwnewton:
> I can open an issue for [uk-biobank](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa) which I am working on right now, the compatibility issue will also be a problem here, I am using a similar script from the on in spine-generic [manual_correction.py](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/master/pipeline_ukbiobank/cli/manual_correction.py)

Oh! It's very interesting to see that similar functionality is duplicated across multiple projects. (From what I can tell, the BIDS file-checking utilities are also duplicated?) 

This makes me wonder if the manual correction code should be part of SCT. (Especially because the manual correction scripts mainly call SCT functions anyway -- it seems like it might be in "SCT's domain", so to speak.)

But, I'm getting on a tangent here... This is out of scope for this PR. :sweat_smile:

joshuacwnewton (approved):
Giving my own approval (because I previously requested changes) so that this can be merged.

Issue #3174, opened 2021-01-16
Systematically check for sform/qform mismatch throughout the code


jcohenadad:
After checking for sform/qform mismatch in `sct_straighten_spinalcord` #2858, the check could be extended to all SCT code. 

More specifically, the field `check_sform` in `spinalcordtoolbox.image.Image()` can be set to `True`. More context in https://github.com/neuropoly/spinalcordtoolbox/pull/2858#issuecomment-761292976

joshuacwnewton:
A barrier to this has been pointed out by @cfhammill in https://github.com/neuropoly/spinalcordtoolbox/issues/3097#issuecomment-745655648.

> The real issue is that we appear to have at least one program in sct that uses the sform, so our plan to require sform = qform is made challenging by the fact that ANTs outputs don't satisfying this requirement

joshuacwnewton:
Once we replace calls to the ANTs binary with `sct_apply_transfo` (see #3272), the sform will be set in more places, so it's possible this `check_sform` can be enabled by default. I'll test it out in that PR.

Issue #3097, opened 2020-12-15
ANTs based tools don't set sform


cfhammill:
Related to #2858 it appears that ANTs based programs fail to set the sform field in the NIfTI headers. For the first example we'll look at `sct_register_multimodal` which uses ANTs under the hood.

```
sct_download_data -d course_london19/
cd course_london19/multi_subject/data/sub-01/anat/
sct_register_multimodal \
  -i sub-01_acq-MTon_MTS.nii.gz \
  -d sub-01_acq-MToff_MTS.nii.gz \
  -param step=1,type=im,algo=rigid,slicewise=1,metric=CC \
  -x spline
```

Then in python check that both input images and the transformation have valid sforms

```python
s = Image("sub-01_acq-MTon_MTS.nii.gz")
t = Image("sub-01_acq-MToff_MTS.nii.gz")
x = Image("warp_sub-01_acq-MToff_MTS2sub-01_acq-MTon_MTS.nii.gz")
print(s.hdr.get_sform())
print(t.hdr.get_sform())
print(x.hdr.get_sform())
```

gives

```
[[ 8.98437500e-01  6.13959088e-08  3.44130327e-04 -1.14131424e+02]
 [-1.00533734e-05  8.86628509e-01  8.08007956e-01 -1.22961884e+02]
 [-6.10132338e-05 -1.45188943e-01  4.93427992e+00 -1.73513947e+01]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  1.00000000e+00]]

[[ 8.98437500e-01  6.13959088e-08  3.44130327e-04 -1.14131424e+02]
 [-1.00533734e-05  8.86628509e-01  8.08007956e-01 -1.22961884e+02]
 [-6.10132338e-05 -1.45188943e-01  4.93427992e+00 -1.73513947e+01]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  1.00000000e+00]]

[[0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 0.]
 [0. 0. 0. 1.]]
```

The last of which is not a valid affine matrix, the 1 is added by nibabel, because the sform field holds only 12 float numbers, the last row `0 0 0 1` is added automatically. So this is I believe the output when the sform is left uninitialized.

To demonstrate this issue effects more than just `sct_register_multimodal` we can fix the header and check that it gets fixed.

```python
from spinalcordtoolbox.image import Image
[...]

PR #3217, opened 2021-02-11
Fix size calculation bug for '-method map' and '-method ml'

joshuacwnewton:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [X] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

Both `-method map` and `-method ml` include data from extra labels in their masks:

https://github.com/neuropoly/spinalcordtoolbox/blob/74f068df9042e6cce8473e67f1a3fe60ddadac70/spinalcordtoolbox/aggregate_slicewise.py#L416-L421

This causes the `Size [vox]` calculation to be larger than expected. 

To fix this, we limit the calculation to only include the first label. (Also, remove `.flatten()`, as it is no longer necessary once we index the array.)

### A note about coupling

This change is coupled very closely to code inside [a separate `extract_metric()` function](https://github.com/neuropoly/spinalcordtoolbox/blob/74f068df9042e6cce8473e67f1a3fe60ddadac70/spinalcordtoolbox/aggregate_slicewise.py#L389-L434): I make the assumption that `mask` will always have an extra dimension for labels, and that only label `[0]` is relevant for size calculations, just like in `extract_metric()`. 

While this assumption holds for our internal use of `aggregate_per_slice_or_level()`, this might be a shaky assumption when thinking about the function in isolation (as part of a public API) since we don't do any input validation to verify `mask`. (I'm not sure the API is ready enough to be thinking about this, though) 

## Linked issues

Fixes #3216.

Issue #3216, opened 2021-02-10
The size of the label is wrong when using -method map


jcohenadad:
### Description

When using `sct_extract_metric -method map`, the column "Size [vox]" does not represent the correct size of the region.

This issue was reported by a user on [the forum](https://forum.spinalcordmri.org/t/measuring-involvement-of-corticospinal-tract/626/6?u=jcohenadad).
 
### Steps to Reproduce

```bash
sct_download_data -d sct_testing_data
cd sct_testing_data/mt
sct_extract_metric -i mtr.nii.gz -l 51,52 -perslice 0 -method map -o map.csv
sct_extract_metric -i mtr.nii.gz -l 51,52 -perslice 0 -method wa -o wa.csv
cat map.csv wa.csv
```

The result is:
**map.csv**
|Timestamp          |SCT Version                                        |Filename                                    |Slice (I->S)|VertLevel|Label         |Size [vox]        |MAP()             |STD()             |
|-------------------|---------------------------------------------------|--------------------------------------------|------------|---------|--------------|------------------|------------------|------------------|
|2021-02-10 11:57:30|git-master-92f004872deccb6ef1d1411751f8cffea14d5f51|/Users/julien/sct_testing_data/mt/mtr.nii.gz|0:4         |         |white matter  |331.1098567905528 |34.2645214894471  |11.457341402914249|
|2021-02-10 11:57:30|git-master-92f004872deccb6ef1d1411751f8cffea14d5f51|/Users/julien/sct_testing_data/mt/mtr.nii.gz|0:4         |         |dorsal columns|331.1098567905528 |35.2217919776565  |6.378461578742057 |

**wa.csv**
|Timestamp          |SCT Version                                        |Filename                                    |Slice (I->S)|VertLevel|Label         |Size [vox]        |WA()              |STD()             |
|-------------------|---------------------------------------------------|--------------------------------------------|------------|---------|--------------|------------------|------------------|------------------|
|2021-02-10 11:57:31|git-master-92f004872deccb6ef1d1411751f8cffea14d5f51|/Users/julien/sct_testing_data/mt/mtr.nii.gz|0:4         |         |white matter  |176.1532667399286 |32.640416229403456|11.457341402914249|
[...]

PR #3272, opened 2021-03-10
Replace ANTs binary call with sct_apply_transfo call to properly set sform

### LINKED ISSUE #3254: sform not preserved when running sct_warp_template
### Description

When applying a transformation to bring the PAM50 to the native space of an image using `sct_warp_template`, [the sform field is not preserved](https://forum.spinalcordmri.org/t/images-with-different-orientations-fields-of-view/633/19?u=joshuacwnewton) (all zeros). We should ensure the qform is copied onto the sform for the generated PAM50 files. 

### Steps to reproduce

1. Download and extract the following images: [javier_20210226.zip](https://github.com/neuropoly/spinalcordtoolbox/files/6109654/javier_20210226.zip)

2. Run the following commands:

```
sct_register_to_template -i PARESP_00003__003_eddy_md_crop.nii -s PARESP_00003__003_WholeSpine_Mask2.nii.gz -l PARESP_00003__003_dmri_labels.nii.gz -c t2 -param step=1,type=imseg,algo=centermassrot,rot_method=pcahog:step=2,type=im,algo=syn,metric=cc,iter=5,smooth=0,slicewise=0 -r 0 -qc qc-optimization;
mv warp_template2anat.nii.gz warp_template2dmri.nii.gz
mv warp_anat2template.nii.gz warp_dmri2template.nii.gz
sct_warp_template -d PARESP_00003__003_eddy_md_crop.nii -w warp_template2dmri.nii.gz -a 1 -s 1 -ofolder label
```

3a. (Optional) Open `label/template/PAM50_wm.nii.gz` in FSLeyes (version 0.34.2) to see the message "Displaying images with different orientations/fields of view"

3b. (Optional) Run `flshd label/template/PAM50_wm.nii.gz` to see the following result:

<details>
<summary> Result of <code> fslhd </code> </summary>

```
filename	PAM50_wm.nii.gz

sizeof_hdr	348
data_type	FLOAT64
dim0		3
dim1		21
dim2		51
dim3		147
dim4		1
dim5		1
dim6		1
dim7		1
vox_units	mm
time_units	Unknown
datatype	64
nbyper		8
bitpix		64
pixdim0		-1.000000
pixdim1		1.550000
pixdim2		1.458333
pixdim3		1.458333
pixdim4		0.000000
pixdim5		0.000000
pixdim6		0.000000
pixdim7		0.000000
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	0
freq_dim	0
slice_dim	0
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Unknown
intent_code	0
intent_name	
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	-1.549350 0.010756 -0.040837 24.244717 
qto_xyz:2	0.000000 1.410237 0.371441 -75.029106 
qto_xyz:3	-0.044884 -0.371286 1.409645 -89.249001 
qform_xorient	Right-to-Left
sform_name	Scanner Anat
sform_code	0
sto_xyz:1	0.000000 0.000000 0.000000 0.000000 
sto_xyz:2	0.000000 0.000000 0.000000 0.000000 
sto_xyz:3	0.000000 0.000000 0.000000 0.000000 
sform_xorient	Unknown
sform_yorient	Unknown
sform_zorient	Unknown
file_type	NIFTI-1+
file_code	1
descrip		
aux_file	
```

</details>

joshuacwnewton:
#### GitHub

#### PR contents

## Description

When `isct_antsApplyTransform` is called directly, the sform of the output file is left unset (all zeros). (See #3097 for context.)

`sct_apply_transfo` contains logic to set the sform copied from the reference image. So, swap in that command wherever the ANTs binary was called directly . Here is the new `fslhd` output for the header of the file mentioned in #3254:

```
qform_name      Aligned Anat
qform_code      2
qto_xyz:1       -1.549350 0.010756 -0.040836 24.244717 
qto_xyz:2       0.000000 1.410237 0.371441 -75.029106 
qto_xyz:3       -0.044884 -0.371286 1.409645 -89.249001 
qto_xyz:4       0.000000 0.000000 0.000000 1.000000 
qform_xorient   Right-to-Left
qform_yorient   Posterior-to-Anterior
qform_zorient   Inferior-to-Superior
sform_name      Aligned Anat
sform_code      2
sto_xyz:1       -1.549350 0.010757 -0.040839 24.244717 
sto_xyz:2       0.000000 1.410237 0.371441 -75.029106 
sto_xyz:3       -0.044887 -0.371286 1.409645 -89.249001 
sto_xyz:4       0.000000 0.000000 0.000000 1.000000 
sform_xorient   Right-to-Left
sform_yorient   Posterior-to-Anterior
sform_zorient   Inferior-to-Superior
```

This PR also updates the ground truth for `batch_processing.sh`. I believe the shift in metrics is to be expected because of the bugfix (see https://github.com/neuropoly/spinalcordtoolbox/pull/3272#discussion_r593548511 for reasoning), but I'd love it if someone could vet my logic and make sure I'm being reasonable here.

## Linked issues

Fixes #3254.


joshuacwnewton:
I've marked this as "Draft" because I notice we have direct calls to `isct_antsApplyTransform` in `straightening.py` and `sct_label_vertebrae` too. I want to investigate those files as well to see if the issue is also present there.

joshuacwnewton:
Oh, how exciting! `Test batch_processing.sh` did its job. :smile: 

I'll have to investigate further, but it looks like switching `isct_antsApplyTransform` for `sct_apply_transfo` caused small shifts in 3 of the results.

jcohenadad:
> Oh, how exciting! Test batch_processing.sh did its job. 😄

🚀

spinalcordtoolbox/straightening.py
```diff
@@ -518,31 +520,26 @@ def straighten(self):
...
-                      '-n', 'BSpline[3]'],
-                     is_sct_binary=True,
-                     verbose=verbose)
+            sct_apply_transfo.main(['-i', 'data.nii',
+                                    '-d', fname_ref,
+                                    '-w', 'tmp.curve2straight.nii.gz',
+                                    '-o', 'tmp.anat_rigid_warp.nii.gz',
+                                    '-x', 'spline',
+                                    '-v', str(verbose)])
+
```
joshuacwnewton, line 529: I [reverted the change above](https://github.com/neuropoly/spinalcordtoolbox/commit/b7909506fa6a9f5a72e1227fcd4ec16cf79b7972), and the `batch_processing.sh` [tests pass](https://github.com/neuropoly/spinalcordtoolbox/runs/2088187862). So, I believe `straightening.py` is the cause.

### Investigation

In `batch_processing.sh`, the highlighted call is used in two scripts:

https://github.com/neuropoly/spinalcordtoolbox/blob/0b7be07b7e64a57a303824e85930fbceb8b890f0/batch_processing.sh#L73

https://github.com/neuropoly/spinalcordtoolbox/blob/0b7be07b7e64a57a303824e85930fbceb8b890f0/batch_processing.sh#L79

In both calls, `straight_ref.nii.gz` (which is just a renamed `tmp.anat_rigid_warp.nii.gz`) is used in many subsequent steps (its header gets copied as a reference for other output images). So, I ran the script twice (before/after changes) to compare the headers for each straightened file: https://www.diffchecker.com/GbBSFdZ0

The headers are the only thing that differ (the data was the same). So, my conclusion is that the shift in `batch_processing.sh` values is not a bad thing, because  `straight_ref.nii.gz` has had the wrong headers this whole time. (Same issue as #3254.)

joshuacwnewton: > So, I ran the script twice (before/after changes) to compare the headers for each straightened file: https://www.diffchecker.com/GbBSFdZ0

The only strange thing here is the sform/qform codes for both straightened files. Neither looks quite right to me.

joshuacwnewton:
Out of curiosity, I'm just going to check to see if fixing the `sform` here (in all places) allows us to also fix #3174.

spinalcordtoolbox/straightening.py
jcohenadad: There are suspicious lines under `image.py`, e.g.

https://github.com/neuropoly/spinalcordtoolbox/blob/60ea08c9d664cf403f2982f8f5c8ae1cc5d704a3/spinalcordtoolbox/image.py#L1417

Relevant doc: https://nipy.org/nibabel/nifti_images.html#default-sform-and-qform-codes
image

Analogous problem as in https://github.com/ivadomed/ivadomed/pull/714

Moreover, I see this line: 
https://github.com/neuropoly/spinalcordtoolbox/blob/60ea08c9d664cf403f2982f8f5c8ae1cc5d704a3/spinalcordtoolbox/image.py#L1330

Which, according to the doc, might be problematic if qform/sform are different:
image

joshuacwnewton: Ahh! Thank you for sharing what was learned in https://github.com/ivadomed/ivadomed/pull/714. You've saved me from having to figure this out all over again. :smile:

joshuacwnewton: Perhaps a separate SCT issue should be created for this? It seems this issue with `image.py` has larger implications outside the scope of this PR (which is addressing uninitialized sforms from `isct_antsApplyTransform`).

joshuacwnewton:
Ah, it looks like I made the tolerances too strict on the `batch_processing.sh` tests. 

https://github.com/neuropoly/spinalcordtoolbox/blob/60ea08c9d664cf403f2982f8f5c8ae1cc5d704a3/unit_testing/batch_processing/test_batch_processing.py#L26

I uploaded results from my local Ubuntu machine, but the GitHub Actions runner returned [slightly different results](https://github.com/neuropoly/spinalcordtoolbox/runs/2122666480#step:5:66).

```
E         +0.7755359385435815
E         -0.7755359256476565 ± 7.8e-09
```

jcohenadad:
> I uploaded results from my local Ubuntu machine, but the GitHub Actions runner returned slightly different results.

yeah we can definitely consider these the "same"

spinalcordtoolbox/straightening.py
joshuacwnewton: Issue here: https://github.com/neuropoly/spinalcordtoolbox/issues/3295

unit_testing/batch_processing/cached_results/mt/mtr_in_wm.csv
```diff
@@ -1,2 +1,2 @@
 Timestamp,SCT Version,Filename,Slice (I->S),VertLevel,Label,Size [vox],MAP(),STD()
-2021-01-25 00:48:42,git-HEAD-42ef57b3a6241a127ed0d7a7eebc909dc46c9d07,/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/sct_example_data/mt/mtr.nii.gz,4:17,2:5,white matter,1866.8310447890826,54.38001541207312,4.851424830774496
+2021-03-16 10:47:36,git-jn/3254-fix-sform-sct_warp_template-9566dce5850797d73d819917f7f49ce74b4a052c,/home/joshua/repos/spinalcordtoolbox/sct_example_data/mt/mtr.nii.gz,4:17,2:5,white matter,980.9416908352437,54.37973987894469,4.85153610548257
```
jcohenadad, line 2: from ~1800 to ~900 that’s a suspiciously large change for « size »

joshuacwnewton: Thanks for pointing this out. I'll try to dig a little deeper here. 

(Also relevant: https://github.com/neuropoly/spinalcordtoolbox/issues/3250)

jcohenadad: if it help: what i would do is open up the WM mask used for this quantification on FSLeyes, before/after the change, and see how they differ

joshuacwnewton: Ah! I think this is actually due to a previously fixed bug. Recall: https://github.com/neuropoly/spinalcordtoolbox/issues/3216 and https://github.com/neuropoly/spinalcordtoolbox/pull/3217

jcohenadad: > Ah! I think this is actually due to a bug that was fixed between #3198 and now. (Issue: #3216 and PR: #3217). mtr_in_wm.csv was generated using the following command (note -method map):

ah yes! good catch!

jcohenadad:
@joshuacwnewton my only concern with this PR is the change in processing time for `sct_warp_template -a 1`, which wraps a LOT of images. I remember i called ANTs binary to save time there.

could you please run sct_warp_template on the sct_example_data/t1 with option "-a 1 -s 1" before/after the change and report the time difference? thanks!

joshuacwnewton:
> could you please run sct_warp_template on the sct_example_data/t1 with option "-a 1 -s 1" before/after the change and report the time difference? thanks!

Indeed, there is a large time difference:

```console
# On master
490.86s user 44.56s system 189% cpu 4:43.22 total

# On jn/3254-fix-sform-sct_warp_template
722.87s user 94.99s system 157% cpu 8:39.29 total
```

An issue was opened for the speed problem a few weeks ago: https://github.com/neuropoly/spinalcordtoolbox/issues/3271. I can start looking into this.

spinalcordtoolbox/scripts/sct_warp_template.py
```diff
@@ -117,14 +118,12 @@ def warp_label(path_label, folder_label, file_label, fname_src, fname_transfo, p
...
-                      os.path.join(path_out, folder_label, template_label_file[i]),
-                      get_interp(template_label_file[i], list_labels_nn)),
-                     is_sct_binary=True,
-                     verbose=verbose)
+            sct_apply_transfo.main(['-i', fname_label,
+                                    '-d', fname_src,
+                                    '-w', fname_transfo,
+                                    '-o', os.path.join(path_out, folder_label, template_label_file[i]),
+                                    '-x', get_interp(template_label_file[i], list_labels_nn),
+                                    '-v', str(verbose)])
```
joshuacwnewton, line 126: This change causes a slowdown from ~4.5m to ~8.5m. 

(Highlighting so that it can be linked to.)

joshuacwnewton: @jcohenadad The tl;dr of my investigation in https://github.com/neuropoly/spinalcordtoolbox/issues/3271#issuecomment-806035150 is this:

* `sct_apply_transfo` has some extra logic involving headers, so it loads some `Images` to access the header.
* SCTs' `Image` class loads the data array _immediately_, even if it isn't needed.
    * By comparison, `nibabel.load` waits until the data is explicitly requested (which is much more efficient). 
* This data loading is the source of the extra ~4s per individual call to `sct_apply_transfo`.

jcohenadad: Excellent investigation, and yes, it makes perfect sense. Thank you for opening this other issue. Now that we know the cause we should easily find a workaround to this increased processing time.

jcohenadad (approved):
I haven't tested but now that batch_processing is included in our CI I am less anxious that something major is broken 😅 
Thanks for this fix @joshuacwnewton !

PR #3299, opened 2021-03-17
Added flag -s-template-id to use another segmentation (e.g. white matter) for registration to template

### LINKED ISSUE #3297: Make it possible to use another segmentation, e.g. white matter, for registration
### Context

Currently, the function assumes that the user inputs the cord segmentation with the flag `-s`. SCT then automatically associates `-s` with the file with ID=3 from **info_label.txt** file:

https://github.com/neuropoly/spinalcordtoolbox/blob/6d70ad82fbd9ba7961fc1ef7fbb7ebca70ac6fc7/spinalcordtoolbox/scripts/sct_register_to_template.py#L342

This ID label is supposed to correspond to the spinal cord mask.

However, in some cases where the white matter mask is available, the user might want to use it for registration, as it would provide better registration performance.

In this case, we need to specify somehow that another segmentation ID file should be used.

### Suggestion

Introduce the flag `-s-template-id`, which would indicate which file to select to be associated with `-s`. By default, the value would be "3" to ensure retro-compatibility.

jcohenadad:
#### GitHub

#### PR contents

## Description

Currently, the function assumes that the user inputs the cord segmentation with the flag `-s`. SCT then automatically associates `-s` with the file with ID=3 from **info_label.txt** file:

https://github.com/neuropoly/spinalcordtoolbox/blob/6d70ad82fbd9ba7961fc1ef7fbb7ebca70ac6fc7/spinalcordtoolbox/scripts/sct_register_to_template.py#L342

This ID label is supposed to correspond to the spinal cord mask.

However, in some cases where the white matter mask is available, the user might want to use it for registration, as it would provide better registration performance.

In this case, we need to specify somehow that another segmentation ID file should be used.

In this PR we introduce the flag `-s-template-id`, which indicates which file to select to be associated with `-s`. By default, the value would be "3" to ensure retro-compatibility.

## Linked issues
Fixes #3297

### DIFF
--- a/spinalcordtoolbox/registration/landmarks.py
+++ b/spinalcordtoolbox/registration/landmarks.py
@@ -41,6 +41,7 @@ ini_param_trans_x = 270.0  # pix
 ini_param_trans_y = -150.0  # pix
 initial_step = 2
 
+
 def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', verbose=1):
     """
     Register two NIFTI volumes containing landmarks
@@ -101,6 +102,7 @@ def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', ve
                                                            points_moving_barycenter[2]))
     text_file.close()
 
+    
 def getNeighbors(point, set_points, k=1):
     '''
     Locate most similar neighbours
--- a/spinalcordtoolbox/registration/landmarks.py
+++ b/spinalcordtoolbox/registration/landmarks.py
@@ -41,7 +41,7 @@ ini_param_trans_x = 270.0  # pix
 ini_param_trans_y = -150.0  # pix
 initial_step = 2
 
-def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', verbose=1, path_qc=None):
+def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', verbose=1):
     """
     Register two NIFTI volumes containing landmarks
     :param fname_src: fname of source landmarks
@@ -84,7 +84,7 @@ def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', ve
     # estimate transformation
     # N.B. points_src and points_dest are inverted below, because ITK uses inverted transformation matrices, i.e., src->dest is defined in dest instead of src.
     # (rotation_matrix, translation_array, points_moving_reg, points_moving_barycenter) = getRigidTransformFromLandmarks(points_dest, points_src, constraints=dof, verbose=verbose, path_qc=path_qc)
-    (rotation_matrix, translation_array, points_moving_reg, points_moving_barycenter) = getRigidTransformFromLandmarks(points_src, points_dest, constraints=dof, verbose=verbose, path_qc=path_qc)
+    (rotation_matrix, translation_array, points_moving_reg, points_moving_barycenter) = getRigidTransformFromLandmarks(points_src, points_dest, constraints=dof, verbose=verbose)
     # writing rigid transformation file
     # N.B. x and y dimensions have a negative sign to ensure compatibility between Python and ITK transfo
     text_file = open(fname_affine, 'w')
@@ -200,7 +200,7 @@ def minimize_transform(params, points_dest, points_src, constraints):
     return SSE(np.matrix(points_dest), points_src_reg)
 
 
-def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_Tz_Rx_Ry_Rz', verbose=0, path_qc=None):
+def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_Tz_Rx_Ry_Rz', verbose=0):
     """
     Compute affine transformation to register landmarks
     :param points_src:
@@ -254,11 +254,7 @@ def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_T
     logger.info(f"Center:\n {points_src_barycenter}")
     logger.info(f"Translation:\n {translation_array}")
 
-    if verbose == 2 and path_qc is not None:
-        try:
-            os.makedirs(path_qc)
-        except FileExistsError:
-            pass
+    if verbose == 2:
 
         import matplotlib
         # use Agg to prevent display
@@ -288,14 +284,14 @@ def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_T
         ax.set_aspect('auto')
         plt.legend()
         # plt.show()
-        plt.savefig(os.path.join(path_qc, 'getRigidTransformFromLandmarks_plot.png'))
+        plt.savefig('getRigidTransformFromLandmarks_plot.png')
 
         fig2 = plt.figure()
         plt.plot(sse_results)
         plt.grid()
         plt.title('#Iterations: ' + str(res.nit) + ', #FuncEval: ' + str(res.nfev) + ', Error: ' + str(res.fun))
         plt.show()
-        plt.savefig(os.path.join(path_qc, 'getRigidTransformFromLandmarks_iterations.png'))
+        plt.savefig(os.path.join('getRigidTransformFromLandmarks_iterations.png'))
 
     # transform numpy matrix to list structure because it is easier to handle
     points_src_reg = points_src_reg.tolist()
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -30,6 +30,7 @@ from spinalcordtoolbox.registration.register import *
 from spinalcordtoolbox.registration.landmarks import *
 from spinalcordtoolbox.types import Coordinate
 from spinalcordtoolbox.utils import *
+from spinalcordtoolbox.utils import Metavar
 from spinalcordtoolbox import __data_dir__
 import spinalcordtoolbox.image as msct_image
 import spinalcordtoolbox.labels as sct_labels
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -139,9 +139,9 @@ def get_parser():
         '-s-template-id',
         metavar=Metavar.int,
         type=int,
-        help="Segmentation from the template to associate with the flag '-s'. Typically, the spinal cord segmentation "
-             "is used, but if it is available, the white matter mask would produced better registration results. The ID"
-             "is indicated in the file 'template/info_label.txt' of the template indicated by the flag '-t'.",
+        help="Segmentation from the template to be used by flag '-s'. By default, the spinal cord segmentation "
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -135,6 +135,15 @@ def get_parser():
         action="help",
         help="Show this help message and exit."
     )
+    optional.add_argument(
+        '-s-template-id',
+        metavar=Metavar.int,
+        type=int,
+        help="Segmentation from the template to associate with the flag '-s'. Typically, the spinal cord segmentation "
+             "is used, but if it is available, the white matter mask would produced better registration results. The ID"
```
kousu, line 143: Is this "would" or "could"?

spinalcordtoolbox/registration/landmarks.py
```diff
@@ -288,14 +284,14 @@ def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_T
...
         # plt.show()
-        plt.savefig(os.path.join(path_qc, 'getRigidTransformFromLandmarks_plot.png'))
+        plt.savefig('getRigidTransformFromLandmarks_plot.png')
 
         fig2 = plt.figure()
         plt.plot(sse_results)
         plt.grid()
         plt.title('#Iterations: ' + str(res.nit) + ', #FuncEval: ' + str(res.nfev) + ', Error: ' + str(res.fun))
         plt.show()
-        plt.savefig(os.path.join(path_qc, 'getRigidTransformFromLandmarks_iterations.png'))
```
kousu, line 298: Cool. I like relying on `pwd` much better. That seems nicely unixey.

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -135,6 +135,15 @@ def get_parser():
...
+             "is indicated in the file 'template/info_label.txt' of the template indicated by the flag '-t'.",
+        default=3
```
kousu, line 145: I think this is probably going to drive people crazy. It's a double-layer of indirection, via an integer. Could it at least be an enum instead of an int?

But, looking at the only template people actually use, `data/PAM50/template/` (to remind myself and the future: this only shows up after `./install_sct`; it's source is https://github.com/sct-data/PAM50/releases/download/r20201104/PAM50-r20201104.zip), the options are:


```
# Template labels
# Keyword=IndivLabels (Please DO NOT change this line)
# ID, name, file
0, T1-weighted template, PAM50_t1.nii.gz
1, T2-weighted template, PAM50_t2.nii.gz
2, T2S-weighted template, PAM50_t2s.nii.gz
3, spinal cord mask (binary), PAM50_cord.nii.gz
4, white matter mask (probabilistic), PAM50_wm.nii.gz
5, gray matter mask (probabilistic), PAM50_gm.nii.gz
6, cerebrospinal fluid mask (binary), PAM50_csf.nii.gz
7, spinal cord mask with discrete vertebral levels, PAM50_levels.nii.gz
8, spinal cord mask with continuous vertebral levels, PAM50_levels_continuous.nii.gz
9, point-wise vertebral body labels, PAM50_label_body.nii.gz
10, point-wise intervertebral disc labels, PAM50_label_disc.nii.gz
11, point-wise posterior intervertebral disc labels, PAM50_label_discPosterior.nii.gz
12, spine mask (binary), PAM50_spine.nii.gz
13, spinal cord centerline, PAM50_centerline.nii.gz
14, point-wise spinal level labels, PAM50_label_spinal_levels.nii.gz
```

So, `info_label.txt` is a manifest, or maybe a dictionary. And it already has unique IDs. So instead of taking an int here, what if you took the string "cord", or "wm" or "gm". That way the scripts people run and what they see on disk match up more directly. You can keep `info_label.txt` for the full names, but for the future I'd also think about dropping its first column and swapping the third and second around so it looks like like the key-value dictionary it is.

You'll have to rework (or maybe just drop?) `get_file_label()`, then.

As written, especially with this explanation, it looks like *every* template has the same set of options. But that's not true, is it? If that was true, why ship `info_label.txt` with each?

kousu (commented):
Is there an example I could see of actually running this? Is there something in one of the SCT courses?

I'm missing a lot of context for understanding what selecting different segmentations should do. I don't think I would know what to look for.

spinalcordtoolbox/scripts/sct_register_to_template.py
jcohenadad: good catch-- could is more accurate-- can you pls update the code-- thx

spinalcordtoolbox/scripts/sct_register_to_template.py
jcohenadad: @kousu these are excellent thoughts-- quick context (no time to elaborate more): a few years ago we did use more intuitive naming (sc, wm, etc.) however for reasons i vaguely remember (we need to dig in the past PRs, i don't have time to do that now) this caused some issues because as we were prototyping register to template functionality and adding more contrast (t1, t2s, etc.) it because cumbersome to use naming (eg: there would be a "label" for "vertebral label" and then we introduced spinal label, so "label" was no more specific enough, etc.), so we ended up using numerical IDs. However, i've been unhappy about this choice for the past few years-- i'd like to come back to your suggestion-- if you or someone could please open an issue to get this started that would be great-- for now, this PR needs go get in because users are waiting after us-- they have urgent analyses to perform-- thanks

jcohenadad:
> Is there an example I could see of actually running this? Is there something in one of the SCT courses?
> 
> I'm missing a lot of context for understanding what selecting different segmentations should do. I don't think I would know what to look for.

i'm really sorry but i don't have the time to provide this guidance now-- the gist of it is: selecting the wm mask (instead of the cord mask) would include the cord contour AND the gm contour, so that registration to template can rely on the cord contour AND gm contour-- the typical usage of reg to template ONLY uses the cord contour-- tehre is a mentino on that in the batch processing-- it is currently commented-- @joshuacwnewton opened an issue to actually update this commented part-- but it relates to this--

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -135,6 +135,15 @@ def get_parser():
         action="help",
         help="Show this help message and exit."
     )
+    optional.add_argument(
+        '-s-template-id',
+        metavar=Metavar.int,
+        type=int,
+        help="Segmentation from the template to associate with the flag '-s'. Typically, the spinal cord segmentation "
+             "is used, but if it is available, the white matter mask would produced better registration results. The ID"
+             "is indicated in the file 'template/info_label.txt' of the template indicated by the flag '-t'.",
```
kousu, line 144:
```suggestion
        help="Segmentation from the template to be used by flag '-s'. By default, the spinal cord segmentation "
             "is used, but if available, a different segmentation could produced better registration results. The ID"
             "is an integer indicated in the file 'template/info_label.txt' of the template indicated by the flag '-t'.",
```

jcohenadad: no that's not it. "-s" is the seg from the subject, while this new flag "-s-template-id" is the seg from the template. that's why i used the term "associate with"-- could you suggest something clearer maybe? if not we stick with associate

joshuacwnewton:
Just commenting to mention that I'd like to review this today/tomorrow. (A personal issue has come up, which is why I haven't reviewed this yet, but I did see it and skim over it.)

kousu (commented):
I want to help you out but I have realized I don't feel qualified to to test it. I don't know how what datasets I would use or what to expect results to look like.

I'm going to leave this with @joshuacwnewton, who reminded me that this should be a feature bump: 5.2.1 -> 5.3.0.

kousu:
I did appease the linter for you though so you don't need to worry about that.

spinalcordtoolbox/scripts/sct_register_to_template.py
kousu: #3303 

I also edited your explanation so that it doesn't talk about white matter, because each template has many or few segmentations and there's no reason any of them have to be white matter or that the user would want to use that one. Unless this is somewhere where my knowledge is failing me.

jcohenadad: thank you @kousu, actually i think i will put back WM, because in 99% of use cases, ppl have the cord seg, in 20% of the cases they have cord and WM, and in <1% of the cases they have another seg (eg: MS lesion, spine seg, etc.).

jcohenadad: thank you @kousu, actually i think i will put back WM, because in 99% of use cases, ppl have the cord seg, in 20% of the cases they have cord and WM, and in <1% of the cases they have another seg (eg: MS lesion, spine seg, etc.).

spinalcordtoolbox/registration/landmarks.py
```diff
@@ -288,14 +286,14 @@ def getRigidTransformFromLandmarks(points_dest, points_src, constraints='Tx_Ty_T
...
+        plt.savefig(os.path.join('getRigidTransformFromLandmarks_iterations.png'))
```
joshuacwnewton, line 296: Nit: There is an extra `os.path.join` here.

spinalcordtoolbox/registration/landmarks.py
```diff
@@ -41,7 +41,8 @@
 ini_param_trans_y = -150.0  # pix
 initial_step = 2
 
-def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', verbose=1, path_qc=None):
+
+def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', verbose=1):
```
joshuacwnewton, line 45: What is the motivation for the changes in `landmarks.py`? 

(I'm not necessarily opposed, they are just unexpected to me given the scope of this PR. I am wondering why they were included when they appear to be unrelated.)

spinalcordtoolbox/registration/landmarks.py
```diff
@@ -101,6 +102,7 @@ def register_landmarks(fname_src, fname_dest, dof, fname_affine='affine.txt', ve
                                                            points_moving_barycenter[2]))
     text_file.close()
 
+
```
joshuacwnewton, line 105: Small nit (see Lintly feedback)

joshuacwnewton (approved):
I don't have much to say besides what Nick has already pointed (int labels vs. descriptive labels). 

These points are straightfoward, so I'll pre-approve. :slightly_smiling_face:

spinalcordtoolbox/registration/landmarks.py
jcohenadad: sorry no time to wait for new ci--

spinalcordtoolbox/registration/landmarks.py
jcohenadad: > is it just because path_qc was unused by calls to register_landmarks?

indeed that's it-- indeed, out of scope, i just did it to save time

jcohenadad:
thank you so much for your review @joshuacwnewton @kousu

PR #3310, opened 2021-03-24
Add entry to download exvivo template

### LINKED ISSUE #3280: Add entry to download exvivo template
### Context

@charleygros created [this exvivo template](https://github.com/sct-data/exvivo-template), which is notably used by our collaborators in Australia (see [user forum](https://forum.spinalcordmri.org/t/registration-to-t1-ex-vivo-template-error/644/6)).

Currently there is no instructions as to how to download this template.

### Suggestion

Add an entry to `sct_download_data` to install this template.

Note: this template should ideally be installed under `$SCT_DIR/data/`. It would be cleaner IMHO.

I've added the label "priority" because currently the installation requires users to install https://github.com/Drulex/ghsplit/, which for some users is a complicated series of steps.

charleygros:
#### GitHub

- [ ] I've applied a [release milestone](https://github.com/neuropoly/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/neuropoly/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 

#### PR contents

## Description
Add an entry to `sct_download_data` to install the exvivo template.

## Linked issues
Resolves #3280

### DIFF
--- a/spinalcordtoolbox/scripts/sct_download_data.py
+++ b/spinalcordtoolbox/scripts/sct_download_data.py
@@ -95,6 +95,9 @@ DICT_URL = {
         "https://osf.io/t97ap/?action=download",
         "https://www.neuro.polymtl.ca/_media/downloads/sct/20190117_c2c3_disc_models.zip",
     ],
+    "exvivo_template": [
+        "https://github.com/sct-data/exvivo-template/archive/refs/tags/r20210317.zip"
+    ],
 }
 
 


jcohenadad (changes requested):
related to https://github.com/neuropoly/spinalcordtoolbox/issues/3312, we might want to change the behaviour to download at a specific location. Should we mimic what is done in `sct_deepseg -install-task`?  @joshuacwnewton @Drulex what do you think?

Drulex:
> related to #3312, we might want to change the behaviour to download at a specific location. Should we mimic what is done in `sct_deepseg -install-task`? @joshuacwnewton @Drulex what do you think?

If the user has the ability to specify a download location, consequently the processing should accept a `template_location` (or similar) input path as well. I agree that there can be a default location if not specified.

Having a mix of static paths and configurable path seems like a confusing thing from a user perspective.

jcohenadad:
> > related to #3312, we might want to change the behaviour to download at a specific location. Should we mimic what is done in `sct_deepseg -install-task`? @joshuacwnewton @Drulex what do you think?
> 
> If the user has the ability to specify a download location, consequently the processing should accept a `template_location` (or similar) input path as well. I agree that there can be a default location if not specified.
> 
> Having a mix of static paths and configurable path seems like a confusing thing from a user perspective.

thanks for the feedback. Currently, the processing does accept any template path: 
- `sct_register_to_template -t <PATH>`
- `sct_warp_template -t <PATH>`

So what you're saying is that by "forcing" installation under `$SCT_DIR/data`, we might make it confusing from the user standpoint. I see how it could be the case indeed. So, to sum up the pros/cons of forcing the installation under SCT's dir (top/bottom: most/least important arguments):

- 👎 : People creating their own template not included in the "official list of template" will not be able to use it;
- 👎 : We have some refactoring to do (likely create a new function "sct_download_template" or something like that, and also change the "-t" flag behaviour to point to the name of a template instead of a path)
- 👎 : The change of behaviour with "-t" will break retro-compatibility
- 👍 : We have a better control of the template version user is using, see [discussion](https://github.com/neuropoly/spinalcordtoolbox/issues/3280#issuecomment-805395427)
- 👍 : We control the installation of the template (I anticipate some people will struggle with downloading, unzipping, pointing to the right folder when running the processing, etc.)

Based on this list, it looks like we might want to keep the status quo after all... @Drulex @charleygros @joshuacwnewton what do you think?

Drulex:
> * +1 : We have a better control of the template version user is using, see [discussion](https://github.com/neuropoly/spinalcordtoolbox/issues/3280#issuecomment-805395427)
> * +1 : We control the installation of the template (I anticipate some people will struggle with downloading, unzipping, pointing to the right folder when running the processing, etc.)

I think we should still have a default location entry for each dataset that will be used unless the user specifies otherwise. This would be the location that we want to "force" (it will be more of a nudge than a force).

joshuacwnewton:
> Having a mix of static paths and configurable path seems like a confusing thing from a user perspective.

I think there's a middle-ground between the two options. We could keep the base directory static (`$SCT_DIR/data`) and always download things there. But, also allow users to specify only the subdirectory (rather than any folder). That way, for these concerns:

> * :-1: : People creating their own template not included in the "official list of template" will not be able to use it;
> * :-1: : We have some refactoring to do (likely create a new function "sct_download_template" or something like that, and also change the "-t" flag behaviour to point to the name of a template instead of a path)

We could say "if you would like to use your own template, copy it to the `data/` folder, then specify the name of the template directory using `-t`."

jcohenadad:
> I think we should still have a default location entry for each dataset that will be used unless the user specifies otherwise. This would be the location that we want to "force" (it will be more of a nudge than a force).

> For these concerns, we could keep the base directory static ($SCT_DIR/data) and always download things there. But, also allow users to specify a subdirectory relative to data/. 

OK, so based on that, are you suggesting to modify `sct_download_data` by, for example, including a dictionary that will list the datasets to install under data (ie: PAM50, exvivo-template, etc.)? Few thoughts:
- we need to keep retro-compatibility for sct_download_data;
- we might want to update install_sct: 
  https://github.com/neuropoly/spinalcordtoolbox/blob/af2d4de531f66cd062af6fefa87c3f84185875ff/install_sct#L659-L660
- If we go with that dictionary strategy, we could also do the same for all binaries

joshuacwnewton:
> OK, so based on that, are you suggesting to modify sct_download_data by, for example, including a dictionary that will list the datasets to install under data (ie: PAM50, exvivo-template, etc.)? Few thoughts:

My thinking was that we would change the default location of `sct_download_data/` to `$SCT_DIR/data/${dataset-name}` for everything (i.e. no dictionary).

So, we would no longer download to the current working directory.

jcohenadad:
> My thinking was that we could change the default location of `sct_download_data/` to `$SCT_DIR/data/${dataset-name}` for everything (i.e. no dictionary).

the problem then is for:
- binaries (need to go under bin) --> solution (acceptable?): under install_sct script, we keep indicating the output folder (which we currently do). 
- course material (during SCT course, we tell ppl to run `sct_download_data -d COURSE_MATERIAL`, and by default it would download locally. That will create a retro-compat issue. Not the end of the world of course, but we want to be 100% sure this is the direction we want to take.

the pros i see with this is that the function is already called "sct_download_data", so it makes sense to put it all into "data/" 😊

joshuacwnewton:
Ahhhhh. I understand the motivation for the dictionary now. I forgot that we use `sct_download_data` for so many purposes. :sweat_smile: 

1. Binaries -> `bin/`
2. Models -> `data/`
3. Templates -> `data/`
4. Testing/example data -> `./` (Though IMO this should go somewhere static, too.)
5. Course data -> `./`

I feel like this is coming back around to the same "sct_download_data is doing too much" discussion from https://github.com/neuropoly/spinalcordtoolbox/issues/2959...

Drulex:
> * This modification will increase maintenance on our end. Are we OK with that? (ie: cost/benefit balance is worth it?)
> * we need to keep retro-compatibility for sct_download_data;
> * we might want to update install_sct:
>   https://github.com/neuropoly/spinalcordtoolbox/blob/af2d4de531f66cd062af6fefa87c3f84185875ff/install_sct#L659-L660
> * If we go with that dictionary strategy, we could also do the same for all binaries

We can have both a new format and the old format in parallel. We can write a compatibility layer that maintains backwards compatibility with old calls.

With a new format we can include more metadata. For example (something quick, not final):
```python
DICT_URL = {
    "sct_example_data":  
    {
        "mirrors": [
            {
                "url": "https://github.com/sct-data/sct_example_data/releases/download/r20180525/20180525_sct_example_data.zip",
                "hash": "1f86b30ddaac071a90be988c964789fb"
            },
            ...
        ],

        "default_install_location": "{}/sct_testing_data".format(os.environ.get('SCT_DIR', sct_dir_local_path)),
        "sct_compatibility": ">5.1",
        ...
    },
    ...
}
```
Just adding the hash is going to benefit greatly in my opinion.

joshuacwnewton:
(Side note: Perhaps we should have had this conversation in #3312 instead of @charleygros's PR. :sweat_smile:)

charleygros:
> 
> 
> (Side note: Perhaps we should have had this conversation in #3312 instead of @charleygros's PR. 😅)

:-)
@jcohenadad Green light for merging this PR? while keep having the ball rolling in #3312?

jcohenadad (approved):

PR #3313, opened 2021-03-24
Unset `PYTHONNOUSERSITE` in environment before calling batch script

### LINKED ISSUE #3200: "PYTHONNOUSERSITE=True" leaks into shell scripts run by 'sct_run_batch'
### Description

I am using `sct_run_batch` and in the shell script called, I call a command from [gradient_unwarp](https://github.com/Washington-University/gradunwarp/tree/v1.2.0). I downloaded SCT 5.1.0 and when calling this line :
~~~
gradient_unwarp.py ${file_t1}.nii.gz ${file_t1}_gradcorr.nii.gz siemens -g ${PATH_GRADCORR_FILE}/coeff.grad -n
~~~
This error occurs:
~~~
+ gradient_unwarp.py sub-1000394_T1w_raw_RPI_r.nii.gz sub-1000394_T1w_raw_RPI_r_gradcorr.nii.gz siemens -g /mnt/y/sebeda/coeff.grad/ICM/coeff.grad -n
  File "/usr/local/bin/gradient_unwarp.py", line 11, in <module>
    from gradunwarp.core import (globals, coeffs, utils)
  File "/usr/local/lib/python3.7/dist-packages/gradunwarp/core/__init__.py", line 8, in <module>
    from .unwarp_resample import Unwarper
  File "/usr/local/lib/python3.7/dist-packages/gradunwarp/core/unwarp_resample.py", line 8, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
~~~
I tried runnig the same command in my terminal, there was no error. I went back to SCT v5.0.1 and there was no error, the command worked.
I am running on WSL1 with Ubuntu 18.04.

### Steps to Reproduce
See [README.md](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/split_processing/README.md) (preprocessing section)

0. Install SCT 5.1.0

1. Initialize shell variable with path to gradient correction file
~~~
PATH_GRADCORR_FILE=<path-gradcorr>
~~~
2. Run preprocessing (see [preprocess_data.sh](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/split_processing/preprocess_data.sh))
~~~
sct_run_batch -jobs -1 -path-data <PATH-DATA> -path-output ~/ukbiobank_preprocess -script preprocess_data.sh -script-args $PATH_GRADCORR_FILE
~~~

**Expected behavior:** Using SCT v5.0.1 or running the command in the terminal:
~~~
 gradient_unwarp.py sub-1000032_T1w_raw_RPI_r.nii.gz sub-1000032_T1w_raw_RPI_r_gradcorr.nii.gz siemens -g /mnt/y/sebeda/coeff.grad/ICM/coeff.grad -n
gradunwarp-INFO: Parsing /mnt/y/sebeda/coeff.grad/ICM/coeff.grad for harmonics coeffs
gradunwarp-INFO: Evaluating spherical harmonics
gradunwarp-INFO: on a 60^3 grid
gradunwarp-INFO: with extents -300.0mm to 300.0mm
gradunwarp-INFO: along x...
...
~~~

**Actual behavior:** Here is the complete log file of the processing error.
[err.preprocess_data_sub-1000394.log](https://github.com/neuropoly/spinalcordtoolbox/files/5875679/err.preprocess_data_sub-1000394.log)

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR adds step 3:

1. `sct_run_batch` is called.
2. `PYTHONNOUSERSITE` is set to True in `compat/launcher.py` for the context of `sct_run_batch`.
3. **`PYTHONNOUSERSITE` is unset for the context of the pipeline shell script.**
4. (Optional) A pipeline script calls `sct_<function>`.
5. (Optional) `PYTHONNOUSERSITE` is set to True in `compat/launcher.py`, but only in the context of `sct_<function>`.

Step 3 is needed because `PYTHONNOUSERSITE=True` would leak into `sct_run_batch` pipeline shell scripts and affect non-SCT commands. Specifically, the command `gradunwarp` couldn't access its dependencies if they were installed in the user's site packages directory.

## Linked issues

Fixes #3200.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -198,7 +198,7 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
     })
     # PYTHONNOUSERSITE is originally set in compat/launcher.py. We unset here so that it doesn't affect non-SCT commands
     # in user's pipeline scripts. See: https://github.com/neuropoly/spinalcordtoolbox/issues/3200.
-    del envir['PYTHONNOUSERSITE']
+    envir.pop('PYTHONNOUSERSITE', None)
 
     # Ship the job out, merging stdout/stderr and piping to log file
     try:
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -196,6 +196,9 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
         'ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS': str(itk_threads),
         'SCT_PROGRESS_BAR': 'off'
     })
+    # PYTHONNOUSERSITE is originally set in compat/launcher.py. We unset here so that it doesn't affect non-SCT commands
+    # in user's pipeline scripts. See: https://github.com/neuropoly/spinalcordtoolbox/issues/3200.
+    del envir['PYTHONNOUSERSITE']
 
     # Ship the job out, merging stdout/stderr and piping to log file
     try:


Drulex:
My understanding:

1. We added `PYTHONNOUSERSITE=True` in https://github.com/neuropoly/spinalcordtoolbox/issues/3067, because conda's `venv` was not properly  isolated from the user's site-packages (which still baffles me. I tried printing `sys.path` on a fresh native `python3 venv` and it did not include the user site-packages directory whereas a fresh `conda venv` does :confused: )
2. This causes an issue with https://github.com/neuropoly/spinalcordtoolbox/issues/3200, because we want to use an external dependency with `sct_run_batch`, so we temporarily unset the variable just inside this script.

If `gradient_unwarp` is meant to be used with `sct` then why is it not installed within `sct`'s virtual environment? Installing it outside and then modifying the environment to look in the user site-packages feels a bit hacky, but I am also worried that it may cause another dependency conflict down the line considering that we don't know what kind of scripts users may feed `sct_run_batch`.

@joshuacwnewton what do you think? Doesn't it make more sense to install dependencies within the virtual environment instead of doing this?

joshuacwnewton:
> If gradient_unwarp is meant to be used with sct then why is it not installed within sct's virtual environment? 

Right now, I don't think this is possible, because `gradunwarp` doesn't use setuptools. :slightly_frowning_face: 

More details here: https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/issues/31

Drulex:
> Right now, I don't think this is possible, because `gradunwarp` doesn't use setuptools.

Perhaps, but that's their problem (unfortunately).

I guess if we really care we can patch it for them? I just want to make sure that we solve the right problem.

joshuacwnewton:
> Perhaps, but that's their problem (unfortunately).

That's fair! 

> I guess if we really care we can patch it for them? I just want to make sure that we solve the right problem.

I appreciate that. Let me try answering from a different angle.

> If gradient_unwarp is meant to be used with sct then why is it not installed within sct's virtual environment?

I'm not sure `gradient_unwarp` is an SCT dependency. I see it more as an external program. It can be used alongside SCT, but it's not necessary for SCT, so I think it should kept separate from SCT's installation environment. 

> and then modifying the environment to look in the user site-packages feels a bit hacky, 

Well, to be fair, setting `PYTHONNOUSERSITE=True` is already kind of a a hacky change. :sweat_smile: 

@kousu was the person who first convinced me `PYTHONNOUSERSITE=True` was not a good solution in https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/issues/31#issuecomment-771647698:

> On our side, I think setting `PYTHONUSERSITE=True` and other global variables like it is not a good solution. Most apps, reasonably, assume that they are running on a standard-ish system. Breaking those assumptions makes SCT an incompatible thing like Matlab that people have to containerize to keep running. If we're going to force people to containerize we should at least provide the docker image ourselves.
> 
> For another hacky solution: We could try to isolate `PYTHONUSERSITE` to just SCT itself, maybe? Like, make sure every: `subprocess.run([...], env={k: os.environ[k] for k in set(os.environ.keys()) - {'PYTHONUSERSITE'}})`. That would be a hacky workaround but at least it wouldn't leak out beyond SCT anymore.

So, we would only modify the environment here because we want to restore how the environment was before SCT's modification, to prevent external programs from being affected. :sweat_smile:

kousu:
:grimacing: aoyoe.

I think if we solve https://github.com/neuropoly/spinalcordtoolbox/issues/1526 this will go away. But for now, sure.

kousu (approved):

kousu:
I knew I was looking for something else...

so is it too late to throw this one out there?

```
import site; site.ENABLE_USER_SITE = False
```

I *think* this is the same as PYTHONNOUSERSITE, but it should only exist in the current process. That should probably replace both #3067 and this.

joshuacwnewton:
> I _think_ this is the same as PYTHONNOUSERSITE, but it should only exist in the current process. That should probably replace both #3067 and this.

Hrm. I dug a little deeper and I'm not sure they are the same. It looks like [`site.ENABLE_USER_SITE`](https://docs.python.org/3/library/site.html#site.ENABLE_USER_SITE) is just a status indicator, rather than having any control over what gets added to `sys.path`:

> Flag showing the status of the user site-packages directory. True means that it is enabled and was added to sys.path. False means that it was disabled by user request (with -s or PYTHONNOUSERSITE).

(Probably should have been called `site.USER_SITE_ENABLED`, eh?)

----

As well, the process for adding the user `site-packages` dir to `sys.path` seems to be done on initialization. From the [`site` docs](https://docs.python.org/3/library/site.html):

> This module is automatically imported during initialization. The automatic import can be suppressed using the interpreter’s -S option.
>
> Importing this module will append site-specific paths to the module search path and add a few builtins, unless -S was used.  In that case, this module can be safely imported with no automatic modifications to the module search path or additions to the builtins. 

So, running `import site` again (a la `import site; site.ENABLE_USER_SITE = False`) I think has no effect.

kousu:
Huhm. But then how come this workaround for this pip bug works? https://github.com/pypa/pip/issues/7953#issuecomment-645133255

With it, I can install `pip install -e`; without it I can't.

I guess maybe pip could be paying attention to it directly instead.

joshuacwnewton:
> I guess maybe pip could be paying attention to it directly instead.

[Indeed!](https://github.com/pypa/pip/blob/7ec0fa5142466f8402bb866a1dcfb8ca6ffdc66b/src/pip/_internal/commands/install.py#L645-L648)

Issue #3311, opened 2021-03-24
Discrepancy between `fslhd` and `sct_image -getorient`


jcohenadad:
### Description

fslhd gives LPI while sct_image -getorient gives RPI.

### Steps to Reproduce

```console
git clone https://github.com/sct-data/exvivo-template.git
cd exvivo-template
git checkout 6a595f97410f4a26882852fb6d1ab40389b78206
cd template
```

With sct_image:
```console
sct_image -i exvivo_t2.nii.gz -getorient
RPI
```

with fslhd:
```console
fslhd exvivo_t2.nii.gz
[...]
qform_xorient	Left-to-Right
qform_yorient	Posterior-to-Anterior
qform_zorient	Inferior-to-Superior
```

Related to https://github.com/sct-data/exvivo-template/issues/16

joshuacwnewton:
I am currently working on #504. Since displaying the header will include this information, we could just deprecate `-getorient` as part of that effort?

joshuacwnewton:
I've looked into this behavior, and it turns out this is due to a new bug introduced in #2858.

https://github.com/neuropoly/spinalcordtoolbox/blob/c09a14a2e12317edebda03cad01c8aa4cf64aa31/spinalcordtoolbox/scripts/sct_image.py#L193-L194

The default value for `arguments.set_sform_to_qform` is actually `False`, and _not_ `None`, so right now this always fires. :scream:

The reason this affects orientation is because calling `arguments.set_sform_to_qform` causes `im.hdr.get_best_affine()` to return the wrong affine, and hence the wrong orientation:

```python
# Before `self.hdr._structarr['sform_code'] = self.hdr._structarr['qform_code']`
> im.hdr.get_best_affine()
[[0.08 0.   0.   0.  ]
 [0.   0.08 0.   0.  ]
 [0.   0.   0.08 0.  ]
 [0.   0.   0.   1.  ]]
> res = "".join(nib.orientations.aff2axcodes(im.hdr.get_best_affine()))
'RAS'
> orientation_string_nib2sct(res)
'LPI'

# After `self.hdr._structarr['sform_code'] = self.hdr._structarr['qform_code']`
> im.hdr.get_best_affine()
[[ -0.08         0.           0.           9.95999978]
 [  0.           0.08         0.          -9.95999978]
 [  0.           0.           0.08       -39.55999912]
 [  0.           0.           0.           1.        ]]
[...]

Issue #3281, opened 2021-03-14
Wrong estimation of orientation


jcohenadad:
### Description

While working on this forum issue, I noticed a strange behaviour for the following dataset: [dwi_1500_orient_b0_mean_orient-wrong_gmseg_manual.nii.gz](https://github.com/neuropoly/spinalcordtoolbox/files/6137029/dwi_1500_orient_b0_mean_orient-wrong_gmseg_manual.nii.gz)

The output of
```console
sct_image -i dwi_1500_orient_b0_mean_orient-wrong_gmseg_manual.nii.gz -getorient 
```
gives `RPI` orientation, while the output of:
```console
fslhd dwi_1500_orient_b0_mean_orient-wrong_gmseg_manual.nii.gz
```
gives `ASL`:
```
qform_xorient	Anterior-to-Posterior
qform_yorient	Superior-to-Inferior
qform_zorient	Left-to-Right
```

jcohenadad:
A particularity of this file is that the qform_code=2, which is incorrect. Also see https://github.com/neuropoly/spinalcordtoolbox/issues/2454

joshuacwnewton:
Ah! This issue might be the same as #3311. ("RPI" is reported both times by `-getorient`)

jcohenadad:
> Ah! This issue might be the same as #3311. ("RPI" is reported both times by `-getorient`)

oops! yes you're right 😅 
closing because duplicate of https://github.com/neuropoly/spinalcordtoolbox/issues/3311

PR #3317, opened 2021-03-29
Add flag to `sct_image` to print image headers 

### LINKED ISSUE #540: add flag to display header
e.g. `sct_image -i IMAGE.NII.GZ -header`
would output something similar than fslhd in the terminal, in addition to creating a csv file.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR:
* Reproduces part of `fslhd`'s C++ source code in a Python script (in accordance with their license).
    * The most relevant C++ source file is [`/fsl/src/NewNifti/NewNifti.cc`](https://github.com/neuropoly/spinalcordtoolbox/files/6218778/NewNifti.cc.txt)
  (see [Fslinstallation/SourceCode](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FslInstallation/SourceCode).)
* Wraps the script in an API function, and call API function via `sct_image -header`.

The usage of the CLI script is:

```console
# fslhd formatting
sct_image -i t2.nii.gz -header fslhd

# fslhd formatting with tweaks (e.g. combining `dim` into a one-line field)
sct_image -i t2.nii.gz -header
# Alternative way to specify this
sct_image -i t2.nii.gz -header sct  

# nibabel formatting (i.e. `print(Nifti1Image.header)`)
sct_image -i t2.nii.gz -header nibabel
```

The output of `-header fslhd` should be identical to `fslhd`: https://www.diffchecker.com/gzaVUfj9

(I've manually run this on a variety of images, both Nifti1 and Nifti2. But, I'm a bit stuck on how to write a good `pytest` test for this, considering we can't easily compare with `fslhd` in our test suite. We could compared against cached output, but there are ~60 fields and many possible images, so doing that feels a little awkward.)

## Linked issues

Fixes #540.


joshuacwnewton:
Since this new functionality effectively replaces `sct_image -getorient`, I'm thinking of removing that option in this PR, which would fix https://github.com/neuropoly/spinalcordtoolbox/issues/3311 and possibly https://github.com/neuropoly/spinalcordtoolbox/issues/3281. But, I would need to investigate those issues further to be certain.

jcohenadad:
> (I've manually run this on a variety of images, both Nifti1 and Nifti2. But, I'm a bit stuck on how to write a good pytest test for this, considering we can't easily compare with fslhd in our test suite. We could compared against cached output, but there are ~60 fields and many possible images, so doing that feels a little awkward.)

i agree caching would be the best approach here. Starting with the most important fields, I would say: dim, q/sform, q/sform_code.

jcohenadad:
@joshuacwnewton i'm testing your PR-- can you pls remind me what i need to do in order to get access to the CLI (install_sct being an overkill)- thanks

contrib/fslhd.py
```diff
@@ -0,0 +1,263 @@
...
+    Software or any derivative of it for research with the final aim of
+    developing software products for sale or license to a third party or
+    (3) use of the Software or any derivative of it for research with the
+    final aim of developing non-software products for sale or license to a
+    third party, or (4) use of the Software to provide any service to an
+    external organisation for which payment is received. If you are
+    interested in using the Software commercially, please contact Oxford
+    University Innovation ("OUI"), the technology transfer company of the
+    University, to negotiate a licence. Contact details are:
+    fsl@innovation.ox.ac.uk quoting Reference Project 9564, FSL.*/
```
jcohenadad, line 73: this collides with SCT's LGVLv3 license, no?

jcohenadad (commented):
quick comment on the usage. I did this:
~~~
julien-macbook:~/code/exvivo-template/template $ sct_header 

--
Spinal Cord Toolbox (git-jn/504-display-header-fc6d0200551b9636b457a8dc8da6770d6eed1357)

sct_header 
--

sct_header: error: the following arguments are required: command

usage: sct_header [-h] [-v {0,1,2}] {display} ...

Utility functions for processing the headers of Nifti1 images.

positional arguments:
  {display}
    display   Subcommand for displaying the header.

optional arguments:
  -h, --help  show this help message and exit
  -v {0,1,2}  Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2:
              Debug mode (default: 1)
~~~

but, it is not clear where to write the image. I suggest writing this instead:
~~~
usage: sct_header [-h] [-v {0,1,2}] {display} IMAGE
~~~

Also: I realized after 5min that this is actually a SUB-command. Our users are definitely not used to this approach, and they will likely not understand the usage at all. They expect to see all options in a flat design (1st level). Can we reconsider the subcommand? is that very useful here?

last point: after i understood it was a subcommand, i ran this and got an error:

<details><summary></summary>

```
julien-macbook:~/code/exvivo-template/template $ sct_header display

--
Spinal Cord Toolbox (git-jn/504-display-header-fc6d0200551b9636b457a8dc8da6770d6eed1357)

sct_header display
--

  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_header.py", line 84, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_header.py", line 72, in main
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1734, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1766, in parse_known_args
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1975, in _parse_known_args
    stop_index = consume_positionals(start_index)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1931, in consume_positionals
    take_action(action, args)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1840, in take_action
    action(self, namespace, argument_values, option_string)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1137, in __call__
    subnamespace, arg_strings = parser.parse_known_args(arg_strings, None)
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 1766, in parse_known_args
  File "/Users/julien/code/sct/python/envs/venv_sct/lib/python3.6/argparse.py", line 2001, in _parse_known_args
    ', '.join(required_actions))
TypeError: sequence item 0: expected str instance, Metavar found
julien-macbook:~/code/exvivo-template/template $ sct_header display

```

</details>

joshuacwnewton:
> @joshuacwnewton i'm testing your PR-- can you pls remind me what i need to do in order to get access to the CLI (install_sct being an overkill)- thanks
> 
> EDIT: i just went ahead with install_sct

Thanks much! `pip install -e .` would do the job here. I've added [a new section](https://github.com/neuropoly/spinalcordtoolbox/wiki/Contributing:-Installing-SCT-for-development#reinstalling-after-changes-to-packages-cli-etc) to the dev Wiki for this. :slightly_smiling_face:

joshuacwnewton:
> Also: I realized after 5min that this is actually a SUB-command. Our users are definitely not used to this approach, and they will likely not understand the usage at all. They expect to see all options in a flat design (1st level). Can we reconsider the subcommand? is that very useful here?

I did it this way because subcommands had come up in past discussion (see https://github.com/neuropoly/spinalcordtoolbox/issues/3054, https://github.com/neuropoly/spinalcordtoolbox/issues/3159) and I wanted to experiment to see if they were a viable option. 

But, you've provided very useful feedback: If subcommands felt unintuitive, then I agree that this is not the best choice, and I'm happy to revert it to our typical approach.

contrib/fslhd.py
joshuacwnewton: Thanks for pointing this out. Truthfully, I'm not 100% sure of the legal implications when multiple licenses interact with one another.

My goal was to follow ["combined work with separation"](https://en.wikipedia.org/wiki/License_compatibility#Kinds_of_combined_works) by putting the derivative work into the `contrib/` folder. But, I'm not a lawyer, so I'm unsure if this is a valid way of combining licenses

joshuacwnewton: The tricky thing here is that FSL doesn't use an off-the-shelf license -- this looks to be their own custom license. This makes it harder to rely on advice online, which mainly talks about interactions between LGPL, GPL, MIT, BSD, etc.

jcohenadad:
> Thanks much! pip install -e . would do the job here.

actually that's what i did (under the sct_venv), but it did't create the launchers

joshuacwnewton:
> > Thanks much! pip install -e . would do the job here.
> 
> actually that's what i did (under the sct_venv), but it did't create the launchers

Interesting... I tried it just now and it worked OK. :thinking: 

<details>
<summary> Steps I tested </summary>

1. Add a dummy command to `setup.py`

```diff
    entry_points=dict(
        console_scripts=[
            '{}=spinalcordtoolbox.compat.launcher:main'.format(x) for x in \
            [
+               'test_command'
                'isct_convert_binary_to_trilinear',
                'isct_minc2volume-viewer',
```

2. Reinstall and try command

```console
> test_command
zsh: command not found: test_command
> pip install -e .
Obtaining file:///home/joshua/repos/spinalcordtoolbox
Installing collected packages: spinalcordtoolbox
  Attempting uninstall: spinalcordtoolbox
    Found existing installation: spinalcordtoolbox dev
    Uninstalling spinalcordtoolbox-dev:
      Successfully uninstalled spinalcordtoolbox-dev
  Running setup.py develop for spinalcordtoolbox
Successfully installed spinalcordtoolbox
> test_command
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/bin/test_command", line 33, in <module>
    sys.exit(load_entry_point('spinalcordtoolbox', 'console_scripts', 'test_command')())
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/compat/launcher.py", line 34, in main
    assert os.path.exists(script)
AssertionError
# Tries to go through `compat/launcher.py`, so it's available
```

</details>

jcohenadad (commented):
the "format" flag is really cool. I am just wondering: do we need to keep both "sct" and "fslhd" formats, considering they are super close? ultimately, people will not use this function to compare the output with fslhd (if they have fslhd, they will just use it). Or, did you create the "fslhd" formatting to facilitate unit testing?

joshuacwnewton:
> the "format" flag is really cool. I am just wondering: do we need to keep both "sct" and "fslhd" formats, considering they are super close? ultimately, people will not use this function to compare the output with fslhd (if they have fslhd, they will just use it). Or, did you create the "fslhd" formatting to facilitate unit testing?

That's a fair point! You're right, most people won't use it. The scenario I was thinking of was: say one of us posts the output of `fslhd` on a GH issue, then another dev starts debugging and realize they don't have `fslhd` installed, then they could use SCT to generate matching output to compare.

joshuacwnewton:
The failing RTD is confusing me, as I'm not able to reproduce the error message locally on this branch. 

I took a look at the `sphinxcontrib-programoutput` repo and noticed that another user had recently opened an issue with similar failing output. I'm not sure if it's the same problem or just coincidence, but linking just in case: https://github.com/NextThought/sphinxcontrib-programoutput/issues/50#issuecomment-811091463

joshuacwnewton:
> i agree caching would be the best approach here. Starting with the most important fields, I would say: dim, q/sform, q/sform_code.

I'm still not quite sure about this kind of test. 

All that `sct_image -header` does is echo the values already stored in the image. So, if any problems were to arise, then the root cause is more likely to be because of the specific image (and not `sct_image -header`). Since most problems would be data-dependent, I'm not sure it would be useful to write a test that caches values for 1 image.

joshuacwnewton:
> i agree caching would be the best approach here. Starting with the most important fields, I would say: dim, q/sform, q/sform_code.

I'm still not quite sure about this kind of test. 

All that `sct_image -header` does is echo the values already stored in the image. So, if any problems were to arise, then the root cause is more likely to be because of the image itself (and not `sct_image -header`). Since most problems would be data-dependent, I'm not sure that a test using cached values would catch anything.

jcohenadad:
> I'm still not quite sure about this kind of test.
> All that sct_image -header does is echo the values already stored in the image. So, if any problems were to arise, then the root cause is more likely to be because of the image itself (and not sct_image -header). Since most problems would be data-dependent, I'm not sure that a test using cached values would catch anything for us.

hum... yes, fair point. How about a simple functional test just to make sure the function doesn't crash?

unit_testing/cli/test_cli_sct_image.py
```diff
@@ -12,3 +14,9 @@ def test_sct_image_backwards_compat(script_runner):
     logger.debug(f"{ret.stderr}")
     assert ret.success
     assert ret.stderr == ''
+
+
+@pytest.mark.parametrize("output_format", ('sct', 'fslhd', 'nibabel'))
+def test_sct_image_show_header_no_checks(output_format):
+    """Run the CLI script without checking results."""
+    sct_image.main(argv=['-i', 'sct_testing_data/t2/t2.nii.gz', '-header', output_format])
```
joshuacwnewton, line 22: > > I'm still not quite sure about this kind of test.
> > All that sct_image -header does is echo the values already stored in the image. So, if any problems were to arise, then the root cause is more likely to be because of the image itself (and not sct_image -header). Since most problems would be data-dependent, I'm not sure that a test using cached values would catch anything for us.
> 
> hum... yes, fair point. How about a simple functional test just to make sure the function doesn't crash?

:wink:

joshuacwnewton:
Note to self: We should also run `sct_image -header` on the relevant image files from #3311 and #3281 to confirm that the problem was with `-getorient`.

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -91,6 +92,15 @@ def get_parser():
...
 
     header = parser.add_argument_group('HEADER OPERATIONS')
+    header.add_argument(
+        "-header",
+        choices=HEADER_FORMATS,
+        # 'const' and 'nargs' used because of https://stackoverflow.com/q/40324356
+        const='sct',
+        nargs='?',
+        help="Print the header using the given output format. Choose 'nibabel' or 'fslhd' if you already have header "
+             "information from those programs and you want to compute the diff between two headers."
```
jcohenadad, line 102: I would remove the blurb about the diff (it will confuse our users). Also i changed the usage to make it clearer that adding the format is optional (our users don't understand the POSIX "[...]" semantic in usage).
```suggestion
        help="Print the header of a NIfTI file. You can select the output format of the header: 'sct' (default), 'nibabel' or 'fslhd'."
```

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -108,7 +118,8 @@ def get_parser():
     orientation = parser.add_argument_group('ORIENTATION OPERATIONS')
     orientation.add_argument(
         '-getorient',
-        help='Get orientation of the input image',
+        help='Get orientation of the input image. NOTE: This option has been deprecated, and will be removed in a '
```
jcohenadad, line 121: i have after thoughts: do we really want to remove this? few considerations:
- we will still have the API (in our code, we need to know what is the orientation)
- some users might be confused by the output of "sct_image -header" (people not used to the concept of qform/sform won't know how to get the orientation from there)
- it is not too much of a burden to maintain
- when we display the qform and sform, users don't know what SCT effectively uses (in case they are different). By selecting the one we rely on (ie: the qform), it will be clear for users what is the _effective_ orientation of the image (from SCT perspective)

joshuacwnewton: These are good points. Especially this one:

> some users might be confused by the output of "sct_image -header" (people not used to the concept of qform/sform won't know how to get the orientation from there)

I had been working under the assumption that our users knew how to parse `fslhd` output. This is really good to know, because it changes the warning message I was going to write for https://github.com/neuropoly/spinalcordtoolbox/issues/3249#issuecomment-803681048. 

Anyways, I'll undo the `-getorient` changes and take some time to investigate the cause of #3311 separately from this PR.

jcohenadad: > Anyways, I'll undo the -getorient changes and take some time to investigate the cause of #3311 separately from this PR.

maybe we can refactor the way we get the orientation, now that we parse the header, and just pick the first letter of the output qform orientation in each axis (ie: the way _we_ assess orientation from fslhd)

jcohenadad (approved):
Awesome contrib! thanks

PR #3320, opened 2021-03-30
Implement QC for sct_dmri_moco and sct_fmri_moco

### LINKED ISSUE #3242: Implement QC for dmri_moco and fmri_moco
Addressing issue #1343 to introduce QC reports for `sct_fmri_moco `and `sct_dmri_moco`:

The report could display GIF images before and after motion correction in an axial mosaic which loops across the 4th dimension (time), as well as a graphic for motion correction parameters.

Mandatory input:
- fname_in1, 4D image after motion correction 

Optional inputs:
- fname_in2, 4D image after motion correction
- fname_seg, segmentation of input image used for cropping
- fps, frame rate for output gif images
- moco_params, TSV file with X and Y motion correction parameters

**Tasks:**
- [x] Introduce new parameter -qc to `sct_qc `for `sct_fmri_moco `and `sct_dmri_moco` #3274
- [x] Create entry in `sct_qc ` and `generate_qc ` for `sct_fmri_moco `and `sct_dmri_moco` 
- [x] Add the function `mosaic_through_time ` in the `Slice `class that will create an axial mosaic for each volume of the input images.
- [x] Add functionnality in `qc.py` to save list of png mosaic files ordered by time.
- [x] Add function in `qc.py ` to generate animated GIF image from the list of png mosaics.
- [x] Add function for white grid overlay in `qc.py `to aid visualization of motion correction.
- [x] Add possibility to crop images to focus on the spinal cord by calculating center of mass (using spinal cord segmentation if provided).
- [ ] Add function to generate moco_params graphic and save as png.
- [ ] Adapt QC report layout to include GIF images and to be able to display up to 3 figures, as shown in preview bellow.

**Proposed layout:**
Mosaics loop through time in GIF format. Overlay is white grid to facilitate visualization.
apercu_moco_qc

claragiroux:
#### GitHub

#### PR contents

## Description

This PR introduces QC reports for `sct_fmri_moco `and `sct_dmri_moco` with the following changes:
- Introducing QC flags to `sct_fmri_moco `and `sct_dmri_moco` and call to generate_qc.
- Creating entry in `sct_qc ` and `generate_qc ` for `sct_fmri_moco `and `sct_dmri_moco` .
- Adding the function `mosaic_through_time ` in the `Slice `class that will create an axial mosaic for each volume of the input images and adapt `Slice `for 4d images.
- Adding the method `grid ` in `QcImage ` to add white grid overlay in `grid `to aid visualization of motion correction.
- Adding the method `_generate_moco_figure ` in `wrapped_f ` to create figures with two mosaics.
- Saving a list of png mosaic files ordered by time for overlay and background images.
- Adding the method `_save_gif ` in `wrapped_f ` to save overlay and background GIF images from the list of png mosaics.

Feature missing from issue #3242 :
For now I have not implemented the graphic with moco params. It requires adaptation of the QC report layout to be able to display 2 images side by side.

## Linked issues

Fixes #3242


jcohenadad (commented):
This is a great first start @claragiroux 🏅 
Below are my comments based on 18a4364efa0477ccf86700b64e07b9b820b9a161

1. The Terminal output displays all sorts of things related to the generation of the QC, but there is no announcement of the QC report generation and because it appears at the end, the user misses the "final" message about the output files and fsleyes syntax.
  <details><summary>See output of sct_fmri_moco -qc</summary>

```console

julien-macbook:~/sct_testing_data/fmri $ sct_fmri_moco -i fmri.nii.gz -qc qc

--
Spinal Cord Toolbox (git-3242-QC-dmri-fmri-moco-18a4364efa0477ccf86700b64e07b9b820b9a161)

sct_fmri_moco -i fmri.nii.gz -qc qc
--


Input parameters:
  Input file ............ fmri.nii.gz
  Group size ............ 1
Creating temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331173441.821411-moco-eq7lzs1r)

Copying input data to tmp folder and convert to nii...

Get dimensions of data...
  68 x 68 x 6

Data orientation: RPI
  Treated as axial

Set suffix of transformation file name, which depends on the orientation:
Orientation is axial, suffix is 'Warp.nii.gz'. The estimated transformation is a 3D warping field, which is composed of a stack of 2D Tx-Ty transformations

Split along T dimension...
Merge within groups: 100%|███████████████████| 30/30 [00:00<00:00, 154.18iter/s]

Merge across groups...

-------------------------------------------------------------------------------
  Estimating motion across groups...
-------------------------------------------------------------------------------

Input parameters:
  Input file ............ datasub-groups.nii
  Reference file ........ datasub_0_mean.nii.gz
  Polynomial degree ..... 2
  Smoothing kernel ...... 0
  Gradient step ......... 1
  Metric ................ MeanSquares
  Sampling .............. None
  Todo .................. estimate_and_apply
  Mask  ................. 
  Output mat folder ..... mat_groups

Data dimensions:
  68 x 68 x 6 x 30

Copy file_target to a temporary file...

Register. Loop across Z (note: there is only one Z if orientation is axial)
Z=0/0: 100%|██████████████████████████████████| 30/30 [00:15<00:00,  1.92iter/s]

-------------------------------------------------------------------------------
  Apply moco
-------------------------------------------------------------------------------

Input parameters:
  Input file ............ data.nii
  Reference file ........ datasub_0_mean.nii.gz
  Polynomial degree ..... 2
  Smoothing kernel ...... 0
  Gradient step ......... 1
  Metric ................ MeanSquares
  Sampling .............. None
  Todo .................. apply
  Mask  ................. 
  Output mat folder ..... mat_final/

Data dimensions:
  68 x 68 x 6 x 30

Copy file_target to a temporary file...

Register. Loop across Z (note: there is only one Z if orientation is axial)
Z=0/0: 100%|██████████████████████████████████| 30/30 [00:03<00:00,  9.08iter/s]
Extract motion parameters...

Generate output files...

Delete temporary files...
rm -rf /var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331173441.821411-moco-eq7lzs1r

Finished! Elapsed time: 20s

Done! To view results, type:
fsleyes ./fmri_moco.nii.gz fmri.nii.gz &

cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap.min.css /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap-table.min.css /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/select2.min.css /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap.min.css.map /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/style.css /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap-theme.min.css /Users/julien/sct_testing_data/fmri/qc/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/select2.min.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/jquery-3.1.0.min.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/bootstrap-table.min.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/main.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/bootstrap.min.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/animation.js /Users/julien/sct_testing_data/fmri/qc/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/axial.png /Users/julien/sct_testing_data/fmri/qc/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/sagittal.png /Users/julien/sct_testing_data/fmri/qc/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/sct_logo.png /Users/julien/sct_testing_data/fmri/qc/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.woff /Users/julien/sct_testing_data/fmri/qc/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.eot /Users/julien/sct_testing_data/fmri/qc/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.woff2 /Users/julien/sct_testing_data/fmri/qc/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.ttf /Users/julien/sct_testing_data/fmri/qc/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.svg /Users/julien/sct_testing_data/fmri/qc/_assets/fonts
```

</details>

<details><summary>For comparison, here is the output of sct_deepseg_sc</summary>

```
julien-macbook:~/sct_testing_data/fmri $ sct_deepseg_sc -i fmri_moco_mean.nii.gz -c t2 -qc qc1

--
Spinal Cord Toolbox (git-3242-QC-dmri-fmri-moco-18a4364efa0477ccf86700b64e07b9b820b9a161)

sct_deepseg_sc -i fmri_moco_mean.nii.gz -c t2 -qc qc1
--

Config deepseg_sc:
  Centerline algorithm: svm
  Brain in image: True
  Kernel dimension: 2d
  Contrast: t2
  Threshold: 0.7
Creating temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331175224.569693-5j6mbs79)
Creating temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331175224.625246-wqgeymz8)
Remove temporary files...
rm -rf /var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331175224.625246-wqgeymz8
Cropping the image around the spinal cord...
Normalizing the intensity...
Segmenting the spinal cord using deep learning on 2D patches...
Reassembling the image...
Resampling the segmentation to the native image resolution using linear interpolation...
Binarizing the resampled segmentation...
Compute shape analysis: 100%|###################| 6/6 [00:00<00:00, 46.92iter/s]
Remove temporary files...
rm -rf /var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20210331175224.569693-5j6mbs79
File /Users/julien/sct_testing_data/fmri/fmri_moco_mean_seg.nii.gz already exists. Will overwrite it.

*** Generate Quality Control (QC) html report ***
Resample images to 0.6x0.6 mm
QcImage: layout with Axial slice
Compute center of mass at each slice
/Users/julien/sct_testing_data/fmri/qc1/julien/sct_testing_data/fmri/sct_deepseg_sc/2021_03_31_175225.422095/bkg_img.png
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap.min.css /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap-table.min.css /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/select2.min.css /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap.min.css.map /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/style.css /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/css/bootstrap-theme.min.css /Users/julien/sct_testing_data/fmri/qc1/_assets/css
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/select2.min.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/jquery-3.1.0.min.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/bootstrap-table.min.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/main.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/bootstrap.min.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/js/animation.js /Users/julien/sct_testing_data/fmri/qc1/_assets/js
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/axial.png /Users/julien/sct_testing_data/fmri/qc1/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/sagittal.png /Users/julien/sct_testing_data/fmri/qc1/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/imgs/sct_logo.png /Users/julien/sct_testing_data/fmri/qc1/_assets/imgs
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.woff /Users/julien/sct_testing_data/fmri/qc1/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.eot /Users/julien/sct_testing_data/fmri/qc1/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.woff2 /Users/julien/sct_testing_data/fmri/qc1/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.ttf /Users/julien/sct_testing_data/fmri/qc1/_assets/fonts
cp /Users/julien/code/sct/spinalcordtoolbox/reports/assets/_assets/fonts/glyphicons-halflings-regular.svg /Users/julien/sct_testing_data/fmri/qc1/_assets/fonts
Successfully generated the QC results in /Users/julien/sct_testing_data/fmri/qc1/_json/qc_2021_03_31_175225.422095.json
Use the following command to see the results in a browser:
open "/Users/julien/sct_testing_data/fmri/qc1/index.html"

Done! To view results, type:
fsleyes /Users/julien/sct_testing_data/fmri/fmri_moco_mean.nii.gz -cm greyscale /Users/julien/sct_testing_data/fmri/fmri_moco_mean_seg.nii.gz -cm red -a 70.0 &
```

</details>

2. The "Command" syntax is wrong:
  image

3. I would switch "before" and "after" (more intuitive to have the "before" before)
   image

4. The generation of QC is fairly long compared to the actual process of the function:
  with QC. 
  ```
  real	0m35.232s
  user	0m47.922s
  sys	0m5.376s
  ```
  without QC:
  ```
  real	0m21.538s
  user	0m25.379s
  sys	0m2.933s
  ```
  Maybe there is room for optimization? 

5. The QC without providing a segmentation is pretty useless (see below). I am wondering if we should maybe "impose" a segmentation with the QC feature, otherwise users might wrongly think that the motion correction failed (because they don't see the spinal cord). 
  image

6. The "current volume" is pretty lonely out there (see below). I would move it to the left. Also, I would remove "Current" (pretty self-explanatory trivial) and indicate the total number. E.g.: `Volume 3/34`
  Screen Shot 2021-03-31 at 6 04 43 PM

claragiroux:
Thanks for the great feedback! @jcohenadad 
I will correct the QC to match your comments and try to find if there is a way to reduce execution time.
Also, calling `moco_wrapper` seems to be changing the verbose to 0, if i reset it to 1 before calling `generate_qc` it resolves the logging issue in your first comment. Is there a way to check the value of the global verbose to test this?

jcohenadad:
> Also, calling `moco_wrapper` seems to be changing the verbose to 0, if i reset it to 1 before calling `generate_qc` it resolves the logging issue in your first comment. Is there a way to check the value of the global verbose to test this?

the key would be to display the QC _before_ the following syntax, which is under `moco_wrapper`:
https://github.com/neuropoly/spinalcordtoolbox/blob/16d3ce661023c9c998cf6df7a0fa772b57033fa2/spinalcordtoolbox/moco.py#L475-L477

So you can move that syntax above into the respective CLIs (dmri/fmri_moco) after the call to QC, and enforce the verbose=1 to fix the issue.

claragiroux:
@jcohenadad I've added new commits following your suggestions.

The following changes were made: 

1. Update sct_fmri_moco, sct_dmri_moco and moco.py to display terminal outputs correctly (like in sct_deepseg_sc).
2. Correct the command syntax in the QC report.
3. The 'before' image is now displayed on top in the QC report.
4. Since saving all the individual png images that were used to build the gif images was taking a long execution time, I changed the gif saving method. Now, the gif is saved directly from a figure object using matplotlib.animation.FuncAnimation. The total execution time is reduced by about 40-50%. 
5. I've made the segmentation input mandatory for the QC of moco functions.
6. `Current volume : #` was changed to `Volume #/total` and placed on the left.

image

jcohenadad:
@claragiroux i'm looking at your PR now. Can you please always keep your branch up-to-date with master? thanks

jcohenadad (changes requested):
Tested 982a55e0949ec380ef365211bb83d115993b80f4: 
- If not specifying an input segmentation, the program crashes at the end with the following message:
~~~
*** Generate Quality Control (QC) html report ***
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_fmri_moco.py", line 195, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_fmri_moco.py", line 188, in main
    subject=qc_subject, process='sct_fmri_moco')
  File "/Users/julien/code/sct/spinalcordtoolbox/reports/qc.py", line 791, in generate_qc
    raise Exception("Segmentation is needed to ensure proper cropping around spinal cord.")
Exception: Segmentation is needed to ensure proper cropping around spinal cord.
~~~
It would be much better to have the program crash at the beginning, while parsing the arguments, so the user doesn't have to wait until the end to realize they need to input a segmentation. Sometimes, this script can take >10min, so it would be frustrating to have it crash after 10min of processing.

Apart from that minor fix, the rest looks good 👍

joshuacwnewton:
I tried running `-i sct_example_data/fmri/fmri_crop.nii.gz -qc qc -qc-seg sct_example_data/fmri/fmri_crop_moco_mean_seg_manual.nii.gz`, but had a bit of trouble with the current layout.

1. The before/after mosaics are currently quite far apart. So, I found that my eyes were darting up-down-up-down when trying to compare each slice, making it difficult to tell what the differences were.

szoter_annotated_image(1)

2. I also found the default FPS to be quite fast... my eyes couldn't dart up and down fast enough to keep up with the changing volumes.

So, I'm wondering: Am I looking at the QC report in the wrong way? (For example, should I not be trying to compare slice-by-slice, and instead just focus on the "big picture" i.e. whether the slices are centred or not?)

spinalcordtoolbox/reports/qc.py
```diff
@@ -419,15 +522,25 @@ def __init__(self, input_file, command, args, orientation, dest_folder, dpi=300,
...
 
     def abs_overlay_img_path(self):
         return os.path.join(self.root_folder, self.overlay_img_path)
 
+    def abs_bkg_gif_path(self):
+        return os.path.join(self.root_folder, self.bkg_img_path)
+
+    def abs_overlay_gif_path(self):
+        return os.path.join(self.root_folder, self.overlay_img_path)
+
```
joshuacwnewton, line 543: In lines 538-543, do we actually need new attributes `abs_bkg_gif_path` and `abs_overlay_gif_path`? Aren't they identical to `abs_bkg_img_path` and `abs_overlay_img_path`?

The distinction between `.gif` and `.png` is already made earlier on when `self.bkg_img_path` and `self.overlay_img_path` are first set, right? (lines 525-530)

spinalcordtoolbox/reports/qc.py
```diff
@@ -456,13 +569,20 @@ def make_content_path(self):
...
+            except OSError as err:
+                if not os.path.isdir(target_img_folder):
+                    raise err
+        else:
+            target_img_folder = os.path.dirname(self.qc_params.abs_bkg_img_path())
+            try:
+                os.makedirs(target_img_folder, exist_ok=True)
+            except OSError as err:
+                if not os.path.isdir(target_img_folder):
+                    raise err
```
joshuacwnewton, line 585: If `abs_bkg_gif_path` turns out not to be necessary, then I think this whole change is also unnecessary, since the existing code will be sufficient.

spinalcordtoolbox/reports/qc.py
```diff
@@ -419,15 +522,25 @@ def __init__(self, input_file, command, args, orientation, dest_folder, dpi=300,
...
         self.mod_date = datetime.datetime.strftime(datetime.datetime.now(), '%Y_%m_%d_%H%M%S.%f')
         self.qc_results = os.path.join(dest_folder, '_json/qc_' + self.mod_date + '.json')
-        self.bkg_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'bkg_img.png')
-        self.overlay_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'overlay_img.png')
+        if command in ['sct_fmri_moco', 'sct_dmri_moco']:
+            self.bkg_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'bkg_img.gif')
+            self.overlay_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'overlay_img.gif')
+        else:
+            self.bkg_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'bkg_img.png')
+            self.overlay_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, 'overlay_img.png')
```
joshuacwnewton, line 530: Nitpick: This could be simplified a little:

```python
    if command in ['sct_fmri_moco', 'sct_dmri_moco']:
        ext = "gif"
    else: 
        ext = "png"
    self.bkg_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, f"bkg_img.{ext}")
    self.overlay_img_path = os.path.join(dataset, subject, contrast, command, self.mod_date, f"overlay_img.{ext}")
```

spinalcordtoolbox/reports/qc.py
```diff
@@ -264,17 +278,20 @@ def wrapped_f(sct_slice, *args):
...
                 [img, mask], centermass = func(sct_slice, *args)
                 self._centermass = centermass
+            if self._fps is not None:
+                [images_after_moco, images_before_moco], centermass = func(sct_slice, *args)
+                self._centermass = centermass
+            else:
+                img, mask = func(sct_slice, *args)
+
+            # Get the aspect ratio (height/width) based on pixel size. Consider only the first 2 slices.
+            self.aspect_img, self.aspect_mask = sct_slice.aspect()[:2]
```
joshuacwnewton, line 294: Hmm... looking at the `__call__` function, it's quite complex right now. It has a lot of interwoven branching paths based on various conditions. By adding moco code using `if self._fps is not None` patches, the function becomes even more complex.

To me, this is a ["code smell"](https://en.wikipedia.org/wiki/Code_smell) indicating that `__call__` should be refactored in a more [modular](https://aist.global/en/importance-of-modular-programming) way.

Because it isn't very modular, I think `__call__` is quite error-prone and hard to maintain. For example, what happens if the `fps` parameter got passed to a non-moco instance of `QcImage`? I have a feeling `__call__` would break.

spinalcordtoolbox/reports/qc.py
```diff
@@ -605,8 +727,8 @@ def layout(qslice):
         print("WARNING! Platform undetectable.")
 
 
-def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args=None, path_qc=None, dataset=None,
-                subject=None, path_img=None, process=None):
+def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args=None, path_qc=None, fps=None,
+                dataset=None, subject=None, path_img=None, process=None):
```
joshuacwnewton, line 731: Nitpick, but it might be safer to add `fps` as the last parameter, just in case there are any calls to `generate_qc` passing parameters positionally.

spinalcordtoolbox/reports/qc.py
```diff
@@ -63,21 +64,23 @@ class QcImage(object):
...
-                 stretch_contrast_method='contrast_stretching', angle_line=None):
+                 stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
         """
         :param qc_report: QcReport: The QC report object
         :param interpolation: str: Type of interpolation used in matplotlib
         :param action_list: list: List of functions that generates a specific type of images
         :param stretch_contrast: adjust image so as to improve contrast
         :param stretch_contrast_method: str: {'contrast_stretching', 'equalized'}: Method for stretching contrast
         :param angle_line: float: See generate_qc()
+        :param fps: float: Number of frames per second for output gif images
```
joshuacwnewton, line 75: Right now, `fps` is only valid for moco, right? It might be good to be explicit about this in the parameter description.

spinalcordtoolbox/reports/qc.py
```diff
@@ -618,6 +740,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
     each slice, for slice that don't have an angle to display, a nan is expected. To be used for assessing cord orientation.
     :param args: args from parent function
     :param path_qc: str: Path to save QC report
+    :param fps: float: Number of frames per second for output gif images
```
joshuacwnewton, line 743: Right now, `fps` is only valid for moco, right? It might be good to be explicit about this in the parameter descriptions.

Maybe even run a check to ensure that `fps` is only passed for `process` == `['sci_dmri_moco', 'sct_fmri_moco']`.

spinalcordtoolbox/scripts/sct_dmri_moco.py
```diff
@@ -161,8 +193,28 @@ def main(argv=None):
...
+
+    if path_qc is not None:
+        if qc_seg is None:
+            raise Exception("-qc-seg is needed to ensure proper cropping around spinal cord in QC report.")
+
     # run moco
-    moco_wrapper(param)
+    fname_output_image = moco_wrapper(param)
+
+    set_global_loglevel(verbose)  # moco_wrapper seems to change verbose to 0
```
joshuacwnewton, line 209: I'm really curious about why this is. This seems worth investigating further.

spinalcordtoolbox/scripts/sct_dmri_moco.py
```diff
@@ -135,7 +137,37 @@ def get_parser():
...
+        '-qc',
+        metavar=Metavar.folder,
+        action=ActionCreateFolder,
+        help="The path where the quality control generated content will be saved."
+    )
+    optional.add_argument(
+        '-qc-seg',
+        metavar=Metavar.file,
+        help="Segmentation of spinal cord to improve cropping in qc report"
+    )
```
joshuacwnewton, line 152: Right now, it's not very clear that `-qc-seg` is mandatory for `-qc`. It might be good to add a note about this in the help descriptions.

When I tried `-qc` for the first time, I only found out `-qc-seg` was necessary via the error message, which felt a bit jarring.

spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -55,6 +56,11 @@ def get_parser():
                         help='If provided, this string will be mentioned in the QC report as the subject the process '
                              'was run on',
                         required=False)
+    parser.add_argument('-fps',
+                        metavar='float',
+                        type=float,
+                        help='The number of frames per second for output gif images',
+                        required=False)
```
joshuacwnewton, line 63: Ditto again for making it clear that `-fps` is only valid for the moco functions. 

(So, maybe add a note about this in the help descriptions, and throw an exception if combined with non-moco functions.)

joshuacwnewton (changes requested):
Thanks for taking the time to make this contribution!

I've taken a look at everything except `spinalcordtoolbox/reports/slice.py` (I ran out of time), but I plan to give it a look soon. :slightly_smiling_face:

spinalcordtoolbox/reports/qc.py
joshuacwnewton: <details>
<summary> Aside about <code>__call__</code> </summary>

Looking at the `__call__` function, it's quite complex right now. It has a lot of interwoven branching paths based on various conditions. By adding moco code using `if self._fps is not None` patches, the function becomes even more complex.

To me, this is a ["code smell"](https://en.wikipedia.org/wiki/Code_smell) indicating that `__call__` should be refactored in a more [modular](https://aist.global/en/importance-of-modular-programming) way. Because it isn't very modular, I think `__call__` is quite error-prone and hard to maintain right now.
</details>

claragiroux:
> I tried running `-i sct_example_data/fmri/fmri_crop.nii.gz -qc qc -qc-seg sct_example_data/fmri/fmri_crop_moco_mean_seg_manual.nii.gz`, but had a bit of trouble with the current layout.
> 
> 1. The before/after mosaics are currently quite far apart. So, I found that my eyes were darting up-down-up-down when trying to compare each slice, making it difficult to tell what the differences were.
> 
> 2. I also found the default FPS to be quite fast... my eyes couldn't dart up and down fast enough to keep up with the changing volumes.
> 
> So, I'm wondering: Am I looking at the QC report in the wrong way? (For example, should I not be trying to compare slice-by-slice, and instead just focus on the "big picture" i.e. whether the slices are centred or not?)

Thanks for pointing this out @joshuacwnewton! I agree that 2 fps might be too fast!
If I put the images closer and change the default fps to 1 do you find it easier to keep up with the changing volumes? Another option would be to put the "After motion correction" title at the bottom so the two mosaics could be even closer.
overlay_img

spinalcordtoolbox/reports/qc.py
claragiroux: You're right, they're not necessary anymore, thanks for pointing this out! I was using them in a first version where I was saving all the indivual frames as png images, but I didn't think to remove `abs_bkg_gif_path` and `abs_overlay_gif_path` after making the change!

spinalcordtoolbox/scripts/sct_dmri_moco.py
jcohenadad: > Right now, it's not very clear that -qc-seg is mandatory for -qc. It might be good to add a note about this in the help descriptions.

+1

jcohenadad:
> Thanks for pointing this out @joshuacwnewton! I agree that 2 fps might be too fast!

hum... i wanted to suggest to put it faster 😉   the way we do this QC, is by focusing on the 'after', not by switching back and forth between the before/after. A proper QCing of a moco algo is to focus on the _same_ slice and see if it is stable. Usually I do that with 5 fps. For reference, see the default fps of FSLeyes's video mode (which was created for moco QC): it is probably around 5fps. 

So i would definitely advise against putting it at 1fps. Either leave it at 2, or put it at 3. 

> Another option would be to put the "After motion correction" title at the bottom so the two mosaics could be even closer.

no, don't do that, it will be confusing. And as i mentioned above, a proper QCing is to not switch between the two views.

spinalcordtoolbox/reports/qc.py
```diff
@@ -264,85 +281,127 @@ def wrapped_f(sct_slice, *args):
...
+            else:
+                if self._angle_line is not None:
+                    [img, mask], centermass = func(sct_slice, *args)
+                    self._centermass = centermass
+                else:
+                    img, mask = func(sct_slice, *args)
+
+                self._make_QC_image_for_3d_volumes(img, mask, slice_orientation=sct_slice.get_name())
+
+        return wrapped_f
```
claragiroux, line 304: @joshuacwnewton 
I made two functions to call in wrapped_f to make it shorter and remove all the `if self._fps is not None`. Is this the kind of changes you were expecting to make it more modular?

spinalcordtoolbox/reports/qc.py
```diff
@@ -264,85 +281,127 @@ def wrapped_f(sct_slice, *args):
...
+
+        w, h = (self.qc_report.qc_params.dpi * size_fig[0], self.qc_report.qc_params.dpi * size_fig[1])
+        self.qc_report.update_description_file((w, h))
+
+    def _func_stretch_contrast(self, img):
+        if self._stretch_contrast_method == "equalized":
+            return self._equalize_histogram(img)
+        else:
+            p2, p98 = np.percentile(img, (2, 98))
+            return skimage.exposure.rescale_intensity(img, in_range=(p2, p98))
```
joshuacwnewton, line 377: I think this might behave differently than the existing code.

* Before, there were exactly 2 options: `equalized` and `contrast_stretching`. If a 3rd value was passed, then an error would occur.
* Now, if a 3rd value is passed, it will fall into the `else:` case, which might be unexpected.

Because of this, I think it would be good to explicitly check the value of `stretch_contrast_method` to ensure that `equalized` and `contrast_stretching` are the only valid options. Maybe in `__init__`, so that it's caught right away?

joshuacwnewton: Also, one small nitpicky thing: it could be good to also split the two `contrast_stretching` lines into a separate method like you did with `self._equalize_histogram`.

joshuacwnewton (changes requested):
Thanks a ton for addressing my feedback! :tada: 

One small request: If possible, instead of one large commit like 5500759, could you keep your changes in individual commits? 

That makes it easier for me to figure out if review comments are "resolved", because I can pick out each individual change and look at it in isolation. :slightly_smiling_face:

spinalcordtoolbox/reports/qc.py
```diff
@@ -62,22 +63,27 @@ class QcImage(object):
                      "#a22abd", "#d58240", "#ac2aff"]
     # _seg_colormap = plt.cm.autumn
 
-    def __init__(self, qc_report, interpolation, action_list, stretch_contrast=True,
-                 stretch_contrast_method='contrast_stretching', angle_line=None):
```
joshuacwnewton, line 66: Ah, one other comment. Could you double check that adding the `process` parameter like this meshes with the existing usage of `QcImage`? For example:

https://github.com/neuropoly/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/reports/qc.py#L565-L566

spinalcordtoolbox/reports/qc.py
```diff
@@ -62,22 +63,27 @@ class QcImage(object):
...
+    def __init__(self, qc_report, interpolation, action_list, process, stretch_contrast=True,
+                 stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
```
joshuacwnewton, line 67: Ah, one other comment. Could you double check that adding the `process` parameter like this meshes with the existing usage of `QcImage`? For example:

https://github.com/neuropoly/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/reports/qc.py#L565-L566

claragiroux: I'm not sure what you mean here! Is it something to do with the order of the parameters? Right now I'm calling `QcImage` like this : 
https://github.com/neuropoly/spinalcordtoolbox/blob/e17b2fc4423ef94087fb10e59802e2c5fdfdfab1/spinalcordtoolbox/reports/qc.py#L693-L694

joshuacwnewton: Ah, my apologies! I was looking at master instead of this branch, and thought there was a call to `QcImage` that hadn't been changed. :facepalm:

spinalcordtoolbox/scripts/sct_dmri_moco.py
claragiroux: I could take a look and try to find where the verbose is changed in `moco_wrapper.` Is there a command to check the global verbose value?

joshuacwnewton: Thank you for offering! I was thinking this over last night, and think I have a rough idea of why it might be. So, I'd be happy to look into this. :slightly_smiling_face: 

I'll try and finish that task + reviewing `slice.py` by the end of today.

joshuacwnewton: Aha. `moco_wrapper` calls the script `sct_apply_transfo`, and that call explicitly sets `verbose=0`:

https://github.com/neuropoly/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/moco.py#L768-L773

joshuacwnewton: My guess is that `sct_apply_transfo` has `-v 0` because it would be too noisy otherwise. (`sct_apply_transfo` gets called repeatedly here)

The problem is, calling `sct_apply_transfo` has the side-effect of changing the global logging level... I'll open a separate issue for this.

joshuacwnewton: Done in https://github.com/neuropoly/spinalcordtoolbox/issues/3341.

@claragiroux Perhaps it could be good to update the `# moco_wrapper seems to change verbose to 0` comments to reference this new issue?

claragiroux: Thanks for the investigation! I'll add the comment to reference this issue when calling `set_global_loglevel(verbose)`.

spinalcordtoolbox/reports/slice.py
```diff
@@ -347,7 +378,10 @@ def get_dim(self, image):
     def get_center(self, img_idx=-1):
         """Get the center of mass of each slice. By default, it assumes that self._images is a list of images, and the
         last item is the segmentation from which the center of mass is computed."""
-        image = self._images[img_idx]
+        if self._image_seg is None:
+            image = self._images[img_idx]
+        else:
+            image = self._image_seg
```
joshuacwnewton, line 384: It might be good to update the help description to reference the new behavior of using the segmented spinal cord by default if it's present.

spinalcordtoolbox/reports/slice.py
```diff
@@ -38,26 +38,34 @@ class Slice(object):
...
         """
-        :param images: list of 3D volumes to be separated into slices.
+        :param images: list of 3D or 4D volumes to be separated into slices.
         """
         logger.info('Resample images to {}x{} mm'.format(p_resample, p_resample))
-        self._images = list()
+        self._images = list()  # 3d volumes
+        self._4d_images = list()  # 4d volumes
+        self._image_seg = None  # for cropping
+        self._absolute_paths = list()  # Used because change_orientation removes the field absolute_path
```
joshuacwnewton, line 47: Adding a link here to https://github.com/neuropoly/spinalcordtoolbox/issues/3304 to remind us to revisit.

spinalcordtoolbox/reports/slice.py
```diff
@@ -270,6 +278,26 @@ def mosaic(self, nb_column=0, size=15, return_center=False):
...
+
+        :return: list of tuples of numpy.ndarray containing the mosaics of each volumes
+        """
+
+        mosaics = list()
+        self._image_seg = self._images[0].copy()  # segmentation used for cropping
+
+        for i, img in enumerate(self._4d_images):
+            # The absolutepath is changed to None after change_orientation see issue #3304
+            img.absolutepath = self._absolute_paths[i]
```
joshuacwnewton, line 292: Adding a link here to https://github.com/neuropoly/spinalcordtoolbox/issues/3304 to remind us to revisit.

spinalcordtoolbox/reports/slice.py
```diff
@@ -38,26 +38,34 @@ class Slice(object):
...
     def __init__(self, images, p_resample=0.6):
         """
-        :param images: list of 3D volumes to be separated into slices.
+        :param images: list of 3D or 4D volumes to be separated into slices.
         """
         logger.info('Resample images to {}x{} mm'.format(p_resample, p_resample))
-        self._images = list()
+        self._images = list()  # 3d volumes
+        self._4d_images = list()  # 4d volumes
+        self._image_seg = None  # for cropping
```
joshuacwnewton, line 46: I have some possible suggestions that could simplify the changes to `Slice`:

* Instead of `self._image_seg`, call it `self._image_ref`
* Set `self._image_ref` to  `self._images[0]` inside `__init__`, then replace any calls to `self._images[0]`.
* Because you no longer need to access  `_images[0]` in `mosaics_through_time`, I think you no longer need to keep both `_images` and `_images_4d`, which lets you use the same variable for both 3D and 4D cases.

spinalcordtoolbox/reports/slice.py
```diff
@@ -294,7 +322,10 @@ def single(self):
         return matrices
 
     def aspect(self):
-        return [self.get_aspect(x) for x in self._images]
+        if len(self._4d_images) == 0:  # For 3D images
+            return [self.get_aspect(x) for x in self._images]
+        else:  # For 4D images
+            return [self.get_aspect(x) for x in self._4d_images]
```
joshuacwnewton, line 328: e.g. this change would no longer be necessary if you re-used `self._images` for the 4D case.

joshuacwnewton (changes requested):
Sorry for the wait! Here's my second pass of review comments. :slightly_smiling_face:

spinalcordtoolbox/reports/qc.py
```diff
@@ -264,85 +283,130 @@ def wrapped_f(sct_slice, *args):
...
+                [images_after_moco, images_before_moco], centermass = func(sct_slice, *args)
                 self._centermass = centermass
+                self._make_QC_image_for_4d_volumes(images_after_moco, images_before_moco)
+
+            else:
+                if self._angle_line is not None:
+                    [img, mask], centermass = func(sct_slice, *args)
+                    self._centermass = centermass
+                else:
+                    img, mask = func(sct_slice, *args)
```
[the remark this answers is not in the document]
claragiroux: I'm quite new to git, is there a way to "remove" changes like this one from the PR, where the only difference is that the code has been moved a few lines?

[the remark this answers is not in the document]
claragiroux: And about 'self._centermass = centermass', it needs to be called before '_make_QC_image_for_4d_volumes' and '_make_QC_image_for_3d_volumes', so I cannot really move it outside!

spinalcordtoolbox/reports/slice.py
claragiroux: I was stuck for a while on this issue when writing 'mosaics_through_time'! I created '_images_4d' mainly to be able to use `mosaic` from without too much changes. My initial version was just like you suggested with 'self._images' being used for both 3D and 4D images, but the code was quite messy with a lot of if/else blocks!

joshuacwnewton: Gotcha! That makes sense... no worries, then! :slightly_smiling_face:

spinalcordtoolbox/reports/qc.py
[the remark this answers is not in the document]
joshuacwnewton: > And about 'self._centermass = centermass', it needs to be called before '_make_QC_image_for_4d_volumes' and '_make_QC_image_for_3d_volumes', so I cannot really move it outside!

Oh, of course! I was mistaken, you're absolutely right. :sweat_smile:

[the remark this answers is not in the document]
joshuacwnewton: > I'm quite new to git, is there a way to "remove" changes like this one from the PR, where the only difference is that the code has been moved a few lines?

If you were working on a team that cared about having a perfectly neat/tidy commit history for PRs, then what you could do is make a commit for the change, then use [`git rebase -i`](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#interactive-rebase) along with the ["fixup"](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history#squash-commits-together) option to combine the new commit with an older commit.

Here at SCT, though, when a PR is merged, all of the commits will get squashed (so that there is a single commit on master representing this PR). Because of that, there's not much benefit to rebasing, so it's OK to just create a new commit that undoes the change.

(One other reason not to use rebases for work that was already pushed is because it can make a PR a little confusing to read through. Because of how rebasing works, the edited commits become entirely new commits, so they will appear at the bottom of the PR, removed from the context of the original discussion.)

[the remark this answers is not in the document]
joshuacwnewton: > I'm quite new to git

Also, I would have never guessed! It's been very nice and easy to collaborate with you. :smiley:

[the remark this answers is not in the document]
claragiroux: Thanks so much! And I definitely learned a lot! 😄

joshuacwnewton (commented):
Thanks for addressing my feedback so quickly! I don't have much else to add. :slightly_smiling_face: 

I do have one last request, though: Could you add a test for this addition? 

At the moment, the easiest way to add a test by adding a 2nd and 3rd command (for `sct_qc -p sct_dmri_moco`/`sct_qc -p sct_fmri_moco`) to the following list:

https://github.com/neuropoly/spinalcordtoolbox/blob/ae361d25b1977ab93cd25d5b891da8c30a9ef82b/testing/test_sct_qc.py#L21

Doing this won't check the results, but it will at least run the command to make sure there are no errors.

(Also, sorry in advance if SCT's testing setup seems a little confusing at the moment... there are some ongoing efforts right now to clean things up: https://github.com/neuropoly/spinalcordtoolbox/issues/3219 and https://github.com/neuropoly/spinalcordtoolbox/issues/3093.)

claragiroux:
> At the moment, the easiest way to add a test by adding a 2nd and 3rd command to the list (for `sct_qc -p sct_dmri_moco`/`sct_qc -p sct_fmri_moco`) that mimics the existing command:

I'm a bit confused as to where I should write the test! Does it go in `tests/test_reports.py`, `sct_testing` or somewhere else?

joshuacwnewton:
> > At the moment, the easiest way to add a test by adding a 2nd and 3rd command to the list (for `sct_qc -p sct_dmri_moco`/`sct_qc -p sct_fmri_moco`) that mimics the existing command:
> 
> I'm a bit confused as to where I should write the test! Does it go in `tests/test_reports.py`, `sct_testing` or somewhere else?

You're right that it's confusing... `tests/test_reports.py` is actually dead code that I want to clean up as part of #3219. :fearful:
Sorry about all of this. :sweat_smile: 

The file you'll want to edit is `testing/test_sct_qc.py`. There's a list called `default_args` that has one set of arguments for `sct_qc`:

https://github.com/neuropoly/spinalcordtoolbox/blob/ae361d25b1977ab93cd25d5b891da8c30a9ef82b/testing/test_sct_qc.py#L21

When `sct_testing` is run, `sct_qc` will be called with each set of arguments included in this list. So, by adding another set of arguments to the list, `sct_qc` can be called to test `sct_dmri_moco` and `sct_fmri_moco`.

claragiroux:
@joshuacwnewton There is currently no segmentation for dmri and fmri images in sct_testing_data and those are needed for the qc. Is there a way to add those segmentations? Otherwise maybe I could I use dummy data? On the wiki it says that some summy data functions are already created, but I couldn't find them! https://github.com/neuropoly/spinalcordtoolbox/wiki/Testing%3A-Datasets#dummy-data)

jcohenadad:
@joshuacwnewton if there are no segmentation, i feel this is a bit out of scope of this PR and should be fixed in subsequent PR (where we update the testing data with seg)

joshuacwnewton:
> @joshuacwnewton There is currently no segmentation for dmri and fmri images in sct_testing_data and those are needed for the qc. Is there a way to add those segmentations? Otherwise maybe I could I use dummy data? On the wiki it says that some dummy data functions are already created, but I couldn't find them! https://github.com/neuropoly/spinalcordtoolbox/wiki/Testing%3A-Datasets#dummy-data)

My apologies. I thought `sct_testing_data` would be sufficient, but I had been looking at `sct_example_data` instead (which contains segmentation for dmri/fmri, but is not easily accessible when writing tests). 

> @joshuacwnewton if there are no segmentation, i feel this is a bit out of scope of this PR and should be fixed in subsequent PR (where we update the testing data with seg)

I agree. I have been thinking about focusing on the testing-related issues after the upcoming release. :slightly_smiling_face:

joshuacwnewton:
I don't have any other review comments to make, then! The only other thing I notice is this comment in the issue description:

> For now I have not implemented the graphic with moco params. It requires adaptation of the QC report layout to be able to display 2 images side by side.

Could we work around this by displaying the moco param chart underneath the slice mosaics? It might not be as compact, but it could save us some trouble. One other benefit to avoiding side-by-side images is that they can also be a bit of a headache when dealing with window sizing and making sure all of the elements fit. (Webdev/UIs are not my specialty, though, so I'm only speaking from limited experience.)

joshuacwnewton:
@jcohenadad Regarding https://github.com/neuropoly/spinalcordtoolbox/pull/3320#issuecomment-819650572, is the moco param  graphic necessary for this PR?

jcohenadad:
> @jcohenadad Regarding #3320 (comment), just to double-check, is the moco param graphic necessary for this PR?

no

joshuacwnewton (approved):
Since https://github.com/neuropoly/spinalcordtoolbox/pull/3253 is currently dealing with the CentOS issues, it's probably easiest to merge this first, then fix d957db7 in that PR.

claragiroux:
> I notice there is one remaining task in the issue description:
> 
> > For now I have not implemented the graphic with moco params. It requires adaptation of the QC report layout to be able to display 2 images side by side.
> 
> Could we work around this by displaying the moco param graphic underneath the slice mosaics? It might not be as compact, but it could save us some trouble.
> 
> Side-by-side layouts can also be a bit of a headache when dealing with window sizing and making sure all of the elements fit. So, if we avoid that, it might make things easier anyway. (Webdev/UIs are not my specialty, though, so I'm only speaking from limited experience.)

Sorry for the delay I totally missed that notification! I will not include the moco params graphic in this PR!

joshuacwnewton:
If there are no other comments, I'd like to merge this, then. :slightly_smiling_face: 

Any other thoughts @jcohenadad @claragiroux?

claragiroux:
> If there are no other comments, I'd like to merge this, then. 🙂
> 
> Any other thoughts @jcohenadad @claragiroux?

Everything is good on my side!

jcohenadad (approved):
all good 👌   thank you for the contribution!

PR #3329, opened 2021-04-01
Implemented weighted median

### LINKED ISSUE #3327: Add median when aggregating metrics across voxels
### Context

Currently users can aggregate metrics using the following options:
~~~
                         - ml: maximum likelihood.
                            This method is recommended for large labels and low noise. Also, this
                            method should only be used with the PAM50 white/gray matter atlas, or
                            with any custom atlas as long as the sum across all labels equals 1, in
                            each voxel part of the atlas.
                          - map: maximum a posteriori.
                            Mean priors are estimated by maximum likelihood within three clusters
                            (white matter, gray matter and CSF). Tract and noise variance are set
                            with flag -p. This method should only be used with the PAM50 white/gray
                            matter atlas, or with any custom atlas as long as the sum across all
                            labels equals 1, in each voxel part of the atlas.
                          - wa: weighted average
                          - bin: binarize mask (threshold=0.5)
                          - max: for each z-slice of the input data, extract the max value for each
                          slice of the input data. (default: wa)
~~~

A [suggestion from a user](https://forum.spinalcordmri.org/t/median-values-instead-of-mean/676) is to add the median.

### Suggestion

Add the option `-method median`, which could compute the median across voxels in the mask (could be soft voxels). We can still output the STD as we do for `wa`.

jcohenadad:
#### GitHub

#### PR contents

## Description

Adds the `-method median` option to `sct_extract_metric`. 

Done:
- [x] Implemented `func_median()` in API [aggregate_slicewise.py](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/aggregate_slicewise.py)
- [x] Added test
- [x] Updated CLI

## Linked issues

Fixes #3327

### DIFF
--- a/spinalcordtoolbox/aggregate_slicewise.py
+++ b/spinalcordtoolbox/aggregate_slicewise.py
@@ -158,7 +158,7 @@ def func_median(data, mask=None, map_clusters=None):
     # Check if mask has an additional dimension (in case it is a label). If so, select the first label
     if mask.ndim == data.ndim + 1:
         mask = mask[..., 0]
-    # data, weights = np.array(data).squeeze(), np.array(weights).squeeze()
+    data, mask = data.reshape(-1), mask.reshape(-1)
     s_data, s_mask = map(np.array, zip(*sorted(zip(data, mask))))
     midpoint = 0.5 * sum(s_mask)
     if any(mask > midpoint):
--- a/spinalcordtoolbox/aggregate_slicewise.py
+++ b/spinalcordtoolbox/aggregate_slicewise.py
@@ -146,6 +146,33 @@ def func_map(data, mask, map_clusters):
     return beta[0], beta
 
 
+def func_median(data, mask=None, map_clusters=None):
+    """
+    Compute weighted median.
+    Code inspired from: https://gist.github.com/tinybike/d9ff1dad515b66cc0d87
+    :param data: nd-array: input data
+    :param mask: (n+1)d-array: input mask
+    :param map_clusters: not used
+    :return:
+    """
+    # Check if mask has an additional dimension (in case it is a label). If so, select the first label
+    if mask.ndim == data.ndim + 1:
+        mask = mask[..., 0]
+    # data, weights = np.array(data).squeeze(), np.array(weights).squeeze()
+    s_data, s_mask = map(np.array, zip(*sorted(zip(data, mask))))
+    midpoint = 0.5 * sum(s_mask)
+    if any(mask > midpoint):
+        w_median = (data[mask == np.max(mask)])[0]
+    else:
+        cs_mask = np.cumsum(s_mask)
+        idx = np.where(cs_mask <= midpoint)[0][-1]
+        if cs_mask[idx] == midpoint:
+            w_median = np.mean(s_data[idx:idx + 2])
+        else:
+            w_median = s_data[idx + 1]
+    return w_median, None
+
+
 def func_ml(data, mask, map_clusters=None):
     """
     Compute maximum likelihood (ML) for the first label of mask.
@@ -424,6 +451,10 @@ def extract_metric(data, labels=None, slices=None, levels=None, perslice=True, p
     elif method == 'wa':
         mask = Metric(data=labels_sum, label=label_struc[id_label].name)
         group_funcs = (('WA', func_wa), ('STD', func_std))
+    # Weighted median
+    elif method == 'median':
+        mask = Metric(data=labels_sum, label=label_struc[id_label].name)
+        group_funcs = (('MEDIAN', func_median), ('STD', func_std))
     # Binarize mask
     elif method == 'bin':
         mask = Metric(data=labels_sum, label=label_struc[id_label].name)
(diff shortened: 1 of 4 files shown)


jcohenadad:
I'm getting this when running with `-method median` on testing data:
~~~
The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
~~~

investigating...

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -146,6 +146,33 @@ def func_map(data, mask, map_clusters):
     return beta[0], beta
 
 
+def func_median(data, mask=None, map_clusters=None):
```
joshuacwnewton, line 149:
```suggestion
def func_median(data, mask, map_clusters=None):
```

Since we use `mask`, I think it isn't optional here.

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -146,6 +146,33 @@ def func_map(data, mask, map_clusters):
...
+    midpoint = 0.5 * sum(s_mask)
+    if any(mask > midpoint):
+        w_median = (data[mask == np.max(mask)])[0]
+    else:
+        cs_mask = np.cumsum(s_mask)
+        idx = np.where(cs_mask <= midpoint)[0][-1]
+        if cs_mask[idx] == midpoint:
+            w_median = np.mean(s_data[idx:idx + 2])
+        else:
+            w_median = s_data[idx + 1]
```
joshuacwnewton, line 172: I tried replacing this snippet with a small package called [`wquantiles`](https://github.com/nudomarinero/wquantiles) (I looked over the code and it seems well-tested + has additional input validation, so I thought it'd be nice for double-checking.) 

```suggestion
    w_median = wquantiles.median(data=data, weights=mask)
```

I got slightly different results! 

```
>       assert agg_metric[list(agg_metric)[0]]['MEDIAN()'] == 39.0
E       assert 39.5 == 39.0
```

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -146,6 +146,33 @@ def func_map(data, mask, map_clusters):
...
+    return w_median, None
```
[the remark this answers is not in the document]
jcohenadad: > I have ideas about how to handle this more elegantly, but that would be a slightly more involved refactor outside the scope of this PR. So, all good for now. 👍

yes, i agree the code is cumbersome to handle and understand.

spinalcordtoolbox/aggregate_slicewise.py
jcohenadad: oopsi! that's a problem. OK, i need to investigate further.

jcohenadad: hum, so I downloaded `wquantiles` and tried this example from the definition of the weighted median on wikipedia and got 3.6 instead of the 4:
```python
>>> data = np.array([1, 2, 3, 4, 5])
>>> mask = np.array([0.15, 0.1, 0.2, 0.3, 0.25])
>>> wquantiles.median(data, mask)
3.6
```
ok... this is wikipedia, so to consider with a grain of salt... i continue my investigations...

jcohenadad: Trying a comparison of `wquantiles` with another Python library for weighted statistics: [robustats](https://github.com/FilippoBovo/robustats)
```
>>> x = np.array([1.1, 5.3, 3.7, 2.1, 7.0, 9.9])
>>> weights = np.array([1.1, 0.4, 2.1, 3.5, 1.2, 0.8])
>>> wquantiles.median(x, weights)                                          
3.0714285714285716  # ⚠️ should be # 2.1
```

jcohenadad: and the wikipedia example with robustats:
```python
>>> data = np.array([1, 2, 3, 4, 5])
>>> mask = np.array([0.15, 0.1, 0.2, 0.3, 0.25])
>>> robustats.weighted_median(data, mask)
4.0
```

jcohenadad: here is the explanation for the discrepancies: https://github.com/nudomarinero/wquantiles/issues/4

jcohenadad: OK, after giving it some thoughts, I think I prefer the non-discrete implementation of the median. So I will go with the **wquantiles** package. Thank you for finding it @joshuacwnewton, and for raising this debate over the definition of the median. I am better informed now 😅

joshuacwnewton: Thank YOU for taking the time to look deeply into this! 

One last nit: I notice you've updated the docstring of the function, but perhaps we should also be explicit about the chosen method in the argparse description for `-method median`, too.

joshuacwnewton (approved):
Giving pre-approval as the one last comment I gave above is just a nit. :slightly_smiling_face:

spinalcordtoolbox/aggregate_slicewise.py
jcohenadad: excellent point, fixed in https://github.com/neuropoly/spinalcordtoolbox/pull/3329/commits/af686d5532143b56a5b65f95a2dc0cbc57a494af

PR #3339, opened 2021-04-09
Fully isolate the conda env by its site.py

kousu:
## Motivation

The fix for https://github.com/neuropoly/spinalcordtoolbox/issues/3067 turned out to be pretty complicated: https://github.com/neuropoly/spinalcordtoolbox/issues/3200

This is a cleaner fix for that issue, inspired by @takluyver's helpful observation https://github.com/conda/conda/issues/7173#issuecomment-389190657 that `site.py` [is the canonical place to manage this sort of thing](https://docs.python.org/3/library/site.html#site.ENABLE_USER_SITE) -- @takluyver, I know you mitigated your opinion later while helping users, but for us, we're using conda to build a sort of .appimage or .app/ folder, with protected, pinned dependencies, and intentionally not exposing conda to our users, so isolating the entire python site is *exactly* what we need!

This turns a 7 line fix into a 1 line fix: just set `site.ENABLE_USER_SITE` for the env. That isolates it to python programs run in the env, and can't leak accidentally.

## Proof

To demonstrate this is right, I ran `sct_run_batch` because that's what we had problems with in #3200, feeding it this script:

```
$ cat ../hello.sh 
#!/bin/sh

(
set | grep PATH=
python3 -c 'import sys; print(sys.path)'
) | tee /tmp/hello.log
```

I also had to make a dataset it would accept; I culled https://github.com/spine-generic/data-multi-subject/ down to just

```
$ ls -l ../datalad/spine-generic/data-multi-subject/
total 128
-rwxr-xr-x 1 kousu kousu  3087 Feb  2 15:10 CHANGES.md
-rwxr-xr-x 1 kousu kousu  3467 Aug 16  2020 dataset_description.json
drwxr-xr-x 3 kousu kousu  4096 Aug 16  2020 derivatives
-rw-r--r-- 1 kousu kousu  2922 Feb  2 15:16 exclude.yml
-rw-r--r-- 1 kousu kousu  1078 Feb  2 15:10 LICENSE
-rwxr-xr-x 1 kousu kousu  1175 Feb  2 15:16 participants.json
-rwxr-xr-x 1 kousu kousu 41123 Feb  2 15:16 participants.tsv
-rwxr-xr-x 1 kousu kousu  1952 Feb  2 15:10 README.md
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl01
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl02
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl03
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl04
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl05
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-ucl06
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf01
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf02
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf03
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf04
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf05
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf06
drwxr-xr-x 4 kousu kousu  4096 Aug 16  2020 sub-unf07
```

This is the command I used to test:

```
$ ./python/envs/venv_sct/bin/sct_run_batch -script ../hello.sh -path-data ../datalad/spine-generic/data-multi-subject/
$ cat /tmp/hello.log
```

0. Reinstall

    ```
    $ git checkout master
    $ rm -rf python
    $ ./install_sct -y
    ```

There are four conditions to consider: what happens when running `sct_run_batch` inside `venv_sct`, what happens outside? x 
what happens before the patch, and what happens after?

(`master` == 3f4b37e09448ba23e9571c4f77645412621c6106 here)

1. *out* of venv *before* patch

    ```
    $ ./python/envs/venv_sct/bin/sct_run_batch -script ../hello.sh -path-data ../datalad/spine-generic/data-multi-subject/
    $ cat /tmp/hello.log
    PATH=/opt/miniconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
    ['', '/usr/lib/python39.zip', '/usr/lib/python3.9', '/usr/lib/python3.9/lib-dynload', '/home/kousu/.local/lib/python3.9/site-packages', '/home/kousu/src/xmpp/bridges/xmpp-smpp', '/home/kousu/src/neuropoly/sysadmin/infersent-data-model1/src', '/usr/lib/python3.9/site-packages']
    ```

2. *in* venv *before* patch

    ```
    $ . python/bin/activate 
    (base) $ conda activate venv_sct
    (venv_sct) $ ./python/envs/venv_sct/bin/sct_run_batch -script ../hello.sh -path-data ../datalad/spine-generic/data-multi-subject/
    (venv_sct) $ cat /tmp/hello.log
    PATH=/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin:/opt/miniconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
    ['', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python36.zip', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/lib-dynload', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/site-packages', '/home/kousu/src/neuropoly/spinalcordtoolbox']
    ```

    ```
    (venv_sct) $ conda deactivate
    (base) $ conda deactivate
    $
    ```

3. *out* of venv *after* patch

    Note: I cheated here; I didn't do a fresh reinstall but patched the existing conda env directly, but I'm 99% sure that this is accurate:

    ```
    $ git checkout ng/nousersite
    $ sed -i 's/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/' "python/envs/venv_sct/lib/python3.6/site.py"
    $ grep ENABLE_USER_SITE "python/envs/venv_sct/lib/python3.6/site.py" | head -n 1 # verify the patch
    ENABLE_USER_SITE = False
    $ ./python/envs/venv_sct/bin/sct_run_batch -script ../hello.sh -path-data ../datalad/spine-generic/data-multi-subject/
    $ cat /tmp/hello.log
    PATH=/opt/miniconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
    ['', '/usr/lib/python39.zip', '/usr/lib/python3.9', '/usr/lib/python3.9/lib-dynload', '/home/kousu/.local/lib/python3.9/site-packages', '/home/kousu/src/xmpp/bridges/xmpp-smpp', '/home/kousu/src/neuropoly/sysadmin/infersent-data-model1/src', '/usr/lib/python3.9/site-packages']
    ```

4. *in* venv *after* patch

    ```
    $ . python/bin/activate 
    (base) $ conda activate venv_sct
    (venv_sct) $ ./python/envs/venv_sct/bin/sct_run_batch -script ../hello.sh -path-data ../datalad/spine-generic/data-multi-subject/
    $ cat /tmp/hello.log
    PATH=/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/bin:/opt/miniconda3/condabin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
    ['', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python36.zip', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/lib-dynload', '/home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/venv_sct/lib/python3.6/site-packages', '/home/kousu/src/neuropoly/spinalcordtoolbox']
    ```

## Analysis

The environment within the venv is identical in both old and new, and same with outside. So this patch seems to safely accomplish the same goal.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -325,6 +325,17 @@ if [ "$(uname)" = "Darwin" ]; then
   (command -v realpath >/dev/null) || realpath() {
     python3 -c 'import sys, os; [print(os.path.realpath(f)) for f in sys.argv[1:]]' "$@"
   }
+
+
+  # glue over the difference between how BSD and GNU sed work with -i
+  # https://code-examples.net/en/q/56e314
+  sed_i() {
+    sed -i '' "$@"
+  }
+else
+  sed_i() {
+    sed -i "$@"
+  }
 fi
 
 # ======================================================================================================================
@@ -557,9 +568,6 @@ fi
 # Install Python
 # ----------------------------------------------------------------------------------------------------------------------
 
-# We make sure that there is no conflict with local python install by unsetting PYTHONPATH and forcing PYTHONNOUSERSITE
-unset PYTHONPATH
-export PYTHONNOUSERSITE=1
 
 # Remove old python folder
 print info "Installing conda..."
@@ -581,6 +589,15 @@ run bash "$TMP_DIR/miniconda.sh" -p "$SCT_DIR/$PYTHON_DIR" -b -f
 
 # create py3.6 venv (for Keras/TF compatibility with Centos7, see issue #2270)
 python/bin/conda create -y -n venv_sct python=3.6
+
+# make sure that there is no conflict with local python install by making venv_sct an isolated environment.
+# workaround for https://github.com/conda/conda/issues/7173
+# see also
+# * https://github.com/neuropoly/spinalcordtoolbox/pull/3067
+# * https://github.com/neuropoly/spinalcordtoolbox/issues/3200
+# this needs to be added very early in python's boot process
+# so using sitecustomize.py or even just appending to the file are impossible.
+sed_i 's/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/' "$SCT_DIR/$PYTHON_DIR/envs/venv_sct/lib/python"*"/site.py"
 
 # activate miniconda
 # shellcheck disable=SC1091
--- a/spinalcordtoolbox/compat/launcher.py
+++ b/spinalcordtoolbox/compat/launcher.py
@@ -24,9 +24,6 @@ def main():
     if "ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS" not in os.environ:
         env["ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS"] = str(multiprocessing.cpu_count())
 
-    # Prevent user site packages from interfering with SCT dependencies (See issue #3067)
-    env["PYTHONNOUSERSITE"] = "True"
-
     command = os.path.basename(sys.argv[0])
     pkg_dir = os.path.dirname(sct.__file__)
 
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -196,9 +196,6 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
         'ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS': str(itk_threads),
         'SCT_PROGRESS_BAR': 'off'
     })
-    # PYTHONNOUSERSITE is originally set in compat/launcher.py. We unset here so that it doesn't affect non-SCT commands
-    # in user's pipeline scripts. See: https://github.com/neuropoly/spinalcordtoolbox/issues/3200.
-    envir.pop('PYTHONNOUSERSITE', None)
 
     # Ship the job out, merging stdout/stderr and piping to log file
     try:


kousu:
Tagging @ahill187 and @dyt811 too because this is definitely relevant to things we've been talking about for shimming-toolbox and ivadomed. Take a skim if you can, I hope it is enlightening.

kousu:
## Didacts

I agree with @joshuacwnewton in https://github.com/conda/conda/issues/7173#issuecomment-389190657 -- this is a bug in conda. With `virtualenv` you need to supply `--system-site-packages` to get this behaviour, whereas conda misleadingly has the opposite default, despite relying on almost all the same mechanisms as virtualenv. Here are all the threads we've found about this:

* https://github.com/conda/conda/issues/7173 (still open)
* https://github.com/mozilla/OpenWPM/issues/689
* https://github.com/conda-forge/python-feedstock/issues/171
* https://github.com/conda/conda/issues/448



So there's definitely an isolation failure here. But I am confused about the original bug though. From the logs we were given, it doesn't seem like it should have happened. In https://forum.spinalcordmri.org/t/attributeerror-str-object-has-no-attribute-decode-v5-0-0/567/12 we see from the install log:

```
Collecting h5py==2.10.0
  Using cached h5py-2.10.0-cp36-cp36m-manylinux1_x86_64.whl (2.9 MB)
[...]
Requirement already satisfied: wheel>=0.26; python_version >= "3" in ./python/envs/venv_sct/lib/python3.6/site-packages (from tensorboard==2.4.0->-r requirements-freeze.txt (line 80)) (0.35.1)
Installing collected packages: six, absl-py, [...], h5py, [...]
Successfully installed Keras-2.1.5 [...] h5py-2.10.0 html5lib-0.9999999 [...]
```

which makes it look like it should have installed h5py to ./python/envs/venv_sct/lib/python3.6/site-packages. Then we see furthermore that the `venv_sct/` *has precendence over ~/.local/*:

```
extop@b:~$ $SCT_DIR/python/envs/venv_sct/bin/python
Python 3.6.12 |Anaconda, Inc.| (default, Sep  8 2020, 23:10:56) 
[GCC 7.3.0] on linux
>>> import sys
>>> sys.path
['', '/home/extop/sct_5.0.1/python/envs/venv_sct/lib/python36.zip', '/home/extop/sct_5.0.1/python/envs/venv_sct/lib/python3.6', '/home/extop/sct_5.0.1/python/envs/venv_sct/lib/python3.6/lib-dynload', '/home/extop/.local/lib/python3.6/site-packages', '/home/extop/sct_5.0.1/python/envs/venv_sct/lib/python3.6/site-packages', '/home/extop/sct_5.0.1']
```

but for some reason the h5py from `~/.local` got used instead?

```
>>> import h5py
>>> h5py.__version__
'3.1.0'
>>> import importlib
>>> importlib.util.find_spec('h5py')
ModuleSpec(name='h5py', loader=<_frozen_importlib_external.SourceFileLoader object at 0x7f595a1eb5f8>, origin='/home/extop/.local/lib/python3.6/site-packages/h5py/__init__.py', submodule_search_locations=['/home/extop/.local/lib/python3.6/site-packages/h5py'])
>>> exit()
```

So was pip lying about installing h5py? Why didn't it say "requirement already satisfied" if it was going to do that? Did it install it but for some reason python is refusing to load that version and is instead falling back to the other one?

So, maybe there are two bugs here? One is just that conda has the opposite default from virtualenv and they should probably add `--system-site-packages` too, but also that even in that mode somehow their version of python silently swaps the order of sys.path?? That seems crazy. D


## Cloud Rap

`virtualenv` has taken care of making sure that *because* `sct_run_batch` and all the rest are in the venv, their shebangs are hardcoded to the venv:

```
$ head -n 2 venv_sct2/bin/sct_run_batch 
#!/home/kousu/src/neuropoly/spinalcordtoolbox/venv_sct2/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'spinalcordtoolbox','console_scripts','sct_run_batch'
```

In our installer we symlink all of these over to $SCT_DIR/bin/ and add *that* to the user's PATH **instead of** adding `venv_sct/bin`; this is important. It means users can run our scripts but without accidentally using our python for their own scripts, and it *should* mean that tools like `gradunwrap` that are also packaged sketchily (#3200) can still remain compatible with us.

takluyver:
> @takluyver, I know you mitigated your opinion later while helping users, but for us, we're using conda to build a sort of .appimage or .app/ folder, with protected, pinned dependencies, and intentionally not exposing conda to our users

Of course - I was thinking about what conda should do by default, but what you describe does sound like an excellent case for disabling user site packages. The [`-I` flag](https://docs.python.org/3/using/cmdline.html#id2) for 'isolated mode' might also be useful if you haven't already found it.

joshuacwnewton (approved):
So, if I'm understanding correctly...

* **Current approach:** Set/unset `PYTHONNOUSERSITE` environment variable, which has the effect of changing `site.ENABLE_USER_SITE`, which then impacts how `sys.path` is created. 
    * This effect is active for any `python` instance that can see `PYTHONNOUSERSITE`.
* **New approach:** Skip directly to the source, patching in a new default value for [`ENABLE_USER_SITE` in `site.py`](https://github.com/python/cpython/blob/6f37ebc61e9e0d13bcb1a2ddb7fc9723c04b6372/Lib/site.py#L80-L82).
    * This is active only for `venv_sct`'s python instance.

-----

Mea culpa! I was convinced `import site; site.ENABLE_USER_SITE = True` wasn't applicable, because I assumed `site.ENABLE_USER_SITE` was just a status indicator and had no effect.

It's interesting to learn that it [_does_ have an effect](https://github.com/python/cpython/blob/6f37ebc61e9e0d13bcb1a2ddb7fc9723c04b6372/Lib/site.py#L331-L344), but only if it's set prior to initialization:

```python
# from site.py
def addusersitepackages(known_paths):
    """Add a per user site-package to sys.path

    Each user has its own python directory with site-packages in the
    home directory.
    """
    # get the per user site-package path
    # this call will also make sure USER_BASE and USER_SITE are set
    user_site = getusersitepackages()

    if ENABLE_USER_SITE and os.path.isdir(user_site):
        addsitedir(user_site, known_paths)
    return known_paths
```

There's probably a lesson here for me that digging into `stdlib` source code is sometimes worth doing, rather than solely trying to infer behavior from documentation. 

I'm really glad you kept digging, so thank you. :slightly_smiling_face:

kousu:
> Of course - I was thinking about what conda should do by default, but what you describe does sound like an excellent case for disabling user site packages. The [`-I` flag](https://docs.python.org/3/using/cmdline.html#id2) for 'isolated mode' might also be useful if you haven't already found it.

No actually I hadn't noticed `-I`! Thanks for the tip. But unless pip can install launchers that use it I wouldn't know how to make use of it.

Do you think conda is going to figure out how to handle this better? It's an obscure pitfall but pretty hard to get out of if you fall into it.

kousu:
For posterity, another bug conda has around its environment handling was this: layering multiple conda installs is all kinds of glitchy: https://github.com/neuropoly/spinalcordtoolbox/issues/3029. It assumes you won't do that, but it never checks that you haven't.

Issue #3227, opened 2021-02-18
Plan for Python 3.6 EOL (December 2021)


joshuacwnewton:
Python 3.6 will no longer receive security fixes [after 2021-12](https://www.python.org/dev/peps/pep-0494/#and-beyond-schedule).

Considering SCT relies exclusively on Python 3.6 to support a specific older version of Tensorflow (see #3046), I'm creating this issue to (at the very least) put this on our radar well in advance.

joshuacwnewton:
(This seems like a pin-worthy issue to me, but let me know if I'm jumping the gun here.)

joshuacwnewton:
I'm still keeping this in the back of my mind. I'm worried that without a plan, SCT might stall on this right up until December 2021, then be forced to scramble.

To start, I've been trying to brainstorm some unpolished ideas for the two strategies mentioned:

> Python 3.7 support would require at least `tensorflow==1.13.1`, or dropping Tensorflow entirely (see #3046).

### Ideas for upgrading Tensorflow

This would be a sort of "intermediate step" -- a bridge between where SCT is now, and a future PyTorch-only version of SCT (i.e. it would buy us time). If we go this route, then we would need a plan for the portion of our userbase whose CPUs don't support AVX (#2562/#2618).

1. Only support `sct_deepseg_<x>` for AVX users, and tell any users who don't have AVX support to switch to `sct_deepseg` moving forward.
2. Follow instructions from this comment: https://github.com/neuropoly/spinalcordtoolbox/issues/2562#issuecomment-686394915 (I missed this comment entirely until just now. This person is suggesting that our Keras models can be used on AVX-less machines even after upgrading Tensorflow. Definitely requires further investigation.)
3. Try to support AVX-less users by using a Tensorflow wheel that doesn't have an AVX dependency (e.g. [`tensorflow-community-wheels`](https://github.com/yaroslavvb/tensorflow-community-wheels))

### Ideas for dropping Tensorflow

IMO this is a long-term goal worth striving for. Notably, though, these ideas would have an impact on our entire user base (rather than just the AVX-less users). But, it might be worth it to "rip off the bandaid" so to speak, rather than devote time to intermediate steps we'll eventually abandon anyway.

4. Drop `sct_deepseg_<x>` entirely (breaking backwards compatibility). Then, encourage users to update their scripts to use `sct_deepseg` moving forward (i.e. #3046)
[...]

Issue #3095, opened 2020-12-14
Split "Command-line Tools" RTD page into 1 page per CLI script


joshuacwnewton:
Previously, in our old [SourceForge](https://sourceforge.net/p/spinalcordtoolbox/wiki/) documentation, we had a dedicated [Tools](https://sourceforge.net/p/spinalcordtoolbox/wiki/tools/) page containing subpages for each SCT script.

By comparison, on ReadTheDocs, we have a single ([Command-Line Tools](https://spinalcordtoolbox.com/en/stable/user_section/command-line.html)) page that contains *all* of the CLI scripts. I think it could be better to split this page up once more and have 1 page per script, for the following reasons:

* The existing page is very lengthy, so it is hard to skim through.
* Our argparse help descriptions are getting very long for some scripts. If we had dedicated RTD pages for each script, we could move some of this information to the script's webpage instead, making `-h` more readable.
    * Lengthy `-h` example: [`sct_register_multimodal`](https://spinalcordtoolbox.com/en/stable/user_section/command-line.html#sct-register-multimodal)
* We have existing "Advanced Usage Tips" pages for some commands. These pages are currently in the "Tutorials" section, but we could move these tips and tricks to each script's dedicated page. That way, the usage information for script is centralized in 1 location.
    * [`sct_propseg`](https://spinalcordtoolbox.com/en/stable/user_section/tutorials/correcting_sct_propseg.html)
    * [`sct_register_to_template`](https://spinalcordtoolbox--3094.org.readthedocs.build/en/3094/user_section/tutorials/customizing_registration.html)

jcohenadad:
@joshuacwnewton this is a brilliant idea. Agreed with all the points you raised. 

Regarding the readability of some usages, eg [sct_register_multimodal](https://spinalcordtoolbox.com/en/stable/user_section/command-line.html#sct-register-multimodal), I am wondering if we could maybe come up with a flag, that would set the number of chars we want to display for the usage, and we would use that flag when generating the content on RTD (currently it's set for 80chars (i think), and we should definitely increase it).

joshuacwnewton:
When you say "number of chars", you're referring to the width/length of each line, right? 

This is a bit tricky, because argparse line length is set when the argparse text formatter is first initialized:

```python
    parser = argparse.ArgumentParser(
        formatter_class=HelpFormatter(width=80)  # Set to value >80
    )
```

[...]

PR #3340, opened 2021-04-09
Don't install all of SCT just for docs

kousu:
I had a spurious failure in https://github.com/neuropoly/spinalcordtoolbox/pull/3339 because in https://readthedocs.org/projects/spinalcordtoolbox/builds/13465331/ ([full log](https://github.com/neuropoly/spinalcordtoolbox/files/6284219/13465331.txt))

> 
> ```
> Collecting joblib
>  Downloading joblib-1.0.1-py3-none-any.whl (303 kB)
> WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))': /whl/cpu/torch_stable.html
> ERROR: Could not install packages due to an OSError: ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))
> ```

We don't need to make readthedocs install pytorch and all the rest just to run sphinx.


kousu:
Okay actually you do need all of SCT just for docs. All the code gets loaded so that its `__doc__`s can be read (by https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html). But I found https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports which alleviates this (though imperfectly: here's someone else working on scientific computing struggling the same way https://github.com/readthedocs/readthedocs.org/issues/5512), and more difficult to work around, we use [`program-output:: sct_<something> -h`](https://sphinxcontrib-programoutput.readthedocs.io/en/latest/) to get output, and automock doesn't survive being called across a subroutine like that.

jcohenadad:
thank you for looking into this @kousu !

joshuacwnewton:
> we use [`program-output:: sct_<something> -h`](https://sphinxcontrib-programoutput.readthedocs.io/en/latest/) to get output, and automock doesn't survive being called across a subroutine like that.

Curious thought: Our scripts now support being called using:

```python
from spinalcordtoolbox.scripts import sct_<something>
sct_<something>.main(argv=[])
```

(This wasn't possible when first we started using `programoutput`.) 

So, I'm wondering if we could make use of passing `-h` in a function call to `main` to get around automock's limitations.

kousu:
Yeah, totally! I asked about this over at `program-output::` and they don't think it can be done with their tool https://github.com/NextThought/sphinxcontrib-programoutput/issues/54 so I think that trying to run the help text in the same process is the way to go.

kousu:
Ah we can do this:


```
(venv_sct) [kousu@requiem spinalcordtoolbox]$ python
Python 3.6.13 |Anaconda, Inc.| (default, Feb 23 2021, 21:15:04) 
[GCC 7.3.0] on linux
>>> import io
>>> import spinalcordtoolbox.scripts.sct_dmri_moco
>>> s = io.StringIO()
>>> spinalcordtoolbox.scripts.sct_dmri_moco.get_parser().print_help(s)
>>> s
<_io.StringIO object at 0x7fa4ec50d048>
>>> str(s)
'<_io.StringIO object at 0x7fa4ec50d048>'
>>> print(s.getvalue())
usage: sct_dmri_moco -i <file> -bvec <file> [-h] [-bval <file>] [-bvalmin <float>] [-g <int>] [-m <file>] [-param <list>] [-x {nn,linear,spline}] [-ofolder <folder>] [-r {0,1}] [-v <int>]

Motion correction of dMRI data. Some of the features to improve robustness were proposed in Xu et al. (http://dx.doi.org/10.1016/j.neuroimage.2012.11.014) and include:
  - group-wise (-g)
  - slice-wise regularized along z using polynomial function (-param). For more info about the method, type: isct_antsSliceRegularizedRegistration
  - masking (-m)
  - iterative averaging of target volume

MANDATORY ARGUMENTS:
  -i <file>             Diffusion data. Example: dmri.nii.gz
  -bvec <file>          Bvecs file. Example: bvecs.txt

OPTIONAL ARGUMENTS:
  -h, --help            Show this help message and exit.
  -bval <file>          Bvals file. Example: bvals.nii.gz
  -bvalmin <float>      B-value threshold (in s/mm2) below which data is considered as b=0. Example: 50.0 (default: 100)
  -g <int>              Group nvols successive dMRI volumes for more robustness. Example: 2 (default: 3)
  -m <file>             Binary mask to limit voxels considered by the registration metric. Example: dmri_mask.nii.gz
  -param <list>         Advanced parameters. Assign value with "=", and separate arguments with ",".
                          - poly [int]: Degree of polynomial function used for regularization along Z. For no regularization set to 0. Default=2.
                          - smooth [mm]: Smoothing kernel. Default=1.
                          - metric {MI, MeanSquares, CC}: Metric used for registration. Default=MI.
                          - gradStep [float]: Searching step used by registration algorithm. The higher the more deformation allowed. Default=1.
                          - sample [None or 0-1]: Sampling rate used for registration metric. Default=None.
  -x {nn,linear,spline}
                        Final interpolation. (default: spline)
  -ofolder <folder>     Output folder. Example: dmri_moco_results/
  -r {0,1}              Remove temporary files. 0 = no, 1 = yes (default: 1)
  -v <int>              Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode (default: 1)

```

So we should be able to replace all the lines in https://github.com/neuropoly/spinalcordtoolbox/blob/6569ce50c4deeb0a6a3eaef26a8e233ab23f21d8/documentation/source/user_section/command-line.rst#main-tools

like this:

```diff

 sct_analyze_lesion
 ==================
 
-.. program-output:: sct_analyze_lesion -h
+.. <something>:: <something> s=io.StringIO(); __import__('spinalcordtoolbox.scripts.sct_analyze_lesion').get_parser().print_help(s); s.getvalue()

But I don't know sphinx so I don't know how to get it to do that. @joshuacwnewton do you have any experience with get sphinx to output plain <code> blocks?

kousu:
Maybe https://pypi.org/project/sphinx_execute_code/ ?

I wonder what the alternatives this is trying to replace are. Maybe one of them would work for us.

joshuacwnewton:
Actually, I wonder if now might be the time to explore https://sphinx-argparse.readthedocs.io/en/stable/. 

I've been curious about this, because it properly formats the output in nice html tables and what have you, but I haven't had a good reason to go down this path until now. Since we'd be replacing these lines anyway, what do you think?

kousu:
Oh sweet I'll give that one a shot. Two shakes.

kousu:
Attempt #1: blank output :(

2021-04-09-134942_450x202_scrot

joshuacwnewton:
> Attempt #1 with sphinx-argparse: blank output :(
> 
> 2021-04-09-134942_450x202_scrot

Oh dear...

As a fallback, I also see that `sphinx-argparse` mentions this project as an alternative: https://sphinxcontrib-autoprogram.readthedocs.io/en/stable/

kousu:
Right now I'm guessing the docs are just confusing. They [say](https://sphinx-argparse.readthedocs.io/en/stable/usage.html#basic-usage) 

>    :func: my_func_that_returns_a_parser

but [their examples](https://sphinx-argparse.readthedocs.io/en/stable/sample.html#source-of-example-file) show that actually `:func:` is supposed to be a pre-created parser instance.

kousu:
Weird. I don't know what's up. I'm sharing what I've tried here: https://github.com/neuropoly/spinalcordtoolbox/pull/3340/commits/1f68b5d06bf8bce594b0f2140eb946570923524b

I copied their example and it works:

https://github.com/neuropoly/spinalcordtoolbox/pull/3340/commits/1f68b5d06bf8bce594b0f2140eb946570923524b#diff-2e0d6a7235f1278c231a3e82a06db174946a9a4da865efe333aaad3c2ca1a94fR128

but ours doesn't:

https://github.com/neuropoly/spinalcordtoolbox/pull/3340/commits/1f68b5d06bf8bce594b0f2140eb946570923524b#diff-2e0d6a7235f1278c231a3e82a06db174946a9a4da865efe333aaad3c2ca1a94fR124

it just skips from "sct_analyze_lesion" to the sample:

2021-04-09-141125_769x391_scrot

kousu:
Oh no, autoprogram needs python3.7: https://github.com/sphinx-contrib/autoprogram/issues/47 annnnd we're not there yet ( #3227  )

kousu:
I don't think `autodoc_mock_imports` is compatible with either argparse:: or autoprogram:: :(. The mocks only trigger for the duration of something handled by the `sphinx.ext.autodoc`.

I'm going to report bugs to both of them. That's unfortunate!

kousu:
Still, migrating to `autoprogram` is easy and makes prettier docs, so we should put that on the list.

joshuacwnewton:
> Oh no, autoprogram needs python3.7: [sphinx-contrib/autoprogram#47](https://github.com/sphinx-contrib/autoprogram/issues/47) annnnd we're not there yet ( #3227 )

https://github.com/neuropoly/spinalcordtoolbox/pull/3361 :tada:

jcohenadad:
> The good news? sphinx-argparse now works as intended, and it looks great!

indeed!

joshuacwnewton:
This PR was originally meant figure out how to build the docs on RTD without actually installing SCT.

We tried 3 different alternatives  in this PR ( `program-output`, `sphinx-argparse`, `autoprogram`). However, all three required us to install SCT first, which would defeat the original purpose of the PR.

---

Even though the original purpose of the PR did not pan out, the PR still acted as a great testing ground for `sphinx-argparse`. 

So, the purpose of this PR morphed into:

1. Adopting the `sphinx-argparse` package to improve the readability of our CLI script argparse help pages.
2. Splitting the command-line pages into 1 page per CLI script (since this was an existing open issue ([#3095](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3095)), and the effort is very closely tied to adopting `sphinx-argparse`).

----

However, actually adopting `sphinx-argparse` is a _huge_ undertaking. It requires editing the argparse help descriptions for _all 49 CLI scripts_ so that they display correctly in both RTD _and_ in the command line. 

- You can see an example of the tweaks that are required in [a9d53a6](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3340/commits/a9d53a6fbce89995205d979619c5c326fa5fa8cb). 
- The end result is [here](https://spinalcordtoolbox--3340.org.readthedocs.build/user_section/command-line.html#sct-analyze-lesion).

I think this end result is quite worth the effort, but whether we finish up the work here depends on whether someone is willing to devote the time to proofread and format every single argparse help description.

----

Given the fact that this PR has veered far from its original purpose, I'm going to close this PR, then leave tasks 1. and 2. for separate PRs. That way, we don't have to wade through the debugging comments when reviewing this sort of work.

PR #3351, opened 2021-04-21
Update moco commands in `batch_processing.sh` to include QC

### LINKED ISSUE #3350: Update `batch_processing.sh` to include QC output for sct_dmri_moco/sct_fmri_moco
Just realized we should have done this as part of #3320. Should be a quick PR to open and approve.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Follow up to #3320 to add the corresponding QC commands to `sct_dmri_moco`/`sct_fmri_moco` in `batch_processing.sh`.

## Linked issues

Fixes #3350.


batch_processing.sh
```diff
@@ -199,6 +199,8 @@ sct_deepseg_sc -i dmri_crop_moco_dwi_mean.nii.gz -c dwi -qc "$SCT_BP_QC_FOLDER"
 # Register template to dwi
 # Tips: Again, here, we prefer to stick to segmentation-based registration. If there are susceptibility distortions in your EPI, then you might consider adding a third step with bsplinesyn or syn transformation for local adjustment.
 sct_register_multimodal -i $SCT_DIR/data/PAM50/template/PAM50_t1.nii.gz -iseg $SCT_DIR/data/PAM50/template/PAM50_cord.nii.gz -d dmri_crop_moco_dwi_mean.nii.gz -dseg dmri_crop_moco_dwi_mean_seg.nii.gz -param step=1,type=seg,algo=centermass:step=2,type=seg,algo=bsplinesyn,metric=MeanSquares,smooth=1,iter=3 -initwarp ../t2/warp_template2anat.nii.gz -initwarpinv ../t2/warp_anat2template.nii.gz -qc "$SCT_BP_QC_FOLDER"
+# Use destination template seg file ('-dseg') to generate QC for sct_dmri_moco
+sct_qc -i dmri_crop.nii.gz -d dmri_crop_moco.nii.gz -s dmri_crop_moco_dwi_mean_seg.nii.gz -p sct_dmri_moco -qc "$SCT_BP_QC_FOLDER"
```
jcohenadad, line 203: You could also get the QC directly from [this line](https://github.com/neuropoly/spinalcordtoolbox/pull/3351/files#diff-86da3fafb905747ce12a7fbc7249f6f44cab684e674a0800cb25edcaf0257a79R193) with:
~~~
sct_crop_image -i dmri.nii.gz -m mask_dmri_mean.nii.gz -o dmri_crop.nii.gz -qc "$SCT_BP_QC_FOLDER" -qc-seg t2_seg_reg.nii.gz
~~~

batch_processing.sh
```diff
@@ -234,6 +236,8 @@ sct_fmri_moco -i fmri_crop.nii.gz -g 1
 # and hence in this case we do it manually. The file is called: fmri_crop_moco_mean_seg_manual.nii.gz
 # Register template->fmri
 sct_register_multimodal -i $SCT_DIR/data/PAM50/template/PAM50_t2.nii.gz -iseg $SCT_DIR/data/PAM50/template/PAM50_cord.nii.gz -d fmri_crop_moco_mean.nii.gz -dseg fmri_crop_moco_mean_seg_manual.nii.gz -param step=1,type=seg,algo=slicereg,metric=MeanSquares,smooth=2:step=2,type=im,algo=bsplinesyn,metric=MeanSquares,iter=5,gradStep=0.5 -initwarp ../t2/warp_template2anat.nii.gz -initwarpinv ../t2/warp_anat2template.nii.gz -qc "$SCT_BP_QC_FOLDER"
+# Use destination template seg file ('-dseg') to generate QC for sct_fmri_moco
+sct_qc -i fmri_crop.nii.gz -d fmri_crop_moco.nii.gz -s fmri_crop_moco_mean_seg_manual.nii.gz -p sct_fmri_moco -qc "$SCT_BP_QC_FOLDER"
```
jcohenadad, line 240: we could also do as suggested in https://github.com/neuropoly/spinalcordtoolbox/pull/3351/files#r618512267, but we could also leave it as it is, because it shows users that this approach is also possible. Moreover, the centerline is not very reliable (consequently, the cord is not well centered in each element of the mosaic)

joshuacwnewton: As you've said, I do like `sct_qc` more than `sct_crop_image` here.

joshuacwnewton: My main concern, though, is about the usefulness of the `-qc` flag in the moco scripts themselves. Since the steps right now are:

```console
# 1. Get motion-corrected image
sct_dmri_moco [...]

# 2. Get cord segmentation by processing motion-corrected image (or manual for fmri)
sct_deepseg_sc [...]

# 3. Use motion-corrected image + cord segmentation to generate qc
sct_qc -p sct_dmri_moco [...]
```

Then I don't quite understand how someone could use the `-qc` option in step 1. They would not yet have the mandatory cord segmentation required for `-qc-seg` (it's generated in step 2, but is needed in step 1).

jcohenadad: > Then I don't quite understand how someone would be able to use the -qc option for the sct_dmri_moco command in step 1. They would not yet have the mandatory cord segmentation required for -qc-seg (as the segmentation is generated in step 2, but is needed in step 1).

the "segmentation" would not be a segmentation per se, but the centerline extracted here:
https://github.com/neuropoly/spinalcordtoolbox/blob/07a383f0bf41d6bfdf02726a722b34e99045845c/batch_processing.sh#L226

In fact, while the flag `-qc-seg` asks for a segmentation, what is essentially needs is an image where the center of mass at each slices corresponds (more or less) to the center of the spinal cord. This is what is used by the QC report to then crop each sub-image of the mosaic for the nice HTML display.

joshuacwnewton: Ahhh, I see. I was too focused on the `-qc-seg` option name, and got the wrong idea. Thank you for the clarification. (I wonder if it would be better to call it `-qc-crop` or something along those lines?)

At the very least, for `batch_processing.sh`, what I can try to do is put in both options, but comment out one of them, and explain about centerline reliability as you mentioned above, that way the user can decide between options.

jcohenadad: > I wonder if it would be better to call it -qc-crop or something along those lines?

in most cases users will (should) input a segmentation, so let's leave it as it is

> At the very least, for batch_processing.sh, what I can try to do is put in both options, but comment out one of them, and explain about centerline reliability as you mentioned above, that way the user can decide between options for how they want to generate the QC for moco functions.

yeah... i don't know... i'm afraid some users might be confused. I'd rather we keep this batch_processing as simple as possible (ie: not too many alternative), because for many of our users not familiar with computers/neuroimaging, this script really is the first entry point.

joshuacwnewton: Thanks for the feedback. I'll refrain from making any further changes, then, and stick with the simplest addition to batch_processing. :)

batch_processing.sh
```diff
@@ -230,8 +232,11 @@ sct_crop_image -i fmri.nii.gz -m mask_fmri_mean.nii.gz -o fmri_crop.nii.gz
 # Tips: Here data have sufficient SNR and there is visible motion between two consecutive scans, so motion correction is more efficient with -g 1 (i.e. not average consecutive scans)
 sct_fmri_moco -i fmri_crop.nii.gz -g 1
 # Segment spinal cord manually
-# Since these data have very poor cord/CSF contrast, it is difficult to segment the cord properly
-# and hence in this case we do it manually. The file is called: fmri_crop_moco_mean_seg_manual.nii.gz
+#   Since these data have very poor cord/CSF contrast, it is difficult to segment the cord properly using sct_deepseg_sc
+#   and hence in this case we do it manually. The file is called: fmri_crop_moco_mean_seg_manual.nii.gz
+#   There is no command for this step, because the file is included in the 'sct_example_data' dataset.
```
joshuacwnewton, line 237: I have made one other small change (just indenting + adding a clarification) because I wanted to make it clear that this comment is a placeholder (corresponding to the sct_deepseg_sc command from the dmri steps).

jcohenadad (approved):

Issue #3046, opened 2020-11-15
Deprecate Tensorflow-based methods (`sct_deepseg_<x>`) and use only PyTorch (`sct_deepseg`)


joshuacwnewton:
### Context

Historically, SCT has used `Keras`/`tensorflow` for its deep learning-based segmentation methods (`sct_deepseg_sc`, `sct_deepseg_gm`, `sct_deepseg_lesion`). However, #2639 introduced the more general `sct_deepseg`, which aims to provide alternatives that use `pytorch` instead. (More context in #2626.) 

But, as noted in #2639, the transition is not complete:

> For now, this new function will not replace `sct_deepseg_X`, but will be complementary to the existing functions, so we can gather feedback from users before making these functions deprecated. The "big" change will likely be associated with a major release.

### Motivation

The main motivation to switch is to unify the development between SCT and ivadomed. Further information here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2626#issuecomment-595560605

Other benefits include:

* Easier Python version support, because Tensorflow is typically the only blocker for upgrading Python versions:
    * `tensorflow==1.5` ties us to Python 3.6, which is quickly approaching EOL ([#3227](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3227)).
    * `tensorflow==1.15` ties us to Python 3.7. ([#3367](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367))
    * etc.

    Since PyTorch is the main library used by sister project `ivadomed`, and their development team has already spent much effort testing Python 3.7/3.8/3.9, becoming PyTorch-only helps to sync us up with `ivadomed`.

* Fewer dependencies, so fewer upstream issues (e.g. [#2987](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2987), [#2562](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2562))
* Security benefits (we're currently quietly ignoring many dependabot vulnerability warnings for Tensorflow.)

### Next steps

There is a brief summary in https://github.com/neuropoly/spinalcordtoolbox/issues/3227#issuecomment-821894992:

> The bigger work is really to tie in the new model training with the git-annex datasets that @kousu is deploying. We also need to make some decisions on how we want to expose the models, what default models to install, how many flavors of these models (eg: SoftSeg vs. HardSeg), etc. 

[...]

Issue #2562, opened 2020-01-04
Illegal instruction when importing Keras/Tensorflow with CPU not supporting AVX


zgm0407:
I run  sct_check_dependencies after launch sct_4.0.0 on Win10's docker DeskTop, it seemed something wrong. how can I resolve the problem?
### Description

~~~
sct@35acfd3d29fa:~/sct_4.0.0/bin$  sct_check_dependencies

--
Spinal Cord Toolbox (4.0.0)

SCT info:
- version: 4.0.0
- path: /home/sct/sct_4.0.0
OS: linux (Linux-4.9.184-linuxkit-x86_64-with-debian-buster-sid)
CPU cores: Available: 2, Used by SCT: 2
RAM: MemTotal:        2047012 kB
total        used        free      shared  buff/cache   available
Mem:           1999         348         167           0        1482        1465
Swap:          1023           2        1021
Check Python executable.............................[OK]
  Using bundled python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 19:07:31)
[GCC 7.3.0] at /home/sct/sct_4.0.0/python/envs/venv_sct/bin/python
Check if data are installed.........................[OK]
Check if numpy is installed.........................[OK] (1.17.0)
Check if cryptography is installed..................[OK] (2.7)
Check if dipy is installed..........................[OK] (0.16.0)
Check if futures is installed.......................[OK]
Check if h5py is installed..........................[OK] (2.9.0)
Check if Keras is installed.........................Illegal instruction
sct@35acfd3d29fa:~/sct_4.0.0/bin$
~~~

jcohenadad:
Hi,
There might be incompatibilities with your CPU. Some info here:
- https://github.com/tensorflow/tensorflow/issues/17411
- https://github.com/keras-team/keras/issues/9630
- https://tech.amikelive.com/node-887/how-to-resolve-error-illegal-instruction-core-dumped-when-running-import-tensorflow-in-a-python-program/

What is your hardware configuration (CPU type)?

Before digging too extensively, could you please try the following:
- Install SCT via virtual machine (instructions [here](https://github.com/neuropoly/spinalcordtoolbox#installation-in-virtual-machine-vm))
- Use another computer (to see if your issue is computer-specific)

zgm0407:
I tried install SCT via virtual machine, but failed again.
SCT info:
- version: 4.2.0
- path: /home/brain/sct_4.2.0
[...]

PR #3361, opened 2021-04-27
(WIP) Test the impact of `tensorflow==1.13.1` + Python 3.7

joshuacwnewton:
See https://github.com/neuropoly/spinalcordtoolbox/issues/3227#issuecomment-821897302 for context:

> Just as a backup plan, I can also quickly test the "upgrading tensorflow" approach to see if we can still use Python 3.7 with SCT's current state (without deprecating anything). Even if we don't end up going that route, it might be useful preparation for the inevitable change later this year.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -628,14 +628,6 @@ else
 fi
 (
   pip install -r "$REQUIREMENTS_FILE" &&
-  # `tensorflow-tensorboard` is installed by `tensorflow==1.5.0` but is not needed,
-  # and conflicts with the other installation of `tensorboard`. So, we uninstall here.
-  # This is hacky, and should be removed as soon as we stop using `tensorflow==1.5.0`.
-  # See https://github.com/neuropoly/spinalcordtoolbox/issues/3035 for more info.
-  pip uninstall -y tensorflow-tensorboard &&
-  pip uninstall -y tensorboard &&
-  pip install tensorboard &&
-  # install sct itself
   print info "Installing spinalcordtoolbox..." &&
   pip install -e .
 ) ||
--- a/install_sct
+++ b/install_sct
@@ -587,8 +587,8 @@ esac
 # Run conda installer
 run bash "$TMP_DIR/miniconda.sh" -p "$SCT_DIR/$PYTHON_DIR" -b -f
 
-# create py3.6 venv (for Keras/TF compatibility with Centos7, see issue #2270)
-python/bin/conda create -y -n venv_sct python=3.6
+# create py3.7 venv (checking to see impact on test suite)
+python/bin/conda create -y -n venv_sct python=3.7
 
 # make sure that there is no conflict with local python install by making venv_sct an isolated environment.
 # workaround for https://github.com/conda/conda/issues/7173
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,7 +47,7 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
+    # # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
     runs-on: ubuntu-18.04
     container: archlinux
     steps:
@@ -64,7 +64,7 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
     runs-on: ubuntu-18.04
     container: debian:sid
     steps:
@@ -81,7 +81,7 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
     runs-on: ubuntu-18.04
     container: debian:testing
     steps:
@@ -98,7 +98,7 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
     runs-on: ubuntu-18.04
     container: debian:10
     steps:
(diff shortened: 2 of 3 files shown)


joshuacwnewton:
It looks like `tensorflow==1.13.1` [might not be good enough](https://github.com/neuropoly/spinalcordtoolbox/pull/3361/checks?check_run_id=2449110461#step:3:2739).

> E   ImportError: TensorBoard logging requires TensorBoard with Python summary writer installed. This should be available in 1.14 or above.

At one point, we weren't pinning Tensorflow at all, so I'll try with the latest 1.x version instead.

Drulex:
> * Upgrading tensorflow means that SCT is no longer available to users whose CPUs do not support the AVX instruction set.

Just to add a correction here: the pre-built versions of `tensorflow  > v1.6` are using the AVX instruction set by default (source: https://www.tensorflow.org/install/source#preconfigured_configurations), but it is still possible to build it without AVX support. 

The AVX instruction set is available on intel CPUs since 2011 so I guess it's a reasonable expectation to assume that it is available. I don't know if `pip` will correctly build the wheel from source considering the CPU architecture, but it is definitely possible (although not very user friendly) to build a later version without AVX support.

kousu:
> > * Upgrading tensorflow means that SCT is no longer available to users whose CPUs do not support the AVX instruction set.
> 
> Just to add a correction here: the pre-built versions of `tensorflow > v1.6` are using the AVX instruction set by default (source: https://www.tensorflow.org/install/source#preconfigured_configurations), but it is still possible to build it without AVX support.
> 
> The AVX instruction set is available on intel CPUs since 2011 so I guess it's a reasonable expectation to assume that it is available. I don't know if `pip` will correctly build the wheel from source considering the CPU architecture, but it is definitely possible (although not very user friendly) to build a later version without AVX support.

It's supposed to but it's a mess, especially on Windows and macOS which don't tend to come with compilers installed. Even if you have one it might be an incompatible compiler, and even if you have a compatible one building tensorflow is a multi-hour process, maybe a multi-day process on an older machine. On the dev side, I want to make sure everything can build from source; but on the user side, we need to find a workaround because that's not a reasonable thing to expect out of them.

Drulex:
> > > * Upgrading tensorflow means that SCT is no longer available to users whose CPUs do not support the AVX instruction set.
> > 
> > 
> > Just to add a correction here: the pre-built versions of `tensorflow > v1.6` are using the AVX instruction set by default (source: https://www.tensorflow.org/install/source#preconfigured_configurations), but it is still possible to build it without AVX support.
> > The AVX instruction set is available on intel CPUs since 2011 so I guess it's a reasonable expectation to assume that it is available. I don't know if `pip` will correctly build the wheel from source considering the CPU architecture, but it is definitely possible (although not very user friendly) to build a later version without AVX support.
> 
> It's supposed to but it's a mess, especially on Windows and macOS which don't tend to come with compilers installed. Even if you have one it might be an incompatible compiler, and even if you have a compatible one building tensorflow is a multi-hour process, maybe a multi-day process on an older machine. On the dev side, I want to make sure everything can build from source; but on the user side, we need to find a workaround because that's not a reasonable thing to expect out of them.

Perhaps, but how many people actually use CPUs that don't support `AVX`? If our friends at google don't build it anymore maybe the occurrence is low. 

In any case we can compile and distribute the binaries ourselves if we want to. Just trying to show that pinning `tensorflow` to a low version is not actually a hard requirement.

kousu:
> Perhaps, but how many people actually use CPUs that don't support `AVX`? If our friends at google don't build it anymore maybe the occurrence is low.

Enough for this emergency downgrade barely a year ago: https://github.com/neuropoly/spinalcordtoolbox/pull/2618

Our userbase is not rich cloud-savvy startups in high-bandwidth zones. Some are on the upside of the digital divide, many more people only get that kind of access indirectly via HPC clusters, and many are just using one old Windows machine in a corner of their lab.

> In any case we can compile and distribute the binaries ourselves if we want to. Just trying to show that pinning `tensorflow` to a low version is not actually a hard requirement.

For sure, this is what I'm advocating. But we've gotta do the work to make that a practical option for the people in the gap. But it's looking like just going to drop tensorflow entirely and hope torch is doing a slightly better job supporting more diverse CPUs.

It's interesting that this is the sort of issue that should be caught by CI but can't be because CI chooses a common denominator. I mean, unless we set up our own fleet of self-hosted CI. But I don't have the capacity to maintain that right now and I don't know who in the lab or at polytechnique would.

joshuacwnewton:
Given the discussion from today's meeting, it became more clear that the goal will be to prioritize #3046 so that the AVX-related workarounds will hopefully not be necessary.

Still, we can still keep an eye on this in the coming month. This PR can still help us bridge the gap (e.g. using it to extend the deprecation period of the old methods past December 2021).

kousu:
> Given the discussion from today's meeting, it became more clear that the goal will be to prioritize #3046 so that the AVX-related workarounds will hopefully not be necessary.
> 
> Still, we can still keep an eye on this in the coming month. This PR can still help us bridge the gap (e.g. maybe by using it to extend the deprecation period for the old methods past December 2021).

Today at our meeting we figured we are going to miss our deadline. So we need to just upgrade tensorflow in the meantime, and work on doing AVX-less tensorflow builds.

We *should* be able to do this by setting some `./configure` flags, waiting for the build to finish, and uploading the results somewhere. We can get Github Actions to do the builds, once we figure out the right flags to pick.

joshuacwnewton:
> "If you see AVX errors, run: pip install -f https://github.com/spinalcordtoolbox/tensorflow/releases -U tensorflow"

Actually, now that I think about it... could we do this step on behalf of the user during `install_sct`?

The AVX error occurs [immediately](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2562#issue-545264737) when Keras is imported. So, maybe we could do something like:

```bash
if [[ $(python -c "import keras") == "*Illegal Instruction*" ]]
then
    echo "CPU does not appear to support AVX instruction set. Installing compiled AVX-less version of Tensorflow..."
    pip install -f https://github.com/spinalcordtoolbox/tensorflow/releases -U tensorflow
fi
```

kousu:
> ```shell
> if [[ $(python -c "import keras") == "*Illegal instruction*" ]]
> then
>     echo "CPU does not appear to support AVX instruction set. Installing compiled AVX-less version of Tensorflow..."
>     pip install -f https://github.com/spinalcordtoolbox/tensorflow/releases -U tensorflow
> fi
> ```

Huh, yeah, I think that should work :). I like that it only applies to people who need the workaround.

When checking text outputs like this you might need to write `"$(LANG=C python -c "import keras" 2>&1)"` to be safe (LANG=C means "use American English", because C was developed at Bell labs, in America and `2>&1` means mix stderr into stdout).

joshuacwnewton:
This PR should be good to go, barring support for AVX-less users. I've opened #3509 to focus on this effort without muddying up this PR further.

joshuacwnewton:
I've compiled an AVX-less Tensorflow 1.15.5 wheel (Python 3.7, Linux) here: https://github.com/spinalcordtoolbox/docker-tensorflow-builder/pull/1

I'll wait until @kousu returns from vacation before figuring out what we want to do with this wheel (see "Task 3" in issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3509).

install_sct
```diff
@@ -634,14 +634,6 @@ else
...
 (
   pip install -r "$REQUIREMENTS_FILE" &&
-  # `tensorflow-tensorboard` is installed by `tensorflow==1.5.0` but is not needed,
-  # and conflicts with the other installation of `tensorboard`. So, we uninstall here.
-  # This is hacky, and should be removed as soon as we stop using `tensorflow==1.5.0`.
-  # See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3035 for more info.
-  pip uninstall -y tensorflow-tensorboard &&
-  pip uninstall -y tensorboard &&
-  pip install tensorboard &&
-  # install sct itself
```
kousu, line 644: bless

PR #3363, opened 2021-04-28
Print errors at the end, and in red.

kousu:
Fixes https://github.com/neuropoly/spinalcordtoolbox/issues/3362

2021-04-28-122817_936x614_scrot

### DIFF
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -135,8 +135,8 @@ class SCTArgumentParser(argparse.ArgumentParser):
             See https://github.com/neuropoly/spinalcordtoolbox/issues/3137.
         """
         # Source: https://stackoverflow.com/a/4042861
-        sys.stderr.write(f'{self.prog}: error: {message}\n\n')
         self.print_help(sys.stderr)
+        sys.stderr.write(f'\n\033[0;01;31m{self.prog}: error: {message}\033[m\n')
         sys.exit(2)
 
 
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -136,7 +136,7 @@ class SCTArgumentParser(argparse.ArgumentParser):
         """
         # Source: https://stackoverflow.com/a/4042861
         self.print_help(sys.stderr)
-        sys.stderr.write(f'\n\033[0;01;31m{self.prog}: error: {message}\033[m\n')
+        printv(f'\n{self.prog}: error: {message}\n', verbose=1, type='error', file=sys.stderr)
         sys.exit(2)
 
 
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -322,24 +322,23 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
     return status, output
 
 
-def printv(string, verbose=1, type='normal'):
+def printv(string, verbose=1, type='normal', file=None):
     """
     Enables to print color-coded messages, depending on verbose status. Only use in command-line programs (e.g.,
     sct_propseg).
     """
-
-    colors = {'normal': bcolors.normal, 'info': bcolors.green, 'warning': bcolors.yellow, 'error': bcolors.red,
+    colors = {'normal': bcolors.normal, 'info': bcolors.green, 'warning': bcolors.yellow + bcolors.bold, 'error': bcolors.red + bcolors.bold,
               'code': bcolors.blue, 'bold': bcolors.bold, 'process': bcolors.magenta}
 
     if verbose:
-        # The try/except is there in case stdout does not have isatty field (it did happen to me)
+        # The try/except is there in case file does not have isatty field (it did happen to me)
         try:
             # Print color only if the output is the terminal
-            if sys.stdout.isatty():
+            if file.isatty():
                 color = colors.get(type, bcolors.normal)
-                print(color + string + bcolors.normal)
+                print(color + string + bcolors.normal, file=file)
             else:
-                print(string)
+                print(string, file=file)
         except Exception as e:
             print(string)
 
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/utils/shell.py
```diff
@@ -135,8 +135,8 @@ def error(self, message):
             See https://github.com/neuropoly/spinalcordtoolbox/issues/3137.
         """
         # Source: https://stackoverflow.com/a/4042861
-        sys.stderr.write(f'{self.prog}: error: {message}\n\n')
         self.print_help(sys.stderr)
+        sys.stderr.write(f'\n\033[0;01;31m{self.prog}: error: {message}\033[m\n')
```
joshuacwnewton, line 139: This reminded me of `printv`, so I looked at how we're currently handling the `error` case:

https://github.com/neuropoly/spinalcordtoolbox/blob/c2d2206e21485ab7585af73f4d5b6e895f5fb6e7/spinalcordtoolbox/utils/sys.py#L331

https://github.com/neuropoly/spinalcordtoolbox/blob/c2d2206e21485ab7585af73f4d5b6e895f5fb6e7/spinalcordtoolbox/utils/sys.py#L340

Currently we're using `print` for every case, even for `'error'`. So, would there be any value to refactoring `printv` to switch between `stdout/stderr`. Then, using it here, too?

kousu: How about that now?

To make it show up on my color scheme I had to bold it, so that's now...inlined for every 'error', and for 'warning' too.

joshuacwnewton: Nice! 

Though, does `file` need to be an argument? Doing it this way requires us to backport `stderr` to every instance of `type=['error', 'warning'])`, I think.

I'm wondering if it would be easier to hardcode it as  'error'/'warning' -> stderr, everything else -> stdout.

kousu: Right, that is a problem.

I don't like `printv` that much; it wraps and combines orthogonal concepts (verbosity, colours and output location). And it replicates features that are in `logging`.

Guessing what the user wants causes more of the same. It would mean users can't `| ` reliably anymore to get the same output (they'd need to `2>&1 | ` instead). We'd have to comb for all uses of `printv(..., type='warning')` and `printv(..., type='error')` and decide if it's safe to split it off like that.

I don't know. This is such a niche detail.

kousu: Maybe I should just not use printv? I dunno.

joshuacwnewton: > Maybe I should just not use printv? I dunno.

Yeah, this sounds like a good call. My suggestion here inadvertently expanded the scope of the change quite a lot.

(Which might be ok if now is the time we want to deal with `printv`, but if not, your original change is the most straightforward.)

kousu: Come to think of it, the current behaviour is to write errors to stdout. We can think about changing that to be more consistent with the rest of unix but this patch doesn't change any current behaviour, only augments it.

joshuacwnewton (approved):

spinalcordtoolbox/utils/sys.py
```diff
@@ -322,24 +322,23 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
...
+    colors = {'normal': bcolors.normal, 'info': bcolors.green, 'warning': bcolors.yellow + bcolors.bold, 'error': bcolors.red + bcolors.bold,
               'code': bcolors.blue, 'bold': bcolors.bold, 'process': bcolors.magenta}
 
     if verbose:
-        # The try/except is there in case stdout does not have isatty field (it did happen to me)
+        # The try/except is there in case file does not have isatty field (it did happen to me)
         try:
             # Print color only if the output is the terminal
-            if sys.stdout.isatty():
+            if file.isatty():
```
joshuacwnewton, line 337: The default value for `file` is None, so if I understand correctly, this will always hit the `else:` case for the existing `printv` calls, thus not adding colour to anything, right?

So, shouldn't the default be `sys.stdout`?

kousu: :sweat_smile: yes of course. Drat. I was hoping to not have to duplicate code from the stdlib.

Oh well, should be fixed now.

joshuacwnewton (approved):

jcohenadad (approved):
Tested and works, thanks!

small comment: in the past (can't remember when), when we entered the command without any input, it would launch the help. Similar to when you write "git": it shows the usage, without raising an error. Currently, with SCT, an error is launched, which is not super-nice. Maybe we should open an issue to address this? IIRC we already talked about this

kousu:
> Currently, with SCT, an error is launched, which is not super-nice. Maybe we should open an issue to address this? IIRC we already talked about this

Git gives an error too, check it out:

```
$ if git; then echo; echo ">>> success <<<"; else echo; echo ">>> error <<<"; fi
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
   clone             Clone a repository into a new directory
   init              Create an empty Git repository or reinitialize an existing one

work on the current change (see also: git help everyday)
   add               Add file contents to the index
   mv                Move or rename a file, a directory, or a symlink
   restore           Restore working tree files
   rm                Remove files from the working tree and from the index
   sparse-checkout   Initialize and modify the sparse-checkout

examine the history and state (see also: git help revisions)
   bisect            Use binary search to find the commit that introduced a bug
   diff              Show changes between commits, commit and working tree, etc
   grep              Print lines matching a pattern
   log               Show commit logs
   show              Show various types of objects
   status            Show the working tree status

grow, mark and tweak your common history
   branch            List, create, or delete branches
   commit            Record changes to the repository
   merge             Join two or more development histories together
   rebase            Reapply commits on top of another base tip
   reset             Reset current HEAD to the specified state
   switch            Switch branches
   tag               Create, list, delete or verify a tag object signed with GPG

collaborate (see also: git help workflows)
   fetch             Download objects and refs from another repository
   pull              Fetch from and integrate with another repository or a local branch
   push              Update remote refs along with associated objects

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
See 'git help git' for an overview of the system.

>>> error <<<
```

The whole usage text *is* its error message. It's just not printing it in red.

jcohenadad:
> Git gives an error too, check it out:

ah! i learnt something today. OK, nevermind, then.

kousu:
It's really cool that everyone here is always teaching each other stuff!

Issue #2133, opened 2019-01-20
Fix broken integrity test for sct_label_utils


jcohenadad:
i deactivated it for emergency reasons (need to create release, test was not passing on travis due to precision issue, etc.)

joshuacwnewton:
I've tried reactivating it, and the test is still broken, but for a seemingly different reason.

https://github.com/neuropoly/spinalcordtoolbox/blob/f497c2f4cf512d184e07761f712631016d3e0454/testing/test_sct_label_utils.py#L57

This call to `sct_label_utils` returns `None` instead of the expected output, so I believe this call is broken/outdated.

jcohenadad:
@joshuacwnewton now that we have a proper API, you can bypass the system call in the integrity test, and just get the coordinates of the output file, and verify that they match with the [hard-coded coordinates](https://github.com/neuropoly/spinalcordtoolbox/blob/f497c2f4cf512d184e07761f712631016d3e0454/testing/test_sct_label_utils.py#L36).

This is what i've tried on my terminal:
```bash
cd sct_testing_data/t2
# run function to test
sct_label_utils -i t2_seg-manual.nii.gz -cubic-to-point -o test_centerofmass.nii.gz
# interrogate output
sct_label_utils -i test_centerofmass.nii.gz -display
# and this is the output:
--
Spinal Cord Toolbox (git-master-f497c2f4cf512d184e07761f712631016d3e0454)

Position=(31,28,25) -- Value= 1
All labels (useful syntax):
31,28,25,1   --> looks good 🎉
```

Issue #3372, opened 2021-05-05
Output of `sct_register_multimodal` no longer matches "goldstandard" `mt0_reg.nii.gz` files


joshuacwnewton:
More info to come. Just opening this issue for now so that I'll have a URL to point to in a comment, but I'll rewrite the description ASAP.

joshuacwnewton:
@jcohenadad I've updated the issue description. :slightly_smiling_face:

kousu:
How was this passing before? It looks to me like you faithfully ported the previous test. But presumably it was passing? Or were our test scripts just not catching its failure?

joshuacwnewton:
> How was this passing before? It looks to me like you faithfully ported the previous test. But presumably it was passing? Or were our test scripts just not catching its failure?

Funny enough, the old test wasn't actually doing the comparison!

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1ee7a54452956f17c59ba048f62ebe54d69bb529/testing/test_sct_register_multimodal.py#L41-L44

The old test required `fname_gt` to be defined, but nothing was specified. (That's why I had to guess at what the proper GT files should have been.)

PR #3373, opened 2021-05-05
Replace all 45 `sct_testing` tests with Pytest equivalents

### LINKED ISSUE #3093: Replace older `sct_testing` tests with new `pytest` tests

joshuacwnewton:
#### GitHub

- [ ] I've applied a [release milestone](https://github.com/neuropoly/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/neuropoly/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 

#### PR contents

## Description

This PR replaces `sct_testing` with `pytest`. The motivation for this is summarized here: https://github.com/neuropoly/spinalcordtoolbox/issues/3093#issuecomment-763792518

45 of the commits in this PR do the following:
   * Removes an original `sct_testing` test.
   * Removes a "backwards compatibility" wrapper test.
   * Adds a new `pytest` test to replace these older tests.

Then, there are 4 additional commits:

* f0defbf: Adds a fixture to ensure that the tests run in the correct directory.
* 762af7f: Adds a custom marker so that `pytest -m sct_testing` behaves identically to `sct_testing`.
* b78ce08: Fixes a bug in `sct_register_multimodal.py` discovered by the new tests.
* d06f961: Removes `spinalcordtoolbox/scripts/sct_testing.py` itself.

## How to review this PR

> This PR is quite large, and I apologize for that. But, I think it's best to do this in one go. If we try to split up the PR, we will have a half-working `sct_testing` if only one PR is merged. (If you have any ideas for how to make this easier to review, please let me know. :sweat_smile:)

To review this PR, please compare each old `sct_testing` test to the new `pytest` test to make sure they match up. (For example, compare `testing/test_sct_analyze_lesion.py` with `unit_testing/cli/test_cli_sct_analyze_lesion.py`.) 

To make this easier, use "Commit" view, because each commit will contain a pair of test files. 

> _**Note:** The old `sct_testing` tests can be a bit confusing. So, to help explain, I have written a Wiki page called [**sct_testing**](https://github.com/neuropoly/spinalcordtoolbox/wiki/Testing:-sct_testing)._

## Linked issues

Fixes #3093.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2133.


unit_testing/cli/test_cli_sct_compute_ernst_angle.py
```diff
@@ -1,14 +1,17 @@
...
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_ernst_angle_value_against_groundtruth():
+    """Run the CLI script and verify that computed ernst angle is equivalent to known ground truth value."""
+    fname_out = 'ernst_angle.txt'
+    sct_compute_ernst_angle.main(argv=['-tr', '2000', '-t1', '850', '-o', fname_out])
+    with open(fname_out, 'r') as f:
+        angle_result = float(f.read())
+    assert angle_result == pytest.approx(84.543553255)
```
cakester, line 17: is this the same as `abs(angle_result - param_test.angle_gt) < param_test.threshold:`? approx considers numbers within a relative tolerance of 1e-6 (i.e. one part in a million) of its expected value to be equal, you threshold is bigger than that

unit_testing/cli/test_cli_sct_compute_mscc.py
```diff
@@ -1,14 +1,19 @@
...
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_mscc_value_against_groundtruth():
+    """Run the CLI script and verify that computed mscc value is equivalent to known ground truth value."""
+    di, da, db = 6.85, 7.65, 7.02
+    # FIXME: The results of "sct_compute_mscc" are not actually verified. Instead, the "mscc" function is called,
+    #        and THOSE results are verified instead.
+    # This was copied as-is from the existing 'sct_testing' test, but should be fixed at a later date.
+    sct_compute_mscc.main(argv=['-di', str(di), '-da', str(da), '-db', str(db)])
+    mscc = sct_compute_mscc.mscc(di=di, da=da, db=db)
+    assert mscc == pytest.approx(6.612133606)
```
cakester, line 19: same thing, pytest.approx has a more strict boundary than the original test

unit_testing/cli/test_cli_sct_dice_coefficient.py
```diff
@@ -1,14 +1,21 @@
...
+def test_sct_dice_coefficient_check_output_against_groundtruth():
+    """Run the CLI script and verify its output matches a known ground truth value."""
+    # FIXME: The results of "sct_dice_coefficient" are not actually verified. Instead, the "compute_dice" function
+    #        is called, and THOSE results are verified instead.
+    # This was copied as-is from the existing 'sct_testing' test, but should be fixed at a later date.
+    path_data = 't2/t2_seg-manual.nii.gz'
+    sct_dice_coefficient.main(argv=['-i', path_data, '-d', path_data])
+    im_seg_manual = Image(path_data)
+    dice_segmentation = compute_dice(im_seg_manual, im_seg_manual, mode='3d', zboundaries=False)
+    assert dice_segmentation == 1.0
```
cakester, line 21: what about the assertion message like the original test `'\nERROR: Dice coefficient should be : ' + str(param_test.dice_value)`

unit_testing/cli/test_cli_sct_deepseg_gm.py
```diff
@@ -1,14 +1,18 @@
...
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_deepseg_gm_check_dice_coefficient_against_groundtruth():
+    """Run the CLI script and verify that dice (computed against ground truth) is within a certain threshold."""
+    fname_out = 'output.nii.gz'
+    fname_gt = 't2s/t2s_uncropped_gmseg_manual.nii.gz'
+    sct_deepseg_gm.main(argv=['-i', 't2s/t2s_uncropped.nii.gz', '-o', fname_out, '-qc', 'testing-qc'])
+    dice_segmentation = compute_dice(Image(fname_out), Image(fname_gt), mode='3d', zboundaries=False)
+    assert dice_segmentation >= 0.85
```
cakester, line 18: i guess the display`# display
    param_test.output += 'Computed dice: '+str(dice_segmentation)
    param_test.output += '\nDice threshold (if computed Dice smaller: fail): '+str(param_test.dice_threshold)` is not important here

unit_testing/cli/test_cli_sct_deepseg_sc.py
```diff
@@ -1,23 +1,12 @@
...
-
-def test_sct_deepseg_sc_o_flag(tmp_path):
-    argv = ['-i', 'sct_testing_data/t2/t2.nii.gz', '-c', 't2', '-o',
-            str(os.path.join(str(tmp_path), 'test_seg.nii.gz'))]
-    sct_deepseg_sc.main(argv)
-    assert os.path.isfile(os.path.join(str(tmp_path), 'test_seg.nii.gz'))
+def test_sct_deepseg_sc_check_output_exists(tmp_path):
+    fname_out = str(tmp_path / 'test_seg.nii.gz')
+    sct_deepseg_sc.main(argv=['-i', 'sct_testing_data/t2/t2.nii.gz', '-c', 't2', '-o', fname_out])
+    assert os.path.isfile(fname_out)
```
cakester, line 12: only one test? how about `default_args = ['-i t2/t2.nii.gz -c t2 -qc testing-qc']  # default parameters` from the deleted file?

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -1,23 +1,32 @@
...
 
-@pytest.mark.script_launch_mode('subprocess')
-def test_sct_detect_pmj_backwards_compat(script_runner):
-    ret = script_runner.run('sct_testing', '--function', 'sct_detect_pmj')
-    logger.debug(f"{ret.stdout}")
-    logger.debug(f"{ret.stderr}")
-    assert ret.success
-    assert ret.stderr == ''
-
-def test_sct_detect_pmj_o_flag(tmp_path):
```
cakester, line 18: where did this test go?

unit_testing/cli/test_cli_sct_dice_coefficient.py
joshuacwnewton: Here is what an example failure would look like (with no assertion message):

```
unit_testing/cli/test_cli_sct_dice_coefficient.py::test_sct_dice_coefficient_check_output_against_groundtruth FAILED [100%]

============================================== FAILURES ==============================================
_____________________ test_sct_dice_coefficient_check_output_against_groundtruth _____________________
  File "/home/joshua/repos/spinalcordtoolbox/unit_testing/cli/test_cli_sct_dice_coefficient.py", line 22, in test_sct_dice_coefficient_check_output_against_groundtruth
    assert dice_segmentation == 1.0
AssertionError: assert 2.0 == 1.0
  +2.0
  -1.0
====================================== short test summary info =======================================
FAILED unit_testing/cli/test_cli_sct_dice_coefficient.py::test_sct_dice_coefficient_check_output_against_groundtruth
========================================= 1 failed in 2.90s ==========================================
```

I was thinking that the pytest output might be clear enough that we don't need an additional message.

unit_testing/cli/test_cli_sct_detect_pmj.py
joshuacwnewton: Ah, thank you for reminding me! 

I have combined this test with the other test in ee44471. (Since the old test was just using checking if `fname_out` exists, and the new test has to open `Image(fname_out),` it does basically the same thing.)

joshuacwnewton: Ah wait, my apologies, there is a slight difference. I will need to make another change.

joshuacwnewton: Ah... interesting. I looked into the `sct_detect_pmj` script, and I believe `-igt` is not actually used anywhere in the script! (For context, it was added quite some time ago in https://github.com/neuropoly/spinalcordtoolbox/pull/1453#discussion_r627384597.)

I will open a separate issue for this.

joshuacwnewton: Done in https://github.com/neuropoly/spinalcordtoolbox/issues/3377.

joshuacwnewton:
For transparency, I'm going to rebase the changes from 0e3434e, 4c8d34d, f6ab574, and ee44471 to make sure that the commit view is still useful for comparing the tests from this PR.

unit_testing/cli/test_cli_sct_extract_metric.py
```diff
@@ -1,14 +1,23 @@
...
+def test_sct_extract_metric_weighted_average_value_against_groundtruth():
+    """Verify that computed WA value is equivalent to known ground truth value."""
+    fname_out = 'quantif_mtr.csv'
+    sct_extract_metric.main(argv=['-i', 'mt/mtr.nii.gz', '-f', 'mt/label/atlas', '-method', 'wa', '-l', '51',
+                                  '-z', '1:2', '-o', fname_out])
+    with open(fname_out, 'r') as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=',')
+        mtr_result = np.float([row['WA()'] for row in reader][0])
+
+    assert mtr_result == pytest.approx(32.6404)
```
cakester, line 23: same thing with pytest.approx bound

joshuacwnewton: Sorry about this! Thanks for catching another instance. :sweat_smile:

unit_testing/cli/test_cli_sct_resample.py
```diff
@@ -1,14 +1,25 @@
...
+    ('t2/t2.nii.gz', '-vox', '120x110x26', (120, 110, 26, 1, 0.5, 0.5, 2.0, 1)),
+])
+def test_sct_resample_output_has_expected_dimensions(path_in, type_arg, dim, expected_dim):
+    """Run the CLI script and verify output file exists."""
+    path_out = 'resampled.nii.gz'
+    sct_resample.main(argv=['-i', path_in, '-o', path_out, type_arg, dim, '-v', '1'])
+    actual_dim = Image(path_out).dim
+
+    for actual_val, expected_val in zip(actual_dim, expected_dim):
+        assert actual_val == pytest.approx(expected_val)
```
cakester, line 25: here is another one

joshuacwnewton: Done in ef29e95.

unit_testing/cli/test_cli_sct_extract_metric.py
joshuacwnewton: Done in 16261d44.

spinalcordtoolbox/scripts/sct_register_multimodal.py
```diff
@@ -348,6 +349,7 @@ def main(argv=None):
     else:
         fname_mask = ''
     padding = arguments.z
+    paramregmulti = deepcopy(DEFAULT_PARAMREGMULTI)
```
kousu, line 352: Good fix, but it raises the deeper question of why the tests are modifying their input at all.

testing/test_sct_register_multimodal.py
```diff
@@ -1,71 +0,0 @@
...
-    data_gt = im_gt.data
-    data_result = Image(fname_result).data
-    # get dimensions
-    nx, ny, nz, nt, px, py, pz, pt = im_gt.dim
-    # set the difference threshold to 1e-3 pe voxel
-    threshold = 1e-3 * nx * ny * nz * nt
-    # check if non-zero elements are present when computing the difference of the two images
-    diff = data_gt - data_result
-    # compare images
-    if abs(np.sum(diff)) > threshold:
```
kousu, line 66: Again shouldn't this be `np.linalg.norm`?

unit_testing/cli/test_cli_sct_register_multimodal.py
```diff
@@ -35,3 +28,36 @@ def test_sct_register_multimodal_mask_no_checks(tmp_path):
...
+    (True, 'step=1,algo=centermassrot,type=imseg,rot_method=pcahog', None),
+    (True, 'step=1,algo=columnwise,type=seg,smooth=1', None),
+])
+def test_sct_register_multimodal_mt0_image_data_within_threshold(use_seg, param, fname_gt):
+    """Run the CLI script and verify that the output image data is close to a reference image (within threshold)."""
+    fname_out = 'mt0_reg.nii.gz'
+
+    argv = ['-i', 'mt/mt0.nii.gz', '-d', 'mt/mt1.nii.gz', '-o', fname_out, '-x', 'linear', '-r', '0', '-param', param]
+    seg_argv = ['-iseg', 'mt/mt0_seg.nii.gz', '-dseg', 'mt/mt1_seg.nii.gz']
+    sct_register_multimodal.main(argv=(argv + seg_argv) if use_seg else argv)
```
kousu, line 49: Is the a chance that  https://github.com/neuropoly/spinalcordtoolbox/issues/3372 was caused by reordering argv?

unit_testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -62,3 +56,49 @@ def test_sct_register_to_template_non_rpi_data(tmp_path, template_lpi):
...
     assert np.linalg.norm(img_orig.data - img_reg.data) < 1
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+@pytest.mark.parametrize("fname_gt, remaining_args", [
+    ('template/template/PAM50_small_cord.nii.gz',
+     ['-l', 't2/labels.nii.gz', '-t', 'template', '-qc', 'qc-testing', '-param',
+      'step=1,type=seg,algo=centermassrot,metric=MeanSquares:step=2,type=seg,algo=bsplinesyn,iter=5,metric=MeanSquares']),
+    (os.path.join(__sct_dir__, 'data/PAM50/template/PAM50_cord.nii.gz'),
```
kousu, line 67: `__sct_dir__` should be taken care of by `run_in_sct_testing_data_dir`?

In https://github.com/neuropoly/spinalcordtoolbox/issues/2669#issuecomment-665153804 et al I want to get rid of `__sct_dir__`.

Also what's `template/template`? Have I just not seen that before? Why isn't that's `PAM50/template/`?

unit_testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -62,3 +56,49 @@ def test_sct_register_to_template_non_rpi_data(tmp_path, template_lpi):
...
+        '-d', fname_gt,
+        '-w', 'warp_anat2template.nii.gz',
+        '-o', 'test_anat2template.nii.gz',
+        '-x', 'nn',
+        '-v', '0'])
+
+    # compute dice coefficient between template segmentation warped to anat and segmentation from anat
+    im_seg = Image(fname_seg)
+    im_template_seg_reg = Image('test_template2anat.nii.gz')
+    dice_template2anat = msct_image.compute_dice(im_seg, im_template_seg_reg, mode='3d', zboundaries=True)
```
kousu, line 97: No reason to overcomplicate this

```diff
-     dice_template2anat = msct_image.compute_dice(im_seg, im_template_seg_reg, mode='3d', zboundaries=True)
+     dice_template2anat = spinalcordtoolbox.image.compute_dice(im_seg, im_template_seg_reg, mode='3d', zboundaries=True)
```

unit_testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -62,3 +56,49 @@ def test_sct_register_to_template_non_rpi_data(tmp_path, template_lpi):
...
+    assert dice_template2anat > dice_threshold
+
+    # compute dice coefficient between anat segmentation warped to template and segmentation from template
+    im_seg_reg = Image('test_anat2template.nii.gz')
+    im_template_seg = Image(fname_gt)
+    dice_anat2template = msct_image.compute_dice(im_seg_reg, im_template_seg, mode='3d', zboundaries=True)
```
kousu, line 103: ditto

unit_testing/cli/test_cli_sct_resample.py
```diff
@@ -1,14 +1,25 @@
...
-    logger.debug(f"{ret.stderr}")
-    assert ret.success
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+@pytest.mark.parametrize("path_in,type_arg,dim,expected_dim", [
+    ('dmri/dmri.nii.gz', '-f', '0.5x0.5x1', (20, 21, 5, 7, 1.6826923, 1.6826923, 17.5, 2.2)),
+    ('t2/t2.nii.gz', '-mm', '0.97x1.14x1.2', (62, 48, 43, 1, 0.96774191, 1.1458334, 1.2093023, 1)),
+    ('t2/t2.nii.gz', '-vox', '120x110x26', (120, 110, 26, 1, 0.5, 0.5, 2.0, 1)),
+])
```
kousu, line 17: Copy the "4d factor" etc comments down here

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
```diff
@@ -1,19 +1,18 @@
...
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_smooth_spinalcord_no_checks():
+    """Run the CLI script without checking results.
+    TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
+    sct_smooth_spinalcord.main(argv=['-i', 't2/t2.nii.gz', '-s', 't2/t2_seg-manual.nii.gz', '-smooth', '0,0,5'])
 
 
 def test_sct_smooth_spinalcord_o_flag(tmp_path):
```
kousu, line 18: Can you delete the `os.unlink`s in this? I guess that `@pytest.mark.usefixtures("run_in_sct_testing_data_dir")` still isn't `tmp_path`, so the output will still muck up the input no matter what :/

conftest.py
```diff
@@ -31,6 +31,16 @@ def pytest_sessionstart():
...
 
 
+@pytest.fixture
+def run_in_sct_testing_data_dir():
+    """Temporarily change the working directory to 'sct_testing_data'. This replicates the behavior of the old
+    `sct_testing`, and is needed to prevent tests from cluttering the working directory with output files."""
+    cwd = os.getcwd()
+    os.chdir(sct_test_path())
+    yield
+    os.chdir(cwd)
```
kousu, line 41: I think you can move this to `unit_testing/cli/conftest.py`, **and mark it with** `autouse=True`. Then you can get rid of all the `@pytest.mark.usefixtures()`.

`conftests` are all run according to the folder structure, the way packages all run their `__init__`s: https://docs.pytest.org/en/latest/how-to/fixtures.html?highlight=conftest#override-a-fixture-on-a-folder-conftest-level

kousu, line 41: This is probably not a real concern but: this chdir can fail if the filesystem has shifted under sct during testing. This is an example of where functional programming's copy-and-edit (which you get for free with unix subprocesses) is safer than set-and-reset.

kousu: Though _explicit is better than implicit_

conftest.py
```diff
@@ -31,6 +31,16 @@ def pytest_sessionstart():
     downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
 
 
+@pytest.fixture
+def run_in_sct_testing_data_dir():
+    """Temporarily change the working directory to 'sct_testing_data'. This replicates the behavior of the old
+    `sct_testing`, and is needed to prevent tests from cluttering the working directory with output files."""
+    cwd = os.getcwd()
+    os.chdir(sct_test_path())
+    yield
```
kousu, line 40: If you do this

```suggestion
def test_data():
    """Get the path to the test data folder.
    
    Temporarily changes the working directory to 'sct_testing_data'. This replicates the behavior of the old
    `sct_testing`, and is needed to prevent tests from cluttering the working directory with output files
    but should not be relied on in new tests."""
    cwd = os.getcwd()
    os.chdir(sct_test_path())
    yield pathlib.Path(os.getcwd())
```

then you can start improving the tests with e.g.

```
im_gt = Image(test_data/'mt/mt1.nii.gz')
```

and then it will be easier to get rid of the `chdir` eventually.

kousu: And it also shortens the fixture name. I think fixtures are supposed to be nouns, not verbs, generally.

kousu (approved):
Thank you for doing this grind.

I wonder what is the point of testing these scripts?  I would think the numerical checks should be handled at a lower level and testing the CLI should just like...make sure the right files come out; maybe that they have the right shape?

Probably about half the tests are unfinished, the ones you've marked with `_no_checks`, but at least they do the minimum of exercises the code and making sure it doesn't crash.

joshuacwnewton:
> I wonder what is the point of testing these scripts? I would think the numerical checks should be handled at a lower level and testing the CLI should just like...make sure the right files come out; maybe that they have the right shape?

This sounds reasonable to me! Especially in a world where we've refactored the processing functionality from the CLI into the API, and the CLI scripts are just light wrappers for the API.

unit_testing/cli/test_cli_sct_compute_hausdorff_distance.py
```diff
@@ -1,14 +1,22 @@
...
-    ret = script_runner.run('sct_testing', '--function', 'sct_compute_hausdorff_distance')
-    logger.debug(f"{ret.stdout}")
-    logger.debug(f"{ret.stderr}")
-    assert ret.success
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_hausdorff_distance_values_against_threshold():
+    """Run the CLI script and verify computed distances are all within a given threshold.
+    TODO: Distances are all 0.0. Is this test checking anything useful?"""
```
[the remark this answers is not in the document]
kousu: 🤔

unit_testing/cli/test_cli_sct_compute_hausdorff_distance.py
```diff
@@ -1,14 +1,22 @@
...
+    sct_compute_hausdorff_distance.main(argv=['-i', 't2s/t2s_gmseg_manual.nii.gz', '-d', 't2s/t2s_gmseg_manual.nii.gz'])
+
+    with open('hausdorff_distance.txt', 'r') as f:
+        hausdorff_distance_lst = []
+        for line in f.readlines():
+            if line.startswith('Slice'):
+                hausdorff_distance_lst.append(float(line.split(': ')[1].split(' -')[0]))
+
+    assert max(hausdorff_distance_lst) <= 1.0
```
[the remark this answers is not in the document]
joshuacwnewton: I think this is because the original test was checking the distance between one file and itsel:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/998226f1e654961da6bdf77ad083b5f74013beb2/testing/test_sct_compute_hausdorff_distance.py#L18

So, it makes sense that the computed distances are 0. If we stick to comparing one file with itself, then the assertion should probably be `assert np.all(hausdorff_distance_list == 0)`. 

(Otherwise, we could find two different files to compare against one another, but... that would be a more involved test, I think.)

unit_testing/cli/test_cli_sct_compute_mscc.py
```diff
@@ -1,14 +1,19 @@
...
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_mscc_value_against_groundtruth():
+    """Run the CLI script and verify that computed mscc value is equivalent to known ground truth value."""
+    di, da, db = 6.85, 7.65, 7.02
+    # FIXME: The results of "sct_compute_mscc" are not actually verified. Instead, the "mscc" function is called,
+    #        and THOSE results are verified instead.
+    # This was copied as-is from the existing 'sct_testing' test, but should be fixed at a later date.
+    sct_compute_mscc.main(argv=['-di', str(di), '-da', str(da), '-db', str(db)])
```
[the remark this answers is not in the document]
joshuacwnewton: > some CPUs might end up rounding 6.85 to 6.84433338383 and print that.

That seems like a pretty significant rounding error for a simple `float -> str` conversion. There's no arithmetic going on, right? Wouldn't it just preserve the number exactly as typed? :thinking:

unit_testing/cli/test_cli_sct_crop_image.py
```diff
@@ -1,14 +1,21 @@
...
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+@pytest.mark.parametrize("path_in,path_out,remaining_args,expected_dim", [
+    ('t2/t2.nii.gz', 't2_crop_xyz.nii', ['-xmin', '1', '-xmax', '-3', '-ymin', '2', '-ymax', '10'], (57, 9, 52)),
+    ('t2/t2.nii.gz', 't2_crop_mask.nii', ['-m', 't2/t2_seg-manual.nii.gz'], (11, 55, 13)),
+    ('t2/t2.nii.gz', 't2_crop_ref.nii', ['-ref', 'mt/mt0.nii.gz'], (37, 55, 34)),
+])
+def test_sct_crop_image_output_has_expected_dimensions(path_in, path_out, remaining_args, expected_dim):
+    """Run the CLI script and verify cropped image has the expected dimensions."""
+    sct_crop_image.main(argv=['-i', path_in, '-o', path_out] + remaining_args)
+    nx, ny, nz, _, _, _, _, _ = Image(path_out).dim
```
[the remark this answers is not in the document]
joshuacwnewton: Aha! Yes, they're necessary to define the relation between voxel coordinates to real world coordinates. Here's a breakdown:

* `nx, ny, nz`: Image dimensions in voxels.
* `nt`: Time dimension (relevant for sequences of scans, esp. dMRI and fMRI).
* `px, py, pz`: The conversion factors between voxels and millimetres (otherwise known as `pixdim` in [`nibabel` docs](https://nipy.org/nibabel/nibabel_images.html#the-image-header)).
* `pt`: The time between scans in milliseconds. 

What I can do is test the remaining 5 values to ensure that `sct_crop_image` doesn't inadvertently change them, since they should be unaffected by cropping.

testing/test_sct_label_utils.py
```diff
@@ -1,58 +0,0 @@
...
-def test_integrity(param_test):
-    """
-    Test integrity of function
-    """
-    # find the test that is performed and check the integrity of the output
-    index_args = param_test.default_args.index(param_test.args)
-
-    # Removed because of:
-    # https://travis-ci.org/spinalcordtoolbox/spinalcordtoolbox/jobs/482061826
-    param_test.output += "NOT TESTED-- SHOULD BE REACTIVATED ASAP"
```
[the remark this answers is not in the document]
joshuacwnewton: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2133 :D

unit_testing/cli/test_cli_sct_compute_hausdorff_distance.py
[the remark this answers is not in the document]
kousu: In that case, rename the test `test_sct_compute_null_hausdorff_distance` for now and we can find a fuller test later. Testing the null cases is important too, but it shouldn't be the only thing we do!

unit_testing/cli/test_cli_sct_compute_mscc.py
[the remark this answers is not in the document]
kousu: No, not necessarily: [What Every Computer Scientist Should Know About Floating Point Arithmetic](https://ece.uwaterloo.ca/~dwharder/NumericalAnalysis/02Numerics/Double/paper.pdf), particularly that far from every real number *exists*. So for example:

```
>>> 6.8500000000000001
6.8500000000000005
```

With only three significant digits it, well, it's probably safe, but it depends on the CPU.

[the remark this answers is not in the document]
joshuacwnewton: > With only three significant digits it, well, it's probably safe

:+1:! 

`6.8500000000000001 -> 6.8500000000000005` is much different than `6.85 -> 6.84433338383`. :stuck_out_tongue_closed_eyes:

unit_testing/cli/test_cli_sct_register_to_template.py
joshuacwnewton: At that point, it's not a CLI test at all! Might as well move it to the API tests, hehe.

unit_testing/cli/test_cli_sct_analyze_texture.py
```diff
@@ -1,14 +1,22 @@
...
-    logger.debug(f"{ret.stdout}")
-    logger.debug(f"{ret.stderr}")
-    assert ret.success
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_analyze_texture_image_data_within_threshold():
+    """Run the CLI script and verify that the output image data is close to a reference image (within threshold)."""
+    sct_analyze_texture.main(argv=['-i', 't2/t2.nii.gz', '-m', 't2/t2_seg-manual.nii.gz', '-feature', 'contrast',
+                                   '-distance', '1', '-ofolder', '.'])
```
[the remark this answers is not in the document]
joshuacwnewton: Agreed! I've opened an issue for this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3422.

unit_testing/cli/test_cli_sct_register_multimodal.py
joshuacwnewton: I've written a comment (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3372#issuecomment-861512461) to hopefully help explain the problem a little better.

unit_testing/cli/test_cli_sct_extract_metric.py
```diff
@@ -1,14 +1,23 @@
...
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_extract_metric_weighted_average_value_against_groundtruth():
+    """Verify that computed WA value is equivalent to known ground truth value."""
+    fname_out = 'quantif_mtr.csv'
+    sct_extract_metric.main(argv=['-i', 'mt/mtr.nii.gz', '-f', 'mt/label/atlas', '-method', 'wa', '-l', '51',
+                                  '-z', '1:2', '-o', fname_out])
+    with open(fname_out, 'r') as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=',')
+        mtr_result = np.float([row['WA()'] for row in reader][0])
```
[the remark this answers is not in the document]
joshuacwnewton: In this case, `np.genfromtext` is a little awkward, because the csv file contains both numeric and non-numeric columns. We want to access specifically the `WA()` column, but if we try using `np.genfromtxt`, we get this:

Screenshot from 2021-06-21 11-14-48

We could access the value using a numeric index, but that could be error prone if we add a new column... so, I think I like using `csv.DictReader` here.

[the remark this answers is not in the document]
joshuacwnewton: Actually, now that I type that out, it's a bit silly that we're only checking the weighted-average column. We could easily be checking all three.

Screenshot from 2021-06-21 11-17-30

unit_testing/cli/test_cli_sct_image.py
```diff
@@ -1,21 +1,43 @@
...
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_image_concat(tmp_path, dmri_t_slices, dmri_in):
+    """Run the CLI script and verify concatenated imaeg matches reference image."""
+    path_out = str(tmp_path / 'dmri_concat.nii.gz')
+    sct_image.main(argv=['-i'] + dmri_t_slices + ['-concat', 't', '-o', path_out])
+    ref = Image(dmri_in)
+    new = Image(path_out)
+    assert np.sum(ref.data - new.data) <= 1e-3
```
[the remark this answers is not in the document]
joshuacwnewton: > and adjust the tolerances to match, if necessary.

Funny enough, the difference between the reference image and the new image is 0.0, because process is just splitting and concatenating images. I'm not sure where the `1e-3` tolerance came from, but I think in this case it might be more reasonable to assert that there should be no difference at all.

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -1,20 +1,37 @@
...
+import spinalcordtoolbox.scripts.sct_label_vertebrae as sct_label_vertebrae
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.sct_testing
+def test_sct_label_vertebrae_consistent_disc(tmp_path):
+    """Check that all expected output labeled discs exist"""
+    command = '-i sct_testing_data/t2/t2.nii.gz -s sct_testing_data/t2/t2_seg-manual.nii.gz -c t2 -discfile sct_testing_data/t2/labels.nii.gz -ofolder ' + str(tmp_path)
+    sct_label_vertebrae.main(command.split())
```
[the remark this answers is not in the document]
joshuacwnewton: Ahh, I see what went wrong here. This test had already been written prior to this PR -- I just slapped a `@pytest.mark.sct_testing` on it, rather than checking the test itself.

Thanks for calling me out on this. :smile:

unit_testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -1,20 +1,37 @@
...
+    ref = Image('sct_testing_data/t2/labels.nii.gz')
+    pred = Image(os.path.join(tmp_path, 't2_seg-manual_labeled_discs.nii.gz'))
+    fp, fn = check_missing_label(pred, ref)
+    assert fp == []
+    assert fn == []
+
+    # Files created in root directory by sct_label_vertebrae
+    os.unlink('straightening.cache')
```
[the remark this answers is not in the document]
joshuacwnewton: Funny story... `straightening.cache` isn't output in the tmpdir! :sweat_smile:

unit_testing/cli/test_cli_sct_maths.py
```diff
@@ -1,14 +1,30 @@
...
+    TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
+    sct_maths.main(argv=['-i', 'mt/mtr.nii.gz', '-add', '1', '-o', 'test.nii.gz'])
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_maths_add_images_no_checks():
+    """Run the CLI script without checking results.
+    TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
+    sct_maths.main(argv=['-i', 'mt/mtr.nii.gz', '-add', 'mt/mtr.nii.gz', 'mt/mtr.nii.gz', '-o', 'test.nii.gz'])
```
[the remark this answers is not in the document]
joshuacwnewton: I definitely agree... but "I feel like we could do better" could be said about a lot of different tests here, and I wanted to keep the scope of the PR somewhat reasonable by focusing on verbatim porting for now. :sweat_smile:  

My plan was to go over all of these tests once more after I get coverage up and running again.

unit_testing/cli/test_cli_sct_register_to_template.py
joshuacwnewton: > `__sct_dir__` should be taken care of by `run_in_sct_testing_data_dir`?

In this case, we use `__sct_dir__` to access the PAM50 template stored in the `data/` folder of SCT's installation directory (i.e. it's separate from `sct_testing_data`).

> Also what's `template/template`? Have I just not seen that before? Why isn't that's `PAM50/template/`?

This is a great question, and it caught me off guard! It looks like this is an `sct_testing_data`-specific thing: https://github.com/spinalcordtoolbox/sct_testing_data/tree/master/template

So, it looks like earlier part of your comment should be using `template/template` instead of `os.path.join(__sct_dir__, 'data/PAM50/template/PAM50_cord.nii.gz').

conftest.py
joshuacwnewton: The tricky thing with this is that it would apply to all of the existing pytest tests (i.e. the non-sct_testing ones).

I don't think we handle `sct_test_data()` / `sct_testing_data` consistently in the test suite. So, I think it would be easiest to handle this in #3422.

conftest.py
joshuacwnewton: Same feelings here about wanting to handle this as part of #3422. I definitely agree that there are improvements to be made re: `sct_testing_data`, but I figure we can standardize the test suite as a whole in a dedicated PR.

unit_testing/cli/test_cli_sct_smooth_spinalcord.py
joshuacwnewton: You're right... The fact that `os.unlinks` are necessary is a good smell that `sct_smooth_spinalcord` could use an `-ofolder` argument. Right now, it pollutes the working directory no matter what we do.

unit_testing/cli/test_cli_sct_compute_hausdorff_distance.py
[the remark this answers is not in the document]
joshuacwnewton: Done in 03ea808.

unit_testing/cli/test_cli_sct_crop_image.py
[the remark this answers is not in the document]
joshuacwnewton: Done in 362455f.

unit_testing/cli/test_cli_sct_denoising_onlm.py
```diff
@@ -0,0 +1,16 @@
+import sys
+import pytest
+import logging
+
+from spinalcordtoolbox.scripts import sct_denoising_onlm
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.skipif(sys.platform == "darwin", reason="Test hangs on macOS 10.15 CI runners (cause currently unknown)")
```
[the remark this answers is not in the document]
joshuacwnewton: Done in c042124.

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -1,23 +1,33 @@
...
+    im_pmj = Image(fname_out)
+    im_pmj_manual = Image(fname_gt)
+
+    # compute Euclidean distance between predicted and GT PMJ label
+    x_true, y_true, z_true = np.where(im_pmj_manual.data == 50)
+    x_pred, y_pred, z_pred = np.where(im_pmj.data == 50)
+    x_true, y_true, z_true = im_pmj_manual.transfo_pix2phys([[x_true[0], y_true[0], z_true[0]]])[0]
+    x_pred, y_pred, z_pred = im_pmj.transfo_pix2phys([[x_pred[0], y_pred[0], z_pred[0]]])[0]
+
+    distance_detection = math.sqrt((x_true - x_pred)**2 + (y_true - y_pred)**2 + (z_true - z_pred)**2)
```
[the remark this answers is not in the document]
joshuacwnewton: Done in 1a284df.

unit_testing/cli/test_cli_sct_dmri_moco.py
```diff
@@ -1,14 +1,66 @@
...
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_dmri_moco_with_mask_check_params(tmp_path, dmri_mask):
+    """Run the CLI script with '-m' option and validate output moco params."""
+    sct_dmri_moco.main(argv=['-i', 'dmri/dmri.nii.gz', '-bvec', 'dmri/bvecs.txt', '-g', '3', '-r', '0',
+                             '-m', dmri_mask, '-ofolder', str(tmp_path)])
+
+    df = read_csv(tmp_path / "moco_params.tsv", sep="\t")
+    lresults = list(df['X'][:])
```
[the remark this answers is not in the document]
joshuacwnewton: Done in a295995.

unit_testing/cli/test_cli_sct_dmri_moco.py
```diff
@@ -1,14 +1,66 @@
...
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_dmri_moco_check_params(tmp_path):
+    """Run the CLI script and validate output moco params."""
+    sct_dmri_moco.main(argv=['-i', 'dmri/dmri.nii.gz', '-bvec', 'dmri/bvecs.txt', '-g', '3', '-x', 'nn', '-r', '0',
+                             '-ofolder', str(tmp_path)])
+
+    df = read_csv(tmp_path / "moco_params.tsv", sep="\t")
+    lresults = list(df['X'][:])
```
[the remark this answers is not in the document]
joshuacwnewton: a295995

unit_testing/cli/test_cli_sct_extract_metric.py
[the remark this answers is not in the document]
joshuacwnewton: d06dfc2

unit_testing/cli/test_cli_sct_image.py
```diff
@@ -31,3 +53,34 @@ def test_sct_image_display_warp_check_output_exists():
...
+
+    parent, filename, ext = extract_fname(fname_out)
+    fname_slices = [os.path.join(parent, filename + '_T' + str(i).zfill(4) + ext) for i in range(7)]
+    return fname_slices
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_image_pad():
+    """Run the CLI script and ."""
```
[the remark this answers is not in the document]
joshuacwnewton: 8fe71ed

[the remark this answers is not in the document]
joshuacwnewton: Would you look at that! The short hash isn't unique. 

8fe71ed2e6e47bc89a4493f38ffd2595c27dd7d3

unit_testing/cli/test_cli_sct_image.py
```diff
@@ -1,21 +1,43 @@
...
-@pytest.mark.script_launch_mode('subprocess')
-def test_sct_image_backwards_compat(script_runner):
-    ret = script_runner.run('sct_testing', '--function', 'sct_image')
-    logger.debug(f"{ret.stdout}")
-    logger.debug(f"{ret.stderr}")
-    assert ret.success
-    assert ret.stderr == ''
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+@pytest.mark.parametrize("path_in", ['t2/t2.nii.gz', 'dmri/dmri.nii.gz'])
```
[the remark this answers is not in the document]
joshuacwnewton: 93d4e36

unit_testing/cli/test_cli_sct_image.py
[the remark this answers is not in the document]
joshuacwnewton: e4fa6c2

unit_testing/cli/test_cli_sct_label_vertebrae.py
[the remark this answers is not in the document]
joshuacwnewton: 1f19ad2

unit_testing/cli/test_cli_sct_label_vertebrae.py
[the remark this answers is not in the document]
joshuacwnewton: Funny story -- `sct_label_vertebrae` _wasn't_ outputting the cache file into the tmpdir.

I fixed this in 1c372b1.

testing/test_sct_register_multimodal.py
joshuacwnewton: 2a41177

unit_testing/cli/test_cli_sct_register_to_template.py
joshuacwnewton: ef1064e

unit_testing/cli/test_cli_sct_resample.py
joshuacwnewton: 00a61ca

joshuacwnewton:
I think I've answered all of your feedback, @kousu. If it's not too much trouble, could you go through your comments and mark them as "resolved" if you're satisfied by the changes? :slightly_smiling_face:

conftest.py
kousu: I think if you specifically put it under `unit_testing/cli`, then it only applies to `unit_testing/cli`. If there's non-sct_testing tests in there already (I haven't looked to see) then you could make a separate folder just for them.

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -1,23 +1,40 @@
...
+    sct_detect_pmj.main(argv=['-i', fname_in, '-o', fname_out, '-c', 't2', '-qc', 'testing-qc'])
+
+    im_pmj = Image(fname_out)
+    im_pmj_manual = Image(fname_gt)
+
+    # np.where outputs a tuple with an array for each axis: https://stackoverflow.com/q/50646102
+    # so, convert ==> `tuple(x_array, y_array, z_array)` -> `list([x1, y1, z1], [x2, y2, z2], ...)`
+    label_value = 50
+    gt = [[x, y, z] for x, y, z in zip(*np.where(im_pmj_manual.data == label_value))]
+    pred = [[x, y, z] for x, y, z in zip(*np.where(im_pmj.data == label_value))]
```
kousu, line 29: I find this a bit easier to read. But I'm familiar with `zip`. But I think it's a pretty basic part of python vocabulary.

```suggestion
    gt = list(zip(*np.where(im_pmj_manual.data == label_value)))
    pred = list(zip(*np.where(im_pmj.data == label_value)))
```

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -1,23 +1,40 @@
...
+
+    # ensure that only one coordinate was predicted
+    assert len(gt) == len(pred) == 1
+
+    # transform pixel coordinates to physical coordinates (units in millimeters)
+    gt_phys = im_pmj_manual.transfo_pix2phys(gt)
+    pred_phys = im_pmj.transfo_pix2phys(pred)
+
+    # ensure prediction is within 10mm of ground truth coordinates
+    distances = np.linalg.norm(gt_phys - pred_phys, axis=1)
```
kousu, line 39: I'm a bit lost here:  why `axis=1`?

unit_testing/cli/test_cli_sct_detect_pmj.py
```diff
@@ -1,23 +1,40 @@
...
+    fname_in = 'template/template/PAM50_small_t2.nii.gz'
+    fname_out = 'PAM50_small_t2_pmj.nii.gz'
+    fname_gt = 'template/template/PAM50_small_t2_pmj_manual.nii.gz'
+    sct_detect_pmj.main(argv=['-i', fname_in, '-o', fname_out, '-c', 't2', '-qc', 'testing-qc'])
+
+    im_pmj = Image(fname_out)
+    im_pmj_manual = Image(fname_gt)
+
+    # np.where outputs a tuple with an array for each axis: https://stackoverflow.com/q/50646102
+    # so, convert ==> `tuple(x_array, y_array, z_array)` -> `list([x1, y1, z1], [x2, y2, z2], ...)`
```
kousu, line 26: This might be easier to parse for people.

```suggestion
    # so, convert ==> `([x1, x2, ...], [y1, y2, ...], [z1, z2, ...])` -> `[ (x1, y1, z1), (x2, y2, z2), ...]`
```

Note that I changed the square brackets to round brackets because of what I suggested about `zip` below. It shouldn't change the meaning though.

unit_testing/cli/test_cli_sct_detect_pmj.py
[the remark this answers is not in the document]
kousu: Follow ups:

* https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373/files#r663981130
* https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373/files#r663984299
* https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373/files#r663982885

unit_testing/cli/test_cli_sct_dmri_moco.py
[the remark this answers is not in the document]
kousu: :tada:

unit_testing/cli/test_cli_sct_extract_metric.py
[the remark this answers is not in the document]
kousu:
```
     results = results[~np.isnan(results)]  # Remove non-numeric fields such as filename, SCT version, etc.
```

Nicely played :clap:

unit_testing/cli/test_cli_sct_detect_pmj.py
joshuacwnewton: Oh! Of course, that makes perfect sense.

joshuacwnewton: Ah, wait, now I remember why I did it this way. I wanted to make it a little more explicit that the result was a list of `[x, y, z]` coordinates.

`list()` is definitely more compact, but maybe a little harder to grok (i.e. more dependent on the comment above to explain).

unit_testing/cli/test_cli_sct_detect_pmj.py
joshuacwnewton: Thanks for pointing this out, it was unnecessarily complex. Fixed in e495cc0.

unit_testing/cli/test_cli_sct_detect_pmj.py
joshuacwnewton: > Note that I changed the square brackets to round brackets because of what I suggested about zip below. It shouldn't change the meaning though.

We can just use tuples straight-up anyway! :) 

done in 1b7e636.

kousu (approved):

unit_testing/cli/test_cli_sct_register_to_template.py
kousu: I don't know what to do with this. We can think about this later.

PR #3376, opened 2021-05-05
Fix `-bvals` filepath handling and update bvals argparse help

### LINKED ISSUE #3374: Argparse help string for `sct_dmri_moco -bvals` is misleading
https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/scripts/sct_dmri_moco.py#L78

`bvals` should not be an image file (`.nii.gz`). See [`read_bvals_bvecs`](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/io/gradients.py#L44-L45) for the expected extensions.

### LINKED ISSUE #3375: `-bvals` does not use absolute path, causing a FileNotFoundError
Repro steps:

```
cd sct_example_data/dmri
sct_dmri_moco -i dmri.nii.gz -bvec bvecs.txt -bval bvals.txt -x spline
```

The `bvecs` file is properly expanded, but the `bvals` filepath is still a relative filepath, so it can't be found inside the tmp directory we're using.

https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/scripts/sct_dmri_moco.py#L187-L188

Screenshot from 2021-05-05 11-53-48

There is additional code to do this in `moco.py`, included in #2634, but it was left commented out:

https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/moco.py#L153-L157

This wasn't caught because we have no test for `-bvals` currently, only `-bvecs`:

https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/testing/test_sct_dmri_moco.py#L31-L33

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR:
* Uses absolute path for `-bvals`.
* Edits the argparse help descriptions to fix references to `bvals.nii.gz`

I'd like to write a test for this, but the timing of this is a bit tricky, as it would clash with #3373. Once that PR is complete, writing tests should be more straightforward.

## Linked issues

Fixes #3374.
Fixes #3375.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_dmri_moco.py
+++ b/spinalcordtoolbox/scripts/sct_dmri_moco.py
@@ -184,8 +184,8 @@ def main(argv=None):
 
     # Fetch user arguments
     param.fname_data = arguments.i
-    param.fname_bvecs = os.path.abspath(arguments.bvec)
-    param.fname_bvals = os.path.abspath(arguments.bval)
+    param.fname_bvecs = arguments.bvec
+    param.fname_bvals = arguments.bval
     param.bval_min = arguments.bvalmin
     param.group_size = arguments.g
     param.fname_mask = arguments.m
--- a/spinalcordtoolbox/scripts/sct_dmri_moco.py
+++ b/spinalcordtoolbox/scripts/sct_dmri_moco.py
@@ -185,7 +185,7 @@ def main(argv=None):
     # Fetch user arguments
     param.fname_data = arguments.i
     param.fname_bvecs = os.path.abspath(arguments.bvec)
-    param.fname_bvals = arguments.bval
+    param.fname_bvals = os.path.abspath(arguments.bval)
     param.bval_min = arguments.bvalmin
     param.group_size = arguments.g
     param.fname_mask = arguments.m
--- a/spinalcordtoolbox/scripts/sct_dmri_moco.py
+++ b/spinalcordtoolbox/scripts/sct_dmri_moco.py
@@ -75,7 +75,7 @@ def get_parser():
         '-bval',
         metavar=Metavar.file,
         default=param_default.fname_bvals,
-        help='Bvals file. Example: bvals.nii.gz',
+        help='Bvals file. Example: bvals.txt',
     )
     optional.add_argument(
         '-bvalmin',
--- a/spinalcordtoolbox/moco.py
+++ b/spinalcordtoolbox/moco.py
@@ -161,6 +161,16 @@ def moco_wrapper(param):
         im_mask.save(os.path.join(path_tmp, file_mask), mutable=True, verbose=param.verbose)
         # Update field in param (because used later in another function, and param class will be passed)
         param.fname_mask = file_mask
+    if param.fname_bvals != '':
+        _, _, ext_bvals = extract_fname(param.fname_bvals)
+        file_bvals = f"bvals.{ext_bvals}"  # Use hardcoded name to avoid potential duplicate files when copying
+        copyfile(param.fname_bvals, os.path.join(path_tmp, file_bvals))
+        param.fname_bvals = file_bvals
+    if param.fname_bvecs != '':
+        _, _, ext_bvecs = extract_fname(param.fname_bvecs)
+        file_bvecs = f"bvecs.{ext_bvecs}"  # Use hardcoded name to avoid potential duplicate files when copying
+        copyfile(param.fname_bvecs, os.path.join(path_tmp, file_bvecs))
+        param.fname_bvecs = file_bvecs
 
     # Build absolute output path and go to tmp folder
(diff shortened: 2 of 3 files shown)


joshuacwnewton:
Oops. The default value is `''`, so `os.path.absolutepath` evaluates to the `cwd`. I'll fix this shortly.

spinalcordtoolbox/scripts/sct_dmri_moco.py
```diff
@@ -185,7 +185,7 @@ def main(argv=None):
     # Fetch user arguments
     param.fname_data = arguments.i
     param.fname_bvecs = os.path.abspath(arguments.bvec)
-    param.fname_bvals = arguments.bval
+    param.fname_bvals = os.path.abspath(arguments.bval)
```
Drulex, line 188: Normally this wouldn't hurt, but in this case if `bval` is not specified by the user it is set to an empty string automatically by `moco.py` (this is not a sane default value): https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/spinalcordtoolbox/moco.py#L56

If expanded it resolves the absolute path of the current working directory (normally) which is wrong.

Looking at the batch processing log (https://github.com/neuropoly/spinalcordtoolbox/runs/2511298710?check_suite_focus=true#step:4:3136) it seems to have been resolved to `sct_example_data/dmri`.. which I don't understand how, maybe because of the use of `chdir()` in the code? (if this is the case then it's definitely a recipe for disaster)

joshuacwnewton: I should have waited for the CI to pass before requesting a review, aha. :sweat_smile: 

> it seems to have been resolved to sct_example_data/dmri.. which I don't understand how, maybe because of the use of chdir() in the code? (if this is the case then it's definitely a recipe for disaster

`batch_processing.sh` does all of its processing directly in `sct_example_data`:

https://github.com/neuropoly/spinalcordtoolbox/blob/67170203934274498b90bc65f6a084f724b6ec19/batch_processing.sh#L186

Drulex: I'm not a fan of the proposed fix. It is necessary only because the code performs unexpected `chdir()` which causes a maintenance and debugging burden.

Drulex: > `batch_processing.sh` does all of its processing directly in `sct_example_data`:

Ah yes I see.

joshuacwnewton: > I'm not a fan of the proposed fix. It is necessary only because the code performs unexpected `chdir()` which causes a maintenance and debugging burden.
> 
> edit: not saying that the fix will not work, but if we should at least consider refactoring the `chdir()` in another issue if not in this one.

I had the exact same thought. See the extended commit message in b6ebe82.

joshuacwnewton: I wasn't sure how to scope this, as I have a feeling `moco.py` could use a larger refactor. It does a lot of file I/O related tasks that might be better suited for `sct_dmri_moco`. But, I didn't want this to spiral. :p

joshuacwnewton: > we should at least consider refactoring the chdir() in another issue if not in this one.

I did a quick regex search for `file_.* =` and `fname_.* =` and there are 30+ instances of relative filepaths being set in `moco_wrapper` -- we'd have to modify all of them if we stopped changing the working directory. It's doable, but it would expand the scope of this PR considerably. And, there are some file paths that are bit tricky to modify... `file` and `fname` don't seem to be used 100% consistently according to the [Naming conventions](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Naming-conventions-(variables,-files)).

I think it might be better to focus on fixing this bug for 5.3.1, and save a refactor for a future PR.

jcohenadad: > I think it might be better to focus on fixing this bug for 5.3.1, and save a refactor for a future PR.

👍

Drulex (approved):
Approving because it fixes the initial issue, but I think it's a good idea to:
- use more sane defaut values (`''` is not the same as `None` which is more appropriate and would avoid potentially silent problems like the one we saw with the initial tentative fix using `abspath()`)
- get rid of `chdir()` as it also brings its own set of potentially hard to debug problems.

PR #3379, opened 2021-05-07
Modify `zeros_like` to no longer make a copy, and to avoid rescaling

### LINKED ISSUE #3110: Function `zeros_like()` performs an unnecessary, noisy rescaling operation
See also https://github.com/neuropoly/spinalcordtoolbox/pull/3104#issuecomment-747763574.

`spinalcordtoolbox/image.py` has a function called `zeros_like`. It creates a copy of the image (to preserve headers, etc.) using the `change_type` function, then sets the data values to 0.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/11249f7b9eab4791cd1834c537cfafd1379bac02/spinalcordtoolbox/image.py#L1227-L1229

`change_type` has some additional rescaling logic to handle quantization errors that also warns the user. Normally this is useful for non-zero images, but since we're just going to be setting data to 0 anyway, it's unnecessary.

So, I think it might be better to do this instead, which should also prevent the warning from firing.

```
dst = img.copy()
dst.data[:] = 0 
dst = dst.change_type(dtype)
return dst 
```

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR follows the suggestion in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3110#issuecomment-747822584 to make 2 small improvements: 

1. No longer make a copy of the image, which is accomplished by using the `.change_type()` method instead of the `change_type()` function. (So, now the docstring is accurate, when it wasn't before.)
2. No longer set data to 0 after changing type. (This prevents the `change_type` function from unnecessarily rescaling the data.)

## Linked issues

Fixes #3110

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1245,7 +1245,10 @@ def zeros_like(img, dtype=None):
     intent and avoid doing a copy, which is slower than initialization with a constant.
 
     """
-    return Image(np.zeros_like(img.data), hdr=img.hdr).change_type(dtype)
+    zimg = Image(np.zeros_like(img.data), hdr=img.hdr)
+    if dtype is not None:
+        zimg.change_type(dtype)
+    return zimg
 
 
 def empty_like(img, dtype=None):
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1245,9 +1245,7 @@ def zeros_like(img, dtype=None):
     intent and avoid doing a copy, which is slower than initialization with a constant.
 
     """
-    dst = change_type(img, dtype)
-    dst.data[:] = 0
-    return dst
+    return Image(np.zeros_like(img.data), hdr=img.hdr).change_type(dtype)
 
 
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/image.py
```diff
@@ -1245,9 +1245,7 @@ def zeros_like(img, dtype=None):
     intent and avoid doing a copy, which is slower than initialization with a constant.
 
     """
-    dst = change_type(img, dtype)
-    dst.data[:] = 0
-    return dst
+    return Image(np.zeros_like(img.data), hdr=img.hdr).change_type(dtype)
```
Drulex, line 1248: LGTM.

Potential nitpick: if you write it this way:
```python
zimg = Image(np.zeros_like(img.data), hdr=img.hdr)
if dtype is not None:
    zimg.change_type(dtype)
return zimg
```
It makes it obvious that `change_type(dtype=None)` has no potential side-effect. Right now one would need to look into `Image.change_type()` to make sure of that.

Drulex (approved):

PR #3387, opened 2021-05-12
Add test for `sct_dmri_display_bvecs`, then add `-v` argument to make the test pass

joshuacwnewton:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [X] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [X] I've assigned a reviewer

#### PR contents

- [X] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [X] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [X] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

#3091 added loglevel setting steps to all SCT scripts. However, `sct_dmri_display_bvecs` never had a verbose argument.

This was never caught because there is no test for `sct_dmri_display_bvecs`. So, this adds an (admittedly very basic) test for the script. (A more thorough test would require some refactoring of `sct_dmri_display_bvecs`.)

This test fails on master, but passes on this branch after `-v` is added.

## Linked issues

Fixes #3386.

PR #3390, opened 2021-05-14
Allow `sct_dmri_display_bvecs.py` to display multi-shell acquisition

valosekj:
#### GitHub

#### PR contents

- [x] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution - create another test(s), see [comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390#discussion_r633883397) besides [current test](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/unit_testing/cli/test_cli_sct_dmri_display_bvecs.py)

## Description

After the fix of recent [bug](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3387) in `sct_dmri_display_bvecs.py` I have played with the script little bit.

I needed to display gradient directions for multi-shell sequence but the current version works only with `bvecs` file and does not take into account information about shells from `bvals` file.

Thus, I little bit extended the behaviour and added optional argument `-bval`.

The original behaviour (when only `-bvec` file is passed) is preserved.

<details>
<summary> Screenshots (now outdated due to colour change) </summary>

Output from original [master](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py) branch (single-shell):

bvecs-single_shell-no_bval_file-master

Output from [jv/sct_dmri_display_bvecs-multishell](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jv/sct_dmri_display_bvecs-multishell) branch with only `-bvec` option (single-shell), note - (almost) same behaviour as current master:

bvecs-single_shell-no_bval_file-jv

Output from [jv/sct_dmri_display_bvecs-multishell](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jv/sct_dmri_display_bvecs-multishell) branch with both `-bval` and `-bvec` options (single-shell), note - axis are changed and legend with bvalue is displayed:

bvecs-single_shell-bval_file_passed-jv

Output from [jv/sct_dmri_display_bvecs-multishell](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jv/sct_dmri_display_bvecs-multishell) branch with both `-bval` and `-bvec` options (multi-shell):

bvecs-multishell-bval_file_passed-jv

</details>

What do you think about that? Is it suitable for master branch or should I keep that only in my fork?


jcohenadad:
> What do you think about that? Is it suitable for master branch or should I keep that only in my fork?

This is super relevant, thank you for your contribution @valosekj 🙏

jcohenadad (commented):
I find the colors a bit distracting though-- it took me a while to realize that the b-shell is encoded by the shape, not the color. My suggestion would be to do it the other way around, ie: use a single color per b-shell (and use the same shape)

valosekj:
> I find the colors a bit distracting though-- it took me a while to realize that the b-shell is encoded by the shape, not the color. My suggestion would be to do it the other way around, ie: use a single color per b-shell (and use the same shape)

You are right. I have similar feeling.

Multi-shell scheme looks much more clear when colors code shell instead of symbols:

bvecs-multishell-bval_file_passed-jv_colors

What about single-shell acquisition? Should I preserve original color coding (i.e., [different colors](https://user-images.githubusercontent.com/39456460/118304189-a5ad2e00-b4e6-11eb-98a9-47a1bbd845a6.png) for each effective direction), or should I switch to same color for whole shell, like this?

bvecs-single_shell-no_bval_file-jv_colors

jcohenadad:
> What about single-shell acquisition? Should I preserve the original color coding (i.e., different colors for each effective direction), or should I switch to same color for whole shell, like this?

switch to single color to make it more consistent. Thanks!

valosekj:
> > What about single-shell acquisition? Should I preserve the original color coding (i.e., different colors for each effective direction), or should I switch to same color for whole shell, like this?
> 
> switch to single color to make it more consistent. Thanks!

Switched in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/8cc3998049134e09de7777f01b9f21d21aa91edb. Output figures now look as [follow](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390#issuecomment-841849490). PR is probably ready for review. I have assigned you and @joshuacwnewton as a reviewers.

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -140,4 +221,3 @@ def main(argv=None):
 if __name__ == "__main__":
     init_sct()
     main(sys.argv[1:])
-
```
joshuacwnewton, line 143: Would you like to try writing a test for this script? There's an existing test in `unit_testing/cli/test_cli_dmri_display_bvecs.py` that you can base the new test off of.

Also, if you want to try something new while writing the test, I've just discovered that `matplotlib` provides the [`@image_comparison` and `@check_figures_equal`](https://matplotlib.org/stable/devel/testing.html#writing-an-image-comparison-test) decorators. I'm really curious to try this out, so I'd be happy to try writing a test for this too. Either way!

joshuacwnewton (changes requested):
I've left a lot of comments, but most of them are nitpicks, so I think you've done a really nice job here. Thank you so much for doing this! :tada:

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
+
+    # Assign scatter markers to unique bvals
+    shell_colors = {}
+    index = 0
+    for unique_bval in np.unique(bvals):
+        # skip b=0
+        if unique_bval < bzero:
+            continue
+        shell_colors[unique_bval] = colors[index]
+        index += 1
```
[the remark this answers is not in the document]
valosekj: Great point! I was considering the proper solution. Yes, there's theoretical possibility to have >6 shells in diffusion sequence. Then, it would fail. I'll check `cycler`, thanks!

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -10,29 +10,45 @@
...
 import matplotlib.pyplot as plt
 from dipy.data.fetcher import read_bvals_bvecs
+from matplotlib.lines import Line2D
 
 from spinalcordtoolbox.utils import SCTArgumentParser, Metavar, init_sct, printv, set_global_loglevel
 
+# Switch matplotlib backend on MacOS in PyCharm
+if sys.platform == 'darwin':
+    import matplotlib
+    matplotlib.use('TkAgg')
```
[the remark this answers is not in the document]
jcohenadad: Interesting, I have just tested this function on the master branch (I haven't tried it for a while 😅 ), and I confirm about the issue-- no image pops up:
~~~
julien-macbook:~/sct_example_data/dmri $ sct_dmri_display_bvecs -bvec bvecs.txt 

--
Spinal Cord Toolbox (git-master-a62fb672c602b9863a8c03521eb272bfd7bad9e1)

sct_dmri_display_bvecs -bvec bvecs.txt
--

Saving figure: bvecs.png

/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py:137: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  plt.show()
~~~

if we fix it in this PR i would explicitly mention it in the body (and title) of the PR

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
+    # count number of bvals for individual shells
+    bvals_per_shell = pd.Series(bvals[bvals > bzero]).value_counts()
+
+    # Create single custom legend for whole figure with several subplots
+    lines = list()  # initialize list for individual symbols in the legend
+    labels = list()  # initialize list for individual text labels in the legend
+    # Loop across legend elements
+    for key, value in shell_colors.items():
+        lines.append(Line2D([0], [0], color=value, marker='o', markersize=8, alpha=0.7, linestyle=''))
+        labels.append('b-values = {} (n = {})'.format(str(int(key)), str(bvals_per_shell[key])))
```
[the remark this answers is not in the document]
valosekj: Excellent, thanks. I'll simplify the code based on this recommendations!

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
+        fname_bvals = arguments.bval
+        # Read bvals and bvecs files
+        bvals, bvecs = read_bvals_bvecs(fname_bvals, fname_bvecs)
+    else:
+        # Read only bvecs file
+        _, bvecs = read_bvals_bvecs(None, fname_bvecs)
 
-    # Read bvecs
-    bvecs = read_bvals_bvecs(fname_bvecs, None)
-    bvecs = bvecs[0]
```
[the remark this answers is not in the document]
valosekj: Yes, the order was wrong. It resulted that bvecs file (file with three lines) was read as a single line file. But code kind of worked 😄

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: To be honest, I have only basic experience with writing tests. But definitely I would like to learn it more deeply! So, we can collaborate on that.

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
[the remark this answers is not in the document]
joshuacwnewton: I also get this error on Ubuntu 20.04, and I can confirm that the change above fixes this for me, too. 

I think I would prefer if this were fixed in a follow-up PR, though. The scope of backend changes is quite a bit bigger than just this script, and probably requires a more thorough investigation (especially given the other backend issues we've had).

I've opened an issue for this here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3392

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: > But definitely I would like to learn it more deeply! So, we can collaborate on that.

I'm glad to hear this! :tada: 

As a starting point, we have some internal documentation on test writing [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing). If anything is confusing about those pages, feel free to message on Slack! (I'm always looking to tweak the pages to make them easier to understand, so any questions are very helpful.)

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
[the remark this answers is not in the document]
valosekj: Hm, I haven't found satisfactory workaround with `cycler` library, it looks it can iterate only across predefined list of colors (don't solve our problem). But I haven't study it so deeply.
Instead I found quick workaround using `matplotlib.cm` and iterator, see commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/a6886da7caa1efb329701d0493bde9c0005162fd (and inspiration [here](https://www.kite.com/python/answers/how-to-set-different-colors-for-each-series-in-a-matplotbib-scatter-plot-in-python))

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -123,14 +195,23 @@ def main(argv=None):
...
     plt.title('3D view (use mouse to rotate)')
-    plt.axis('off')
+    plt.axis('on')
     # plt.draw()
 
+    # add legend with b-values if bvals file was passed
+    if arguments.bval is not None:
+        create_custom_legend(fig, shell_colors, bvals)
+    else:
+        plt.tight_layout()
```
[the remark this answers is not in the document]
valosekj: Solved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/1d9ecdb7125784ba2c18afeefb21258976255ab3

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
+    labels = list()  # initialize list for individual text labels in the legend
+    # Loop across legend elements
+    for key, value in shell_colors.items():
+        lines.append(Line2D([0], [0], color=value, marker='o', markersize=8, alpha=0.7, linestyle=''))
+        labels.append('b-values = {} (n = {})'.format(str(int(key)), str(bvals_per_shell[key])))
+
+    plt.tight_layout()
+    # Insert legend below subplots, NB - this line has to be below the plt.tight_layout()
+    legend = fig.legend(lines, labels, loc='lower left', bbox_to_anchor=(0.1, 0),
+                        bbox_transform=plt.gcf().transFigure, ncol=len(lines), fontsize=10)
```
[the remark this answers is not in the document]
valosekj: Solved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/1d9ecdb7125784ba2c18afeefb21258976255ab3

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
     set_global_loglevel(verbose=verbose)
 
     fname_bvecs = arguments.bvec
+    if arguments.bval is not None:
+        fname_bvals = arguments.bval
+        # Read bvals and bvecs files
+        bvals, bvecs = read_bvals_bvecs(fname_bvals, fname_bvecs)
+    else:
+        # Read only bvecs file
+        _, bvecs = read_bvals_bvecs(None, fname_bvecs)
```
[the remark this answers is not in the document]
valosekj: Solved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/9e14961c1c708608bd93cb476c8ed11d4dff4439

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +68,93 @@ def get_parser():
...
+
+    # bvals file was passed
+    if arguments.bval is not None:
+        # multiply unit b-vectors by b-values
+        x, y, z = bvecs[0] * bvals, bvecs[1] * bvals, bvecs[2] * bvals
+    # bvals file was not passed
+    else:
+        x, y, z = bvecs[0], bvecs[1], bvecs[2]
+        # create dummy unit bvals array (necessary fot scatter plots)
+        bvals = np.repeat(1, bvecs.shape[1])
```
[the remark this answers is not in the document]
valosekj: I am not sure, because there has to be check for bvecs transposition:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a6886da7caa1efb329701d0493bde9c0005162fd/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py#L130-L131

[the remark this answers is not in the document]
joshuacwnewton: Ah, good point. Hrm.... Well, could that statement be moved too?

I also realize that this check could be simplified as well:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a6886da7caa1efb329701d0493bde9c0005162fd/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py#L133-L141

Since for the `else:` case, if `bvals` becomes a dummy array of 1s, then multiplying would be the same as `bvecs[0], bvecs[1], bvecs[2]`, I think? So perhaps all together it could look something like:

```python
    fname_bvecs = arguments.bvec
    fname_bvals = arguments.bval
    bvals, bvecs = read_bvals_bvecs(fname_bvals, fname_bvecs)
    # if first dimension is not equal to 3 (x,y,z), transpose bvecs file
    if bvecs.shape[0] != 3: 
        bvecs = bvecs.transpose() 
    # create dummy unit bvals array if bvals wasn't specified (necessary for scatter plots)
    if bvals is None:
        bvals = np.repeat(1, bvecs.shape[1])
    # multiply unit b-vectors by b-values
    x, y, z = bvecs[0] * bvals, bvecs[1] * bvals, bvecs[2] * bvals

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (joshuacwnewton) [reply spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py:27]: @valosekj Could you remove the code to change the backend, and also temporarily comment out "`plt.show()`"?

There are also 15 instances of `plt.show()` in SCT (with many also commented out), so I'd like to try to address them all at once in a separate PR. (If there's a fix for one script, it should fix all of them.)

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py (hunk 1 above)
Comment (joshuacwnewton) [reply spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py:143]: Actually, now that I think about it a little more... I take back my suggestion about `@image_comparison` and `@check_figures_equal`... I'm not sure if they're very useful in this instance.

This is a pretty self-contained script: No other SCT code depends on it, and it doesn't really depend on any SCT code. So, the only time the image contents are likely to change is if we're making an intentional change... in which case, testing against the image contents isn't really useful.

I think it might be good enough to just check for the existence of `bvecs.png` (like what is done in the existing test).

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (valosekj) [reply spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py]: @joshuacwnewton Sure, commented in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/59615de72bc5f7276ecae59a389aa2b08cd691bc

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (valosekj) [reply spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py]: Thanks one more time, simplified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/6384472341666c62e66cb0f647b68f50084dafb4

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (valosekj) [reply spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py]: > Since for the else: case, if bvals becomes a dummy array of 1s, then multiplying would be the same as bvecs[0], bvecs[1], bvecs[2], I think?

@joshuacwnewton of course, you are right! :-) Refactored in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/057227447e8924fc214bfc211437cba0870e70c9

File: spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py [hunk 10]
```diff
@@ -52,37 +61,82 @@ def get_parser():
...
+    # multiply unit b-vectors by b-values
+    x, y, z = bvecs[0] * bvals, bvecs[1] * bvals, bvecs[2] * bvals
+
+    # Set different color for each shell (bval)
+    shell_colors = {}
+    # Create iterator with different colors from brg palette
+    colors = iter(cm.brg(np.linspace(0, 1, len(np.unique(bvals)))))
+    for unique_bval in np.unique(bvals):
+        # skip b=0
+        if unique_bval < bzero:
```
joshuacwnewton, line 137: This suggestion is outside the scope of this PR, but it's a tiny change that might be nice to do while we're here.

Since `bzero` is a module-scoped (global) constant, what do you think of renaming it to `BZERO_THRESH`? While reviewing, I got a bit confused wondering where `bzero` was defined (because it was all lowercase). So, the rename might make it clearer.

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +61,82 @@ def get_parser():
...
+    # if bvals file was not passed, create dummy unit bvals array (necessary fot scatter plots)
+    if bvals is None:
+        bvals = np.repeat(1, bvecs.shape[1])
+    # multiply unit b-vectors by b-values
+    x, y, z = bvecs[0] * bvals, bvecs[1] * bvals, bvecs[2] * bvals
+
+    # Set different color for each shell (bval)
+    shell_colors = {}
+    # Create iterator with different colors from brg palette
+    colors = iter(cm.brg(np.linspace(0, 1, len(np.unique(bvals)))))
```
joshuacwnewton, line 134: Nice!

I tried this out with >6 unique bvals (just manually editing the bvals.txt file in sct_example_data), and I noticed that because the colors in the BRG colormap are a 3-color gradient, the intermediate colors were a little hard for me to tell apart:

image

I looked at [the matplotlib colormap list](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/files#) and tried out a few other colomaps. `cm.nipy_spectral` felt the most distinct to me, but this is a bit personal preference too.

image

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -52,37 +61,82 @@ def get_parser():
...
+    # Loop across legend elements
+    lines = [Line2D([0], [0], color=color, marker='o', markersize=8, alpha=0.7, linestyle='')
+             for color in shell_colors.values()]
+    labels = [f'b-values = {shell} (n = {bvals_per_shell[shell]})'
+              for shell in shell_colors.keys()]
+
+    # Insert legend below subplots, NB - this line has to be run after the plt.tight_layout() which is called before
+    # function call
+    legend = fig.legend(lines, labels, loc='lower left', bbox_to_anchor=(0.1, 0),
+                        bbox_transform=plt.gcf().transFigure, ncol=len(lines), fontsize=10)
```
joshuacwnewton, line 100: Currently, the legend doesn't wrap, so if there are a lot of entries, it will go off the image. :sweat_smile: 

You can see examples of this in <insert link to other review comment>

joshuacwnewton (changes requested):
Thanks for addressing the first feedback so quickly! :)

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: Thanks for great suggestion! Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/64508ff17fc3fda2c2c3efc3a2d2bee938c02917

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: Yeah. Thanks for this catch! I have fixed this by setting fix number of columns in the legend to 3. See, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/323fa5d8f8389043525e9ebf240851fb91800ba2

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: @joshuacwnewton Thanks for this test with >6 unique bvals! Indeed, the workaround with `linspace` has this disadvantage that color palette is changing unpredictably based on number of unique bvals. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d68efd6b68951cb54d3303b9767c7ac85458455d/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py#L134

I will try to figure out some more elegant solution. But usually, common spherical multi-shell DWI sequences have up to 3 unique bvalues. (but of course, some [DSI q-space](https://www.researchgate.net/profile/Jian-Cheng/publication/230569009/figure/fig3/AS:437671634968579@1481360457930/Several-kinds-of-sampling-in-q-space-The-black-dot-in-q-0-0-0T-is-the-baseline.png) can have much more bvalues. So it is question how much time to invest to this thing 😄)

joshuacwnewton: Actually, now that I think about it, I should probably ask: Is it even important to tell the b-values apart? 

Maybe I misunderstood the purpose of the image. Since the bvals are ordered in the legend by their magnitude, it's true that the first colormap does effectively convey the increase (from low bvals to high bvals).

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: Splendid! :)

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: What do you think of something like this, @valosekj? https://stackoverflow.com/questions/61084381/create-gradient-legend-matplotlib

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: @joshuacwnewton Hm, however I have spotted another glitch regarding to the legend 🙉  When the legend contains more than 5-6 rows, it overlaps axis in figure. It is possibly caused by fixed position of the legend relative to the axis:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/323fa5d8f8389043525e9ebf240851fb91800ba2/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py#L105

<details> 
<summary> see here </summary>

bvecs

</details> 

I have tried to play with `tight_layout()` command, but without success...
But again, I don't assume that so many bvals will be commonly passed.

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: @joshuacwnewton The purpose of the image (at least for me) is to display how individual diffusion directions (bvecs file) and individual diffusion weighting (bvals file) are distributed over the q-space. Because from these two text files (bvals and bvecs) I don't get this information easily.
Important (again for me) is to be able to easily distinguish between individual shells (b-values).

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: > But again, I don't assume that so many bvals will be commonly passed.

In that case, perhaps we can leave it as-is for now, but add a comment explaining the limitation (e.g. maybe in the docstring of the legend function) to refresh our memory if we ever need to return to this in the future (e.g. if a user runs into this issue, then mentions it on the forum).

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: > Important (again at least for me) is to be able to easily distinguish between individual shells (b-values).

I appreciate the clarification! In that case, do you think switching the colormap to `cm.nipy_spectral` might help to make the shells more distinguishable? (see: figure in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390#discussion_r634835644)

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
joshuacwnewton: @valosekj I think the easiest way to add a test would be to just copy and paste the existing test, but change the name + the arguments to include bvals from the example dataset. 

In this case, we don't have to get too fancy, hehe. :sweat_smile:

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: 👍🏻 Color map switched in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/eef85738bde5a304736d8bacab68ac0dd92e945b

spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
valosekj: Thanks 👍🏻 I added this limitation into docstring in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3390/commits/a497719eec1164e5b8de57768eb7e20d9f12b85a

joshuacwnewton (approved):
LGTM! Thank you again for the contribution. :smile:

jcohenadad:
thank you @valosekj !

joshuacwnewton:
Depending on how we want to handle the next release, we can either:

1. Wait to merge this, release 5.3.1 ASAP with recent bugfixes, then include this as part of 5.4.0.
2. Merge this now, and make the next release 5.4.0 instead.

I don't have super strong feelings either way...

joshuacwnewton:
> Depending on how we want to handle the next release, I think we can either:
> 
>     1. Wait to merge this, release 5.3.1 ASAP with recent bugfixes, then include this as part of 5.4.0.
> 
>     2. Merge this now, and make the next release 5.4.0 instead.
> 
> I don't have super strong feelings either way, so I'm happy to go with either option.

As discussed in the most recent SCT meeting, we will not use a patch release, and instead release v5.4 next. So, there is no need to delay merging this PR. :slightly_smiling_face:

joshuacwnewton:
Thank you again for taking the time to make this contribution, @valosekj! :tada:

PR #3094, opened 2020-12-13
Add multiple tutorial pages covering the "Registration" material from the SCT course

joshuacwnewton:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [ ] I've assigned a reviewer

#### PR contents

- [X] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [X] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [X] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR creates RTD pages from the material from the "Registration" section, which gets us 2/3 of the way through the course material. (Timestamps: [0:42:47 - 2:06:51](https://www.youtube.com/watch?v=whbtjYNtHko&t=2567s), PDF Pages: [42-80](https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT_Course_20200121).)

This section was much longer than I expected, so I had to split it into 4 new tutorial pages + several side pages (warping fields, qc reports, pam50 template, `sct_register_to_template` guide, etc.)

You can view the specific tutorial pages here: https://spinalcordtoolbox--3094.org.readthedocs.build/en/3094/user_section/tutorials.html

## Linked issues

Partially addresses #2857.
Follows #2991.
Fixes #3265.

PR #3410, opened 2021-06-05
Don't set the global loglevel when CLI scripts are called in-code

### LINKED ISSUE #3341: Calling an SCT script using `.main()` changes the global logging level
At the start of each CLI script, we set the global logging level using the parsed verbose argument. For example:

https://github.com/neuropoly/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/scripts/sct_apply_transfo.py#L341-L344

This works as intended when calling scripts via the command line. However, when calling a script in code via `.main()`, changing the global logging level like this feels more like an unexpected side effect. 

For example, take this call in `moco.py`:

https://github.com/neuropoly/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/moco.py#L768-L773

`-v 0` is set for `sct_apply_transfo`, but because it gets set globally, `-v 0` leaks outside of this call to the rest of moco, undoing whatever was initially set for `param.verbose` in moco.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Currently, we set the global loglevel at the start of every CLI script. We do this to avoid having to repeatedly set the loglevel for every single child module.

However, this is a problem for in-code CLI script calls like this one:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/916ab591ca2e35cab9f5f27ec2439ce8fdb0b9d2/spinalcordtoolbox/moco.py#L768-L773

Because we set the loglevel globally, the `-v 0` change leaks outside of `sct_apply_transfo`. This means that any `logger.info` messages that come afterwards won't show up.

So, this PR makes sure that the global loglevel is _only set_ when we're calling a CLI script directly from the command line. (3c21fe8)

#### Note about "55 files changed"

(Note: Most of the file changes are due to a7bf27c. The actual substantial changes are quite small.)

## Linked issues

Fixes #3341.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -21,7 +21,7 @@ from skimage.measure import label
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.centerline.core import ParamCenterline, get_centerline
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, ActionCreateFolder
-from spinalcordtoolbox.utils.sys import init_sct, printv, set_global_loglevel
+from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, extract_fname, copy, rmtree
 
 
@@ -515,7 +515,7 @@ def main(argv=None):
     parser = get_parser()
     arguments = parser.parse_args(argv)
     verbose = arguments.v
-    set_global_loglevel(verbose=verbose)
+    set_loglevel(verbose=verbose)
 
     fname_mask = arguments.m
     fname_sc = arguments.s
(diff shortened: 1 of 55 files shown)


joshuacwnewton:
@ahill187 Could you check if this fixes the problem you noticed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3094#discussion_r643240157?

kousu (approved):

kousu:
This is a hacky workaround. @joshuacwnewton and I talked about some better solutions:

- figure out a way to not use a global logger at all: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2853
- previously, this *used* to spawn a subprocess, which isolated the verbosity level changes
- if we can't rely on clone-and-set like the subprocess case, wrap every main() in a set-reset (even write it as a [@contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)) that saves and restores the global log level
- separating the algorithm part from the `main()` CLI part; then instead of calling `sct_apply_transfo.main(argv=['-i', file_src, ...` just call `sct_apply_transfo(file_src, ....`, which sidesteps the issue

Also related: how many "levels" do we want to support (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2676)? Particularly, this part needs to be rethought:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c21fe8781aef003566c64d2b4dec0f6ef308613/spinalcordtoolbox/utils/sys.py#L64-L67

There's way more work to solve here than in just one PR, so this is good enough for now! Thanks for all your hard work.

PR #3418, opened 2021-06-11
Display soft segmentation in qc report

### LINKED ISSUE #3417: Display non binary segmentation in QC report
## Context
Qc report with `sct_deepseg_sc` only displays a binary segmentation even if the input segmentation is non binary. Since we are currently working on spinal cord softseg (see [here](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord) and [here](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/issues/2)), it would be useful to display softseg in the QC report with `sct_deepseg_sc`. 

## Suggestion
The QC report could display something like this:
image

sandrinebedard:
#### GitHub

#### PR contents

## Description

The purpose of this PR is to add the option to display a soft segmenation in the qc report.

What is done:

- [x] Add a colormap to look like the red colormap in FSLeyes (like the image in #3417).
- [x] Remove binarization of the segmentation.

What is left:

- [x] I wasn't sure what was the best way to differentiate softseg and binary seg --> Either add a flag/check if the segmentation is soft or not to binarize the segmentation after linear interpolation.
- [x] Choose the colormap (I commented 2 other options).

#### What it looks like with current color map:
On T1w:
image

On T2w:
image

## Linked issues

Resolves #3417

### DIFF
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -63,9 +63,9 @@ class QcImage(object):
                      "#a22abd", "#d58240", "#ac2aff"]
     _seg_colormap = ["#4d0000", "#ff0000"]
 
-    
+
     def __init__(self, qc_report, interpolation, action_list, process, stretch_contrast=True,
-                stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
+                 stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
         """
         :param qc_report: QcReport: The QC report object
         :param interpolation: str: Type of interpolation used in matplotlib
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -61,13 +61,11 @@ class QcImage(object):
                      "#9f89b0", "#e08e08", "#3d2b54",
                      "#7d0434", "#fb1849", "#14aab4",
                      "#a22abd", "#d58240", "#ac2aff"]
-    # _seg_colormap = ["#660000", "#ff0000"]  # Dark red
-    # _seg_colormap = ["#000000", "#ff0000"]  # With Black
-    _seg_colormap = ["#4d0000", "#ff0000"]  # Darker red
+    _seg_colormap = ["#4d0000", "#ff0000"]
 
     
     def __init__(self, qc_report, interpolation, action_list, process, stretch_contrast=True,
-                 stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
+                stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
         """
         :param qc_report: QcReport: The QC report object
         :param interpolation: str: Type of interpolation used in matplotlib
@@ -136,9 +134,8 @@ class QcImage(object):
         img = np.ma.masked_equal(mask, 0)
         ax.imshow(img,
                   cmap=color.LinearSegmentedColormap.from_list("", self._seg_colormap),
-                  norm=color.Normalize(vmin=0, vmax=1),
+                  norm=color.Normalize(vmin=0.5, vmax=1),
                   interpolation=self.interpolation,
-                  # alpha=1,
                   aspect=float(self.aspect_mask))
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
Hmm. My first impression is that the dark red can be difficult to see in some of the slices in the mosaic view. This wasn't a problem when zoomed in (e.g. the image in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3417), but here the images are quite small, and the backing anatomical image can be dark grey/black, so for specific slices it doesn't contrast very well. :sweat: 


What would it look like if we inverted the colormap? (Going from dark -> light red instead of light -> dark red.)

jcohenadad:
My primary concern is that 99% of the people seeing this will think the cord is over-segmented and won't understand the concept of a non-binary segmentation. So my suggestion would be to keep the soft colormap but still threshold the segmentation at, let's say, 0.5.

sandrinebedard:
> My primary concern is that 99% of the people seeing this will think the cord is over-segmented and won't understand the concept of a non-binary segmentation. So my suggestion would be to keep the soft colormap but still threshold the segmentation at, let's say, 0.5.

@jcohenadad But currently, the user has to input a non-binary seg in sct_qc or use the flag -thr -1 in sct_deepseg_sc to do that, so won't they be aware of what is a softseg in that case? (this may be a more important concern once we implement softseg in sct_deepseg I agree)

A colorbar could also help with this!

I'll test out applying the threshold at 0.5!

joshuacwnewton:
> hmm, I feel this will be more confusing for users since the binary segmentation is normally red, but the contrast is definitely better, I can maybe try another color than dark red?

Totally agree with everything you've said. :slightly_smiling_face: 

Maybe we could go from "medium red" to "lighter red" so to speak?

jcohenadad:
> hmm, I feel this will be more confusing for users since the binary segmentation is normally red, but the contrast is definitely better, I can maybe try another color than dark red?

i wouldn't go that route-- that's too confusing for users

jcohenadad:
> (this may be a more important concern once we implement softseg in sct_deepseg I agree)

indeed. But even for your investigations, i think the 0.5 threshold is a good compromise. It will give you an idea of how well the segmentation fits the anatomy.

in fact, we are currently using non-binary overlay on the QC for `sct_warp_template` function. You can reuse the same function to display the overlay (appears in blue, but you can replace with red for the SC seg)

sandrinebedard:
> Maybe we could go from "medium red" to "lighter red" so to speak?

Here I applied the threshold at 0.5 too. Here is the color map from red to light (well pink actually )

On T2w:
image
On T1w:
image

I think that going with brighter or darker red doesn't solve the problem since the contrast will always be bad for one image contrast ... But with dark red, the contrast was still better I think. 

We could try with red and a different bright color too?

jcohenadad:
> I think that going with brighter or darker red doesn't solve the problem since the contrast will always be bad for one image contrast ... But with dark red, the contrast was still better I think.
We could try with red and a different bright color too?

I would not go from dark to brighter red, this is very unusual for the community. I think that the original red->dark was the most appropriate.

sandrinebedard:
> > I think that going with brighter or darker red doesn't solve the problem since the contrast will always be bad for one image contrast ... But with dark red, the contrast was still better I think.
> > We could try with red and a different bright color too?
> 
> I would not go from dark to brighter red, this is very unusual for the community. I think that the original red->dark was the most appropriate.

Oh yes, I absolutely agree, I was refering to that too (it gets confusing depending on the way we see it : 1-->0 (red-->dark) or 0-->1 (dark --> red)) 🙂

sandrinebedard:
I put back the original colormap and the threshold of 0.5 is applided!
To take into account softseg, I had to remove the binarization of the segmentation and ony apply the threshold, do you think that there could be a problem? (because of the linear interpolation that the segmentation was binarized)

jcohenadad:
> do you think that there could be a problem? 

hum... could you please run deepseg_sc with QC and check if the output looks the same as it was before? If not, we might need to do something about it (eg: binarize "on-demand")

sandrinebedard:
> could you please run deepseg_sc with QC and check if the output looks the same as it was before? 

Looks exactly the same as before! (I generated QC for 15 sc binary segmentations and they all looked the same as before)

image

jcohenadad (approved):
thank you for your contribution @sandrinebedard!

joshuacwnewton:
@sandrinebedard Just to double-check the title, is there anything still "WIP" for this PR?

sandrinebedard:
> @sandrinebedard Just to double-check the title, is there anything still "WIP" for this PR?

No! (just changed the name 😊)

joshuacwnewton:
Thank you again for the contribution! :tada:

Issue #1784, opened 2018-05-27
ImportError: libSM.so.6: cannot open shared object file: No such file or directory


jcohenadad:
On joplin (Ubuntu) after fresh git clone of master/fc72f34d4251de09d3e92071c66b40ac7fe4c92a and while running `batch_processing`:
~~~
Label discs...

Generate output files...
  File created: ./t2_seg_labeled.nii.gz
  File created: ./t2_seg_labeled_discs.nii.gz
  File created: ./warp_curve2straight.nii.gz
  File created: ./warp_straight2curve.nii.gz
  File created: ./straight_ref.nii.gz

Remove temporary files...
rm -rf /tmp/sct-180527181207-label_vertebrae-95hPlw
Traceback (most recent call last):
  File "/home/jcohen/code/sct-release/scripts/sct_label_vertebrae.py", line 953, in <module>
    main()
  File "/home/jcohen/code/sct-release/scripts/sct_label_vertebrae.py", line 361, in main
    generate_qc(fname_in, labeled_seg_file, args, path_qc)
  File "/home/jcohen/code/sct-release/scripts/sct_label_vertebrae.py", line 370, in generate_qc
    import spinalcordtoolbox.reports.qc as qc
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/spinalcordtoolbox/reports/qc.py", line 19, in <module>
    import matplotlib.pyplot as plt
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/matplotlib/pyplot.py", line 114, in <module>
    _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/matplotlib/backends/__init__.py", line 32, in pylab_setup
    globals(),locals(),[backend_name],0)
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/matplotlib/backends/backend_qt4agg.py", line 18, in <module>
    from .backend_qt5agg import FigureCanvasQTAggBase as _FigureCanvasQTAggBase
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/matplotlib/backends/backend_qt5agg.py", line 15, in <module>
    from .backend_qt5 import QtCore
  File "/home/jcohen/code/sct-release/python/lib/python2.7/site-packages/matplotlib/backends/backend_qt5.py", line 31, in <module>
[...]

PR #3420, opened 2021-06-16
Stop trying to detect headless systems using `DISPLAY`, and no longer set `MPLBACKEND`

### LINKED ISSUE #2693: DISPLAY variable is set....................[FAIL]
I installed the most recent version on osx 10.15.4 (19E287).
I get an error message when running `sct_check_dependencies`.

```
Spinal Cord Toolbox (git-master-36238247360b0ccd76d9cd679eaebc4e2c754cc1)

SCT info:
- version: git-master-36238247360b0ccd76d9cd679eaebc4e2c754cc1
- path: /Users/alfoi/code/spinalcordtoolbox
OS: osx (Darwin-19.4.0-x86_64-i386-64bit)
CPU cores: Available: 8, Used by SCT: 8
RAM: Primary memory available: 16.00 gigabytes
  Using bundled python 3.6.10 |Anaconda, Inc.| (default, Mar 25 2020, 18:53:43) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] at /Users/alfoi/code/spinalcordtoolbox/python/envs/venv_sct/bin/python
Check if numpy is installed.........................[OK] (1.18.4)
Check if cryptography is installed..................[OK] (2.9.2)
Check if dipy is installed..........................[OK] (1.1.1)
Check if Keras (2.1.5) is installed.................[OK]
Check if matplotlib is installed....................[OK] (3.2.1)
Check if nibabel is installed.......................[OK] (3.1.0)
Check if pandas is installed........................[OK] (1.0.3)
Check if psutil is installed........................[OK] (5.7.0)
Check if pyqt5 (5.11.3) is installed................[OK]
Check if pytest is installed........................[OK] (5.4.1)
Check if pytest-cov is installed....................[OK] (2.8.1)
Check if requests is installed......................[OK] (2.23.0)
Check if scipy is installed.........................[OK] (1.4.1)
Check if scikit-image is installed..................[OK] (0.16.2)
Check if scikit-learn is installed..................[OK] (0.22.2.post1)
Check if tensorflow (1.5.0) is installed............[OK]
Check if xlrd is installed..........................[OK] (1.2.0)
Check if xlutils is installed.......................[OK]
Check if tqdm is installed..........................[OK] (4.46.0)
Check if urllib3[secure] is installed...............[OK] (1.25.9)
Check if DISPLAY variable is set....................[FAIL]
```

### LINKED ISSUE #3388: Using `matplotlib.pyplot.show()` causes MacOS 10.15 runners to hang
If a pytest test calls a CLI script that uses `pyplot.show()`, the macOS runner will hang indefinitely. I experienced this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3387#issuecomment-839915047 and one other test (see linked comment). To confirm the cause, I temporarily commented out `pyplot.show()` in 3b5aefe and the runner no longer stalled out.

When doing a project-wide search for `plt.show()`, I notice there are 9 instances where the command is commented out, however there are still 5 other instances, hence this issue.

There's an existing [internal Wiki entry](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Moving-away-from-pyplot) about `pyplot` which recommends to refactor to matplotlib's OO interface (I didn't write this -- renaming a page changes the author on GitHub Wikis). My question is, is a complete refactor necessary? If it's just `plt.show()` that's the problem, could we get away with leaving the existing `pyplot` code and simply removing `plt.show()` (assuming we're also saving the plot as an image)?

### LINKED ISSUE #3392: `matplotlib` defaults to non-interactive backend, causing `UserWarning` for `plt.show()`
The easiest way to demonstrate is using `sct_dmri_display_bvecs`:

```
--
Spinal Cord Toolbox (git-master-a62fb672c602b9863a8c03521eb272bfd7bad9e1)

sct_dmri_display_bvecs -bvec sct_example_data/dmri/bvecs.txt
--

Saving figure: bvecs.png

/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py:137: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  plt.show()
```

This warning occurs and no figure is shown.

We explicitly set the backend type using the `MPLBACKEND` environment variable. (Option 2. in [this list](https://matplotlib.org/2.0.2/faq/usage_faq.html#what-is-a-backend).)

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a62fb672c602b9863a8c03521eb272bfd7bad9e1/install_sct#L266-L272

We started using the `Agg` backend in #2157 (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2238/commits/cd521dbdce9fb74332528a6a0d3b253012dfd492) to avoid crashes on headless systems (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2138). 

Moving forward, I can think of two options, depending on whether we need interactive `matplotlib` windows in SCT.

* If we want interactive windows, we should figure out how to set `MPLBACKEND=Agg` (a non-interactive backend) for headless systems only. 
* If we don't care about interactive windows, and instead only want to generate images to disk, then we should remove any instances of `plt.show()`.

joshuacwnewton:
#### GitHub

#### PR contents

### Context

Historically, to avoid crashes on headless systems (#1784), we've set `MPLBACKEND` to a non-GUI backend in the user's RC file. We've also tried to detect headless systems using `DISPLAY`, then if we are headless, set `MPLBACKEND` in the CLI launcher. 

However, #1784 appears to have been fixed upstream (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3401#issuecomment-861614835), so in theory, these heavy-handed `MPLBACKEND` workarounds are no longer needed.

### PR Changes

* Avoid setting `MPLBACKEND` at all within SCT, and let matplotlib choose an appropriate backend itself (Fixes #3392)
* Set `MPLBACKEND=Agg` for the test suite only, since showing interactive plots would interrupt the tests (Fixes #3388)

Additionally, since checking for headless systems using `$DISPLAY` is no longer relevant, this PR refactors the `sct_check_dependencies` GUI checks to be more useful. (Fixes #2693)

## Linked issues

Fixes #2693.
Fixes #3388.
Fixes #3392.

Supersedes #3401.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -26,7 +26,12 @@ logger = logging.getLogger(__name__)
 
 
 def pytest_sessionstart():
-    """ Download sct_testing_data prior to test collection. """
+    """Perform actions that must be done prior to test collection."""
+    # Use a non-interactive backend so that no GUI plots will interrupt the test suite.
+    # (NB: We do this here to ensure it is set before `matplotlib` is first imported.)
+    os.environ["MPLBACKEND"] = 'Agg'
+
+    # Download sct_testing_data prior to test collection
     logger.info("Downloading sct test data")
     downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
 
--- a/conftest.py
+++ b/conftest.py
@@ -29,7 +29,8 @@ def pytest_sessionstart():
     """Perform actions that must be done prior to test collection."""
     # Use a non-interactive backend so that no GUI plots will interrupt the test suite.
     # (NB: We do this here to ensure it is set before `matplotlib` is first imported.)
-    os.environ["MPLBACKEND"] = 'Agg'
+    if 'MPLBACKEND' not in os.environ:
+        os.environ["MPLBACKEND"] = 'Agg'
 
     # Download sct_testing_data prior to test collection
     logger.info("Downloading sct test data")
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -317,16 +317,28 @@ def main(argv=None):
     if complete_test:
         print((status, output), '\n')
 
-    print_line('Check if DISPLAY variable is set')
+    print_line('Check if figure can be opened with matplotlib')
     try:
-        os.environ['DISPLAY']
-        print_ok()
+        import matplotlib
+        import matplotlib.pyplot as plt
+        # If matplotlib is using a GUI backend, the default 'show()` function will be overridden
+        # See: https://github.com/matplotlib/matplotlib/issues/20281#issuecomment-846467732
+        fig = plt.figure()
+        if getattr(fig.canvas.manager.show, "__func__", None) != matplotlib.backend_bases.FigureManagerBase.show:
+            print_ok(f" (Using GUI backend: '{matplotlib.get_backend()}')")
+        else:
+            print_fail(f" (Using non-GUI backend '{matplotlib.get_backend()}')")
+    except Exception as err:
+        print_fail()
+        print(err)
 
-        # Further check with PyQt specifically
-        print_line('Check if figure can be opened with PyQt')
-        from PyQt5.QtWidgets import QApplication, QLabel
+    print_line('Check if figure can be opened with PyQt')
+    if sys.platform == "linux" and 'DISPLAY' not in os.environ:
+        print_fail(" ($DISPLAY not set on X11-supporting system)")
+    else:
         try:
-            app = QApplication([])
+            from PyQt5.QtWidgets import QApplication, QLabel
+            _ = QApplication([])
             label = QLabel('Hello World!')
             label.show()
             label.close()
@@ -334,9 +346,6 @@ def main(argv=None):
         except Exception as err:
             print_fail()
             print(err)
-
-    except KeyError:
-        print_fail()
 
     print('')
     sys.exit(e + install_software)
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -41,7 +41,6 @@ jobs:
         run: |
           PIP_PROGRESS_BAR=off ./install_sct -y
           cat ~/.bashrc | grep "export SCT_DIR" | cut -d " " -f 2 >> $GITHUB_ENV
-          cat ~/.bashrc | grep "export MPLBACKEND" | cut -d " " -f 2 >> $GITHUB_ENV
           cat ~/.bashrc | grep "export PATH" | grep -o "/.*" | cut -d ':' -f 1 >> $GITHUB_PATH
 
       - name: Run batch_processing.sh
--- a/install_sct
+++ b/install_sct
@@ -265,10 +265,8 @@ function edit_shellrc() {
     # Switch between shell
     if [[ "$THE_RC" == "bash" ]]; then
       echo "export SCT_DIR=$SCT_DIR"
-      echo "export MPLBACKEND=Agg"
     elif [[ "$THE_RC" == "csh" ]]; then
       echo "setenv SCT_DIR $SCT_DIR"
-      echo "setenv MPLBACKEND Agg"
     fi
     # add line
     echo ""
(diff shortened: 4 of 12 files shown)


conftest.py
```diff
@@ -26,7 +26,12 @@
 
 
 def pytest_sessionstart():
-    """ Download sct_testing_data prior to test collection. """
+    """Perform actions that must be done prior to test collection."""
+    # Use a non-interactive backend so that no GUI plots will interrupt the test suite.
+    # (NB: We do this here to ensure it is set before `matplotlib` is first imported.)
+    os.environ["MPLBACKEND"] = 'Agg'
```
joshuacwnewton, line 32: Note: This change is slightly different than what @kousu suggested in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3388#issuecomment-861620956.

If we set `MPLBACKEND` in `tests.yml`, then interactive plots will still interrupt the test suite when `pytest` is run locally.

kousu: Ah, true. :+1: 

Can we do both? If we later handle https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3392 / https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3388 by making sure the non-interactive tests never have to call any non-interactive code we can get rid of this smoothly.

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -317,16 +317,28 @@ def main(argv=None):
...
-        print_line('Check if figure can be opened with PyQt')
-        from PyQt5.QtWidgets import QApplication, QLabel
+    print_line('Check if figure can be opened with PyQt')
+    if sys.platform == "linux" and 'DISPLAY' not in os.environ:
+        print_fail(" ($DISPLAY not set on X11-supporting system)")
+    else:
         try:
-            app = QApplication([])
+            from PyQt5.QtWidgets import QApplication, QLabel
+            _ = QApplication([])
```
joshuacwnewton, line 341: We can see the results of these checks here:

* [Ubuntu](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/2840568568#step:3:2774): Here we can see that matplotlib defaults to a non-GUI backend because `DISPLAY` is not set.
* [MacOS 10.15](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/2840568255#step:3:2021): Here we can see that matplotlib defaults to a GUI backend (#3388), but this is OK for our CI test suite, because we set MPLBACKEND in `pytest`.

conftest.py
joshuacwnewton: Sorry, I'm still a bit unclear still on the benefits of doing both. 

In the future, if we remove interactive plotting code, wouldn't we not need `MPLBACKEND=Agg` in either place?

kousu: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3388#issuecomment-846091012

> to test [..] I suggest unconditionally setting the backend to Agg (that is what we do in the Matplotlib tests). It avoids a whole class of pain points (related to making sure you have the event loop running enough, etc) and will make your tests run faster!

joshuacwnewton: Right!

In that case, what's the benefit of setting it in the CI config instead of setting it within `pytest`?

kousu: Well, it'll keep CI faster, for one.

I guess I was thinking that we know the CI machines are headless, but the matplotlib doesn't, and this is how we communicate that fact to it globally.

Communicating the same via pytest feels like more of a workaround. Maybe you want to run the tests locally in your GUI just to see what they do sometimes?

kousu (approved):
Otherwise, looks good. Merge if you want :)

conftest.py
joshuacwnewton: > Maybe you want to run the tests locally in your GUI just to see what they do sometimes?

Ah, good point. And if we're setting it in pytest, there's not a great way to unset it beyond going it and modifying that LOC, which is a little problematic.

At the very least, to address this we could do this instead:

```suggestion
    # Use a non-interactive backend so that no GUI plots will interrupt the test suite.
    # (NB: We do this here to ensure it is set before `matplotlib` is first imported.)
    if 'MPLBACKEND' not in os.environ:
        os.environ["MPLBACKEND"] = 'Agg'
```

That way, a dev can specify e.g. `export MPLBACKEND=TkAgg` locally to effectively turn GUI testing back on.

PR #3429, opened 2021-06-21
(WIP) Measure CSA based on distance from pontomedullary junction (PMJ)

### LINKED ISSUE #3063: Measure CSA based on distance from spinal cord landmark such as the pontomedullary junction (PMJ)
## Context

Currently, CSA is measured and aggregated using vertebral levels. This is inaccurate, as people's positioning on the scanner (eg head tilting) impacts the location of the spinal level wrt. vertebral level.

We use the vertebral levels because they are easy to "see" and label, and also because this is what people have been doing so far. 

But some groups (eg UCSF, Queens U.) are now using the distance between a spinal (vs. vertebral) landmark in order to report CSA value.

## Proposed solution

Implement a way to compute CSA as a function of the distance to the PMJ. 

Some details:
- The PMJ can be detected with `sct_detect_pmj`. We should check the precision of the detection is sufficient.
- The computed distance should be along the spinal cord centerline (ie: not a "straight line" distance).

The input would be:
- the distance to the PMJ
- the extent of the mask in mm in the S-I direction

The output would be a binary mask with value 1. 

Illustration of the proposed approach (source: [Google Drawing](https://docs.google.com/drawings/d/1X3IIxPnRtuf2wE9ZVqqNUUvFyKhle-KRSAiWMLwJ4wE/edit)):

CSA_length_PMJ

sandrinebedard:
#### GitHub

#### PR contents

## Description

This PR addresses #3063. The goal is to measure CSA based on a distance from pontomedullary junction (PMJ). To take into account the curvature of the spinal cord, the distance is measured from the projected PMJ on the continuation of the centerline as illustrated [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3063#issuecomment-862648093).

I have integrated the feature in `sct_process_segmentation` minimally to use existing useful functions. I wasn't sure exaclty how we wanted to intergrate it/ what features we wanted, so this is still work in progress!

#### Main changes include:
* New input arguments in `sct_process_segmentation` :
    * `-pmj` --> PMJ label
    * `-distance` --> distance from PMJ
    * `-extent` --> S-I size of the mask to compute CSA 
*  New script `csa_pmj.py` (could be good to rename the script since it doesn't really compute CSA)
    (This script inlcude functions to get the distance from projected PMJ and get slices corresponding to distance from PMJ)
    * Extrapolate centerline
    * Get intersection of continuation of the centerline and plane that passes through PMJ.
    * Computes distance from projected PMJ for each slices.
    * Gets slices based on extent and desired distance from PMJ.
    * Returns slices and mask.
* Add option `distance_pmj` in `aggregate_slicewise.py` to specify the distance and add a column for distance from PMJ in `csa.csv` file.

Outputs:
* Extrapolated centerline
* Binary mask of extent
* In `csa.csv`
    * Distance from PMJ
    * csa value (and the other measures that sct_process_segmentation outputs) 

#### Whats left:

- [ ] Update documentation

## Linked issues

Resolves #3063


spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -527,7 +541,7 @@ def save_as_csv(agg_metric, fname_out, fname_in=None, append=False):
     if not append or not os.path.isfile(fname_out):
         with open(fname_out, 'w') as csvfile:
             # spamwriter = csv.writer(csvfile, delimiter=',')
-            header = ['Timestamp', 'SCT Version', 'Filename', 'Slice (I->S)', 'VertLevel']
+            header = ['Timestamp', 'SCT Version', 'Filename', 'Slice (I->S)', 'VertLevel', 'DistancePMJ']
```
joshuacwnewton, line 544: Just a quick note: It looks like a couple of tests are [failing](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/2876872450#step:5:322) due to the addition of a new column in the csv output of `aggregate_slicewise.py`. 

Could you update these tests to account for the new column?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -320,17 +352,30 @@ def main(argv=None):
...
                                          verbose=verbose)
+    if fname_pmj is not None:
+        mask, pmj_slices = get_slices_for_pmj_distance(fname_segmentation, fname_pmj,
+                                                       distance_from_pmj, extent_mask,
+                                                       param_centerline=param_centerline,
+                                                       verbose=verbose)
+
+        fname_mask_out = add_suffix(arguments.i, '_mask_csa')
+        mask.save(fname_mask_out)
+        pmj_slices = [parse_num_list(pmj_slices[0]), pmj_slices[-1]]
```
jcohenadad, line 363: Do you think there is a possibility to instead update the variable `slices`, so that the API `aggregate_slicewise.py` would not need to be updated at all?

sandrinebedard: I though about this, it was my first idea, but we cannot also output the distance from PMJ that the user specified this way, I think that is necessary to have in the .csv file.
Also, if we want to add the possibility to input multiple distances (so various range of slices) it wouldn't work with `slices ` (from my understanding) (if this is a feature we want of course)

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: of course, I am on it 🙂

spinalcordtoolbox/scripts/sct_process_segmentation.py
jcohenadad: > I thought about this, it was my first idea, but we cannot also output the distance from PMJ that the user specified this way, I think that is necessary to have in the .csv file.

If the goal is to just populate the CSV file then this single argument could be passed to the API

> Also, if we want to add the possibility to input multiple distances (so various range of slices) it wouldn't work with slices (from my understanding) if this is a feature we want of course

I don't think we will be needing this feature.

sandrinebedard: > If the goal is to just populate the CSV file then this single argument could be passed to the API

okay! done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/0696bb0abe42ee8b9a9bf72ecb87a467092ccd7e 😊

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -232,15 +232,16 @@ def func_wa(data, mask=None, map_clusters=None):
...
+def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], distance_pmj=None, perslice=None,
+                                 perlevel=False, vert_level=None, group_funcs=(('MEAN', func_wa),), map_clusters=None):
     """
     The aggregation will be performed along the last dimension of 'metric' ndarray.
 
     :param metric: Class Metric(): data to aggregate.
     :param mask: Class Metric(): mask to use for aggregating the data. Optional.
     :param slices: List[int]: Slices to aggregate metric from. If empty, select all slices.
     :param levels: List[int]: Vertebral levels to aggregate metric from. It has priority over "slices".
+    :param distance_pmj: float: Distance from Ponto-Medullary Junction (PMJ).
```
jcohenadad, line 244:
```suggestion
    :param distance_pmj: float: Distance from Ponto-Medullary Junction (PMJ) in mm.
```

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Extract min and max index in Z direction
+    data_seg = im_segr.data
+    X, Y, Z = (data_seg > NEAR_ZERO_THRESHOLD).nonzero()
+    min_z_index, max_z_index = min(Z), max(Z)
+
+    # Compute the spinal cord centerline based on the spinal cord segmentation
+    _, arr_ctl, arr_ctl_der, fit_results = get_centerline(im_segr, param=param_centerline, verbose=verbose)
+
+    # Get tagent vector at the top of the centerline | Maybe take a fiew slices to get the vector ?
```
jcohenadad, line 49:
```suggestion
    # Get tangent vector at the top of the centerline | Maybe take a few slices to get the vector ?
```

jcohenadad (commented):
As a reminder: we will also need to update the documentation (which @joshuacwnewton is currently working on)

jcohenadad:
> If input -distance in not specfied, compute CSA perslice and sepecify distance from PMJ for each slice.

That's a cool idea but I would leave this for a future issue so this PR doesn't get too big.

jcohenadad (commented):
it would be great if you could also add a few unit tests on your code, just to validate that the distance to PMJ is properly calculated (eg: you can "manually" measure it on the testing dataset).

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,179 @@
...
+    :param segmentation: input segmentation. Could be either an Image or a file name.
+    :param pmj: label of PMJ.
+    :param distance: float: Distance from PMJ.
+    :param param_centerline: see centerline.core.ParamCenterline()
+    :param verbose:
+    :return mask:
+    :return slices:
+
+    """
+    # TODO: loop through distance (when multiple distances are given)
```
jcohenadad, line 30: I think we can remove this for now?

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,179 @@
...
+    plane_pt = arr_ctl[:, -1] + [0, 0, min_z_index]
+    line_pt = np.argwhere(data_pmj != 0)[0]  # list(im_pmjr.getNonZeroCoordinates()[0])[:-1]
+
+    # Get intersection of plane passing through PMJ and continuation of centerline
+    init_point = get_intersection_plane_line(plane_pt, tangent_vect, line_pt, tangent_vect)
+    # Get the distance from PMJ for all slices of the centerline
+    arr_length = get_distance_from_pmj(arr_ctl, init_point, px, py, pz)
+
+    # Check if distance is out of bound
+    if distance > arr_length[0][0]:
```
jcohenadad, line 63: you might also want to check if `distance < arr_length[0][-1]`

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -320,17 +352,27 @@ def main(argv=None):
...
                                          param_centerline=param_centerline,
                                          verbose=verbose)
+    if fname_pmj is not None:
+        mask, slices = get_slices_for_pmj_distance(fname_segmentation, fname_pmj,
+                                                   distance_pmj, extent_mask,
+                                                   param_centerline=param_centerline,
+                                                   verbose=verbose)
+
+        fname_mask_out = add_suffix(arguments.i, '_mask_csa')
+        mask.save(fname_mask_out)
```
jcohenadad, line 362: hum, the mask has a different orientation than the input segmentation:

**sct_image -i t2_seg.nii.gz -header**
~~~
qto_xyz:1	0.799726 0.000000 -0.020942 -19.452049 
qto_xyz:2	0.000000 0.800000 0.000000 -115.151794 
qto_xyz:3	0.020942 -0.000000 0.799726 -122.996811 
qform_xorient	Left-to-Right
~~~

sct_image -i t2_seg_mask_csa.nii.gz -header
~~~
qto_xyz:1	-0.799726 0.000000 -0.020939 30.930708 
qto_xyz:2	-0.000000 0.800000 0.000000 -115.151794 
qto_xyz:3	-0.020939 -0.000000 0.799726 -121.677490 
qform_xorient	Right-to-Left
~~~

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,179 @@
...
+    data_seg = im_segr.data
+    X, Y, Z = (data_seg > NEAR_ZERO_THRESHOLD).nonzero()
+    min_z_index, max_z_index = min(Z), max(Z)
+
+    # Compute the spinal cord centerline based on the spinal cord segmentation
+    _, arr_ctl, arr_ctl_der, fit_results = get_centerline(im_segr, param=param_centerline, verbose=verbose)
+
+    # Get tagent vector at the top of the centerline | Maybe take a fiew slices to get the vector ?
+    iz = max_z_index
+    tangent_vect = np.array([arr_ctl_der[0][iz - min_z_index] * px, arr_ctl_der[1][iz - min_z_index] * py, pz])
```
jcohenadad, line 51: You got the intuition right that "take a few slices to get the vector" would be more robust to issues in the latest segmentation, but even so, I think that taking the tangent is not a robust approach as it is way to sensitive to the top segmented slices, which are usually only partially segmented. To demonstrate the sensitivity, I slightly modified the top two slices of the segmentation produced by `sct_deepseg_sc` on the T2w scan from the **sct_example_data**. The original segmentation is in white, and the modified one in red:
image

When running the following command on the original segmentation:
~~~
sct_process_segmentation -i t2_seg.nii.gz -pmj t2_pmj.nii.gz -distance 55 -extent 20 -o csa_pmj.csv
~~~

I get this:


Timestamp | SCT Version | Filename | Slice (I->S) | VertLevel | DistancePMJ | MEAN(area) | STD(area) | MEAN(angle_AP) | STD(angle_AP) | MEAN(angle_RL) | STD(angle_RL) | MEAN(diameter_AP) | STD(diameter_AP) | MEAN(diameter_RL) | STD(diameter_RL) | MEAN(eccentricity) | STD(eccentricity) | MEAN(orientation) | STD(orientation) | MEAN(solidity) | STD(solidity) | SUM(length)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
2021-06-21 13:52:13 | git-sb/3063-csa-measure-from-pmj-122c615148a38e8550a0ae9e1d4ed1375857015b | /Users/julien/sct_example_data/t2/t2_seg.nii.gz | 176:200 |   | 55.0 | 73.09505206059430 | 2.058273644130360 | 0.6042205187778950 | 0.509096773612399 | 5.99722237889052 | 1.3648000670811700 | 8.044099345457310 | 0.2133923010218110 | 11.551670079850800 | 0.11679222483932000 | 0.7170797849523140 | 0.019380548885043800 | 1.1437363154081200 | 1.0298888376877800 | 0.9723314116348530 | 0.009082028290704460 | 20.117826899911600

While when running it on the modified segmentation, I get a very different selection of slices (shifted by 10 slices, which represents 8.0 mm), even though the distance to the PMJ is the same:



Timestamp | SCT Version | Filename | Slice (I->S) | VertLevel | DistancePMJ | MEAN(area) | STD(area) | MEAN(angle_AP) | STD(angle_AP) | MEAN(angle_RL) | STD(angle_RL) | MEAN(diameter_AP) | STD(diameter_AP) | MEAN(diameter_RL) | STD(diameter_RL) | MEAN(eccentricity) | STD(eccentricity) | MEAN(orientation) | STD(orientation) | MEAN(solidity) | STD(solidity) | SUM(length)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
2021-06-21 18:06:28 | git-sb/3063-csa-measure-from-pmj-0696bb0abe42ee8b9a9bf72ecb87a467092ccd7e | /Users/julien/sct_example_data/t2/t2_seg_modif.nii.gz | 186:210 |   | 55.0 | 74.49899710166940 | 2.0654612628143000 | -0.4655824057061610 | 0.6187649443692570 | 4.602201395154190 | 0.7021559581881710 | 8.21724057257737 | 0.19672289786466700 | 11.508958552747000 | 0.09988903157047450 | 0.699731378857051 | 0.01719526973665410 | 1.219835288456290 | 0.9493392141345070 | 0.9747126663782220 | 0.009245100001059110 | 20.06805395574310


For convenience, here is the original segmentation: 
[t2_seg.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/6689746/t2_seg.nii.gz)

And the modified segmentation: 
[t2_seg_modif.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/6689748/t2_seg_modif.nii.gz)

My suggestion would be to leverage the algorithms behind the [`get_centerline()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/centerline/core.py#L86) method, which includes smoothing and extrapolation.

spinalcordtoolbox/csa_pmj.py
sandrinebedard: yes!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: ahh yes, I can maybe put back the mask to native orientation before saving it?

spinalcordtoolbox/csa_pmj.py
sandrinebedard: thank you for testing this, I'll look into extrapolation of the centerline 🙂

spinalcordtoolbox/csa_pmj.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/6bccae41cf8296490437e0adf5f2e03a7a084cc6

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/2155bf2def2d64f893165839c76efc3de4b6cde6

spinalcordtoolbox/csa_pmj.py
sandrinebedard: I've implemented the extrapolation of the centerline in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/df7cbc986444e7e90feb14868c1b6ef808f244c6

I tested out setting the parameter `param.minmax = None` in `get_centerline()` to directly extrapolate the centerline while generating it. However, with the modified segmentation you did, the result of the extrapolated centerline is really bad. So, what I've comme up with is that I extrapolate on the centerline that is generated only on the slices available in the segmentation minus the top slices of the centerline to exclude those biased values of the centerline.
I just wonder if this would really be generalizable (how many slices, etc.)? It is usually the case that top slices of the segmentation aren't good, so the top of centerline is also biased. Also, even if we don't include the top slices for extrapolation, we still have the general direction of the centerline and that may be enough here
After this, I was a bit out of ideas 😬, I tested out other methods, but the modified segmentation always had a big impact on extrapolation...

PS: The extrapolation was good on the original segmentation

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def get_min_distance(pmj, centerline, px, py, pz):
+    distance = np.sqrt(((centerline[0, :] - pmj[0]) * px) ** 2 +
+                       ((centerline[1, :] - pmj[1]) * py) ** 2 +
+                       ((centerline[2, :] - pmj[2]) * pz) ** 2)
+    return int(centerline[2, distance.argmin()])
+
+
+def validate_length(z_min, z_max, arr, value):  # TODO: Find a better way for this
```
jcohenadad, line 173: I suggest to add a warning in case slices are removed (so users are aware)

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,206 @@
...
+        raise ValueError("Input distance of " + str(distance) + " mm is out of bound for minimum distance of " + str(arr_length[0][-1]) + " mm")
+
+    # Get Z index of corresponding distance from PMJ with the specified extent
+    z_index_extent_min = get_nearest_index(arr_length, distance + extent/2)
+    z_index_extent_max = get_nearest_index(arr_length, distance - extent/2)
+    # Check if extent corresponds to the lenght, if not add or remove a slice
+    z_index_extent_min, z_index_extent_max = validate_length(z_index_extent_min, z_index_extent_max, arr_length, extent)  # Find a quicker way to solve this
+    # Check if min Z index is available in the segmentation, if not, use the min_z_index of segmentation
+    if z_index_extent_min < min_z_index:
+        z_index_extent_min = min_z_index
```
jcohenadad, line 80: you might want to add a warning to let the user know that the requested I-S length will be considered because...

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,206 @@
...
+    arr_length = get_distance_from_pmj(arr_ctl_extra, pmj_index, px, py, pz)
+
+    # Check if distance is out of bound
+    if distance > arr_length[0][0]:
+        raise ValueError("Input distance of " + str(distance) + " mm is out of bound for maximum distance of " + str(arr_length[0][0]) + " mm")
+
+    if distance < arr_length[0][-1]:
+        raise ValueError("Input distance of " + str(distance) + " mm is out of bound for minimum distance of " + str(arr_length[0][-1]) + " mm")
+
+    # Get Z index of corresponding distance from PMJ with the specified extent
```
jcohenadad, line 73: I ran the code in the debugger and found that the requested `extent=20` yielded the slices '183:207', which gives an extent of 24 (instead of 20). I haven't dug to much into the code but maybe there is an indexing issue somewhere?

sandrinebedard: ahh, right now the extent is in mm (along the centerline) not in number of slices, what would be better?

jcohenadad: ahh i see, my mistake. We should definitely keep the extent in mm to be consistent with the distance from PMJ (and also to be agnostic to image resolution)

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: I updated the tests for `test_aggregate_slicewise.py`  in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/3e63ff34b4abf2604dfbafdc48a4f8363ca2001c
and `test_batch_processing.py` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/3e63ff34b4abf2604dfbafdc48a4f8363ca2001c
The tests still doesn't pass because `batch_processing/catched_results/` needs to be updated so the csv files have the right new format I think (can you confirm @joshuacwnewton ?)

spinalcordtoolbox/csa_pmj.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/387ee6962b903e1d3495b0fbae1c689ae800884a

sandrinebedard:
## Update on the PR
I think I've addressed almost every comment (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429#discussion_r658907970 is left), here is an update of what is done and what is left
### What is done
* Extrapolate centerline (in `get_centerline()`, with `ParamCenterline.minmax = False` )(https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/65a79684d2165234ab05063035b302dfee772f7b)
* Added unit tests to check if the distance is properly calculated (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/db93a025e2b07f09c25f754ff160344b45c48541)
* Save the centerline (with native orientation), related to [comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429#commitcomment-52669024), (done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429/commits/9be1b5b10d3a9d018876d131bb30b23abaf78054)

### What is left

* Should I add a warning if the specified distance is smaller than half of the extent ? (Example: distance = 7 mm and extent = 20 mm --> 7>10) (same thing for the other extremity)
    * This raises the question of how we want to address the issues at the extremities of the segmentation (if asked distance with extent includes empty slices, ...)
* Related to [#3429 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429#discussion_r658907970), I wanted to discuss the use of this function. When getting the slice index corresponding to a distance (with the extent), sometimes, with the nearest distance, we get a 19 mm length, or 21 of length, that was why I added this function to check if the length was 20 (add or remove a slice to get 20 mm of length). However, when doing  that, we are not centered on the specified distance:
    * Do we prefer to have the distance centered in the mask or to have exaclty the asked extent?
    * There may be a better way to get direclty the right length?

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,206 @@
...
+        centerline = centerline[:, :-5]  # Remove last slices (maybe always do that?)
+    x_mean = centerline[0, :]
+    y_mean = centerline[1, :]
+    z_mean = centerline[2, :]
+    z_ref = np.array(range(im_seg.dim[2]))
+    x_centerline_fit, x_centerline_deriv = curve_fitting.bspline(z_mean, x_mean, z_ref, smooth, pz=pz, deg_bspline=3)
+    y_centerline_fit, y_centerline_deriv = curve_fitting.bspline(z_mean, y_mean, z_ref, smooth, pz=pz, deg_bspline=3)
+    return np.array([x_centerline_fit, y_centerline_fit, z_ref])
+
+
```
jcohenadad, line 153: doesn't seem to be used anymore
```suggestion
```

jcohenadad: ah! i just noticed the "# TO REMOVE" comment 😉

jcohenadad:
> Related to #3429 (comment), I wanted to discuss the use of this function. When getting the slice index corresponding to a distance (with the extent), sometimes, with the nearest distance, we get a 19 mm length, or 21 of length, that was why I added this function to check if the length was 20 (add or remove a slice to get 20 mm of length). However, when doing that, we are not centered on the specified distance:
> - Do we prefer to have the distance centered in the mask or to have exaclty the asked extent?
 > - There may be a better way to get direclty the right length?

Yes, there might be a better way to get directly the right length. I am investigating (in a parallel branch https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb%2F3063-csa-measure-from-pmj) the use of the [`spinalcordtoolbox.types.Centerline`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/types.py#L173) class, which has a lot of useful methods, including length along the centerline and index of the nearest point to the centerline (in our case: the PMJ). This class is pretty old (~4y) and lacks some documentation, so I prefer to be the one doing the digging because I remember the context of the class, which helps me understand quickly what each method does.

spinalcordtoolbox/csa_pmj.py
```diff
@@ -0,0 +1,206 @@
...
+    # Update dimensions from resampled image
+    nx, ny, nz, nt, px, py, pz, pt = im_segr.dim
+
+    # Extract min and max index in Z direction
+    data_seg = im_segr.data
+    X, Y, Z = (data_seg > NEAR_ZERO_THRESHOLD).nonzero()
+    min_z_index, max_z_index = min(Z), max(Z)
+
+    # Remove top slices  | TODO: check if center of mass of top slices is close to other slices, if not, remove
+    im_segr.data[:, :, max_z_index - 4:max_z_index + 1] = 0
```
jcohenadad, line 49: This code deals with cases where the top slices of the segmentations are only partial, which could yield wrong estimate of the extrapolation. Example below:

with this code:
fig_centerline_210628-125903372620_bspline


without this code (notice the ^106 for the Y component):
fig_centerline_210628-150754841347_bspline

However, I am not a big fan of this code because we could imagine cases where more than the top four slices show a not well defined segmentation. In general, extrapolation seems poorly robust so I would be in favor of avoiding it.

One possibility I'm thinking about is including the PMJ in the segmentation and then interpolating between the segmentation and the PMJ. That way we would avoid robustness issues. I'll investigate a bit and document in this thread...

sandrinebedard: Great thanks for investigating!

> One possibility I'm thinking about is including the PMJ in the segmentation and then interpolating between the segmentation and the PMJ. 

The one thing maybe to look out for with this approach is to validate that the curvature of the spinal cord will be taken into account and not biased with the PMJ coordinate (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3063#issuecomment-862648093)

jcohenadad: > The one thing maybe to look out for with this approach is to validate that the curvature of the spinal cord will be taken into account and not biased with the PMJ coordinate ([#3063 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3063#issuecomment-862648093))

Yes, this is a very valid point. I am wondering what is the most valid approach: using the projection on an extrapolated centerline (where the extrapolation can be problematic) or the length to the PMJ (even though it is slightly displaced anteriorly compared to the centerline). My intuition says that the latter is better. This is something we could/should validate by acquiring high resolution T2w scans where we can see the nerve rootlets.

sandrinebedard: > This is something we could/should validate by acquiring high resolution T2w scans where we can see the nerve rootlets.
yes that will definitely be interesting!

jcohenadad:
Following up on https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429#discussion_r660050894, here are results from 753b0477f1ed76383401c8b37bebf6244dcb3f89 (WIP branch) comparing the three interpolation methods (the most right point is the PMJ):

**linear 🎉 **
fig_centerline_210628-182055296954_linear

**bspline 👎  (not robust-- see Y plot)**
fig_centerline_210628-182042410095_bspline

**nurbs 👎  (not regularized enough-- slower to computed-- discontinuities which will affect length calculation)**
fig_centerline_210628-182216416142_nurbs

> ℹ️ ‎ In this case we can use `param_centerline.minmax = True`

I'll continue implementation using linear interpolation (with smoothing of 50mm).

jcohenadad:
Preliminary results using the `Centerline` class and computation of distance from the PMJ along the centerline. One advantage of using the `Centerline` class is that the code is much reduced (see: ).

There is a slight difference between the two implementations: [see difference](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/6fa315e..05976283f)

Results are similar thought, in this example image, the mask is only shifted by two slices:

SCT Version | Filename | Slice (I->S) | VertLevel | DistancePMJ | MEAN(area)
-- | -- | -- | -- | -- | --
git-sb/3063-csa-measure-from-pmj-6fa315e20c69ddbe22a12743f5f7d48b3a07accd | /Users/julien/sct_example_data/t2/t2_seg.nii.gz | 182:206 |   | 55.0 | 73.91137423841470
git-jca/sb/3063-csa-measure-from-pmj-e943df8045edc19292194a9bd29686ee9269067c* | /Users/julien/sct_example_data/t2/t2_seg.nii.gz | 184:208 |   | 55.0 | 74.13788779984930

Also see the generated mask (green: Sandrine | red: Julien | yellow: overlap):
Screen Shot 2021-06-28 at 9 32 16 PM

@sandrinebedard A few comments:
- I went with raising an exception if the range of slices is not fully covered by the segmentation. I think it is better than raising a warning, because most people won't see this warning (whereas if there is an error, they will see it);
- I was not sure why you resampled the data to be isotropic resolution in the axial plane?
- Having access to the two implementations will be useful for the paper. We can validate the two methods further in more datasets (where we have ground truth from the nerve rootlets) and then decide which one to recommend.

sandrinebedard:
> Preliminary results using the Centerline class and computation of distance from the PMJ along the centerline. One advantage of using the Centerline class is that the code is much reduced (see: ).

great, I can modify my code to use it too, but did we want to keep 2 branches? how do we want to keep both methods?

> @sandrinebedard A few comments:
> 
> * I went with raising an exception if the range of slices is not fully covered by the segmentation. I think it is better than raising a warning, because most people won't see this warning (whereas if there is an error, they will see it);

yes that's better!

> * I was not sure why you resampled the data to be isotropic resolution in the axial plane?

I think I just took that part of code to get centerline in [`process_seg.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/634670e4e0e282d27abaf457def5feb49da9a1f6/spinalcordtoolbox/process_seg.py#L47-L70), but don't think it is necessary just for the centerline.

> * Having access to the two implementations will be useful for the paper. We can validate the two methods further in more datasets (where we have ground truth from the nerve rootlets) and then decide which one to recommend.
> * Validation should also be done in a variety of datasets, eg: with problematic segmentation at the edge, with highly anisotropic resolution scans, with very short segmentation, etc.

Great!

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Ah, good call! I can rewrite this test so that it's no longer dependent on specific column indexes. I'll open a PR, so feel free to ignore it for now. :smile:

sandrinebedard: great thank you!

joshuacwnewton: #3436 has been merged! Feel free to mark this as resolved once this PR is updated with master. :slightly_smiling_face:

sandrinebedard: will do,  thanks for the fix!

sandrinebedard:
@jcohenadad 
I looked at your method in https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb%2F3063-csa-measure-from-pmj

A few things I noticed:

1.  argmin() in these [lines ](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/05976283f121899e8903e8bb8222dce004940236/spinalcordtoolbox/csa_pmj.py#L58-L59) doesn't give the right index if the segmentation doesn't start at the index 0. Since [`    param_centerline.minmax = True`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/05976283f121899e8903e8bb8222dce004940236/spinalcordtoolbox/csa_pmj.py#L49), the centerline doesn't include slices without segmentation. I removed the first slice of a segmentation and got for `-distance 55` slices 129:148 for the croped segmentation and 128:147 for the other. We should get the exact same mask since the length is measurd from PMJ so the segmentation bellow shouln't matter.

    In **red**:  mask with original segmentation, | in **yellow**: mask with cropped segmentation
    image


    However, when I check the length value at these slices, it is the same, that shoudn't be so sice the mask is shifted of 1 slice...

    Maybe [ctl_seg_with_pmj.incremental_length](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/05976283f121899e8903e8bb8222dce004940236/spinalcordtoolbox/csa_pmj.py#L56) is the problem, it starts at the first index , we should probabily use [incremental_inverse](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f02309a7279e5bea0cf0e321ac8192a2900c7e0b/spinalcordtoolbox/types.py#L278-L291), I'll explore that!

    #### For reference here are the segmentations I used: 
    cropped segmentation: 
    [sub-amu01_T1w_seg_crop.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/6743586/sub-amu01_T1w_seg_crop.nii.gz)

    segmentation: 
    [sub-amu01_T1w_seg.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/6743589/sub-amu01_T1w_seg.nii.gz)

2. To check if  if the range of slices is not fully covered by the segmentation, in these [lines ](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/05976283f121899e8903e8bb8222dce004940236/spinalcordtoolbox/csa_pmj.py#L61-L62), if I input `distance 400`, (something longer then the max length from PMJ), zmin and zmax will both be 0, (the first slice) and sice it is include in the segmentation, no exception is raised

I can fix those things in your branch if you don't mind 🙂

jcohenadad:
great investigations!

> I can fix those things in your branch if you don't mind 🙂

yes please, go ahead!

sandrinebedard:
Here are the results of the comparison of 10 subjects from UK-Biobank with the methods from both branches (this one and [jca/sb/3063-csa-measure-from-pmj](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb%2F3063-csa-measure-from-pmj))

### Method # 1  - extrapolation of the centerline
Extrapolation of the centerline was good for 4/10 subjects
Failed for 6/10:
* Bad in R-L direction: 2/6
* Bad in A-P direction: 4/6

I compared the output mask of method # 1 and # 2 for the 4 successful centerline extrapolation and it is is pretty similar (it differs by max 2 slices) that is great!

### Method # 2 - linear interpolation with PMJ and centerline
Centerline is good for all 10 subjects 🎉

### Conclusion
Method # 2 is definitely more robust!

I summarized the results in google slides [here](https://docs.google.com/presentation/d/1DXM-HGBXxzqmp61_6v_sQu44nBj8WWP4KZSPgdUV25g/edit#slide=id.ge39d03680e_0_83) with more details.
PS: Slides 8 and 9 are nice to see to check the difference of the extent mask for `-distance 55` accross subjects!

jcohenadad:
Wow this is a great investigation @sandrinebedard 🏅  Super useful! Indeed, method 2 seems more robust so let's use that for the UKB study, and we will do another more thorough study where we compare both methods, and also include multiple scans from the same individuals (with head tilting).

joshuacwnewton:
Just to check in on the status of this PR: 

> Wow this is a great investigation @sandrinebedard :medal_sports: Super useful! Indeed, method 2 seems more robust so let's use that for the UKB study, and we will do another more thorough study where we compare both methods, and also include multiple scans from the same individuals (with head tilting).

If method 2 is the preferable method, should we close this PR and open a new one based off of the work in `jca/sb/3063-csa-measure-from-pmj`?

jcohenadad:
> Just to check in on the status of this PR:
> 
> > Wow this is a great investigation @sandrinebedard 🏅 Super useful! Indeed, method 2 seems more robust so let's use that for the UKB study, and we will do another more thorough study where we compare both methods, and also include multiple scans from the same individuals (with head tilting).
> 
> If method 2 is the preferable method, should we close this PR and open a new one based off of the work in `jca/sb/3063-csa-measure-from-pmj`?

Good question. We are planning to write a paper that compares the two methods, and for the sake of reproducibility and transparency, we would like to keep both codebases visible. What would be the best course of action? keeping the two branches forever? or implementing the two functions? (that seems unnecessary if ultimately one will be abandoned). Eager to hear your thoughts...

joshuacwnewton:
> We are planning to write a paper that compares the two methods, and for the sake of reproducibility and transparency, we would like to keep both codebases visible.

Aha! Thank you for the extra context. :slightly_smiling_face: 

> What would be the best course of action? keeping the two branches forever? or implementing the two functions? (that seems unnecessary if ultimately one will be abandoned). Eager to hear your thoughts...

If it's for the purposes of a paper, I would suggest not storing the code in the `spinalcordtoolbox` repo if we can help it, and instead putting the source code in a dedicated, separate place. 

> (I say this because I'm reminded of the `flirtsch` scripts. Right now, we [can't move or modify them](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3461#discussion_r672328846) because there is a direct link in Tanguy's paper. My goal would be to avoid this, and make sure we have freedom to make any changes we like in the future. So, it would be a shame if there were more parts of SCT repo (branches, folders, etc.) that we would have to tiptoe around.)

I'm wondering if in the long-term, we could store the `csa_pmj.py` script in a dedicated repo in, say, @sct-pipeline. 

@sandrinebedard Looking at this PR, you've done an excellent job at keeping the core logic of your method separate from other parts of SCT, so moving the script would be easy enough. :slightly_smiling_face:

The only downside is, this PR integrates the method with `sct_process_segmentation` for some convenience functionality (e.g. slice metric aggregation, CSV file generation, etc.). So, hypothetically, I'm wondering... would the paper need to use the extra functionality from `sct_process_segmentation`? Or, would `csa_pmj.py` be enough by itself to perform the comparison? If only `csa_pmj.py` is needed, then it would be much easier to provide reproduction steps in a separate repo.

sandrinebedard:
> The only downside is, this PR integrates the method with `sct_process_segmentation` for some convenience functionality (e.g. slice metric aggregation, CSV file generation, etc.). So, hypothetically, I'm wondering... would the paper need to use the extra functionality from `sct_process_segmentation`? Or, would `csa_pmj.py` be enough by itself to perform the comparison? If only `csa_pmj.py` is needed, then it would be much easier to provide reproduction steps in a separate repo, but I understand if that makes things trickier. 😅

But this part is also intergrated in `jca/sb/3063-csa-measure-from-pmj` that we would keep, only `csa_pmj.py` varies for both branches, the intergration in `sct_process_segmentation` is the same 🙂

jcohenadad:
> If it's for the purposes of a paper, I would suggest not storing the code in the spinalcordtoolbox repo if we can help it, and instead putting the source code in a dedicated, separate place.

> (I say this because I'm reminded of the flirtsch schedule files. Right now, we can't move or modify them because there is a direct link in Tanguy's paper. My goal would be to avoid this, and make sure we have freedom to make any changes we like in the future. So, it would be a shame if there were more parts of SCT repo (branches, folders, etc.) that we would have to tiptoe around.)

> I'm wondering, then, if we could store the csa_pmj.py script in a dedicated repo (in, say, @sct-pipeline) for the long-term.

i was also thinking that. So let's do this!

sandrinebedard:
> > If it's for the purposes of a paper, I would suggest not storing the code in the spinalcordtoolbox repo if we can help it, and instead putting the source code in a dedicated, separate place.
> 
> > (I say this because I'm reminded of the flirtsch schedule files. Right now, we can't move or modify them because there is a direct link in Tanguy's paper. My goal would be to avoid this, and make sure we have freedom to make any changes we like in the future. So, it would be a shame if there were more parts of SCT repo (branches, folders, etc.) that we would have to tiptoe around.)
> 
> > I'm wondering, then, if we could store the csa_pmj.py script in a dedicated repo (in, say, @sct-pipeline) for the long-term.
> 
> i was also thinking that. So let's do this!

I've created a new pull request #3478 with the branch  `jca/sb/3063-csa-measure-from-pmj` and created a new repo in @sct-pipeline --> [pmj-based-csa](https://github.com/sct-pipeline/pmj-based-csa) and added the script `csa_pmj.py`, was there anything else before closing this PR?

joshuacwnewton:
> I've created a new pull request #3478 with the branch jca/sb/3063-csa-measure-from-pmj and created a new repo in @sct-pipeline --> pmj-based-csa and added the script csa_pmj.py, was there anything else before closing this PR?

Just a small, quick thing: Could you update both branches to the at the same place against `master`? Once that's done, I think we should be able to use this link to compare the two branches: 

* https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/sb/3063-csa-measure-from-pmj...jca/sb/3063-csa-measure-from-pmj

Although honestly, I'm not sure if that link is _that_ useful, but I'm curious to try it out. :yum:

sandrinebedard:
> Just a small, quick thing: Could you make sure both branches are up-to-date with `master`? Once that's done, I think we should be able to use this link to compare the two branches directly:
> 
> * [sb/3063-csa-measure-from-pmj...jca/sb/3063-csa-measure-from-pmj](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/sb/3063-csa-measure-from-pmj...jca/sb/3063-csa-measure-from-pmj) (The link looks a little "off" right now because they're both synced with different `master` commits.)

I just merged master into both branches, but the link to compare still doesn't seem to work 😬

joshuacwnewton:
> I just merged master into both branches, but the link to compare still doesn't seem to work grimacing

Ah, my apologies then. It looks like GitHub's diff compares changes across all of the commits (including merge commits). I mistakenly thought that would compare the state of both branches.

Thankfully, this comparison is still possible in PyCharm! :slightly_smiling_face: 

Screenshot from 2021-07-19 16-35-11

joshuacwnewton:
> Was there anything else before closing this PR?

I think it should be safe to close this PR, since the only real difference between the two branches was `csa_pmj.py`, and your original script is now backed up in the @sct-pipeline repo. :slightly_smiling_face: 

So, we can focus on #3478 and the `jca/sb/3063-csa-measure-from-pmj` branch moving forward. Thanks for taking the time to tidy everything up! :smile:

PR #3447, opened 2021-07-11
Remove misleading info about unsupported installation/usage methods (pip, API)

### LINKED ISSUE #3385: "Install with pip (experimental)" doesn't actually install SCT's dependencies
We have a set of steps for installing with pip [here](https://spinalcordtoolbox.com/en/latest/user_section/installation.html#install-with-pip-experimental). However, `pip install -e .` doesn't install SCT's dependencies, because `install_requires` is never specified in [`setup.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/setup.py). 

Instead, we install the dependencies via `install_sct`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0e8fa62535e5496ccbee114793764b7dc03f72ee/install_sct#L637

I think we should either:

* Try to parse requirements.txt/requirements-freeze.txt in `setup.py`, not `install_sct`, to facilitate `pip` installs a little better. (FWIW, this is currently a TODO for #1526.)
* Remove the "experimental" installation instructions for SCT, as to not mislead anyone. (I think this might have been the case in Eva's forum post here: https://forum.spinalcordmri.org/t/issue-when-running-sct-deepseg-sc-within-a-google-colab-notebook/701/2.)

### LINKED ISSUE #3439: Update the documentation to be clearer that the Python API is unavailable
Until we sort out #1526, there isn't much use advertising the Python API, as there isn't an easy way to access SCT as a Python package. It's misleading, and leads to situations like https://forum.spinalcordmri.org/t/using-sct-via-python/575/3.

We should consider hiding this section entirely in the interim, as well as removing reference to pip installs (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3385).

joshuacwnewton:
#### GitHub

#### PR contents

- [X] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

Pip installation + Python API usage is not fully available for SCT users, but we have several pages/sections on our documentation that advertise these features. So, this PR hides/removes information related to those topics to prevent users from being misled. (Both of the linked issues stemmed from users on the forum following this information and running into problems.)

## Linked issues

Fixes #3385.
Fixes #3439.


Drulex (commented):
LGTM.

One thing: since we don't need the API section (for the moment) maybe it's worth removing the step that builds SCT in order to speed up the doc build? I think it was needed for the `autodoc` extension.

Noticed here: https://readthedocs.org/projects/spinalcordtoolbox/builds/14214516/ that it takes 195s to build the documentation.

joshuacwnewton:
> One thing: since we don't need the API section (for the moment) maybe it's worth removing the step that builds SCT in order to speed up the doc build? I think it was needed for the `autodoc` extension.

IIUC currently the most expensive part of the docs building is the [Command-Line Tools page](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/documentation/source/user_section/command-line.rst), because calling `sct_function -h` is quite expensive. (When I build the docs locally, this is the bit that takes the longest, and I usually end up excluding it temporarily when working on the docs. When I do so, it the build time decreases dramatically.)

So, I'm not sure we can remove building the docs with the changes from this PR, since `api.rst` is the only thing being excluded. :sweat:

Drulex:
> IIUC currently the most time-consuming part of the docs building is the [Command-Line Tools page](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/documentation/source/user_section/command-line.rst), because calling `sct_function -h` is quite expensive. (When I build the docs locally, this is the bit that takes the longest. I usually end up excluding it temporarily when working on the docs, because the build time decreases dramatically when I do.)

Ahh yes indeed, somehow I though `autodoc` was used for the API doc instead of CLI doc. Sorry about that, approving.

joshuacwnewton:
> Ahh yes indeed, somehow I though `autodoc` was used for the API doc instead of CLI doc.

Ah, `autodoc` is used that way! The CLI doc uses e.g.

```
.. program-output:: sct_analyze_lesion -h
```

Drulex (approved):

PR #3465, opened 2021-07-15
Add QC report for `sct_process_segmentation` for PMJ-based CSA

### LINKED ISSUE #3431: Add visualization of the centerline smoothed in RL direction in `sct_qc`
## Context
With the new method measuring CSA from distance of PMJ (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429), the distance is measured along the extrapolated centerline. Visual assessment of the extrapolated centerline is important to detect if there is failure.

## Suggestion
As suggested in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429#commitcomment-52668990, a solution to visually assess the quality of the extrapolated centerline would be to smooth the centerline in the RL direction and display the sagittal view in the QC.

sandrinebedard:
#### GitHub

#### PR contents

## Description
This PR addresses #3431. It modifies the QC report for `sct_process_segmentation` to visualize the extrapolated centerline smoothed in R-L direction, the PMJ label and the coverage mask.

Currently, the QC report for `sct_process_segmentation ` was only available inside `sct_process_segmentation` and not when calling `sct_qc`. I left it that way, we could also add it. 

**I created this feature from the branch [jca/sb/3063-csa-measure-from-pmj](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb%2F3063-csa-measure-from-pmj), I'll merge inside the branch and not the master when it is ready.**

The overlay includes smoothed centerline, coverage mask and PMJ label (in one overlay), do we want to seprate them ?(that would require some further changes). Also, to use the QC with the[ extrapolation method of the centerline](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/sb/3063-csa-measure-from-pmj) (what is implemented in #3429 ), there is very little chance that PMJ label is at the same R-L slice than the extrapolated centerline, it will definitely be a problem for that since the image is flatten from the centerline.

Here is what it looks like:

image

To test it out, run `sct_process_segmentation` with `-i, -pmj, -distance, -qc, and -qc-image`

## Linked issues
Resolves #3431

### DIFF
--- a/spinalcordtoolbox/csa_pmj.py
+++ b/spinalcordtoolbox/csa_pmj.py
@@ -80,7 +80,7 @@ def get_slices_for_pmj_distance(segmentation, pmj, distance, extent, param_cente
     mask.change_orientation(native_orientation)
 
     np.savetxt("centerline.csv", arr_ctl, delimiter=",")
-    
+
     # Get corresponding slices
     # TODO: why the "-1"?
     slices = "{}:{}".format(zmin, zmax-1)
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -131,7 +131,6 @@ class QcImage(object):
 
     def listed_seg(self, mask, ax):
         """Create figure with red segmentation. Common scenario."""
-        #mask = mask/mask.max()  # TODO: move when generating smooth centerline
         img = np.ma.masked_equal(mask, 0)
         ax.imshow(img,
                   cmap=color.LinearSegmentedColormap.from_list("", self._seg_colormap),
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -850,6 +850,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
     # Metric outputs (only graphs)
     elif process in ['sct_process_segmentation']:
         plane = 'Sagittal'
+        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
         fname_list = [fname_in1]
         fname_list.extend(fname_seg)
         qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -266,7 +266,7 @@ class QcImage(object):
 
     def smooth_centerline(self, mask, ax):
         """Display smoothed centerline"""
-        mask = mask/mask.max()  # TODO: move when generating smooth centerline
+        mask = mask/mask.max()
         mask[mask < 0.05] = 0  # Apply 0.5 threshold
         img = np.ma.masked_equal(mask, 0)
         ax.imshow(img,
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -851,6 +851,8 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
         plane = 'Sagittal'
         dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
         fname_list = [fname_in1]
+        # fname_seg should be a list of 4 images: 3 for each of the `qcslice_operations`, plus an extra
+        # centerline image, which is needed to make `Sagittal.get_center_spit` work correctly
         fname_list.extend(fname_seg)
         qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
         qcslice_operations = [QcImage.smooth_centerline, QcImage.highlight_pmj, QcImage.listed_seg]
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -854,10 +854,9 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
         # fname_seg should be a list of 4 images: 3 for each of the `qcslice_operations`, plus an extra
         # centerline image, which is needed to make `Sagittal.get_center_spit` work correctly
         fname_list.extend(fname_seg)
-        qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
+        qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)
         qcslice_operations = [QcImage.smooth_centerline, QcImage.highlight_pmj, QcImage.listed_seg]
         def qcslice_layout(x): return x.single()
-        # assert os.path.isfile(path_img)  # TODO: REMOVE
     else:
         raise ValueError("Unrecognized process: {}".format(process))
 
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -779,7 +779,6 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
     qcslice_operations = None
     qcslice_layout = None
 
-
     # Get QC specifics based on SCT process
     # Axial orientation, switch between two input images
     if process in ['sct_register_multimodal', 'sct_register_to_template']:
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -62,7 +62,7 @@ class QcImage(object):
                      "#7d0434", "#fb1849", "#14aab4",
                      "#a22abd", "#d58240", "#ac2aff"]
     _seg_colormap = ["#4d0000", "#ff0000"]
-
+    _ctl_colormap = ["#ff000099", '#ffff00']
 
     def __init__(self, qc_report, interpolation, action_list, process, stretch_contrast=True,
                  stretch_contrast_method='contrast_stretching', angle_line=None, fps=None):
@@ -131,6 +131,7 @@ class QcImage(object):
 
     def listed_seg(self, mask, ax):
         """Create figure with red segmentation. Common scenario."""
+        #mask = mask/mask.max()  # TODO: move when generating smooth centerline
         img = np.ma.masked_equal(mask, 0)
         ax.imshow(img,
                   cmap=color.LinearSegmentedColormap.from_list("", self._seg_colormap),
@@ -264,6 +265,19 @@ class QcImage(object):
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
 
+    def smooth_centerline(self, mask, ax):
+        """Display smoothed centerline"""
+        mask = mask/mask.max()  # TODO: move when generating smooth centerline
+        mask[mask < 0.05] = 0  # Apply 0.5 threshold
+        img = np.ma.masked_equal(mask, 0)
+        ax.imshow(img,
+                  cmap=color.LinearSegmentedColormap.from_list("", self._ctl_colormap),
+                  norm=color.Normalize(vmin=0, vmax=1),
+                  interpolation=self.interpolation,
+                  aspect=float(self.aspect_mask))
+        ax.get_xaxis().set_visible(False)
+        ax.get_yaxis().set_visible(False)
+
     # def colorbar(self):
     #     fig = plt.figure(figsize=(9, 1.5))
     #     ax = fig.add_axes([0.05, 0.80, 0.9, 0.15])
@@ -294,10 +308,9 @@ class QcImage(object):
                 [images_after_moco, images_before_moco], centermass = func(sct_slice, *args)
                 self._centermass = centermass
                 self._make_QC_image_for_4d_volumes(images_after_moco, images_before_moco)
-
             else:
                 if self._angle_line is None:
-                    img, mask = func(sct_slice, *args)
+                    img, *mask = func(sct_slice, *args)
                 else:
                     [img, mask], centermass = func(sct_slice, *args)
                     self._centermass = centermass
@@ -311,7 +324,7 @@ class QcImage(object):
         Create overlay and background images for all processes that deal with 3d volumes
         (all except sct_fmri_moco and sct_dmri_moco)
 
-        :param img: list of mosaic images after motion correction
+        :param img: list of mosaic images after motion correction  # Doesn't seem correct?
         :param mask: list of mosaic images before motion correction
         :return:
         """
@@ -337,17 +350,17 @@ class QcImage(object):
         logger.info(self.qc_report.qc_params.abs_bkg_img_path())
         self._save(fig, self.qc_report.qc_params.abs_bkg_img_path(), dpi=self.qc_report.qc_params.dpi)
 
-        for action in self.action_list:
+        fig = Figure()
+        fig.set_size_inches(size_fig[0], size_fig[1], forward=True)
+        FigureCanvas(fig)
+        for i, action in enumerate(self.action_list):
             logger.debug('Action List %s', action.__name__)
             if self._stretch_contrast and action.__name__ in ("no_seg_seg",):
-                print("Mask type %s" % mask.dtype)
-                mask = self._func_stretch_contrast(mask)
-            fig = Figure()
-            fig.set_size_inches(size_fig[0], size_fig[1], forward=True)
-            FigureCanvas(fig)
-            ax = fig.add_axes((0, 0, 1, 1))
-            action(self, mask, ax)
-            self._save(fig, self.qc_report.qc_params.abs_overlay_img_path(), dpi=self.qc_report.qc_params.dpi)
+                print("Mask type %s" % mask[i].dtype)
+                mask[i] = self._func_stretch_contrast(mask[i])
+            ax = fig.add_axes((0, 0, 1, 1), label=str(i))
+            action(self, mask[i], ax)
+        self._save(fig, self.qc_report.qc_params.abs_overlay_img_path(), dpi=self.qc_report.qc_params.dpi)
 
         self.qc_report.update_description_file(img.shape)
 
@@ -767,6 +780,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
     qcslice_operations = None
     qcslice_layout = None
 
+
     # Get QC specifics based on SCT process
     # Axial orientation, switch between two input images
     if process in ['sct_register_multimodal', 'sct_register_to_template']:
@@ -836,7 +850,13 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
         def qcslice_layout(x): return x.single()
     # Metric outputs (only graphs)
     elif process in ['sct_process_segmentation']:
-        assert os.path.isfile(path_img)
+        plane = 'Sagittal'
+        fname_list = [fname_in1]
+        fname_list.extend(fname_seg)
+        qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
+        qcslice_operations = [QcImage.smooth_centerline, QcImage.highlight_pmj, QcImage.listed_seg]
+        def qcslice_layout(x): return x.single()
+        # assert os.path.isfile(path_img)  # TODO: REMOVE
     else:
         raise ValueError("Unrecognized process: {}".format(process))
 
@@ -855,7 +875,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, angle_line=None, args
         qcslice_layout=qcslice_layout,
         stretch_contrast_method='equalized',
         angle_line=angle_line,
-        fps=fps,
+        fps=fps
     )
 
 
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -323,8 +323,8 @@ class QcImage(object):
         Create overlay and background images for all processes that deal with 3d volumes
         (all except sct_fmri_moco and sct_dmri_moco)
 
-        :param img: list of mosaic images after motion correction  # Doesn't seem correct?
-        :param mask: list of mosaic images before motion correction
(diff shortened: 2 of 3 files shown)


spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -320,27 +367,54 @@ def main(argv=None):
...
+        # Save mask
+        fname_mask_out = add_suffix(arguments.i, '_mask_csa')
+        mask.save(fname_mask_out)
+        # Save extrapolated centerline
+        fname_ctl = add_suffix(arguments.i, '_centerline_extrapolated')
+        im_ctl.save(fname_ctl)
+
+        # Generated centerline smoothed in RL direction for visualization (and QC report)
+        fname_ctl_smooth = add_suffix(fname_ctl, '_smooth')
+        subprocess.run(['sct_maths', '-i', fname_ctl, '-smooth', '10,1,1', '-o', fname_ctl_smooth], stdout=subprocess.PIPE, check=True)
```
sandrinebedard, line 384: I wanted to generate the smoothed centerline, I wasn't sure if there was another wat to call `sct_maths `for it? (I wanted to use the API` maths.py`, but I would have still repeated some lines from `sct_maths.py`)

joshuacwnewton:
> I created this feature from the branch `jca/sb/3063-csa-measure-from-pmj`, I'll merge inside the branch and not the master when it is ready.

@sandrinebedard I'd like to try temporarily changing the base branch of this PR to `jca/sb/3063-csa-measure-from-pmj` just to make reviewing a little easier. (If it gets confusing, it's easy to change it back.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: At the very least, you can import `sct_maths` then access it via `sct_maths.main()` to avoid the subprocess call.

joshuacwnewton: At the very least, you can import `sct_maths` then access it via `sct_maths.main()` to avoid the subprocess call.

testing/api/test_csa_pmj.py
```diff
@@ -1,44 +0,0 @@
...
-    fname_ctl = sct_test_path('t2', 't2_centerline-manual.nii.gz')
-    im_ctl = Image(fname_ctl).change_orientation('RPI')
-    _, _, _, _, px, py, pz, _ = im_ctl.dim
-    # Create an array like output of get_centerline()
-    arr_ctl = np.array(np.nonzero(im_ctl.data))
-    arr_ctl = arr_ctl[::, arr_ctl[2, ].argsort()]
-    return arr_ctl, px, py, pz
-
-
-def test_distance_from_pmj(centerline):
```
joshuacwnewton, line 29: Can we modify @jcohenadad's branch (`jca/sb/3063-csa-measure-from-pmj`) so that these tests don't fail? That way we won't need a commit to remove this test, I think?

spinalcordtoolbox/csa_pmj.py
```diff
@@ -33,7 +32,7 @@ def get_slices_for_pmj_distance(segmentation, pmj, distance, extent, param_cente
     im_seg.change_orientation('RPI')
     im_pmj = Image(pmj).change_orientation('RPI')
     if not im_seg.data.shape == im_pmj.data.shape:
-        raise RuntimeError(f"segmentation and pmj should be in the same space coordinate.")
+        raise RuntimeError(f'{"segmentation and pmj should be in the same space coordinate."}')
```
joshuacwnewton, line 35: Same sort of comment -- can we fix these linting issues in @jcohenadad's branch, that way we won't need to do it here?

joshuacwnewton (changes requested):
I've changed the base branch so that this PR will show only the commits that are different from `jca/sb/3063-csa-measure-from-pmj`. But, l'm wondering if we could maybe tidy up the branches for both this PR and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429 a little more.

Something like:

• Pick a single branch to use for https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429 (either your branch or Julien's branch, but not both)
• Move some of the fixes from this PR into https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429 (linting, test deletion, etc. see below).
• Clean up this PR so that it only contains the QC-related changes.

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: great i'll do that! thanks!

testing/api/test_csa_pmj.py
sandrinebedard: actually because it uses the class Centerline, (not possible for the other method), these function don't exists in `jca/sb/3063-csa-measure-from-pmj` so we can't really test them

sandrinebedard: but I'll remove them in the branch `jca/sb/3063-csa-measure-from-pmj`

spinalcordtoolbox/csa_pmj.py
joshuacwnewton: Ah, I should also mention: Instead of adding `{` to the string, I think you can just get rid of the `f""` entirely and make it a normal string.

sandrinebedard: yes on it!

testing/api/test_csa_pmj.py
joshuacwnewton: > actually because it uses the class Centerline, (not possible for the other method), these function don't exists in jca/sb/3063-csa-measure-from-pmj so we can't really test them

Oh! I must have misunderstood, sorry. I didn't realize the tests were only valid for your branch (and not Julien's branch). Thank you for the clarification.

sandrinebedard:
> I've changed the base branch so that this PR will show only the commits that are different from `jca/sb/3063-csa-measure-from-pmj`. But, l'm wondering if we could maybe tidy up the branches for both this PR and #3429 a little more.
> 
> Something like:
> 
> • Pick a single branch to use for #3429 (either your branch or Julien's branch, but not both)
> * EDIT: What I mean to say is, we should probably close #3429, then open a new PR based from Julien's branch.
> • Move some of the fixes from this PR into the branch for #3429 (linting, test deletion, etc. see below).
> • Clean up this PR so that it only contains the QC-related changes.

@joshuacwnewton I've cleaned up the PR, thanks for your insights on this!

joshuacwnewton:
> @joshuacwnewton I've cleaned up the PR, thanks for your insights on this!

Hmm, it's strange. I see the changes you made on the [`jca/sb/3063-csa-measure-from-pmj`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb/3063-csa-measure-from-pmj), yet those same changes are still showing up in this PR.

I would have expected GitHub to remove the linting/test/etc. changes from the "[Files Changed](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465/files)" view of this PR, because they're already there in the other branch! :stuck_out_tongue_closed_eyes:

sandrinebedard:
> > @joshuacwnewton I've cleaned up the PR, thanks for your insights on this!
> 
> Hmm, it's strange. I see the changes you made on the [`jca/sb/3063-csa-measure-from-pmj`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/jca/sb/3063-csa-measure-from-pmj), yet for some reason those same changes are still showing up in this PR. 🤔
> 
> I would have expected GitHub to remove the linting/test/etc. changes from the "[Files Changed](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465/files)" view of this PR, because they're already there in the other branch! 😝

oupps this was my mistake, should be good right now 😊

sandrinebedard:
@jcohenadad if you have a bit of time, could you take a quick look at this 😊, thanks

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -385,12 +400,21 @@ def main(argv=None):
...
+    # QC report (only for PMJ-based CSA)
     if path_qc is not None:
-        generate_qc(fname_segmentation, args=arguments, path_qc=os.path.abspath(path_qc), dataset=qc_dataset,
-                    subject=qc_subject, path_img=_make_figure(metrics_agg_merged, fit_results),
-                    process='sct_process_segmentation')
+        if fname_pmj is not None:
+            if arguments.qc_image is not None:
+                generate_qc(fname_in1=get_absolute_path(arguments.qc_image),
+                            fname_seg=[fname_ctl_smooth, fname_pmj, fname_mask_out, fname_ctl_smooth],  # Repeat centerline to flatten image along the centerline
+                            args=arguments,
```
joshuacwnewton, line 409: Passing `arguments` will cause all of the non-passed arguments to appear as well. (See screenshot.)

Looking at some other calls to `generate_sc`, it looks like `args=sys.argv[1:]` lets us pass only the args that were specified by the user.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c79a02021509da8b3c84878c6029a42a26a3d2d8/spinalcordtoolbox/scripts/sct_deepseg_gm.py#L127

spinalcordtoolbox/reports/qc.py
```diff
@@ -337,17 +349,17 @@ def _make_QC_image_for_3d_volumes(self, img, mask, slice_orientation):
...
-            fig.set_size_inches(size_fig[0], size_fig[1], forward=True)
-            FigureCanvas(fig)
-            ax = fig.add_axes((0, 0, 1, 1))
-            action(self, mask, ax)
-            self._save(fig, self.qc_report.qc_params.abs_overlay_img_path(), dpi=self.qc_report.qc_params.dpi)
+                print("Mask type %s" % mask[i].dtype)
+                mask[i] = self._func_stretch_contrast(mask[i])
+            ax = fig.add_axes((0, 0, 1, 1), label=str(i))
+            action(self, mask[i], ax)
+        self._save(fig, self.qc_report.qc_params.abs_overlay_img_path(), dpi=self.qc_report.qc_params.dpi)
```
joshuacwnewton, line 362: I'm wondering if there's a way to make the QC image a little smaller. It's so big that it pushes the text underneath off the page. (See screenshot.)

joshuacwnewton (changes requested):
I tested this using the following commands:

```
sct_detect_pmj -i sct_example_data/t2/t2.nii.gz -c t2
sct_deepseg_sc -i sct_example_data/t2/t2.nii.gz -c t2
sct_process_segmentation -i t2_seg.nii.gz -pmj t2_pmj.nii.gz -distance 50 
                         -qc ~/qc_SingleSubj -qc-image sct_example_data/t2/t2.nii.gz
```

And this is what I see:

Screenshot from 2021-07-21 13-35-04

I want to do a more thorough review to try and better understand the QC plotting code, but I wanted to get a quick review in for now (just to share some exact commands used to test).

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: > Looking at some other calls to generate_sc, it looks like args=sys.argv[1:] lets us pass just the args that were specified by the user.

I tested it out in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465/commits/a8170907e41d4d67bbe960f3ceb8d24865c2e2b8 and you are right, it lets us pass just the args that were specified by the user 🙂

spinalcordtoolbox/reports/qc.py
sandrinebedard: I reduced dpi to 100 instead of 300 as it is done for `sct_label_vertebrae` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465/commits/98df805a3bb1488912e37205af85dd75305a9927

joshuacwnewton: Much better! 

Screenshot from 2021-07-21 15-48-52

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -184,6 +188,12 @@ def get_parser():
         metavar=Metavar.folder,
         action=ActionCreateFolder,
         help="The path where the quality control generated content will be saved."
+             "QC report is only availbale for PMJ-based CSA."
```
jcohenadad, line 191:
```suggestion
             " The QC report is only available for PMJ-based CSA (with flag '-pmj')."
```
A space was needed at the beginning of the line.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -184,6 +188,12 @@ def get_parser():
...
+    )
+    optional.add_argument(
+        '-qc-image',
+        metavar=Metavar.str,
+        help='Input image to display in QC report. To be used with flag -qc '
```
jcohenadad, line 196:
```suggestion
        help="Input image to display in QC report. Typically, it would be the image that was used for spinal cord segmentation. This flag is mandatory if using flag '-qc'."
```

jcohenadad (approved):
I only left minor comments (no need for re-review). This is a great contribution @sandrinebedard !!! Thank you 🙏

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Ahh, this is a good clarification!

With this new wording, though, I wonder if people might misread it and accidentally supply the segmented spinal cord here too. This wording might be a little more explicit:

```suggestion
        help="Input image to display in QC report. Typically, it would be the "
             "source anatomical image used to generate the spinal cord "
             segmentation. This flag is mandatory if using flag '-qc'."
```

spinalcordtoolbox/reports/qc.py
joshuacwnewton, line 362: I tried taking a look at this section using the commands from my previous review, and I noticed that `self.action_list` had 3 entries, but `mask` had 4 volumes, so I think the last volume never gets used.

Screenshot from 2021-07-22 11-55-08

The 4 volumes correspond to `fname_list[1:4]`:

Screenshot from 2021-07-22 12-01-41

So, the 4th volume is `t2_seg_centerline_extrapolated_smooth.nii.gz`... but it's repeated? So how did this get here... (Answer: Link to other review comment.)

spinalcordtoolbox/reports/qc.py
```diff
@@ -836,7 +848,14 @@ def qcslice_layout(x): return x.single()
         def qcslice_layout(x): return x.single()
     # Metric outputs (only graphs)
     elif process in ['sct_process_segmentation']:
-        assert os.path.isfile(path_img)
+        plane = 'Sagittal'
+        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
+        fname_list = [fname_in1]
+        fname_list.extend(fname_seg)
```
joshuacwnewton, line 854: I think this line is the cause of the extra image in `mask` that is never used...

joshuacwnewton (changes requested):
I just had a few questions about some of the changes QC code (just so  I can understand a little better).

spinalcordtoolbox/reports/qc.py
```diff
@@ -264,6 +264,19 @@ def grid(self, mask, ax):
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
 
+    def smooth_centerline(self, mask, ax):
+        """Display smoothed centerline"""
+        mask = mask/mask.max()  # TODO: move when generating smooth centerline
```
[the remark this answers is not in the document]
sandrinebedard: Ahhh forgot to remove this comment! (It was moved to there 😅)

spinalcordtoolbox/reports/qc.py
sandrinebedard: Yeah it is repeated, the problem is that the last volume is used to select the sagittal slice for the Qc, so I want it to be the smoothed centerline, but if I leave it at the end, it will be displayed over PMJ and the mask (not the desired display), this was the solution that I found for this, there may be a better way if you have any ideas 🙂

joshuacwnewton: Ohh, I see. Thank you for clarifying! I'll try to take a look and see if there's a better way.

If there isn't a better way, perhaps at the very least we could leave a comment explaining this behavior. (Tracing the QC module is quite a pain, isn't it... :sweat_smile:. The wrapped functions are a little hard to understand! So, if there are any comments we can leave here that might help people navigate this code in the future, that might be good.)

sandrinebedard: Found the comment, I knew I specified it somewhere 😅:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/754cbe86eec066035481a328684f768d5f0f9875/spinalcordtoolbox/scripts/sct_process_segmentation.py#L408

I could also add it in qc.py

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -385,12 +400,21 @@ def main(argv=None):
...
-    # QC report (only show CSA for clarity)
+    # QC report (only for PMJ-based CSA)
     if path_qc is not None:
-        generate_qc(fname_segmentation, args=arguments, path_qc=os.path.abspath(path_qc), dataset=qc_dataset,
-                    subject=qc_subject, path_img=_make_figure(metrics_agg_merged, fit_results),
-                    process='sct_process_segmentation')
+        if fname_pmj is not None:
+            if arguments.qc_image is not None:
+                generate_qc(fname_in1=get_absolute_path(arguments.qc_image),
+                            fname_seg=[fname_ctl_smooth, fname_pmj, fname_mask_out, fname_ctl_smooth],  # Repeat centerline to flatten image along the centerline
```
joshuacwnewton, line 408: Here is a more verbose comment that explains the entire situation!

```suggestion
                            # NB: For this QC figure, the centerline has to be first in the list in order for the PMJ
                            # marker to be properly layered on top of the centerline. However, Sagittal.get_center_spit
                            # is called during QC, and it uses `fname_seg[-1]` to center the slices. `fname_mask_out`
                            # doesn't work for this, so we have to repeat `fname_ctl_smooth` at the end of the list.
                            fname_seg=[fname_ctl_smooth, fname_pmj, fname_mask_out, fname_ctl_smooth], 
```

spinalcordtoolbox/reports/qc.py
```diff
@@ -836,7 +848,14 @@ def qcslice_layout(x): return x.single()
...
     elif process in ['sct_process_segmentation']:
-        assert os.path.isfile(path_img)
+        plane = 'Sagittal'
+        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
+        fname_list = [fname_in1]
+        fname_list.extend(fname_seg)
+        qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
+        qcslice_operations = [QcImage.smooth_centerline, QcImage.highlight_pmj, QcImage.listed_seg]
+        def qcslice_layout(x): return x.single()
+        # assert os.path.isfile(path_img)  # TODO: REMOVE
```
joshuacwnewton, line 858: `#TODO: REMOVE`?

spinalcordtoolbox/reports/qc.py
```diff
@@ -836,7 +848,14 @@ def qcslice_layout(x): return x.single()
         def qcslice_layout(x): return x.single()
     # Metric outputs (only graphs)
     elif process in ['sct_process_segmentation']:
-        assert os.path.isfile(path_img)
+        plane = 'Sagittal'
+        dpi = 100  # bigger picture is needed for this special case, hence reduce dpi
+        fname_list = [fname_in1]
+        fname_list.extend(fname_seg)
+        qcslice_type = qcslice.Sagittal([Image(fname) for fname in fname_list], p_resample=None)  # TODO: add multiple overlays
```
joshuacwnewton, line 855: `# TODO: add multiple overlays`?

spinalcordtoolbox/reports/qc.py
joshuacwnewton, line 854:
```suggestion
        # fname_seg should contain 4 images: 3 for each of the `qcslice_operations`, plus an extra
        # centerline image, which is needed to make `Sagittal.get_center_spit` work correctly
        fname_list.extend(fname_seg)
```

spinalcordtoolbox/reports/qc.py
joshuacwnewton: You were absolutely right... this is necessary, sadly. :sweat_smile: 

I've added some suggestions for comments to make this a little clearer perhaps: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465#discussion_r675561784 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465#discussion_r675574281)

spinalcordtoolbox/reports/qc.py
```diff
@@ -311,7 +323,7 @@ def _make_QC_image_for_3d_volumes(self, img, mask, slice_orientation):
         Create overlay and background images for all processes that deal with 3d volumes
         (all except sct_fmri_moco and sct_dmri_moco)
 
-        :param img: list of mosaic images after motion correction
+        :param img: list of mosaic images after motion correction  # Doesn't seem correct?
```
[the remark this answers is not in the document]
joshuacwnewton: Oh, I see now! The docstring was just copied and pasted from `_make_QC_image_for_4d_volumes`. Oops. :sweat_smile: 

How about this?

```
        :param img: The base anatomical image to display underneath the overlays
        :param mask: A list of images to be processed and overlaid on top of `img`
```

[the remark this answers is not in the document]
sandrinebedard: sounds good!

spinalcordtoolbox/reports/qc.py
sandrinebedard: I'll remove the TODO, it is done :)

spinalcordtoolbox/reports/qc.py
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3465/commits/1abdc8daf9d95550e4eabf3b82e2f7474d71ab8c

PR #3467, opened 2021-07-15
Add ref Harrison et al.

alexfoias:


jcohenadad:
no this was only a review-- they don't use SCT

Issue #3432, opened 2021-06-27
Output csa perslice and distance from PMJ for each slices if `-distance` not specified


sandrinebedard:
## Context
This is a feature related to [Measure CSA based on distance from pontomedullary junction (PMJ)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3429)
Currently, CSA is computed at a distance from PMJ and averaged across slices of corresponding extent.

## Suggestion
If the flag `-distance` is not specified, CSA could be computed per slice and the distance from PMJ could be specified at each slice in csa.csv.

PR #3478, opened 2021-07-19
Measure CSA based on distance from pontomedullary junction (PMJ)

### LINKED ISSUE #3063: Measure CSA based on distance from spinal cord landmark such as the pontomedullary junction (PMJ)

### LINKED ISSUE #3431: Add visualization of the centerline smoothed in RL direction in `sct_qc`

sandrinebedard:
#### GitHub

#### PR contents

## Description

This PR addressed #3063. The goal is to measure CSA based on a distance from pontomedullary junction (PMJ) To take into account the curvature of the spinal cord, we use linear interpolation between PMJ and centerline. We can after measure the distance along the centerline.

This PR adds the API `csa_pmj.py` and intergrates the method in `sct_process_segmentation` and `aggregate_slicewise.py`.
It also adds QC report for this method.

**Note: we will need to update the documentation**

## Linked issues

Resolves #3063 and closes #3431


sandrinebedard:
I've cleaned up and added some comments to the code, I think it would be ready to merge! We also discussed that we need to update the documentation with this new feature, do we want to address that soon? (I can also open an issue for this)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -320,27 +369,60 @@ def main(argv=None):
                                          angle_correction=angle_correction,
                                          param_centerline=param_centerline,
                                          verbose=verbose)
+    if fname_pmj is not None:
+        im_ctl, mask, slices, centerline = get_slices_for_pmj_distance(fname_segmentation, fname_pmj,
+                                                                       distance_pmj, extent_mask,
+                                                                       param_centerline=param_centerline,
+                                                                       verbose=verbose)
```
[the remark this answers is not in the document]
joshuacwnewton: Maybe we could add a comment in the main help description to indicate what the correct usage is supposed to be:

```diff
            "Compute the following morphometric measures based on the spinal cord segmentation:\n"
            "  - area [mm^2]: Cross-sectional area, measured by counting pixels in each slice. Partial volume can be "
            "accounted for by inputing a mask comprising values within [0,1].\n"
            "  - angle_AP, angle_RL: Estimated angle between the cord centerline and the axial slice. This angle is "
            "used to correct for morphometric information.\n"
            "  - diameter_AP, diameter_RL: Finds the major and minor axes of the cord and measure their length.\n"
            "  - eccentricity: Eccentricity of the ellipse that has the same second-moments as the spinal cord. "
            "The eccentricity is the ratio of the focal distance (distance between focal points) over the major axis "
            "length. The value is in the interval [0, 1). When it is 0, the ellipse becomes a circle.\n"
            "  - orientation: angle (in degrees) between the AP axis of the spinal cord and the AP axis of the "
            "image\n"
            "  - solidity: CSA(spinal_cord) / CSA_convex(spinal_cord). If perfect ellipse, it should be one. This "
            "metric is interesting for detecting non-convex shape (e.g., in case of strong compression)\n"
            "  - length: Length of the segmentation, computed by summing the slice thickness (corrected for the "
            "centerline angle at each slice) across the specified superior-inferior region.\n"
+           "
+           "To select the region to compute metrics over, specify any of the following options:
+           "   - '-z': Select axial slices based on slice index.
+           "   - '-pmj' + '-distance' + '-extent': Select axial slices based on distance from pontomedullary junction.
+           "   - '-vert' + '-vertfile': Select a region based on vertebral levels, instead of individual slices.

```

Just so no one accidentally mixes and matches the mutually-exclusive arguments groups.

testing/api/test_aggregate_slicewise.py
```diff
@@ -316,7 +316,7 @@ def test_save_as_csv_extract_metric(tmp_path, dummy_data_and_labels):
     with open(path_out, 'r') as csvfile:
         spamreader = csv.reader(csvfile, delimiter=',')
         next(spamreader)  # skip header
-        assert next(spamreader)[1:-1] == [__version__, '', '0:4', '', 'label_0', '2.5', '38.0']
+        assert next(spamreader)[1:-1] == [__version__, '', '0:4', '', '', 'label_0', '2.5', '38.0']
 
 
 def test_dimension_mismatch_between_metric_and_vertfile(dummy_metrics, dummy_vert_level):
```
joshuacwnewton, line 322: Can we add some tests that demonstrate typical usage of this feature? 

e.g. How would we expect the user to specify the `-pmj`, `-distance`, and `-extent` arguments? Then validating the expected output.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -156,11 +162,40 @@ def get_parser():
...
+             "To be used with flag -pmj"
+    )
+    optional.add_argument(
+        '-extent',
+        type=float,
+        metavar=Metavar.float,
+        default=20,
+        help="Extent of the mask to average CSA (mm). "
+             "To be used with flag -pmj and -distance"
+    )
```
joshuacwnewton, line 185: Do we need both of these arguments?

Say the user might specify something like `-distance 50 -extent 30`. If this just means `-pmj-region 50:80`, then maybe it would be simpler/clearer to just have the user specify the region in a single argument.

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -293,17 +295,20 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], persli
...
         else:
             # slicegroups = [(0, 1, 2, 3, 4, 5, 6, 7, 8)]
             slicegroups = [tuple(slices)]
     agg_metric = dict((slicegroup, dict()) for slicegroup in slicegroups)
-
     # loop across slice group
     for slicegroup in slicegroups:
+        # add distance from PMJ info
+        if distance_pmj is not None:
+            agg_metric[slicegroup]['DistancePMJ'] = [distance_pmj]
```
joshuacwnewton, line 309: I'm a bit confused about the way we're using `distance_pmj` inside `aggregate_per_slice_or_level`. 

If I'm understanding correctly, we're just echoing the same `distance_pmj` value specified by the user, and we're not showing the offset for each slice, right?

Is that useful enough to deserve its own column in the CSV file?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -90,7 +96,7 @@ def get_parser():
         '-z',
         metavar=Metavar.str,
         type=str,
-        help="Slice range to compute the metrics across (requires '-p csa'). Example: 5:23"
+        help="Slice range to compute the metrics across (requires '-p csa'). Example: 5:23"  # TODO: -p csa dose not exists anymore
```
joshuacwnewton, line 99: Note to address this TODO.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -18,19 +18,25 @@
```
joshuacwnewton, line 18: As far as documentation goes, I agree that we could provide a new page in [4. Computing shape metrics for PAM50-registered data](https://spinalcordtoolbox.com/en/latest/user_section/tutorials/shape-metric-computation.html).

Something like "4.5 CSA (Region based on distance from PMJ)"

joshuacwnewton (commented):
Sorry for not giving this feedback sooner... I probably should have looked at this much earlier, before the QC PR. :sweat_smile:

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -320,27 +369,60 @@ def main(argv=None):
...
+        mask.save(fname_mask_out)
+        # Save extrapolated centerline
+        fname_ctl = add_suffix(arguments.i, '_centerline_extrapolated')
+        im_ctl.save(fname_ctl)
+        # Save array of the centerline in a .csv file
+        fname_ctl_csv, _ = splitext(add_suffix(arguments.i, '_centerline_extrapolated'))
+        np.savetxt(fname_ctl_csv + '.csv', centerline, delimiter=",")
+        # Generated centerline smoothed in RL direction for visualization (and QC report)
+        fname_ctl_smooth = add_suffix(fname_ctl, '_smooth')
+        sct_maths.main(['-i', fname_ctl, '-smooth', '10,1,1', '-o', fname_ctl_smooth])
```
joshuacwnewton, line 388: All of these steps are just for visualization/feedback, right? 

Do they _need_ to be saved to files? I ask because it seems kind of noisy to pollute the user's working directory, especially since the `sct_process_segmentation` is mainly just meant for metric calculation.

So, I'm wondering if maybe the QC report good enough by itself for this kind of feedback.

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: > If I'm understanding correctly, we're just echoing the same distance_pmj value specified by the user, and we're not showing the individual offset for each slice, right?

yes that's right

> Is that useful enough to deserve its own column in the CSV file? (Since it's the same value for every row?)

It would be useful if a person would run `sct_process_segmentation` with different distances and use the flag `-append`, so the CSV file would have various distances. Also, if we want to add this feature https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3432

spinalcordtoolbox/scripts/sct_process_segmentation.py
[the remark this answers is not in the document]
sandrinebedard: yes good idea! There is also `-perlevel` that can be added with option 3, also I think `-perslice` can be mixed with `-z`, so many options 😄 we should add them too I think?

testing/api/test_aggregate_slicewise.py
sandrinebedard: > Can we add some tests that demonstrate typical usage of this feature?

yes I'll write up some tests!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: well it would actually be `-distance 50 -extent 30` --> `-pmj-region 35:65`, the extent is centered on the distance

I personnaly find it more intuitive to keep `-distance `and `-extent` since there is a default value of `-extent `and when using floats too. For example, I am using this feature in https://github.com/sct-pipeline/ukbiobank-spinalcord-csa, I decided `-distance 64`  since it is the mean distance from PMJ to C2-C3 disc, it is directly the input of `-distance`, so just one number to change if the mean changes. I would have to do the maths otherwise (±10, not very complicated I agree 😂)

maybe @jcohenadad has an opinion on this

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: The mask and extrapolated centerline are only relevant (for now) for th QC report. Maybe I will use them for a follow up project on this to validate the method, but I don't know yet, haven't planed that much of it!
I use the CSV file of the extrapolated centerline for my research after (e.g. compute distance between C2-C3 disc and PMJ)
we could hide them all in the behind `-verbose 2`, 

this way, we would save the centerline and the mask in a tmp directory?

testing/api/test_aggregate_slicewise.py
sandrinebedard: should we also add the usage to `batch_processing.sh`?

spinalcordtoolbox/scripts/sct_process_segmentation.py
jcohenadad: yes, I also think I prefer to have them separate, because of the default value for extent and also we don't need to parse it out (although it's trivial to do...). 

however we might want to consider using "pmj" as a prefix for these flags, eg: `-pmj-distance` and `-pmj-extent`

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Ahhhhh! Okay, this makes much more sense. Especially now that I know that the distance is centered in the region, too.

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Thank you for the clarification! Now I understand the intended usage much better (fix `-extent` and adjust `-distance` when fine-tuning).

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: > this way, we would save the centerline and the mask in a tmp directory?

I think that sound reasonable! :smile: And maybe move the file-saving over to the part of the PR where QC is generated, just to make it clearer that they only need to be generated if QC is requested.

jcohenadad (approved):
Amazing contribution @sandrinebedard !!

spinalcordtoolbox/scripts/sct_process_segmentation.py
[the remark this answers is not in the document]
sandrinebedard: I added the descriptions in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/3dd2374d161ffad8b3a3aa1805baa2e6fd7524c1

testing/api/test_aggregate_slicewise.py
sandrinebedard: I added a few tests in:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/cd420b4ca247115a37212f50eda794af0ed77476, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/d151a92858f42fcbbf6acd40cc4f9b0807f6f70d, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/0e32c430aed61ef4b54b80d9e0d511fd0b42e7e6

I want to add the method in `batch_processing.sh` with the related tests, but I can also address it in an other PR

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: I changed the flags for `-pmj-distance` and `-pmj-extent` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/d3d6c3e3c6a26df276786056fe0e730867474df9

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/e17592974a7def4671cd353e666ac91a5c8fb1cc

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3478/commits/32e3140cafffec2d72127c074c8986b6e180cf8d

joshuacwnewton (approved):
I have a few nitpicky suggestions, but for the most part, this LGTM! 

I'd say feel free to merge after this! :slightly_smiling_face:

PR #3138, opened 2020-12-29
Add a `-k` flag to keep existing installation files, rather than cleaning them

kousu:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

    Add a -c (clean) flag.
    
    This replaces -b (clean binaries) and -d (clean data) with a single -b flag, and expands
    it to cover cleaning miniconda and the deepseg models.
    With this, iterating on sct is a lot easier because the installer
    doesn't take 3-10 minutes every time.
    
    It should be safe to expand this to cover miniconda (in most case):
    pip handles dependency versioning fine, and is smart enough to recognize
    what's changed if anything has changed.
    
    It is unsafe to use it with binaries/data/deepseg models because we do
    not have those versioned. But it was already unsafe to use -b and -d in
    anyway for exactly that reason; most of our users will be installing SCT to
    fresh directories ~/sct_$VERSION each time anyway, so I think the tradeoff
    is worth it.

## Linked issues

Depends on #3102 being merged first.

PR #3480, opened 2021-07-20
Skip redownloading `sct_testing_data/` if the directory already exists

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Currently, in pytest, we redownload the test data on every run to ensure that we're working with a fresh copy of the test data. This is wasteful and unnecessary when debugging. 

Ideally we would skip this if the existing test data is fine, and in #2959, we outlined some rough plans to use checksums to verify the integrity of the test data. But, #2959 has stalled (likely due to the workload required for it), so in the meantime, this PR adds a simpler way to skip test data downloading.

I'll also document this on the Test data Wiki page.

## Linked issues

N/A, born from Slack message.

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -25,7 +25,7 @@ from spinalcordtoolbox.scripts import sct_download_data as downloader
 logger = logging.getLogger(__name__)
 
 
-def pytest_sessionstart(session):
+def pytest_sessionstart():
     """Perform actions that must be done prior to test collection."""
     # Use a non-interactive backend so that no GUI plots will interrupt the test suite.
     # (NB: We do this here to ensure it is set before `matplotlib` is first imported.)
@@ -33,16 +33,9 @@ def pytest_sessionstart(session):
         os.environ["MPLBACKEND"] = 'Agg'
 
     # Download sct_testing_data prior to test collection
-    if not session.config.getoption('--skip-data-dl'):
+    if not os.path.exists(sct_test_path()):
         logger.info("Downloading sct test data")
         downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
-
-
-def pytest_addoption(parser):
-    """Pytest hook to add custom command line option(s)."""
-    parser.addoption(
-        "--skip-data-dl", action="store_true", help="Opt out of redownloading sct_testing_data when running tests."
-    )
 
 
 @pytest.fixture
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -25,7 +25,7 @@ from spinalcordtoolbox.scripts import sct_download_data as downloader
 logger = logging.getLogger(__name__)
 
 
-def pytest_sessionstart():
+def pytest_sessionstart(session):
     """Perform actions that must be done prior to test collection."""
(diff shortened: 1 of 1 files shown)


kousu (changes requested):
We had simpler solution for this but we seem to have forgotten it in the shuffle!

In my PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3152/files#diff-a31c7ed5d35f5ed8233994868c54d625b18e6bacb6794344c4531e62bd9dde59R147

```
    if not os.path.exists(sct_test_path()):
        downloader.main(['-d', 'sct_testing_data', '-o', sct_test_path()])
```

but that PR was cancelled because it was also trying to deal with that weird `pytest-xdist` bug and it got pretty complicated.

Also in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2903 which added this `downloader.main` line in the first place, @Drulex rightly [said](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2903#issuecomment-696935732)

> Ideally we should add some caching/integrity checking in the download function, but in the meantime I am assuming that people who really care about this (developers) can simply delete the folder if they want to test with "fresh" data.

and he had the identical check in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2903/commits/ae3cf882b14cc61018a8c2cad9c96591243a1cfc

but it was removed before the PR was committed

https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2903/commits/a349f504c005976b2fc19b58a833424ac4dccef0

which...smells like it was also trying to work around the `xdist` bug.

so I think the workaround fix should be to add that `if not os.path.exists(sct_test_path()):` line back.

kousu:
This sort of fix thing is exactly what #2669 will solve. By making everything into versioned packages, we can let `pip` take care of figuring out if it needs to reinstall the data, instead of doing it unconditionally. We'd do something like:

```
pip install -e .[tests]
```

which would then via

```
# setup.py
extras_require={
  'tests': ["spinalcordtoolbox-data-test"],
},
```

go and get the latest test data.

kousu:
This is a similar idea I had: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3138. It was originally about making `./install_sct` work sort of like `pip install -U` will eventually work: install over top of an existing install, but leave all the pre-existing dataset dependencies in place. @jcohenadad didn't like the risk of mixing up datasets (because we *don't* currently version them) so I inverted it into an option `./install_sct -C`. It didn't cover testing data though.

I think #2669 is the good long term solution that covers both cases.

kousu (approved):

Drulex (approved):
LGTM. FYI at some point I also solved this a bit differently via https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2958 which gave more control.

kousu:
Huh https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2958/ looks like the first commit on this branch: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3480/commits/9e2ebe7c4fa803a50b2005d816e2bd62e9924403

I like the "just delete it" solution better, because every added option is an extra piece of vocabulary to learn.

PR #3499, opened 2021-08-06
Introduce -uthr (upper threshold) to complement -thr (lower threshold) in sct_maths

### LINKED ISSUE #3319: Introduce `-uthr` (upper threshold) to complement `-thr` (lower threshold)
`fslmaths` has both `-thr` and `-uthr`, with sample usage:

```bash
# Thresholding with FSL
fslmaths t1_seg_labeled.nii.gz -thr 3 -uthr 6 -bin t1_seg_labeled_c3c6.nii.gz
```

Output:

image

`sct_maths` currently only has "-thr" for lower threshold, why not introducing "uthr" for upper threshold. 

(A user has requested this feature in [this forum post](https://forum.spinalcordmri.org/t/generate-spinal-cord-mask-over-specific-region-i-e-levels-of-the-spinal-cord/673/5?u=jcohenadad).)

Drulex:
…) #3319

#### GitHub

#### PR contents

## Description
Add `-uthr` for upper threshold in `sct_maths` to complement `-thr` (lower threshold). Add corresponding unit test.

## Linked issues
closes #3319

### DIFF
--- a/testing/api/test_math.py
+++ b/testing/api/test_math.py
@@ -7,7 +7,7 @@ import datetime
 
 import numpy as np
 
-import spinalcordtoolbox as sct
+import spinalcordtoolbox.math as sct_math
 from spinalcordtoolbox.testing.create_test_data import dummy_blob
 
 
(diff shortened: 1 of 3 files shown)


Drulex:
FYI the `sct_maths` code uses the `math` API as if the `data` param is passed by value, but in fact it is passed by reference since it is a mutable object (`np.array`).

For example:
```python
[...]
elif arguments.thr is not None or arguments.uthr is not None:
        data_out = sct_math.threshold(data, arguments.thr, arguments.uthr)
```
^ there is no need to do `data_out = ...` because `data` is modified directly inside the `threshold()` function since it is passed by reference.

I didn't change anything since it's like this everywhere in `math`, but it may be worth it to clean up the code and remove the return values so that it is more intuitive. (or make copies everywhere, but I don't think it's necessary).

joshuacwnewton:
> FYI the sct_maths code uses the math API as if the data param is passed by value, but in fact it is passed by reference since it is a mutable object (np.array).

Good catch!

> it may be worth it to clean up the code and remove the return values so that it is more intuitive. (or make copies everywhere, but I don't think it's necessary).

I agree that it would be good to change this. I've opened an issue for this here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3500

testing/api/test_math.py
```diff
@@ -7,7 +7,7 @@
 
 import numpy as np
 
-import spinalcordtoolbox as sct
+import spinalcordtoolbox.math as sct_math
```
joshuacwnewton, line 10: Thanks for fixing this other bug!

It must have been silently there this whole time, but the tests never failed because `math` must be imported somewhere earlier in the test suite.

joshuacwnewton (approved):

Issue #3564, opened 2021-11-03
WSL builds failing with `OSError: [Errno 40] Too many levels of symbolic links` for ncurses-6.3


joshuacwnewton:
I've hit "Re-run jobs" a few times and no dice: https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4087550544?check_suite_focus=true#step:6:340

I'll try to dig in deeper as soon as I can.

joshuacwnewton:
ncurses+miniconda might be relevant here, see also: https://github.com/rstudio/reticulate/issues/982

joshuacwnewton:
[Diffing the logs](https://www.diffchecker.com/utCMgTTf) against the last working CI run, I see some key differences:

```diff
# OS Type changed
- fv-az30-432 
+ fv-az71-870

# ncurses-6.3 now gets installed via conda instead of ncurses-6.2
- ncurses            pkgs/main/linux-64::ncurses-6.2-he6710b0_1
+ ncurses            pkgs/main/linux-64::ncurses-6.3-h7f8727e_0
```

Since this change is pretty recent (<24hr), this is probably an upstream breaking change of some kind. Still digging.

joshuacwnewton:
Possible solution: https://github.com/conda/conda/issues/10333#issuecomment-919739093

> I changed the installation path from /mnt/d/... to ~/..., and it works now

Since our current install path path is `/mnt/d/a/spinalcordtoolbox/spinalcordtoolbox/`, this looks possibly applicable.

kousu:
We don't even _use_ curses, why is it there?

It's not installed by default by conda:

```
(base) [kousu@requiem ~]$ conda create -n whatever 
Collecting package metadata (current_repodata.json): done
Solving environment: done


==> WARNING: A newer version of conda exists. <==
  current version: 4.9.2
  latest version: 4.10.3

Please update conda by running

    $ conda update -n base -c defaults conda



## Package Plan ##

  environment location: /home/kousu/src/neuropoly/spinalcordtoolbox/python/envs/whatever



Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate whatever
#
# To deactivate an active environment, use
#
#     $ conda deactivate

(base) [kousu@requiem ~]$ conda list whatever
# packages in environment at /home/kousu/src/neuropoly/spinalcordtoolbox/python:
#
[...]

PR #3567, opened 2021-11-05
Update install_sct

kousu:
re: Maxime's install problems on WSL.

I want to see how CI's WSL handles this.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -594,7 +594,7 @@ esac
 run bash "$TMP_DIR/miniconda.sh" -p "$SCT_DIR/$PYTHON_DIR" -b -f
 
 # create conda environment
-python/bin/conda create -y -n venv_sct python=3.7
+python/bin/conda create -y -n venv_sct python=3.7.11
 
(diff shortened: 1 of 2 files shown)


install_sct
```diff
@@ -591,7 +591,7 @@ osx)
 esac
 
 # Run conda installer
-run bash "$TMP_DIR/miniconda.sh" -p "$SCT_DIR/$PYTHON_DIR" -b -f
+run bash "$TMP_DIR/miniconda.sh" -p "./$PYTHON_DIR" -b -f
 
 # create conda environment
 python/bin/conda create -y -n venv_sct python=3.7.11
```
joshuacwnewton, line 597: Yeah, #3564 looks like it will get in the way of debugging Maxime's issue, because the `symbolic links` error happens prior to the rest of the install. 

I'd maybe try pinning `ncurses=6.2` when making the conda environment (at least temporarily)? 

```suggestion
python/bin/conda create -y -n venv_sct python=3.7.11 ncurses=6.2
```

joshuacwnewton:
We can close this now that #3574 has been merged.

PR #3568, opened 2021-11-05
Revert TensorFlow/Keras/Python upgrade to retain support for M1 Macs

joshuacwnewton:
#### GitHub

- [ ] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR reverts:

* #3361
* #3561

This brings us back to TensorFlow 1.5, which restores compatibility with AVX-less CPUs. 

However, this will also re-open #3227, so this shouldn't be considered a long-term solution. Ideally, we would find a way to upgrade Tensorflow/Python _and_ retain support for M1 macs.

## Linked issues

This is a temporary solution to #3566, so I believe that issue should stay open even if this is merged.

PR #3569, opened 2021-11-07
Check CPU attributes to determine whether to install an AVX-less version of TensorFlow (M1 Macs, older Linux CPUs)

### LINKED ISSUE #3566: TensorFlow 1.15 requires AVX support, which is incompatible with M1 Macs
In #3361, we upgraded Tensorflow from 1.5 to 1.15. This was necessary to move away from Python 3.6, but doing so introduced a new requirement: Users' CPUs must support the AVX instruction set. 

At the time this was acceptable to us, because the vast majority of CPUs produced after 2011 support AVX. However, M1 Macs [do not support AVX](https://www.anaconda.com/blog/apple-silicon-transition):

> For the last 10 years, Intel and AMD have rolled out support for “vector instructions,” which operate on multiple pieces of data simultaneously. Specifically, these additions to the x86-64 instruction set are AVX, AVX2, and AVX-512 (which itself has different variants). As you might imagine, these instructions can be handy when working with array data, and several libraries have adopted them when compiling binaries for x86-64. **The problem is that Rosetta2 does not support any of the AVX family of instructions and will produce an illegal instruction error if your binary tries to use them.** Because of the existence of different Intel CPUs with different AVX capabilities, many programs can already dynamically select between AVX, AVX2, AVX-512, and non-AVX code paths, detecting the CPU capabilities at runtime. These programs will work just fine under Rosetta2 because the CPU capabilities reported to processes running under Rosetta2 do not include AVX. **However, suppose an application or library does not have the capability to pick non-AVX code paths at runtime.** In that case, it will not work under Rosetta2 if the packager assumed the CPU would have AVX support when they compiled the library. **TensorFlow wheels, for example, do not work under Rosetta2 on the M1.**

This means that, unless we revert back to Tensorflow v1.5, SCT's `sct_deepseg_<x>` functions will not be compatible with M1 Macs.

* Related: #3046

joshuacwnewton:
#### GitHub

#### PR contents

## Description

**Context**: If the PyPI version of TensorFlow 1.15 is installed on an M1 Mac, or a Linux machine with an older CPU, the installation will fail. (This is because the PyPI version of Tensorflow 1.15 is dependent on the "AVX instruction set", which M1 Macs and older CPUs do not support.) To fix this, it's possible to install a version of TensorFlow compiled to _not_ use AVX instructions.

This PR provides 2 improvements on the previous approach for installing AVX-less TensorFlow:

* **Older Linux CPUs**: Before, we checked to see if an error was thrown when trying to import Keras/Tensorflow. Instead, this PR checks the CPU attributes directly. (Reason: "Illegal instruction" exceptions were difficult to catch using bash conditionals.)
* **M1 Macs**: This PR adds new support for M1 Macs.

## Linked issues

Fixes #3566
Closes https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3568.


install_sct
```diff
@@ -641,19 +641,16 @@ fi
 
 
 ## Check if installed TF/Keras requirements throw instruction set errors
-if python3 -c "import keras; print(keras.__version__)" 2>&1 | grep -q 'The TensorFlow library was compiled to use'; then
+if python3 -c "import keras; print(keras.__version__)" 2>&1 | grep -q 'The TensorFlow library was compiled to use\|Illegal instruction'; then
```
joshuacwnewton, line 644: This conditional [doesn't seem to trigger](https://neuropoly.slack.com/archives/CAW9X21D5/p1636243175074100) on @mathieuboudreau's M1 Mac, and I can't figure out why.

This is the testing I did for this conditional back in #3361:

```console
joshua@tadpole:~$ qemu-x86_64 --cpu Westmere-v1 /home/joshua/repos/spinalcordtoolbox_test-python-3.7-tensorflow-1.15/python/envs/venv_sct/bin/python3 -c "import keras; print(keras.__version__)" 2>&1 |       grep 'The TensorFlow library was compiled to use\|Illegal instruction'
2021-11-07 09:56:12.161788: F tensorflow/core/platform/cpu_feature_guard.cc:37] The TensorFlow library was compiled to use AVX instructions, but these aren't available on your machine.
```

My best guess is that  `python3 -c "import keras; print(keras.__version__)"` behaves differently when called on an M1 Mac compared to calling it via `qemu-x86_64`. Maybe it triggers some sort of crash that prevents the output from being captured?

joshuacwnewton: @kousu Is there a better way to perform this check? (Shot in the dark, but... would using a subprocess help here? I'm not really sure how to isolate the Python call.)

joshuacwnewton: @mathieuboudreau: Could you share the output when you run the following commands on your M1 Mac?

```
# Without subshell
$SCT_DIR/python/envs/venv_sct/bin/python3 -c "import keras; print(keras.__version__)" 2>&1 | grep 'The TensorFlow library was compiled to use\|Illegal instruction' || echo "Illegal instruction not matched" 

# With subshell (not sure if this makes any difference)
echo $($SCT_DIR/python/envs/venv_sct/bin/python3 -c "import keras; print(keras.__version__)" 2>&1 | grep 'The TensorFlow library was compiled to use\|Illegal instruction' || echo "Illegal instruction not matched")
```

mathieuboudreau:
```shell
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ # Without subshell
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ $SCT_DIR/python/envs/venv_sct/bin/python3 -c "import keras; print(keras.__version__)" 2>&1 | grep 'The TensorFlow library was compiled to use\|Illegal instruction' || echo "Illegal instruction not matched" 

Illegal instruction not matched
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ 
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ # With subshell (not sure if this makes any difference)
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ echo $($SCT_DIR/python/envs/venv_sct/bin/python3 -c "import keras; print(keras.__version__)" 2>&1 | grep 'The TensorFlow library was compiled to use\|Illegal instruction' || echo "Illegal instruction not matched")
Illegal instruction not matched
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ 
```

kousu: I think the problem is you need to use `egrep` (or `grep -E`). `|` is a regexp character and `grep` doesn't do those by default.

You could also split it up like

```
if ( (python3 -c "import keras; print(keras.__version__)" 2>&1 | grep -q 'The TensorFlow library was compiled to use) || (python3 -c "import keras; print(keras.__version__)" 2>&1 | grep -q 'Illegal instruction') ); then 
```

joshuacwnewton: > I think the problem is you need to use `egrep` (or `grep -E`). `|` is a regexp character and `grep` doesn't do those by default.

I had been following [this blog post](https://www.thegeekstuff.com/2011/10/grep-or-and-not-operators/) which suggests that escaping `|` (i.e. `\|`) allows it to be used with regular ol' `grep`.

mathieuboudreau: @joshuacwnewton let me know when you want me to test it again!

kousu: macs might have BSD grep, not Linux grep? It's a different dialect.

https://ss64.com/osx/grep.html says macOS has `grep -E` so..that should work?

mathieuboudreau: I'm willing to schedule a videocall if you want to tinker with it.

joshuacwnewton: Aside from the grep issues, we can still quickly test the wheel itself manually for now. :smile: 

@mathieuboudreau Could you try the following steps and see if it fixes the issue?

```bash
conda activate venv_sct
pip uninstall tensorflow
pip install https://github.com/spinalcordtoolbox/docker-tensorflow-builder/releases/download/v1.15.5-py3.7/tensorflow-1.15.0-cp37-cp37m-macos_arm64.whl
sct_check_dependencies
```

If all goes well, you shouldn't need to reinstall SCT from scratch -- you should just be able to immediately run `sct_check_dependencies` without error.

mathieuboudreau: @joshuacwnewton I get an error, see below:

```shell

(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ source ${SCT_DIR}/python/etc/profile.d/conda.sh
(base) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ conda activate venv_sct
(venv_sct) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ pip uninstall tensorflow
Found existing installation: tensorflow 1.15.5
Uninstalling tensorflow-1.15.5:
  Would remove:
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/estimator_ckpt_converter
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/freeze_graph
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/saved_model_cli
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/tensorboard
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/tf_upgrade_v2
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/tflite_convert
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/toco
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/bin/toco_from_protos
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/site-packages/tensorflow-1.15.5.dist-info/*
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/site-packages/tensorflow/*
    /Users/sct_m1/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/site-packages/tensorflow_core/*
Proceed (y/n)? y
  Successfully uninstalled tensorflow-1.15.5
(venv_sct) Mathieus-MacBook-Pro:spinalcordtoolbox sct_m1$ pip install https://github.com/spinalcordtoolbox/docker-tensorflow-builder/releases/download/v1.15.5-py3.7/tensorflow-1.15.0-cp37-cp37m-macos_arm64.whl
ERROR: tensorflow-1.15.0-cp37-cp37m-macos_arm64.whl is not a supported wheel on this platform.

```

kousu: :( drat. That seemed so promising.

That thread *said* they'd got it working on an M1, right?

mathieuboudreau: Which thread? Is it possible that I need to be on the most recent OS for it to work?

mathieuboudreau: I'm still on Big Sur.

mathieuboudreau: Looks good! 🚀
Screen Shot 2021-11-09 at 11 00 58 AM

Now I'm hitting an FSLeyes error, it seems it hasn't been installed. Was I supposed to install it first? I know there's a separate hack needed to get FSLeyes working on recent Mac OS's.

joshuacwnewton: That's not an error, thankfully. FSLeyes is an optional dependency.

But, maybe the wording needs to be tweaked to make it a little friendlier. It does look like an error, doesn't it? :sweat_smile:

mathieuboudreau:
Yup haha. Also the fact that's the last test made me think that the tests stopped after it "failed", maybe having an additional message afteewards would help know when the tests are officially all done.

Télécharger Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

install_sct
joshuacwnewton: OK, I think the reason we couldn't catch the error is because we're piping the output of `python3 -c` into `grep`. But, the `python3` command naturally fails (due to the illegal instruction), _and_ we have `pipefail` set in `install_sct`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7fd2ea718751dd858840c3823c0830a910d9777c/install_sct#L36

So, I think the command we actually want to use is:

```bash
if grep -q 'The TensorFlow library was compiled to use\|Illegal instruction' <<< "$(python3 -c "import keras; print(keras.__version__)" 2>&1)"; then
```

Using a subshell + a [here string](https://tldp.org/LDP/abs/html/x17837.html) should let us more reliably catch the output of the check... I hope? But, @kousu is a more skilled bash wizard than I, so... maybe I'm missing something here. :P

----

@mathieuboudreau Sorry to bother again, but could you try installing SCT from this branch and see if you run into the error again?

joshuacwnewton: Scratch that -- @valosekj gave this a try, and this solution still didn't catch the error. :thinking: 

image

mathieuboudreau: Sorry, I missed this in my emails! Feel free to ping me on Slack next time if it's urgent!

install_sct
```diff
@@ -175,6 +175,26 @@ function fetch_os_type() {
...
+    ARCH='x32'
+  elif [[ "$uname_arch" = arm* ]]; then
+    ARCH='a32'
+  elif test "$uname_arch" = aarch64; then
+    ARCH='a64'
+  else
+    exit 1
+  fi
+}
+
```
joshuacwnewton, line 197: Oops, please ignore this. I committed this accidentally. :sweat: 

Removing in a follow-up commit...

install_sct
joshuacwnewton: I think this has been resolved now by 978d1cd. :slightly_smiling_face:

mathieuboudreau: Want a fresh test?

joshuacwnewton: That would be wonderful, thank you!

mathieuboudreau: Looks good on my end! 

Flawless victory <3 
Screen Shot 2021-11-30 at 2 48 44 PM

valosekj: Installation finished successfully for me as well! On both M1 MacBook Air and M1 Mac mini. Thanks a lot @joshuacwnewton!

jcohenadad (approved):

PR #3573, opened 2021-11-11
Fix multiple WSL issues related to installation and display

### LINKED ISSUE #3501: Error is thrown for WSL during matplotlib GUI check, but isn't caught by try/except
Currently, in `sct_check_dependencies`, we try to catch exceptions raised when opening a plot with matplotlib:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65c0b8be3b1be19fbb901be817a198f08a23c8c7/spinalcordtoolbox/scripts/sct_check_dependencies.py#L322-L335

By catching exceptions this way, we can simply print `FAIL` rather than erroring out (that way the installer will successfully conclude without an "Installation failed!" message).

However, a user on the forum reported this installation failure:

```
Check if figure can be opened with matplotlib.......qt.qpa.screen: QXcbConnection: Could not connect to display 127.0.0.1:0.0
Could not connect to any X display.

Installation validation Failed!


Installation failed!

Please copy the output of this Terminal (starting with the command install_sct) and upload it as a .txt attachment in a new topic on SCT's forum:

```

Questions:

* Why wasn't the error caught by our try/except?
* Why is this error being thrown in the first place? (Especially given the changes in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3420/.)

### LINKED ISSUE #3564: WSL builds failing with `OSError: [Errno 40] Too many levels of symbolic links` for ncurses-6.3

### LINKED ISSUE #3571: Error installing SCT on WSL2 - Miniconda related
### Description

When running the `./install_sct` script on WSL I ran into the systematic issue of conda failing to get anything.

At the end I pinpointed the issue to this [issue](https://github.com/conda/conda/issues/9948) from conda itself.
The current workaround I adopted (one suggested in https://github.com/conda/conda/issues/9948#issuecomment-655087415) is to download an old version of `miniconda` and then run the `conda update` command. This worked for me.

### Steps to Reproduce the error

0. Install SCT <5.4>
```./install_sct```

**Installation interrupted:** 

An HTTP error occurred when trying to retrieve this URL. HTTP errors are often intermittent, and a simple retry will get you on your way. If your current network has https://www.anaconda.com blocked, please file a support request with your network engineering team. 'https://repo.anaconda.com/pkgs/main/linux-64'

### Steps to fix the error

Modified the `install_sct ` script in the **Dowload Miniconda section** (lines 583-600 approx).
Replaced :
` download "$TMP_DIR/"miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh`
by this line:
`download "$TMP_DIR/"miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-4.7.12.1-Linux-x86_64.sh`

And, after the installation (of miniconda) I added this line:
`python/bin/conda update conda`

Here's my install_sct_mod [script](https://gist.github.com/sercharpak/bd3c329671c2deecdee8d6e3489f6019)

### LINKED ISSUE #3572: WSL2 - GUI - FSLeyes
**Add additional instructions for WSL2 & Fsleyes.**

When updating from WSL1 to WSL2 the GUI management becomes a little bit problematic.

After installing `fsleyes `as recommended (for instance, through _conda-forge_) the following error appears:

```
libGL error: No matching fbConfigs or visuals found
The program 'fsleyes' received an X Window System error.
This probably reflects a bug in the program.
The error was 'GLXBadContext'.
  (Details: serial 8748 error_code 154 request_code 149 minor_code 5)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

Found a workaround following this [issue](https://github.com/microsoft/WSL/issues/5535) and setting the shortcut of xlaunch as indicated in this [tutorial](https://www.digitalcitizen.life/shortcut-arguments-parameters-windows).

You could summarize the fix with:

1. Adding: 
```
export DISPLAY=$(awk '/nameserver / {print $2; exit}' /etc/resolv.conf 2>/dev/null):0
export LIBGL_ALWAYS_INDIRECT=0
```
to `~/.bashrc`.

2. Source the file:
```
source ~/.bashrc
```

3. Setting up your firewall to accept xlaunch  following this [guide](https://github.com/cascadium/wsl-windows-toolbar-launcher#firewall-rules)

4. When opening xlaunch, select the tick for and un-select the tick for 

a. 
image
b. 
image
c.
image

### PR DESCRIPTION
**Pull request by joshuacwnewton**
#### GitHub

#### PR contents

## Description

This PR fixes a number of WSL-related issues discovered over the past week or so:

* 458226b5bebb266a67050a49e479d110915a793b: This commit rewrites the WSL instructions so that they address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3572 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3501. 
    * Note: I've also tested these steps for both WSL1 and WSL2, and both SCT and FSLeyes worked perfectly for me.
* a477fc6894e73ae2f102aebe7f82d7a9e60ce8f9: This commit fixes an issue specific to our WSL runners (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3564). 
    * By working inside `~`, rather than GitHub Action's default working directory (`/mnt/d/a/spinalcordtoolbox/spinalcordtoolbox`), we avoid the Windows part of the filesystem. (I'm not sure if there's a cleaner way to do this -- maybe by supplying `path: /root` to `actions/checkout@v2`?)

You can see this branch working in our test suite for WSL in https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/1446773364.

## Linked issues

Fixes #3501.
Fixes #3564.
Fixes #3572.
Closes #3567.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -305,14 +305,21 @@ jobs:
         # NB: mesa-utils is needed for PyQT testing (https://github.com/Microsoft/WSL/issues/1246#issuecomment-356425862)
         sudo apt update && sudo DEBIAN_FRONTEND=noninteractive apt install -y gcc git curl mesa-utils
     - uses: actions/checkout@v2
+    - name: Copy SCT from $GITHUB_WORKSPACE to ~
+      run: |
+        cd ..
+        cp -r spinalcordtoolbox ~/spinalcordtoolbox
     - name: Install SCT
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -i
     - name: Check dependencies
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -c
     - name: Run pytest test suite
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -t
 
   windows-wsl-ubuntu-20_04:
@@ -346,12 +353,19 @@ jobs:
         # NB: mesa-utils is needed for PyQT testing (https://github.com/Microsoft/WSL/issues/1246#issuecomment-356425862)
         sudo apt update && sudo DEBIAN_FRONTEND=noninteractive apt install -y gcc git curl mesa-utils
     - uses: actions/checkout@v2
+    - name: Copy SCT from $GITHUB_WORKSPACE to ~
+      run: |
+        cd ..
+        cp -r spinalcordtoolbox ~/spinalcordtoolbox
     - name: Install SCT
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -i
     - name: Check dependencies
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -c
     - name: Run pytest test suite
       run: |
+        cd ~/spinalcordtoolbox
         ./.ci.sh -t
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -356,7 +356,7 @@ jobs:
     - name: Copy SCT from $GITHUB_WORKSPACE to ~
       run: |
         cd ..
-        cp -r spinalcordtoolbox ~/spinalcordtoolbox
+        cp -rp spinalcordtoolbox ~/spinalcordtoolbox
     - name: Install SCT
       run: |
         cd ~/spinalcordtoolbox
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -38,7 +38,7 @@ Basic installation (No GUI)
 
    - After this command finishes, you will be prompted to restart your computer.
 
-   - After restarting, the installation should automatically resume, and you will be able to create an user account inside Ubuntu by selecting a username and password.
+   - After restarting, the installation should automatically resume, and you will be able to create a user account inside Ubuntu by selecting a username and password.
 
 #. Decide between WSL Version 1 or WSL Version 2.
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -162,7 +162,7 @@ If you would like to use SCT's GUI features, or if you would like to try FSLeyes
 
       source ${SCT_DIR}/python/etc/profile.d/conda.sh
       conda create -c conda-forge -p ~/fsleyes_env fsleyes -y
-      ln -s ~/fsleyes_env/bin/fsleyes /usr/local/bin/fsleyes
+      sudo ln -s ~/fsleyes_env/bin/fsleyes /usr/local/bin/fsleyes
 
    These instructions will install FSLeyes into a fresh ``conda`` environment, then create a link to FSLeyes so that you can use the ``fsleyes`` command without having to activate the conda environment each time.
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -171,11 +171,10 @@ Option 2: Install with Docker
 Basic Installation (No GUI)
 ***************************
 
-First, `install Docker <https://docs.docker.com/install/>`_. Then, follow the examples below to create an OS-specific SCT installation.
+First, `install Docker <https://docs.docker.com/install/>`_. Then, follow either of the examples below to create an OS-specific SCT installation.
 
-
-Docker Image: Ubuntu
-^^^^^^^^^^^^^^^^^^^^
+Option 1: Ubuntu Docker Image
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code:: bash
 
@@ -198,8 +197,8 @@ Docker Image: Ubuntu
    docker ps -a  # list all containers
    docker commit <CONTAINER_ID> <YOUR_NAME>/ubuntu:ubuntu16.04
 
-Docker Image: CentOS7
-^^^^^^^^^^^^^^^^^^^^^
+Option 2: CentOS7 Docker Image
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code:: bash
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -51,6 +51,7 @@ Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it pos
 
    .. code-block:: sh
 
+      cd ~
       git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git
       cd spinalcordtoolbox
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -34,7 +34,7 @@ Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it pos
 
       wsl --set-version Ubuntu 1
 
-    You can then run ``wsl --list --verbose`` again to check that the distribution has changed from WSL2 to WSL1.
+   You can then run ``wsl --list --verbose`` again to check that the distribution has changed from WSL2 to WSL1.
 
 #. Environment preparation
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -24,7 +24,7 @@ Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it pos
 
    Because of this, your best guarantee for a stable installation of SCT is to convert the WSL2 Ubuntu environment to use WSL1 before proceeding:
 
-   .. code:
+   .. code::
 
       wsl --set-version Ubuntu 1
 
--- a/documentation/source/user_section/installation/windows.rst
+++ b/documentation/source/user_section/installation/windows.rst
@@ -14,9 +14,19 @@ Basic installation (No GUI)
 
 Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it possible to run native Linux programs, such as SCT.
 
-#. Make sure that your version of Windows is up to date before continuing.
+#. Make sure that your version of Windows 10 is up to date before continuing.
 
-#. Install Windows Subsystem for Linux (WSL)
+   - In Windows, search for "System Information" and open the app. In the "Version" field, make sure that you are running "Build 19041" or higher.
+
+   - Then, search for "Powershell" in your Start Menu, then right-click and "Run as administrator". Then run the following command:
+
+     .. code::
+
+        wsl --update
+
+   - If this command is successful, then you can proceed to the next step. Otherwise, please update your version of Windows and try again. Additionally, please make sure that you have sufficient administrative privileges for your Windows system.
+
+#. Installing an Ubuntu distribution in Windows Subsystem for Linux (WSL)
 
    - In Windows, search for "Powershell" in your Start Menu, then right-click and "Run as administrator".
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: documentation/source/user_section/installation/windows.rst [hunk 1]
```diff
@@ -4,60 +4,38 @@
 Installation for Windows
 ************************
```
joshuacwnewton, line 6: ## Discussion: SCT Course

This PR needs to be merged and included in a stable release for it to be available for our users prior to the SCT Course. (This means v5.5). 

However, that would mean creating a new release with Python 3.7, which runs right into the M1 Mac issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3566). Now, we have a almost-working fix for that too (in #3569), but the `grep` issue in that PR has yet to be addressed.

So... as far as making sure SCT is usable for both Windows and Mac M1 users... we could either:

* Backport these fixes to SCT v5.4, quietly, amend that release, and stick with Python 3.6/Tensorflow 1.5.
* Try to get #3569 ready and merged, then create a new release ASAP.

jcohenadad: > Backport these fixes to SCT v5.4, quietly amend that release, and stick with Python 3.6/Tensorflow 1.5.

I would feel more at ease with this solution because *think* it will create the least collateral damage few days before the SCT course. The risk with going with a new version (option 2) is that it has not been thoroughly tested, and there is much more actions to do (more stressful, more prone to errors).

documentation/source/user_section/installation/windows.rst
```diff
@@ -4,60 +4,55 @@
...
 #. Install Windows Subsystem for Linux (WSL)
 
-   - Install `Xming <https://sourceforge.net/projects/xming/>`_.
+   - Follow the instructions on Microsoft's `Install WSL <https://docs.microsoft.com/en-us/windows/wsl/install>`_ page.
+
+#. Decide between WSL Version 1 or WSL Version 2.
+
+   By default, Microsoft's instructions will create an Ubuntu environment using Version 2 of WSL. While version 2 has been tested to work with SCT, our development team tests more thoroughly using Version 1 of WSL (due to `better support from GitHub Actions <https://github.com/actions/virtual-environments/issues/50>`_).
+
+   Because of this, your best guarantee for a stable installation of SCT is to convert the WSL2 Ubuntu environment to use WSL1 before proceeding. To do this, first close Ubuntu, then run the following command in a Windows command prompt:
```
jcohenadad, line 25: "your best guarantee for a stable installation of SCT is to convert the WSL2 Ubuntu environment to use WSL1 before proceeding."

i find this sentence difficult to understand (as an ESL 😅 ), and the person I guided through the installation also find it confusing (ie: does it mean that we should use WSL2 or WSL1?)

joshuacwnewton: Thanks for the feedback! I'll try to simplify the language.

install_sct
```diff
@@ -591,11 +591,21 @@ osx)
...
+# Syntax explanation: https://stackoverflow.com/a/6085237
+find "$SCT_DIR/$PYTHON_DIR" -type f -exec touch {} +
+
 # create conda environment
+print info "Creating conda environment..."
 python/bin/conda create -y -n venv_sct python=3.7
 
+# Reapply the touch fix to avoid connection issues on WSL1
+find "$SCT_DIR/$PYTHON_DIR" -type f -exec touch {} +
+
```
joshuacwnewton, line 608: Based on discussion in our last SCT Course meeting, I think the least invasive method would be to split the "touch" lines into a separate PR, and only backport these changes. 

(That's because the rest of the fixes are for the CI runners (not necessary to backport, since our CI runs off `master`) and documentation (not necessary to backport, since our docs are pointed at `master`).)

That would let us keep the changes to `5.4` to a minimum.

kousu:
I think you can use some Actions syntax to apply to the entire script instead of repeating `cd ~/spinalcordtoolbox` or spending time and risking permission glitches copying the repo from one place to another.

I patched this in directly and now let's keep an eye on this run to see how it handles: https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4184366172?check_suite_focus=true

joshuacwnewton:
I tried this too, but sadly I got the same error as you did. 😓 

> Error: The directory name is invalid.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4184366172?check_suite_focus=true#step:3:11

install_sct
joshuacwnewton: 5.4 has been updated to include the fix from #3574.

See:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/commits/5.4

documentation/source/user_section/installation/windows.rst
```diff
@@ -4,60 +4,82 @@
...
 
    .. code-block:: sh
 
-      git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git sct
-      cd sct
+      cd ~
+      git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git
+      cd spinalcordtoolbox
 
    To select a `specific release <https://github.com/spinalcordtoolbox/spinalcordtoolbox/releases>`_, replace X.Y.Z below with the proper release number. If you prefer to use the development version, you can skip this step.
```
jcohenadad, line 84: I know it's not part of this PR, but do we want to remove the "Z" now that we decided to remove patches? (and maybe give an example because some people might not understand what we mean by X.Y., eg: 5.4)

documentation/source/user_section/installation/windows.rst
```diff
@@ -86,18 +117,57 @@ Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it pos
...
+   .. code::
+
+      source ${SCT_DIR}/python/etc/profile.d/conda.sh
+      conda create -c conda-forge -p ~/fsleyes_env fsleyes -y
+      sudo ln -s ~/fsleyes_env/bin/fsleyes /usr/local/bin/fsleyes
+
+   These instructions will install FSLeyes into a fresh ``conda`` environment, then create a link to FSLeyes so that you can use the ``fsleyes`` command without having to activate the conda environment each time.
+
+
 Option 2: Install with Docker
```
jcohenadad, line 166: Should we move this section outside of the "windows" section, given that this is OS-agnostic?

documentation/source/user_section/installation/windows.rst
```diff
@@ -86,18 +117,57 @@ Windows Subsystem for Linux (WSL) is available on Windows 10 and it makes it pos
...
 
-In the context of SCT, it can be used:
-
-- To run SCT on Windows, until SCT can run natively there
-- For development testing of SCT, faster than running a full-fledged
-  virtual machine
-- <your reason here>
-
 Basic Installation (No GUI)
 ***************************
```
jcohenadad, line 173: should we say something like "Below we show you two examples of Docker config (pick one)".

to make sure people won't actually do both sequentially

documentation/source/user_section/installation/windows.rst
jcohenadad: Thinking about it: given that we are pressed with time, feel free to ignore this comment for now. We can reorganize the Docker section later in time. The important thing for now is to have the WSL instructions.

kousu:
> I tried this too previously (with `/root` instead of `~/`, because `~` evaluates to `/root`), but sadly I got the same error as you did. sweat
> 
> > Error: The directory name is invalid.
> 
> See: [3de2eba](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/3de2ebaaf32a57b724350b36c53dd8f5c0a8b987) + https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4172032301?check_suite_focus=true

Okay okay hang on i have another idea one sec

documentation/source/user_section/installation/windows.rst
joshuacwnewton: > Thinking about it: given that we are pressed with time, feel free to ignore this comment for now. We can reorganize the Docker section later in time. The important thing for now is to have the WSL instructions.

I agree -- the Docker section needs a lot more work and testing than we likely have time for right now.

.github/workflows/tests.yml
```diff
@@ -346,12 +353,19 @@ jobs:
         # NB: mesa-utils is needed for PyQT testing (https://github.com/Microsoft/WSL/issues/1246#issuecomment-356425862)
         sudo apt update && sudo DEBIAN_FRONTEND=noninteractive apt install -y gcc git curl mesa-utils
     - uses: actions/checkout@v2
+    - name: Copy SCT from $GITHUB_WORKSPACE to ~
+      run: |
+        cd ..
+        cp -r spinalcordtoolbox ~/spinalcordtoolbox
```
kousu, line 359: If we don't do

```suggestion
        cp -rp spinalcordtoolbox ~/spinalcordtoolbox
```

there will be even more mysterious permission errors down the road.

kousu (changes requested):

.github/workflows/tests.yml
joshuacwnewton: Thank you for the input re: bash wizardry

.github/workflows/tests.yml
```diff
@@ -305,14 +305,21 @@ jobs:
...
```
kousu, line 311: So, the reason you can't `defaults: working-directory: ~/` is because `working-directory:` is applied by the runner itself, before it shells out to the shell (defined by `defaults: shell:` or just the system default), and that means that in this case, you need to use Windows syntax, and can only pick Windows folders: `defaults: working-directory: C:\Users\runneradmin`.

And `actions/checkout` too: it's unaware that we've installed WSL and, being a self-contained "Action", uses its own scripts and shells, so it can *only* check out into Windows folders.

Hence to get this to work we would have to:

- a: patch `Vampire/setup-wsl` to add a `wsl-working-directory:` option to instruct `wsl-bash` to `cd` a second time once it the Github Runner launches it
- b. replace `actions/shell` with a direct `git clone` (run under `wsl-bash`)

In light of all that, your workarounds seem like the best option unfortunately.

I wish we had a good issue we could at least link to explain these odd lines, Maybe the writeup I just did at https://github.com/Vampire/setup-wsl/issues/12 but I don't know if it's clear enough.

kousu (approved):

.github/workflows/tests.yml
joshuacwnewton: Thank you for digging into this and debugging how the working directory is set. When I first started fiddling with this, `setup-wsl` felt like an opaque thing to me, so I'm glad we have a better understanding of how things work now.

joshuacwnewton:
I think this is good to go (and if any additional feedback comes in from testing over the weekend, we can always follow this PR up another time).

@jcohenadad After this, I'm happy to direct my focus towards the slides + the sct_tutorial_data changes, so feel free to ping me over there if you need input / reviews on anything. :slightly_smiling_face:

.github/workflows/tests.yml
kousu: https://github.com/Vampire/setup-wsl/discussions/13#discussioncomment-1641892 says we should be able to add

```
shell: wsl-bash
wsl-shell-command: bash -c "cd && bash --noprofile --norc -euo pipefail '{0}'"
```

to get it to behave; and we don't need to set `working-directory` or patch it to add `wsl-working-directory`, it'll just be in `$HOME` by default.

and also that my attempts to use `working-directory` *should* have worked, I was just using the wrong path (I don't think I was but I can check again).

That doesn't deal with `actions/checkout`, though.

Issue #3583, opened 2021-11-16
Support side-by-side versions


kousu:
From @valosekj during the 2021 SCT course:

**Feature justification: Is your feature request related to a problem? Please describe.**

> What is the best way how to work with more SCT versions on single machine? Currently, I have one stable version (installed e.g. in ~/sct_v5.3.0) which I use for processing of some longitudinal data and second version (installed e.g. in ~/sct_latest) which is up-to-date with GitHub master version. I have then symbolic link pointing to one of the version (sct -> ~/sct_latest or sct -> ~/sct_v5.3.0). When I want to switch from one version to another, I have to either change the symbolic link or directly modify export PATH="/usr/local/lib/sct/bin:$PATH" in my .zshrc file. Is there any more elegant workaround? 

**Describe the solution you'd like**

We're currently internally using a `conda` env; we're sort of overriding it by editing the user's `$PATH`

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b7e747993cd6fc38d272bd8599834389d4934979/install_sct#L260-L276

which means you can only have one version available.

I believe users who need it should be able to directly use the existing conda envs:

```
source ~/sct_v5.3.0/python/bin/activate && conda activate venv_sct
```

```
source ~/sct_latest/python/bin/activate && conda activate venv_sct
```

but I'm not sure if there might be conflicts from the top-level `$PATH` edit we also do.

Can we make this more elegant for users?

**Describe alternatives you've considered**

If we get pip working (#1526) then it would be relatively simple for users to use standard tools (`python -m venv`, `virtualenv`, `conda env create`) to make isolated versions.

kousu:
Also, for completeness, the current recommended way to do this is to manually edit your `~/.zshrc` or `~/.bashrc` as needed. e.g. to choose the `master` copy:

```diff
--- ~/.bashrc
+++ ~/.bashrc
@@ -121,11 +121,11 @@
 
 # SPINALCORDTOOLBOX (installed on 2021-11-16 11:18:38)
 export PATH="/home/GRAMES.POLYMTL.CA/u108545/sct_5.4/bin:$PATH"
 export SCT_DIR=/home/GRAMES.POLYMTL.CA/u108545/sct_5.4
 
 
 
-# SPINALCORDTOOLBOX (installed on 2021-11-16 11:34:46)
[...]

PR #3586, opened 2021-11-16
Generate issue log automatically, then direct users to share the log when posting on the forum

kousu:
Many people post install errors on the forum without all the details
attached. This logs *the entire install log* to a file for them,
which should make getting errors reliably easier.

This kind of a silly solution but I think it will help.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -120,8 +120,8 @@ function finish() {
     echo ""
   else
     print error "Installation failed!\n
-Please find the file $(pwd)/install_sct_log.txt and \n
-upload it as a .txt attachment in a new topic on SCT's forum:\n
+Please find the file $(pwd)/install_sct_log.txt, 
+then upload it as a .txt attachment in a new topic on SCT's forum:
 --> http://forum.spinalcordmri.org/c/sct"
   fi
   # clean tmp_dir
--- a/install_sct
+++ b/install_sct
@@ -115,7 +115,6 @@ function finish() {
   cd "$SCT_SOURCE"
   if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
-    rm -r install_sct_log.txt  # if there's no errors, there's no need to keep this log around
   elif [[ "$value" -eq 99 ]]; then
     # Showing usage with -h
     echo ""
--- a/install_sct
+++ b/install_sct
@@ -29,6 +29,8 @@
 # Authors: PO Quirion, J Cohen-Adad, J Carretero
 # License: see the file LICENSE.TXT
 
+( # | tee install_sct_log.txt
+
 # stricter shell mode
 # https://sipb.mit.edu/doc/safe-shell/
 set -eo pipefail  # exit if non-zero error is encountered (even in a pipeline)
@@ -113,13 +115,14 @@ function finish() {
   cd "$SCT_SOURCE"
   if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
+    rm -r install_sct_log.txt  # if there's no errors, there's no need to keep this log around
   elif [[ "$value" -eq 99 ]]; then
     # Showing usage with -h
     echo ""
   else
     print error "Installation failed!\n
-Please copy the output of this Terminal (starting with the command install_sct) and upload it as a .txt attachment in a new topic on SCT's forum:\n
---> http://forum.spinalcordmri.org/c/sct"
+Please find the file $(pwd)/install_sct_log.txt and \n
+attach it to a new issue in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/new so we can help you."
   fi
   # clean tmp_dir
   rm -rf "$TMP_DIR"
@@ -757,3 +760,5 @@ else
     die "Installation validation Failed!"
   fi
 fi
+
+) 2>&1 | tee install_sct_log.txt
--- a/install_sct
+++ b/install_sct
@@ -121,7 +121,8 @@ function finish() {
   else
     print error "Installation failed!\n
 Please find the file $(pwd)/install_sct_log.txt and \n
-attach it to a new issue in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/new so we can help you."
+upload it as a .txt attachment in a new topic on SCT's forum:\n
+--> http://forum.spinalcordmri.org/c/sct"
   fi
   # clean tmp_dir
   rm -rf "$TMP_DIR"


install_sct
```diff
@@ -113,13 +115,13 @@ function finish() {
   cd "$SCT_SOURCE"
   if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
+    rm -r install_sct_log.txt  # if there's no errors, there's no need to keep this log around
```
kousu, line 118: Is this a good idea? Maybe we should just leave it around in all cases, that's a little more consistent.

install_sct
```diff
@@ -113,13 +115,14 @@ function finish() {
...
   elif [[ "$value" -eq 99 ]]; then
     # Showing usage with -h
     echo ""
   else
     print error "Installation failed!\n
-Please copy the output of this Terminal (starting with the command install_sct) and upload it as a .txt attachment in a new topic on SCT's forum:\n
---> http://forum.spinalcordmri.org/c/sct"
+Please find the file $(pwd)/install_sct_log.txt and \n
+attach it to a new issue in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/new so we can help you."
```
joshuacwnewton, line 125: This bumps into the question of "What do we want to use issues for, and what do we want to use the forum for?"

I think currently, it's something like:

* GitHub: All dev activity.
* Forum: All user activity.

But the proposed is:

* GitHub: Installation issues, dev activity
* Forum: Usage issues/questions

I _think_ I like the proposed split? It basically means that GitHub becomes "the place where broken things get reported and fixed", and removes some clutter from the forum, letting it stay focused on: "OK, I have a working installation. Now, how do I use SCT effectively?"

install_sct
jcohenadad: i don't see it as a problem to leave it around indeed. More pros than cons

install_sct
jcohenadad: hum... on the other hand, we are dealing with users who can easily get overwhelmed with +1 platform to deal with.

kousu: I figured this line would be controversial :P. No worries, we can back it out.

Before we do though, can we take a survey? Of people who've had installation problems in the last year or so, people whose first post(s) were about it:

* https://forum.spinalcordmri.org/u/aran/activity
* https://forum.spinalcordmri.org/u/davidcadotte/activity
* https://forum.spinalcordmri.org/u/annekbaker/activity
* https://forum.spinalcordmri.org/u/abdullah_althubaiti/activity
* https://forum.spinalcordmri.org/u/alexander_zhang/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/b_zeinali/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/salmanshaz/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/hussnain/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/vb922/activity (installation are their only posts)
* https://forum.spinalcordmri.org/u/mahdi/activity
* https://forum.spinalcordmri.org/u/deng/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/jetpack/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/willk/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/pelegrini/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/gmotzespina/activity
* https://forum.spinalcordmri.org/u/gali/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/sahar/activity
* https://forum.spinalcordmri.org/u/idris/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/mvaro2002/activity
* https://forum.spinalcordmri.org/u/felixi/activity
* https://forum.spinalcordmri.org/u/jtalbott/activity
* https://forum.spinalcordmri.org/u/amijas/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/deshengkong/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/spine/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/brma/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/marco/activity
* https://forum.spinalcordmri.org/u/georgebouz/activity (installation is their only post)
* https://forum.spinalcordmri.org/u/miki/activity (installation is their only post; but deviating from the normal bug report, [this](https://forum.spinalcordmri.org/t/install-previous-versions/240) is actually #3583, but I'd still want that in the bug tracker)
* https://forum.spinalcordmri.org/u/petermultivac/activity


Of them, people whose weren't:

* https://forum.spinalcordmri.org/u/rosella_tro/activity
* https://forum.spinalcordmri.org/u/zmzforever/activity
* https://forum.spinalcordmri.org/u/kim/activity (however, their *second* post was, and their first was about a different code issue)
* https://forum.spinalcordmri.org/u/chang/activity
* https://forum.spinalcordmri.org/u/hanna/activity
* https://forum.spinalcordmri.org/u/weizx/activity
* https://forum.spinalcordmri.org/u/daniel_papp/activity
* https://forum.spinalcordmri.org/u/ihattan/activity
* https://forum.spinalcordmri.org/u/merve/activity

I think this shows that people are generally _creating_ accounts to report bugs. Even amongst the people who weren't, a couple only logged back in after a couple of years because they hit a problem while upgrading SCT.

A couple of the issues were user oversights like forgetting `./` in `./install_sct`, but most of them were legitimate issues. Even the user oversights are bug-ish since they're opportunities to clarify our documentation. So, I hate to be "that person" because I'm not a huge fan of Github in principle, but my thinking was, if they need to create an account anyway, it might as well be here with the rest of the bugs.

kousu: A different approach entirely: we could (with consent) automatically upload the failed installation logs somewhere, and have a bot post them to Github for us with a `triage` tag.

But then I guess we leave the user without any recourse to tech support :thinking:

jcohenadad: i do see the value of removing the 'noise' in the forum. On the other hand, I'm afraid to see the other effect: topics that should be in the forum (ie: useful for the community) be put on GH. 

having the clear distinction: GH is for the devs | Forum is for users, makes it less confusing

install_sct
```diff
@@ -757,3 +760,5 @@ else
     die "Installation validation Failed!"
   fi
 fi
+
+) 2>&1 | tee install_sct_log.txt
```
joshuacwnewton, line 764: Just to make sure I'm understanding correctly... from [What is the purpose of 'tee'?](https://superuser.com/a/1356843)

> `do_something | tee -a logfile` puts the output into logfile and to stdout, while `do_something >> logfile` puts it only into the logfile.
>
> The purpose of tee is to produce a one input, multiple output scenario - just like in a 'T' crossing.

So, we're using `tee` to print the output of `install_sct` to stdout _and_ the logfile. 

But... does this print output all at once, or does it print output bit by bit as the script runs in the subshell? Testing locally... it works as intended, so... I'm guessing I don't understand pipes as well as I thought I did, aha. :blush:

install_sct
joshuacwnewton: FWIW, I think it's probably safe enough to stick with what we're doing anyway, rather than rock the boat (especially for a PR that's mainly about generating a log file.)

So, I'm :+1: on not changing anything here.

install_sct
kousu: Well it's sort of up to `tee` but as far as I know `tee` moves data as fast as it gets it.

It's possible a crash _could_ leave some data inside of `tee`'s internal buffers before getting flushed to the logfile/stdout, but the window is very short. Maybe a system under heavy load would be prone to that.

install_sct
```diff
@@ -118,7 +120,8 @@ function finish() {
     echo ""
   else
     print error "Installation failed!\n
-Please copy the output of this Terminal (starting with the command install_sct) and upload it as a .txt attachment in a new topic on SCT's forum:\n
+Please find the file $(pwd)/install_sct_log.txt and \n
+upload it as a .txt attachment in a new topic on SCT's forum:\n
```
joshuacwnewton, line 124: Is this how we want it to look?

Screenshot from 2021-11-17 14-01-05

I'm guessing `print` preserves newlines, so we're doubling up on them here.


```suggestion
Please find the file $(pwd)/install_sct_log.txt, 
then upload it as a .txt attachment in a new topic on SCT's forum:
```

How's this? 

Screenshot from 2021-11-17 14-05-32

kousu: Yeah that seems good. Thanks for testing that. I thought the extra newlines were a bit weird in the old version but I didn't want to think about futzing with them...

jcohenadad (approved):
👍

install_sct
```diff
@@ -757,3 +760,5 @@ else
     die "Installation validation Failed!"
   fi
 fi
+
```
joshuacwnewton, line 763: One last thought before we merge -- would there be a easy way to get `set -x` functionality in the install log, but not in the stdout? (I'm guessing that would require a more complicated solution than `tee`, aha)

I ask because often when debugging a user's issue I feel like I want them to re-run with verbose mode on, but that can be a bit of a hassle. It would be nice if we could have access to that by default.

kousu: Maybe with `| egrep -v '^\+ '`? Maybe with, uh, there's some syntax like `exec 4<>/path/to/fifo` that opens temporary extra streams you can use besides stdout and stderr. But either seems fragile.

joshuacwnewton: Probably not a huge deal, then. :stuck_out_tongue_closed_eyes:

Issue #3587, opened 2021-11-16
matplotlib.backends.backend_qt4agg removed


kousu:
### Description

matplotlib 3.5 was released 16 hours ago, and (https://matplotlib.org/3.5.0/api/prev_api_changes/api_changes_3.5.0.html#modules)

> The Qt4-based backends, qt4agg and qt4cairo, have been removed. Qt4 has reached its end-of-life in 2015 and there are no releases of either PyQt4 or PySide for recent versions of Python. Please use one of the Qt5 or Qt6 backends.

and this is crashing our tests and blocking CI.

### Steps to Reproduce

0. Install SCT from `master`
1. Run the tests

**Expected behavior:** the tests should pass 

**Actual behavior:**

https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4231487281?check_suite_focus=true#step:7:28

```
==================================== ERRORS ====================================
___________________ ERROR collecting testing/api/test_gui.py ___________________
ImportError while importing test module '/__w/spinalcordtoolbox/spinalcordtoolbox/testing/api/test_gui.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
python/envs/venv_sct/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
testing/api/test_gui.py:6: in <module>
    from spinalcordtoolbox.gui import base, centerline, sagittal
spinalcordtoolbox/gui/centerline.py:15: in <module>
    from spinalcordtoolbox.gui import widgets
spinalcordtoolbox/gui/widgets.py:10: in <module>
    from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg as FigureCanvas
E   ModuleNotFoundError: No module named 'matplotlib.backends.backend_qt4agg'
=========================== short test summary info ============================
ERROR testing/api/test_gui.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
======================== 25 warnings, 1 error in 5.72s =========================
Error: Process completed with exit code 2.
```

kousu:
Question for myself: why are we touching `matplotlib.backends` directly at all? Isn't that supposed to be hidden behind `matplotlib.use()`?

kousu:
More direct repro:

```
./install_sct -y
source python/bin/activate && conda activate venv_sct
[...]

PR #3588, opened 2021-11-16
Upgrade to Qt5.

kousu:
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3587

### DIFF
--- a/spinalcordtoolbox/gui/widgets.py
+++ b/spinalcordtoolbox/gui/widgets.py
@@ -7,8 +7,8 @@
 import logging
 from time import time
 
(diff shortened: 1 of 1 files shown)


kousu:
Not sure if this is enough? It gets the tests passing, anyway, but it seemed _too_ easy.

joshuacwnewton (approved):
I did some searching, and found this issue: https://github.com/matplotlib/matplotlib/issues/6209/#issue-142961227

The discussion in that issue seems to suggest that Qt4 was actually using Qt5 functions under the hood, and that the two backends overlap almost entirely. So, I think this change should be A-OK.

kousu:
:pray:

PR #3591, opened 2021-11-17
`install_sct`: Replace `sed -i` with `perl -pi -e`

### LINKED ISSUE #3590: `sed_i` polyfill does not work as expected in `install_sct`
Currently, we use the following polyfill, to make sure `sed -i` (in-place) works across platforms.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c9e4e80128f31755c6c9348e13de90b2364d5aee/install_sct#L330-L338

However, for [a user on the forum](https://forum.spinalcordmri.org/t/installation-failed-in-mac-10-14-6/776), this polyfill doesn't work:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c9e4e80128f31755c6c9348e13de90b2364d5aee/install_sct#L616

Error message:

```
sed: can't read s/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/: No such file or directory
```

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Perl is available in all of our supported OS's, and should work identically. So, this avoids the platform-specific issues that come from trying to use 2 different `sed`s.

## Linked issues

Fixes #3590.
Closes #3592.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -325,18 +325,18 @@ if [ "$(uname)" = "Darwin" ]; then
   (command -v realpath >/dev/null) || realpath() {
     python3 -c 'import sys, os; [print(os.path.realpath(f)) for f in sys.argv[1:]]' "$@"
   }
-
-
-  # glue over the difference between how BSD and GNU sed work with -i
-  # https://code-examples.net/en/q/56e314
-  sed_i() {
-    sed -i '' "$@"
-  }
-else
-  sed_i() {
-    sed -i "$@"
-  }
 fi
+
+sed_i() {
+  # glue over the difference between how BSD and GNU sed work with -i
+  # From https://code-examples.net/en/q/56e314:
+  #   "My recommendation, for what it is worth, is not to update-in-place
+  #    with sed (which has complex failure modes), but to generate new files
+  #    and rename them afterwards. In other words: avoid -i."
+  tmp_file="$TMP_DIR/tmp_file_sed_i"
+  sed "$1" "$2" > "$tmp_file"
+  cp -f "$tmp_file" "$2"
+}
 
 # ======================================================================================================================
 # SCRIPT STARTS HERE
--- a/install_sct
+++ b/install_sct
@@ -327,17 +327,6 @@ if [ "$(uname)" = "Darwin" ]; then
   }
 fi
 
-sed_i() {
-  # glue over the difference between how BSD and GNU sed work with -i
-  # From https://code-examples.net/en/q/56e314:
-  #   "My recommendation, for what it is worth, is not to update-in-place
-  #    with sed (which has complex failure modes), but to generate new files
-  #    and rename them afterwards. In other words: avoid -i."
-  tmp_file="$TMP_DIR/tmp_file_sed_i"
-  sed "$1" "$2" > "$tmp_file"
-  cp -f "$tmp_file" "$2"
-}
-
 # ======================================================================================================================
 # SCRIPT STARTS HERE
 # ======================================================================================================================
@@ -613,7 +602,7 @@ find "$SCT_DIR/$PYTHON_DIR" -type f -exec touch {} +
 # * https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3200
 # this needs to be added very early in python's boot process
 # so using sitecustomize.py or even just appending to the file are impossible.
-sed_i 's/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/' "$SCT_DIR/$PYTHON_DIR/envs/venv_sct/lib/python"*"/site.py"
+perl -pi -e 's/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/' "$SCT_DIR/$PYTHON_DIR/envs/venv_sct/lib/python"*"/site.py"
 
 # activate miniconda
 # shellcheck disable=SC1091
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
I've included two potential fixes based on discussion from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3590:

* 88c2d27 
* And https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3591/files (i.e. 1cad3c0)

kousu:
Ah the perl option!

I think if this has reasonably wide support it's better than #3592 because it's less code!

joshuacwnewton:
> Ah the perl option!
> 
> I think if this has reasonably wide support it's better than #3592 because it's less code!

Here's an in-depth explanation of the options, too: https://www.linuxquestions.org/questions/programming-9/perl-replace-text-in-file-476382/#post2390981

I think it's cute that it spells out `-pie`, to boot. :blush:

kousu:
By wide support I mean: macOS comes with perl. Debian and Ubuntu too. Unsure about CentOS though. And the WSL distros though Ubuntu-based might not have the full complement of Ubuntu tools by default.

If CI passes on all platforms then I think it's safe: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/1472992331

joshuacwnewton:
> If CI passes on all platforms then I think it's safe: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/1472992331

LGTM!

kousu (approved):
let's do it

install_sct
```diff
@@ -325,17 +325,6 @@ if [ "$(uname)" = "Darwin" ]; then
   (command -v realpath >/dev/null) || realpath() {
     python3 -c 'import sys, os; [print(os.path.realpath(f)) for f in sys.argv[1:]]' "$@"
   }
-
```
joshuacwnewton, line 328: Bizarrely, `test_batch_processing.sh` is now failing:

```python
___________ test_batch_processing_results[mt/mtr_in_wm.csv-0-MAP()] ____________
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/batch_processing/test_batch_processing.py", line 42, in test_batch_processing_results
    assert (get_csv_float_value(csv_output, row, column) == pytest.approx(  # Default rel_tolerance: 1e-6
AssertionError: assert 54.42425821653468 == 54.37973987894469 ± 5.4e-05
  +54.42425821653468
  -54.37973987894469 ± 5.4e-05
___________ test_batch_processing_results[dmri/fa_in_cst.csv-0-WA()] ___________
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/batch_processing/test_batch_processing.py", line 42, in test_batch_processing_results
    assert (get_csv_float_value(csv_output, row, column) == pytest.approx(  # Default rel_tolerance: 1e-6
AssertionError: assert 0.780605541771859 == 0.7793999429926204 ± 7.8e-07
  +0.780605541771859
  -0.7793999429926204 ± 7.8e-07
___________ test_batch_processing_results[dmri/fa_in_cst.csv-1-WA()] ___________
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/batch_processing/test_batch_processing.py", line 42, in test_batch_processing_results
    assert (get_csv_float_value(csv_output, row, column) == pytest.approx(  # Default rel_tolerance: 1e-6
AssertionError: assert 0.77461183127549 == 0.7755359256476565 ± 7.8e-07
  +0.77461183127549
  -0.7755359256476565 ± 7.8e-07
```

So, we're seeing meaningfully different batch processing results (despite this branch [having passed a few days ago](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4241779744?check_suite_focus=true)). 

I assume this is an upstream change, and unrelated to this PR, but I'll investigate.

joshuacwnewton: My best guess is that this maybe a... one-off glitch? Currently, `test_batch_processing.sh` is [passing on `master`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4255751692?check_suite_focus=true), too. 

So, the changes from switching `pip` versions shouldn't have affected anything...

joshuacwnewton: Still, the fact that a one-off glitch could affect batch processing results is quite curious to me... 

I tried [diffing the `batch_processing.sh` logs](https://www.diffchecker.com/9jqmFWwV), and the only meaningful change I noticed was [in the T2 registation](https://www.diffchecker.com/3Ih4BiJj). But, the T2 CSA results were unaffected! It was the MT and dMRI results that changed. So... I'm at a bit of a loss.

I think the best course of action would be to wait and see if [re-running the processing](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4267235151?check_suite_focus=true) fixes things, and just... moving on? :thinking:

joshuacwnewton: Yep! Restarting the job fixed it. How odd.

jcohenadad: > But, the T2 CSA results were unaffected! 

indeed, the T2 CSA does not use the PAM50 template registration (it only uses the segmentation and the vertebral levels-- which in this case match that one from the registered PAM50_levels)

one thing concerning is that the straightening produced different results:
image

that would explain the PAM50 registration difference (and different MT and DTI results)-- this is strange... i'm pretty sure there is no stochastic computation for the straightening

joshuacwnewton: I'm glad that the restarted job shows precisely the same results as before, at least. :thinking:

joshuacwnewton: Or... not? It [happened again](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4267499139?check_suite_focus=true) after merging master once more. This is quite odd.

joshuacwnewton: My best guess is that the error is unrelated to the changes in this PR, and we happened to get unlucky with a broken VM that failed in the same way twice in a row, with all other VMs functioning normally.

I'm going to leave any follow-up investigations for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597, in case this pops up again in the future.

joshuacwnewton:
Merging this PR because @kousu was able to reproduce the `batch_processing.sh` issue [on `master`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597#issuecomment-974586455), and we can follow up on that issue in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597.

Issue #3443, opened 2021-07-07
Remove mirror neuro.polymtl.ca


alexfoias:
Remove mirror neuro.polymtl.ca because our dokuwiki migrated to gitbook.io

taowa:
I'd like to try to *replace* neuro.polymtl.ca and not remove it outright. OSF is the only other mirror for at least a couple models, and I think that we should try to make them available ourselves. I outlined one of our options for doing that on `#website`, but am open to others.

joshuacwnewton:
> I'd like to try to _replace_ neuro.polymtl.ca and not remove it outright. OSF is the only other mirror for at least a couple models, and I think that we should try to make them available ourselves. I outlined one of our options for doing that on `#website`, but am open to others.

The mirrors are mainly relevant for SCT's custom download manager, `sct_download_data`. But, just to mention one alternative, @kousu has also been taking a stab at hosting/versioning/distributing the models via `pip` as a more long-term solution. It's a little nonstandard, but would solve a big packaging blocker for #1526. 

(There's further context in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2669, https://github.com/spinalcordtoolbox/spinalcordtoolbox-data-template/pull/1.)

kousu:
Thanks for poking me about that @joshuacwnewton.

And there's another problem: the backup mirror doesn't seem to be working? There's been a few reports on the forum like https://forum.spinalcordmri.org/t/installation-failed-at-the-end/496:

> Exception: (‘Download error’, [SSLError(MaxRetryError(‘HTTPSConnectionPool(host=‘osf.io’, port=443): Max retries exceeded with url: /mka78/?action=download (Caused by SSLError(SSLError(“bad handshake: SysCallError(-1, ‘Unexpected EOF’)”,),))’,),), HTTPError(‘403 Client Error: Forbidden for url: https://www.neuro.polymtl.ca/_media/downloads/sct/20200420_sct_binaries_linux.tar.gz 1’,)])

It seems to be mixing up the two domains somehow, but only when it has to fall back. If that's true, it means www.neuro.polymtl.ca/_media/downloads hasn't been used in months, maybe years. I haven't debugged this. I tried to reproduce it a bit but just didn't finish before going to do something else.

joshuacwnewton:
> And there's another problem: the backup mirror doesn't seem to be working?

[...]

Issue #3333, opened 2021-04-06
Remove `gm_model`


kousu:
`sct_download_data -d gm_model` used to be used by `sct_segment_graymatter` but that was removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1622 / https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2549. As part of #2669 I discovered this one hadn't even been ported to https://github.com/sct-data, and I would rather not resurrect it for nothing, so remove the remaining references to it.

PR #3596, opened 2021-11-18
Update outdated mirror links

alexfoias:
* Model/datasets were migrated from the `sct-data` org to the `spinalcordtoolbox` org
* The dokuwiki migrated so we have to remove the `neuro.polymtl.ca` mirrors.
* SCT Course datasets are now managed in [`sct_tutorial_data`](https://github.com/spinalcordtoolbox/sct_tutorial_data) repo -- these can be used to replace some of the dokuwiki links.

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3443


spinalcordtoolbox/scripts/sct_download_data.py
```diff
@@ -25,75 +25,61 @@
...
-        "https://www.neuro.polymtl.ca/_media/downloads/sct/20180612_sct_course-paris18.zip",
     ],
     "course_london19": [
         "https://osf.io/4q3u7/?action=download",
-        "https://www.neuro.polymtl.ca/_media/downloads/sct/20190121_sct_course-london19.zip",
     ],
     "course_beijing19": [
         "https://osf.io/ef4xz/?action=download",
-        "https://www.neuro.polymtl.ca/_media/downloads/sct/20190802_sct_course-beijing19.zip",
     ],
```
joshuacwnewton, line 67: Some of these datasets now only have a single mirror. Would it be worthwhile to create GitHub repos (or at least, `.zip` downloads) for everything here?

* Binaries: Maybe we could put the binary-generating code into a dedicated repo (currently it lurks in the `/install` folder) then attach the generated binaries as `.zip` files to releases on that repo.
* Old course datasets: These can be added to `sct_tutorial_data` quite easily.
* `gm_model`: Doesn't this project already have a corresponding repo somewhere?
* `exvivo_template`: Do we want to upload this to OSF?

joshuacwnewton (commented):
Thanks for noticing this!

spinalcordtoolbox/scripts/sct_download_data.py
joshuacwnewton: > Binaries: Maybe we could put the binary-generating code into a dedicated repo (currently it lurks in the /install folder) then attach the generated binaries as .zip files to releases on that repo.

It looks like @kousu had this idea before me in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2668.

--> @kousu how critical would you say it is to have 2 mirrors for our binaries? If so, do you have any interest in returning to those draft PRs from last year that stalled out?

kousu: Yes of course I have an interest in returning to them.

joshuacwnewton: I guess since the mirrors are already broken, this PR doesn't have to wait on those changes, though -- we'll have 1 mirror either way. :stuck_out_tongue_closed_eyes:

joshuacwnewton (approved):
I think the remaining suggestions in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3596#discussion_r753706703 can be addressed by smaller follow-up PRs, just to keep things moving. :slightly_smiling_face:

spinalcordtoolbox/scripts/sct_download_data.py
joshuacwnewton: > * [ ] `gm_model`: Doesn't this project already have a corresponding repo somewhere? Maybe we could create a release from that project to store this model.

Oh! @kousu noticed this one  too: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3333

PR #3603, opened 2021-11-23
Update help for 2+ labels usage, and change link from iCloud to RTD

### LINKED ISSUE #3601: Update usage
it says: "More information about label creation can be found at https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%%5FCourse%%5F20200121 (p47)"

we should change that

also: 
"                        If you are using more than 2 labels, all discs covering the region of interest should be provided. E.g., if you are interested in levels C2 to C7, then you should provide disc labels 2,3,4,5,6,7. For more information about label creation, please refer to https://www.icloud.com/keynote/0th8lcatyVPkM_W14zpjynr5g#SCT%5FCourse%5F20200121 (p47)
"

### LINKED ISSUE #3602: update usage to inform user that using more than 2 labels will produce weird results
usage says:
~~~
If more than two labels (only with the parameter '-disc') are used, a non-linear registration will be applied to align the each intervertebral disc between the subject and the template, as described in sct_straighten_spinalcord. This the most accurate and preferred method. This feature does not work with the parameter '-ref subject', where only a rigid registration is performed.
~~~

we should add a disclaimer--  something like:
~~~
Due to the non-linear registration in the S-I direction, the warping field will be cropped above the top label and below the bottom label. This will result in a strange-looking registered image that has the same value above the top label and below the bottom label. But if you are not interested in these regions, you do not need to worry about it.
~~~

context:
- https://forum.spinalcordmri.org/t/labels-for-lumbar-spinal-level/737/5?u=jcohenadad

joshuacwnewton:
#### GitHub

#### PR contents

## Description

These changes should hopefully make it a little easier to understand the proper labeling for registration.

## Linked issues

Fixes #3601 
Fixes #3602


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -102,19 +102,21 @@ def get_parser():
...
             "mis-alignment of discs because a subject''s intervertebral discs distance might differ from that of the "
             "template.\n"
             "\n"
-            "If more than two labels (only with the parameter '-disc') are used, a non-linear registration will be "
-            "applied to align the each intervertebral disc between the subject and the template, as described in "
-            "sct_straighten_spinalcord. This the most accurate and preferred method. This feature does not work with "
-            "the parameter '-ref subject', where only a rigid registration is performed.\n"
+            "If more than two labels are used, a non-linear registration will be applied to align the each "
+            "intervertebral disc between the subject and the template, as described in "
+            "sct_straighten_spinalcord. This the most accurate and preferred method. Note: \n"
```
jcohenadad, line 113:
```suggestion
            "sct_straighten_spinalcord. Note: \n"
```
i don't think we want to endorse this as the best method-- there are serious cons associated with it

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -102,19 +102,21 @@ def get_parser():
...
+            "  - Using more than two labels will only work with disc labels (i.e. '-ldisc')\n"
```
jcohenadad, line 114: is that true? i'm wondering why it wouldn't work with -l labels 🤔

joshuacwnewton: Good question -- I was just going off of what was written previously ("`If more than two labels (only with the parameter '-disc') are used`") but I can double-check!

joshuacwnewton: I followed the tutorial steps, swapping out the typical two-label `sct_label_utils` step (`3,9`) with this:

```
sct_label_utils -i t2_seg_labeled.nii.gz -vert-body 3,4,5,6,7,8,9 -o t2_labels_vert.nii.gz
```

When running `sct_register_to_template`, I get this error:

```
ERROR: Label file has 7 label(s). It must contain one or two labels.
```

The error message comes from this part of the codebase:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8985ba391ff149cb7af8a086dd7d0d28be90f010/spinalcordtoolbox/scripts/sct_register_to_template.py#L916-L918

However, since it's just a `printv` statement, the error doesn't actually halt the program... so it keeps going! (Log [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/7591237/sct_register_to_template_with_7_labels.txt).)

Results:

template2anat.nii.gz (overlaid on t2.nii.gz):

template2anat

anat2template.nii.gz (overlaid on PAM50_t2.nii.gz):

anat2template

Apart from the seemingly spurious error message (specifically for `body` labels), it does seem like it works fine? I wonder why that error message was put there in the first place.

joshuacwnewton: The context here is from a commit made 3 years ago: https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/e64928c4f976f18351a67b95333867984533a25e

It looks like we used to only permit exactly 2 labels, then we opened it up to 1 or 2. But we've never seemed to permit 2+ labels... :thinking:

jcohenadad: thank you for these investigations, @joshuacwnewton . I _think_ that this error message was set because the motivation for allowing more than 2 labels was based on older investigations when using multiple disc labels-- given that it was motivated by disc labeling, we did not really consider if this could also apply to vert labels (and since we didn't want to deal with too many things at that time we just decided to say "let's authorize this only for disc labels"). but is looks like we can now remove that limitation

jcohenadad (approved):

Issue #3312, opened 2021-03-24
Force location of data downloading for some datasets


charleygros:
**Feature justification: Is your feature request related to a problem? Please describe.**

When using `sct_download_data`, some data (e.g. `exvivo_template` or deepseg models) should be installed under specific folders, eg `$SCT_DIR/data/`. Here is a full list of our current uses of `sct_download_data` (from https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3310#issuecomment-805972981):

> 1. Binaries -> `bin/`
> 2. Models -> `data/`
> 3. Templates -> `data/`
> 4. Testing/example data -> `./`
> 5. Course data -> `./`

More context [here](https://github.com/neuropoly/spinalcordtoolbox/issues/3280#issuecomment-805395427).

**Describe the solution you'd like**
If `-o` is not specified by the user, we can use the default location for each dataset, that could be integrated in a dictionary, similarly than [here ](https://github.com/neuropoly/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_download_data.py#L24) with the URL.

joshuacwnewton:
This is especially relevant for our test suite, as downloading the test data to the working directory results in many duplicated copies of `sct_testing_data` if you run `pytest` in different locations.

This cropped up in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3436#discussion_r660865597, so there's motivation to fix this before releasing https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373.

Issue #2911, opened 2020-09-23
Tests don't cleanup after themselves


Drulex:
After a successful run of `pytest` I noticed that 172 files were created in my working directory. Not a big deal, but it would be nice to cleanup after running the tests.

joshuacwnewton:
To add to this, it's not just the unit tests but the integration tests as well. (See: the output of `sct_testing_data/` before and after running `sct_testing`.)

At the very least, moving forward I believe that we should recommend that new tests clean up after themselves. I plan to update the Testing page on the dev Wiki, so I'll make a comment here when I do that.

joshuacwnewton:
I wanted to find out which tests were to blame, so I modified the `check_testing_data_integrity` function that @Drulex wrote so that it would run once per test:

<details>
<summary> Source code for modified fixture, click to show </summary>

I had some fun with sets! :D

```python
@pytest.fixture(autouse=True)
def test_data_integrity():
    checksums_before = checksum_walk()  # Run before test executes
    yield
    checksums_after = checksum_walk()  # Run after test executes
    
    # Use only filenames to see which files were added/removed
    files_before, files_after = set(checksums_before.keys()), set(checksums_after.keys())
    files_added = files_after - files_before
    files_removed = files_before - files_after

    # Use filenames AND checksums to see which files have changed
    files_changed = set(checksums_before) - set(checksums_after)

    assert not files_added or files_removed or files_changed, \
        f"Data integrity check: {len(files_added)} ADDED, {len(files_removed)} REMOVED, {len(files_changed)} CHANGED"


def checksum(fname: os.PathLike) -> str:
    with open(fname, 'rb') as f:
        data = f.read()
    return md5(data).hexdigest()


def checksum_walk(input_path="."):
    files_checksums = dict()
    for root, _, files in os.walk(input_path):
        for f in files:
            fname = os.path.join(root, f)
            chksum = checksum(fname)
            files_checksums[fname] = chksum

    return files_checksums
```
</details>

Here are the tests that don't clean up after themselves:

<details>
<summary> Large table hidden, click to show </summary>

|File                                                                                                                            |Test|# Added|
|--------------------------------------------------------------------------------------------------------------------------------|----|------|
|test_aggregate_slicewise.py                                                                                                     |test_save_as_csv|1     |
|test_image.py                                                                                                                   |test_change_orientation|6     |
|test_image.py                                                                                                                   |test_more_change_orientation|7     |
[...]

PR #3605, opened 2021-11-26
Bring back `sct_testing` as a light wrapper for `pytest`

### LINKED ISSUE #3532: Bring back `sct_testing` CLI command to wrap `pytest`
After #3093, SCT's previous custom testing framework `sct_testing` was dropped in favor of `pytest`.

The problem, however, is that the `pytest` command is only available if the SCT venv is active. This is fine for devs, but most users will not use SCT this way -- they will only use the CLI interface. So, for our users' sake, we should bring back the `sct_testing` command, but only as a light wrapper for `pytest`, that way users don't have to touch any virtual environments.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR brings back `sct_testing`, but as a very light wrapper for `pytest`. This lets users run tests locally without having to activate SCT's virtual environment, while also preserving the `sct_testing` command (in case of past scripts, forum posts, etc.).

Additionally, this PR modifies two existing tests to use `tmp_path`, to ensure that no files are created in the working directory. That way, a user running this new `sct_testing` command won't generate files in their working directory beyond `sct_testing_data`. (I verified this using the procedure outlined in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2911#issuecomment-779482628.) 

However, because `sct_testing_data` still gets downloaded to the working directory, we may also want to consider addressing https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3312 in this PR, or soon after this PR is merged. (It depends on how we want to group/scope this changeset.)

## Linked issues

Fixes #3532.


spinalcordtoolbox/scripts/sct_testing.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+import sys
+
+from spinalcordtoolbox import __sct_dir__
+
+if __name__ == "__main__":
+    # Treat `sct_testing` as an alias to `pytest`. If no arguments are passed,
+    # simply point pytest at the repo (to avoid looking for tests in the pwd).
+    sys.exit(pytest.main(sys.argv[1:] if sys.argv[1:] else [__sct_dir__]))
```
joshuacwnewton, line 19: Do we want a proper argparse interface for this?

Right now, the way it works is this:

* Running `sct_testing` with no args will run all the tests.
* Running `sct_testing` with arguments will just pass those args to `pytest`.

My thinking is that in 99% of cases, our users will just run `sct_testing` to run all of the tests. 

But, our old `sct_testing` command had options like `-function sct_deepseg`, etc. I could replicate that functionality too, but it would make this PR a little more complicated.

jcohenadad: i agree, no strong need for argparse

RignonNoel:
I like the idea of the `sct_testing` command. And the PR looks good for me.

However, I'm a bit worried about this question:

> However, because `sct_testing_data` still gets downloaded to the working directory, we may also want to consider addressing #3312 in this PR, or soon after this PR is merged. (It depends on how we want to group/scope this changeset.)

Having `sct_testing` back will make the usage of the command outside the working directory a LOT more easy! The problem of the local copy of `sct_testing_data` will highly increase.

Actually when you run pytest you need to manually put the path to the folder of test, so most of the time you are inside the working directory or really not far away. With the command we will copy the test data everywhere in the host system.

I think it should be part of this PR, without it the users will not know about this problem and we will be responsible on the mess in their file systems...

joshuacwnewton:
> With the `sct_testing` command we will copy the test data everywhere in the host system.
> 
> I think it should be part of this PR, without it the users will not know about this problem and we will be responsible on the mess in their file systems...

I appreciate your opinion here! I agree. And, fixing that `sct_testing_data` issue has been a desire of mine for a long time... I'll get on that immediately! :)

joshuacwnewton:
Re-requesting a review now that #3650 is merged. Running `sct_testing` after re-installing SCT from this branch should no longer pollute the working directory. :tada:

(Famous last words...)

RignonNoel (approved):

RignonNoel:
Everything works smooth on my side when I try your branch! And `sct_testing_data` is under the `data` folder and it feel warm in my heart because of that :p

Little detail: the `.gitignore` contain a reference to `sct_testing_data`, maybe we should remove it at long term, but for the moment I approve like that since most of the people will have this folder from legacy version and if we remove it from `.gitignore` they will not like it.

PR #3638, opened 2022-01-08
Convert input file to float32 before compute MTR

### LINKED ISSUE #3636: Inconsistent MTR values when calculated from short dtype
### Description

> Note: This issue was originally created in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3635#issue-1086726238

using sct_computre_mtr, I have an issue with the results when using short (4) NIFTI images for MT0 and MT1. The values I obtained are inconsistent (very small negatives or very big positives) for many slices but not all slices. At the end, when comparing with manual computing of MTR, a lot of voxels have wrong values.
I changed the type of the MT0/MT1 images from short to float and the result is correct, meaning similar to what I got with manual computing of MTR, and this for all slices and voxels.
Maybe is there a reason for this behavior?

RignonNoel:
#### GitHub

**Contributing documentation link is broken**

#### PR contents

## Description
When we use `sct_compute_mtr`, we have an issue with the results when using non `float32` NIFTI images for MT0 and MT1. The values obtained are inconsistent. We need to convert the type of input to float32 before compute the MTR result.

## Linked issues
Fixes #3636 

## Actual status of this PR

Some things need to be cleaned-up but i need some informations/help to know the way we work here:

 - The testing dataset are dynamically download from [sct_testing_data repository](https://github.com/spinalcordtoolbox/sct_testing_data) but i need to test the script with a int_16 file in place of a float_32. What is the process to add a new file of testing (ie: PR ? Request to the team ?) or do i just convert the file first directly in the test ?
 
  - I don't have the PR labels information since the link for the contribution guideline is broken on severals places: https://www.neuro.polymtl.ca/software/contributing#pr_labels
  
  - The actual `sct_compute_mtr` tests does not check the results. Do you have some tips before i go on it ? I will be able to add them also on the initial test already existing.

### DIFF
--- a/spinalcordtoolbox/qmri/mt.py
+++ b/spinalcordtoolbox/qmri/mt.py
@@ -44,8 +44,14 @@ def compute_mtr(nii_mt1, nii_mt0, threshold_mtr=100):
     :param threshold_mtr: float: value above which number will be clipped
     :return: nii_mtr
     """
+    # Convert input to avoid numerical errors from int16 data
+    # Related issue: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3636
+    nii_mt1.change_type('float32')
+    nii_mt0.change_type('float32')
+
     # Initialize Image object
     nii_mtr = nii_mt1.copy()
+
     # Compute MTR
     nii_mtr.data = divide_after_removing_zero(100 * (nii_mt0.data - nii_mt1.data), nii_mt0.data, threshold_mtr)
     return nii_mtr
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -1,25 +1,12 @@
 import pytest
 import logging
 import numpy
+import nibabel as nib
 
 from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.scripts import sct_compute_mtr
 
 logger = logging.getLogger(__name__)
-
-
-@pytest.mark.sct_testing
-@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
-def test_sct_compute_mtr_no_checks(tmp_path):
-    """Run the CLI script without checking results."""
-    mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
-
-    sct_compute_mtr.main(argv=['-mt0', 'mt/mt0.nii.gz', '-mt1', 'mt/mt1.nii.gz', '-o', mtr_output_path])
-
-    # Comparate ground truth mtr file to new generated one
-    ground_truth_mtr = Image('mt/mtr.nii.gz')
-    output_mtr = Image(mtr_output_path)
-    assert numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data) <= 0.001
 
 
 @pytest.mark.sct_testing
@@ -48,3 +35,47 @@ def test_sct_compute_mtr_with_int16_image_type(tmp_path):
             output_mtr.data[numpy.isfinite(output_mtr.data)])
 
     assert numpy.abs(numpy.average(diff)) <= 0.5
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_mtr_with_dummy_int16_image_type(tmp_path):
+    """Run the CLI script with dummy int_16 image type"""
+
+    mt0_path = str(tmp_path / 'mt0_int16.nii.gz')
+    mt1_path = str(tmp_path / 'mt1_int16.nii.gz')
+    mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
+    mtr_truth_path = str(tmp_path / 'mtr_dummy_truth.nii.gz')
+
+    # Generate dummy files of inputs and output comparison
+    nx, ny, nz = 9, 9, 9  # image dimension
+
+    data = numpy.zeros((nx, ny, nz), dtype=numpy.int16)
+    data[4, 4, 4] = 100
+    affine = numpy.eye(4)
+    mt0 = nib.nifti1.Nifti1Image(data, affine)
+    nib.save(mt0, mt0_path)
+
+    data = numpy.zeros((nx, ny, nz), dtype=numpy.int16)
+    data[4, 4, 4] = 70
+    affine = numpy.eye(4)
+    mt1 = nib.nifti1.Nifti1Image(data, affine)
+    nib.save(mt1, mt1_path)
+
+    data = numpy.zeros((nx, ny, nz), dtype=numpy.float32)
+    data[4, 4, 4] = 30
+    affine = numpy.eye(4)
+    mtr_truth = nib.nifti1.Nifti1Image(data, affine)
+    nib.save(mtr_truth, mtr_truth_path)
+
+    # Compute the MTR
+    sct_compute_mtr.main(argv=['-mt0', mt0_path, '-mt1', mt1_path, '-o', mtr_output_path])
+
+    # Compare the dummy output with the mtr file newly generated
+    truth_mtr = Image(mtr_truth_path)
+    output_mtr = Image(mtr_output_path)
+
+    diff = (truth_mtr.data[numpy.isfinite(truth_mtr.data)] -
+            output_mtr.data[numpy.isfinite(output_mtr.data)])
+
+    assert numpy.abs(numpy.average(diff)) <= 30
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -32,7 +32,7 @@ def test_sct_compute_mtr_with_int16_image_type(tmp_path):
     mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
 
     # Generate int16 test file based on sct_testing_data existing one
-    mt0 = Image('mt/mt0.nii.gz')
+    mt0 = Image('mt/mt0_reg_slicereg_goldstandard.nii.gz')
     mt0.save(mt0_path, dtype='int16')
     mt1 = Image('mt/mt1.nii.gz')
     mt1.save(mt1_path, dtype='int16')
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -43,4 +43,8 @@ def test_sct_compute_mtr_with_int16_image_type(tmp_path):
     ground_truth_mtr = Image('mt/mtr.nii.gz')
     output_mtr = Image(mtr_output_path)
 
-    assert numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data) <= 0.001
+    # NB: numpy.isfinite is used to exclude nan/inf elements from comparison
+    diff = (ground_truth_mtr.data[numpy.isfinite(ground_truth_mtr.data)] -
+            output_mtr.data[numpy.isfinite(output_mtr.data)])
+
+    assert numpy.abs(numpy.average(diff)) <= 0.5
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -1,6 +1,8 @@
 import pytest
 import logging
+import numpy
 
+from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.scripts import sct_compute_mtr
 
 logger = logging.getLogger(__name__)
@@ -8,7 +10,37 @@ logger = logging.getLogger(__name__)
 
 @pytest.mark.sct_testing
 @pytest.mark.usefixtures("run_in_sct_testing_data_dir")
-def test_sct_compute_mtr_no_checks():
-    """Run the CLI script without checking results.
-    TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
-    sct_compute_mtr.main(argv=['-mt0', 'mt/mt0.nii.gz', '-mt1', 'mt/mt1.nii.gz'])
+def test_sct_compute_mtr_no_checks(tmp_path):
+    """Run the CLI script without checking results."""
+    mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
+
+    sct_compute_mtr.main(argv=['-mt0', 'mt/mt0.nii.gz', '-mt1', 'mt/mt1.nii.gz', '-o', mtr_output_path])
+
+    # Comparate ground truth mtr file to new generated one
+    ground_truth_mtr = Image('mt/mtr.nii.gz')
+    output_mtr = Image(mtr_output_path)
+    assert numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data) <= 0.001
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_mtr_with_int16_image_type(tmp_path):
+    """Run the CLI script with int_16 image type"""
+
+    mt0_path = str(tmp_path / 'mt0_int16.nii.gz')
+    mt1_path = str(tmp_path / 'mt1_int16.nii.gz')
+    mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
+
+    # Generate int16 test file based on sct_testing_data existing one
+    mt0 = Image('mt/mt0.nii.gz')
+    mt0.save(mt0_path, dtype='int16')
+    mt1 = Image('mt/mt1.nii.gz')
+    mt1.save(mt1_path, dtype='int16')
+
+    sct_compute_mtr.main(argv=['-mt0', mt0_path, '-mt1', mt1_path, '-o', mtr_output_path])
+
+    # Comparate ground truth mtr file to new generated one
+    ground_truth_mtr = Image('mt/mtr.nii.gz')
+    output_mtr = Image(mtr_output_path)
+
+    assert numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data) <= 0.001
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -48,7 +48,7 @@ def test_sct_compute_mtr_with_dummy_highvalue_int16_data(tmp_path):
     mtr_truth_path = str(tmp_path / 'mtr_dummy_truth.nii.gz')
 
     # Generate dummy files of inputs and output comparison
-    nx, ny, nz = 9, 9, 9  # image dimension # center location
+    nx, ny, nz = 9, 9, 9  # image dimension
 
     data = numpy.zeros((nx, ny, nz), dtype=numpy.int16)
     data[4, 4, 4] = 1000
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -36,8 +36,6 @@ def test_sct_compute_mtr_results_are_identical(tmp_path):
                              output_mtr.data[numpy.isfinite(output_mtr.data)])
 
 
-@pytest.mark.sct_testing
-@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
 def test_sct_compute_mtr_with_dummy_highvalue_int16_data(tmp_path):
     """
     Test that high-valued int16 data doesn't result in clipping. See also:
--- a/testing/cli/test_cli_sct_compute_mtr.py
+++ b/testing/cli/test_cli_sct_compute_mtr.py
@@ -11,30 +11,29 @@ logger = logging.getLogger(__name__)
 
 @pytest.mark.sct_testing
 @pytest.mark.usefixtures("run_in_sct_testing_data_dir")
-def test_sct_compute_mtr_with_int16_image_type(tmp_path):
-    """Run the CLI script with int_16 image type"""
+def test_sct_compute_mtr_results_are_identical(tmp_path):
+    """
+    Test the CLI script to ensure that MTR computed with sct_testing_data
+    images remains the same.
+    """
 
-    mt0_path = str(tmp_path / 'mt0_int16.nii.gz')
-    mt1_path = str(tmp_path / 'mt1_int16.nii.gz')
+    mt0_path = 'mt/mt0_reg_slicereg_goldstandard.nii.gz'
+    mt1_path = 'mt/mt1.nii.gz'
+    mtr_groundtruth_path = 'mt/mtr.nii.gz'
     mtr_output_path = str(tmp_path / 'mtr_output.nii.gz')
 
-    # Generate int16 test file based on sct_testing_data existing one
-    mt0 = Image('mt/mt0_reg_slicereg_goldstandard.nii.gz')
-    mt0.save(mt0_path, dtype='int16')
-    mt1 = Image('mt/mt1.nii.gz')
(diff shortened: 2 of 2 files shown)


jcohenadad:
> What is the process to add a new file of testing (ie: PR ? Request to the team ?) or do i just convert the file first directly in the test ?

yes, with PR, or you can also convert in the test

> I don't have the PR labels information since the link for the contribution guideline is broken on severals places: https://www.neuro.polymtl.ca/software/contributing#pr_labels

oopsi! something is broken... here is the list of labels for SCT: https://github.com/spinalcordtoolbox/spinalcordtoolbox/labels?q=category

would you mind opening an issue so this can be fixed? (for other projects as well)? thanks!

> The actual sct_compute_mtr tests does not check the results. Do you have some tips before i go on it ? I will be able to add them also on the initial test already existing.

ah! good question. I guess checking for the quantitative results for this function is on our list of 1000+ things to do 😉  . I'm not sure what is the best course of action here-- to be pragmatic, so far we've incrementally added unit and functional tests when one was fixing a specific part of SCT. So I suggest you do the same here (ie: add a quantitative check). 
There are a few ways you can do this. For example: open the output NIfTI file and check the value of a few selected voxels (and you can hardcode the MTR value in the test). Maybe there is a better / more elegant approach than this 🤷‍♂️ 

also useful is to look at what the functional batch_processing test does (similar goal, but all functional tests are concentrated in this function):
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/afdb05da537a75a34f3434f5512473f4e34f55bf/batch_processing.sh#L253

spinalcordtoolbox/qmri/mt.py
```diff
@@ -44,8 +44,13 @@ def compute_mtr(nii_mt1, nii_mt0, threshold_mtr=100):
     :param threshold_mtr: float: value above which number will be clipped
     :return: nii_mtr
     """
+    # Convert input before compute MTR
+    nii_mt1.change_type('float32')
+    nii_mt0.change_type('float32')
```
joshuacwnewton, line 49: While this may be an appropriate fix, I'm also curious to find out what the root cause of the problem actually is!

((For example, is there an issue with our `divide_after_removing_zero` function? I'd love to find out the place where the errors get introduced at... :thinking:))

So, I think it could be a good idea to create some int16 data (as you mentioned), then explore the behavior of the functions to determine the place where the differences get introduced...

spinalcordtoolbox/qmri/mt.py
```diff
@@ -44,8 +44,13 @@ def compute_mtr(nii_mt1, nii_mt0, threshold_mtr=100):
     :param threshold_mtr: float: value above which number will be clipped
     :return: nii_mtr
     """
+    # Convert input before compute MTR
```
joshuacwnewton, line 47: Nitpick, but we might want to use a more descriptive comment here... right now, I think this comment isn't adding much?

```suggestion
    # Convert input to avoid numerical errors from int16 data (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3636)
```

spinalcordtoolbox/qmri/mt.py
joshuacwnewton: It's probably simplest to just open the existing `mt.nii.gz` files using SCT's [`Image`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e106992c42ac8be6e74e9bf8cba53a9a71b42dd6/spinalcordtoolbox/image.py#L248) class, convert them, then re-save them to a temporary directory using pytest's [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture)?

RignonNoel: For sure! I tried to see things easily to begin with but i never say no to some analysis ;)
I will do the investigation and report on this PR as soon as I have something.

@joshuacwnewton Do you have some idea how i can have some others file of tests ? When i do it with `sct_testing_data/mt/mt0.json` and `sct_testing_data/mt/mt1.json` it just say 0 voxels for both int16 and float32 input...

@jcohenadad Maybe the file you got from the issue's author are public and we can add them in the testing base ? I'm not aware of the intellectual property of this kind of file.

joshuacwnewton: Ahh, so. The json files are different than the image files. The files you want end in the `.nii.gz` extension. (`.nii` refers to the [NIfTI file format](https://radiopaedia.org/articles/nifti-file-format), and `.gz` is additional compression to save space.)

Screenshot from 2022-01-10 14-03-15

RignonNoel: @joshuacwnewton Yes, I got that and did some personal time research on NifTI & DICOM. I also played a bit with FSLEyes to make sure I was able to read them. 

But actually when I do:

```
❯ sct_compute_mtr -mt0 sct_testing_data/mt/mt0.nii.gz -mt1 sct_testing_data/mt/mtr.nii.gz
```

I got:
```
--
Spinal Cord Toolbox (git-enhancement-help_linux_user_to_find_wsl_section-ea2ad7aaa76e2e1c8a9f6ba92762fa9ee3329cb0*)

sct_compute_mtr -mt0 sct_testing_data/mt/mt0.nii.gz -mt1 sct_testing_data/mt/mtr.nii.gz
--


Compute MTR...
Found 0 voxels with value=0. These will be replaced by nan.
Threshold to clip values: +/- 100
File ./mtr.nii.gz already exists. Will overwrite it.

Done! To view results, type:
fsleyes sct_testing_data/mt/mt0.nii.gz sct_testing_data/mt/mtr.nii.gz ./mtr.nii.gz &
```

But it say `Found 0 voxels`, and I guess that it's not good since the issue say that a lot of voxels have wrong values. 

---

The two mains problems I have are:

 - Understanding of the validation process: I do not know how to validate if the output files are correct or not
 - Miss knowledge on input files: As you say, there are some `.json` files also, but i'm not sure what they are. From what I learn on IRM lately, i guess they are the IRM process configuration to take images (ie: echo time, repetition time, etc..)

joshuacwnewton: Oh, sorry! I thought you were passing `.json` files to `sct_compute_mtr`. I misunderstood. :)

> ```
> Found 0 voxels with value=0. These will be replaced by nan.
> ```

This seems like a bug in the function to me! The function seems like it's working properly, so I feel like this log message should maybe be skipped to prevent confusion, in the case where there are no voxels that need to be replaced. 

> Understanding of the validation process: I do not know how to validate if the output files are correct or not

In `sct_testing_data`, there is already a pre-computed [`mtr.nii.gz`](https://github.com/spinalcordtoolbox/sct_testing_data/blob/master/mt/mtr.nii.gz) image, too. So, one suggestion I have is to treat the existing image like "ground truth", then output the newly calculated image to a different folder (temporary directory), and compare the ground truth to the newly calculated file, and make sure that the two files are within a specific threshold.

Similar to other tests we have? For example: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0bef2991f0c05251fbf43df8d1fc0360960565c2/testing/cli/test_cli_sct_analyze_texture.py#L19-L22

jcohenadad:
> @jcohenadad Maybe the file you got from the issue's author are public and we can add them in the testing base ? I'm not aware of the intellectual property of this kind of file.

not public but there is no way to identify the participant, so the risk of sharing is negligible: [Archive.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/7842263/Archive.zip)

RignonNoel:
@joshuacwnewton I implemented a `tmp_path` as you proposed in place of adding new files inside the repository used for `sct_testing_data`.

I also tried to add the check to see the difference between the two vectors but it seems that the content of the ouput mtr is only filled with `0`:

```
print(output_mtr.data)
```
```
[[[0. 0. 0. 0. 0.]
  [0. 0. 0. 0. 0.]
  [0. 0. 0. 0. 0.]
  
  [... lot of others zeros ...]
  
  [0. 0. 0. 0. 0.]
  [0. 0. 0. 0. 0.]
  [0. 0. 0. 0. 0.]]]
 ```
 
 Do you have some idea of what I'm doing wrong based on my actual written test ?

**Comment by joshuacwnewton** [comment]
Ah! My apologies for missing your last comment, @RignonNoel. When I checked my notifications, I only saw the reply in the thread here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3638#discussion_r781480436 and didn't scroll down enough to see your further remarks. :facepalm: 

I'll take a look at what you've done right away. :)

File: testing/cli/test_cli_sct_compute_mtr.py [hunk 3]
```diff
@@ -9,6 +11,33 @@
...
+    assert numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data) <= 0.001
+
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_mtr_with_int16_image_type(tmp_path):
+    """Run the CLI script with int_16 image type"""
+
+    mt0_path = str(tmp_path / 'mt0_int16.nii.gz')
+    mt1_path = str(tmp_path / 'mt0_int16.nii.gz')
```
joshuacwnewton, line 29: > I also tried to add the check to see the difference between the two vectors but it seems that the content of the ouput mtr is only filled with 0:

I think the highlighted section above might be the cause! (Same filepath for both `mt0_path` and `mt1_path`, hehe.)

RignonNoel: ... I will hide myself forever after that 🤦‍♂️ 

You were right, it was the only reason. Since the generation is good now I took the time to comparate files and it seems that the actual `sct_testing_data/mt/mtr.nii.gz` is not the same results from what I get by computing mtr from `sct_testing_data/mt/mt0.nii.gz` and `sct_testing_data/mt/mt1.nii.gz`.

I don't know if the problem is the function or the "ground truth" since the only test we had (ie: `test_sct_compute_mtr_no_checks`) did not check the results before.

@joshuacwnewton Do we have other way to get a mtr to create a ground truth ? (concurrent tool ? database ?)

joshuacwnewton: > Since the generation is good now I took the time to comparate files and it seems that the actual `sct_testing_data/mt/mtr.nii.gz` is not the same results from what I get by computing mtr from `sct_testing_data/mt/mt0.nii.gz` and `sct_testing_data/mt/mt1.nii.gz`.

Thank you for testing this! 

To be completely transparent, the dataset `sct_testing_data` (as well as many of the original tests) were created [6 years ago](https://github.com/spinalcordtoolbox/sct_testing_data/tree/master/mt), but I only joined the team 1.5 years ago. So, often I am finding things out just as you are! :p 

I was making an assumption when I suggested using `mtr.nii.gz`, but my assumption may have been wrong here. :sweat_smile: 

---

My first thought was that maybe the `mtr.nii.gz` file was actually generated by comparing `mt1.nii.gz` and `
mt0_reg_slicereg_goldstandard.nii.gz `, instead. Since, MTR is usually calculated between _coregistered_ MT0 and MT1 images (you can find more info in our tutorial [here](https://spinalcordtoolbox.com/user_section/tutorials/registration-to-template/mtr-computation.html)). 

If I make this change to your test, then we're able to get closer to matching MTR images: 

Screenshot from 2022-01-12 14-02-04

While not identical, this could be good enough for us to write a comparison for now? e.g. within 0.1 rather than 0.001. 

(Still, I don't know the exact conditions the `mtr.nii.gz` file was generated in... :thinking:)

---

However, the other problem I'm noticing is that the MTR images contain both `nan` and `-inf` values. 

Screenshot from 2022-01-12 14-11-25

I wasn't expecting that! And it makes performing `numpy.linalg.norm(ground_truth_mtr.data - output_mtr.data)` a bit harder. For now, I think we can use [`numpy.isfinite`](https://stackoverflow.com/a/11620982) to just compare the finite values. (Though, the `nan`/`inf` behavior might require deeper investigation later on.)

When I make all of the changes above, I think I've got a working test. Would it be okay if I push the changes to your branch in a separate commit? :)

jcohenadad: > (Still, I don't know the exact conditions the mtr.nii.gz file was generated in... 🤔)

IIRC it was using the batch_processing, so yes: after co-registering the data together. So if the goal is to assess this numerical results, i would personally not use the existing mtr data as ground truth

joshuacwnewton: > So if the goal is to assess this numerical results, i would personally not use the existing mtr data as ground truth

Apologies, but I'm not sure I understand...

My expectation was that we would want to make sure the MTR results are roughly the same as what they were 6 years ago. Wouldn't comparing the current MTR output to the previous output be a worthwhile test?

Or are you suggesting that it would be better to try and find some objective measure of "correct" results to use as ground truth to compare against instead? (This seems much more difficult given our present testing data... Hmm.)

jcohenadad: sorry for my lack of clarity: i meant that we should compare an mtr that is computed from the same input data (the current mtr is not computed from the same input data, because IIRC one has been coregistered and resampled). So one solution would be to compute another mtr, from the native data (unregistered) and add it to the testing data (eg could be called mtr_not_registered.nii.gz)

RignonNoel: @joshuacwnewton  Yes feel free to add commits, I'm not conservative on my PR ;)

@jcohenadad But if we compute a new mtr from native data (uneregistered) we will not be sure that the mtr is good since we want to test our function. Will you be able to do an exceptional manual validation to make it a ground truth ? Or do you propose to compute another mtr but with an other tool ?

jcohenadad: > @jcohenadad But if we compute a new mtr from native data (uneregistered) we will not be sure that the mtr is good since we want to test our function. Will you be able to do an exceptional manual validation to make it a ground truth ? Or do you propose to compute another mtr but with an other tool ?

right, chicken-and-egg problem 😉   the mtr we currently consider 'ground truth' isn't necessarily better than the one i am proposing to create because it would be using the same code (ie: the code is so trivial that we trust it... by maybe we shouldn't?). Alternatively you could create a synthetic NIfTI mt0 and mt1 and use that for the functional test, eg we do that for some tests, eg [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ca81d72c9e3eef265ba3a0ff63d8297b4372f3d/spinalcordtoolbox/testing/create_test_data.py#L120) and [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ca81d72c9e3eef265ba3a0ff63d8297b4372f3d/testing/api/test_resampling.py#L21).

RignonNoel: @jcohenadad Even if we use synthetic NifTI image for mt0 and mt1, the mtr groundtruth will always depend of the same code we want to test.

I don't have the medical knowledge, but if you say that there is **no other way** to compute an MTR that we can use as a ground truth I think it's OK to generate a new MTR based on float32 image as you did before (no reason to think it does not work lately) and store it as a ground truth. We can just add some comments in the test to explain the situation so future contributor can take it in account if one day they found the ground truth to be no so ground.

I just need a greenlight from you @jcohenadad since it's more a data research question than a technical software issue.

jcohenadad: > @jcohenadad Even if we use synthetic NifTI image for mt0 and mt1, the mtr groundtruth will always depend of the same code we want to test.

given that mtr is based on a [simple equation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ca81d72c9e3eef265ba3a0ff63d8297b4372f3d/spinalcordtoolbox/qmri/mt.py#L50) we could something like: mt0=100, mt1=70, "assess mtr = 100*(100-70)/100 = 30"

RignonNoel: @jcohenadad Oh! I didn't saw it was that simple 😄 

I will try to create an `mt0` `mt1` and `mtr` and compare it with the system output. Will be a good way for me to rampup on NifTI image data and we will have a ground truth for the future!

testing/cli/test_cli_sct_compute_mtr.py
```diff
@@ -1,14 +1,50 @@
...
+
+    # Comparate ground truth mtr file to new generated one
+    ground_truth_mtr = Image('mt/mtr.nii.gz')
+    output_mtr = Image(mtr_output_path)
+
+    # NB: numpy.isfinite is used to exclude nan/inf elements from comparison
+    diff = (ground_truth_mtr.data[numpy.isfinite(ground_truth_mtr.data)] -
+            output_mtr.data[numpy.isfinite(output_mtr.data)])
+
+    assert numpy.abs(numpy.average(diff)) <= 0.5
```
joshuacwnewton, line 50: What I've done here is separate from the suggestion made by @jcohenadad earlier. But, I still wanted to push these changes to demonstrate that it's possible to compare the MTR results from `sct_testing_data` to the `mtr.nii.gz` file.

I'm not sure how _useful_ this test is, but at least it's here so we can talk about it and improve it if need be. :)

joshuacwnewton: We can always keep this test in addition to the test with custom values (100, 70, etc.), maybe?

RignonNoel:
@joshuacwnewton Hi! I did some research and found how to generate files with `nibabel` and `numpy` to compare the results with your proposition.

Everything looks good for me now on the tests, I think it's good to keep a dummy comparison on the side to make sure we have a base test not depending of the testing data that can change with the time.

---

Let me know for an approval ;)

spinalcordtoolbox/qmri/mt.py
```diff
@@ -44,8 +44,14 @@ def compute_mtr(nii_mt1, nii_mt0, threshold_mtr=100):
     :param threshold_mtr: float: value above which number will be clipped
     :return: nii_mtr
     """
+    # Convert input to avoid numerical errors from int16 data
+    # Related issue: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3636
+    nii_mt1.change_type('float32')
+    nii_mt0.change_type('float32')
```
joshuacwnewton, line 50: Just to be absolutely thorough, I copied the 2 new tests over to `master` (to make sure that they properly capture the initial buggy behavior reported in #3636). 

However, the results of the tests are the same on both `master` and this branch. In other words, the `float32` changes seem to have no impact, at least for the test data we've tried.

In hindsight, I realize that we probably should have been working with this user's exact data to try and isolate the problem precisely. @jcohenadad, is that something you would be able to share privately?

joshuacwnewton: No problem whatsoever! This is partially on me, too -- I wrote the initial test with the `sct_testing_data` files as well, without considering the problem at hand. :sweat_smile: 

Regardless, the time was not wasted at all! What you've learned will still be very valuable for working with image data in the future. :)

joshuacwnewton: I can confirm that the `float32` change does indeed have a large effect on the user's data. Note the screenshots below:

<details>

sct_compute_sct on `master`:

master

sct_compute_sct with the float32 fix:

with_float32_fix

Difference between the two images:

diff

</details>

Investigating deeper, the root cause of the issue here is actually... dead simple. :facepalm: 

It's not _just_ that the user's data is `int16`, but that the user's data spans the range `[0, 4095]`:

<details>

Screenshot from 2022-01-25 12-36-15

</details>

When we combine these large values with the [`100 *` step](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/26b7fe7aa1784f61de4058b5d14a29789f9defe0/spinalcordtoolbox/qmri/mt.py#L50) in our MTR calculation code, it results in the data getting clipped by the `2^15` limits of the `int16` datatype:

<details>

Screenshot from 2022-01-25 12-39-13

</details>

By changing the datatype to `float32`, the data  is no longer clipped, and we can compute the MTR as expected:

<details>

Screenshot from 2022-01-25 12-40-44

</details>

So, the original fix works as intended! And, the only reason our tests didn't capture the original buggy behavior is that our values weren't large enough. :laughing:

I'll fix up the tests so that they fail on `master` but pass on this branch, and we should be good to go. :)

RignonNoel: @joshuacwnewton Good catch! In this context should we enhance a bit my fix and transform ALL the images input into a biggest dtype before calculation ? Today it was a bug due to int16 but tomorrow it could be an error with a int64 for example.

```
(2, 'uint8', np.uint8, "NIFTI_TYPE_UINT8"),\
(4, 'int16', np.int16, "NIFTI_TYPE_INT16"),\
(8, 'int32', np.int32, "NIFTI_TYPE_INT32"),\
(16, 'float32', np.float32, "NIFTI_TYPE_FLOAT32"),\
(32, 'complex64', np.complex64, "NIFTI_TYPE_COMPLEX64"),\
(64, 'float64', np.float64, "NIFTI_TYPE_FLOAT64"),\
(256, 'int8', np.int8, "NIFTI_TYPE_INT8"),\
(512, 'uint16', np.uint16, "NIFTI_TYPE_UINT16"),\
(768, 'uint32', np.uint32, "NIFTI_TYPE_UINT32"),\
(1024,'int64', np.int64, "NIFTI_TYPE_INT64"),\
(1280, 'uint64', np.uint64, "NIFTI_TYPE_UINT64"),\
(1536, 'float128', _float128t, "NIFTI_TYPE_FLOAT128"),\
(1792, 'complex128', np.complex128, "NIFTI_TYPE_COMPLEX128"),\
(2048, 'complex256', _complex256t, "NIFTI_TYPE_COMPLEX256"),
```

joshuacwnewton: That's a good question... Perhaps a bit out of scope for this PR, though? I feel like I want to give it more time and thought. :thinking:

RignonNoel: The actual fix is an enhancement, and it will fix the problem for the user that created the issue. 
We can take more time and create a new issue for the generic problem.

testing/cli/test_cli_sct_compute_mtr.py
```diff
@@ -1,14 +1,78 @@
...
+    sct_compute_mtr.main(argv=['-mt0', mt0_path, '-mt1', mt1_path,
+                               '-o', mtr_output_path, '-thr', '2000'])
+
+    # Compare ground truth mtr file to new generated one
+    ground_truth_mtr = Image(mtr_groundtruth_path)
+    output_mtr = Image(mtr_output_path)
+
+    # NB: numpy.isfinite is used to exclude nan/inf elements from comparison
+    assert numpy.array_equal(ground_truth_mtr.data[numpy.isfinite(ground_truth_mtr.data)],
+                             output_mtr.data[numpy.isfinite(output_mtr.data)])
```
joshuacwnewton, line 36: Good news! I was able to fix up the other test, too, so that we can compare the files in `sct_testing_data` _exactly_. 

This confirms that `mtr.nii.gz` was indeed created from `mt0_reg_slicereg_goldstandard.nii.gz` and `mt1.nii.gz` 6 years ago. :P

spinalcordtoolbox/qmri/mt.py
joshuacwnewton: As an aside, I've fixed up the test to use high-valued data in ab3d0be!

And, just to demonstrate that this modified test does indeed capture the buggy behavior on `master`, here's what the results look like when the data gets clipped:

Screenshot from 2022-01-25 13-47-55

On this branch, however, the two arrays are identical. :smile:

joshuacwnewton (approved):
I think this looks good to me! But, I also made some changes, and I can't really self-review here, so could you perhaps look over the new tests, @RignonNoel?

If they look good to you, then I think we can merge this. :)

PR #3646, opened 2022-01-13
Move dependencies from `requirements.txt` to `setup.py :: install_requires`

abelfodil:
#### GitHub

#### PR contents

## Description

Require `setup.py` to install SCT dependencies.

## Ideas (outside the scope of this PR)
- We should not install inference libraries, this side-steps the `-f` issue.
- If a user runs a command requiring inference, we should ask them if we can install inference tools


requirements.txt
```diff
@@ -1,41 +1,2 @@
...
-torch==1.5.0+cpu; sys_platform != "darwin"
-torch==1.5.0; sys_platform == "darwin"
-torchvision==0.6.0+cpu; sys_platform != "darwin"
-torchvision==0.6.0; sys_platform == "darwin"
-xlwt
-tqdm
-transforms3d
-urllib3[secure]
-pytest_console_scripts
-wquantiles
```
abelfodil, line 41: This was converted to `setup.py` using a regex

install_sct
```diff
@@ -637,9 +637,8 @@ fi
 (
   # We use "--ignore-installed" to preserve the version of `certifi` installed into the conda
   # env, which prevents https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3609
-  pip install -r "$REQUIREMENTS_FILE" --ignore-installed certifi &&
```
joshuacwnewton, line 640: Keep in mind that SCT is currently using 2 different types of requirements file:

* `requirements.txt`: This is the default one, present on `master`, with flexible requirements.
* `requirements_freeze.txt`: This is a set of frozen dependencies for stable releases. It is generated at time of release to ensure a stable set of requirements.

You can see the logic to parse `requirements_freeze.txt` files here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ca81d72c9e3eef265ba3a0ff63d8297b4372f3d/install_sct#L629-L636

Because of this, the changes in this PR will currently break any future releases we create. 

But, I think this is fixable! I think we could put the `requirements_freeze.txt` logic inside of `setup.py`. Something like "if requirements_freeze.txt is present, read the dependencies from file, otherwise use the default requirements".

abelfodil: Gotcha, I will submit the logic to parse the frozen file inside setup.py tonight. May I ask, how is the frozen file generated? Do you create a venv, install packages and pip freeze?

abelfodil: @joshuacwnewton if the build passes (at least It Works On My Machine™), it's ready for review.

I added support for frozen requirements and improved dependency check.

joshuacwnewton: Thank you! I'll take a look at this first thing tomorrow. :)

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -27,8 +27,10 @@
 
 import requirements
 
+from dependencies import dependencies
```
joshuacwnewton, line 30: One quick comment: I noticed our RTD documentation build is failing for this PR, and the error seems to be:

WARNING: Unexpected return code 1 from command Command(command=('sct_check_dependencies', '-h'), shell=False, hide_standard_error=False, working_directory='/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/checkouts/3646/documentation/source') (output='Traceback (most recent call last):\n  File "/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/envs/3646/lib/python3.7/site-packages/spinalcordtoolbox/scripts/sct_check_dependencies.py", line 30, in <module>\n    from dependencies import dependencies\nModuleNotFoundError: No module named \'dependencies\'')

Highlighting the line here.

abelfodil: I updated the PR but the CI has been stuck for 18h, is that normal?

joshuacwnewton: Very sorry about this! 

Screenshot from 2022-01-14 12-24-29

This is something that GitHub enforces, and there's no settings in our admin panel to stop it from happening. As soon as your first SCT PR is merged, CI runs will start immediately. :)

.github/workflows/create-release.yml
```diff
@@ -40,10 +40,8 @@ jobs:
       run: |
         source python/etc/profile.d/conda.sh
         conda activate venv_sct
-        pip freeze | grep -v "-e git+" | grep -v "torch" > requirements-freeze.txt
+        pip freeze | grep -v "-e git+" > requirements-freeze.txt
         conda deactivate
-        echo "# Platform-specific torch requirements (See SCT Issue #2745)" >> requirements-freeze.txt
-        grep "torch" requirements.txt >> requirements-freeze.txt
```
joshuacwnewton, line 46: Hrmm... The reason I did the `torch` grepping initially is because our old `requirements.txt` file had some platform-specific branching paths, depending on whether the user is on macOS or Linux:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/129e537b83e445fe96aed8b1971d748d8424c25e/requirements.txt#L32-L35

However, if we just `pip freeze`, then we won't get those platform-specific statements:

```console
(venv_sct) joshua@tadpole:~/repos/spinalcordtoolbox$ pip freeze
absl-py==1.0.0
astor==0.8.1
[...]
torch==1.5.0+cpu
torchvision==0.6.0+cpu
[...]
```

So we would need a way to fetch these platform-specific torch requirements from the `setup.py`. :thinking:

requirements.txt
```diff
@@ -1,41 +1,2 @@
...
-requirements-parser
-scipy
-scikit-image
-scikit-learn
-tensorflow~=1.15.0
-# PyTorch's Linux/Windows distribution is very large due to its GPU support,
-# but we only need that for training models. For users, use the CPU-only version
-# (only available directly from the PyTorch project).
-# The macOS version has never had GPU support, so doesn't need the workaround.
 -f https://download.pytorch.org/whl/cpu/torch_stable.html
```
joshuacwnewton, line 1: Hmm... I wonder if `requirements.txt` is even necessary anymore if its contents are just:

```
-f https://download.pytorch.org/whl/cpu/torch_stable.html
-e .
```

Since we want to encourage use of `install_sct` (for now), it might be better to get rid of this entirely, so people don't try to run `pip install -r requirements.txt` and end up with a broken installation.

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -157,7 +157,7 @@ def add_bash_profile(string):
 
 def get_dependencies(requirements_txt=None):
     if requirements_txt is None:
-        requirements_txt = sct_dir_local_path("requirements.txt")
+        requirements_txt = sct_dir_local_path("generated_requirements.txt")
```
joshuacwnewton, line 160: Oh, of course! Good catch here. I didn't even consider that `sct_check_dependencies` relies on `requirements.txt`. 

This actually seems like it fixes a bug, I think? Right now, when `sct_check_dependencies` is run for stable releases, it's only checking against the flexible requirements. Now, with your new changes, it will check against the stricter frozen requirements, which is what we want!

Thank you for this!

documentation/source/user_section/installation/mac.rst
```diff
@@ -111,13 +111,13 @@ Procedure:
 
    .. code:: sh
 
-      pip install -e .
+      pip install -e . -f https://download.pytorch.org/whl/cpu/torch_stable.html
```
joshuacwnewton, line 114: Oh, huh! I forgot that "Option 3: Install with pip (experimental)" was a part of our docs, to be completely honest.

Part of me suspects that those instructions don't even work? My hunch is that no one has actually tried them in a long time... But that's probably out of scope for this PR for now, aha.

joshuacwnewton (commented):
I really appreciate the work you've done on this PR! Most of the changes look great to me. I just had a few small comments. :tada:

.github/workflows/create-release.yml
abelfodil: I wrongly assumed they weren't needed after we froze packages oops

requirements.txt
abelfodil: Should we include this in another PR (same one that removes `pip` from doc)?

.github/workflows/create-release.yml
abelfodil: Fixed

install_sct
```diff
@@ -625,21 +625,11 @@ python -m pip install -U "pip!=21.2.*"
 
 ## Install the spinalcordtoolbox into the Conda venv
 print info "Installing Python dependencies..."
```
joshuacwnewton, line 627: This log message now looks a little weird, since there is just an empty space:

Screenshot from 2022-01-18 10-26-16

We can probably merge the message into the `Installing spinalcordtoolbox...` message so that we just have 1 message that says:

```
Installing spinalcordtoolbox and its Python dependencies...
```

requirements.txt
joshuacwnewton: I think we could do it in this PR! (And save the documentation update for a separate PR.)

The only other thing we'd have to do, I think, is update `.readthedocs.yml`, which currently relies on the presence of `requirements.txt`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e49161416a8e79bc75ce74b55c0ddf5563689327/.readthedocs.yml#L13-L20

I think it's just as simple as going:

```diff
-    - requirements: requirements.txt
```

Since ReadTheDocs already runs `pip install .[docs]` during docs builds (which will install SCT, its dependencies, and the docs `extra_requirements`):

Screenshot from 2022-01-18 10-38-06

setup.py
```diff
@@ -13,6 +13,70 @@
...
+    'urllib3[secure]',
+    'pytest_console_scripts',
+    'wquantiles',
+]
+
+frozen_dependencies = get_dependencies(path.join(here, 'requirements-freeze.txt'))
+
+dependencies = frozen_dependencies or DEFAULT_REQUIREMENTS
+
+with open(path.join(here, "generated_requirements.txt"), 'wt') as f:
```
joshuacwnewton, line 76: This is just a nitpick, but what do you think of the name `installed_requirements.txt` instead?

The reason I suggest this is because SCT also generates an install log that gets created by `install_sct`. So, if we choose a similar name, then it might be easier to understand what the file is:

Screenshot from 2022-01-18 10-47-36

vs.

Screenshot from 2022-01-18 10-49-59

joshuacwnewton:
Small note: I just realized that this PR is accomplishing the exact same thing as @kousu's old PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2846.

Since this PR is more active (and much more up to date), I'm going to close https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2846 and we can continue to focus on this PR.

setup.py
abelfodil: Yup makes sense

requirements.txt
abelfodil: Ok I tried removing it, but it fails to install torch+cpu for the doc. I couldn't find an argument to include the torch repo unfortunately :(

joshuacwnewton: > Ok I tried removing it, but it fails to install torch+cpu for the doc.

_(Linking the build for posterity: https://readthedocs.org/projects/spinalcordtoolbox/builds/15808655/)_

joshuacwnewton: Ahh of course, since we can't use `-f` inside `setup.py`, and we're relying on external commands (e.g. `-f` in `install_sct`) to make that work.

joshuacwnewton: Perhaps we could open an upstream issue in ReadTheDocs to ask if there's any way they support `-find-links` option for `setup.py`? But, for now, I think we can save this for a follow-up PR, and not delay this PR too much.

joshuacwnewton (approved):
Thank you for this!! And thank you for the quick responses to feedback. LGTM! :heart:

Issue #3367, opened 2021-04-29
Add support for Python 3.9 by 2024-10-14


joshuacwnewton:
> **Note**: This issue used to cover the tasks of supporting both Python 3.8 and Python 3.9. However, Python 3.8 support was added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3916, which was accomplished by removing our dependency on `tensorflow`. So, I've edited this issue to hide any mentions of `tensorflow`/Python 3.8, so that we can focus purely on Python 3.9.

Current blockers for Python 3.9:

- [ ] PyQt5 (`v5.14.2`): Unable to install wheel on Windows, and the srcbuild process requires `qmake`. ([#3367 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1531670177))
    - Potential fix 1: Renaming the Python 3.8 `.whl` file to trick `pip`. ([#3367 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1531718166))
    - Potential fix 2: Updating PyQt5 to a version with Python 3.9 wheels. ([#3367 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1531769467))
    - Potential fix 3: _Downgrading_ PyQt5 to a version with Python 3.9 wheels (???). ([#3367 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1533181462))

<details>
<summary> Past blockers for Python 3.9 (Completed) </summary>

- [x] `tensorflow==1.15.0`
    * This would be pinned due to #3361, however [wheels are not provided for Py3.9](https://pypi.org/simple/tensorflow/). 
    * The first Py3.9-supporting version is `tensorflow==2.5.0`. We've used 2.0.1 before (see #2592) but it's unclear if 2.5.0 is viable. More testing is needed.
    * Alternatively, #3046 will drop the dependency on tensorflow altogether.
- [x] `onnxruntime==1.4.0`
    * EDIT: Good news! This will be fixed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3433.
    * ~~We pin this due to #3037, however [wheels are not provided](https://pypi.org/simple/onnxruntime/) for Py3.9.~~
[...]

Issue #614, opened 2015-10-21
sct_get_centerline documentation not clear


benjamindeleener:
The documentation of ```sct_get_centerline``` from labels is not clear.

The old function was:
```
sct_get_centerline_from_labels -i seg_and_labels.nii.gz
```

Now, we need to put a ```-l``` option:
```
sct_get_centerline -i seg_and_labels.nii.gz -method labels
ERROR: Needs input label (option -l).
```

What is it? And why do we need to provide it?

benjamindeleener:
`sct_get_centerline` has been removed.

PR #3653, opened 2022-01-14
Upgrade Tensorflow to 2.7

abelfodil:
#### GitHub

#### PR contents

## Description

- Builds upon https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3646 (don't review the file changes just yet)
- Uses Keras that is shipped with TF2
- Upgrade to TF 2.7.0 (I just put the latest version but I can change it to an earlier one)
- Update documentation for hard-core pythonistas
- I'm waiting for the pipeline to run to validate the changes (installation worked tho)

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -628,6 +628,7 @@ print info "Installing spinalcordtoolbox and its Python dependencies..."
 (
   # We use "--ignore-installed" to preserve the version of `certifi` installed into the conda
   # env, which prevents https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3609
+  conda install -c anaconda tensorflow-mkl
   pip install -e . --ignore-installed certifi -f https://download.pytorch.org/whl/cpu/torch_stable.html
 ) ||
   die "Failed running pip install: $?"
--- a/install_sct
+++ b/install_sct
@@ -628,7 +628,6 @@ print info "Installing spinalcordtoolbox and its Python dependencies..."
 (
   # We use "--ignore-installed" to preserve the version of `certifi` installed into the conda
   # env, which prevents https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3609
-  conda install -c anaconda tensorflow-mkl
   pip install -e . --ignore-installed certifi -f https://download.pytorch.org/whl/cpu/torch_stable.html
 ) ||
   die "Failed running pip install: $?"
@@ -645,6 +644,7 @@ linux*)
   fi
   ;;
 osx)
+  conda install -c anaconda tensorflow-mkl
   if sysctl machdep.cpu.brand_string | grep -q "Apple M1"; then
     print info "AVX is not supported on M1 Macs! Installing a version of TensorFlow compiled for M1 Macs..."
     pip uninstall tensorflow -y
--- a/setup.py
+++ b/setup.py
@@ -49,8 +49,8 @@ DEFAULT_REQUIREMENTS = [
     'scikit-image',
     'scikit-learn',
     'intel-tensorflow~=2.7.0; sys_platform != "darwin"',
-    'tensorflow~=2.7.0; sys_platform == "darwin"',
-    'tensorflow-metal; sys_platform == "darwin"',
+    # 'tensorflow~=2.7.0; sys_platform == "darwin"',
+    # 'tensorflow-metal; sys_platform == "darwin"',
     'PyYAML',
     # PyTorch's Linux/Windows distribution is very large due to its GPU support,
     # but we only need that for training models. For users, use the CPU-only version
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ DEFAULT_REQUIREMENTS = [
     'scikit-image',
     'scikit-learn',
     'intel-tensorflow~=2.7.0; sys_platform != "darwin"',
-    'tensorflow-macos~=2.7.0; sys_platform == "darwin"',
+    'tensorflow~=2.7.0; sys_platform == "darwin"',
     'tensorflow-metal; sys_platform == "darwin"',
     'PyYAML',
     # PyTorch's Linux/Windows distribution is very large due to its GPU support,
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,8 @@ DEFAULT_REQUIREMENTS = [
     'scikit-image',
     'scikit-learn',
     'intel-tensorflow~=2.7.0; sys_platform != "darwin"',
-    'tensorflow~=2.7.0; sys_platform == "darwin"',
+    'tensorflow-macos~=2.7.0; sys_platform == "darwin"',
+    'tensorflow-metal; sys_platform == "darwin"',
     'PyYAML',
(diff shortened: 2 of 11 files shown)


dev/install/install_dependencies_with_workarounds.py
```diff
@@ -1,35 +1,11 @@
...
 # coding: utf-8
-# Check if the platform requires conda tensorflow.
+# Check if the platform requires conda skimage
 #
 # Author: Christian S. Perone
 
-import platform, subprocess
+import subprocess
 
-def tf_workarounds():
```
abelfodil, line 9: not sure if we should remove

install_sct
```diff
@@ -625,46 +625,14 @@ python -m pip install -U "pip!=21.2.*"
...
   # We use "--ignore-installed" to preserve the version of `certifi` installed into the conda
   # env, which prevents https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3609
-  pip install -r "$REQUIREMENTS_FILE" --ignore-installed certifi &&
   print info "Installing spinalcordtoolbox..." &&
-  pip install -e .
+  pip install -e . --ignore-installed certifi -f https://download.pytorch.org/whl/cpu/torch_stable.html
 ) ||
   die "Failed running pip install: $?"
 
-## Check if CPU supports AVX instruction set, and install precompiled TensorFlow if it isn't,
```
abelfodil, line 646: not sure if we should remove

dev/install/install_dependencies_with_workarounds.py
joshuacwnewton: Ah, so. The `dev/` folder is largely... dead code? There are a lot of things in that folder that were once relevant, but are no longer touched. SCT hasn't exactly had a great procedure for pruning bits that are no longer needed?

So, as far as changes that affect present-day SCT, you can safely ignore almost everything in `/dev/`

install_sct
joshuacwnewton: This will need to be updated (rather than removed). Though, I'm fairly certain that we'll be able to find (or compile ourselves) new `.whl` files for whichever version of TF we upgrade, though. There are lots of people out there who have an interest in providing support for AVX-less CPUs. :)

For now, you can just ignore this? I think it'll be easiest to just focus on getting TF working for our CI runners. And then, once we can confirm that it's possible to upgrade, we can go back and fill in the gaps for AVX-less users.

setup.py
```diff
@@ -52,7 +48,9 @@ def get_dependencies(requirements_path=None):
     'scipy',
     'scikit-image',
     'scikit-learn',
-    'tensorflow~=1.15.0',
+    'intel-tensorflow~=2.7.0; sys_platform != "darwin"',
```
abelfodil, line 51: Fix for https://github.com/onnx/onnx-tensorflow/issues/535

abelfodil:
If the observable behavior differs from what we want, we could call https://www.tensorflow.org/api_docs/python/tf/compat/v1/disable_v2_behavior

setup.py
```diff
@@ -52,7 +48,9 @@ def get_dependencies(requirements_path=None):
...
+    'tensorflow~=2.7.0; sys_platform == "darwin"',
```
abelfodil, line 52: @joshuacwnewton do you happen to know if there's a MacOS TF2 version with MKL support? Right now the vanilla CPU version can't run ConvXD layers correctly :(

I got around this issue for Windows and Linux by using `intel-tensorflow`. Unfortunately I don't have a Mac machine that I can legally use for this purpose.

abelfodil: Current `tensorflow-metal` package requires Python 3.{8,9} so using it is out of the question :( Looking for alternatives for `tensorflow` accelerations

joshuacwnewton: One small note:: The only thing keeping us from upgrading to Python 3.8/3.9 _is_ TensorFlow (2.2 and 2.5 respectively). (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367)

So, upgrading both at the same time wouldn't be out of the question!

joshuacwnewton: One other comment I want to make: In SCT, we're merely using TF/Keras to perform inference using pre-trained models (as opposed to training models from scratch). So, we actually don't have much of a need for GPU-acceleration? In fact, it may even be preferable to stick to CPU-only libraries, since that would typically reduce the installation size of SCT.

So, if you have a working CPU-based upgrade to TF, then I would say that's more than good enough. :)

joshuacwnewton: Here's my attempt at summarizing the exploration you've done (just to make sure I'm understanding OK):

---

Upgrading stock `tensorflow` to 2.X causes a error to be thrown re: `NHWC` vs. `NCHW` data formatting. Specifically:

```
CPU implementation of Conv3D currently only supports the NHWC tensor format.
```

(Ah! I'm just realizing I came across the exact same error in my previous investigations, too: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-907357925)

This is the root issue, and there are [three possible workarounds](https://github.com/onnx/onnx-tensorflow/issues/535#issuecomment-568743917) for this issue (according to [the linked comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3653#discussion_r787271777), anyway):

  1. Using an MKL-based version of Tensorflow:
      * [`intel-tensorflow`](https://pypi.org/project/intel-tensorflow/) is a pip package that provides MKL support, but only for Linux/Windows, and not macOS. 
      * [`tensorflow-mkl`](https://anaconda.org/anaconda/tensorflow-mkl) is a conda package that provides MKL support for macOS, but only for TF 2.0.0. (Dead-end for macOS support.)
  2. Using a GPU-based version of Tensorflow:
       * [`tensorflow-metal`](https://pypi.org/project/tensorflow-metal/) is a pip package that provides GPU support for macOS, but only for Python 3.8/3.9. (And it would also [involve PluggableDevices](https://pypi.org/project/tensorflow-metal/), which may require more complicated prerequisites on the user side of things.)
   3. Trying to... modify the data format on the model side of things?
       * As far as I can tell, we haven't explored this option yet, right?

joshuacwnewton: There's a potential "workaround 4." in [this comment](https://github.com/onnx/onnx-tensorflow/issues/535#issuecomment-683366977):

> setting `os.environ["CUDA_VISIBLE_DEVICES"]="-1"` (as mentioned in this reply [#614 (comment)](https://github.com/onnx/onnx-tensorflow/issues/614#issuecomment-625251872)) fixes the issue for me.

That linked reply (from onnx-tensorflow issue 614) seems pretty illuminating to me! I didn't really understand _why_ the NWHC/NCHW formatting issues were a problem in the first place, but that explanation maybe clarifies things a little? :eyes:

abelfodil: That's a good summary of the situation. For "workaround 4", I don't think it works because `tensorflow-cpu` package fails as well.

joshuacwnewton:
Closing this PR, as it's been superseded by #3738, which removes Tensorflow entirely from SCT.

abelfodil:
That's great!

PR #3658, opened 2022-01-20
Add utility function to strip '.py' extensions

mguaypaq:
#### GitHub

#### PR contents

- [X] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution

## Description

The previous use of `str.strip(".py")` is a bug waiting to happen,
because it strips from both sides of the string (instead of just
the end), and more importantly, treats `".py"` as a set of
characters rather than a substring to look for (see [the docs](https://docs.python.org/3/library/stdtypes.html#str.strip)):

```
$ python3
>>> "ppp.yyy".strip(".py")
''
>>> "no_dot_py".strip(".py")
'no_dot_'
>>> ".py_prefix".strip(".py")
'_prefix'
```

All uses of `.strip(".py")` are either in `utils/sys.py` itself or import this module, so it seemed like the natural place to add this utility function.

### DIFF
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -87,6 +87,7 @@ def set_loglevel(verbose):
         # the loglevel changes from leaking: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3341
         pass
 
+
 def strip_py(fname):
     """
     Strip ".py" suffix from the string `fname`, if present.
@@ -95,6 +96,7 @@ def strip_py(fname):
     :return: a stripped copy of `fname`
     """
     return fname[:-3] if fname.endswith('.py') else fname
+
 
 # TODO: add test
 def init_sct():
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -100,6 +100,7 @@ def removesuffix(self: str, suffix: str) -> str:
     else:
         return self[:]
 
+
 # TODO: add test
 def init_sct():
     """
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -87,6 +87,14 @@ def set_loglevel(verbose):
         # the loglevel changes from leaking: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3341
         pass
 
+def strip_py(fname):
+    """
+    Strip ".py" suffix from the string `fname`, if present.
+
+    :param fname: str: a filename
+    :return: a stripped copy of `fname`
+    """
+    return fname[:-3] if fname.endswith('.py') else fname
 
 # TODO: add test
 def init_sct():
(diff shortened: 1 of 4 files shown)


joshuacwnewton (commented):
<details>

I think your concerns are valid! Just as an alternate suggestion, though, how do you feel about using [`str.removesuffix`](https://docs.python.org/3/library/stdtypes.html?highlight=rstrip#str.removesuffix) instead? It's also mentioned in the docs for [`str.rstrip`](https://docs.python.org/3/library/stdtypes.html?highlight=rstrip#str.rstrip). 

That way there's one fewer utility function we have to maintain in our codebase.

</details>

EDIT: I missed the "New in version 3.9". Oops! We can't use the feature that I suggested, since we're on Python 3.7 atm.

spinalcordtoolbox/utils/sys.py
```diff
@@ -88,6 +88,16 @@ def set_loglevel(verbose):
...
 
 
+def strip_py(fname):
+    """
+    Strip ".py" suffix from the string `fname`, if present.
+
+    :param fname: str: a filename
+    :return: a stripped copy of `fname`
+    """
+    return fname[:-3] if fname.endswith('.py') else fname
```
joshuacwnewton, line 98: How would you feel about adopting the more general version of the function (`str.removesuffix`), since the source code is provided in the [PEP proposal](https://www.python.org/dev/peps/pep-0616/)?

```python
def removesuffix(self: str, suffix: str, /) -> str:
    """
    Source: https://www.python.org/dev/peps/pep-0616/

    TODO: Replace with built-in str.removesuffix method after upgrading to Python 3.9
    """
    # suffix='' should not call self[:-0].
    if suffix and self.endswith(suffix):
        return self[:-len(suffix)]
    else:
        return self[:]
```

testing/api/test_utils.py
```diff
@@ -7,6 +7,23 @@
...
+
+    # A short string
+    assert utils.strip_py("a") == "a"
+
+    # Permutations and prefixes should not be removed
+    assert utils.strip_py("ppp.yyy") == "ppp.yyy"
+    assert utils.strip_py("no_dot_py") == "no_dot_py"
+    assert utils.strip_py(".py_prefix") == ".py_prefix"
+
+
```
joshuacwnewton, line 26: I appreciate your commitment to writing thorough tests! I'm wondering if this is absolutely necessary, though? 

This is a very straightforward utility function, composed of basic built-in Python logic. So, once we run this test once to demonstrate that the function works as intended, I'm wondering if the test is all that useful anymore as something we run on every PR. It's a very isolated function, and one that isn't likely to change...

I think I'd almost prefer having one fewer test to have to maintain? Though, I'm probably nitpicking here, aha. If anything, my comment here is just me trying to reason about like, "What our motivation is to write tests? What makes a useful test?" Etc. :thinking:

spinalcordtoolbox/utils/sys.py
mguaypaq: Yes!!! I was initially so surprised that this _still_ didn't seem to exist as a built-in, but I think I was just looking at old docs. And your doc-comment, with the official source and the TODO to eventually get rid of the code, are perfect. Thank you! I'll use this version instead.

testing/api/test_utils.py
mguaypaq: You're right, I'll get rid of this test. I like your reasoning. In the spirit of sharing thoughts, my motivations were pretty weak:
- Partly this was an exercise in just finding our existing test suite and adding to it.
- Partly this was documentation of cases which failed with the previous code, especially since that code looked quite natural.
- And finally, since we just turned on codecov I kind of just wanted to see the number go up once :stuck_out_tongue_winking_eye:

PR #3660, opened 2022-01-24
Allow `install_sct` to run from any directory

### LINKED ISSUE #3647: `install_sct` should run from any directory
Hi, thanks for reporting an issue, please take some time to consider the applicable guidelines:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/CONTRIBUTING.rst#reporting-a-bug-or-requesting-a-feature
(about how to suitably describe issues or requests) prior to deleting this blurb when about to submit the issue.

### Description

I would like to execute the script from anywhere. Having to `cd` or `pushd` is a minor inconvenience.

Instead, the source directory can be retrieved using any of the methods [here](https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel).

The fixes would be to either:
-  `pushd` at the beginning and `popd` at the end of the script
- Set the source dir as the working dir for the different commands (ex: `pip install -e $SCT_SOURCE_LOC`, etc.)

This would make the script more robust.

### Steps to Reproduce

1. Go outside the repo

2. Try to install SCT

Installation succeeds

Error message telling me "The install_sct script must be executed from the source directory"

mguaypaq:
#### GitHub

#### PR contents

## Description

Allow the `install_sct` script to run from any location, by first changing directory to the script location, and changing back to the original directory at the end.

Possibly this could have been integrated into the existing `finish()` function, but this way seems nicely self-contained to me. The downside is that it's not logged in `install_sct_log.txt`.

The script passes ShellCheck version 0.8.0, and a test run from a different directory worked fine on my machine.

## Linked issues

Fixes #3647

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -46,7 +46,6 @@ shopt -s failglob # error if a glob doesn't find any files, instead of remaining
 
 # Where tmp file are stored
 TMP_DIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'TMP_DIR')"
-# Source directory
 SCT_SOURCE="$PWD"
 DATA_DIR="data"
 PYTHON_DIR="python"
--- a/install_sct
+++ b/install_sct
@@ -29,6 +29,11 @@
 # Authors: PO Quirion, J Cohen-Adad, J Carretero
 # License: see the file LICENSE.TXT
 
+# Change directory to where the script is, for the duration of the script.
+# This lives outside the subshell below to guarantee the corresponding
+# popd gets run, no matter how the installation script exits.
+pushd -- "${BASH_SOURCE[0]%/*}/" || exit 1
+
 ( # | tee install_sct_log.txt
 
 # stricter shell mode
@@ -39,10 +44,9 @@ shopt -s failglob # error if a glob doesn't find any files, instead of remaining
 
 # set -v  # v: verbose
 
-
 # Where tmp file are stored
 TMP_DIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'TMP_DIR')"
-# Start Directory So we go back there at the end of the Script
+# Source directory
 SCT_SOURCE="$PWD"
 DATA_DIR="data"
 PYTHON_DIR="python"
@@ -111,8 +115,6 @@ function run() {
 function finish() {
   # Catch the last return code
   value="$?"
-  # Get back to starting point
-  cd "$SCT_SOURCE"
   if [[ "$value" -eq 0 ]]; then
     print info "Installation finished successfully!"
   elif [[ "$value" -eq 99 ]]; then
@@ -120,7 +122,7 @@ function finish() {
     echo ""
   else
     print error "Installation failed!\n
-Please find the file $(pwd)/install_sct_log.txt, 
+Please find the file \"$SCT_SOURCE/install_sct_log.txt\",
 then upload it as a .txt attachment in a new topic on SCT's forum:
 --> http://forum.spinalcordmri.org/c/sct"
   fi
@@ -405,8 +407,7 @@ check_requirements
 if [[ -e "spinalcordtoolbox/version.txt" ]]; then
   SCT_VERSION="$(cat spinalcordtoolbox/version.txt)"
 else
-  die "ERROR: version.txt not found. \n
-The install_sct script must be executed from the source directory"
+  die "ERROR: version.txt not found."
 fi
 
 # Get installation type (from git or from package) if not already specified
@@ -744,3 +745,5 @@ else
 fi
 
 ) 2>&1 | tee install_sct_log.txt
+
+popd
--- a/install_sct
+++ b/install_sct
@@ -744,6 +744,6 @@ else
 fi
 
 ) 2>&1 | tee install_sct_log.txt
-status=$?
+status=${PIPESTATUS[0]}
 popd >/dev/null
-exit $status
+exit "$status"
--- a/install_sct
+++ b/install_sct
@@ -32,7 +32,7 @@
 # Change directory to where the script is, for the duration of the script.
 # This lives outside the subshell below to guarantee the corresponding
 # popd gets run, no matter how the installation script exits.
-pushd -- "${BASH_SOURCE[0]%/*}/" || exit 1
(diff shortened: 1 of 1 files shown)


install_sct
```diff
@@ -29,6 +29,11 @@
 # Authors: PO Quirion, J Cohen-Adad, J Carretero
 # License: see the file LICENSE.TXT
 
+# Change directory to where the script is, for the duration of the script.
+# This lives outside the subshell below to guarantee the corresponding
+# popd gets run, no matter how the installation script exits.
+pushd -- "${BASH_SOURCE[0]%/*}/" || exit 1
```
joshuacwnewton, line 35: IIUC `|| exit 1` is allowing us to exit prematurely with a non-zero error code in the case that `pushd` fails. 

But, what if we instead moved the `set` commands outside of the subshell? 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ebea2b27eedf50ea043a592b2e2d9a4db179b689/install_sct#L39-L43

I _think_ that the subshell should inherit the behavior of the parent shell? See, e.g.; https://stackoverflow.com/questions/22630363/how-does-set-e-work-with-subshells

install_sct
```diff
@@ -744,3 +745,5 @@ else
 fi
 
 ) 2>&1 | tee install_sct_log.txt
+
+popd
```
joshuacwnewton, line 749: By default, `pushd` and `popd` echo the directory stack back to `stdout`. This looks a bit odd without the extra context, I think? So, we might want to consider:

1. Suppressing the messages using `> /dev/null`, and/or
2. Adding a more informative message, such as "Temporarily changing directories to <>" and "Returning to the original directory..."

install_sct
```diff
@@ -29,6 +29,11 @@
 # Authors: PO Quirion, J Cohen-Adad, J Carretero
```
joshuacwnewton, line 29: One other note: The failing test is due to a platform specific issue with one of GitHub's VMs for CI runners. (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597)

We haven't figured it out quite yet, but it will cause intermittent errors depending on the machine we get from GitHub. It's luck of the draw, but you can generally just ignore it and hit "re-run jobs" when you see it come up.

joshuacwnewton (changes requested):
Thank you for taking this on!

install_sct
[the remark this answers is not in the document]
mguaypaq: Yes, `--` is useful when, like me, you get paranoid about shell safety! In this case, `pushd` can also accept switches which start with `-` or `+`, so if the expanded path somehow started with one of those characters (which can _very_ technically happen), you would get either strange behaviour or an error message. (For example, on my machine `pushd +/` fails with the message `bash: pushd: +/: invalid number`.)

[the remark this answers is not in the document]
mguaypaq: That's correct! There's a surprisingly large amount of string mangling you can do with variable expansions in bash (see `Parameter Expansion` in [the bash man-page](https://man7.org/linux/man-pages/man1/bash.1.html#EXPANSION)) (it's usually a bad idea to do complicated string mangling in bash, but possible).

Another bit of bash trivia: command substitution (for example, `"$(pwd)"` or `"$(dirname path/to/a/file)"`) automatically strips any trailing newline characters from the command output. This means that you can get some really subtle failures if you use `dirname` and you have a directory name which ends in a newline:
```
$ foo=$'dir_ends_with_newline\n/file'
$ echo "${foo%/*}/"  # this keeps the newline
dir_ends_with_newline
/
$ echo "$(dirname "$foo")/"  # this drops the newline
dir_ends_with_newline/
```

mguaypaq: I wondered about whether to do that, then had trouble convincing myself whether these settings are inherited by subshells, then thought that I don't want to inflict that question on future readers of the code. Also, I don't want to encourage doing more stuff outside of the `install_sct_log.txt`-generating subshell. I'm open to other arguments, though.

install_sct
mguaypaq: Thanks for the context! That will definitely save me some time.

install_sct
mguaypaq: Good point, I added `>/dev/null` to both.

install_sct
joshuacwnewton: Sounds reasonable to me! Thank you for sharing your perspective here. :)

joshuacwnewton (approved):
Thanks for taking the time to use `shellcheck` here, btw! I appreciate that you've got experience with bash wizardry. :stuck_out_tongue_winking_eye: 

I'm pre-approving here just because the last suggestion I have is so tiny.

install_sct
```diff
@@ -744,5 +744,6 @@ else
 fi
```
joshuacwnewton, line 745: Just leaving a common to acknowledge -- it's a bit weird that the Ubuntu 20.04 CI runners keep stalling out during the test suite!

The most recent run [on `master`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/1750446931) looks fine? So, maybe it was a one-off glitch. I'll try restarting the job on this PR, but... :thinking:

joshuacwnewton: Weird... I restarted the job and it worked fine. Yet, on a separate PR (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3638), this behavior cropped up again just now! So, it looks like it might be intermittent?

install_sct
```diff
@@ -744,3 +744,6 @@ else
...
 
 ) 2>&1 | tee install_sct_log.txt
+status=$?
```
joshuacwnewton, line 747: Will `$?` propagate the exit status of the subshell, or will it just propagate the exit status of the following `tee` command?

mguaypaq: Good catch, thanks! And as a result, TIL about `${PIPESTATUS[0]}` (see [this stackexchange question](https://stackoverflow.com/questions/1221833/pipe-output-and-capture-exit-status-in-bash#1221870)). I'll use that instead.

joshuacwnewton: Oh, good find!! :heart:

install_sct
mguaypaq: Yes, I just added a comment on #3661 with my observations.

PR #3677, opened 2022-02-02
Stop appending `.nii.gz` to the centerline output filepath if `-o` has an extension

### LINKED ISSUE #3530: Additional extension in output file
### Description

When running something like this:
~~~
 sct_get_centerline -i sag_cervical_seg.nii.gz -method fitseg -o sag_cervical_centerline.nii.gz
~~~

I expect output file to be `sag_cervical_centerline.nii.gz`, but instead it is `sag_cervical_centerline.nii.gz.nii.gz`

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR adds a failing test to reproduce the behavior in #3530, then changes the behavior of the script to stop appending `.nii.gz` (which makes the test pass).

## Linked issues

Fixes #3530.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_get_centerline.py
+++ b/spinalcordtoolbox/scripts/sct_get_centerline.py
@@ -154,7 +154,7 @@ def main(argv=None):
         file_output = os.path.join(path_data, file_data+ext_data)
     else:
         path_data, file_data, ext_data = extract_fname(fname_data)
-        file_output = os.path.join(path_data, file_data+'_centerline'+ext_data)
+        file_output = os.path.join(path_data, file_data+'_centerline.nii.gz')
 
     if method == 'viewer':
         # Manual labeling of cord centerline
--- a/spinalcordtoolbox/scripts/sct_get_centerline.py
+++ b/spinalcordtoolbox/scripts/sct_get_centerline.py
@@ -175,8 +175,6 @@ def main(argv=None):
     im_centerline.save(file_output)
     np.savetxt(file_output + '.csv', arr_centerline.transpose(), delimiter=",")
 
-    display_viewer_syntax([fname_input_data, file_output], colormaps=['gray', 'red'], opacities=['', '1'])
-
     path_qc = arguments.qc
     qc_dataset = arguments.qc_dataset
     qc_subject = arguments.qc_subject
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -172,20 +172,20 @@ def main(argv=None):
...
     qc_dataset = arguments.qc_dataset
     qc_subject = arguments.qc_subject
 
     # Generate QC report
     if path_qc is not None:
-        generate_qc(fname_input_data, fname_seg=file_output + '.nii.gz', args=sys.argv[1:], path_qc=os.path.abspath(path_qc),
+        generate_qc(fname_input_data, fname_seg=file_output, args=sys.argv[1:], path_qc=os.path.abspath(path_qc),
                     dataset=qc_dataset, subject=qc_subject, process='sct_get_centerline')
-    display_viewer_syntax([fname_input_data, file_output + '.nii.gz'], colormaps=['gray', 'red'], opacities=['', '0.7'])
+    display_viewer_syntax([fname_input_data, file_output], colormaps=['gray', 'red'], opacities=['', '0.7'])
```
joshuacwnewton, line 188: Looking at the changes again, I'm realizing that we're actually duplicating our call to `display_viewer_syntax` unnecessarily. 

Now is probably a good time to remove this, aha. :sweat_smile:

joshuacwnewton: Done in a6c8728.

RignonNoel (commented):
LGTM, however it change the interface of the script, so I don't know if it's really what you want since it's a breaking change.

For me there is two option:

1. Allow people to put it with OR without extension to support the old interface (we could use the ext of the input file by default if there is none)
2. Follow a deprecation process (Idk if we have one ?)

RignonNoel:
> We actually just want to make sure we always output compressed NiFTI files, even if the input data file is `.nii`.

@joshuacwnewton But what about the case where the user give a `-o myfile.nii` ? Actually you're creating a `.nii` and not a `.nii` even if your output is comrpessed.

Should we not just remove the extension of the `-o` if there is one and export always as a `.nii.gz` ?

Or... We do a breaking change by adding an error message if somebody give an ext in the `-o` param

joshuacwnewton:
Apologies, my commit message was unclear! :sweat_smile: 

> We actually just want to make sure we always output compressed NiFTI files, even if the input data file is `.nii`.

What I should have said was:

* If the user doesn't specify the output extension (either by not providing `-o`, or by providing `-o filename`), then default to `.nii.gz`.
* If the user _does_ specify the output extension (either by `filename.nii` or `filename.nii.gz`), then we should respect their choice. (Just in case the user wants uncompressed NIfTI files.)

RignonNoel:
Ok, in this case we are on the same page and everything seems good for me.

Issue #3672, opened 2022-01-27
`np.genfromtxt` reads SCT version number as value during stable releases


joshuacwnewton:
This is a tiny bug, and only affects the test suite (and not one that actually affects our processing). But, running `sct_testing` for a stable release will cause the following error:

```
=================================== FAILURES ===================================
_________________ test_sct_extract_metric_against_groundtruth __________________
Traceback (most recent call last):
  File "/home/joshua/sct_5.5/testing/cli/test_cli_sct_extract_metric.py", line 20, in test_sct_extract_metric_against_groundtruth
    assert results == pytest.approx([176.1532, 32.6404, 11.4573], abs=0.001)  # Size[vox], WA, and STD respectively
AssertionError: assert array([  5.5 ... 11.4573414 ]) == approx([176.1...73 ± 1.0e-03])
  +array([  5.5       , 176.15326674,  32.64041623,  11.4573414 ])
  -approx([176.1532 ± 1.0e-03, 32.6404 ± 1.0e-03, 11.4573 ± 1.0e-03])
=========================== short test summary info ============================
FAILED ../../sct_5.5/testing/cli/test_cli_sct_extract_metric.py::test_sct_extract_metric_against_groundtruth
ERROR ../../sct_5.5/testing/cli/test_cli_sct_extract_metric.py::test_sct_extract_metric_against_groundtruth
=================== 1 failed, 38 warnings, 1 error in 0.57s ====================
```

The issue here is with this test:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fc34ee93cf0073a2abaca9f786aa0f794101e0c4/testing/cli/test_cli_sct_extract_metric.py#L13-L20

When `np.genfromtxt` is used to read from the output .csv file from the test (repeated here):

|Timestamp          |SCT Version|Filename                                                |Slice (I->S)|VertLevel|DistancePMJ|Label       |Size [vox]        |WA()              |STD()             |
|-------------------|-----------|--------------------------------------------------------|------------|---------|-----------|------------|------------------|------------------|------------------|
|2022-01-27 15:50:40|5.5        |/home/joshua/sct_5.5/data/sct_testing_data/mt/mtr.nii.gz|1:2         |         |           |white matter|176.15326673992857|32.640416229403456|11.457341402914249|

The "SCT Version" cell value `5.5` is interpreted as a numerical value instead of a string.

----

[...]

PR #3678, opened 2022-02-02
Test release path in CI

### LINKED ISSUE #3669: Consider testing stable release install path in CI
Part of why the bug in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3667 slipped through is that the "stable release" installation path isn't actively tested at the moment.

Fixing this could be done quite simply, I think. We could just add a switch to the pre-existing `create-release.yml` workflow that follows all of the steps right up until committing the files to the `release` branch. Then, instead of creating a new release, it simply tries re-installing SCT. 

While not _super_ thorough, it would give us some peace of mind about the "stable release" installation path, as well as help to ensure that the `create-release.yml` workflow itself is working as intended, too.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Draft PR while I test whether or not this works. Will update once ready.

## Linked issues

Fixes #3669.


joshuacwnewton:
Side-comment: It looks like codecov is showing that coverage decreased for this change. But, this change affects only a GitHub Actions `.yml` file. So, I wonder if there's a way to specify that `codecov` should only be looking at the `./spinalcordtoolbox` package directory, rather than the full SCT repo. :thinking:

.github/workflows/create-release.yml
```diff
@@ -45,23 +57,52 @@ jobs:
...
+      # Relevant documentation: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files
+      # This workaround should be replaced by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3198#discussion_r568225392
+
+    - name: Run sct_check_dependencies
+      run: |
+        sct_check_dependencies
+
+    - name: Run sct_testing
+      run: |
+        sct_testing
```
joshuacwnewton, line 88: Hmm. I'm surprised that https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3672 isn't triggered by this CI run. I assumed that if `sct_testing` was called post-install from `release`, then this issue would crop up. :thinking: 

I'll look into this and see what I can find.

joshuacwnewton: Update: While I'm not sure if this is the cause of the issue, I did find one bug in `install_sct`. Currently, to determine whether this is a `release` installation, we use the following logic: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f068452c43e19669faca45c222f3d6fca60ca743/install_sct#L412-L420

Checking for `.git` works if you're downloading the source code from the Releases page, but _doesn't_ work if you've checked out the `release` branch (which we do in this GitHub Workflow). So, I think it might be more reliable to check for the presence of `requirements-freeze.txt`, which is the only release-specific file.

This could have some effect on whether the tests are truly being run from the "release" branch as I hoped.

joshuacwnewton: Ah, and fixing the problem in this branch does nothing, because `create-release.yml` checks out `master` first. :laugh:

joshuacwnewton: Excellent! I switched the main branch to this PR, and now the test fails as expected.

This gives us some assurance that #3672 would have been caught by this worklflow, and that we can catch any other release-related issues ahead of time.

joshuacwnewton: I think I'm going to test a fix for #3672 in this branch for convenience sake, since we have the CI ready for debugging purposes, and merging this PR as-is will just cause a failure on `master`.

.github/workflows/create-release.yml
```diff
@@ -6,10 +6,29 @@ on:
...
+    # https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onschedule
+    # > Scheduled workflows run on the latest commit on the default or base branch
+    # i.e. this can only run on master
+    - cron: '0 11 * * *'
 
 env:
-  MAIN_BRANCH: "master"
+  # FIXME: If you see this comment in production, revert to "master" ASAP.
+  # --> Temporarily switched to PR branch to demo an `install_sct` fix.
+  MAIN_BRANCH: "jn/3669-test-release-path-in-CI"
```
joshuacwnewton, line 24: TODO: Remember to revert this back to `master` once this PR is ready to merge.

.github/workflows/create-release.yml
joshuacwnewton: Ok, I've got a fix up in [93cc464](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3678/commits/93cc4640faf39de501baf933f292ef0b69667a17). :)

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -544,7 +544,7 @@ def save_as_csv(agg_metric, fname_out, fname_in=None, append=False):
 
     # populate data
     with open(fname_out, 'a') as csvfile:
-        spamwriter = csv.writer(csvfile, delimiter=',')
+        spamwriter = csv.writer(csvfile, delimiter=',', quoting=csv.QUOTE_NONNUMERIC)
```
mguaypaq, line 547: That a useful option! I'll note that `delimiter=','` is the default already, but it doesn't hurt to be explicit.

.github/workflows/create-release.yml
```diff
@@ -6,10 +6,29 @@ on:
       milestone_title:
         description: 'Milestone title'
         required: true
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
```
mguaypaq, line 14: I'm not sure I understand [the docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-activity-types-and-filters-with-multiple-events) completely, but it looks like this will trigger the "Create release" workflow on every merge into master and on every PR, **in addition** to every day at 11am? Is there a reason for wanting both?

.github/workflows/create-release.yml
```diff
@@ -45,23 +65,52 @@ jobs:
         echo "# Platform-specific torch requirements (See SCT Issue #2745)" >> requirements-freeze.txt
         grep "torch" requirements.txt >> requirements-freeze.txt
 
-    - name: Update version.txt
+    - name: Update version.txt (using test value for CI)
+      run: |
+        echo "0.0" > spinalcordtoolbox/version.txt
+      # Only attach to release if workflow is run manually. (This allows the workflow to double as a PR test.)
```
mguaypaq, line 71: Since this condition is the opposite of the condition below, I assume the comment should also be the opposite? Or else I don't quite understand the phrase "only attach to release".
```suggestion
      # Only attach to release if workflow is run automatically. (This allows the workflow to double as a PR test.)
```

mguaypaq (approved):
So, even though this workflow creates a merge-commit when it's run automatically, that's ok because it doesn't get pushed to the repo in that case.

This seems like a fine way to do it. I'm marking this as approved, assuming you explain and/or fix the triggers and that TODO comment you left before merging.

.github/workflows/create-release.yml
mguaypaq: To clarify, I was suggesting that maybe the daily run itself is sufficient (but I haven't done releases before, so @joshuacwnewton's opinion counts for a lot more!). I agree that regularly checking dependencies is good!

joshuacwnewton: In 99% of cases, the "merge into master" run is probably overkill, yes! 

* Our PRs already have a requirement for being up to date with master, so we already effectively test the build on master right before the PR is merged, anyway.
* If somehow there ends up being some quirk or issue that would only reveal itself on `master` branch specifically, then we would see at most a ~24h delay before finding out that something had broke (because of the cron job).

At the same time, maybe that one extra run is still useful feedback? Say we merge a PR at noon, then notice the build fails at 1pm, and then fix it by 2pm... it's probably an unlikely scenario, but it would save us a day's worth of delay. :thinking:

.github/workflows/create-release.yml
joshuacwnewton: Good catch! This was a copy-paste that I forgot to update. :sweat_smile: 

In fact, we're not even attaching the file to the release at all, so the whole comment is misleading. It would probably be better as something like:

```suggestion
      # If the workflow *hasn't* been run manually, then it's being run as a PR test, so use a dummy value instead
```

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Ah, good to know! 

Honestly, this section could be cleaned up a little more (e.g. the fact that we use `spamwriter`, which I think is copied and pasted right out of a Python tutorial...) but I wanted to keep my changes minimal and easy to review.

I'm never quite sure when a good time is to clean up little things like this. :thinking:

.github/workflows/create-release.yml
joshuacwnewton: In fact, this PR is a good example of a case where keeping the `master` check is useful?

Since I had to do a little hacky branch workaround to test the changes in this PR (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3678#discussion_r799585067), I'm still not 100% sure if this workflow will work on `master`. I'm 99% sure they will? But just in case there's something I've missed, it will be nice to get immediate feedback.

Issue #3633, opened 2021-12-19
Wrong measures for lesion length


jcohenadad:
### Description

Estimation of lesion length can be wrong. 

This issue was raised by a user on the forum: https://forum.spinalcordmri.org/t/negative-values-with-sct-analyze-lesion/794

### Steps to Reproduce

```console
sct_download_data -d sct_testing_data
cd sct_testing_data/t2
# create fake lesion spanning 4 voxels
sct_label_utils -i t2.nii.gz -create 29,27,25,1:29,28,25,1:29,29,25,1:29,30,25,1 -o lesion.nii.gz
# run morphometrics
sct_analyze_lesion -m lesion.nii.gz -s t2_seg-manual.nii.gz -i t2.nii.gz 
```

**Expected behavior:**
~~~
  Volume [mm^3] = 4.0+/-0.0
  (S-I) Length [mm] = 4.0
  Equivalent Diameter [mm] = 1
~~~

**Actual behavior:** <What actually happens>
~~~
  Volume [mm^3] = 4.0+/-0.0
  (S-I) Length [mm] = -0.93+/-0.0
  Equivalent Diameter [mm] = -0.0+/-0.0
~~~

### Possible issues

I suspect the issue is related to a wrong estimation of objects that are one voxel wide. Deeper investigations are warranted, and appropriate unit tests should be developed.

### Bonus changes

We could also take this opportunity to output CSV instead of XSL, and PKL

mguaypaq:
The immediate cause seems to be [this line in the script](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f068452c43e19669faca45c222f3d6fca60ca743/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L443), which wrongfully converts an angle from radians to degrees:
```python3
            self.angles[iz] = math.degrees(angle)
```
I'll investigate more to check that the computation is correct, but already just fixing this seems to give the right lesion length:
```
Averaged measures...
  Volume [mm^3] = 4.0+/-0.0
  (S-I) Length [mm] = 4.0+/-0.0
  Equivalent Diameter [mm] = 0.56+/-0.0

Total volume = 4.0 mm^3
Lesion count = 1
```
I haven't yet looked into the "Equivalent Diameter" computation, which says 0.56 instead of 1, as in the "expected behavior" in the issue description.

mguaypaq:
As for changing the output from XSL and PKL to CSV:
[...]

PR #3681, opened 2022-02-04
Fix computation of estimated lesion length and diameter

mguaypaq:
#### GitHub

#### PR contents

## Description

The estimated lesion length and diameter are currently computed incorrectly, because:
- the code wrongly converts some angles from radians to degrees, and
- it uses the wrong formula to compute the radius from the area of a circle.

The angles in question are the angle between the centerline of the spine and a vertical vector at each slice (in the S-I direction). They are computed using `np.arccos`, which works in radians, and only used by `np.cos`, which also works in radians. (Come to think of it, maybe we could skip both `arccos` and `cos`, since they're inverse functions, but then the array couldn't be called something nice and understandable like `angles`.)

Since these are basic computational errors, it's not clear what an appropriate unit test would be. Maybe just the reproduction steps from the linked issue?

Also, I intend to go through the file and apply many little fixes (fix typos, simplify a few computations, etc.), but that should be in a separate refactoring PR. This PR is just a minimal fix to get the formulas correct.

## Linked issues

Part of #3633

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -9,7 +9,6 @@
 # About the license: see the file LICENSE.TXT
 
 import os
-import math
 import sys
 import pickle
 import shutil
@@ -439,8 +438,7 @@ class AnalyzeLeion:
                 [x_centerline_deriv[iz] * px, y_centerline_deriv[iz] * py, pz]))
 
             # compute the angle between the normal vector of the plane and the vector z
-            angle = np.arccos(np.vdot(tangent_vect, np.array([0, 0, 1])))
-            self.angles[iz] = math.degrees(angle)
+            self.angles[iz] = np.arccos(np.vdot(tangent_vect, np.array([0, 0, 1])))
 
     def label_lesion(self):
         printv('\nLabel connected regions of the masked image...', self.verbose, 'normal')
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -255,7 +255,7 @@ class AnalyzeLeion:
 
     def _measure_diameter(self, im_data, p_lst, idx):
         area_lst = [np.sum(im_data[:, :, zz]) * np.cos(self.angles[zz]) * p_lst[0] * p_lst[1] for zz in range(im_data.shape[2])]
-        diameter_cur = 2 * np.sqrt(max(area_lst) / (4 * np.pi))
+        diameter_cur = 2 * np.sqrt(max(area_lst) / np.pi)
         self.measure_pd.loc[idx, 'max_equivalent_diameter [mm]'] = diameter_cur
         printv('  Max. equivalent diameter : ' + str(np.round(diameter_cur, 2)) + ' mm', self.verbose, type='info')
 


mguaypaq:
Also for the record, manually running the reproduction steps now gives me the correct lesion length, and a sensible diameter:
```
Averaged measures...
  Volume [mm^3] = 4.0+/-0.0
  (S-I) Length [mm] = 4.0+/-0.0
  Equivalent Diameter [mm] = 1.13+/-0.0

Total volume = 4.0 mm^3
Lesion count = 1
```
The diameter isn't just `1.0`, as you might expect for a 1-voxel-wide lesion, but that's because voxels are cubes, and cylinders are not cubes. `1.13` is approximately `2/sqrt(pi)`.

joshuacwnewton (commented):
I notice the test for `sct_analyze_lesion` is... quite awful! 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f068452c43e19669faca45c222f3d6fca60ca743/testing/cli/test_cli_sct_analyze_lesion.py#L13-L16

It uses the same spinal cord segmentation file for both `-s` (the segmentation file) but also `-m` (the _lesion mask_), so the test can't possibly be doing anything useful! So, I'm thinking it might be good to fix up the test for this PR, too. 

I think I'd like to take that on before I review your changes? I've never actually worked on `sct_analyze_lesion` before. So, my thought process is: if I spend the time learning how to write a good test for this script, then it'll force me to figure out what the script is actually doing, and make my review a lot more effective.

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -1,16 +1,71 @@
...
+                                  '-ofolder', str(tmp_path)])
+
+    # Load analysis results from pickled pandas.Dataframe
+    _, fname, _ = extract_fname(path_lesion)
+    with open(tmp_path/f"{fname}_analyzis.pkl", 'rb') as f:
+        measurements = pickle.load(f)['measures']
+
+    # Validate analysis results
+    for key, expected_value in expected_measurements.items():
+        assert(measurements.at[0, key] == expected_value)
```
joshuacwnewton, line 71: There's a slight snag here. For the `length [mm]` measurement, there's a mismatch between what gets logged in the console, and the _actual_ value.

Log:

```
Averaged measures...
  Volume [mm^3] = 4.0+/-0.0
  (S-I) Length [mm] = 4.0+/-0.0
  Equivalent Diameter [mm] = 1.13+/-0.0

Total volume = 4.0 mm^3
Lesion count = 1
```

Actual value:

Screenshot from 2022-02-05 10-44-43

This is why the test fails. But, I'm pretty sure this is an issue with the script, rather than the test.

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -1,16 +1,71 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture()
+def dummy_lesion(tmp_path):
+    """Define a fake voxel lesion using the specified dimensions."""
+    # TODO: Parametrize this fixture to test other lesion dimensions
+    starting_coord = [29, 27, 25]
+    dim = [1, 4, 1]
```
joshuacwnewton, line 18: Also, I tried to write the fixture in such a way that it should be easy to play around with the dimensions to test lesions of differing sizes. :)

joshuacwnewton (commented):
I took the `1x4x1` lesion example from the issue and turned it into a test!

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -1,16 +1,71 @@
...
+        'volume [mm3]': dim[0] * dim[1] * dim[2],
+    }
+
+    return path_out, measurements
+
+
 @pytest.mark.sct_testing
-@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
-def test_sct_analyze_lesion_output_file_exists():
+def test_sct_analyze_lesion_output_file_exists(dummy_lesion, tmp_path):
```
joshuacwnewton, line 56: Whoops. Forgot to rename the test.

```suggestion
def test_sct_analyze_lesion_results_with_dummy_data(dummy_lesion, tmp_path):
```

testing/cli/test_cli_sct_analyze_lesion.py
mguaypaq: Yes, I think this is because the logged values are rounded ([using `np.round` here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f0b244cd174463a3825620bc488767c409beccdc/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L219)).

As a separate point, it's a bit alarming that the uncertainties as `+/-0.0`. Surely they should be rounded up to `0.1`?

Based on the computations in the script, we expect:
- the volume to be exactly 4.0, aka the number of voxels in the lesion;
- the length to be a little bit more than 4.0, because it's measured along the curve of the spinal cord centerline (if I understand correctly); if so, that's another bug uncovered, and [this line](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f0b244cd174463a3825620bc488767c409beccdc/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L251) should be _dividing_ by the cosine, rather than multiplying;
- the diameter to be a little bit less than 2/sqrt(pi) = 1.1283791670955126, again because there's a correction factor that takes into account the spinal cord centerline.

joshuacwnewton: Hrmm. I tried playing with some different lesion sizes and got consistently less than the expected length, regardless of where I placed the lesion. For reference, here is the t2 image, centerline, and a sample lesion I tried:

Screenshot from 2022-02-08 13-18-55

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -439,8 +437,7 @@ def angle_correction(self):
                 [x_centerline_deriv[iz] * px, y_centerline_deriv[iz] * py, pz]))
 
             # compute the angle between the normal vector of the plane and the vector z
-            angle = np.arccos(np.vdot(tangent_vect, np.array([0, 0, 1])))
-            self.angles[iz] = math.degrees(angle)
+            self.angles[iz] = np.arccos(np.vdot(tangent_vect, np.array([0, 0, 1])))
```
joshuacwnewton, line 440: Just double-checked, and your change makes sense to me. Since `self.angles` is eventually consumed by `np.cos`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6f388046fa7d21d4315513eba8a2ec0b94267adf/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L251

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6f388046fa7d21d4315513eba8a2ec0b94267adf/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L256

With your change, we're basically going from:

* Before: `np.arccos` (rad) -> `math.degrees` (deg) -> `np.cos` [(expects rad)](https://numpy.org/doc/stable/reference/generated/numpy.cos.html)
* After: `np.arccos` (rad) -> `np.cos` (expects rad)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -256,7 +254,7 @@ def _measure_length(self, im_data, p_lst, idx):
 
     def _measure_diameter(self, im_data, p_lst, idx):
         area_lst = [np.sum(im_data[:, :, zz]) * np.cos(self.angles[zz]) * p_lst[0] * p_lst[1] for zz in range(im_data.shape[2])]
-        diameter_cur = 2 * np.sqrt(max(area_lst) / (4 * np.pi))
+        diameter_cur = 2 * np.sqrt(max(area_lst) / np.pi)
```
joshuacwnewton, line 257: This change also makes sense to me, as it lines up with what I wrote in my test:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6f388046fa7d21d4315513eba8a2ec0b94267adf/testing/cli/test_cli_sct_analyze_lesion.py#L40-L46

My best guess is that the original `4 * pi` equation was derived from the formula for the area of a sphere?

```
   a = 4*pi*r^2 
-> a = 4*pi*(d/2)^2 
-> d = 2*sqrt(a/4*pi) 
```

But, the `max(area_lst)` used in this formula is computed using the 2D area, _not_ the 3D area:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6f388046fa7d21d4315513eba8a2ec0b94267adf/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L256

So, I agree that your change makes sense.

joshuacwnewton (approved):
Giving this a soft approval, but wanting to wait for what comes of your chat with @charleygros before merging, just in case my understanding was off at all.

mguaypaq:
After discussion with @charleygros, he confirms that the lesion length should be computed by _dividing_ by the cosine, rather than multiplying. In the original example, which is 4 voxels tall, the computed lesion length should be slightly _larger_ than 4.0, which is what I now get (4.001988).

As a sanity check, this is consistent with the diameter computation: there, we compute the cross-sectional area, and we adjust for the centerline angle by _multiplying_ by the cosine. If we multiply the length by the cross-sectional area, the two angle adjustments cancel out, which they should, because a rotation of the lesion should keep the volume constant.

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton: Resolved by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3681#issuecomment-1035562858.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -250,13 +248,13 @@ def _measure_volume(self, im_data, p_lst, idx):
         printv('  Volume : ' + str(np.round(vol_tot_cur, 2)) + ' mm^3', self.verbose, type='info')
 
     def _measure_length(self, im_data, p_lst, idx):
-        length_cur = np.sum([np.cos(self.angles[zz]) * p_lst[2] for zz in np.unique(np.where(im_data)[2])])
+        length_cur = np.sum([p_lst[2] / np.cos(self.angles[zz]) for zz in np.unique(np.where(im_data)[2])])
         self.measure_pd.loc[idx, 'length [mm]'] = length_cur
         printv('  (S-I) length : ' + str(np.round(length_cur, 2)) + ' mm', self.verbose, type='info')
 
     def _measure_diameter(self, im_data, p_lst, idx):
         area_lst = [np.sum(im_data[:, :, zz]) * np.cos(self.angles[zz]) * p_lst[0] * p_lst[1] for zz in range(im_data.shape[2])]
```
joshuacwnewton, line 256: I noticed you changed `_measure_length` from `np.cos(self.angles[zz]) * p_lst[2]` to `p_lst[2] / np.cos(self.angles[zz])`.

Is there any chance the same needs to be done for the `_measure_diameter` instance of `np.cos`, too?

joshuacwnewton, line 256: Took me a little while to wrap my head around the angle correction stuff. Had to visualize? Especially to understand why the angle was originally computed via `arccos(tangent_vector <dot> unit_z_vector)`.

Screenshot from 2022-02-11 11-29-44

But this makes sense to me now, I think. I can also adjust the text to confirm this behavior, too.

joshuacwnewton (approved):
LGTM! Ready to merge once I push my changes to the test, I think.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Done in [1908f1a](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3681/commits/1908f1a2fe226567e709f5900fb601c8a0b4cd55).

jcohenadad:
Thank you for this contribution @mguaypaq !

PR #3683, opened 2022-02-09
Add studies

alexfoias:
Requested on https://neuropoly.slack.com/archives/CB27THD2T/p1644359291940379 & https://neuropoly.slack.com/archives/CB27THD2T/p1644422371821379

### DIFF
--- a/documentation/source/overview/studies.rst
+++ b/documentation/source/overview/studies.rst
@@ -8,6 +8,8 @@ The following studies have used SCT:
 2022
 ----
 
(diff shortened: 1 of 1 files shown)


jcohenadad (approved):

joshuacwnewton:
> @joshuacwnewton can we merge without it knowing we have a problem actually on this script ?

Yes, we can. If you see issues with `batch_processing.sh`, it is 99% likely to be due to # issue is due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597.

I think it might be best to set the test as "non-required" for now, so that we can monitor the results without it blocking PR merges.

PR #3684, opened 2022-02-09
Jupyter notebooks to aid SCT classes/workshops, and for users without  CLI background

etetteh:
@jcohenadad for your review, please.

### DIFF
--- a/sct_tutorial/SCT_Tutorial_01_Segmentation.ipynb
+++ b/sct_tutorial/SCT_Tutorial_01_Segmentation.ipynb
@@ -1,0 +1,445 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ac33cc75",
+   "metadata": {},
+   "source": [
+    "# [Spinal Cord Toolbox (SCT)](https://spinalcordtoolbox.com/index.html)\n",
+    "\n",
+    "<img src=\"https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/overview.png\" align=\"center\" width=\"750px\">"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f72f60fa",
+   "metadata": {},
+   "source": [
+    "## Installation Instructions\n",
+    "The instructions below are for MacOS and Linux platforms. \\\n",
+    "For more installation instructions, such as installing on Windows or Docker, check [here](https://spinalcordtoolbox.com/user_section/installation.html).\n",
+    "\n",
+    "### Run the following (starting with $) in your terminal\n",
+    "Skip if you already have SCT and FSLeyes installed."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d2688852-4d7a-44ab-a205-5685f9aab3f4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# First we'll clone the SCT repository from GitHub \n",
+    "$git -C spinalcordtoolbox pull || git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git spinalcordtoolbox\n",
+    "\n",
+    "# Change our directory to the newly created directory\n",
+    "$cd spinalcordtoolbox\n",
+    "\n",
+    "#start installation\n",
+    "$./install_sct -y"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "af21fb47-3348-4571-a272-9494f159efb5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install the following\n",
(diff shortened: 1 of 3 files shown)


sct_tutorial/SCT_Tutorial_01_Segmentation.ipynb
```diff
@@ -0,0 +1,445 @@
+{
```
joshuacwnewton, line 1: ## Cons

* Difficult to maintain
    * Jupyter notebooks are tricky to version-track using Git
    * This would add a fouth source of truth for the tutorials (in addition to the `.rst` pages, the `batch_single_subject.sh` script, and the SCT Course slides). Making a change to the tutorials would now require keeping even more things in sync with one another.
 * Might make using SCT more complicated for users
    * Users would now have extra options for what the "proper" way to use SCT is (too many choices --> confusion)
    * Documentation might be more challenging to navigate for users (Which page/section do we put these notebooks in? Will users know to go to the tutorial pages, or the notebooks?)
    * They will likely have to learn CLI skills anyway if they want to fully take advantage of SCT (e.g. batch scripting), so providing this resource might be delaying the inevitable for them

joshuacwnewton, line 1: ## Alternate proposal: A single notebook with commands only

As a way to keep all of the "pros" of this idea, but mitigate the "cons", what if we just had a single notebook containing no theory, and instead just the commands as well as section titles? 

Because it would be barebones and simplified, it would be much easier to maintain, and effectively serve the same purpose as the existing [`batch_single_subject.sh`](https://github.com/spinalcordtoolbox/sct_tutorial_data/blob/master/single_subject/batch_single_subject.sh) script in `sct_tutorial_data`. It could even replace that script!

That way, instead of copying and pasting commands from the script into the terminal (like we did in the last SCT Course), we would execute the commands one by one directly in the notebook. But, the overall flow of the course would stay the same (using the slide deck, pausing for "hands-on" portions, etc).

joshuacwnewton (commented):
Thank you so much for taking the effort to convert these tutorials! I think this is a very intriguing idea, and raises a lot of interesting questions.

I have a few thoughts on this -- I've organized them into pros/cons in the comments below to try to kickstart discussion.

jcohenadad:
a quick addition to @joshuacwnewton 's comment:

> Difficult to maintain

I agree with this. As an example, we used to have notebooks in https://axondeepseg.readthedocs.io/, but it was a real pain to maintain them, and at the end we decided to get rid of them.

etetteh:
I agree with the proposal, and the "pros" are the main reasons for the design of these notebooks. I might not have figured in the "pros" during the design, especially the versioning.

Furthermore, I am open to looking into the proposal to modify these notebooks, and to inform subsequent ones. 
Thank you, and let me know if you or anyone has specific design ideas.

joshuacwnewton:
> I agree with the proposal, and the "pros" are the main reasons for the design of these notebooks. I might not have figured in the "cons" during the design, especially the versioning.

In the future, for efforts like this, I would recommend [opening an issue](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/new/choose) on SCT's GitHub repo if you have an idea for something you'd like to work on. That way, the team can discuss beforehand whether the idea is something we want to spend time/effort implementing and maintaining in the first place. It also provides a way for the team to talk about design details before diving into the work itself.

> Furthermore, I am open to looking into the proposal to modify these notebooks, and to inform subsequent ones.
Thank you, and let me know if you or anyone has specific design ideas.

I can't speak for @jcohenadad, but I'm personally leaning towards pausing this effort for now, but revisiting it in advance of the next SCT Course. The reason I suggest pausing the effort is because of how far away in time the next course will be. (The next Martinos workshop is [scheduled for 2023](https://education.martinos.org/home/martinos-spinal-cord-workshop/), and who knows what could change inside SCT between now and then). When that time comes, though, we'll likely have several meetings ahead of time to prepare course materials, and I think your approach could go a long way towards solving https://github.com/spinalcordtoolbox/sct_tutorial_data/issues/7. 

Also, I very much appreciate the work you did troubleshooting the installation of SCT for Jupyter notebooks! If we do decide to adopt Jupyter notebooks in the future, this will save us some trouble, since we already have a proof of concept we can revisit.

etetteh:
Thank you @joshuacwnewton,  and I agree with pausing it for now. 
I've also taken note of opening an issue first to brainstorm ideas.

joshuacwnewton:
One other small comment: I notice that you created this pull request from the `master` branch of your fork of SCT.

In the future, I would recommend instead [creating a branch](https://www.atlassian.com/git/tutorials/using-branches/) for your work, that way your local copy of `master` stays clean and unmodified.

joshuacwnewton:
Temporarily closing this PR, with the intent to revisit it in the future. :)

PR #3700, opened 2022-02-22
Refactor `-param` parsing to address several bugs

### LINKED ISSUE #3698: `gaussian_std` parameter is ignored
The script `sct_label_vertebrae` accepts a parameter call `gaussian_std`, with default value `1.0`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L44
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L185-L188

However, the library code ignores this value, and hard-codes the value `999` instead:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/vertebrae/core.py#L164-L168

I can think of a few ways to address this discrepancy:

1. Replace the hard-coded value by `param.gaussian_std` in the library code, keep the default value of `1.0`. This would be a change in behaviour, probably not what we want.
2. Replace the hard-coded value by `param.gaussian_std` in the library code, update the default value to `999`.
3. Get rid of the `gaussian_std` parameter, if it's not being used. We can still accept the command-line parameter for backwards compatibility, and emit a warning that it's ignored and deprecated.

From looking at #1422, it looks like `gaussian_std` has never been taken into account, and from looking at #1423, it looks like it was already obsolete 5 years ago, so I'm leaning towards option 3.

mguaypaq:
#### PR contents

## Description

This PR began with the `gaussian_std` parameter in `-param`. This parameter has been silently ignored since it was introduced, and `sct_label_vertebrae` uses `gaussian_std=999` instead.

Getting rid of the Gaussian altogether should have minimal impact, since using 999 is very close to 1.0:
```py3
>>> from scipy.signal import gaussian
>>> gaussian(20*2, std=20*999)[:20]
array([0.99999952, 0.99999957, 0.99999962, 0.99999966, 0.9999997 ,
       0.99999974, 0.99999977, 0.9999998 , 0.99999983, 0.99999986,
       0.99999989, 0.99999991, 0.99999993, 0.99999995, 0.99999996,
       0.99999997, 0.99999998, 0.99999999, 1.        , 1.        ])
```

However, in the course of this PR, several additional bugs were discovered within `-param`:

* Any comma-separated parameters passed to `-param` beyond the first would be ignored.
* The `laplacian` parameter accidentally overwrote the `laplacian` _function_ due to a name collision.
* Several unrelated parameters were grouped into `-param` for no particular reason.
* Errors with `-param` parsing would not display the argparse `-help`

So, `-param` has been refactored further to address these issues.

## Linked issues

Fixes #3698

### DIFF
--- a/spinalcordtoolbox/scripts/sct_label_vertebrae.py
+++ b/spinalcordtoolbox/scripts/sct_label_vertebrae.py
@@ -55,6 +55,7 @@ class Param:
             if key in ['shift_AP', 'size_AP', 'size_RL', 'size_IS', 'shift_AP_visu']:
                 setattr(self, key, value)
             elif key == 'gaussian_std':
+                # TODO: remove 'gaussian_std' completely for release 6.0
                 printv('WARNING: gaussian_std parameter is currently ignored, '
                        'and will be removed in a later version.', 1, type='warning')
             else:
--- a/spinalcordtoolbox/scripts/sct_label_vertebrae.py
+++ b/spinalcordtoolbox/scripts/sct_label_vertebrae.py
@@ -31,39 +31,33 @@ from spinalcordtoolbox.math import threshold, laplacian
 from spinalcordtoolbox.scripts import sct_straighten_spinalcord, sct_apply_transfo
 
 
-# PARAMETERS
-class Param:
-    # The constructor
-    def __init__(self):
-        self.shift_AP = 32  # shift the centerline towards the spine (in voxel).
-        self.size_AP = 11  # window size in AP direction (=y) (in voxel)
-        self.size_RL = 1  # window size in RL direction (=x) (in voxel)
-        self.size_IS = 19  # window size in IS direction (=z) (in voxel)
-        self.shift_AP_visu = 15  # shift AP for displaying disc values
+# for vertebral_detection
+param_default = {
+    'shift_AP': 32,  # shift the centerline towards the spine (in voxel).
+    'size_AP': 11,  # window size in AP direction (=y) (in voxel)
+    'size_RL': 1,  # window size in RL direction (=x) (in voxel)
+    'size_IS': 19,  # window size in IS direction (=z) (in voxel)
+    'shift_AP_visu': 15,  # shift AP for displaying disc values
+}
 
-    # update constructor with user's parameters
-    def update(self, param_user):
-        for param in param_user.split(','):
-            try:
-                key, value = param.split('=', maxsplit=1)
-                value = int(value)
-            except ValueError:
-                printv('ERROR: Wrong usage.', 1, type='error')
-                sys.exit(1)
-            if key in ['shift_AP', 'size_AP', 'size_RL', 'size_IS', 'shift_AP_visu']:
-                setattr(self, key, value)
-            elif key == 'gaussian_std':
-                # TODO: remove 'gaussian_std' completely for release 6.0
-                printv('WARNING: gaussian_std parameter is currently ignored, '
-                       'and will be removed in a later version.', 1, type='warning')
-            else:
-                printv('ERROR: Wrong usage.', 1, type='error')
-                sys.exit(1)
+
+def vertebral_detection_param(string):
+    """Custom parser for vertebral_detection advanced parameters."""
+    param = param_default.copy()
+    for key_value in string.split(','):
+        key, value = key_value.split('=', maxsplit=1)
+        if key in param:
+            param[key] = int(value)
+        elif key == 'gaussian_std':
+            # TODO(issue#3706): remove 'gaussian_std' completely for v5.7
+            printv('WARNING: gaussian_std parameter is currently ignored, '
+                   'and will be removed in a later version.', 1, type='warning')
+        else:
+            raise ValueError(f'Unknown parameter: {key}')
+    return param
 
 
 def get_parser():
-    # initialize default param
-    param_default = Param()
     parser = SCTArgumentParser(
         description=(
             "This function takes an anatomical image and its cord segmentation (binary file), and outputs the "
@@ -180,12 +174,13 @@ def get_parser():
     optional.add_argument(
         '-param',
         metavar=Metavar.list,
-        type=list_type(',', str),
-        help=f"Advanced parameters. Assign value with \"=\"; Separate arguments with \",\"\n"
-             f"  - shift_AP [mm]: AP shift of centerline for disc search. Default={param_default.shift_AP}.\n"
-             f"  - size_AP [mm]: AP window size for disc search. Default={param_default.size_AP}.\n"
-             f"  - size_RL [mm]: RL window size for disc search. Default={param_default.size_RL}.\n"
-             f"  - size_IS [mm]: IS window size for disc search. Default={param_default.size_IS}.\n"
+        type=vertebral_detection_param,
+        default=param_default.copy(),
+        help=f'Advanced parameters. Assign value with "="; Separate arguments with ","\n'
+             f'  - shift_AP [mm]: AP shift of centerline for disc search (default: {param_default["shift_AP"]})\n'
+             f'  - size_AP [mm]: AP window size for disc search (default: {param_default["size_AP"]})\n'
+             f'  - size_RL [mm]: RL window size for disc search (default: {param_default["size_RL"]})\n'
+             f'  - size_IS [mm]: IS window size for disc search (default: {param_default["size_IS"]})\n',
     )
     optional.add_argument(
         '-r',
@@ -235,7 +230,6 @@ def main(argv=None):
     initcenter = ''
     fname_initlabel = ''
     file_labelz = 'labelz.nii.gz'
-    param = Param()
 
     fname_in = os.path.abspath(arguments.i)
     fname_seg = os.path.abspath(arguments.s)
@@ -268,8 +262,6 @@ def main(argv=None):
     if arguments.initlabel is not None:
         # get absolute path of label
         fname_initlabel = os.path.abspath(arguments.initlabel)
-    if arguments.param is not None:
-        param.update(arguments.param[0])
     remove_temp_files = arguments.r
     clean_labels = arguments.clean_labels
 
@@ -414,7 +406,7 @@ def main(argv=None):
 
         # detect vertebral levels on straight spinal cord
         init_disc[1] = init_disc[1] - 1
-        vertebral_detection('data_straightr.nii', 'segmentation_straight.nii', contrast, param, init_disc=init_disc,
+        vertebral_detection('data_straightr.nii', 'segmentation_straight.nii', contrast, arguments.param, init_disc=init_disc,
                             verbose=verbose, path_template=path_template, path_output=path_output, scale_dist=scale_dist)
 
     # un-straighten labeled spinal cord
--- a/spinalcordtoolbox/scripts/sct_label_vertebrae.py
+++ b/spinalcordtoolbox/scripts/sct_label_vertebrae.py
@@ -12,6 +12,7 @@
 
 import sys
 import os
+import argparse
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


mguaypaq:
Also, I don't believe the `verbose=2` plots are currently tested (out of scope for this PR, since they're not currently saved to the right path), but I had a look at the plots generated by the following steps, and they seem fine to me:

```bash
cd $(mktemp -d)
curl -LO https://github.com/spinalcordtoolbox/sct_tutorial_data/releases/download/r20220125/data_spinalcord-segmentation.zip
unzip data_spinalcord-segmentation.zip
cd single_subject/data/t2
sct_deepseg_sc -i t2.nii.gz -c t2 -qc qc
sct_label_vertebrae -i t2.nii.gz -s t2_seg.nii.gz -c t2 -qc qc -r 0 -v 2
for png in /tmp/sct-*-label_vertebrae-*/fig_pattern_disc2.png; do xdg-open "$png"; done
```

fig_pattern_disc2

joshuacwnewton (approved):
I like the changes you've made! I concur that our use of `scipy.gaussian` was having no effect at all.

It's tempting for me to suggest more changes (minor refactors, tests for `verbose=2`, etc.), but on reflection I feel that this is probably a relatively low-priority part of SCT's codebase, so I'm going to give my approval here.

(To explain further: I imagine the vast majority of our users use `sct_label_vertebrae` with its default values -- it's rare that folks will adjust CLI options on their own, as it usually only happens after they receive guidance on the forums, e.g. https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3702. Given that these were "advanced" parameters, i.e. an even _more_ niche part of the options, I don't think we need to worry too much about it at the moment. If we were to spend our time writing tests, for example, it would probably be better to focus on the more high-traffic areas of SCT's codebase, e.g. CLI scripts that aren't currently tested at all.)

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -35,26 +35,32 @@
...
-            obj = object.split('=')
-            if obj[0] == 'gaussian_std':
-                setattr(self, obj[0], float(obj[1]))
+                sys.exit(1)
+            if key in ['shift_AP', 'size_AP', 'size_RL', 'size_IS', 'shift_AP_visu']:
+                setattr(self, key, value)
+            elif key == 'gaussian_std':
+                # TODO: remove 'gaussian_std' completely for release 6.0
+                printv('WARNING: gaussian_std parameter is currently ignored, '
+                       'and will be removed in a later version.', 1, type='warning')
```
joshuacwnewton, line 60: Ah, one other comment. I would love to maybe come up with a procedure for returning to deprecation notices like these?

Maybe like, a specific syntax we could include in the comment, like:


```suggestion
                 # DEP: remove 'gaussian_std' completely two releases after v5.5
                printv('WARNING: gaussian_std parameter is currently ignored, '
                       'and will be removed in a later version.', 1, type='warning')
```

Then we could include instructions in the "preparing for a new release" dev Wiki page to do a project-wide search for `# DEP:` a week or two prior to the release due-date.

mguaypaq: I'm in favour of using a standardized syntax, yes. How about `# TODO(v6.0): remove gaussian_std completely`? I think `TODO` is already a widely recognized string, and the project doesn't seem to contain `# TODO(` as a substring yet.

From [our versioning guidelines](https://intranet.neuro.polymtl.ca/geek-tips/contributing.html#versioning-a-href-versioning-id-versioning-a), it sounds like we're doing something like [semantic versioning](https://semver.org/spec/v2.0.0.html), so I assumed the deprecation would last until the next major release, v6.0 in this case. Then again, I don't know if we have any sort of timeline for when v6.0 happens. Would you instead prefer v5.7 (so the deprecation warning appears during v5.6), or maybe v5.8 (so the warning appears during v5.6 and v5.7)?

(As an aside, I tend to prefer deadlines (for example, "by v5.7") instead of timeouts ("two releases after v5.5"), as expressed in [this little blog post](https://guxi.me/posts/timeout-vs-deadline/).)

joshuacwnewton: > From [our versioning guidelines](https://intranet.neuro.polymtl.ca/geek-tips/contributing.html#versioning-a-href-versioning-id-versioning-a), it sounds like we're doing something like [semantic versioning](https://semver.org/spec/v2.0.0.html), so I assumed the deprecation would last until the next major release, v6.0 in this case.

So... SCT's versioning is a bit tricky at the moment. Before I was hired, we were indeed following semantic versioning! But, IMO there were four main issues with that:

1. SCT's dev team is small, and we're not currently fast/organized enough to plan out both bugfix/patch releases and feature releases. (At least, not without some changes to my/our workflow?)
2. SCT's feature development is more sporadic, and often dependent on whether or not we have dedicated grad students developing new image processing approaches. So, it can be hard(er) to manage/schedule patch vs. feature releases?
3. SCT is not pip installable, which means:
   - Our userbase is has ended up composed of primarily non-developer folks, who are less likely to have semantic expectations for version numbers.
   - There isn't an easy way to update SCT beyond wiping the installation folder entirely, then reinstalling from scratch. (This makes patch releases... functionally useless? Since installing a patch release is equally as laborious as installing, say, a minor release.) If SCT _was_ pip installable, being able to upgrade via pip to the newest patch version would make semantic versioning (and more frequent releases via PyPI) a lot more appealing, I think?
4. Our users tend to stick with a specific version for conducting a study. i.e. they're not likely to stay on top of new versions as rapidly as typical consumers of Python packages. (You'll often see bug reports or questions on the forum which start with "I'm using SCT 4.X..." because that's the version they used a year or two ago in their last study.) 

Historically, all of these factors combined resulted in SCT only ever releasing once there were significant enough changes to _warrant_ a new release. So, you'd see 0-1 patch releases for every minor release -- basically nonexistent. 

During a meeting last year, I asked "Does it even make sense to follow semantic versioning anymore?" with the end result being us ditching semantic versioning to stick entirely to minor/major releases. 

(But, even with minor/major releases only, it's unclear to me what change would be dramatic enough to warrant a 6.0 release. If anything, our release numbering is more of a marketing tool at the moment than a system that conveys developer intent? I expect that once something radical happens (pip installable SCT, windows release, etc.) then that's when we'll choose 6.0, as a marker for a new era of SCT, aha.)

Still, if you have thoughts or opinions about this, I think our approach to versioning definitely worth revisiting. And, at the very least, there should be a dev wiki section documenting the methodology here. 😅

joshuacwnewton: > (But, even with minor/major releases only, it's unclear to me what change would be dramatic enough to warrant a 6.0 release. If anything, our release numbering is more of a marketing tool at the moment than a system that conveys developer intent? I expect that once something radical happens (pip installable SCT, windows release, etc.) then that's when we'll choose 6.0, as a marker for a new era of SCT, aha.)

Given what I've written here, I also think 5.7 also makes a lot of sense for this specific PR, as you've mentioned...)

mguaypaq: Thanks for the writeup! That makes a lot of sense. I'm happy to follow this versioning scheme, knowing what it means and the intent behind it.

joshuacwnewton: Thank you for pointing out the fuzzy versioning methodology in the first place! It helps a lot to have to put these things into words, rather than having them be these ephemeral, unspoken thing. :)

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -35,26 +35,32 @@
...
             else:
-                setattr(self, obj[0], int(obj[1]))
+                printv('ERROR: Wrong usage.', 1, type='error')
+                sys.exit(1)
```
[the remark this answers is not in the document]
mguaypaq: Well, I think I managed to refactor the parsing code more completely. The parser now complains more nicely when `-param` is used incorrectly.

As a bonus, I fixed a bug where the old code only ever looked at the first `key=value` pair passed to `-param`.

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -31,39 +31,33 @@
...
+    for key_value in string.split(','):
+        key, value = key_value.split('=', maxsplit=1)
+        if key in param:
+            param[key] = int(value)
+        elif key == 'gaussian_std':
+            # TODO(issue#3706): remove 'gaussian_std' completely for v5.7
+            printv('WARNING: gaussian_std parameter is currently ignored, '
+                   'and will be removed in a later version.', 1, type='warning')
+        else:
+            raise ValueError(f'Unknown parameter: {key}')
```
joshuacwnewton, line 56: I like the refactoring you've done! I noticed, though, that with the newly-proposed method for parsing, the failure given by `argparse` is a little opaque. Some (separate) examples:

```
sct_label_vertebrae: error: argument -param: invalid vertebral_detection_param value: 'test'
sct_label_vertebrae: error: argument -param: invalid vertebral_detection_param value: 'test=None,shift_AP=None'
sct_label_vertebrae: error: argument -param: invalid vertebral_detection_param value: 'shift_AP=None,shift_AP=None'
sct_label_vertebrae: error: argument -param: invalid vertebral_detection_param value: 'shift_AP=9,size_AP=6.0'
```

Regardless of what the actual error is, `argparse` interprets it as a generic type-related error and obscures what the actual issue is.

For niche parsing requirements like this, I tend to like using a [custom action](https://me.micahrl.com/blog/python-argparse-custom-actions-types/)? For example:

```python
import argparse

class ParseParamAction(argparse.Action):
    """Custom parser for vertebral_detection advanced parameters."""
    def __call__(self, parser, namespace, values, option_string=None):
        try:
            param = param_default.copy()
            for key_value in values.split(','):
                key, value = key_value.split('=', maxsplit=1)
                if key in param:
                    param[key] = int(value)
                elif key == 'gaussian_std':
                    # TODO(issue#3706): remove 'gaussian_std' completely for v5.7
                    printv('WARNING: gaussian_std parameter is currently ignored, '
                           'and will be removed in a later version.', 1, type='warning')
                else:
                    raise ValueError(f'Unknown parameter for "-param" argument: {key}')
            return param
        except Exception as e:
            parser.error(e)

    optional.add_argument(
        '-param',
        metavar=Metavar.list,
        action=ParseParamAction,
        default=param_default.copy(),
        help=f'Advanced parameters. Assign value with "="; Separate arguments with ","\n'
             f'  - shift_AP [mm]: AP shift of centerline for disc search (default: {param_default["shift_AP"]})\n'
             f'  - size_AP [mm]: AP window size for disc search (default: {param_default["size_AP"]})\n'
             f'  - size_RL [mm]: RL window size for disc search (default: {param_default["size_RL"]})\n'
             f'  - size_IS [mm]: IS window size for disc search (default: {param_default["size_IS"]})\n',
    )
```

With this, `argparse` will raise a more specific error while still showing the `-h`:

Screenshot from 2022-03-04 20-47-33

joshuacwnewton (commented):
Admittedly, when I saw the new changes, I was a bit nervous about the expanding scope of this PR? 

But, now that I've read the bugs you uncovered, I think the refactoring is very much worth the effort. And, I definitely think it's easy enough to rebrand this PR as a collection of bugfixes for `-param`, which is still (relatively) tiny in scope. :D

If it's okay, I think I might rename this PR, just to reflect the new changes?

spinalcordtoolbox/scripts/sct_label_vertebrae.py
joshuacwnewton: Admittedly, though, custom actions aren't well documented in `argparse`'s documentation, though. So, my apologies for throwing a (slightly) arcane pattern at you, aha. And, feel free to tweak this suggestion if needed!

mguaypaq: That's a very fair point about the specificity of error messages, thanks for bringing it up.

Indeed, my previous draft used a custom Action, yes! I abandoned it because:
1. I didn't know about `parser.error`, so the errors raised by my custom Action had an ugly traceback and no usage message. Now I know!
2. I was uneasy about the handling of the other arguments to `Action.__init__`, like `nargs`, that don't really make sense for this custom Action.

But, good news, I just found that it's possible to have custom error messages from a type function! We just have to use `argparse.ArgumentTypeError` (not to be confused with `argparse.ArgumentError`):
```py3
>>> import argparse
>>> p = argparse.ArgumentParser('my_program')
>>> def f(string):
...     raise argparse.ArgumentTypeError(f'a custom message, {len(string)}')
... 
>>> _ = p.add_argument('--foo', type=f)
>>> p.parse_args(['--foo', 'lalala'])
usage: my_program [-h] [--foo FOO]
my_program: error: argument --foo: a custom message, 6
```

Given this, I'll keep the simple type function approach, but make better error messages.

joshuacwnewton: Perfect!!

PR #3699, opened 2022-02-22
Demonstrate native Windows installation with all bugfixes combined

joshuacwnewton:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR combines _all_ of the platform-specific Windows fixes. (This is just to demonstrate how all the fixes will work together.)

However, this work should be probably be reviewed in chunks? I imagine the best route forward is:

* Close this PR.
* Open a separate PR to merge `install_sct.bat`, CI runners, and bare-minimum fixes required for proper install (`sys.py`-related bugfixes)
* Merge test-specific bugfixes (probably 5-8 separate PRs for this)

## Linked issues

Addresses #3682.

Issue #3682, opened 2022-02-09
Support installing SCT natively on Windows


joshuacwnewton:
## Context

Currently, it is not possible to install SCT natively on Windows. Because of this, some sort of virtualization software (e.g. WSL, Docker, a VM)  is needed to run SCT on Windows.

Setting up this virtualization software is often challenging and confusing for our users. So, there is a pressing need to improve this experience, for example by streamlining the documentation (#3610). This issue, however, is for the more radical approach: ditching WSL/Docker/VMs entirely by installing SCT natively on Windows.

It's possible that this would require more work than is feasible, so I'm opening this issue as a place to record an initial investigation.

## Remaining work

More details will be filled in soon. This is just a rough brainstorming sketch:

- [ ] `install_sct` vs. pip-installable SCT (#1526)
- [ ] Tricky platform-specific dependencies (torch, tensorflow, etc.)
- [ ] Platform-specific binaries (ANTs, etc.)

joshuacwnewton:
## Bare-minimum pip install test (no PyPI, binaries, PATH modifications, etc.)

The first thing I wanted to try is simply installing SCT and its dependencies as-is to see what sort of snags we run into:

1. Download and install Python 3.7.9 executable via [this page](https://www.python.org/downloads/release/python-379/)
2. Clone `spinalcordtoolbox` repo
3. Setup a `virtualenv` in the SCT directory by running:
    ```bash
    cd <sct-repo-clone-location>
    python -m venv venv_sct
    cd venv_sct/Scripts
    activate
    # upgrade because default p3.7 pip won't resolve dependency conflicts correctly
    python -m pip install --upgrade pip setuptools wheel
    ```
4. Install SCT's dependencies by running:
    ```bash
    cd ../..
    pip install -r requirements.txt
    ```
5. Install SCT by running:
    ```
    pip install -e .
    ```
6. Verify installation with
    ```
    sct_check_dependencies
    ```

<details>
<summary> Log of the above steps </summary>

```console
(venv_sct) C:\Users\Joshua\repos\spinalcordtoolbox\venv_sct\Scripts>    python -m pip install --upgrade pip setuptools wheel
Collecting pip
  Using cached pip-22.0.3-py3-none-any.whl (2.1 MB)
Collecting setuptools
  Using cached setuptools-60.8.2-py3-none-any.whl (1.1 MB)
Collecting wheel
[...]

Issue #3545, opened 2021-10-19
SCT's current distribution method prevents us from checking download statistics


joshuacwnewton:
Currently, users download stable releases via the "Source code" links on the [Releases](https://github.com/spinalcordtoolbox/spinalcordtoolbox/releases) page (which point to a tag on the `release` branch).

![Screenshot from 2021-10-19 13-16-43](https://user-images.githubusercontent.com/16181459/137959666-d6f9c754-d4b4-44b0-b60b-24ffb92b70fb.png)

The problem is: Download statistics aren't tracked for "Source code" downloads (https://github.com/isaacs/github/issues/593). For us to be able to track the number of downloads, we would need to attach additional assets to our releases. 

* This means we have [no data for the past 5 years of releases](https://hanadigital.github.io/grev/?user=spinalcordtoolbox&repo=spinalcordtoolbox), because they didn't include assets. This is a big problem for demonstrating the impact of SCT, which is critical for securing funding.

* This also means that our current Downloads badge is _severely_ underestimating the number of downloads for SCT.

   ![total.svg](https://img.shields.io/github/downloads/spinalcordtoolbox/spinalcordtoolbox/total.svg)

To fix this, we could modify `install_sct` to download the source code, that way the script could be distributed and run by itself (e.g. `install_sct-5.4.sh` attached to the release). Then we could track the number of downloads for that script.

But, this doesn't help with the problem that we're in the dark for download statistics over the past 5 years.

joshuacwnewton:
> Would that somewhat go against the parallel effort to "pipify" SCT?

I think that the amount of work needed would be roughly the same before/after the change.

* Current:
    1. User manually downloads source code
    2. Run install_sct
    3. Rest of steps (install dependencies, install SCT, etc.)
* Proposed: 
    1. Run install_sct
    2. Automatically download source code
    3. Rest of steps (install dependencies, install SCT, etc.)

Since we are just swapping the first two steps, the rest of the steps are still the same, so the majority of the work needed to "pipify" SCT remains unchanged.

joshuacwnewton:
As part of our Native Windows releases, we began trialling `install_sct.bat` as a separate release asset, rather than asking users to download the full source code zip.

I'm happy to say that this has indeed allowed us to track some statistics for downloads:

![image](https://user-images.githubusercontent.com/16181459/205134004-56599d4e-2979-4377-bba8-4ec5aea43adf.png)

[...]

Issue #2669, opened 2020-04-21
Packaging Dependencies


kousu:
Currently `./install_sct` uses two package managers (actually four if you count conda and brew): `pip` and `sct_download_data`. The latter we are maintaining ourselves, and we use it for dependencies that we "own", namely the large pre-compiled binary programs (particularly `bin/isct_antsApplyTransforms`, `bin/isct_antsRegistration`, and `bin/isct_antsSliceRegularizedRegistration`), some tools from `dev/`, and M Sdika's [ctrDetect](https://www.creatis.insa-lyon.fr/site7/en/ctrDetect) which we don't own but which also isn't packaged by anything other package manager, and also all of our large pre-trained machine learning models, and maintaining the contents and metadata is a fairly manual process documented here:

* https://github.com/neuropoly/spinalcordtoolbox/wiki/Binaries-Compilation
* https://github.com/neuropoly/spinalcordtoolbox/wiki/Compile-binaries
* https://github.com/neuropoly/spinalcordtoolbox/wiki/Package-data-&-binaries

Motivated by #947 / [forum365](http://forum.spinalcordmri.org/t/sct-fmri-moco-very-very-slow/365/2) / #2642, where the issue was that we were shipping very old binaries and easily update them, I want to automate building and packaging these dependencies as much as possible.

Here is my plan

kousu:
I think [this comment in the wiki](https://github.com/neuropoly/spinalcordtoolbox/wiki/Binaries-Compilation#diy) is worth preserving, to demonstrate why getting this automated and repeatable (and *repeated* regularly) is important:

> If you want to build your binaries, as of 2017-12-05, you're on your own, but we are working (#1526) on handling bundled vs. system dependencies and building the binaries as part of the installation process (using distribution packages and ultimately setuptools). Feel free to contribute!

kousu:
# Dependency Graph

This is the conceptual dependency graph; eventually I want to encode this as data, using a package management formats (probably `requirements.txt` but I'm open to suggestions).

* TODO

kousu:
@jcohenadad says that isct_dice_coefficient is unused. So really `spinalcordtoolbox-dev` is just `spinalcordtoolbox-propseg`. It still needs to be recompiled.

kousu:
[...]

PR #3701, opened 2022-02-22
Add minimally functional Windows installer and GitHub Actions CI runner

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR accomplishes the following 3 tasks:

1. Adds an `install_sct.bat` script that will take care of the necessary steps to install SCT on Windows.
2. Fixes the bare-minimum number of bugs in order to ensure SCT can install correctly on Windows.
3. Adds a native Windows CI runner that installs and tests SCT on Windows.

Note: There will be _many_ test failures on this PR. (There are currently [50 failures in this CI run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5292198694?check_suite_focus=true), but I've set it up so that the runner will mark as "Passed" as to not cause noise for other PRs, see 096f5c7.)

We do have fixes to get the number of failures from 50 down to ~9~ **0** (see PR #3699), but I'm thinking that we can use this PR as a starting point, then merge the test-specific bugfixes one by one _after_ we set up this foundation first. (That way, we can isolate the effects of individual changes, and more easily review them in context.)

## Linked issues

Partially addresses #3682.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -424,4 +424,4 @@ jobs:
       shell: cmd
       run: sct_check_dependencies
     - name: Run pytest test suite
-      run: sct_testing
+      run: sct_testing || EXIT 0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -384,3 +384,39 @@ jobs:
       run: |
         cd ~/spinalcordtoolbox
         ./.ci.sh -t
+
+  windows-native:
+    name: Windows (Native)
+    runs-on: windows-2019
+    defaults:
+      run:
+        shell: cmd
+    steps:
+    - name: Get branch name (merge)
+      if: github.event_name != 'pull_request'
+      shell: bash
+      run: echo "BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+    - name: Get branch name (pull request)
+      if: github.event_name == 'pull_request'
+      shell: bash
+      run: echo "BRANCH_NAME=$(echo ${GITHUB_HEAD_REF})" >> $GITHUB_ENV
+    - uses: actions/checkout@v2
+      with:
+        path: checkout/spinalcordtoolbox
+    - name: Install SCT
+      # NB: In real-world usage, it's assumed that the user would download the
+      # install_sct.bat file from the Downloads page, and _not_ the full repo.
+      run: |
+        set HOMEPATH=%GITHUB_WORKSPACE%
+        cd checkout/spinalcordtoolbox
+        .\install_sct.bat %BRANCH_NAME%
+    - name: Add SCT to path
+      # NB: I'm not sure what GHA's syntax is for cmd.exe, so we use bash just for this one change
+      # In a user install, the user would perform this step using the Windows PATH-changing GUI.
+      shell: bash
+      run: echo "$GITHUB_WORKSPACE/spinalcordtoolbox/venv_sct/Scripts" >> $GITHUB_PATH
+    - name: Check dependencies
+      shell: cmd
+      run: sct_check_dependencies
+    - name: Run pytest test suite
+      run: sct_testing
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -411,9 +411,10 @@ jobs:
     - name: Install SCT
       # NB: In real-world usage, it's assumed that the user would download the
       # install_sct.bat file from the Downloads page, and _not_ the full repo.
+      # We emulate that here by copying just the install_sct.bat to the PWD
       run: |
         set HOMEPATH=%GITHUB_WORKSPACE%
-        cd checkout/spinalcordtoolbox
+        copy checkout\spinalcordtoolbox\install_sct.bat .\
         install_sct.bat %BRANCH_NAME%
     - name: Add SCT to path
       # NB: I'm not sure what GHA's syntax is for cmd.exe, so we use bash just for this one change
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,6 +39,11 @@ env:
     # This will be defined for non-debian platforms below too, but there's no harm in that.
     # (TravisCI quietly defined this on all their platforms, but we have to give it manually on GithubCI.)
     DEBIAN_FRONTEND: 'noninteractive'
+    # Turns on color output for pytest. See: https://github.com/pytest-dev/pytest/issues/7443#issuecomment-656642591
+    PY_COLORS: "1"
+    # Disable progress bars for less verbose output
+    PIP_PROGRESS_BAR: "off"
+    SCT_PROGRESS_BAR: "off"
 
 jobs:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -414,7 +414,7 @@ jobs:
       run: |
         set HOMEPATH=%GITHUB_WORKSPACE%
         cd checkout/spinalcordtoolbox
-        .\install_sct.bat %BRANCH_NAME%
+        install_sct.bat %BRANCH_NAME%
     - name: Add SCT to path
       # NB: I'm not sure what GHA's syntax is for cmd.exe, so we use bash just for this one change
       # In a user install, the user would perform this step using the Windows PATH-changing GUI.
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -2,6 +2,9 @@
 :: e.g.
 ::        install_sct.bat 5.5
 
+:: Turn off command echoing for nicer output
+@echo off
+
 :: Fetch version number (should be git branch, git tag, etc.)
 set git_ref=%1
 
@@ -12,31 +15,53 @@ FOR /F "tokens=*" %%G IN ('cd') do (SET PWD=%%G)
 cd %HOMEPATH%
 
 :: Download SCT and check out the branch requested by the user
+echo:
+echo ### Downloading SCT source code to %HOMEPATH%...
 git clone -b %git_ref% --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git || goto :error
 
 :: Create and activate virtual environment to install SCT into
+echo:
+echo ### Using Python to create virtual environment...
 cd spinalcordtoolbox
 python -m venv venv_sct || goto :error
 CALL venv_sct\Scripts\activate.bat || goto :error
-
-:: Upgrade pip because default p3.7 pip won't resolve dependency conflicts correctly
-python -m pip install --upgrade pip || goto :error
+echo Virtual environment created and activated successfully!
 
 :: Install SCT and its requirements
+echo:
+echo ### Installing SCT and its dependencies...
+:: Pip needs to be upgraded because default p3.7 pip won't resolve dependency conflicts correctly
+python -m pip install --upgrade pip || goto :error
 pip install -r requirements.txt || goto :error
 pip install -e . || goto :error
 
 :: Install external dependencies
+echo:
+echo ### Downloading model files and binaries...
 FOR %%D IN (PAM50 gm_model optic_models pmj_models deepseg_sc_models deepseg_gm_models deepseg_lesion_models c2c3_disc_models binaries_win) DO sct_download_data -d %%D -k || goto :error
 
 :: Deactivate virtual environment
 deactivate
 
+:: Give further instructions that the user add the Scripts directory to their PATH
+echo:
+echo ### Installation finished!
+echo:
+echo To use SCT's command-line scripts in Command Prompt, please follow these instructions:
+echo:
+echo 1. Open the ^Start Menu -^> Type 'path' -^> Open 'Edit the System Environment Variables'
+echo 2. Click the 'Environment variables...' button
+echo 3. Under the section 'User variables for ____', highlight the 'Path' entry, then click the 'Edit...' button.
+echo 4. Click 'New', then copy and paste this directory:
+echo:
+echo    %CD%\%venv_sct\Scripts
+echo:
+echo: 5. Click 'OK' three times. You can now access SCT's scripts in the Command Prompt.
+
 :: Return to initial directory
 cd %PWD%
 
 
-:: Give further instructions that the user add the Scripts directory to their PATH
 
 :error
 cd %PWD%
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -8,11 +8,8 @@
 :: Fetch version number (should be git branch, git tag, etc.)
 set git_ref=%1
 
-:: Save present working directory to return to later
-FOR /F "tokens=*" %%G IN ('cd') do (SET PWD=%%G)
-
 :: Go to user's home directory
-cd %HOMEPATH%
+pushd %HOMEPATH%
 
 :: Download SCT and check out the branch requested by the user
 echo:
@@ -59,12 +56,12 @@ echo:
 echo: 5. Click 'OK' three times. You can now access SCT's scripts in the Command Prompt.
 
 :: Return to initial directory
-cd %PWD%
+popd
 
 exit /b 0
 
 :error
-cd %PWD%
+popd
 deactivate || BREAK
 echo Failed with error #%errorlevel%.
 exit /b %errorlevel%
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -43,7 +43,7 @@ echo ### Installation finished!
 echo:
 echo To use SCT's command-line scripts in Command Prompt, please follow these instructions:
 echo:
-echo 1. Open the ^Start Menu -^> Type 'path' -^> Open 'Edit the System Environment Variables'
+echo 1. Open the Start Menu -^> Type 'path' -^> Open 'Edit the System Environment Variables'
 echo 2. Click the 'Environment variables...' button
 echo 3. Under the section 'User variables for ____', highlight the 'Path' entry, then click the 'Edit...' button.
 echo 4. Click 'New', then copy and paste this directory:
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -37,9 +37,6 @@ echo:
 echo ### Downloading model files and binaries...
 FOR %%D IN (PAM50 gm_model optic_models pmj_models deepseg_sc_models deepseg_gm_models deepseg_lesion_models c2c3_disc_models binaries_win) DO sct_download_data -d %%D -k || goto error
 
-:: Deactivate virtual environment
-deactivate
-
 :: Give further instructions that the user add the Scripts directory to their PATH
 echo:
 echo ### Installation finished!
@@ -55,8 +52,9 @@ echo    %CD%\%venv_sct\Scripts
 echo:
 echo: 5. Click 'OK' three times. You can now access SCT's scripts in the Command Prompt.
 
-:: Return to initial directory
+:: Return to initial directory and deactivate the virtual environment
 popd
+deactivate
 
 exit /b 0
 
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -14,7 +14,7 @@ pushd %HOMEPATH%
 
 rem Download SCT and check out the branch requested by the user
 echo:
-echo ### Downloading SCT source code to %HOMEPATH%...
+echo ### Downloading SCT source code to %HOMEPATH%\spinalcordtoolbox...
 git clone -b %git_ref% --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git || goto error
 
 rem Create and activate virtual environment to install SCT into
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -1,22 +1,22 @@
-:: Usage: install_sct.bat <version>
-:: e.g.
-::        install_sct.bat 5.5
+rem Usage: install_sct.bat <version>
+rem e.g.
+rem        install_sct.bat 5.5
 
-:: Turn off command echoing for nicer output
+rem Turn off command echoing for nicer output
 @echo off
 
-:: Fetch version number (should be git branch, git tag, etc.)
+rem Fetch version number (should be git branch, git tag, etc.)
 set git_ref=%1
 
-:: Go to user's home directory
+rem Go to user's home directory
 pushd %HOMEPATH%
 
-:: Download SCT and check out the branch requested by the user
+rem Download SCT and check out the branch requested by the user
 echo:
 echo ### Downloading SCT source code to %HOMEPATH%...
 git clone -b %git_ref% --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git || goto error
 
-:: Create and activate virtual environment to install SCT into
+rem Create and activate virtual environment to install SCT into
 echo:
 echo ### Using Python to create virtual environment...
 cd spinalcordtoolbox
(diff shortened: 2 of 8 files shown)


install_sct.bat
```diff
@@ -0,0 +1,45 @@
...
+FOR %%D IN (PAM50 gm_model optic_models pmj_models deepseg_sc_models deepseg_gm_models deepseg_lesion_models c2c3_disc_models binaries_win) DO sct_download_data -d %%D -k || goto :error
+
+:: Deactivate virtual environment
+deactivate
+
+:: Return to initial directory
+cd %PWD%
+
+
+:: Give further instructions that the user add the Scripts directory to their PATH
```
joshuacwnewton, line 39: Todo reminder (whoops, forgot about this)

install_sct.bat
```diff
@@ -0,0 +1,45 @@
...
+::        install_sct.bat 5.5
+
+:: Fetch version number (should be git branch, git tag, etc.)
+set git_ref=%1
+
+:: Save present working directory to return to later
+FOR /F "tokens=*" %%G IN ('cd') do (SET PWD=%%G)
+
+:: Go to user's home directory
+cd %HOMEPATH%
```
joshuacwnewton, line 12: For simplicity's sake (just to get us off the ground), I've hardcoded the script to install to the user's home directory.

While this isn't really kosher as far as Windows applications go, my thought process was that... because our end-goal is a pip-installable SCT, we're going to be discarding a lot of this fluff so that the user can install directly into a virtual environment. So, rather than go down the path of a Windows application, I'd rather be transparent about what's actually happening as far as virtual environments go.

install_sct.bat
```diff
@@ -0,0 +1,45 @@
...
+
+:: Download SCT and check out the branch requested by the user
+git clone -b %git_ref% --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git || goto :error
```
joshuacwnewton, line 15: My thought process with cloning is that we would attach `install_sct.bat` to a release, and the user would download _just_ the lightweight script.

Ideally, we would do this for `install_sct`, too. That way we would address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3545 for all platforms.

install_sct.bat
```diff
@@ -0,0 +1,45 @@
+:: Usage: install_sct.bat <version>
+:: e.g.
+::        install_sct.bat 5.5
+
+:: Fetch version number (should be git branch, git tag, etc.)
+set git_ref=%1
```
joshuacwnewton, line 6: This usage could maybe be tweaked a little. e.g. Hardcoding the version to install for each release?

The main reason I did it this way was to allow for testing PRs, etc.

install_sct.bat
```diff
@@ -0,0 +1,45 @@
...
+cd spinalcordtoolbox
+python -m venv venv_sct || goto :error
+CALL venv_sct\Scripts\activate.bat || goto :error
+
+:: Upgrade pip because default p3.7 pip won't resolve dependency conflicts correctly
+python -m pip install --upgrade pip || goto :error
+
+:: Install SCT and its requirements
+pip install -r requirements.txt || goto :error
+pip install -e . || goto :error
```
joshuacwnewton, line 27: At one point I tested installing SCT in non-editable mode (`pip install .`) but that caused a mess in terms of the files outside of the `spinalcordtoolbox/` folder (`bin/`, `data/`, `contrib/`, etc.)

I foresee this being another blocker for #1526, so we should keep this in mind. (I think Nick already somewhat took this into account during his efforts for #2669?)

spinalcordtoolbox/compat/launcher.py
```diff
@@ -34,4 +35,4 @@ def main():
             mpi_flags = "-n 1"
         cmd = ["mpiexec"] + mpi_flags.split() + cmd
 
-    os.execvpe(cmd[0], cmd[0:], env)
+    return subprocess.call(" ".join(cmd), shell=True, env=env)
```
joshuacwnewton, line 38: From discussions in https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/22#discussion_r811346511, I learned that `shell=True` is perhaps not so good.

I'm not sure why I have this here, but I should remember to test with `shell=False` instead.

spinalcordtoolbox/reports/qc.py
```diff
@@ -660,7 +653,13 @@ def _update_html_assets(self, json_data):
...
             template = Template(template_index.read())
             output = template.substitute(sct_json_data=json.dumps(json_data))
-            io.open(os.path.join(dest_path, 'index.html'), 'w', encoding="utf-8").write(output)
+
+        # Use locking when writing to `index.html` only (since this is the only file that multiple processes
+        # would simultaneously write to)
+        dest_file = io.open(os.path.join(dest_path, 'index.html'), 'w', encoding="utf-8")
+        portalocker.lock(dest_file, portalocker.LOCK_EX)
+        dest_file.write(output)
+        dest_file.close()
```
joshuacwnewton, line 662: `portalocker.LOCK_EX` refers to an exclusive lock, see: https://portalocker.readthedocs.io/en/latest/portalocker.constants.html

spinalcordtoolbox/reports/qc.py
```diff
@@ -660,7 +653,13 @@ def _update_html_assets(self, json_data):
         with io.open(os.path.join(assets_path, 'index.html'), encoding="utf-8") as template_index:
             template = Template(template_index.read())
             output = template.substitute(sct_json_data=json.dumps(json_data))
-            io.open(os.path.join(dest_path, 'index.html'), 'w', encoding="utf-8").write(output)
+
+        # Use locking when writing to `index.html` only (since this is the only file that multiple processes
+        # would simultaneously write to)
+        dest_file = io.open(os.path.join(dest_path, 'index.html'), 'w', encoding="utf-8")
+        portalocker.lock(dest_file, portalocker.LOCK_EX)
```
mguaypaq, line 660: I'm seeing some possible race conditions with the current ordering. The biggest one would be:
1. Process A starts writing to A.json
2. Process B reads all json files, including this probably-malformed-right-now A.json, and crashes.

Another would be:
1. Process C creates C.json, and reads all json files.
2. Process D creates D.json, reads all json files, produces a new index.html, with all the data.
3. Process C overwrites this with a new index.html file, which doesn't include D.json.

With portalocker, my understanding is that we can only lock files, not directories, but that's not a problem: we could lock any particular file, like index.html, as long as all the processes use the same file. But the ordering should probably be something like:
1. Open index.html for updating, mode `'r+'` (not writing! That would immediately delete the contents), and grab/wait for an exclusive lock.
2. Write a new json file.
3. Read all json files.
4. Write the new index.html.
5. Release the lock on index.html.
6. (I'm not sure if this step needs to hold the lock, since these files are constant) copy the other asset files.

That way, all the modifications to (the collection of json files and their contents) happen within the critical region, so each process has a consistent view, and the last update to index.html contains everything.

spinalcordtoolbox/reports/qc.py
mguaypaq: I believe that on Unix-likes, closing a file handle also releases fcntl locks, but I'm not sure about Windows. The [portalocker examples section](https://portalocker.readthedocs.io/en/latest/#examples) seems to say that it's best to
1. `flush()` the file
2. `unlock()` it explicitly
3. `close()` it.

mguaypaq (changes requested):
Ok, this is a lot of comments, but overall I have to say I'm extremely pleased with this. Thank you so much! Also, thanks to this I'm learning tons of Windows stuff :book:

I set my review as "request changes" only because of the file locking issue; I'd like to have another look when you're done with it. For everything else, feel free to incorporate and/or ignore it as you see fit.

.github/workflows/tests.yml
```diff
@@ -384,3 +389,39 @@ jobs:
...
+    - name: Install SCT
+      # NB: In real-world usage, it's assumed that the user would download the
+      # install_sct.bat file from the Downloads page, and _not_ the full repo.
+      run: |
+        set HOMEPATH=%GITHUB_WORKSPACE%
+        cd checkout/spinalcordtoolbox
+        .\install_sct.bat %BRANCH_NAME%
+    - name: Add SCT to path
+      # NB: I'm not sure what GHA's syntax is for cmd.exe, so we use bash just for this one change
+      # In a user install, the user would perform this step using the Windows PATH-changing GUI.
```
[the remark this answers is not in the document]
joshuacwnewton: > A possible caveat: it looks like the Github Actions runner runs `cmd` with the `/d` switch, which [disables execution of AutoRun commands](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd). So even if we set `Autoexec.nt`, we probably also need to set `$GITHUB_PATH` as done here.

Caveat 2:

> If you place the path command in your Autoexec.nt file, the Windows operating system automatically appends the specified MS-DOS subsystem search path every time you log on to your computer. _**Cmd.exe does not use the Autoexec.nt file.**_ When started from a shortcut, Cmd.exe inherits the environment variables set in My Computer/Properties/Advanced/Environment. 
> 
> (Emphasize mine)

 😢... 

I feel this, though. I wanted a programmatic way to do this too, like what we currently do with `.bashrc`, etc.

At the same time, from a security perspective, it's nice that this is a step the user has some control over? I just hope it doesn't trip anyone up. 😅

spinalcordtoolbox/reports/qc.py
joshuacwnewton: > 2. Process B reads all json files, including this probably-malformed-right-now A.json, and crashes.

My goodness... I think you may have predicted the [exact error](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5379684935?check_suite_focus=true) that happened to me earlier today?

```
________________________________ test_many_qc _________________________________
"""
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\multiprocessing\pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\multiprocessing\pool.py", line 44, in mapstar
    return list(map(*args))
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\testing\api\test_qc_parallel.py", line 15, in gen_qc
    qc.generate_qc(fname_in1=t2_image, fname_seg=t2_seg, path_qc=path_qc, process="sct_deepseg_gm")
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 878, in generate_qc
    fps=fps
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 717, in add_entry
    layout(qcslice)
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 317, in wrapped_f
    self._make_QC_image_for_3d_volumes(img, mask, slice_orientation=sct_slice.get_name())
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 364, in _make_QC_image_for_3d_volumes
    self.qc_report.update_description_file(img.shape)
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 646, in update_description_file
    self._update_html_assets(get_json_data_from_path(path_json))
  File "d:\a\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\spinalcordtoolbox\reports\qc.py", line 888, in get_json_data_from_path
    results.append(json.load(fjson))
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\json\__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\json\__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\json\decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
"""
```

When writing this original `portalocker` fix, I thought that the `index.html` file was the only one that was repeatedly read and written to. But, as you've identified, the `.json` files are read and written to as well. Good catch!! Pat yourself on the back for that, gosh.

Your proposed steps make sense to me, too. Locking first before touching any `.json`/`.html` files does seem to me like it will prevent any issues with concurrent reads/writes of `.json` files.

joshuacwnewton:
Thank you for such a thorough review, @mguaypaq. I really appreciate your careful eye here. 

I also appreciate the nudges for little cleanups! I've been hesitant to make those changes myself, as I wanted to avoid causing headaches for reviewers by expanding the scope of my PRs too much. However, I think these types of changes make a lot more sense when the suggestions come _from the reviewer_! As, that way I'm not bombarding you with irrelevant changes; these are issues you're noticing yourself. :)

install_sct.bat
```diff
@@ -12,31 +15,53 @@ FOR /F "tokens=*" %%G IN ('cd') do (SET PWD=%%G)
...
 :: Deactivate virtual environment
 deactivate
 
+:: Give further instructions that the user add the Scripts directory to their PATH
+echo:
+echo ### Installation finished!
+echo:
+echo To use SCT's command-line scripts in Command Prompt, please follow these instructions:
+echo:
+echo 1. Open the ^Start Menu -^> Type 'path' -^> Open 'Edit the System Environment Variables'
```
joshuacwnewton, line 49: TIL that Windows uses `^` as its escape char, as `\` is taken up for path separation. 🤔

.github/workflows/tests.yml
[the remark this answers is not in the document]
joshuacwnewton: I've added what I hope are clear steps for manually setting the PATH in [fd9e06c](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3701/commits/fd9e06c73d2e0b013305621618d04e7d5a6f8ee4).

spinalcordtoolbox/reports/qc.py
joshuacwnewton: Done in [f2c2453](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3701/commits/f2c24536d0c57d04c7fb286673fb4dc42d7f43da).

.github/workflows/tests.yml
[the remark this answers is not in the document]
joshuacwnewton: Huh. On closer inspection, it looks like this isn't showing up in GHA's logs? https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5383645638?check_suite_focus=true#step:5:436

However, it does in my own Command Prompt. I'm curious to see if helps to swap the steps so that `deactivate` comes _after_. 🤔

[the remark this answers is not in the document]
joshuacwnewton: Yep, [8b400a1](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3701/commits/8b400a11dcaacbc5eb3fa57723f7ea6575e4b324) fixes it:

image

install_sct.bat
```diff
@@ -0,0 +1,66 @@
...
+echo:
+echo    %CD%\%venv_sct\Scripts
+echo:
+echo: 5. Click 'OK' three times. You can now access SCT's scripts in the Command Prompt.
+
+:: Return to initial directory and deactivate the virtual environment
+popd
+deactivate
+
+exit /b 0
```
mguaypaq, line 59: I'm almost certain that [`deactivate` is a batch script](https://github.com/conda/conda/blob/master/conda/shell/condabin/deactivate.bat), and that calling a batch script without using `call` is like `goto`: [it doesn't return](https://stackoverflow.com/questions/11638705/why-does-calling-a-nested-batch-file-without-prepending-call-to-the-line-exit).

(You wouldn't run into this if you were copy-pasting each line into the command line manually, which is very confusing when testing things.)

So technically, the `exit /b 0` is currently never executed and redundant, but I suspect the intention would be clearer by saying `call deactivate` instead.

install_sct.bat
```diff
@@ -0,0 +1,66 @@
...
+
+:error
+popd
+deactivate || BREAK
```
mguaypaq, line 63: This should also be `call deactivate`, I believe.

.github/workflows/tests.yml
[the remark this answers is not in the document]
mguaypaq: Those are great instructions! And if anything is still unclear, the only way to know will be if an actual user gets confused and reports it to us.

mguaypaq (approved):
Thanks for the additional changes! Here's another round of comments. There are still minor things to fix, but I'm marking this as "approved". Feel free to ping me if you want a final review at the end, but I think it's optional at this point.

install_sct.bat
```diff
@@ -0,0 +1,66 @@
+:: Usage: install_sct.bat <version>
```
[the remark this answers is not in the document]
joshuacwnewton: Oh! Thank you. That whole SO question + answer is useful, too, as it lead me to [this batch guide](https://www.tacktech.com/display.cfm?ttid=307). 😄

install_sct.bat
```diff
@@ -0,0 +1,66 @@
...
+@echo off
+
+:: Fetch version number (should be git branch, git tag, etc.)
+set git_ref=%1
+
+:: Go to user's home directory
+pushd %HOMEPATH%
+
+:: Download SCT and check out the branch requested by the user
+echo:
```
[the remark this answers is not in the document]
joshuacwnewton: I got the suggestion from [this SO answer](https://stackoverflow.com/a/132804), but given your docs link, I imagine that's a better way to go?

install_sct.bat
```diff
@@ -0,0 +1,66 @@
...
+echo ### Installation finished!
+echo:
+echo To use SCT's command-line scripts in Command Prompt, please follow these instructions:
+echo:
+echo 1. Open the ^Start Menu -^> Type 'path' -^> Open 'Edit the System Environment Variables'
+echo 2. Click the 'Environment variables...' button
+echo 3. Under the section 'User variables for ____', highlight the 'Path' entry, then click the 'Edit...' button.
+echo 4. Click 'New', then copy and paste this directory:
+echo:
+echo    %CD%\%venv_sct\Scripts
```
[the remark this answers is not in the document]
joshuacwnewton: Oh! That's very much a typo. Good catch. It should be `%CD%\venv_sct\Scripts`.

install_sct.bat
joshuacwnewton: I think it's because [`start`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/start) is a command in the batch language?

It's also possible that this is not actually necessary, aha. My IDE added syntax highlighting to `Start` and I felt the urge to escape it, but I can't remember if I double checked it or not.

install_sct.bat
joshuacwnewton: The reason I added `exit /b 0` was because at some point in my testing, the lines from the `:error` sections were being executed? 

I'll have to double-check, but for some reason I have the memory that calling `deactivate` _did_ return. Hm.. 

(Still, I really like that you pointed this out. You're getting me to think about aspects of batch scripting + venvs on Windows that I otherwise didn't consider!)

install_sct.bat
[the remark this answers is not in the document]
mguaypaq: Wow, after skimming [this forum discussion](https://www.dostips.com/forum/viewtopic.php?p=4554) very briefly, I don't know what to believe anymore. Maybe just keep `echo:`, since it's really not clear that any other suggestion is better, even the official one.

install_sct.bat
joshuacwnewton: Yep, good call. `^Start` isn't necessary; `Start` is fine. :)

install_sct.bat
mguaypaq: I guess the no-op would be to put nothing at all, instead of `|| something`? I don't believe `cmd.exe` has an equivalent of `set -e` anyway, it just ignores errors by default.

If you want to suppress output, I would have suggested redirecting output, but I believe `call` doesn't allow pipes and input/output redirection. In any case, if I remember my DOS/Windows days, I was personally never surprised to see extra errors at the end, after a real error :stuck_out_tongue: 

It occurs to me, however, that `popd` and `call deactivate` are very likely to overwrite `%errorlevel%` in between `:error` and `echo`/`exit`. If so, maybe save it into another variable right after `:error`?

install_sct.bat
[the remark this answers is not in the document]
joshuacwnewton: Follow-up to this: PyCharm [automatically adds `@REM`](https://stackoverflow.com/questions/64730481/how-to-change-toggle-comment-from-rem-to-rem-in-vs-code) when I use the commenting keyboard shortcut. I was curious about [`@REM` vs `REM`](https://softwareengineering.stackexchange.com/a/343864):

```batch
REM This will be shown during execution
@REM This will not!
@ECHO OFF
REM Nor this!
```

Probably Batch 101, but TIL. :)

install_sct.bat
joshuacwnewton: Ah, you're absolutely right. `deactivate` on its own doesn't return, but `call deactivate` returns. So, we should either use `deactivate` or `call deactivate ;; exit /b 0` but _not_ `deactivate ;; exit /b 0`. Subtle!

install_sct.bat
joshuacwnewton: Ahh, good catch.

spinalcordtoolbox/reports/qc.py
```diff
@@ -613,6 +612,17 @@ def update_description_file(self, dimension):
 
         :param: dimension 2-tuple, the dimension of the image frame (w, h)
         """
+        dest_path = self.qc_params.root_folder
+        html_path = os.path.join(dest_path, 'index.html')
+        # NB: We use 'r+' because it allows us to open an existing file for locking *without* immediately truncating the
+        # existing contents prior to opening. We can then use this file to overwrite the contents later in the function.
+        mode = "r+" if os.path.isfile(html_path) else "w"
```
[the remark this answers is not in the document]
joshuacwnewton: > My understanding is that a lock doesn't prevent opening and/or modifying the file; it only prevents another process from also locking the file. (That seems to be the case on Linux, but things could be very different on Windows.)

Your suspicions were correct. Opening a second file pointer in `'w'` mode causes `PermissionError: [Errno 13] Permission denied`. 😢 

However, I think there might be a simpler approach, specifically because of what you wrote here (emphasis mine): 

> Open the file with mode `'a'`, **which creates it if need be, doesn't erase it**, and allows us to grab a lock.

This made me realize that we're basically just looking for the `touch` command, but in Python, aren't we?

And, what do you know, `pathlib` has [`touch`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch):

> Create a file at this given path. [...] If the file already exists, the function succeeds if `exist_ok` is true (and its modification time is updated to the current time), otherwise [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) is raised.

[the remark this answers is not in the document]
joshuacwnewton: Actually, even simpler (since `touch` involves updating modification times, and maybe that ends up doing something weird, who knows).

```python
if not os.path.isfile(html_path):
    open(html_path, 'a').close()
```

install_sct.bat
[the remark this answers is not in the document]
joshuacwnewton: Oops. TIL to be careful about using `@` in git commits, lest I accidentally tag https://github.com/Rem. 😆

joshuacwnewton:
Repinging @mguaypaq on this just because I want to be extra safe when it comes to the file locking stuff. 😅

spinalcordtoolbox/reports/qc.py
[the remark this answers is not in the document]
mguaypaq: You don't even need the conditional:
```py3
# make sure the file exists
open(html_path, 'a').close()
```

Also, if you're opening in `'r+'` mode (which sounds like a good idea), make sure to call ['dest_file.truncate()`](https://docs.python.org/3/library/io.html#io.IOBase.truncate) after writing, otherwise there may be junk at the end:
```py3
dest_file = open(html_file, 'r+')
portalocker.lock(dest_file, portalocker.LOCK_EX)
...
dest_file.write(output)
dest_file.truncate()
dest_file.flush()
...
portalocker.unlock(dest_file)
dest_file.close()
```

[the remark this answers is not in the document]
joshuacwnewton: Aha, thank you! :)

mguaypaq (approved):
LGTM!

I added two small commits, if you agree with them you can go ahead and merge this :rocket:

Issue #3271, opened 2021-03-10
Investigate why 'sct_apply_transfo' is much slower than 'isct_antsApplyTransform'


joshuacwnewton:
In #2175, a call to `sct_apply_transfo` was replaced with a direct call to `isct_antsApplyTransform` for the following reason:

>  In a typical dataset, execution time goes from 4.5min to 2min. Fixes #2173.

However,`sct_apply_transfo` contains useful fixes for issues in the `isct_antsApplyTransform` binary (e.g. https://github.com/neuropoly/spinalcordtoolbox/issues/3254#issuecomment-795862992). So, they're not 1:1 replacements, and it seems we need `sct_apply_transfo` to be used.

Since we use `sct_apply_transfo` in many places (registration, vertebral labeling, etc.), fixing the speed problem could benefit many scripts.

joshuacwnewton:
I'm curious if the slowdown is because, in the past, we used `run_proc('sct_apply_transfo')`, which would use a subprocess. 

Now that we can call `main` directly (see https://github.com/neuropoly/spinalcordtoolbox/pull/3091) perhaps this issue has already been addressed.

joshuacwnewton:
> I'm curious if the slowdown is because, in the past, we used `run_proc('sct_apply_transfo')`, which would use a subprocess.
> 
> Now that we can call `sct_apply_transfo.main()` directly (see #3091) perhaps this issue has already been addressed. Perhaps fixing the slowdowns is as simple as replacing any `run_proc` calls.

Nix this theory. I tried running `sct_warp_template -a 1 -s 1` before and after this change: https://github.com/neuropoly/spinalcordtoolbox/pull/3272#discussion_r600663231

Even though `sct_apply_transfo.main` was used, there was still a significant slowdown.

joshuacwnewton:
I ran cProfile on the calls from https://github.com/neuropoly/spinalcordtoolbox/pull/3272#discussion_r600663231.

<details>
<summary> <code>run_proc(isct_antsApplyTransform)</code> </summary>

```python
         1333 function calls in 4.200 seconds
   Ordered by: cumulative time
   List reduced from 108 to 20 due to restriction <20>
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    4.200    4.200 /home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/sys.py:267(run_proc)
[...]

PR #3718, opened 2022-03-07
Improve how image overwriting is handled for memory-mapped data arrays on Windows

### LINKED ISSUE #3695: On Windows, saving an `Image` to a file with an open file pointer causes `OSError: [Errno 22] Invalid argument`
> * **Note:** This issue is the cause of 4 of the 13 remaining test failures described in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3682#issuecomment-1042079185.

In many places within SCT, we open an image file using our `Image` class, do some processing, then save the image to the same filename, overwriting it. On our unix platforms, this doesn't seem to cause any issues. However, on Windows, _sometimes_ an error like this is thrown:

```
OSError: [Errno 22] Invalid argument: 'C:\\Users\\Joshua\\AppData\\Local\\Temp\\sct-20220218135625.068144-label_vertebrae-ezuo8o0i\\segmentation_labeled.nii'
```

I was able to narrow down the cause of this vague error by trying to call `os.remove` instead of `Image.save()`:

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Joshua\\AppData\\Local\\Temp\\sct-20220218135625.068144-label_vertebrae-ezuo8o0i\\segmentation_labeled.nii'
```

So, I believe the cause of the initial error is due to attempting to overwrite a file that has a currently-open file pointer. 

I then found that open file-pointers are often caused via either:

1. The `im_file` property of SCT's `Image` objects, which [stores the original `nibabel` object](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/image.py#L405) inside of our `Image` objects.
    * I was able to get rid of some errors in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/e50e7f3ea17a85a84963e78cd94609f9317bf858 by refactoring `Image` to no longer store the `nibabel` object.
2. Nibabel loading the data array as an `np.memmap` object, rather than loading it into memory, then we [store it](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/image.py#L406).
    * I was able to get rid of some additional errors in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/23738acb278291cb37ad867a2cada82c0ad7850d by conditionally loading the `np.memmap`'d data into memory before saving.

However, even with these two changes, 4 test failures still remain. Related to 2., there is a much bigger hammer we could use here: _Unconditionally_ avoiding memory-mapped data using the following option:

```diff
- self.im_file = nib.load(path)
+ self.im_file = nib.load(path, mmap=False)
```

Making this change indeed fixes all of the `Invalid argument` errors. However, I'm unsure of the implications here, and wanted to discuss beforehand to see if we could identify any other approaches.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR applies two separate fixes relating to memory-mapped data arrays:

1. Avoid `mmap` on Windows.
2. Stop storing nibabel objects (`self.im_file`) in `Image` classes
   - This change has a complicated history, and is related to [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3271#issuecomment-806197324).
   - It also may not even be necessary, given 1.? I would love to compare the test failures with and without this commit.

## Linked issues

Fixes #3695.


spinalcordtoolbox/image.py
```diff
@@ -265,7 +265,7 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
         """
 
         # initialization of all parameters
-        self.im_file = None
```
mguaypaq, line 268: From a quick code search, it seems that the `im_file` attribute is indeed only used in these two files, and mostly just for its `im_file.affine` attribute.

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -311,9 +312,9 @@ def main(argv=None):
     elif arguments.header is not None:
         header = im_in.header
         # Necessary because of https://github.com/nipy/nibabel/issues/480#issuecomment-239227821
-        if hasattr(im_in, "im_file"):
```
mguaypaq, line 314: I'm curious as to why the `hasattr` check was here to begin with. It looks like you added it in #3317, do you remember why?

mguaypaq (approved):
LGTM. Not keeping around an open file object seems like a good idea in general.

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: I _think_ my logic was that if `sct_image -header` was ever called on an image that wasn't loaded from a file, then the `Image` object wouldn't have the `im_in` attribute?

But, that makes no sense, since `im_in` will only ever be created from files:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/598aa433e19375abed4175d5b7350a9e6b5f15be/spinalcordtoolbox/scripts/sct_image.py#L206

So (prior to this PR) it would _always_ have the `im_in` attribute, making the `hasattr` check useless. 🤔

spinalcordtoolbox/image.py
```diff
@@ -364,7 +364,6 @@ def __deepcopy__(self, memo):
     def copy(self, image=None):
         from copy import deepcopy
         if image is not None:
-            self.im_file = deepcopy(image.im_file)
```
joshuacwnewton, line 367: Just realizing I missed copying the affine here, since `self.im_file` is being replaced by `self.affine`.

PR #3719, opened 2022-03-07
Add better support for shell script batch processing on Windows

### LINKED ISSUE #3682: Support installing SCT natively on Windows

### LINKED ISSUE #3696: `sct_run_batch` shell script usage is incompatible with native Windows installations
> * **Note:** This issue is the cause of 3 of the 13 remaining test failures described in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3682#issuecomment-1042079185.

## Context

`sct_run_batch` operates by parallel-executing shell scripts containing SCT CLI commands:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1766ed3f63b4d83fecb7a382b357a2427d8e6c2/spinalcordtoolbox/scripts/sct_run_batch.py#L208-L211

Naturally, shell scripts aren't executable on Windows, so we run into the following error in our test suite:

```
___________________________ test_separate_sessions ____________________________
"""
  File "c:\users\joshua\repos\spinalcordtoolbox\spinalcordtoolbox\scripts\sct_run_batch.py", line 211, in run_single
    stderr=subprocess.STDOUT)
  File "C:\Users\Joshua\AppData\Local\Programs\Python\Python37\lib\subprocess.py", line 488, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Users\Joshua\AppData\Local\Programs\Python\Python37\lib\subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "C:\Users\Joshua\AppData\Local\Programs\Python\Python37\lib\subprocess.py", line 1207, in _execute_child
    startupinfo)
OSError: [WinError 193] %1 is not a valid Win32 application
```

(There are 2 other tests that fail for similar reasons.)

We could skip or rewrite these tests for Windows, of course. But, beyond the tests themselves, we would need to come up with a way to run typical batch-processing-style workflows on Windows. 

## Possible solutions

1.  Using a [bash port for Windows](https://stackoverflow.com/a/26523241) (win-bash, Git Bash, Cygwin):

    > The top Google hit is http://win-bash.sourceforge.net/ but you also need to understand that most Bash scripts expect a Unix-like environment; so just installing Bash is probably unlikely to allow you to run a script you found on the net, unless it was specifically designed for this particular usage scenario. The usual solution to that is https://www.cygwin.com/ but there are many possible alternatives, depending on what exactly it is that you want to accomplish.

2. Alternatively, our users may instead be able to write [batch scripts](https://en.wikipedia.org/wiki/Batch_file) (`.bat`, not `.sh`) to accomplish a similar purpose, but this too would need to be tested.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR fixes 4 separate issues relating to `sct_run_batch` and its tests:

1. `sct_run_batch` needs to explicitly call `bash.exe` on Windows, rather than implicitly treating `.sh` scripts as executable.
    - And for this to happen, we need to explicitly _avoid_ Windows' default System32 `bash.exe`, because it's a misleading trap!
2. Many `sct_run_batch` scripts written for our UNIX installs rely on the `SCT_DIR` environment variable we set during installation, but that we _don't_ set during Windows installation.
3. `test_sct_run_batch` was doing funky things with TemporaryFiles
4. `test_sct_run_batch` included an assertion that doesn't work on Windows

## Linked issues

Fixes #3696.
Fixes #3682. 🎉 

- [x] Dependent on #3725 (to partially fix `test_only_one_include` test failure)
- [x] Dependent on #3720 (to partially fix `test_separate_sessions` test failure)

### DIFF
--- a/testing/cli/test_sct_run_batch.py
+++ b/testing/cli/test_sct_run_batch.py
@@ -104,10 +104,6 @@ def test_non_executable_task(tmp_path, dummy_script):
     # > (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value).
     # > All other bits are ignored.
 
-    # More broadly speaking, however: Don't the other tests also accomplish this same functionality?
-    # i.e. Wouldn't _all_ dummy scripts we create start out as non-executable? I suspect that this test
-    # could be combined with the other tests and there wouldn't be a meaningful loss of coverage.
-
     sct_run_batch.main(['-include', '^t.*',
                         '-subject-prefix', '',
                         '-path-data', str(data), '-path-out', str(out),


joshuacwnewton:
Now that all of the other bugfixes have been merged, I believe we can finally take out the bit of code that hides errors, and actually, fully test Windows (Native). I'll do that now!

testing/cli/test_sct_run_batch.py
```diff
@@ -12,46 +13,43 @@
...
     SUBJECT=$1
     echo $SUBJECT
     """
-    # indexing removes beginning newline
-    script.write(dedent(script_text)[1:])
-    script.flush()
+    with open(path_out, 'w') as script:
+        # indexing removes beginning newline
+        script.write(dedent(script_text)[1:])
+        script.flush()
```
mguaypaq, line 28: Is flushing still necessary, given the `with open()` context?

testing/cli/test_sct_run_batch.py
```diff
@@ -1,8 +1,9 @@
 import glob
 import os
 import sys
+import json
+
 import pytest
-from tempfile import NamedTemporaryFile, TemporaryDirectory
```
mguaypaq, line 5: I'm really glad that you're using the `tmp_path` fixture instead now.

testing/cli/test_sct_run_batch.py
```diff
@@ -12,46 +13,43 @@
...
+def test_config_with_args_warning(tmp_path):
+    data = str(tmp_path / 'data')
+    out = str(tmp_path / 'out')
+    config_path = str(tmp_path / 'config.json')
 
-        cnf_text = '{{"jobs": 1, "path_data": "{}", "path_output": "{}"}}'.format(data, out)
-        config.write(cnf_text)
-        config.flush()
+    with open(config_path, 'w') as config:
+        cnf_text = {"jobs": 1, "path_data": data, "path_output": out}
```
mguaypaq, line 38: This is a tiny nitpick, and I like that you're using `json.dump`, but maybe rename `cnf_text` since it's no longer `_text`.

testing/cli/test_sct_run_batch.py
```diff
@@ -12,46 +13,43 @@
...
+        json.dump(cnf_text, config)
 
-        with pytest.warns(UserWarning, match=r'-config.*discouraged'):
-            # I'm not sure how to check that argparse is printing the right error here, but I trust
-            with pytest.raises(FileNotFoundError):
-                sct_run_batch.main(['-c', config.name, '-include', 'something', '-script', 'script'])
+    with pytest.warns(UserWarning, match=r'-config.*discouraged'):
+        # I'm not sure how to check that argparse is printing the right error here, but I trust
+        with pytest.raises(FileNotFoundError):
+            sct_run_batch.main(['-c', config_path, '-include', 'something', '-script', 'script'])
```
mguaypaq, line 44: Why not use your new `dummy_script` fixture? From the name of this test, it's more about the `UserWarning` than about the `FileNotFoundError`. (I'm assuming the `FileNotFoundError` is about `-script script`, rather than, say, non-existent data.)

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
-    with \
-        TemporaryDirectory() as data,\
-            TemporaryDirectory() as out:
-        with pytest.raises(AssertionError, match='Only one'):
-            sct_run_batch.main(['-include', 'arg', '-include-list',
-                                'arg2', '-path-data', data, '-path-out', out
-                                , '-script', out])
+def test_only_one_include(tmp_path):
+    data = tmp_path / 'data'
+    data.mkdir()
```
mguaypaq, line 70: This line is different from the previous test setups. I guess it's to prevent the `FileNotFoundError`?

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
+    out = tmp_path / 'out'
+    with pytest.raises(AssertionError, match='Only one'):
+        sct_run_batch.main(['-include', 'arg', '-include-list', 'arg2',
+                            '-path-data', str(data), '-path-out', str(out), '-script', str(out)])
```
mguaypaq, line 74: It would be nice to use your `dummy_script` fixture here instead of `'-script', str(out)`

testing/cli/test_sct_run_batch.py
```diff
@@ -12,46 +13,43 @@
...
     echo $SUBJECT
     """
-    # indexing removes beginning newline
-    script.write(dedent(script_text)[1:])
-    script.flush()
+    with open(path_out, 'w') as script:
+        # indexing removes beginning newline
+        script.write(dedent(script_text)[1:])
+        script.flush()
+    return path_out
```
mguaypaq, line 29: After reading `test_non_executable_task` below, it occurs to me that maybe this fixture also needs a `chmod` to make the new script executable?
Edit: `sct_run_batch` explicitly sets the executable bit, so maybe it's not needed, but it would be good manners.

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
-        """
-        script.write(dedent(script_text)[1:]) #indexing removes beginning newline
+    out = tmp_path / 'out'
+
+    script_text = """
+    #!/bin/bash
+    echo $SUBJECT
+    """
+    with open(tmp_path / "dummy_nonexectuable_script.sh", 'w') as script:
+        script.write(dedent(script_text)[1:])  # indexing removes beginning newline
```
mguaypaq, line 86: You don't need this anymore, given the `dummy_script` fixture. (Also, this version is missing the `SUBJECT=$1` line. This wouldn't cause the bash script to return an error though, so I don't think this difference is what the test is about.)

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
         script.flush()
 
-        assert not os.access(script.name, os.X_OK), "Script already executable"
+    # The assertion below is meant to ensure that the script is not executable at first,
```
mguaypaq, line 89: Instead of asserting it, how about doing `os.chmod` to ensure it?

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
+    # so that we can test whether `sct_run_batch` properly changes the permissions. However,
+    # checking against os.X_OK isn't compatible with Windows. So, I've commented it out:
+
+    # assert not os.access(script.name, os.X_OK), "Script already executable"
+
+    # More broadly speaking, however: Don't the other tests also accomplish this same functionality?
+    # i.e. Wouldn't _all_ dummy scripts we create start out as non-executable? I suspect that this test
+    # could be combined with the other tests and there wouldn't be a meaningful loss of coverage.
```
mguaypaq, line 97: I'm similarly puzzled.

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
 
-        sct_run_batch.main(['-include', '^t.*',
-                            '-subject-prefix', '',
-                            '-path-data', data_path, '-path-out', out,
-                            '-script', script.name,
-                            '-continue-on-error', 0])
+    sct_run_batch.main(['-include', '^t.*',
+                        '-subject-prefix', '',
+                        '-path-data', data_path, '-path-out', str(out),
+                        '-script', script.name,
```
mguaypaq, line 102: I'm a little bit surprised that `script.name` is still fine to use after the file is closed (by being done with the `with` block above), but I guess nothing **un**sets it. Either way, shouldn't it be `dummy_script` instead, as returned by the fixture?

testing/cli/test_sct_run_batch.py
```diff
@@ -61,70 +59,73 @@ def test_config_extra_value_warning():
...
-        assert 'sub-01' in open(file_log, "r").read()
-
-
-def test_separate_sessions():
+    data = tmp_path / 'data'
+    data.mkdir()
+    out = tmp_path / 'out'
+    # Create dummy BIDS directory with sessions
+    os.makedirs(os.path.join(data, 'sub-01', 'anat'))
+    os.makedirs(os.path.join(data, 'sub-02', 'anat'))
```
mguaypaq, line 113: Since `tmp_path` is a pathlib-type object, you could use the same idiom as above, `data / 'sub-01' / 'anat'` instead of `os.path.join(...)`. I believe `os.makedirs()` will accept it as-is, even without wrapping it in a `str(...)`.

mguaypaq (changes requested):
Thanks again for all these Windows compatibility fixes!
Looking at `test_sct_run_batch.py`, it's a strange test suite, but after this PR it should work as well on Windows as it already does on other platforms :stuck_out_tongue:
I'm putting "Request changes" because you still have a few inconsistencies in the use of `dummy_script`.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -202,10 +222,29 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
         'ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS': str(itk_threads),
         'SCT_PROGRESS_BAR': 'off'
     })
+    if 'SCT_DIR' not in envir:  # For native Windows installations, the install script won't add SCT_DIR
+        envir['SCT_DIR'] = __sct_dir__
+
+    cmd = [script_full, subj_dir] + script_args.split(' ')
+
+    # Make sure that a Windows-compatible bash port is used to run shell scripts
+    if sys.platform == "win32":
```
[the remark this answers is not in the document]
joshuacwnewton: I'm sorry! I think I had settled on this batch of changes (across all of the different PRs) days/weeks ago, and by the time I got the memo on `.startswith`, I had moved onto other things and forgot to retroactively amend all the existing PRs I had opened. :sweat_smile: 

Thank you for your diligence, though!

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -202,10 +222,29 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
...
+        with open(script_full) as f:
+            shebang = f.readline()
+        if script_full.endswith('.sh') or "sh" in shebang:
```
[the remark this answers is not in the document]
joshuacwnewton: Thanks for pointing this out!! How about the `^#!.*\/.*sh\b.*` regex?

image

testing/cli/test_sct_run_batch.py
joshuacwnewton: [Definitely not.](https://stackoverflow.com/a/51300280) Good catch. :)

testing/cli/test_sct_run_batch.py
joshuacwnewton: Good call! 

In fact, this even revealed another bug in the test, since now this test actually _runs_ `sct_run_batch` using the `path_data` and `path_output` inputs. (And, I forgot to `mkdir` the `data` and `out` directories, aha.)

testing/cli/test_sct_run_batch.py
joshuacwnewton: Close! The previous `FileNotFoundError` was for the `-script script` nonsense, but `data.mkdir` is to prevent:

```
NotADirectoryError: [WinError 267] The directory name is invalid
```	

Which happens when you pass a non-existent folder to `path_data`.

joshuacwnewton: And, as you sniffed out, the fact that `data.mkdir` was missing from the previous test causes an error: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3719#discussion_r836806343

testing/cli/test_sct_run_batch.py
joshuacwnewton: Definitely agree!

testing/cli/test_sct_run_batch.py
joshuacwnewton: This works (as in, doesn't throw an error on Windows like the assertion does), but it's worth noting:

> Although Windows supports [chmod()](https://docs.python.org/3/library/os.html#os.chmod), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

I agree that this is a cleaner cross-platform way of ensuring that the script is non-executable, though. :)

spinalcordtoolbox/scripts/sct_run_batch.py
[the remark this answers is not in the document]
mguaypaq: It's very much a matter of personal taste, but since you asked: I'm partial to `re.match(r'#!\S*sh\b', first_line, re.ASCII)`, personally, because
* `re.match()` looks at the start of the string, hence no initial `^`.
* `/` isn't technically required?
* `.*` matches whitespace, but I think we want the first word to end in `sh`, so `\S*` is more accurate.
* I just checked in my terminal, and non-ascii whitespace seems to be non-word-splitting, hence the `re.ASCII` flag.

testing/cli/test_sct_run_batch.py
joshuacwnewton: I _think_ with the new `chmod` commands (to both ensure that `dummy_script` starts as executable, and ensure that it becomes non-executable during this test), then my concerns are assuaged?

joshuacwnewton:
Thank you so much for such a thorough review, @mguaypaq! 🎉 

Apologies for being so inconsistent across my rewrite of the tests -- things should be a little more even now. :)

testing/cli/test_sct_run_batch.py
```diff
@@ -60,71 +72,72 @@ def test_config_extra_value_warning(tmp_path):
...
+    """
+    Test that individual subjects (i.e. not in 'ses' subfolders) can be processed individually by sct_run_batch.
+    """
     data = tmp_path / 'data'
     data.mkdir()
     out = tmp_path / 'out'
-    # Create dummy BIDS directory with sessions
-    os.makedirs(os.path.join(data, 'sub-01', 'anat'))
-    os.makedirs(os.path.join(data, 'sub-02', 'anat'))
+    # Create dummy BIDS directory without sessions
```
mguaypaq, line 121: Thanks for correcting this comment!

mguaypaq (approved):
This is great, thanks! I like that you added some actual docstring descriptions to the tests. Also, I'm learning that `pathlib` is convenient and has a very clean syntax.

joshuacwnewton:
> Also, I'm learning that `pathlib` is convenient and has a very clean syntax.

Very much so! I like it a lot, but at the same time, the conversion between `Path` objects and strings can be a bit fussy. I'm still not entirely decided on `Pathlib` vs. `os.path`? 

I feel like `Pathlib` makes sense for out tests (almost entirely because of `tmp_path` being a `Path` object), but `os.path` makes sense for our API (because we're often working with strings due to `argparse`, unless we were to implement some custom path parsing action that spits out `Path`s).

But I'm going off on a tangent now... merging this PR! :)

Issue #732, opened 2016-02-11
Call sct tools within the code without system calls


poquirion:
- [x] finalise debug the main()/get_parser() paradigm
- [ ] rewrite the the lines where you have system calls

jcohenadad:
important to printout the "bash equivalent" command for easier debugging within bash terminal.

poquirion:
I am thinking of adding a dev or verbose option to get those print for developer only

PR #3724, opened 2022-03-07
Replace `run_proc` calls with `main` calls across SCT's scripts

### LINKED ISSUE #3711: Replace existing `run_proc` calls with `sct_script.main()` calls
About a year ago, we modified our CLI scripts so that their `main()` functions could be called in-code. For example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c5f194410605dd9c5035d704e1656930830a9cdc/spinalcordtoolbox/scripts/sct_image.py#L557

However, before we had this capability, we would use `run_proc`, which is basically just a wrapper for `subprocess`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c5f194410605dd9c5035d704e1656930830a9cdc/spinalcordtoolbox/scripts/sct_smooth_spinalcord.py#L210

We've slowly been phasing out these `run_proc` calls with `main()` calls, because they're faster and easier to trace with a debugger. We just haven't opened an issue for this yet, I think?

joshuacwnewton:
## Description

In many places, we use `run_proc` to execute an SCT CLI script in a subprocess. But, ever since https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3091, we've gained the ability to call the script within the same process using the script's `main()` function instead.

So, this PR replaces existing `run_proc` calls with `main` calls, so the majority of SCT's processing is done within the same process.

## Linked issues

Fixes #3711.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/732.


joshuacwnewton:
Huh. Now that some of the other bugfixes have been merged (e.g. #3718), I'm realizing that the changes in this PR don't... actually fix any failing tests? 

Right now, there are 9 failing tests on master, and 9 failing tests on this branch, and they fail in identical ways...

So, I'm not really sure what I was going on about in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3711#issuecomment-1060810502? 

I'm going to temporarily close this PR just to indicate that it's not critical for Windows compatibility, and to save @mguaypaq some review effort. But, we may still want to consider addressing #3711 outside of the Windows effort...

spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
```diff
@@ -113,28 +115,29 @@ def main():
...
-              "-i", "data.nii",
-              "-x", "linear",
-              "-vox", str(nx * interp_factor) + 'x' + str(ny * interp_factor) + 'x' + str(nz * interp_factor),
-              "-o", "data_up.nii"], verbose)
+    sct_resample.main([
+        "-i", "data.nii",
+        "-x", "linear",
+        "-vox", str(nx * interp_factor) + 'x' + str(ny * interp_factor) + 'x' + str(nz * interp_factor),
+        "-o", "data_up.nii",
+    ])
```
mguaypaq, line 123: Here (and a few other places), you're dropping the `verbose` parameter instead of passing it in with a `-v` option, is that intentional?

I guess `run_proc` itself didn't pass it on before, so the only difference is that the diagnostic log message generated by `run_proc` doesn't get printed anymore.

spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
```diff
@@ -113,28 +115,29 @@ def main():
...
-              "-i", "data_up.nii",
-              "-s", "data_up.nii",
-              "-smooth", str(smoothing_sigma),
-              "-r", str(remove_temp_files),
-              "-v", str(verbose)], verbose)
+    sct_smooth_spinalcord.main(["-i", "data_up.nii",
+                                "-s", "data_up.nii",
+                                "-smooth", str(smoothing_sigma),
+                                "-r", str(remove_temp_files),
+                                "-v", str(verbose), ])
```
mguaypaq, line 131: I think the trailing comma here is unusual, since the closing brackets are on the same line. Using the same indentation style as a few lines below would be my favourite fix:
```suggestion
    sct_smooth_spinalcord.main([
        "-i", "data_up.nii",
        "-s", "data_up.nii",
        "-smooth", str(smoothing_sigma),
        "-r", str(remove_temp_files),
        "-v", str(verbose),
    ])
```

spinalcordtoolbox/scripts/isct_test_ants.py
```diff
@@ -108,14 +110,15 @@ def main():
     #
     # # Apply deformation to input image
     # printv('\nApply transformation to input image...', verbose)
-    # run_proc('sct_apply_transfo -i data_src.nii.gz -o data_src_warp.nii.gz -d data_dest.nii.gz -w warp_curve2straight.nii.gz -p nn', verbose)
+    # sct_apply_transfo(["-i", "data_src.nii.gz", "-o", "data_src_warp.nii.gz", "-d", "data_dest.nii.gz", "-w", "warp_curve2straight.nii.gz", "-p", "nn"])
```
mguaypaq, line 113: I like that you're also fixing the commented out code, but there's a missing `.main` here:
```suggestion
    # sct_apply_transfo.main(["-i", "data_src.nii.gz", "-o", "data_src_warp.nii.gz", "-d", "data_dest.nii.gz", "-w", "warp_curve2straight.nii.gz", "-p", "nn"])
```

mguaypaq (approved):
LGTM, thanks for the cleanup.

spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
joshuacwnewton: Ah oops! That definitely wasn't intentional. :sweat_smile:

spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
joshuacwnewton: Ah! Thank you for raising this point. I'm curious about the before/after output logging -- I want to double-check this before merging. :slightly_smiling_face:

joshuacwnewton: Ah, right. So... it turns out that switching from `run_proc` to `main()` does significantly increase the amount of logging that occurs. (For example, here's a diff of the output for [`sct_label_vertebrae`](https://www.diffchecker.com/wVzo0QkQ) before and after.)

I'm fairly certain the reason for this is because of:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7ce5ea59e213510ff0566eef99d3cefe811f0051/spinalcordtoolbox/utils/sys.py#L406-L407

While the default value of `verbose` is:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7ce5ea59e213510ff0566eef99d3cefe811f0051/spinalcordtoolbox/utils/sys.py#L368

(And, the default `-v` value on (most? all?) scripts is also 1.)

Meaning that, for the previous `run_proc` behavior, the full output of each subcall to a CLI script is hidden by default, and only shown when `-v 2` is passed.

Now, however, with the new `.main()` calls, the full output of each subcall is shown by default, which results in _much_ more verbose output.

joshuacwnewton: I'm fairly certain this is responsible for the uptick in `display_viewer_syntax` output I noticed as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3877.

But, that also means the scope of the problem is much larger than just `display_viewer_syntax`.

joshuacwnewton: Just for comparison's sake, here is a diff comparison between the old `run_proc` and `main(['-v', '0'])`: https://www.diffchecker.com/z11yJBpK

It's not a 1:1 match, but it's a lot closer? As for the differences: It looks like we don't propagate `-v 0` perfectly throughout our scripts, because I would assume `-v 0` would silence everything but warnings/errors, yet we're still getting "info" messages.

joshuacwnewton: I would maybe suggest to merge this as-is, then follow up with a PR that adds `-v 0` to all internal calls to SCT scripts, while also making sure that we propagate `verbose` to `printv` properly.

mguaypaq: Oh that's tricky! `run_proc` does a bunch of filtering and post-processing of the called command's output. The real problem here is that we're not supposed to be calling scripts, even through the `main()` function, and should be using something in the API layer. (Another effect of calling `main()` is that `init_sct()` and `set_loglevel()` have to be careful about when to modify global state.) But that's certainly out of scope for this PR.

I have a feeling that [this older comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3341#issuecomment-817348176) is pointing in the right direction: we could be making better use of [the `logging` module](https://docs.python.org/3/library/logging.html)'s abilities. Possibly even the colour-coding currently handled by `printv()` could be turned into a log formatter. But that would be a much larger project.

Since some calls to `run_proc()` have already been converted to `main()`, we could finish the transition with this PR, and deal with the remaining issues separately.

But having just looked around for uses of `run_proc()` without `is_sct_binary=True`, there's still:
* A recursive call inside `run_proc()` to `sct_download_data`, and
* A few calls from `sct_check_dependencies`, which would be out of scope for this PR.

joshuacwnewton: > A recursive call inside `run_proc()` to `sct_download_data`

Thank you for acknowledging this! I tried converting this one, too, but when trying to run `pytest`, I ran into a circular import error (due to 2 instances of `from module import function` importing):

```
(venv_sct) joshua@tadpole:~/repos/spinalcordtoolbox/data/sct_testing_data/t2$ sct_testing

[...]

  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/__init__.py", line 1, in <module>
    from .utils import __version__, __sct_dir__, __data_dir__, __deepseg_dir__  # noqa: F401
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/__init__.py", line 1, in <module>
    from .fs import *  # noqa: F401, F403
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/fs.py", line 11, in <module>
    from .sys import printv
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/sys.py", line 21, in <module>
    from spinalcordtoolbox.scripts import sct_download_data
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_download_data.py", line 17, in <module>
    from spinalcordtoolbox.download import install_data
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/download.py", line 13, in <module>
    from spinalcordtoolbox.utils import sct_progress_bar, tmp_create
ImportError: cannot import name 'sct_progress_bar' from 'spinalcordtoolbox.utils' (/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/__init__.py)
```

I can think of the following options for solving this:

1. Use the CLI `main()` call, but lazy-load the `sct_download_data` script.
2. Use the CLI `main()` call, but stop using `from module import function` imports for all of the `utils` imports in `sct_download_data` _and_ `download.py`.
3. Get rid of the CLI `main()` call and replace it with a `download.py` API call, then stop using `from module import function` imports for all of the `utils` imports in _just_ `download.py`
4. Get rid of data-downloading functionality from `run_proc` entirely. _(Should downloading datasets really be the responsibility of this function?)_

Options 3./4. feels like the best long-term solution, since ideally, we would be eliminating every call to CLI `.main()` functions. 

But, in the short-term 1. feels like the lightest way to keep existing functionality?

joshuacwnewton: > A few calls from sct_check_dependencies, which would be out of scope for this PR.

Yes, exactly. My goal for this PR was simply to split the `sct_<CLI script>` calls from all other `run_proc` calls. And, the calls in `sct_check_dependencies` are for:

* `date`
* `whoami`
* `pwd`
* `fsleyes`
* `isct_propseg`
* `isct_test_ants`

The first 5 don't even have main calls, because they're external binaries. However, `isct_test_ants` does indeed have a `main()` function we can call, so thank you for spotting this, too!

joshuacwnewton: Actually, I think I want to suggest removing the data-downloading function from `run_proc` entirely. As currently written:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7ce5ea59e213510ff0566eef99d3cefe811f0051/spinalcordtoolbox/utils/sys.py#L376-L377

The conditional checks to see if the `bin/` directory exists. But, `bin/` is where _SCT's CLI script entry points_ are contained! So if `bin/` doesn't exist, then SCT as a whole wouldn't even function in the first place.

In other words, if we're running code here, then `bin/` will _always_ exist, meaning this LOC has probably never run at all.

joshuacwnewton: > Since some calls to run_proc() have already been converted to main(), we could finish the transition with this PR, and deal with the remaining issues separately.

I've opened https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3881 as a follow-up for this PR.

mguaypaq: Oh wow, my eyes just skipped over the `os.path.isdir(__bin_dir__)` test completely; I had assumed that there was actually a case where some binaries could be missing and `run_proc` would helpfully try to fetch them. In that case option 4 is great, let's remove that dead code!

mguaypaq (approved):

PR #3727, opened 2022-03-07
Stop using a tempfile for data reorientation

### LINKED ISSUE #3714: `reorient_data()` repeatedly overwrites the same temporary file, causing `FileExistsError`
> **Note:** This issue represents a Windows-specific bug that has already been addressed in #3699.

`sct_analyze_texture` contains the following function:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2c2a662baf6eadc0c28711886ec354a4c8d14016/spinalcordtoolbox/scripts/sct_analyze_texture.py#L278-L283

This is a little hard to parse, but what's happening here is:

1. The input file `self.fname_metric_lst[f]` gets renamed to a temporary filename
2. The temporary file is then re-saved back to the original filename (`self.fname_metric_lst[f]`), but with its orientation changed.

The problem is with step 1. That temporary file persists, and will remain present for each `f in self.fname_metric_lst`. Because of this, the following error is thrown on Windows:

```
FileExistsError: [WinError 183] Cannot create a file when that file already exists
```

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR deletes an unused temporary file in order to avoid overwriting the same file repeatedly. Without doing this, on Windows machines, the following error is thrown:

    FileExistsError: [WinError 183] Cannot create a file when that file already exists

I don't think the intermediate file is used for anything, so I think this is OK to do.

I wonder if using a temporary file is even necessary, though? Why can't we just reorient in-place? 🤔 

## Linked issues

Fixes #3714.


mguaypaq (changes requested):
I agree with your question, "Why can't we just reorient in-place?"
Looking at `Image.change_orientation()`, it returns a new Image object with an empty path field (as documented):
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/84060e21e95ec75a0b0c2df195ac1d95a8ca5fab/spinalcordtoolbox/image.py#L1061-L1063

So I think you could get rid of the rename, but you do have to specify the filename when saving the re-oriented copy.

Alternatively, the docstring for `Image.change_orientation` explicitly suggests setting `im_dst` to the input image for in-place operation. From skimming the code, I don't see any obvious reason why this would fail, so that might be the best option.

(As to "why is the code written this way", this business about renaming the file dates back to the initial commit from 5 years ago, so I don't know.)

joshuacwnewton:
> Looking at `Image.change_orientation()`, it returns a new Image object with an empty path field (as documented):

It looks like there's two `change_orientation` functions: The `image.py` function, as well as the `Image` class method:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/84060e21e95ec75a0b0c2df195ac1d95a8ca5fab/spinalcordtoolbox/image.py#L425-L437

And, it looks like the method supplies `self` for `im_dst`. So, I think [73c7880](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3727/commits/73c7880cd60b23eb76e1fc10f00097675331cf11) should be sufficient?

mguaypaq (commented):
This looks so clean! Thanks :smile:

Issue #3059, opened 2020-11-24
clean_labeled_sgementation add extr slices to the labeled segmentation


lrouhier:
### Description
When using `sct_label_vertebrae`, you can notice that the slice at the top does not really align with the first disc. I think that this is due to the `clean_labeled_segmentation` function and more precisely [these lines]
(https://github.com/neuropoly/spinalcordtoolbox/blob/5eae6da1e97a42cfaf8780a18fa61ecf3c95d6f1/spinalcordtoolbox/vertebrae/core.py#L318-L327) ). 
The first line actually creates a dilation with factor 2, which would explain the two added slices at the top of the segmentation.
This gif is here to explain: 
![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/22171999/99007032-64c16600-2511-11eb-8b28-372faedd4b82.gif)

here you see what we obtain with the previous  `run sct_label_vertebrae -i sub-amu01_T2w.nii.gz -s ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_seg.nii.gz -c t2 -discfile ../../derivatives/labels/sub-amu01/anat/sub-amu01_T2w_labels-disc-manual_modified.nii.gz -r 0 -ofolder keep_files_modified` 
* in pale green is the segmentation obtained with the current code. 
* In yellow you can see what we get when we comment [this line](https://github.com/neuropoly/spinalcordtoolbox/blob/99ec9229dbbecff6f9435c9a30adf4676a7408e3/scripts/sct_label_vertebrae.py#L434) so we don't use clean_labeled_segmentation. 

This was first found in PR #2990 (see https://github.com/neuropoly/spinalcordtoolbox/pull/2990#issuecomment-725706405)

PR #3739, opened 2022-03-21
Improve the label cleaning function

### LINKED ISSUE #3702: Consider setting `-clean-labels 1` by default
Context: https://forum.spinalcordmri.org/t/seg-and-label-seg-mismatch/820/5?u=jcohenadad

Considering that:
- these segmentation issues could happen (ie: spurious voxels in the labeled segmentation)
- many users are not as careful as the user who reported the issue (ie: it requires QCing the selected slices, understand what it means)

I suggest to set `-clean-labels 1` by default.

I would like it if we could trace back the decision we set it to 0 (maybe there was a good reason to do so...).

mguaypaq:
#### PR contents

## Description
This PR adds an intermediate value for the option `sct_label_vertebrae -clean-labels` and sets it as the new default.
The option exists because the set of voxels in the generated labeled segmentation may not match the set of voxels in the original (unlabeled) segmentation: the generated segmentation may have extra voxels and/or missing voxels, compared to the original segmentation. "Cleaning" tries to make the generated segmentation match the original.
Before this PR, the choices of argument for `-clean-labels` were:
- `0`: do no cleaning (the old default).
- `1`: add missing voxels and remove extra voxels.

However, both `-clean-labels 0` and `-clean-labels 1` can have surprising behavior (see [a forum post](https://forum.spinalcordmri.org/t/seg-and-label-seg-mismatch/820/5) and #3059).
After this PR, the choices of argument for `-clean-labels` will be:
- `0`: do no cleaning.
- `1`: remove extra voxels (the new default).
- `2`: also add missing voxels.

The new default value would have prevented the issue described on the forum post.

This PR also changes the algorithm for adding missing voxels. The previous code claims to use the nearest label value, but instead it only slightly enlarges the region covered by each label value, with higher label values encroaching on top of lower label values. This may fix the strange behavior of `-clean-labels 2`, making it safer to use.

## Linked issues
Fixes #3702.
Related to #3059.


mguaypaq:
Note to self: I should update the tests and check for any usage of `-clean-labels` in ReadTheDocs.

mguaypaq:
@joshuacwnewton while writing the PR description and labels, I got the feeling that this is really two PRs. I can split the change in two if you prefer. I'm also open to the suggestion that `-clean-labels 2` should be the new default (and the added cleaning option should be removed), if the weirdness from #3059 is indeed fixed by the new algorithm.

joshuacwnewton (commented):
> @joshuacwnewton while writing the PR description and labels, I got the feeling that this is really two PRs. I can split the change in two if you prefer.

@mguaypaq: Given the above comment of yours, I wanted to try splitting your commit [2e0cec2](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3739/commits/2e0cec218378a439bde4ababf71ad0bc50509b34) into multiple different commits, while still preserving the exact state of the work you did:

```console
$ git diff baac3be71ca70c3e53e68e17ac6d1ea4d6f36829..2e0cec218378a439bde4ababf71ad0bc50509b34

(returns nothing)
```

My goal was to use this as an exercise to better understand all of the different changes in your PR, so that I could review it more effectively. I hope you don't mind? If this was too intrusive of me, my sincere apologies!!

I also had a few small comments on the changes you made, too. :)

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -308,36 +309,44 @@ def get_z_and_disc_values_from_label(fname_label):
...
+    :return: None
+    """
+    img_labeled_seg = Image(fname_labeled_seg)
+    img_seg = Image(fname_seg)
+    # img_seg voxel values are either 0 or 1
+    img_labeled_seg.data *= img_seg.data
+    img_labeled_seg.save()
+
+
+def clean_extra_and_missing_labels(fname_labeled_seg, fname_seg):
```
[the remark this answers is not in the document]
mguaypaq: Done, thanks for the suggestion. There's already a `TODO` comment at the top of the file to stop doing file I/O in the API, so this is progress towards that.

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -308,36 +309,50 @@ def get_z_and_disc_values_from_label(fname_label):
...
-    for i_vox in range(len(ind_nonzero[0])):
-        # assign closest label value for this voxel
-        ix, iy, iz = ind_nonzero[0][i_vox], ind_nonzero[1][i_vox], ind_nonzero[2][i_vox]
-        img_labeled_seg_corr.data[ix, iy, iz] = data_labeled_seg_dil[ix, iy, iz]
-    # save new label file (overwrite)
-    img_labeled_seg_corr.absolutepath = fname_labeled_seg_new
-    img_labeled_seg_corr.save()
+    # img_seg voxel values are either 0 or 1
+    img_labeled_seg.data *= img_seg.data
+    img_labeled_seg.save()
```
joshuacwnewton, line 325: I might be blanking here, but given the refactor into `expand_labels` and `crop_labels`, do we still need this old `clean_extra_labels` function?

testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -52,23 +52,22 @@ def test_sct_label_vertebrae_high_value_warning(caplog, tmp_path):
...
+                                  '-initz', '40,3',
+                                  '-clean-labels', str(i),
+                                  '-ofolder', str(tmp_path/str(i))])
+        im_labeled_seg = Image(str(tmp_path/str(i)/'t2_seg-manual_labeled.nii.gz'))
+        # binarization (because labels are 2.0, 3.0, 4.0)
+        im_labeled_seg.data = im_labeled_seg.data > 0.5
+        dice_score[i] = compute_dice(im_labeled_seg, im_seg)
+    # the cleaner version should be closer to the segmentation
+    assert dice_score[2] >= dice_score[1]
+    assert dice_score[1] >= dice_score[0]
```
joshuacwnewton, line 70: Love the way this is written! ❤️ 

It's such a subtle thing, but not using named variables here makes it really intuitive to see "ah, we're comparing the dice for `2` vs. `1`, and `1` vs. `0`".

joshuacwnewton, line 70: Ack, I did a poor job documenting the chat you had with Julien during the last meeting. 

Could you remind me why we're going with `1` as a default (even though `2` produces a better segmentation, as shown here)?

Does it have to do with `2` being a little more invasive/transformative? (It's additive, rather than simply subtracting non-segmentation voxels, so it should be opt-in?)

joshuacwnewton (approved):
This mostly LGTM!

The one change I requested is simple enough, and I imagine we're not going to reverse the decision on the default value, so I'm giving approval. 🙂

testing/cli/test_cli_sct_label_vertebrae.py
mguaypaq: That's it exactly: `-clean-labels 2` is inventing more data, so it should be opt-in; whereas `-clean-labels 1` is just trimming labels that fall outside the segmentation, which is less surprising. (And `-clean-labels 0` is more of a raw/debugging thing, which should also be opt-in.)

spinalcordtoolbox/vertebrae/core.py
mguaypaq: Thanks for catching this! I just forgot to remove the function after rewriting it. Fixed now.

testing/cli/test_cli_sct_label_vertebrae.py
joshuacwnewton: Perfect! Thank you. 🙂

PR #3757, opened 2022-04-08
Refactor `sys.platform` checks to include Windows and use `.startswith()` idiom

### LINKED ISSUE #3742: Ctrl+F the SCT repo for any instances of `sys.platform` that don't account for Windows
This is a quick sanity check task for Windows compatibility (just writing this down now before I forget).

Recently, we've addressed specific issues related to checking `sys.platform` (i.e. checking if we're on macOS or Linux) where we weren't handling the Windows case. But, in general, we should make sure that all instances of checking `sys.platform` take Windows into account.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR looks for all instances where we check `sys.platform` to make sure that:

* Windows platforms are included whenever necessary
* `sys.platform ==` is discarded in favor of `sys.platform.startswith()`

In the process of doing this, I noticed that there were a few spots where the same check (determining whether to use `xdg-open` or `open` for QC reports, etc.) was being repeated across multiple locations. But, we already have an existing function for this called `display_open()` that we can re-use in each place.

## Linked issues

Fixes #3742.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -235,6 +235,8 @@ def main(argv=None):
         os_running = 'linux'
     elif sys.platform.startswith('win32'):
         os_running = 'windows'
+    else:
+        os_running = 'unknown'
 
     print('OS: ' + os_running + ' (' + platform.platform() + ')')
     print('CPU cores: Available: {}, Used by ITK functions: {}'.format(psutil.cpu_count(), int(os.getenv('ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS', 0))))
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -340,7 +340,7 @@ def main(argv=None):
         print(err)
 
     print_line('Check if figure can be opened with PyQt')
-    if sys.platform == "linux" and 'DISPLAY' not in os.environ:
+    if sys.platform.startswith("linux") and 'DISPLAY' not in os.environ:
         print_fail(" ($DISPLAY not set on X11-supporting system)")
     else:
         try:
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -212,7 +212,6 @@ def main(argv=None):
     # initialization
     install_software = 0
     e = 0
-    os_running = 'not identified'
 
     # complete test
     if complete_test:
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -536,7 +536,7 @@ def main(argv=None):
                           status_message + timing_message)
 
     display_open(file=os.path.join(path_qc, "index.html"),
-                 message="To open the Quality Control (QC) report on a web-browser")
+                 message="To open the Quality Control (QC) report in a web-browser")
 
     if arguments.zip:
         file_zip = 'sct_run_batch_{}'.format(time.strftime('%Y%m%d%H%M%S'))
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -33,7 +33,7 @@ from textwrap import dedent
 import yaml
 import psutil
 
-from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar
+from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, display_open
 from spinalcordtoolbox.utils.sys import send_email, init_sct, __get_commit, __get_git_origin, __version__, __sct_dir__, set_loglevel
 from spinalcordtoolbox.utils.fs import Tee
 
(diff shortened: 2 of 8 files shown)


spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -235,6 +235,8 @@ def main(argv=None):
         os_running = 'linux'
     elif sys.platform.startswith('win32'):
         os_running = 'windows'
+    else:
+        os_running = 'unknown'
```
mguaypaq, line 239: This case was previously covered above, under `# initialization`, by setting the default value `os_running = 'not identified'`. I like your version better because a missing `else` branch is usually a red flag; it's nice to have all the logic for setting `os_running` together. But please remove the (now redundant) default initialization.

spinalcordtoolbox/utils/shell.py
```diff
@@ -17,17 +17,26 @@
 logger = logging.getLogger(__name__)
 
 
-def display_open(file):
+def display_open(file, message="\nDone! To view results"):
```
mguaypaq, line 20: I'm noticing that the default message starts with a `\n` newline, but in the two new calls to the function, there's no initial newline. Is that intentional?
(I would have expected the message parameter to be without newline, and the function's f-strings to add it, but that's just me.)

spinalcordtoolbox/utils/shell.py
```diff
@@ -17,17 +17,26 @@
...
+        if "DOCKER" not in os.environ and "IS_WSL" not in os.environ and "WSL_DISTRO_NAME" not in os.environ:
+            cmd_open = 'xdg-open'
+    elif sys.platform.startswith('darwin'):
+        cmd_open = 'open'
+    elif sys.platform.startswith('win32'):
+        cmd_open = 'start'
+
+    if cmd_open:
+        printv(f'{message}, type:')
+        printv(f"{cmd_open} {file}\n", verbose=1, type='info')
```
mguaypaq, line 36: I'm surprised to see back-to-back calls to `printv` with seemingly different values for `verbose`. It would be very strange to see only half of the message! But thankfully the default value is `verbose=1` anyway.
(The different values for `type=` are to get different colouring, which is useful.)
```suggestion
        printv(f'{message}, type:')
        printv(f"{cmd_open} {file}\n", type='info')
```

spinalcordtoolbox/utils/shell.py
```diff
@@ -17,17 +17,26 @@
...
     else:
-        printv('\nDone! To view results, open the following file:')
-        printv(file + '\n', verbose=1, type='info')
+        printv(f'{message}, open the following file:')
+        printv(f"{file}\n", verbose=1, type='info')
```
mguaypaq, line 39:
```suggestion
        printv(f"{file}\n", type='info')
```

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -535,10 +535,8 @@ def reset_streams():
         send_notification('sct_run_batch: Run completed',
                           status_message + timing_message)
 
-    open_cmd = 'open' if sys.platform == 'darwin' else 'xdg-open'
-
-    print(f'To open the Quality Control (QC) report on a web-browser, run the following:\n'
-          f'{open_cmd} {os.path.join(path_qc, "index.html")}')
+    display_open(file=os.path.join(path_qc, "index.html"),
+                 message="To open the Quality Control (QC) report on a web-browser")
```
mguaypaq, line 539: Grammar nit, while we're at it.
```suggestion
                 message="To open the Quality Control (QC) report in a web-browser")
```

mguaypaq (approved):
Thanks for the cleanup, for revamping `display_open`, and using it more widely!

mguaypaq:
Hmm, the failing tests seem to be unrelated to this pull request:
* The CentOS build is failing to install SimpleITK because of [missing compiler dependencies](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5892021026?check_suite_focus=true#step:5:592).
* The Windows Native build is failing to install SimpleITK because of [filename too long](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5892021952?check_suite_focus=true#step:5:434).
* The ReadTheDocs build failed to install SimpleITK because of [timeout or excessive memory consumption](https://readthedocs.org/api/v2/build/16603839.txt).

My added commit should really not be the culprit, and the tests all passed on the previous commit, so you could merge this while ignoring the failures.

mguaypaq:
Re: the failures described in [the previous comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3757#issuecomment-1093403149), they're being investigated in issue #3758.

spinalcordtoolbox/scripts/sct_check_dependencies.py
joshuacwnewton: Oh, good catch!! I definitely missed that.

spinalcordtoolbox/utils/shell.py
joshuacwnewton: Ahh, another good catch! Thank you. :)

spinalcordtoolbox/utils/shell.py
joshuacwnewton: Oh! This is definitely a silly error on my part. Thank you for catching that. :)

joshuacwnewton:
> Re: the failures described in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3757#issuecomment-1093403149, they're being investigated in issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3758.

All resolved! :)

spinalcordtoolbox/scripts/sct_check_dependencies.py
mguaypaq: Both `e` and `install_software` seem to be 0-1 variables, where 1 indicates that some kind of failure was encountered. I don't think it's a great pattern, but given the current use, they do need to be initialized to zero.

Issue #3508, opened 2021-08-24
Investigate merging `batch_processing.sh` with `batch_single_subject.sh`


joshuacwnewton:
Currently, we maintain two different shell scripts and two different datasets:

1. [`batch_processing.sh`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/batch_processing.sh): This is the script we use to show off pipelines for typical SCT usage. (Uses `sct_example_data`.)
2. [`batch_single_subject.sh`](https://github.com/spinalcordtoolbox/sct_tutorial_data/blob/master/single_subject/batch_single_subject.sh): This is the script that was in the 2020 SCT Course zip. It contains all of the commands in the SCT tutorials, and I've been updating it as part of the effort to migrate the SCT Course to RTD. (Uses `sct_tutorial_data`.)

On reflection, I think these are parallel efforts doing (more or less) exactly the same thing! So, we might be able to merge them, and have a single script + dataset cover everything. (In other words, we would be keeping `batch_processing.sh` in sync with the tutorial commands.)

* This would save us the maintenance effort of having to update multiple different scripts each time SCT syntax/usage changes. (See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3507#pullrequestreview-737689729.) 
* This would also simplify our documentation, as we would have a single page for new SCT users, rather than having both ["Getting Started"](https://spinalcordtoolbox.com/en/latest/user_section/getting-started.html) and ["Tutorials"](https://spinalcordtoolbox.com/en/latest/user_section/tutorials.html).

joshuacwnewton:
Good news! Comparing the two scripts, they're structured mostly identically. The same core sections are present in both scripts. (e.g. T2 labeling/segmentation/registration/metrics, T2* GM/WM seg/reg/metrics, dMRI/fMRI preprocessing/moco, etc.)

There _have_ been many little tweaks between the two scripts over the years, however. For example:

* `batch_processing.sh` generates the metric file `csa_c2c3.csv`, while `batch_single_subject.sh` generates `csa_c3c4.csv`
* `batch_processing.sh` uses C2 and C5 manual vertebral labels, while `batch_single_subject.sh` uses C2 and T2 labels.
* `batch_processing.sh` has the new dMRI/fMRI QC commands, while `batch_single_subject.sh` does not.
* etc.
 
[...]

Issue #3471, opened 2021-07-17
Add full-sized images to `sct_testing_data` (or combine with `sct_example_data`)


joshuacwnewton:
Currently, SCT maintains 2 different datasets for testing

* [`sct_example_data`](https://github.com/spinalcordtoolbox/sct_example_data): Full-sized images, used mainly for `batch_processing.sh`, but not actually available during regular testing.
* [`sct_testing_data`](https://github.com/spinalcordtoolbox/sct_testing_data): A set of pre-processed images (cropped, resampled, and run through SCT tools e.g. segmentation), created to speed up testing. This is downloaded at the start of every `pytest` run.

The problem is, sometimes during testing, we really do want access to the full-sized images (see for example https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3468#discussion_r671696767), and `sct_testing_data` starts to feel unrepresentative of real-world data.

----

At that point, if we are including full-sized images in `sct_testing_data`, then is there still a benefit is there to keeping the two datasets separate? I'm wondering if it would just be easier to keep both types of data (raw, processed) in the same dataset, then use that dataset for both `batch_processing.sh` _and_ our test suite. 

If we do merge them, we would have to think about dataset structure. For example, maybe we could follow a BIDS-like approach and use a `derivatives` folder for the pre-processed images?

joshuacwnewton:
One problem with this is that we're currently redownloading the test dataset on _every_ pytest run. If we increase the size of `sct_testing_data`, then redownloading starts to become a problem.

Related: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2959

joshuacwnewton:
> One problem with this is that we're currently redownloading the test dataset on every pytest run. If we increase the size of sct_testing_data, then redownloading starts to become a problem.

This was fixed by #3480. :slightly_smiling_face:

Issue #3136, opened 2020-12-28
Centralize SCT's test data code (dummy fixtures, 'sct_testing_data/')


joshuacwnewton:
Currently, SCT tests use test data from a few different locations:

* `sct_testing_data/`: Contains real anatomical images
* `spinalcordtoolbox/testing/create_test_data.py`: Contains a few `dummy_<imagetype>` functions
* `unit_testing/test_aggregate_slicewise.py` (and many others): Per-test-module `pytest.fixtures` for dummy image data. There are lots of these scattered about.

I have few questions... My hope is to write some clearer recommendations in the [Tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) developer Wiki page.

* When should we be using real anatomical data vs. dummy data? (I imagine it's context-dependent, but are there any broad recommendations we can give to make writing tests easier for contributors? This would be useful in cases like https://github.com/neuropoly/spinalcordtoolbox/pull/3120#discussion_r549472092)
* Are we even using all of the dummy data functions? 
* Can we centralize the dummy data functions? (It seems a bit scattered to have dummy data in both in the `spinalcordtoolbox` module _and_ in each individual test file. I'm wondering if we could move the fixtures to a centralized file containing test data fixtures, so they're available to all tests. That way, we're more likely to reuse existing dummy data, rather than creating data each time we write a new test.)

joshuacwnewton:
In https://github.com/neuropoly/spinalcordtoolbox/pull/3152#discussion_r553622125, I proposed that some module-scoped dummy test images should instead be prepared in `conftest.py`. This would allow us to use `pytest.mark.parametrize` without having to rely on fixture functions.

joshuacwnewton:
> Can we centralize the dummy data functions? (It seems a bit scattered to have dummy data in both in the `spinalcordtoolbox` module _and_ in each individual test file. I'm wondering if we could move the fixtures to a centralized file containing test data fixtures, so they're available to all tests. That way, we're more likely to reuse existing dummy data, rather than creating data each time we write a new test.)

I've come across the [`pytest-cases`](https://smarie.github.io/python-pytest-cases/#why-pytest-cases) plugin. It provides a really nice way of organizing test data (test cases). It's also well-documented and well-maintained.

[...]

PR #3771, opened 2022-04-15
Set a more permissive threshold for reading the qform

### LINKED ISSUE #3703: ValueError: w2 should be positive
### Description

nibabel's `quaternions.py` function has a check to ensure the quaternions are correct wrt. the affine matrix. However, due to precision issues, this check could fail, hence the presence of the method `nib.Nifti1Header.quaternion_threshold`. In rare instances, the threshold is not sufficient and loading the data fails. 

Full context: https://forum.spinalcordmri.org/t/sct-crop-image-valueerror-w2-should-be-positive/826/10

Related issue: https://github.com/nipy/nibabel/issues/626

### Steps to Reproduce

This bug is data-dependent. The data were sent privately to @jcohenadad. 

### Proposed solutions

1. Add a user-accessible flag to control `nib.Nifti1Header.quaternion_threshold`. Cons: the flag would need to be added to all CLI functions, which is probably not a good idea.
2. Add an environment variable that would be read by SCT's `image.py` API. Cons: environment variables are not common in SCT (ie: users are not aware of them, and don't use them often).
3. Hard-code the line `nib.Nifti1Header.quaternion_threshold = -1e-06` when opening an image. Cons: (i) it bypasses nibabel's decision to _not_ set that threshold value; (ii) it _might_ have an impact on cross-compatibility.

I would personally go with (3).

mguaypaq:
## Description
When cropping a nifti file which has a slightly bad qform (which could happen because of rounding errors), nibabel throws an error. This PR implements proposed solution number 3 from issue #3703, which is to set the image's `quaternion_threshold` to `-1e-6` before getting its qform. (This value comes from a suggestion in https://github.com/nipy/nibabel/issues/626 and seems reasonable, see [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3703#issuecomment-1049221586).)

The PR also adds a test case for an image with a slightly bad qform, which fails before the main code change, and passes afterwards. See https://github.com/spinalcordtoolbox/sct_testing_data/pull/10 for the actual test file. Either this PR can wait until we do a new release of `sct_testing_data`, or else we can split off the new test case as a future PR.

And while I was editing the code, I took the opportunity to simplify the code surrounding `BoundingBox` (with individual refactoring steps broken down as separate commits).

## Linked issues
Fixes #3703

### DIFF
--- a/spinalcordtoolbox/cropping.py
+++ b/spinalcordtoolbox/cropping.py
@@ -13,9 +13,7 @@ logger = logging.getLogger(__name__)
 
 
 class BoundingBox(object):
-    """
-    """
-    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, zmin=None, zmax=None):
+    def __init__(self, xmin, xmax, ymin, ymax, zmin, zmax):
         self.xmin = xmin
         self.xmax = xmax
         self.ymin = ymin
@@ -30,7 +28,12 @@ class ImageCropper(object):
         :param img_in:
         """
         self.img_in = img_in
-        self.bbox = BoundingBox()
+        # the default is to keep the whole image
+        self.bbox = BoundingBox(
+            0, img_in.dim[0]-1,
+            0, img_in.dim[1]-1,
+            0, img_in.dim[2]-1,
+        )
 
     def crop(self, background=None):
         """
--- a/spinalcordtoolbox/cropping.py
+++ b/spinalcordtoolbox/cropping.py
@@ -106,7 +106,7 @@ class ImageCropper(object):
         if not len(img_ref.data.shape) == len(self.img_in.data.shape):
             logger.error("Inconsistent dimensions: n_dim(img_ref)={}; n_dim(img_in)={}"
                          .format(len(img_ref.data.shape), len(self.img_in.data.shape)))
-            raise Exception(ValueError)
+            raise ValueError('Inconsistent dimensions')
         # Fill reference data with ones
         img_ref.data[:] = 1
         # Resample new image (in reference coordinates) into input image
--- a/spinalcordtoolbox/cropping.py
+++ b/spinalcordtoolbox/cropping.py
@@ -23,7 +23,7 @@ class BoundingBox(object):
         self.zmin = zmin
         self.zmax = zmax
 
-    def get_minmax(self, img=None):
+    def get_minmax(self, img):
         """
         Get voxel-based bounding box from coordinates. Replaces '-1' with max dim along each axis, '-2' with max dim
         minus 1, etc.
@@ -31,34 +31,26 @@ class BoundingBox(object):
         :param img: Image object to get dimensions
         :return:
         """
-        def _get_min_value(input):
-            if input is None:
-                return 0
-            else:
-                return input
+        def _get_min_value(val):
+            if val is None:
+                val = 0
+            return val
 
-        def _get_max_value(input, dim):
-            # If empty, return maximum dimension (i.e. no change)
-            if input is None:
-                return dim
-            # If input is "-1", return maximum dimension (i.e. no change). If input is "-2", returns maximum
-            # dimension minus one, etc.
-            elif np.sign(input) == -1:
-                return input + dim
-            # If user specified a non-negative value, use that
-            else:
-                return input
+        def _get_max_value(val, dim):
+            if val is None:
+                val = dim
+            elif val < 0:
+                val += dim
+            return val
 
-        xyz_to_num = {'x': 0, 'y': 1, 'z': 2}
-        bbox_voxel = BoundingBox()
-        for attr, value in self.__dict__.items():
-            if attr[-3:] == 'min':
-                bbox_voxel.__setattr__(attr, _get_min_value(self.__getattribute__(attr)))
-            elif attr[-3:] == 'max':
-                bbox_voxel.__setattr__(attr, _get_max_value(self.__getattribute__(attr), img.dim[xyz_to_num[attr[0]]]))
-            else:
-                raise Exception(ValueError)
-        return bbox_voxel
+        return BoundingBox(
+            xmin=_get_min_value(self.xmin),
+            ymin=_get_min_value(self.ymin),
+            zmin=_get_min_value(self.zmin),
+            xmax=_get_max_value(self.xmax, img.dim[0]),
+            ymax=_get_max_value(self.ymax, img.dim[1]),
+            zmax=_get_max_value(self.zmax, img.dim[2]),
+        )
 
 
 class ImageCropper(object):
--- a/spinalcordtoolbox/cropping.py
+++ b/spinalcordtoolbox/cropping.py
@@ -57,11 +57,13 @@ class ImageCropper(object):
             # set a more permissive threshold for reading the qform
             # (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3703 for details)
             img_out.hdr.quaternion_threshold = -1e-6
-            # adapt the origin in the sform and qform matrix
+            # adapt the origin in the qform matrix
             new_origin = np.dot(img_out.hdr.get_qform(), [bbox.xmin, bbox.ymin, bbox.zmin, 1])
             img_out.hdr.structarr['qoffset_x'] = new_origin[0]
             img_out.hdr.structarr['qoffset_y'] = new_origin[1]
             img_out.hdr.structarr['qoffset_z'] = new_origin[2]
+            # adapt the origin in the sform matrix
+            new_origin = np.dot(img_out.hdr.get_sform(), [bbox.xmin, bbox.ymin, bbox.zmin, 1])
             img_out.hdr.structarr['srow_x'][-1] = new_origin[0]
             img_out.hdr.structarr['srow_y'][-1] = new_origin[1]
             img_out.hdr.structarr['srow_z'][-1] = new_origin[2]
--- a/spinalcordtoolbox/cropping.py
+++ b/spinalcordtoolbox/cropping.py
@@ -54,6 +54,9 @@ class ImageCropper(object):
             data_crop = self.img_in.data[bbox.xmin:bbox.xmax+1, bbox.ymin:bbox.ymax+1, bbox.zmin:bbox.zmax+1]
             img_out = Image(param=data_crop, hdr=self.img_in.hdr)
 
+            # set a more permissive threshold for reading the qform
+            # (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3703 for details)
+            img_out.hdr.quaternion_threshold = -1e-6
             # adapt the origin in the sform and qform matrix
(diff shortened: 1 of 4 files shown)


testing/cli/test_cli_sct_crop_image.py
```diff
@@ -13,6 +13,7 @@
     ('t2/t2.nii.gz', 't2_crop_xyz.nii', ['-xmin', '1', '-xmax', '-3', '-ymin', '2', '-ymax', '10'], (57, 9, 52)),
     ('t2/t2.nii.gz', 't2_crop_mask.nii', ['-m', 't2/t2_seg-manual.nii.gz'], (11, 55, 13)),
     ('t2/t2.nii.gz', 't2_crop_ref.nii', ['-ref', 'mt/mt0.nii.gz'], (37, 55, 34)),
+    ('qform/bad_quaternion.nii.gz', 'bad_quaternion_crop.nii', [], (2, 3, 4)),
```
joshuacwnewton, line 16: I feel a bit hesitant about adding this image to `sct_testing_data`:

- It's a little opaque to have to read the [`README.md`](https://github.com/spinalcordtoolbox/sct_testing_data/pull/10/files#diff-a3ac5c739d67ad8ef086f9ebcec1ebbbac57dd64a47823561e969a3985f2e760) to find out what `bad_quaternion.nii.gz` is actually doing.
- A new image might be overkill considering the only condition we're testing is a slightly off quarternion.
- `sct_testing_data` already feels quite bloated with niche, corner-case images (not the end of the world, but...)

So, my thinking here is that it might be better to take the existing file `t2.nii.gz` and tweak the quarternion on the fly, then save that modified image to a temporary directory. (That way, we can see exactly what it is we're testing inside the test itself, and we don't need to add a new image to `sct_testing_data`.)

Something like:

```python
@pytest.fixture()
def bad_quaternion(input_file, tmp_path):
    """This function modifies the quaternion of the image so that the 'quatern' values (b, c, d)
    no longer add to 1."""
	im_in = Image(input_file)
    im_in.header['quatern_d'] += 0.00097656
    fname_out = tmp_path / 'bad_quaternion.nii.gz'
    im_in.save(fname_out)
    return fname_out
```

However, this runs into a completely separate issue: pytest [doesn't allow you to use fixtures in `pytest.mark.parametrize`'d tests](https://github.com/pytest-dev/pytest/issues/349). 

This is... the one glaring fault I have with Pytest, and it's a bit awkward to work around, but... maybe we could just... create a separate test for this, instead of re-using the parametrized test?

joshuacwnewton (commented):
I just had a quick initial comment...

I plan to do a more in-depth review in a little while, but I figured this would be a good starting point (since it applies to https://github.com/spinalcordtoolbox/sct_testing_data/pull/10 as well).

testing/cli/test_cli_sct_crop_image.py
mguaypaq: Thanks for your thoughts. Here are some of mine, with no particular conclusion yet.
* I'd be happy to not add this image to `sct_testing_data`, if only because it seems like a heavy process to create a new release of another repository to add a test case in this repository.
* Relatedly, I was a bit surprised to see that we don't have anywhere in this repository to have small test data files? Could we consider adding a `testing/data/` directory beside `testing/api/` and `testing/cli/`?
* Alternatively, we could take the stance that all of our testing data should be programmatically generated. But then silent bugs in the generation code would mean that we're not really testing what we _think_ we're testing. For example, what if the image saving code automatically normalizes the quaternions? Or what if we're trying to test behaviour on ill-formed data files?
* (A nitpick: I know the example `bad_quaternion` fixture is only for illustrative purposes, but I'll note that it's the sum of _squares_ of quaternions that we're concerned with, and that typically the sum is _at most_ 1, so it would be more robust to overwrite all three quaternions with known values.)
* Regarding the corner-case-ness of `sct_testing_data`, I would have assumed that that was the point, especially since we also have `sct_example_data` for more realistic or typical data. In fact I would argue that corner cases provide a lot of value for automated testing, since they guard against regressions.

joshuacwnewton: Thank you for your thoughts!! I very much agree with all of what you said. 

To be completely honest, testing data is... something that I have a lot of half-formed thoughts about? I agree that it's not exactly clear when SCT devs should use `sct_example_data`, `sct_testing_data`, `sct_tutorial_data`, a programmatically-generated dummy file, or an in-repo file. 

This has been the source of a few different ideas and issues (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3471, https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3136, and tangentially related https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3508), but, I haven't really sat down to settle on a clear set of recommendations. 

The closest I got was the ["Testing: Datasets"](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing%3A-Datasets) wiki page, but even that is... not clear enough in its recommendations, I think.

joshuacwnewton: In the short-term, could we use a programmatically generated "bad quaternion" file, but then following the 5.6 release, sit down to properly outline how we want to approach testing data when working together?

joshuacwnewton (changes requested):
I really like a lot of the clean-up and refactoring you've done! The majority of these changes are sensible, and I like the simplification of `get_bbox_from_minmax`'s usage a lot.

I think I may have identified some backwards-incompatible changes, though?

spinalcordtoolbox/cropping.py
```diff
@@ -13,9 +13,7 @@
 
 
 class BoundingBox(object):
-    """
-    """
-    def __init__(self, xmin=None, xmax=None, ymin=None, ymax=None, zmin=None, zmax=None):
+    def __init__(self, xmin, xmax, ymin, ymax, zmin, zmax):
```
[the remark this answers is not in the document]
joshuacwnewton: <details>
<summary> Aside </summary>

I felt that this was a little difficult + time-intensive to review, I think because a few different changes were spread across these two commits, and I had to jump back and forth between the commits to fully understand what had changed.

</details>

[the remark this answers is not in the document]
joshuacwnewton: > Case in point: I should have mentioned that the default values for `arguments.xmin`, `arguments.xmax`, etc. are not `None`, but rather 0 and -1, see [the argument declarations here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2df887bfdeee82039964aaa2ae9b4f739a5a5c31/spinalcordtoolbox/scripts/sct_crop_image.py#L71-L85).

Oh!! Of course! Thank you for pointing this out for me. I indeed completely missed that. 😅 

> I did wonder about that, and reasoned that this would match the behaviour at the command line when not passing any arguments, but I agree that raising an explicit error is better. I'll change it.

Now that I understand the default behaviour when not passing any arguments, your original reasoning makes more sense to me! 

But, that makes me wonder... Should we even consider "calling `sct_crop_image` with no arguments" valid usage? 🤔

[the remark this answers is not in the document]
joshuacwnewton: > Sorry for the difficult review. I'd be happy to try out different workflows and find out what's easiest (and part of that is getting feedback about when it's hard, so thanks!)
> 
> For example, it may be easier if I separate the refactoring and the bug-fixing into different PRs (as I did in #3774). I naturally do both at the same time in my process for understanding the code, but I can still draft separate PRs.
> 
> I do think some of the tangled-ness comes from the original code, though, so probably refactoring PRs will be somewhat harder to review in general. I'll try to explicitly state my findings about the code (like invariants, assumptions, etc.), which I didn't do enough here.

Oh gosh, no worries whatsoever. Like you said, a lot of the difficulty was inherent to the... messiness of the existing codebase, hehe. 

In past months, I've tended to just... work around the mess, and only make very surgical, light changes? So, having someone take the time to actually untangle it all is simultaneously a big boon and also a bit of a new challenge, review-wise. But, I think it'll be worth it in the end. :)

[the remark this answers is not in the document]
mguaypaq: I think "calling `sct_crop_image` with no cropping-related arguments"  should be fine, since there are documented default values for each of `xmin`, `xmax`, etc. Also, it does emit a log message with the bounding box it's using, so it wouldn't be a hard-to-debug silent problem.

Personally, I like having sensible defaults in my tools. A tool that throws an error when I ask it for a no-op means that it's harder to use in a pipeline or a script.

mguaypaq:
[This commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3771/commits/2df887bfdeee82039964aaa2ae9b4f739a5a5c31) introduces a change in the computation of the s-form, and re-introduces the flakiness from #3597 into the batch processing test. This is because the floating-point values excluded are compared exactly, rather than with a tolerance:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8148af6bcc0497f0f84a080cb6ba7f3791620b47/testing/batch_processing/test_batch_processing.py#L49-L50

I'll add a commit to address this.

testing/batch_processing/test_batch_processing.py
```diff
@@ -46,7 +46,9 @@ def test_batch_processing_results(csv_filepath, row, column):
     # These values should be considered a failure, but because the cause is known, skipping the
     # test saves us the intermittent interruptions on PRs until we can fix the issue.
     # TODO: Remove this skip once we can implement a long-term workaround for the issue
-    if actual_value in [0.780605541771859, 0.77461183127549, 54.42425823974834]:
+    if (actual_value == pytest.approx(0.780605541771859) or
+            actual_value == pytest.approx(0.77461183127549) or
+            actual_value == pytest.approx(54.42425823974834)):
```
joshuacwnewton, line 51: > [This commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3771/commits/14eb91cbba05bde51a91bbd41e5d5ed077340b6e)  introduces a change in the computation of the s-form, and re-introduces the flakiness from #3597 into the batch processing test.

If I'm inferring correctly from this, it sounds like the tested failed once here due to the Platinum CPU. Or... did it? Because of the rebase, I can't see the failing CI...

I ask because I'm not sure I understand how the sform change re-introduced flakiness into the batch processing test. I'd love to see an example of how the test failed, and what tolerance the values differed by.

joshuacwnewton (approved):
Everything else (besides that one comment about `pytest.approx`) looks good to me, so I'm pre-approving.

testing/batch_processing/test_batch_processing.py
mguaypaq: I could have sworn I saw a failure at the "Validate new results" stage which showed numbers very close to the Platinum CPU ones (which, incidentally, are also the numbers my own laptop produces, although it has a different CPU)... in fact that test run passed when I triggered a re-run, whereas the failure you linked would have failed again.

Ah! But I did a force-push to this branch, and I think all the test runs from those previous commits are gone as a result.

Anyway, since my CPU produces very similar results to the Platinum CPU, I can confirm that this PR changes the numbers a little bit (but still within the floating point comparison threshold). Does that address your concern?

Issue #3673, opened 2022-01-31
Watch out for 'negative' discs


jcohenadad:
### Description

It seems like if C2-C3 is wrongly detected and has a bunch of discs above, then the function will iterate and eventually go through negative discs, which seems to have caused the [issue reported by a user](https://forum.spinalcordmri.org/t/inconsistency-in-spinalcord-segmentation-leads-to-error-in-in-sct-label-vertebrae-version-5-3-0/809/2). 

### Steps to Reproduce

It would be nice to be able to reproduce this issue via our testing data. Eg: manually initialize a C2-C3 disc relatively 'low' in the c-spine, and then run the algorithm that iterates across discs.

### Possible fix

If the cause of the bug was properly identified, implement some 'safety checks' so the user knows what's going on (in this case: problem with C2-C3 detection).

mguaypaq:
After some more investigation, the cause is not as expected. In fact, the specific error message about `UnboundLocalError: local variable 'approx_distance_to_next_disc' referenced before assignment` can only happen in the **first** iteration.

The source is closer to this part of the log:
```
Get z and disc values from straight label...
Loaded labelz_straight.nii.gz (/tmp/sct-20220128144857.530973-label_vertebrae-t2yf2k9o/labelz_straight.nii.gz) orientation RPI shape (115, 141, 624)
.. [530, 0]
```

The label should be `[530, 3]`, not `[530, 0]`, because we see `C2-C3 detected...` in the log, so the saved label is created by this code:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/vertebrae/detect_c2c3.py#L127

And indeed, using a label of `[..., 0]` would cause the `UnboundLocalError`.

So, probably because of straightening, or because of this code which loads the label, there's a problem with the label:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/vertebrae/core.py#L286-L299

Without the original image and segmentation, this may be hard to investigate.

joshuacwnewton:
[...]

PR #3775, opened 2022-04-20
Output a better error message when the initial disc is invalid

mguaypaq:
## Description
When calling `vertebral_detection` with an invalid initial disc (that is, a disc label that's not in the PAM50 template), the result is a very unfriendly error message:

```python
  File "/icometrix-install/spinalcordtoolbox-5.3.0/spinalcordtoolbox/vertebrae/core.py", line 206, in vertebral_detection
    approx_distance_to_next_disc = list_distance[list_disc_value_template.index(current_disc - 1)]
ValueError: -2 is not in list
During handling of the above exception, another exception occurred:
  File "/icometrix-install/spinalcordtoolbox-5.3.0/spinalcordtoolbox/scripts/sct_label_vertebrae.py", line 482, in <module>
    main(sys.argv[1:])
  File "/icometrix-install/spinalcordtoolbox-5.3.0/spinalcordtoolbox/scripts/sct_label_vertebrae.py", line 419, in main
    verbose=verbose, path_template=path_template, path_output=path_output, scale_dist=scale_dist)
  File "/icometrix-install/spinalcordtoolbox-5.3.0/spinalcordtoolbox/vertebrae/core.py", line 208, in vertebral_detection
    logger.warning('Disc value not included in template. Using previously-calculated distance: %s', approx_distance_to_next_disc)
UnboundLocalError: local variable 'approx_distance_to_next_disc' referenced before assignment
```

(This can only happen in the first iteration through the loop in `vertebral_detection()`. In further iterations, the error is caught, a warning is displayed, and the previous iteration's value of `approx_distance_to_next_disc` is used.)

This PR adds a check to detect this condition up-front and output a (hopefully) friendlier message to the user.

This is not a complete fix for the linked issue (#3673, see also [the original forum post](https://forum.spinalcordmri.org/t/inconsistency-in-spinalcord-segmentation-leads-to-error-in-in-sct-label-vertebrae-version-5-3-0/809)), but is likely the best we can do in time for the 5.6 release.

## Linked issues
Part of #3673.

### DIFF
--- a/spinalcordtoolbox/vertebrae/core.py
+++ b/spinalcordtoolbox/vertebrae/core.py
@@ -141,8 +141,8 @@ def vertebral_detection(fname, fname_seg, contrast, param, init_disc, verbose=1,
         logger.info('Current disc: %s (z=%s). Direction: %s', current_value, current_z, direction)
         try:
             # get z corresponding to current disc on template
-            current_z_template = list_disc_z_template[current_value]
-        except IndexError:
+            current_z_template = list_disc_z_template[list_disc_value_template.index(current_value)]
+        except (ValueError, IndexError):
             # in case reached the bottom (see issue #849)
             logger.warning('Reached the bottom of the template. Stop searching.')
             break
--- a/spinalcordtoolbox/vertebrae/core.py
+++ b/spinalcordtoolbox/vertebrae/core.py
@@ -86,7 +86,7 @@ def vertebral_detection(fname, fname_seg, contrast, param, init_disc, verbose=1,
     xct = int(np.round(nxt / 2))  # direction RL
     yct = int(np.round(nyt / 2))  # direction AP
 
-    # define mean distance (in voxel) between adjacent discs: [C1/C2 -> C2/C3], [C2/C3 -> C4/C5], ..., [L1/L2 -> L2/L3]
+    # define mean distance (in voxel) between adjacent discs: [C1/C2 -> C2/C3], [C2/C3 -> C3/C4], ..., [L1/L2 -> L2/L3]
     centerline_level = data_disc_template[xct, yct, :]
     # attribute value to each disc. Starts from max level, then decrease.
     min_level = centerline_level[centerline_level.nonzero()].min()
--- a/spinalcordtoolbox/vertebrae/core.py
+++ b/spinalcordtoolbox/vertebrae/core.py
@@ -130,9 +130,7 @@ def vertebral_detection(fname, fname_seg, contrast, param, init_disc, verbose=1,
     # FIND DISCS
     # ===========================================================================
     logger.info('Detect intervertebral discs...')
-    # assign initial z and disc
-    current_z = init_disc[0]
-    current_disc = init_disc[1]
+    current_z, current_value = init_disc
     # create list for z and disc
     list_disc_z = []
     list_disc_value = []
@@ -140,39 +138,39 @@ def vertebral_detection(fname, fname_seg, contrast, param, init_disc, verbose=1,
     direction = 'superior'
     search_next_disc = True
     while search_next_disc:
-        logger.info('Current disc: %s (z=%s). Direction: %s', current_disc, current_z, direction)
+        logger.info('Current disc: %s (z=%s). Direction: %s', current_value, current_z, direction)
         try:
             # get z corresponding to current disc on template
-            current_z_template = list_disc_z_template[current_disc]
-        except:
+            current_z_template = list_disc_z_template[current_value]
+        except IndexError:
             # in case reached the bottom (see issue #849)
             logger.warning('Reached the bottom of the template. Stop searching.')
             break
         # find next disc
         # N.B. Do not search for C1/C2 disc (because poorly visible), use template distance instead
-        if current_disc != 1:
+        if current_value != 1:
             current_z = compute_corr_3d(data, data_template, x=xc, xshift=0, xsize=param['size_RL'],
                                         y=yc, yshift=param['shift_AP'], ysize=param['size_AP'],
                                         z=current_z, zshift=0, zsize=param['size_IS'],
                                         xtarget=xct, ytarget=yct, ztarget=current_z_template,
-                                        zrange=zrange, verbose=verbose, save_suffix='_disc' + str(current_disc),
+                                        zrange=zrange, verbose=verbose, save_suffix='_disc' + str(current_value),
                                         path_output=path_output)
 
         # display new disc
         if verbose == 2:
             ax_disc.scatter(yc + param['shift_AP_visu'], current_z, c='yellow', s=10)
-            ax_disc.text(yc + param['shift_AP_visu'] + 4, current_z, str(current_disc) + '/' + str(current_disc + 1),
+            ax_disc.text(yc + param['shift_AP_visu'] + 4, current_z, str(current_value) + '/' + str(current_value + 1),
                          verticalalignment='center', horizontalalignment='left', color='yellow', fontsize=7)
 
         # append to main list
         if direction == 'superior':
             # append at the beginning
             list_disc_z.insert(0, current_z)
-            list_disc_value.insert(0, current_disc)
+            list_disc_value.insert(0, current_value)
         elif direction == 'inferior':
             # append at the end
             list_disc_z.append(current_z)
-            list_disc_value.append(current_disc)
+            list_disc_value.append(current_value)
 
         # adjust correcting factor based on already-identified discs
         if len(list_disc_z) > 1:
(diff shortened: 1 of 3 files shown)


mguaypaq:
Looks like `test_sct_label_vertebrae_high_value_warning` is failing. I'll investigate.

mguaypaq:
Ok, this is ready for review again. I'm very open to the suggestion that we shouldn't merge the new test cases into master, but I wanted to include them in the PR mostly as an illustration of various corner cases (and proof that the new errors/warnings do get triggered).

I tried to plug as many holes as possible with decent error messages, but in my mind this further cements the idea that we should take a good look at the vertebral detection code after the 5.6 release.

testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -51,6 +51,45 @@ def test_sct_label_vertebrae_high_value_warning(caplog, tmp_path):
     assert 'Disc value not included in template.' in caplog.text
 
 
+def test_sct_label_vertebrae_initial_disc_no_inferior(caplog, tmp_path):
```
joshuacwnewton, line 54: I really love the in-depth descriptions you put into your commits. For example, the description in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/a442e1291c060600bed126aeef617737c0296cd6 really aided me when trying to trace the behavior of `sct_label_vertebrae`.

I'm wondering... would you be able to adapt these descriptions into docstrings for the tests? That way, we have a loose description of the script's behavior for each corner case test (i.e. -1, 0, 19, 20, 21) attached to the tests themselves.

joshuacwnewton (approved):
> I'm very open to the suggestion that we shouldn't merge the new test cases into master, but I wanted to include them in the PR mostly as an illustration of various corner cases (and proof that the new errors/warnings do get triggered).

I found these corner case tests to be _very_ valuable for reviewing. 

(What I did locally was to rebase the "adding tests" commit to be the first commit, then step through your changes to see what the effect on the tests would be before/after, making it act as a sort of TDD-esque changeset.) 

* test_sct_label_vertebrae_initial_disc_no_inferior: ValueError: 20 is not in list (Fixed by a442e12) 
* test_sct_label_vertebrae_initial_disc_zero: spinalcordtoolbox.vertebrae.core.EmptyArrayError: Center of mass can't be calculated on empty arrays. (Fixed by 16dbd6f)
* test_sct_label_vertebrae_initial_disc_too_low: ValueError: min() arg is an empty sequence (Fixed by b8e146f) 
* test_sct_label_vertebrae_initial_disc_too_high: ValueError: 21 is not in list (Fixed by b8e146f)

So, thank you for including these tests. :) 

I've included one comment, but it's just about documentation for the tests, so I'm going to pre-approve.

testing/cli/test_cli_sct_label_vertebrae.py
```diff
@@ -51,6 +51,45 @@ def test_sct_label_vertebrae_high_value_warning(caplog, tmp_path):
...
+    with pytest.raises(SystemExit) as excinfo:
+        sct_label_vertebrae.main(['-i', sct_test_path('t2', 't2.nii.gz'),
+                                  '-s', sct_test_path('t2', 't2_seg-manual.nii.gz'),
+                                  '-c', 't2', '-initz', '40,0',
+                                  '-ofolder', str(tmp_path)])
+    assert excinfo.value.code == 1
+    assert 'Missing label or zero label for initial disc.' in capsys.readouterr().out
+
+
+def test_sct_label_vertebrae_initial_disc_too_low(capsys, tmp_path):
```
joshuacwnewton, line 73: Actually, wait, sorry. It occurs to me that none of the error messages observed from the tests (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3775#pullrequestreview-950071054) match the error message in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3775#issue-1210242621.

Did we want to try and create a test that (would have) thrown the `UnboundLocalError`?

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -194,28 +195,32 @@ def vertebral_detection(fname, fname_seg, contrast, param, init_disc, verbose=1,
...
         elif direction == 'inferior':
             try:
-                approx_distance_to_next_disc = list_distance[list_disc_value_template.index(current_disc)]
+                approx_distance_to_next_disc = list_distance[list_disc_value_template.index(current_value)]
             except ValueError:
                 logger.warning('Disc value not included in template. Using previously-calculated distance: %s', approx_distance_to_next_disc)
             # assign new current_z and disc value
             current_z = current_z - approx_distance_to_next_disc
-            current_disc = current_disc + 1
+            current_value = current_value + 1
```
joshuacwnewton, line 211: Ah! So, there was a small, curious thing that caught my attention:

* The error I observed from [`test_sct_label_vertebrae_initial_disc_too_low`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3775#pullrequestreview-950071054) prior to your try/except changes was `ValueError: min() arg is an empty sequence`
* However, the user's error from [the reported traceback](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3775#issue-1210242621) was an `UnboundLocalError`

Either way, your fix _does_ address the issue (by catching a `current_value` that doesn't exist in the template). But, the test itself produced slightly different behavior _prior_ to your changes. Curious!

I tried investigating further by tracing your test and seeing how it differs from the user's report. And it turns out, when we supply `-initz 40,-1` as we do in that test, the `current_value` actually gets parsed as _254_???

image

I think this is because of this line:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L383

Which creates a `uint8` label image from the `initz` argument (i.e. the image is bound to `[0, 255]`). So, I'm guessing that by supplying -1, it wraps around and creates a labeled image with the value `254`? (I tried changing the test to -2 and it produced an image with label value `253`.)

But, as I mentioned, your check still catches the value `254` as it would `-1`, so... I'm not sure. Is this something we want to investigate while we're here? If not, this has my approval. :)

joshuacwnewton: Oh wait, I can just quickly update `create_labels_along_segmentation` to not use `uint8`, to see what would happen if we did in fact supply `-1`. One sec. :V

joshuacwnewton: So, if I change this line to use `dtype=np._int`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/labels.py#L111

We get the `-1` label in the image as expected. However, then when the image gets dilated here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L402

The dilation causes the `-1` value to disappear. So, again, another little factor preventing your `-1` test from actually doing what it's supposed to. :P

joshuacwnewton: Ah, we use [`skimage.morphology.dilation`](https://scikit-image.org/docs/dev/api/skimage.morphology.html#skimage.morphology.dilation) for our dilate function, which inherently returns a `uint8` image.

So... I guess negative label values are impossible via `fname_labelz`? But... the user's error certainly suggests that it's possible...

joshuacwnewton: Again though, just to re-emphasize, I think we've covered all of our bases with the new corner cases (0, 19, 20, 21), so fixating on the `-1` test isn't really that important?

I just wanted to highlight that the test wasn't doing what we thought it was doing. :P

mguaypaq: That's a great investigation! I didn't think about numpy's wrap-around integer types here, but it's a good source of surprises and bugs. And you're right that image dilation does surprising things with negative numbers.

But! There's one more place where negative numbers can creep in :stuck_out_tongue: After straightening of `fname_labelz`, the label is read from here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L414
and then we subtract 1 before handing to `vertebral_detection`:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/scripts/sct_label_vertebrae.py#L433
Note that by this point `init_disc[1]` is an honest-to-goodness Python int, because it's explicitly cast here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bad57807faf05643e2697c911f3403c4f57d9f66/spinalcordtoolbox/vertebrae/core.py#L298

So, I think `-1` is in fact the *only* possible negative value, and it happens with `-initz *,0` (or I guess, any positive or negative multiple of 256?). This also explains why you saw 254 instead of 255 when using `-initz *,-1`.

----

~~Now, I want to argue that the `min() arg is an empty sequence` is actually already addressed by this PR.~~ The error is raised on this line after breaking out of the main loop:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f7832ad5670fc6cb875c8b413a37c15a1e5166c/spinalcordtoolbox/vertebrae/core.py#L232
The argument to `min` can only be empty if this code in the loop never executes:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f7832ad5670fc6cb875c8b413a37c15a1e5166c/spinalcordtoolbox/vertebrae/core.py#L168-L176
And the only `break` before that is here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f7832ad5670fc6cb875c8b413a37c15a1e5166c/spinalcordtoolbox/vertebrae/core.py#L145-L151
However, ~~the new error checking code prevents this from happening on the first iteration,~~ because we now check whether the initial `current_value` is there:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f7832ad5670fc6cb875c8b413a37c15a1e5166c/spinalcordtoolbox/vertebrae/core.py#L134-L136
Edit: the check which would actually prevent this is `current_value in range(-1-len(list_disc_value_template), len(list_disc_value_template))`. But for the PAM50 template, `list_disc_value_template == [0, 1, 2, ..., 19]`, so the new error checking code accidentally catches this condition.

I think we're hitting diminishing returns here; the real solution is not to add more error checking code, but to actually fix the underlying code after the 5.6 release. So, I think I'll add some comments and/or docstrings for the new corner-case tests and call this PR done, if that's ok with you.

joshuacwnewton: >  So, I think I'll add some comments and/or docstrings for the new corner-case tests and call this PR done, if that's ok with you.

Perfectly reasonable to me! That was very much my feelings too after writing https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3775#discussion_r856302643.

joshuacwnewton (approved):

Issue #3396, opened 2021-05-19
Remove dependency on PyTorch by relying only on `onnxruntime` and ONNX models


joshuacwnewton:
Currently, SCT depends on at least some version of PyTorch (CPU or GPU, either is OK) so that we can run inference on `.pt` models trained using `ivadomed`. But, some of ivadomed's models are already available in the ONNX representation, so theoretically we could run inference without having to rely on `torch` at all.

#### Motivation

 * `torch` is _heavy_: 600MB for the CPU version, and 1.3GB for the GPU version. Removing this dependency makes SCT much lighter.
 * Not depending on `torch` makes a `pip` install much more viable for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1526. We wouldn't have to expose any CPU/GPU installation workarounds to the user, making for a more seamless installation procedure.

#### Blockers

- [ ] Some, but not all models are in the ONNX format. See, for example, this issue: https://github.com/ivadomed/ivadomed/issues/749
- [ ] `spinalcordtoolbox` currently lists `ivadomed` as a requirement, which in turn depends on `torch`. So, even if SCT removes `torch` from its requirements, we'll still be tied to it via `ivadomed`. Notably, though, we only use `ivadomed` in a single place for a single purpose: inference in `sct_deepseg`.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/9a4b45715240d6c08b8d71587ffbe99a41326cc6/spinalcordtoolbox/scripts/sct_deepseg.py#L18

jcohenadad:
> theoretically we could run inference without having to rely on torch at all.

That would be amazing. But, what library would be used then, if not pytorch?

joshuacwnewton:
> That would be amazing. But, what library would be used then, if not pytorch?

Just to clarify, PyTorch would still be used within `ivadomed` to train/evaluate the models. But, using the ONNX representations, we would simply use `onnxruntime` to perform inference.

This is already done within `ivadomed` -- we would basically just extract the ONNX inference function from ivadomed, and leave the rest of the project out of SCT.

kousu:
pytorch or tensorflow, especially when used in GPU mode, are like *compilers*. Listing them in our dependencies is like packaging your entire build toolchain along with your software: you can do it, but 99% of the time it's a lot of dead weight.

[...]

PR #3790, opened 2022-05-09
Use CPU `--extra-index-url` to remove strict dependency pinning for `torch`

kousu:
#### GitHub

#### PR contents

## Description

At some point recently torch put up a full pip repo (`pip --extra-index-url https://download.pytorch.org/whl/cpu`) instead of just a list of files (`pip --find-links https://download.pytorch.org/whl/torch_stable.html`), meaning the officially stated version of torch (1.8.0, 1.9.0, 1.10, ...) is now orthogonal from the build variant (`cpu`, `cu101`, `cu113`, ...), so we can unpin the version and leave the choice of version encapsulated in ivadomed, especially because `torch` is not something we use directly at all, it's just something `ivadomed` calls.

The version has more to do with what APIs we can call, and the build variant has more to do with what platform the end user is actually installing on, so they are a lot better separated.

## Linked issues

Sidesteps https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3396

Also helps https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1526; this isn't 100% clean but it will mean our install instructions to users when we do that will be:

```
pip install spinalcordtoolbox
```

or:

```
pip install --extra-index-url https://download.pytorch.org/whl/cpu spinalcordtoolbox
```

which is pretty good, until https://github.com/pytorch/pytorch/issues/26340 is finally resolved. If it's ever resolved.

I've given the same treatment to ivadomed directly at https://github.com/ivadomed/ivadomed/pull/1129.

See discussions at 

* https://github.com/ivadomed/ivadomed/pull/1000
* https://docs.google.com/document/d/19mEHKOZKFdprb3UPQYxmfO46mn0bPQfOgE-WPdeF7AQ/edit#heading=h.cxdfvr56zc7g

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,8 @@ Keras==2.3.1
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu
-ivadomed@git+https://github.com/ivadomed/ivadomed@ng/cu113
+ivadomed@git+https://github.com/ivadomed/ivadomed
+#ivadomed>=2.9.6
 matplotlib
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
 # Installing "Microsoft Visual C++ Redistributable for Visual Studio" will fix this too, but
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,8 +6,7 @@ Keras==2.3.1
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu
-ivadomed@git+https://github.com/ivadomed/ivadomed
-#ivadomed>=2.9.6
+ivadomed~=2.9.6
 matplotlib
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
 # Installing "Microsoft Visual C++ Redistributable for Visual Studio" will fix this too, but
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ Keras==2.3.1
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu
-ivadomed
+ivadomed@git+https://github.com/ivadomed/ivadomed@ng/cu113
 matplotlib
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
 # Installing "Microsoft Visual C++ Redistributable for Visual Studio" will fix this too, but
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,9 @@ dipy
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.3.1
+# PyTorch's Linux distribution is very large due to its GPU support,
+# but we only need that for training models. Our users only need CPU.
+--extra-index-url https://download.pytorch.org/whl/cpu
 ivadomed
 matplotlib
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
@@ -34,16 +37,6 @@ scipy
 scikit-image
 scikit-learn
 tensorflow~=1.15.0
-# PyTorch's Linux/Windows distribution is very large due to its GPU support,
-# but we only need that for training models. For users, use the CPU-only version
-# (only available directly from the PyTorch project).
-# The macOS version has never had GPU support, so doesn't need the workaround.
--f https://download.pytorch.org/whl/torch/
--f https://download.pytorch.org/whl/torchvision/
-torch==1.5.0+cpu; sys_platform != "darwin"
-torch==1.5.0; sys_platform == "darwin"
-torchvision==0.6.0+cpu; sys_platform != "darwin"
-torchvision==0.6.0; sys_platform == "darwin"
 xlwt
 tqdm
 transforms3d


kousu:
If I run:

```
p115628@joplin:~/spinalcordtoolbox$ ./install_sct 
[...]
Using requirements.txt (git installation)

Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/cpu
[...]
Collecting torch>=1.1
  Downloading https://download.pytorch.org/whl/cpu/torch-1.11.0%2Bcpu-cp37-cp37m-linux_x86_64.whl (169.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 169.1/169.1 MB 8.3 MB/s eta 0:00:00
[...]
Successfully installed Deprecated-1.2.13 Keras-2.3.1 PyQt5_sip-4.19.19 PyWavelets-1.3.0 SimpleITK-2.1.1.2 absl-py-1.0.0 astor-0.8.1 attrs-21.4.0 bids-validator-1.9.3 certifi-2021.10.8 cffi-1.15.0 charset-normalizer-2.0.12 click-8.1.3 coverage-6.3.2 cryptography-37.0.2 csv-diff-1.1 cycler-0.11.0 dictdiffer-0.9.0 dipy-1.5.0 docopt-0.6.2 fonttools-4.33.3 formulaic-0.3.4 gast-0.2.2 google-pasta-0.2.0 grpcio-1.46.0 h5py-2.10.0 humanize-4.1.0 idna-3.3 imageio-2.19.1 importlib-metadata-4.11.3 iniconfig-1.1.1 interface-meta-1.3.0 ivadomed-2.9.5 joblib-1.1.0 keras-applications-1.0.8 keras-preprocessing-1.1.2 kiwisolver-1.4.2 loguru-0.6.0 markdown-3.3.7 matplotlib-3.5.2 networkx-2.6.3 nibabel-3.2.2 num2words-0.5.10 numpy-1.18.5 onnxruntime-1.7.0 opt-einsum-3.3.0 packaging-21.3 pandas-1.3.5 pillow-9.1.0 pluggy-1.0.0 portalocker-2.4.0 protobuf-3.20.1 psutil-5.9.0 py-1.11.0 pyOpenSSL-22.0.0 pybids-0.15.1 pycparser-2.21 pyparsing-2.4.7 pyqt5-5.11.3 pytest-7.1.2 pytest-cov-3.0.0 pytest_console_scripts-1.3.1 python-dateutil-2.8.2 pytz-2022.1 pyyaml-6.0 raven-6.10.0 requests-2.27.1 requirements-parser-0.5.0 scikit-image-0.19.2 scikit-learn-1.0.2 scipy-1.7.3 seaborn-0.11.2 setuptools-62.1.0 six-1.16.0 sqlalchemy-1.3.24 tensorboard-1.15.0 tensorflow-1.15.5 tensorflow-estimator-1.15.1 termcolor-1.1.0 threadpoolctl-3.1.0 tifffile-2021.11.2 tomli-2.0.1 torch-1.11.0+cpu torchio-0.18.76 tqdm-4.64.0 transforms3d-0.3.1 types-setuptools-57.4.14 typing-extensions-4.2.0 urllib3-1.26.9 werkzeug-2.1.2 wheel-0.37.1 wquantiles-0.6 wrapt-1.14.1 xlwt-1.3.0 zipp-3.8.0
```

Notice it got "torch-1.11.0+cpu" in the end and the download was only 169MB.

kousu:
Without looking deeper yet, the test failures are probably because this did

> Collecting torch>=1.1

but ivadomed wants 1.8.0:

https://github.com/ivadomed/ivadomed/blob/bd904e5e139fa6a437abfc225216f6057824f1e3/requirements.txt#L12

(it looks like their setup.py is broken on this point? maybe? and we were hiding the bug by pinning torch ourselves?)

kousu:
Indeed, it was because ivadomed's package had forgotten to depend on either `torch` or `torchvision`: https://github.com/ivadomed/ivadomed/issues/1130, and that was percolating up into our tests.

I've got a fix in https://github.com/ivadomed/ivadomed/pull/1129, so if that gets published then this can too.

kousu:
Until https://github.com/ivadomed/ivadomed/pull/1130 is fixed, you should be able to test this by temporarily pointing SCT at [my patch](https://github.com/ivadomed/ivadomed/pull/1129):

```diff
diff --git a/requirements.txt b/requirements.txt
index b91216c5..f958ef8a 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ dipy
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.3.1
-ivadomed
+ivadomed@git+https://github.com/ivadomed/ivadomed@ng/cu113
 matplotlib
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
 # Installing "Microsoft Visual C++ Redistributable for Visual Studio" will fix this too, but
```

This will install the fixed ivadomed.

kousu:
CI is [green](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/2341278644):

Screenshot 2022-05-17 at 17-20-55 Loosen pytorch pinning  by kousu · Pull Request #3790 · spinalcordtoolbox_spinalcordtoolbox

And it [installs the right torch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/6478353169?check_suite_focus=true#step:3:443):

```
 Collecting torch==1.8.1
  Downloading https://download.pytorch.org/whl/cpu/torch-1.8.1%2Bcpu-cp37-cp37m-linux_x86_64.whl (169.1 MB)
Collecting torchvision==0.9.1
  Downloading https://download.pytorch.org/whl/cpu/torchvision-0.9.1%2Bcpu-cp37-cp37m-linux_x86_64.whl (13.3 MB)
```

It's sort of surprising that we could have been downgrading torch all the way to 1.5.0 without breaking anything. I guess we really don't use much of ivadomed at all?

requirements.txt
```diff
@@ -3,7 +3,10 @@ dipy
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.3.1
-ivadomed
+# PyTorch's Linux distribution is very large due to its GPU support,
+# but we only need that for training models. Our users only need CPU.
+--extra-index-url https://download.pytorch.org/whl/cpu
+ivadomed@git+https://github.com/ivadomed/ivadomed@ng/cu113
```
kousu, line 9: Remember to revert back to

```suggestion
ivadomed
```

once ivadomed publishes this fix.

kousu:
@kanishk16 discovered something: there's even more indecies we could use to pin specific torch/cuda/non-cuda versions like

* `--extra-index-url https://download.pytorch.org/whl/lts/1.8/cpu`
* `--extra-index-url https://download.pytorch.org/whl/lts/1.8/cu111`

I think probably sticking with `/whl/cpu` is enough for us, but if we did need to pin torch again for some reason it's good to know we could.

joshuacwnewton (approved):
I'm going to give this my approval, just because I've been heavily involved with the discussion over on the ivadomed side of things, and understand the implications of this change.

But, I'm also going to mark this as draft, just so that it's clear that we're waiting on upstream `ivadomed` change before we can merge this.

requirements.txt
kousu: Release is out, I've put this back how it should be.

I've added a minimum version now, to make sure anyone upgrading overtop is sure to get the working version of ivadomed. Not that that could possibly be an issue until #1526. But it's good to follow the conventions.

kousu:
Okay [well](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/6716594655?check_suite_focus=true), PyPI is having a hard day:

Screen Shot 2022-06-02 at 5 40 46 PM

But I'm confident once that's resolved this will pass CI.

joshuacwnewton:
Yep, this is a PyPI thing: https://status.python.org/incidents/bt0tzq7x2fk2?u=szv0ry9nhcbg

requirements.txt
```diff
@@ -3,7 +3,10 @@ dipy
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.3.1
-ivadomed
+# PyTorch's Linux distribution is very large due to its GPU support,
+# but we only need that for training models. Our users only need CPU.
+--extra-index-url https://download.pytorch.org/whl/cpu
+ivadomed~=2.9.6
```
joshuacwnewton, line 9: Previously, `ivadomed` was unpinned, so we would have gotten e.g. `2.10.0` when the `ivadomed` team decided to bump the minor version. But, with `~=`, we're locked to `2.9.X`. 

Was there any specific motivation for going with `~=` as opposed to `>=`?

requirements.txt
```diff
@@ -3,7 +3,10 @@ dipy
 # https://github.com/tensorflow/tensorflow/issues/44467
 h5py~=2.10.0
 Keras==2.3.1
-ivadomed
+# PyTorch's Linux distribution is very large due to its GPU support,
+# but we only need that for training models. Our users only need CPU.
+--extra-index-url https://download.pytorch.org/whl/cpu
```
joshuacwnewton, line 8: Up until now, I've been inferring the behavior of `--extra-index-url` based on what I've seen in the demos.

> Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/cpu

But, I'm realizing I don't actually understand *why* the `--extra-index-url` is prioritized over [`--index-url`](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-0) (i.e. https://pypi.org/simple)?

In other words, why exactly is it that the CPU version of `torch` get installed, rather than the GPU version?

joshuacwnewton: Hmm. Reading https://github.com/pypa/pip/issues/8606, there's a deep discussion there, with the tl;dr being that pip treats the index URLs [with equal priority](https://github.com/pypa/pip/issues/8606#issuecomment-665554122). But, that doesn't exactly answer my question...

The closest thing I've come to a description of what we're trying to do is: https://github.com/pypa/pip/issues/8606#issuecomment-965659563

> As anecdotal usage feedback, I see requirements.txt files with `--extra-index-url` used as a prefix to specific packages as if it were a per-requirement option; of course, this isn't actually how `--extra-index-url` works, but the invocation succeeds and generally does what the user expects (apart from edge cases where everything explodes in exciting nasal-daemon sorts of ways).

joshuacwnewton: But again, I don't really get how pypi decides, when all that `ivadomed` specifies is `torch~=1.8.1`... 

* PyPI: [torch-1.8.1-cp37-cp37m-manylinux1_x86_64.whl](https://files.pythonhosted.org/packages/56/74/6fc9dee50f7c93d6b7d9644554bdc9692f3023fa5d1de779666e6bf8ae76/torch-1.8.1-cp37-cp37m-manylinux1_x86_64.whl#sha256=95b7bbbacc3f28fe438f418392ceeae146a01adc03b29d44917d55214ac234c9)
* torch/cpu: [torch-1.8.1+cpu-cp37-cp37m-linux_x86_64.whl](https://download.pytorch.org/whl/cpu/torch-1.8.1%2Bcpu-cp37-cp37m-linux_x86_64.whl)

Yet, the [latter gets installed](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/6717573228?check_suite_focus=true#step:3:454):

```
Collecting torch~=1.8.1
  Downloading https://download.pytorch.org/whl/cpu/torch-1.8.1%2Bcpu-cp37-cp37m-linux_x86_64.whl (169.1 MB)
```

joshuacwnewton (changes requested):
Sorry to delay this a little more. I had a realization that I don't understand this as well as I thought I did.

joshuacwnewton:
@kousu Oh noooo, I wasn't quick enough :(

requirements.txt
joshuacwnewton: From https://stackoverflow.com/a/67358322:

> Pip does not really prioritize one index over the other in theory. In practice, because of a coincidence in the way things are implemented in code, it might be that one is always checked first, but it is not a behavior you should rely on.

joshuacwnewton: Is it because we're not explicitly specifying `--index-url` in `requirements.txt`? Hmm...

kousu: That''''''s a really good point. I read further down the thread and found some answers, from [ComputeCanada of all people](https://github.com/ComputeCanada/software-stack/blob/main/pip-which-version.md#build-tags-and-local-versions):

> If such a tag is found, it is used as a tie breaker between two otherwise identical packages.

So the existence of `+cpu` is what makes that wheel win.

kousu: ComputeCanada is in the process of adopting torch's solution:

> As of July 2021, we have yet to repackage all of our existing python wheels to add local version specifiers, since this will be a long endeavour, but the newly compiled ones automatically include it. We are therefore still relying on a pip<21 constraint, but we should be able to lift that constraint in the future.

Of course the better solution would be if like, torch was split up in such a way that you could just say `torch[cpu]` or `torch[gpu]` directly.

joshuacwnewton: To get even closer to the source on this, the ComputeCanada docs say:

> These options were pointed to us in https://github.com/pypa/pip/issues/10156, by [Pradyun Gedam](https://github.com/pradyunsg).

https://github.com/pypa/pip/issues/10156#issuecomment-878571384

> Another option is wheel build tags -- which might be an even better answer. If a wheel has a build tag, that wheel is preferred over a wheel that doesn't.

joshuacwnewton: Aha, even closer to the source, https://github.com/pypa/pip/issues/10156#issuecomment-878606128:

> None the less, I think the local version will be prefered over build tags. The order of preference is specified at: https://github.com/pypa/pip/blob/main/src/pip/_internal/index/package_finder.py#L530

requirements.txt
kousu: Because it's supposed to be [the safest](https://peps.python.org/pep-0440/#compatible-release)

> It matches any candidate version that is expected to be compatible with the specified version.

Is that actually true? It depends on how closely ivadomed follows semver.

But in any case it should balance between having to manually update pinnings every week and having surprise breakage in the future.

There's also the future to think about: if we just depend on `ivadomed` then someone trying to use SCT 5.4 in 3 years will get an ivadomed that's 3 years newer, with its APIs all shifted around probably.

requirements.txt
kousu: https://peps.python.org/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering is probably the closest thing to   the final word on priority.

requirements.txt
kanishk16: > There's also the future to think about: if we just depend on `ivadomed` then someone trying to use SCT 5.4 in 3 years will get an ivadomed that's 3 years newer, with its APIs all shifted around probably.

Just thinking out loud, there should be a strong reason to use an older version of SCT, as a matter of fact for any library unless it's LTS

joshuacwnewton: > There's also the future to think about: if we just depend on `ivadomed` then someone trying to use SCT 5.4 in 3 years will get an ivadomed that's 3 years newer, with its APIs all shifted around probably.

This isn't quite true, as SCT's stable releases use frozen dependencies. When 5.4 was created, for that specific version, `ivadomed` was pinned to:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/55fbd43cfee489486ea91f502b0117d7063732a0/requirements-freeze.txt#L34

To be clear: The SCT project maintains two sets of requirements, that are roughly equivalent to [`install_requires` vs. `requirements.txt`](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/)

One set is very loosely pinned:

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

The other is precisely pinned:

> Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving [repeatable installations](https://pip.pypa.io/en/latest/user_guide/#repeatability) of a complete environment.

So, my thought process for pinning dependencies has always been to try to leave them unpinned on `master` (in order to regularly test upstream dependencies), while also freezing them in place for stable releases.

joshuacwnewton: In this case, since our default requirements file is being fed into `install_requires`, I would want to follow PyPA's recommendations:

> Additionally, it’s best practice to indicate any known lower or upper bounds. 
>
> For example, it may be known, that your project requires at least v1 of ‘A’, and v2 of ‘B’, so it would be like so:
> 
> ```
> install_requires=[
>    'A>=1',
>    'B>=2'
> ]
> ```

Since on master, from this point on, `ivadomed=2.9.6` is now a lower bound, I would want to follow their example and pin `ivadomed>=2.9.6`.

kousu: > Just thinking out loud, there should be a strong reason to use an older version of SCT, as a matter of fact for any library unless it's LTS

We're thinking about scientific reproducibility. We want people in 20 years to be able to redo our studies. Or at least have a high chance of making them work.

> The SCT project maintains two sets of requirements, that are roughly equivalent to [`install_requires` vs. `requirements.txt`](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/)

We do, but no one else does. Surely we're going against some convention with that choice What does npm do? What does golang do? `apt`? `pacman`?

> I would want to follow PyPA's recommendations:

I don't know if I trust PyPA's own recommendations. They can't seem to agree on a lot of things. For example, the [setuptools docs don't make any recommendations](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html). Why did they add `~=` and call it the "compatible" clause if no one should use it? Isn't the point of a major version upgrade supposed to be that there's breaking changes and that downstream should watch out?

Though I do see `>=` used in [jaraco/skeleton](https://github.com/jaraco/skeleton/blob/1a6b828304e7a8896b55d9ebf83f481ba7ebd568/setup.cfg#L29-L43), which is afaict the closest thing that that exists to documenting modern best practices.

Ah I don't know. I can't get my head around how to use pip properly. (and I know I'm not alone). Let's just do whatever works. I guess `>=` because jaraco says it's right.

requirements.txt
kousu: Sorry to bump an old thread but I've noticed the way CC's solution falls down. I mean, they're aware and @kanishk16 just quoted it:

>  Note that if one does not require an explicit version when installing the package, and if a more recent version of the package is uploaded on PyPI, pip will still prefer the more recent version.

But to be really clear, this means currently, asking for the cu101 build of torch instead falls over to the PyPI build:

```
$ pip install torch --extra-index-url https://download.pytorch.org/whl/cu101
Defaulting to user installation because normal site-packages is not writeable
Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/cu101
Collecting torch
  Downloading torch-1.11.0-cp310-cp310-manylinux1_x86_64.whl (750.6 MB)
     |█                               | 25.2 MB 10.7 MB/s eta 0:01:08^C
```


Though asking for the [`+cu102`](https://download.pytorch.org/whl/cu102/torch) build and above currently works as expected:

```
$ pip install torch --extra-index-url https://download.pytorch.org/whl/cu102
Defaulting to user installation because normal site-packages is not writeable
Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/cu102
Collecting torch
  Downloading https://download.pytorch.org/whl/cu102/torch-1.11.0%2Bcu102-cp310-cp310-linux_x86_64.whl (750.6 MB)
     |▊                               | 17.1 MB 928 kB/s eta 0:13:10^C
```

But at some point there will be a torch v1.12 and a torch v1.13 on PyPI that aren't not mirrored in https://download.pytorch.org/whl/cu102/torch, and that that point there will be a problem.

**But** this is safe for us because:

2. `ivadomed` currently [*is* pinning `torch`'s version](https://github.com/ivadomed/ivadomed/blob/f063485b083fc5503e1a1ee4ae0d7e947131603f/requirements.txt#L24) to `~=1.8.1`.

   It's possible with `~=` that they could upload a 1.8.3 to PyPI which would then take precedence over https://download.pytorch.org/whl/cpu/torch, but I think it's unlikely for them to do that. I think it's less likely they'll decide to skip uploading to the side-repos for a patch version than for a minor or major version. 
1. unlike the CUDA builds, torch is publishing a `+cpu` for every single version they put out, so we can trust that `+cpu` will break the tie in favor of the CPU build.

So leaning on local version tags is fragile, but should be okay for us.

Issue #3779, opened 2022-04-21
Github action Tests / Windows (Native) (pull_request) does not work for PR from forked repository


RignonNoel:
### Description

Actuallly the `install_sct.bat` allow us to install from a specific branch name and it's this feature that we use in the Github action to verify that every PR opened are installable via the `.bat` script.

However, all contributions does not come from SCT maintainers and using fork is most of the time showed as a good practice to avoid giving too much permissions to contributors and respect least privileges patterns.

We should enhance a bit the `.bat` AND the github action to allow us install SCT from a fork.

### Steps to Reproduce

0. Fork SpinalCordToolbox repo
1. Open a new branch on your fork
2. Create a dumb commit with a little change in the `README`
3. Open a PR on official SCT repo
4. Wait for github action to fail

**Expected behavior:** 

Github action should pass since the change are valid and does not break the install on native windows.

**Actual behavior:**

Github action fail and the PR contribution should be refused.
We need to give contributor access to the repo without maybe knowing him and ask him to move his branch to the official repo before creating a new PR.

RignonNoel:
@joshuacwnewton @mguaypaq I don't know which milestone you would like to put on this one. It seems like we can live with it for a moment and wait after the 5.6 release is officialy finished.

Since it's a CI issue, it could be solve after 5.6 but before 5.7 without stress.

joshuacwnewton:
Thank you for pointing this out! 😄 

This is only somewhat related, but we've run into a similar issues in the past with our linting action (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3286, https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3462). I realize now that it's easy to forget about contributors who are working from forks.

> Since it's a CI issue, it could be solve after 5.6 but before 5.7 without stress.

I think this makes perfect sense. :)

RignonNoel:
> I realize now that it's easy to forget about contributors who are working from forks, aha.

Ahah, I perfectly understand. It's why I never work directly on an official repository. It allow me to be on the same position as any other contributor and I've much less stress when I want to force push change everywhere on my branches 😆

PR #3795, opened 2022-05-18
Try testing Windows/macOS in Release workflow

joshuacwnewton:
Draft PR to debug an attempt to rewrite the `create-release.yml` to test macOS/Windows

Will update once this PR is ready for review.


joshuacwnewton:
This isn't working for the Windows installer, because running `install_sct.bat <git_ref>` only works if there's a `<git_ref>` that corresponds to the changes being made.

But, here, we want to test `requirements-freeze.txt` _without_ pushing to git first. 

This is very much the same root cause as https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3779. So, I think what needs to be done is update `install_sct.bat` so that it can re-use an existing `.git` repository (if it exists), rather than it being mandatory to `git clone` each time.

I'm going to close this for now, but the work here will be valuable once https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3779 is addressed.

PR #3738, opened 2022-03-17
Replace Tensorflow/Keras-based inference (`.h5`) with onnxruntime (`.ONNX`)

joshuacwnewton:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [x] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [x] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [x] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR eliminates our dependency on Tensorflow/keras by replacing our Keras `.h5` models with equivalent `.ONNX` models.

The changes include:

* Keras model code cleanup (Some issues were fixed in the Keras model code, then the source files were moved into `dev/`)
* `deepseg_` module code cleanup (Source files were moved to a central `deepseg_` module)
* A Jupyter notebook was added to `dev/` to generate ONNX models from the Keras code
* Keras inference was replaced with ONNX inference
* The `CroppedRegion` class in `sct_deepseg_gm` was rewritten (to better support fixed-input ONNX models)
* Every trace of Keras/Tensorflow code was removed from the `spinalcordtoolbox/` package
* Keras/Tensorflow-related dependencies were removed.

All of these changes are explained by individual commits, so I highly recommend using the "Commit" view to review this PR.

## Validation

A second Jupyter notebook,  `keras_vs_onnx_model_output_comparison.ipynb`, is used to ensure that the model output is identical between the `.h5` and `.ONNX` versions.

[...]

PR #3807, opened 2022-06-14
Integration of DL multimodal registration in SCT

### LINKED ISSUE #3760: Integrate EvanBeal's contrast agnostic registration model into SCT's registration scripts
This issue is for planning the integration of @EvanBeal's [contrast agnostic registration model project](https://github.com/ivadomed/multimodal-registration) into SCT's registration scripts.

### Proposed changes (copied from 2022-04-05 meeting minutes)

- Include preprocessing steps
- Load the models weights and build the architecture (using [VoxelMorph](https://github.com/voxelmorph/voxelmorph) functions)
- Use the models to produce the warping field
- Transform the warping field so it can be used with sct_apply_transfo
- Call this function from the [register function](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_register_to_template.py#L1189) of `sct_register_to_template` 
    - And/or `sct_register_multimodal`? (TODO: Needs clarification)

### Remaining work

- [ ] Investigate if model can be converted to ONNX, or if we need to depend on VoxelMorph directly
   - NB: If it can't be converted, then SCT will have to depend on Tensorflow once more, which conflicts with the vision for #3735.

EvanBeal:
#### GitHub

#### PR contents

## Description

This PR is about the integration of the [contrast agnostic deep learning based registration project](https://github.com/ivadomed/multimodal-registration) into SCT. 

This is done by adding a new choice of algorithm to use when registering the data via the [`sct_register_multimodal`](https://spinalcordtoolbox.com/user_section/command-line.html#sct-register-multimodal) command line. 

Example of use of this new feature:
```
sct_register_multimodal -i moving_data.nii.gz -d fixed_data.nii.gz -param step=1,type=im,algo=dl
``` 

The changes include:
- 2 new functions in `registration/algorithms.py` to perform registration using contrast agnostic deep learning models 
- 1 new if condition in the `register()` function of `registration/core.py` to be able to use this registration strategy via the `sct_register_multimodal` command line
- 2 new dependencies, added in the `requirements.txt` file, for `nilearn` (used for a preprocessing task) and `voxelmorph` (used for the registration models architecture)

## Testing the PR

To test the method now, you need to download the deep learning models and put them in a folder called `test_data_models` in your SCT folder and then use the code present on this [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/master...jn/3760-add-voxelmorph-registration) (with the correct installation of the packages needed).

First, download and extract `cascaded_models_pytorch.zip` (from [here](https://github.com/ivadomed/multimodal-registration/releases/tag/r20220512)) to a `test_data_models/` folder inside your SCT folder (same location than the `data`, `documentation` or `dev` folder).

Then, choose some data to do your test. For instance, considering the [`spine generic dataset`](https://github.com/spine-generic/data-multi-subject) you can choose to register the [T2w](https://github.com/spine-generic/data-multi-subject/blob/master/sub-geneva06/anat/sub-geneva06_T2w.nii.gz) data of `sub-geneva06` to the [T1w](https://github.com/spine-generic/data-multi-subject/blob/master/sub-geneva06/anat/sub-geneva06_T1w.nii.gz) of the same subject. 

It is **highly** recommended to crop the data before using it, otherwise it is very likely that the process will be killed because it will use too much memory.  

To do so, you can use `sct_deepseg_sc` and `sct_create_mask` to crop around the spinal cord:
```
# Cropping around the spinal cord (T1w)
sct_deepseg_sc -i T1w.nii.gz -c t1 -centerline cnn
sct_create_mask -i T1w.nii.gz -p centerline,T1w_seg.nii.gz -size 35mm -f cylinder -o mask_T1w.nii.gz
sct_crop_image -i T1w.nii.gz -m mask_T1w.nii.gz

# Cropping around the spinal cord (T2w)
sct_deepseg_sc -i T2w.nii.gz -c t2 -centerline cnn
sct_create_mask -i T2w.nii.gz -p centerline,T2w_seg.nii.gz -size 35mm -f cylinder -o mask_T2w.nii.gz
sct_crop_image -i T2w.nii.gz -m mask_T2w.nii.gz
```

Eventually, you can perform registration:
```
sct_register_multimodal -i T2w_crop.nii.gz -d T1w_crop.nii.gz -param step=1,type=im,algo=dl
```

## TODO

- [x] Test the implementation using multiple different images
- [x] Add the deep learning (PyTorch) models at the right place into SCT
- [x] Change the code in function of the location of the deep learning models 

## Linked issues

Resolves #3760.


requirements.txt
```diff
@@ -44,4 +45,5 @@ tqdm
 transforms3d
 urllib3[secure]
 pytest_console_scripts
-wquantiles
+voxelmorph
```
joshuacwnewton, line 48: It looks like installation is currently failing due to the following error:

```
An error occured while importing module voxelmorph -> voxelmorph requires tensorflow version 2.4 or greater, but found version 1.15.5
Full traceback: Traceback (most recent call last):
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_check_dependencies.py", line 272, in main
    module = module_import(module_name, suppress_stderr)
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_check_dependencies.py", line 104, in module_import
    module = importlib.import_module(module_name)
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/site-packages/voxelmorph/__init__.py", line 52, in <module>
    raise ImportError(f'voxelmorph requires tensorflow version {minv} or greater, '
ImportError: voxelmorph requires tensorflow version 2.4 or greater, but found version 1.15.5
```

The good news is, our dependency on `tensorflow` will be removed shortly. So, once https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3738 is merged, we can update this PR against `master`, and the installation errors should be fixed.

joshuacwnewton: #3738 has been merged, so I've gone ahead and updated master. We'll see how the affects the installation on this branch. :)

joshuacwnewton: Huh. Now we're getting a different error:

```
An error occured while importing module voxelmorph -> Please install tensorflow to use this neurite backend
Full traceback: Traceback (most recent call last):
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/site-packages/neurite/__init__.py", line 46, in <module>
    import tensorflow
ModuleNotFoundError: No module named 'tensorflow'
```

So, `voxelmorph` is depending on `neurite`, which depends on `tensorflow`. But, we use parts of `voxelmorph` that don't depend on `tensorflow`... 🤔 

That's not good! `voxelmorph` should be installable without needing to have `tensorflow` installed. I'll dig a little deeper into this.

joshuacwnewton: The relevant backend-importing code from `neurite` is [here](https://github.com/adalca/neurite/blob/741c95d794f82f91d568eb28ef0c8d1f36d509cd/neurite/__init__.py#L32-L51) with the relevant part being the [`get_backend()`](https://github.com/adalca/neurite/blob/741c95d794f82f91d568eb28ef0c8d1f36d509cd/neurite/py/utils.py#L20) function:

> ```
>     return 'pytorch' if os.environ.get('NEURITE_BACKEND') == 'pytorch' else 'tensorflow'
> ```

Awkward. `neurite` defaults to the `tensorflow` backend, and to prevent it from doing so, we have to manually set an environment variable.

This is doable? But, it would be nice if the upstream dependencies could have handled this a little more gracefully.

joshuacwnewton: Perfect! Apart from the linting issues, this PR is ready for testing. :)

EvanBeal: Perfect ! Thank you very much @joshuacwnewton for having looked into this 👌

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -1296,6 +1296,19 @@ def register(src, dest, step, param):
             verbose=param.verbose,
         )
 
+    # deep learning registration using cascaded models
+    elif step.algo == 'dl':
+        # check if user provided a mask-- if so, inform it will be ignored
+        if fname_mask:
+            printv('\nWARNING: algo ' + step.algo + ' will ignore the provided mask.\n', 1, 'warning')
+
+        warp_forward_out, warp_inverse_out = register_step_dl_multimodal_cascaded_reg(
```
joshuacwnewton, line 1305: This function is causing another linting issue (GitHub Actions run [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/6889986178?check_suite_focus=true#step:5:17)):

```
 spinalcordtoolbox/scripts/sct_register_to_template.py:1305:46: F405 'register_step_dl_multimodal_cascaded_reg' may be undefined, or defined from star imports: spinalcordtoolbox.registration.landmarks, spinalcordtoolbox.registration.register
```

But, this isn't your fault -- This was an existing issue! F405 was triggered because we use a star import here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65768163e2e2b00bde46cbf9bc9ee13539fcd5f9/spinalcordtoolbox/scripts/sct_register_to_template.py#L29-L30

(See https://www.flake8rules.com/rules/F405.html for more info.)

For now, you can fix this error by explicitly importing your function:

```diff
 from spinalcordtoolbox.registration.register import * 
+from spinalcordtoolbox.registration.register import register_step_dl_multimodal_cascaded_reg
 from spinalcordtoolbox.registration.landmarks import * 
```

But, outside of this PR, SCT should eventually remove the star-importing in `sct_register_to_template`.

spinalcordtoolbox/registration/register.py
```diff
@@ -329,6 +335,220 @@ def register_step_label(src, dest, step, verbose=1):
...
 
+def register_step_dl_multimodal_cascaded_reg(src, dest, step, verbose=1):
+    """
+    """
+    # Preprocessing steps - TODO maybe not include these here so there is an option for the user to test different ones
+    logger.info(f"\nPreprocess data by setting an isotropic resolution of 1mm "
+                f"and bringing the source image into same space as moving image...")
+    # Isotropic resolution
+    dest_iso_res = image.add_suffix(dest, '_iso_res')
+    run_proc(['sct_resample', '-i', dest, '-o', dest_iso_res, '-mm', '1x1x1'])
```
joshuacwnewton, line 346: I would recommend replacing any `run_proc` calls with calls to `main()`. For example:

```suggestion
    sct_resample.main(['-i', dest, '-o', dest_iso_res, '-mm', '1x1x1'])
```

You can import `sct_resample` using:

```python
from spinalcordtoolbox.scripts import sct_resample
```

joshuacwnewton (changes requested):
Thank you again for this contribution! It's very exciting to see the integration of this project into SCT. ❤️ 😄 

I've added some suggestions that will help us get started.

(There are additional structural suggestions I want to make, but I want to save them until after we can confirm that `algo=dl` works across our platforms.)

spinalcordtoolbox/registration/register.py
```diff
@@ -329,6 +335,220 @@ def register_step_label(src, dest, step, verbose=1):
     return warp_forward_out, warp_inverse_out
 
 
+def register_step_dl_multimodal_cascaded_reg(src, dest, step, verbose=1):
+    """
+    """
+    # Preprocessing steps - TODO maybe not include these here so there is an option for the user to test different ones
+    logger.info(f"\nPreprocess data by setting an isotropic resolution of 1mm "
```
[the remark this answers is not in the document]
EvanBeal: Done in this [commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807/commits/ca4d4159b58dca54f520cc8d422e52922a828db9)

spinalcordtoolbox/scripts/sct_register_to_template.py
EvanBeal: > But, outside of this PR, SCT should eventually remove the star-importing in sct_register_to_template

Yes I realized this as well and thought that it was outside the range of this PR. But indeed it's better to start with the correct way of doing it by explicitly importing the register function in this case :) 

So, this is done in this [commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807/commits/f583d1648ea4270b25a5d19dbf1ebcb080c066fe).

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -1296,6 +1296,19 @@ def register(src, dest, step, param):
             verbose=param.verbose,
         )
 
+    # deep learning registration using cascaded models
+    elif step.algo == 'dl':
```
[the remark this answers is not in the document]
EvanBeal: Thanks for this recommendation ! I don't know a lot about how the tests are working in SCT (and in general 😅), so if you have any recommendations, advices or suggestions to increase the testing coverage of this additional code, that would be awesome :)
For now, I have added your suggestion with this simple test in this [commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807/commits/1d5734435ea0fd526600b92fef218e300814875a).

EvanBeal:
Thanks a lot @joshuacwnewton for your review :)
I agree with all your points and have added the different modifications you suggested ! 
There are still a lot of checks failing (maybe weirdly even more than before) but from what I see and understand it doesn't seem to be necessarily linked to the modifications done in this PR, so I don't really know how this should be resolved. What do you think ?

spinalcordtoolbox/registration/register.py
```diff
@@ -29,6 +29,7 @@
 import spinalcordtoolbox.image as image
 from spinalcordtoolbox.math import laplacian
 from spinalcordtoolbox.registration.landmarks import register_landmarks
+from spinalcordtoolbox.scripts import sct_resample, sct_register_multimodal
```
joshuacwnewton, line 32: > There are still a lot of checks failing (maybe weirdly even more than before) but from what I see and understand it doesn't seem to be necessarily linked to the modifications done in this PR, so I don't really know how this should be resolved. What do you think ?

Ahhhh, I see the problem. This line has caused a [`circular import`](https://stackoverflow.com/q/22187279), In other words:

* `register.py` imports `sct_register_multimodal`
* But, `sct_register_multimodal` imports functions from `register.py`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b12fe5d3b615431355dd01a845fe596ea5f9dbba/spinalcordtoolbox/scripts/sct_register_multimodal.py#L41

There are 2 fixes we could apply here:

1. Try to not use `sct_register_multimodal`, but instead try to pre-register the images using the `register_wrapper()` function directly.
    * The problem with this is that right now, `register_wrapper()` is *also* in a CLI script right now (`sct_register_to_template`), so I fear that this would also cause a circular import error (since, `register.py` would import from `sct_register_to_template`, and `sct_register_to_tempalte` would import from `register.py`).
    * So, I think to solve the root cause, we would need to perform a deeper refactor to better separate the CLI/API. 
    * But, I think we can save this until after we finish up this PR.
2. The alternative is to apply the workaround from [this SO post](https://stackoverflow.com/a/22210807):

	  > [Circular imports] can work just fine, if you set them up correctly.
	  > 	  
	  > The easiest way to do so is to use import `my_module` syntax, rather than `from my_module import some_object`. The former will almost always work, even if `my_module` included imports us back. The latter only works if `my_object` is already defined in `my_module`, which in a circular import may not be the case.

I'm going to try and add a commit that does this to see if I can fix the issue.

joshuacwnewton: Even after fixing the first circular import, it looks like there are even more circular imports.

Perhaps tackling 1. now *is* the best option, rather than putting it off.

joshuacwnewton: Update: Moving `register_wrapper` out of `sct_register_to_template` is a lot harder than I anticipated, because `register_wrapper` is designed to use a `Param` object that gets defined inside of `sct_register_to_template`. 

But, I think I have a workable solution here...

EvanBeal: > Ahhhh, I see the problem. The change from https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/7e43cf8d90b7557c03257e7d1088497b378f20c8 has caused a [circular import](https://stackoverflow.com/q/22187279)

Oh yes you're right, good catch !

> Okay! I've opened https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3811, which moves register_wrapper() out of sct_register_to_template. This allows us to use register_wrapper() to perform the -identity step, rather than sct_register_multimodal, which will fix the circular import error.

Ok perfect, that's really cool thank you ;)

> To confirm that the tests all pass, I ran [this GitHub Actions run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/2511107436).

Looks nice !!

> Once https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3811 is merged, we can remake the PR using the new branch.

Alright ! Thank you very much for looking into this in depth :)

joshuacwnewton: #3811 has been merged, so I've done two things:

* Rebased this PR onto master (which incorporates the changes from [#3811](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3811)).
* Added commit [1dfe5dc](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807/commits/1dfe5dc80beef565e77d8ea6495d23cf73e8c4b2), which was originally from my test branch.

This PR should now be ready for a proper review!

EvanBeal:
Thank you @joshuacwnewton for having fixed the circular importing issue ! Indeed everything seems to be working now to do a proper review 👍

To test different scenarios and the registration results on different images I think that we could use shell scripts (to facilitate the process and the reproducibility). 

So, here is an example that could be used as a starting point and therefore may be useful for the testing process. This example has been done to perform intra-subject registration (by default T2w registration to T1w) for all the subjects of a dataset organized with the BIDS convention (here the dataset is named `test_data`). It uses `sct_deepseg_sc` and `sct_create_mask` to crop around the spinal cord and then it performs the registration with `sct_register_multimodal -i $file_mov -d $file_fx -param step=1,type=im,algo=dl`. Eventually, it produces a quality control report to observe the spinal cord registration.

To run this script: 
```
sct_run_batch -jobs 1 -path-data test_data -path-out res_registration -script test_registration.sh
```

<details>
  <summary>Here is the script (test_registration.sh)
</summary>
  
```
#!/bin/bash
#
# Register moving volumes to fixed volumes using sct_register_multimodal command line with algo=dl
#
# Usage:
# sct_run_batch -jobs 1 -path-data test_data -path-out res_registration -script test_registration.sh
#
#
# Author: Evan Béal

set -x
# Immediately exit if error
set -e -o pipefail

# Exit if user presses CTRL+C (Linux, OSX)
trap "echo Caught Keyboard Interrupt within script. Exiting now.; exit" INT

# Retrieve input params
SUBJECT=$1

# Save script path
PATH_SCRIPT=$PWD

# get starting time:
start=`date +%s`

# FUNCTIONS
# ==============================================================================

# Perform segmentation.
segment(){
  local file="$1"
  local contrast="$2"

  echo "Proceeding with automatic segmentation"
  # Segment spinal cord
  sct_deepseg_sc -i $file -c $contrast -qc ${PATH_QC}_seg -qc-subject ${SUBJECT}
}

# SCRIPT STARTS HERE
# ==============================================================================
# Display useful info for the log, such as SCT version, RAM and CPU cores available
sct_check_dependencies -short

SUBJECT_ID=$(dirname "$SUBJECT")
SES=$(basename "$SUBJECT")

# PARAMETERS TO SPECIFY
# ==============================================================================
# Choose the name and the extension of the fixed volume (ex: T1w (or acq-ax_T1w) and .nii.gz) and its contrast (ex: t1)
FX_NAME="T1w"
FX_EXT=".nii.gz"
FX_CONTRAST="t1"  # This parameter is used for SC segmentation and should be one of {t1, t2, t2s, dwi}
# Choose the name and the extension of the moving volume (ex: T2w and .nii.gz) and its contrast (ex: t2)
MOV_NAME="T2w"
MOV_EXT=".nii.gz"
MOV_CONTRAST="t2"  # This parameter is used for SC segmentation and should be one of {t1, t2, t2s, dwi}
# Specify if multi-sessions are available per subject: if yes set the value to 1, if no set the value to 0
MULT_SESSIONS=0
# ==============================================================================

# Go to folder where data will be copied and processed
cd ${PATH_DATA_PROCESSED}
# Copy source images
mkdir -p ${SUBJECT}
rsync -avzh $PATH_DATA/$SUBJECT/ ${SUBJECT}
# Go to anat folder where all structural data are located
echo $PWD
cd ${SUBJECT}/anat/

if [ $MULT_SESSIONS == 1 ]
then
  file_fx_before_proc="${SUBJECT_ID}_${SES}_${FX_NAME}${FX_EXT}"
  file_mov_before_proc="${SUBJECT_ID}_${SES}_${MOV_NAME}${MOV_EXT}"
else
  file_fx_before_proc="${SES}_${FX_NAME}${FX_EXT}"
  file_mov_before_proc="${SES}_${MOV_NAME}${MOV_EXT}"
fi

# Cropping around the spinal cord

# 1. Segment the spinal cord
segment $file_fx_before_proc $FX_CONTRAST
segment $file_mov_before_proc $MOV_CONTRAST

if [ $MULT_SESSIONS == 1 ]
then
  file_fx_seg="${SUBJECT_ID}_${SES}_${FX_NAME}_seg${FX_EXT}"
  file_mov_seg="${SUBJECT_ID}_${SES}_${MOV_NAME}_seg${MOV_EXT}"
else
  file_fx_seg="${SES}_${FX_NAME}_seg${FX_EXT}"
  file_mov_seg="${SES}_${MOV_NAME}_seg${MOV_EXT}"
fi

# 2. Create a mask around the spinal cord segmentation
sct_create_mask -i $file_fx_before_proc -p centerline,$file_fx_seg -size 35mm -f cylinder -o mask_fx.nii.gz
sct_create_mask -i $file_mov_before_proc -p centerline,$file_mov_seg -size 35mm -f cylinder -o mask_mov.nii.gz

# 3. Crop the image using the mask
sct_crop_image -i $file_fx_before_proc -m mask_fx.nii.gz
sct_crop_image -i $file_mov_before_proc -m mask_mov.nii.gz

if [ $MULT_SESSIONS == 1 ]
then
  file_fx="${SUBJECT_ID}_${SES}_${FX_NAME}_crop${FX_EXT}"
  file_mov="${SUBJECT_ID}_${SES}_${MOV_NAME}_crop${MOV_EXT}"
else
  file_fx="${SES}_${FX_NAME}_crop${FX_EXT}"
  file_mov="${SES}_${MOV_NAME}_crop${MOV_EXT}"
fi

# Registration
sct_register_multimodal -i $file_mov -d $file_fx -param step=1,type=im,algo=dl

if [ $MULT_SESSIONS == 1 ]
then
  file_fx_reg="${SUBJECT_ID}_${SES}_${FX_NAME}_crop_reg${FX_EXT}"
  file_mov_reg="${SUBJECT_ID}_${SES}_${MOV_NAME}_crop_reg${MOV_EXT}"
else
  file_fx_reg="${SES}_${FX_NAME}_crop_reg${FX_EXT}"
  file_mov_reg="${SES}_${MOV_NAME}_crop_reg${MOV_EXT}"
fi

# Quality Control
# Generate QC report to assess registration
sct_qc -i ${file_fx} -s ${file_fx_seg} -d ${file_mov} -p sct_register_multimodal -qc ${PATH_QC}_reg -qc-subject ${SUBJECT}
sct_qc -i ${file_fx} -s ${file_fx_seg} -d ${file_mov_reg} -p sct_register_multimodal -qc ${PATH_QC}_reg -qc-subject ${SUBJECT}


# Verify presence of output files and write log file if error
# ------------------------------------------------------------------------------
FILES_TO_CHECK=(
  "${PATH_DATA_PROCESSED}/${SUBJECT}/anat/${file_fx}"
  "${PATH_DATA_PROCESSED}/${SUBJECT}/anat/${file_fx_reg}"
  "${PATH_DATA_PROCESSED}/${SUBJECT}/anat/${file_mov}"
  "${PATH_DATA_PROCESSED}/${SUBJECT}/anat/${file_mov_reg}"
)

pwd
for file in ${FILES_TO_CHECK[@]}; do
  if [[ ! -e $file ]]; then
    echo "${file} does not exist" >> $PATH_LOG/_error_check_output_files.log
  fi
done

# Display useful info for the log
end=`date +%s`
runtime=$((end-start))
echo
echo "~~~"
echo "SCT version: `sct_version`"
echo "Ran on:      `uname -nsr`"
echo "Duration:    $(($runtime / 3600))hrs $((($runtime / 60) % 60))min $(($runtime % 60))sec"
echo "~~~"
```
</details>

EvanBeal:
> My one concern is that I'm not sure that this new script fits very well into into SCT's existing automatic testing infrastructure.

Yes I totally agree. The purpose of this script was more to allow us to manually test this PR than to be used for automatic testing. 
It wasn't really clear indeed in my comment because I didn't think about automatic tests at all while writing it to be honest ^^'

> And rewrite the section so that it contains a shortened version of your script

Yes that's a good idea 👌

spinalcordtoolbox/registration/core.py
```diff
@@ -392,6 +392,19 @@ def register(src, dest, step, param):
             verbose=param.verbose,
         )
 
+    # deep learning registration using cascaded models
+    elif step.algo == 'dl':
+        # check if user provided a mask-- if so, inform it will be ignored
+        if fname_mask:
+            printv('\nWARNING: algo ' + step.algo + ' will ignore the provided mask.\n', 1, 'warning')
```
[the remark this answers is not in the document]
joshuacwnewton: Actually, on second thought, this is subtly different behavior vs. the stated purpose of `-m`:

```
  -m <file>             Mask that can be created with sct_create_mask to improve accuracy over region 
                        of interest. This mask will be used on the destination image. Example: mask.nii.gz
```

 `-m` isn't used by the deepreg method in this way, so it might be best to keep things simple...

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -329,6 +337,220 @@ def register_step_label(src, dest, step, verbose=1):
     return warp_forward_out, warp_inverse_out
 
 
+def register_step_dl_multimodal_cascaded_reg(src, dest, step, verbose=1):
+    """
+    """
+    # Preprocessing steps - TODO maybe not include these here so there is an option for the user to test different ones
```
joshuacwnewton, line 343: Note to resolve this TODO

joshuacwnewton: @EvanBeal Did you have any thoughts about which options you would want to expose to the user to be able to change? (e.g. changing isotropic resolution, perhaps?)

It's tricky, because the current `-param` options don't give us much flexibility in terms of exposing options to the user. The closest I can think of is like, `shrink`? 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2ea3643585099860751ce2bf5fb76919dd580f86/spinalcordtoolbox/scripts/sct_register_multimodal.py#L189-L191

But the usage here is kind of arcane. For example, if we want to use 0.9mm isotropic resolution, we would have to specify a `shrink` value of 1/0.9 == `1.11..`. 

It might be easiest to just... leave the defaults as is, then. :sweat:

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -329,6 +337,220 @@ def register_step_label(src, dest, step, verbose=1):
...
+
+    # Compose the two warping fields together (from the first and second registration)
+
+    # Forward registration
+    warp_data = warp_data_exp_first + warp_data_exp_second
+    warp = Nifti1Image(warp_data, fx_preproc_nii.affine)
+    # Set the intent code to vector
+    warp.header['intent_code'] = 1007
+    # Save the composed warping field [forward]
+    save(warp, fname_warp_forward)
```
joshuacwnewton, line 544: Which resources did you use to write this section? (e.g. did you use any NIfTI documentation for figuring out that `1007` is the intent code for vector arrays?)

I ask because I just want to make sure we're not missing anything for creating these NIfTI files... want to make sure we're setting everything correctly. So, it would be nice to have a link to refer to that explains what we're doing here.

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -329,6 +337,220 @@ def register_step_label(src, dest, step, verbose=1):
...
+
+    # Reverse registration
+    warp_data_rev = -warp_data  # TODO - observe if it provides good results
```
joshuacwnewton, line 547: Adding a note for this TODO here.

joshuacwnewton: @EvanBeal Do you think this TODO is still needed? Has this worked the way you expected it to?

joshuacwnewton (changes requested):
I've added a few more comments/discussion points for this PR.

I don't really have many more structural comments on the DL functions themselves to make. Since, this addition is kind of like a black box in some ways: It's an isolated set of steps that are applicable only in the context of your method. So, I don't see much point in, for example, factoring out steps into smaller re-usable chunks.  

The only other changes I can think of would be to try and tweak and improve the method itself? But again, I imagine you've already done much of this R&D during your time in the lab. So, perhaps this PR should mainly focus on getting the first draft out there to users, rather than focusing on evaluating the performance.

Given that, we might not even need to test the PR that thoroughly, beyond ensuring that output files are produced correctly. :thinking:

spinalcordtoolbox/registration/algorithms.py
EvanBeal: > Did you have any thoughts about which options you would want to expose to the user to be able to change? (e.g. changing isotropic resolution, perhaps?)

The preprocessing that could be changed by the user is indeed this isotropic resolution of 1 mm. I have set this value by default as I have done most of the tests in my project with this preprocessing but in fact as the registration models are trained on synthetic data, they are not specific to data with an isotropic resolution of 1 mm. Therefore, we could expect good registration results for data with a different preprocessing as well, but there are more uncertainties as it has not been test in depth.

So, I think that the ideal solution would have been to set the isotropic resolution of 1 mm and then have an option to potentially change it (or maybe even remove this preprocessing step). But I agree that it's quite complex because we don't really want to add a param (and need to refactor everything) just for this special case and as you said using the `shrink` parameter may be a bit tricky. 

Therefore, I guess that for now we could either leave it as it stands or remove it and mention that the user can do this preprocessing step before using the method if he doesn't observe satisfactory results. 

I think that the first option of leaving it as it is now is probably the best to facilitate the use of the method.

spinalcordtoolbox/registration/core.py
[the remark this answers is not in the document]
EvanBeal: Yes it was a good idea but you are right `-m` is in fact used to focus the registration on a specific area, while our recommendations are mostly to reduce the images size and are not used to lead to more accurate registration results. So I agree with you, I think that we can't really use `-m` here.

install_sct
```diff
@@ -674,7 +674,7 @@ else
   print info "Installing data..."
   run rm -rf "$SCT_DIR/$DATA_DIR"
   run mkdir -p "$SCT_DIR/$DATA_DIR"
-  for data in PAM50 gm_model optic_models pmj_models deepseg_sc_models deepseg_gm_models deepseg_lesion_models c2c3_disc_models; do
+  for data in PAM50 gm_model optic_models pmj_models deepseg_sc_models deepseg_gm_models deepseg_lesion_models c2c3_disc_models deepreg_models; do
```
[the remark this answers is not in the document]
EvanBeal: I don't have a strong opinion on this but I think that your suggestion of "automatically download the models at runtime" may be interesting to not increase the installation size of SCT with models that may be never used by the users. 

> Doing this might cause issues for user-run batch scripts, though, since they might simultaneously download copies of the same files in parallel

Oh yes, in this case it can be problematic but I would agree that it is probably a minimal concern. But then again, I don't know much about it ^^'

spinalcordtoolbox/registration/algorithms.py
EvanBeal: Thanks for pointing this out ! You're right it appears in the code as a black box but to decide to set the `intent_code` to `1007`, I have used the information found on [this](https://brainder.org/2012/09/23/the-nifti-file-format/#:~:text=As%20above%2C%20vector%20per%20voxel%2C%20stored%20in%20the%205th%20dimension.) website and confirmed it with the documentation found [here](http://users.bmap.ucla.edu/~mchamber/npl/nifti_8h_source.html#l00470:~:text=470%C2%A0%23define%20NIFTI_INTENT_VECTOR%20%20%20%201007%20%20%20/*%20for%20any%20other%20type%20of%20vector%20*/).

I have added these two links in the code with this [commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807/commits/92a1a9084fe5a50b3b031888ca4b6ee25e85333e).

spinalcordtoolbox/registration/algorithms.py
EvanBeal: Yes it seems to work as expected so I have removed it now.

spinalcordtoolbox/registration/algorithms.py
joshuacwnewton: I checked the links that you provided, and noticed that the spec says:

```
#define NIFTI_INTENT_DISPVECT  1006   /* specifically for displacements */
#define NIFTI_INTENT_VECTOR    1007   /* for any other type of vector */
```

My guess would have been that a warping field represents displacement vectors? So, I would have guessed that the intent code would be 1006. :thinking: 

Just to check, I've run `sct_register_to_template` using an ANTs algorithm (slicereg) then ran `sct_image -header` to check the metadata:

```console
joshua@tadpole:/tmp/pytest-of-joshua/pytest-1/test_sct_register_multimodal_mcurrent$ sct_image -i warp_mt02mt1.nii.gz -header

--
Spinal Cord Toolbox (git-jn/3760-add-voxelmorph-registration-1dfe5dc80beef565e77d8ea6495d23cf73e8c4b2*)

sct_image -i warp_mt02mt1.nii.gz -header
--

sizeof_hdr	348
data_type	FLOAT32
dim		[5, 40, 40, 5, 1, 3, 1, 1]
vox_units	mm
time_units	Unknown
datatype	16
nbyper		4
bitpix		32
pixdim		[-1.0, 0.84375, 0.84375, 17.0, 0.0, 0.0, 0.0, 0.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	0
freq_dim	0
slice_dim	0
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Vector
intent_code	1007
intent_name	
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	-0.843379 -0.000000 -0.503756 10.726137 
qto_xyz:2	-0.001874 0.841376 1.273740 -17.241285 
qto_xyz:3	-0.024932 -0.063247 16.944728 -25.065205 
qform_xorient	Right-to-Left
sform_name	Unknown
sform_code	0
sto_xyz:1	0.000000 0.000000 0.000000 0.000000 
sto_xyz:2	0.000000 0.000000 0.000000 0.000000 
sto_xyz:3	0.000000 0.000000 0.000000 0.000000 
sform_xorient	Unknown
sform_yorient	Unknown
sform_zorient	Unknown
file_type	NIFTI-1+
file_code	1
descrip		
aux_file
```

Comparing to files generated by `algo=dl`:

```console
joshua@tadpole:/tmp/pytest-of-joshua/pytest-0/test_sct_register_multimodal_mcurrent$ sct_image -i warp_mt02mt1.nii.gz -header

--
Spinal Cord Toolbox (git-jn/3760-add-voxelmorph-registration-1dfe5dc80beef565e77d8ea6495d23cf73e8c4b2*)

sct_image -i warp_mt02mt1.nii.gz -header
--

sizeof_hdr	348
data_type	FLOAT32
dim		[5, 40, 40, 5, 1, 3, 1, 1]
vox_units	mm
time_units	Unknown
datatype	16
nbyper		4
bitpix		32
pixdim		[-1.0, 0.84375, 0.84375, 17.0, 0.0, 0.0, 0.0, 0.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	0
freq_dim	0
slice_dim	0
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Vector
intent_code	1007
intent_name	
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	-0.843379 -0.000000 -0.503756 10.726137 
qto_xyz:2	-0.001874 0.841376 1.273740 -17.241285 
qto_xyz:3	-0.024932 -0.063247 16.944728 -25.065205 
qform_xorient	Right-to-Left
sform_name	Unknown
sform_code	0
sto_xyz:1	0.000000 0.000000 0.000000 0.000000 
sto_xyz:2	0.000000 0.000000 0.000000 0.000000 
sto_xyz:3	0.000000 0.000000 0.000000 0.000000 
sform_xorient	Unknown
sform_yorient	Unknown
sform_zorient	Unknown
file_type	NIFTI-1+
file_code	1
descrip		
aux_file
```

They are identical. `1007` seems to be the correct code, and there is no metadata missing! :)

batch_processing.sh
```diff
@@ -126,11 +126,23 @@ cd ..
...
+sct_deepseg_sc -i t1.nii.gz -c t1 -centerline cnn
+sct_create_mask -i t1.nii.gz -p centerline,t1_seg.nii.gz -size 35mm -f cylinder -o mask_t1.nii.gz
+sct_crop_image -i t1.nii.gz -m mask_t1.nii.gz
+# 2. t2w preprocessing (cropping around spinal cord)
+cp ../t2/t2.nii.gz ./t2.nii.gz
+sct_deepseg_sc -i t2.nii.gz -c t2
+sct_create_mask -i t2.nii.gz -p centerline,t2_seg.nii.gz -size 35mm -f cylinder -o mask_t2.nii.gz
+sct_crop_image -i t2.nii.gz -m mask_t2.nii.gz
+# 3. Perform registration
+sct_register_multimodal -i t1_crop.nii.gz -d t2_crop.nii.gz -param step=1,type=im,algo=dl
```
joshuacwnewton, line 140: I've added the registration pipeline for T1/T2 data in `sct_example_data`. Here are some comparison screenshots:

`t1_crop_reg.nii.gz` + `t2_crop.nii.gz`:

ezgif-3-8492cdd8c6

Zoomed in, with increased contrast on the T1 image:

ezgif-3-8e868e9e6e

jcohenadad: this is amazing! the QC you performed/documented is excellent (i wouldn't have done a better job)

joshuacwnewton (approved):
I'm going to give this a soft approval now, since I'm not sure I have anything else to add.

But, given that SCT's biweekly meeting is coming up tomorrow, I'd like to use that as an opportunity to resolve some of the remaining open questions about this contribution:

* Model installation: `install_sct` vs. check if model files are present at runtime
* Model testing: Should we do any manual testing with different image files? Or release the model as-is, and let our users do the testing?
* Model advertisement: How should we get the word out to our users? New tutorial? Include as part of our SCT course? etc.

PR #3833, opened 2022-07-07
Use `shutil.copyfile` for output files if src/dest are on different filesystems

### LINKED ISSUE #3832: `shutil.move` fails when source/destination are on different filesystems
## Context

Currently, we use `shutil.move()` in our `generate_output_file()` function:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/50f030fa55609f4e1020705acc12161b51b4ba05/spinalcordtoolbox/image.py#L1480-L1481

Specifically, often a CLI script will do its most of its processing in `/tmp/`, then use `generate_output_file()` to rename+move any output files from `tmp/` into the user-specified output folder, e.g.:

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Before this PR, if you try to run a CLI script within WSL from an external HDD, trying to generate output files will fail:

```
Generate output files...
File warp_template2anat.nii.gz already exists. Deleting it..
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 566, in move
    os.rename(src, real_dst)
OSError: [Errno 18] Invalid cross-device link: '/tmp/sct-20220707125005.221636-register_to_template-5_vcb25c/warp_template2anat.nii.gz' -> 'warp_template2anat.nii.gz'

During handling of the above exception, another exception occurred:

  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_register_to_template.py", line 800, in <module>
    main(sys.argv[1:])
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_register_to_template.py", line 769, in main
    generate_output_file(os.path.join(path_tmp, "warp_template2anat.nii.gz"), os.path.join(path_output, "warp_template2anat.nii.gz"), verbose=verbose)
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/[image.py](http://image.py/)", line 1481, in generate_output_file
    shutil.move(fname_in, fname_out)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 580, in move
    copy_function(src, real_dst)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 267, in copy2
    copystat(src, dst, follow_symlinks=follow_symlinks)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 206, in copystat
    follow_symlinks=follow)
PermissionError: [Errno 1] Operation not permitted
```

After this PR, the error is mitigated by copying the files instead:

```
Generate output files...
File warp_template2anat.nii.gz already exists. Deleting it..
File created: warp_template2anat.nii.gz
File warp_anat2template.nii.gz already exists. Deleting it..
File created: warp_anat2template.nii.gz
File template2anat.nii.gz already exists. Deleting it..
File created: template2anat.nii.gz
File anat2template.nii.gz already exists. Deleting it..
File created: anat2template.nii.gz
File warp_curve2straight.nii.gz already exists. Deleting it..
File created: warp_curve2straight.nii.gz
File warp_straight2curve.nii.gz already exists. Deleting it..
File created: warp_straight2curve.nii.gz
File straight_ref.nii.gz already exists. Deleting it..
File created: straight_ref.nii.gz

Delete temporary files...
rm -rf

/tmp/sct-20220707130619.241170-register_to_template-1kj50w47

Finished! Elapsed time: 86s
```

For regular usage, I also verified that src/dest have the same device ID (e.g. in our test suite).

## Linked issues

Fixes #3832.


mguaypaq (changes requested):
I don't think this is the right fix. The [docs for shutil.move](https://docs.python.org/3/library/shutil.html#shutil.move) mention that it already has a mechanism to handle different filesystems; it detects this case and uses a `copy_function` which can be supplied as an argument. Indeed, the stack trace in the PR description shows that it's inside the default `copy_function=copy2` that the problem arises, as `copystat` fails to copy some file metadata from the source to the destination. Passing a different `copy_function` seems like the right move here. Some suggestions:
- The docs suggest using `shutil.copy` instead of `shutil.copy2`, with the slight cost that this doesn't even try to copy the file metadata.
- Another possibility: a small custom function which does a `try shutil.copy2` with a fallback of `shutil.copy`? It might be tricky to reason about all the failure cases there, though, so maybe not.

joshuacwnewton:
> The [docs for shutil.move](https://docs.python.org/3/library/shutil.html#shutil.move) mention that it already has a mechanism to handle different filesystems; it detects this case and uses a `copy_function` which can be supplied as an argument. Indeed, the stack trace in the PR description shows that it's inside the default `copy_function=copy2` that the problem arises, as `copystat` fails to copy some file metadata from the source to the destination.

Thanks for catching this! I didn't realize that `shutil.move()` has a mechanism to handle this. I'd be happy to try another `copy_function`.

----

Just to explain my thought process a little, though: I had interpreted the stack trace a little differently? Given the mention of `During handling of the above exception, another exception occurred`, I had presumed that the first exception was the root cause, and that the second exception was a side-effect of the first exception.

As well, I had connected the presence of `os.rename()` within that first exception with [other reports](https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link) of `shutil.move` being insufficient:

> `os.rename` only works if source and destination are on the same file system. You should use `shutil.move` instead. 

> My version if `shutil.move()` is actually implemented in terms of `os.rename()`

> As Kilian says `shutil.move()` with different file systems gives the same error as `os.rename()`. Easiest to use `shutil.copy()` and `os.remove()` instead.

Reading those reports, I presumed that `shutil.move()` wasn't able to handle differing filesystems. :thinking:

joshuacwnewton:
Regarding your proposed solution:

> Passing a different `copy_function` seems like the right move here. Some suggestions:
> 
>     * The docs suggest using `shutil.copy` instead of `shutil.copy2`, with the slight cost that this doesn't even try to copy the file metadata.

Given that `generate_output_file` is always moving a newly-created file from one place to another, I wouldn't *think* that losing the metadata is a huge deal here? We'd be losing:

> the permission bits, last access time, last modification time, and flags from src to dst. 

The times here definitely don't matter, I think, since the file was newly-created. I'm not quite sure about the permission bits and flags, though. I'm not sure what defaults come out of `nibabel`, vs. the defaults after the metadata is lost. :thinking:

mguaypaq:
On second thought, my second suggestion is definitely overkill for this situation. Just using `shutil.copy` instead of `shutil.copy2` seems like the right move, if you can check that it fixes things in your reproduction of the bug.

joshuacwnewton:
Subbing in `shutil.copy` instead of `shutil.copy2` results in a similar error as before, this time in a different context:

<details>

```
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 566, in move
    os.rename(src, real_dst)
OSError: [Errno 18] Invalid cross-device link: '/tmp/sct-20220711095838.154535-register_to_template-r49rswb1/warp_template2anat.nii.gz' -> 'warp_template2anat.nii.gz'

During handling of the above exception, another exception occurred:

  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_register_to_template.py", line 800, in <module>
    main(sys.argv[1:])
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_register_to_template.py", line 769, in main
    generate_output_file(os.path.join(path_tmp, "warp_template2anat.nii.gz"), os.path.join(path_output, "warp_template2anat.nii.gz"), verbose=verbose)
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/image.py", line 1484, in generate_output_file
    shutil.move(fname_in, fname_out, copy_function=shutil.copy)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 580, in move
    copy_function(src, real_dst)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 249, in copy
    copymode(src, dst, follow_symlinks=follow_symlinks)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 144, in copymode
    chmod_func(dst, stat.S_IMODE(st.st_mode))
PermissionError: [Errno 1] Operation not permitted: 'warp_template2anat.nii.gz'
```
</details>

It turns out that even the bare `copy` function [still copies _some_ file metadata](https://docs.python.org/3/library/shutil.html#shutil.copy):

> [copy()](https://docs.python.org/3/library/shutil.html#shutil.copy) copies the file data and the file’s permission mode (see [os.chmod()](https://docs.python.org/3/library/os.html#os.chmod)). Other metadata, like the file’s creation and modification times, is not preserved. To preserve all file metadata from the original, use [copy2()](https://docs.python.org/3/library/shutil.html#shutil.copy2) instead.

Next, I tried passing the `shutil.copyfile` function to `copy_function=` instead, since that function had worked for me in [my initial (flawed) solution](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3833/commits/a6dbf484dcaa7461d7ca9ec6f5f3dce5986b6acd). Luckily, it works as expected:

<details>
```
Generate output files...
File warp_template2anat.nii.gz already exists. Deleting it..
File created: warp_template2anat.nii.gz
File warp_anat2template.nii.gz already exists. Deleting it..
File created: warp_anat2template.nii.gz
File template2anat.nii.gz already exists. Deleting it..
File created: template2anat.nii.gz
File anat2template.nii.gz already exists. Deleting it..
File created: anat2template.nii.gz
File warp_curve2straight.nii.gz already exists. Deleting it..
File created: warp_curve2straight.nii.gz
File warp_straight2curve.nii.gz already exists. Deleting it..
File created: warp_straight2curve.nii.gz
File straight_ref.nii.gz already exists. Deleting it..
File created: straight_ref.nii.gz
```
</details>

I was curious why `copyfile` doesn't run into the same issues as the other copy functions. But, [the docs say](https://docs.python.org/3/library/shutil.html#shutil.copyfile):

> Copy the contents (no metadata) of the file named src to a file named dst and return dst in the most efficient way possible.

No metadata1

PR #3835, opened 2022-07-08
Fix executable file permissions and remove unnecessary header declarations

mguaypaq:
## Description
I noticed that a few of the scripts under `scripts/` were not executable, and that a few files outside of there *were* executable, so I decided to fix the inconsistencies. This ended up being bigger than I expected, sorry. This PR may be easier to review commit-by-commit.

While I was fixing top-of-file metadata (removing spurious shebang lines), I went ahead and removed the old Python2-style utf-8 encoding declarations, which [shouldn't be needed anymore](https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3). If that's controversial, it's a single commit that can be easily dropped. If we keep it, I should remember to:
- [x] Remove the utf-8 coding declaration from [the CLI template on the dev wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-CLI-script-structure#1-setup-description-and-attribution).

## Linked issues
None.


spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8
```
joshuacwnewton, line 2: > They used to be necessary in python2, but utf-8 is the default in python3. 

[Indeed!](https://peps.python.org/pep-3120/)

I imagine this is a holdover from SCT's Python2 days, and was left behind after the transition to Python3. 

I'm guilty of this too -- I've definitely been copying and pasting the file header declarations into new modules without really thinking about it? Basically treating old modules as templates for new modules. (And, from the git blames on various files, it seems like this is what other SCT devs have done too.)

Thank you for taking the time to stop and question whether these declarations are even necessary. :)

joshuacwnewton (approved):
These changes look sensible and well-scoped to me! Thank you for taking the time to make these smaller, easy-to-review maintenance changes. :)

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: > Remove the utf-8 coding declaration from [the CLI template on the dev wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-CLI-script-structure#1-setup-description-and-attribution).

Done!

PR #3836, opened 2022-07-11
Refactor release workflow into 3 stages to properly test PR branches 

### LINKED ISSUE #3140: Make CI test the `SCT_INSTALL_TYPE=package` path

### LINKED ISSUE #3794: ModuleNotFoundError for `msvc_runtime` when running `sct_check_dependencies`
Forum thread: https://forum.spinalcordmri.org/t/windows-os-specific-sct-installation/878?u=joshuacwnewton

Full traceback:

```python
Check if msvc-runtime is installed..................[FAIL]
An error occured while importing module msvc-runtime -> No module named 'msvc_runtime'
Full traceback: Traceback (most recent call last):
  File "c:\users\chauc\spinalcordtoolbox\spinalcordtoolbox\scripts\sct_check_dependencies.py", line 271, in main
    module = module_import(module_name, suppress_stderr)
  File "c:\users\chauc\spinalcordtoolbox\spinalcordtoolbox\scripts\sct_check_dependencies.py", line 103, in module_import
    module = importlib.import_module(module_name)
  File "C:\Users\chauc\AppData\Local\Programs\Python\Python37\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'msvc_runtime'
```

I can reproduce this issue with my own local installation of SCT on my Windows development machine.

My initial hunch is that this is an issue with `sct_check_dependencies`, and not an actual installation issue. (In other words, the underlying installation is okay.) I think this might be specifically due to a mismatch between the names `msvc_runtime` and `msvc-runtime`.

### LINKED ISSUE #3799: `create-release.yml` is not currently testing changes from PR branches
In order to create a release, currently we check out the `release` branch and merge `master` into it. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/20d2f599c6bccffbcc0a852ad6acd3b71d509457/.github/workflows/create-release.yml#L52-L55

We turned this into a test in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3678. But, because we're still just merging `master` into `release`, we're testing `master`, and not the PR branch itself.

To fix this, we should probably adopt https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3732, so that we no longer have to do this "merge master into release" step, and can more easily test the PR branch directly.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

Currently, `create-release.yml` workflow isn't actually testing PR branches, nor is it testing macOS/Windows, because it uses a single job with a single OS (`ubuntu-latest`) and single git branch (`master`).

To make testing easier, this PR splits the workflow into 3 distinct stages, because doing so gives us more control over which OS and git branch we use at each stage of the release process. 

With the new changes, this PR should now catch the bug #3794 and fail accordingly. I can then follow up with a fix for #3794.

## Linked issues

Fixes #3140.
Fixes #3799.
Will eventually fix  #3794.


.github/workflows/create-release.yml
```diff
@@ -59,27 +50,64 @@ jobs:
...
+        path: requirements-freeze.txt
+
+  test-release:
+    # The purpose of this step is to run tests using the requirements-freeze file
+    # In 99% of cases, this will be identical to running tests using `requirements.txt`. But, sometimes
+    # there are unique bugs that arise from using the `requirements-freeze.txt` file.
+    needs: generate-requirements-freeze
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15, windows-2019]
```
joshuacwnewton, line 66: Huzzah! The [`windows-2019`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/7286144257?check_suite_focus=true) workflow did in fact fail with the following exception:

```
Check if msvc-runtime is installed..................[FAIL]
An error occured while importing module msvc-runtime -> No module named 'msvc_runtime'
```

We can now rework the `generate-requirements-freeze` step to properly copy over the correct lines from `requirements.txt` to fix #3794.

.github/workflows/create-release.yml
```diff
@@ -58,28 +49,66 @@ jobs:
         conda deactivate
         echo "# Platform-specific torch requirements (See SCT Issue #2745)" >> requirements-freeze.txt
         grep "torch" requirements.txt >> requirements-freeze.txt
```
mguaypaq, line 51: Maybe out of scope, but I'd be curious to see how `grep "torch"` interacts with the recent changes in #3790 (which remove the lines that used to begin with "torch", and adds an "--extra-index-url" line which happens to contain the string "torch" in its url).

mguaypaq (approved):
This is nice! And I'm learning about `upload-artifact` and `download-artifact`, seems convenient.

requirements.txt
```diff
@@ -9,7 +9,7 @@ matplotlib
 # we'd prefer not to burden users with another prerequisite. 
 # NB: It's possible this fix gets included upstream in a future skimage release, see:
 #     https://github.com/scikit-image/scikit-image/issues/4780#issuecomment-824678872
-msvc-runtime; sys.platform == 'win32'
+msvc-runtime==14.29.30133; sys.platform == 'win32'
```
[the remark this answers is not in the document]
joshuacwnewton: > but it's nice being able to rely on the fact that requirements-freeze.txt includes a specific version for each package listed.

That was my thinking, too! 

It felt odd to have every package have a strictly pinned version #, yet have this one library stay unpinned. Not that I would expect anything to break because of it, but having a strictly pinned version would give me peace of mind for SCT's stable releases. :)

.github/workflows/create-release.yml
joshuacwnewton: Thank you for keeping an eye on this!

Thankfully, it works as we'd want it to:

```console
joshua@tadpole:~/repos/spinalcordtoolbox$ grep "torch" requirements.txt
--extra-index-url https://download.pytorch.org/whl/cpu
```

But, that almost feels by luck, rather than by design? :P 

Grepping within `requirements.txt` files feels so odd...

joshuacwnewton: One other nice bonus of using the `upload-artifact` and `download-artifact` actions is that it actually creates file download [within the workflow itself](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/2651333502):

image

Downloading that file shows us the same result:

image

joshuacwnewton: Ah, but the echo'd comment *is* out of date -- we're no longer concerned with platform-specific torch requirements.

mguaypaq: I'm glad it's intentional! I did get the feeling that luck might be involved, hence the question.

I think grepping through config files is a time-honored tradition, although there are ways to make it more robust. Purely as an example, since [pip supports end-of-line comments](https://pip.pypa.io/en/latest/reference/requirements-file-format/#comments), we could instead grep for a token like `append_to_freeze` that we would put in an end-of-line comment on the lines we want.

.github/workflows/create-release.yml
```diff
@@ -29,57 +29,87 @@ env:
...
     - name: Update requirements-freeze.txt
       run: |
         source python/etc/profile.d/conda.sh
         conda activate venv_sct
         pip freeze | grep -v "-e git+" | grep -v "torch" | grep -v "certifi @ file" > requirements-freeze.txt
         conda deactivate
-        echo "# Platform-specific torch requirements (See SCT Issue #2745)" >> requirements-freeze.txt
+        echo "# Prioritize CPU version of torch (See https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3790#discussion_r888573886)" >> requirements-freeze.txt
         grep "torch" requirements.txt >> requirements-freeze.txt
+        echo "# Preserve platform-specific requirements missed by `pip freeze`" >> requirements-freeze.txt
```
joshuacwnewton, line 52: Oopsie... Using backticks to specify code/commands is a habit of mine with Markdown. 

But, here, the backticks evaluated `pip freeze` and stuck the output into the command. :rofl: 

image

```suggestion
        echo "# Preserve platform-specific requirements missed by 'pip freeze'" >> requirements-freeze.txt
```

.github/workflows/create-release.yml
joshuacwnewton: > Purely as an example, since [pip supports end-of-line comments](https://pip.pypa.io/en/latest/reference/requirements-file-format/#comments), we could instead grep for a token like append_to_freeze that we would put in an end-of-line comment on the lines we want.

Ah! I absolutely love this idea! There are all sorts of benefits here:

* `create-release.yml` only needs a single grep pattern (i.e. it doesn't need to be continuously updated)
* So, any future changes would only occur within the `requirements.txt` file, which is where we're already making our changes, giving us a single location to focus on (preventing us from accidentally missing the workflow file when making changes to the requirements file)
* The pattern can be chosen to be unique enough to avoid accidental/lucky pattern-matching

PR #3837, opened 2022-07-11
Remove coverage via `codecov`

mguaypaq:
This was added in #3463 about six months ago, but (according to an informal poll on Slack) none of the current developers have been paying attention to the resulting bot-generated comments on all the pull requests, so it's just noise.


joshuacwnewton (approved):
The present changes LGTM! Pretty much a 1:1 reversal of #3463. :) 

But, I do want to point out that we also have a `setup.cfg` that contains:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/820d375c4abcfa8418881d54b04a0763c3f7e3a6/setup.cfg#L13-L18

This must have been added prior to #3463? These settings are more general, applying to the [Coverage.py](https://coverage.readthedocs.io/en/6.4.1/config.html) library rather than Codecov specifically (meaning it might be relevant for folks who want to run coverage locally).

That said, maybe we're better off letting devs handle their own configuration locally (via, say, a [`.coveragerc`](https://coverage.readthedocs.io/en/6.4.1/config.html) file they add to their local directory). 

My thinking is: the only way coverage will ever get utilised is if someone feels strongly enough to take up the mantle on their own, and in that case, they might be opinionated/experienced enough to handle the configuration on their own, too?

So, maybe we remove these `setup.cfg` lines as well... :thinking:

mguaypaq:
Done! Yes, these lines seem to be from 2017, for a different code coverage tool. You're right, we might as well let individual developers set their own configuration for tools they run locally.

PR #3842, opened 2022-07-13
Reduce flake8 warnings throughout the codebase

mguaypaq:
Sorry for the large PR. I tried to at least split it up into commits thematically. If it's too annoying to have all the whitespace fixes together (rather than having them fixed as we touch the code for other issues), I can rebase to remove that commit.

With this PR, the number of warnings reported by `flake8` on the `spinalcordtoolbox/` folder goes down from 367 to 30. Having looked at them, the remaining warnings will probably require a bit more thought, so I don't think they should be part of a bulk PR like this. (This is kind of my goal for this PR: removing all the low-effort noise from the list of warnings, so that the remaining warnings are worth looking at.)

Also, I put 5.7 as the milestone by default, but we can also wait until the next release if we don't want such a big change so close to the release.


spinalcordtoolbox/utils/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .fs import *
-from .sys import *
-from .sys import __version__, __sct_dir__, __data_dir__, __deepseg_dir__
-from .shell import *
+from .fs import *  # noqa: F401, F403
+from .sys import *  # noqa: F401, F403
+from .sys import __version__, __sct_dir__, __data_dir__, __deepseg_dir__  # noqa: F401
+from .shell import *  # noqa: F401, F403
```
joshuacwnewton, line 4: I was curious if there was an alternate pattern we could use here that doesn't require the use of `noqa` comments, since I figure if `noqa` comments are necessary, then perhaps we're doing something in a non-standard way?

The closest I got to a useful suggestion was [proposing the use of `__all__`](https://stackoverflow.com/a/31079085) to define the public API for a module. But, `__all__` seems intended for trying to [distinguish between public and private interfaces](https://peps.python.org/pep-0008/#public-and-internal-interfaces) in a module? Whereas here, we're just trying to alias submodules to `import utils`.

Given that we've talked about not necessarily being perfectly Pythonic when it comes to API design (as we don't intend to expose our API to downstream developers), I think going with the `noqa` comments is the most sensible thing to do, especially given the scope of the PR.

spinalcordtoolbox/process_seg.py
```diff
@@ -193,7 +193,7 @@ def _properties2d(image, dim):
                   'eccentricity': region.eccentricity,
                   'orientation': orientation,
                   'solidity': solidity  # convexity measure
-    }
+                  }
```
joshuacwnewton, line 197: Nit, but perhaps this might be better off as:

```python
   # Fill up dictionary
    properties = {
        'area': area,
        'diameter_AP': diameter_AP,
        'diameter_RL': diameter_RL,
        'centroid': region.centroid,
        'eccentricity': region.eccentricity,
        'orientation': orientation,
        'solidity': solidity  # convexity measure
    }

File: spinalcordtoolbox/registration/algorithms.py [hunk 3]
```diff
@@ -1275,16 +1275,16 @@ def register2d(fname_src, fname_dest, fname_mask='', fname_warp='warp_forward.ni
...
+               '--dimensionality', '2',
+               '--transform', paramreg.algo + '[' + str(paramreg.gradStep) + ants_registration_params[paramreg.algo.lower()] + ']',
+               '--metric', paramreg.metric + '[dest_Z' + num + '.nii' + ',src_Z' + num + '.nii' + ',1,' + metricSize + ']',  # [fixedImage,movingImage,metricWeight +nb_of_bins (MI) or radius (other)
+               '--convergence', str(paramreg.iter),
+               '--shrink-factors', str(paramreg.shrink),
+               '--smoothing-sigmas', str(paramreg.smooth) + 'mm',
+               '--output', '[' + prefix_warp2d + ',src_Z' + num + '_reg.nii]',    # --> file.mat (contains Tx,Ty, theta)
+               '--interpolation', 'BSpline[3]',
+               '--verbose', '1',
+               ] + masking
```
joshuacwnewton, line 1287: Similar nitpick for this one:

```python
        cmd = [
            'isct_antsRegistration',
            '--dimensionality', '2',
            '--transform', paramreg.algo + '[' + str(paramreg.gradStep) + ants_registration_params[paramreg.algo.lower()] + ']',
            '--metric', paramreg.metric + '[dest_Z' + num + '.nii' + ',src_Z' + num + '.nii' + ',1,' + metricSize + ']',  # [fixedImage,movingImage,metricWeight +nb_of_bins (MI) or radius (other)
            '--convergence', str(paramreg.iter),
            '--shrink-factors', str(paramreg.shrink),
            '--smoothing-sigmas', str(paramreg.smooth) + 'mm',
            '--output', '[' + prefix_warp2d + ',src_Z' + num + '_reg.nii]',  # --> file.mat (contains Tx,Ty, theta)
            '--interpolation', 'BSpline[3]',
            '--verbose', '1',
        ] + masking
```

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1313,14 +1313,14 @@ def register2d(fname_src, fname_dest, fname_mask='', fname_warp='warp_forward.ni
...
-                 '-s', '0',
-                ], is_sct_binary=True)
+                          '-d', '2',
+                          '-t', 'SyN[1,1,1]',
+                          '-c', '0',
+                          '-m', 'MI[dest_Z' + num + '.nii,src_Z' + num + '.nii,1,32]',
+                          '-o', 'warp2d_null',
+                          '-f', '1',
+                          '-s', '0',
+                          ], is_sct_binary=True)
```
joshuacwnewton, line 1323: Ditto again here:

```python
                run_proc([
                    'isct_antsRegistration',
                    '-d', '2',
                    '-t', 'SyN[1,1,1]',
                    '-c', '0',
                    '-m', 'MI[dest_Z' + num + '.nii,src_Z' + num + '.nii,1,32]',
                    '-o', 'warp2d_null',
                    '-f', '1',
                    '-s', '0',
                 ], is_sct_binary=True)
```

spinalcordtoolbox/scripts/sct_compute_mtsat.py
```diff
@@ -199,10 +199,10 @@ def main(argv=None):
...
 
     display_viewer_syntax([arguments.omtsat, arguments.ot1map],
-                              colormaps=['gray', 'gray'],
-                              minmax=['-10,10', '0, 3'],
-                              opacities=['1', '1'],
-                              verbose=verbose)
+                          colormaps=['gray', 'gray'],
+                          minmax=['-10,10', '0, 3'],
+                          opacities=['1', '1'],
+                          verbose=verbose)
```
joshuacwnewton, line 205: Ditto:

```python
    display_viewer_syntax(
        [arguments.omtsat, arguments.ot1map],
        colormaps=['gray', 'gray'],
        minmax=['-10,10', '0, 3'],
        opacities=['1', '1'],
        verbose=verbose
    )
```

spinalcordtoolbox/scripts/sct_deepseg_gm.py
```diff
@@ -125,12 +125,11 @@ def main(argv=None):
                     dataset=qc_dataset, subject=qc_subject, process='sct_deepseg_gm')
 
     display_viewer_syntax([input_filename, format(out_fname)],
-                              colormaps=['gray', 'red'],
-                              opacities=['1', '0.7'],
-                              verbose=verbose)
+                          colormaps=['gray', 'red'],
+                          opacities=['1', '0.7'],
+                          verbose=verbose)
```
joshuacwnewton, line 130:
```python
    display_viewer_syntax(
        [input_filename, format(out_fname)],
        colormaps=['gray', 'red'],
        opacities=['1', '0.7'],
        verbose=verbose
    )

File: spinalcordtoolbox/aggregate_slicewise.py [hunk 7]
```diff
@@ -326,7 +326,7 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], distan
                 else:
                     mask_slicegroup = np.ones(data_slicegroup.shape)
                 # Ignore nonfinite values
-                i_nonfinite = np.where(np.isfinite(data_slicegroup) == False)
+                i_nonfinite = np.where(np.isfinite(data_slicegroup) == False)  # noqa: E712
```
joshuacwnewton, line 329: This seems reasonable, as per https://github.com/PyCQA/pycodestyle/issues/450#issuecomment-390169995!

spinalcordtoolbox/types.py
```diff
@@ -771,7 +771,7 @@ def display(self, mode='absolute'):
         # ax.set_aspect('equal')
         plt.xlabel('z')
         plt.ylabel('x')
-        ax = plt.subplot(212)
+        ax = plt.subplot(212)  # noqa: F841
```
joshuacwnewton, line 774: Is the reason these got `noqa` comments because of the commented out `# ax.set_aspect('equal')`? (i.e. wanting to allow that line to be uncommented without throwing an error?)

joshuacwnewton (approved):
As you've said, all of the warnings/errors you've fixed are straightforward and easy to repair in bulk!

I had a few observations and/or nitpicks, but for the most part, this LGTM! The nitpicks are tiny and subjective, too, so either way, feel free to merge once you're satisfied. :)

spinalcordtoolbox/utils/__init__.py
mguaypaq: I thought about `__all__` as well, but (apart from any further subtleties) we don't have an explicit list to give it, since we're importing `*`. We're definitely not being pythonic here. Possibly I should just open an issue to review who imports what from where and get rid of this `__init__.py` file altogether.

PR #3858, opened 2022-08-16
Call `parser.error` properly

mguaypaq:
This PR removes the `raise parser.error(...)` anti-pattern from the codebase.

The function `parser.error()` from `argparse` never returns with a value; it either exits the program, or raises an exception. There's no need to raise the non-existent return value.

All other `raise` statements look like they raise actual exceptions (maybe not always the most appropriate one, but at least an exception). While checking this, I couldn't ignore two cases of `raise Exception(TypeError)`, so I fixed those as well.

### DIFF
--- a/spinalcordtoolbox/resampling.py
+++ b/spinalcordtoolbox/resampling.py
@@ -50,7 +50,7 @@ def resample_nib(image, new_size=None, new_size_type=None, image_dest=None, inte
     elif type(image) == Image:
         img = nib.nifti1.Nifti1Image(image.data, image.hdr.get_best_affine())
     else:
-        raise Exception(TypeError)
+        raise TypeError(f'Invalid image type: {type(image)}')
 
     if image_dest is None:
         # Get dimensions of data
@@ -108,7 +108,7 @@ def resample_nib(image, new_size=None, new_size_type=None, image_dest=None, inte
         elif type(image_dest) == Image:
             reference = nib.nifti1.Nifti1Image(image_dest.data, image_dest.hdr.get_best_affine())
         else:
-            raise Exception(TypeError)
+            raise TypeError(f'Invalid image type: {type(image_dest)}')
 
     if img.ndim == 3:
         # we use mode 'nearest' to overcome issue #2453
--- a/spinalcordtoolbox/scripts/sct_compute_snr.py
+++ b/spinalcordtoolbox/scripts/sct_compute_snr.py
@@ -149,7 +149,7 @@ def main(argv=None):
     # Check parameters
     if method in ['diff', 'single']:
         if not fname_mask:
-            raise parser.error(f"Argument '-m' must be specified when using '-method {method}'.")
+            parser.error(f"Argument '-m' must be specified when using '-method {method}'.")
 
     # Load data
     im_data = Image(fname_data)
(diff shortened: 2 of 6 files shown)


spinalcordtoolbox/scripts/sct_compute_snr.py
```diff
@@ -149,7 +149,7 @@ def main(argv=None):
     # Check parameters
     if method in ['diff', 'single']:
         if not fname_mask:
-            raise parser.error(f"Argument '-m' must be specified when using '-method {method}'.")
+            parser.error(f"Argument '-m' must be specified when using '-method {method}'.")
```
joshuacwnewton, line 152: This makes sense to me!

> [`raise`](https://docs.python.org/3/tutorial/errors.html#raising-exceptions)
>
> The [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) statement allows the programmer to force a specified exception to occur. The sole argument to [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) indicates the exception to be raised. **This must be either an exception instance or an exception class (a class that derives from [Exception](https://docs.python.org/3/library/exceptions.html#Exception)).**

> [`ArgumentParser.error(message)`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error)
>
> This method prints a usage message including the message to the standard error and terminates the program with a status code of 2.

Just for fun, I tried using `raise` with a dummy function and got this:

```
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/code.py", line 90, in runcode
    exec(code, self.locals)
TypeError: exceptions must derive from BaseException
```

I was curious why `raise parser.error` doesn't throw the same error, but then I realized... `raise parser.error` _doesn't even get to `raise`_; the program `exit`s before the `raise` statement can be invoked. :facepalm: 

Definitely unnecessary, then. :slightly_smiling_face:

spinalcordtoolbox/resampling.py
```diff
@@ -50,7 +50,7 @@ def resample_nib(image, new_size=None, new_size_type=None, image_dest=None, inte
     elif type(image) == Image:
         img = nib.nifti1.Nifti1Image(image.data, image.hdr.get_best_affine())
     else:
-        raise Exception(TypeError)
+        raise TypeError(f'Invalid image type: {type(image)}')
```
joshuacwnewton, line 53: I'm curious if `try`/`except TypeError` even catches `Exception(TypeError)`?

Bit besides the point, though, since we're replacing it with the proper syntax, anyway. :P

joshuacwnewton (approved):

spinalcordtoolbox/scripts/sct_compute_snr.py
mguaypaq: Exactly!

And FYI, you can get the same result with `raise None`, because functions that don't return anything explicitly (such as your dummy function) actually return `None`.

(Anecdotally, I'm guilty of writing the opposite bug at least once: creating an Exception object without raising it. Python will happily run the statement, create the exception object, then discard it, as if discarding an unused return value of a function call.)

spinalcordtoolbox/resampling.py
mguaypaq: Nope: `except TypeError` catches any exception which inherits from TypeError, and the raised exception is an Exception (which happens to contain the class object `TypeError` as one of its fields).
```py3
>>> try: raise Exception(TypeError)
... except TypeError: print('yes')
... else: print('nope')
... 
Exception: <class 'TypeError'>
```

PR #3859, opened 2022-08-16
Modify `-include-list` and `-exclude-list` to check against parts of a directory, too

### LINKED ISSUE #3598: `include` and `include_list` function differently for `ses` folders
Context here: https://forum.spinalcordmri.org/t/sct-run-batch-include-list/770/3?u=joshuacwnewton

> I have tried your solution:
>
> ```
> ("include_list" : ["sub-0001", "sub-0002"]) 
> ```
> 
> but it didn't work. Since our dataset is longitudinal (organised in this way: sub-0001/ses-01/anat, …) I then tried to include also ses into the list:
> 
> ```
> "include_list" : ["sub-0001/ses-01","sub-0002/ses-01"]
> ```
> 
> In this configuration, it works. So, it seems that include_list required ses for longitudinal studies. 
> 
> On the other hand, `include` accepts both variants (with ses:  `"include" : "sub-0001/ses-01"` and also without ses:  `"include" : "sub-0001"`)

So, for datasets organized with `ses` folders, it seems that you can't just specify `include_list: sub-0001` -- you have to specify the `ses` part of the path, too.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR:

* Does a bit of refactoring to make the filtering behavior a little easier to test.
* Adds a new test that fails for `-include-list` and `-exclude-list`.
* Updates the behavior of `-include-list` and `-exclude-list` to check against the sub-parts of a directory (e.g. `sub-01` within `sub-01/ses-01`)

## Linked issues

Fixes #3598.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -184,7 +184,7 @@ def _filter_directories(dir_list, include=None, include_list=None, exclude=None,
     """
     # Handle inclusions (regex OR explicit list, but not both)
     if include is not None:
-        dir_list = [f for f in dir_list if re.search(include, f) is not None]
+        dir_list = [f for f in dir_list if re.search(include, f)]
     elif include_list is not None:
         dir_list = [f for f in dir_list
                     # Check if include_list specified entire path (e.g. "sub-01/ses-01")
@@ -194,7 +194,7 @@ def _filter_directories(dir_list, include=None, include_list=None, exclude=None,
 
     # Handle exclusions (regex OR explicit list, but not both)
     if exclude is not None:
-        dir_list = [f for f in dir_list if re.search(exclude, f) is None]
+        dir_list = [f for f in dir_list if not re.search(exclude, f)]
     elif exclude_list is not None:
         dir_list = [f for f in dir_list
                     # Check if exclude_list specified entire path (e.g. "sub-01/ses-01")
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -176,22 +176,26 @@ def _find_nonsys32_bash_exe():
 
 
 def _filter_directories(dir_list, include=None, include_list=None, exclude=None, exclude_list=None):
-    # Handle inclusion lists
+    """
+    Filter a list of directories using inclusion/exclusion regex patterns or explicit lists.
+
+    NB: Only one of [include, include_list] and only one of [exclude, exclude_list] should be passed.
+        (Currently, this requirement is handled at the argument-parsing level, because we use `parser.error`.)
+    """
+    # Handle inclusions (regex OR explicit list, but not both)
     if include is not None:
         dir_list = [f for f in dir_list if re.search(include, f) is not None]
-
-    if include_list is not None:
+    elif include_list is not None:
         dir_list = [f for f in dir_list
                     # Check if include_list specified entire path (e.g. "sub-01/ses-01")
                     if f in include_list
                     # Check if include_list specified a subdirectory (e.g. just "sub-01" or just "ses-01")
                     or any([p in include_list for p in pathlib.Path(f).parts])]
 
-    # Handle exclusions
+    # Handle exclusions (regex OR explicit list, but not both)
     if exclude is not None:
         dir_list = [f for f in dir_list if re.search(exclude, f) is None]
-
-    if exclude_list is not None:
+    elif exclude_list is not None:
         dir_list = [f for f in dir_list
                     # Check if exclude_list specified entire path (e.g. "sub-01/ses-01")
                     if f not in exclude_list
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -174,6 +174,31 @@ def _find_nonsys32_bash_exe():
     return shutil.which('bash', path=nonsys32_paths)
 
 
+def _filter_directories(dir_list, include=None, include_list=None, exclude=None, exclude_list=None):
+    # Handle inclusion lists
+    assert not ((include is not None) and (include_list is not None)),\
+        'Only one of `include` and `include-list` can be used'
+
+    if include is not None:
+        dir_list = [f for f in dir_list if re.search(include, f) is not None]
+
+    if include_list is not None:
+        # TODO decide if we should warn users if one of their inclusions isn't around
+        dir_list = [f for f in dir_list if f in include_list]
+
+    # Handle exclusions
+    assert not ((exclude is not None) and (exclude_list is not None)), \
+        'Only one of `exclude` and `exclude-list` can be used'
+
+    if exclude is not None:
+        dir_list = [f for f in dir_list if re.search(exclude, f) is None]
+
+    if exclude_list is not None:
+        dir_list = [f for f in dir_list if f not in exclude_list]
+
+    return dir_list
+
+
 def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_data_processed, path_results, path_log,
                path_qc, itk_threads, continue_on_error=False):
     """
@@ -446,26 +471,9 @@ def main(argv=None):
                 for isess in session_dirs:
                     subject_dirs.append(os.path.join(isub, isess))
 
-    # Handle inclusion lists
-    assert not ((arguments.include is not None) and (arguments.include_list is not None)),\
-        'Only one of `include` and `include-list` can be used'
-
-    if arguments.include is not None:
-        subject_dirs = [f for f in subject_dirs if re.search(arguments.include, f) is not None]
-
-    if arguments.include_list is not None:
-        # TODO decide if we should warn users if one of their inclusions isn't around
-        subject_dirs = [f for f in subject_dirs if f in arguments.include_list]
-
-    # Handle exclusions
-    assert not ((arguments.exclude is not None) and (arguments.exclude_list is not None)),\
-        'Only one of `exclude` and `exclude-list` can be used'
-
-    if arguments.exclude is not None:
-        subject_dirs = [f for f in subject_dirs if re.search(arguments.exclude, f) is None]
-
-    if arguments.exclude_list is not None:
-        subject_dirs = [f for f in subject_dirs if f not in arguments.exclude_list]
+    subject_dirs = _filter_directories(subject_dirs,
+                                       arguments.include, arguments.exclude,
+                                       arguments.include_list, arguments.exclude_list)
 
     # Determine the number of jobs we can run simultaneously
     if arguments.jobs < 1:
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -174,6 +175,33 @@ def _find_nonsys32_bash_exe():
     return shutil.which('bash', path=nonsys32_paths)
 
 
+def _filter_directories(dir_list, include=None, include_list=None, exclude=None, exclude_list=None):
+    # Handle inclusion lists
+    if include is not None:
+        dir_list = [f for f in dir_list if re.search(include, f) is not None]
+
+    if include_list is not None:
+        # TODO decide if we should warn users if one of their inclusions isn't around
```
joshuacwnewton, line 184: We might want to address this existing `# TODO` here?

----

My vote would be to remove the TODO without warning...

I'm not sure how much a warning will help, because if it's just a warning, then the processing will continue as normal, and I suspect that the user won't catch it due to the text that will be printed underneath it.

In other words, it seems like the user will be just as likely to notice the error without the message, once they realize that X subject wasn't processed.

testing/cli/test_sct_run_batch.py
```diff
@@ -87,6 +87,39 @@ def test_only_one_include(tmp_path, dummy_script):
                             '-path-data', str(data), '-path-out', str(out), '-script', dummy_script])
 
 
+def test_directory_inclusion_exclusion():
+    """
+    Test that -include, -include-list, -exclude, and -exclude-list properly filter directories.
```
joshuacwnewton, line 92: I was unsure if writing this test was worth the effort. It looks a little silly, and the corresponding function felt a little silly to write, too.

Is this overkill? It's hard to sense what's worth writing a test for and what's not...

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -174,6 +175,33 @@ def _find_nonsys32_bash_exe():
     return shutil.which('bash', path=nonsys32_paths)
 
 
+def _filter_directories(dir_list, include=None, include_list=None, exclude=None, exclude_list=None):
+    # Handle inclusion lists
+    if include is not None:
+        dir_list = [f for f in dir_list if re.search(include, f) is not None]
+
+    if include_list is not None:
```
mguaypaq, line 183: I wonder if this should be an `elif`? Or better yet, either a function docstring or a code comment should remind us close to here of the expectation that at most one of `include`, `include_list` is used, and at most one of `exclude`, `exclude_list` is used.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -174,6 +175,33 @@ def _find_nonsys32_bash_exe():
     return shutil.which('bash', path=nonsys32_paths)
 
 
+def _filter_directories(dir_list, include=None, include_list=None, exclude=None, exclude_list=None):
+    # Handle inclusion lists
+    if include is not None:
+        dir_list = [f for f in dir_list if re.search(include, f) is not None]
```
mguaypaq, line 181: The result of `re.search` is guaranteed to be truthy if it's not None, so you can drop the `is not None` if you want.

spinalcordtoolbox/scripts/sct_run_batch.py
mguaypaq: I agree, an outright error would be upsetting, and a warning would get lost in the sea of output. And the argument descriptions are explicitly worded to indicate that these arguments are filters.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -174,6 +175,33 @@ def _find_nonsys32_bash_exe():
...
+
+    if include_list is not None:
+        # TODO decide if we should warn users if one of their inclusions isn't around
+        dir_list = [f for f in dir_list
+                    # Check if include_list specified entire path (e.g. "sub-01/ses-01")
+                    if f in include_list
+                    # Check if include_list specified a subdirectory (e.g. just "sub-01" or just "ses-01")
+                    or any([p in include_list for p in pathlib.Path(f).parts])]
```
mguaypaq, line 189: I like this list comprehension! And the use of `pathlib.Path(f).parts`, very clear. Nitpick: you can get rid of the inner list comprehension, and just write `any(p in include_list for ...)`.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -174,6 +175,33 @@ def _find_nonsys32_bash_exe():
...
+    # Handle exclusions
+    if exclude is not None:
+        dir_list = [f for f in dir_list if re.search(exclude, f) is None]
+
+    if exclude_list is not None:
+        dir_list = [f for f in dir_list
+                    # Check if exclude_list specified entire path (e.g. "sub-01/ses-01")
+                    if f not in exclude_list
+                    # Check if exclude_list specified a subdirectory (e.g. just "sub-01" or just "ses-01")
+                    and all([p not in exclude_list for p in pathlib.Path(f).parts])]
```
mguaypaq, line 200: Ditto about dropping the square brackets for the inner list comprehension.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -446,26 +474,15 @@ def reset_streams():
...
-
-    if arguments.include_list is not None:
-        # TODO decide if we should warn users if one of their inclusions isn't around
-        subject_dirs = [f for f in subject_dirs if f in arguments.include_list]
-
-    # Handle exclusions
-    assert not ((arguments.exclude is not None) and (arguments.exclude_list is not None)),\
-        'Only one of `exclude` and `exclude-list` can be used'
+    if (arguments.include is not None) and (arguments.include_list is not None):
+        parser.error('Only one of `include` and `include-list` can be used')
```
mguaypaq, line 478: Thanks for turning this into a proper error rather than an `assert`. I've always liked the description in [this style guide](https://google.github.io/styleguide/pyguide.html#244-decision), that asserts should only be used for internal invariants (because they [may be optimized out](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)).

testing/cli/test_sct_run_batch.py
mguaypaq: Since this PR originates from a user request on a forum post, I think it would be good to test that specific scenario in this test, possibly with a link back to the issue in a code comment, so that we avoid future regressions. (Future user requests could turn into added cases.)

For the other test cases, they do look like overkill. This is a self-contained function, so as long as the code is reasonable and it's been tested at least once, it shouldn't suddenly break because of unrelated code changes.

mguaypaq (approved):
Thanks for the clean fix!

spinalcordtoolbox/scripts/sct_run_batch.py
joshuacwnewton: I like this suggestion. If we're enforcing mutual exclusion at the function level, it makes me want to put the exception checks there, too. But, then we lose easy access to `parser.error`... :thinking:

spinalcordtoolbox/scripts/sct_run_batch.py
joshuacwnewton: > I've always liked the description in [this style guide](https://google.github.io/styleguide/pyguide.html#244-decision),

Aside: Thank you for linking this style guide! It's interesting to see the internal perspectives of a mega-corporation like Google. There's a whole lot to unpack, but it seems like a valuable resource. :)

> (because they [may be optimized out](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)).

Huh. I'd never heard of [`-O`](https://docs.python.org/3/using/cmdline.html#cmdoption-O) before.

> Remove assert statements and any code conditional on the value of [`__debug__`](https://docs.python.org/3/library/constants.html#debug__). Augment the filename for compiled ([bytecode](https://docs.python.org/3/glossary.html#term-bytecode)) files by adding .opt-1 before the .pyc extension (see [PEP 488](https://www.python.org/dev/peps/pep-0488)). See also [PYTHONOPTIMIZE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE).

I wonder how this would be used in practice... :thinking:

spinalcordtoolbox/scripts/sct_run_batch.py
mguaypaq: Ha, you can also do both if you want: use `parser.error` in the parsing code, then use an `assert` in the function, because then it would be an internal invariant. But that's optional: your current comments are clear enough.

spinalcordtoolbox/scripts/sct_run_batch.py
mguaypaq: I think it's pretty rare to use `-O` for Python code, so `assert` statements are very likely to run in practice. This is definitely a tiny, technical, theoretical detail, essentially trivia.

mguaypaq (approved):

valosekj:
Hi @joshuacwnewton!

I wanted to try the changes proposed in this PR. But I am struggling with how exactly to specify the `include-list`.

I updated my SCT with this PR:

```
git checkout master
git pull
```

My dataset looks like this:
```
.
├── sub-0001
│   ├── ses-01
│   │   ├── anat
│   │   │   ├── sub-0001_ses-01_T2w.json
│   │   │   └── sub-0001_ses-01_T2w.nii.gz
│   │   └── dwi
│   │       ├── ...
│   └── ses-02
│       ├── anat
│       │   ├── sub-0001_ses-02_T2w.json
│       │   └── sub-0001_ses-02_T2w.nii.gz
│       └── dwi
│           ├── ...
├── sub-0002
...
```

When I run `sct_run_batch -c ~/config.json` with the `include` flag, everything works fine, and both sessions for `sub-0001` are processed.

```
{
  "path_data"   : "./bids",
  "path_output" : "./results",
  "script"      : "~/process_data.sh",
  "include"     : "0001",
  "jobs"        : -1
}
```

But, if try `config.json` with the `include_list` flag, the `sct_run_batch` finishes with error.

```
{
  "path_data"   : "./bids",
  "path_output" : "./results",
  "script"      : "~/process_data.sh",
  "include_list": ["0001","0002"],
  "jobs"        : -1
}
```

Based on [your comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3598#issuecomment-1216681299), I also tried to specify a precise subject/session combination. But I got the same error.

```
{
  "path_data"   : "./bids",
  "path_output" : "./results",
  "script"      : "~/process_data.sh",
  "include_list": ["sub-0001/ses-01","sub-0001/ses-02","sub-0002/ses-01","sub-0002/ses-02"],
  "jobs"        : -1
}
```

Error:

```
  File "/usr/local/lib/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 598, in <module>
    main(sys.argv[1:])
  File "/usr/local/lib/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 509, in main
    arguments.include_list, arguments.exclude_list)
  File "/usr/local/lib/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 232, in _filter_directories
    dir_list = [f for f in dir_list if not re.search(exclude, f)]
  File "/usr/local/lib/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 232, in <listcomp>
    dir_list = [f for f in dir_list if not re.search(exclude, f)]
  File "/usr/local/lib/sct_latest/python/envs/venv_sct/lib/python3.7/re.py", line 185, in search
    return _compile(pattern, flags).search(string)
  File "/usr/local/lib/sct_latest/python/envs/venv_sct/lib/python3.7/re.py", line 278, in _compile
    return _cache[type(pattern), pattern, flags]
TypeError: unhashable type: 'list' 
```

PR #3862, opened 2022-08-17
Add `-ignore-ses` flag to process `sub-` directories even when `ses-` subdirectories are present

### LINKED ISSUE #3809: Make it possible to loop across `sub-*` without considering `ses-*` if present
As of SCT v5.4, whenever a `ses-` folder is present under a `sub-*` folder, it will consider it and aggregate subject and session, eg: `sub-01/ses-01` and use this instance inside the script being run.

A problem could occur if a script is supposed to act at the subject levels instead of at the session level. For example, if a script is responsible for registering data between session 1 and session 2, there is no way of doing it with the current implementation of how sessions are dealt with.

Note that this was not a problem in SCT v5.3.0 and before.

My suggestion would be to add a flag that would create a clause here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5caa7d93c67454ac5835d14590d1202195c6b147/spinalcordtoolbox/scripts/sct_run_batch.py#L433-L447

That would bypass this 'is there a session' finding, and force each subject to be treated as a single instance for the script. It would then be the responsibility of the script to dive into session folders appropriately.

joshuacwnewton:
#### GitHub

#### PR contents

## Description

This PR provides an option that allows users to process `sub-` directories on their own, even when `ses-` subdirectories are present.

Without `-ignore-ses`:

image

With `-ignore-ses`:

image

I've used `action='store_true'` here because it preserves the existing behavior, so there should be no changes unless users choose to opt-in.

## Linked issues

Fixes #3809.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -191,13 +191,13 @@ def _parse_dataset_directory(path_data, subject_prefix="sub-", ignore_ses=False)
           TODO: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3415
     """
     dirs = []
-    subject_dirs = sorted([f for f in os.listdir(path_data)
-                           if f.startswith(subject_prefix)
-                           and os.path.isdir(os.path.join(path_data, f))])
+    subject_dirs = sorted([d.name for d in os.scandir(path_data)
+                           if d.name.startswith(subject_prefix)
+                           and d.is_dir()])
     for sub_dir in subject_dirs:
-        session_dirs = sorted([f for f in os.listdir(os.path.join(path_data, sub_dir))
-                               if f.startswith('ses-')
-                               and os.path.isdir(os.path.join(path_data, sub_dir, f))])
+        session_dirs = sorted([d.name for d in os.scandir(os.path.join(path_data, sub_dir))
+                               if d.name.startswith('ses-')
+                               and d.is_dir()])
         if session_dirs and not ignore_ses:
             # There is a 'ses-' subdirectory AND arguments.ignore_ses = False, so we concatenate: e.g. sub-XX/ses-YY
             for sess_dir in session_dirs:
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -191,18 +191,20 @@ def _parse_dataset_directory(path_data, subject_prefix="sub-", ignore_ses=False)
           TODO: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3415
     """
     dirs = []
-    subject_dirs = sorted([f for f in os.listdir(path_data) if f.startswith(subject_prefix)])
+    subject_dirs = sorted([f for f in os.listdir(path_data)
+                           if f.startswith(subject_prefix)
+                           and os.path.isdir(os.path.join(path_data, f))])
     for sub_dir in subject_dirs:
-        # Only consider folders
-        if os.path.isdir(os.path.join(path_data, sub_dir)):
-            session_dirs = sorted([f for f in os.listdir(os.path.join(path_data, sub_dir)) if f.startswith('ses-')])
-            if session_dirs and not ignore_ses:
-                # There is a 'ses-' subdirectory AND arguments.ignore_ses = False, so we concatenate: e.g. sub-XX/ses-YY
-                for sess_dir in session_dirs:
-                    dirs.append(os.path.join(sub_dir, sess_dir))
-            else:
-                # Otherwise, consider only 'sub-' directories and don't include 'ses-' subdirectories: e.g. sub-XX
-                dirs.append(sub_dir)
+        session_dirs = sorted([f for f in os.listdir(os.path.join(path_data, sub_dir))
+                               if f.startswith('ses-')
+                               and os.path.isdir(os.path.join(path_data, sub_dir, f))])
+        if session_dirs and not ignore_ses:
+            # There is a 'ses-' subdirectory AND arguments.ignore_ses = False, so we concatenate: e.g. sub-XX/ses-YY
+            for sess_dir in session_dirs:
+                dirs.append(os.path.join(sub_dir, sess_dir))
+        else:
+            # Otherwise, consider only 'sub-' directories and don't include 'ses-' subdirectories: e.g. sub-XX
+            dirs.append(sub_dir)
 
     return dirs
 
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -191,14 +191,13 @@ def _parse_dataset_directory(path_data, subject_prefix="sub-", ignore_ses=False)
           TODO: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3415
     """
     dirs = []
-    subject_dirs = [f for f in os.listdir(path_data) if f.startswith(subject_prefix)]
+    subject_dirs = sorted([f for f in os.listdir(path_data) if f.startswith(subject_prefix)])
     for sub_dir in subject_dirs:
         # Only consider folders
         if os.path.isdir(os.path.join(path_data, sub_dir)):
-            session_dirs = [f for f in os.listdir(os.path.join(path_data, sub_dir)) if f.startswith('ses-')]
+            session_dirs = sorted([f for f in os.listdir(os.path.join(path_data, sub_dir)) if f.startswith('ses-')])
             if session_dirs and not ignore_ses:
                 # There is a 'ses-' subdirectory AND arguments.ignore_ses = False, so we concatenate: e.g. sub-XX/ses-YY
-                session_dirs.sort()
                 for sess_dir in session_dirs:
                     dirs.append(os.path.join(sub_dir, sess_dir))
             else:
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -123,6 +123,11 @@ def get_parser():
                         help='Optional space separated list of subjects to exclude. Only process '
                         'a subject if they are not on this list. Inclusions are processed before exclusions. '
                         'Cannot be used with either `exclude`.', nargs='+')
+    parser.add_argument('-ignore-ses', action='store_true',
```
mguaypaq, line 126: I try to write flags and conditionals in a "positive" direction, to avoid the inevitable double negations that come up (like "not ignore"). How about something like `-only-sub`? Then again, maybe users that run into problems with `ses` folders will scan the options for something about `ses` rather than something about `sub`.

Sorry for the bike-shedding :sweat:

mguaypaq (approved):
Would this be worth adding a test? From the description I see that you did test this manually, so maybe it's not required.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -437,14 +442,14 @@ def reset_streams():
         # Only consider folders
         if os.path.isdir(os.path.join(path_data, isub)):
             session_dirs = [f for f in os.listdir(os.path.join(path_data, isub)) if f.startswith('ses-')]
-            if not session_dirs:
-                # There is no session folder, so we consider only sub- directory: sub-XX
-                subject_dirs.append(isub)
-            else:
-                # There is a session folder, so we concatenate: sub-XX/ses-YY
+            if session_dirs and arguments.ignore_ses is not True:
```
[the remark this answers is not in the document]
joshuacwnewton: Thank you for this! I think I may have reflexively wrote `is not True` because it sounded better in terms of spoken English. (Even though, as you've pointed out, what sounds good in English isn't necessarily functionally correct. :stuck_out_tongue_winking_eye:)

joshuacwnewton:
> Would this be worth adding a test? From the description I see that you did test this manually, so maybe it's not required.

Ah! Thank you for reminding me. Yes, I think it would be good to do a similar thing as was done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3859, i.e. putting the filtering logic into its own function, then writing a test for said function.

joshuacwnewton:
While looking at the function that was created in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3862/commits/7af966781d3b2f98eeaec86f83778797ae0f2d79, I got an itch to rewrite the parsing logic (even though admittedly it's out of scope for this PR... :sweat_smile:).

I think that the changes I made in 2971227 make the logic a little easier to read and understand? And with the test added, I feel confident that everything works as it did before the rewrite. But, maybe I got a little ahead of myself. :sweat_smile:

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -175,6 +180,35 @@ def _find_nonsys32_bash_exe():
...
+    # Find subject directories within the root dataset folder
+    subject_dirs = sorted([f for f in os.listdir(path_data)
+                           if f.startswith(subject_prefix) and os.path.isdir(os.path.join(path_data, f))])
+
+    # Find session subdirectories within subject directories
+    session_dirs = []
+    for sub_dir in subject_dirs:
+        path_sub = os.path.join(path_data, sub_dir)
+        subject_session_dirs = sorted([os.path.join(sub_dir, f) for f in os.listdir(path_sub)
+                                       if f.startswith("ses-") and os.path.isdir(os.path.join(path_sub, f))])
```
mguaypaq, line 202: Thanks for adding the `isdir` check for `ses-` subfolders, which mirrors the check for `sub-` folders.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -175,6 +180,35 @@ def _find_nonsys32_bash_exe():
...
+    Parse a dataset directory to find subject directories (or session subdirectories, if present).
+
+    Notes:
+        - The dataset is assumed to be structured in a (somewhat) BIDS-compliant way, see:
+          https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-name-structure
+        - This function is a rudimentary version of the library PyBIDS: https://github.com/bids-standard/pybids.
+          TODO: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3415
+    """
+    # Find subject directories within the root dataset folder
+    subject_dirs = sorted([f for f in os.listdir(path_data)
```
mguaypaq, line 194: A drive-by suggestion: I noticed in the docs for `listdir()` that [`scandir()`](https://docs.python.org/3/library/os.html#os.scandir) is a good option when you also want file attributes (like `is_dir()`) while iterating through filenames. Also, the `os.DirEntry` objects it returns conveniently have both `name` (the base filename) and `path` (the full path including the filename) as attributes.

However, you can't sort a list of `os.DirEntry` objects without specifying a key, for example:
```py3
subject_dirs = [d for d in os.scandir(path_data)
                if d.name.startswith(subject_prefix)
                and d.is_dir()]
subject_dirs.sort(key=lambda d: d.name)
```
(Alternatively, wait until the end of the function to sort the list of full paths.)

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -175,6 +180,35 @@ def _find_nonsys32_bash_exe():
...
+        path_sub = os.path.join(path_data, sub_dir)
+        subject_session_dirs = sorted([os.path.join(sub_dir, f) for f in os.listdir(path_sub)
+                                       if f.startswith("ses-") and os.path.isdir(os.path.join(path_sub, f))])
+        session_dirs.extend(subject_session_dirs)
+
+    # Determine which directories to return
+    if session_dirs and not ignore_ses:
+        return session_dirs
+    else:
+        return subject_dirs
```
mguaypaq, line 209: Unfortunately, this is a significant behaviour change, and I believe it will affect existing datasets.

Specifically, if a dataset contains some `sub-` folders with no `ses-` subfolders, and other `sub-` folders with `ses-` subfolders, then:
* the old code returned a mix of `sub-` folders and `ses-` folders;
* the new code silently drops the `sub-` folders with no `ses-` subfolders.

I think you need something like a `result_dirs` variable which gets either an `append()` or an `extend()` for each `subject_dir`.

testing/cli/test_sct_run_batch.py
```diff
@@ -115,6 +115,29 @@ def test_directory_inclusion_exclusion():
...
+    """
+    Test that subject and session subdirectories are parsed correctly within a sample dataset structure.
+    """
+
+    parse_dir = sct_run_batch._parse_dataset_directory  # for brevity
+
+    subject_names = ['sub-001', 'sub-002', 'sub-003']
+    for sub in subject_names:
+        (tmp_path / sub).mkdir()
+        (tmp_path / f"{sub}.txt").touch()  # Add false positive files to ensure only directories are returned
```
mguaypaq, line 128: :heart: testing the `is_dir()` condition!

mguaypaq (changes requested):
Thanks for the refactor! I do think it's cleaner, but also that it introduces a bug (see comment below).

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -180,6 +180,34 @@ def _find_nonsys32_bash_exe():
     return shutil.which('bash', path=nonsys32_paths)
 
 
+def _parse_dataset_directory(path_data, subject_prefix="sub-", ignore_ses=False):
```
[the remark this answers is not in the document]
joshuacwnewton: Absolutely. 

In fact, if we were to go strictly by [the BIDS specification](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-name-structure), I'm not even sure we we would want to have a `subject_prefix` parameter in the first place. But. something tells me there are datasets out there that aren't 100% BIDS compliant that we've wanted to support. (Perhaps they have custom `subject` names, but then, we can safely assume that if someone has `sub-` directories, they're putting in the extra effort to be BIDS compliant?)

spinalcordtoolbox/scripts/sct_run_batch.py
joshuacwnewton: Ahhhh, you're right! Thank you so much for catching this. :heart:

joshuacwnewton: I've reverted the rewrite from [2971227](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3862/commits/297122754a6cc5ad2809460dc0a73e3f85cbb425), then re-commited the smaller, more atomic changes that had been included in [2971227](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3862/commits/297122754a6cc5ad2809460dc0a73e3f85cbb425).

spinalcordtoolbox/scripts/sct_run_batch.py
joshuacwnewton: > However, you can't sort a list of os.DirEntry objects without specifying a key

Once we do the `name.startswith` and `is_dir` checks, do we even need the entries to be `os.DirEntry` objects anymore? In other words, could we do something like this?

```python
subject_dirs = sorted([d.name for d in os.scandir(path_data)
                       if d.name.startswith(subject_prefix)
                       and d.is_dir()])
```

mguaypaq: And in fact, you could even save yourself some path concatenation later by using `d.path` instead of `d.name`, so:
```py3
subject_dirs = sorted(d.path for d in os.scandir(path_data)
                      if d.name.startswith(subject_prefix)
                      and d.is_dir())
```

mguaypaq: Oh, except that the function itself returns relative paths rather than absolute paths, so using `d.path` doesn't work, sorry!

mguaypaq (approved):
Looks good!

One last suggestion: can you add a test case with "mixed" `sub-` folders (that is, one with `ses-` and one with no `ses-` inside)? I feel like that's a very likely bug to crop up again in a future refactoring.

Issue #3759, opened 2022-04-12
CentOS 7 test failure for `sct_compute_snr` on `master` branch


joshuacwnewton:
* Most recent success: https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5992281364?check_suite_focus=true
    * This was only 1 hour prior, so the window for changes is very small.
* Failure: https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5978103780?check_suite_focus=true
* I've [restarted the build](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5993118832?check_suite_focus=true), and will follow-up if the failure repeats itself.

```
=================================== FAILURES ===================================
__________________ test_sct_compute_snr_single_3d_mask_object __________________
Traceback (most recent call last):
  File "/__w/spinalcordtoolbox/spinalcordtoolbox/testing/cli/test_cli_sct_compute_snr.py", line 267, in test_sct_compute_snr_single_3d_mask_object
    assert snr == pytest.approx(np.sqrt(2*SIGNAL_OBJECT**2), rel=0.05)
AssertionError: assert 1500.3838068681146 == 1414.213562373095 ± 7.1e+01
  comparison failed
  Obtained: 1500.3838068681146
  Expected: 1414.213562373095 ± 7.1e+01
=========================== short test summary info ============================
FAILED testing/cli/test_cli_sct_compute_snr.py::test_sct_compute_snr_single_3d_mask_object
==== 1 failed, 370 passed, 11 skipped, 88441 warnings in 372.84s (0:06:12) =====
```

joshuacwnewton:
I think this might have just been a fluke? The restarted run passes, and the installation logs don't look meaningfully different between the passing and failing runs.

We can re-open if we encounter this again. 🤔

PR #3866, opened 2022-08-24
Replace "argv=None" default by "argv: Sequence[str]" type hint

### LINKED ISSUE #3176: Replace argv=None with type hinting in script main functions
In https://github.com/neuropoly/spinalcordtoolbox/issues/3085#issuecomment-740083842, we decided on `None` as a default value for `argv` in all of our scripts.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/37f0bde2ca521424598ba9665853b2c99944996b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L509

But, on second thought, why does main need a default value? We expect that `argv` should always be passed, and `None` can cause issues like: https://github.com/neuropoly/spinalcordtoolbox/pull/3156#issuecomment-759593822. To fix #3156, I rewrote the `main()` declaration as follows:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/51511fc387113c853c290b0a5d81c85a02fd9f93/spinalcordtoolbox/scripts/sct_label_utils.py#L234-L239

(We also expect `argv` to be a list of strings (i.e. unparsed CLI arguments), so I added type hinting for this. An empty list `[]` will pass the type-hinting check, too. This is good, as calling `main(argv=[])` fulfills scripts where calling no args is valid, too.) 

I think it could be good to add this to the remaining scripts, too.

mguaypaq:
## Description
This PR replaces the `def main(argv=None)` pattern in SCT scripts with ~~`def main(argv: List[str])`~~ `def main(argv: Sequence[str])`, since we never call `main()` without specifying `argv`, and since the object passed in is always a list of strings. ~~It uses `List` rather than the more generic `Sequence` because one script (sct_extract_metric.py) relies on specific list methods (namely, `append`).~~

I've also updated the [dev wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming:-CLI-script-structure) to reflect this change.

## Linked issues
Fixes #3176.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_snr.py
+++ b/spinalcordtoolbox/scripts/sct_compute_snr.py
@@ -15,7 +15,7 @@
 ########################################################################################
 
 import sys
-from typing import List
+from typing import Sequence
 
 import numpy as np
 
@@ -130,7 +130,7 @@ def weighted_std(values, weights):
     return np.sqrt(variance)
 
 
-def main(argv: List[str]):
+def main(argv: Sequence[str]):
     parser = get_parser()
     arguments = parser.parse_args(argv)
     verbose = arguments.v
--- a/spinalcordtoolbox/scripts/sct_compute_snr.py
+++ b/spinalcordtoolbox/scripts/sct_compute_snr.py
@@ -15,6 +15,7 @@
 ########################################################################################
 
 import sys
+from typing import List
 
 import numpy as np
 
@@ -129,7 +130,7 @@ def weighted_std(values, weights):
     return np.sqrt(variance)
 
 
-def main(argv=None):
+def main(argv: List[str]):
     parser = get_parser()
     arguments = parser.parse_args(argv)
     verbose = arguments.v
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -11,7 +11,7 @@
 import os
 import sys
 import pickle
-from typing import List
+from typing import Sequence
 
 import numpy as np
 import pandas as pd
@@ -504,7 +504,7 @@ class AnalyzeLeion:
         os.chdir(self.tmp_dir)  # go to tmp directory
 
 
-def main(argv: List[str]):
+def main(argv: Sequence[str]):
     """
     Main function
     :param argv:
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -11,6 +11,7 @@
 import os
 import sys
 import pickle
+from typing import List
 
 import numpy as np
 import pandas as pd
@@ -503,7 +504,7 @@ class AnalyzeLeion:
         os.chdir(self.tmp_dir)  # go to tmp directory
 
 
-def main(argv=None):
+def main(argv: List[str]):
     """
     Main function
     :param argv:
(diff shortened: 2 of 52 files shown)


joshuacwnewton:
> It uses `List` rather than the more generic `Sequence` because one script (sct_extract_metric.py) relies on specific list methods (namely, `append`).

Exactly! I initially tried `Sequence` in the previous PR and the type checking helper got mad at me because of the list-specific methods. 😝

Now that I think about it, though... is there any merit to keeping `Sequence`, and instead amending sct_extract_metric to convert to `list()` prior to the check against `-list-labels`? 

> I've also updated the [dev wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming:-CLI-script-structure) to reflect this change.

Thank you for staying on top of the wiki! ♥️

----

One other question: How do you feel about docstrings for `main`? I added them with the previous commits, mainly to clarify that it expects a list of unparsed strings, but perhaps this is a little overkill/self-explanatory?

mguaypaq:
> Now that I think about it, though... is there any merit to keeping `Sequence`, and instead amending sct_extract_metric to convert to `list()` prior to the check against `-list-labels`?

Yes, on second thought, it's actually kind of rude of that one `main()` function to modify `argv` in-place, and it would be better to make a copy. And while thinking of this, I figured out that the current in-place modification has a very subtle bug: if `-list-labels` is used more than once, only the first instance of the argument would be moved to the end of the list. I'll fix both issues. And I'll switch to `Sequence` everywhere.

> One other question: How do you feel about docstrings for `main`? I added them with the previous commits, mainly to clarify that it expects a list of unparsed strings, but perhaps this is a little overkill/self-explanatory?

I didn't mind having it there, but yes I think it's a little overkill. The type hint makes it pretty clear, and the only use of `argv` in `main()` is pretty close to the top in case there's any doubt about what should be in the list of strings. For something like this which is a boilerplate pattern throughout the codebase, I think the dev wiki is a good place to centralize the explanation.

> Thank you for staying on top of the wiki! :heart:

Haha I blame the checklist in the PR template for reminding me :wink:

joshuacwnewton (approved):
I've removed the `main()` function docstrings, as discussed above. Other than that, this LGTM!

spinalcordtoolbox/scripts/sct_compute_snr.py
```diff
@@ -129,7 +130,7 @@ def weighted_std(values, weights):
     return np.sqrt(variance)
 
 
-def main(argv=None):
+def main(argv: Sequence[str]):
```
joshuacwnewton, line 133: Hmm. It looks like we encountered https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3759 again:

```
 =================================== FAILURES ===================================
__________________ test_sct_compute_snr_single_3d_mask_object __________________
  File "/home/runner/sct_0.0/testing/cli/test_cli_sct_compute_snr.py", line 267, in test_sct_compute_snr_single_3d_mask_object
    assert snr == pytest.approx(np.sqrt(2*SIGNAL_OBJECT**2), rel=0.05)
AssertionError: assert 1512.0527270220502 == 1414.213562373095 ± 7.1e+01
  comparison failed
  Obtained: 1512.0527270220502
  Expected: 1414.213562373095 ± 7.1e+01
FAILED ../../../sct_0.0/testing/cli/test_cli_sct_compute_snr.py::test_sct_compute_snr_single_3d_mask_object
```

I've retriggered the workflow, and I'm presuming it's going to pass (based on what's written in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3759), but it's strange to see the same error twice...

mguaypaq: Thanks for the cross-reference!

PR #3870, opened 2022-09-01
Fix `include_list`/`exclude` argument ordering in call to `_filter_directories`

### LINKED ISSUE #3598: `include` and `include_list` function differently for `ses` folders

joshuacwnewton:
## Description

This issue is a follow-on for https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3859. In that PR, a bug was merged to `master`, because the ordering of the arguments for the `_filter_directories` function call was incorrect:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/51df5fc0569c0c04571e44601b7ddf95c981631b/spinalcordtoolbox/scripts/sct_run_batch.py#L213

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/51df5fc0569c0c04571e44601b7ddf95c981631b/spinalcordtoolbox/scripts/sct_run_batch.py#L507-L509

Thank you to @valosekj for catching this. :heart: 

## Linked issues

Fixes #3598.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -505,8 +505,8 @@ def main(argv: Sequence[str]):
         parser.error('Only one of `exclude` and `exclude-list` can be used')
 
     subject_dirs = _filter_directories(subject_dirs,
-                                       arguments.include, arguments.exclude,
-                                       arguments.include_list, arguments.exclude_list)
+                                       include=arguments.include, include_list=arguments.include_list,
+                                       exclude=arguments.exclude, exclude_list=arguments.exclude_list)
 
     # Determine the number of jobs we can run simultaneously
     if arguments.jobs < 1:


valosekj (approved):
Thank you for your prompt fix @joshuacwnewton !! 👍🏻 

Works fine for both

 ``` 
"include_list": ["sub-0001/ses-01","sub-0001/ses-02","sub-0002/ses-01","sub-0002/ses-02"],
```

and 

```
"include_list": ["sub-0001","sub-0002"],
```

mguaypaq (approved):
Thanks for the double fix: giving the arguments in the right order, _and_ passing them by keyword.

PR #3805, opened 2022-06-02
Repair commenting out pre-4.0 sct settings.

kousu:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [x] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [x] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3804

This was inadvertently broken in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3591.

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3804

PR #3872, opened 2022-09-02
Properly escape `\` in the regex for removing `sct_env` from old RC files

### LINKED ISSUE #3871: `sct_env`-removing regex pattern is malformed, and causes an error when invoked
@valosekj privately reported the following error when trying to reinstall a newer version of SCT:

```
In case an old version SCT is already installed (4.0.0-beta.1 or before), remove 'sct_env' declaration in RC file

Unmatched ( in regex; marked by <-- HERE in m/^(# )?( <-- HERE .*bin/ at -e line 1.

Installation failed!
```

This is referring to the following section of code in `install_sct`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a4b10b65a5bf1156d23969478a4f1959a5064ff4/install_sct#L695-L698

@kousu and I tested this regex pattern quite thoroughly in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3805#discussion_r888226125. But, I realize now that there's a subtle difference between the regex pattern that @kousu tested:

```console
parker:~ p115628$ perl -pi -e 's/^(# )?(.*bin\/sct_env)/# \2/' test_file.txt 
```

And the actual regex pattern that got committed:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a4b10b65a5bf1156d23969478a4f1959a5064ff4/install_sct#L697

While @kousu indeed escaped the `/` in `bin/sct_env` in testing, `\/` was never actually committed. :sweat_smile: 

Fixing this, then, should just be a matter of escaping the `/` character.

joshuacwnewton:
## Description

`install_sct` uses a regex-based `perl -pi -e` find and replace command to comment out any instances of the `sct_env` environment variable being set (which is no longer used).

This PR amends the `perl -pi -e` statement so that it uses the pattern that was tested by @kousu in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3805#discussion_r888226125.

(The reason for this bug is because, in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3805#discussion_r888226125, the PR "suggestion" was subtly different than what was actually tested.)

## Linked issues

Fixes #3871.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -694,7 +694,7 @@ conda deactivate >/dev/null 2>&1
 if [[ -e "$RC_FILE_PATH" ]]; then
     if grep "sct_env" "$RC_FILE_PATH"; then
       print info "In case an old version SCT is already installed (4.0.0-beta.1 or before), remove 'sct_env' declaration in RC file"
-      perl -pi -e 's/^(# )?(.*bin/sct_env)/# \2/' "$RC_FILE_PATH"
+      perl -pi -e 's/^(# )?(.*bin\/sct_env)/# \2/' "$RC_FILE_PATH"
     fi
     if grep "^export MPLBACKEND=Agg" "$RC_FILE_PATH"; then
       print info "Commenting out 'export MPLBACKEND=Agg' from previous SCT installation"


mguaypaq (approved):
I don't think this warrants an automated test, but just to confirm, I did copy-paste the command in a terminal and checked that it correctly comments out any line containing `bin/sct_env`.

Issue #3873, opened 2022-09-06
Warn when labels exist outside of the segmentation (due to disappearing during straightening step)


joshuacwnewton:
From https://forum.spinalcordmri.org/t/labels-for-lumbar-spinal-level/737/17?u=joshuacwnewton:

> Tracing the source code, I found that the label in your label image disappears after [the straightening step.](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c26086ec3e779d4f8f025651e46de2f962698d98/spinalcordtoolbox/scripts/sct_register_to_template.py#L593-L601) 
> 
> Internally, in order to straighten the label image, we reuse the same warping field that was used to straighten the segmentation. (As well, the segmentation image gets cropped to the extent of the segmentation (shown in grey below)):
> 
> ![image](https://user-images.githubusercontent.com/16181459/188667906-1bebafbd-6134-42d9-9e4e-c7945e1d14ea.png)
> 
> So, because the label was outside of this cropped region (as you surmised), the straightening step causes the label to be lost (because the warping field, shown below, isn't defined for that region):
> 
> ![image](https://user-images.githubusercontent.com/16181459/188667983-1351a948-5135-4c75-b149-0af366146e7c.png)

I imagine this is by design? (I'm assuming it's not possible to determine the straightening transform for regions outside of the segmentation.)

That said, we could be doing a better job of warning users when this might happen:

* For example, we could check the coordinates of the nonzero labels, then check the dimensions of the warping field, then throw a warning/exception if we can determine that that the label won't be transformed as part of the straightening. 
* Or, more simply, we could compare the # of labels before straightening to the # of labels after straightening, and emit a warning/error if a label was lost in the straightening.

PR #3874, opened 2022-09-06
Ensure that calling `printv(type='error')` actually exits the program

### LINKED ISSUE #3861: Replace `printv(type='error')` with proper exceptions
In many places in SCT's code, we have `printv` statements that have `type='error'` specified. For example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/98872496fb8325686c1d3e2c9602e20e4d960c26/spinalcordtoolbox/utils/fs.py#L98-L102

The intent here is to exit the program, but `printv` doesn't actually raise an exception, so execution will continue until... something else breaks. This will usually result in some sort of unexpected/unintentional/buggy behavior.

We've been slowly tackling these as we've come across them in other PRs, but we might want to try to tackle these all in one go by replacing them with proper exceptions, then getting rid of `type='error'` entirely.

----

By my count, there are 61 instances:

image

Given the scope of these changes, we might want to also consider simply raising an exception within `printv`. It won't be as nice as raising properly-typed exceptions at the source, but it will make the change a one-liner, and let us more easily see what might break as a result of these changes. Could be a good intermediate/testing step?

joshuacwnewton:
## Description

In many places in our codebase, we call `printv` using the argument `type='error'`, with the intent to halt execution. But, currently, the only thing that `printv(type='error')` does is print a message in red text. In other words, the script will continue, until some other unintended error is thrown later in processing (if that even happens!).

This PR applies a quick and dirty change to cause printv to throw an exception instead of continuing. However, there are some caveats here:

* This small change is actually affecting [61 instances](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3861) of `printv(type='error')` in the codebase. Yet, our test suite only seems to cover 2 of these instances (2c2821dc943c649db2f500327382f6c66d1c9f32, 83aea4a93d907d3f7830899b9f36fc832072f754). Meaning that, we now have 59 other new (unexpected?) errors that will now be thrown when they weren't before. 
    * Can we be certain that these new exceptions won't break existing workflows? Is it possible there have been silent error messages buried in logs that have become the norm for working scripts?
* ~~This change necessarily uses a generic `Exception`, rather than a tailored `Exception`. Meaning that, we don't quite have the capability to catch specific errors in `try: except:` blocks.~~ Addressed by [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3874#issuecomment-1239598714).
* Lastly, the errors are always thrown from `spinalcordtoolbox.utils.sys` instead of the calling script. (I'm not sure if this has any meaningful effect, though.)

Given these concerns, we may want to consider scrapping this PR, then meticulously going through each of these 61 instances, replacing `printv` with `raise <type>Exception` and reviewing the error message for each instance. 

## Linked issues

Fixes #3861.


spinalcordtoolbox/utils/shell.py
```diff
@@ -199,8 +199,10 @@ def error(self, message):
         """
         # Source: https://stackoverflow.com/a/4042861
         self.print_help(sys.stderr)
-        printv(f'\n{self.prog}: error: {message}\n', verbose=1, type='error', file=sys.stderr)
-        sys.exit(2)
+        message_formatted = (ANSIColors16.Bold + ANSIColors16.LightRed
+                             + f'\n{self.prog}: error: {message}\n\n'
+                             + ANSIColors16.ResetAll)
+        self.exit(2, message_formatted)
```
joshuacwnewton, line 205: NB: The formatting change here was done in an effort to better mimic the parent method from `argparse`:

```python
    def error(self, message):
        """error(message: string)

        Prints a usage message incorporating the message to stderr and
        exits.

        If you override this in a subclass, it should not return -- it
        should either exit or raise an exception.
        """
        self.print_usage(_sys.stderr)
        args = {'prog': self.prog, 'message': message}
        self.exit(2, _('%(prog)s: error: %(message)s\n') % args)
```

joshuacwnewton: Additionally, I found that I needed to add a second `\n` in order to preserve the same formatting as before. 

I'm not sure why that is? Maybe it's because `printv` inherently adds a newline to the end of the message, since it's a print function, but `self.exit` does not? Not sure, tut I thought I'd acknowledge the change just in case it's not, say, consistent across platforms.

mguaypaq: About the underscore: yes, it's even documented that `gettext.gettext` is often imported as simply `_` ([link](https://docs.python.org/3/library/gettext.html#gettext.gettext)). It's not my favourite convention though, since it's so sneaky! But I believe this is a cross-language convention (for example C and PHP both seem to use it), so it's not going away.

mguaypaq (commented):
After thinking about it, I see two sensible approaches:
1. Modify all 61 instances to throw an appropriate `Exception` subclass, because as you note we can't do that within `printv(type='error')`.
2. Modify `printv(type='error')` to call `sys.exit(1)` after printing the red error message.

But given that `printv` is documented as a convenience function to be used only in scripts (rather than in the API modules), I think `sys.exit` makes more sense than showing a traceback. (Either way, this behaviour should be mentioned in the docstring.)

(Aside: I think `raise`ing from `spinalcordtoolbox.utils.sys` specifically is not a problem, because the 'real' culprit would still show up in the traceback.)

As for whether this might unduly break existing scripts:
* We've seen several instances now where `printv(type='error')` is clearly used with the intention of exiting the program, so I would assume that's also the intent for most, if not all, of the 61 instances.
* I also suspect that most (but not all) of the 61 instances would fail not too long after `printv(type='error')` as things stand right now, just with a worse error message.
* For the not-currently-failing instances, they're very likely to be silently wrong, and being loudly wrong would be an improvement.
* I also suspect that existing workflows which would be broken because of this change, could stay on SCT v5.7 until we fix any problems that come up. And we'll only get those bug reports if we start failing on `printv(type='error')`.

mguaypaq:
Also: I like that you modified `SCTArgumentParser.error()` specifically, because that code runs in a special context (argument parsing, which should try hard to be user-friendly).

joshuacwnewton:
> But given that `printv` is documented as a convenience function to be used only in scripts (rather than in the API modules), I think `sys.exit` makes more sense than showing a traceback.

This is an interesting perspective, and not one that I think had come up in past API vs. CLI discussions! 

Is the idea here that raising an exception has value because it displays a traceback, and because specific exception types can be caught and handled by the caller of the function? (Plus, these features are only valuable when the source of the error is nested 1+ layers deep in the call stack.) 

So, because the CLI scripts should (in theory) only ever be the top layer of the call stack, exceptions lose this utility, to the point of even being unnecessarily verbose? And, a hypothetical ideal would be to only ever `sys.exit` at the CLI layer, while saving exceptions for the API layer.

mguaypaq:
> Just to make sure I'm understanding correctly: Is the idea here that raising an exception only has value because it displays a traceback, and because specific exception types can be caught and handled by the caller of the function? (Plus, these features are only valuable when the source of the error is nested 1+ layers deep in the call stack.)

That's right: in my mind, for these reasons, exceptions are a 'developer-facing' tool, whereas printed messages and program exit codes are 'user-facing' tools.

> So, because the CLI scripts should (in theory) only ever be the top layer of the call stack, exceptions lose this utility, to the point of even being unnecessarily verbose? And, a hypothetical ideal would be to only ever `sys.exit` at the CLI layer, while saving exceptions for the API layer.

Exactly. Because they're top-of-stack, CLI scripts _have_ to decide how to present things to the user. And the API layer is _not_ top-of-stack, so it can (and probably should) delegate those decisions to the calling function.

joshuacwnewton:
I really like this distinction. It makes a lot of sense to me for printv to throw `sys.exit` instead?

In that case, rather than phasing out `printv(type='error')`, we may in fact want to encourage its usage even more, by replacing existing exceptions/tracebacks in CLI scripts with `printv`.

((Even right now, I'm separately working on https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3873, and adding a new exception, but the traceback is quite user unfriendly? I very much wish I could be using `printv(type='error')` instead.))

mguaypaq (approved):
This looks great, thanks!

PR #3876, opened 2022-09-07
`shell.py`: Let `Metavar` inherit from `str` to silence PyCharm type warning

joshuacwnewton:
## Description

(This is a tiny developer QOL change.)

Currently, SCT uses a custom Enum class called `Metavar` to print `<type>` strings in argparse help descriptions:

image

PyCharm doesn't like the fact that we're using a bare `Enum`, though:

image

We can fix this by adding `str` to the list of types that `Metavar` inherits, though:

image

## Linked issues

N/A.

### DIFF
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -244,7 +244,7 @@ def list_type(delimiter, subtype):
     return list_typecast_func
 
 
-class Metavar(Enum):
+class Metavar(str, Enum):
     """
     This class is used to display intuitive input types via the metavar field of argparse
     """


spinalcordtoolbox/utils/shell.py
```diff
@@ -244,7 +244,7 @@ def list_typecast_func(string):
     return list_typecast_func
 
 
-class Metavar(Enum):
+class Metavar(str, Enum):
```
joshuacwnewton, line 247: Source on this change: https://stackoverflow.com/a/58608362

joshuacwnewton: I also noticed that the example in ^ doesn't include a `__str__` method. However, excluding this method results in the following:

image

mguaypaq: It looks like Python 3.11 has [`enum.StrEnum`](https://docs.python.org/3.11/library/enum.html#enum.StrEnum), so when we upgrade we should be able to use that (and it looks like [it won't need `__str__`](https://docs.python.org/3.11/library/enum.html#notes)? I'm not sure).

mguaypaq (approved):
LGTM

PR #3878, opened 2022-09-07
Simplify `add_suffix`

mguaypaq:
The strings are already concatenated by `+`, and `os.path.join` with a single string argument just returns it unchanged.


joshuacwnewton (approved):
Yes, this makes perfect sense! `os.path.join` was doing nothing here. :)

Plus, any pathlike components should be contained entirely in `stem`, with `suffix` and `ext` not needing to use platform-specific path separators or anything like that, so regular old string concatenation is fine here.

Issue #3877, opened 2022-09-07
`display_viewer_syntax` prints unnecessary output during intermediate CLI scripts


joshuacwnewton:
Many SCT CLI scripts end their processing with a call to the function `display_viewer_syntax`, which outputs some handy commands for opening output images in image viewers:

![image](https://user-images.githubusercontent.com/16181459/188932961-8d1ef80d-a5ed-4e32-830b-43cb0cfda9fe.png)

However, in many places SCT scripts will call the `main` functions of _other_ SCT scripts for intermediate processing. For example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c26086ec3e779d4f8f025651e46de2f962698d98/spinalcordtoolbox/scripts/sct_register_to_template.py#L545-L550

In these cases, you can still see the `display_viewer_syntax` output for the _subcalls_, despite those calls being intermediate processing. For example, here's some output that very clearly mentions `/tmp/` files:

![image](https://user-images.githubusercontent.com/16181459/188933282-2a535656-0934-4921-b63b-d611cda653b4.png)

This extra output clutters the logs and makes it more difficult to tell what the proper output files are. So, it would be nice if we could silence the `display_viewer_syntax` output if the script isn't being run from the CLI.

joshuacwnewton:
One thing we could do here is move `display_viewer_syntax` to be explicitly within the `if __name__ == "__main__":` block. For example, `sct_register_to_template`:

Before:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c26086ec3e779d4f8f025651e46de2f962698d98/spinalcordtoolbox/scripts/sct_register_to_template.py#L794-L800

After:

```python
def main(argv):
    # Some processing
    # [...]
    return fname_data, fname_template, fname_template2anat, fname_anat2template

if __name__ == "__main__":
    init_sct()
    fname_data, fname_template, fname_template2anat, fname_anat2template = main(sys.argv[1:])
    display_viewer_syntax([fname_data, fname_template2anat])
    display_viewer_syntax([fname_template, fname_anat2template])
```

This would have the added benefit of nudging us to define the output files for each `main()` call, too. That feels like a good practice to me? It's nice to have motivation to definie which files each CLI script intends to generate. I imagine it makes verifying the expected output of each script a bit easier in tests, too?

joshuacwnewton:
[...]

PR #3880, opened 2022-09-09
Throw error when labels are lost during the straightening transform

### LINKED ISSUE #3873: Warn when labels exist outside of the segmentation (due to disappearing during straightening step)

joshuacwnewton:
## Description

This PR adds a new error to halt registration when labels are lost during the straightening transform, and gives some advice to address situations like https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3873#issue-1363434092.

I've thrown an error here because, if not caught, this problem would result in an error later in processing due to a mismatch between template labels and image labels. (See [this forum post](https://forum.spinalcordmri.org/t/labels-for-lumbar-spinal-level/737/11?u=joshuacwnewton) for an example.) The purpose of this change is just to make the root cause of the issue clearer.

One caveat here is that the output is a little ugly, specifically because of issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3877.

<details>

image

</details>

However, this is very easily solved by e.g. setting `-v 0` in the call to `sct_apply_transfo`:

<details>

image

</details>

I plan to do this next as part of a fix for #3877.

## Linked issues

Fixes #3873.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -592,6 +592,7 @@ def main(argv: Sequence[str]):
 
             # Apply straightening to labels
             printv('\nApply straightening to labels...', verbose)
+            label_vals_src = {coord.value for coord in Image(ftmp_label).getCoordinatesAveragedByValue()}
             sct_apply_transfo.main(argv=[
                 '-i', ftmp_label,
                 '-o', add_suffix(ftmp_label, '_straight'),
@@ -599,6 +600,12 @@ def main(argv: Sequence[str]):
                 '-w', 'warp_curve2straight.nii.gz',
                 '-x', 'nn'])
             ftmp_label = add_suffix(ftmp_label, '_straight')
+            label_vals_out = {coord.value for coord in Image(ftmp_label).getCoordinatesAveragedByValue()}
+            missing_labels = label_vals_src - label_vals_out
+            if missing_labels:
+                raise RuntimeError(f"Labels {missing_labels} were lost during straightening transform. This can be "
+                                   f"caused by the labels being outside the ROI of the spinal cord segmentation. "
+                                   f"Please make sure all labels are within the ROI of the spinal cord segmentation.")
 
             # Compute rigid transformation straight landmarks --> template landmarks
             printv('\nEstimate transformation for step #0...', verbose)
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -592,13 +592,22 @@ def main(argv: Sequence[str]):
...
                 '-x', 'nn'])
             ftmp_label = add_suffix(ftmp_label, '_straight')
+            label_vals_out = {coord.value for coord in Image(ftmp_label).getCoordinatesAveragedByValue()}
+            missing_labels = label_vals_src - label_vals_out
+            if missing_labels:
+                printv(
+                    f"ERROR: Labels {missing_labels} were lost during straightening transform. This can be caused by "
+                    f"the labels being outside the ROI of the spinal cord segmentation. Please make sure all labels "
+                    f"are within the ROI of the spinal cord segmentation.", type='error'
+                )
```
joshuacwnewton, line 611: In https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3873#issue-1363434092, I proposed 2 possible checks:


>  * For example, we could check the coordinates of the nonzero labels, then check the dimensions of the warping field, then throw a warning/exception if we can determine that that the label won't be transformed as part of the straightening.
>  * Or, more simply, we could compare the # of labels before straightening to the # of labels after straightening, and emit a warning/error if a label was lost in the straightening.


I was hesitant to go for the former option (checking the coordinates of the label relative to the warping field) because of the dilation step; there's a possible case where the single-voxel label is outside of the warping field, but the dilation step expands the label to be inside the warping field. So, I wasn't sure: should I check against the undilated label or the dilated label? Neither guarantees for sure that the label will be lost/kept.

So, that's why I felt that checking the presence of labels post-straightening was more foolproof.

mguaypaq (approved):
This is great: you're testing for exactly the right condition that will lead to a confusing error, and outputting a very helpful message instead.

PR #3882, opened 2022-09-14
Address `[Errno 24] Too many open files` during motion correction

### LINKED ISSUE #2149: OSError: Too many open files when running moco on large files
When running `sct_fmri_moco` in large files with hundreds of volumes, the function crashes (see below). 
Note: this issue would likely occur on `sct_dmri_moco` as well. To be tested.

<Details>

~~~
Merging volumes...
  File "/Users/julien/code/sct/scripts/sct_fmri_moco.py", line 396, in <module>
    main()
  File "/Users/julien/code/sct/scripts/sct_fmri_moco.py", line 209, in main
    fmri_moco(param)
  File "/Users/julien/code/sct/scripts/sct_fmri_moco.py", line 328, in fmri_moco
    im_mean_list.append(Image(file_data + '_mean_' + str(iGroup) + ext_data))
  File "/Users/julien/code/sct/spinalcordtoolbox/image.py", line 235, in __init__
    self.loadFromPath(param, verbose)
  File "/Users/julien/code/sct/spinalcordtoolbox/image.py", line 321, in loadFromPath
    self.data = self.im_file.get_data()
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/nibabel/spatialimages.py", line 572, in get_data
    data = np.asanyarray(self._dataobj)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/numpy/core/numeric.py", line 533, in asanyarray
    return array(a, dtype, copy=False, order=order, subok=True)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/nibabel/arrayproxy.py", line 145, in __array__
    raw_data = self.get_unscaled()
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/nibabel/arrayproxy.py", line 140, in get_unscaled
    mmap=self._mmap)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/nibabel/volumeutils.py", line 509, in array_from_file
    offset=offset)
  File "/Users/julien/code/sct/python/lib/python2.7/site-packages/numpy/core/memmap.py", line 273, in __new__
    self.filename = os.path.abspath(filename.name)
  File "/Users/julien/code/sct/python/lib/python2.7/posixpath.py", line 364, in abspath
    cwd = os.getcwd()
OSError: [Errno 24] Too many open files
Total processing time: 1 min 56 s
~~~

</Details><p>

To reproduce (on an OSX 4-core 2.6GHz  16GB RAM):
~~~
cd sct_example_data/fmri
sct_image -i fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz,fmri.nii.gz -concat t -o fmri_concat.nii.gz 
sct_fmri_moco -i fmri_concat.nii.gz
~~~

joshuacwnewton:
### Context

On Unix systems, when motion correction is run on a 4D image with many 3D volumes (e.g. `[x, y, z, 300]`), and the number of volumes exceeds the open file descriptor limit (e.g. `ulimit -Sn ==> 256`), then SCT will throw an `[Errno 24] Too many open files` error.

The reason for this is because of this LOC:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7ce5ea59e213510ff0566eef99d3cefe811f0051/spinalcordtoolbox/moco.py#L640

`Image`, by default, will load the data arrays of `.nii` images as `np.memmap` objects, which keeps an open file descriptor. So, for a 300-volume dMRI image, at least 300 file descriptors will be kept open at once.

### Proposed solutions

To fix this, we can tackle the problem a number of ways:

1. Rely on the workaround of `ulimit -Sn` to increase the limit on open file descriptors. (Presented here as 6900c92cdc354dfbae1bd3e3572f272051eb6af3)
2. Avoid `np.memmap` entirely, but only in the offending L640. (Presented here as 0055c0e160cb2caf47dfc4a83cc826589b2ce5be)
3. Avoid `np.memmap` _conditionally_, based on whether or not `Image.loadFromPath` throws an `OSError`. (Presented here as cf82f856cd30cc5013bdcb39eb5275549dc60825).
4. Change the way we perform concatenation, but only in the offending L640. (Not immediately presented here, due to time complexity concerns, as well as [numerical differences](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2149#issuecomment-1245534459).)
5. Change the way we perform concatenation, by rewriting how `concat_data` processes filenames/`Image` objects. (Not immediately presented here due to increased scope caused by this change.)

I'm not quite sure of the best solution here, because each approach has various caveats and tradeoffs. I would be happy to discuss further, though!

## Linked issues

Fixes #2149

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -248,14 +248,17 @@ class Image(object):
     a few methods (load, save) that deal with image dtype.
     """
 
-    def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, dim=None, verbose=1,
-                 check_sform=False):
+    def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, dim=None,
+                 mmap=(not sys.platform.startswith('win32')), verbose=1, check_sform=False):
         """
         :param param: string indicating a path to a image file or an `Image` object.
         :param hdr: a nibabel header object to use as the header for the image (overwritten if `param` is provided)
         :param orientation: a three character orientation code (e.g. RPI).
         :param absolutepath: a relative path to associate with the image.
         :param dim: The dimensions of the image, defaults to automatically determined.
+        :param mmap: Whether to load data arrays as np.memmaps. Defaults to False (i.e. np.array) on Windows, and True
+          (i.e. np.memmap) on every other platform. (The context for defaulting to False on Windows is:
+          https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3695)
         :param verbose: integer how verbose to be 0 is silent 1 is chatty.
         :param check_sform: whether or not to check whether the sform matches the qform. If this is set to `True`,
           `Image` will fail raise an error if they don't match.
@@ -280,7 +283,7 @@ class Image(object):
         # load an image from file
         if isinstance(param, str):
             try:
-                self.loadFromPath(param, verbose)
+                self.loadFromPath(param, mmap, verbose)
             except OSError as e:
                 if str(e) == "[Errno 24] Too many open files":
                     raise OSError("[Errno 24] Too many open files. Please try increasing your system's file descriptor "
@@ -397,7 +400,7 @@ class Image(object):
         self.hdr.set_qform(self.hdr.get_sform())
         self.hdr._structarr['qform_code'] = self.hdr._structarr['sform_code']
 
-    def loadFromPath(self, path, verbose):
+    def loadFromPath(self, path, mmap, verbose):
         """
         This function load an image from an absolute path using nibabel library
 
@@ -406,7 +409,7 @@ class Image(object):
         """
 
         self.absolutepath = os.path.abspath(path)
-        im_file = nib.load(self.absolutepath, mmap=(not sys.platform.startswith('win32')))
+        im_file = nib.load(self.absolutepath, mmap=mmap)
         self.affine = im_file.affine.copy()
         self.data = im_file.get_data()
         self.hdr = im_file.header.copy()
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -287,10 +287,9 @@ class Image(object):
                 self.loadFromPath(param, mmap, verbose)
             except OSError as e:
                 if e.errno == errno.EMFILE:
-                    raise OSError("[Errno 24] Too many open files. Please try increasing your system's file descriptor "
-                                  "limit by using the command `ulimit -Sn`.")
-                else:
-                    raise e
+                    e.strerror += (". Please try increasing your system's file descriptor "
+                                   "limit by using the command `ulimit -Sn`.")
+                raise e
         # copy constructor
         elif isinstance(param, type(self)):
             self.copy(param)
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -286,16 +286,8 @@ class Image(object):
                 self.loadFromPath(param, mmap, verbose)
             except OSError as e:
                 if str(e) == "[Errno 24] Too many open files":
-                    try:
-                        # Only load images into memory if there are no more available file descriptors
-                        self.loadFromPath(param, False, verbose)
-                    except OSError as e:
-                        # If for some reason, an OSError is still thrown, then give the user a helpful error message
-                        if str(e) == "[Errno 24] Too many open files":
-                            raise OSError("[Errno 24] Too many open files. Please try increasing your system's file "
-                                          "descriptor limit by using the command `ulimit -Sn`.")
-                        else:
-                            raise e
+                    raise OSError("[Errno 24] Too many open files. Please try increasing your system's file descriptor "
+                                  "limit by using the command `ulimit -Sn`.")
                 else:
                     raise e
         # copy constructor
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -286,8 +286,16 @@ class Image(object):
                 self.loadFromPath(param, mmap, verbose)
             except OSError as e:
                 if str(e) == "[Errno 24] Too many open files":
-                    raise OSError("[Errno 24] Too many open files. Please try increasing your system's file descriptor "
-                                  "limit by using the command `ulimit -Sn`.")
+                    try:
+                        # Only load images into memory if there are no more available file descriptors
+                        self.loadFromPath(param, False, verbose)
+                    except OSError as e:
+                        # If for some reason, an OSError is still thrown, then give the user a helpful error message
+                        if str(e) == "[Errno 24] Too many open files":
+                            raise OSError("[Errno 24] Too many open files. Please try increasing your system's file "
+                                          "descriptor limit by using the command `ulimit -Sn`.")
+                        else:
+                            raise e
                 else:
                     raise e
         # copy constructor
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -12,6 +12,7 @@
 
 import sys
 import os
+import errno
 import itertools
 import warnings
 import logging
@@ -285,7 +286,7 @@ class Image(object):
             try:
                 self.loadFromPath(param, mmap, verbose)
             except OSError as e:
-                if str(e) == "[Errno 24] Too many open files":
+                if e.errno == errno.EMFILE:
                     raise OSError("[Errno 24] Too many open files. Please try increasing your system's file descriptor "
                                   "limit by using the command `ulimit -Sn`.")
                 else:
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -279,7 +279,14 @@ class Image(object):
 
         # load an image from file
         if isinstance(param, str):
-            self.loadFromPath(param, verbose)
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/image.py
```diff
@@ -279,7 +282,22 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
...
+                    try:
+                        # Only load images into memory if there are no more available file descriptors
+                        self.loadFromPath(param, False, verbose)
+                    except OSError as e:
+                        # If for some reason, an OSError is still thrown, then give the user a helpful error message
+                        if str(e) == "[Errno 24] Too many open files":
+                            raise OSError("[Errno 24] Too many open files. Please try increasing your system's file "
+                                          "descriptor limit by using the command `ulimit -Sn`.")
+                        else:
+                            raise e
```
joshuacwnewton, line 298: One thing I want to note here is, if we are catching exceptions and replacing the tracebacks with a friendlier output message, then presumably that would be the job of a non-exception function like `printv`.

But, we've also got a soft-ban on `printv` in the API. So, I guess I should be using `logger.error` + `sys.exit` instead?

spinalcordtoolbox/image.py
```diff
@@ -279,7 +282,22 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
 
         # load an image from file
         if isinstance(param, str):
-            self.loadFromPath(param, verbose)
+            try:
+                self.loadFromPath(param, mmap, verbose)
+            except OSError as e:
+                if str(e) == "[Errno 24] Too many open files":
```
mguaypaq, line 288: Nitpick: checking the string representation of the error may be fragile, it might be better to compare the numeric `errno` attribute:
```py3
import errno  # can be imported at the top of the module
if e.errno == errno.EMFILE:
    # Too many open files
```
This should even [work on Windows](https://docs.python.org/3/library/exceptions.html#OSError.winerror), although the `ulimit` suggestion may not be applicable there.

spinalcordtoolbox/image.py
```diff
@@ -279,7 +282,22 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
...
+                    try:
+                        # Only load images into memory if there are no more available file descriptors
+                        self.loadFromPath(param, False, verbose)
+                    except OSError as e:
+                        # If for some reason, an OSError is still thrown, then give the user a helpful error message
+                        if str(e) == "[Errno 24] Too many open files":
+                            raise OSError("[Errno 24] Too many open files. Please try increasing your system's file "
```
mguaypaq, line 295: Another nitpick: since [`OSError` has a bunch of other useful fields](https://docs.python.org/3/library/exceptions.html#OSError), it would be nice to preserve them.

One way would be to use second constructor, which accepts the other fields (but `winerror` only exists on Windows, so that gets complicated).

Another way, which is maybe a little gross, but seems appropriate to me, would be to modify `e` directly before re-raising it:
```py3
except OSError as e:
    if e.errno == errno.EMFILE:
        e.strerror += '. Please try...'
    raise  # with no argument, this re-raises the original exception
```

spinalcordtoolbox/image.py
mguaypaq: I think raising an exception is appropriate here. The `Image` constructor lives in the API layer, so it doesn't have the whole picture, and `sys.exit` would be a bit rude.

If we want to try a non-mmap approach when encountering the "Too many open files" error, I think this would normally be handled in the calling code, which actually has access to the other open files and can close them before retrying. This is only possible by throwing an exception, which can be caught by the calling code. And if it's not caught, at least there'll be a useful suggestion displayed to the user in the traceback.

mguaypaq (approved):
Thanks for the fix. The core seems to be that we should use `Image(..., mmap=False)` when we expect a large number of files could be opened (like, say, in a loop), so that they don't stay open after then call to `Image()`, which is very reasonable. I also like the practical suggestion you added in the error message, in case we run into this problem elsewhere.

spinalcordtoolbox/image.py
joshuacwnewton: Fascinating! Thank you for teaching me about the [`errno`](https://docs.python.org/3/library/errno.html) library. :slightly_smiling_face:

spinalcordtoolbox/image.py
joshuacwnewton: Oh gosh, I had no idea `OSError` had attributes! Thank you for pointing this out to me. :slightly_smiling_face:

spinalcordtoolbox/image.py
```diff
@@ -279,7 +282,22 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
...
         # load an image from file
         if isinstance(param, str):
-            self.loadFromPath(param, verbose)
+            try:
+                self.loadFromPath(param, mmap, verbose)
+            except OSError as e:
+                if str(e) == "[Errno 24] Too many open files":
+                    try:
+                        # Only load images into memory if there are no more available file descriptors
+                        self.loadFromPath(param, False, verbose)
```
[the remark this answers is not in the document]
joshuacwnewton: > Edit: Or is the constrained resource actually memory mappings, even though it's reported as too many open files? Otherwise how does this fix work?

As far as I understand, ever since https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3718, we no longer keep any open file descriptors for `Image` objects (when loaded as `ndarray`s. In other words, it's only when the data array is loaded as a memory mapped array that we get an open file descriptor.

So, my understanding was that, by forcing a data array to be loaded as an `ndarray`, we ensure that we can have 300+ `Image` objects and 0 open file descriptors.

> For another thing, running out of file descriptors is a little bit like running out of memory: unless we actively free up some resources, everything is likely to start failing. For example, opening a file to write the results of whatever computation we're doing will likely fail. Also, import statements for not-already-imported modules would also fail, etc.

That was exactly my thinking in the extended commit message of https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/cf82f856cd30cc5013bdcb39eb5275549dc60825, too. 

I'll revert this solution and stick with the others.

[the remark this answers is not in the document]
mguaypaq: Oh I hadn't seen your lovely detailed commit messages! Sorry, I usually look for those but forgot in this PR.

spinalcordtoolbox/image.py
joshuacwnewton: Works perfectly!

spinalcordtoolbox/image.py
joshuacwnewton: Works perfectly here, too!

joshuacwnewton: > Another way, which is maybe a little gross, but seems appropriate to me, would be to modify e directly before re-raising it

Just out of curiosity, which part of this feels gross to you? Or, phrased another way, which aspect of this proposed change is giving you pause?

mguaypaq: I guess it's an instinctive reaction to mutating the innards of a object that we didn't create, but actually it's a documented public field of `OSError`, so it's really fine. And I can't imagine that a human-facing text attribute like this would have any guarantees that other code relies on. So it's pythonic.

Another option I thought of (which is almost certainly overkill), especially if we want to handle a specific situation without giving up any information along the way, would be to use a custom Exception subclass (something like `class MyException(Exception): pass`) and then to `raise MyException("some message") from e`, using the `raise ... from ...` form of [the `raise` statment](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement). Then, the custom exception is easy to catch (because it can't be some other kind of `OSError`), but it still contains the original `OSError` object in its `__cause__` field, and they both get displayed in a sensible way if uncaught.

Actually after testing this out with a toy example, I see that a `raise` statement inside an `except:` handler already saves the original exception object, but in the `__context__` field instead of `__cause__`. Both get displayed if uncaught, but with a slightly different text about the relationship between the two exceptions.

joshuacwnewton: > Actually after testing this out with a toy example, I see that a `raise` statement inside an `except:` handler already saves the original exception object, but in the `__context__` field instead of `__cause__`. Both get displayed if uncaught, but with a slightly different text about the relationship between the two exceptions.

This makes perfect sense! `__context__` must be tied to the "During handling of the above exception, another exception occurred" text I sometimes see during tracebacks, in which both the parent and child tracebacks are displayed.

In that case, I think I agree even more that we should modify the exception in place:

* If our goal was to display an entirely different error message describing an entirely different contextual error (e.g. "Too many motion-corrected volumes were opened during motion correction"), then I think raising a new exception would make sense?
* But since we're just trying to amend an existing error with useful advice, it feels a bit clunky to display what is basically the same traceback + error message twice, with the second error message being identical except for a bit of extra text.

mguaypaq (approved):

Issue #3883, opened 2022-09-21
3 SCT scripts are hardcoding FSLeyes commands, rather than using `display_viewer_syntax`


joshuacwnewton:
SCT has a function called `display_viewer_syntax`, which will generate commands to open output images in `fsleyes`, `itksnap`, and more, depending on which viewers the user has installed. For example:

![image](https://user-images.githubusercontent.com/16181459/191509358-bc6a11a0-54a1-46c0-b377-c774a31a969a.png)

However, some functions (such as `sct_analyze_lesion`) are using hardcoded `fsleyes` command strings instead:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/464721865ed5f7adbf6a755f8619816c75e8ef75/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L557-L561

This means that if a user doesn't have `fsleyes`, they will be unable to run the printed command. (This is especially an issue for native Windows users, where FSLeyes isn't available, but where they may use ITKSnap instead.)

We should do a quick sweep to make sure each SCT CLI script is using `display_viewer_syntax`. 

From my quick search, I think only 3 scripts are currently doing this?

![image](https://user-images.githubusercontent.com/16181459/191509162-028e2a3b-bafc-46ce-a67e-28b885e2c3e5.png)

joshuacwnewton:
Ah, there's a fourth. `isct_convert_binary_to_trilinear` is hardcoding a `fslview` command, even! 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/464721865ed5f7adbf6a755f8619816c75e8ef75/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py#L160-L162

Goodness, that's a relic from before FSLeyes even existed.

PR #3887, opened 2022-09-22
Enforce `type=int` for arguments that use `choices=(0, 1)`

### LINKED ISSUE #3885: `choices`-based integer arguments must contain `type=int` or else an error will be thrown
While fixing up `isct_convert_binary_to_trilinear` as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3883#issuecomment-1253881044, I ran across this error:

```
sct_smooth_spinalcord: error: argument -r: invalid choice: '1' (choose from 0, 1)
```

The source is this call:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/464721865ed5f7adbf6a755f8619816c75e8ef75/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py#L127-L133

We pass a string `'1'`, but `sct_smooth_spinalcord` expects an int. So, I tried passing an integer `1` instead:

```
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/argparse.py", line 1787, in parse_known_args
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/argparse.py", line 1828, in _parse_known_args
    option_tuple = self._parse_optional(arg_string)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/argparse.py", line 2115, in _parse_optional
    if not arg_string[0] in self.prefix_chars:
TypeError: 'int' object is not subscriptable
```

Oh. Right. `argparse` expects an unparsed list of argument strings, and _not_ typed values.

So, where's the real problem? It's here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/464721865ed5f7adbf6a755f8619816c75e8ef75/spinalcordtoolbox/scripts/sct_smooth_spinalcord.py#L103-L108

But, `sct_smooth_spinalcord` provides only two valid choices: integer 1 and integer 0. Yet, it doesn't have `type=int` specified. So, the string `'1'` or string `'0'` never gets converted to one of the two valid choices.

This means that you can even trigger this error from the command line:

```console
(venv_sct) joshua@tadpole:~/repos/spinalcordtoolbox$ sct_smooth_spinalcord -r 0

[...]

sct_smooth_spinalcord: error: argument -r: invalid choice: '0' (choose from 0, 1)
```

(Note, though, that the default value *is* an integer, which is why this bug has persisted silently for so long. If you don't specify `-r` in your command, then no error will occur.)

We should double-check to make sure that any arguments with integer choices also have `type=int` associated with them.

joshuacwnewton:
## Description

For `argparse`-based arguments, if the argument has e.g. `choices=(0, 1)`, then `type=int` must be set. This is because unparsed arguments are passed as strings (e.g. `'0'`, `'1'`), which must be converted to integers to be considered a valid choice.

I did a project-wide regex search for `choices=.*0, 1.*` (this search also included results for `choices=(0, 1, 2)`). Then, I checked each search result to see if `type=int` was specified. I found two instances of this not being the case (`sct_smooth_spinalcord`, `sct_merge_images`).

This PR fixes both arguments so that they now work as intended. 

## Linked issues

Fixes #3885.
Needed for the PR that will address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3883, because `isct_convert_binary_to_trilinear` has a call to `sct_smooth_spinalcord` that fails.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_merge_images.py
+++ b/spinalcordtoolbox/scripts/sct_merge_images.py
@@ -32,7 +32,7 @@ class Param:
     def __init__(self):
         self.fname_out = 'merged_images.nii.gz'
         self.interp = 'linear'
-        self.rm_tmp = True
+        self.rm_tmp = 1
         self.verbose = 1
         self.almost_zero = 0.00000001
 
@@ -97,7 +97,7 @@ def get_parser():
     misc = parser.add_argument_group('MISC')
     misc.add_argument(
         "-r",
-        type=bool,
+        type=int,
         help='Remove temporary files.',
         required=False,
         default=Param().rm_tmp,
--- a/spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
+++ b/spinalcordtoolbox/scripts/sct_smooth_spinalcord.py
@@ -102,6 +102,7 @@ def get_parser():
         help="Output filename. Example: smooth_sc.nii.gz. By default, the suffix '_smooth' will be added to the input file name."),
     optional.add_argument(
         '-r',
+        type=int,
         choices=[0, 1],
         default=1,
         help="Whether to remove temporary files. 0 = no, 1 = yes"


spinalcordtoolbox/scripts/sct_merge_images.py
```diff
@@ -97,7 +97,7 @@ def get_parser():
     misc = parser.add_argument_group('MISC')
     misc.add_argument(
         "-r",
-        type=bool,
+        type=int,
         help='Remove temporary files.',
         required=False,
         default=Param().rm_tmp,
```
joshuacwnewton, line 103: This one was sort of weird!

The default value for `-r` (`Param().rm_tmp`) was `True`, yet its `choices=(0, 1)`, so I'm not sure how this one ever actually... worked? Shouldn't a boolean value not be a valid choice?

Still, its associated check is:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/43a50029187b4a9512cb58149210d45ad3f3227f/spinalcordtoolbox/scripts/sct_merge_images.py#L194-L196

So, there should be no functional change going from `bool` to `int`.

mguaypaq (approved):
Thanks for the fix, and for searching through the SCT code base for all instances of the problem!

PR #3891, opened 2022-09-26
`install_sct.bat`: Use delayed expansion (`!!`) for `git_ref`

### LINKED ISSUE #3890: Parse-time expansion of `%git_ref%` does not work reliably, causing installer crash
SCT's `install_sct.bat` installer for native Windows installations has the following feature:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/43a50029187b4a9512cb58149210d45ad3f3227f/install_sct.bat#L31-L36

The purpose of this feature is to allow us to pass git refs corresponding to PR heads (so that we can test contributions on Windows):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/43a50029187b4a9512cb58149210d45ad3f3227f/install_sct.bat#L40-L41

However, for some reason, the value of `git_ref` is being expanded as `''` instead of the default value, despite no arguments being passed:

```
### Downloading SCT source code (@ ) to \Users\user\spinalcordtoolbox...
Cloning into 'spinalcordtoolbox'...
warning: Could not find remote branch --single-branch to clone.
fatal: Remote branch --single-branch not found in upstream origin
Failed with error #128.
```

joshuacwnewton:
## Description

This PR fixes an issue where the value of `git_ref` would be set, but not properly expanded. The reason for this is because `git_ref` is set _and expanded_ all within the scope of a larger outer `if` statement, causing [this issue](https://stackoverflow.com/questions/9102422/windows-batch-set-inside-if-not-working).

Because parse-time expansion (`%git_ref%`) won't work, this PR instead uses [execution-time expansion](https://ss64.com/nt/delayedexpansion.html) (`!git_ref!`) to allow the variable to be properly expanded.

## Linked issues

Fixes #3890.


mguaypaq (approved):
TIL about EnableDelayedExpansion!

PR #3894, opened 2022-10-03
Improve handling of case where `algo` is specified alongside `type='label'`

### LINKED ISSUE #3893: label-based registration outputs mat instead of txt file causing crash of ANTs function
### Description

When using `sct_register_multimodal` with the type `label`, the function crashes:

~~~
Description: itk::ERROR: TransformFileReaderTemplate(0x600002cf3a80): Transform IO: MatlabTransformIOTemplate
failed to read file: warp_forward_0.mat
~~~

### Steps to Reproduce

~~~bash
# Download testing data
# go to the test data
cd $SCT_DIR/data/sct_testing_data/t2
# Run command
# note: here we register the image on itself, which is irrelevant in general, but convenient here to show the crash
sct_register_multimodal -i t2.nii.gz -d t2.nii.gz -ilabel labels.nii.gz -dlabel labels.nii.gz -param step=0,type=label,algo=rigid
~~~

```console
julien-macbook:~/code/sct/data/sct_testing_data/t2 $ sct_register_multimodal -i t2.nii.gz -d t2.nii.gz -ilabel labels.nii.gz -dlabel labels.nii.gz -param step=0,type=label,algo=rigid

--
Spinal Cord Toolbox (git-master-00ca428f20b522dd940db46cbf41f1ae233fb5a2)

sct_register_multimodal -i t2.nii.gz -d t2.nii.gz -ilabel labels.nii.gz -dlabel labels.nii.gz -param step=0,type=label,algo=rigid
--


Input parameters:
  Source .............. t2.nii.gz
  Destination ......... t2.nii.gz
  Init transfo ........ 
  Mask ................ 
  Output name ......... 
  Remove temp files ... 1
  Verbose ............. 1

Check if input data are 3D...
Creating temporary folder (/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20220929175346.938519-register-1zhn1y9q)

Copying input data to tmp folder and convert to nii...

--
ESTIMATE TRANSFORMATION FOR STEP #0
Registration parameters:
  type ........... label
  algo ........... rigid
  slicewise ...... 0
  metric ......... MI
  samplStrategy .. None
  samplPercent ... 0.2
  iter ........... 0
  smooth ......... 0
  laplacian ...... 0
  shrink ......... 1
  gradStep ....... 0.5
  deformation .... 1x1x0
  init ........... 
  poly ........... 5
  filter_size .... 5
  dof ............ Tx_Ty_Tz_Rx_Ry_Rz
  smoothWarpXY ... 2
  rot_method ..... pca
Labels src: [[6.2192840576171875, -0.7866058349609375, 24.678329467773438], [6.2192840576171875, 0.2133941650390625, -10.321670532226562]]
Labels dest: [[6.2192840576171875, -0.7866058349609375, 24.678329467773438], [6.2192840576171875, 0.2133941650390625, -10.321670532226562]]
Degrees of freedom (dof): Tx_Ty_Tz_Rx_Ry_Rz
Optimization terminated successfully.
         Current function value: 0.000000
         Iterations: 1
         Function evaluations: 55
Matrix:
 [[ 1.  0.  0.]
 [ 0.  1.  0.]
 [-0.  0.  1.]]
Center:
 [ 6.21928406 -0.28660583  7.17832947]
Translation:
 [[0. 0. 0.]]

--
ESTIMATE TRANSFORMATION FOR STEP #1

Apply transformation from previous step

Parse list of warping fields...

Get dimensions of data...
  60 x 55 x 52 x 1

Apply transformation...

Apply transformation and resample to destination space...
/Users/julien/code/sct/bin/isct_antsApplyTransforms -d 3 -i src.nii -o src_reg.nii -t warp_forward_0.mat -r dest_RPI.nii -n 'BSpline[3]' # in /private/var/folders/s8/4qnm5q1n261ch35b5kkclsb00000gn/T/sct-20220929175346.938519-register-1zhn1y9q
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_register_multimodal.py", line 443, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_register_multimodal.py", line 421, in main
    path_out=path_out)
  File "/Users/julien/code/sct/spinalcordtoolbox/registration/core.py", line 176, in register_wrapper
    '-w'] + warp_forward
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_apply_transfo.py", line 360, in main
    transform.apply()
  File "/Users/julien/code/sct/spinalcordtoolbox/scripts/sct_apply_transfo.py", line 224, in apply
    ] + fname_warp_list_invert + ['-r', fname_dest] + interp, is_sct_binary=True)
  File "/Users/julien/code/sct/spinalcordtoolbox/utils/sys.py", line 414, in run_proc
    raise RuntimeError(output)
RuntimeError: Transform reader for warp_forward_0.mat caught an ITK exception:

itk::ExceptionObject (0x7fe65eb04b20)
Location: "unknown"
File: /Users/runner/work/build_ANTs/build_ANTs/antsbin/ITKv5/Modules/IO/TransformBase/src/itkTransformFileReader.cxx
Line: 144
Description: itk::ERROR: TransformFileReaderTemplate(0x600002cf3a80): Transform IO: MatlabTransformIOTemplate
failed to read file: warp_forward_0.mat
```

</details>

### Suggestion

The crash occurs because the function `isct_antsApplyTransforms` is expecting a file ending with '.txt' suffix. A simple fix is to rename the file:
~~~
cp warp_forward_0.mat warp_forward_0.txt
antsApplyTransforms -d 3 -i src.nii -o src_reg.nii -t warp_forward_0.txt -r dest_RPI.nii -n 'BSpline[3]'
~~~
--> that works

Note: this is already what we do here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c70dad6be6721e52eac93336afda31563b43465f/spinalcordtoolbox/registration/core.py#L433-L437

which is why `sct_register_to_template` does not crash (even though it relies on landmark-based registration), but for some reasons when using `sct_register_multimodal` it does crash. 

I think it should be a very simple fix

joshuacwnewton:
#### GitHub

#### PR contents

## Description

`sct_register_multimodal` performs multi-step registration. To be able to specify multiple parameters corresponding to each step, we use the `-param` argument, which takes a string formatted as `step=0,type=<>,algo=<>:step=1,type=<>,algo=<>` and can continue for as many steps the user wishes.

An ugly error occurs when the user erroneously specifies both `type='label'` and `algo=rigid` in a step:

```bash
# NB This isn't actually valid usage? See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3893#issuecomment-1265503675
sct_register_multimodal -param step=0,type=label,algo=rigid -i t2.nii.gz -d t2.nii.gz -ilabel labels.nii.gz -dlabel labels.nii.gz
```

* `type='label'` will output the transformation in the form of a `.txt` file.
* But, specifying `algo='rigid'` will cause the `.txt` file to incorrectly be renamed to a `.mat` file.
* The `.mat` file is incompatible with further transformation steps, hence the following error:

```
Description: itk::ERROR: TransformFileReaderTemplate(0x600002cf3a80): Transform IO: MatlabTransformIOTemplate
failed to read file: warp_forward_0.mat
```

To fix this, I've added a number of changes:

1. Prioritizing `type='label'`'s rename operation (`.txt`) over `algo='rigid'`'s rename operation (`.mat`)
2. Alternatively, rewriting the 'rename' block of code so that filename extensions are explicitly preserved. (Supersedes fix 1.)
3. Adding a warning when the user erroneously specifies `algo` alongside `type=label`.

## Linked issues

Fixes #3893.

### DIFF
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -424,18 +424,11 @@ def register(src, dest, step, param):
                '\nERROR: ANTs failed. Exit program.\n', 1, 'error')
     else:
         # rename warping fields
-        if step.type in ['label']:
-            # if label-based registration is used --> outputs .txt file
-            warp_forward = 'warp_forward_' + str(step.step) + '.txt'
+        _, _, output_ext = extract_fname(warp_forward_out)
+        if output_ext in ['.txt', '.mat']:
+            warp_forward = 'warp_forward_' + str(step.step) + output_ext
             os.rename(warp_forward_out, warp_forward)
-            warp_inverse = '-warp_forward_' + str(step.step) + '.txt'
-        elif (step.algo.lower() in ['rigid', 'affine', 'translation'] and
-                step.slicewise == '0'):
-            # if ANTs is used with affine/rigid --> outputs .mat file
-            warp_forward = 'warp_forward_' + str(step.step) + '.mat'
-            os.rename(warp_forward_out, warp_forward)
-            warp_inverse = '-warp_forward_' + str(step.step) + '.mat'
-
+            warp_inverse = '-warp_forward_' + str(step.step) + output_ext
         else:
             warp_forward = 'warp_forward_' + str(step.step) + '.nii.gz'
             warp_inverse = 'warp_inverse_' + str(step.step) + '.nii.gz'
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -344,7 +344,7 @@ def register(src, dest, step, param):
     # # landmark-based registration
     if step.type in ['label']:
         if step.algo:
-            printv(f"Specifying 'algo={step.algo}' has no effect for 'type=label' registration.", type='warning')
+            printv(f"Parameter 'algo={step.algo}' has no effect for 'type=label' registration.", type='warning')
         warp_forward_out, warp_inverse_out = algorithms.register_step_label(
             src=src,
             dest=dest,
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -15,6 +15,8 @@
 
 import os
 
+import logging
+
 from spinalcordtoolbox.registration import algorithms
 
 from spinalcordtoolbox.image import Image, add_suffix, generate_output_file
@@ -22,6 +24,8 @@ from spinalcordtoolbox.utils.fs import extract_fname, rmtree, tmp_create
 from spinalcordtoolbox.utils.shell import printv
 from spinalcordtoolbox.utils.sys import run_proc
 from spinalcordtoolbox.scripts import sct_apply_transfo
+
+logger = logging.getLogger(__name__)
 
 
 def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg='', fname_dest_seg='', fname_src_label='',
@@ -339,6 +343,8 @@ def register(src, dest, step, param):
 
     # # landmark-based registration
     if step.type in ['label']:
+        if step.algo:
+            logger.warning(f"Specifying 'algo={step.algo}' has no effect for 'type=label' registration.")
         warp_forward_out, warp_inverse_out = algorithms.register_step_label(
             src=src,
             dest=dest,
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -344,7 +344,7 @@ def register(src, dest, step, param):
     # # landmark-based registration
     if step.type in ['label']:
         if step.algo:
-            logger.warning(f"Specifying 'algo={step.algo}' has no effect for 'type=label' registration.")
+            printv(f"Specifying 'algo={step.algo}' has no effect for 'type=label' registration.", type='warning')
         warp_forward_out, warp_inverse_out = algorithms.register_step_label(
             src=src,
             dest=dest,
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -424,17 +424,18 @@ def register(src, dest, step, param):
                '\nERROR: ANTs failed. Exit program.\n', 1, 'error')
     else:
         # rename warping fields
-        if (step.algo.lower() in ['rigid', 'affine', 'translation'] and
+        if step.type in ['label']:
+            # if label-based registration is used --> outputs .txt file
+            warp_forward = 'warp_forward_' + str(step.step) + '.txt'
+            os.rename(warp_forward_out, warp_forward)
+            warp_inverse = '-warp_forward_' + str(step.step) + '.txt'
+        elif (step.algo.lower() in ['rigid', 'affine', 'translation'] and
                 step.slicewise == '0'):
             # if ANTs is used with affine/rigid --> outputs .mat file
             warp_forward = 'warp_forward_' + str(step.step) + '.mat'
             os.rename(warp_forward_out, warp_forward)
             warp_inverse = '-warp_forward_' + str(step.step) + '.mat'
-        elif step.type in ['label']:
-            # if label-based registration is used --> outputs .txt file
-            warp_forward = 'warp_forward_' + str(step.step) + '.txt'
-            os.rename(warp_forward_out, warp_forward)
-            warp_inverse = '-warp_forward_' + str(step.step) + '.txt'
+
         else:
             warp_forward = 'warp_forward_' + str(step.step) + '.nii.gz'
             warp_inverse = 'warp_inverse_' + str(step.step) + '.nii.gz'
--- a/testing/cli/test_cli_sct_register_multimodal.py
+++ b/testing/cli/test_cli_sct_register_multimodal.py
@@ -145,5 +145,6 @@ def test_sct_register_multimodal_with_labels(capsys, tmp_path, algo):
                                   '-ofolder', str(tmp_path)])
     for file in ['t2_dest_reg.nii.gz', 't2_src_reg.nii.gz', 'warp_t22t2.nii.gz']:
         assert os.path.isfile(tmp_path / file)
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/registration/core.py
```diff
@@ -424,17 +430,11 @@ def register(src, dest, step, param):
...
-            warp_forward = 'warp_forward_' + str(step.step) + '.txt'
+        _, _, output_ext = extract_fname(warp_forward_out)
+        if output_ext in ['.txt', '.mat']:
+            warp_forward = 'warp_forward_' + str(step.step) + output_ext
             os.rename(warp_forward_out, warp_forward)
-            warp_inverse = '-warp_forward_' + str(step.step) + '.txt'
+            warp_inverse = '-warp_forward_' + str(step.step) + output_ext
         else:
             warp_forward = 'warp_forward_' + str(step.step) + '.nii.gz'
             warp_inverse = 'warp_inverse_' + str(step.step) + '.nii.gz'
```
joshuacwnewton, line 440: I've added extended commit messages to my commits, but there's one snippet I wrote that I want to explicitly want to highlight here:

> That said, this section of code still gives me pause:
> 
> - Why is it that the 'else:' case renames both warp_forward and warp_inverse, but the 'if:' case only renames 'warp_forward'?
> - In that case, why even create the 'warp_inverse' variable at all?
> - Does the 'warp_inverse' file just not exist?
> - If we know that the 'warp_inverse' file doesn't get created in the '.mat' and '.txt' cases, then why not return e.g. 'None' for such a filename?
> 
> This requires a deeper dive into how the registration commands determine their output files, though, and I don't want to balloon the scope of this PR too much.

I'm not sure if now's the best time to refactor this section of code, but it definitely confused me / set off red flags when I first tried to read through it.

spinalcordtoolbox/registration/core.py
```diff
@@ -339,6 +343,8 @@ def register(src, dest, step, param):
 
     # # landmark-based registration
     if step.type in ['label']:
+        if step.algo:
```
jcohenadad, line 346: 👍

testing/cli/test_cli_sct_register_multimodal.py
```diff
@@ -118,3 +118,33 @@ def test_sct_register_multimodal_with_softmask(tmp_path):
     # im_t2_seg, _, _ = deep_segmentation_spinalcord(Image(fname_t2), contrast_type='t2', ctr_algo='svm')
     # dice_score_t1_reg = dice(im_t2_seg.data, im_t1_reg_seg.data)
     # assert dice_score_t1_reg > 0.9
+
+
+@pytest.mark.parametrize('algo', [',algo=rigid', ''])
+def test_sct_register_multimodal_with_labels(caplog, tmp_path, algo):
```
jcohenadad, line 124: ❤️

jcohenadad (approved):
Thank you so much @joshuacwnewton !

spinalcordtoolbox/registration/core.py
joshuacwnewton: Ah, one thing to note here: 

Even if `algo` isn't explicitly specified in the `-param` command, `sct_register_multimodal` will default to `algo=syn`, which meaning that the warning will always trigger. For example:

```
sct_register_multimodal -param step=0,type=label
```

image

Maybe the warning is still worth including?

But, at the same time, I think it speaks to what I wrote in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3893#issuecomment-1265512260, too, in that it's a bit weird that `type='label'` acts like an `algo`, but isn't.

jcohenadad: if we decide to attack this beast, we should also include the other (related) issues, eg: when selecting algo=syn, the field 'poly' has nothing to do with it. There are many other examples like that...

mguaypaq: From your screencap, it looks like the warning is coloured, but then I don't understand your commit message:

> Is printv the right function for this? We're in the API, so maybe
logging should be used instead. But, I like the colors...

Is the screencap from a previous version of the code?

Regardless, I think it's fine to use `printv` in this case, since the rest of the function uses `printv` a lot. Consistency is very important! It matters most within a single function, and a lot within a single file. If/when this part of the code gets an overhaul, we can decide how to deal with the fact that it's calling `printv` at the API layer. Possibly the chatty parts of the code would move out of the API; or possibly all of it would get converted to logging instead.

(Sorry for the self-contradiction with my previous statements about `printv`! It's just that _either_ `printv` _or_ logging is better than a mix of both within the same function/file.)

mguaypaq (approved):
Thanks for making this code incrementally more robust, adding a warning for the users, and writing a test! You're right that the registration code in general looks like it could use a lot more work, but this PR is a step in the right direction.

spinalcordtoolbox/registration/core.py
joshuacwnewton: Ah gosh, sorry. My commit message has made a mess of things. Let me try and clear up what happened:

* I originally used `printv`, thinking it was the only way to get yellow text (hence the commit message)
* I then realized that Pytest's `caplog.text` only captures logging warnings/messages, so I swapped out `printv` for `logging`, then amended the commit (while forgetting to amend the commit message!)
* I _then_ realized that `logger.warning` actually _does_ print yellow text too (which is what you see in the screencap). The fact that they both print yellow text makes the commit message seem even more misleading.

I'm happy to switch to using `printv` in the API function, I'd just need to find a way for Pytest to pick up that sort of output. :)

spinalcordtoolbox/registration/core.py
mguaypaq: These are great questions, but they would take more digging than we probably want to do in this PR, as you said.

PR #3895, opened 2022-10-03
Enforce use of `display_viewer_syntax` in four scripts using hardcoded `fsleyes` commands

### LINKED ISSUE #3883: 3 SCT scripts are hardcoding FSLeyes commands, rather than using `display_viewer_syntax`

joshuacwnewton:
## Description

Most SCT scripts use a function called `display_viewer_syntax` at the end of processing. This function searches for the presence of image viewing software like `fsleyes` or `itksnap`, then outputs commands that let the user open output files in those viewers.

However, some scripts still hardcode a command for `fsleyes`/`fslview`. So, this PR replaces those commands with `display_viewer_syntax`.

> ((NB: The `isct_convert_binary_to_trilinear` script is looking a bit neglected. I've made sure it works, because it's listed in our `setup.py`, and thus is available to our users. That said, the script doesn't use argparse, and its main function isn't properly exposed, so we don't currently test it. 
> 
> We might want to reflect on whether or not these `isct_` scripts are worth maintaining. If they are, we should fix them up and turn them into a properly formatted + tested `sct_` scripts. If they aren't, then we should move them to `dev/`.))

## Linked issues

Fixes #3883.

### DIFF
--- a/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
+++ b/spinalcordtoolbox/scripts/isct_convert_binary_to_trilinear.py
@@ -23,6 +23,7 @@ import numpy as np
 from spinalcordtoolbox.image import Image, generate_output_file, convert
 from spinalcordtoolbox.utils.sys import init_sct, __data_dir__, printv
 from spinalcordtoolbox.utils.fs import tmp_create, check_file_exist, rmtree, extract_fname
+from spinalcordtoolbox.utils.shell import display_viewer_syntax
 
 from spinalcordtoolbox.scripts import sct_resample, sct_smooth_spinalcord
 
@@ -160,8 +161,7 @@ def main():
     printv('\nFinished! Elapsed time: ' + str(int(np.round(elapsed_time))) + 's')
 
     # to view results
-    printv('\nTo view results, type:')
-    printv('fslview ' + file_data + ' ' + file_data + suffix + ' &\n')
+    display_viewer_syntax([fname_data, fname_out], verbose=verbose)
 
 
 # printv(usage)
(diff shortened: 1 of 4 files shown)


mguaypaq (approved):
Thanks for the fix, including figuring out the missing grayscale and opacity parameters.

> We might want to reflect on whether or not these isct_ scripts are worth maintaining. If they are, we should fix them up and turn them into a properly formatted + tested sct_ scripts. If they aren't, then we should move them to dev/.))

Good question, this is worth discussing at the next dev meeting. In particular, your bugfix of the `isct_convert_binary_to_trilinear` means that this script has been completely broken for quite some time, so I can't imagine it gets much use.

_Out of scope for this PR:_ I'm a bit surprised that `sct_smooth_spinalcord` produces a `.nii.gz` even when the input is a plain `.nii` file. Is this intended? My impression from the existence of [`Image.add_suffix()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1604a48c79bc96b3aea405ad54db1f40a877e27/spinalcordtoolbox/image.py#L1402-L1417) is that we usually try to match input and output formats. As a further aside, `Image.add_suffix()` uses an almost-but-not-quite-identical implementation of [`utils.sys.extract_fname()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f1604a48c79bc96b3aea405ad54db1f40a877e27/spinalcordtoolbox/utils/fs.py#L83-L95), which puzzles me.

PR #3906, opened 2022-10-11
Switch to higher-quality "Tools used" graph on the studies page

### LINKED ISSUE #3897: Clean up the "Pathology" and "Tools" columns in "studies_citing_SCT"
In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3896, we began displaying statistics graphs from our "studies_citing_SCT" Google Sheets spreadsheet. For example:

image

But, I noticed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3857#issuecomment-1261200743 that there were some issues with the data we're keeping:

> One thing we might want to do before publishing is to go through the Pathology and Tools charts and make sure they've been entered and parsed correctly. Some immediate thoughts:
> 
> 1. We use acronyms for pathology, but it might not be 100% clear what each of these is referring to, so we may want to swap e.g. HC for Healthy Control for clarity purposes, assuming it fits space-wise on the chart.
> 2. Many of the entries for "tools used" are something like "sct_register_to_template, sct_dmri_compute_dti", but I don't think these are parsed as individual tools for the chart, meaning we get a lot of duplicate/overlapping one-off entries. So, we might want to figure out how to get Google Sheets to properly parse comma-delimited entries as individual counts for the chart.
> 3. Also, a lot of "tools used" entries are tool names ("sct_X") while others are descriptors ("Segmentation", "CSA") so we might want to unify these, too.

This is a follow-up issue to address these concerns.

joshuacwnewton:
## Description

Currently, we display a pie chart of "Tools used" from the [studies_citing_SCT](https://docs.google.com/spreadsheets/d/1Utmf-btG7uXmu7RCBBx7fIKobm6yvXstX2Psc-UR2HA/edit?usp=sharing) spreadsheet. However, this pie chart has two main problems:

1. It contains a mix of both descriptors (e.g. Segmentation, Registration) and literal tool names (e.g. sct_deepseg, sct_register_to_template)
2. It contains comma-delimited entries (e.g. "labeling, moco") that are treated as their own entry, rather than parsed as individual parts.

To fix these problems, I created a new column in the spreadsheet, copied the existing data, and applied the following changes:

1. Replaced tool names (e.g. sct_deepseg) with their broader purpose (e.g. Segmentation)
2. In a separate postprocessing sheet, I concatenated the cells, then split them based on `','`, so that each individual part is counted as a unique entry.

The result is a pie graph with no redundant entries:

image

This PR simply updates the link to this new graph.

## Linked issues

Fixes #3897. (NB: I also did a find and replace for HC -> Healthy Control, which fixes the "Pathology" column, too.)

### DIFF
--- a/documentation/source/overview/studies.rst
+++ b/documentation/source/overview/studies.rst
@@ -12,7 +12,7 @@ The graphs below summarize the pathologies, tools, species, and count per year o
 
        <iframe width="600" height="371" seamless frameborder="0" scrolling="no" src="https://docs.google.com/spreadsheets/d/e/2PACX-1vSwyEvoiTOMflrJveD277xWYSb_1QSwkpxWsZoMSucgHBS7BHcgfvzGG21--1bLRFO_DIV4EhL9lBl2/pubchart?oid=1220039972&amp;format=interactive"></iframe>
 
-       <iframe width="600" height="371" seamless frameborder="0" scrolling="no" src="https://docs.google.com/spreadsheets/d/e/2PACX-1vSwyEvoiTOMflrJveD277xWYSb_1QSwkpxWsZoMSucgHBS7BHcgfvzGG21--1bLRFO_DIV4EhL9lBl2/pubchart?oid=366269532&amp;format=interactive"></iframe>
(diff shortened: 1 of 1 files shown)


jcohenadad (approved):
Fantastic! thank you so much!

The only remaining duplicates I see are: 'Spinal cord compression' which can fall into 'DCM' or 'SCI'.

So: we need to figure out what this '1.4%' is (traumatic SCI or non-traumatic DCM), and then put it in the right category.

also: we might want to rename 'DCM' for 'Non-traumatic compression'

joshuacwnewton:
Ah, good call! Thankfully, the % of papers is small so it shouldn't take too long to check, and this work can be done directly in the spreadsheet. I'll take a look today. :)

PR #3907, opened 2022-10-11
Refactor data downloading API to remove dependency on CLI script

### LINKED ISSUE #3902: API for `sct_download_data`
From https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2959#issuecomment-1270851866

Goal: ensure that a dataset can be downloaded purely via the API, without the need to touch the CLI script.

Project:

*    Move dataset [name/URL dict](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2b62a4ddfcd89b3fb18db90b3c926c1dd82c0ad5/spinalcordtoolbox/scripts/sct_download_data.py#L29) to the API
*    Ensure that it's possible to pass a dataset name to an API function and download a dataset
* Replace `subprocess.Popen(['sct_download_data', ...])` calls with `sct_download.download()` (or whatever we end up calling the function)

joshuacwnewton:
## Description

Currently, it is necessary to call the `sct_download_data` CLI script in order to download a dataset using only its name.

This PR reduces the role of the CLI script, moving functionality into the API instead, and then removes the only instance of an `sct_download_data.main()` call.

(This PR also fixes up a few small linting warning/errors in the files that it touches.)

## Linked issues

Fixes #3902.


spinalcordtoolbox/download.py
```diff
@@ -195,3 +349,18 @@ def install_data(url, dest_folder, keep=False):
...
     shutil.rmtree(extraction_folder)
+
+
+def install_named_dataset(dataset_name, dest_folder=None, keep=False):
+    """
+    A light wrapper for the 'install_data' function to allow downstream consumers to download
+    datasets using only the dataset's name (i.e. without needing to access DATASET_DICT fields).
+    """
+    if dataset_name not in DATASET_DICT.keys():
+        raise ValueError(f"{dataset_name} is not contained in list of datasets. Choose from: {DATASET_DICT.keys()}")
```
kousu, line 360: A perhaps more readable error:

```suggestion
        raise ValueError(f"{dataset_name} is not contained in list of datasets. Choose from:\n\n {pprint.pformat(set(DATASET_DICT.keys()))}")
```

kousu (approved):
Nice and smooth

spinalcordtoolbox/download.py
joshuacwnewton: Ah, good catch! I didn't think to check the output of printing a `keys` object as-is:

Before:

image

After:

image

kousu: It doesn't look as pretty as I was hoping and `dict_keys()` isn't as distracting as I thought it might be.

This is good to merge either way.

joshuacwnewton: As a compromise, I think I might just use `list(DATASET_DICT.keys())` to remove the `dict_keys` bit but preserve the condensed list.

image

joshuacwnewton: Oh! TIL that sorting in Python treats capital letters as a different set than lowercase letters. We get the same sorting even using `sorted()`.

joshuacwnewton: Neat. In Python 3.3+, you can use [`key=str.casefold`](https://stackoverflow.com/a/10269828) to take care of this.

joshuacwnewton:
The failure here is due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3918, which I believe will be fixed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3919, so I'm going to merge this using admin privileges.

PR #3908, opened 2022-10-11
Remove unused `isct_` scripts from SCT's CLI and test suite

### LINKED ISSUE #3904: Move unused `isct_` scripts into `spinalcordtoolbox-dev`
In https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3884, we removed a handful of scripts that had become neglected/broken because they weren't exposed to users via `setup.py`. 

As a follow-up to that issue, we also noticed that there were a few additional scripts that _were_ included in `setup.py`, but were still unused/neglected:

- `isct_convert_binary_to_triliinear`
- `isct_minc2volume-viewer`

These two scripts have an `isct_` prefix, meaning that they're intended to be used internally, and not meant for our users. 

However, these scripts aren't actually used inside of SCT, and haven't recently been used by NeuroPoly developers/students, either. (This is demonstrated by the fact that, in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3895, we noticed that `isct_convert_binary_to_trilinear` was completely broken, and could not be run without error.)

During the October 6th, 2022 SCT meeting, we decided that these scripts are not critical, and can be moved from the main SCT repo into the `spinalcordtoolbox-dev` repo.

joshuacwnewton:
## Description

In our October 6th, 2022 dev meeting, we took at look at 2 internal CLI scripts that are no longer in use, and decided that these scripts can be archived in `spinalcordtoolbox-dev`.

This PR moves the scripts to https://github.com/spinalcordtoolbox/spinalcordtoolbox-dev/commit/e49a827533d244e5172f244c0768be830079ccda, and removes any mention of them within the repository (`setup.py`, `testing/`).

_NB: The "no_args" test used to be written in a convoluted matter to try and account for these old scripts. Now that they've been removed, the test has been greatly simplified._

## Linked issues

Fixes #3904.


joshuacwnewton:
> Thanks for the cleanup! I think the documentation is failing to build because the two scripts are still [referenced in command-line.rst](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/jn/3904-move-unused-isct_scripts_to_dev/documentation/source/user_section/command-line.rst#internal-commands). (I didn't see any other references to them in the repository.)

Ah! Good catch! I had done a project-wide search for the script names, but I think I had accidentally toggled `.py` only, thereby excluding our documentation. Oops. :sweat:

PR #3915, opened 2022-10-13
Move `isct_test_ants` functionality into `sct_check_dependencies`

### LINKED ISSUE #3905: Move `isct_test_ants` functionality from the CLI script into `sct_check_dependencies` / `testing/`
As part of our analysis of existing `isct_` scripts (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3904), we also took a look at one other `isct_` script called [`isct_test_ants`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/74d4367fea98ab27bbabf7db36c25de1251b48fa/spinalcordtoolbox/scripts/isct_test_ants.py).

This script isn't really a CLI script in the traditional sense, and instead acts more like a `pytest` test that validates the `isct_antsRegistration` binary that ships with SCT. (The test [performs registration](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2b62a4ddfcd89b3fb18db90b3c926c1dd82c0ad5/spinalcordtoolbox/scripts/isct_test_ants.py#L86-L95) using dummy data, then [checks the DICE score](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2b62a4ddfcd89b3fb18db90b3c926c1dd82c0ad5/spinalcordtoolbox/scripts/isct_test_ants.py#L116-L125) of the registered images, which is very similar to the sort of test you might find for, say, `sct_register_multimodal`.)

In addition, right now this test is only run via `sct_check_dependencies`, and isn't actually something a user or developer would call via the command line. So, there isn't much point to it being its own CLI script.

There are a number of ways we could handle this:

1. The most straightforward way to handle this would be to simply fold the functionality into `sct_check_dependencies`.
    * This would result in no change in user experience, but would remove `isct_test_ants` from the list of scripts exposed to users.
    * However, this would be storing test-related code in SCT's Python package, when all of our other tests are currently stored in `testing/`. 
2. So, alternatively, we could move the test into `testing/` and have it run via `pytest`. 
    * However, that would mean that the ANTs binary wouldn't be tested when users run `sct_check_dependencies`.
    * Maybe this is okay, though? Maybe the responsibility of `sct_check_dependencies` should only be to check the _integrity_ of the binaries so that we know they were downloaded correctly (see [#2489 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2489#issuecomment-1270853033)), rather than the _functionality_ of the binaries (which is more the responsibility of `pytest`).

My suggestion would be to start with 1., then later on when we implement integrity-checking for downloaded data, replace/remove the ANTs test from `sct_check_dependencies`.

joshuacwnewton:
## Description

Following PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3908, there is now one remaining `isct_` script called `isct_test_ants`. This script isn't meant to be run by users, and in practice, is only used as a part of the `sct_check_dependencies` command.

This PR removes the `isct_test_ants` script and turns it into a function instead. Then, in `sct_check_dependencies`, this PR replaces the `run_proc('isct_test_ants')` call with a call to the new function.

----

Notably, however, I encountered one issue with the change: `run_proc` with `verbose=0` silences all output. But, switching to a function call (rather than a subprocess call) results in new, noisy output:

image

So, this PR adds an additional 4 commits to make sure that the output is properly silenced when `verbose=0` is set. 

image

## Linked issues

Fixes #3905.


spinalcordtoolbox/utils/sys.py
```diff
@@ -122,7 +122,7 @@ def set_loglevel(verbose):
     """
     dict_log_levels = {
         '0': 'WARNING', '1': 'INFO', '2': 'DEBUG',  # Older scheme
-        'False': 'INFO', 'True': 'DEBUG',           # Newer scheme (See issue #2676)
+        'False': 'WARNING', 'True': 'DEBUG',        # Newer scheme (See issue #2676)
```
joshuacwnewton, line 125: This change (aa1cc3f) requires a bit of context, I think:

* In issue [#2676](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2676), a proposal was made to switch from `-v {0,1,2}` to `-v {True/False}`. 
    * This effort was ultimately abandoned due to the complexity of verbosity in SCT -- it's not easy to cleanly map a 3-tier verbosity system to a 2-tier verbosity system. 
    * But, there are still remnants of this effort in this codebase: here in `sys.py`, we attempt to translate `True/False` into `DEBUG`/`WARNING`.
* While most scripts continue to use the 3-tier `{0,1,2}` system, `sct_check_dependencies` uses the 2-tier `{True/False}` verbosity system instead, confusingly called `-complete`

    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d17e52f90a971cb778d93d88306b102254e9ba2f/spinalcordtoolbox/scripts/sct_check_dependencies.py#L184-L187

* Because `False` maps to `INFO`, then it means that `INFO` messages will be shown _regardless of the verbosity level_. This probably isn't great! My assumption would be that `-verbose False` should map to `-verbose 0`, and thus shouldn't `INFO` messages.
* By making this change, it correctly silences any `logging` output globally for `sct_check_dependencies`

joshuacwnewton: Alternatively, we could try to unify scripts around the 3-tier verbosity system, and remove `False`/`True` from `sys.py` entirely.

mguaypaq (commented):
Thanks for the PR. I agree that we shouldn't be exposing `isct_*` scripts to the users, but I wonder if bringing the test "in-process" is the right approach, given that we're getting some problems with chatty functions, and ([elsewhere](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3916#discussion_r1003609419)) interactions between different tests. Maybe we should instead be factoring out more of the tests into their own script/executable, so that we can properly sandbox them.

Of course we should fix our non-verbose-respecting functions, but I think we'd want `sct_check_dependencies` to not assume that everything is well-behaved (because it's most useful when there are bugs!). And not every chatty test is necessarily something that we can fix, for example just today I noticed this chattiness of the matplotlib test on my laptop:

```
Check if figure can be opened with matplotlib.......Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
libGL error: MESA-LOADER: failed to open iris: /usr/lib/dri/iris_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: iris
libGL error: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
[OK] (Using GUI backend: 'QtAgg')
```

I'm going to leave this as "Comment" rather than "Approve" because I think we'll want a bit more discussion before merging.

spinalcordtoolbox/utils/sys.py
joshuacwnewton: As for the scope of this PR, the only problematic `False/True` script is `sct_check_dependencies` itself. 

(Because `verbose=False`, the log level for the root logger gets set to `INFO`, which then propagates to the inner scripts.)

joshuacwnewton:
> Thanks for the PR. I agree that we shouldn't be exposing `isct_*` scripts to the users, but I wonder if bringing the test "in-process" is the right approach, given that we're getting some problems with chatty functions, and ([elsewhere](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3916#discussion_r1003609419)) interactions between different tests. Maybe we should instead be factoring out more of the tests into their own script/executable, so that we can properly sandbox them.
> 
> Of course we should fix our non-verbose-respecting functions, but I think we'd want `sct_check_dependencies` to not assume that everything is well-behaved (because it's most useful when there are bugs!). And not every chatty test is necessarily something that we can fix, for example just today I noticed this chattiness of the matplotlib test on my laptop:

I very much agree with what @mguaypaq has written here, and spoke to him in private to quickly brainstorm some alternatives.

My proposal, then, is the following: I think we should turn the ANTs tests/GUI tests/etc. into pytest tests. But, instead of only running these tests in our usual test suite, we would also use `run_proc("sct_testing", os.path.join(__sct_dir__, path_to_test))` as a way of calling specific pytest tests via sct_check_dependencies.

That way, `sct_check_dependencies` would essentially act as a way to run a subset of important tests from the test suite, rather than being a completely separate testing entity as it currently is now.

I think doing so would solve all four problems described in this PR and the associated issue:

1. The `isct_test_ants` script is exposed to users when it shouldn't be.
2. Several `sct_check_dependencies` testing functions can be noisy when run in-process.
3. There is a lack of isolation between conflicting checks (e.g. matplotlib/pyqt5)
4. There is testing logic stored in `sct_check_dependencies`/API, rather than in testing/ where it should be.

As such, I'm going to close this PR for now, then open a new PR containing `pytest` versions of these tests.

PR #3917, opened 2022-10-14
Exclude original file extension (e.g. `.nii.gz`) from the output `.csv` filename

### LINKED ISSUE #3913: Minor fix: naming convention output by `sct_get_centerline`
Running`sct_get_centerline -i input.nii.gz -c contrast` yields the following output: 
* `input_centerline.nii.gz`
* `input_centerline.nii.gz.csv`. \
It is a little bit confusing to have a NIFTI extension preceding a CSV extension; could be worth changing such that the output is named as follows:
* `input_centerline.nii.gz`
* `input_centerline.csv`.

joshuacwnewton:
## Description

The SCT CLI script `sct_get_centerline` outputs 2 files: An `.nii.gz` file and a `.csv` file. 

The filename for the `.csv` file is derived from the filename chosen for the `.nii.gz` file:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d17e52f90a971cb778d93d88306b102254e9ba2f/spinalcordtoolbox/scripts/sct_get_centerline.py#L181

However, simply adding `.csv` to the filename results in outputs like `t2s_centerline.nii.gz.csv`, rather than the expected `t2s_centerline.csv`. 

So, this PR properly excludes the original file extension from the `.csv` filename.

## Linked issues

Fixes #3913.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_get_centerline.py
+++ b/spinalcordtoolbox/scripts/sct_get_centerline.py
@@ -178,7 +178,9 @@ def main(argv: Sequence[str]):
 
     # save centerline as nifti (discrete) and csv (continuous) files
     im_centerline.save(file_output)
-    np.savetxt(file_output + '.csv', arr_centerline.transpose(), delimiter=",")
+    # Replace `.nii.gz` extension with `.csv` extension
+    path_output, filename_output, _ = extract_fname(file_output)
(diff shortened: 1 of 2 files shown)


mguaypaq (approved):
Thanks for the fix and the test!

PR #3919, opened 2022-10-14
Upgrade Ubuntu 18.04 -> 20.04/22.04 in test suite

### LINKED ISSUE #3918: CentOS Stream 8 CI jobs are being cancelled after some time (with no logs to debug)
Example: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/3250781727/jobs/5334915466

Context:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d17e52f90a971cb778d93d88306b102254e9ba2f/.github/workflows/tests.yml#L165-L170

The error message that GitHub gives is:

```
CentOS Stream 8
GitHub Actions has encountered an internal error when running your job.
```

There doesn't seem to be any incidents @ http://githubstatus.com.

I might try switching from `ubuntu-18.04` to `ubuntu-20.04`, given that [support for 18.04 will be removed in Dec 2022](https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/) anyway.

joshuacwnewton:
## Description

In our test suite, we have two kinds of jobs:

1. OSs natively supported by GitHub Actions
2. OSs that must be run via docker containers.

For this 2nd kind, we use `ubuntu-18.04` as the base runner, then run `docker` on top of that using the `container:` keyword. 

However:

* The CentOS Stream 8 CI job is having some troubles, and I want to test whether upgrading to `ubuntu-20.04` fixes this.
* Additionally, support for 18.04 will be removed on Dec 1, 2022. And, ArchLinux is already conducting a brownout for `ubuntu-18.04`:
   ```
   This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on December 1st, 2022. For more details, see 
   ```

## Linked issues

Fixes #3918.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -383,7 +383,7 @@ jobs:
       run:
         shell: cmd
     steps:
-    - uses: actions/checkout@v3ush
+    - uses: actions/checkout@v3
     - name: Install SCT
       # NB: In real-world usage, it's assumed that the user would download the
       # install_sct.bat file from the Downloads page, and _not_ the full repo.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,7 +53,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container: archlinux
     steps:
       - name: Dependencies
@@ -76,7 +76,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container: debian:sid
     steps:
       - name: Dependencies
@@ -99,7 +99,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container: debian:testing
     steps:
       - name: Dependencies
@@ -122,7 +122,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container: debian:10
     steps:
       - name: Dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -164,7 +164,7 @@ jobs:
 
   centos-8:
     name: CentOS Stream 8
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
joshuacwnewton, line 167: Well what do you know, the [CentOS Stream 8 job](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/3250896472/jobs/5335180455) is running and hasn't been canceled. Hooray! :)

.github/workflows/tests.yml
```diff
@@ -145,7 +145,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: ubuntu-18.04  # use an older ubuntu's kernel to closer emulate the older Debian
+    runs-on: ubuntu-20.04
```
joshuacwnewton, line 148: I'm going to also point a manual run of `tests.yml` at this PR's branch so that I can preemptively test the remaining OSs, too (i.e. the ones that aren't tested by default on PRs).

.github/workflows/tests.yml
joshuacwnewton: Just to make sure this wasn't a weird fluke/quirk, I'm going to try and re-run CentOS Stream 8 jobs on the other PRs.

joshuacwnewton: Ah, it looks like it was just a momentary fluke. Those jobs are now running, too, despite still using `ubuntu-18.04`.

I'll close the original issue, but I'll keep this PR open, because this is something we'll have to do eventually this year anyway.

joshuacwnewton: Re-running Stream 8 jobs on other PRs seems to work at first, but again, the jobs get cancelled midway through (with no logs to show for it).

.github/workflows/tests.yml
joshuacwnewton: Everything here passed, although we're getting a warning here:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout
```

I presume this is just a matter of going from `actions/checkout@v2` to `actions/checkout@v3`

.github/workflows/tests.yml
joshuacwnewton: Re-running the Stream 8 jobs a _third_ time, hours later, does in fact lead to passes on all PRs. :facepalm:

joshuacwnewton: Just kidding... 4 days later, we're getting Stream 8 jobs cancelled.

I'm almost certain at this point, though, that it's due to the `ubuntu-18.04` brownout, and that this PR fixes the issue.

.github/workflows/tests.yml
```diff
@@ -299,7 +299,7 @@ jobs:
     - name: Use unix line endings
       shell: bash
       run: |
-        # Github's actions/checkout@v2 when run on Windows mangles the line-endings to DOS-style
+        # Github's actions/checkout@v3 when run on Windows mangles the line-endings to DOS-style
```
mguaypaq, line 302: The linked issue seems to still be revelant with `v3`, so updating this comment seems fine :+1:

mguaypaq (approved):
LGTM

PR #3929, opened 2022-11-01
Ensure bundled GLIBC is up-to-date in order to resolve `MESA-LOADER` errors

### LINKED ISSUE #3927: Non-fatal `libGL error: failed to load driver` errors are thrown during `sct_check_dependencies`
Originally reported by @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3915#pullrequestreview-1157166675:

> Just today I noticed this chattiness of the matplotlib test on my laptop:
> 
> ```
> Check PropSeg compatibility with OS ................[OK]
> Check if figure can be opened with matplotlib.......Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
> libGL error: MESA-LOADER: failed to open iris: /usr/lib/dri/iris_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
> libGL error: failed to load driver: iris
> libGL error: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
> libGL error: failed to load driver: swrast
> [OK] (Using GUI backend: 'QtAgg')
> Check if figure can be opened with PyQt.............[OK]
> ```

joshuacwnewton:
## Description

For Ubuntu 22.04 (and possibly other OSs), in `sct_check_dependencies`, the matplotlib/pyqt5 checks will throw a non-fatal error about missing `.so` files that should be in `/usr/lib/dri`. 

```
libGL error: MESA-LOADER: failed to open iris: /usr/lib/dri/iris_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
libGL error: failed to load driver: iris
libGL error: MESA-LOADER: failed to open swrast: /usr/lib/dri/swrast_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
```

However (on my machine) these files do exist! They're just in `/usr/lib/x86_64-gnu-linux/dri/` instead of `/usr/lib/dri`.

Separately, there is another file in SCT's venv called `libstdc++.so`, which is [tied to the loading of the above `.so` driver files](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927#issuecomment-1298613727). By default, the version installed is `6.0.29`. But, when this file is updated to `6.0.30`, suddenly `MESA-LOADER` can successfully locate the needed `.so` files, thus eliminating the errors.

So, this PR introduces a conda command to update the GLIBC version before installing SCT's dependencies.

## Linked issues

Fixes #3927.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -620,6 +620,9 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
   exit 1
 fi
 
+# Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
+conda install -y -c conda-forge libstdcxx-ng
+
 # Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
 python -m pip install -U "pip!=21.2.*"
 
--- a/install_sct
+++ b/install_sct
@@ -621,7 +621,9 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
 fi
 
 # Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
-conda install -y -c conda-forge libstdcxx-ng
+if [[ $OS == linux* ]]; then
+  conda install -y -c conda-forge libstdcxx-ng
+fi
 
 # Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
 python -m pip install -U "pip!=21.2.*"
--- a/install_sct
+++ b/install_sct
@@ -574,7 +574,7 @@ run mkdir -p "$SCT_DIR/$PYTHON_DIR"
 # Download miniconda
 print info "Downloading Miniconda..."
 case $OS in
-linux*)
+linux)
   download "$TMP_DIR/"miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
   ;;
 osx)
@@ -621,7 +621,7 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
 fi
 
 # Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
-if [[ $OS == linux* ]]; then
+if [[ $OS == linux ]]; then
   conda install -y -c conda-forge libstdcxx-ng
 fi
 


install_sct
```diff
@@ -620,6 +620,9 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
   exit 1
 fi
 
+# Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
+conda install -y -c conda-forge libstdcxx-ng
```
joshuacwnewton, line 624: Ah oops! The `libstdcxx-ng` is for Linux only, and not [macOS](https://stackoverflow.com/a/68845839).

joshuacwnewton: Fixed in [7eaadee](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3929/commits/7eaadeed1b415889fe03cfb846459a5450d18f7f).

install_sct
```diff
@@ -621,7 +621,9 @@ if [ "$ACTUAL_PYTHON" != "$EXPECTED_PYTHON"  ]; then
 fi
 
 # Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
-conda install -y -c conda-forge libstdcxx-ng
+if [[ $OS == linux* ]]; then
```
mguaypaq, line 624: Is there a reason to use the pattern `linux*` rather than just the literal `linux`, here and on [line 577](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7eaadeed1b415889fe03cfb846459a5450d18f7f/install_sct#L577)? As far as I can tell, the only possible values for `$OS` are the literals [`osx`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7eaadeed1b415889fe03cfb846459a5450d18f7f/install_sct#L167) and [`linux`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7eaadeed1b415889fe03cfb846459a5450d18f7f/install_sct#L172).

mguaypaq (approved):
Thanks for hunting down this fix!

install_sct
joshuacwnewton: Ahh, thank you for asking about this! 

I had noticed than an earlier line in the script mentions `linux*`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6c3a3dc2bcc819dddf03026e45fe4b61c1f1282b/install_sct#L576-L583

And I had presumed that there were multiple `linux`-es? But, as you've pointed out, there aren't! 

_(At least, not anymore! Doing a `git blame` on that original `linux*` line shows that it in a much older iteration of `install_sct`, we used to treat `centos6` as a separate case from other `linux` distros.)_

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/95954e209444eea033c066404149fbf726777c33/install_sct#L248-L257

I can change both instances of `linux*` to `linux`. :)

PR #3932, opened 2022-11-04
Convert `isct_test_ants` script into a `pytest` test, then call via `sct_check_dependencies`

### LINKED ISSUE #3905: Move `isct_test_ants` functionality from the CLI script into `sct_check_dependencies` / `testing/`

joshuacwnewton:
## Description

Currently, we have a user-facing CLI script called `isct_test_ants.py`. However, its contents are more akin to a test, rather than a traditional CLI script.

So, this PR migrates the test to `testing/`, and calls this test via `sct_check_dependencies` to preserve existing behavior.

## Linked issues

Supersedes https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3915.
Fixes #3905.

### DIFF
--- a/testing/dependencies/test_ants.py
+++ b/testing/dependencies/test_ants.py
@@ -1,56 +1,30 @@
 #!/usr/bin/env python
 #########################################################################################
-#
-# This function test the integrity of ANTs output, given that some versions of ANTs give a wrong BSpline transform,
-# notably when using sct_ANTSUseLandmarkImagesToGetBSplineDisplacementField.
-#
-# ---------------------------------------------------------------------------------------
 # Copyright (c) 2014 Polytechnique Montreal <www.neuro.polymtl.ca>
 # Authors: Julien Cohen-Adad
-# Modified: 2014-07-02
-#
-# About the license: see the file LICENSE.TXT
 #########################################################################################
 
-import sys
-import getopt
 import os
 
 import numpy as np
 import nibabel as nib
 
-from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv
+from spinalcordtoolbox.utils.sys import run_proc, printv
 from spinalcordtoolbox.utils.fs import tmp_create, rmtree
 
 from spinalcordtoolbox.scripts import sct_dice_coefficient
 
 
-# main
-# =======================================================================================================================
-def main():
-
+def test_ants_registration():
+    """
+    This function test the integrity of ANTs output, given that some versions of ANTs give a wrong BSpline transform,
+    notably when using sct_ANTSUseLandmarkImagesToGetBSplineDisplacementField.
+    """
     # Initialization
     size_data = 61
     size_label = 1  # put zero for labels that are single points.
     dice_acceptable = 0.39  # computed DICE should be 0.931034
-    test_passed = 0
-    remove_temp_files = 1
     verbose = 1
-
-    # Check input parameters
-    try:
-        opts, args = getopt.getopt(sys.argv[1:], 'hvr:')
-    except getopt.GetoptError:
-        usage()
-        raise SystemExit(2)
-    for opt, arg in opts:
-        if opt == '-h':
-            usage()
-            return
-        elif opt in ('-v'):
-            verbose = int(arg)
-        elif opt in ('-r'):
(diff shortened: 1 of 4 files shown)


testing/dependencies/test_ants.py
```diff
@@ -1,57 +1,31 @@
...
-
+def test_ants_registration():
+    """
+    This function test the integrity of ANTs output, given that some versions of ANTs give a wrong BSpline transform,
+    notably when using sct_ANTSUseLandmarkImagesToGetBSplineDisplacementField.
+    """
     # Initialization
     size_data = 61
     size_label = 1  # put zero for labels that are single points.
     dice_acceptable = 0.39  # computed DICE should be 0.931034
```
joshuacwnewton, line 26: As per @mguaypaq's previous review on https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3915#discussion_r1006114514:

> This is a separate issue, but `0.39` looks like a typo, given that it's supposed to be a threshold close to `0.931034`? I'd be curious to know if this test actually rejects the (unspecified) versions of ANTs that we're trying to guard against.

Interestingly, the current dice score is.... nowhere near the reported ideal of `0.931034`.

image

The similarity here does lead me to think that `0.39` is intentional, though. But, `0.39` is so low... :thinking: 

That said, this test has been like this since its inception, all the way back in 2016 (https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/4e4951d36dab6ae70a2c2389130e96df701c27f4), so. Hm!

mguaypaq: Thanks for the investigation. I think this is out of scope for this PR, but I've added an agenda item to the next SCT dev meeting to see if anyone remembers any context. If any followup is needed it can be in a separate PR.

mguaypaq (approved):
I like how much simpler this is as a test, rather than as a CLI script.

PR #3933, opened 2022-11-04
Add missing spaces in multiline strings, and missing commas in lists

mguaypaq:
Two missing spaces when the parts of the string get concatenated.

### DIFF
--- a/testing/cli/test_cli_sct_merge_images.py
+++ b/testing/cli/test_cli_sct_merge_images.py
@@ -13,4 +13,4 @@ def test_sct_merge_images_no_checks():
     TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
     sct_merge_images.main(argv=['-i', 'template/template/PAM50_small_cord.nii.gz', 't2/t2_seg-manual.nii.gz',
                                 '-w', 'mt/warp_template2mt.nii.gz', 't2/warp_template2anat.nii.gz',
-                                '-d' 'mt/mt1.nii.gz'])
+                                '-d', 'mt/mt1.nii.gz'])
--- a/spinalcordtoolbox/moco.py
+++ b/spinalcordtoolbox/moco.py
@@ -764,7 +764,7 @@ def register(param, file_src, file_dest, file_mat, file_out, im_mask=None):
                'mutual information. Either the mask you provided is '
                'too small, or the subject moved a lot. If you see too '
                'many messages like this try with a bigger mask. '
-               'Using previous transformation for this volume (if it'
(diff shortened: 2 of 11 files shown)


joshuacwnewton (approved):
Good catch!

I think I might have also come up with a regex that will help us search for instances of this project-wide:

```
\w'\n\s+'
```

(In other words, a word character (i.e. non-space), then a `'`, then a newline, then one-or-more whitespace characters, then another `'`)

Testing this on the incorrect description yields a match: https://regexr.com/71mm3

Testing this using PyCharm's regex search yields some false-positives (because the `n` in the string `'\n'` matches `\w`), but also a few extra hits, e.g.:

image

Up to you if you'd want to expand the scope of this PR to hit all of these instances, but I'll pre-approve here since the changes are ultra straightforward.

mguaypaq:
Thanks for the regex idea! I used a slightly different regex to find more problems and fewer false positives: `[^\\]\w['"]\s+['"]\w` (which is a pain to properly escape if you're using it in bash: `'[^\\]\w['"'"'"]\s+['"'"'"]\w'` or ``).

This found some more spacing problems in usage strings, and some missing commas in an example json snippet, and two broken tests.

testing/cli/test_cli_sct_merge_images.py
```diff
@@ -13,4 +13,4 @@ def test_sct_merge_images_no_checks():
     TODO: Check the results. (This test replaces the 'sct_testing' test, which did not implement any checks.)"""
     sct_merge_images.main(argv=['-i', 'template/template/PAM50_small_cord.nii.gz', 't2/t2_seg-manual.nii.gz',
                                 '-w', 'mt/warp_template2mt.nii.gz', 't2/warp_template2anat.nii.gz',
-                                '-d' 'mt/mt1.nii.gz'])
+                                '-d', 'mt/mt1.nii.gz'])
```
joshuacwnewton, line 16: I was curious how the tests were even passing if they were broken... :thinking: 

Without the comma, the two strings get concatenated together like this:

image

But it looks like argparse is smart enough to parse `-dmt/mt1.nii.gz` as `-d mt/mt1.nii.gz`:

image

Still, good catch!

mguaypaq: Ah! This is pretty common (although not universal) with "short" options (meaning, a single dash followed by a single letter, followed directly by the argument). For example, `git -amwip` is really parsed as `git -a -m "wip"`. I didn't know argparse supported this pattern.

The other test was more broken, though: when I tested it in the terminal, I got different error messages with spaced and non-spaced arguments (but still the same exit code, which is what the test looks at).

PR #3940, opened 2022-11-10
Set `draft: true` in the release creation workflow

### LINKED ISSUE #3787: Consider using "Draft Release" for `create-release.yml` GitHub Actions workflow
I just noticed this option when editing the 5.6 release:

image

This would be a nice way to run the release workflow, but not immediately have the release show up in the repo. (Which would allow the release author to edit the description and etc. without feeling pressured for time.)

joshuacwnewton:
## Description

Tiny change to enable draft releases by default, which would allow the release author to edit the description and etc. without feeling pressured for time.

From [ncipollo/release-action](https://github.com/ncipollo/release-action):

> Input name | Description | Required | Default Value
> -- | -- | -- | --
> draft | Optionally marks this release as a draft release. Set to true to enable. | false | ""

## Linked issues

Fixes #3787.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -173,3 +173,4 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         bodyFile: ".github/workflows/release-body.md"
         artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat"
+        draft: true


mguaypaq (approved):
Great! During the last release I had been wondering why it wasn't set to draft, so I'm glad for this change :slightly_smiling_face:

Issue #3843, opened 2022-07-14
PAM50 atlas warped to the subject's space gives corrupted values


jcohenadad:
### Description

This issue was originally reported here: https://github.com/sct-pipeline/gm-challenge/issues/77

The white matter atlas is supposedly encoded between 0 and 1. After estimating warping fields between the subject's space and the PAM50 space, those warping fields are applied to the PAM50 objects (including the WM atlas), and the values should _still_ range between 0 and 1, which does not seem to be the case in this reported bug (one voxel had value `-1`). 

### Suggestions

1. Make sure that the WM atlas has ALL values between 0 and 1. 
2. Reproduce the bug and... find a solution 😉  (likely something got corrupted with the warping field)

valosekj:
Just for clarification (I am the author of the original issue https://github.com/sct-pipeline/gm-challenge/issues/77); I did not work with PAM50 labels. Instead, I segmented the SC and GM using `sct_deepseg_sc` and `sct_deepseg_gm` functions.

## Steps to reproduce the bug:

Download data:
[sct_issue_3843.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/9109177/sct_issue_3843.zip)

Run:
```
# Segment spinal cord
sct_deepseg_sc -i sub-001_T2star.nii.gz -c t2s

# Segment gray matter
sct_deepseg_gm -i sub-001_T2star.nii.gz

# Generate white matter segmentation
sct_maths -i sub-001_T2star_seg.nii.gz -sub sub-001_T2star_gmseg.nii.gz -o sub-001_T2star_wmseg.nii.gz

# Erode white matter mask to minimize partial volume effect
sct_maths -i sub-001_T2star_wmseg.nii.gz -erode 1 -o sub-001_T2star_wmseg_erode.nii.gz

# Compute SNR, CNR, and contrast
# (the compute_snr script is from the gm-challenge package) 
compute_cnr --data1 sub-001_T2star.nii.gz --data2 .nii.gz --mask-noise sub-001_T2star_wmseg_erode.nii.gz --mask-wm sub-001_T2star_wmseg.nii.gz --mask-gm sub-001_T2star_gmseg.nii.gz --subject sub-001 --output ./results.csv
```
[...]

PR #3270, opened 2021-03-10
Create 2d masks in memory instead of via intermediate files

jcohenadad:
Fixes #3269

#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've applied a [release milestone](https://github.com/neuropoly/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/neuropoly/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

Issue #3075, opened 2020-12-03
ERROR: VirtualSampledPointSet must have 1 or more points. When using Gaussian masks in SliceReg


jcohenadad:
### Context

When using `sct_register_multimodal` with `algo=slicereg` and a Gaussian (ie non-binary) mask, the function crashes.

<details><summary>Output error</summary>

```
--
ESTIMATE TRANSFORMATION FOR STEP #1

Apply transformation from previous step

Parse list of warping fields...

Get dimensions of data...
  192 x 260 x 320 x 1

Apply transformation...

Apply transformation and resample to destination space...
Copy affine matrix from destination space to make sure qform/sform are the same.
Done! To view results, type:
fsleyes dest_RPI.nii src_reg.nii &
Registration parameters:
  type ........... im
  algo ........... slicereg
  slicewise ...... 0
  metric ......... CC
  iter ........... 10
  smooth ......... 0
  laplacian ...... 0
  shrink ......... 1
  gradStep ....... 0.5
  deformation .... 1x1x0
  init ...........
  poly ........... 5
  filter_size .... 5
  dof ............ Tx_Ty_Tz_Rx_Ry_Rz
  smoothWarpXY ... 2
  rot_method ..... pca

Traceback (most recent call last):
  File "/home/charley/sct-dev/scripts/sct_register_multimodal.py", line 413, in <module>
    main()
  File "/home/charley/sct-dev/scripts/sct_register_multimodal.py", line 389, in main
    path_out=path_out)
  File "/home/charley/sct-dev/scripts/sct_register_to_template.py", line 1038, in register_wrapper
    warp_forward_out, warp_inverse_out = register(src=src, dest=dest, step=step, param=param)
  File "/home/charley/sct-dev/scripts/sct_register_to_template.py", line 1204, in register
    verbose=param.verbose,
  File "/home/charley/sct-dev/spinalcordtoolbox/registration/register.py", line 189, in register_step_ants_slice_regularized_registration
    status, output = run_proc(cmd, verbose, is_sct_binary=True)
  File "/home/charley/sct-dev/spinalcordtoolbox/utils/sys.py", line 295, in run_proc
    raise RuntimeError(output)
RuntimeError: setting mask mask_crop.nii.gz
WARNING: In /__w/ANTs/ANTs/antsbin/staging/include/ITK-5.1/itkObjectToObjectMetric.hxx, line 579
[...]

PR #3944, opened 2022-11-11
Update header `dtype` property on save/load to match the datatype of the data array

### LINKED ISSUE #3232: Scaling issue with nibabel.save() when header datatype is `int`, but array datatype is `float`
### Description

### Steps to Reproduce

~~~
~~~

    ```python
    
    ```

### LINKED ISSUE #3943: Re-using the header from `im_in` can cause datatype mismatches in `im_out`
> NB: This is a very subtle bit of behavior, but it may have implications for how we end up solving https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3232.

In `sct_maths`, we re-use the header from the input image when creating the output image:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/scripts/sct_maths.py#L452-L456

Say our input image is a binary mask of type `uint8` with values `{0, 1}`. And, say we then perform a `-sub` operation such that negative numbers are introduced to the output image. Right now, the output image will be saved with a header that reads `uint8`, even though the array now contains signed values! 

> _**(In fact, this is exactly what happened in #3843!)**_

Mismatches like this can cause issues later down the line. For example, say we then try to [cast the array to the datatype listed in the header](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3232#issuecomment-1287313511) (`uint8`). In that case, the negative values will be erased (!!!).

So, it might be worthwhile to make sure that when saving an image from `sct_maths`, the header isn't just copied from `im_in`; we should make sure the header datatype matches the datatype of the newly-created output array.

joshuacwnewton:
## Background

`Image` objects contain 2 places where datatype information (such as `int16`, `float64`, etc.) can be found:

1. **`Image.data.dtype`**: Property that returns the current type of the image's data array
    - Note: The datatype of the array may change during mathematical operations (e.g. `3` (int) / `2` (int) = `1.5` (float))
2. **`Image.hdr.get_data_dtype()`**: The metadata contained in the header of the image
    - Unfortunately, the datatype stored in the header _won't_ be automatically updated if the array changes.

Because the array datatype changes automatically, but the header datatype must be manually updated, it's not uncommon to encounter a mismatch between the datatypes in the array and header. And, mismatches can cause issues such as:

* #3232 

## Description

To ensure that there are no datatype mismatches, this PR adds a new method to the `Image` class called `fix_header_dtype`, which updates the header `dtype` property to match the datatype of the array itself. This method is then called on both load and save.

I chose this approach over my [initial idea](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944#discussion_r1021695531) of modifying the _array_ to match the _header_, because existing header metadata [can't always be trusted](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944#discussion_r1021695531), and modifying the data array would be [a destructive operation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944#discussion_r1021695531).

> _NB: In order to allow `self.fix_header_dtype` to operate in-place without arguments, the `save()` method of the `Image` class had to be rewritten to change how the `mutable` argument worked (see [4012d95](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944/commits/4012d951dcd9de4aebcf035cf69eab18f5dc3ede) for further context)._

## Linked issues

Fixes #3232.
Fixes #3943.


spinalcordtoolbox/image.py
```diff
@@ -411,9 +413,23 @@ def loadFromPath(self, path, mmap, verbose):
...
+        self.affine = im_file.affine.copy()
+
+        # Check to see if there's a mismatch between the array's datatype and the header datatype
+        dtype_apparent = np.result_type(im_file.dataobj)
+        dtype_header = self.hdr.get_data_dtype()
+        if dtype_apparent != dtype_header:
+            logger.warning(f"DTYPEWARN: Image header specifies datatype {dtype_header}, but array appears to be of "
+                           f"type {dtype_apparent}. Array will be loaded as type {dtype_header}.")
+        # Convert the data array to the header's datatype
+        self.data = im_file.get_fdata().astype(dtype_header)
```
joshuacwnewton, line 426: In the above changes, I attempt to cast the data array to the type specified by the header.

When I first made these changes, I had hoped that the header's datatype property would be a useful source of truth, and that it would help to dictate what type the array "should" have. 

In reality, however:

1. In many cases, the header metadata is not to be trusted. For example:
    - In SCT's codebase, we often copy headers from one image to another (`im = Image(data, old_header)`) to preserve metadata.
    - We will also often modify data arrays directly (`im.data = updated_array`) without updating the header. 
    - This means that the header datatype property will often be outdated and incorrect.
2. In addition, casting the array datatype using `.astype(dtype_header)` is a very destructive operation, because it can often impose an incorrect datatype on an otherwise OK array.

So, I think instead, it's much safer to treat the _array_ as the source of truth, and update the _header_ on save/load to reflect the datatype of the array.

spinalcordtoolbox/image.py
```diff
@@ -501,9 +517,17 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
+            dtype_data = data.dtype
+            dtype_header = hdr.get_data_dtype()
+            if dtype_data != dtype_header:
+                logger.warning(f"DTYPEWARN: Image header specifies datatype '{dtype_header}', but array is of type"
+                               f"type '{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
+                if dtype_data == 'bool':
+                    # Sometimes we create boolean arrays, apparently! But, using 'bool' for nibabel headers triggers:
+                    # `nibabel.spatialimages.HeaderDataError: data dtype "bool" not supported`, so use `int16` instead
+                    dtype_data = np.int16
+                hdr.set_data_dtype(dtype_data)
```
joshuacwnewton, line 530: The test suite [reveals](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/3448711626/jobs/5755958619#step:5:834) that in several places, we create boolean arrays, then attempt to save them as image files.

Now, `nibabel` can handle the actual boolean array just fine when saving. But, if you try and call `hdr.set_data_dtype('bool')`, then you'll get an error, because `bool` isn't a valid header dtype value. 

So, just to mitigate some CI errors, I've substituted `int16` in, so that the image gets saved as an `int16` image. But, I  think the actual long-term solution deserves a little more discussion:

- What datatype should boolean arrays be cast to?
- When/where should this casting happen? 
- Should we avoid creating images from boolean arrays at all?

spinalcordtoolbox/image.py
joshuacwnewton: > So, I think instead, it's much safer to treat the array as the source of truth, and update the header on save/load to reflect the datatype of the array.

Upon making these changes, we now only have 2 failing tests:

```
_______________________________________________________________ test_sct_dmri_moco_check_params ________________________________________________________________
  File "/home/joshua/repos/spinalcordtoolbox/testing/cli/test_cli_sct_dmri_moco.py", line 21, in test_sct_dmri_moco_check_params
    assert allclose(lresults, lgroundtruth)
AssertionError: assert False
 +  where False = allclose(array([ 4.75290417e-04, -1.19489128e-05, -1.19489128e-05, -1.19489128e-05,\n       -1.29636934e-01, -1.29636934e-01, -1.29636934e-01]), 
                                 [0.00047529041677414337, -1.1970542445283172e-05, -1.1970542445283172e-05, -1.1970542445283172e-05, -0.1296642741802682, -0.1296642741802682, ...])
__________________________________________________________ test_sct_dmri_moco_with_mask_check_params ___________________________________________________________
  File "/home/joshua/repos/spinalcordtoolbox/testing/cli/test_cli_sct_dmri_moco.py", line 44, in test_sct_dmri_moco_with_mask_check_params
    assert allclose(lresults, lgroundtruth)
AssertionError: assert False
 +  where False = allclose(array([ 0.00804089,  0.00395558,  0.00395558,  0.00395558, -0.01534442,\n       -0.01534442, -0.01534442]), [0.008032332623754357, 0.0037734940916436697, 0.0037734940916436697, 0.0037734940916436697, -0.01502861167728611, -0.01502861167728611, ...])
```

The differences are tiny in magnitude; I'm fairly certain these failures are due to the fact that we're hardcoding these parameters.

I've had reservations about these tests before:
> If this check fails, though, it doesn't give us much qualitative feedback:
> 
> * The test fails in an identical way for changes of any magnitude (e.g. 1e-4 and 1e2), so it's hard to gauge the severity of a failing test.
> * The test displays the moco params, but doesn't provide any message describing the values, so it's not immediately obvious what it means for these values to change.
>  * Checking against these values only tests if we're preserving the current state of the moco algorithm. But, I think it would be valuable to also try to capture the goal of motion correction, too. (e.g. Say we started with misaligned data, and had ground truth for aligned data. Then we could check if the moco algorithm performs the realignment to an acceptable degree.)
>
> _Source: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3075#issuecomment-871366336_

So, I'm not concerned that this test is failing.

joshuacwnewton: tl;dr: I think the failing test is actually due to the current "ground truth" values being wrong (due to corruption from scaling issues). And, I think the new values are actually the correct ones!

<details> 
<summary> Investigation </summary>

I set breakpoints within both of the new datatype mismatch checks. 

Immediately, I found `float/int` mismatches on these lines

  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L244
  
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L283
  
    (The `.mean` call causes the `int16`-typed `im` array to change to `float64`)
  
  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L612-L614

    (The mathematical operation causes the `int16`-typed `data_targetz` array to change to `float32`.)
  
  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L377-L378
  
    (Originally, `im_moco` has a `float64` header. However, `im_data` has an `int16` header which gets copied over.)

In all four cases, having a `float` array and an `int` header is what what triggers: 

* #3232. 
 
So, it is likely that the values were being slightly corrupted! 

However, after this PR, the images are saved with `float64` in both the array _and_ header, avoiding any scaling issues.

</details>

testing/cli/test_cli_sct_dmri_moco.py
```diff
@@ -16,8 +16,8 @@ def test_sct_dmri_moco_check_params(tmp_path):
                              '-ofolder', str(tmp_path)])
 
     lresults = genfromtxt(tmp_path / "moco_params.tsv", skip_header=1, delimiter='\t')[:, 0]
-    lgroundtruth = [0.00047529041677414337, -1.1970542445283172e-05, -1.1970542445283172e-05, -1.1970542445283172e-05,
-                    -0.1296642741802682, -0.1296642741802682, -0.1296642741802682]
+    lgroundtruth = [4.75290417e-04, -1.19489128e-05, -1.19489128e-05, -1.19489128e-05,
+                    -1.29636934e-01, -1.29636934e-01, -1.29636934e-01]
```
joshuacwnewton, line 20: > So, I think instead, it's much safer to treat the array as the source of truth, and update the header on save/load to reflect the datatype of the array.

Upon making the changes proposed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944#discussion_r1021695531, we now only have 2 failing tests:

```
_______________________________________________________________ test_sct_dmri_moco_check_params ________________________________________________________________
  File "/home/joshua/repos/spinalcordtoolbox/testing/cli/test_cli_sct_dmri_moco.py", line 21, in test_sct_dmri_moco_check_params
    assert allclose(lresults, lgroundtruth)
AssertionError: assert False
 +  where False = allclose(array([ 4.75290417e-04, -1.19489128e-05, -1.19489128e-05, -1.19489128e-05,\n       -1.29636934e-01, -1.29636934e-01, -1.29636934e-01]), 
                                 [0.00047529041677414337, -1.1970542445283172e-05, -1.1970542445283172e-05, -1.1970542445283172e-05, -0.1296642741802682, -0.1296642741802682, ...])
__________________________________________________________ test_sct_dmri_moco_with_mask_check_params ___________________________________________________________
  File "/home/joshua/repos/spinalcordtoolbox/testing/cli/test_cli_sct_dmri_moco.py", line 44, in test_sct_dmri_moco_with_mask_check_params
    assert allclose(lresults, lgroundtruth)
AssertionError: assert False
 +  where False = allclose(array([ 0.00804089,  0.00395558,  0.00395558,  0.00395558, -0.01534442,\n       -0.01534442, -0.01534442]), 
                                 [0.008032332623754357, 0.0037734940916436697, 0.0037734940916436697, 0.0037734940916436697, -0.01502861167728611, -0.01502861167728611, ...])
```

The differences are (relatively) small in magnitude. In addition, I've had [issues with these 2 moco tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3075#issuecomment-871366336) in the past:

> If this check fails, it doesn't give us much qualitative feedback:
> 
> * The test fails in an identical way for changes of any magnitude (e.g. 1e-4 and 1e2), so it's hard to gauge the severity of a failing test.
> * The test displays the moco params, but doesn't provide any message describing the values, so it's not immediately obvious what it means for these values to change.
>  * Checking against these values only tests if we're preserving the current state of the moco algorithm. But, I think it would be valuable to also try to capture the goal of motion correction, too. (e.g. Say we started with misaligned data, and had ground truth for aligned data. Then we could check if the moco algorithm performs the realignment to an acceptable degree.)

These ground truth values have been manually updated in the past (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2642). In fact, the ground truth values have even been updated as a direct result of the same issue (#3232) that this PR is trying to fix (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3270#discussion_r593780155). 

So, I'm not overly surprised that these tests are failing once more.

joshuacwnewton: tl;dr: I think the failing test is actually due to the current "ground truth" values being wrong (due to corruption from scaling issues). And, I think the new values are actually the correct ones!

<details> 
<summary> Investigation </summary>

I set breakpoints within both of the new "datatype mismatch" checks. 

Immediately, I found `float/int` mismatches on these lines

  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L244
  
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L283
  
    (The `.mean` call causes the `int16`-typed `im` array to change to `float64`)
  
  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L612-L614

    (The mathematical operation causes the `int16`-typed `data_targetz` array to change to `float32`.)
  
  * https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d89f8a3de5253cdca5c5041a467422aa8214fa24/spinalcordtoolbox/moco.py#L377-L378
  
    (Originally, `im_moco` has a `float64` header. However, `im_data` has an `int16` header which gets copied over.)

In all four cases, having a `float` array and an `int` header is what what triggers: 

* #3232. 
 
So, it is likely that the values were being slightly corrupted, and that our supposedly-correct ground truth values... aren't correct? 

After this PR, though, the images are saved with `float64` in both the array _and_ header, which avoids any scaling issues.

</details>

Still, I'd like to dig just a little bit further to see _exactly_ where the changes are introduced. (Are all 4 datatype mismatches contribute? Does one instance have a more outsized effect than the other?)

testing/batch_processing/cached_results/dmri/fa_in_cst.csv
```diff
@@ -1,3 +1,3 @@
 Timestamp,SCT Version,Filename,Slice (I->S),VertLevel,Label,Size [vox],WA(),STD()
-2021-03-16 10:49:07,git-jn/3254-fix-sform-sct_warp_template-9566dce5850797d73d819917f7f49ce74b4a052c,/home/joshua/repos/spinalcordtoolbox/sct_example_data/dmri/dti_FA.nii.gz,2:14,,WM left lateral corticospinal tract,70.35704290421518,0.7793999429926204,0.13533664759850947
-2021-03-16 10:49:07,git-jn/3254-fix-sform-sct_warp_template-9566dce5850797d73d819917f7f49ce74b4a052c,/home/joshua/repos/spinalcordtoolbox/sct_example_data/dmri/dti_FA.nii.gz,2:14,,WM right lateral corticospinal tract,71.21054427612819,0.7755359256476565,0.13641569050448102
+2021-03-16 10:49:07,git-jn/3254-fix-sform-sct_warp_template-9566dce5850797d73d819917f7f49ce74b4a052c,/home/joshua/repos/spinalcordtoolbox/sct_example_data/dmri/dti_FA.nii.gz,2:14,,WM left lateral corticospinal tract,70.35704290421518,0.7793405440192693,0.13533664759850947
+2021-03-16 10:49:07,git-jn/3254-fix-sform-sct_warp_template-9566dce5850797d73d819917f7f49ce74b4a052c,/home/joshua/repos/spinalcordtoolbox/sct_example_data/dmri/dti_FA.nii.gz,2:14,,WM right lateral corticospinal tract,71.21054427612819,0.7755611783165834,0.13641569050448102
```
joshuacwnewton, line 3: The ground truth values here seem to be incorrect/flawed in much the same way that the `sct_dmri_moco` ground truth values are off.

I'm going to try to dig a little bit more into what exactly is causing the values to change.

spinalcordtoolbox/image.py
joshuacwnewton: I've rewritten this PR to modify the header, rather than the array.

spinalcordtoolbox/image.py
```diff
@@ -303,6 +303,14 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
         else:
             raise TypeError('Image constructor takes at least one argument.')
 
+        # Check to see if there's a mismatch between the array's datatype and the header datatype
+        dtype_data = self.data.dtype
+        dtype_header = self.hdr.get_data_dtype()
+        if dtype_data != dtype_header:
+            logger.warning(f"Image header specifies datatype '{dtype_header}', but array is of type"
+                           f"type '{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
```
valosekj, line 311:
```suggestion
            logger.warning(f"Image header specifies datatype '{dtype_header}', but array is of type"
                           f"'{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
```

joshuacwnewton: Ahh, good catch!

spinalcordtoolbox/image.py
```diff
@@ -501,9 +509,17 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
             hdr.set_data_shape(data.shape)
-            # Update dtype if provided (but not if based on SCT-specific values: 'minimize')
-            if (dtype is not None) and (dtype not in ['minimize', 'minimize_int']):
-                hdr.set_data_dtype(dtype)
+            # Update dtype to match that of the data array
+            dtype_data = data.dtype
+            dtype_header = hdr.get_data_dtype()
+            if dtype_data != dtype_header:
+                logger.warning(f"Image header specifies datatype '{dtype_header}', but array is of type"
+                               f"type '{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
```
valosekj, line 517:
```suggestion
                logger.warning(f"Image header specifies datatype '{dtype_header}', but array is of type"
                               f"'{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
```

valosekj (commented):
The lines 306-312 and 513-522 look very similar to me. They are actually doing the same thing, aren't they? Can they not be replaced by a single function/method to reduce code duplications?

spinalcordtoolbox/image.py
mguaypaq: For boolean arrays, I suspect `np.uint8` is the natural header dtype to use, rather than `np.int16`, assuming we want to use integer types at all.

spinalcordtoolbox/image.py
mguaypaq: I agree that the array dtype is much more reliable as a source of truth, so updating the header's dtype makes the most sense.

mguaypaq:
@valosekj says:
> They are actually doing the same thing, aren't they? Can they not be replaced by a single function/method to reduce code duplications?

I agree, a method like `self.fix_header_dtype()` called from both places would make sense.

spinalcordtoolbox/image.py
valosekj: > For boolean arrays, I suspect `np.uint8` is the natural header dtype to use, rather than `np.int16`, assuming we want to use integer types at all.

`np.uint8` sounds good. It is used for boolean data by the `nilearn`; see [here](https://github.com/nilearn/nilearn/pull/3286)

spinalcordtoolbox/image.py
```diff
@@ -401,6 +404,25 @@ def set_qform_to_sform(self):
...
         self.hdr._structarr['qform_code'] = self.hdr._structarr['sform_code']
 
+    def fix_header_dtype(self, header='self', array='self'):
+        """
+        Change the header dtype to the match the datatype of the array.
+
+        Note: This method defaults to operating on the image's `data` and `hdr` properties,
+              but the fix can also be applied to supplied arrays and headers, too."""
+        hdr = self.hdr if header == 'self' else header
+        data = self.data if array == 'self' else array
```
joshuacwnewton, line 414: > They are actually doing the same thing, aren't they? Can they not be replaced by a single function/method to reduce code duplications?

> I agree, a method like `self.fix_header_dtype()` called from both places would make sense.

I've done this in [65e9b31](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944/commits/65e9b313f47039f1d94c1a7ce2bd80453afc7416). The tricky thing, though, was that:

- The `load` LOCs acted on the properties of the image itself (so, a `fix_header_dtype` method could act on `self.data` and `self.hdr`)
- The `save` LOCs acted on _copies_ of the properties, meaning that `self.data` and `self.hdr` were off-limits.

To account for both cases while still using a class method, I had to use a sort of awkward pattern. So, we may also want to consider making this an external function instead? I'm not sure... :thinking:

spinalcordtoolbox/image.py
joshuacwnewton: `np.uint8` was my gut reaction too, but then @mguaypaq brought up an interesting point in a Slack message:

> * unsigned types are pretty terrible. For example:
>    ```python
>    >>> import numpy as np
>    >>> x = np.zeros(1, dtype=np.uint8); x
>    array([0], dtype=uint8)
>    >>> x - 1
>    array([255], dtype=uint8)
>     ```

Given that our binary arrays tend to be masks, and given that it's possible for masks to be used in array arithmetic (e.g. https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3843), I was a little hesitant to use an unsigned type? 

That said, `uint8` _is_ used quite heavily throughout the SCT codebase for binary images. So, it's not like using `uint8` would be out of place. Perhaps we should stick with `uint8`, then discuss SCT's overall handling of `uint8` images in a separate issue?

spinalcordtoolbox/image.py
joshuacwnewton: The second, copy-pasted typo was removed by the refactoring of the two separate checks.

spinalcordtoolbox/image.py
mguaypaq: I'm thinking more and more about this possible pattern:
- For masks and labels that we generate, store them as `uint8` (assuming we never see very large labels, i.e., bigger than 255?)
- When doing any arithmetic at all, convert all inputs to `float64`, and store the results as `float64`.

At this point my reasons for this are more intuitions than anything very solid, but do you have any immediate thoughts?

mguaypaq: (There's maybe the question of whether some operations, like `min` and `max`, count as arithmetic, but I think that can be resolved later.)

mguaypaq: I'll note that, despite its bad reputation, the `float64` type can handle arithmetic with integer values just fine. In particular, doing additions, subtractions and multiplications with any integers that would fit inside 32 bits without overflow gives the same result in `float64`, without any loss of precision. Similarly, divisions where the result is an integer (so, no fractional part) also have no loss of precision.

mguaypaq: Ah, except that `sct_image -type`, which is definitely user-facing, lets the user choose the storage type from `('uint8', 'int16', 'int32', 'float32', 'complex64', 'float64', 'int8', 'uint16', 'uint32', 'int64', 'uint64')`. The internal methods also accept `minimize` and `minimize_int`, but I can't find any use of them in SCT (so maybe they can be dropped?).

joshuacwnewton: I've typed up a response, but... I'm wondering if it might be better to move this discussion into a separate issue, given that the task of standardizing datatypes is much greater in scope than this PR accounts for.

(For the record, though, I do agree with the general notion of using `uint8` for labels/masks, and `float64` for arithmetic.)

spinalcordtoolbox/image.py
mguaypaq: I decided to refactor `Image.save()` so that `fix_header_dtype()` can be simplified, I hope that's ok.

mguaypaq (approved):
Thanks so much for fixing this long-standing issue!

I think this is a good approach: the dtype from self.data is a better source of truth than self.hdr.get_data_dtype(), and it's a good idea to fix any mismatches at load time and at save time.

Also, the updated tests make a lot of sense: the data there was most likely the result of bugs in the previous code.

Since I made some extra changes, I'll let @joshuacwnewton do a final review and merge this PR if he agrees with them.

spinalcordtoolbox/image.py
```diff
@@ -503,53 +503,44 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
 
         :param mutable: whether to update members with newly created path or dtype
         """
+        if not mutable:
+            self.copy().save(path, dtype, verbose, mutable=True)
+            return self
```
joshuacwnewton, line 506: This is a really interesting pattern! It took me a minute to wrap my head around, though, so I had one small tweak that might make it a bit easier to grok?

```suggestion
        if mutable:
            pass  # changes to the path/dtype/header will be applied in-place
        else:
            # otherwise, we work with a copy (to preserve the state of the original image object)
            self.copy().save(path, dtype, verbose, mutable=True)
            return self
```

I mainly suggest this because `mutable` is off by default, and will only ever be specified via `mutable=True`. So, I find it helpful to clarify the positive case (`if mutable`) rather than the negative case (`if not mutable`).

mguaypaq: Sure, if there's an `else` branch it's generally good to make the `if` condition "positive". Instead of `pass` though, we might as well put the rest of the code there. I'll make the change.

mguaypaq: Done!

spinalcordtoolbox/image.py
```diff
@@ -503,53 +503,44 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
+        if path is None:
+            # Use self.absolutepath by default
+            if self.absolutepath is None:
+                raise ValueError("Don't know where to save the image (no absolutepath or path parameter)")
+            path = self.absolutepath
+        elif os.path.isdir(path):
             # Save to destination directory with original basename
+            if self.absolutepath is None:
+                raise ValueError("Don't know where to save the image (no absolutepath and path parameter is dir)")
             path = os.path.join(os.path.abspath(path), os.path.basename(self.absolutepath))
```
joshuacwnewton: I had a little trouble understanding these `path` changes, too. (I couldn't easily tell what all of the cases were?)

This is mostly personal preference, but... I wanted to see if I could clarify each case?

- I changed the conditionals to check the positive case (`if self.absolutepath`)
- I added `else:` statements to give each case the same indentation
- I moved the comments around a little for visual clarity

```python
       # Case 1: `path` not specified
       if path is None:
            if self.absolutepath:  # Fallback to the original filepath
                path = self.absolutepath  
            else:
                raise ValueError("Don't know where to save the image (no absolutepath or path parameter)")
        # Case 2: `path` points to an existing directory
        elif os.path.isdir(path):
            if self.absolutepath:  # Use the original filename, but save to the directory specified by `path`
                path = os.path.join(os.path.abspath(path), os.path.basename(self.absolutepath))
            else:
                raise ValueError("Don't know where to save the image (path parameter is dir, but absolutepath is missing)")
        # Case 3: `path` points to a file
        elif :
            self.absolutepath = path
```

mguaypaq: Case 3 isn't really a case, actually: the result of the `if path is None` and `elif os.path.isdir(path)` cases is to make sure that the `path` variable is set to the actual filename we want to use, in case it's not already usable. After that, the `self.absolutepath = path` assignment should be unconditional: regardless of how we got the final value of `path`, we need to record it as this image's filename.

If you want to keep the last `else` for clarity, you could replace it with either `pass` or `path = path` (with a code comment in either case). Maybe an extra blank line before the `self.absolutepath = path` and/or a code comment as well?

mguaypaq: (But overall this is a fine suggestion, feel free to commit it yourself.)

joshuacwnewton: There is one corner case we're not checking for, though: 

`os.path.isdir(path)` checks if paths are [_existing_](https://docs.python.org/3/library/os.path.html#os.path.isdir) directories or not. Meaning, the user could erroneously pass a nonexistent directory, "Case 2" would be skipped, our code would presume that this directory is actually a file (Case 3), and `nib.save()` would likely fail.

So, it might be beneficial to distinguish between "nonexistent directory" and "file"?

```python
       # Case 1: `path` not specified
       if path is None:
            if self.absolutepath:  # Fallback to the original filepath
                path = self.absolutepath  
            else:
                raise ValueError("Don't know where to save the image (no absolutepath or path parameter)")
        # Case 2: `path` points to an existing directory
        elif os.path.isdir(path):
            if self.absolutepath:  # Use the original filename, but save to the directory specified by `path`
                path = os.path.join(os.path.abspath(path), os.path.basename(self.absolutepath))
            else:
                raise ValueError("Don't know where to save the image (path parameter is dir, but absolutepath is missing)")
        # Case 3: `path` has no extension, so it is a *nonexistent* directory
        elif os.path.splitext(path)[1] == '':
            if self.absolutepath:  # Use the original filename, but save to the directory specified by `path`
                os.mkdir(os.path.abspath(path))
                path = os.path.join(os.path.abspath(path), os.path.basename(self.absolutepath))
            else:
                raise ValueError("Don't know where to save the image (path parameter is dir, but absolutepath is missing)")     
        # Case 4: `path` points to a file
        else:
            self.absolutepath = path
```

Maybe checking the extension is a bit faulty, though? 

(e.g. What happens if, say, we're doing some filename manipulation, and we accidentally forget to append the extension? My proposed approach will quietly create a new directory and save the file using the absolutepath without error.)

So, maybe we _want_ `nib` to throw the error, rather than trying to catch it ourselves...

spinalcordtoolbox/image.py
```diff
@@ -503,53 +503,44 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
-            self.absolutepath = path
-            self.data = data
+        if self.hdr is not None:
+            self.hdr.set_data_shape(self.data.shape)
+            self.fix_header_dtype()
 
+        # nb. that copy() is important because if it were a memory map, save() would corrupt it
+        dataobj = self.data.copy()
+        affine = None
+        header = self.hdr.copy() if self.hdr is not None else None
```
joshuacwnewton: Thank you for explicitly laying out `dataobj`, `affine`, and `header` like this!

Now that I'm looking at it, it's curious to me that we're setting `affine = None`. I'm guessing the idea here is that we want to prompt `nibabel` to determine the affine from the state of the `dataobj`/`header`?

I tried digging into the `nibabel` source code for some time, but I got a little lost, as their saving logic is quite convoluted when there's multiple layers of inheritance going on. 

This probably isn't something that's super necessary to interrogate? I mostly wanted some sort of comment to explain why we're doing this, since otherwise it looks a little strange.

spinalcordtoolbox/image.py
```diff
@@ -503,53 +503,44 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
+        nib.save(nib.nifti1.Nifti1Image(dataobj, affine, header), path)
         if not os.path.isfile(path):
-            raise RuntimeError("Couldn't save {}".format(path))
-
+            raise RuntimeError(f"Couldn't save {path}")
```
joshuacwnewton: Thanks for another `.format` -> `f""` update! :>

spinalcordtoolbox/image.py
```diff
@@ -404,24 +404,22 @@ def set_qform_to_sform(self):
...
+        if dtype_data == bool:
+            dtype_data = np.uint8
+
+        dtype_header = self.hdr.get_data_dtype()
+        if dtype_header != dtype_data:
             logger.warning(f"Image header specifies datatype '{dtype_header}', but array is of type "
                            f"'{dtype_data}'. Header metadata will be overwritten to use '{dtype_data}'.")
-            hdr.set_data_dtype(dtype_data)
+
+        self.hdr.set_data_dtype(dtype_data)
```
joshuacwnewton, line 422: This is a tiny tiny tiny nitpicky question, but I noticed the call to `self.hdr.set_data_dtype` was moved outside of the `if` statement.

Is there any reason to call `self.hdr.set_data_dtype()` unconditionally, even when the two datatypes are the same?

spinalcordtoolbox/image.py
```diff
@@ -404,24 +404,22 @@ def set_qform_to_sform(self):
...
-        dtype_header = hdr.get_data_dtype()
-        dtype_data = data.dtype
-        if dtype_data != dtype_header:
-            if dtype_data == 'bool':
-                # Sometimes we create boolean arrays, apparently! But, using 'bool' for nibabel headers triggers:
-                # `nibabel.spatialimages.HeaderDataError: data dtype "bool" not supported`, so use `uint8` instead
-                dtype_data = np.uint8
+        """
+        # Using bool for nibabel headers is unsupported, so use uint8 instead:
+        # `nibabel.spatialimages.HeaderDataError: data dtype "bool" not supported`
```
joshuacwnewton, line 412: Thank you for updating this comment! My first draft of this was a little unprofessional, so I appreciate cleaning up the language here. :)

joshuacwnewton (commented):
Thank you for the refactor, @mguaypaq! I'm much, much happier with how things look now, especially now that we're not dealing with partial copies anymore. :) 

My comments/questions here are mainly nitpicks and readability changes -- otherwise, the core logic itself LGTM? I can't actually give GitHub Approval :tm: (since it's my PR), but consider this a soft "Approval" :)

spinalcordtoolbox/image.py
joshuacwnewton: Ahhh, of course! Case 3 isn't a case, you're right. Thanks for the correction there!

spinalcordtoolbox/image.py
joshuacwnewton: > Instead of pass though, we might as well put the rest of the code there. I'll make the change.

My only concern with this is readability -- with big if/else blocks, I find it a little cumbersome to have to scroll all the way to the bottom to see what the `else:` is... I like being able to read both if/else comments in one small, sequential set of lines?

That said, taking a look at your changes, the `if mutable` block does all fit into one 1080p screen, so... maybe it's not too bad?

spinalcordtoolbox/image.py
```diff
@@ -481,54 +501,45 @@ def save(self, path=None, dtype=None, verbose=1, mutable=False):
...
-            raise RuntimeError("Couldn't save {}".format(path))
-
+            if os.path.isfile(path) and verbose:
+                logger.warning("File %s already exists. Will overwrite it.", path)
+            if os.path.isabs(path):
+                logger.debug("Saving image to %s orientation %s shape %s",
+                             path, self.orientation, self.data.shape)
+            else:
+                logger.debug("Saving image to %s (%s) orientation %s shape %s",
+                             path, os.path.abspath(path), self.orientation, self.data.shape)
```
joshuacwnewton, line 524: I just realized... we're setting `self.absolutepath` to be `path` without actually... calling `abspath`?

I'm wondering if we'd be better off printing these debug messages, and _then_ setting `self.absolutepath`:

```diff
-           self.absolutepath = path
            if os.path.isfile(path) and verbose:
                logger.warning("File %s already exists. Will overwrite it.", path)
            if os.path.isabs(path):
                logger.debug("Saving image to %s orientation %s shape %s",

                             path, self.orientation, self.data.shape)
            else:
                logger.debug("Saving image to %s (%s) orientation %s shape %s",
                             path, os.path.abspath(path), self.orientation, self.data.shape)
+           self.absolutepath = os.path.abspath(path)	
```

spinalcordtoolbox/image.py
joshuacwnewton: > (But overall this is a fine suggestion, feel free to commit it yourself.)

Done in[22a24a2](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944/commits/22a24a26e8831d06599318e4e135bef5e6c02f7e). :)

mguaypaq: I think trying to guess whether a path is a nonexistent file or a nonexistent directory based on extension is more likely to mask errors than to be helpful, yes. (To be honest I'm surprised we even have special logic for the existing directory case.)

spinalcordtoolbox/image.py
mguaypaq: There's no particular reason. I think how it happened is that:
1. I first rewrote the code without any reference to `dtype_header`, so that `set_data_dtype` was called unconditionally.
2. Then I noticed the conditional `logger.warning` in the old code and inserted it back.

Either way should be fine. Feel free to hoist it back into the conditional. After all, if you wondered why it was this way, that's proof that it's confusing to the reader.

spinalcordtoolbox/image.py
mguaypaq: I think your version was fine really, I guess I was just in a rewriting mood, sorry.

spinalcordtoolbox/image.py
mguaypaq: Good catch! :+1:

spinalcordtoolbox/image.py
joshuacwnewton: No no no, oh gosh. I was actually feeling self-conscious about what I had written, even before your rewrite? 

I wanted to fix it up before you did, so I'm glad you did. :smile:

PR #3953, opened 2022-11-21
Minimize memory usage for `algo=dl` (and add a warning for potential OOM killer issues)

### LINKED ISSUE #3867: `algo=dl` abruptly stops with no error message when used on images that are too large
As discovered in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3865#issuecomment-1232161658, when the following command is run:

```
sct_register_multimodal -i _VISTA_Lspine_SENSE_20210922113031_2001.nii -d _VISTA_Tspine_SENSE_20210922113031_1901.nii -param step=1,type=im,algo=dl
```

The output of `sct_register_multimodal` abruptly stops when trying to predict using the second VxmDense model:

```
Load data to register...

 Creating data tensors...

 Input shape for fixed tensor: torch.Size([1, 1, 64, 304, 304])

 Input shape for moving tensor: torch.Size([1, 1, 64, 304, 304])

 Creating VxmDense models with arguments:

{'inshape': [64, 304, 304], 'int_steps': 5, 'int_downsize': 2, 'unet_half_res': True, 'nb_unet_features': ([256, 256, 256, 256], [256, 256, 256, 256, 256, 256])}

 Loading first VxmDense model...

 Loading second VxmDense model...

 Predicting using first VxmDense model...
```

There is no indication of what went wrong, so some investigation is needed.

> **NB:** This dataset is internal/private, and can be found on `duke` in `projects/template_dog_virginiatech`

joshuacwnewton:
## Description

This PR amends the `algo='dl'` feature introduced in SCT v5.7 (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807) by doing 2 things:

1. Minimize memory usage by factoring model inference into a function
    * `algo='dl'` runs inference using two successive PyTorch models.
    * When inference for both models is kept inside a single function, memory usage stacks.
    * However, when model inference is factored out into separate functions, memory usage resets between models.
    
2. Add a warning to preempt a possible OOM-killed process
    * By taking measurements, we can estimate how much memory will be used for images of a given size.
    * Then, if the estimated RAM usage will come close to (or exceed) the amount of free memory, we can emit a warning before the OOM killer kills the process.

## Linked issues

Fixes #3867.

### DIFF
--- a/spinalcordtoolbox/registration/algorithms.py
+++ b/spinalcordtoolbox/registration/algorithms.py
@@ -449,8 +449,11 @@ def register_dl_multimodal_cascaded_reg(fname_src, fname_dest, fname_warp_forwar
     logger.info("\n Creating VxmDense models with arguments:")
     logger.info(f"\n{reg_args}")
 
-    moved, warp_data_first = register_dl_first_model(input_moving, input_fixed, reg_args, device)
-    moved_final, warp_data_second = register_dl_second_model(moved, input_fixed, reg_args, device)
+    logger.info("\n Predicting using first VxmDense model...")
+    moved, warp_data_first = register_dl_inference('pt_cascaded_first_model.pt', input_moving, input_fixed, reg_args, device)
+
+    logger.info("\n Predicting using second VxmDense model...")
+    moved_final, warp_data_second = register_dl_inference('pt_cascaded_second_model.pt', moved, input_fixed, reg_args, device)
 
     # ---- Warping fields ---- #
     # Modify the warp data so it can be used with sct_apply_transfo()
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/registration/algorithms.py
```diff
@@ -447,56 +448,22 @@ def register_dl_multimodal_cascaded_reg(fname_src, fname_dest, fname_warp_forwar
...
-    # ---- Registration ---- #
     logger.info("\n Predicting using first VxmDense model...")
-    moved, warp_tensor_first = pt_first_model(input_moving, input_fixed, registration=True)
-    warp_data_first = warp_tensor_first[0].permute(1, 2, 3, 0).detach().numpy()
+    moved, warp_data_first = register_dl_inference('pt_cascaded_first_model.pt', input_moving, input_fixed, reg_args, device)
 
     logger.info("\n Predicting using second VxmDense model...")
-    moved_final, warp_tensor_second = pt_second_model(moved, input_fixed, registration=True)
-    warp_data_second = warp_tensor_second[0].permute(1, 2, 3, 0).detach().numpy()
+    moved_final, warp_data_second = register_dl_inference('pt_cascaded_second_model.pt', moved, input_fixed, reg_args, device)
```
joshuacwnewton, line 466: I ran inference using 5 images of varying sizes, then tracked RAM usage using my system monitor.

1. When inference for both models is kept within a single function:

    image

    Here, you can see 5 groupings, each with two successive jumps in RAM usage (one for the first model, and one for the second). The RAM usage here stacks!

2. Factoring out model inference into its own separate functions:

    image

    Here, you can also see 5 groupings. However, this time, the RAM usage resets after the first model's inference, such that there is no stacking, and the RAM usage is roughly equivalent for both the first and second models.

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -447,56 +448,22 @@ def register_dl_multimodal_cascaded_reg(fname_src, fname_dest, fname_warp_forwar
...
-    pt_second_model.to(device)
-    pt_second_model.eval()
+    volume_m = (new_img_shape[0] * new_img_shape[1] * new_img_shape[2] / 1e6)
+    ram_estimated = round((volume_m * 3.52) + 0.0538, 2)
+    ram_free = round(psutil.virtual_memory().free / 1024 ** 3, 2)
+    if ram_estimated + 1 > ram_free:
+        logger.warning(f"\nWARNING: System has {ram_free}GB of free memory, but inference may require as much as "
+                       f"{ram_estimated}GB to register images of size {new_img_shape}. As a result, the script may "
+                       f"quit unexpectedly due to lack of memory. To reduce memory requirements, please consider "
+                       f"cropping your images around the spinal cord prior to registration.")
```
joshuacwnewton, line 460: I've added this warning that tries to estimate when RAM issues will occur:

image

The estimated RAM value is based off of measurements taken by varying the size of the input image (across each dimension), then using `psutil` to check free RAM before and after inference.

joshuacwnewton: Thankfully, RAM appears to vary linearly with total volume, regardless of which dimension is varied:

image

Hence the equation used above:

```python
    volume_m = (new_img_shape[0] * new_img_shape[1] * new_img_shape[2] / 1e6)
    ram_estimated = round((volume_m * 3.52) + 0.0538, 2)
```

joshuacwnewton: > ```python
> if ram_estimated + 1 > ram_free:
> ```

Note that I've made the check a little more sensitive than it actually is (by `~1GB`) in an attempt to warn users when they get close to the limit, even if they don't exceed it.

spinalcordtoolbox/registration/algorithms.py
jcohenadad: super interesting!

spinalcordtoolbox/registration/algorithms.py
jcohenadad: awesome! very elegant solution, we are all Math lovers here

mguaypaq (approved):
This seems like a reasonable warning! And it's fine if the memory usage model isn't totally accurate, since it won't stop the code from running.

The `ram_estimated` computation seems to say that the model uses 53.8MB of fixed data structures, plus 3.52KB of extra memory for each voxel. I'm not surprised by the fixed part, and I guess a few KB per voxel is high but not beyond the pale.

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -554,6 +521,34 @@ def register_dl_multimodal_cascaded_reg(fname_src, fname_dest, fname_warp_forwar
...
+    weights_model = []
+    for k in trained_state_dict:
+        weights_model.append(trained_state_dict[k])
+    i = 0
+    i_max = len(list(pt_model.named_parameters()))
+    torchparam = pt_model.state_dict()
+    for k, v in torchparam.items():
+        if i < i_max:
+            torchparam[k] = weights_model[i]
+            i += 1
```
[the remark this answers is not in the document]
joshuacwnewton: --> https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3964

PR #3958, opened 2022-11-22
Allow `path_output` parameter to start with `~`

### LINKED ISSUE #3957: Can not use "~" in config file for path_output arg in `sct_run_batch`
### Description

When using the config file in `sct_run_batch`, with the parameter `path_output`, using "~" in the desired path results in : *No such file or directory*, even if the path should be created if it doesn't exist. It works when using it for `path_data`. If we specify the entire path for `path_output`, we don't have this error.

### Steps to Reproduce

With the following config file:

```python
(base) sebeda@joplin:~/dcm_metric_normalization$ cat etc/config_process_data_bug.json
{
  "path_data"   : "~/data_nvme_sebeda/data-multi-subject",
  "path_output" : "~/data_nvme_sebeda/dcm/preprocess_sg_20221122",
  "script"      : "./bin/process_data.sh",
  "jobs"        : 1
}
```

We get that the output path doesn't exist, which is right, but is should just be created instead.

```python
(base) sebeda@joplin:~/dcm_metric_normalization$ sct_run_batch -c etc/config_process_data_bug.json

--
Spinal Cord Toolbox (git-master-c57b43f715cb85dad87a33623c7caf0ed47543ca)

sct_run_batch -c etc/config_process_data_bug.json
--

configuring
INFO SYSTEM
-----------
OS: linux (Linux-5.15.0-53-generic-x86_64-with-debian-bookworm-sid)
CPU cores: Available: 128 | Threads used by ITK Programs: 1
RAM: Total 257591 MB | Available 204485 MB | Used 36371 MB

INPUT ARGUMENTS
---------------
batch_log: sct_run_batch_log.txt
config: etc/config_process_data_bug.json
continue_on_error: 1
email_from: null
email_host: smtp.gmail.com:587
email_to: null
exclude: null
exclude_list: null
ignore_ses: false
include: null
include_list: null
itk_threads: 1
jobs: 1
path_data: ~/data_nvme_sebeda/data-multi-subject
path_output: ~/data_nvme_sebeda/dcm/preprocess_sg_20221122
path_segmanual: .
script: ./bin/process_data.sh
script_args: ''
subject_prefix: sub-
v: 1
zip: false

SCRIPT
------
git commit: 90ee556288943e029c48a49b23d0604f900d3623*
git origin: https://github.com/sct-pipeline/dcm_metric_normalization.git
Copying script to output folder...
  File "/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_run_batch.py", line 598, in <module>
    main(sys.argv[1:])
  File "/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_run_batch.py", line 482, in main
    script_copy = os.path.abspath(shutil.copy(script, arguments.path_output))
  File "/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 248, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/python/envs/venv_sct/lib/python3.7/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '~/data_nvme_sebeda/dcm/preprocess_sg_20221122'
```

mguaypaq:
The `path_output` variable is defined so that it correctly handles paths which start with `~`, via `os.path.expanduser`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c57b43f715cb85dad87a33623c7caf0ed47543ca/spinalcordtoolbox/scripts/sct_run_batch.py#L407-L418

But the raw value of `arguments.path_output` is used once later on, causing problems:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c57b43f715cb85dad87a33623c7caf0ed47543ca/spinalcordtoolbox/scripts/sct_run_batch.py#L482

(A quick search didn't turn up any other bad uses of `arguments.path_*` parameters.)

## Linked issues
Fixes #3957.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -479,7 +479,7 @@ def main(argv: Sequence[str]):
     else:
         try:
             # Copy the script and record the new location
-            script_copy = os.path.abspath(shutil.copy(script, arguments.path_output))
+            script_copy = os.path.abspath(shutil.copy(script, path_output))
             print("{} -> {}".format(script, script_copy))
             script = script_copy
         except shutil.SameFileError:


joshuacwnewton (approved):
LGTM!

Some tangential thoughts, though:

I was looking over `sct_run_batch`, thinking about how this slipped through... and it's tricky! `sct_run_batch` uses mostly `arguments.<>`, with these expanded path variables being the only exception.

So, I was wondering... is there any merit to putting the `os.path.abspath(os.path.expanduser())` part inside the parser itself? Then, making a point to not even create these intermediate `path_` variables?

I'm not sure, honestly. Part of me likes having some distinction between what the user originally passes (`arguments`) vs. what the expanded version becomes...

But it would also be nice to be able to use `arguments.<>` all throughout the script without having to think about gotchas like this. :thinking: 

Either way though, I think this is mergeable. :)

mguaypaq:
That's a good idea! `os.abspath(os.expanduser(...))` does seem like a useful `argparse` type converter, maybe one that we could use consistently in other scripts too. While looking at argument processing in `sct_run_batch`, I'm tempted to review this bit of code too, which seems like it might be a good fit for an `argparse` custom action:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6649647dc27325687d0d8b3f464cc592964331da/spinalcordtoolbox/scripts/sct_run_batch.py#L344-L371

But I think I'll keep my changes small and local, and just use `os.abspath(os.expanduser(...))` as a custom type converter in `sct_run_batch`.

mguaypaq:
On second thought, there are a bunch of `path_*` variables that are computed from `arguments.path_output` rather than being themselves part of `arguments`. I think I'd rather have consistency between the `path_*` variables.

PR #3962, opened 2022-11-23
Pin `flake8<6.0.0` in our linting workflow

### LINKED ISSUE #3961: `flake8: error: unrecognized arguments: --diff` during linting CI test
Our linting check has started failing on recent PRs. ([Example run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/3535457184/jobs/5933500966)) 

The error given is:

```bash
##[debug]Evaluating condition for step: 'Lint with flake8'
##[debug]Evaluating: success()
##[debug]Evaluating success:
##[debug]=> true
##[debug]Result: true
##[debug]Starting: Lint with flake8
##[debug]Loading inputs
##[debug]Loading env
Run git diff origin/master... | flake8 --diff
##[debug]/usr/bin/bash -e /home/runner/work/_temp/4a22afc9-d7d5-45e4-90ec-ba759be1d28b.sh
usage: flake8 [options] file file ...
flake8: error: unrecognized arguments: --diff
Error: Process completed with exit code 2.
##[debug]Finishing: Lint with flake8
```

Looking at the installed version of `flake8` before/after failing runs shows that automatic upgrade to `flake8=6.0.0` was the cause here. I'll see if I can find an upstream issue to link this to.

I presume we can pin to `flake8<6.0.0`, but I'm curious if `flake8` provides an alternative option to `--diff`... :thinking:

joshuacwnewton:
## Description

v6.0.0 of `flake8` breaks our CI linting workflow by removing the `--diff` option. There is no officially-endorsed alternative for linting diffs. So, for now, preserve the old behavior until we work out a better solution. (--> https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3961#issuecomment-1325669386?)

## Linked issues

Fixes #3961.

### DIFF
--- a/.github/workflows/lint_code.yml
+++ b/.github/workflows/lint_code.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         python-version: 3.8
     - name: Install dependencies
-      run: pip install flake8
+      run: pip install flake8<6.0.0
     # Note: flake8 picks up project-wide configuration options from 'setup.cfg' in SCT's root directory
     - name: Lint with flake8
       # NB: 'master...branch' is syntactic sugar for 'git diff `git merge-base master branch`..branch'
--- a/.github/workflows/lint_code.yml
+++ b/.github/workflows/lint_code.yml
@@ -14,7 +14,7 @@ jobs:
       with:
         python-version: 3.8
     - name: Install dependencies
-      run: pip install flake8<6.0.0
+      run: pip install 'flake8<6.0.0'
     # Note: flake8 picks up project-wide configuration options from 'setup.cfg' in SCT's root directory
     - name: Lint with flake8
       # NB: 'master...branch' is syntactic sugar for 'git diff `git merge-base master branch`..branch'


.github/workflows/lint_code.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       with:
         python-version: 3.8
     - name: Install dependencies
-      run: pip install flake8
+      run: pip install flake8<6.0.0
```
mguaypaq, line 17: I think the `<` character is being interpreted by the shell as a redirection, does this escape work?
```suggestion
      run: pip install 'flake8<6.0.0'
```

mguaypaq (approved):
Merging this would unblock us for the release, so let's go for it.

mguaypaq:
From the [commit message](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3962/commits/0614acecc4dc938f4adaad5e1312694b045de3cb):
> I wonder if there's a better place to store this dependency. Could we have some sort of list of dev dependencies? Do devs rely on any other packages besides `flake8`? Local docs builds, maybe?

Then again, storing the version constraint in `lint-code.yml` means that it's right next to the reason we have it: the use of the `--diff` flag right below. But if we ever get an official git commit hook that runs linting on our development clones of the repo, it would be important to list the dependency on `flake8` more centrally.

joshuacwnewton:
I think the main reason I had suggested that was because of the `pip install flake8` in dev Wiki instructions for adding the precommit hook? 

But, it's simple enough to just update the Wiki page to also say `flake8<6.0.0`, too. :)

PR #3963, opened 2022-11-23
List changed files for flake8

### LINKED ISSUE #3961: `flake8: error: unrecognized arguments: --diff` during linting CI test

mguaypaq:
As of version 6.0.0, flake8 no longer supports the --diff option. This PR instead supplies flake8 with the list of changed Python files, so that they can be checked in their entirety.

This is a draft PR to see if it even works.

If we decide to do this, I should also update [the dev wiki instructions for local linting](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Installing-SCT-for-development#step-5-linting-automatically).

## Linked issues
Alternative to PR #3962.
Fixes #3961.


.github/workflows/lint_code.yml
```diff
@@ -17,6 +17,4 @@ jobs:
       run: pip install flake8
     # Note: flake8 picks up project-wide configuration options from 'setup.cfg' in SCT's root directory
     - name: Lint with flake8
-      # NB: 'master...branch' is syntactic sugar for 'git diff `git merge-base master branch`..branch'
-      # NB: You can exclude the second ref ('branch') and git diff will take the currently checked-out ref
-      run: git diff origin/master... | flake8 --diff
+      run: git diff --name-only --diff-filter=d --merge-base master | grep '\.py$' | xargs --delimiter='\n' --no-run-if-empty flake8
```
joshuacwnewton, line 20: Dissecting this command:

* [`git diff --merge-base master`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgt--merge-baseltcommitgt--ltpathgt82308203): Generates diff between working tree and `master`. 
    * (With no other options, this will output the diff [as a patch by default](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---patch), which is very verbose.)
* [`--name-only`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---name-only):  Will suppress the patch output, and display only filenames of changed files.
* [`--diff-filter=d`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203): Display all kinds of changed files (e.g. added, copied, modified, renamed, etc.) **except** files that are deleted (lowercase `d` == exclusion). This is necessary because trying to run `flake8` on a nonexistent file will result in error.
* `| grep '\.py$'`: Filter the output of `git diff` to include only filenames that end in `.py`.
* `| xargs flake8`: Run `flake8` once per filename, using each filename as an argument.
* `--delimiter='\n'`: This is necessary because, by default, `xargs` uses whitespace to delimit input. But, `git diff` outputs each filename on its own line.
* `--no-run-if-empty`: This is necessary because running `flake8` with no arguments will attempt to process the entire working directory. (And, "no arguments" might happen if a PR is opened that only modifies non-Python files.)

This all seems very reasonable to me!

.github/workflows/lint_code.yml
```diff
@@ -17,6 +17,4 @@ jobs:
       run: pip install flake8
```
joshuacwnewton, line 17: I think on master, `pip install flake8` is actually pinned to `<6.0.0`? So when we merge master, we'll probably have to add another commit to unpin `flake8`.

joshuacwnewton (commented):
I can't see anything wrong with the changes themselves? The only catch here is that this PR behaves differently than the current <6.0.0 `flake8 diff` behavior -- it lints whole files, rather than merely the changed LOCs.

We could take the time to address linting issues project-wide? If we did that, then linting whole files would behave identically to the current behavior. 

But, right now our project has a _lot_ of linting issues... so, I think this PR would result in future PRs failing the linting check almost 100% of the time.

.github/workflows/lint_code.yml
mguaypaq: A small correction: `| xargs flake8` tries to run `flake8` with many files at once, not once per file.

For example:
```sh
$ seq 10 | xargs echo cmd
cmd 1 2 3 4 5 6 7 8 9 10
$ seq 10 | xargs --max-args=3 echo cmd
cmd 1 2 3
cmd 4 5 6
cmd 7 8 9
cmd 10
```

joshuacwnewton: Oops! Thank you for the correction.

.github/workflows/lint_code.yml
mguaypaq: Good point. Although that could even be a separate PR if `flake8` version 6.0.0 introduces other breaking changes.

mguaypaq:
Thanks for the link to prior discussion about _Black_, and for the experimental commit! That would be an interesting issue to revisit at an upcoming dev meeting.

But actually, since #3842, there's not many linting issues left in the SCT codebase. (There's a few more since #3998 because of the now-enforced maximum line length, but that's ok.) So what I'll do is that I'll try to address the remaining linting issues before continuing with this PR.

mguaypaq:
Now that #4040 is fixed, we should be able to merge this!

.github/workflows/check-code-quality.yml
joshuacwnewton, line 1: Your solution (using `git diff`, filtering for Python files, passing them to `flake8`) is excellent!! :hearts: 

Though, it's naturally geared towards our CI + PR diffs. So, separately, I had one other idea geared more towards usage on local dev machines. 

If we make the following change to `setup.cfg`:

```ini
[flake8]
extend-exclude = python  # Exclude SCT's miniconda dir to avoid scanning dependency `.py` files
```

A bare `flake8` command can be run in the root of the SCT repo, regardless of branch or state, which is a lot simpler to remember for local sanity checks/git hooks/etc.

----

Do you think this might be worth including in this PR, as a sort of "here are the changes we can now make after fixing all of the linting issues" grouping of changes?

joshuacwnewton (approved):
Giving my approval in advance since my suggestion is such a simple change, and everything else LGTM.

PR #3960, opened 2022-11-23
Ensure that the output of `get_centerline()` can be saved using `-r 0`

joshuacwnewton:
## Description

In SCT, the `get_centerline()` function is used by many CLI scripts for detecting the centerline of the spinal cord.

In many cases, CLI scripts will even offer users external arguments for centerline parameters, which will then be fed into `get_centerline()`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c4f05f6d923dc7e2e78bc3e345cdaa111916bb34/spinalcordtoolbox/scripts/sct_deepseg_sc.py#L51-L60

Separately, many CLI scripts have an `-r` argument, which allows users to keep temporary files by setting `-r` to `0`.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c4f05f6d923dc7e2e78bc3e345cdaa111916bb34/spinalcordtoolbox/scripts/sct_deepseg_sc.py#L95-L100

However, right now, the `get_centerline()` function doesn't actually use the value of `-r`.  So, no intermediate centerline files will be saved to disk (regardless of the value of `-r`), which can make debugging difficult.

> _**NB:** The only way that centerline files can currently be saved is if if the `'optic'` method is chosen, and `-v 2` is set. But, `'optic'`is only a subset of the available centerline methods, and using `-v` is also a bit unintuitive when `-r` exists._

To address this issue, this PR makes the following changes:

- Adds a `remove_temp_files` parameter to `get_centerline()` (89a3a05a6e3ae6bbb4a1c5e5fb59f27c23f10b2c)
- Makes sure that setting `remove_temp_files == 0` causes a centerline file to be saved (f9feeb98e25ad26bfa2e20e59dda5d5e25efe742)
- Propagates the value of `remove_temp_files` into the inner OptiC centerline detection function, too, replacing the `verbose == 2` behavior (13fa59438b3106574dd51da263568506ce54d4e1)
- Then, we make sure that any relevant calls to `get_centerline()` actually specify `remove_temp_files`:
    - This was done for 7 total CLI scripts -- the ones where centerline parameters can be modified: `sct_deepseg_sc`, `sct_deepseg_lesion`, `sct_propseg`, `sct_get_centerline`, `sct_process_segmentation`, `sct_register_to_template`, and `sct_straighten_spinalcord`.
[...]

PR #2147, opened 2019-01-24
Major refactoring of centerline routine

jcohenadad:
Major refactoring to clarify code, centerline centerline-related stuff inside spinalcordtoolbox.centerline, minimize i/o, pythonize code, add unit tests. More specificially:

- Major refactoring to centralize functions related to curve fitting inside `spinalcordtoolbox/centerline`. Notably: `sct_straighten_spinalcord.smooth_centerline()`, msct_smooth.py, msct_nurbs.py.
- Fixed error of compatibility between 'extract_centerline' and 'centerline when using viewer' in `sct_get_centerline` and `sct_deepseg_sc`. Fixes #2146 
- `extract_centerline()` now also output the real centerline coordinate, not only the discrete version. Fixes #1938 
- Fixed inconsistencies with the flag -init in `sct_get_centerline`. Fixes #1886
- Removed duplicate: sct_label_utils.extract_centerline and sct_process_segmentation.extract_centerline. Fixes #1937
- Address smoothing issue in `process_segmentation.extract_centerline`. Fixes #1936
- Removed unnecessary i/o in `process_segmentation.extract_centerline`. Fixes #1935
- Removed duplication: sct_get_centerline.viewer and sct_process_segmentation.extract_centerline. Fixes #1934
- Fixed loss of precision with Hanning filtering. Fixes #2043 
- Harmonize output name. Fixes #1777 
- Deal with NaN in image. Fixes #1774
- Make sure FSLeyes integration is not broken with changes in output path

## Results batch_processing

jca/2146-centerline/4f52278bb9aa9c96d2be8bacd601b458d6205346
~~~
t2/CSA:          78.8522376127
mt/MTR(WM):      54.1816425058
t2s/CSA_GM:      11.8920875133
t2s/CSA_WM:      65.8072385202
dmri/FA(CST_r):  0.793649553099
dmri/FA(CST_l):  0.764177717674
~~~

master/4fa355badbfccf9017632f7a15bc0f9a248ee431
~~~
t2/CSA:          78.7586037178
mt/MTR(WM):      54.1523235253
t2s/CSA_GM:      11.871046642
t2s/CSA_WM:      65.595794105
dmri/FA(CST_r):  0.783606378857
dmri/FA(CST_l):  0.76378467981
~~~

PR #3975, opened 2022-12-06
Refactor `get_centerline` function to distinguish between OptiC and non-OptiC methods

### LINKED ISSUE #3974: Temporary centerline image (`nii`) is not saved during `-r 0` when `algo='optic'`
In #3960, we added a conditional in the `get_centerline()` function that would save the centerline image if `-r 0`:

> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/812a8e00bd8862c3c9f2d099e24e0143a5337976/spinalcordtoolbox/centerline/core.py#L241-L244

However, what we missed was that the `optic` method has its own return statement (note the comment): 

> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/812a8e00bd8862c3c9f2d099e24e0143a5337976/spinalcordtoolbox/centerline/core.py#L147-L150
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/812a8e00bd8862c3c9f2d099e24e0143a5337976/spinalcordtoolbox/centerline/core.py#L159-L163

Because `'optic'` exits the function prematurely, it skips the temporary file creation. :facepalm:

joshuacwnewton:
## Description

In #3960, I (and both reviewers) overlooked the fact that the `optic` method had its own separate return call. Because of this return call, the PR did not work as intended; the `optic` codeblock exited before the temporary file could be saved.

I believe a big part of why we missed the second `return` statement was due to the layout of the `get_centerline` function. As-is, it's not clear which parts of the function are "optic" lines of code, which parts are "non-optic" lines of code, and which parts apply to both categories.

* Before: [`get_centerline`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/464724d1d60637b0bbe3797c9fca0ad787e2f5ab/spinalcordtoolbox/centerline/core.py#L85-L249) (`master`)

To fix the problem, this PR refactors the `get_centerline` function to make the distinctions between methods clearer. 

* After: [`get_centerline`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d544d96b87c9a34ab38895df469595340652740a/spinalcordtoolbox/centerline/core.py#L85-L255) (`jn/3974-refactor-get_centerline_function`)

In the process, these changes eliminate the intermediate `return` call, which then enables `get_centerline` to properly produce a centerline file for the `optic` method.

> _NB: While working on this PR, I also noticed a small bug with a call to `os.path.join`, which I fixed in d544d96b87c9a34ab38895df469595340652740a._

## Reviewing this PR

While this looks like a big PR, the large diff is mainly due to moving sections of code around. The actual changes are relatively small and straightforward, so please use the "Commits" view to look at each change in isolation.

## Linked issues

Fixes #3974.

### DIFF
--- a/spinalcordtoolbox/centerline/core.py
+++ b/spinalcordtoolbox/centerline/core.py
@@ -181,18 +181,6 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
         # reorient centerline to native orientation
         im_centerline.change_orientation(native_orientation)
         im_seg.change_orientation(native_orientation)
-        # TODO: Reorient centerline in native orientation. For now, we output the array in RPI. Note that it is tricky to
-        #   reorient in native orientation, because the voxel center is not in the middle, but in the top corner, so this
-        #   needs to be taken into accound during reorientation. The code below does not work properly.
-        # # Get a permutation and inversion based on native orientation
-        # perm, inversion = _get_permutations(im_seg.orientation, native_orientation)
-        # # axes inversion (flip)
-        # # ctl = np.array([x_centerline_fit[::inversion[0]], y_centerline_fit[::inversion[1]], z_ref[::inversion[2]]])
-        # ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
-        # ctl_deriv = np.array([x_centerline_deriv[::inversion[0]], y_centerline_deriv[::inversion[1]], np.ones_like(z_ref)])
-        # return im_centerline, \
-        #        np.array([ctl[perm[0]], ctl[perm[1]], ctl[perm[2]]]), \
-        #        np.array([ctl_deriv[perm[0]], ctl_deriv[perm[1]], ctl_deriv[perm[2]]])
 
         # Compute fitting metrics
         fit_results = FitResults()
--- a/spinalcordtoolbox/centerline/core.py
+++ b/spinalcordtoolbox/centerline/core.py
@@ -126,7 +126,7 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
     else:
         px, py, pz = im_seg.dim[4:7]
 
-        # Take the center of mass at each slice to avoid: https://stackoverflow.com/questions/2009379/interpolate-question
+        # Take the center of mass at each slice to avoid: https://stackoverflow.com/q/2009379
         x_mean, y_mean, z_mean = find_and_sort_coord(im_seg)
 
         # Crop output centerline to where the segmentation starts/end
--- a/spinalcordtoolbox/centerline/core.py
+++ b/spinalcordtoolbox/centerline/core.py
@@ -246,7 +246,8 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
     # Save centerline image to tmp_folder (but only if user hasn't opted to `remove_temp_files`)
     if not remove_temp_files:
         tmp_folder = tmp_create("centerline")
-        im_centerline.save(os.path.join(tmp_folder, add_suffix(im_seg.absolutepath, "_ctr")), mutable=True)
+        im_centerline.save(os.path.join(tmp_folder,
+                                        add_suffix(os.path.basename(im_seg.absolutepath), "_ctr")), mutable=True)
 
     return (im_centerline,
             np.array([x_centerline_fit, y_centerline_fit, z_ref]),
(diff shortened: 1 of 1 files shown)


joshuacwnewton:
NB: Sorry in advance about the diff for commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3975/commits/ae95b5572c8da6d44d9c02602dc84a3c75e02fad. 

It's literally just adding an `else:` and indenting everything within the block, but GitHub does a poor job at intelligently displaying this diff. :sweat:

spinalcordtoolbox/centerline/core.py
```diff
@@ -156,92 +114,140 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+        #   needs to be taken into accound during reorientation. The code below does not work properly.
+        # # Get a permutation and inversion based on native orientation
+        # perm, inversion = _get_permutations(im_seg.orientation, native_orientation)
+        # # axes inversion (flip)
+        # # ctl = np.array([x_centerline_fit[::inversion[0]], y_centerline_fit[::inversion[1]], z_ref[::inversion[2]]])
+        # ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
+        # ctl_deriv = np.array([x_centerline_deriv[::inversion[0]], y_centerline_deriv[::inversion[1]], np.ones_like(z_ref)])
+        # return im_centerline, \
+        #        np.array([ctl[perm[0]], ctl[perm[1]], ctl[perm[2]]]), \
+        #        np.array([ctl_deriv[perm[0]], ctl_deriv[perm[1]], ctl_deriv[perm[2]]])
```
joshuacwnewton, line 195: I'm wondering whether or not we should remove this dead code in the process.

The `TODO` is a bit confusing to me, because it implies we _don't_ reorient the centerline... when... we very clearly do, not 2 lines above?

I'm wondering whether the problem described in this comment is actually still a problem... :thinking:

valosekj (dismissed):
Thank you @joshuacwnewton!

I tested your branch, and it seems to be working as expected now-- `centerline` temp folder is created and not deleted:

```
$ sct_deepseg_lesion -i sub-mon001_ses-M0_T2w.nii.gz -c t2 -r 0

--
Spinal Cord Toolbox (git-jn/3974-refactor-get_centerline_function-d544d96b87c9a34ab38895df469595340652740a*)

sct_deepseg_lesion -i sub-mon001_ses-M0_T2w.nii.gz -c t2 -r 0
--


Method:
	Centerline algorithm: svm
	Assumes brain section included in the image: True

Image header specifies datatype 'int16', but array is of type 'float64'. Header metadata will be overwritten to use 'float64'.
Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221207140131.723604-qm0v7k81)

Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221207140134.767863-kbe2hv5_)
Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221207140139.284801-centerline-0t2l154s)
```

Thanks to that, I can easily find the `.nii` centerline:

```
$ ls /var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221207140139.284801-centerline-0t2l154s
image_in_RPI_res_ctr.nii
```

spinalcordtoolbox/centerline/core.py
mguaypaq: This file was born with this commented out code in the gigantic PR #2147 four years ago, but even in that PR I don't see any deleted code that it would correspond to. If this TODO points to a genuine bug, I don't think this comment is going to be the reason we find/fix it. Let's get rid of it!

mguaypaq (approved):
Thanks for breaking up this change into such easy-to-review commits with good descriptions!

I like this reorganization, especially: having a single return at the end, and fitting the optic and non-optic methods into a single if-else mould.

joshuacwnewton:
> Hm, I just tested the `sct_deepseg_lesion` with `-centerline cnn` method, and it fails:

Good catch, @valosekj! 

I had made the assumption that `im_seg` would always have an `absolutepath`. However, for `sct_deepseg_lesion -centerline cnn`, this assumption doesn't hold:

* `im_seg` corresponds to `im_heatmap` from the outer context.
* `im_heatmap` is generated by the [`heatmap`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fd2c839a84da564bcaf7a25ed58bbcda0e9d0874/spinalcordtoolbox/deepseg_/sc.py#L208-L304) function
* The heatmap image is initially created via the [`change_type`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fd2c839a84da564bcaf7a25ed58bbcda0e9d0874/spinalcordtoolbox/deepseg_/sc.py#L211) function
* The `change_type` function intentionally [does not preserve `absolutepaths`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fd2c839a84da564bcaf7a25ed58bbcda0e9d0874/spinalcordtoolbox/image.py#L1164)!

I can fix this issue by accounting for the case when `im_seg.absolutepath` isn't defined.

Issue #3113, opened 2020-12-20
Display vertebral level when using -perslice 1 option


jcohenadad:
When running `sct_process_segmentation` (or `sct_extract_metric`) with option `perslice 1`, each row corresponds to a slice. However, when doing so, the column "VertLevel" is empty:

Filename | Slice (I->S) | VertLevel | MEAN(area)
-- | -- | -- | --
/Users/julien/temp/20201217_inspired/results6/data_processed/HC02/bl/cord/processing_sct/t2_tra_seg.nii.gz | 0 |   | 70.41316501066880
/Users/julien/temp/20201217_inspired/results6/data_processed/HC02/bl/cord/processing_sct/t2_tra_seg.nii.gz | 1 |   | 76.68867922117070
/Users/julien/temp/20201217_inspired/results6/data_processed/HC02/bl/cord/processing_sct/t2_tra_seg.nii.gz | 2 |   | 76.84621033344560
/Users/julien/temp/20201217_inspired/results6/data_processed/HC02/bl/cord/processing_sct/t2_tra_seg.nii.gz | 3 |   | 81.27146085321930
/Users/julien/temp/20201217_inspired/results6/data_processed/HC02/bl/cord/processing_sct/t2_tra_seg.nii.gz | 4 |   | 82.83952355053640

It shouldn't be empty, because we _know_ the information about vertebral level, if user input vertebral labeling file, as was done here, with the flag `-vertfile`. 

So as a feature, we should fill this column "VertLevel" when the vertebral level for each slice is known.

joshuacwnewton:
I very much agree with this suggestion. The blank space stood out to me too while writing #3094.

We already have the capacity to get the level values, e.g. when we do `-perslice 1 -vert`:

https://github.com/neuropoly/spinalcordtoolbox/blob/997c82ebaa2aa3c0502b9fc3d8f753d0efbc08a7/spinalcordtoolbox/aggregate_slicewise.py#L260

But, for `-perslice 1 -z` we don't bother getting the level values for some reason.

https://github.com/neuropoly/spinalcordtoolbox/blob/997c82ebaa2aa3c0502b9fc3d8f753d0efbc08a7/spinalcordtoolbox/aggregate_slicewise.py#L269

So, I think a small tweak to this line in  `aggregate_per_slice_or_level()` is all we need.

Issue #3976, opened 2022-12-06
`sct_process_segmentation -perslice 1` - include `VertLevel` information across all available levels


valosekj:
### Description

When running `sct_process_segmentation -perslice 1`, is it possible to include the `VertLevel` information across all available levels?

I intuitively thought that it could have been possible by running:

```
sct_process_segmentation -i t1w_seg-manual.nii.gz -perslice 1 -vertfile t1w_seg-manual_labeled.nii.gz
```

However, the output `csa.csv` file has an empty `VertLevel` column:

```
Timestamp              SCT Version                                             Filename                                                                              Slice (I->S)  VertLevel  DistancePMJ  MEAN(area)          STD(area)  MEAN(angle_AP)        STD(angle_AP)  MEAN(angle_RL)        STD(angle_RL)  MEAN(diameter_AP)   STD(diameter_AP)  MEAN(diameter_RL)   STD(diameter_RL)  MEAN(eccentricity)  STD(eccentricity)  MEAN(orientation)      STD(orientation)  MEAN(solidity)      STD(solidity)  SUM(length)
"2022-12-06 16:57:33"  "git-master-812a8e00bd8862c3c9f2d099e24e0143a5337976*"  "/Users/valosek/Desktop/process_seg_issue/sct_testing_data/t1/t1w_seg-manual.nii.gz"  "0"           ""         ""           70.18783267281653   0.0        2.759028956254652     0.0            13.426794943347216    0.0            7.642569740858666   0.0               11.967398081968513  0.0               0.7695257203552127  0.0                1.1073798178329701     0.0               0.9349333333333333  0.0            1.029293815567324
"2022-12-06 16:57:33"  "git-master-812a8e00bd8862c3c9f2d099e24e0143a5337976*"  "/Users/valosek/Desktop/process_seg_issue/sct_testing_data/t1/t1w_seg-manual.nii.gz"  "1"           ""         ""           65.19992482576312   0.0        2.6477986592570417    0.0            12.883732671268943    0.0            7.170174386686721   0.0               11.479623207575598  0.0               0.7809447173833011  0.0                0.7045383853767712     0.0               0.9698067632850241  0.0            1.0269219965887146
[...]

PR #3977, opened 2022-12-07
Bring metrics in PAM50 anatomical dimensions in `sct_process_segmentation`

### LINKED ISSUE #3733: Add normalized metrics for degenerative cervical myelopathy (DCM)
# DCM project
This project aims to add relevant metrics to assess cervical spinal cord compression for degenerative cervical myolopathy in collaboration with Allan Martin.

Objectives:
1. Normalize with PAM50 template relevant metrics to assess compression.
2. Add other relevant metrics.
3. Compare on a pathological dataset the normalized MSCC vs non-normalized. 

## Normalize metrics with PAM50 template

Normalize each metric from `sct_process_segmentation` with the corresponding level in PAM50 template. After, use the .csv file in `sct_compute_mscc` and specify the compressed level, upper and lower.

image

In `sct_process_segmentation`, add the option `-normalize PAM50`. Maybe only use with perlevel?

In `sct_compute_mscc`:

**Input** : 
* `.csv` output file from `sct_process_segementation`
*  Compressed level(s)

### CSA for R-L

Compute CSA for right and left spinal cord by seperating from mid-point of R-L diameter.

image

**TODO**: Check in scikit-image how to extract the mid point (we use scikit-image to get the axes for AP/RL diam, so it’s a matter of fetching more outputs from that library)

### Triangular compression
Compression can also have a triangular shape. However, metrics like symmetry/solidity would not be able to detect this compression.

**Proposition:**
1. With the spinal cord segmentation, scale compressed spinal cord to PAM50 using CSA. 
2. Register to PAM50 template (center of mass or centermassrot)
3. Compute similarity metric (Dice, Hausdorff )

image

**TODO**: Make sure there isn’t a metric that can already assess this triangulation (another that is currently not implemented in process_seg)

sandrinebedard:
#### GitHub

#### PR contents

## Description
This PR adresses #3733 . The concept is to bring metrics generated in `sct_process_segmentation` to the PAM50 anatomical space in order to do slicewise comparison. The metrics need to be computed slicewise and the vertfile needs to be specified. 2 .csv files will be output.

- Adds the option `-normalize PAM50` in `sct_process_segmentation`
- Brings the metrics in the PAM50 anatomical dimensions in `metrics_to_PAM50.py`
    - Interpolates metrics perlevel, except the first and last where scaling is applied
- Aggregates metrics and save in a second .csv file.

To run the script:
```
sct_process_segmentation -i <segmentation> -perslice 1 -vertfile <labeled segmentation> -normalize-PAM50 1
```

## Linked issues

Fixes #3733. (We should reopen the issue after this PR is merged, as there are other ideas in the issue.)


jcohenadad (commented):
for the `SUM(length)` column, put `n/a` instead

sandrinebedard:
TODO: change the output csv file with `_PAM50` only!

joshuacwnewton (commented):
Just a quick note (will add a deeper review tomorrow)

> Note : -vert needs to be specified for now see issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3976 --> https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3113

Based on a quick glance at #3113, I feel like it should be relatively quick to solve (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3113#issuecomment-748633121).

So, I think it might be worth trying to open a separate PR to fix #3113, then updating this PR to remove the requirement for `-vert`?

sandrinebedard:
> So, I think it might be worth trying to open a separate PR to fix #3113, then updating this PR to remove the requirement for `-vert`?

Yes good idea! I'll work on it!

sandrinebedard:
The test suite won't work until https://github.com/spinalcordtoolbox/sct_testing_data/pull/11 is merged

testing/batch_processing/cached_results/t2/csa_pam50_PAM50.csv
```diff
@@ -0,0 +1,885 @@
...
+"2023-01-19 14:37:23","git-sb/3733-bring-metrics-to-pam50-786f6302859e392e257a498576df5a0f55c2315e","/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/data/sct_example_data/t2/t2_seg.nii.gz","81","20","","","","","","","","","","","","","","","","","",""
+"2023-01-19 14:37:23","git-sb/3733-bring-metrics-to-pam50-786f6302859e392e257a498576df5a0f55c2315e","/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/data/sct_example_data/t2/t2_seg.nii.gz","82","20","","","","","","","","","","","","","","","","","",""
+"2023-01-19 14:37:23","git-sb/3733-bring-metrics-to-pam50-786f6302859e392e257a498576df5a0f55c2315e","/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/data/sct_example_data/t2/t2_seg.nii.gz","83","20","","","","","","","","","","","","","","","","","",""
+"2023-01-19 14:37:23","git-sb/3733-bring-metrics-to-pam50-786f6302859e392e257a498576df5a0f55c2315e","/home/GRAMES.POLYMTL.CA/sebeda/spinalcordtoolbox/data/sct_example_data/t2/t2_seg.nii.gz","84","20","","","","","","","","","","","","","","","","","",""
```
joshuacwnewton, line 6: I notice that the generated `_PAM50.csv` file has a lot of extra rows:

- the PAM50 template spans `VertLevel = 1-20`
- `t2.nii.gz` spans only `VertLevel 2-5`
- So, most of the columns are blank, because `VertLevel 1` and `VertLevel 6-20` have no metrics.

Is there a way to generate this CSV file without generating the unnecessary rows?

joshuacwnewton (changes requested):
To start out, my comments are mainly focused on `sct_process_segmentation`. My goal is to see if we can make the changes simpler/smaller by reusing the existing functionality the best we can.

(I haven't looked at the `metrics_to_PAM50.py` script yet (I'm treating it like a black box for now), but I'll give a more thorough review of that script once we iron out the changes to the CLI script.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -422,37 +439,60 @@ def main(argv: Sequence[str]):
...
     metrics_agg_merged = merge_dict(metrics_agg)
+    # Save metrics in PAM50 anatomical dimensions
+    if arguments.normalize == 'PAM50':
+        metrics_agg_merged_PAM50_space = merge_dict(metrics_agg_PAM50_space)
+        file_out_norm_PAM50 = add_suffix(file_out, '_PAM50')
+        save_as_csv(metrics_agg_merged_PAM50_space, file_out_norm_PAM50, fname_in=fname_segmentation, append=append)
+        display_open(file_out_norm_PAM50)
+
     # Normalize CSA values (MEAN(area))
     if arguments.normalize is not None:
```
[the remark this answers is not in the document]
joshuacwnewton: Actually, this suggestion might be unnecessary if we adopt the other suggestion from https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977#discussion_r1086893697.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -422,37 +439,60 @@ def main(argv: Sequence[str]):
...
+                                                                            distance_pmj=distance_pmj, perslice=perslice,
+                                                                            perlevel=perlevel, fname_vert_level=fname_vert_level_PAM50,
+                                                                            group_funcs=group_funcs, length_pmj=length_from_pmj)
+
     metrics_agg_merged = merge_dict(metrics_agg)
+    # Save metrics in PAM50 anatomical dimensions
+    if arguments.normalize == 'PAM50':
+        metrics_agg_merged_PAM50_space = merge_dict(metrics_agg_PAM50_space)
+        file_out_norm_PAM50 = add_suffix(file_out, '_PAM50')
+        save_as_csv(metrics_agg_merged_PAM50_space, file_out_norm_PAM50, fname_in=fname_segmentation, append=append)
```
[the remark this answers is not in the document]
sandrinebedard: yes we need both files, afterwards for when computing the MSCC script : see  PR #4003, I am not against having to run it twice, but it will still be recomputing the same morphometrics on the segmentation before adjusting with PAM50

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -232,7 +237,9 @@ def get_parser():
              "For more details on the subjects and methods used to create the models, go to: "
              "https://github.com/sct-pipeline/ukbiobank-spinalcord-csa#readme \n"  # TODO add ref of the paper
              "Given the risks and lack of consensus surrounding CSA normalization, we recommend thoroughly reviewing "
-             "the literature on this topic before applying this feature to your data.\n"
+             "the literature on this topic before applying this feature to your data.\n\n"
+             "To normalize every metric with the PAM50 template perslice:"
+             "-normalize PAM50"
```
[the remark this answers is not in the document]
sandrinebedard: yes it is a good idea!

spinalcordtoolbox/scripts/sct_process_segmentation.py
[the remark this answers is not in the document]
joshuacwnewton: That's a good point -- We would be recomputing the same `metrics` object for both calls:

<details>

Without `-normalize PAM50`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/72b6317c65272292a563e3b8778b2e7f80dfe448/spinalcordtoolbox/scripts/sct_process_segmentation.py#L419-L423

With `-normalize PAM50`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/72b6317c65272292a563e3b8778b2e7f80dfe448/spinalcordtoolbox/scripts/sct_process_segmentation.py#L419-L428

</details>

That said, we kind of already repeat morphometrics computations? For example, in the ["Computing metrics"](https://spinalcordtoolbox.com/user_section/tutorials/registration-to-template/shape-metric-computation.html) section of the tutorials:

image

For several of these commands, they perform the same morphometric computation -- the only difference is the aggregation.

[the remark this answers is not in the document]
joshuacwnewton: One other factor is the idea that "1 call to `sct_process_segmentation` == 1 output csv file". I think that feels intuitive to me? And I think there's a benefit to keeping that pattern, even if it means repeating some of the computation.

testing/batch_processing/cached_results/t2/csa_pam50_PAM50.csv
sandrinebedard: the way to do it is to specify `-vert` ironically haha! But in the context we want to use it, it is nice to have the exact same size of .csv file --> we want to use the normalized values in a large database to then compare with patients

spinalcordtoolbox/scripts/sct_process_segmentation.py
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/2478d5d75100e4fb5bd88836764c4c5d4089f109, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/5cc852019e2ef44f511ab424eb003f6ae29d052a and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/55d0cb95f9ad60f9b7358458cd3c329ef0a03541

testing/batch_processing/cached_results/t2/csa_pam50_PAM50.csv
joshuacwnewton: > But in the context we want to use it, it is nice to have the exact same size of .csv file --> we want to use the normalized values in a large database to then compare with patients

Ahhhh, I see! I suppose it is a little extra work to manipulate the CSV files if they're all different sizes... :thinking:

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -422,37 +444,41 @@ def main(argv: Sequence[str]):
...
+            if not os.path.isfile(path_model):
+                parser.error('Invalid choice of predictors in -normalize. Please specify sex and brain-volume or sex, brain-volume and thalamus-volume.')
+            # Get normalization model
+            # Models are generated with https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/master/pipeline_ukbiobank/cli/compute_stats.py
+            # TODO update link with release tag.
+            data_predictors = pd.read_csv(path_model, index_col=0)
+            # Add interaction term
+            data_subject['inter-BV_sex'] = data_subject['brain-volume']*data_subject['sex']
+            for line in metrics_agg_merged.values():
+                line['MEAN(area)'] = normalize_csa(line['MEAN(area)'], data_predictors, data_subject)
```
joshuacwnewton, line 481: Small nit: I think the `if arguments.normalize != 'PAM50':` can be removed too, and these changes can be undone (back to their original state).

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -232,7 +234,15 @@ def get_parser():
              "For more details on the subjects and methods used to create the models, go to: "
              "https://github.com/sct-pipeline/ukbiobank-spinalcord-csa#readme \n"  # TODO add ref of the paper
              "Given the risks and lack of consensus surrounding CSA normalization, we recommend thoroughly reviewing "
-             "the literature on this topic before applying this feature to your data.\n"
+             "the literature on this topic before applying this feature to your data.\n\n"
```
joshuacwnewton, line 237: Nit: There's an extra `\n` from the old description.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -373,6 +383,7 @@ def main(argv: Sequence[str]):
 
     file_out = os.path.abspath(arguments.o)
     append = bool(arguments.append)
+
```
joshuacwnewton, line 386: Nit: There are a few of these whitespace changes remaining in the file. Could you please revert them to minimize the amount of noise in the PR? 

(See also: https://medium.com/google-developer-experts/how-to-pull-request-d75ac81449a5#e07e)

joshuacwnewton (commented):
Thanks so much for the `sct_process_segmentation` cleanup! :smile: 

I just had a few small nitpicks for now, but I'll start taking a look at the `metrics_to_PAM50.py` file, too.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -47,11 +49,11 @@ def __call__(self, parser, namespace, values, option_string=None):
         pred = values[::2]
         val = values[1::2]
         if len(pred) != len(val):
-            parser.error("Values for normalization need to be specified for each predictor.")
+            raise parser.error("Values for normalization need to be specified for each predictor.")
```
joshuacwnewton, line 52: I don't think these `raise` statements are necessary, either? 

The [`.error`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) method of `ArgumentParser` objects will exit on their own, and don't need `raise` like other error types do.

spinalcordtoolbox/scripts/sct_process_segmentation.py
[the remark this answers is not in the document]
sandrinebedard: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/41c9b6dc3778b0c85cdc0893f2d5e3d641ae13e3

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/f35f8a13d0aa8bb9096ca8af49b60337389f4531

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/f80dfbc653bb7ae7f4a5d021537b48e43b22f1b7

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/9e06ffcab367ca361319f3ec584bbd19fe23c2a9

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/f35f8a13d0aa8bb9096ca8af49b60337389f4531

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Sorry to nitpick more, but I think only one of the whitespace lines was fixed -- there are 3 more in the file, hehe. :sweat_smile:

sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/1e56d4d334f5d4476a8d56c4d20e932faa403348

testing/batch_processing/cached_results/t2/csa_pam50_PAM50.csv
sandrinebedard: hmm, we could force the parameter `-vert 1:20` when we specify `normalize-PAM50 1` since it would still have the same size of .csv files

joshuacwnewton: Just out of curiosity, what kind of data analysis tool would you be using for subject comparison of the csv files in your project?

Say the csv files are different sizes -- couldn't you like, do the same thing that this PR is doing now (initializing all 991 slices with `np.nan`, then copying over values slice-by-slice), but move that init-and-copy logic into your analysis script?

sandrinebedard: the idea here, we want to create a database of healthy suvjects with their metrics in the normative PAM50 space, so we will use it, but the idea would be to make it public, that's why we discussed with julien that it would be better to have the same format for all files! (we use it in #4003 by the way!)

spinalcordtoolbox/metrics_to_PAM50.py
```diff
@@ -0,0 +1,121 @@
+# Functions to interpolate metrics (from sct_process_segmentation) into the PAM50 anatomical dimensions
```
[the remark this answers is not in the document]
sandrinebedard: we can create an issue about this, good idea!

testing/batch_processing/cached_results/t2/csa_pam50_PAM50.csv
joshuacwnewton: Ahhhh, gotcha!

Just to clarify, what do you mean by "database"? Just like, a collection of output CSV files in a repo? Or, something more structured?

sandrinebedard: For now just a collection of CSV files (not even on a repo for now), we will also include a participants.tsv with information on the subjects, or now we are populating it! Maybe when we are more advance we will have something more sophisticated!

joshuacwnewton: Ahhh, it makes perfect sense to want to have extra rows, then.

Hmm. What if we used `-verbose 2` for this? For example:

```diff
    if normalize_pam50:
        fname_vert_level_PAM50 = os.path.join(__data_dir__, 'PAM50', 'template', 'PAM50_levels.nii.gz')
-       metrics_PAM50_space = interpolate_metrics(metrics, fname_vert_level_PAM50, fname_vert_level)
+       keep_blank_rows = True if verbose == 2 else False
+       metrics_PAM50_space = interpolate_metrics(metrics, fname_vert_level_PAM50, fname_vert_level, keep_blank_rows)
        metrics = metrics_PAM50_space  # Set metrics to the metrics in PAM50 space to use instead
        fname_vert_level = fname_vert_level_PAM50  # Set vertebral levels to PAM50
```

That way, you can specify `-v 2` for your project to get the "full" output, but the default behavior (`-v 1`) is to output only the slices that are available in the image (which I think matches user expectations for `sct_process_segmentation`).

sandrinebedard: yes we could do that, but this CSV file would also be an input of the updated version of `sct_compute_mscc` in #4003 just to consider! (if we do this just to make it compatible with the other feature)

joshuacwnewton: Thank you for the heads up! I haven't looked at #4003 yet, but I'll take some time to read it over and make sure I understand it before proposing further changes.

joshuacwnewton:
> The test suite won't work until [spinalcordtoolbox/sct_testing_data#11](https://github.com/spinalcordtoolbox/sct_testing_data/pull/11) is merged.

To address this, I've updated this PR to use [`sct_testing_data-PR3977.zip`](https://github.com/spinalcordtoolbox/sct_testing_data/releases/download/r20210330230310/sct_testing_data-PR3977.zip), a manually-created (temporary) archive containing the new `sct_testing_data` file, to be used only for testing purposes.

Once https://github.com/spinalcordtoolbox/sct_testing_data/pull/11 is _actually_ merged, we can create a new `sct_testing_data` release, then update the link.

batch_processing.sh
```diff
@@ -93,6 +93,8 @@ sct_process_segmentation -i t2_seg.nii.gz -vert 2:3 -o csa_c2c3.csv
 sct_detect_pmj -i t2.nii.gz -c t2 -qc "$SCT_BP_QC_FOLDER" 
 # Compute cross-section area at 60 mm from PMJ averaged on a 30 mm extent
 sct_process_segmentation -i t2_seg.nii.gz -pmj t2_pmj.nii.gz -pmj-distance 60 -pmj-extent 30 -qc "$SCT_BP_QC_FOLDER" -qc-image t2.nii.gz -o csa_pmj.csv
+# Compute morphometrics in PAM50 anatomical dimensions
+sct_process_segmentation -i t2_seg.nii.gz -vertfile t2_seg_labeled.nii.gz -perslice 1 -normalize PAM50 -o csa_pam50.csv
```
joshuacwnewton, line 97: It looks like the `pytest` tests now pass. However, the `batch_processing.sh` test fails, due to this mentioning `-normalize PAM50` instead of `-normalize-PAM50`.

```suggestion
# Compute morphometrics in PAM50 anatomical dimensions
sct_process_segmentation -i t2_seg.nii.gz -vertfile t2_seg_labeled.nii.gz -perslice 1 -normalize PAM50 -o csa_pam50.csv
```
```suggestion
# Compute morphometrics in PAM50 anatomical dimensions
sct_process_segmentation -i t2_seg.nii.gz -vertfile t2_seg_labeled.nii.gz -perslice 1 -normalize-PAM50 -o csa_pam50.csv
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -393,10 +403,13 @@ def main(argv: Sequence[str]):
     fname_pmj = arguments.pmj
     distance_pmj = arguments.pmj_distance
     extent_pmj = arguments.pmj_extent
+    normalize_pam50 = arguments.normalize_PAM50
     path_qc = arguments.qc
     qc_dataset = arguments.qc_dataset
     qc_subject = arguments.qc_subject
 
+    if normalize_pam50 and (fname_vert_level is None or not perslice):
+        parser.error("Option '-normalize PAM50' requires options '-vertfile' and '-perslice 1'.")
```
joshuacwnewton, line 412: It might be good to split this into two different errors?

- for `fname_vert_level`, we require `-vertfile` to _exist_, so it might be better to move this part of the error into the section where we check for file existence, then word it similar to the `-vert` warning:

  https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/cb7cd19eb9478a0a7301374e997f8c22768bc2e3/spinalcordtoolbox/scripts/sct_process_segmentation.py#L393-L394

  "The vertebral level file must exist to use `-normalize-PAM50`."

- For `perslice`, we can keep the existing check.

joshuacwnewton: Also, small nit, but this still says `-normalize PAM50`.

batch_processing.sh
joshuacwnewton: Fixed in [99caabe](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/99caabe4ad9bd5e145cc10ca5d148d70d6b0c6c8).

batch_processing.sh
```diff
@@ -93,6 +93,8 @@ sct_process_segmentation -i t2_seg.nii.gz -vert 2:3 -o csa_c2c3.csv
 sct_detect_pmj -i t2.nii.gz -c t2 -qc "$SCT_BP_QC_FOLDER" 
 # Compute cross-section area at 60 mm from PMJ averaged on a 30 mm extent
 sct_process_segmentation -i t2_seg.nii.gz -pmj t2_pmj.nii.gz -pmj-distance 60 -pmj-extent 30 -qc "$SCT_BP_QC_FOLDER" -qc-image t2.nii.gz -o csa_pmj.csv
+# Compute morphometrics in PAM50 anatomical dimensions
+sct_process_segmentation -i t2_seg.nii.gz -vertfile t2_seg_labeled.nii.gz -perslice 1 -normalize-PAM50 1 -o csa_pam50.csv
```
joshuacwnewton, line 97: I notice that in this sample command, we use `-vertfile t2_seg_labeled.nii.gz`.

But, for `sct_process_segmentation`, the default value for `-vertfile` is `./label/template/PAM50_levels.nii.gz`. (i.e. the template level file warped to the subject space).

The two labeled segmentations are very similar, but not equal:

ezgif-5-00081ef359

What is the difference between using the label files generated by `sct_label_vertebrae` vs the label files generated by `sct_warp_template`? Is there a "correct" choice for performing the normalization here?

batch_processing.sh
joshuacwnewton: It looks like this is throwing a new error now:

```
set +v
~~~
Version:         git-HEAD-fd16b0e4c927eeabb282304998e38ddb797b8206
Ran on:          Linux fv-az133-476 5.15.0-1031-azure
Duration:        0hrs 18min 45sec
---
t2/csa_c2c3.csv:         73.87711295363036  [Row 1, MEAN(area)]
t2/csa_pmj.csv:          73.89298021190447  [Row 1, MEAN(area)]
  File "testing/batch_processing/test_batch_processing.py", line 68, in <module>
    display_batch_processing_results()
  File "testing/batch_processing/test_batch_processing.py", line 63, in display_batch_processing_results
    value = get_csv_float_value(OUTPUT_DIR/csv_filepath, row, column)
  File "testing/batch_processing/test_batch_processing.py", line 28, in get_csv_float_value
    with open(csv_filepath, newline='') as csvfile:
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_example_data/t2/csa_pam50_PAM50.csv'
```

I guess because we got rid of the extra `_PAM50.csv` file, we need to just use the normal output file (`csa_pam50.csv`).

joshuacwnewton: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/1a44606f4be4a7ff34def3b23621f49d35c1d151.

joshuacwnewton: Ah shoot, we have yet another error:

```
  File "testing/batch_processing/test_batch_processing.py", line 68, in <module>
    display_batch_processing_results()
  File "testing/batch_processing/test_batch_processing.py", line 63, in display_batch_processing_results
    value = get_csv_float_value(OUTPUT_DIR/csv_filepath, row, column)
  File "testing/batch_processing/test_batch_processing.py", line 31, in get_csv_float_value
    return float(value)
ValueError: could not convert string to float: ''
batch_processing.sh: line 270: ./venv_sct/Scripts/python.exe: No such file or directory
```

The problem is because the cached .csv file starts at slice 80:

image

While the generated .csv file starts at slice 0:

image

Meaning that we index different rows for both the cached and generated csv files.

joshuacwnewton: We can fix this by either adding rows to the cached file, or changing the code to exclude PAM50 slices without vertebral levels...

Of course, all of this is kind of moot if we end up implementing https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977#discussion_r1089294999.

sandrinebedard: if we change the test for batch processing, on which row to check, would it work?

batch_processing.sh
sandrinebedard: no necessarly, in this case, the registration to template was really good, but not all analysis require registration to template, so just using  `sct_label_vertebrae` is fine for this!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: don in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/d7c5cf5f83151b647a060e3801296f1b33a56ea2

batch_processing.sh
joshuacwnewton: Any row with metrics (so, vert levels 2,3,4,5) should work. We just need to make sure that the cached file and the generated file have matching rows.

sandrinebedard: hmm, and how come the cached file is different? How is it generated?

joshuacwnewton: I'm not sure, since I think it was uploaded by you in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/fff1bdf9b3d2996876a7e5fcb496b0d1de68691a. :sweat_smile: 

I think the best plan is to figure out what we want the default `csa_pam50.csv` to look like:

1. Only slices present in the vertebral segmentation for `im`
2. Only slices present in the vertebral segmentation for `PAM50_levels.nii.gz`
3. _All_ slices present in `PAM50_levels.nii.gz`

Then, we update the behavior in the code, and regenerate the csv file from scratch by running the relevant section in `batch_processing.sh`.

joshuacwnewton: Personally I think I'm still in favor of having just the `im` slices by default, then using the PAM50 slices for `-v 2`. 

If we make that change, we'd have to regenerate the cached file anyway?

sandrinebedard: > Personally I think I'm still in favor of having just the `im` slices by default, then using the PAM50 slices for `-v 2`.
> 
> If we make that change, we'd have to regenerate the cached file anyway?

Let's go with that! In that case, it gives us flexibility for further features using that

sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/c7459d7d94f6b20f0709d662152c7eaac631d7c5

Now we can adjust the cached file!

sandrinebedard: just to be sure, how can I create the cached file? I can run batch_processing and put that file as a chached file?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -425,13 +426,17 @@ def main(argv: Sequence[str]):
...
     if normalize_pam50:
         fname_vert_level_PAM50 = os.path.join(__data_dir__, 'PAM50', 'template', 'PAM50_levels.nii.gz')
         metrics_PAM50_space = interpolate_metrics(metrics, fname_vert_level_PAM50, fname_vert_level)
+        if verbose == 2:
+            # Get all available vertebral levels from PAM50 template to only include slices from available levels in .csv file
+            levels = get_all_vertebral_level(Image(fname_vert_level_PAM50))
+        else:
+            # Get all available vertebral levels to only include slices from available levels in .csv file
+            levels = get_all_vertebral_level(Image(fname_vert_level))
         metrics = metrics_PAM50_space  # Set metrics to the metrics in PAM50 space to use instead
```
joshuacwnewton, line 438: If a user specifies `-z` + `-normalize-PAM50`, do they have to specify `-z` in terms of the original image's slices, or the PAM50's slices?

Since we're setting `metrics=metrics_PAM50_space`, I think `-z` has to be in terms of PAM50 slices, and not image slices...

Is this the behavior that we want?

joshuacwnewton: I guess it's sort of impossible to keep track of both the original image's slices _and_ the PAM50 slices once normalization is done... :sweat: 

So, there's probably not much we can do here...

batch_processing.sh
joshuacwnewton: Yes exactly!

(Normally what I do to save a bit of time is only run the commands that are necessary, and skip the T1/dMRI/fMRI processing.

spinalcordtoolbox/template.py
```diff
@@ -50,3 +50,17 @@ def get_vertebral_level_from_slice(im_vertlevel, idx_slice):
...
+    Find all available vertebral levels.
+    :return: list: List of all available vertebral levels.
+    """
+    data_vertlevel = im_vertlevel.data
+    vert_level = list(np.unique(data_vertlevel))
+    # Sort vertebral levels
+    vert_level.sort()
+    # Remove 0 value since not a level
+    vert_level.remove(0)
+    return vert_level
```
joshuacwnewton, line 66: There is a similar method already present in `Image` objects: `.getNonZeroCoordinates`.

So, what if we made this function into a `Image.getNonZeroValues` method? That way, the interface is somewhat consistent....

joshuacwnewton: (Side note: The `camelCase` of the existing method isn't great? But I figure it would be good to keep similar methods organized in similar ways. That way, if a future dev decides to one day refactor `.getNonZeroCoordinates`, then the similar method will be nearby, and will likely be included in the refactoring effort.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: since this is always perslice, the fact that a user wants to specify -z doesn't have much of an impact, there is no averaging. They will just reduce teh size of the .csv  file. The  behaviour doesn't bother me in that sence!

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -425,13 +426,17 @@ def main(argv: Sequence[str]):
...
-
     if normalize_pam50:
         fname_vert_level_PAM50 = os.path.join(__data_dir__, 'PAM50', 'template', 'PAM50_levels.nii.gz')
         metrics_PAM50_space = interpolate_metrics(metrics, fname_vert_level_PAM50, fname_vert_level)
+        if verbose == 2:
+            # Get all available vertebral levels from PAM50 template to only include slices from available levels in .csv file
+            levels = get_all_vertebral_level(Image(fname_vert_level_PAM50))
+        else:
+            # Get all available vertebral levels to only include slices from available levels in .csv file
+            levels = get_all_vertebral_level(Image(fname_vert_level))
```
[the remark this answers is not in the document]
sandrinebedard: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/2e3f37b99628619bb21fd716e81eabc044f69384

testing/cli/test_cli_sct_process_segmentation.py
```diff
@@ -96,6 +97,38 @@ def test_sct_process_segmentation_check_normalize_missing_predictor(dummy_3d_mas
...
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'), '-normalize-PAM50', '1',
+                                        '-perslice', '1', '-vertfile', sct_test_path('t2', 't2_seg-manual_labeled.nii.gz'), '-o', filename])
+    with open(filename, "r") as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=',')
+        rows = list(reader)
+        row = rows[827]
+        assert row['Slice (I->S)'] == '827'
+        assert float(row['MEAN(area)']) == pytest.approx(71.96880493869594)
+        assert row['VertLevel'] == '5'
```
joshuacwnewton, line 111: It looks like with the new changes, this test is now failing:

```
_____________ test_sct_process_segmentation_check_normalize_PAM50 ______________
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/cli/test_cli_sct_process_segmentation.py", line 108, in test_sct_process_segmentation_check_normalize_PAM50
    row = rows[827]
IndexError: list index out of range
```

With the smaller CSV files, we need to pick a new row to check. :stuck_out_tongue_closed_eyes:

testing/cli/test_cli_sct_process_segmentation.py
```diff
@@ -96,6 +97,38 @@ def test_sct_process_segmentation_check_normalize_missing_predictor(dummy_3d_mas
...
+        assert e.value.code == 2
+
+
+def test_sct_process_segmentation_check_normalize_PAM50_missing_vertfile(tmp_path):
+    """ Run sct_process_segmentation with -normalize PAM50 when missing -vertfile"""
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    with pytest.raises(SystemExit) as e:
+        sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'), '-normalize-PAM50', '1',
+                                            '-perslice', '1', '-o', filename])
+        assert e.value.code == 2
```
joshuacwnewton, line 129: Since the "missing vertfile" error was changed to a `FileNotFoundError`, we need to switch the `pytest.raises` statement, as the test now fails to catch the error:

```
_____ test_sct_process_segmentation_check_normalize_PAM50_missing_vertfile _____
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/cli/test_cli_sct_process_segmentation.py", line 127, in test_sct_process_segmentation_check_normalize_PAM50_missing_vertfile
    sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'), '-normalize-PAM50', '1',
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 394, in main
    raise FileNotFoundError("The vertebral level file must exist to use -normalize-PAM50.")
FileNotFoundError: The vertebral level file must exist to use -normalize-PAM50.
```

joshuacwnewton (changes requested):
The new changes LGTM! But, as a result of the newest changes, we now have test failures.

testing/cli/test_cli_sct_process_segmentation.py
sandrinebedard: changed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/660581eae89cd0cb454988de0bc07addf5fa3c58

testing/cli/test_cli_sct_process_segmentation.py
sandrinebedard: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/55699d2571a351a354c4ca662704e8474ceeb08b

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -8,8 +8,7 @@
 #
```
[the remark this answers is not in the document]
joshuacwnewton: Resolved by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/5722cab228cd976b91d3b6f1a6c2a3ba2f2f6a78.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -397,6 +410,8 @@ def main(argv: Sequence[str]):
     qc_dataset = arguments.qc_dataset
     qc_subject = arguments.qc_subject
 
+    if normalize_pam50 and not perslice:
+        parser.error("Option '-normalize-PAM50' requires option '-perslice 1'.")
```
joshuacwnewton, line 414: Just to double-check: Why is `-perslice 1` mandatory?

Out of curiosity, I tried removing this error, then specifying `-perlevel 1` instead of `-perslice 1`, and the results seem okay?

```
sct_process_segmentation -i t2_seg.nii.gz -vertfile t2_seg_labeled.nii.gz -perlevel 1 -normalize-PAM50 1 -o csa_pam50.csv
```

image

Is it worth considering removing hard requirement for `-perslice 1`?

spinalcordtoolbox/download.py
```diff
@@ -32,7 +32,8 @@
     },
     "sct_testing_data": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/sct_testing_data/releases/download/r20210330230310/sct_testing_data-r20210330230310.zip",
+            # FIXME: This link should NOT be merged to master. It should be replaced via a new sct_testing_data release.
+            "https://github.com/spinalcordtoolbox/sct_testing_data/releases/download/r20210330230310/sct_testing_data-PR3977.zip",
```
joshuacwnewton, line 36: Reminder for myself to merge the `sct_testing_data` PR, create a new release, then update this link.

joshuacwnewton (commented):
I think I'm just about ready to approve this? It mostly LGTM! I just had one last question (and one more task for myself.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: No, we want afterwards to do slicewise comparisons. The idea of puting everything in the PAM50 anatomical space is that we will have a slice correspondance between subjects! If we average across levels, using the pam50 anatomical dimensions is not useful, we could compare perlevel in the native space .

joshuacwnewton: That's very fair! I can see why it's only really useful to use `-normalize-PAM50` with `-perlevel 1`.

Still... it isn't exactly a _mandatory_ step? (It's not like `-normalize-PAM50` will fail without `-perlevel 1`, because the normalization step is completely separate from the aggregation step.) 

Technically, saying that it is a hard requirement is a bit misleading... instead, `-perlevel 1` is more like a strong recommendation?. Because of this, I feel like it may be better to remove the strict requirement, and instead explain the `-perlevel 1` recommendation in the usage:

```diff
    optional.add_argument(
        '-normalize-PAM50',
        metavar=Metavar.int,
        type=int,
        choices=[0, 1],
        default=0,
-       help="Set to 1 to bring the metrics in the PAM50 anatomical dimensions perslice. -vertfile and -perslice need to be specified."
+       help="Set to 1 to bring the metrics in the PAM50 anatomical dimensions. -vertfile must be specified.\n"
+            "NB: It is strongly recommended to always use this option in combination with '-perslice 1', as this will enable you to "
+            "perform slicewise comparisons of normalized metrics between different subjects."
    )
```

That way, the user needs think about the aggregation setting that they are choosing, and will hopefully have a better understanding of _why_ they should be setting `-perlevel 1`.

sandrinebedard: I am not sure I would want to remove the mandatory `-perslice 1` since the option` -normalize-PAM50` is kind of useless without... Maybe @jcohenadad and @valosekj you have an opinion on this?

valosekj: I agree with Sandrine. The option `-normalize-PAM50` is designed to bring morphometric metrics from the subject native space to the PAM50. This normalization is done using a linear interpolation of [individual slices per level](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5722cab228cd976b91d3b6f1a6c2a3ba2f2f6a78/spinalcordtoolbox/metrics_to_PAM50.py#L10-L12). This means that we need individual slices (the option `-perslice 1`) for each level. Without the option `-perslice 1`, we would have only a single value for each level, and interpolation would not be possible.

joshuacwnewton: > This means that we need individual slices (the option `-perslice 1`) for each level. Without the option `-perslice 1`, we would have only a single value for each level, and interpolation would not be possible.

Just to clarify one thing: the PAM50 normalization step occurs before any aggregation steps. So, regardless of `-perslice 1` the normalization will be performed with the full per-slice metrics.

Whether or not we specify `-perslice 1` only affects how the per-slice metrics then get aggregated for the output CSV file.

valosekj: > > This means that we need individual slices (the option `-perslice 1`) for each level. Without the option `-perslice 1`, we would have only a single value for each level, and interpolation would not be possible.
> 
> Just to clarify one thing: the PAM50 normalization step occurs before any aggregation steps. So, the setting of `-perslice` has no effect on the normalization/interpolation steps; they will be performed with the full per-slice metrics regardless of how `-perslice` is set..
> The option `-perslice 1` only affects how the per-slice metrics get aggregated for the output CSV file. (So, the discussion is whether we should make it mandatory for the output CSV to be per-slice, or whether it should be explained in the usage only.)

I see! Now I get what you mean! Exactly; metrics are: 1) computed slice-wise by [`compute_shape`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5722cab228cd976b91d3b6f1a6c2a3ba2f2f6a78/spinalcordtoolbox/scripts/sct_process_segmentation.py#L423), 2) interpolated by [`interpolate_metrics`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5722cab228cd976b91d3b6f1a6c2a3ba2f2f6a78/spinalcordtoolbox/scripts/sct_process_segmentation.py#L430), and 3) aggregated using [`aggregate_per_slice_or_level`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5722cab228cd976b91d3b6f1a6c2a3ba2f2f6a78/spinalcordtoolbox/scripts/sct_process_segmentation.py#L463). 
So, yes, as you mentioned, technically, `-normalize-PAM50` works also with `-perlevel 1` (instead of `-perslice 1`) because interpolation/normalization is done slice-wise regardless of how `-perslice` is set.

sandrinebedard: yes, it is correct in the sens of the interpolation, however, it would give the same results if it is `-perlevel 1` than if we don't use `-normalize-PAM50` since it is averaged on the entire level. The only thing that would differ are the slices numbers. This is what I mean when it is useless. I will discuss in our meeting with Julien tomorrow and get back to you @joshuacwnewton !

joshuacwnewton: Ahhhh, I see!!! In that case, I think it's probably fine to keep things as they are!

spinalcordtoolbox/download.py
joshuacwnewton: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977/commits/28ad78bfeefb883d705068475935f18734ea4b50.

joshuacwnewton (approved):
Assuming the tests pass with the new dataset release, this LGTM!

Links to next steps:

- Incorporating changes into https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4019 and

PR #3986, opened 2022-12-23
Improve the readability of temporary directory names throughout SCT

### LINKED ISSUE #3978: `sct_propseg` - wrong name of temp centerline folder
### Description

Temporary folder created by `sct_propseg` to store a centerline includes misleading `-label_vertebrae-` string in its name:

```
$ sct_propseg -i sub-mon001_ses-M0_T2w.nii.gz -c t2

--
Spinal Cord Toolbox (git-master-3cfab1dbff6a1467ca2fad03d070eb2762dfa5e1*)

sct_propseg -i sub-mon001_ses-M0_T2w.nii.gz -c t2
--

Image header specifies datatype 'int16', but array is of type 'float64'. Header metadata will be overwritten to use 'float64'.
Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221208085800.685980-propseg-36oudp7g)
Image header specifies datatype 'int16', but array is of type 'float64'. Header metadata will be overwritten to use 'float64'.
Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221208085800.797748-label_vertebrae-pf839gpv)
Creating temporary folder (/var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221208085800.798363-qkrg225b)
Remove temporary files...
...
```

Since the temporary folder contains a centerline:

```
$ ls /var/folders/v8/8v423b3x4g58mp5g1p33wsh00000gn/T/sct-20221208085800.797748-label_vertebrae-pf839gpv
centerline_optic.nii.gz
```

It should include `-centerline-` in its name instead of `-label_vertebrae-`.

Maybe related to https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3975

joshuacwnewton:
## Description

This PR originated in issue #3978, when @valosekj noticed that `sct_propseg`'s centerline folder name was erroneously called "label_vertebrae".

Separately, however, I've held the opinion that tempdir names in SCT [aren't as clear as they could be](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3978#issuecomment-1342822951). I've been wanting to update _every_ tempdir name; issue #3978 is just the straw that broke the camel's back.

So, this PR aims to improve the quality of tempdir names in SCT. This PR contains two main parts:

1. Updating the `tmp_create` function in `fs.py`, to improve the general structure of tmpdir names.
2. Updating the `basename` parameters in each _call_ to `tmp_create`, so that they all follow a standard.

The "naming standardization" is explained in commit 0c3abd0310b51d21682342ea77d041033d6a9654:

1. Only lowercase names
2. Words are delimited with `-`, not `_` (due to the fact that `_` is already in use to delimit other parts of the temporary directory name e.g. date, time, etc.)
3. Basenames should roughly match the name of the CLI script or API function that the temporary directory belongs to
   (minus `sct_` for CLI scripts).
4. If the function name itself doesn't provide enough context, then another identifier can be added (e.g. adding `propseg-` to the propseg-specific `rescale-header` function.)

Lastly, #3978 itself is fixed by 80b62205c698eb10cca74afac5130a07023a217a.

## Linked issues

Fixes #3978.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -50,7 +50,7 @@ def check_and_correct_segmentation(fname_segmentation, fname_centerline, folder_
     """
     printv('\nCheck consistency of segmentation...', verbose)
     # creating a temporary folder in which all temporary files will be placed and deleted afterwards
-    path_tmp = tmp_create(basename="propseg")
+    path_tmp = tmp_create(basename="propseg-check-segmentation")
     im_seg = convert(Image(fname_segmentation))
     im_seg.save(os.path.join(path_tmp, "tmp.segmentation.nii.gz"), mutable=True, verbose=0)
     im_centerline = convert(Image(fname_centerline))
@@ -421,7 +421,7 @@ def func_rescale_header(fname_data, rescale_factor, verbose=0):
     header_rescaled.set_qform(qform)
     # the data are the same-- only the header changes
     img_rescaled = nib.nifti1.Nifti1Image(img.get_data(), None, header=header_rescaled)
-    path_tmp = tmp_create(basename="propseg")
+    path_tmp = tmp_create(basename="propseg-rescale-header")
     fname_data_rescaled = os.path.join(path_tmp, os.path.basename(add_suffix(fname_data, "_rescaled")))
     nib.save(img_rescaled, fname_data_rescaled)
     return fname_data_rescaled
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -563,7 +563,6 @@ def propseg(img_input, options_dict):
         printv('ERROR: your input image needs to be 3D in order to be segmented.', 1, 'error')
 
     path_data, file_data, ext_data = extract_fname(fname_data)
-    path_tmp = tmp_create(basename="label_vertebrae")
 
     # rescale header (see issue #1406)
     if rescale_header != 1.0:
@@ -611,6 +610,7 @@ def propseg(img_input, options_dict):
     # If using OptiC
     elif use_optic:
         image_centerline = optic.detect_centerline(image_input, contrast_type, remove_temp_files)
+        path_tmp = tmp_create(basename="propseg-centerline-optic")
         fname_centerline_optic = os.path.join(path_tmp, 'centerline_optic.nii.gz')
         image_centerline.save(fname_centerline_optic)
         cmd += ["-init-centerline", fname_centerline_optic]
--- a/spinalcordtoolbox/utils/fs.py
+++ b/spinalcordtoolbox/utils/fs.py
@@ -13,12 +13,11 @@ from .sys import printv
 logger = logging.getLogger(__name__)
 
 
-def tmp_create(basename=None):
+def tmp_create(basename):
     """Create temporary folder and return its path
     """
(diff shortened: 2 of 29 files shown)


spinalcordtoolbox/utils/fs.py
```diff
@@ -13,14 +13,12 @@
...
 
 
-def tmp_create(basename=None):
+def tmp_create(basename):
     """Create temporary folder and return its path
     """
-    prefix = "sct-%s-" % datetime.datetime.now().strftime("%Y%m%d%H%M%S.%f")
-    if basename:
-        prefix += "%s-" % basename
+    prefix = f"sct_{datetime.datetime.now().strftime('%Y-%m-%d_%H:%M:%S')}_{basename}_"
```
joshuacwnewton, line 19: Oops! I think I've caused an error on native Windows machines by [using the `:` character](https://stackoverflow.com/a/31976060) in a directory name.

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\sct_2022-12-23_20:36:28_install-data_qdfid74d'
```

I'll swap this out for `-`.

joshuacwnewton, line 19: Oops! I think I've caused an error on native Windows machines by [using the `:` character](https://stackoverflow.com/a/31976060) in a directory name.

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\sct_2022-12-23_20:36:28_install-data_qdfid74d'
```

I'll swap this out for `-`.

spinalcordtoolbox/scripts/sct_propseg.py
```diff
@@ -611,6 +610,7 @@ def propseg(img_input, options_dict):
     # If using OptiC
     elif use_optic:
         image_centerline = optic.detect_centerline(image_input, contrast_type, remove_temp_files)
+        path_tmp = tmp_create(basename="propseg-centerline-optic")
```
mguaypaq, line 613: This isn't a new issue, but I'm noticing it now, so I might as well comment: I'm a little concerned that this call to `tmp_create()` doesn't have any corresponding logic to remove it if `remove_temp_files` is set. That means that `sct_propseg` leaves the generated centerline lying around, even after it's done using it.

But, as you noted in your commit message, fixing this would be out of scope for this PR.

mguaypaq (approved):
Thanks for this comprehensive fix to the temporary directory names. In particular, I appreciate that you spelled out the guidelines you used, and that you made all the prefixes unique.

spinalcordtoolbox/utils/fs.py
```diff
@@ -16,7 +16,7 @@
 def tmp_create(basename):
     """Create temporary folder and return its path
     """
-    prefix = f"sct_{datetime.datetime.now().strftime('%Y-%m-%d_%H:%M:%S')}_{basename}_"
+    prefix = f"sct_{datetime.datetime.now().strftime('%Y-%m-%d_%H.%M.%S')}_{basename}_"
```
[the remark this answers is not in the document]
joshuacwnewton: I think I wanted to mimic human-readable times (`HH:MM.SS`), and found `_`/`-` a little less easy to grok?

But, I think you're right here. Better to be safe than sorry... I wouldn't want a folder to be misinterpreted as a file.

valosekj:
Amazing! Thank you for this fix! 🎉

Issue #3968, opened 2022-11-29
`automate-changelog-PR` workflow works but returns an error


mguaypaq:
See [this run of the workflow](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/3570136396), and several previous ones, which has this error message:
> Invalid request. For 'properties/milestone', 5.8 is not an integer, string, or null.

I suspect this is caused by the change from version `5.3.0`, which would parse as a string, and version `5.4`, which would parse as a floating point number.

The workflow still works fine, so this is low priority, but it still bugs me a bit.

PR #3990, opened 2023-01-06
Use milestone due date

mguaypaq:
This is a follow-up to [changelog#33](https://github.com/neuropoly/changelog/pull/33), to use the new command-line flag `--use-milestone-due-date`, so that the changelog text refers to the planned release date, rather than whatever date the draft is created on.

### DIFF
--- a/.github/workflows/automate-changelog-PR.yml
+++ b/.github/workflows/automate-changelog-PR.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Update '${{ github.event.inputs.fname_changelog }}' for '${{ github.event.repository.name }}'
       run: |
         cd ${{ github.event.repository.name }}
-        changelog spinalcordtoolbox/spinalcordtoolbox --update --name "${{ github.event.inputs.fname_changelog }}" --milestone ${{ github.event.inputs.milestone_title }}
+        changelog spinalcordtoolbox/spinalcordtoolbox --update --name "${{ github.event.inputs.fname_changelog }}" --milestone ${{ github.event.inputs.milestone_title }} --use-milestone-due-date
         rm -f "${{ github.event.inputs.fname_changelog }}.bak"
     - name: Create pull request for updated changelog
       uses: peter-evans/create-pull-request@v3


joshuacwnewton (approved):
LGTM!

Also, since we're modifying the changelog workflow file anyway, I'm wondering if this might be a good time to fix that pesky bug that was causing errors for float-like milestone names (`X.Y`)? --> https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3968

mguaypaq:
> Also, since we're modifying the changelog workflow file anyway, I'm wondering if this might be a good time to fix that pesky bug that was causing errors for float-like milestone names (`X.Y`)? --> #3968

I think that's out of scope for this PR. There are still a lot of unknowns for me to figure out before I can fix #3968:
* Should we upgrade the [create-pull-request](https://github.com/marketplace/actions/create-pull-request) action from v3 to v4? I'd have to review the changelog to see what changed, and carefully read our workflow script to see if we're affected. Also, v3 is running on Node.js 12, which is [deprecated and will be removed next summer](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).
* How is the milestone argument actually passed through all the layers of indirection?
  * The workflow file is YAML, and I'm not sure what the quoting and/or escaping and/or number interpretation rules are for that.
  * Plus, how does Github Actions supply the value? Is it just string-interpolating the YAML file? Or something else?
  * This is passed to Node.js, somehow. Does Node.js do anything funny? Is the shell involved in-between somewhere?
  * It looks like this is handled by a third-party library instead of directly by the action?
  * The [Github REST API for updating an issue / pull request](https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#update-an-issue) says that the `milestone` can be null or string or integer, but how is the string interpreted? According to the [json schema for milestones](https://docs.github.com/en/rest/issues/milestones?apiVersion=2022-11-28#list-milestones), some likely candidates would be `node_id` (an opaque string) or `title`, but there are others.
  * Besides, it's not clear that the Github Action would even let us pass a string, because [somewhere along the line](https://github.com/peter-evans/create-pull-request/blob/v3/src/main.ts#L27), the value seems to go through `Number()` coercion.
* How about just using the [Github CLI](https://github.com/cli/cli#github-actions) directly instead, which is available on Github Actions runners already? I feel like I could understand and debug that more easily than any Node.js action.

PR #3995, opened 2023-01-11
Remove `gm_model` from `download.py` module and install scripts

joshuacwnewton:
## Description

This download was only ever used in the context of `sct_segment_graymatter`, which was removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1622 / https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2549.

The presence of this download is causing errors in our CI (due to OSF issues and a lack of GitHub mirror). So, I wanted to expedite the removal of the model, hence this PR.

## Linked issues

Fixes #3333.
Fixes #3994.

Issue #3994, opened 2023-01-11
OSF mirror download issues for `gm_model` in test suite (400/403 Client Error)


joshuacwnewton:
I've noticed a few installation failures in our CI, all pertaining to `sct_download_data`. Errors persist even after restarting the jobs.

All errors occur for the following `sct_download_data` entry:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a15070f635264ad05aae41db83ce31160600e930/spinalcordtoolbox/download.py#L54-L59

Some sample errors:

```python
Exception: ('Download error', [HTTPError('400 Client Error: Bad Request for url: https://files.osf.io/v1/resources/6zbyf/providers/osfstorage/583c9a8c594d9000451f6fd6?action=download&direct&version=2')])
```

```python
Exception: ('Download error', [HTTPError('403 Client Error: Forbidden for url: https://files.osf.io/v1/resources/6zbyf/providers/osfstorage/583c9a8c594d9000451f6fd6?action=download&direct&version=2')])
```

```python
  File "/__w/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.8/zipfile.py", line 1336, in _RealGetContents
    raise BadZipFile("File is not a zip file")
```

One the one hand, my hunch is that these are fluky service outages on OSF's end. 

On the other hand, this may be a good nudge to ensure we have a GitHub backup mirror for `gm_model`, since right now, it's our only OSF-only upload.

joshuacwnewton:
Wait a minute. Where is the `gm_model` actually used in SCT?

The contents are as follows:

```console
(master) joshua@tadpole:~/repos/spinalcordtoolbox/data$ ls gm_model/
fitted_data.pklz  fitted_model.pklz  info.txt  intensities.pklz  slices.pklz
```

But, note that these `.pklz` file are distinct from [`deepseg_gm_models`](https://github.com/spinalcordtoolbox/deepseg_gm_models), which contains `.hdf5` model files instead.

A quick project-wide search for `gm_model` only returns usages for `deepseg_gm_models`...

Were these model files only used in the old, now-removed `segment_graymatter` function?

joshuacwnewton:
Oh.

* https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3333

Closing this issue to instead work on the issue above.

joshuacwnewton:
[...]

PR #3996, opened 2023-01-11
Simplify weight-loading model code for `algo='dl'`

### LINKED ISSUE #3964: Investigate weight-loading model code for `algo='dl'` method
The algo='dl' feature was introduced in SCT v5.7 (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3807), and was tweaked in PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3953.

During the review process for the second PR, a few suspect lines of code were identified:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8dad90a8c6515f7e6a2bfae046dcbc2f773bb926/spinalcordtoolbox/registration/algorithms.py#L525-L543

> This seems like a convoluted way to do a fairly simple thing, but I can't say that I understand PyTorch and VoxelMorph enough to suggest a much cleaner way. Certainly out of scope for this PR, which is moving this bit of code rather than changing it.
> > _Originally posted by @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3953#discussion_r1029891356_

This issue is for following up on that discussion: 

- [x] Try to better understand PyTorch conventions
- [x] Update the code if need be

joshuacwnewton:
## Description

In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3953#discussion_r1029891356, @mguaypaq identified a snippet of code that looked convoluted and unclear, especially given the lack of inline documentation.

After some investigation in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3964#issuecomment-1376296542, we determined the purpose of the code (to load trained weights *except* the unnamed `.grid` parameters), as well as a simpler way to achieve the same behavior.

This PR makes the switch, as per https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3964#issuecomment-1376299648.

## Linked issues

Fixes #3964.


joshuacwnewton:
The CI failures in this PR are due to #3994, which will be fixed as part of #3995.

mguaypaq (approved):

Issue #3984, opened 2022-12-20
Extend `-create-viewer` to allow users to create compression disc labels (all `1` values)


valosekj:
## Feature justification

Within our `metric-normalization / automatic compression detection` [project](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977), we are looking for an optimal way, how to store the information about the level/slice of the SC compression.
It would be great to have a nii file containing single point(s) (non-zero voxels) corresponding to individual compression(s) (one patient usually has more than one compression). It could be something similar as in case of `disc-labels`, where individual non-zero voxels corresponds to individual intervertebral discs.

Points for a discussion:
- voxel values - if more than one compression is presented, which voxel value to set? `1` for the first compression, `2` for the second compression, etc.? Or the same value (`1`) across all compressions.
- a single compression can cover several slices. In such cases, it will be necessary to label all affected slices. [Such approach we currently use in our draft](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/983114df0692e8c237c39a3c6ff21af8fb2a7447/spinalcordtoolbox/scripts/sct_compute_mscc.py#L53-L58). But instead of nii file, we use txt file.
- how to effectively label the compressions across a large dataset? One possibility might be to extend the [current viewer](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ee19076472c7e42da1e15b4483ff0720772df27/spinalcordtoolbox/scripts/sct_label_utils.py#L344) used for disc labeling.

jcohenadad:
I like the voxel value approach because:
- it can be used later on to train models 
- xxx

However, we should not do '1', '2' , etc because the concept of "first/second compression" is relative. And there is no pros in distinguishing with different labels. So: we keep single voxel of value: '1' for each compression detected. So if there are two compressions: one voxel at each compression. --> we need feedback from Allan. 

in case a compression site spans multiple slices along the superior-inferior direction (which is often the case), the reported voxel will be that corresponding to the most compressed location. However, if there is a distal second compression (eg: at another level), another voxel will be done.

valosekj:
> However, we should not do '1', '2' , etc because the concept of "first/second compression" is relative. And there is no pros in distinguishing with different labels. So: we keep single voxel of value: '1' for each compression detected. So if there are two compressions: one voxel at each compression.

What do you thing about something like this?

[...]

Issue #3201, opened 2021-01-27
merge sct_dice_coefficient inside sct_maths


jcohenadad:
`sct_dice_coefficient` is effectively doing maths on the images (measuring overlap), so I would suggest to simply merge it under a new flag in `sct_maths`. The less CLIs we have, the clearer it is from UX standpoint (I think). So, something like:
~~~
sct_maths -i SEG1 -dice SEG2
~~~

joshuacwnewton:
> `sct_dice_coefficient` is effectively doing maths on the images (measuring overlap), so I would suggest to simply merge it under a new flag in `sct_maths`.

I haven't looked closely at the internals of each script, but would this logic also apply to the `sct_compute_<>` series of scripts? (`ernst_angle`, `hausdoff_distance`, etc.)

jcohenadad:
exception for `ernst_angle` as it is not based on an image, but is just a convenient script to compute the [Ernst angle](http://mriquestions.com/optimal-flip-angle.html) when doing MRI acquisition

PR #4003, opened 2023-01-19
Add new CLI script to compute normalized metric ratios (MSCC, etc.) for compressed levels

### LINKED ISSUE #3733: Add normalized metrics for degenerative cervical myelopathy (DCM)

sandrinebedard:
#### GitHub

#### PR contents

## Description
This issue is related to #3733. It computes normalized metrics ratio  (MSCC and others) using a reference dataset of healthy subject in the PAM50 anatomical space. It calls `sct_process_segmentation` to generate the spinal cord morphometrics perslice in the native space and in the PAM50 anatomical dimensions (if `-normalize` 1) to compute the metric ratio.

It relies on 3 inputs:

- A label file identifying compression sites.
- Spinal cord segmentation
- Vertebral labels

## How to test

1. Prepare the development environment by running:
    ~~~bash
    cd $SCT_DIR
    git checkout sb/3733-bring-metrics-to-pam50-add_mscc 
    git pull
    ./install_sct
    ~~~
2. Download the reference dataset of healthy controls to `$SCT_DIR/data/` by running:
    ```
    sct_download_data -d PAM50_normalized_metrics
    ```
3. Generate the 3 necessary input files from this subject file: [sub-twh018_T2w.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/10982357/sub-twh018_T2w.nii.gz)
    1. Compression labels (`sub-twh018_T2w_compression_labels.nii.gz`)
    2. Spinal cord segmentation
    3. Vertebral labels
    ~~~bash
    # NB: These steps approximate what a tutorial would look like for using this function.

    # Identify compression sites using labels
    # NB: These steps are a workaround since -create-viewer can't create multiple labels with the same value 
    #     See also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3984
    sct_label_utils -create-viewer 1 -i sub-twh018_T2w.nii.gz -o labels1.nii.gz
    sct_label_utils -create-viewer 1 -i sub-twh018_T2w.nii.gz -o labels2.nii.gz
    sct_label_utils -create-viewer 1 -i sub-twh018_T2w.nii.gz -o labels3.nii.gz
    sct_maths -i labels1.nii.gz -add labels2.nii.gz labels3.nii.gz -o sub-twh018_T2w_compression_labels.nii.gz

    # Generate spinal cord segmentation
    sct_deepseg_sc -i sub-twh018_T2w.nii.gz -c t2
    # NB: Some manual corrections were performed on the segmentation afterwards,
    #     due to the presence of spinal cord compression.

    # Get verterbral labeling
    sct_label_vertebrae -i sub-twh018_T2w.nii.gz -s sub-twh018_T2w_seg.nii.gz -c t2

    ~~~
    > (Alternatively, you can just download the 3 generated files directly here : 
[sub-twh018_T2w_testing_data.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/11063797/sub-twh018_T2w_testing_data.zip)

4. Run the following example commands:
    **Compute ratio of (diameter ap):**
    ~~~
    sct_compute_compression -i sub-twh018_T2w_seg.nii.gz -l sub-twh018_T2w_compression_labels.nii.gz -vertfile sub-twh018_T2w_seg_labeled.nii.gz -normalize 1 -o test_mscc.csv
    ~~~
    **Compute ratio of area (CSA):**
    ~~~
    sct_compute_compression -i sub-twh018_T2w_seg.nii.gz -l sub-twh018_T2w_compression_labels.nii.gz  -vertfile sub-twh018_T2w_seg_labeled.nii.gz -metric area -normalize 1 -o test_csa.csv
    ~~~
    **Test specifying sex and age range**
    ~~~
    sct_compute_compression -i sub-twh018_T2w_seg.nii.gz -l sub-twh018_T2w_compression_labels.nii.gz  -vertfile sub-twh018_T2w_seg_labeled.nii.gz  -normalize 1 -o test_mscc_age_sex.csv -sex F -age 35 60
    ~~~
    **Test without normalization**
    ~~~
    sct_compute_compression -i sub-twh018_T2w_seg.nii.gz -l sub-twh018_T2w_compression_labels.nii.gz  -vertfile sub-twh018_T2w_seg_labeled.nii.gz  -normalize 0 -o test_mscc_no_norm.csv
    ~~~
## What is left
- Remove default value for normalize
- Adapt for multiples labeled slices for one compression
- Add in documentation & tutorial

## Linked issues
Fixes #3733.


joshuacwnewton:
I've done some git surgery to clean up this PR.

> **Note**: The reason for the cleanup is because this PR's branch contained outdated, duplicated work. This is because it was developed alongside PR #3977. However, PR #3977 was reviewed, modified, and finally merged to master. So, the two branches `sb/3733-bring-metrics-to-pam50` and `sb/3733-bring-metrics-to-pam50-add-mscc` fell out of sync, with this branch containing outdated versions of the same work.

Here are the steps I performed:

<details>
<summary> git surgery </summary>

.
Before making any changes, this is what the `sb/3733-bring-metrics-to-pam50-add-mscc` branch looked like ([BACKUP](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/master...sb/3733-bring-metrics-to-pam50-add_mscc_BACKUP)):

- 23 commits (11-19-2022 to 12-08-2022)
    - These commits are duplicated with `sb/3733-bring-metrics-to-pam50` (PR [#3977](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977)).
- 21 commits (12-12-2022 to 01-11-2023)
    - These commits are unique to `sb/3733-bring-metrics-to-pam50-add-mscc`.
- 1 merge commit from `sb/3733-bring-metrics-to-pam50` (01-19-2023)
- 1 merge commit from `master` (01-19-2023)
- 27 commits (01-24-2023 to 02-01-2023)
    - These commits are unique to `sb/3733-bring-metrics-to-pam50-add-mscc`.
- 1 merge commit from `master` (02-08-2023)

My goal was to remove the 23 duplicate commits, as well as all 3 merge commits, then **rebase** the remaining (21+27) commits onto master. This way, the `-add-mscc` branch would contain only unique commits, while still being up to date with `master`.

To do this, I ran the following commands:

```bash
# Backup the existing state of the branch
git checkout sb/3733-bring-metrics-to-pam50-add-mscc
git checkout -b sb/3733-bring-metrics-to-pam50-add-mscc_BACKUP
git push -u origin sb/3733-bring-metrics-to-pam50-add-mscc_BACKUP
# Remove all commits from the existing branch
git checkout sb/3733-bring-metrics-to-pam50-add-mscc
git reset --hard fd2c839a84da564bcaf7a25ed58bbcda0e9d0874
# Move branch pointer to the tip of master
git fetch
git rebase master
# Cherry-pick the first 21 unique commits
# NB: This command caused a conflict, because one of the "unique commits" 
#     (205cd79c1958061334acd10d4a3e27195e03556f) had modified the 
#     `metrics_to_PAM50.py` file. However, this commit copied work from 
#      PR #3977, so it is unnecessary, and was dropped using `git cherry-pick --skip`.
git cherry-pick f50918a4307cbcd5ed1d70474e7dc1ee54b10add^..4af14db08fdc50c85cc8251a5782989bb1c606f7
# Cherry-pick the last 27 unique commits
git cherry-pick 32027fcbae779fa3544a6a6d2054298f07bda1dd^..018bd546f7ba4634e6678f2f2d2b1607b9875a95
# Force-push the `sb/3733-bring-metrics-to-pam50-add-mscc` branch to update PR #4003 
git push --force
```

After the above steps, the branch has 47 unique commits: 21 + 27, minus the 1 commit that was found to be unnecessary.

</details>

As a result, this branch now contains only unique work: `sct_compute_mscc` is modified, but the other files are left untouched. :tada:

You will need to update your local branches to reflect this new history. Assuming that you have no un-pushed work on your local branch, the easiest way to do this is to run the following commands:

```bash
git fetch
git checkout master
git branch -D sb/3733-bring-metrics-to-pam50-add-mscc
git checkout -b sb/3733-bring-metrics-to-pam50-add-mscc
```

Please let me know if you run into any issues due to my changes.

spinalcordtoolbox/scripts/sct_compute_mscc.py
```diff
@@ -67,28 +125,371 @@ def get_parser():
     return parser
 
 
-def mscc(di, da, db):
-    return (1 - float(di) / ((da + db) / float(2))) * 100
```
joshuacwnewton, line 71: I notice that this PR is completely removing the older, simpler `sct_compute_mscc` functionality.

In that case, we may want to keep in mind the following issue: 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3201

sandrinebedard:
We need to migrate https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4045 of the `data_normalize_metric_pam50` or git clone the repo insdide the data folder before testing!

jcohenadad (commented):
I suggest to update the body of the PR to add the following step-by-step to properly test this PR:
- after checkout, run `pip install -e .` to make sure the script is callable.

_TO BE CONTINUED_

valosekj:
Tested using:

```
cd $SCT_DIR
git checkout sb/3733-bring-metrics-to-pam50-add_mscc 
git pull
./install_sct
sct_compute_compression -i sub-twh018_T2w.csv -i-PAM50 sub-twh018_T2w_PAM50.csv -l sub-twh018_T2w_compression_labels.nii.gz
```

(I also added these commands to the [PR description](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003#issue-1549957398))

Working fine!

---

I also tested another metric (`MEAN(area)`). First, I tried:

```
sct_compute_compression -i sub-twh018_T2w.csv -i-PAM50 sub-twh018_T2w_PAM50.csv -l sub-twh018_T2w_compression_labels.nii.gz -metric MEAN(area) 
```

But, my `zsh` returned me an error: `zsh: number expected`. To fix this, I had to use `'` quotes around `MEAN(area)`:

```
sct_compute_compression -i sub-twh018_T2w.csv -i-PAM50 sub-twh018_T2w_PAM50.csv -l sub-twh018_T2w_compression_labels.nii.gz -metric 'MEAN(area)'
```

I think that this behaviour is caused by the presence of `(` and `)`.

I understand that [this syntax](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f4ba815d0baaf2512e2812354718f2c1449cdc2/spinalcordtoolbox/scripts/sct_compute_compression.py#L70) is used to be compatible with CSV column names generated by `sct_process_segmentation`. But it might be confusing for some users.

sandrinebedard:
> think that this behaviour is caused by the presence of `(` and `)` in `MEAN(area)`.
> 
> I understand that [this syntax](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f4ba815d0baaf2512e2812354718f2c1449cdc2/spinalcordtoolbox/scripts/sct_compute_compression.py#L70) is used to be compatible with CSV column names generated by `sct_process_segmentation`. But it might be confusing for some users.

What do you think if we remove the MEAN() of the argument and add it the script?

It would be:
- area
- diameter_AP
- diameter_RL
- eccentricity
- solidity

testing/cli/test_cli_sct_compute_compression.py
```diff
@@ -0,0 +1,155 @@
...
+    """ Run sct_compute_compression and chexk mscc and normalized mscc"""
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    sct_compute_compression.main(argv=['-i', dummy_metrics_csv, '-i-PAM50', dummy_metrics_csv_pam50, '-l', dummy_3d_label_label,
+                                       '-o', filename])
+    with open(filename, "r") as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=',')
+        row = next(reader)
+        assert row['Compression Level'] == '7.0'
+        assert float(row['MEAN(diameter_AP) ratio']) == pytest.approx(11.471813876181447)
+        assert float(row['Normalized MEAN(diameter_AP) ratio']) == pytest.approx(11.372089241980499)
```
sandrinebedard, line 142: I am a bit reluctant of testing the output values of the normalization since it depends on the control data.. Each time we will update https://github.com/spinalcordtoolbox/data_normalize_metric_pam50/ the test will need to be modified. I think I would remove it. @joshuacwnewton what do you think?

valosekj:
> What do you think if we remove the MEAN() of the argument and add it the script?
> 
> It would be:
> 
> * area
> * diameter_AP
> * diameter_RL
> * eccentricity
> * solidity

This sounds good!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+        default='diameter_AP',
+        choices=['diameter_AP', 'area', 'diameter_RL', 'eccentricity', 'solidity'],
+        metavar=Metavar.file,
+    )
+    optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        '-file-participants',
+        metavar=Metavar.file,
+        default='participants.tsv',
+        help='participants.tsv file of helathy controls.'
```
valosekj, line 78:
```suggestion
        help='participants.tsv file of healthy controls.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-sex',
+        type=str,
+        choices=['F', 'M'],
+        help='Select sex of healthy subjet to use for normalization. Requires participant.tsv'
```
valosekj, line 84:
```suggestion
        help='Select sex of healthy subjects to use for normalization. Requires participant.tsv'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-age',
+        type=int,
+        nargs=2,
+        metavar="[0 100]",
+        help='Select age range of healthy subjet to use for normalization. Requires participant.tsv'
```
valosekj, line 91:
```suggestion
        help='Select age range of healthy subjects to use for normalization. Requires participant.tsv'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+# ==========================================================================================
+def get_parser():
+    parser = SCTArgumentParser(
+        description='Compute spinal cord compression morpometrics such as Maximum Spinal Cord Compression as in'
+                    'Miyanji F, Furlan JC, Aarabi B, Arnold PM, Fehlings MG. Acute cervical traumatic spinal cord injury:'
+                    ' MR imaging findings correlated with neurologic outcome--prospective study with 100 consecutive patients.'
+                    ' Radiology 2007;243(3):820-827.'
+                    'Can be computed using AP diameter or other morphometrics (CSA, RL diameter, eccentricity and solidity).'
+                    'Morphometrics are normalized with a reference database where morphometrics from sct_process_segmentation '
+                    'are brought to PAM50 anatomical space using -normalize-PAM50.'
```
jcohenadad, line 40:
```suggestion
        description='Compute spinal cord compression morphometrics such as the Maximum Spinal Cord Compression [1], '
                    'the antero-posterior (AP) diameter, or other relevant spinal cord morphometrics that are output'
                    ' by the function sct_process_segmentation (CSA, RL diameter, eccentricity, solidity, etc.).'
                    ' Metrics are normalized using a database of spinal cord morphometrics built from healthy control'
                    ' subjects. This database uses the PAM50 template as an anatomical reference system.'
                    
                    '[1]: Miyanji F, Furlan JC, Aarabi B, Arnold PM, Fehlings MG. Acute cervical traumatic spinal cord injury:'
                    ' MR imaging findings correlated with neurologic outcome--prospective study with 100 consecutive' 
                    ' patients. Radiology 2007;243(3):820-827.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+        help='.nii file with compression labels. Each compression is denoted by a single voxel of value `1`. '
+             'Example: sub-001_T2w_compression_labels.nii.gz'
+    )
+    mandatoryArguments.add_argument(
+        '-i-PAM50',
+        metavar=Metavar.file,
+        required=True,
+        help='Input csv file name in PAM50 space. Example: csa_PAM50.csv.'
+        'Use sct_process_segmentation with -normalize-PAM50 to generate the file.'
+    )
```
jcohenadad, line 63: Do we want to have it mandatory? What if users want to compute MSCC without the normalization (ie: the traditional way)?

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+                    'Morphometrics are normalized with a reference database where morphometrics from sct_process_segmentation '
+                    'are brought to PAM50 anatomical space using -normalize-PAM50.'
+    )
+
+    mandatoryArguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatoryArguments.add_argument(
+        '-i',
+        metavar=Metavar.file,
+        required=True,
+        help='Input csv file name in native space. Example: csa.csv.'
```
jcohenadad, line 48:
```suggestion
        help='CSV file obtained with sct_process_segmentation. Example: csa.csv.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    mandatoryArguments.add_argument(
+        '-l',
+        metavar=Metavar.file,
+        required=True,
+        help='.nii file with compression labels. Each compression is denoted by a single voxel of value `1`. '
+             'Example: sub-001_T2w_compression_labels.nii.gz'
```
jcohenadad, line 55:
```suggestion
        help='NIfTI file that includes labels at the compression sites. Each compression site is denoted by a single voxel of value `1`. '
             'Example: sub-001_T2w_compression_labels.nii.gz'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    mandatoryArguments.add_argument(
+        '-i-PAM50',
+        metavar=Metavar.file,
+        required=True,
+        help='Input csv file name in PAM50 space. Example: csa_PAM50.csv.'
+        'Use sct_process_segmentation with -normalize-PAM50 to generate the file.'
```
jcohenadad, line 62:
```suggestion
        help='CSV morphometric file in the PAM50 space, obtained by running: '
        'sct_process_segmentation -normalize-PAM50.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    mandatoryArguments.add_argument(
+        '-metric',
+        required=False,
+        help='Metric name to normalize in .csv file output from sct_process_segmentation.'
```
jcohenadad, line 67:
```suggestion
        help='Metric to normalize in .csv file output from sct_process_segmentation.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+        '\nChoices: MEAN(area), MEAN(diameter_RL), MEAN(eccentricity), MEAN(solidity)\n',
```
jcohenadad:
```suggestion
        help='Metric to normalize. '
        'Choices: MEAN(area), MEAN(diameter_AP), MEAN(diameter_RL), MEAN(eccentricity), MEAN(solidity)\n',
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+        default='MEAN(diameter_AP)',
+        choices=['MEAN(diameter_AP)', 'MEAN(area)', 'MEAN(diameter_RL)', 'MEAN(eccentricity)', 'MEAN(solidity)'],
+        metavar=Metavar.file,
+    )
+    optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        '-file-participants',
+        metavar=Metavar.file,
+        default='participants.tsv',
+        help='participants.tsv file of helathy controls.'
```
jcohenadad:
```suggestion
        help='participants.tsv file of healthy controls.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-sex',
+        type=str,
+        choices=['F', 'M'],
+        help='Select sex of healthy subjet to use for normalization. Requires participant.tsv'
```
jcohenadad, line 84:
```suggestion
        help='Sex of healthy subject to use for the normalization. Requires the flag "-file-participants"'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-age',
+        type=int,
+        nargs=2,
+        metavar="[0 100]",
+        help='Select age range of healthy subjet to use for normalization. Requires participant.tsv'
+        'E.g.:-age 60 80'
```
jcohenadad, line 92:
```suggestion
        help='Age range of healthy subject to use for the normalization. Requires the flag "-file-participants".'
        'Example: "-age 60 80". By default, all ages are considered.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-o',
+        metavar=Metavar.file,
+        help="Output csv file name. If not provided, _compression_metrics suffix is added to the file name provided by -i flag."
```
jcohenadad, line 97:
```suggestion
        help='Output CSV file name. If not provided, the suffix "_compression_metrics" is added to the file name provided by the flag "-i"'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,490 @@
...
+    )
+    optional.add_argument(
+        '-subject',
+        metavar=Metavar.file,
+        help="Name of subject. Default: filename of -i"
```
jcohenadad, line 102: I don't understand this flag. How this subject will be used? for normalization? only with one subject?

sandrinebedard: It is just the subjects name to include in the output csv file. If it is not specified the filename of `-i` argument is used

spinalcordtoolbox/scripts/sct_compute_compression.py
jcohenadad: I also suggest you add documentation about how one can interpret a result like this:

|   Subject             |   Compression Level  |   MEAN(diameter_AP) ratio  |   Normalized MEAN(diameter_AP) ratio  |
|-----------------------|----------------------|----------------------------|---------------------------------------|
|   sub-twh018_T2w.csv  |   4                  |   2.4655002897446400       |   10.375583376646100                  |
|   sub-twh018_T2w.csv  |   5                  |   7.177562376846180        |   11.525718035912400                  |
|   sub-twh018_T2w.csv  |   7                  |   4.854757889876880        |   1.8346037155670000                  |

Ie: explain what "MEAN(diameter_AP) ratio" is (eg: 1 means no compression would, 2 means that ...). 

Also, it would be nice to include the equation of the ratio, and the normalization in the usage.

spinalcordtoolbox/scripts/sct_compute_compression.py
sandrinebedard: it is not related to the output file, but the column `subject` in the output .csv file like you see in : https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003#discussion_r1114902193.

Maybe it is not that useful either and I can modify the column with filename instead

jcohenadad: ahhh! ok gotcha! So, I suggest to replace the column name 'Subject' by 'File', because this is exactly what it is. And I would suggest to not have the 'subject' feature for now.

joshuacwnewton:
I've added some small changes just to address the errors in the CI tests. 

I plan on doing a more thorough review of this PR shortly after. :slightly_smiling_face:

sandrinebedard:
The last changes are now pushed. However, we discussed with @valosekj and @jcohenadad about a limitation of averaging the metrics across the vertebral bodie above and below. The compression labels are usually located at interverterbral discs and can be classified with the level above or below depending on where the compression label is positionned. If it is associated with the level below, there is high chances that slices in the level above present compression and are considered as not compressed.

Here is the proposed solution:

Average on a certain extent 1cm above and 1 cm below the compression
image

For that, we need the spinal cord centerline to compute the distance along the centerline. I need to check if we also get it from `sct_process_segmentation`.

sandrinebedard:
A question remains, is it really necessary to compute the distance along the centerline or can we assume orthogonality.

@valosekj pointed out:

Do we really need to use the centerline? Since FOV is usually placed orthogonally to the SC, the effect of the curvature should not be significant. Then, we could rely just on a certain distance in PAM50 space (which is easy to compute).
I am also thinking; if we choose a distance, for example, 1cm, it is highly probable the distance will "hit" the slice center. And we will be deciding whether to include or discard such a slice.
image

@jcohenadad do you have an opinion on this?

jcohenadad:
> if we choose a distance, for example, 1cm, it is highly probable the distance will "hit" the slice center. And we will be deciding whether to include or discard such a slice.

I would use rounding, ie: if the distance from the center covers more than half of the slice, include the slice, otherwise not. Sketch below:

image

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,497 @@
...
+        compressed_levels_dict = get_slices_in_PAM50(compressed_levels_dict, df_metrics, df_metrics_PAM50)
+
+        # Get data from healthy control and average them
+        df_avg_HC = average_hc(path_ref, metric, list_HC)
+
+    # Loop through all compressed levels (compute one MSCC per compressed level)
+    for level in compressed_levels_dict.keys():
+        # Get metric of patient with compression
+        if arguments.i_PAM50:
+            ap, slices_avg = average_compression_PAM50(slice_thickness, slice_thickness_PAM50, metric, df_metrics_PAM50,
```
valosekj, line 471: Since we allowed to normalize also other metrics (`['diameter_AP', 'area', 'diameter_RL', 'eccentricity', 'solidity']`) using the`-metric` input argument, we should probably rename the `ap` variable. The `ap` referred to  diameter_AP. But now, it can contain any of the listed metrics.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,497 @@
...
+        default='diameter_AP',
+        choices=['diameter_AP', 'area', 'diameter_RL', 'eccentricity', 'solidity'],
+        metavar=Metavar.file,
+    )
+    optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        '-i-PAM50',
+        metavar=Metavar.file,
+        help='CSV morphometric file in the PAM50 space, obtained by running: '
+        'sct_process_segmentation -normalize-PAM50.'
```
valosekj, line 74: Maybe we could mention that it is necessary to provide this file if one wants to do normalization using the database of healthy control. 

```suggestion
        help='CSV morphometric file in the PAM50 space, obtained by running: '
        'sct_process_segmentation -normalize-PAM50. Use this flag if you want to normalize the metrics using a database of healthy controls.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,497 @@
...
+    )
+    optional.add_argument(
+        '-file-participants',
+        metavar=Metavar.file,
+        default='participants.tsv',
+        help='participants.tsv file of healthy controls.'
```
valosekj, line 80:
```suggestion
        help='participants.tsv file of healthy controls. Use this flag in combination with the "-sex" and "age" flags.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,497 @@
...
+# ==========================================================================================
+def get_parser():
+    parser = SCTArgumentParser(
+        description='Compute spinal cord compression morphometrics such as the Maximum Spinal Cord Compression [1], '
+                    'the antero-posterior (AP) diameter, or other relevant spinal cord morphometrics that are output'
+                    ' by the function sct_process_segmentation (CSA, RL diameter, eccentricity, solidity, etc.).'
+                    ' Metrics are normalized using a database of spinal cord morphometrics built from healthy control'
+                    ' subjects. This database uses the PAM50 template as an anatomical reference system.'
+                    ' \nEquation:    ratio = (1 - mi/((ma+mb)/2))\n'
+                    'mi: metric at compression level. ma: metric of level above compression level. mb: metric of level below compression level.'
```
valosekj, line 40:
```suggestion
                    'Metrics are normalized using the non-compressed levels above and below the compression site:'
                    ' \nEquation:    ratio = (1 - mi/((ma+mb)/2))\n'
                    'mi: metric at compression level. ma: metric of level above compression level. mb: metric of level below compression level.\n'
                    'Additionally, if the "-i-PAM50" flag is used, metrics are normalized using a database of spinal cord morphometrics built from healthy control'
                    ' subjects. This database uses the PAM50 template as an anatomical reference system.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,497 @@
+#!/usr/bin/env python
+#########################################################################################
+#
+# Compute maximum spinal cord compression using AP diameter or other morphometrics.
+#
+# ---------------------------------------------------------------------------------------
+# Copyright (c) 2015 Polytechnique Montreal <www.neuro.polymtl.ca>
```
valosekj, line 7: I guess we can update or remove `2015`

spinalcordtoolbox/scripts/sct_compute_compression.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/634e2f3a7c105d4c62931df3845878bde9c525c2

sandrinebedard:
I modified the function to average across a specific extent at a distance from the compression. If we are using normalization (`-i-PAM50`), the distance is taken in the PAM50 space. If not, the distance is measured along the spinal cord centerline, the spinal cord ssegmentation is now necessary. However, it may be better to also output it from `sct_process_segmentation` directly.

@joshuacwnewton everything should be up to date for you to check 🙂

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,644 @@
...
+                    'with 100 consecutive patients. Radiology 2007;243(3):820-827.'
+    )
+
+    mandatoryArguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
+    mandatoryArguments.add_argument(
+        '-i',
+        metavar=Metavar.file,
+        required=True,
+        help='NIfTI file of spinal cord segmentation to compute morphometrics. If no normalization is wanted (if -i-PAM50 is not specified),'
+             ' metric ratio will take average along the centerline.'
```
valosekj, line 64: Maybe, since there is now no input .csv file, the help could be simplified:

```suggestion
        help='Spinal cord segmentation mask to compute morphometrics from. Example: t2_seg.nii.gz'
         'Note: If no normalization is wanted (i.e., if the "-i-PAM50" flag is not specified),'
             ' metric ratio will take average along the segmentation centerline.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,644 @@
...
+    )
+    mandatoryArguments.add_argument(
+        '-vertfile',
+        metavar=Metavar.file,
+        required=True,
+        help='Vertebral labeling file.'
+             'The input and the vertebral labelling file must be in the same voxel coordinate system'
+             'and must match the dimensions between each other.'
```
valosekj, line 72:
```suggestion
        help='Vertebral labeling file. Example: sub-001_T2w_seg_labeled.nii.gz'
             'Note: The input and the vertebral labelling file must be in the same voxel coordinate system'
             'and must match the dimensions between each other.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,644 @@
...
+    )
+    mandatoryArguments.add_argument(
+        '-l',
+        metavar=Metavar.file,
+        required=True,
+        help='NIfTI file that includes labels at the compression sites. Each compression site is denoted by a single voxel of value `1`. '
+             'The input and the compression label file must be in the same voxel coordinate system'
+             'and must match the dimensions between each other.'
+             'Example: sub-001_T2w_compression_labels.nii.gz'
```
valosekj, line 81: Just reorganize to match the previous arguments.

```suggestion
        help='NIfTI file that includes labels at the compression sites. Each compression site is denoted by a single voxel of value `1`.  Example: sub-001_T2w_compression_labels.nii.gz'
             'Note: The input and the compression label file must be in the same voxel coordinate system'
             'and must match the dimensions between each other.'
```

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+        # Get centerline object
+        centerline = get_centerline_object(img_seg, verbose=verbose)
+        # Get healthy slices to average for level above and below
+        z_range_above, z_range_below = get_slices_upper_lower_level_from_centerline(centerline, distance, extent, slice_compressed, z_ref)
+
+    # Step 4. Compute MSCC metrics for each compressed level
+    # ------------------------------------------------------
+    # Loop through all compressed levels (compute one MSCC per compressed level)
+    for level in compressed_levels_dict.keys():
+        # Get metric of patient with compression
```
valosekj, line 641: Maybe, the `printv` from line 660 could be moved here to have `Level:` followed by `metric_a = ...` when using `-v 2`:

image


```suggestion
        printv('\nLevel: {}'.format(level), verbose=verbose, type='info')
        # Get metric of patient with compression
```

valosekj: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/2a587a74209da12d8dc638c022ea32474cc71293

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+                list_HC = select_HC(fname_partcipants, sex, age)
+        else:
+            list_HC = None
+        if age:
+            age.sort()
+            if any(n < 0 for n in age):
+                parser.error('Age range needs to be positive, {} was specified'.format(age))
+            # Put age range in order
+            else:
+                age.sort()
```
joshuacwnewton, line 609: I think this `age.sort()` call is duplicated.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+            # Get metrics of healthy controls
+            metrics_HC = average_metric(df_avg_HC, metric, z_range_above, z_range_below, slices_avg)
+            logger.debug('\nmetric_a_HC =  {}, metric_b_HC = {}, metric_i_HC = {}'.format(metrics_HC[0], metrics_HC[1], metrics_HC[2]))
+            # Compute Normalized Ratio
+            metric_ratio_norm_result = metric_ratio_norm(metrics_patient, metrics_HC)
+        else:
+            slices_avg = compressed_levels_dict[level]
+            metrics_patient = average_metric(df_metrics, metric, z_range_above, z_range_below, slices_avg)
+            metric_ratio_norm_result = None
+        logger.debug('\nmetric_a =  {}, metric_b = {}, metric_i = {}'.format(metrics_patient[0], metrics_patient[1], metrics_patient[2]))
```
joshuacwnewton, line 655: I think this section would be a tiny bit easier to read if this `logger.debug` statement was moved to the same place as the `# Display results` section below. (Just to keep the three print-related statements together.)

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton, line 1: As a script-wide comment, I notice that in many places. `"".format()` is used. 

I think it would be a bit cleaner if we removed all instances of `"".format()` from this script and replaced them with f-strings (`f""`).

joshuacwnewton (changes requested):
My first review is mainly just nitpicks and easy fixes.

I have a follow-up review that I'll add afterwards that will add comments related to script organization.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+             'Each compression site is denoted by a single voxel of value `1`.'
+             ' Example: sub-001_T2w_compression_labels.nii.gz '
+             'Note: The input and the compression label file must be in the same voxel coordinate system '
+             'and must match the dimensions between each other.'
+    )
+    mandatoryArguments.add_argument(
+        '-metric',
+        required=False,
+        help='Metric to normalize. '
+             'Choices: area, diameter_AP, diameter_RL, eccentricity, solidity\n',
```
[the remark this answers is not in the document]
sandrinebedard: hmm on my side it only displays the default: 
image
It should be there 2 times (since I wrote it) if it showed. Maybe because of the default.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+    else:
+        path, file_name, ext = extract_fname(get_absolute_path(arguments.i))
+        fname_out = os.path.join(path, file_name + '_compression_metrics' + '.csv')
+
+    # Step 1. Load subject input files and get csv files (label image, segmentation)
+    # -----------------------------------------------------------
+    img_seg = Image(fname_segmentation).change_orientation('RPI')
+    img_labels = Image(fname_labels).change_orientation('RPI')
+    # Check if compression labels are in the same dimensions than spinal cord segmentation
+    if check_if_shape_mismatch(img_labels, img_seg):
```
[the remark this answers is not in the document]
sandrinebedard: changed in:  https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/36f7bd09890079032fe5f69f4a6f094d62e05d31 , https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/4d7780a429605bb25d5d0202bfe0d165128af764 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/3b5aa87493d66ae34cf8519e1c7bb308a0cd7df1

spinalcordtoolbox/scripts/sct_compute_compression.py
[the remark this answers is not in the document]
joshuacwnewton: Oh! Can you try removing the `metavar.Str` part? I think the presence of that line causes `<str>` to appear instead of the choices.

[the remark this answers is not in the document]
sandrinebedard: it works: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/e029d754f54df2a6f352d9538d63e33b0221b856

spinalcordtoolbox/scripts/sct_compute_compression.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/1fd5f62373fe0a6d67c85b88ab639f69caf2d604

spinalcordtoolbox/scripts/sct_compute_compression.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/40f92891ffc868f2c76106dced4d6a49bf83714a

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,669 @@
...
+                    'Additionally, if the "-normalize" flag is used, metrics are normalized using a database built '
+                    'from healthy control subjects. This database uses the PAM50 template as an anatomical reference '
+                    'system.\n'
+                    '\n'
+                    'Reference: Miyanji F, Furlan JC, Aarabi B, Arnold PM, Fehlings MG. Acute cervical traumatic '
+                    'spinal cord injury: MR imaging findings correlated with neurologic outcome--prospective study '
+                    'with 100 consecutive patients. Radiology 2007;243(3):820-827.'
+    )
+
+    mandatoryArguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
```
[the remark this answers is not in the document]
sandrinebedard: done!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,653 @@
...
+    # Step 1. Load subject input files and get csv files (label image, segmentation)
+    # -----------------------------------------------------------
+    img_seg = Image(fname_segmentation).change_orientation('RPI')
+    img_labels = Image(fname_labels).change_orientation('RPI')
+    img_vertfile = Image(fname_vertfile).change_orientation('RPI')
+    # Check if segmentation, compression labels, and vertebral label files have same dimensions
+    if not img_seg.data.shape == img_labels.data.shape == img_vertfile.data.shape:
+        raise ValueError(f"Shape mismatch between compression labels [{img_labels.data.shape}], vertebral labels [{img_vertfile.data.shape}]"
+                         f" and segmentation [{img_seg.data.shape}]). "
+                         f"Please verify that your compression labels and vertebral labels were done in the same space as your input segmentation.")
```
joshuacwnewton, line 557: I think it might be good to move this whole section (input loading and shape validation) into the "Step 0: Argument validation section"

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,653 @@
...
+        sex = arguments.sex
+        age = arguments.age
+
+        # Step 2. Load reference input files (PAM50, healthy controls)
+        # ------------------------------------------------------------
+        # Select healthy controls based on sex and/or age range
+        path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics')
+        if sex or age:
+            if not os.path.isfile(get_absolute_path(os.path.join(path_ref, arguments.file_participants))):
+                raise FileNotFoundError('participants.tsv file must exists to select sex or age.')
```
joshuacwnewton, line 584: I think this input check should also be moved to the "argument validation section"

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,653 @@
...
+            else:
+                fname_partcipants = get_absolute_path(os.path.join(path_ref, arguments.file_participants))
+                list_HC = select_HC(fname_partcipants, sex, age)
+        else:
+            list_HC = None
+        if age:
+            age.sort()
+            if any(n < 0 for n in age):
+                parser.error(f'Age range needs to be positive, {age} was specified')
```
joshuacwnewton, line 593: This whole section too, I think, can be moved to step 0.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,653 @@
...
+    slice_compressed = get_compressed_slice(img_labels, verbose)
+    compressed_levels_dict = get_verterbral_level_from_slice(slice_compressed, df_metrics)
+    # Initialize variables if normalization
+    if arguments.normalize:
+        fname_metrics_PAM50 = os.path.join(path, file_name + '_metrics_PAM50' + '.csv')
+        # Call sct_process_segmentation to get morphometrics perslice in PAM50 space
+        sct_process_segmentation.main(argv=['-i', fname_segmentation, '-vertfile', fname_vertfile, '-normalize-PAM50', '1',
+                                      '-perslice', '1', '-o', fname_metrics_PAM50])
+        sex = arguments.sex
+        age = arguments.age
```
joshuacwnewton, line 576: I think these assignments should be moved to Step 0, too.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,653 @@
...
+        csv_writer.writerow(line)
+
+
+def main(argv: Sequence[str]):
+    parser = get_parser()
+    arguments = parser.parse_args(argv)
+    verbose = arguments.v
+    set_loglevel(verbose=verbose)    # values [0, 1, 2] map to logging levels [WARNING, INFO, DEBUG]
+
+    # Step 0: Argument validation
```
joshuacwnewton, line 534: Just to make things a tiny bit clearer, I think we can rename this :

```suggestion
    # Step 0: Argument loading and validation
```

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton, line 1: My old refactoring PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4066 suggested to break the script down into Steps 0-4. 

However, in the time since:

- We now compute metrics within the script, rather than loading them from file.
- The script now contains large conditionals based on whether or not  `arguments.normalize` was specified. This makes things a bit awkward for the "Step 2" and "Step 3" comments, as they are in the middle of a big conditional, 
- There is now a bunch of code not covered by any "Step".

Because of these changes, I think we need to rework the "Step" comments so that they reflect the current state of the script:

- Step 0: Argument loading and validation
- Step 1: Get subject metrics and compressed slices
- Step 2: Get normalization metrics and slices (using PAM50 and reference dataset)
    - (To be added right before the `if arguments.normalize:` branch)
- Step 2: Get normalization metrics and slices (using non-compressed subject slices)
    - (To be added right before the `else:` branch)
- Step 3: Compute MSCC metrics for each compressed level

In other words, we rename Step 1, we rework Step 2/3 to account for the big `if arguments.normalize` conditional, and we rename Step 4 to be the new Step 3. 

(And then apply these same changes to the function comments, too.)

joshuacwnewton (changes requested):
This review is focuses on script organization:

- Reworking the "Step 0-4" comments so that they match the updated state of the script.
- Moving `arguments`-related code to Step 0:
   - Any code that loads or transforms `arguments` (sorting, changing to string, etc.) I think would be good to group together at the start of the script.
   - Any code that validates `arguments` and possibly throws an error I think would be good to group together at the start of the script, to quit early and avoid unnecessary processing (since we now include expensive `sct_process_segmentation` calls.)

spinalcordtoolbox/scripts/sct_compute_compression.py
sandrinebedard: All addressed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/26d6afc48d52980beb1bbad711d9306eb9077f8a and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/4d6d247e500d3921ad89eb44f8cdc4658026be8a

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,651 @@
...
+    # Select healthy controls based on sex and/or age range
+    path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics')
+    if sex or age:
+        if not os.path.isfile(get_absolute_path(os.path.join(path_ref, arguments.file_participants))):
+            raise FileNotFoundError('participants.tsv file must exists to select sex or age.')
+        else:
+            fname_partcipants = get_absolute_path(os.path.join(path_ref, arguments.file_participants))
+            list_HC = select_HC(fname_partcipants, sex, age)
+    else:
+        list_HC = None
```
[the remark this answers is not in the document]
joshuacwnewton: (This would also mean that `select_HC` is a "Step 2" function, instead of a "Step 0" function.)

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,651 @@
...
+    # Check if segmentation, compression labels, and vertebral label files have same dimensions
+    if not img_seg.data.shape == img_labels.data.shape == img_vertfile.data.shape:
+        raise ValueError(f"Shape mismatch between compression labels [{img_labels.data.shape}], vertebral labels [{img_vertfile.data.shape}]"
+                         f" and segmentation [{img_seg.data.shape}]). "
+                         f"Please verify that your compression labels and vertebral labels were done in the same space as your input segmentation.")
+    path, file_name, ext = extract_fname(get_absolute_path(arguments.i))
+    fname_metrics = os.path.join(path, file_name + '_metrics' + '.csv')
+    # Call sct_process_segmentation to get morphometrics perslice in native space
+    sct_process_segmentation.main(argv=['-i', fname_segmentation, '-vertfile', fname_vertfile, '-perslice', '1', '-o', fname_metrics])
+    metric = 'MEAN(' + arguments.metric + ')'  # Adjust for csv file columns name
```
joshuacwnewton, line 561: Could this line be moved up to where all of the other variables (`age`, `sex`, etc.) are assigned?

joshuacwnewton (changes requested):
Thank you for the prompt reorganization! :hearts: 

I just had some small additions to my last review -- just to make sure that all of the related code is grouped together.

(Sorry for the nitpicks! After this review I'll get to the bigger/more important feedback.)

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,651 @@
...
+    img_vertfile = Image(fname_vertfile).change_orientation('RPI')
+    # Check if segmentation, compression labels, and vertebral label files have same dimensions
+    if not img_seg.data.shape == img_labels.data.shape == img_vertfile.data.shape:
+        raise ValueError(f"Shape mismatch between compression labels [{img_labels.data.shape}], vertebral labels [{img_vertfile.data.shape}]"
+                         f" and segmentation [{img_seg.data.shape}]). "
+                         f"Please verify that your compression labels and vertebral labels were done in the same space as your input segmentation.")
+    path, file_name, ext = extract_fname(get_absolute_path(arguments.i))
+    fname_metrics = os.path.join(path, file_name + '_metrics' + '.csv')
+    # Call sct_process_segmentation to get morphometrics perslice in native space
+    sct_process_segmentation.main(argv=['-i', fname_segmentation, '-vertfile', fname_vertfile, '-perslice', '1', '-o', fname_metrics])
```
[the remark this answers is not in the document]
joshuacwnewton: ((Also, since defining the csv file and calling `sct_process_segmentation` are tightly related, it might be good to move the comment up, so that we group all the related lines of code.))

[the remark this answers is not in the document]
sandrinebedard: everything should be good now!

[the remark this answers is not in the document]
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/dc83afcf4b467f2d3732f43f951d1670e0ca8b7d

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+             ' Example: sub-001_T2w_compression_labels.nii.gz '
+             'Note: The input and the compression label file must be in the same voxel coordinate system '
+             'and must match the dimensions between each other.'
+    )
+    mandatory.add_argument(
+        '-metric',
+        required=False,
+        help='Metric to normalize. ',
+        default='diameter_AP',
+        choices=['diameter_AP', 'area', 'diameter_RL', 'eccentricity', 'solidity'],
```
joshuacwnewton, line 90: I notice that in the `compute_shape()` function in `process_seg`, there is a similar list of metrics:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a029f36f7bae57cdf238ba13bc416f5edbc8dcbc/spinalcordtoolbox/process_seg.py#L33-L43

And also in the `save_csv()` function in `aggregate_slicewise.py`: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a029f36f7bae57cdf238ba13bc416f5edbc8dcbc/spinalcordtoolbox/aggregate_slicewise.py#L540-L546

Some naive questions here:

- Why are some metrics included, but others excluded?
- Is there any way to reuse a single list of metrics? e.g. by making it a module-level constant in `process_seg.py`, then importing it here? (Just to make sure the available metrics don't become out of sync.)

joshuacwnewton (commented):
This review is mainly focused on `argparse` (the arguments and their descriptions). I had some thoughts/ideas, some of which might be good to discuss at today's meeting.

spinalcordtoolbox/download.py
```diff
@@ -159,7 +159,13 @@
             "https://github.com/ivadomed/multimodal-registration/releases/download/r20220512/models.zip"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "deepreg_models"),
-    }
+    },
```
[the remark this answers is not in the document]
valosekj: Yeah, thanks!
I cloned the repo manually; thus it contained `-` instead of `_`. But the correct way should have been to use `sct_download_data`.

I reverted the https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/37447dd3dd34234ab29adea2ba4be4bfb7df9bc6 commit and added a check if `PAM50_normalized_metrics` exists in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/274107a8df2b9c02636dded7af4a5d42b6f55b17.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,654 @@
...
+
+# Functions for Step 1 (Load subject input files and get csv files)
+# ==========================================================================================
+def get_slice_thickness(img):
+    """
+    Get slice thickness from the input image.
+    :param img: Image: source image
+    :return float: slice thickness in mm.
+    """
+    return img.dim[5]
```
valosekj, line 166: Should this be [6] instead [5]?
```suggestion
    return img.dim[6]
```

Because, for instance, for axial image with slice thickness 3.6, `img.dim` is `(320, 320, 24, 1, 0.5, 0.5, 3.6000001, 1)`

sandrinebedard: you are absolutly right! Thanks!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,631 @@
...
+    level_compression = df_metrics.loc[idx, ['VertLevel', 'Slice (I->S)']]
+    if level_compression.empty:
+        raise ValueError(f"Slice {slices} doesn't have a computed metric")
+    # Check if level_compression['VertLevel'] is nan
+    for _, row in level_compression.iterrows():
+        if np.isnan(row['VertLevel']):
+            raise ValueError(f"Slice {int(row['Slice (I->S)'])} doesn't have computed vertebral level. "
+                             f"Check vertebral labeling file.")
+    level_slice_dict = {}
+    for level in np.unique(level_compression['VertLevel']):
```
sandrinebedard, line 202: This will ignore other slices in the same level that could refer to another compression. 

E.g.: If a compression is at C5/C6 and a second at C6/C7, bot compressions labels could be inside C6 at its extremities.

I need to adjust this. Also, we need to consider is multiple labels refer to the same compressions (the slices will be consecutive).

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+    )
+    optional = parser.add_argument_group("\nOPTIONAL ARGUMENTS")
+    optional.add_argument(
+        '-normalize',
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=1,  # Do we want 1 as default?
+        help='Set to 1 to normalize the metrics using a database of healthy controls. '
+    )
```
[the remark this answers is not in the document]
valosekj: Good points! 

You are right; the normalization with non-compressed levels above and below is always performed.

We renamed the `-normalize` arg to `-normalize-hc` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/02eafd02f8d4adc29b8074052d15c093bece5f22 to make clear that this arg is related to normalization to the database of HC. We also made the arg mandatory to force people to select the option.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+                    'below the compression level.\n'
+                    '\n'
+                    'Additionally, if the "-normalize" flag is used, metrics are normalized using a database built '
+                    'from healthy control subjects. This database uses the PAM50 template as an anatomical reference '
+                    'system.\n'
+                    '\n'
+                    'Reference: Miyanji F, Furlan JC, Aarabi B, Arnold PM, Fehlings MG. Acute cervical traumatic '
+                    'spinal cord injury: MR imaging findings correlated with neurologic outcome--prospective study '
+                    'with 100 consecutive patients. Radiology 2007;243(3):820-827.'
+    )
```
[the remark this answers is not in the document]
valosekj: After https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/02eafd02f8d4adc29b8074052d15c093bece5f22 commit (see also [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003#discussion_r1218126165)), we think that description is clear.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+        ' By default, both sexes are used.'
+    )
+    optional.add_argument(
+        '-age',
+        type=int,
+        nargs=2,
+        metavar="[0 100]",
+        help='Age range of healthy subjects to use for the normalization. Requires the flag "-file-participants". '
+        'Example: "-age 60 80". By default, all ages are considered.'
+    )
```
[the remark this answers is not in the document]
valosekj: Clarified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/04b776ba6216bc34f0075c454d84a96415504df1.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+             ' metric ratio will take the average along the segmentation centerline.'
+    )
+    mandatory.add_argument(
+        '-vertfile',
+        metavar=Metavar.file,
+        required=True,
+        help='Vertebral labeling file. Example: sub-001_T2w_seg_labeled.nii.gz'
+             'Note: The input and the vertebral labelling file must be in the same voxel coordinate system'
+             'and must match the dimensions between each other.'
+    )
```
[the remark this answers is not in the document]
valosekj: We do not say this in other functions --> we think it is not necessary to say it here.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+    mandatory.add_argument(
+        '-l',
+        metavar=Metavar.file,
+        required=True,
+        help='NIfTI file that includes labels at the compression sites. '
+             'Each compression site is denoted by a single voxel of value `1`.'
+             ' Example: sub-001_T2w_compression_labels.nii.gz '
+             'Note: The input and the compression label file must be in the same voxel coordinate system '
+             'and must match the dimensions between each other.'
+    )
```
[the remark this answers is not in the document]
valosekj: We think it is not necessary to say it here. It will be covered in the tutorial.

spinalcordtoolbox/scripts/sct_compute_compression.py
valosekj: Good point! 
The thing is that it makes sense to do the normalization only for some metrics (`choices=['diameter_AP', 'area', 'diameter_RL', 'eccentricity', 'solidity']`).

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,631 @@
...
+    :param level: int: Level of compression.
+    :param metric: str: metric to perform normalization
+    :param metric_ratio: float:
+    :param metric_ratio_norm:
+    :param filename: str: input filename
+    :return:
+    """
+    if not os.path.isfile(fname_out):
+        with open(fname_out, 'w') as csvfile:
+            header = ['filename', 'Compression Level', metric + ' ratio', 'Normalized ' + metric + ' ratio']
```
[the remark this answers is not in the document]
valosekj: Already resolved.

spinalcordtoolbox/scripts/sct_compute_compression.py
valosekj: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/074c19e51222294ee40d47fe32e325fd705b017f

testing/cli/test_cli_sct_compute_compression.py
```diff
@@ -0,0 +1,155 @@
...
+    """ Run sct_compute_compression and check mscc and normalized mscc"""
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    sct_compute_compression.main(argv=['-i', dummy_3d_mask_nib, '-l', dummy_3d_compression_label,
+                                       '-vertfile', dummy_3d_vert_label, '-normalize-hc',  '1', '-o', filename])
+    with open(filename, "r") as csvfile:
+        reader = csv.DictReader(csvfile, delimiter=',')
+        row = next(reader)
+        assert float(row['compression_level']) == 5.0
+        assert float(row['diameter_AP_ratio']) == pytest.approx(12.525502319807725)
+        assert float(row['normalized_diameter_AP_ratio']) == pytest.approx(16.985020560800656)
```
joshuacwnewton, line 141: I notice that the `diameter_AP_ratio` is different between these two tests. This is because, in the `-normalize-hc 1` case, we use the metrics computed from `sct_process_segmentation -normalize-PAM50` instead:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/246d97b5e92621f703749763790841383a0acc8f/spinalcordtoolbox/scripts/sct_compute_compression.py#L627

Using the same name (`diameter_AP_ratio`) in both cases is a bit confusing, because `diameter_AP_ratio` implies that no normalization occurs. But, in fact, for `-normalize-hc 1`, `diameter_AP_ratio` is normalized to the PAM50 template, making it different that the same-name `diameter_AP_ratio` from `-normalize-hc 0`. This might create a trap for users, who don't realize that the two quantities are different.

I think it might be good to find a way to communicate this distinction to the user. For example:

- Have 3 keys instead of 2: `diameter_AP_ratio`, `diameter_AP_ratio_normalized_PAM50`, and `diameter_AP_ratio_normalized_HC`
  - `diameter_AP_ratio` is the computed same way regardless of whether `-normalize-hc` is 0 or 1.
  - The two `normalized` metrics are only computed for `-normalize-hc 1`.
- Then, add in the description for `-normalize-hc` that two forms of normalization are performed (PAM50 template normalization, PAM50 + HC database normalization).

joshuacwnewton: This would require a bit of refactoring, and I understand that @sandrinebedard and @valosekj have ISMRM right now and are going on vacation soon, so if this is too much to complete right now, I'm happy to take care of this. (Only if you agree with the suggestion, of course!)

joshuacwnewton (changes requested):
This mostly LGTM! 

I mainly have nitpicks/small changes, but there is one big change I want to suggest, too.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+    # Check if path_ref with normalized metrics exists
+    if not os.path.isdir(path_ref):
+        raise FileNotFoundError(f"Directory with normalized PAM50 metrics {path_ref} does not exist.\n"
+                                f"You can download it using 'sct_download_data -d PAM50_normalized_metrics'.")
+
+    # Print warning if sex or age are specified without normalized-hc
+    if sex and not arguments.normalize_hc:
+        logger.warning("The 'sex' flag requires the '-normalize-hc 1'.")
+    if age and not arguments.normalize_hc:
+        logger.warning("The 'age' flag requires the '-normalize-hc 1'.")
```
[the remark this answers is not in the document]
valosekj: Great suggestion, agree!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+    img_vertfile = Image(fname_vertfile).change_orientation('RPI')
+    if not img_seg.data.shape == img_labels.data.shape == img_vertfile.data.shape:
+        raise ValueError(f"Shape mismatch between compression labels [{img_labels.data.shape}], vertebral labels [{img_vertfile.data.shape}]"
+                         f" and segmentation [{img_seg.data.shape}]). "
+                         f"Please verify that your compression labels and vertebral labels were done in the same space as your input segmentation.")
+    path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics')
+    # Check if path_ref with normalized metrics exists
+    if not os.path.isdir(path_ref):
+        raise FileNotFoundError(f"Directory with normalized PAM50 metrics {path_ref} does not exist.\n"
+                                f"You can download it using 'sct_download_data -d PAM50_normalized_metrics'.")
```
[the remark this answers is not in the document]
valosekj: Great catch! You are right!

testing/cli/test_cli_sct_compute_compression.py
```diff
@@ -0,0 +1,155 @@
...
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+def test_sct_compute_compression_value_against_groundtruth():
+    """Run the CLI script and verify that computed mscc value is equivalent to known ground truth value."""
+    di, da, db = 6.85, 7.65, 7.02
+    # FIXME: The results of "sct_compute_mscc" are not actually verified. Instead, the "mscc" function is called,
+    #        and THOSE results are verified instead.
+    # This was copied as-is from the existing 'sct_testing' test, but should be fixed at a later date.
+    # sct_compute_compression.main(argv=['-di', str(di), '-da', str(da), '-db', str(db)])
```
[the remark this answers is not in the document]
valosekj: Agree!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -0,0 +1,650 @@
...
+    # TODO adjust for multiple slices for one compresssion (that can have multiple levels too)
+    #slices_combined = []
+    #for slice in slices:
+    #    slices_same_compression = [slice_1 for slice_1 in slices if np.abs((slice - slice_1)) == 1]
+    #    if slices_same_compression:
+    #        if slices_same_compression[0] < slices_same_compression[1]:
+    #            slices_same_compression.append(slice)
+    #            slices_combined.append(slices_same_compression)
+    #    else:
+    #        slices_combined.append(slice)
```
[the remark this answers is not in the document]
valosekj: Yes, I think so. We still assume that the compression covers only a single slice and only a single vertebral level.

testing/cli/test_cli_sct_compute_compression.py
valosekj: This is an excellent point!
When processing the `sci-zurich` dataset, I manually set the output filename using the `-o` flag to make it clear whether `-normalize-hc` was used. See [these lines](https://github.com/sct-pipeline/dcm-metric-normalization/blob/c44dd1f8549766df6e02d2ed2cbe36eef79c678f/scripts/process_data_dcm-zurich.sh#L195-L196). But definitely, using dedicated keys would make it cleaner.

I am just thinking about the keys' names. Maybe I would go with the following:
- `diameter_AP_ratio` - obtained using `-normalize-hc 0` -- metrics computed in the native space without normalization
- `diameter_AP_ratio_PAM50` - obtained using `-normalize-hc 1` -- metrics computed in the PAM50 space without normalization
- `diameter_AP_ratio_PAM50_normalized` - obtained using `-normalize-hc 1` -- metrics computed in the PAM50 space with normalization to HC database

But I would also wait for @sandrinebedard opinion.

joshuacwnewton: > We could always begin the refactoring with your proposed names? Then, if there is further discussion on what the names should be, it would be a simple update, since it is just a find and replace.

I'm going to try and tackle this later today, then push to this PR. If there are any issues/concerns, we can always revert the commits or modify them. But, still, I'm just happy to try to lessen the menial work on you right now, since this is kind of boring refactoring work, hehe. :)

joshuacwnewton: Done!

I've tried my best to break each change down into individual commits, that way every bit of refactoring is straightforward and easy to understand. You should be able to look through each commit, one at a time, to understand what I've done. :slightly_smiling_face:

valosekj: Thank you very much for this refactoring, @joshuacwnewton! I also really appreciate your detailed commit messages making the code review very convenient!

I tested your changes, and their work perfectly!

The command I used:
```
sct_compute_compression -i sub-260155_acq-axial_T2w_label-SC_mask.nii.gz  -vertfile sub-260155_acq-axial_T2w_label-SC_mask_labeled.nii.gz  -l sub-260155_acq-axial_T2w_label-compression-manual.nii.gz  -normalize-hc 1 -o sub-260155_acq-axial_T2w_diameter_AP_normalize-hc.csv
```

The output CSV file looks like:

image

I have just one minor comment here. The last column is now named `diameter_AP_ratio_normalized`. But maybe, we could call it `diameter_AP_ratio_PAM50_normalized` (as I mentioned in 
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003#discussion_r1234325392) to make explicitely clear that the normalization is done on metrics computed in PAM50 space.

joshuacwnewton: Oh, oops! That was definitely an accidental mistake on my part. Thank you for catching that!

joshuacwnewton: Re: @valosekj's https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003#discussion_r1234325392:

> But I would also wait for @sandrinebedard opinion.

Do you have any thoughts about the this refactor, @sandrinebedard? Anything you object to, or that should be changed?

If there are no objections, then I think this PR might be ready to merge. :slightly_smiling_face:

sandrinebedard: everything looks good to me!

valosekj (approved):

testing/cli/test_cli_sct_compute_compression.py
joshuacwnewton: Sorry for the very late reply to this older comment! I missed it until just now.

I think it's okay to test these values? 

The reason I think so is because,  every time we update the `data_normalize_metric_pam50` repo, we would need to create a new release anyway for the dataset, and then update the link in `sct_download_data`?

So, we would already be creating a PR in SCT, and I don't think it's too much extra work to update the values at the same time.

joshuacwnewton (approved):

joshuacwnewton:
Thank you so much @sandrinebedard and @valosekj for this massive effort! :hearts:

valosekj:
@joshuacwnewton thank you for your detailed review and help! 💯

PR #4010, opened 2023-01-24
Update `get_centerline` in `core.py`

### LINKED ISSUE #4006: Describe outputs of different Centerline extraction methods with more detail
# Feature justification
I think it would be easier for SCT developers to have a concise description of what objects are returned and files are written by the various Centerline extraction methods. For instance, the space and orientation in which objects & files are defined can be confusing and lead to downstream errors.

# Potential solution
Adding more detailed information about what objects are returned & files written in the comments section at the beginning each function. 

# Obtaining centerlines & understanding outputs

## Environment
* Python 3.8.15
* SCT: 5.8
* Shell: /bin/bash

## Data & Set up
_Note: unless otherwise specified, commands below are run from command-line._
	1. Download data [here](https://spinalcordtoolbox.com/user_section/tutorials/segmentation/before-starting.html)
	2. `cd single_subject/data/t2/`
	3. Segment spinal cord: `sct_deepseg_sc -i t2.nii.gz -c t2 -ofolder $PWD -qc $PWD/qc`

### A. `sct_get_centerline`
**Command:**
  ```
  sct_get_centerline -i t2.nii.gz -c t2
  ```
**Writes files:**
		* `t2_centerline.csv`: centerline coordinates defined in voxel ("pix") space and RPI orientation 
		* `t2_centerline.nii.gz`: centerline mask defined in voxel ("pix") space and input (e.g. _t2.nii.gz_) orientation (not RPI) 

### B. `spinalcordtoolbox.centerline.core.get_centerline`
**Command:**
  ```
  python
  from spinalcordtoolbox.centerline.core import get_centerline 
  from spinalcordtoolbox.centerline.core import ParamCenterline 
  from spinalcordtoolbox.image import * 
  from spinalcordtoolbox.types import Centerline
  param=ParamCenterline(algo_fitting = 'linear')
  im_centerline, arr_ctl, arr_ctl_der,fit_results = get_centerline(Image("t2_seg.nii.gz"),param=param)
  centerline = Centerline(arr_ctl[0],arr_ctl[1],arr_ctl[2],arr_ctl_der[0],arr_ctl_der[1],arr_ctl_der[2])
  ```
**Returns objects:**
	* `im_centerline`: centerline Image object defined in voxel ("pix") space and input orientation (e.g. _t2_seg.nii.gz_)
	* `arr_ctl`: centerline coordinates defined in voxel ("pix") space and RPI orientation
	* `centerline`: centerline object that stores points (centerline.points) defined in voxel ("pix") space and RPI orientation
	* `arr_ctl_der`
	* `fit_results`

### C. `spinalcordtoolbox.straightening._get_centerline`
**Command:**
```
python
from spinalcordtoolbox.straightening import _get_centerline 
from spinalcordtoolbox.centerline.core import get_centerline 
from spinalcordtoolbox.centerline.core import ParamCenterline 
from spinalcordtoolbox.types import Centerline
from spinalcordtoolbox.image import * 
param=ParamCenterline(algo_fitting = 'linear')
centerline = _get_centerline(Image("t2_seg.nii.gz"),param_centerline=param,verbose=1)
```
**Returns objects:**
	*`centerline`*: Centerline object that stores centerline derivatives (centerline.derivatives) and centerline coordinates (centerline.points) defined in physical ("phys") space and input (e.g. _t2_seg.nii.gz_) orientation

### D. `spinalcordtoolbox.types.save_centerline`
**Prerequisite:** generate Centerline object, either via `spinalcordtoolbox.straightening._get_centerline` or `spinalcordtoolbox.centerline.core.get_centerline`
**Command:**
```
python
centerline.save_centerline(fname_output='ctl')
centerline.save_centerline()
```
**Writes files:**
	*`ctl.npz`: File format that stores all of the Centerline object information (not just the coordinates)
	*`centerline.sct.npz`: default nomenclature if no `fname_output` given

NadiaBlostein:
Integrating `spinalcordtoolbox.straightening._get_centerline()` into `spinalcordtoolbox.core.get_centerline`, as discussed in [issue 4006](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4006).

Fixes #4006 (partially; there are still many instances that call `spinalcordtoolbox.straightening._get_centerline()` so it cannot be removed yet.

## Description
Added optional argument to `get_centerline` such that users can obtain Centerline coordinates in physical space and native orientation if they so chose. The method still defaults to outputting coordinates in pixel space and RPI orientation. It is therefore no longer necessary for users to call `straightening._get_centerline` and the latter can be replaced wherever it is being used.

#### Environment
* Python 3.8.15
* SCT: 5.8
* Shell: /bin/bash

#### 1. Data & Set up
	1. Download data [here](https://spinalcordtoolbox.com/user_section/tutorials/segmentation/before-starting.html)
	2. `cd single_subject/data/t2/`
	3. Segment spinal cord: `sct_deepseg_sc -i t2.nii.gz -c t2 -ofolder $PWD -qc $PWD/qc`
#### 2. Run in Python:
```
from spinalcordtoolbox.centerline.core import get_centerline
import numpy as np
from spinalcordtoolbox.centerline.core import ParamCenterline

def test_arrays(arr_ctl,deriv_ctl,settings='default',):
	print(f'\n{settings} settings, arr_ctl.shape: {arr_ctl.shape} & deriv_ctl.shape: {deriv_ctl.shape}')
	print(f'arr_ctl[0,0:5: {arr_ctl[0,0:5]} & deriv_ctl[0,0:5: {deriv_ctl[0,0:5]}')
	print(f'arr_ctl[1,0:5: {arr_ctl[1,0:5]} & deriv_ctl[1,0:5: {deriv_ctl[1,0:5]}')
	print(f'arr_ctl[2,0:5: {arr_ctl[2,0:5]} & deriv_ctl[2,0:5: {deriv_ctl[2,0:5]}')

params=ParamCenterline()
im_seg=Image('t2_seg.nii.gz')

im_ctl,arr_ctl,deriv_ctl,_ = get_centerline(im_seg,params)
test_arrays(arr_ctl,deriv_ctl)

im_ctl,arr_ctl,deriv_ctl,_ = get_centerline(im_seg,params,space='pix')
test_arrays(arr_ctl,deriv_ctl,settings='pix')

im_ctl,arr_ctl,deriv_ctl,_ = get_centerline(im_seg,params,space='phys')
test_arrays(arr_ctl,deriv_ctl,settings='phys')
```

### DIFF
--- a/spinalcordtoolbox/centerline/core.py
+++ b/spinalcordtoolbox/centerline/core.py
@@ -82,7 +82,7 @@ def find_and_sort_coord(img):
     return np.array(arr_sorted_avg)
 
 
-def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files=1):
+def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files=1,space='pix'):
     """
     Extract centerline from an image (using optic) or from a binary or weighted segmentation (using the center of mass).
 
@@ -90,6 +90,7 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
     :param param: ParamCenterline() class:
     :param verbose: int: verbose level
     :param remove_temp_files: int: Whether to remove temporary files. 0 = no, 1 = yes.
+    :param space: string: Defining space and orientation in which to output Centerline information. 'pix' = pixel space / RPI, 'phys' = physical space /native.
     :return: im_centerline: Image: Centerline in discrete coordinate (int)
     :return: arr_centerline: 3x1 array: Centerline in continuous coordinate (float) for each slice in RPI orientation.
     :return: arr_centerline_deriv: 3x1 array: Derivatives of x and y centerline wrt. z for each slice in RPI orient.
@@ -237,10 +238,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
         fname_ctr = (add_suffix(os.path.basename(im_seg.absolutepath), "_ctr") if im_seg.absolutepath
                      else "centerline.nii.gz")
         im_centerline.save(os.path.join(tmp_folder, fname_ctr), mutable=True)
-
+    
+    # Obtain centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) in voxel ("pix") space and RPI orientation
+    arr_ctl=np.array([x_centerline_fit, y_centerline_fit, z_ref])
+    arr_ctl_der=np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
+    
+    # If specified by user, adjust centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation
+    if  space!='pix':
+        arr_ctl=np.array([x_centerline_fit, y_centerline_fit, z_ref])
+        arr_ctl_der=np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
+        nx, ny, nz, nt, px, py, pz, pt = im_seg.change_orientation(native_orientation).dim
+        # Transform centerline to physical coordinate system
+        arr_ctl_phys = im_seg.transfo_pix2phys(
+            [[arr_ctl[0][i], arr_ctl[1][i], arr_ctl[2][i]] for i in range(len(arr_ctl[0]))])
+        x_centerline, y_centerline, z_centerline = arr_ctl_phys[:, 0], arr_ctl_phys[:, 1], arr_ctl_phys[:, 2]
+        # Adjust derivatives with pixel size
+        x_centerline_deriv, y_centerline_deriv, z_centerline_deriv = arr_ctl_der[0][:] * px, \
+        arr_ctl_der[1][:] * py, \
+        arr_ctl_der[2][:] * pz
+        arr_ctl=np.array([x_centerline, y_centerline, z_centerline])
+        arr_ctl_der=np.array([x_centerline_deriv, y_centerline_deriv, z_centerline_deriv])
+        
     return (im_centerline,
-            np.array([x_centerline_fit, y_centerline_fit, z_ref]),
-            np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)]),
+            arr_ctl,
+            arr_ctl_der,
             fit_results)
 
 


spinalcordtoolbox/centerline/core.py
```diff
@@ -82,14 +82,15 @@ def find_and_sort_coord(img):
     return np.array(arr_sorted_avg)
 
 
-def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files=1):
+def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files=1,space='pix'):
```
jcohenadad, line 85:
```suggestion
def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files=1, space='pix'):
```
PEP8 nitpick

spinalcordtoolbox/centerline/core.py
```diff
@@ -237,10 +238,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
                      else "centerline.nii.gz")
         im_centerline.save(os.path.join(tmp_folder, fname_ctr), mutable=True)
-
+    
+    # Obtain centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) in voxel ("pix") space and RPI orientation
+    arr_ctl=np.array([x_centerline_fit, y_centerline_fit, z_ref])
+    arr_ctl_der=np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
+    
+    # If specified by user, adjust centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation
+    if  space!='pix':
```
jcohenadad, line 247:
```suggestion
    if  space != 'pix':
```
PEP8 nitpick https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements

spinalcordtoolbox/centerline/core.py
```diff
@@ -237,10 +238,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+        arr_ctl=np.array([x_centerline_fit, y_centerline_fit, z_ref])
+        arr_ctl_der=np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
```
jcohenadad, line 249:
```suggestion
        arr_ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
```

jcohenadad (commented):
thank you for looking into this @NadiaBlostein ❤️ !
Sorry i don't have the time to test the code and make 'smart' comments, so i'm only going with PEP8 nitpicks 😅

joshuacwnewton:
To expand on what @jcohenadad has mentioned:

SCT has automated tests that check your code. In particular, we have a "Python Linting" check that analyzes your code for style issues. If the check has found issues, then you will see an :x:, along with a "Details" link:

image

If you click "Details", then it will provide feedback for how you can improve the appearance of your code:

```
spinalcordtoolbox/centerline/core.py:85:83: E231 missing whitespace after ','
spinalcordtoolbox/centerline/core.py:241:1: W293 blank line contains whitespace
spinalcordtoolbox/centerline/core.py:243:12: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:244:16: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:245:1: W293 blank line contains whitespace
spinalcordtoolbox/centerline/core.py:246:180: E501 line too long (183 > 179 characters)
spinalcordtoolbox/centerline/core.py:247:7: E271 multiple spaces after keyword
spinalcordtoolbox/centerline/core.py:247:14: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:248:16: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:249:20: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:257:9: E122 continuation line missing indentation or outdented
spinalcordtoolbox/centerline/core.py:258:9: E122 continuation line missing indentation or outdented
spinalcordtoolbox/centerline/core.py:259:16: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:260:20: E225 missing whitespace around operator
spinalcordtoolbox/centerline/core.py:261:1: W293 blank line contains whitespace
```

You can then google each of these codes to find out what needs to be fixed. (For example, googling "E231 python" gives you https://www.flake8rules.com/rules/E231.html. So, you just need to go to line 85 in the file, then add some whitespace, and etc.)

Could you please fix all of these style issues? Thank you! :hearts:

joshuacwnewton:
> Could you please fix all of these style issues (and not just the ones pointed out by Julien)? Thank you! :hearts:

Small reminder ping @NadiaBlostein. :)

spinalcordtoolbox/centerline/core.py
```diff
@@ -238,9 +239,29 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+        # Transform centerline to physical coordinate system
+        arr_ctl_phys = im_seg.transfo_pix2phys(
+            [[arr_ctl[0][i], arr_ctl[1][i], arr_ctl[2][i]] for i in range(len(arr_ctl[0]))])
+        x_centerline, y_centerline, z_centerline = arr_ctl_phys[:, 0], arr_ctl_phys[:, 1], arr_ctl_phys[:, 2]
+        # Adjust derivatives with pixel size
+        x_centerline_deriv, y_centerline_deriv, z_centerline_deriv = arr_ctl_der[0][:] * px, \
+            arr_ctl_der[1][:] * py, \
+            arr_ctl_der[2][:] * pz
+        arr_ctl = np.array([x_centerline, y_centerline, z_centerline])
+        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, z_centerline_deriv])
```
joshuacwnewton, line 260: In this block of code, we create and assign the variables `arr_ctl` and `arr_ctl_der` 3 times.

But, I think we only need to create these variables once, at the very end? For the other 2 times, we can just access the `x`, `y`, and `z` variables directly.

So, if we replace all the instances of `arr_ctl[0]`, `arr_ctl[1]`, and `arr_ctl[2]` with the `x`, `y`, and `z` variables, we can get something like this:

```python
    # If specified by user, adjust centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation
    if space != 'pix':
        # Transform centerline to physical coordinate system
        arr_ctl = im_seg.transfo_pix2phys([[x_centerline_fit[i], y_centerline_fit[i], z_ref[i]]
                                           for i in range(len(x_centerline_fit))]).T
        # Adjust derivatives with pixel size
        nx, ny, nz, nt, px, py, pz, pt = im_seg.change_orientation(native_orientation).dim
        arr_ctl_der = np.array([x_centerline_deriv * px, y_centerline_deriv * py, np.ones_like(z_ref) * pz])
    else:
        arr_ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
```

I think this produces identical output, but in fewer lines?

joshuacwnewton, line 260: In this block of code, we create and assign the variables `arr_ctl` and `arr_ctl_der` 3 separate times.

But, I think we only need to create these variables once, at the very end? For the other 2 times, we can just access the `x`, `y`, and `z` variables directly.

So, if we replace all the instances of `arr_ctl[0]`, `arr_ctl[1]`, and `arr_ctl[2]` with the `x`, `y`, and `z` variables, we can get something like this:

```python
    # If specified by user, adjust centerline coordinates (`Centerline.points`) and derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation
    if space != 'pix':
        # Transform centerline to physical coordinate system
        arr_ctl = im_seg.transfo_pix2phys([[x_centerline_fit[i], y_centerline_fit[i], z_ref[i]]
                                           for i in range(len(x_centerline_fit))]).T
        # Adjust derivatives with pixel size
        nx, ny, nz, nt, px, py, pz, pt = im_seg.change_orientation(native_orientation).dim
        arr_ctl_der = np.array([x_centerline_deriv * px, y_centerline_deriv * py, np.ones_like(z_ref) * pz])
    else:
        arr_ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
```

I think this produces identical output, but in fewer lines?

joshuacwnewton (commented):
Thank you so much for taking this on!! :blush:

And, thank you so much for promptly responding to the linting feedback! :tada:

spinalcordtoolbox/centerline/core.py
NadiaBlostein: Yup, I had just copy-pasted `straightening._get_centerline`. This is a great catch!!! Thank you!

spinalcordtoolbox/centerline/core.py
```diff
@@ -238,9 +239,21 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+        arr_ctl_der = np.array([x_centerline_deriv * px, y_centerline_deriv * py, np.ones_like(z_ref) * pz])
+    else:
+        arr_ctl = np.array([x_centerline_fit, y_centerline_fit, z_ref])
+        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
+
     return (im_centerline,
-            np.array([x_centerline_fit, y_centerline_fit, z_ref]),
-            np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)]),
+            arr_ctl,
+            arr_ctl_der,
```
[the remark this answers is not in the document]
joshuacwnewton: So, for example, in a call like this: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7cb30ddbaa88ded9c3ae6250c72832f5c40f4d64/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L429-L430

We would replace `arr_ctl` and `arr_ctl_der` with a single `centerline` object that has an `arr_ctl_der` property:

```python
 _, centerline, _ = get_centerline(im_seg, param=ParamCenterline(), verbose=1) 
 x_centerline_deriv, y_centerline_deriv, z_centerline_deriv = centerline.arr_ctl_der 
```

And for a call like this:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7cb30ddbaa88ded9c3ae6250c72832f5c40f4d64/spinalcordtoolbox/straightening.py#L152

We would replace it with:

```python
 _, centerline, _ = get_centerline(img_ctl, self.param_centerline, verbose, remove_temp_files) 
```

Thus serving both purposes with a single `get_centerline` function.

joshuacwnewton (commented):
This PR looks good to me as-is! :tada: 

But, I'm also interested in investigating what remaining work is necessary to completely replace `straightening._get_centerline` with `centerline/core.get_centerline`. 

(I'm thinking it might be good to tackle everything in a single PR, that way we can test all of the changes together in one branch.)

spinalcordtoolbox/centerline/core.py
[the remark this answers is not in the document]
NadiaBlostein: Yeah that would make a lot of sense. I am guessing that there is an automatic way to find every instance of
```
 _, arr_ctl, arr_ctl_der, _ = get_centerline(im_seg, param=ParamCenterline(), verbose=1)
```
and replace each one with
```
 _, centerline,_ = get_centerline(im_seg, param=ParamCenterline(), verbose=1)
arr_ctl = centerline.arr_ctl
arr_ctl_der = centerline.arr_ctl_der
```
Is this an error-prone process? The `centerline.core.get_centerline` function is quite ubiquitous (unlike `straightening._get_centerline()`, which I have only personally only encountered in the 2 places described in the issue [4006](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4006)).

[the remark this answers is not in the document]
joshuacwnewton: > I am guessing that there is an automatic way to find every instance of [old cpde] and replace each one with [new code]

My IDE of choice is [PyCharm](https://www.jetbrains.com/pycharm/), and technically they have a way to do Find+Replace over multiple lines using Regex patterns.

But, as you say, doing this automatically would likely be error-prone? (Especially since there are some cases where only 0/1 of `{arr_ctl,arr_ctl_der}` are used, rather than both.)

So my first instinct would be to do this replacement work manually by searching for the more simple `"= get_centerline("` term. This returns 19 instances for me:

<details>

image

</details>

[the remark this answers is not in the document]
joshuacwnewton: (But of course, we would first need to update the `Centerline` class to store the arrays, as well as update the `get_centerline` function to produce a `Centerline` object.)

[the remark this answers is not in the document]
NadiaBlostein: Hey! Sorry for the delay on this. Doesn't the Centerline class already store the arrays? (`centerline.points` and `centerline.derivatives`)? So I made a mistake above, and after updating `get_centerline` to produce a `Centerline` object, I will replace this:

` _, arr_ctl, arr_ctl_der, _ = get_centerline(im_seg, param=ParamCenterline(), verbose=1)`

with this:

```
 _, centerline,_ = get_centerline(im_seg, param=ParamCenterline(), verbose=1)
arr_ctl = centerline.points
arr_ctl_der = centerline.derivatives
```

If this seems correct to you, I will move forward with this (and double check the shapes of everything along the way to make sure not to miss any transposes)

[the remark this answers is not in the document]
joshuacwnewton: I have tried to implement the above plan in the commits below. :smile:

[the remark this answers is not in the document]
joshuacwnewton: You're absolutely right that `points` and `derivatives` are basically the same as `arr_ctl` and `arr_ctl_der`. Except, they're actually transposes of one another? 

So, in [1da5a3c](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010/commits/1da5a3c25718e427b7e6f181e49fa52db393c049), I added properties that essentially just call `.T` (transpose) on the `points` and `derivatives` arrays to produce `arr_ctl` and `arr_ctl_der`.

spinalcordtoolbox/flattening.py
```diff
@@ -35,7 +35,8 @@ def flatten_sagittal(im_anat, im_centerline, verbose):
     nx, ny, nz, nt, px, py, pz, pt = im_anat.dim
 
     # smooth centerline and return fitted coordinates in voxel space
-    _, arr_ctl, _, _ = get_centerline(im_centerline, param=ParamCenterline(), verbose=verbose)
+    _, centerline, _ = get_centerline(im_centerline, param=ParamCenterline(), verbose=verbose)
+    arr_ctl = centerline.arr_ctl
```
benjamindeleener, line 39: I don't think it is useful to keep arr_ctl

spinalcordtoolbox/labels.py
```diff
@@ -370,7 +370,8 @@ def continuous_vertebral_levels(img: Image) -> Image:
     #   a. extract centerline
     #   b. for each slice, extract corresponding level
     nx, ny, nz, nt, px, py, pz, pt = img.dim
-    _, arr_ctl, _, _ = get_centerline(img, param=ParamCenterline())
+    _, centerline, _ = get_centerline(img, param=ParamCenterline())
+    arr_ctl = centerline.arr_ctl
```
benjamindeleener, line 374: arr_ctl is not used anywhere else. It is possible to extract directly the x, y and z components?

spinalcordtoolbox/registration/labeling.py
```diff
@@ -65,7 +65,8 @@ def project_labels_on_spinalcord(fname_label, fname_seg, param_centerline, remov
     im_seg.change_orientation("RPI")
 
     # smooth centerline and return fitted coordinates in voxel space
-    _, arr_ctl, _, _ = get_centerline(im_seg, param_centerline, remove_temp_files=remove_temp_files)
+    _, centerline, _ = get_centerline(im_seg, param_centerline, remove_temp_files=remove_temp_files)
+    arr_ctl = centerline.arr_ctl
```
benjamindeleener, line 69: Same comment as before on arr_ctl

spinalcordtoolbox/types.py
```diff
@@ -274,6 +274,18 @@ def __init__(self, points_x=None, points_y=None, points_z=None, deriv_x=None, de
         if self.compute_init_distribution:
             self.compute_vertebral_distribution(disks_levels=self.disks_levels, label_reference=self.label_reference)
 
+    # NB: Prior to https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010, the get_centerline() function
+    #     would return two 3xN arrays: `arr_ctl` and `arr_ctl_der`. These arrays are almost identical to
+    #     `Centerline.points` and `Centerline.derivatives`, except transposed. So, to preserve backwards-compatibility,
+    #     we provide some properties that mimic the old behavior.
+    @property
+    def arr_ctl(self):
```
benjamindeleener, line 282: Not sure if the name is appropriate. If I am not wrong, it means "array control" and control is related to a specific technique for generating the centerline. I would replace "arr_ctl" with "points" or "positions" or "centers" or anything more appropriate. The same comment goes for "arr_ctl_der".

benjamindeleener: I am actually wrong, "ctl" probably means "centerline". However, it proves that it is confusing.

spinalcordtoolbox/flattening.py
joshuacwnewton: I agree!

My only reason for including the redundant variables in commit [784704c](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010/commits/784704c4a15f24a3f9cd54c2d49cca976cf88f69) was to make it very clear to reviewers that the changes preserved the existing functionality.

But, as you point out, there is definitely some cleanup that can be done.

spinalcordtoolbox/labels.py
joshuacwnewton: Definitely! 

This was done for the same reasoning as https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010#discussion_r1106271333, and there is definitely cleanup that can be done here.

spinalcordtoolbox/types.py
joshuacwnewton: I agree that the `ctl` variable name isn't obvious at first. 

The reason I used it, however, was to preserve the existing conventions. (`ctl` is present throughout SCT's codebase -- almost everywhere that `get_centerline` is called.)

image

mguaypaq (commented):
I'm going to mark this as "comment" rather than "approve" for now, since I'm raising a question about the overall architecture of the change. Ping me if you'd like to keep the current organization, in which case I'll review the rest of the details.

spinalcordtoolbox/centerline/core.py
```diff
@@ -237,9 +241,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+        arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
+
+    centerline = Centerline(
+        points_x=arr_ctl[0].tolist(),
+        points_y=arr_ctl[1].tolist(),
+        points_z=arr_ctl[2].tolist(),
+        deriv_x=arr_ctl_der[0].tolist(),
+        deriv_y=arr_ctl_der[1].tolist(),
+        deriv_z=arr_ctl_der[2].tolist()
+    )
```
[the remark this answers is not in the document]
joshuacwnewton: Ahhh, this is a very good point. I like this idea a lot!!

Essentially, we would make it so that wherever `straightening._get_centerline()` used to be called, we call `get_centerline(space='phys')` instead, then feed the resulting `arr_ctl` and `arr_ctl_der` to `Centerline` at the site of the caller, rather than within the function itself.

In other words:

* We keep the thrust of commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010/commits/784704c4a15f24a3f9cd54c2d49cca976cf88f69
* But we reverse all of the fiddly changes to https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010/commits/415db4d1a0407d3fdda00b7932b72970441aff99, since these calls were just fine working in pixel space without the `Centerline` objects.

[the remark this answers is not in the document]
joshuacwnewton: The only downside is that this would hinder efforts to encourage developers to use `Centerline` objects, since `get_centerline` wouldn't return them. So, the onus is on the caller to know that `Centerline` exists, and when to import it and use it.

But, maybe this is a good thing, given your observation that `Centerline` may not be compatible with pixel-space arrays.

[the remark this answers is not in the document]
joshuacwnewton: I've reverted the commits that modify `get_centerline`, preserving the existing function calls. Then, I redid the commit that removes `_get_centerline`.

You're absolutely correct -- this drastically reduces the scope of this PR, and only creates `Centerline` objects for `space="phys"` specifically, which avoids the concerns described in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010#discussion_r1107400160.

spinalcordtoolbox/centerline/core.py
```diff
@@ -82,21 +83,24 @@ def find_and_sort_coord(img):
...
     :param im_seg: Image(): Input segmentation or series of points along the centerline.
     :param param: ParamCenterline() class:
     :param verbose: int: verbose level
     :param remove_temp_files: int: Whether to remove temporary files. 0 = no, 1 = yes.
+    :param space: string: Defining space and orientation in which to output Centerline information. 'pix' = pixel space / RPI, 'phys' = physical space /native.
     :return: im_centerline: Image: Centerline in discrete coordinate (int)
-    :return: arr_centerline: 3x1 array: Centerline in continuous coordinate (float) for each slice in RPI orientation.
-    :return: arr_centerline_deriv: 3x1 array: Derivatives of x and y centerline wrt. z for each slice in RPI orient.
+    :return: centerline: Centerline: An object built from the continuous centerline coordinates that provides additional
+             centerline-specific properties beyond what NIfTI images can store (e.g. disks_levels, label_reference)
```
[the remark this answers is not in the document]
joshuacwnewton: I think that this comment deserves deeper investigation, and that we should open an issue for it.

As far as this PR is concerned, though, now that https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010#discussion_r1108902353 has been implemented, we no longer create `Centerline` objects for `space="pix"`, meaning we dodge the concerns you've raised in the short term.

spinalcordtoolbox/flattening.py
joshuacwnewton: This concern has been addressed, because the highlighted changes have been reverted in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010#discussion_r1108902353.

spinalcordtoolbox/labels.py
joshuacwnewton: This concern has been addressed, because the highlighted changes have been reverted in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4010#discussion_r1108902353.

spinalcordtoolbox/centerline/core.py
```diff
@@ -238,9 +241,22 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
         im_centerline.save(os.path.join(tmp_folder, fname_ctr), mutable=True)
 
+    # If 'phys' is specified, adjust centerline coordinates (`Centerline.points`) and
+    # derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation.
+    if space == 'phys':
+        # Transform centerline to physical coordinate system
+        arr_ctl = im_seg.transfo_pix2phys([[x_centerline_fit[i], y_centerline_fit[i], z_ref[i]]
+                                           for i in range(len(x_centerline_fit))]).T
+        # Adjust derivatives with pixel size
+        nx, ny, nz, nt, px, py, pz, pt = im_seg.change_orientation(native_orientation).dim
```
joshuacwnewton, line 251: Small suggestion: We can indicate unused variables via `_`.

```suggestion
        _, _, _, _, px, py, pz, _ = im_seg.change_orientation(native_orientation).dim
```

spinalcordtoolbox/csa_pmj.py
```diff
@@ -44,16 +45,18 @@ def get_slices_for_pmj_distance(segmentation, pmj, distance, extent, param_cente
...
     param_centerline.smooth = 50
     param_centerline.minmax = True
-    # Compute spinalcordtoolbox.types.Centerline class
-    ctl_seg_with_pmj = _get_centerline(im_seg_with_pmj, param_centerline, verbose=verbose)
+    # Compute spinalcordtoolbox.types.Centerline class from get_centerline wit physical coordinates
+    _, arr_ctl_phys, arr_ctl_der_phys, _ = get_centerline(im_seg_with_pmj, param_centerline,
+                                                          verbose=verbose, space="phys")
+    ctl_seg_with_pmj = Centerline(points_x=arr_ctl_phys[0].tolist(), deriv_x=arr_ctl_der_phys[0].tolist(),
+                                  points_y=arr_ctl_phys[1].tolist(), deriv_y=arr_ctl_der_phys[1].tolist(),
+                                  points_z=arr_ctl_phys[2].tolist(), deriv_z=arr_ctl_der_phys[2].tolist())
```
mguaypaq, line 58: Maybe this is too tricky/obscure, but I believe the following shorter syntax would actually work:
```suggestion
    ctl_seg_with_pmj = Centerline(*arr_ctl_phys, *arr_ctl_der_phys)
```

To unpack this a little bit:
- You can use an iterable to supply positional arguments to a python function. This mirrors the syntax for defining a function which accepts a variable number of arguments:
  ```py3
  def my_function(a, b, c, *args): pass
  ```
  For function definitions, you can only have one `*args`, and it must be the last positional argument (so, keyword-only arguments can still go after), and `args` will be an actual `list`. But for function calls, you can put `*some_iterable` anywhere in the list, or even more than once, and it can be any iterable.
- For numpy arrays like `arr_ctl_phys` and `arr_ctl_der_phys`, iterating over them produces sub-arrays in the order `arr_ctl_phys[0], arr_ctl_phys[1], ...`, and the ordering happens to match the positional arguments to the `Centerline` constructor.
- This doesn't account for the `.tolist()` part, but the `Centerline` constructor happens to only use these parameters by iterating over them once. So they would act the same even without `.tolist()`.

But, again, maybe too obscure.

mguaypaq (approved):
Looks good, thanks!

spinalcordtoolbox/csa_pmj.py
joshuacwnewton: !!! This actually works? This is lovely. 

I've used `*` syntax before when supplying arguments, but I hadn't realized it could be used on numpy arrays, too! 

Thanks for the tip. :hearts:

joshuacwnewton (approved):

spinalcordtoolbox/csa_pmj.py
joshuacwnewton: Quick test:

image

image

NadiaBlostein:
Thank you for taking this on!!! 😍

PR #4016, opened 2023-01-29
Faithfully display `sct_qc` arguments, rather than making a fake command

### LINKED ISSUE #3356: `sct_qc` displays the wrong arguments/command
When using `sct_qc`, it will take the script name from the `-p` argument, then incorrectly combine it with the other arguments from `sct_qc`. Example:

```bash
# Input command 
sct_qc -i dmri_crop.nii.gz -d dmri_crop_moco.nii.gz -s dmri_crop_moco_dwi_mean_seg.nii.gz -p sct_dmri_moco -qc "$SCT_BP_QC_FOLDER"

# What it actually displays (these are not valid arguments for 'sct_dmri_moco`)
sct_dmri_moco -i dmri_crop.nii.gz -d dmri_crop_moco.nii.gz -s dmri_crop_moco_dwi_mean_seg.nii.gz
```

Screenshot from 2021-04-21 17-27-17

I imagine that this is not specific to `sct_dmri_moco`, but I have not checked yet.

joshuacwnewton:
## Description

Rather than generating a fake, incorrect set of arguments that mimic the original command, such as:

```
sct_dmri_moco -i dmri_crop.nii.gz -d dmri_crop_moco.nii.gz
              -s dmri_crop_moco_dwi_mean_seg.nii.gz
```

We instead try to faithfully display the arguments that were passed to sct_qc itself:

```
sct_dmri_moco ("sct_qc -i dmri_crop.nii.gz -d dmri_crop_moco.nii.gz
                       -s dmri_crop_moco_dwi_mean_seg.nii.gz
                       -p sct_dmri_moco -qc .")
```

## Linked issues

Fixes #3356.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_qc.py
+++ b/spinalcordtoolbox/scripts/sct_qc.py
@@ -81,16 +81,10 @@ def main(argv: Sequence[str]):
     set_loglevel(verbose=verbose)
 
     from spinalcordtoolbox.reports.qc import generate_qc
-    # Build args list (for display)
-    args_disp = '-i ' + arguments.i
-    if arguments.d:
-        args_disp += ' -d ' + arguments.d
-    if arguments.s:
-        args_disp += ' -s ' + arguments.s
     generate_qc(fname_in1=arguments.i,
                 fname_in2=arguments.d,
                 fname_seg=arguments.s,
-                args=args_disp,
+                args=f'("sct_qc {" ".join(argv)}")',
                 path_qc=arguments.qc,
                 dataset=arguments.qc_dataset,
                 subject=arguments.qc_subject,
--- a/spinalcordtoolbox/scripts/sct_qc.py
+++ b/spinalcordtoolbox/scripts/sct_qc.py
@@ -11,7 +11,7 @@ import os
 import sys
 from typing import Sequence
 
-from spinalcordtoolbox.utils import init_sct, set_loglevel, SCTArgumentParser
+from spinalcordtoolbox.utils import init_sct, set_loglevel, SCTArgumentParser, list2cmdline
 
 
 def get_parser():
@@ -84,7 +84,7 @@ def main(argv: Sequence[str]):
     generate_qc(fname_in1=arguments.i,
                 fname_in2=arguments.d,
                 fname_seg=arguments.s,
-                args=f'("sct_qc {" ".join(argv)}")',
+                args=f'("sct_qc {list2cmdline(argv)}")',
                 path_qc=arguments.qc,
                 dataset=arguments.qc_dataset,
                 subject=arguments.qc_subject,


spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -81,16 +81,10 @@ def main(argv: Sequence[str]):
...
-    args_disp = '-i ' + arguments.i
-    if arguments.d:
-        args_disp += ' -d ' + arguments.d
-    if arguments.s:
-        args_disp += ' -s ' + arguments.s
     generate_qc(fname_in1=arguments.i,
                 fname_in2=arguments.d,
                 fname_seg=arguments.s,
-                args=args_disp,
+                args=f'("sct_qc {" ".join(argv)}")',
```
mguaypaq, line 87: I'm probably just being paranoid, but how about using `spinalcordtoolbox.utils.list2cmdline` here? It's what `generate_qc` would end up doing if we passed a list instead of a string, and it would take care of shell quoting any weird arguments. (This would also need an extra import at the top of the file.)
```suggestion
                args=f'(sct_qc {list2cmdline(argv)})',
```

mguaypaq (approved):
You're right, it's better to report a true command line rather than trying to rebuild a fictional one.

spinalcordtoolbox/scripts/sct_qc.py
joshuacwnewton: In truth I hadn't realized this function existed! 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/842b0422f1b6f1743be42d366927e1af5c7a9f8f/spinalcordtoolbox/utils/sys.py#L364-L365

Looking at [`shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote), it seems to be designed to add safety to  strings that will be executed as command line arguments. My hunch was that it was originally designed for `run_proc`, given its usage there?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/842b0422f1b6f1743be42d366927e1af5c7a9f8f/spinalcordtoolbox/utils/sys.py#L385-L388

_(In fact, we should probably be using `shlex.quote` directly on `cmd`, too!)_

But, since we're not executing anything in the context of QC reports, I'm wondering if using this function is overkill?

joshuacwnewton: > It's what `generate_qc` would end up doing if we passed a list instead of a string, and it would take care of shell quoting any weird arguments.

Although, as you say, `generate_qc` does use this function too... I wonder why? Does it have a meaningful effect there?

I'm wondering if `list2cmdline` should just be inlined in `run_proc`, and QC should be using `" ".join`?

mguaypaq: > I'm wondering if `list2cmdline` should removed, with the functionality inlined in `run_proc`, and `generate_qc` should be using `" ".join` instead? Just to convey that `shlex.quote` should only be used in the context of commands to execute.

I don't think that would work. The most common "weird" character handled by `shlex.quote` is a space character. So, calling `shlex.quote` before or after `" ".join` would have a drastically different effect! For example:
```bash
$ args() { printf 'got %s arguments:\n' "$#"; printf '|%s|\n' "$@"; }
$ args 'a' 'b' ' c'
got 3 arguments:
|a|
|b|
| c|
$ args 'a b  c'
got 1 arguments:
|a b  c|
```

mguaypaq: > But, since we're not executing anything in the context of QC reports (we're just echoing what was passed to the command line), I'm wondering if using this function is overkill? It might even mislead the user as to what was actually passed as arguments?

That's fair. If we don't intend/expect the command to be copy-pasted into a terminal, then maybe we can be careless with the quoting. The only downside is that doing `" ".join` on the arguments loses information in the (unlikely?) case that an argument actually contained a space character.

joshuacwnewton: > So, calling `shlex.quote` before or after `" ".join` would have a drastically different effect!

Ah, my apologies. I think I may have miscommunicated? My suggestion was:

- Replace `list2cmdline` with `" ".join` inside of `generate_qc`.
- Then, after the above change, `list2cmdline` would only be used in 1 place (`run_proc`). So, get rid of the function itself, move the functionality directly into `run_proc`.

I didn't mean to suggest using `shlex.quote` in tandem with `" ",join`? Or maybe I'm misunderstanding what you're saying.

joshuacwnewton: > If we don't intend/expect the command to be copy-pasted into a terminal, then maybe we can be careless with the quoting.

OH! Of course!! This is a very very good point. I completely missed that the displayed string could be reused in that way.

PR #4021, opened 2023-01-30
Use `filters.rank` instead of `dilate`/`erode` for `uint8`/`uint16` images

### LINKED ISSUE #4020: Bypass dilation if the goal is to crop an image with larger input mask
## Context

Usually in pipelines we segment the cord, dilate it, and then used the dilated cord as a mask to crop an image. See [example](https://github.com/ivadomed/model_seg_ms_mp2rage/blob/33a3a717e9b9677622b57d09980d121bf5ec0bad/preprocessing/preprocess_data.sh#L123-L127). 

While this approach works well, it is highly inefficient because the dilation can take a _looooong_ time (>1min), which is unnecessary because the cropping algorithm only needs xmin/xmax, ymin/ymax, etc.

## Solution

From the input segmentation (or whatever binary mask) and an input 'dilation', which adds a value to the max coordinate and subtract a value to the min coordinate (ie: exactly what the dilation filter would do), get the updated box coordinates for the cropping.

Example:
```
sct_crop_image -i IMAGE -m MASK -dilate 20x20x5 -o IMAGE_CROPED
```

Here, if the mask defines xmin and xmax, the output image is cropped around xmin-20 and xmax+20.

joshuacwnewton:
## Description

When images are `uint8` or `uint16`, using `filters.rank.{maximum,mininum}` is *much* faster than using `dilation` or `erosion`, especially in our use-case (large 3D images).

Source: [`skimage.morphology.dilation`](https://scikit-image.org/docs/stable/api/skimage.morphology.html#dilation)

> For uint8 (and uint16 up to a certain bit-depth) data, the lower algorithm complexity makes the [skimage.filters.rank.maximum](https://scikit-image.org/docs/stable/api/skimage.filters.rank.html#skimage.filters.rank.maximum) function more efficient for larger images and footprints.

Anecdotally, on my personal laptop, testing on a `sct_example_data/t2` segmentation using `-dilate 32` caused the processing time to reduce from ~260s to ~6s. (And my CPU kind of sucks, so I imagine this would be even faster on better computers.)

## Linked issues

Related to #4020.

### DIFF
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -5,7 +5,7 @@ import logging
 
 import numpy as np
 from skimage.morphology import erosion, dilation, disk, ball, square, cube
-from skimage.filters import threshold_local, threshold_otsu
+from skimage.filters import threshold_local, threshold_otsu, rank
 from scipy.ndimage.filters import gaussian_filter, gaussian_laplace
 from scipy.stats import pearsonr, spearmanr
 from dipy.denoise.noise_estimate import estimate_sigma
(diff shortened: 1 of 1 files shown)


joshuacwnewton:
@jcohenadad If it's not too much trouble, could you try out this PR using your existing MPRAGE scripts?

mguaypaq (approved):
This looks like a straightforward improvement, and [recommended by the skimage docs](https://scikit-image.org/docs/stable/api/skimage.filters.rank.html#maximum) too.

joshuacwnewton:
I'll unlink #4020, though, since we're going to go ahead with the feature addition to `sct_crop_image`.

Issue #4012, opened 2023-01-24
`sct_get_centerline -method fitseg` - allow to output soft centerline


valosekj:
## Feature justification

Within the [centerline detection](https://github.com/neuropoly/idea-projects/issues/15) project, we use [`sct_get_centerline -method fitseg`](https://github.com/ivadomed/model_sc_centerline/blob/70f6e189fde93070e917bd96175266bf7c426d14/scripts/get_centerline_from_segmentation.sh#L90-L92) to obtain SC centerline from existing ground truth SC segmentations in order to train a DL model for centerline detection. 

Currently, `sct_get_centerline -method fitseg` outputs binarized centerline. But, for DL training, it would be beneficial to have a probabilistic (soft) centerline (i.e., values can lie between 0 and 1) as well.

Possibly related to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4006

joshuacwnewton:
This issue was brought up in the [2023-01-26](https://docs.google.com/document/d/1ItApJQfajO2lRzOU2yenWbeRg6alfsdut3J4AVVdo78/edit#heading=h.2xefygcyqgee) SCT meeting, and it was suggested that this might be as simple as a one-line change (updating to linear interpolation). 

See the `-centerline-algo` parameter, which is applied to `-method fitseg`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d09af3ee707ca166c3d1611295a7575ffdd4cd0/spinalcordtoolbox/scripts/sct_get_centerline.py#L61-L66

naga-karthik:
Thanks @valosekj for digging into the code! Indeed these lines are the culprit for binarized outputs. 
```console

im_centerline.data[round_and_clip(x_centerline_fit, clip=[0, im_centerline.data.shape[0]]), 
                    round_and_clip(y_centerline_fit, clip=[0, im_centerline.data.shape[1]]), 
                    z_ref] = 1 
```

> If we used in this step the float centerline coordinates to set the intensity of the surrounding voxels to interval 0 to 1, we could obtain a soft segmentation.

Following up on this - there is the question of "how many" surrounding coordinates we should consider to get the soft segmentation. One suggestion is to use 3 voxels per extracted centerline coordinate - one would the `.floor()`, one would be the `.ceil()` and the third would be the original (un-rounded, float) coordinate whose intensity we set as 1. In code, this looks something like:

```python

[...]

PR #4026, opened 2023-02-02
Add `-centerline-soft` option to output a non-binary "soft" centerline

valosekj:
#### GitHub

#### PR contents

## Description

This PR allows outputting of soft (i.e., non-binary) centerline when running `sct_get_centerline -method fitseg` based on the suggestion in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4012#issuecomment-1411156250. 

The soft centerline will be used for training of the [DL centerline detection model](https://github.com/neuropoly/idea-projects/issues/15).

The function can be run by:

```bash
git fetch
git checkout jv/4012_allow_soft_centerline
git pull
sct_get_centerline -i <filename>_seg.nii.gz -method fitseg -centerline-soft 1
```

The current binary centerline (contains only value `1`):
bin

The proposed soft centerline (contains values in the range `0` to `1`):
soft

Example data:
[sub-cal072_ses-M0_T2w_seg.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/10573622/sub-cal072_ses-M0_T2w_seg.nii.gz)

Thanks to @naga-karthik and @NathanMolinier for their collaboration! 🍾 

## Linked issues
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4012

---

Question:

- do we want to include the suffix `probseg` (i.e., to `_centerline_probseg.nii.gz`) to make explicit that the output is a soft segmentation (see [intranet BIDS convention](https://intranet.neuro.polymtl.ca/data/dataset-curation.html#derivatives-structure))?


joshuacwnewton:
> * do we want to include the suffix `propseg` (i.e., to `_centerline_propseg.nii.gz`) to make explicit that the output is a soft segmentation (see [intranet BIDS convention](https://intranet.neuro.polymtl.ca/data/dataset-curation.html#derivatives-structure))?

I think there might be a typo in that page? (In one place, it says "propseg", but later on in the document, it says "pro**b**seg", as in "probabilistic segmentation". I'm assuming we want the latter, instead of `sct_propseg`? :stuck_out_tongue_closed_eyes:)

spinalcordtoolbox/centerline/core.py
```diff
@@ -13,7 +13,7 @@
 
 class ParamCenterline:
     """Default parameters for centerline fitting"""
-    def __init__(self, algo_fitting='bspline', degree=5, smooth=20, contrast=None, minmax=True):
+    def __init__(self, algo_fitting='bspline', degree=5, smooth=20, soft=0, contrast=None, minmax=True):
```
joshuacwnewton, line 16: Very small nitpick, but I think it's generally best practice to add new parameters to the very end of the function? 

That way, we can trust that existing calls that specify arguments positionally (without keywords): 

```python
ParamCenterline(algo_fitting, degree, smooth, contrast, mixmax)
```

Won't accidentally assign the wrong value to the wrong parameter. (In this case, `contrast` would get applied to `soft`.)

spinalcordtoolbox/centerline/core.py
```diff
@@ -175,9 +176,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
         im_centerline.data = np.zeros(im_centerline.data.shape)
         # Assign value=1 to centerline. Make sure to clip to avoid array overflow.
         # TODO: check this round and clip-- suspicious
```
joshuacwnewton, line 178: We may want to move / rewrite this comment, since it only applies to to the `binary` case?

valosekj:
> > * do we want to include the suffix `propseg` (i.e., to `_centerline_propseg.nii.gz`) to make explicit that the output is a soft segmentation (see [intranet BIDS convention](https://intranet.neuro.polymtl.ca/data/dataset-curation.html#derivatives-structure))?
> 
> I think there might be a typo in that page? (In one place, it says "propseg", but later on in the document, it says "pro**b**seg", as in "probabilistic segmentation". I'm assuming we want "prob" instead of "prop" -- propseg refers to [mesh propagation](https://spinalcordtoolbox.com/user_section/tutorials/segmentation/sct_propseg.html)? 😝)

Yeah! Good catch! You are right, [`probseg`](https://bids-specification.readthedocs.io/en/stable/glossary.html#probseg-suffixes) is correct. Thank you! I actually did this typo also in the past 😅

spinalcordtoolbox/centerline/core.py
```diff
@@ -175,9 +176,30 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+                               z_ref] = 0.5 * (1 - (x_centerline_fit - x_centerline_fit.astype(int)) +
+                                               (1 - (y_centerline_fit - y_centerline_fit.astype(int))))
+            im_centerline.data[np.floor(x_centerline_fit).astype(int),
+                               np.ceil(y_centerline_fit).astype(int),
+                               z_ref] = 0.5 * (1 - (x_centerline_fit - x_centerline_fit.astype(int)) +
+                                               (y_centerline_fit - y_centerline_fit.astype(int)))
+            im_centerline.data[np.ceil(x_centerline_fit).astype(int),
+                               np.floor(y_centerline_fit).astype(int),
+                               z_ref] = 0.5 * (x_centerline_fit - x_centerline_fit.astype(int) +
+                                               (1 - (y_centerline_fit - y_centerline_fit.astype(int))))
```
joshuacwnewton, line 201: I think we could maybe save some space, and make things a little more readable, by precomputing some of the repeated quantities?

```python
            # Compute integer pixel locations (all together, these form a 2x2 grid around centerline)
            x_ceil = np.ceil(x_centerline_fit).astype(int)
            y_ceil = np.ceil(y_centerline_fit).astype(int)
            x_floor = np.floor(x_centerline_fit).astype(int)
            y_floor = np.floor(y_centerline_fit).astype(int),
            # Compute the difference between the centerline and its rounded values
            x_diff = x_centerline_fit - x_centerline_fit.astype(int)
            y_diff = y_centerline_fit - y_centerline_fit.astype(int)
            # Assign soft segmentation values to the 2x2 grid based on the computed differences
            im_centerline.data[x_ceil, y_ceil, z_ref]   = 0.5 * (x_diff       + y_diff)
            im_centerline.data[x_floor, y_floor, z_ref] = 0.5 * ((1 - x_diff) + (1 - y_diff))
            im_centerline.data[x_floor, y_ceil, z_ref]  = 0.5 * ((1 - x_diff) + y_diff))
            im_centerline.data[x_ceil, y_floor, z_ref]  = 0.5 * (x_diff       + (1 - y_diff))
```

I think this makes it easier to intuitively see how the soft segmentation is computed.

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -178,6 +186,10 @@ def main(argv: Sequence[str]):
     else:
         printv("ERROR: The selected method is not available: {}. Please look at the help.".format(method), type='error')
         return
+    if arguments.centerline_soft == 1:
+        param_centerline.soft = 1
+    else:
+        param_centerline.soft = 0
```
joshuacwnewton, line 192: Since the choices are just `[0, 1]`, I don't think we need an if-else?

```suggestion
param_centerline.soft = arguments.centerline_soft
```

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -71,6 +71,14 @@ def get_parser():
...
         help="Degree of smoothing for centerline fitting. Only for -centerline-algo {bspline, linear}."
     )
+    optional.add_argument(
+        "-centerline-soft",
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=0,
+        help="Binary or soft centerline. 0 = binarized, 1 = soft"
+    )
```
joshuacwnewton, line 81: Since the script is called `sct_get_centerline`, I don't think we necessarily need to include the word "centerline" in the argument name?

It should be relatively understandable if the argument is just called `-soft`, I think?

joshuacwnewton (changes requested):
Thank you for taking this on!! This looks like a wonderful contribution. :tada:

spinalcordtoolbox/centerline/core.py
joshuacwnewton: Maybe we could say like, `x__ceil` or something, and use extra brackets, as a way to better align the calls? Hehe...

```python
      # Assign soft segmentation values to the 2x2 grid based on the computed differences
      im_centerline.data[x__ceil, y__ceil, z_ref] = 0.5 * ((x_diff)     + (y_diff))
      im_centerline.data[x_floor, y_floor, z_ref] = 0.5 * ((1 - x_diff) + (1 - y_diff))
      im_centerline.data[x_floor, y__ceil, z_ref] = 0.5 * ((1 - x_diff) + (y_diff))
      im_centerline.data[x__ceil, y_floor, z_ref] = 0.5 * ((x_diff)     + (1 - y_diff))
```

Probably not too important, but I really like when everything aligns all nicely. :stuck_out_tongue_closed_eyes:

jcohenadad (approved):
thank you team! amazing contribution indeed

spinalcordtoolbox/centerline/core.py
valosekj: Good point! Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/97dc82f07cc874850b3816c85f1ef1a5f2a62c7e

spinalcordtoolbox/centerline/core.py
valosekj: Awesome suggestion! 👍🏻  Applied in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/5ec33d6416c936ebe2af657224f311ee13a2aa93

spinalcordtoolbox/scripts/sct_get_centerline.py
valosekj: You are right! `if-else` removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/5693e0af70cbe5258e5317ec59df089d87702e30

spinalcordtoolbox/centerline/core.py
valosekj: Right! Comment moved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/30cb6c83801ab8098ef04e1243cdbd2197a535de. I also removed the TODO. After playing with this part of the code, I don't think that "round and clip" is suspicious -- it is working as expected.

spinalcordtoolbox/scripts/sct_get_centerline.py
valosekj: Yeah. Good point! This is exactly what we were considering. But since other arguments such as [-centerline-algo](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7cb30ddbaa88ded9c3ae6250c72832f5c40f4d64/spinalcordtoolbox/scripts/sct_get_centerline.py#L62) and [-centerline-smooth](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7cb30ddbaa88ded9c3ae6250c72832f5c40f4d64/spinalcordtoolbox/scripts/sct_get_centerline.py#L68) also contains the word "centerline", we decided to include the word also to the `-centerline-soft` argument.

valosekj:
@joshuacwnewton thank you for the review! 💯  I addressed/answered your suggestions and comments.

spinalcordtoolbox/scripts/sct_get_centerline.py
joshuacwnewton: Ahh, very good point!

spinalcordtoolbox/centerline/core.py
```diff
@@ -173,11 +175,28 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
+            x_floor = np.floor(x_centerline_fit).astype(int)
+            y_floor = np.floor(y_centerline_fit).astype(int),
+            # Compute the difference between the centerline and its rounded values
+            x_diff = x_centerline_fit - x_centerline_fit.astype(int)
+            y_diff = y_centerline_fit - y_centerline_fit.astype(int)
+            # Assign soft segmentation values to the 2x2 grid based on the computed differences
+            im_centerline.data[x_ceil, y_ceil, z_ref] = 0.5 * (x_diff + y_diff)
+            im_centerline.data[x_ceil, y_floor, z_ref] = 0.5 * (x_diff + (1 - y_diff))
+            im_centerline.data[x_floor, y_ceil, z_ref] = 0.5 * ((1 - x_diff) + y_diff)
+            im_centerline.data[x_floor, y_floor, z_ref] = 0.5 * ((1 - x_diff) + (1 - y_diff))
```
joshuacwnewton, line 198: Also, I just want to sanity check that the distribution of `x/y_diff` and `(1 - x/y_diff)` is correct:

Say, for example, that `x_centerline_fit` and `y_centerline_fit` are both at voxel location `10.75`:

- `ceil` == `11`
- `floor` == `10`
- `diff` == `0.75`
- `[11,11]` = `0.25 * (0.75 + 0.75)` = `0.375`
- `[11,10]` = `0.25 * (0.75 + 0.25)` = `0.25`
- `[10,11]` = `0.25 * (0.25 + 0.75)` = `0.25`
- `[10,10]` = `0.25 * (0.25 + 0.25)` = `0.125`

That makes sense to me? The voxels sum to 1, and it's more heavily weighted towards `[11,11]`.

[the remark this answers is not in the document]
joshuacwnewton: Also, it might be worth writing a test for this? e.g. running `sct_get_centerline` twice (`-soft 0` and `-soft 1`), then verifying that the two centerlines A) overlap and B) the soft centerline adds to `1`.

That way, we can modify this code in the future and verify that it still produces expected output.

joshuacwnewton: Actually, the more I think about it, the more I'm wondering whether the weighting is correct? 

For example, say it was `[10.9, 10.9]` instead:

- `[11,11]` = `0.25 * (0.90 + 0.10)` = `0.45`
- `[11,10]` = `0.25 * (0.90 + 0.10)` = `0.25`
- `[10,11]` = `0.25 * (0.10 + 0.90)` = `0.25`
- `[10,10]` = `0.25 * (0.10 + 0.10)` = `0.05`

Or even say it was `[10.99, 10.99]` instead:

- `[11,11]` = `0.25 * (0.99 + 0.99)` = `0.495`
- `[11,10]` = `0.25 * (0.99 + 0.01)` = `0.25`
- `[10,11]` = `0.25 * (0.01 + 0.99)` = `0.25`
- `[10,10]` = `0.25 * (0.01 + 0.01)` = `0.005`

Maybe my math skills are lacking here, but this doesn't make intuitive sense to me?

Since `[10.99,10.99]` is practically `[11,11]`, shouldn't most of the weighting be put on the `[11,11]` voxel? Instead, there's an upper bound of `0.5` for each voxel, and the corner voxels will always add to `0.25`, which doesn't seem right  to me... :thinking:

joshuacwnewton (changes requested):

spinalcordtoolbox/centerline/core.py
naga-karthik: Hey @joshuacwnewton, thanks for looking into the math in this much detail! We actually considering the 0.25 weighting for pixel too but, as you also ended up finding through your calculations, if the actual centerline coordinates are `[10.9, 10.9]` then we would assume that the pixel `[11, 11]` will have the highest intensity. We are somewhat seeing that when we choose 0.5 as the weighting. With 0.25, the intensities are just too small for a pixel and it didn't make sense to us as well!

@NathanMolinier had suggested an alternative where the diagonals sum up to 1, that is, if the `[11, 11]` has the highest intensity, then the diagonally opposite pixel (usually the one with (floor, floor)) should have the least intensity. This is what we implemented here.

naga-karthik: Moreover, from the screenshots that @valosekj posted at https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026#issue-1568820188, looking at the intensity of the binarized centerline, we were thinking that our soft centerline should have similar intensity at that pixel if not as high as the binarized. Hence, if we choose 0.25 as the weighting, the 2 pixels (left and the top) will have similar intensities, and we wouldn't be sure where the actual centerline is. 

(Also, note that I'm just trying to explain our thought process as we did the math on a paper. The intention is not to sound too defensive! Let me know if any of what I said doesn't make sense! Thanks again for looking into the details, this is really useful!

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -178,6 +186,8 @@ def main(argv: Sequence[str]):
     else:
         printv("ERROR: The selected method is not available: {}. Please look at the help.".format(method), type='error')
         return
+    # Binarized or soft centerline
+    param_centerline.soft = arguments.centerline_soft
```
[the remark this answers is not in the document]
valosekj: Good point! Like this https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/0f228da127cc78408462b824624564ea7e09a7ba?

spinalcordtoolbox/centerline/core.py
joshuacwnewton: I have two thoughts here:

#### 1.

I don't think my concern from https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026#discussion_r1095891109 has been addressed?

My understanding of partial volume calculations was that the "soft" segmentation must sum to the same values as the "binarized" segmentation (in this case, 1). But, right now the 2x2 grid sums to 2 instead, which breaks the requirement for partial volume calculations. (I think?)

#### 2.

> Hence, if we choose 0.25 as the weighting, the 2 pixels (left and the top) will have similar intensities, and we wouldn't be sure where the actual centerline is.

You're right, this is a problem. However, I think that the problem you describe is independent of the constant scaling factor; whether we use `0.25` or `0.5` shouldn't matter, because the scaling factor is applied equally to all 4 voxel locations. It's something that can be factored out?

To fix the problem of "unclear weighting", I think the actual "distance" computation (i.e. x_diff, y_diff) needs to be updated, rather than the constant scaling factor.

mguaypaq: I believe the following formulas would be better, because:
- the weights would still add up to 1 (up to a tiny floating point error); and
- the initial floating point coordinates of the centerline will be the weighted average of the rounded coordinates (with the given weights/probabilities). (In other words, coordinates like `[10.9, 10.9]` and `[11.1, 11.1]` would put almost all the weight on coordinate `[11, 11]`.)

Additionally:
- In both cases (`param.soft == 0` and `param.soft == 1`), it would be wise to clip the centerline coordinates to the ranges `[0, xdim-1]` and `[0, ydim-1]`, because other values (including `xdim` and `ydim` themselves) would cause array indexing errors.
- Note that the expected relationship `ceil(x) == floor(x)+1` fails when `x` happens to be an integer.
  - This is actually good for us, because using just `floor(x)` and `floor(x)+1` would run the risk of getting outside the bounds `[0, xdim-1]`.
  - But we have to be careful, otherwise we might _overwrite_ one of the probabilities with another one, and the total would become less than 1. The way to fix this is to add, rather than assign, the probabilities.

```py3
x_max, y_max = im_centerline.data.shape[:2]
x_centerline_fit = x_centerline_fit.clip(0, x_max-1)
y_centerline_fit = y_centerline_fit.clip(0, y_max-1)

x_floor = np.floor(x_centerline_fit).astype(int)
y_floor = np.floor(y_centerline_fit).astype(int)

x_ceil = np.ceil(x_centerline_fit).astype(int)
y_ceil = np.ceil(y_centerline_fit).astype(int)

x_frac = x_centerline_fit - x_floor
y_frac = y_centerline_fit - y_floor

im_centerline.data[x_floor, y_floor, z_ref] += (1-x_frac) * (1-y_frac)
im_centerline.data[x_floor, y_ceil, z_ref] += (1-x_frac) * y_frac
im_centerline.data[x_ceil, y_floor, z_ref] += x_frac * (1-y_frac)
im_centerline.data[x_ceil, y_ceil, z_ref] += x_frac * y_frac
```

joshuacwnewton: So, in this case, for the coordinate `[10.99,10.99]`, we get:

<details>

* `floor`: 10
* `ceil`: 11
* `frac`: 0.99
* `[10,10]`: 0.01 * 0.01 = 0.0001
* `[10,11]`: 0.01 * 0.99 = 0.0099
* `[11,10]`: 0.99 * 0.01 = 0.0099
* `[11,11]`: 0.99 * 0.99 = 0.9801

And, 0.0001 + 0.0099 + 0.0099 + 0.9801 = 1!

</details>

This matches my expectations exactly! :tada:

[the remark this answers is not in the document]
joshuacwnewton: Conversation superseded by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026#discussion_r1096093021.

spinalcordtoolbox/scripts/sct_get_centerline.py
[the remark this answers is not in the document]
joshuacwnewton: Perfect!

spinalcordtoolbox/centerline/core.py
naga-karthik: Thank you @mguaypaq for your suggestion! I think this covers the issues we have been having! 

There is one question though - Do you think the `x_max - 1` in `x_centerline_fit = x_centerline_fit.clip(0, x_max-1)` is necessary? Because, in the original implementation, we have this `round_and_clip(x_centerline_fit, clip=[0, im_centerline.data.shape[0]])` and it is clipping until what would be `x_max`.

mguaypaq: > There is one question though - Do you think the `x_max - 1` in `x_centerline_fit = x_centerline_fit.clip(0, x_max-1)` is necessary? Because, in the [original implementation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f228da127cc78408462b824624564ea7e09a7ba/spinalcordtoolbox/centerline/core.py#L181-L182), we have this `round_and_clip(x_centerline_fit, clip=[0, im_centerline.data.shape[0]])` and it is clipping until what would be `x_max`.

Yes, I think the `-1` is necessary, and I think that's a (very minor) bug in the original. For example:
```py3
>>> import numpy as np
>>>
>>> # Make a 2x4 numpy array
>>> data = np.arange(8).reshape((2, 4)); data
array([[0, 1, 2, 3],
       [4, 5, 6, 7]])
>>>
>>> # Get coordinates from its shape
>>> x_max, y_max = data.shape; data.shape
(2, 4)
>>>
>>> # We can access the last element with -1's
>>> data[x_max-1, y_max-1]
7
>>>
>>> # But without the -1's we get an IndexError
>>> data[x_max, y_max]
IndexError: index 2 is out of bounds for axis 0 with size 2
```

valosekj: Thanks all for the fruitful discussion! I implemented [Mathieu's solution](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026#discussion_r1096093021) in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/fcdcef65c0d2b0cef47235e88b5b83e699653079. 


> Yes, I think the `-1` is necessary, and I think that's a (very minor) bug in the original. For example:

Just for clarification. What do you mean by "the original"? Because [round_and_clip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f228da127cc78408462b824624564ea7e09a7ba/spinalcordtoolbox/centerline/core.py#L276) function `-1` contains:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f228da127cc78408462b824624564ea7e09a7ba/spinalcordtoolbox/centerline/core.py#L276

joshuacwnewton (changes requested):
This mostly looks good to me! I'm almost ready to give my approval.

I just had one last comment regarding the `display_viewer_syntax` command that gets run at the end of `sct_get_centerline`.

spinalcordtoolbox/centerline/core.py
mguaypaq: > Because [round_and_clip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f228da127cc78408462b824624564ea7e09a7ba/spinalcordtoolbox/centerline/core.py#L276) function `-1` contains:

I stand corrected! I just read the help strings for `np.round` and `np.clip` and assumed that a function called `round_and_clip` operated the same way, but I should have read the code for `round_and_clip` as well.

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -153,7 +161,8 @@ def main(argv: Sequence[str]):
     param_centerline = ParamCenterline(
         algo_fitting=arguments.centerline_algo,
         smooth=arguments.centerline_smooth,
-        minmax=True)
+        minmax=True,
+        soft=arguments.centerline_soft)
 
     # Output folder
     if arguments.o is not None:
```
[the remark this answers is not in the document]
valosekj: > `-cm red` is good for binarized centerlines, but isn't the best for soft centerlines:

Good catch! We also noticed this! We were manually changing the colormap in FSLeyes to `red-yellow`.

> The only problem here is that for some reason, the display_viewer_syntax function only provides a subset of the valid FSLeyes colors. 
>
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a4430850c4068ad2da4b5c6ca247a017fa6e1b03/spinalcordtoolbox/utils/shell.py#L115-L116

Indeed, `dict_fsleyes` does not contain all available colormaps. All available colormaps can be obtained by running `fsleyes -h`. It would be great if we could fetch the list of all available colormaps directly from fsleyes instead of specifying them manually.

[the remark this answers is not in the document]
joshuacwnewton: Ah! If you were using red-yellow, then we can just use that for now, rather than reworking the colormap logic?

```python
    cm_ctl = 'red-yellow' if arguments.centerline_soft else 'red'
    display_viewer_syntax([fname_input_data, file_output], colormaps=['gray', cm_ctl], opacities=['', '0.7'], verbose=verbose)
```

I'll open a new issue to address the limited colormap selection.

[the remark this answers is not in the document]
joshuacwnewton: > But, this isn't really a "true" segmentation... So, we might also want to look into the available options for ITKSnap to see if there are different flags we can use for soft segmentations.

To follow up on my comment about ITKSnap, it looks like there are no options for displaying a "soft" segmentation:

```
Image Options:
   -g FILE              : Load the main image from FILE
   -s FILE              : Load the segmentation image from FILE
   -l FILE              : Load label descriptions from FILE
   -o FILE [FILE+]      : Load additional images from FILE
                        :   (multiple files may be provided)
   -w FILE              : Load workspace from FILE
                        :   (-w cannot be mixed with -g,-s,-l,-o options)
```

So, I think sticking with `-s` is the best we can do on that front.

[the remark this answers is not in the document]
valosekj: > To follow up on my comment about ITKSnap, it looks like there are no options for displaying a "soft" segmentation:
> ....
> So, I think sticking with `-s` is the best we can do on that front.

I think so too.

spinalcordtoolbox/centerline/core.py
joshuacwnewton: Aside: The existence of the `round_and_clip` function is confusing to me. What is its purpose? To trip up developers? :stuck_out_tongue_winking_eye: 

It's literally just a mashup of the `round` and `clip` functions -- it will do one or the other, but not both, so there's not really any added benefit/functionality. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f228da127cc78408462b824624564ea7e09a7ba/spinalcordtoolbox/centerline/core.py#L275-L278

Why not just call `np.round` and `np.clip` individually when needed?

Plus, `round_and_clip` is only truly called in the context of `centerline/core.py`, so its usage isn't even that widespread:

image

Given that it's a source of confusion, I would suggest to get rid of it entirely?

spinalcordtoolbox/scripts/sct_get_centerline.py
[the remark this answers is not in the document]
joshuacwnewton: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/04681854bbeb35eaf1a0428d5567958a8bb98e61.

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -71,6 +71,14 @@ def get_parser():
...
         help="Degree of smoothing for centerline fitting. Only for -centerline-algo {bspline, linear}."
     )
+    optional.add_argument(
+        "-centerline-soft",
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=0,
+        help="Binary or soft centerline. 0 = binarized, 1 = soft. Only relevant with -method fitseg."
+    )
```
joshuacwnewton, line 81: Sorry to suggest one more change, but. Could you write a test for this new option?

It would be nice to run the command twice (`-centerline-soft {0,1}`) then compare the outputs to make sure:

- The softseg output is equal to 1 on all slices.
- The softseg overlaps with the binary segmentation on all slices.

This might be a little overkill? But it would allow us to make changes to this code in the future with confidence.

joshuacwnewton (changes requested):

spinalcordtoolbox/scripts/sct_get_centerline.py
valosekj: Sure, great suggestion! I will write tests next week.

valosekj: Hey @joshuacwnewton! @NathanMolinier and I tried to implement some first tests in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/e8b25c9921e24b721b010ecd245e87e978e0e989 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/120c94bc3f0414e0b4093816babb3079d271e374. It is not the final version yet, we just wanted to check if CI passes. But the CI has not started; instead, it is: "Waiting for status to be reported". Could you please restart it?
Also, there are some conflicts with master, could you please resolve them? Thank you!

joshuacwnewton: I believe the two problems are one and the same! (I think the conflicts need to be resolved for the CI to run.)

I'll take care of this right now. :)

joshuacwnewton: Note: The reason for the conflict was because of #4031, which removes the `round_and_clip` function. Resolving the conflict is as easy as replacing the `round_and_clip` parts of your PR with the version on master:

image

NathanMolinier:
Hey @joshuacwnewton, what are the next steps for this pull request ? Maybe you should check the tests we created with @valosekj ?

testing/cli/test_cli_sct_get_centerline.py
```diff
@@ -31,5 +33,54 @@ def test_sct_get_centerline_output_file_exists_with_o_arg(tmp_path, ext):
     ensure that the correct output file is created either way."""
     sct_get_centerline.main(argv=['-i', 't2s/t2s.nii.gz', '-c', 't2s', '-qc', str(tmp_path),
                                   '-o', os.path.join(tmp_path, 't2s_centerline'+ext)])
-    for file in [os.path.join('t2s', 't2s_centerline.nii.gz'), os.path.join('t2s', 't2s_centerline.csv')]:
+    for file in [os.path.join(tmp_path, 't2s_centerline.nii.gz'), os.path.join(tmp_path, 't2s_centerline.csv')]:
         assert os.path.exists(file)
```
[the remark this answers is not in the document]
joshuacwnewton: To make sure that the changes in this PR are focused on only the new additions, I would suggest to revert https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4026/commits/6956f709ef0919528e82905bc874e5aa22f75f09, and instead follow up in a separate PR with the fixes to the old tests.

Related: 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2911

testing/cli/test_cli_sct_get_centerline.py
```diff
@@ -31,5 +33,54 @@ def test_sct_get_centerline_output_file_exists_with_o_arg(tmp_path, ext):
...
+
+@pytest.mark.sct_testing
+@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
+@pytest.mark.parametrize('ext', ["", ".nii.gz"])
+def test_sct_get_centerline_soft_output_file_exists(tmp_path, ext):
+    """This test checks the output soft centerline using default usage of the CLI script."""
+    sct_get_centerline.main(argv=['-i', 't2/t2_seg-manual.nii.gz', '-method', 'fitseg', '-centerline-soft', '1', '-o',
+                                  os.path.join(tmp_path, 't2_seg_centerline_soft'+ext), '-qc', str(tmp_path)])
+    for file in [os.path.join(tmp_path, 't2_seg_centerline_soft.nii.gz'), os.path.join(tmp_path, 't2_seg_centerline_soft.csv')]:
+        assert os.path.exists(file)
```
joshuacwnewton, line 48: I think we could maybe remove this test?

```suggestion
```

My thinking here is:

- The other two tests for `-centerline-soft` will load the `.nii.gz` output file.
- This means that the only unique behavior this test checks for is the existence of the `.csv` file.
- But, this is *basically* covered by the previous tests, too? 
- (All that `-centerline-soft` does is modify the internals of `centerline/core.py` -- it doesn't make many changes to `sct_get_centerline` itself. So the method for generating CSV files is basically the same with or without `-centerline-soft`.)
- Therefore, this test is basically redundant with the previous tests? If one passes, then the other will too.

testing/cli/test_cli_sct_get_centerline.py
```diff
@@ -31,5 +33,54 @@ def test_sct_get_centerline_output_file_exists_with_o_arg(tmp_path, ext):
...
+    im_soft = Image(os.path.join(tmp_path, 't2_seg_centerline_soft.nii.gz'))  # Read soft centerline file
+    im_bin = Image(os.path.join(tmp_path, 't2_seg_centerline_bin.nii.gz'))  # Read binary centerline file
+    # Find the maximum intensity voxel across all slices in soft centerline and binary centerline
+    max_over_slices_soft = np.apply_over_axes(np.max, im_soft.data, [0, 2])
+    max_over_slices_bin = np.apply_over_axes(np.max, im_bin.data, [0, 2])
+    # Find the coordinates of the maximum in each slice in both soft and binary centerline
+    max_coords_over_slices_soft = np.transpose(np.where(im_soft.data == max_over_slices_soft))
+    max_coords_over_slices_bin = np.transpose(np.where(im_bin.data == max_over_slices_bin))
+    # Test if the maximum are the same between soft and binary centerline for each slices
+    assert (max_coords_over_slices_soft == max_coords_over_slices_bin).all()
```
joshuacwnewton, line 86: I think that we could probably merge these two tests together?

```suggestion


@pytest.mark.sct_testing
@pytest.mark.usefixtures("run_in_sct_testing_data_dir")
def test_sct_get_centerline_soft_sums_to_one_and_overlaps_with_bin(tmp_path):
    """
    This test checks two necessary conditions of the soft centerline:

    1) The sum of the output intensities of the soft centerline is equal to 1 on all slices
    2) The output maximum of the soft centerline overlaps with the binary segmentation on all slices.
    """
    # Condition 1: All slices of the soft centerline sum to 1
    sct_get_centerline.main(argv=['-i', 't2/t2_seg-manual.nii.gz', '-method', 'fitseg', '-centerline-soft', '1', '-o',
                                  os.path.join(tmp_path, 't2_seg_centerline_soft.nii.gz')])
    im_soft = Image(os.path.join(tmp_path, 't2_seg_centerline_soft.nii.gz'))
    # Sum soft centerline intensities in the (x,y) plane, across all slices
    sum_over_slices = np.apply_over_axes(np.sum, im_soft.data, [0, 2]).flatten()
    # Test if the summation for each slice is equal to 1
    assert (sum_over_slices == 1).all()

    # Condition 2: The max voxels of the soft centerline overlap with the binary centerline
    sct_get_centerline.main(argv=['-i', 't2/t2_seg-manual.nii.gz', '-method', 'fitseg', '-centerline-soft', '0', '-o',
                                  os.path.join(tmp_path, 't2_seg_centerline_bin.nii.gz')])
    im_bin = Image(os.path.join(tmp_path, 't2_seg_centerline_bin.nii.gz'))
    # Find the maximum intensity voxel across all slices in soft centerline and binary centerline
    max_over_slices_soft = np.apply_over_axes(np.max, im_soft.data, [0, 2])
    max_over_slices_bin = np.apply_over_axes(np.max, im_bin.data, [0, 2])
    # Find the coordinates of the maximum in each slice in both soft and binary centerline
    max_coords_over_slices_soft = np.transpose(np.where(im_soft.data == max_over_slices_soft))
    max_coords_over_slices_bin = np.transpose(np.where(im_bin.data == max_over_slices_bin))
    # Test if the maximum are the same between soft and binary centerline for each slices
    assert (max_coords_over_slices_soft == max_coords_over_slices_bin).all()
```

My thinking here is that both tests use the exact same calls to `sct_get_centerline`. So, the actual output file is the same between tests. So, we can check multiple of its properties in one test.

joshuacwnewton (changes requested):
The new additions to the test suite look good! I just had some small suggestions -- the bulk of the actual changes are great! :tada:

batch_processing.sh
```diff
@@ -68,6 +68,12 @@ cd t2
 sct_deepseg_sc -i t2.nii.gz -c t2 -qc "$SCT_BP_QC_FOLDER"
 # Tips: If you are not satisfied with the results you can try with another algorithm:
 # sct_propseg -i t2.nii.gz -c t2 -qc "$SCT_BP_QC_FOLDER"
+# Fit binarized centerline from SC seg (default settings)
+sct_get_centerline -i t2_seg.nii.gz -method fitseg -qc "$SCT_BP_QC_FOLDER"
+# Fit binarized centerline from SC seg (trying `-centerline-soft` and `o` flags)
+sct_get_centerline -i t2_seg.nii.gz -method fitseg -centerline-soft 0 -o t2_seg_centerline_bin.nii.gz -qc "$SCT_BP_QC_FOLDER"
```
joshuacwnewton, line 74: One last nitpick: I don't think we need to include both "default settings" (the first call) _and_  `-centerline-soft 0` (the second call), since they're basically doing the exact same thing? 

So, I think we can remove the `-centerline-soft 0` call:

```suggestion
# Fit binarized centerline from SC seg (default settings)
sct_get_centerline -i t2_seg.nii.gz -method fitseg -qc "$SCT_BP_QC_FOLDER"
```

[`batch_processing.sh`](https://spinalcordtoolbox.com/user_section/getting-started.html#batch-processing-example) is less of an exhaustive test, and is more meant to be a place where we showcase "typical CLI usage" for our users -- and our users won't be specifying `-centerline-soft 0` since that's the default value. :sweat_smile:

joshuacwnewton (approved):
I have one last tiny comment. Once that comment is addressed, I think this will be ready to merge!! (So, I'm going to pre-approve.)

Thank you so much for the contribution, @valosekj and @NathanMolinier!! :hearts:

PR #4027, opened 2023-02-06
Use Miniconda instead of built-in Python for Windows installations

### LINKED ISSUE #3942: Consider using Miniconda as the basis for Windows installations, rather than pure Python
When designing the `install_sct.bat` script for native Windows installations, I used Python's virtual environments, rather than Miniconda (which we had been using for our `install_sct` unix installations). 

My reasoning for this was that, at the time, we were working towards a pip-installable SCT, and so I wanted to keep the install script as light and "standard" as possible (so that the transition to `pip` would be as straightforward as possible).

However, in the following months, we've stopped actively pursing a pip-installable SCT due to a host of newly-identified blockers. And, in the meanwhile, the use of virtualenvs has caused a few minor bits of awkwardness:

- On one OS, we have a Python virtual environment, while on the others, we use Conda environments. Meaning that, documentation pages such as [Installing SCT for development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Installing-SCT-for-development) have to account for both environment types, making the documentation less straightfoward.
- Upgrading Python versions requires a documentation update to instruct the user to change their available version of Python, rather than that being a knob that we ourselves can control via the `conda create` command. (See also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3924.)
   - To add to this: What happens if the user changes their Python version? What happens if they've already installed a newer version of Python? Can we trust that the user will fiddle with their PATH to make SCT's chosen version of Python available on the command line?

So, perhaps instead of instructing the user to install Python, we can instruct them to install Miniconda, and then use `conda create` in the `install_sct.bat` script in much the same way we do for unix platforms. 

(This hasn't really been a big problem presently, but the more I think about it, the more I anticipate that it will become a problem in the future.)

joshuacwnewton:
## Description

Currently, our Windows install script (`install_sct.bat`) requires users to install Python 3.8 system-wide. `install_sct.bat` will then use the system Python to create a virtual environment that SCT will be installed into.

Relying on the system Python is fragile and error-prone, though. It makes upgrading Python versions difficult from SCT's end, and it requires users to juggle SCT's required Python version with any existing/future Python installations they may use on their computer.

So, this PR adds the same solution that we currently use in our Linux/macOS installations: Miniconda. This lets us automatically download and install an isolated, portable version of Python that we can update from our end whenever we like, solving the above issues and removing a prerequisite from the installation steps.

## Linked issues

Fixes #3942.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -71,13 +71,6 @@ jobs:
       run:
         shell: bash
     steps:
-
-    # NB: For Windows installations, users must install Python beforehand
-    - uses: actions/setup-python@v4
-      if: matrix.os == 'windows-2019'
-      with:
-        python-version: '3.8'
-
     - name: Checkout spinalcordtoolbox
       uses: actions/checkout@v2
 
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -39,12 +39,6 @@ jobs:
       - name: Checkout SCT
         uses: actions/checkout@v2
 
-      # NB: For Windows installations, users must install Python beforehand
-      - uses: actions/setup-python@v4
-        if: matrix.os == 'windows-2019'
-        with:
-          python-version: '3.8'
-
       - name: Install SCT (Unix)
         if: matrix.os != 'windows-2019'
         run: ./install_sct -y
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -384,11 +384,6 @@ jobs:
         shell: cmd
     steps:
     - uses: actions/checkout@v3
-    # NB: For Windows installations, users must install Python beforehand
-    - uses: actions/setup-python@v4
-      if: matrix.os == 'windows-2019'
-      with:
-        python-version: '3.8'
     - name: Install SCT
       # NB: In real-world usage, it's assumed that the user would download the
       # install_sct.bat file from the Downloads page, and _not_ the full repo.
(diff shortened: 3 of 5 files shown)


mguaypaq (approved):
I'm not really a Windows or Conda expert, but this looks sensible, the provided links to documentation agree, and the tests pass :slightly_smiling_face:

We should make a separate draft PR to update the [Windows installation docs](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a4430850c4068ad2da4b5c6ca247a017fa6e1b03/documentation/source/user_section/installation/windows.rst?plain=1#L33-L46), which can be merged as soon as this release goes live. (Although I guess we'll have to add yet another section, since Python will still be a prerequisite for installing older versions.)

joshuacwnewton:
> We should make a separate draft PR to update the [Windows installation docs](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a4430850c4068ad2da4b5c6ca247a017fa6e1b03/documentation/source/user_section/installation/windows.rst?plain=1#L33-L46), which can be merged as soon as this release goes live. (Although I guess we'll have to add yet another section, since Python will still be a prerequisite for installing older versions.)

Very good point. 

I wish we didn't have to do that, though... I wish I had done it this way from the get-go? Part of me wishes we could just backport this change to the older versions, and quietly ignore that we even recommended installing Python in the first place. :sweat:

mguaypaq:
> Part of me wishes we could just backport this change to the older versions

If the changes were only to files within `.github` and the release asset `install_sct.bat`, maybe we could get away with this. But because there's also a change in `batch_processing.sh`, I'm hesitant.

If it's any consolation, SCT isn't the only software with different installation instructions for different versions, for example [fsleyes does this too](https://open.win.ox.ac.uk/pages/fsl/fsleyes/fsleyes/userdoc/install.html).

PR #4045, opened 2023-02-21
Add dataset `data_normalize_metric_pam50`

### LINKED ISSUE #4044: Add  `data_normalize_metric_pam50` in `sct_download_data`
## Description

Related to the `sct_compute_compression `function #4003 , we added some morphometrics of healthy controls to normalize compression morphometrics in https://github.com/spinalcordtoolbox/data_normalize_metric_pam50. We ow need to add the download in `sct_download_data` during installation.

mguaypaq:
Fixes #4044.

For testing, I successfully downloaded the data with:
```
sct_download_data -d data_normalize_metric_pam50
```

Question for the reviewer(s): is `data_normalize_metric_pam50` the name we want to use for this dataset?

### DIFF
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -159,7 +159,13 @@ DATASET_DICT = {
             "https://github.com/ivadomed/multimodal-registration/releases/download/r20220512/models.zip"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "deepreg_models"),
-    }
+    },
+    "data_normalize_metric_pam50": {
+        "mirrors": [
+            "https://github.com/spinalcordtoolbox/data_normalize_metric_pam50/archive/refs/tags/r20230221.zip"
+        ],
+        "default_location": os.path.join(__sct_dir__, "data", "data_normalize_metric_pam50"),
+    },
 }
 
 


sandrinebedard:
Will it be downloaded with the SCT installation?

> Question for the reviewer(s): is `data_normalize_metric_pam50` the name we want to use for this dataset?

Yes it is fine!

spinalcordtoolbox/download.py
```diff
@@ -159,7 +159,13 @@
             "https://github.com/ivadomed/multimodal-registration/releases/download/r20220512/models.zip"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "deepreg_models"),
-    }
+    },
+    "data_normalize_metric_pam50": {
+        "mirrors": [
+            "https://github.com/spinalcordtoolbox/data_normalize_metric_pam50/archive/refs/tags/r20230221.zip"
```
joshuacwnewton, line 165: From what I can tell, this link comes from the "Source code (.zip)" download for the [`r20230221`](https://github.com/spinalcordtoolbox/data_normalize_metric_pam50/releases/tag/r20230221) tag.

However, is there merit to instead making a full-fledged release (like we have for other datasets)? And, if we do make a release, should we create a separate `.zip` asset (like we have done for other datasets)?

image

joshuacwnewton: For some historical context, NeuroPoly uses (or, used to use) an internally-developed Python package called [`distriploy`](https://github.com/neuropoly/distriploy) to upload zipped archives to GitHub and OSF. That's why you'll sometimes see `.distriploy.yml` files in repos such as `PAM50` and `sct_testing_data`.

Distriploy defaults to making separate `.zip` assets and attaching them to the release (even if they're identical to the Source Code .zip), which is why we see this behavior elsewhere.

joshuacwnewton: From what I can tell:

- Separate assets are static, while "Source code (.zip)" will adapt to whatever the state of the tag is.
- Separate assets provide download statistics, while "Source code (.zip)" do not.
- Separate assets are guaranteed to have their checksums preserved, while "Source code (.zip)" archives may change (see [recent controversy](https://news.ycombinator.com/item?id=34586917), though this was rolled back).

I don't know how much we care about any of these pros/cons? Part of me thinks that it's fine to just use the "Source code (.zip)" archives, especially in cases like this where the contents of the repo are _exactly_ what we want to be downloading.

spinalcordtoolbox/download.py
```diff
@@ -159,7 +159,13 @@
             "https://github.com/ivadomed/multimodal-registration/releases/download/r20220512/models.zip"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "deepreg_models"),
-    }
+    },
+    "data_normalize_metric_pam50": {
```
joshuacwnewton, line 163: > Question for the reviewer(s): is `data_normalize_metric_pam50` the name we want to use for this dataset?

Some nitpicks:

- Elsewhere in `sct_download_data`, we use `PAM50` (capitalized), so it might be good to unify that.
- Part of me wants to use a name that represents a noun, similar to the other dataset names? For example, rather than "data normalize metric pam50", we would say "PAM50-normalized metrics", which could then become the dataset name `PAM50_normalized_metrics` .

mguaypaq:
> Will it be downloaded with the SCT installation?

Not by default. Should it be? If so, we just need to add it to `install_sct` here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/158abd122c134c4cb9b447bc8623c75a3c9bf8e2/install_sct#L684-L689

and to `install_sct.bat` here:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/158abd122c134c4cb9b447bc8623c75a3c9bf8e2/install_sct.bat#L97-L100

spinalcordtoolbox/download.py
```diff
@@ -159,7 +159,13 @@
...
+        "mirrors": [
+            "https://github.com/spinalcordtoolbox/data_normalize_metric_pam50/archive/refs/tags/r20230221.zip"
+        ],
+        "default_location": os.path.join(__sct_dir__, "data", "data_normalize_metric_pam50"),
+    },
 }
```
joshuacwnewton, line 171: Side-question: Would it be beneficial to combine this PR with #4003?

The reason I suggest this is because of the following scenario:

- We merge this PR to master.
- [#4003](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003) is updated with master to include the new dataset link.
- We start playing around with [#4003](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003) to compute compression metrics from the CSV dataset.
- During testing, we identify some flaw or concern with the dataset (perhaps some issue with how the metrics were generated via [#3977](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977)).
- We need to regenerate the dataset and create a new release link, and thus a new change to `master`.

By comparison, if this dataset link were to be included as part of #4003, then it could potentially save a bit of clutter on master, and keep related development grouped in the same location, since any changes to the dataset would be part of #4003's development.

(But, maybe the scenario I describe is unlikely, or easy enough to handle if something goes wrong?)

spinalcordtoolbox/download.py
sandrinebedard: > (But, even if we do use the Source code .zips, we still might want to consider an actual release instead of just a tag?)

yes agreed, I actually deleated the relase associated with teh tag, I thought it didn't work!

We can go with the source.zip for now, and if the cons become a problem, we can always change that in the futur

spinalcordtoolbox/download.py
sandrinebedard: I am fine with renaming the repo and folder!

spinalcordtoolbox/download.py
sandrinebedard: yes, we can include this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003, I agree!

sandrinebedard:
> > Will it be downloaded with the SCT installation?
> 
> Not by default. Should it be? If so, we just need to add it to `install_sct` here:
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/158abd122c134c4cb9b447bc8623c75a3c9bf8e2/install_sct#L684-L689
> 
> and to `install_sct.bat` here:
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/158abd122c134c4cb9b447bc8623c75a3c9bf8e2/install_sct.bat#L97-L100

OK great! We discussed this in the sct_dev meeting and with the project team for the compression metrics, we want it within the sct installation!

sandrinebedard:
Closing since I merged this branch into https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003. --> https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4003/commits/124cf661520019ade1ef7cbb5204082f6a10ebf2

PR #4049, opened 2023-02-22
Allow `install_sct` to be run standalone (without downloading "Source code" archive)

### LINKED ISSUE #3545: SCT's current distribution method prevents us from checking download statistics

joshuacwnewton:
## Description

Currently, for Linux/macOS installations of SCT, users must navigate to the Release page, then download a "Source code (.zip)" archive. They must then `cd` into the folder and run `install_sct`. 

This is a bit unwieldy, and has also [prevented us from collecting download statistics](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3545) for the past several years. Without download statistics, it's more difficult to demonstrate the use and impact of SCT (for the purposes of e.g. grant proposals).

As a test, the Windows installer (`install_sct.bat`) was designed to be run standalone. This allowed us to upload the install script to the release as an asset, which [_does_ grant us download statistics](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3545#issuecomment-1334195311).

As a followup, then, this PR updates `install_sct` so that the Linux/macOS installers so that they can be run standalone, and thus uploaded to releases to benefit from the same statistics gathering.

## Future work

- [ ] Add steps to our CI to upload `install_sct` as a release asset (changing the default `git_ref` value from `master` to `<release milestone name>`)
  - We could potentially even upload `install_sct` twice as two separate release artifacts, Linux and macOS, to collect better data on who uses which OS. 
    (This would also make the installation process for each OS a bit more unified, since `install_sct.bat` gets renamed to `install_sct-5.8_win.bat`. Similarly, `install_sct` could become `install_sct-5.8_linux.sh` and `install_sct-5.8_macos.sh`.)
    (This could also lead us down a path where we remove all of the OS-specific if/else blocks that pollute `install_sct`, since we would maintain dedicated installers for Linux vs. macOS.)
  - This could be possibly be done as part of this PR, too?
- [ ] Update the documentation to describe the new installation process (and clarify that `git` is now a dependency).
- [ ] Better unify the installation process for Linux/macOS and Windows
   - [ ] Windows: Using a more sensible installation directory by default (`~/sct_<version>` instead of `~/spinalcordtoolbox`)
   - [ ] Windows: Allowing users to choose a custom installation directory.
   - [ ] Windows: Git cloning into a temporary directory first, then copying the files over to the target installation directory.
   - [ ] etc.

## Linked issues

Fixes #3545.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -22,6 +22,7 @@
 #  -d   Prevent the (re)-installation of the data/ directory
 #  -b   Prevent the (re)-installation of the SCT binaries files
 #  -c   Prevent checks from being run to validate the installation
+#  -g   Specify git ref (branch, tag, commit SHA) to clone when 'install_sct' is run by itself
 #  -v   Full verbose
 #
 #
@@ -57,6 +58,9 @@ NONINTERACTIVE=""
 NO_DATA_INSTALL=""
 NO_SCT_BIN_INSTALL=""
 NO_INSTALL_VALIDATION=""
+# CLI option -g: The git ref to use during the git clone (if `install_sct` is run by itself without source files)
+# (Default value: 'master', however this value is updated on stable release branches.)
+SCT_GIT_REF="master"
 
 # ======================================================================================================================
 # FUNCTIONS
@@ -348,7 +352,7 @@ for arg in "$@"; do
   esac
 done
 
-while getopts ":iydbcvh" opt; do
+while getopts ":iydbcgvh" opt; do
   case $opt in
   i)
     SCT_INSTALL_TYPE="in-place"
@@ -368,6 +372,10 @@ while getopts ":iydbcvh" opt; do
   c)
     echo " no checks will be run (installation will not be validated)"
     NO_INSTALL_VALIDATION="yes"
+    ;;
+    g)
+    echo " if script has been run by itself, git clone will use ref ${OPTARG}"
+    SCT_GIT_REF=${OPTARG}
     ;;
   v)
     echo " Full verbose!"
--- a/install_sct
+++ b/install_sct
@@ -352,7 +352,7 @@ for arg in "$@"; do
   esac
 done
 
-while getopts ":iydbcgvh" opt; do
+while getopts ":iydbcg:vh" opt; do
   case $opt in
   i)
     SCT_INSTALL_TYPE="in-place"
@@ -373,7 +373,7 @@ while getopts ":iydbcgvh" opt; do
     echo " no checks will be run (installation will not be validated)"
     NO_INSTALL_VALIDATION="yes"
     ;;
-    g)
+  g)
     echo " if script has been run by itself, git clone will use ref ${OPTARG}"
     SCT_GIT_REF=${OPTARG}
     ;;
--- a/install_sct
+++ b/install_sct
@@ -411,11 +411,15 @@ print info "
 fetch_os_type
 check_requirements
 
-# Catch SCT version
+# Check to see if the PWD contains the project source files (using `version.txt` as a proxy for the entire source dir)
 if [[ -e "spinalcordtoolbox/version.txt" ]]; then
   SCT_VERSION="$(cat spinalcordtoolbox/version.txt)"
+# If version.txt isn't present, then the installation script is being run without source files, so clone SCT
 else
-  die "ERROR: version.txt not found."
+  SCT_SOURCE="$TMP_DIR/spinalcordtoolbox"
+  echo "Source files not present. Cloning source files to $SCT_SOURCE."
+  git clone -b "$SCT_GIT_REF" --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git "$SCT_SOURCE"
+  SCT_VERSION=$(cat "$SCT_SOURCE/spinalcordtoolbox/version.txt")
 fi
 
 # Get installation type (from git or from package) if not already specified
--- a/install_sct
+++ b/install_sct
@@ -412,25 +412,28 @@ fetch_os_type
 check_requirements
 
 # Check to see if the PWD contains the project source files (using `version.txt` as a proxy for the entire source dir)
+# If it exists, then we can reliably access source files (`version.txt`, `requirements-freeze.txt) from the PWD.
 if [[ -e "spinalcordtoolbox/version.txt" ]]; then
   SCT_VERSION="$(cat spinalcordtoolbox/version.txt)"
-# If version.txt isn't present, then the installation script is being run without source files, so clone SCT
+  # Get installation type if not already specified
+  if [[ -z "$SCT_INSTALL_TYPE" ]]; then
+    # The file 'requirements-freeze.txt` only exists for stable releases
+    if [[ -e "requirements-freeze.txt" ]]; then
+      SCT_INSTALL_TYPE="package"
+    # If it doesn't exist, then we can assume that a dev is performing an in-place installation from master
+    else
+      SCT_INSTALL_TYPE="in-place"
+    fi
+  fi
+# If version.txt isn't present, then the installation script is being run by itself (i.e. without source files).
+# So, we need to clone SCT to a TMPDIR to access the source files, and update SCT_SOURCE accordingly.
 else
   SCT_SOURCE="$TMP_DIR/spinalcordtoolbox"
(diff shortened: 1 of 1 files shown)


install_sct
```diff
@@ -348,7 +352,7 @@ for arg in "$@"; do
   esac
 done
 
-while getopts ":iydbcvh" opt; do
+while getopts ":iydbcgvh" opt; do
```
mguaypaq, line 355: I believe you need a colon after `g` in the option spec to capture the following argument as `$OPTARG`:

```suggestion
while getopts ":iydbcg:vh" opt; do
```

install_sct
```diff
@@ -403,22 +411,29 @@ print info "
...
   fi
+# If version.txt isn't present, then the installation script is being run by itself (i.e. without source files).
+# So, we need to clone SCT to a TMPDIR to access the source files, and update SCT_SOURCE accordingly.
+else
+  SCT_SOURCE="$TMP_DIR/spinalcordtoolbox"
+  echo "Source files not present. Cloning source files to $SCT_SOURCE."
+  git clone -b "$SCT_GIT_REF" --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git "$SCT_SOURCE"
+  SCT_VERSION=$(cat "$SCT_SOURCE/spinalcordtoolbox/version.txt")
+  # Since we're git cloning into a TMPDIR, this can never be an "in-place" installation, so we force "package" instead.
+  SCT_INSTALL_TYPE="package"
```
mguaypaq, line 436: Would it be simpler to `git clone` into the current directory? A quick test on my machine shows that `git` will happily do so.
```suggestion
# So, we need to clone SCT to access the source files.
else
  echo "Source files not present. Cloning source files."
  git clone -b "$SCT_GIT_REF" --single-branch --depth 1 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git .
  SCT_VERSION=$(cat "spinalcordtoolbox/version.txt")
```
(The above suggestion isn't a complete implementation of this, though.)

mguaypaq (commented):
A couple of comments for right now (I'll finish this review a bit later).

install_sct
mguaypaq: If you do this, I would suggest keeping the `version.txt` test for the `git clone` logic, and then doing the rest of the logic (setting `SCT_VERSION`, checking for `requirements-freeze.txt`, etc.) unconditionally, afterwards.

mguaypaq (changes requested):
I guess my comments were more of a "request changes" nature. But overall, this looks good, thanks!

install_sct
joshuacwnewton: There are a couple of factors here that make this a bit tricky, I think?

- A user may download the install script, and then run it by double-clicking on it (in much the same way they're instructed to do for the Windows batch scripts). In that case, the current working directory may be `Downloads/`, `Desktop/`, etc. 
- Additionally, the script distinguishes between `$SCT_SOURCE` and `SCT_DIR`:
    - `$SCT_SOURCE`, in this case, would be the PWD, because it contains the source files.
    - `$SCT_DIR` is the directory that the user has chosen to install SCT to (typically e.g. `~/sct_5.8/`)
    - Once `$SCT_DIR` has been chosen, the source files are copied from `$SCT_SOURCE` to `$SCT_DIR`.
- So, if we were to clone into `.`, would the copy step inadvertently copy user files from, say, `Downloads/` to `$SCT_DIR`?

The one other benefit to using a tmpdir (as opposed to, say, `./spinalcordtoolbox`) is that it avoids the risk cloning to a non-empty folder, since the generated tmpdir will be unique on each repeated run of `./install_sct`.

install_sct
joshuacwnewton: Thank you for pointing this out! I definitely didn't understand `getopts` syntax as well as I thought I did.

From what I can tell, adding a `:` makes the argument mandatory? But, in most cases, we'll be leaning on the default value, rather than expecting the user to specify the value themselves. (The option, in this case, is mainly to provide support for running the install script standalone within CI workflows, so that we can `git clone` PR branches, as we do in the Windows.)

However, `getopts` [doesn't support optional arguments](https://stackoverflow.com/questions/11517139/optional-option-argument-with-getopts). There do seem to be workarounds. though?

- Write a function that provides this functionality: https://stackoverflow.com/a/57295993
- Use `getopt` (no `s`) instead: https://stackoverflow.com/a/52674277
    - Though, there are big cross-platform compatibility concerns here, so I think `getopt` is a no-go:
      > getopt vs getopts .. very different cross platform compliance
    
      > While I too find GNU getopt superior, it's not built in on BSD systems (like macOS) so if your goal is cross platform compatibility, go with getopts

We could also ditch the notion of an option entirely, and instead use an environment variable for this, given that this is only really relevant for our CI in particular? (In fact, I'm a little surprised I didn't do this in the Windows install script...)

install_sct
```diff
@@ -22,6 +22,7 @@
 #  -d   Prevent the (re)-installation of the data/ directory
 #  -b   Prevent the (re)-installation of the SCT binaries files
 #  -c   Prevent checks from being run to validate the installation
+#  -g   Specify git ref (branch, tag, commit SHA) to clone when 'install_sct' is run by itself
```
joshuacwnewton, line 25: From https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4049#discussion_r1117229028:

> Running the install script standalone within CI workflows

Reminder note to self to update the CI to use the standalone installation method, as a way of testing out the changes I've made.

joshuacwnewton: Ah, I was mistaken here! We don't actually use the _["standalone installation method, specifying a PR git ref"]_ installation in our CI anymore, because it used to break PRs from forks:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/33e134edf001670725d5fb36a06b8f37245d9580/.github/workflows/tests.yml#L387-L393

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3779

So... we don't actually need a `-g` option at all?

install_sct
joshuacwnewton: Scratch this -- We don't need a `-g` option at all. I was mistaken: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4049#discussion_r1118965144

install_sct
joshuacwnewton: Worryingly, this means we're not testing the "standalone installation" method at all in our CI... hmm...

install_sct
mguaypaq: I think "mandatory argument" means something different here: not that `-g` must always be specified, but rather that _its_ argument (the thing after `-g`, in this case the git ref we want to clone) has to be specified if `-g` is present. In our case, it would make no sense to say `-g` without giving a git ref, so it shouldn't matter to us.

joshuacwnewton: > I think "mandatory argument" means something different here: not that `-g` must always be specified, but rather that _its_ argument (the thing after `-g`, in this case the git ref we want to clone) has to be specified if `-g` is present.

You're right! My apologies... I had sworn I had come across a `getopts` example where an error was thrown when an argument wasn't passed, but perhaps I was reading some custom bit of bash that someone has written. My bad. :sweat:

mguaypaq (approved):

PR #4055, opened 2023-03-02
Attach `install_sct` as an asset during release creation CI workflow

### LINKED ISSUE #3545: SCT's current distribution method prevents us from checking download statistics

joshuacwnewton:
## Description

Since `install_sct` can now be run standalone (#4049), the next step is to adopt what we currently do for our Native Windows installer:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2850ac8aca9d8d199165caac5e84bc1d904c7bd8/.github/workflows/create-release.yml#L173-L185

(i.e. Update the install script to specify the stable release version, then attach the install script as a release asset, so that users don't have to download the "Source code"  archive, and so we can collect download statistics.)

## Linked issues

Fixes #3545.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -190,5 +190,7 @@ jobs:
         tag: ${{ github.event.inputs.milestone_title }}
         token: ${{ secrets.GITHUB_TOKEN }}
         bodyFile: ".github/workflows/release-body.md"
-        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,install_sct-${{ github.event.inputs.milestone_title }}_macos.sh"
+        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,\
+                    install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,\
+                    install_sct-${{ github.event.inputs.milestone_title }}_macos.sh"
         draft: true
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -175,6 +175,14 @@ jobs:
         cp install_sct.bat install_sct-${{ github.event.inputs.milestone_title }}_win.bat
         perl -pi -e 's/set git_ref=master/set git_ref=${{ github.event.inputs.milestone_title }}/' install_sct-${{ github.event.inputs.milestone_title }}_win.bat
 
+    - name: Create ${{ github.event.inputs.milestone_title }} Linux/macOS install scripts
+      # NB: `install_sct` works for both Linux and macOS platforms. So, here, we simply duplicate the script for both
+      # platforms (which will allow us to collect some rudimentary statistics + make our installation steps clearer).
+      run: |
+        cp install_sct install_sct-${{ github.event.inputs.milestone_title }}_linux.sh
+        perl -pi -e 's/SCT_GIT_REF="master"/SCT_GIT_REF="${{ github.event.inputs.milestone_title }}"/' install_sct-${{ github.event.inputs.milestone_title }}_linux.sh
+        cp install_sct-${{ github.event.inputs.milestone_title }}_linux.sh install_sct-${{ github.event.inputs.milestone_title }}_macos.sh
+
     - uses: ncipollo/release-action@v1
       name: Create release
       id: create_release
@@ -182,5 +190,5 @@ jobs:
         tag: ${{ github.event.inputs.milestone_title }}
         token: ${{ secrets.GITHUB_TOKEN }}
         bodyFile: ".github/workflows/release-body.md"
-        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat"
+        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,install_sct-${{ github.event.inputs.milestone_title }}_macos.sh"
         draft: true


.github/workflows/create-release.yml
```diff
@@ -175,12 +175,20 @@ jobs:
         cp install_sct.bat install_sct-${{ github.event.inputs.milestone_title }}_win.bat
         perl -pi -e 's/set git_ref=master/set git_ref=${{ github.event.inputs.milestone_title }}/' install_sct-${{ github.event.inputs.milestone_title }}_win.bat
 
+    - name: Create ${{ github.event.inputs.milestone_title }} Linux/macOS install scripts
+      # NB: `install_sct` works for both Linux and macOS platforms. So, here, we simply duplicate the script for both
+      # platforms (which will allow us to collect some rudimentary statistics + make our installation steps clearer).
+      run: |
+        cp install_sct install_sct-${{ github.event.inputs.milestone_title }}_linux.sh
+        perl -pi -e 's/SCT_GIT_REF="master"/SCT_GIT_REF="${{ github.event.inputs.milestone_title }}"/' install_sct-${{ github.event.inputs.milestone_title }}_linux.sh
```
joshuacwnewton, line 183: This command was tested using:

```console
(venv_sct) jn/3545-add-install_sct-as-a-release-asset|joshua@tadpole:~/repos/spinalcordtoolbox$ cp install_sct install_sct-6.0_linux.sh
(venv_sct) jn/3545-add-install_sct-as-a-release-asset|joshua@tadpole:~/repos/spinalcordtoolbox$ perl -pi -e 's/SCT_GIT_REF="master"/SCT_GIT_REF="6.0"/' install_sct-6.0_linux.sh
(venv_sct) jn/3545-add-install_sct-as-a-release-asset|joshua@tadpole:~/repos/spinalcordtoolbox$ grep SCT_GIT_REF install_sct-6.0_linux.sh
SCT_GIT_REF="6.0"
```

.github/workflows/create-release.yml
```diff
@@ -175,12 +175,20 @@ jobs:
...
+
     - uses: ncipollo/release-action@v1
       name: Create release
       id: create_release
       with:
         tag: ${{ github.event.inputs.milestone_title }}
         token: ${{ secrets.GITHUB_TOKEN }}
         bodyFile: ".github/workflows/release-body.md"
-        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat"
+        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,install_sct-${{ github.event.inputs.milestone_title }}_macos.sh"
```
joshuacwnewton, line 193: I wanted to split this line into multiple lines (for line length purposes) but I was having trouble wrapping my head around the intersection between:

1. The syntax for specifying multiple artifacts to `ncipollo/release-action`
    * The repo for this action states "This may be a single path or a comma delimited list of paths (or globs)", but provides no tangible examples of how a list should be expressed. (Do we use `[`/`]` brackets? Where do quotes go, if at all?)
    * Searching GitHub-wide for "ncipollo/release-action artifacts:" gave many examples, but going through a few of the 100+ pages, I couldn't find any that showed what "list of paths" looks like.
    * The closest I came was [this issue comment](https://github.com/ncipollo/release-action/issues/42#issuecomment-826150550), which offhandedly refers to the syntax `"string1,string2"` all within double-quotes, which is what I used here.
2. YAML multiline statements
    - Multiline YAML is so notoriously complex and error-prone that there's a whole site dedicated to it: https://yaml-multiline.info/
    - I figured that `>` would be what we would want, and so I tried [this syntax](https://github.com/joshuacwnewton/test-github-actions/commit/1853f59e2d2462c62c1c42ad3ff2f4f10b520f4c), but I got some [strange warnings](https://github.com/joshuacwnewton/test-github-actions/actions/runs/4307222314/jobs/7512024597#step:13:26), and the assets were not attached to the dummy 6.0 release.
    - I presume I maybe did something wrong in my "test github actions workflow", so I then got stuck in the weeds with the [PWDs of various GH Action steps](https://github.com/ncipollo/release-action/issues/42). Because this was rather time-consuming to test, I figured it would be safer to keep the string on a single line (at least for this first PR draft).

.github/workflows/create-release.yml
```diff
@@ -175,12 +175,20 @@ jobs:
         cp install_sct.bat install_sct-${{ github.event.inputs.milestone_title }}_win.bat
         perl -pi -e 's/set git_ref=master/set git_ref=${{ github.event.inputs.milestone_title }}/' install_sct-${{ github.event.inputs.milestone_title }}_win.bat
 
+    - name: Create ${{ github.event.inputs.milestone_title }} Linux/macOS install scripts
+      # NB: `install_sct` works for both Linux and macOS platforms. So, here, we simply duplicate the script for both
+      # platforms (which will allow us to collect some rudimentary statistics + make our installation steps clearer).
+      run: |
+        cp install_sct install_sct-${{ github.event.inputs.milestone_title }}_linux.sh
```
joshuacwnewton, line 182: When renaming the file, I chose to add the `.sh` file extension, largely for symmetry with the `install_sct-6.0_win.bat` script. (I've always found it a little weird that we chose to drop the file extension for `install_sct`?)

I wanted to point this out so that we maybe discuss/brainstorm the consequences of this change? (Were there benefits to excluding the extension in the past?)

mguaypaq: As I understand it, from sources like ["command names considered harmful"](https://www.talisman.org/~erlkonig/documents/commandname-extensions-considered-harmful/), it's typical Unix practice to not use file extensions for any executables that live on the `$PATH`, because unlike Windows, file extensions are not used to determine how to run the file (rather that's the job of the shebang line inside the file).

But! There's some nuance here: this is Unix-specific, and it's only for things on the `$PATH`. For an example of another case, the [Google style guide](https://google.github.io/styleguide/shellguide.html#s2.1-file-extensions) recommends using extensions for shell *libraries* (that are sourced from other shell scripts with `.` or `source`, so that they can declare functions and modify variables).

In our case, I would argue that:
- We're in a non-Unix-specific context (the release download page, right next to a Windows executable), so the extension is helpful to the reader.
- We're *not* on the `$PATH`, and after downloading the file, it might not even have executable permissions. So users might have to either `chmod +x install_sct` or invoke it through the shell, like `bash install_sct`.
- We might even want to use `.bash` instead of `.sh`, because our script probably fails if it's run with `sh install_sct` instead of `bash install_sct`.
- Probably for all these reasons, I think specifically for *installers* it's common to put the `.sh` extension.

.github/workflows/create-release.yml
mguaypaq, line 193: By my reading of the https://yaml-multiline.info/ site, maybe this would work? It would be a flow scalar, double-quoted, with newlines escaped so that they don't turn into spaces.
```suggestion
        artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,\
                    install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,\
                    install_sct-${{ github.event.inputs.milestone_title }}_macos.sh"
```
Looking at [the source of release-action](https://github.com/ncipollo/release-action/blob/d51f1872a4078bf85434f4b20c9bdbc4eb02f295/src/ArtifactGlobber.ts#L20-L22), I think the allowed syntax is that the list is split on both commas and newlines, and that leading whitespace is then trimmed for each path. But I'm not sure. I admit I'm a bit puzzled by the strange warnings you got.

mguaypaq (approved):

.github/workflows/create-release.yml
joshuacwnewton: > * We might even want to use `.bash` instead of `.sh`, because our script probably fails if it's run with `sh install_sct` instead of `bash install_sct`.

This is a really good point!

Though, from what I understand, `install_sct` makes a point to support non-bash shells (e.g. `zsh`), especially since macOS [defaults to `zsh`](https://www.theverge.com/2019/6/4/18651872/apple-macos-catalina-zsh-bash-shell-replacement-features) these days? And, I'm slightly concerned that a `.bash` extension may scare away/confuse macOS users who don't understand the nuances of shell types. 

`.sh` may be inaccurate, but it signals that it's a shell-agnostic script to some degree, I think?

.github/workflows/create-release.yml
joshuacwnewton: Delightful! Thank you for the suggestion. :hearts:

joshuacwnewton:
Hmm. The required status checks seem to not be triggering. I'm going to close and reopen this PR to try to re-trigger the workflows.

PR #4066, opened 2023-03-15
Refactor the `sct_compute_compression` script to make it easier to understand

joshuacwnewton:
## Description

This PR proposes some changes to the following currently-open PR:

- #4003. 

This commits in this PR are only **refactoring** changes (rearranging code, renaming functions/variables, etc.), My goal was to make PR #4003 easier to read and understand, so that it will be easier to add additional reviews to the PR in the future.

> (Also, by organizing the script into sections, it helped me understand the script a lot better myself! :smile:)

## Reviewing this PR

> **Note**: The diffs here are quite large, but this is because a lot of code was rearranged. Little to no logical changes were actually made here. To make things easier, please use the "Commits" view to look at each change individually. 

Also, of course, feel free to push back against or discuss any proposed changes if something doesn't feel right. :slightly_smiling_face: 

## Linked issues

Part of #3733.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -389,11 +389,9 @@ def main(argv: Sequence[str]):
     sex = arguments.sex
     age = arguments.age
     if age:
+        age.sort()
         if any(n < 0 for n in age):
             parser.error('Age range needs to be positive, {} was specified'.format(age))
-        # Put age range in order
-        else:
-            age.sort()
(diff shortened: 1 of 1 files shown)


sandrinebedard (commented):
This looks great! Just from the look of it there is not the check for the  `-i_PAM50` argument right?  This removes the option to just compute MSCC without normalization however. Hmm, I think I forgot to push my last changes on my local computer... I got stuck with a submission deadline and forgot... so sorry. I will add it tomorrow

PR #4068, opened 2023-03-16
Remove each instance of the `-igt` argument from SCT scripts

### LINKED ISSUE #3377: Argument `-igt` is not used and should be removed
7 scripts currently have an `-igt` argument:

Image

This was an `sct_testing`-specific argument, but its usage was commented out in #2315, so the argument is not being used.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e52c8d4d461d72e42366751677d06f71cf7bb207/scripts/sct_testing.py#L554-L561

#2315 also has a TODO to remove the argument, but this TODO must have slipped through the cracks. :sweat_smile: 

> TODO:
> 
>   * Removed flag `-igt` in functions, because ground truth data are now input directly into the testing framework (no need to expose this to the user).

joshuacwnewton:
## Description

Argument `-igt` is a relic from the old `sct_testing` days, where we maintained our own testing framework. The argument is no longer used anywhere. So, this PR removes every instance of `-igt` in SCT scripts.

I also checked for any instances of `arguments.igt`, `.igt`, etc. to look for usage in code, but there are none.

## Linked issues

Fixes #3377.


mguaypaq (approved):
In fact, `arguments.igt` doesn't seem to appear in the git history at all? Strange!

PR #4071, opened 2023-03-20
Replace `-setorient-data` with `-flip` and `-transpose` options

### LINKED ISSUE #4038: Replace `-setorient-data` with dedicated "flip" and "transpose" commands
> **Note**: I recently received a private message on the forum from a user who has expressed confusion over the `-setorient-data` command. I have previously helped this user in [this forum post](https://forum.spinalcordmri.org/t/how-to-use-sct-image-to-fix-the-orientation-of-the-header-and-data-array-setorient-data-vs-setorient/848/1). The private message plus the discussion in that forum post indicates that this feature is not very user-friendly or self-evident, hence this issue.

### Background 

NIfTI images contain two main parts: the "data array" and the "metadata header".

* The data array, by itself, cannot tell you anything about how the image is oriented relative to physical space. 
* Because of this, we need separate metadata to map the discrete axes (`[x, y, z]`) to physical axes (`[S/I, A/P, L/R]`). 
* For example, an image with a header that says "RPI" means that `x` maps to the L/R axis, and increasing `x` coordinates corresponds to increasing _towards_ the `R` direction _from_ the `L` direction. Likewise, `y -> A/P` and `z -> S/I`.)

### SCT context

Currently, the `sct_image` CLI script has two semi-related options for changing the orientation of an image:

* `-setorient`: Changes the header metadata to the specified orientation, then performs flips/transposes to the data array to keep the array in sync with the header.
* `-setorient-data`: Only performs flips/transposes to the data array.

In 99% of cases, `-setorient` should be used, because the array and header are typically already in sync, and we want to update them together to _keep_ them in sync.

(Notably, in such cases, using `-setorient-data` instead would be very dangerous, as it would cause a mismatch between the data array and the header orientation. **We include a strong warning on the `-setorient-data` option because of this.**)

For the other 1% of cases, `-setorient-data` is used to _fix mismatches between the data and the header_. For example, if the header metadata says one orientation, but the array is incorrectly oriented, then `-setorient-data` can be used to manipulate the data array by performing flips and transposes.

### Issue

There are two main issues here:

1. Because of the similarity in syntax between `-setorient` and `-setorient-data`, users may unwittingly choose the latter when they should be using the former, since it can be hard to distinguish between the two options. (This is dangerous!)
2. The actual usage of `-setorient-data` can be unintuitive and confusing. To use this option, you have to:
    1. Note the current orientation in the header metadata. (e.g. `RPI`)
    2. Decide what operations you want to perform on the data. (e.g. flip the x axis + transpose the x axis from `[x,y,z]` to `[y,z,x]`)
    3. Convert those operations into an orientation string _relative to the existing orientation string_. (e.g. `RPI` -> `LPI` performs a flip in the x-axis, then `LPI` -> `PIL` transposes the x axis, so you would enter `-setorient-data PIL` if your header says `RPI`.)
   
    In other words, usage is data-dependent; the command `-setorient-data PIL` will do _different things for different images_ depending on what the header of the image says.

### Proposed solution

Given the above issues, I would propose to remove the `-setorient-data` CLI option, and encourage users to modify the data array directly using dedicated `-flip` and `-transpose` options. That way, there is no need to spend the mental effort trying to convert the desired operations into data-specific orientation strings -- the operations can be expressed in explicit terms.

> (e.g. flip the x axis + transpose the x axis from `[x,y,z]` to `[y,z,x]`)

For the above example, the usage would be clear: `-flip x -transpose y,z,x` (or perhaps `-flip 0 -transpose 1,2,0` to match numpy convention). And, notably, these options would do the same thing every time, independent of the image's header metadata.

joshuacwnewton:
## Description

`-setorient-data` is essentially a way to perform flips and transposes on an image's data array. However, its usage requires the
user to specify an orientation string *relative to the orientation string already in the header*, which has historically been [confusing
to our users](https://forum.spinalcordmri.org/t/how-to-use-sct-image-to-fix-the-orientation-of-the-header-and-data-array-setorient-data-vs-setorient/848/1).

To make things clearer, we can instead provide flip and transpose functionality directly to our users, so that they explicitly know
what operations they are applying to their data.

> **Note**: I considered writing a test for these new options, however they are dead simple, as they call `numpy` functions directly in one-liners. The corresponding test would just be... calling those same `numpy` functions again, and seeing if things match? And that seems like a pretty useless test to me. But, maybe there's a smarter way to write a useful test. Please, share your thoughts!

## Linked issues

Fixes #4038.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_image.py
+++ b/spinalcordtoolbox/scripts/sct_image.py
@@ -24,7 +24,8 @@ from spinalcordtoolbox.image import (Image, concat_data, add_suffix, change_orie
                                      create_formatted_header_string, HEADER_FORMATS,
                                      stitch_images, generate_stitched_qc_images)
 from spinalcordtoolbox.reports.qc import generate_qc
-from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, display_viewer_syntax, ActionCreateFolder
+from spinalcordtoolbox.utils.shell import (SCTArgumentParser, Metavar, display_viewer_syntax, ActionCreateFolder,
+                                           list_type)
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, extract_fname, rmtree
 
@@ -167,6 +168,25 @@ def get_parser():
             'SLA', 'SLP', 'SRA', 'SRP', 'SAL', 'SAR', 'SPL', 'SPR',
             'ILA', 'ILP', 'IRA', 'IRP', 'IAL', 'IAR', 'IPL', 'IPR',
         ],
+        required=False)
+    orientation.add_argument(
+        '-flip',
+        help="Flip an axis of the image's data array. (This will not change the header orientation string.)\n"
+             "(WARNING: This option should only be used to fix the data array when it does not match the orientation "
+             "string in the header. We recommend that you investigate and understand where the mismatch originated "
+             "from in the first place before using this option.)\n"
+             "Example: For an image with 'RPI' in its header, `-flip x` will flip the LR axis of the data array.",
+        choices=['x', 'y', 'z', 't'],
+        required=False)
+    orientation.add_argument(
+        '-transpose',
+        help="Transpose the axes of the image's data array. (This will not change the header orientation string.)\n "
+             "(WARNING: This option should only be used to fix the data array when it does not match the orientation "
+             "string in the header. We recommend that you investigate and understand where the mismatch originated "
+             "from in the first place before using this option.)\n"
+             "Example: For a 3D image with 'RPI' in its header, `-transpose z,y,x` will swap the LR and SI axes of "
+             "the data array.'",
+        type=list_type(',', str),
         required=False)
 
     multi = parser.add_argument_group('MULTI-COMPONENT OPERATIONS ON ITK COMPOSITE WARPING FIELDS')
@@ -335,6 +355,23 @@ def main(argv: Sequence[str]):
     elif arguments.setorient is not None:
         printv(im_in.absolutepath)
         im_out = [change_orientation(im_in, arguments.setorient)]
+
+    elif arguments.flip is not None:
+        if dim_list.index(arguments.flip) >= len(im_in.data.shape):
+            parser.error(f"Cannot flip axis '{arguments.flip}' on an image with n_dim = {len(im_in.data.shape)}.")
+        im_out = im_in.copy()
+        im_out.data = np.flip(im_out.data, axis=dim_list.index(arguments.flip))
+        im_out = [im_out]
+
+    elif arguments.transpose is not None:
+        if len(arguments.transpose) != len(im_in.data.shape):
+            parser.error(f"Transpose argument '{','.join(arguments.transpose)}' must match number of image dimensions "
+                         f"in the input image ({len(im_in.data.shape)}).")
+        im_out = im_in.copy()
+        transpose_numerical = [dim_list.index(axis)  # Convert [x, y, z, t] to [0, 1, 2, 3]
+                               for axis in arguments.transpose]
+        im_out.data = np.transpose(im_out.data, axes=transpose_numerical)
+        im_out = [im_out]
 
     elif arguments.header is not None:
         header = im_in.header
--- a/spinalcordtoolbox/scripts/sct_image.py
+++ b/spinalcordtoolbox/scripts/sct_image.py
@@ -29,6 +29,8 @@ from spinalcordtoolbox.utils.shell import (SCTArgumentParser, Metavar, display_v
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, extract_fname, rmtree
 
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -169,23 +170,23 @@ def get_parser():
...
-            'SLA', 'SLP', 'SRA', 'SRP', 'SAL', 'SAR', 'SPL', 'SPR',
-            'ILA', 'ILP', 'IRA', 'IRP', 'IAL', 'IAR', 'IPL', 'IPR',
-        ],
+        '-flip',
+        help="Flip an axis of the image's data array. (This will not change the header orientation string.)\n"
+             "(WARNING: This option should only be used to fix the data array when it does not match the orientation "
+             "string in the header. We recommend that you investigate and understand where the mismatch originated "
+             "from in the first place before using this option.)\n"
+             "Example: For an image with 'RPI' in its header, `-flip x` will flip the LR axis of the data array.",
+        choices=['x', 'y', 'z', 't'],
```
mguaypaq, line 179: Could we make `dim_list` a module-level constant, and use it here? After all, the code later in the file assumes that all the choices appear in `dim_list`, so it would be a single source of truth.
```suggestion
        choices=dim_list,
```

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -355,8 +356,22 @@ def main(argv: Sequence[str]):
...
+        im_out = [im_out]
+
+    elif arguments.transpose is not None:
+        if len(arguments.transpose) != len(im_in.data.shape):
+            parser.error(f"Transpose argument '{','.join(arguments.transpose)}' must match number of image dimensions "
+                         f"in the input image ({len(im_in.data.shape)}).")
+        im_out = im_in.copy()
+        transpose_numerical = [dim_list.index(axis)  # Convert [x, y, z, t] to [0, 1, 2, 3]
+                               for axis in arguments.transpose]
+        im_out.data = np.transpose(im_out.data, axes=transpose_numerical)
```
mguaypaq, line 373: I'll note that numpy has decent error messages for some of the possible errors in the axis list:
```py3
ValueError: repeated axis in transpose
numpy.AxisError: axis 3 is out of bounds for array of dimension 2
ValueError: axes don't match array
```

mguaypaq (approved):
This looks fine, thanks! And I agree that there's not much point in writing automated tests for such thin wrappers around basic numpy functions, which numpy itself has tests for. As long as we try running each of them manually once, to make sure there are no problems with argument parsing, I'll be happy.

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: Just to clarify, is this a suggestion to allow numpy to handle the errors instead of trying to catch them ourselves?

```python
sct_image -i t2.nii.gz -transpose z,x,x -o repeated_axes.nii.gz
# Before/After (this case wasn't covered by my error message)
"ValueError: repeated axis in transpose"

sct_image -i t2.nii.gz -transpose z,y,x,t -o too_many_axes.nii.gz
# Before (ours)
"sct_image: error: Transpose argument 'z,y,x,t' must match number of image dimensions in the input image (3)."
# After (numpy)
"ValueError: axes don't match array"

sct_image -i t2.nii.gz -flip t -o 3d_doesnt_have_t_dim.nii.gz
# Before (ours)
"sct_image: error: Cannot flip axis 't' on an image with n_dim = 3."
# After (numpy)
"numpy.AxisError: axis 3 is out of bounds for array of dimension 3"
```

I agree that numpy basically covers the cases that we'd worry about, and that my messages aren't super necessary. The only issues I have are:

- The arguments are currently specified in terms of axes labels (x,y,z,t) rather than 0-based indexing (0,1,2,3), so while numpy's error messages are informative for devs, they're might be a bit confusing for users, due to the mention of e.g. "axis 3"
- Numpy's error messages are also a bit terse in general?
- They also spit out a bit of an ugly traceback, with `_wrapfunc` and `bound` functions in the stack trace. 

I'm fine with removing my added input checking, but am curious about what your thoughts are, too. :)

mguaypaq: No no, your existing messages are fine: they prevent the ugly/confusing traceback, and expressing things in terms of (x, y, z, t) is nice.

But they don't catch every possible user error, for example:
```
sct_image -i t2.nii.gz -transpose z,x,x -o repeated_axes.nii.gz
sct_image -i t2.nii,gz -transpose x,y,t -o 3d_with_no_t.nii.gz
```
I was just noting that, even in these non-covered cases, the user will get a reasonable error message (directly from numpy).

mguaypaq: If you just wanted to _check_ that the axes are valid for transposition, I think this bit of code would do it exactly:
```py3
sorted(arguments.transpose) == sorted(DIM_LIST[:len(im_in.data.shape)])
```
But if you want to output sensible messages for different kinds of errors (wrong number of axes, repeated axis, out-of-range axis, invalid letter for an axis), that would be a bunch of code.

I think the most common source of errors is most likely to be wrong number of axes, which your code handles, so that should be fine.

Issue #3395, opened 2021-05-18
Improve `sct_label_vertebrae` by skipping straightening (using orthogonal planes)


jcohenadad:
### Context

Currently, in `sct_label_vertebrae` there is a necessary step which involves straightening the spinal cord (ie: bringing it from the natural curved space to a "straight" spinal cord). This step is required by:
- the current template-based disc detection and;
- the option `sct_label_vertebrae -discfile`, which uses a disc label and produces a labeled segmentation.

The problem is that the straightening is **very** long, and is also prone to errors and inaccuracies (notably related to a parameter defining the smoothness of the centerline).

### Suggestion

Implement a function that computes the plane orthogonal to the spinal cord centerline and that passes through a label. That way, straightening can be avoided.

<img width="300" alt="image" src="https://user-images.githubusercontent.com/39456460/226650261-12e19e62-146b-4d94-b9f3-8ed7cc66c6b3.png">

Note that this approach will also become useful once the [new deep learning-based disc detection method](https://github.com/ivadomed/ivadomed/pull/787) is available in SCT.

joshuacwnewton:
Migrated to https://github.com/neuropoly/idea-projects/issues/7.

jcohenadad:
@NadiaBlostein Could you please point us to your Venus-SCT code that does compute the plane orthogonal to the spinal cord centerline? This will be useful for @valosekj who will work on this issue. 

An alternative approach, instead of computing the orthogonal plane to the centerline, would be to compute the shortest distance between each disc label and a point on the centerline. Maybe this could be implemented by iterating across all points on the centerline, computing the distance, and keeping the shortest distance as the candidate. @sandrinebedard @benjamindeleener do you have any other suggestions?

NadiaBlostein:
[Here](https://github.com/sct-pipeline/venus-integration) it is!

valosekj:
I discovered that the functionality performing labeling from discs (without applying the straightening) is available within the [labels.labelize_from_discs](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/47ab308361c6e9d98f4a6c1d58e98d03363f5814/spinalcordtoolbox/labels.py#L219-L241) function.

It can be run by:

```
sct_label_utils.py -i sub-001_seg.nii.gz -disc labels.nii.gz -o sub-001_seg_labeled.nii.gz
```

[...]

PR #4080, opened 2023-03-27
Add new `project_centerline` option in `sct_label_utils` to project an image on the spinal cord centerline 

NathanMolinier:
#### GitHub

#### PR contents

## Description

This PR enables to project discs coordinates on the spinal cord centerline without using any straightening steps. 

Concerning the type of projection used, this was addressed in this issue #3395: for each point, the projection was computed by minimizing its distance with the spinal cord centerline.

## Usage
Users can now provide a `DISC_FILE` and a spinal cord `SEGMENTATION` using `sct_label_utils` with the new flag 
`-project-discs` to get a new `OUT_FILE` with the discs coordinates projected onto the spinal cord centerline.
```bash
sct_label_utils -i SEGMENTATION -project-discs DISC_FILE -o OUT_FILE
```
#### **Example**:
```bash
sct_label_utils -i sub-amu01_T2w_seg.nii.gz -project-discs sub-amu01_T2w_labels-disc-manual.nii.gz -o projected_discs.nii.gz
```

<details><summary>Output</summary>

Capture d’écran 2023-03-27 à 16 50 30  |  Red dots correspond to discs coordinates in the original input `DISC_FILE` while the green dots correspond to new projected coordinates on the spinal cord centerline.
:-------------------------:|:-------------------------:

Note: Missing dots are actually located on another sagittal plane.
</details>

## Linked issues

Fixes : #3395 and [#2](https://github.com/spinalcordtoolbox/disc-labeling-hourglass/issues/2)


NathanMolinier:
I believe additional testing steps are needed

spinalcordtoolbox/labels.py
```diff
@@ -464,3 +465,69 @@ def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
             logger.warning(f"Label {label} not found in input image!")
 
     return out
+
+
+def project_discs(img: Image, ref: Image, num_disc=True) -> Image:
```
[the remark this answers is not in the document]
jcohenadad: I guess this would be a sensible thing to do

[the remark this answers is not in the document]
NathanMolinier: I tested the script in a case where the input labels are ones and it worked without needing to specify this parameter. I is maybe not so necessary

spinalcordtoolbox/labels.py
```diff
@@ -464,3 +465,69 @@ def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
...
+        if num_disc:
+            disc = point[-1]
+            point = point[:3]
+
+        # Define distance function between the point and the centerline
+        def distance(t):
+            return np.linalg.norm(np.array(point) - np.array(centerline[int(t)]))
+
+        # Minimize distance function to find the parameter t of the projection
+        result = minimize_scalar(distance, bounds=(0, centerline.shape[0]-1), method='bounded')
```
mguaypaq, line 512: I don't think `minimize_scalar` is the right approach: it's designed for finding the minimum of a continuous function on a real interval, but what you really have is a discontinuous function on a discrete interval (really, a finite list of points). It would be better to just iterate over the centerline (as you wrote in the documentation, actually!).

You can either write out the double loop manually (outer loop over the labelled points, inner loop over the centerline points), or there are a few ways to replace the inner loop with a call to [the builtin `min` function](https://docs.python.org/3/library/functions.html#min) (by taking the minimum of a list of (distance, index) pairs, or by using the `key=` parameter of `min`).

spinalcordtoolbox/labels.py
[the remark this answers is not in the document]
mguaypaq: Yes, when you find out that you didn't use an argument after all, it's a good idea to remove it, and replace the variable with its default value in the code. After that, it's often possible to simplify the code a lot!

spinalcordtoolbox/labels.py
```diff
@@ -464,3 +465,69 @@ def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
...
     return out
+
+
+def project_discs(img: Image, ref: Image, num_disc=True) -> Image:
+    """
+    Project discs coordinates on the spinal cord centerline. This projection is obtained by iterating along
+    the centerline to identify the shortest distance with each referenced coordinates.
+    Typically, user inputs a segmentation image, and labels with disks position, and this function computes
+    the identification to the closest coordinates of each labels on the centerline.
+    Labels are assumed to be non-zero and incremented from top to bottom
```
mguaypaq, line 476: I don't think you actually assume that the labels increment from top to bottom.

mguaypaq (changes requested):
Thanks for this prototype! I like that the changes to `sct_label_utils.py` are so small, and that you included type annotations on the new `project_discs` function.

Here's a few comments to start with. I expect you'll change a good fraction of the code to address them, so I'll save the lower level comments for another round.

Also, as a general comment, I think this function isn't really specific to discs? It should work for any sort of labels, whether they come from discs or other anatomical features, as long as they're reasonably close to the centerline. Maybe you could rename it to `-project-centerline`?

NathanMolinier:
> Also, as a general comment, I think this function isn't really specific to discs? It should work for any sort of labels, whether they come from discs or other anatomical features, as long as they're reasonably close to the centerline. Maybe you could rename it to `-project-centerline`?

I agree this function could also be available for other features. See commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4080/commits/431c8958180530b3087cb115652187d2da14e140

spinalcordtoolbox/labels.py
NathanMolinier: I generally avoid creating nested loops to improve computation time but if you feel it is the right approach, I can change that.

spinalcordtoolbox/labels.py
NathanMolinier: I just forgot to erase this line when copying. See https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4080/commits/59e758e3e04d1217ce52ab8f24645776ef316214

spinalcordtoolbox/labels.py
```diff
@@ -464,3 +464,79 @@ def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
...
+    return out
+
+
+def project_point_on_line(point, line):
+    """
+    Project the input point on the referenced line by finding the minimal distance
+
+    :param point: coordinates of a point and its value: point = numpy.array([x y z value])
+    :param line: list of points coordinates which compose the line
+    :returns: closest coordinate to the referenced point on the line: projected_point = numpy.array([X Y Z value])
```
mguaypaq, line 529: Is there a reason to weave `value` through this function? It seems like it would be simpler to just pass `np.array([x, y, z])` and return `np.array([X, Y, Z])`.

(The caller already has to manipulate the input to change it from a `Coordinate` instance to a numpy array, so it might as well do the extra processing before calling `project_point_on_line`. Plus, the caller then has to separate out the `value` from the result anyway.)

spinalcordtoolbox/labels.py
```diff
@@ -464,3 +464,79 @@ def remove_other_labels_from_image(img: Image, labels: Sequence[int]) -> Image:
...
 
     return out
+
+
+def project_centerline(img: Image, ref: Image) -> Image:
+    """
+    Project an input image on the spinal cord centerline. This projection is obtained by iterating along
+    the centerline to identify the shortest distance with each referenced coordinates.
+    Typically, user inputs a segmentation image, and labels with disks position, and this function computes
+    the identification to the closest coordinates of each labels on the centerline.
```
mguaypaq, line 474: I like that you're explaining the typical use case here.

mguaypaq (commented):
I think this is nearly done! My comments are just about small things to change. I'm marking this review as "comments" instead of "approve" just to have a final look before merging. You can click "re-request review" next to my name to let me know when you're ready.

testing/cli/test_cli_sct_label_utils.py
```diff
@@ -48,3 +48,39 @@ def test_create_seg_mid(tmp_path):
...
+         '-p', 'coord,20x15',
+         '-f', 'box'
+         ])
+
+    # Create an image with a point at the coordinates x=1, y=1, and z=1 with the value=1
+    ref = str(tmp_path/'t2_seg_labeled.nii.gz')
+    sct_label_utils.main([
+        '-i', img,
+        '-o', ref,
+        '-create', '1,1,1,1'
```
mguaypaq, line 74: It would be nice to use different values for the coordinates, for example `1,2,3,4`, so that the test has a chance to detect more possible errors, like swapped coordinates or something.

mguaypaq (approved):
The code looks good, and thanks for adding a test! I just have a small suggested improvement for the test, after that you can go ahead and merge this into master.

testing/cli/test_cli_sct_label_utils.py
NathanMolinier: I added 2 other points

PR #4089, opened 2023-04-11
Make matplotlib less verbose

### LINKED ISSUE #4086: Getting a lot of "findfont: score("
### Description

I'm getting a LOT of this output:
~~~
findfont: score(FontEntry(fname='/System/Library/Fonts/NotoSansArmenian.ttc', name='Noto Sans Armenian', style='normal', variant='normal', weight=900, stretch='normal', size='scalable')) = 10.525
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/NotoSansPhagsPa-Regular.ttf', name='Noto Sans PhagsPa', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Noteworthy.ttc', name='Noteworthy', style='normal', variant='normal', weight=300, stretch='normal', size='scalable')) = 10.145
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Savoye LET.ttc', name='Savoye LET', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Brush Script.ttf', name='Brush Script MT', style='italic', variant='normal', weight=400, stretch='normal', size='scalable')) = 11.05
findfont: score(FontEntry(fname='/System/Library/Fonts/SFCompactItalic.ttf', name='.SF Compact', style='italic', variant='normal', weight=1000, stretch='normal', size='scalable')) = 11.62
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Telugu Sangam MN.ttc', name='Telugu Sangam MN', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/Library/Fonts/Arial Unicode.ttf', name='Arial Unicode MS', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/ITFDevanagari.ttc', name='ITF Devanagari', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Ayuthaya.ttf', name='Ayuthaya', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/NotoSansLepcha-Regular.ttf', name='Noto Sans Lepcha', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Thonburi.ttc', name='Thonburi', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Trattatello.ttf', name='Trattatello', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/NotoSansRejang-Regular.ttf', name='Noto Sans Rejang', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: score(FontEntry(fname='/System/Library/Fonts/Supplemental/Farisi.ttf', name='Farisi', style='normal', variant='normal', weight=400, stretch='normal', size='scalable')) = 10.05
findfont: Matching sans\-serif:style=normal:variant=normal:weight=normal:stretch=normal:size=12.0 to DejaVu Sans ('/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.8/site-packages/matplotlib/mpl-data/fonts/ttf/DejaVuSans.ttf') with score of 0.050000.
Create the straight space and the safe zone
~~~

using 7ead83200d7ad9ee5e0d112e77a1d7b894add738

on macos

mguaypaq:
Otherwise it can easily fill the terminal scrollback buffer with irrelevant messages about fonts.

(This PR is basically following the recommendations in [matplotlib's `set_loglevel` documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.set_loglevel.html).)

Fixes #4086.

### DIFF
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -141,6 +141,9 @@ def set_loglevel(verbose):
     # Set the logging level globally, but only when scripts are directly invoked (e.g. from the command line)
     if caller_module_name == "__main__":
         logging.root.setLevel(getattr(logging, log_level))
+        # matplotlib is particularly chatty, so keep it at the default level
+        # see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4086
+        logging.getLogger('matplotlib').setLevel(logging.INFO)
     else:
         # NB: Nothing will be set if we're calling a CLI script in-code, i.e. <sct_cli_script>.main(). This keeps
         # the loglevel changes from leaking: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3341
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -143,7 +143,7 @@ def set_loglevel(verbose):
         logging.root.setLevel(getattr(logging, log_level))
         # matplotlib is particularly chatty, so keep it at the default level
         # see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4086
-        logging.getLogger('matplotlib').setLevel(logging.INFO)
+        logging.getLogger('matplotlib').setLevel(logging.WARNING)
     else:
         # NB: Nothing will be set if we're calling a CLI script in-code, i.e. <sct_cli_script>.main(). This keeps
         # the loglevel changes from leaking: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3341


spinalcordtoolbox/utils/sys.py
```diff
@@ -141,6 +141,9 @@ def set_loglevel(verbose):
     # Set the logging level globally, but only when scripts are directly invoked (e.g. from the command line)
     if caller_module_name == "__main__":
         logging.root.setLevel(getattr(logging, log_level))
+        # matplotlib is particularly chatty, so keep it at the default level
+        # see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4086
+        logging.getLogger('matplotlib').setLevel(logging.INFO)
```
joshuacwnewton, line 146: Just so I'm understanding here: Is the core issue because:

-  _We_ set the root loglevel to `debug` (via `-v 2`), and thus accidentally increased the logging coming from matlplotlib? 
- Or, is it that Matplotlib's `findfont: score(` messages are `INFO` messages, meaning that they sneak through regardless of what `-v` is set to in a script.

If it's the latter, we may want to use `logging.WARNING` instead, But if it's the former, then this LGTM.

joshuacwnewton (approved):
Pre-approving because I think I already have a sense of what the answer to my question will be, and if so, this LGTM. :)

spinalcordtoolbox/utils/sys.py
mguaypaq: Good suggestion, I'll change it to `logging.WARNING`.

The specific `findfont: score` messages are `logging.DEBUG` level, since they're produced [here](https://github.com/matplotlib/matplotlib/blob/b3bd929cf07ea35479fded8f739126ccc39edd6d/lib/matplotlib/font_manager.py#L1390). This is in the `matplotlib.font_manager` module, which [uses its `__name__` for its logger](https://github.com/matplotlib/matplotlib/blob/b3bd929cf07ea35479fded8f739126ccc39edd6d/lib/matplotlib/font_manager.py#L47).

My understanding is that in the normal situation, as per [the defaults](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) (that is, the root logger is at `logging.WARNING` and all other loggers are at `logging.NOTSET`):
1. The message gets created with a `logging.DEBUG` level in the `matplotlib.font_manager` logger.
2. The `matplotlib.font_manager` logger is at `logging.NOTSET`, so the message percolates up.
3. The `matplotlib` logger is at `logging.NOTSET`, so it percolates up again.
4. The root logger is at `logging.WARNING`, so the message gets discarded.

When we set the root logger to `logging.DEBUG`, that step 4 suddenly lets the message through to the console. If we want to get only the "default" logging with matplotlib, then setting `matplotlib` to `logging.WARNING` would make make the message get filtered out in step 3, correctly.

PR #4090, opened 2023-04-13
Add support to `-create-viewer` for creating labels with duplicate values (e.g. `1,1,1`)

### LINKED ISSUE #3984: Extend `-create-viewer` to allow users to create compression disc labels (all `1` values)

joshuacwnewton:
## Description

This PR allows users to run commands such as:

```
sct_label_utils -i t2.nii.gz -create-viewer 1,1,1
```

To place labels with duplicate values:

image

~~One caveat: the `□ Label 1` checkboxes don't really work at all. (To fix this, I would have to dig into the PyQt Controller/Dialog code to see how checkboxes are currently handled, which would take quite a bit of time.)~~ EDIT: Fixed!

Still, the basic functionality (placing labels, undoing them) works OK. So, I wanted to get a PR started, so that @valosekj or @sandrinebedard could try it out in the context of their research. 

## Linked issues

Fixes #3984.


valosekj:
@joshuacwnewton Thank you for starting to work on it! Tested for a sagittal image, and working fine!

For the disc labeling of axial images, it is better to have both sagittal and axial images overlaid at the same time. For example, in FSLeyes, see [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3984#issuecomment-1479957357). I am not sure how difficult it would be to implement such functionality to the SCT viewer, though.

sandrinebedard:
> > For the disc labeling of axial images, it is better to have both sagittal and axial images overlaid at the same time. For example, in FSLeyes, see [#3984 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3984#issuecomment-1479957357). I am not sure how difficult it would be to implement such functionality to the SCT viewer, though.
> 
> I think it would be quite tricky. As long as we are able to change sagittal slices (eg: go along right-left to find the maximum compression point), I think it should be OK to do the labeling on the sagittal scan only.

However, we want to label the axial image only, not the sagittal image since we are using the lablels after to compute MSCC, which is from the axial image (segmentation) such as I did for the dcm-zurich dataset.

jcohenadad:
> However, we want to label the axial image only, not the sagittal image since we are using the lablels after to compute MSCC, which is from the axial image (segmentation) such as I did for the dcm-zurich dataset.

Ah! Gotcha. Hum... that seems like a very niche feature. The SCT get centerline viewer does provide the sag and axial view (from a single input), but not an overlay between two images (in this case, an axial volume and a sagittal volume).

sandrinebedard:
If we want to label the axial image, we need to see the axial view.
Here is the procedure I did for the dcm-zurich 

- Open the axial and sagittal images. 
- Looked following the clinical data provided
- Located the compression in the sagittal view
- In the axial view, toggled to the max compressed slice
- Placed label at the center of the slice of the axial image

In the case that we have 3D images, it would be fine to label on the sagittal view.

sandrinebedard:
> > However, we want to label the axial image only, not the sagittal image since we are using the lablels after to compute MSCC, which is from the axial image (segmentation) such as I did for the dcm-zurich dataset.
> 
> Ah! Gotcha. Hum... that seems like a very niche feature. The SCT get centerline viewer does provide the sag and axial view (from a single input), but not an overlay between two images (in this case, an axial volume and a sagittal volume).

Yes, I used `manual_correction.py` script from https://github.com/spinalcordtoolbox/manual-correction/blob/main/manual_correction.py where it openned both images and created the label file, so it may be fine to keep this feature as it is only in the sagittal view.

valosekj:
> Yes, I used `manual_correction.py` script from https://github.com/spinalcordtoolbox/manual-correction/blob/main/manual_correction.py where it openned both images and created the label file, so it may be fine to keep this feature as it is only in the sagittal view.

Just for info; the labeling procedure has been added to `manual-correction`'s wiki, see [here](https://github.com/spinalcordtoolbox/manual-correction/wiki#manual-labeling-of-spinal-cord-compression). Thank you Sandrine for adding it!

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -374,8 +375,19 @@ def launch_sagittal_viewer(img: Image, labels: Sequence[int], msg: str, previous
 
     params.subtitle = msg
 
+    # NB: `num_points` is used as the limit for how many points *should* be in the image.
+    # If this number is exceeded, `SagittalController` will throw a `TooManyPoints` warning.
```
mguaypaq, line 379: In my testing, I'm not seeing the associated message appearing in the GUI:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bb89863a06036d6762af1d18f7899f128687a9c6/spinalcordtoolbox/gui/base.py#L391-L392

But I don't think Exceptions use this supposed `message` member? Indeed:
```py3
Python 3.8.15 (default, Nov 24 2022, 15:19:38) 
[GCC 11.2.0] :: Anaconda, Inc. on linux
>>> from spinalcordtoolbox.gui.base import TooManyPointsWarning
>>> str(TooManyPointsWarning())
''
>>> raise TooManyPointsWarning
spinalcordtoolbox.gui.base.TooManyPointsWarning
>>> raise TooManyPointsWarning("an actual message")
spinalcordtoolbox.gui.base.TooManyPointsWarning: an actual message
```

Plus, I don't think it should subclass `StopIteration`, that's a very specific exception type used for Python's control flow handling.

In any case, the exception _does_ serve the purpose of preventing extra labels being created once the limit is reached.

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -374,8 +375,19 @@ def launch_sagittal_viewer(img: Image, labels: Sequence[int], msg: str, previous
...
+    # To determine what `num_points should be, we consider two cases:
+    # - Case 1: `previous_points` is set: This means we're adding OR correcting existing labels.
+    #            Presumably, we're dealing with all unique labels. So, we just take the length
+    #            of the union of both the specified labels and the existing (previous) labels.
     if previous_points is not None:
         params.message_warn = 'Please select the label you want to add \nor correct in the list below before clicking \non the image'
+        previous_point_values = [coord.value for coord in previous_points]
+        params.num_points = len(set(labels).union(set(previous_point_values)))
```
mguaypaq, line 387: If there are no repeated values in `labels` and `previous_point_values`, I think this constraint is a no-op? And if there _are_ repeated values, I think the limit might make it impossible to put all the labels in. For example, if:
```py3
previous_point_values = [1, 1, 1, 1]
labels = [1, 2, 3]
```
then it would be impossible put _any_ labels in.

Maybe just leave `num_points` unset in this case?

spinalcordtoolbox/scripts/sct_label_utils.py
mguaypaq: In contrast, I do see the following message if I create a label and try to undo too many times:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bb89863a06036d6762af1d18f7899f128687a9c6/spinalcordtoolbox/gui/base.py#L364

(And if I try to undo _before_ creating any labels, the program just crashes, hmmm.)

mguaypaq (commented):
Thanks for the prototype! It's already enough to get the task done, as long as the user is careful. I know it's not finished, but here's some comments for now anyway.

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -374,8 +375,19 @@ def launch_sagittal_viewer(img: Image, labels: Sequence[int], msg: str, previous
 
     params.subtitle = msg
 
+    # NB: `num_points` is used as the limit for how many points *should* be in the image.
+    # If this number is exceeded, `SagittalController` will throw a `TooManyPoints` warning.
+    # To determine what `num_points should be, we consider two cases:
+    # - Case 1: `previous_points` is set: This means we're adding OR correcting existing labels.
+    #            Presumably, we're dealing with all unique labels. So, we just take the length
```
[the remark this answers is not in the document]
joshuacwnewton: My thought process was:

- Trying to update duplicate labels is _hard_ given our current codebase. It was making my head hurt. (Exactly as you've identified.)
- Adding duplicate labels is a use-case that only came up in the context of the (internal to NeuroPoly) compression project. It isn't something users have asked for. We only really need this feature to provide an SCT-only tutorial, rather than asking users to download FSLeyes. So, I made the (hopeful, optimistic, less-work) presumption that users aren't going to try to correct duplicate labels, and only follow what the tutorial shows. (This could very well be a faulty assumption!)

So, I very much took the easy way out here. I'm very thankful you've pointed this out, though.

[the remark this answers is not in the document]
mguaypaq: That's ok! Thanks for sharing your thought process. And like I said, this seems to already be working for the internal use-case, so it's not urgent to get a solid UI.

spinalcordtoolbox/utils/shell.py
```diff
@@ -373,17 +373,19 @@ def parse_num_list(str_num):
...
         if m is not None:
             a = int(m.group("first"))
             b = int(m.group("last"))
-            list_num += [x for x in range(a, b + 1) if x not in list_num]
+            list_num += [x for x in range(a, b + 1)]
             continue
         raise ValueError("unexpected group element {} group spec {}".format(element, str_num))
 
+    if not allow_duplicates:
+        list_num = list(set(list_num))
```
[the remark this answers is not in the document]
joshuacwnewton: > In general, `dict` is ordered (by insertion order), but `set` is unordered, so `list(set(list_num))` may not be in the same order as `list_num`, even if it has no duplicates.

Good catch! Using a `set` indeed [does not preserve the original order](https://stackoverflow.com/questions/480214/how-do-i-remove-duplicates-from-a-list-while-preserving-order). And, `sct_compute_snr` is enough to make me want to preserve the existing behavior.

As you point out a `dict` is ordered, so we could use the following solution instead ([source](https://stackoverflow.com/a/17016257)):

```python
>>> items = [1, 2, 0, 1, 3, 2]
>>> list(dict.fromkeys(items))  # Or [*dict.fromkeys(items)] if you prefer
[1, 2, 0, 3]
```

```suggestion
        list_num = list(dict.fromkeys(list_num))  # dict --> remove duplicates and preserve order
```

spinalcordtoolbox/scripts/sct_label_utils.py
joshuacwnewton: > (And if I try to undo before creating any labels, the program just crashes, hmmm.)

This is strange. Supposedly, the case you describe is handled by the `else` in [`.undo()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e7c82c359bfb0a24cc852fa4124701bd1b31d659/spinalcordtoolbox/gui/base.py#L355-L364):

The traceback here is:

```
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/gui/sagittal.py", line 85, in on_undo
    self.labels.label = self._controller.label
AttributeError: 'SagittalController' object has no attribute 'label'
```

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e7c82c359bfb0a24cc852fa4124701bd1b31d659/spinalcordtoolbox/gui/sagittal.py#L77-L81

I can confirm that this happens for both `-create-viewer 1,1,1` and `-create-viewer 1,2,3`, so this seems to be a bug outside the scope of the changes in this PR. I'll create a new issue.

spinalcordtoolbox/utils/shell.py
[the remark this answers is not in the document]
mguaypaq: That's quite elegant! I actually didn't know about `dict.fromkeys`, TIL.

joshuacwnewton:
> When the time comes to write the tutorial, we can further test this PR, because we will figuring out how we should instruct users to create compression labels.

Given that the 6.0 release is coming up, I've marked this PR as ready for review, as it will need to be included in 6.0 (even if the tutorial itself is merged post-6.0).

spinalcordtoolbox/gui/base.py
joshuacwnewton, line 1: I'm going to take a look at the checkbox-related code to see if I can understand why they aren't working, and what I might be able to do to fix it. I'll update with replies to this thread when I have a better idea of how it works.

joshuacwnewton: Here is what I've gleamed so far:

- The label checkboxes are handled by [`gui/widgets.py :: VertebraeWidget`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/widgets.py#L23-L95).
- The checkboxes are stored [in the `self._check_boxes` dict](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/widgets.py#L28), with the label _value_ being [used as the key](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/widgets.py#L50).
    - Normally, this is fine, because all of the values are different. So, checkboxes can be successfully accessed from the dict [via their value](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/widgets.py#L81) (e.g. `checkbox for label 7` == `self.check_boxes[7]`).
    - However, in the case of labels `1,1,1`, this means that only 1 checkbox actually gets stored in `self.check_box`, since they all get assigned to `self.check_boxes[1]`. So, trying to access the checkbox for any of the three `1` labels will return the single stored checkbox.
- I think the key to fixing this is to turn `VertebraeWidget.check_boxes` into a list instead, and rely on using an index to access each checkbox instead. (e.g. for `1,1,1`, we would access them using `self.check_boxes[0,1,2]`).
    - This is kind of partially implemented? At several places in the `gui` code, [indices are used](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/sagittal.py#L71-L73) instead of values.
    - The trouble is, the "active label" is passed along and [stored within several GUI layers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/spinalcordtoolbox/gui/sagittal.py#L51). So, it may be tricky to ensure that changes to the checkboxes in `gui/widgets.py` (e.g. dict --> list) don't end up causing unintended side-effects to other parts of the GUI.
    - (For example, if we switch to primarily using indices to access labels, then the "active label index" will need to be stored in various places in addition to the "active label value".)

joshuacwnewton: I think I have a working prototype in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090/commits/256fe0edcd96211034d6b8d1882b8d965acdac5c, but I'm _very_ suspicious about the side-effects of this change. I think that commit needs a heightened level of scrutiny during review?

joshuacwnewton: From my extended commit message:

> (For example, does changing BaseController.undo() affect the CenterlineController (in addition to the SagittalController? Do CenterlineController's `self.points` need to be updated to include an index, too?)

Aha, [it does](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/5458575735/jobs/9933793320?pr=4090#step:5:465)! I didn't realize we had a test for `CenterlineController` -- I'll update it accordingly.

valosekj: Thanks, @joshuacwnewton, for digging into the code! 💯 
I tested your latest commit (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090/commits/1088b87f8ac99f6a4d68c1d9d1505746f4d45182). One finding from my side is that when I clicked on any checkbox to redo the label, the GUI crashed.

<details><summary>Error message:</summary>

```console
$ sct_label_utils -i t2.nii.gz -create-viewer 1,1,1

--
Spinal Cord Toolbox (git-jn/3984-allow-create-viewer-to-place-duplicate-labels-1088b87f8ac99f6a4d68c1d9d1505746f4d45182)

sct_label_utils -i t2.nii.gz -create-viewer 1,1,1
--

  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/gui/widgets.py", line 59, in on_select_label
    self.label = label.label
AttributeError: 'QCheckBox' object has no attribute 'label'
```

</details>

<details><summary>GIF:</summary>

Kapture 2023-07-05 at 07 19 32

</details>

I would also like to point out the previous discussion that we usually use two images at a time (axial and sagittal) for compression labeling; see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#issuecomment-1509036229 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#issuecomment-1511608687.

joshuacwnewton: > One finding from my side is that when I clicked on any checkbox to redo the label, the GUI crashed.

Thank you so much for catching this! I'll fix this right away.

joshuacwnewton: > I would also like to point out the previous discussion that we usually use two images at a time (axial and sagittal) for compression labeling; see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#issuecomment-1509036229 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#issuecomment-1511608687.

If I understand correctly, are you saying that `-create-viewer 1,1,1`, as it is now, is basically useless to you and @sandrinebedard? Or, in other words, it is inferior enough compared to the FSLeyes labeling procedure that you are unlikely to ever use it?

I agree. In fact, I kind of wasn't even expecting this feature to be seriously used by you or Sandrine? My thinking was that we would write the tutorial so that it would look something like this:

> You can add compression labels to your images using `sct_label_utils -create-viewer`:
>
> > [Insert screenshots and demonstration, showing how to use the [left-right sagittal slice toggling](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#issuecomment-1508448933)]
>
> Alternatively, if you need more precise labeling, you can use FSLeyes to create these compression labels instead. FSLeyes allows you to examine an image in 3D to precisely determine which slices have compression using both axial and sagittal views. To add labels using FSLeyes, you can use the following procedure:
>
> > [Insert steps from the manual correction Wiki]

In other words, the SCT feature would only exist as a "simple" (admittedly inferior) way of labeling the images. The purpose of the feature would be to allow users to try out the entire pipeline using only pure SCT. The labels wouldn't be perfect, but they would be close enough to allow users to then try out the steps that come after, without needing to install another program.

Then, if users find that they are seriously adopting the pipeline for their analysis, they can then switch to the more "advanced" method for labeling, and we can offload that responsibility to the FSLeyes, whose GUI is much more rigorously designed and tested.

(My thinking was that adding the "basic" version to SCT would be simple enough, since this PR is just adapting the existing `-create-viewer` feature. But, for more complex usage, we would have to do much more GUI design work, and at that point it might just be better to leave the responsibility to FSLeyes instead.)

joshuacwnewton: I believe I have fixed the behaviors with re-selecting existing checkboxes in my latest 2 commits (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090/commits/54abf899f8ed9a0867a85454fed544203a6defe1, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090/commits/28a2d58c5383696f4568eb6e26197e80e633c16f)

valosekj: > I believe I have fixed the behaviors with re-selecting existing checkboxes in my latest 2 commits ([54abf89](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/54abf899f8ed9a0867a85454fed544203a6defe1), [28a2d58](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/28a2d58c5383696f4568eb6e26197e80e633c16f))

Yes, it is fixed now! Thanks!

valosekj: > If I understand correctly, are you saying that -create-viewer 1,1,1, as it is now, is basically useless to you and @sandrinebedard, since it does not provide an axial view? i.e. It is inferior enough compared to the FSLeyes labeling procedure that you are unlikely to ever use it?

I am not discarding any ideas, but yes, for me and Sandrine, workflow with FSLeyes is better.

joshuacwnewton: Thank you for your input! I can add an action item to tomorrow's agenda where we can brainstorm around what the tutorial might look like.

If we decide that the tutorial should be written specifically with FSLeyes in mind, then we could consider discarding this PR entirely, assuming there are no other benefits to having duplicate labels.

joshuacwnewton:
Given that in the most recent SCT meeting, we decided that we would focus on FSLeyes in the tutorial, rather than highlighting a pure SCT labeling method, I think I'm leaning towards closing this PR.

I think that this feature adds quite a bit of extra complexity to juggling the labels (see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#discussion_r1169131641). So, if we don't _need_ to support duplicate labels... then I'd rather not make things messier than they need to be? We know `-create-viewer` works well enough at is, and I don't want to risk breaking something. :sweat_smile: 

Besides, if we end up needing to add duplicate labels using `-create-viewer` in the future, this PR will always be available for inspiration, especially to highlight the refactoring that needs to be done in order to make the checkboxes work for duplicate labels (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4090#discussion_r1252374947).

mguaypaq:
Since the GUI code is currently fragile and convoluted, another viable workaround could be to create distinct labels, then post-process them to make them equal (either directly with nibabel, or thresholding with `sct_maths`, etc.)

joshuacwnewton:
Very clever idea! I've updated my closing comment for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3984#issuecomment-1625601151 to mention this.

PR #4091, opened 2023-04-17
Remove all 70,000+ warnings from test suite

### LINKED ISSUE #3126: Migrate from np.matrix -> np.array
CI is reporting a bunch of error messages like

```
  /mnt/d/a/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/centerline/nurbs.py:629: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    R = np.matrix(R)
```

This is used in many places, but the migration should be straightforward, though tedious.

### LINKED ISSUE #3216: The size of the label is wrong when using -method map

### LINKED ISSUE #3854: Deal with nibabel's deprecation of `get_header` and `get_data`
`sct_create_mask` generates a couple of deprecation warnings, coming from how we use `nibabel`:
```
.../spinalcordtoolbox/spinalcordtoolbox/scripts/sct_create_mask.py:227: DeprecationWarning: get_header method is deprecated.
Please use the ``img.header`` property instead.

* deprecated from version: 2.1
* Will raise <class 'nibabel.deprecator.ExpiredDeprecationError'> as of version: 4.0
  hdr = centerline.get_header()  # get header
/home/mguaypaq/neuropoly/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_create_mask.py:230: DeprecationWarning: get_data() is deprecated in favor of get_fdata(), which has a more predictable return type. To obtain get_data() behavior going forward, use numpy.asanyarray(img.dataobj).

* deprecated from version: 3.0
* Will raise <class 'nibabel.deprecator.ExpiredDeprecationError'> as of version: 5.0
  data_centerline = centerline.get_data()  # get centerline
```
We should search through the code and update our usage, as explained in the deprecation messages.

_Originally posted by @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3831#discussion_r930599374_

joshuacwnewton:
## Description

This PR addresses each `DeprecationWarning`, `FutureWarning`, `RuntimeWarning`, etc. caught by SCT's test suite.

- Most of these changes are very straightforward (simple import changes, etc.)
- 10dbc05ab4af7f9d46f16e803b77c6d01a32ae48 fixes [#3854](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3854) by implementing the change suggested in both the `DeprecationWarning` and the dev's comments [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3854#issuecomment-1313976579).
- dee9e9fa90dedf44d6c7819d835996dd3abdd5ef fixes [#3126](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3126). In addition to the `matrix` -> `array` swap, it was also necessary to change from `*` to `@` for matmul, and to change 0xN arrays to 1xN to match the expected vector shapes.
- For any fixes that _weren't_ straightforward to implement, I opted to use `filterwarnings` to ignore them, while also providing extensive documentation on where and how these warnings are occurring (c9a244fd96863e857443674ffb75e3ed4acb3c2d). 
    - We should still address these, but perhaps in a follow-up PR where more investigation can be done. (It should be easy to comment out an `ignore`, then investigate a specific type of warning to fix it.)

Some notes:
- In theory, this _should_ allow us to turn on the setting that upgrades warnings into errors (that way we can quickly catch future deprecations as they're released, as well as when we write new tests). However, I found that neither `filterwarnings = error` nor `-Werror` seem to play nicely with the `ignore` lines,  despite this being explicitly mentioned in [this blog post](https://til.simonwillison.net/pytest/treat-warnings-as-errors).
- 70,000 _sounds_ like a lot of warnings, but repeated warnings stack. If you have an oft-traversed bit of code, a single warning can multiply to become 100 or 1000 warnings.
   - Plus, ~53,000 of those warnings were coming purely from `sct_compute_hausdorff_distance`. (How???)  
   - Seriously, revert commit 60c2fe576947cad393d9fd3d10c49670aaeb21b0 then run the hausdorff test. It's wild!
    image

## Linked issues

Fixes #3126.
Fixes #3854.


spinalcordtoolbox/centerline/nurbs.py
```diff
@@ -642,27 +642,27 @@ def reconstructGlobalApproximation(self, P_x, P_y, P_z, p, n, w):
             for k in range(m - 1):
                 somme += w[k] * self.evaluateN(Nik[i], ubar[k], u) * self.Tk(k, P_x, Nik, ubar[k], u) / denU[k]
             Tx.append(somme)
-        Tx = np.matrix(Tx)
+        Tx = np.expand_dims(np.array(Tx), 0)
```
mguaypaq, line 645: I was tempted to say, since we only ever want the transpose `Tx.T`, why not do `expand_dims(..., 1)` instead? But I think this code is tricky enough that it's better to replicate its workings as exactly as possible, like you did here.

setup.cfg
mguaypaq, line 1: Thanks for providing such good context for the warning filters in your comments!

spinalcordtoolbox/deepseg_/gm.py
```diff
@@ -182,7 +182,7 @@ def segment_volume(ninput_volume, model_name,
     model_abs_path = gmseg_model_challenge.get_file_path(model_path)
 
     # Padding/cropping data to match the input dimensions of the model
-    volume_data = ninput_volume.get_data()
+    volume_data = np.asanyarray(ninput_volume.dataobj)
```
mguaypaq, line 185: I'm torn on the replacement using `np.asanyarray(im.dataobj)`, because in theory I agree with the deprecation message that using `get_fdata` to get a consistent, predictable data type would be better. But it's true that this way replicates the old behaviour more closely, so it's less risky.

In practice, I guess this refactoring is fine because it's safe, and we can always grep the source code later for uses of `dataobj` and audit them more deeply.

spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
```diff
@@ -91,7 +91,7 @@ def transitions(self, neighbours):
         """
         # now = time.time()
         n = neighbours + neighbours[0:1]      # P2, P3, ... , P8, P9, P2
-        s = np.sum((n1, n2) == (0, 1) for n1, n2 in zip(n, n[1:]))  # (P2,P3), (P3,P4), ... , (P8,P9), (P9,P2)
+        s = np.sum(np.fromiter([(n1, n2) == (0, 1) for n1, n2 in zip(n, n[1:])], dtype=bool))  # (P2,P3), (P3,P4), ... , (P8,P9), (P9,P2)
```
mguaypaq, line 94: I think using Python's builtin `sum` (which treats `True` as 1 and `False` as 0) is the simpler fix here:
```suggestion
        s = sum((n1, n2) == (0, 1) for n1, n2 in zip(n, n[1:]))  # (P2,P3), (P3,P4), ... , (P8,P9), (P9,P2)
```
(Besides, adding _both_ `np.fromiter` and square brackets is overkill.)

mguaypaq (approved):
Thanks for this massive fix! :heart: :tada:

spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
joshuacwnewton: I very much agree with the sum change! 

But, a note about `np.fromiter` + square brackets: I tried to get rid of the square brackets at first, but ran into:

image

```
E     File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py", line 94
E       s = np.sum(np.fromiter((n1, n2) == (0, 1) for n1, n2 in zip(n, n[1:]), dtype=bool))  # (P2,P3), (P3,P4), ... , (P8,P9), (P9,P2)
E                              ^
E   SyntaxError: Generator expression must be parenthesized
```

This was news to me! :stuck_out_tongue_closed_eyes:

spinalcordtoolbox/process_seg.py
```diff
@@ -129,7 +129,7 @@ def compute_shape(segmentation, angle_correction=True, param_centerline=None, ve
     metrics = {}
     for key, value in shape_properties.items():
         # Making sure all entries added to metrics have results
-        if not value == []:
+        if (isinstance(value, list) and len(value) > 0) or (isinstance(value, np.ndarray) and value.size > 0):
             metrics[key] = Metric(data=np.array(value), label=key)
```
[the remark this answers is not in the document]
joshuacwnewton: Thank you for the suggestion! I was puzzling over this a little, but I like your approach much more.

spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
mguaypaq: Ah, that's true! But still, parenthesizing with `[...]` turns it into a bona fide list, whereas parenthesizing with `(...)` as the `SyntaxError` suggests keeps it as a generator.

joshuacwnewton: Aha!! Thank you, I didn't realize. 

I presumed that using `(...)` would turn it into a tuple (i.e. there wasn't a significant difference between `[...]` and `(...)`). But, TIL that [there is no "tuple comprehension"](https://stackoverflow.com/q/16940293), and that `(...)` indeed returns a generator instead.

PR #4092, opened 2023-04-19
Rename FSLeyes script from `sct_plugin.py` to `sct_fsleyes_script.py`

### LINKED ISSUE #3419: Consider renaming `sct_plugin.py` to something that is less misleading
FSLeyes has "Load plugin" and "Run script" options. However, `sct_plugin.py` is not actually a plugin, so it's not compatible with "Load plugin". Instead, users must choose "Run script".

This is definitely unintuitive, and has caused problems on the forum, e.g. https://forum.spinalcordmri.org/t/problem-with-fsleyes/715

It might be clearer if the file was named, e.g., `sct_fsleyes_script.py`.

Related: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2594#issuecomment-639067039

joshuacwnewton:
## Description

SCT has a script called `sct_plugin.py` that allows users to run certain SCT commands inside of FSLeyes. To activate this script, they must go to the "File" menu, then select "Run script". However, the file menu also contains a "Load plugin" option:

image

The fact that our "script" is called a "plugin" is confusing to users, leading some to choose the wrong menu option, hence this PR.

> **Note**: I'm very open to discussion on the name. We could call it `sct_script.py` instead, or something else entirely.

## Linked issues

Fixes #3419.


jcohenadad (approved):
perfect! `sct_fsleyes_script` is exactly how I would have called it myself

PR #4094, opened 2023-04-20
Use `sct_progress_bar` for `algo=columnwise` and `register2d` slicewise loops

### LINKED ISSUE #2059: Use `sct_progress_bar` for `algo=columnwise`
looks nicer...

----

Relevant code:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5bd7c7b6393bf4e29c6d69c97e018f1f399cc46a/spinalcordtoolbox/registration/algorithms.py#L988-L991

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5bd7c7b6393bf4e29c6d69c97e018f1f399cc46a/spinalcordtoolbox/utils/sys.py#L341-L343

joshuacwnewton:
## Description

In `algorithms.py`, doing a search for "`for iz in`"  or  "`range(nz)`" etc. reveals 4 instances where we're looping over `z` slices. In 2/4 cases, we use `sct_progress_bar`. But, in the last two cases (`algo=columnwise`, `register2d`) we're still using home-grown logging solutions:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5bd7c7b6393bf4e29c6d69c97e018f1f399cc46a/spinalcordtoolbox/registration/algorithms.py#L988-L991

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5bd7c7b6393bf4e29c6d69c97e018f1f399cc46a/spinalcordtoolbox/registration/algorithms.py#L1260-L1263

So, this simple PR updates these loops to use `sct_progress_bar` (while also adding a newline to preserve the existing look of logging.)

Before:

image

After: 

image

## Linked issues

Fixes #2059.


spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1261,9 +1261,9 @@ def register2d(fname_src, fname_dest, fname_mask='', fname_warp='warp_forward.ni
         list_warp_inv = []
 
     # loop across slices
-    for i in range(nz):
+    print()  # Add newline between last log message and the progress bar logging
+    for i in sct_progress_bar(range(0, nz), unit='iter', unit_scale=False, desc="Registering slice", ncols=100):
```
joshuacwnewton, line 1265: One note: For `register2d`, because we're calling an `ANTs` binary on every loop, we print out a very verbose message between each loop:

image

This happens both before and after these changes:

image

We could consider silencing these messages with `verbose=0`, which would instead get us this:

image

Much nicer! But, I wanted to bring this up before actually making the change.

mguaypaq: I think it's a good idea, but it would be nice to preserve debugging output if `verbose == 2`. So I added some commits to this PR to do just that.

mguaypaq (approved):
Looks good, thanks! Since I added some commits, I'll let you do the merging so you can review them first.

spinalcordtoolbox/registration/algorithms.py
joshuacwnewton: Thank you for cleaning up the verbose code! However, I think we need to set `verbose` in one extra place to solve this issue.

In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4094/commits/05a0165362bc59c3d78731c6d83036edb7cad30d, `verbose` is set in the `cmd` list (which sets the verbosity value that gets passed to the ANTs binary itself.) However, `run_proc` has its own separate `verbose` parameter that needs to be set to suppress the blue messages about which command has been run.

I've added this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4094/commits/722bf1cae126e7e0ed2aeb1c9e0305622e4a1637. If it looks good to you, feel free to merge this PR. :)

PR #4095, opened 2023-04-20
Replace `algo=rigid` with `algo=slicereg` for mt0-mt1 registration

### LINKED ISSUE #1804: Use `algo=slicereg` instead of `algo=rigid` in `batch_processing.sh` for mt0-mt1 reg
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1b5c49c3904dffe63546adbadc47759a53fbc8ca/batch_processing.sh#L163-L166

joshuacwnewton:
## Description

This PR changes the algorithm used in `batch_processing.sh` for mt0-mt1 registration to follow @jcohenadad's recommendations from #1804.

This is what we're already doing [in the tutorials](https://spinalcordtoolbox.com/user_section/tutorials/multimodal-registration/mtr-computation/coregistering-mt0-mt1.html). So, this is just unifying `batch_processing.sh` with the script in the tutorial datatset. (Related: #3508)

NB: We also remove `slicewise=1` here because that parameter only applies to non-slicereg ANTs algorithms. Slicereg, by default, is a slicewise registration technique, so `slicewise` has no effect.

### NB: Updating cached `batch_processing.sh` results

Because this slightly changes the MTR results, the `batch_processing.sh` test will fail. This is to be expected. I will add another commit that updates the cached results once I generate the `.csv` file on my local computer.

## Linked issues

Fixes #1804.


jcohenadad (approved):
> I will add another commit that updates the cached results once I generate the .csv file on my local computer.

I believe this is already done, right?

thank you for this PR!

mguaypaq (approved):
This looks fine, thanks!

(I'm a bit surprised to see branch information and a local dev path in the cached results, but I guess it doesn't hurt since we don't test those values, and using a non-redacted file is closer to reality.)

PR #4102, opened 2023-04-24
Add sagittal mosaic QC and use it for `sct_deepseg_lesion` QC report

### LINKED ISSUE #3803: Add QC reporting to `sct_deepseg_lesion`
Additional context given by @jcohenadad:

> note that this particular QC is quite tricky (although not impossible)-- because lesion seg could happen on sag or ax images, and the QC only shows one orientation (so far)-- so which one to choose is a question-- also, we'll need to input the cord seg in the QC in order to crop around the cord (we already do that for the cord seg qc), but then display the lesion seg

Specifically, I believe this is referring to the fact that `sct_deepseg_lesion` can be run on images of multiple orientations:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2e5a817d831f9d8a439b52fca630df29cbf34e69/spinalcordtoolbox/scripts/sct_deepseg_lesion.py#L37-L45

valosekj:
#### GitHub

#### PR contents

## Description

The PR adds QC for lesions.

Checklist based on https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3803#issuecomment-1507041090:

> - what orientation to show? 
>   - the native orientation of the image
>   - if sagittal: crop it around the cord (via the seg) to magnify the ROI. Show all slices from left to right (and display them next to each other). This would be a **NEW** feature.
>  - [x] if axial: show ALL slices as done for SC seg (feature already exists).
> - Input segmentation required?
>   - yes: otherwise, some axial images would be HUGE (eg 512x512) and small lesions would not be easily viewable on the QC page. 
>    - There is the option to crop around the lesion, but if false neg, then nothing to crop (and lesion will be missed)

So far, I have implemented QC for an axial plane. Example:

Kapture 2023-04-24 at 11 15 18

Command:

```console
sct_qc -i dcm-zurich-lesions/sub-05/anat/sub-05_acq-ax_T2w.nii.gz -p sct_deepseg_lesion -plane Axial -d dcm-zurich-lesions/derivatives/labels/sub-05/anat/sub-05_acq-ax_T2w_label-lesion.nii.gz -s dcm-zurich-lesions/derivatives/labels/sub-05/anat/sub-05_acq-ax_T2w_label-SC_mask-manual.nii.gz -qc ./
```

## Linked issues
Resolves #3803


spinalcordtoolbox/reports/slice.py
```diff
@@ -444,3 +456,44 @@ def get_center(self, img_idx=-1):
...
+        Mosaic images are cropped based on the bounding box of the spinal cord segmentation.
+
+        :param nb_column: number of mosaic columns
+        :param dilation: a tuple containing (x,y,z) integers for dilating the bounding box used for cropping.
+        :return: tuple of numpy.ndarray containing the mosaics of each slice pixels
+        """
+        # 0. Use the segmentation image to initialize the image cropper
+        self._image_seg = self._images[-1]
+        cropper = ImageCropper()
+        cropper.get_bbox_from_mask(self._image_seg)
```
joshuacwnewton, line 472: By using the segmentation image to get the bounding box, and cropping the images based on this bounding box, I believe I now have a working sagittal QC:

image

(The lesion segmentation toggle works exactly as expected, too!)

spinalcordtoolbox/reports/slice.py
```diff
@@ -444,3 +456,44 @@ def get_center(self, img_idx=-1):
...
+        # 1c. Calculate number of rows to display.
+        nb_slices = cropper.bbox.zmax - cropper.bbox.zmin + 1 + (2 * dilation[2])  # SAL -> LR axis provides nb_slices
+        nb_row = math.ceil(nb_slices // nb_column) + 1
+        # 1d. Compute the matrix size of the overall mosaic image
+        matrix_sz = (int(size_h * nb_row), int(size_w * nb_column))
+
+        # 2. Use the previously-defined segmentation-based cropper to crop the images to the necessary size
+        matrices = list()
+        for image in self._images:
+            image_cropped = cropper.crop(img_in=image, dilate=dilation)
```
joshuacwnewton, line 490: As mentioned in commit 8ce10aa8f980a2ffc5c9d3f1e94affff2763e00a, using the `ImageCropper` class doesn't really mesh well with the `Axial` approach of implementing a `self.crop` method.

So, I'm opening this thread just to discuss how we should design this approach: Should we make an effort to keep the cropping methods consistent between all `Slice` subclasses?

spinalcordtoolbox/reports/slice.py
joshuacwnewton: The command I used to generate this image was exactly the same command in the PR description, just with `-plane Sagittal` instead of `-plane Axial`:

```
-i dcm-zurich-lesions/sub-05/anat/sub-05_acq-ax_T2w.nii.gz -p sct_deepseg_lesion -plane Sagittal -d dcm-zurich-lesions/derivatives/labels/sub-05/anat/sub-05_acq-ax_T2w_label-lesion.nii.gz -s dcm-zurich-lesions/derivatives/labels/sub-05/anat/sub-05_acq-ax_T2w_label-SC_mask-manual.nii.gz -qc ./dcm-zurich-lesions-qc
```

spinalcordtoolbox/reports/slice.py
```diff
@@ -444,3 +456,44 @@ def get_center(self, img_idx=-1):
         size_y = self.axial_dim(image)
         size_x = self.coronal_dim(image)
         return np.ones(dim) * size_x / 2, np.ones(dim) * size_y / 2
+
+    def mosaic(self, nb_column=0, dilation=(0, 0, 0)):
```
joshuacwnewton, line 460: Oh, I've also added this `dilation` parameter which we can use to dilate the bounding box, just in case we want to capture voxels outside of the SC segmentation bounding box, too.

The only issue here is that the `mosaic` method signatures don't match the base `Slice` class anymore. Is there a way to define the base `mosaic` abstract method to accommodate for different signatures?

spinalcordtoolbox/reports/slice.py
valosekj: This is amazing! Just tested and working fine!
Just one small thing, we could add `A`, `P`, etc. labels to the first sagittal image. Similarly, as is done for axial QC: 
image

joshuacwnewton: I would love to add these labels! Looking into it, though, the current implementation makes this really tricky. :sweat:

For Axial mosaics, the `_add_orientation_label` method assumes a square size of 30x30, and so it _hardcodes the label locations_ (yikes!):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a058fc2937cdf9a36a555e79af4420dc52f9e3a7/spinalcordtoolbox/reports/qc.py#L423-L426

To be able to generalize this to Sagittal images of varying shapes, we would need to have access to the bounding box dimensions. But... we lose that context once we enter the `_add_orientation_label` method, which is called way after generating the mosaic image.

So, I think in order to properly position the orientation labels, we would need to move the orientation label step inside of the `mosaic` method, which would require a bigger refactor. :sweat:

spinalcordtoolbox/reports/slice.py
mguaypaq: Is `self.crop` even really part of the public API for `Slice`? Also, it's a `@staticmethod`, so it could very well be a standalone function rather than a method on `self`. Anyway, I don't see much benefit to trying for consistency here.

spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -20,18 +21,19 @@ def get_parser():
         epilog='Examples:\n'
                'sct_qc -i t2.nii.gz -s t2_seg.nii.gz -p sct_deepseg_sc\n'
                'sct_qc -i t2.nii.gz -s t2_seg_labeled.nii.gz -p sct_label_vertebrae\n'
-               'sct_qc -i t2.nii.gz -s t2_seg.nii.gz -p sct_deepseg_sc -qc-dataset mydata -qc-subject sub-45'
+               'sct_qc -i t2.nii.gz -s t2_seg.nii.gz -p sct_deepseg_sc -qc-dataset mydata -qc-subject sub-45\n'
+               'sct_qc -i t2.nii.gz -s t2_seg.nii.gz -d t2_lesion.nii.gz -p sct_deepseg_lesion -plane Axial'
```
mguaypaq, line 25: @valosekj Thanks for adding an example of the new parameters!

spinalcordtoolbox/reports/slice.py
```diff
@@ -158,10 +159,10 @@ def add_slice(matrix, i, column, size, patch):
         :param patch: patch to insert
         :return: matrix
         """
-        start_col = (i % column) * size * 2
+        start_col = (i % column) * patch.shape[1]
         end_col = start_col + patch.shape[1]
 
-        start_row = int(i / column) * size * 2
+        start_row = int(i / column) * patch.shape[0]
```
mguaypaq, line 165: Unlike `math.ceil` used elsewhere, I think this _should_ use floor division:
```suggestion
        start_row = (i // column) * patch.shape[0]
```

mguaypaq (commented):
Here's a first round of comments, sorry I didn't have time to finish the review today.

@joshuacwnewton Thanks for your annotations on the `Sagittal.mosaic` method! They were helpful in navigating the code, and seeing how your changes fit in.

Also I find `ImageCropper` to be somewhat strange as a class. I almost think it would be better named as `BoundingBoxGetter`, because the current name suggests that it's supposed to be a reusable object, but it looks more like a builder pattern to me (hence the `img_in` attribute), and it seems to exist mostly to support a GUI bounding box picker. Maybe worth discussing and sharing notes on how we both view it before refactoring.

spinalcordtoolbox/reports/qc.py
```diff
@@ -845,6 +845,20 @@ def qcslice_layout(x): return x.single()
...
+            # Then, the input image (fname_in1) is overlaid by the lesion (fname_in2).
+            qcslice_type = qcslice.Axial([Image(fname_in1), Image(fname_in2), Image(fname_seg)])
+            qcslice_operations = [QcImage.listed_seg]
+            def qcslice_layout(x): return x.mosaic()[:2]
+        elif plane == 'Sagittal':
+            # Note, spinal cord segmentation (fname_seg) is used to crop the input image.
+            # Then, the input image (fname_in1) is overlaid by the lesion (fname_in2).
+            qcslice_type = qcslice.Sagittal([Image(fname_in1), Image(fname_in2), Image(fname_seg)])
+            qcslice_operations = [QcImage.listed_seg]
+            def qcslice_layout(x): return x.mosaic()[:2]
```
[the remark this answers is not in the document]
joshuacwnewton: Nitpicky, inconsequential design question: How terrible would it be to use `eval` here? 

```python
valid_planes = ['Axial', 'Sagittal']
if plane not in valid_planes:
    raise ValueError(f"Invalid plane '{plane}'. Valid choices are {valid_planes}.")
SliceSubtype = eval(f"qcslice.{plane}")
qcslice_type = SliceSubtype([Image(fname_in1), Image(fname_in2), Image(fname_seg)])
```

I'm always hesitant to use `eval`, but given that we're limiting the input to only two valid options, it feels somewhat harmless to use here? But I'm perfectly happy being explicit, too.

spinalcordtoolbox/reports/slice.py
```diff
@@ -444,3 +456,44 @@ def get_center(self, img_idx=-1):
...
+        :return: tuple of numpy.ndarray containing the mosaics of each slice pixels
+        """
+        # 0. Use the segmentation image to initialize the image cropper
+        self._image_seg = self._images[-1]
+        cropper = ImageCropper()
+        cropper.get_bbox_from_mask(self._image_seg)
+
+        # 1. Compute the sizes of the patches, as well as the overall image
+        # 1a. Compute width and height of mosaic squares. (This is assumed to be a square for Axial slices.)
+        size_h = cropper.bbox.xmax - cropper.bbox.xmin + 1 + (2 * dilation[0])  # SAL -> SI axis provides height
```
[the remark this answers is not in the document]
joshuacwnewton: Ahhh, I see your point here.

- `ImageCropper.crop()` limits the dilation from going outside of the image dimensions:
   https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f0a96439dbc822136b42653e803b5f89cfa3b3ed/spinalcordtoolbox/cropping.py#L48-L53
- But, `size_h`/`size_w` don't account for this limiting, so the computed dimensions can be larger than the cropped images
- And, trickily, `dilation` is only applied during the actual cropping, which makes it hard to "precompute" the correct size.

It seems like to get the correct size, we would need to perform dilation _before_ cropping, then fetch the size as some sort of `@property` of the bounding box.  But, that has its own caveats, because making `dilation` a method instead of a crop parameter means you can no longer vary the dilation across multiple crops. (Assuming this pattern is even used/wanted?)

I agree with your mention re: architecture. This messiness feels very tied to the points you raised in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4102#pullrequestreview-1427320288 regarding the structure of `ImageCropper`/`BoundingBox`.

spinalcordtoolbox/reports/qc.py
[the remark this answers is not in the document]
mguaypaq: That's an interesting thought. I'm very hesitant to use `eval` at all, but the limited input makes it less alarming. Even better would be to use [`getattr(qcslice, plane)`](https://docs.python.org/3/library/functions.html#getattr).

But as requested by @jcohenadad, the `plane` value should be lower-cased, whereas the class names are capitalized, so I don't think it'll work. (And for the record, I think doing the case modification programmatically would bring us over the line of "too much magic", so no `SliceSubtype = getattr(qcslice, plane.capitalize())`.)

spinalcordtoolbox/reports/slice.py
mguaypaq: Several things:
- The first technical answer is that Python itself doesn't care about the method signature, and neither does the `@abc.abstractmethod` decorator (it only checks whether _something_ called `mosaic` is defined in the derived class before allowing instances to be created). But that's not a very satisfying answer.
- A second answer, which isn't actually applicable here, is that if you want a fully general method signature in Python, you can do this:

  ```py3
  class Foo:
      def bar(self, *args, **kwargs):
          ...
  ```
- A third answer, which is more satisfying and applicable here, is given by [Mypy's rules for deciding which methods have compatible type signatures](https://mypy.readthedocs.io/en/stable/class_basics.html#overriding-statically-typed-methods). Basically, they implement the idea that every valid method call to the parent class's method has to still be a valid method call to the child class's method (and not necessarily the reverse). In particular, this means that:
  - Positional arguments that appear in the parent should appear in the child, with the same names and in the same order. Arguments that are optional in the parent (that is, arguments with a default value) should also be optional in the child. But mandatory arguments in the parent can be optional in the child, and the child can have more optional arguments.
  - Keyword-only arguments in the parent should be accepted in the child.
  - For each argument accepted in the parent, its type should be a subtype of what the child accepts for that argument.
  - Going in the other direction, the _return_ type of the child should be a subtype of what the parent's return type.

  (Aside: this is why the "fully general" method signature from the second answer above isn't applicable. A child method which is "fully general" is compatible with _any_ parent method, but not the other way around. In fact, a parent method which is "fully general" is maximally restrictive on the signature of a child method that wants to override it.)
- More concretely, just looking at the three method signatures here, we could make them compatible by putting only the common elements in the parent signature, since the other elements are optional, like so:
  ```py3
  class Slice:
      def mosaic(self, nb_column): pass

  class Axial(Slice):
      def mosaic(self, nb_column=0, size=15, return_center=False): pass

  class Sagittal(Slice):
      def mosaic(self, nb_column=0, dilation=(0, 0, 0)): pass
  ```
- But, looking at how this method is actually called in our codebase, I'm only seeing two kinds of call:
  - `mosaic()`, with no arguments at all; and
  - `mosaic(return_center=True)`, which uses a named argument, and only gets called on instances of the `Axial` subclass.

  So, possibly we should simplify the signatures to:
  ```py3
  class Slice:
      def mosaic(self): pass

  class Axial(Slice):
      def mosaic(self, return_center=False): pass

  class Sagittal(Slice):
      def mosaic(self): pass
  ```

  (One of the calls to `mosaic(return_center=True)` is done by the `Slice.mosaics_through_time` method, which should probably be moved to `Axial.mosaics_through_time`, unless you want to add `return_center` functionality to `Sagittal.mosaic` as well.)

spinalcordtoolbox/reports/slice.py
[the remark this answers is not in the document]
mguaypaq: I think the simplest short-term fix here is to remove the optional argument `dilation=(0,0,0)` from `Sagittal.mosaic` for now, sadly. We could probably re-add it after a refactoring of the `ImageCropper` code to make it less corner-casey.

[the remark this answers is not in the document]
joshuacwnewton: My main reason for adding the dilation feature was just in case we wanted to crop outside the spinal cord segmentation for lesion detection purposes. (Admittedly I don't understand spinal cord lesions well, but I was wondering if maybe it would be possible for a lesion to exist outside of the detected spinal cord segmentation, and so we would want some buffer to examine the surrounding areas.)

But, I don't know if this is relevant enough to make this a necessity? So, I would want to ask @valosekj, since I imagine he has more of an understanding of lesion segmentation than I do, and can try the QC on more images.

spinalcordtoolbox/reports/slice.py
joshuacwnewton: Thank you for such a deep and thorough answer!

> * But, looking at how this method is actually called in our codebase, I'm only seeing two kinds of call:

This is a very good point. `nb_column`, `size`, and `dilation` are definitely speculative features? They assume that we may one day want to tweak how the axial/sagittal grids are displayed.

But, as you point out, so far we haven't needed to change these values, and thus they've gone unused. 

So, I would be OK with a follow-up PR that removes these extraneous arguments. (I suggest a separate PR as that way we have a clear place to view the precise changes, if/when we ever need to re-add these features).

joshuacwnewton: > (One of the calls to mosaic(return_center=True) is done by the Slice.mosaics_through_time method, which should probably be moved to Axial.mosaics_through_time, unless you want to add return_center functionality to Sagittal.mosaic as well.)

I absolutely agree here. `return_center` is definitely only relevant for axial slices.

spinalcordtoolbox/reports/slice.py
mguaypaq: Ok, after trying to get the information about label placement from `Sagittal.mosaic` to `QcImage._add_orientation_label`, I think adding orientation labels is a complicated enough change that we should do it in a followup PR, possibly after a cleanup PR as well.

mguaypaq (approved):
I think once the existing review discussions are resolved, this should be good to go, so I'm going to mark this PR as approved. Thanks again!

spinalcordtoolbox/reports/slice.py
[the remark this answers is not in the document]
valosekj: >  I was wondering if maybe it would be possible for a lesion to exist outside of the detected spinal cord segmentation, and so we would want some buffer to examine the surrounding areas.)

Great point! From the anatomical point of view, both multiple sclerosis (MS) and spinal cord injury (SCI) lesions should be presented within the spinal cord. But, as you correctly assume, the spinal cord segmentation algorithm (`sct_deepseg_sc`) sometimes fails at the lesion site (due to rapid change of intensities in the lesions relative to the spinal cord). In other words, the lesion is not included within the spinal cord segmentation mask. 
But, the new contrast-agnostic model works pretty well even in these cases. So I believe we can remove the optional argument `dilation=(0,0,0)` (as mentioned by @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4102#discussion_r1199288297).

mguaypaq (approved):
That's a nice extra change for the usage string, thanks!

PR #4106, opened 2023-04-25
Fix Debian 9 CI failures by adding `archive` links to `/etc/apt/sources.list`

### LINKED ISSUE #4105: Debian 9/stretch has moved to `archive.debian.org` causing CI runs to fail
CI run failures: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/workflows/tests.yml?query=is%3Afailure+branch%3Amaster

```
Ign:1 http://deb.debian.org/debian stretch InRelease
Ign:2 http://security.debian.org/debian-security stretch/updates InRelease
Err:3 http://security.debian.org/debian-security stretch/updates Release
  404  Not Found [IP: 151.101.2.132 80]
Ign:4 http://deb.debian.org/debian stretch-updates InRelease
Err:5 http://deb.debian.org/debian stretch Release
  404  Not Found
Err:6 http://deb.debian.org/debian stretch-updates Release
  404  Not Found
Reading package lists...
E: The repository 'http://security.debian.org/debian-security stretch/updates Release' does not have a Release file.
E: The repository 'http://deb.debian.org/debian stretch Release' does not have a Release file.
E: The repository 'http://deb.debian.org/debian stretch-updates Release' does not have a Release file.
```

Context: https://lists.debian.org/debian-devel-announce/2023/03/msg00006.html

> Date: Mon, 27 Mar 2023 00:40:23 +0200
> 
> the stretch, stretch-debug and stretch-proposed-updates suites have now
> also been imported on archive.debian.org. People still interested in
> these should update their sources.list.
> 
> I plan to remove the suites from the main archive in about a month
> (2023-04-23 or later).
>
> The stretch-backports, stretch-backports-sloppy and related debug suites
> will likely move soon as well and might be removed from the main archive
> around the same time.

The timing here (April 23rd) lines up with when we started getting failures.

We could consider:

- Retaining CI runs for Debian 9, and modifying `sources.list` to point to `archive.debian.org`
- Removing CI runs for Debian 9, and adding Debian 11

joshuacwnewton:
## Description

Debian 9 hit EOL in June 2022. Despite this, SCT has still supported Debian 9 because it's been easy to do so: the distribution channels for Debian 9 packages had remained on http://deb.debian.org/ for some time after EOL status, requiring no action on our part. However, over March-April 2023, these channels were migrated to http://archive.debian.org and removed from http://deb.debian.org/.

This removal has caused our CI runs for Debian 9 to fail. To fix this, we must overwrite the existing entries in `/etc/apt/sources.list`, and add new sources that point to the archive location. This PR does exactly that.

Long-term, we should think about our plans to phase EOL distributions out of our test suite (as well as adding newly-released distributions). It would be nice to have a more concrete schedule/plan. (Do we support three versions of Debian (9-11)? Four (9-12)? What about other OSs such as Ubuntu or macOS?) Right now, we handle this ad hoc, when it would be nice to plan ahead.

## Linked issues

Fixes #4105.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -157,8 +157,8 @@ jobs:
         run: |  
           echo "deb http://archive.debian.org/debian/ stretch main non-free contrib" > /etc/apt/sources.list
           echo "deb-src http://archive.debian.org/debian/ stretch main non-free contrib" >> /etc/apt/sources.list
-          echo "deb http://archive.debian.org/debian-security/ stretch-updates main non-free contrib" >> /etc/apt/sources.list
-          echo "deb-src http://archive.debian.org/debian-security/ stretch-updates main non-free contrib" >> /etc/apt/sources.list
+          echo "deb http://archive.debian.org/debian-security/ stretch/updates main non-free contrib" >> /etc/apt/sources.list
+          echo "deb-src http://archive.debian.org/debian-security/ stretch/updates main non-free contrib" >> /etc/apt/sources.list
           apt update && apt install -y libglib2.0-0 libgl1-mesa-dev procps gcc git curl
       - uses: actions/checkout@v3
       - name: Install SCT
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -152,12 +152,8 @@ jobs:
         # NB: libgl1-mesa-dev is needed for PyQT testing (https://stackoverflow.com/q/33085297/7584115)
         # NB: We echo "archive.debian.org" to address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4105
         run: |  
-          echo "Before:"
           cat /etc/apt/sources.list
           sed -i -e 's/\(deb\|security\)\.debian\.org/archive.debian.org/' /etc/apt/sources.list
-          sed -i -e 's/stretch-updates/stretch-proposed-updates/' /etc/apt/sources.list
-          echo
-          echo "After:"
           cat /etc/apt/sources.list
           apt update && apt install -y libglib2.0-0 libgl1-mesa-dev procps gcc git curl
       - uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -152,9 +152,10 @@ jobs:
         # NB: libgl1-mesa-dev is needed for PyQT testing (https://stackoverflow.com/q/33085297/7584115)
         # NB: We echo "archive.debian.org" to address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4105
         run: |  
-          cat /etc/apt/sources.list
-          sed -i -e 's/\(deb\|security\)\.debian\.org/archive.debian.org/' /etc/apt/sources.list
-          cat /etc/apt/sources.list
+          echo "deb http://archive.debian.org/debian/ stretch main non-free contrib" > /etc/apt/sources.list
+          echo "deb-src http://archive.debian.org/debian/ stretch main non-free contrib" >> /etc/apt/sources.list
+          echo "deb http://archive.debian.org/debian-security/ stretch/updates main non-free contrib" >> /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -144,12 +144,21 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
```
joshuacwnewton, line 147: TODO: [fa81a72](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4106/commits/fa81a72774fa259a774df740bece6aa168293a81) will need to be reverted before this PR is merged.

(Debian 9 tests typically only run on `master`; they have been temporarily enabled only to test this PR.)

joshuacwnewton (commented):
Given that the current changes to `tests.yml` result in a passing run, I've marked this as ready for review.

.github/workflows/tests.yml
```diff
@@ -144,12 +144,18 @@ jobs:
...
     steps:
       - name: Dependencies
-        run: |  # NB: libgl1-mesa-dev is needed for PyQT testing (https://stackoverflow.com/q/33085297/7584115)
+        # NB: libgl1-mesa-dev is needed for PyQT testing (https://stackoverflow.com/q/33085297/7584115)
+        # NB: We echo "archive.debian.org" to address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4105
+        run: |  
+          echo "deb http://archive.debian.org/debian/ stretch main non-free contrib" > /etc/apt/sources.list
+          echo "deb-src http://archive.debian.org/debian/ stretch main non-free contrib" >> /etc/apt/sources.list
+          echo "deb http://archive.debian.org/debian-security/ stretch/updates main non-free contrib" >> /etc/apt/sources.list
+          echo "deb-src http://archive.debian.org/debian-security/ stretch/updates main non-free contrib" >> /etc/apt/sources.list
```
joshuacwnewton, line 158: For these commands, I followed https://www.howtoforge.com/using-old-debian-versions-in-your-sources.list as a rough guide, but it would be nice to double-check if everything here is 100% necessary.

mguaypaq: Looking at my own `sources.list`, I see essentially the same four lines, so this seems fine to me.

But, a suggestion: if we literally just want to replace `deb.debian.org` and `security.debian.org` with `archive.debian.org`, why not do exactly that:
```sh
sed -i -e 's/\(deb\|security\)\.debian\.org/archive.debian.org/' /etc/apt/sources.list
```

mguaypaq (approved):
Thanks for testing this, and for the nice PR description!

.github/workflows/tests.yml
joshuacwnewton: Good thinking! I feel like pattern-based techniques (regex, sed, etc.) are definitely a blind spot of mine. Too rarely do I consider them as a first option? 

The only thing I'm unsure about is whether the default entries for Debian 9 map to `archive` links. If [this link](https://www.geekersdigest.com/apt-etc-apt-sources-list-for-every-debian-release/) is to be trusted, then there is an additional source:

```diff
deb http://deb.debian.org/debian/ stretch main contrib non-free
+deb http://deb.debian.org/debian/ stretch-updates main contrib non-free
deb http://security.debian.org/ stretch/updates main contrib non-free
```

Checking http://archive.debian.org/debian/ for `stretch-updates` shows no such distribution:

image

So, I have a feeling that the proposed `sed` command might result in a 404 not found. (But, nonetheless, I'd be happy to try it, as it's a much more elegant solution.)

joshuacwnewton: I wonder if `stretch-updates` is supposed to map to `stretch-proposed-updates`? :thinking:

PR #4112, opened 2023-05-04
Allow lesion volume calculation without providing SC seg

valosekj:
#### GitHub

#### PR contents

## Description
This PR drops the requirement of the spinal cord segmentation (or centerline) for `sct_analyze_lesion` function. The segmentation/centerline is used for angle correction. However, no angle correction is actually performed when calculating the lesion volume.

Thus, this PR introduces the possibility of running `sct_analyze_lesion` without providing segmentation/centerline. In this case, only lesion volume is calculated.

Example:

```
$ sct_analyze_lesion -m sub-edm143_ses-M0_PSIR_lesion.nii.gz
...
Lesion count = 1

Measures on lesion #1...
  Volume : 39.7 mm^3

Orient output image to initial orientation...
File sub-edm143_ses-M0_PSIR_lesion_label.nii.gz already exists. Will overwrite it.


Averaged measures...
  Volume [mm^3] = 39.7+/-0.0
  (S-I) Length [mm] = nan+/-nan
  Equivalent Diameter [mm] = nan+/-nan

Total volume = 39.7 mm^3
Lesion count = 1
...
```

Context: https://github.com/ivadomed/canproco/issues/27

---

<s>Sidenote: I had to manually install `xlsxwriter` into `venv_sct`:

```
conda activate venv_sct
conda install xlsxwriter
```
</s>

UPDATE: this should not be a problem for future users since `xlsxwriter` is automatically installed when installing SCT; see this comment: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4112#discussion_r1195198446

## Linked issues

Resolves: https://github.com/ivadomed/canproco/issues/27


valosekj:
I added a test. The PR is now ready for review @mguaypaq, @Nilser3.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -4,7 +4,7 @@
 #
 # Copyright (c) 2014 Polytechnique Montreal <www.neuro.polymtl.ca>
 # Author: Charley
-# Modified: 2017-08-19
+# Modified: 2023-05-16 by Jan Valosek
```
valosekj, line 7: Do we want to keep this line? Since we use git/github, is it necessary?

jcohenadad: not necessary

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton, line 1: > Sidenote: I had to manually install xlsxwriter into venv_sct:
>
> ```
> source ${SCT_DIR}/python/etc/profile.d/conda.sh
> conda activate venv_sct
> conda install xlsxwriter
> ```

Just a quick reply to the side note you mentioned above:

- The need for `xlsxwriter` comes from [#4091](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4091), specifically https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4091/commits/776da15d38fd149418df2fad17ff7feb8dcddc78.
- Even though SCT uses a conda environment, we don't actually use conda packages, hehe. (Conda is only used as a means of controlling/isolating the Python installation used for SCT.) So, I would recommend `pip install xlsxwriter`, or even just `pip install -r requirements.txt`, if you want to refresh the environment with the most recent packages. (I recommend this because in my experience, mixing conda packages with pip packages can be a source of weird dependency-related issues.)
- Perhaps in the future, whenever we make a change to `requirements.txt`, I can ping `#sct_dev` alerting folks that they will have to update their SCT environments if they plan to work off of master.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: I reset the "Modified" line because all script headers will be cleaned globally in a separate PR (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4117).

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -402,9 +404,13 @@ def measure(self):
...
-            self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
-            self._measure_length(im_lesion_data_cur, p_lst, label_idx)
-            self._measure_diameter(im_lesion_data_cur, p_lst, label_idx)
+            # No angle correction is needed for the volume measurement
+            if self.fname_sc is None:
+                self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
+            else:
+                self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
+                self._measure_length(im_lesion_data_cur, p_lst, label_idx)
+                self._measure_diameter(im_lesion_data_cur, p_lst, label_idx)
```
joshuacwnewton, line 413: Very small nitpick, but I would swap the two cases, that way we match the other `is not None` check above:

```suggestion
            # If SC segmentation is provided, we can compute length and diameter
            if self.fname_sc is not None:
                self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
                self._measure_length(im_lesion_data_cur, p_lst, label_idx)
                self._measure_diameter(im_lesion_data_cur, p_lst, label_idx)
            # Otherwise, without SC seg, we can only compute the volume
            else:
                self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -44,19 +44,20 @@ def get_parser():
...
         "-h",
         "--help",
         action="help",
         help="show this help message and exit")
+    optional.add_argument(
+        "-s",
+        required=False,
+        help="Spinal cord centerline or segmentation file, which will be used to correct morphometric measures with "
+             "cord angle with respect to slice. (e.g.'t2_seg.nii.gz')\nIf not provided, only lesion volume will be "
+             "computed.",
```
joshuacwnewton, line 59: Small nitpick, but I think we can be a little clearer about what will and won't be computed:

```suggestion
        help="Spinal cord centerline or segmentation file, which will be used to correct morphometric measures with "
             "cord angle with respect to slice. (e.g. 't2_seg.nii.gz')\n"
             "If provided, then the lesion volume, length, and diameter will be computed. Otherwise, if not provided, "
             "then only the lesion volume will be computed.",
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: We may even want to simplify this further so that there's only a single call to `self._measure_volume`:

```suggestion
            # If SC segmentation is provided, we can compute length, diameter, and volume
            # Otherwise, without SC seg, we can only compute the volume
            if self.fname_sc is not None:
                self._measure_length(im_lesion_data_cur, p_lst, label_idx)
                self._measure_diameter(im_lesion_data_cur, p_lst, label_idx)
            self._measure_volume(im_lesion_data_cur, p_lst, label_idx)
```

But, I'll leave it up to you as to which one you prefer, as I'm okay with either. :slightly_smiling_face:

joshuacwnewton (approved):
Thank you for this PR! The changes look very straightforward, and most of my comments are just minor nitpicks and small bits of simplification.

Because the core changes LGTM, I'm pre-approving. :slightly_smiling_face:

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -89,3 +90,37 @@ def test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements(dummy_les
...
+    # Validate analysis results
+    for key, expected_value in expected_measurements.items():
+        if key == 'volume [mm3]':
+            np.testing.assert_equal(measurements.at[0, key], expected_value)
+        else:
+            # The max_equivalent_diameter and length are nan because no segmentation is provided
+            if key == 'max_equivalent_diameter [mm]':
+                assert math.isnan(measurements.at[0, key]) is True
+            elif key == 'length [mm]':
+                assert math.isnan(measurements.at[0, key]) is True
```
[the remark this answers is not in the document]
mguaypaq: For normal code, I would agree that such duplication means we should look for a common abstraction, but for testing code I'm not so sure. Normally, I would advocate for the most straightforward code possible in tests, and in particular, as little control flow as possible (ifs, loops, etc.). For testing code, if that means a small amount of copy-pasted code, I think that's fine.

Here, you're right that `args` could be parametrized easily, but I don't think the diameter/length can be parametrized straightforwardly, since it involves `nan` comparisons. (It might be possible to do something tricky with `nan` comparisons, like `assert not X >= expected`, but I don't think that would be nice to read and reason about when debugging a failing test.)

On a different note, though, I will point out that assertions of the form:
```py3
assert X is True
assert Y is False
```
can be better written as the shorter:
```py3
assert X
assert not Y
```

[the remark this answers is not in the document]
valosekj: I agree with @mguaypaq. Even the [first parametrization](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/86c96028ad112fad614596cb22df6e527978793e/testing/cli/test_cli_sct_analyze_lesion.py#L96-L102) had to be explained to me by @mguaypaq. So, I would vote for keeping the tests split to be understandable even for less experienced users (like me 😄). 

Regarding the`assert` simplification:

> On a different note, though, I will point out that assertions of the form:
> 
> ```python
> assert X is True
> assert Y is False
> ```
> 
> can be better written as the shorter:
> 
> ```python
> assert X
> assert not Y
> ```

I have done this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4112/commits/86c96028ad112fad614596cb22df6e527978793e, together with simplification of `if-else` logic.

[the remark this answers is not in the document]
joshuacwnewton: I am okay with this plan. :)

> but I don't think the diameter/length can be parametrized straightforwardly, since it involves `nan` comparisons.

Just to clarify my thought process: rather than checking against `nan`, I was going to propose not checking length/diameter at all for the no-seg case. So, the changes to the `if` statements would be rather simple, e.g.:

```diff
-            if key == 'max_equivalent_diameter [mm]':
+            if path_seg and key == 'max_equivalent_diameter [mm]':
```

I was thinking that because the user isn't expecting length/diameter if they're not providing a seg, then they don't really care whether it's nan? But, maybe there's more value in ensuring that these values are nan than I'm giving credit for.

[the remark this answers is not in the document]
joshuacwnewton: Anyway, I think this PR is good to go now?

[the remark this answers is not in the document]
valosekj: > Anyway, I think this PR is good to go now?

Agree --> merging

PR #4114, opened 2023-05-04
Ensure CI avoids older ReadTheDocs docker images to address OpenSSL mismatch

### LINKED ISSUE #4113: ReadTheDocs builds are failing due to `urllib3`/`ssl`-related extension error for `sphinx.builders.linkcheck`
It looks like we started getting errors a day ago on master, and then more recently all builds within the past 5 hours or so: https://readthedocs.org/projects/spinalcordtoolbox/builds/?page=1

<details>

image

</details>

```
Extension error:
Could not import extension sphinx.builders.linkcheck (exception: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168)
```

I imagine this is the sort of issue that has an upstream element to it, so I'll do a little digging.

joshuacwnewton:
## Description

This follows the advice from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4113#issuecomment-1535392170 to address the recently-failing RTD builds.

## Linked issues

Fixes #4113.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,9 @@
 # Required
 version: 2
 
+build:
+  os: ubuntu-22.04
+
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: documentation/source/conf.py
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,6 +7,8 @@ version: 2
 
 build:
   os: ubuntu-22.04
+  tools:
+      python: 3.8
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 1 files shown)


joshuacwnewton:
As an update, there has been more activity on both the urllib3 and RTD side of things. 

But, tl;dr, the RTD devs [have affirmed](https://github.com/readthedocs/readthedocs.org/issues/10290#issuecomment-1535421149) that using `build.os` is the right way to go about fixing this issue, so this PR is in the clear.

mguaypaq (approved):
Thanks for the investigation and the fix!

PR #4115, opened 2023-05-12
Delete `create_test_data.py` and move `dummy_` functions into their respective test files

### LINKED ISSUE #3136: Centralize SCT's test data code (dummy fixtures, 'sct_testing_data/')

joshuacwnewton:
## Description

This is a minor organizational PR: 

- Currently, within the core `spinalcordtoolbox` package, there is a non-module `testing/` folder containing a single file: `create_test_data.py`.
- This file contains a few functions that create dummy test data -- these dummy functions are imported by only 1 or 2 tests each.
- So, to avoid keeping test-specific functions in the main package, I've moved the dummy functions to their respective test files in `testing/`.

## Future work

In issue #3136, and in commit c3e60e2bb9936bcec66919e8ca00c5a10b24b9b9, I've talked about potentially centralizing dummy test data functions/fixtures into a single file. 

I had proposed this because:

- I suspect that there's a lot of duplicated effort in creating dummy test data -- there might be similar functions scattered throughout various tests that create similar fake test data, that we could potentially de-duplicate.
- It might also ease the burden for new contributors -- if there is a place where they can find existing functions for generating dummy segmentations, centerlines, etc., then it might be easier to write new tests, knowing they don't have to figure out what test data to use from scratch.

However, now I'm not so sure that it's worth the effort to disentangle the dummy test data from their specific tests. I'd almost rather spend the effort actually writing tests, as opposed to restructuring how our tests are organized. Hence, why I've linked this PR to close issue #3136, rather than keeping it open.

## Linked issues

Fixes #3136.


mguaypaq (approved):
I don't really have an opinion on our current (or future) organization for tests, but this seems fine.

PR #1311, opened 2017-05-08
Moved the QC assets into the spinalcordtoolbox package

peristeri:
It makes it easier to access from within the package and solved the bug when
installing SCT in a different path.

Fixes #1300

PR #4129, opened 2023-05-30
Remove unused QC report assets

mguaypaq:
The other files under `_assets/` are all used, either directly from `index.html`, or indirectly through `bootstrap.min.css`. But even looking through the git history, I couldn't find any references to these files.


joshuacwnewton (approved):
It's interesting... some, but not all of these files were introduced in the very first QC PR:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/1250

The others are later introduced in:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/1311

Which, contrary to its PR title, doesn't just move the assets -- it adds new ones. (???)

It feels like these were accidentally-committed files, given the haphazard nature of those PRs. So, I'm very OK with removing them. 

(Testing the QC reports, everything looks A-OK as well.)

PR #4133, opened 2023-06-07
Ensure that sagittal images (and softsegs) are resampled correctly for lesion QC

### LINKED ISSUE #4125: Default resampling (`0.6mm`) for `sct_deepseg_lesion` QC report distorts lesion softsegs
Newly implemented sagittal lesion QC (PR [#4102](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4102)) works only with binary lesion masks. 

If the lesion mask is non-binary (i.e., soft), the output QC is corrupted:

image

Example subject: `data.neuro.polymtl.ca/sci-zurich/sub-zh41/ses-01`

Command:

```console
sct_qc -i sub-zh41/ses-01/anat/sub-zh41_ses-01_acq-sag_T2w.nii.gz -s derivatives/labels/sub-zh41/ses-01/anat/sub-zh41_ses-01_acq-sag_T2w_seg-manual.nii.gz -d derivatives/labels/sub-zh41/ses-01/anat/sub-zh41_ses-01_acq-sag_T2w_lesion-manual.nii.gz -p sct_deepseg_lesion -plane sagittal -qc qc
```

I temporarily solved the issue by [lesion mask binarization](https://github.com/ivadomed/model_seg_sci/blob/3bac2896f00bea21fe0dd3d86b2c7d31f029417b/generate_new_lesions/generate_new_lesion.py#L288-L295):

```python
# Binarize new_lesion (sct_qc supports only binary masks)
new_lesion_bin_path = new_lesion_path.replace('.nii.gz', '_bin.nii.gz')
os.system(f'sct_maths -i {new_lesion_path} -bin 0 -o {new_lesion_bin_path}')
# Example: sct_qc -i t2.nii.gz -s t2_seg.nii.gz -d t2_lesion.nii.gz -p sct_deepseg_lesion -plane axial
os.system(f'sct_qc -i {new_target_path} -s {new_sc_path} -d {new_lesion_bin_path} -p sct_deepseg_lesion '
          f'-plane sagittal -qc {args.dir_save.replace("labelsTr", "qc")} -qc-subject {subject_name_out}')
# Remove binarized lesion
os.remove(new_lesion_bin_path)
```

joshuacwnewton:
## Description

In #4125, we discovered 2 issues with the current `sct_deepseg_lesion` QC:

- For Sagittal images, resampling is applied to the wrong axes, because the code assumes Axial slices by default.
- The method for checking if an image is a segmentation is flawed; instead of inspecting the image contents, the code just assumes that only the last image is a segmentation.
   - This means that earlier segmentation images _aren't_ considered segmentations, resulting in no linear interpolation and no thresholding for lesion segmentations.. 

This PR fixes each issue by:

- Applying the resampling to the correct axes for the Sagittal case.
- Detecting segmentations by inspecting the image contents, ensuring correct interpolation/thresholding for non-last segmentations (including softsegs).

## Linked issues

Fixes #4125.

### DIFF
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -59,7 +59,7 @@ class Slice(object):
             if p_resample:
                 # Check if image is a segmentation (binary or soft) by making sure:
                 # - 0/1 are the two most common voxel values
-                # - 0/1 account for >99% of voxels (to allow for some soft voxels)
+                # - 0/1 account for >95% of voxels (to allow for some soft voxels)
                 unique, counts = np.unique(img.data, return_counts=True)
                 unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
                 binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -63,7 +63,7 @@ class Slice(object):
                 unique, counts = np.unique(img.data, return_counts=True)
                 unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
                 binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
-                binary_percentage = ((counts[0] + counts[1]) / np.sum(counts))
+                binary_percentage = np.sum(counts[0:2]) / np.sum(counts)
                 if binary_most_common and binary_percentage > 0.99:
                     # If a segmentation, use linear interpolation and apply thresholding
                     type_img = 'seg'
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -64,7 +64,7 @@ class Slice(object):
                 unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
                 binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
                 binary_percentage = np.sum(counts[0:2]) / np.sum(counts)
-                if binary_most_common and binary_percentage > 0.99:
+                if binary_most_common and binary_percentage > 0.95:
                     # If a segmentation, use linear interpolation and apply thresholding
                     type_img = 'seg'
                 else:
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -57,8 +57,15 @@ class Slice(object):
             self._absolute_paths.append(img.absolutepath)  # change_orientation removes the field absolute_path
             img.change_orientation('SAL')
             if p_resample:
-                if i == len(images) - 1 and img.dim[3] == 1:
-                    # Last volume, if it is 3d, corresponds to a segmentation, therefore use linear interpolation here
+                # Check if image is a segmentation (binary or soft) by making sure:
+                # - 0/1 are the two most common voxel values
+                # - 0/1 account for >99% of voxels (to allow for some soft voxels)
+                unique, counts = np.unique(img.data, return_counts=True)
+                unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
+                binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
+                binary_percentage = ((counts[0] + counts[1]) / np.sum(counts))
+                if binary_most_common and binary_percentage > 0.99:
+                    # If a segmentation, use linear interpolation and apply thresholding
                     type_img = 'seg'
                 else:
                     # Otherwise it's an image: use spline interpolation
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -293,15 +293,14 @@ class Slice(object):
         nii = Nifti1Image(image.data, image.hdr.get_best_affine())
         # If no reference image is provided, resample to specified resolution
         if image_ref is None:
-            slice_type = self.get_name()
-            if slice_type == 'Axial':
-                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
-                nii_r = resample_nib(nii, new_size=[image.dim[4], p_resample, p_resample], new_size_type='mm',
-                                     interpolation=dict_interp[type_img])
-            elif slice_type == 'Sagittal':
-                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
-                nii_r = resample_nib(nii, new_size=[p_resample, p_resample, image.dim[6]], new_size_type='mm',
-                                     interpolation=dict_interp[type_img])
+            # Resample each slice to p_resample x p_resample mm (orientation is SAL by convention in QC module)
+            if isinstance(self, Axial):
+                new_size = [image.dim[4], p_resample, p_resample]
+            elif isinstance(self, Sagittal):
+                new_size = [p_resample, p_resample, image.dim[6]]
+            else:
+                raise TypeError(f"Unexpected slice type: {type(self)}")
+            nii_r = resample_nib(nii, new_size=new_size, new_size_type='mm', interpolation=dict_interp[type_img])
 
         # Otherwise, resampling to the space of the reference image
         else:
--- a/spinalcordtoolbox/reports/slice.py
+++ b/spinalcordtoolbox/reports/slice.py
@@ -286,9 +286,16 @@ class Slice(object):
         nii = Nifti1Image(image.data, image.hdr.get_best_affine())
         # If no reference image is provided, resample to specified resolution
         if image_ref is None:
-            # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
-            nii_r = resample_nib(nii, new_size=[image.dim[4], p_resample, p_resample], new_size_type='mm',
-                                 interpolation=dict_interp[type_img])
+            slice_type = self.get_name()
+            if slice_type == 'Axial':
+                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
+                nii_r = resample_nib(nii, new_size=[image.dim[4], p_resample, p_resample], new_size_type='mm',
+                                     interpolation=dict_interp[type_img])
+            elif slice_type == 'Sagittal':
+                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
+                nii_r = resample_nib(nii, new_size=[p_resample, p_resample, image.dim[6]], new_size_type='mm',
+                                     interpolation=dict_interp[type_img])
+
         # Otherwise, resampling to the space of the reference image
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/reports/slice.py
```diff
@@ -57,8 +57,15 @@ def __init__(self, images, p_resample=0.6):
...
-                if i == len(images) - 1 and img.dim[3] == 1:
-                    # Last volume, if it is 3d, corresponds to a segmentation, therefore use linear interpolation here
+                # Check if image is a segmentation (binary or soft) by making sure:
+                # - 0/1 are the two most common voxel values
+                # - 0/1 account for >99% of voxels (to allow for some soft voxels)
+                unique, counts = np.unique(img.data, return_counts=True)
+                unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
+                binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
+                binary_percentage = ((counts[0] + counts[1]) / np.sum(counts))
+                if binary_most_common and binary_percentage > 0.99:
```
joshuacwnewton, line 67: I'm not 100% sure this is a foolproof method for detecting segmentations. Is there any chance a non-segmentation would be 99% 0/1 voxels? Or, is there any chance that a softseg would have <99% 0/1 voxels?

spinalcordtoolbox/reports/slice.py
```diff
@@ -286,9 +293,16 @@ def _resample_slicewise(self, image, p_resample, type_img, image_ref=None):
         nii = Nifti1Image(image.data, image.hdr.get_best_affine())
         # If no reference image is provided, resample to specified resolution
         if image_ref is None:
-            # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
-            nii_r = resample_nib(nii, new_size=[image.dim[4], p_resample, p_resample], new_size_type='mm',
-                                 interpolation=dict_interp[type_img])
+            slice_type = self.get_name()
+            if slice_type == 'Axial':
```
joshuacwnewton, line 297: Note: The `.get_name()` method is removed in #4126.

So, we should either a) keep it in #4126, b) find another way to get the slice type, or c) make these class methods instead (which seems like a lot of duplicated code) to me.

I'm leaning towards a) because it requires the fewest changes, but if there was a way to do c) without duplicating too much code, it seems like the most... Pythonic? (Though, it would require a greater refactor.)

spinalcordtoolbox/reports/slice.py
joshuacwnewton, line 1: Even with these two fixes, we may want to simply turn of resampling entirely for this QC? (Or, alternatively, select a different resampling value to ensure that that the sagittal mosaic fits on screen:

- Turning off resampling ensures that we display segs/softsegs accurately, but may show the image as too large to fix on one screen.
- Switching resampling to a value that produces a smaller size would help the mosaic fit on screen, but may make the image harder to inspect. (Though, before the resampling bugs were fixed, a small mosaic was already given the :+1: by @valosekj? So, it's possible that resampling to a smaller image is already the best option.)

spinalcordtoolbox/reports/slice.py
valosekj: Yeah, this is tricky. It really depends on the character of the lesion (how it was segmented, how the rater considered the partial volume effect).
I didn't calculate the voxel intensity values distribution for soft seg, though. Maybe we could decrease it to 95% to be less conservative?

spinalcordtoolbox/reports/slice.py
joshuacwnewton: Sounds good!!

spinalcordtoolbox/reports/slice.py
joshuacwnewton: Ah, wait, I have an idea!

joshuacwnewton: I've tried thinking of a way to make 3. work. For example, we could create a method in each slice called `get_slice_dim`, and it returns the dimensions used for slices in a given SAL image. (e.g. for Axial it returns `[1, 2]` (AP, LR) and for Sagittal it returns `[0, 1]` (SI, AP). Then, those values are used to determine `new_size`, and the resampling call stays the same.

But, this just felt a bit too convoluted. Plus, `Slice` objects already have some suspect Sagittal and Axial-specific methods anyway. So, I think we can save Pythonic inheritance patterns for a separate refactoring PR of `slice.py`.

naga-karthik:
Hey @joshuacwnewton, thank you for your efforts in improving the sagittal QCs! I have tested on a few more subjects. While the sagittal images look quite good, there are some subjects where the image is missing or the contrast is quite dark. You can find the examples below: 

<details>

<summary> missing subject 1 </summary>

Screen Shot 2023-06-13 at 2 16 28 PM

</details>

<details>

<summary> missing subject 2 </summary>

Screen Shot 2023-06-13 at 2 17 04 PM

</details>

<details>

<summary> dark contrast </summary>

Screen Shot 2023-06-13 at 2 16 01 PM

</details>

Note that I didn't have these issues when I was using the sagittal QC from the `master` branch. Curious to know what you think the issue might be!

joshuacwnewton:
Thank you for taking the time to test this PR! I'll try to debug. :heart: 

Which dataset contains these subjects? And, is the dataset on duke, or data.neuro.polymtl.ca, or...?

naga-karthik:
Hmm, seems like the issue exists on `master` as well? It is strange because I was on an earlier version of `master` and I pulled the latest version which is running into issues. 

<details>

<summary> missing subject </summary>

Screen Shot 2023-06-13 at 2 57 58 PM

</details>

Were there any recent changes made to `master` regarding the sagittal QC?

> Which dataset contains these subjects? And, is the dataset on duke, or data.neuro.polymtl.ca, or...?

So, the results are from a combination of two datasets: `data-multi-subject` (spine generic) and `sci-zurich`. If you want to debug with the exact subjects, you can find them on duke here: `duke/temp/muena/Dataset553_temp`

joshuacwnewton:
> If you want to debug with the exact subjects, you can find them on duke here: `duke/temp/muena/Dataset553_temp`

Thanks for this!! I used the following command to try to reproduce for the subject in ["missing subject 1"](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4133#issuecomment-1589818856): 

```
sct_qc -i imagesTr/sub-oxfordFmrib09_sub-zh71_ses-01_026_0000.nii.gz \
       -s masksTr/sub-oxfordFmrib09_sub-zh71_ses-01_026.nii.gz \
       -d labelsTr/sub-oxfordFmrib09_sub-zh71_ses-01_026.nii.gz \
       -p sct_deepseg_lesion -plane sagittal -qc ~/Desktop/qc
```

On master, I see this: 

<details>

Before toggle:

image

After toggle:

image

</details>

And, on this branch (`jn/4125-fix-sagittal_qc_resampling`), I see the following:

<details>

image

</details>

No missing subject for me! 

Can you let me know what you did to produce the "missing subject" QC? (Was it an individual command, or was it an `sct_run_batch` script?)

naga-karthik:
Hey @joshuacwnewton, thank you for digging in! I apologize, it might have been some error in reporting on my end. I re-ran the QCs from your branch and the result looks good this time. No missing subjects, no dark contrasts. See for example the same subject from your above comment: 

<details>

<summary> Details </summary>

Screen Shot 2023-06-13 at 10 35 45 PM

</details>

I am really not sure what happened earlier. Do you think that the resampling of some specific subjects might have caused some internal error and hence they show up on the QC? 

> Can you let me know what you did to produce the "missing subject" QC? (Was it an individual command, or was it an sct_run_batch script?)

If it helps, I ran the individual `sct_qc` command on each subject in all my experiments (albeit, they were run inside a loop)

joshuacwnewton:
> I am really not sure what happened earlier. Do you think that the resampling of some specific subjects might have caused some internal error and hence they show up on the QC?

Truthfully, I don't have a good theory here either. It's hard to say without seeing the output log from the processing. :thinking:

spinalcordtoolbox/reports/slice.py
joshuacwnewton: #4126 has been merged, and `get_name()` has been kept for now, meaning that we're going for 1. at the moment.

mguaypaq: > Plus, `Slice` objects already have some suspect Sagittal and Axial-specific methods anyway (e.g. `get_sagittal_dim` and `get_axial_dim`? why not just `get_dim` as an abstract method?). So, I think we can save aiming for Pythonic inheritance patterns for a separate refactoring PR of `slice.py`.

That sounds like a good idea, yes.

I'll point out another, slightly more Pythonic (but still not great) way to do Option 2: testing the slice object with `isinstance(self, Axial)`, rather than doing a string comparison which boils down to `type(self).__name__ == 'Axial'`.

spinalcordtoolbox/reports/slice.py
```diff
@@ -57,8 +57,15 @@ def __init__(self, images, p_resample=0.6):
             self._absolute_paths.append(img.absolutepath)  # change_orientation removes the field absolute_path
             img.change_orientation('SAL')
             if p_resample:
-                if i == len(images) - 1 and img.dim[3] == 1:
-                    # Last volume, if it is 3d, corresponds to a segmentation, therefore use linear interpolation here
+                # Check if image is a segmentation (binary or soft) by making sure:
+                # - 0/1 are the two most common voxel values
+                # - 0/1 account for >99% of voxels (to allow for some soft voxels)
```
mguaypaq, line 62: I guess this works for now, but doing this kind of auto-detection is always fragile. Really, the calling code should _know_ which images are intended to be segmentations without guessing, and the information shouldn't get lost along the control flow. But I understand that this is hard to do with the current code architecture, and out of scope for this PR.

spinalcordtoolbox/reports/slice.py
```diff
@@ -286,9 +293,16 @@ def _resample_slicewise(self, image, p_resample, type_img, image_ref=None):
...
-                                 interpolation=dict_interp[type_img])
+            slice_type = self.get_name()
+            if slice_type == 'Axial':
+                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
+                nii_r = resample_nib(nii, new_size=[image.dim[4], p_resample, p_resample], new_size_type='mm',
+                                     interpolation=dict_interp[type_img])
+            elif slice_type == 'Sagittal':
+                # Resample to px x p_resample x p_resample mm (orientation is SAL by convention in QC module)
+                nii_r = resample_nib(nii, new_size=[p_resample, p_resample, image.dim[6]], new_size_type='mm',
+                                     interpolation=dict_interp[type_img])
```
mguaypaq, line 304: Since the only difference between the two branches is `new_size`, how about this?
```suggestion
            # Resample each slice to p_resample x p_resample mm (orientation is SAL by convention in QC module)
            if isinstance(self, Axial):
                new_size = [image.dim[4], p_resample, p_resample]
            elif isinstance(self, Sagittal):
                new_size = [p_resample, p_resample, image.dim[6]]
            else:
                raise TypeError(f"Unexpected slice type: {type(self)}")
            nii_r = resample_nib(nii, new_size=new_size, new_size_type='mm', interpolation=dict_interp[type_img])
```

mguaypaq (approved):
Thanks for the fix, although I would love to have a more principled fix after we improve the QC report code architecture.

Issue #4134, opened 2023-06-09
Using nibabel's `set_qform`/`set_sform` without specifying `code` will overwrite `0` codes with `2`


joshuacwnewton:
Nibabel headers have methods called `set_sform` and `set_qform`. SCT uses these methods in places like reorientation: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/922b8091319051f66d140baf56d2bd0e1ba0a6e8/spinalcordtoolbox/image.py#L1116-L1117

However, in the above example, since we specify `affine` but _don't_ specify `code`, the following logic is invoked by nibabel:

```python
        """
        code : None, string or integer, optional
            String or integer giving meaning of transform in *affine*.
            The default is None.  If code is None, then:

            * If affine is None, `code`-> 0
            * If affine not None and existing sform code in header == 0,
              `code`-> 2 (aligned)
            * If affine not None and existing sform code in header != 0,
              `code`-> existing sform code in header
        """
```

So, a common situation that will happen is:

1. An image's qform/sform codes will be overwritten to nibabel's defaults of `0` and `2`, for example:
    * https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3005
    * https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3295
2. Then, `set_qform` will be used, and the `0` value will be overwritten with `2`.

This is exactly what happens in the following issue: 

* https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4132

Now, we don't necessarily need to specify `code`? Because technically, if none of the codes are set to `0` in the first place, then no codes will be overwritten.

But, by making sure we specify `code`, we can A) be more explicit about _where_ the codes are coming from, and B) reveal more instances of default  `0`/`2` codes, which will help us track down cases like #3005 and #3295.

Issue #3005, opened 2020-11-05
Resample alters the qform_code of the output NIfTI header


jcohenadad:
with SCT version: git-lr/missing_disc_label_vertebrae-d5bd44bc312de4db31d1556bea66f15b917e9b7c

When resampling a data to a fixed resolution (specified in mm), the qform_code changes, which is not expected behaviour and could have [consequences down the line](https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html).

Syntax: 
~~~
sct_resample -i ${file}.nii.gz -mm $interp -o ${file}_r.nii.gz
~~~

Header before resampling:
~~~
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	-0.999111 -0.000000 -0.025460 112.883957 
qto_xyz:2	0.000000 0.625000 -0.000000 -67.302994 
qto_xyz:3	-0.040734 0.000000 0.624481 -99.376259 
qto_xyz:4	0.000000 0.000000 0.000000 1.000000 
qform_xorient	Right-to-Left
qform_yorient	Posterior-to-Anterior
qform_zorient	Inferior-to-Superior
sform_name	Scanner Anat
sform_code	1
sto_xyz:1	-0.999111 0.000000 -0.025461 112.883957 
sto_xyz:2	0.000000 0.625000 0.000000 -67.302994 
sto_xyz:3	-0.040736 0.000000 0.624481 -99.376259 
sto_xyz:4	0.000000 0.000000 0.000000 1.000000 
sform_xorient	Right-to-Left
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
file_type	NIFTI-1+
file_code	1
descrip		TE=1.9;Time=141035.000
aux_file	
~~~

Header after resampling:
~~~
qform_name	Unknown
qform_code	0
qto_xyz:1	-0.999111 0.000000 -0.040737 112.883957 
qto_xyz:2	0.000000 1.000000 0.000000 -67.302994 
qto_xyz:3	-0.040734 0.000000 0.999170 -99.376259 
qto_xyz:4	0.000000 0.000000 0.000000 1.000000 
qform_xorient	Right-to-Left
qform_yorient	Posterior-to-Anterior
qform_zorient	Inferior-to-Superior
sform_name	Unknown
sform_code	2
sto_xyz:1	-0.999111 0.000000 -0.040738 112.883957 
[...]

Issue #4025, opened 2023-02-02
`sct_deepseg_sc` outputs segmentation with different `pixdim` than input image


valosekj:
### Description

When running:

```
sct_deepseg_sc -i sub-001_T1w.nii.gz -c t1
```

The output SC seg has a different `pixdim` that the input image:

```
$ sct_image -i sub-001_T1w.nii.gz -header | grep pixdim
pixdim		[-1.0, 0.351562, 0.351562, 2.5, 0.889114, 0.0, 0.0, 0.0]
```

```
$ sct_image -i sub-001_T1w_seg.nii.gz -header | grep pixdim
pixdim		[-1.0, 0.351562, 0.351563, 2.499999, 1.0, 1.0, 1.0, 1.0]
```

Notice the difference `2.5, 0.889114, 0.0, 0.0, 0.0` vs `2.499999, 1.0, 1.0, 1.0, 1.0`.

This difference is consequently raising errors ([here](https://github.com/MIC-DKFZ/nnUNet/blob/d1431c223407c3a66222409cb67c19ae87b2ce6a/nnunet/preprocessing/sanity_checks.py#L150-L152) and [here](https://github.com/MIC-DKFZ/nnUNet/blob/d1431c223407c3a66222409cb67c19ae87b2ce6a/nnunet/preprocessing/sanity_checks.py#L62-L63)) when using [`nnU-Net`](https://github.com/MIC-DKFZ/nnUNet) package.

Example data availbele here.

joshuacwnewton:
I can reproduce this issue using testing data, too: 

```console
master|joshua@tadpole:~/repos/spinalcordtoolbox$ cd data/sct_example_data/t1
master|joshua@tadpole:~/repos/spinalcordtoolbox/data/sct_example_data/t1$ sct_deepseg_sc -i t1.nii.gz -c t1

--
Spinal Cord Toolbox (git-master-9f4d05ad2ce9bd48fd5a396310597fb5db645253)

sct_deepseg_sc -i t1.nii.gz -c t1
--

[...] 

Done! To view results, type:
fsleyes /home/joshua/repos/spinalcordtoolbox/data/sct_example_data/t1/t1.nii.gz -cm greyscale /home/joshua/repos/spinalcordtoolbox/data/sct_example_data/t1/t1_seg.nii.gz -cm red -a 70.0 &

[...]

PR #4135, opened 2023-06-09
Ensure that qform/sform codes are preserved when generating segmentation

### LINKED ISSUE #4132: `sct_deepseg_sc` doesn't preserve s/qform codes, and instead sets both to `2`
If an input image has, for example, s/qform codes equal to `1`, then `sct_deepseg_sc` would overwrite both codes with `2`. 

> **Note**: These codes mean:
> 
> image
> 
> In theory, since the segmentation is in the same space as the input image, it should still be considered "Scanner anat" aligned, right?

Changing the codes like this can cause issues down the line when using software such as SimpleITK:

- SITK uses the codes to determine whether it should rely on the sform or qform matrix. 
- So, if two images are read in with different codes (anat with code `1` and seg with code `2`), then SITK will fetch the sform for the first image, but the qform for the second image. And, if the sform/qform don't match, then SITK will throw an error. 
- However, if the segmentation image had its code `1` preserved, then the sform would be fetched for both images, resulting in no mismatch within SITK.

This chain of events is what caused the issue in this comment: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4025#issuecomment-1445193122. 

Further details about how SITK works can be found in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4025#issuecomment-1445368452 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4025#issuecomment-1578885999.

joshuacwnewton:
## Description

During `sct_deepseg_sc`, we resample images to 0.5mm prior to inference, then resample back to the native resolution after inference. However, resampling has a known issue: 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3005

In this case, we _definitely_ want to preserve the codes, because we're resampling back to the native space. So, the segmentation image should be aligned to the same space as the input image, and thus the codes should not change. If we don't preserve the codes, we run into #4132, which can cause issues later down the line when working with e.g. SimpleITK.

To fix this, I could have changed the behavior of the resampling code unconditionally to address #3005 outright. However, there are some cases where the overwriting of the codes may be desirable behavior (see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3005#issuecomment-1404347614). So, I decided to make the `preserve_codes` behavior opt-in, and enable it only in the `sct_deepseg_sc` case. ((And, as an added benefit, creating an option ensures the behavior is documented in-code via the docstring.))

I also added a test that fails before this PR, but passes after this PR.

## Linked issues

Partially addresses #3005.
Fixes #4132.

> **Note**: Issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4134 was also discovered as part of the investigation for #4132. However, fixing it is not necessary for this specific PR, as addressing the resampling behavior skips over the orientation behavior described in #4134.

### DIFF
--- a/spinalcordtoolbox/deepseg_/sc.py
+++ b/spinalcordtoolbox/deepseg_/sc.py
@@ -437,7 +437,8 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
 
     # Resample image to 0.5mm in plane
     im_image_res = \
-        resampling.resample_nib(im_image, new_size=[0.5, 0.5, im_image.dim[6]], new_size_type='mm', interpolation='linear')
+        resampling.resample_nib(im_image, new_size=[0.5, 0.5, im_image.dim[6]], new_size_type='mm',
+                                interpolation='linear', preserve_codes=True)
 
     fname_orient = 'image_in_RPI_res.nii'
     im_image_res.save(fname_orient)
@@ -454,7 +455,8 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
 
     if ctr_algo == 'file':
         im_ctl = \
-            resampling.resample_nib(im_ctl, new_size=[0.5, 0.5, im_image.dim[6]], new_size_type='mm', interpolation='linear')
+            resampling.resample_nib(im_ctl, new_size=[0.5, 0.5, im_image.dim[6]], new_size_type='mm',
+                                    interpolation='linear', preserve_codes=True)
 
     # crop image around the spinal cord centerline
     logger.info("Cropping the image around the spinal cord...")
@@ -514,7 +516,7 @@ def deep_segmentation_spinalcord(im_image, contrast_type, ctr_algo='cnn', ctr_fi
 
     # resample to initial resolution
     logger.info("Resampling the segmentation to the native image resolution using linear interpolation...")
-    im_seg_r = resampling.resample_nib(im_seg, image_dest=im_image, interpolation='linear')
+    im_seg_r = resampling.resample_nib(im_seg, image_dest=im_image, interpolation='linear', preserve_codes=True)
 
     if ctr_algo == 'viewer':  # for debugging
         im_labels_viewer.save(add_suffix(fname_orient, '_labels-viewer'))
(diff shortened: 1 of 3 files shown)


valosekj (approved):
Thanks for this fix @joshuacwnewton! 💯 

Tested on an example data from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4025#issue-1568352679. ITK header read by [read_itk.py](https://gist.github.com/valosekj/a03195d9060b0e164faff95102129feb#file-read_itk-py).

Seems to be working for me (`direction` are the same for `jn/4132-fix-deepseg_sc-sqform-overwriting-2` branch)!

`master`:

```
$ sct_deepseg_sc -i sub-001_T1w.nii.gz -c t1

$ python read_itk.py sub-001_T1w.nii.gz sub-001_T1w_seg.nii.gz
origin:
(-90.7466812133789, 103.52926635742188, -57.17771911621094)
(-90.7466812133789, 103.52926635742188, -57.17771911621094)
spacing:
(0.3515625, 0.3515625, 2.5)
(0.3515625, 0.3515625298023224, 2.4999990463256836)
direction:
(0.9999999977894414, -6.478192795954982e-09, 6.649148620669932e-05, 2.0651083342197973e-05, -0.9505162149977084, -0.3106749763676334, -6.320324401990454e-05, -0.31067495075469176, 0.950516204300512)
(0.9999999999453735, 1.0322302829269651e-05, 1.6441191221002595e-06, 1.032230276040639e-05, -0.9505162024305571, -0.31067497705397884, 1.6441191998447065e-06, -0.31067498903265794, 0.9505162064003994)
size:
(512, 512, 42)
(512, 512, 42)
```

`jn/4132-fix-deepseg_sc-sqform-overwriting-2`:

```
$ sct_deepseg_sc -i sub-001_T1w.nii.gz -c t1

$ python read_itk.py sub-001_T1w.nii.gz sub-001_T1w_seg.nii.gz
origin:
(-90.7466812133789, 103.52926635742188, -57.17771911621094)
(-90.7466812133789, 103.52926635742188, -57.17771911621094)
spacing:
(0.3515625, 0.3515625, 2.5)
(0.3515625, 0.3515625298023224, 2.4999990463256836)
direction:
(0.9999999977894414, -6.478192795954982e-09, 6.649148620669932e-05, 2.0651083342197973e-05, -0.9505162149977084, -0.3106749763676334, -6.320324401990454e-05, -0.31067495075469176, 0.950516204300512)
(0.9999999977894414, -6.478192795954982e-09, 6.649148620669932e-05, 2.0651083342197973e-05, -0.9505162149977084, -0.3106749763676334, -6.320324401990454e-05, -0.31067495075469176, 0.950516204300512)
size:
(512, 512, 42)
(512, 512, 42)
```

PR #4143, opened 2023-06-26
`install_sct`: Use explicit paths to python/pip executables

### LINKED ISSUE #3029: Activating conda env doesn't override user's pyenv during install_sct

### LINKED ISSUE #4142: If present, `.local/bin/pip` takes priority over Miniconda `pip` during SCT installation
### Background context

SCT uses Miniconda to provide an isolated Python environment, separate from the user's existing Python installations. 

However, in some cases, an existing Python will still take priority, _even if the conda environment is activated_: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3029

To mitigate this problem, we added a check to ensure that the active `python` interpreter corresponds to the one in `venv_sct`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/117e1cbd6be53db336a5ce16854cfa58abaa402e/install_sct#L617-L624

### Problem

@naga-karthik discovered a [rare corner case](https://stackoverflow.com/a/56889729) where the `python` executable can be correct, but the `pip` executable can be incorrect, rendering the `python` check useless:

```console
> which -a python
/home/<user>/.conda/envs/newenv/bin/python
/usr/bin/python
> which -a pip
/home/<user>/.local/bin/pip   
/home/<user>/.conda/envs/newenv/bin/pip
/usr/bin/pip
```

To fix this issue, we could do the same thing we did in the `python` case (check the executable, throw an error if there is a mismatch).

However, the more I think about it, the more I wonder why we don't just use the explicit path to `python`/`pip` instead:

```diff
-python -m pip install -U "pip!=21.2.*"
+python/envs/venv_sct/bin/python -m pip install -U "pip!=21.2.*"

-pip install -r "$REQUIREMENTS_FILE" --ignore-installed certifi &&
+python/envs/venv_sct/bin/pip install -r "$REQUIREMENTS_FILE" --ignore-installed certifi &&

-pip install -e .
+python/envs/venv_sct/bin/pip install -e .
```

This would ensure that the correct executables are used regardless of whether the conda environment activated correctly. And, it would prevent an ugly error from being thrown, saving us some debugging time with users.

(You could argue that hiding a failed `conda` activation from the user is just kicking the can down the road. But, 99% of users won't be activating SCT's conda environment anyway? So, I don't see much harm in explicitly specifying the correct executables.)

joshuacwnewton:
## Description

This PR ensures that we always invoke the correct `python`/`pip` executables during installation, as there are cases when system `python`/`pip` can be prioritized incorrectly, _even when the conda environment is activated_. (#3029, #4142)

Previously, we tried to mitigate this issue in the `python` case by throwing an error:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/117e1cbd6be53db336a5ce16854cfa58abaa402e/install_sct#L617-L624

We could add a second check for `pip` to fix #4142. But, by throwing an error, we halt the installation and create debugging work that is often unnecessary, since even in the case of a mismatch, we can still just access the correct executables directly. So, this PR removes this check/error as well.

And, because we remove the `python` check, we also no longer need the `realpath` MacOS polyfill introduced in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3103.

## Linked issues

Fixes #4142.
Fixes #3029. (Previously fixed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3103)

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -599,7 +599,7 @@ set -u # reactivate safeties
 
 # Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
 if [[ $OS == linux ]]; then
-  python/bin/conda install -y -c conda-forge libstdcxx-ng
+  conda install -y -c conda-forge libstdcxx-ng
 fi
 
 # Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
--- a/install_sct
+++ b/install_sct
@@ -625,11 +625,11 @@ fi
 
 # Ensure that packaged GLIBC version is up to date (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3927)
 if [[ $OS == linux ]]; then
-  conda install -y -c conda-forge libstdcxx-ng
+  python/bin/conda install -y -c conda-forge libstdcxx-ng
 fi
 
 # Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
-python -m pip install -U "pip!=21.2.*"
+python/envs/venv_sct/bin/python -m pip install -U "pip!=21.2.*"
 
 ## Install the spinalcordtoolbox into the Conda venv
 print info "Installing Python dependencies..."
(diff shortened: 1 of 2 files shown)


mguaypaq (approved):
I like it! This seems more robust. It's easy to forget that we don't _have_ to rely on `$PATH` for running executables.

There was a stray instance of just calling `python`, so I updated that too in an extra commit.

Thanks for noticing that the macOS polyfill was no longer needed, and cleaning it up!

PR #4150, opened 2023-07-10
Make sure that `$SCT_DIR` is set for both Windows installations and `batch_processing.sh`

### LINKED ISSUE #4146: Change relative PATH for absolute PATH in example script
While helping a user (https://forum.spinalcordmri.org/t/multimodal-registration/1097), I noticed that our batch example script has relative PATH pointing to SCT installation:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e740edf4c8408ffa44ef7ba23ad068c6d07e4b87/batch_processing.sh#L114

Obviously, if SCT PATH does not match this relative PATH, the file won't be found.

So, I suggest we replace the relative PATH with $SCT_DIR.

I am wondering how this has never been an issue before (ie: some users must have run into an issue?), and why this has been implemented in the first place 🤔

joshuacwnewton:
## Description

When we first introduced our Windows installation, we decided not to instruct users to set `$SCT_DIR`, as it was already a lot to ask users to update the `$PATH`, and we figured that `$SCT_DIR` wasn't entirely necessary for SCT to run.

That said, having `$SCT_DIR` set makes things a _lot_ easier (in terms of the SCT course, tutorials, `batch_processing.sh`, etc.), since it means we can more easily access assets from within the SCT repo (namely the PAM50 template). Plus, having consistent instructions for all OSs will help a lot when actually teaching the SCT course.

So, this PR brings the Windows installation back in line with Linux/macOS installations, and updates `batch_processing.sh` accordingly so that it can use `$SCT_DIR`, fixing #4146.

## Linked issues

Fixes #4146.

### DIFF
--- a/batch_processing.sh
+++ b/batch_processing.sh
@@ -58,7 +58,7 @@ start=$(date +%s)
 if [[ "$SCT_BP_DOWNLOAD" == "1" ]]; then
   sct_download_data -d sct_example_data
 fi
-cd data/sct_example_data
+cd "$SCT_DIR/data/sct_example_data"
 
 
 # t2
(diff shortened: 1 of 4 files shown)


batch_processing.sh
```diff
@@ -58,7 +58,7 @@ start=$(date +%s)
 if [[ "$SCT_BP_DOWNLOAD" == "1" ]]; then
   sct_download_data -d sct_example_data
 fi
-cd data/sct_example_data
+cd "$SCT_DIR/data/sct_example_data"
```
mguaypaq, line 61: Could we also get rid of these two lines near the bottom?
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4a5e9edb25501c11ca76d55cdbb9156bba94b51f/batch_processing.sh#L258-L259

Since the script now does `cd` to an absolute path, there's no meaningful sense in which this is going "back".

(Also, normally scripts get executed in a subshell, which means that any `cd` in the script doesn't affect the user's current path after the script exits. (An exception would be if the script is run via the shell builtin command `source`.))

mguaypaq (approved):
Looks good, thanks!

batch_processing.sh
joshuacwnewton: Oh, oops! My Ctrl+F missed an instance of `../../..`. Thanks for catching this! :hearts:

joshuacwnewton: Ah! It looks like removing the `cd ../../..` revealed another spot where relative paths were being used: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8a40bc28c5ba61d2de8991c0eadc398a9aa7c220/batch_processing.sh#L270-L271

I'll make sure to remove the `.` and replace them with `$SCT_DIR`

joshuacwnewton: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4150/commits/4ab2471931e01c66530471171f255190050e919a. :)

PR #4153, opened 2023-07-11
Use a more appropriate filename for the first QC image for `sct_image -stitch`

### LINKED ISSUE #4152: QC report for `sct_image -stitch` puts the wrong filepath in generated YML files
When running our existing test for `sct_image -stitch`, then opening the QC report, then marking the image as FAIL and downloading the YML file, the following is observed:

```yml
FILES_REG:
    - /home/joshua/repos/spinalcordtoolbox/data/sct_testing_data/concatenated_input_images.nii.gz
```

However, this is the wrong image path. Instead of the stitched image being listed as the image path, the concatenated image (created just for QC) is listed instead.

Swapping the order of the images gives us this instead:

```yml
FILES_REG:
    - /home/joshua/repos/spinalcordtoolbox/data/sct_testing_data/stitched.nii.gz
```

This is what we want! However, the order of the images was actually intentionally swapped in a previous PR to make the stitching QC report easier to view: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4051

So, we have a bit of a dilemma... either the filepath is displayed correctly, or the images are displayed in a helpful order, but not both? :thinking:

joshuacwnewton:
## Description

This is a small change to fix a small bug. Currently, we are using a generic name for the concatenated images used in the QC report. The problem is, this results in the following QC:

image

And the following YML when exporting QC fails:

```yml
FILES_REG:
    - /home/joshua/repos/spinalcordtoolbox/data/sct_testing_data/concatenated_input_images.nii.gz
```

This generic name is basically useless, since it doesn't tell you which input images were actually used?

So, I've updated the filename so that it shows the original input filenames:

image

This seems more useful in QC? 

But, the only issue is, `t2_roi1-t2_roi2.nii.gz` isn't actually a file -- they're two separate files (`t2_roi1.nii.gz` and `t2_roi2.nii.gz`). Do we want our YML file to point to an actual file for further processing? If so, then maybe we can just use the name of the first input image instead.

## Linked issues

Fixes #4152.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_image.py
+++ b/spinalcordtoolbox/scripts/sct_image.py
@@ -439,7 +439,8 @@ def main(argv: Sequence[str]):
             printv("Generating QC Report...", verbose=verbose)
             # specify filenames to use in QC report
             path_tmp = tmp_create(basename="stitching-qc")
-            fname_qc_concat = os.path.join(path_tmp, "concatenated_input_images.nii.gz")
+            fname_qc_concat = "-".join([os.path.basename(p).split(".")[0] for p in fname_in]) + ".nii.gz"
+            fname_qc_concat = os.path.join(path_tmp, fname_qc_concat)
             fname_qc_out = os.path.join(path_tmp, os.path.basename(fname_out))
             # generate 2 images to compare in QC report
             # (1. naively concatenated input images, and 2. stitched image) padded so both have same dimensions


valosekj:
I have tried the PR with the following code:

```console
$ cd dcm-zurich/sourcedata/sub-250791/anat
$ sct_image -i sub-250791_acq-axialTop_T2w.nii.gz sub-250791_acq-axialBottom_T2w.nii.gz -stitch -qc qc -qc-subject sub-250791 -o sub-250791_stitched.nii.gz
```

Since I used `-o` flag, I would expect that `sub-250791_stitched.nii.gz` will be presented in the html report:

image

joshuacwnewton:
Ack, sorry about that. I confused myself a bit. This PR is incorrect, and I should have stuck with the initial solution (swapping the order of the images).

Terribly sorry for wasting your time here! :sweat:

PR #4156, opened 2023-07-12
Ensure that the `straightening.cache` file is output to the PWD during template registration

### LINKED ISSUE #4138: `straightening.cache` file is output to the working directory, even when `-ofolder` is used
For scripts that have an image-straightening step (e.g. `sct_register_to_template`, `sct_label_vertebrae`), we allow re-use of straightened image files to save time. To facilitate this, we use a `straightening.cache` file.

In `sct_register_to_template`, if `-ofolder` is used, the 3 straightened image files will be output to `-ofolder`, but the `straightening.cache` file will _not_. Instead, it will be created in the PWD.

By comparison, `sct_label_vertebrae` will correctly output all 4 straightening files to the `-ofolder` path.

joshuacwnewton:
## Description

This PR adds a check that tests for the existence of straightening files generated by `sct_register_to_template` and `sct_label_vertebrae`. Notably, this test initially fails for `sct_register_to_template` due to #4138.

Then, this PR updates the cache-saving behavior so that A) the cache file is first saved to the temporary directory used by both scripts, then B) the cache file is copied to the output directory along with the other straightening files.

## Linked issues

Fixes #4138.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_label_vertebrae.py
+++ b/spinalcordtoolbox/scripts/sct_label_vertebrae.py
@@ -326,7 +326,7 @@ def main(argv: Sequence[str]):
             '-r', str(remove_temp_files),
             '-v', '0',
         ])
-        cache_save(os.path.join(path_output, fname_cache), cache_sig)
+        cache_save(fname_cache, cache_sig)
 
     # resample to 0.5mm isotropic to match template resolution
     printv('\nResample to 0.5mm isotropic...', verbose)
@@ -467,6 +467,7 @@ def main(argv: Sequence[str]):
     generate_output_file(os.path.join(path_tmp, "segmentation_labeled.nii"), fname_seg_labeled)
     generate_output_file(os.path.join(path_tmp, "segmentation_labeled_disc.nii"), os.path.join(path_output, file_seg + '_labeled_discs' + ext_seg))
     # copy straightening files in case subsequent SCT functions need them
+    generate_output_file(os.path.join(path_tmp, "straightening.cache"), os.path.join(path_output, "straightening.cache"), verbose=verbose)
     generate_output_file(os.path.join(path_tmp, "warp_curve2straight.nii.gz"), os.path.join(path_output, "warp_curve2straight.nii.gz"), verbose=verbose)
     generate_output_file(os.path.join(path_tmp, "warp_straight2curve.nii.gz"), os.path.join(path_output, "warp_straight2curve.nii.gz"), verbose=verbose)
     generate_output_file(os.path.join(path_tmp, "straight_ref.nii.gz"), os.path.join(path_output, "straight_ref.nii.gz"), verbose=verbose)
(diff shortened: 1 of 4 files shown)


mguaypaq (approved):
This seems fine, although our tests caught a corner case, which I think I fixed.

joshuacwnewton:
Hrm. Just to make sure that I understand why the test failed:

- `sct_label_vertebrae` attempted to re-use some existing straightening files from the PWD.
- But, the "reuse existing straightening file" code only copies the image files, and not the `straightening.cache` file.
- So, when the code later on tries to copy all of the straightening files to the output directory (as per the purpose of this PR), it fails, because `straightening.cache` was never generated in the tmpdir.
- Your fix ensures that the pre-existing `straightening.cache` file in the PWD gets copied to the tmpdir, which will later on get copied from the tmpdir to the output directory.

If that's the case, then the fix from https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4156/commits/b44f463bedc7709f94c9521a626170eb075037e4 makes perfect sense to me. Thank you for this!

(Though, as a side note, I didn't realize that the `sct_label_vertebrae` tests were reusing straightening files from previous tests!! I'm thankful that it caught the bug I introduced, but it means that one of the `sct_label_vertebrae` tests is still saving files to the PWD. Hrm...)

mguaypaq:
> Hrm. Just to make sure that I understand why the test failed: [...]

Yep! That's it exactly.

> (Though, as a side note, I didn't realize that the `sct_label_vertebrae` tests were reusing straightening files from previous tests!! I'm thankful that this behavior caught the bug I almost introduced, but it means that one of the `sct_label_vertebrae` tests is still saving files to the PWD. Hrm...)

I had the same worry, but I don't think that's the case. If you look at the test in question, it's a single test that calls `sct_label_vertebrae` with the same images 3 times, so it makes sense that it would re-use the straightening results (which are the same for all 3 calls).

Wait, actually, you're right that `sct_label_vertebrae` looks for `straightening.cache` in the PWD, so that's still fishy.

joshuacwnewton:
I guess perhaps we could merge this PR as-is for now, but tackle this further as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2911?

i.e. when we change the test suite so that tests are inherently run in a temporary directory, we make sure to explicitly test these scripts' ability to load cached files (as opposed to testing this implicitly, as we're doing now.)

mguaypaq:
Well, we can't really claim to be fixing #4138 unless we understand what's going on here, I think.

joshuacwnewton:
Ah, sorry. I thought we understood what was going on?

- [#4138](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4138) focuses on whether `sct_register_to_template` is outputting the `straightening.cache` file to the `-ofolder` directory.
- But, the new test failures were due to one of the `sct_label_vertebrae` tests not using `-ofolder` at all (presumably the test `test_sct_label_vertebrae_initfile_qc_no_checks`), resulting in all 4 straightening files being output to the PWD (`sct_testing_data`), causing those files to later be re-used when further tests are run.

So, I figured that the former issue had been solved by this PR, while the latter issue was more of a "tests don't clean up after themselves" issue (hence #2911).

mguaypaq:
Ah ok, makes sense.

PR #4052, opened 2023-02-26
Add `-qc-subject` and `-qc-dataset` flags to `sct_image`

valosekj:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR adds `-qc-subject` and `-qc-dataset` flags to `sct_image`.

Note: QC reporting is only available for `sct_image -stitch`

## Linked issues

This PR addresses a comment from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4051:

> Footnote question; sct_image supports only the -qc flag but not -qc-subject?

Issue #4051, opened 2023-02-26
`sct_image -stitch` - QC showing empty slice


valosekj:
### Description

When running `sct_image -stitch` with the flag `-qc` to generate the quality control report, I noticed that the QC is showing different results than the actual output image really is.

Html QC (notice the empty slice between the top and bottom images):

<img width="700" alt="image" src="https://user-images.githubusercontent.com/39456460/221416735-32a5021b-6d3c-4cbf-8b4a-e03fe3c4a8cd.png">

The stitched nii file in FSLeyes (notice that both images are combined fine and there is no empty slice):

<img width="700" alt="image" src="https://user-images.githubusercontent.com/39456460/221416777-c509caef-b680-4e3d-9b33-b055e366954f.png">

### Problem

Due to the empty slice, it is difficult for me to assess the QC and say if the stitching works fine or not. Opening the stitched image in FSLeyes is possible but time-consuming across a large dataset.

### Steps to Reproduce

0. Download example data [here](https://www.dropbox.com/sh/r1c23df6ey6cac1/AAC41SMQ-7UxyepyB3mMdcCGa?dl=0)

1. Run:

```console
sct_image -i sub-294305_acq-axialTop_T2w.nii.gz sub-294305_acq-axialBottom_T2w.nii.gz -stitch -qc qc -o sub-294305_stitched.nii.gz
```

2. Compare QC vs FSLeyes:

```
open qc/index.html
fsleyes sub-294305_stitched.nii.gz
```

---

Footnote question; `sct_image` supports only the `-qc` flag but not `-qc-subject`?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/33e134edf001670725d5fb36a06b8f37245d9580/spinalcordtoolbox/scripts/sct_image.py#L87-L93

joshuacwnewton:
Ah, this is intentional! The QC report has two different views:

1. The original images (separated by an empty slice to clearly separate the original images)
2. The stitched image (no empty slices)

You can toggle the two different views by pressing the right arrow key.

---

That said, if this isn't clear, then we should probably make some sort of change to make it more apparent. (For example, starting on the stitched image, then pressing right to show the original images.)

valosekj:
> You can toggle the two different views by pressing the right arrow key. (That way, you can examine the quality of the stitching by comparing before and after.)

[...]

PR #4157, opened 2023-07-12
Swap order of `sct_image -stitch` QC images to fix incorrect YML path

### LINKED ISSUE #4152: QC report for `sct_image -stitch` puts the wrong filepath in generated YML files

joshuacwnewton:
## Description

This change undoes a previous commit from PR #4052. (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4052/commits/ba8d4d57b82d233a82c482f4fcccb82a11bc06e8)

The purpose of the previous commit was to fix a QC report usability issue (#4051). However, this change inadvertently introduced a second bug: incorrect QC report filepaths (#4152).

Currently, I don't think there is a good way to fix both issues at once, since our QC reporting code is built around the assumption that the first image will be an anat image, and the second image will be the overlay on top of the first image. (But, maybe I'm missing something? Maybe there is a good way to fix both at once...)

Still, if we have to fix one of the two issues, I think that the QC path issue is the more critical of the two? But maybe this is debatable, too. There may be an argument for prioritizing a more usable QC report over having the correct filepath. (Up for discussion!)

## Linked issues

Fixes #4152.
Reopens #4051.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_image.py
+++ b/spinalcordtoolbox/scripts/sct_image.py
@@ -447,7 +447,7 @@ def main(argv: Sequence[str]):
             im_concat.save(fname_qc_concat)
             im_out_padded.save(fname_qc_out)
             # generate the QC report itself
-            generate_qc(fname_in1=fname_qc_concat, fname_in2=fname_qc_out, args=sys.argv[1:],
+            generate_qc(fname_in1=fname_qc_out, fname_in2=fname_qc_concat, args=sys.argv[1:],
                         path_qc=os.path.abspath(arguments.qc), dataset=arguments.qc_dataset,
                         subject=arguments.qc_subject, process='sct_image_stitch')
         else:


valosekj (approved):
Tested using:

```console
$ cd dcm-zurich/sourcedata/sub-250791/anat
$ sct_image -i sub-250791_acq-axialTop_T2w.nii.gz sub-250791_acq-axialBottom_T2w.nii.gz -stitch -qc qc -qc-subject sub-250791 -o sub-250791_stitched.nii.gz
```

Now the filename in HTML QC and YML file is correct! 

image

PR #4161, opened 2023-07-14
Exit on errors anywhere in the linter script

mguaypaq:
Otherwise errors in `git diff` or `grep` are silently ignored.
For example, [it happened in #4156](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/5534105505/jobs/10098608485).

(This is currently a draft PR, so that I can test the result.)


mguaypaq:
This is maybe related to us doing a shallow checkout, hence not having access to the `master` or even `origin/master` revisions? I'll try to reproduce the issue locally for some quicker feedback.

mguaypaq:
Ha, doing `set -eu -o pipefail` accidentally surfaced another mistake, but with the `grep` part of the pipeline: normally `grep` considers the empty list of matches to be an error. Thankfully it has a separate exit code for "no match" (exit code 1) versus other errors (exit code 2).

Anyway, I think the current patch fixes the problems:
- If there's some kind of error with `git diff`, it will noisily fail instead of silently generating an empty list of files to check.
- If there's some kind of error with `grep`, it will noisily fail.
- If the list of files to check legitimately contains no `.py` files, then the check will succeed.

.github/workflows/check-code-quality.yml
```diff
@@ -16,8 +16,17 @@ jobs:
...
+    # We allow (($? == 1)) after grep to allow the empty list as a result (but not (($? == 2)), which would be some other error with grep).
+    # We specify --no-run-if-empty so that flake8 doesn't check the whole repo when no .py files got changed.
     - name: Lint Python files with flake8
-      run: git diff --name-only --diff-filter=d --merge-base master | grep '\.py$' | xargs --delimiter='\n' --no-run-if-empty flake8
+      run: |
+        set -eu -o pipefail
+        git diff --name-only --diff-filter=d \
+          --merge-base refs/remotes/origin/master -- \
+          | (grep '\.py$' || (($? == 1)) ) \
+          | xargs --delimiter='\n' --no-run-if-empty flake8
```
joshuacwnewton, line 28: So, IIUC, the changes here are threefold:

1. Adding `set -eu -o pipefail`
   - `set -o pipefail` makes perfect sense, for exactly the reason you mention in the comment.
   - But, is `set -eu` necessary? There's only a single command pipeline, so I think `set -e` is redundant with `set -o pipefail`? And there are no variables that we need to worry about being unset (`set -u`).
2. Changing `-merge-base master` to `-merge-base refs/remotes/origin/master --`
   - You theorized that the error might be due to us doing a shallow checkout. Were you able to confirm whether it's actually necessary to explicitly reference `ref/remotes/origin`? If true, then maybe we should update the comment from the earlier LOC: "`        fetch-depth: 0  # Fetches all references, which is needed to 'git diff' with origin/master`". 
   - (_Is_ `fetch-depth: 0` needed if we're explicitly referencing `origin/master`?)
   - Does `--` have a purpose here, when it's followed immediately by `|`? As far as I understand, `--` [means different things](https://stackoverflow.com/a/22750480) to different git commands. Checking the man page for [`git diff`](https://git-scm.com/docs/git-diff), it seems to only be used as a separator for additional arguments.
3. Changing `grep '\.py$'` to `(grep '\.py$' || (($? == 1)) )`
   - Again, your comment here makes perfect sense, based on the [Exit Status](https://www.gnu.org/software/grep/manual/html_node/Exit-Status.html) page. ((Returning `true` if the error code is 1.))

joshuacwnewton (approved):
I have some short feedback, but the changes are so minimal and nitpicky that I'm pre-approving. :)

.github/workflows/check-code-quality.yml
joshuacwnewton: Thank you for clarifying! This all seems reasonable to me. :)

PR #4164, opened 2023-07-19
Add prototype for generating JSON sidecar files using `launcher.py`

### LINKED ISSUE #3394: Output JSON sidecar files after processing a la BIDS
In the context processing pipeline and databases, in order to keep track of the software and version used to generate derivatives, it would be useful if SCT would output a JSON sidecar (following [BIDS](https://bids.neuroimaging.io/) convention). E.g.:
```
sct_deepseg_sc -i -FILE.nii -c t2
```
Would output:
```
FILE_seg.nii
FILE_seg.json
```
And the content of the JSON could be the following (inspired by [this link](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html)):
```json
{
  "Name": "FMRIPREP Outputs",
  "BIDSVersion": "1.4.0",
  "DatasetType": "derivative",
  "GeneratedBy": [
    {
      "Name": "sct",
      "Version": "5.3.0",
    },
    {
      "Author": "John Doe",
      "Description": "Manual correction of spinal cord segmentation",
      "Date": "2021-05-10 10:10:28",
    }
  ],
  "SourceDatasets": [
    {
      "DOI": "XXX",
      "URL": "data.neuro.polymtl.ca:datasets/uk-biobank-processed",
      "Version": "1.0.1"
    }
  ]
}
```

Another useful link: https://hackmd.io/@effigies/bids-derivatives-readme

joshuacwnewton:
## Description

This PR:

- Adds a new function called `generate_json_sidecar`, which follows the template laid out in @jcohenadad's comment [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3394#issue-894485189)).
- Updates `compat/launcher.py` so that SCT scripts are run in-process, allowing us to fetch return values (such as the output folder and output files).
- Updates `sct_deepseg_sc` to pass the necessary return values, which in turn enables a JSON sidecar file to be created.

To test, just run `sct_deepseg_sc -c t2 -i t2.nii.gz` in the `sct_testing_data/t2/` folder. 

As a result, a file called `t2_seg.json` will be created alongside `t2_seg.nii.gz` with the following contents:

```json
{
    "Name": "SCT Outputs",
    "BIDSVersion": "1.4.0",
    "DatasetType": "derivative",
    "GeneratedBy": [
        {
            "Name": "sct_deepseg_sc",
            "Version": "git-jn/3394-sidecar-launcher.py-prototype-00774fbb288a500150c8043d581702953352c4b3",
            "Date": "2023-07-19 11:28:08"
        }
    ]
}
```

## Linked issues

Fixes #3394.

### DIFF
--- a/spinalcordtoolbox/compat/launcher.py
+++ b/spinalcordtoolbox/compat/launcher.py
@@ -9,26 +9,26 @@ import sys
 import os
 import subprocess
 import multiprocessing
+import importlib
 
 from spinalcordtoolbox import __file__ as package_init_file
+from spinalcordtoolbox.utils.sys import init_sct
 
 
-def main():
+def main(argv=None):
     """
     Compatibility entry point to run scripts
     """
-
-    # Force scripts to not use graphical output
-    env = dict()
-    env.update(os.environ)
+    if argv is None:
+        argv = sys.argv[1:]
 
     if "ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS" not in os.environ:
-        env["ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS"] = str(multiprocessing.cpu_count())
+        os.environ["ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS"] = str(multiprocessing.cpu_count())
 
     # Needed to allow `sct_check_dependencies` to import voxelmorph/neurite without
     # failing due to a missing `tensorflow` dependency (since the backend defaults to TF)
-    env['VXM_BACKEND'] = 'pytorch'
-    env['NEURITE_BACKEND'] = 'pytorch'
+    os.environ['VXM_BACKEND'] = 'pytorch'
+    os.environ['NEURITE_BACKEND'] = 'pytorch'
 
     command = os.path.basename(sys.argv[0])
     pkg_dir = os.path.dirname(package_init_file)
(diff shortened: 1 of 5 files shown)


spinalcordtoolbox/compat/launcher.py
```diff
@@ -9,39 +9,47 @@
...
-
     mpi_flags = os.environ.get("SCT_MPI_MODE", None)
     if mpi_flags is not None:
         if mpi_flags == "yes":  # compat
             mpi_flags = "-n 1"
-        cmd = ["mpiexec"] + mpi_flags.split() + cmd
-
-    return subprocess.run(cmd, env=env).returncode
+        cmd = ["mpiexec"] + mpi_flags.split() + [sys.executable, script] + argv
+        return subprocess.run(cmd, env=os.environ).returncode
```
joshuacwnewton, line 45: Unfortunately, this method doesn't generate a JSON sidecar when SCT is run in `MPI_MODE`, as a subprocess appears to be necessary to run the [`mpiexec`](https://www.mpich.org/static/docs/v3.4.x/www1/mpiexec.html) executable.

Maybe there's a way we can get around this? But perhaps it's just easier to ditch the `launcher.py` idea, and instead call `generate_json_sidecar` directly from each scritp.

joshuacwnewton:
For now, I think I'm going to close this PR in favor of the alternate idea for generating JSON sidecar files: An explicit function call at the end of each script.

The main reason being that, in this PR, switching from subprocess to in-process appears to be a rather significant change that has far-reaching consequences beyond just JSON sidecar generation.

IMO this is definitely an interesting prototype, though, and is potentially worth revisiting in the future for the other benefits (in-process speedup, moving setup/teardown steps to launcher, explicitly specifying output files at the end of each script's `main()` function, etc.)

Issue #4070, opened 2023-03-16
Reconsider the default location for CLI script output files


joshuacwnewton:
A user on the forum has raised a concern that prompted the following discussion:

> > **[kuech](https://forum.spinalcordmri.org/u/kuech)**: On a side note, sct_straighten_spinalcord mv or cp files from the /tmp into the current directory ./. I’m not sure if this is intentional but it is a bit weird, since I usually run the bash scripts from a separate folder, to have these files pop up there or even in the home directory.
>
> **[jcohenadad](https://forum.spinalcordmri.org/u/jcohenadad)**: This is an excellent comment, which we have been discussing internally for the past few years. Some users prefer to have outputs ‘locally’ (ie: `curdir`), but some, like yourself, don’t work locally and as a result having files created locally is indeed impractical. We will have another internal discussion, and we might end up opting for outputting every file at the location of the `-i` file by default.

In summary, there are two choices here:

1. Output files to the location where the script was called (i.e. the current working directory).
2. Output files to the location of the input image (`-i`). 
    - Note: Often the input image is located in the working directory, meaning that 2 == 1 in many cases.

We may want to consider choosing one or the other by default, then standardizing on it across our CLI scripts.

joshuacwnewton:
Some notes:

- Many (but not all) scripts include an `-ofolder` option, allowing users to customize the output directory.
- The vast majority of scripts will use `.` or `os.getcwd()` as the default value for `-ofolder`.
- However, there is one exception: In `sct_propseg`, if `-ofolder` isn't specified, then the parent folder of the **output image file** (?) will be used as the output folder. 
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/05c6162db1f422ff2831d8a812d16fe1b289ab93/spinalcordtoolbox/scripts/sct_propseg.py#L451-L454

mguaypaq:
I thought `-ofolder` was only used for scripts which may produce several output files, with `-o` being sufficient for the scripts which produce a single output file? Although these scripts somehow have both options:
```
sct_deepseg_sc.py
sct_detect_pmj.py
sct_propseg.py
sct_register_multimodal.py
sct_straighten_spinalcord.py
```

I'd like to clarify the interactions between `-ofolder` and `-o`, and default/absolute/relative paths. In isolation, I think you're suggesting this behaviour for `-ofolder`:
- If `-ofolder` is an absolute path, use it.
[...]

PR #4170, opened 2023-07-24
Remove parent parts from `fname_out` to fix buggy `-o` behavior

### LINKED ISSUE #4167: `sct_propseg` issue with '.' in command line
### Description

There is a problem with `sct_propseg`when using './folder/data/image.nii.gz' for the output. The '.' is the reason for the problem. 

### Steps to Reproduce

0. Install SCT 6.0
1. Perform the following command on Joplin : 
  `sct_propseg -i sub-edm143_ses-M0_PSIR.nii.gz -c t1 -o ./test/sub-edm143_ses-M0_PSIR_seg.nii.gz`

More information can be found [here](https://github.com/ivadomed/canproco/issues/37#issuecomment-1646115232)

I also noted that, by iterating over files in a Python script, the following poses a problem `/home/GRAMES.POLYMTL.CA/` as it contains '.' in the file name. 

**Expected behavior:** 

I expected to get the output of the segmented image

**Actual behavior:** 

Segmentation incomplete because of output folder

joshuacwnewton:
## Description

With command:

```
sct_propseg -i ./t2.nii.gz -c t2 -o ./test/t2_seg.nii.gz
```

We see the following behavior:

- `-o` becomes `fname_out` (but isn't processed in any way to remove anything before the filename)
- `folder_output` is derived from `fname_out` using `str(pathlib.Path(fname_out).parent)`
- Then, `folder_output` is combined with the unprocessed `fname_out`, leading to `'test/./test/t2_seg.nii.gz'`

To fix this, we process `fname_out` to only use the actual filename, since subfolders are captured within `folder_output`.

## Linked issues

Fixes #4167.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_propseg.py
+++ b/spinalcordtoolbox/scripts/sct_propseg.py
@@ -441,6 +441,7 @@ def propseg(img_input, options_dict):
         folder_output = os.path.abspath(arguments.ofolder)
     else:
         folder_output = str(pathlib.Path(fname_out).parent)
+        fname_out = pathlib.Path(fname_out).name
     if not os.path.isdir(folder_output) and os.path.exists(folder_output):
         logger.error("output directory %s is not a valid directory" % folder_output)
     if not os.path.exists(folder_output):


mguaypaq (approved):
This is a nicely surgical quick fix, thanks.

I spent quite some time this afternoon thinking about the various corner cases in the existing logic, and at various points I thought I had suggestions, but in the end I think this is the correct quick fix.

And now that SCT v6.0 is out, I hope we can reach a conclusion on #4070, and do a more comprehensive fix for 6.1 or 6.2, since the current state of `-i`, `-o` and `-ofolder` is confusing users.

valosekj:
Thanks for the quick fix!

Issue #4116, opened 2023-05-15
Installation on Compute Canada fails due to missing steps in experimental pip-based installation method


rohanbanerjee:
### Description

I have setup the [template generation pipeline](https://github.com/neuropoly/template/tree/nb/updated_all) in Compute Canada. I have installed SCT on it using the guidelines provided [here](https://spinalcordtoolbox.com/user_section/installation/mac.html?highlight=install#option-3-install-with-pip-experimental). A part of the template generation pipeline is [straightening](https://github.com/neuropoly/template/blob/cf6b81be7e8a1a591a159b335c5f19aa14ff6f9d/preprocessing.py#L527) the spinal cord and when the pipeline is reaching that step, it is running until the warping field generation step but running into an error when it's trying to apply the transformation to the input image. Attaching the error traceback below:
<img width="1605" alt="Screen Shot 2023-05-15 at 4 26 48 PM" src="https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/25586344/1315a5b4-540e-4a09-80ce-e5aa6dc58612">

I am using the 3.8 version of python on CC and the latest sct version. 
  
Any insight would be helpful, TIA

jcohenadad:
- Inconsistency in the code: https://github.com/neuropoly/template/issues/19
- if you are using system call (see previous bullet point), i suggest you install SCT from package (option 1) or dev (option 2). 
- When reporting terminal output, I suggest to copy/paste the terminal content instead of uploading a screenshot (within which ASCII chars search is not possible)
- "I am using the 3.8 version of python on CC and the latest sct version.": please be always specific, as "the latest" changes every time there is a new version

joshuacwnewton:
> I have installed SCT on it using the guidelines provided [here](https://spinalcordtoolbox.com/user_section/installation/mac.html?highlight=install#option-3-install-with-pip-experimental).

Seeing as you have installed SCT via pip ("experimental"), this is likely the problem.

jcohenadad:
@rohanbanerjee just to follow up on this: in the end, were you able to properly install (and use) SCT?

joshuacwnewton:
Given that @SomeoneInParticular has reworked the installation instructions for Compute Canada (Apptainer) in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4857, as well as moved the experimental pip installation to the very bottom of the page, I'm going to close this as completed.

[...]

PR #3821, opened 2022-06-28
Move CSA normalization models to `./data/` at the root of the SCT repo

jqmcginnis:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR fixes issue #3558, c.f. linked issues.

## Linked issues

Resolves #3558 by changing the path from:
```
path_model = os.path.join(__sct_dir__, 'spinalcordtoolbox', 'data', 'csa_normalization_models',
                                  '_'.join(sorted(data_subject.columns)) + '.csv')
```
To:
```
path_model = os.path.join(__sct_dir__, 'data', 'csa_normalization_models',
                                  '_'.join(sorted(data_subject.columns)) + '.csv')
```

PR #4171, opened 2023-07-24
Inline small `.yml` "model" files

### LINKED ISSUE #3903: Inline small models
As part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2669, several of the trained models we use are actually very small, and could be bundled directly into SCT:

* https://github.com/spinalcordtoolbox/pmj_models
* https://github.com/spinalcordtoolbox/c2c3_disc_models
* https://github.com/spinalcordtoolbox/optic_models/tree/master/src/spinalcordtoolbox/data/svm/optic

Move these files into `spinalcordtoolbox/models/{pmj,c2c3_disc,optic}`, and drop them from the `sct_download_data` line. This will reduce our dependency on maintaining that script.

joshuacwnewton:
## Description

This PR is a follow-up to https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3821. For that PR, we decided to store the `csa_normalization_models` .csv files directly in the `data/` folder, rather than downloading them to the `data/` folder on installation.

The same purpose is applied here to the existing `.yml` model files.

There is maybe some discussion to be had about the [`linux.rst`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/documentation/source/user_section/installation/linux.rst#option-5-hard-core-installation-less-sct-usage) page: Do we bother updating the list of models? Or, do we just remove the "experimental" instructions, because of the difficulty in maintaining them? Related:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4116

## Linked issues

Fixes #3903.


mguaypaq (approved):
LGTM, although the third commit message should say `pmj_models` (which can just be fixed when doing "squash and merge").

Issue #4005, opened 2023-01-20
Refactor the `aggregate_per_slice_or_level` function to make it easier to understand and modify


joshuacwnewton:
[`aggregate_per_slice_or_level`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/377bfb88a9845bb8348e760aea5d70bb4f4bcb42/spinalcordtoolbox/aggregate_slicewise.py#L233-L366) is a lengthy function with many nested conditionals. This complexity makes it difficult to understand and make changes to. Even once a change has been proposed, reviewers then have to parse this complicated nest of conditionals to understand the impact that the change has on the rest of the function. This makes a relatively simple change like #4004 a headache to review.

Additionally, `aggregate_per_slice_or_level` is a core component of `sct_extract_metric` and `sct_process_segmentation`, and these functions have been an active focus for quite a lot of past/present/future changes to SCT:

- Adding new features/metrics to `sct_process_segmentation`
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3888
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3946
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3977
- `sct_process_segmentation`/`sct_analyze_lesion` refactor effort
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2677
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378
  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2707

Considering that this is a high-activity part of SCT, it's worth taking account the cost of [tech debt](https://www.maxcountryman.com/articles/a-framework-for-prioritizing-tech-debt):

> Is this part of the system active or dormant–are we building new features on it today or is it in maintenance mode?
> 
>  * If it's dormant, how much time is spent on maintenance each engineering cycle?
>  * If it's active, take note of that: you're building on a foundation that may not be stable.

So, I think `aggregate_per_slice_or_level` would be a good candidate for refactoring to make it easier to work with.

joshuacwnewton:
[...]

PR #4172, opened 2023-07-26
Use pandas for `.csv` saving to correctly merge existing output metric columns

### LINKED ISSUE #4169: `sct_compute_compression.py` - do not append different metrics into the same file
When running [`sct_compute_compression.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_compute_compression.py) function multiple times for different metrics (for example once for `area` and second time for `diameter_AP`), the function appends the output into the same CSV file, for example:

```console
sct_compute_compression -i sub-tokyoIngenia02_T2w_seg-manual.nii.gz -vertfile sub-tokyoIngenia02_T2w_seg-manual_labeled.nii.gz  -l sub-tokyoIngenia02_T2w_label-compression.nii.gz -normalize-hc 1 -metric area
sct_compute_compression -i sub-tokyoIngenia02_T2w_seg-manual.nii.gz -vertfile sub-tokyoIngenia02_T2w_seg-manual_labeled.nii.gz  -l sub-tokyoIngenia02_T2w_label-compression.nii.gz -normalize-hc 1 -metric diameter_AP
```

image

This does not make sense because column names correspond to the metric, which was computed first (`area` in our case). We should not append different metrics into a single CSV file. A possible solution would be to check the column names and either add new columns or create a new CSV file.

Here is the zip file with files to reproduce the above-mentioned commands:
[sub-tokyoIngenia02_T2w.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/12148124/sub-tokyoIngenia02_T2w.zip)

joshuacwnewton:
## Description

Currently, the existing behavior for saving `sct_compute_compression` CSV files doesn't account for differing column names. So, if a user runs `sct_compute_compression` more than once with different metrics, new columns won't be created -- the data will be incorrectly appended to the existing columns.

So, this PR adds a test for that use-case, and then uses Pandas to correctly merge two different tables with different columns. We also use Pandas to combine rows that have the same filename/compression level/slice, eliminating any duplicate rows with `NaN` values.

((Side note: One day, I would like to use the exact same simplification on `aggregate_per_slice_or_level` too. See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4005#issuecomment-1398737874 for my thoughts.))

## Linked issues

Fixes #4169.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -497,11 +497,6 @@ def save_df_to_csv(dataframe, fname_out):
     """
     Save .csv file of MSCC results.
     :param fname_out:
-    :param level: int: Level of compression.
-    :param metric: str: metric to perform normalization
-    :param metric_ratio: float:
-    :param metric_ratio_norm:
-    :param filename: str: input filename
     :return:
     """
     if os.path.isfile(fname_out):
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -507,7 +507,7 @@ def save_csv(fname_out, level, slices, metric, metric_ratio, metric_ratio_PAM50,
     """
     if not os.path.isfile(fname_out):
         with open(fname_out, 'w') as csvfile:
-            header = ['filename', 'compression_level', 'slice(I->S)', metric + '_ratio', metric + '_ratio_PAM50', metric + '_ratio_PAM50_normalized']
+            header = ['filename', 'compression_level', 'slice(I->S)', f'{metric}_ratio', f'{metric}_ratio_PAM50', f'{metric}_ratio_PAM50_normalized']
             writer = csv.DictWriter(csvfile, fieldnames=header)
             writer.writeheader()
     with open(fname_out, 'a') as csvfile:
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -512,7 +512,7 @@ def save_df_to_csv(dataframe, fname_out):
         dataframe = dataframe.groupby(
             [dataframe.iloc[:, 0], dataframe.iloc[:, 1], dataframe.iloc[:, 2]]
         ).max()  # Use max for a tiebreaker (this should never happen unless the user computes the same metric twice)
-    dataframe.to_csv(fname_out, index=False)
+    dataframe.to_csv(fname_out, na_rep='n/a', index=False)
 
 
 def main(argv: Sequence[str]):
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -9,7 +9,6 @@
 import sys
 import os
 import numpy as np
-import csv
 import logging
 from typing import Sequence
 import pandas as pd
@@ -494,7 +493,7 @@ def metric_ratio_norm(metrics_patients, metrics_HC):
     return metric_ratio(ma, mb, mi)
 
 
-def save_csv(fname_out, level, slices, metric, metric_ratio, metric_ratio_PAM50, metric_ratio_norm, filename):
+def save_df_to_csv(dataframe, fname_out):
     """
     Save .csv file of MSCC results.
     :param fname_out:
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -190,8 +189,8 @@ def get_verterbral_level_from_slice(slices, df_metrics):
                              f"Check vertebral labeling file.")
     level_slice_dict = {}
     # TODO adjust for multiple slices for one compresssion (that can have multiple levels too)
-    #slices_combined = []
-    #for slice in slices:
+    # slices_combined = []
+    # for slice in slices:
```
joshuacwnewton, line 193: Looks like @mguaypaq's PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4161 has caught a [linting issue](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/5671517730/job/15368809512) that slipped through!

(Thanks again for fixing the linting workflow! :hearts:)

sandrinebedard (approved):
I tested the following options and it works well!


```
sct_compute_compression -i t2_compressed_seg.nii.gz -vertfile t2_compressed_seg_labeled.nii.gz -l t2_compressed_labels-compression.nii.gz -metric diameter_AP -normalize-hc 0 -o ap_ratio_test2.csv
```
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/sb199/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/sb199/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#0563C1" vlink="#954F72">



filename | compression_level | slice(I->S) | diameter_AP_ratio | diameter_AP_ratio_PAM50 | diameter_AP_ratio_PAM50_normalized
-- | -- | -- | -- | -- | --
t2_compressed_seg.nii.gz | 3.0 | 99 | 3.3493565531077407 |   |  
t2_compressed_seg.nii.gz | 4.0 | 118 | 11.314949666366614 |   |  
t2_compressed_seg.nii.gz | 5.0 | 140 | 10.780650030491945 |   |  
t2_compressed_seg.nii.gz | 6.0 | 159 | 21.88085837700506 |   |  



</body>

</html>


Now changing `-metric` for `area`

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/sb199/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/sb199/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#0563C1" vlink="#954F72">



filename | compression_level | slice(I->S) | diameter_AP_ratio | diameter_AP_ratio_PAM50 | diameter_AP_ratio_PAM50_normalized | area_ratio | area_ratio_PAM50 | area_ratio_PAM50_normalized
-- | -- | -- | -- | -- | -- | -- | -- | --
t2_compressed_seg.nii.gz | 3.0 | 99 | 3.3493565531077407 |   |   | -19.129842145003085 |   |  
t2_compressed_seg.nii.gz | 4.0 | 118 | 11.314949666366614 |   |   | -7.845701761693569 |   |  
t2_compressed_seg.nii.gz | 5.0 | 140 | 10.780650030491945 |   |   | -7.187642897256286 |   |  
t2_compressed_seg.nii.gz | 6.0 | 159 | 21.88085837700506 |   |   | 16.203723269553862 |   |  



</body>

</html>

valosekj (approved):
Working for me as well! Thanks for the quick implementation!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -505,15 +504,15 @@ def save_csv(fname_out, level, slices, metric, metric_ratio, metric_ratio_PAM50,
...
-        csv_writer.writerow(line)
+    if os.path.isfile(fname_out):
+        # Concatenate existing CSV file to the new CSV file
+        dataframe_old = pd.read_csv(fname_out)
+        dataframe = pd.concat([dataframe_old, dataframe], axis=0, ignore_index=True)
+        # Merge rows that have the same 'filename', 'compression_level', and 'slice(I->S)'
+        dataframe = dataframe.groupby(
+            [dataframe.iloc[:, 0], dataframe.iloc[:, 1], dataframe.iloc[:, 2]]
+        ).max()  # Use max for a tiebreaker (this should never happen unless the user computes the same metric twice)
+    dataframe.to_csv(fname_out, index=False)
```
valosekj, line 515: Maybe we could use `na_rep='n/a'` to have `n/a` instead of empty cells.

```suggestion
    dataframe.to_csv(fname_out, na_rep='n/a', index=False)
```

Now:

image

With `na_rep='n/a'`:

image

mguaypaq (approved):
Thanks for the fix, and the added test!

I think if we're careful with our dataframe indexes, we can use [Pandas' builtin `combine_first` method](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.combine_first.html), which is conceptually what we're trying to do. I'll merge this PR first, so that @valosekj can use it today, and propose my changes in a separate PR.

PR #4176, opened 2023-08-01
Use the `pandas.DataFrame.combine_first` method

mguaypaq:
This is a followup suggestion for #4172.

The Pandas builtin method [`combine_first`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.combine_first.html) does exactly what we want: the row and column indexes of the resulting DataFrame is the union of the two input DataFrames.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -25,6 +25,7 @@ logger = logging.getLogger(__name__)
 
 
 NEAR_ZERO_THRESHOLD = 1e-6
+INDEX_COLUMNS = ['filename', 'compression_level', 'slice(I->S)']
 
 
 # PARSER
@@ -500,14 +501,12 @@ def save_df_to_csv(dataframe, fname_out):
     :return:
     """
     if os.path.isfile(fname_out):
-        # Concatenate existing CSV file to the new CSV file
-        dataframe_old = pd.read_csv(fname_out)
-        dataframe = pd.concat([dataframe_old, dataframe], axis=0, ignore_index=True)
-        # Merge rows that have the same 'filename', 'compression_level', and 'slice(I->S)'
-        dataframe = dataframe.groupby(
-            [dataframe.iloc[:, 0], dataframe.iloc[:, 1], dataframe.iloc[:, 2]]
-        ).max()  # Use max for a tiebreaker (this should never happen unless the user computes the same metric twice)
-    dataframe.to_csv(fname_out, na_rep='n/a', index=False)
+        # Combine the data with the existing CSV file.
+        # Rows with the same (filename, compression_level, slice) triple are merged together.
+        # Metric values from the new dataframe take priority over old CSV file values.
+        dataframe_old = pd.read_csv(fname_out, index_col=INDEX_COLUMNS)
+        dataframe = dataframe.combine_first(dataframe_old)
+    dataframe.to_csv(fname_out, na_rep='n/a')
 
 
 def main(argv: Sequence[str]):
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -650,10 +649,12 @@ def main(argv: Sequence[str]):
...
-    df_metric_ratios = pd.DataFrame(rows, columns=['filename', 'compression_level', 'slice(I->S)',
-                                                   f'{arguments.metric}_ratio',
-                                                   f'{arguments.metric}_ratio_PAM50',
-                                                   f'{arguments.metric}_ratio_PAM50_normalized'])
+    metric_columns = [
+        f'{arguments.metric}_ratio',
+        f'{arguments.metric}_ratio_PAM50',
+        f'{arguments.metric}_ratio_PAM50_normalized',
+    ]
+    df_metric_ratios = pd.DataFrame.from_records(rows, index=INDEX_COLUMNS, columns=INDEX_COLUMNS + metric_columns)
```
joshuacwnewton, line 657: I was curious what the difference was between `.from_records()` and simply using the `pd.Dataframe` constructor.

From what I can tell, `from_records()` allows us to select existing columns a a `MultiIndex`? (Whereas `pd.Dataframe`'s `index` argument functions differently -- it's meant to specify the actual index values for each row. Meaning, we'd have to make our own `MultiIndex` and pass it to the constructor.)

`from_records()` seems much simpler/cleaner, so :+1:  from me. :)

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -500,14 +501,12 @@ def save_df_to_csv(dataframe, fname_out):
...
-        # Merge rows that have the same 'filename', 'compression_level', and 'slice(I->S)'
-        dataframe = dataframe.groupby(
-            [dataframe.iloc[:, 0], dataframe.iloc[:, 1], dataframe.iloc[:, 2]]
-        ).max()  # Use max for a tiebreaker (this should never happen unless the user computes the same metric twice)
-    dataframe.to_csv(fname_out, na_rep='n/a', index=False)
+        # Combine the data with the existing CSV file.
+        # Rows with the same (filename, compression_level, slice) triple are merged together.
+        # Metric values from the new dataframe take priority over old CSV file values.
+        dataframe_old = pd.read_csv(fname_out, index_col=INDEX_COLUMNS)
+        dataframe = dataframe.combine_first(dataframe_old)
```
joshuacwnewton, line 508: From what I understand, `combine_first` is basically like [`combine`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.combine.html#pandas.DataFrame.combine), but without the need to specify a tiebreaker function, since the original dataframe will be prioritized every time? Nice!

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -500,14 +501,12 @@ def save_df_to_csv(dataframe, fname_out):
...
     """
     if os.path.isfile(fname_out):
-        # Concatenate existing CSV file to the new CSV file
-        dataframe_old = pd.read_csv(fname_out)
-        dataframe = pd.concat([dataframe_old, dataframe], axis=0, ignore_index=True)
-        # Merge rows that have the same 'filename', 'compression_level', and 'slice(I->S)'
-        dataframe = dataframe.groupby(
-            [dataframe.iloc[:, 0], dataframe.iloc[:, 1], dataframe.iloc[:, 2]]
-        ).max()  # Use max for a tiebreaker (this should never happen unless the user computes the same metric twice)
-    dataframe.to_csv(fname_out, na_rep='n/a', index=False)
```
joshuacwnewton, line 510: I'm happy that we no longer need to use `iloc` (instead explicitly using the column names), and I'm happy we no longer have to specify `index=False` -- we have a proper index now thanks to `.from_records(index=)` + `read_csv(index_col=)`.

joshuacwnewton (approved):
Thank you for this cleanup! This feels much closer to proper pandas usage -- using more appropriate builtins, without the need for hacky `index` workarounds. :hearts:

My comments are just to double-check my own understanding, but I think this PR is good to go as-is. :)

spinalcordtoolbox/scripts/sct_compute_compression.py
mguaypaq: That's it exactly, yes.

spinalcordtoolbox/scripts/sct_compute_compression.py
mguaypaq: Yup! Having both `combine` and `combine_first` available is very convenient.

spinalcordtoolbox/scripts/sct_compute_compression.py
mguaypaq: Yes, I was motivated to do this change when I started messing around with pandas DataFrames in the interactive interpreter to understand #4172 better. Just seeing the index column with 0, 1, 2, ... appearing, and how we had to work around it, made me want to dig a bit deeper.

PR #4194, opened 2023-08-14
Speed up `-remove-reference` by removing unnecessary iteration

### LINKED ISSUE #4191: Very long computation time for sct_label_utils -remove-reference
As detailed in this [comment](https://github.com/ivadomed/canproco/issues/38#issuecomment-1673920787), the computation of ` sct_label_utils -remove-reference ` takes a lot of time. It uses sct version 6.0. 

**Description:** 

As suggested in this [comment](https://github.com/ivadomed/canproco/pull/40#discussion_r1287272491), I tried using `sct_label_utils -remove-reference` to get a common list of vertebral levels in both segmentation file, in order to use them during registration.  I apply it to both images using the other as a reference. I also overwrite the images. 

>os.system('sct_label_utils -i ' + first_vert_output_path + ' -remove-reference ' + second_vert_output_path + ' -o ' + first_vert_output_path)
    t1 = time()
    print("time to remove-reference",t1-t0)
    os.system('sct_label_utils -i ' + second_vert_output_path + ' -remove-reference ' + first_vert_output_path + ' -o ' + second_vert_output_path)
    t2 = time()
    print("time to remove-reference",t2-t1)

However, I am surprised by a very long computation time for the images (which is much longer than the previous method which consisted of loading the data, removing outliers and saving it. 

For the first image: 147 seconds. For the second image: 161 seconds

Terminal output: 
image

I can send you both files if requested.

joshuacwnewton:
## Description

This is a very simple optimization:

- **Before:** Iterate through all possible "src-ref" pairs of labels, comparing their values (`O(n * m)`)
- **After:** Pre-fetch the set of ref labels, then only iterate through the "src" labels (`O(n + m)`)

This works because we only ever access the `value` from the ref labels -- we don't care about the actual coordinates of the ref labels. So, there was no need to compare all of the individual `ref_coord` objects!

## Linked issues

Fixes #4191.

### DIFF
--- a/spinalcordtoolbox/labels.py
+++ b/spinalcordtoolbox/labels.py
@@ -358,12 +358,13 @@ def remove_missing_labels(img: Image, ref: Image):
     :param ref: reference image
     :returns: image with labels missing from reference removed
     """
+    # Get all label values in the reference image
+    ref_values = set(coord.value for coord in ref.getNonZeroCoordinates())
+    # Transfer only the source labels that match values present in the reference image
     out = zeros_like(img)
-
     for src_coord in img.getNonZeroCoordinates():
-        for ref_coord in ref.getNonZeroCoordinates():
-            if src_coord.value == ref_coord.value:
-                out.data[src_coord.x, src_coord.y, src_coord.z] = src_coord.value
+        if src_coord.value in ref_values:
+            out.data[src_coord.x, src_coord.y, src_coord.z] = src_coord.value
     return out
 
 


mguaypaq (approved):
Oh wow that should be a huge speed improvement, yes! (And I suspect @NathanMolinier would appreciate getting rid of a nested loop :wink:)

NathanMolinier:
> Oh wow that should be a huge speed improvement, yes! (And I suspect @NathanMolinier would appreciate getting rid of a nested loop 😉)

PR #4207, opened 2023-08-30
`sct_analyze_lesion` - add function to compute axial damage ratio + minor improvements

valosekj:
#### GitHub

#### PR contents

## Description

This PR:

1. Does minor `sct_analyze_lesion` improvements such as adding docstrings, comments clarification etc.; see commits https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/37d2c6641303e2c6c7a23fae2c1756af656458d8, https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/9b7bf30cbe67b53f1f456fb94f5fbe97eabe9532, https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/1e152cda12e009b609e9037f3ad28165a58bafd5, https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/9b5ee2914e1e8731b2aa776bcabc6d0f2e8521f2
2. Introduce a new function to calculate the maximum axial damage ratio; see commits https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/da9ac4023d4dec1d9bd1d940cd2eba0785eff15a and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4207/commits/3e62a3310a35735635680aebcee0534f46a615f4

The maximum axial damage ratio is a useful measurement in spinal cord injury (SCI) patients and is computed as the ratio of the lesion area divided by the spinal cord area:

image

REF: Smith, A. C. et al. (2021). Axial MRI biomarkers of spinal cord damage to predict future walking and motor function: A retrospective study. Spinal Cord, 59(6), 693-699. [https://doi.org/10.1038/s41393-020-00561-w](https://doi.org/10.1038/s41393-020-00561-w)

---

The PR can be tested, for example, on `sci-zurich` or `sci-colorado` datasets:
(`sct_analyze_lesion` requires lesion mask and spinal cord mask)

```console
$ cd $SCT_DIR
$ git pull
$ git checkout jv/sct_analyze_lesion_improvements
$ sct_analyze_lesion -m <DATA>/sci-colorado/derivatives/labels/sub-5416/anat/sub-5416_T2w_lesion-manual.nii.gz -s <DATA>/sci-colorado/derivatives/labels/sub-5416/anat/sub-5416_T2w_seg-manual.nii.gz
...
Measures on lesion #1...
  (S-I) length : 12.06 mm
  Max. equivalent diameter : 6.7 mm
  Maximum axial damage ratio : 0.45
  Volume : 314.29 mm^3
...
```

---

TODO:
 - [ ] add test(s)


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -218,12 +218,23 @@ def pack_measures(self):
...
+        Print total results to CLI
+        """
+
         printv('\n\nAveraged measures...', self.verbose, 'normal')
-        for stg, key in zip(['  Volume [mm^3] = ', '  (S-I) Length [mm] = ', '  Equivalent Diameter [mm] = '], ['volume [mm3]', 'length [mm]', 'max_equivalent_diameter [mm]']):
-            printv(stg + str(np.round(np.mean(self.measure_pd[key]), 2)) + '+/-' + str(np.round(np.std(self.measure_pd[key]), 2)), self.verbose, type='info')
 
-        printv('\nTotal volume = ' + str(np.round(np.sum(self.measure_pd['volume [mm3]']), 2)) + ' mm^3', self.verbose, 'info')
-        printv('Lesion count = ' + str(len(self.measure_pd['volume [mm3]'].values)), self.verbose, 'info')
+        for label, key in zip(['Volume [mm^3]', '(S-I) Length [mm]', 'Equivalent Diameter [mm]'], self.measure_keys):
```
joshuacwnewton, line 228: We could maybe simplify things a little more by making `self.measure_keys` equivalent to the labels, so that all we have to do is iterate through `self.measure_keys` (acting as both the labels and keys)?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -218,12 +218,23 @@ def pack_measures(self):
...
+            mean_value = np.round(np.mean(self.measure_pd[key]), 2)
+            std_value = np.round(np.std(self.measure_pd[key]), 2)
+            measure_info = f'{label} = {mean_value} +/- {std_value}'
```
joshuacwnewton, line 231: Your refactor in 9b5ee2914e1e8731b2aa776bcabc6d0f2e8521f2 is wonderful! Things are much clearer now.

However, I notice the old `label` strings had two spaces (`'  '`) at the start of each label (e.g. `'  Volume [mm^3] = '`). So, if we want the output to be exactly the same, I think we need to do this:

```suggestion
            measure_info = f'  {label} = {mean_value} +/- {std_value}'
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -247,19 +260,61 @@ def _measure_within_im(self, im_lesion, im_ref, label_lst):
...
+        axial_damage_ratio_dict = {}
+        # Get slices with lesion
+        lesion_slices = np.unique(np.where(im_data)[2])
+        for slice in lesion_slices:
+            # Lesion area
+            lesion_area = np.sum(im_data[:, :, slice]) * p_lst[0] * p_lst[1]
+            # Spinal cord area
+            sc_area = np.sum(im_sc_data[:, :, slice]) * p_lst[0] * p_lst[1]
+            # Compute the axial damage ratio slice by slice
+            axial_damage_ratio_dict[slice] = lesion_area / sc_area
```
joshuacwnewton, line 296: I notice that the area computation is based on z slices, and so it doesn't take into account the angle correction like the other functions (length/diameter) do. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/da9ac4023d4dec1d9bd1d940cd2eba0785eff15a/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L320-L325

Is this something we want to consider for consistency sake?

joshuacwnewton (changes requested):
For the most part, this LGTM! 

I just had two small nitpicky comments, and one larger discussion point about angle correction. :)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -217,12 +219,23 @@ def pack_measures(self):
...
+        Print total results to CLI
+        """
+
         printv('\n\nAveraged measures...', self.verbose, 'normal')
-        for stg, key in zip(['  Volume [mm^3] = ', '  (S-I) Length [mm] = ', '  Equivalent Diameter [mm] = '], ['volume [mm3]', 'length [mm]', 'max_equivalent_diameter [mm]']):
-            printv(stg + str(np.round(np.mean(self.measure_pd[key]), 2)) + '+/-' + str(np.round(np.std(self.measure_pd[key]), 2)), self.verbose, type='info')
 
-        printv('\nTotal volume = ' + str(np.round(np.sum(self.measure_pd['volume [mm3]']), 2)) + ' mm^3', self.verbose, 'info')
-        printv('Lesion count = ' + str(len(self.measure_pd['volume [mm3]'].values)), self.verbose, 'info')
+        for label, key in zip(['Volume [mm^3]', '(S-I) Length [mm]', 'Equivalent Diameter [mm]'], self.measure_keys):
```
joshuacwnewton, line 228: It looks like the linting error is because we now use the variable name `label`. However, we're also importing the `label` module:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3e62a3310a35735635680aebcee0534f46a615f4/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L15

So, I think we need to pick a different variable name here. Maybe `for measure_str, key in`?

(Although, we wouldn't need to use a second variable at all if https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4207/files#r1312014325 is implemented, too.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: > I notice that the area computation is based on z slices, and so it doesn't take into account the angle correction like the other functions (length/diameter) do.

This is an excellent point! Thank you for pointing this out. Indeed, I was considering the angle correction. Eventually, I decided not to use it because the `axial damage ratio` is just a ratio of lesion and spinal cord masks.  So I assumed that since both lesion and spinal cord masks would be corrected using the same angle correction, it would not impact the resulting `axial damage ratio`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Good idea! Simplified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4207/commits/68683a37a939904149da14effa08f111da1283c0

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Resolved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4207/commits/68683a37a939904149da14effa08f111da1283c0 --  no need for the second variable

naga-karthik:
Looks good to me! Thanks for implementing this feature @valosekj ! Just a minor question - does this assume that the images are oriented in RPI (because the summation is done along the z-axis) or does the function reorient it internally? 

I am also wondering if -- because these are binarized lesion and SC masks, would using `np.count_nonzero()` be more appropriate? Would it result in a different value compared to `np.sum()`?

valosekj:
> Just a minor question - does this assume that the images are oriented in RPI (because the summation is done along the z-axis) or does the function reorient it internally?

Good question! The `sct_analyze_lesion.py` function has built-in reorientation to RPI:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/68683a37a939904149da14effa08f111da1283c0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L159-L160

And then back to the original orientation:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/68683a37a939904149da14effa08f111da1283c0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L176-L177

> I am also wondering if -- because these are binarized lesion and SC masks, would using `np.count_nonzero()` be more appropriate? Would it result in a different value compared to `np.sum()`?

Good point! Currently, only binary lesion mask is supported:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/68683a37a939904149da14effa08f111da1283c0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L118

I have on my TODO list to extend the functionality also to soft masks.

joshuacwnewton (approved):

joshuacwnewton:
@valosekj Do you need anything from me on this PR? Or, should I wait until you add soft segmentation functionality to re-review?

valosekj:
> @valosekj Do you need anything from me on this PR? Or, should I wait until you add soft segmentation functionality to re-review?

I am currently re-exploring the angle correction we discussed at https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4207#discussion_r1312039247. Sorry, I did not let you know about this.

valosekj:
Okay, after an in-person discussion with @mguaypaq, we concluded that there is really no need for the angle correction since the damage ratio is computed as a slice-wise **_ratio_** of lesion and spinal cord masks.

@joshuacwnewton, since I added a few (mostly minor) commits from your last review, could you please re-review the PR and merge? Thanks!

joshuacwnewton:
LGTM! I just added `max_axial_damage_ratio` to the existing `sct_analyze_lesion` test, but I think it should be good to merge now. :)

PR #4208, opened 2023-08-30
`sct_run_batch` - add note that `~` should not be used in paths passed using the `-script-args` arg

valosekj:
This PR is just a very minor clarification of the `sct_run_batch` `-script-args` arg.

@naga-karthik encountered errors when using `"~"`, `"${HOME}"` fixed the issues. The reason is that the quotation marks in `"~"` prevent tilde expansion.


mguaypaq (approved):
I'm not really sure why `${HOME}` would work in this context without `~` also working, but I'm certainly willing to believe that this is the case. Thanks for the clarification!

Issue #3283, opened 2021-03-14
General discussion about qform/sform and qform_code


jcohenadad:
### Description

There has been a lot of issues related to the NIfTI's qform/sform and qform_code header fields. The purpose of this issue (which is not really an issue) is to centralize all the relevant information and cross-reference related issues.

Here are relevant links:
- https://github.com/ANTsX/ANTs/wiki/How-does-ANTs-handle-qform-and-sform-in-NIFTI-1-images%3F
- https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html
- https://nipy.org/nibabel/nifti_images.html#the-qform-affine
- https://community.mrtrix.org/t/nifti-transform-ambiguity/2254

Related issues:
- https://github.com/neuropoly/spinalcordtoolbox/issues/2454
- https://github.com/neuropoly/spinalcordtoolbox/issues/3281

joshuacwnewton:
I've been thinking about this a lot recently, too, see https://github.com/neuropoly/spinalcordtoolbox/pull/3270#discussion_r593781030.

As a bare minimum, I was thinking that we could come up with a set of checks to see whether the header complies with NIFTI specifications. Then, apply these checks when saving/loading images to catch issues. And, perhaps emit only `DEBUG` messages for now.

But, this would just be for getting an idea of the current state of SCT. (What problems exist, which parts of SCT are affected, etc.) I'm not sure what to _do_ about those problems yet. I'm glad you've shared approaches from other software.

joshuacwnewton:
One of the qform/sform issues linked above (#3232) sparked a discussion thread over at nibabel: https://github.com/nipy/nibabel/issues/1001

That issue contains some (admittedly frank) comments that nonetheless I appreciate hearing.

> The problem boils down to a combination of two coding patterns that are in common use:
> 
> * Using get_fdata to retrieve the image data, so that it is always loaded and manipulated as float64, regardless of the storage type.
> 
> * When creating a new image from an existing image (e.g. after some processing step), using the header from the original image to preserve the NIFTI header metadata (not that there's much of use in there, but that's another matter):

That last comment piqued my interest. Are our current practices frowned upon? I'm curious how the `nibabel` developers expect the `nibabel.Nifti1Image` class to be handled vs. how we're actually handling it.

[...]

PR #4211, opened 2023-08-31
Raise exception if sform/qform don't match during registation

### LINKED ISSUE #4184: Registration to template results in blank images when input anat image has mismatched affines
A user on the forum reported [empty registration results](https://forum.spinalcordmri.org/t/registration-tips/1102/4?u=joshuacwnewton), and @jcohenadad identified [mismatched sform/qform matrices](https://forum.spinalcordmri.org/t/registration-tips/1102/7?u=joshuacwnewton) as a potential cause. I was able to [confirm this](https://forum.spinalcordmri.org/t/registration-tips/1102/8?u=joshuacwnewton), and now I want to identify _why_ the mismatch causes this issue.

So far, I have drawn the following conclusions:

- The issue occurs when checking out the `5.8` tag, so at the very least, this isn't a new issue for v6.0 (last ~6mo of development).
- ~~The issues begin (at the very least) with the straightening step. (I tried using `warp_curve2straight.nii.gz` on the anat image for both cases (matching xforms vs. mismatched xforms), and the output of `sct_apply_transfo` is blank for the mismatched case but correct for the matched case.)~~ This was a red herring. The straightening actually [occurs OK](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4184#issuecomment-1670042547), since the affine mismatch gets fixed internally prior to the straightening step.

Potentially related to: 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2429

joshuacwnewton:
## Description

This PR adds a test to cover the situation described in #4184 (mismatched qform/sform in input anat image). The command in the test should fail to raise an exception on `master`, but 626dd733b419c2ac3870a836c58bf02509f1aa5e should cause the test to pass.

There is more work that could be done (e.g. papering over the mismatch ourselves in `sct_register_to_template`), but I think for now, I want to leave that for #3283, as discussed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4184#issuecomment-1670042547.

## Linked issues

Fixes #4184.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_register_to_template.py
+++ b/spinalcordtoolbox/scripts/sct_register_to_template.py
@@ -406,12 +406,16 @@ def main(argv: Sequence[str]):
 
     # copy files to temporary folder
     printv('\nCopying input data to tmp folder and convert to nii...', verbose)
-    Image(fname_data).save(os.path.join(path_tmp, ftmp_data))
-    Image(fname_seg).save(os.path.join(path_tmp, ftmp_seg))
-    Image(fname_landmarks).save(os.path.join(path_tmp, ftmp_label))
-    Image(fname_template).save(os.path.join(path_tmp, ftmp_template))
-    Image(fname_template_seg).save(os.path.join(path_tmp, ftmp_template_seg))
-    Image(fname_template_labeling).save(os.path.join(path_tmp, ftmp_template_label))
+    try:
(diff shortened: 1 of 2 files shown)


mguaypaq (approved):
This looks fine, thanks for the test and the more stringent input testing.

PR #4218, opened 2023-09-14
Add start date information to the logs

### LINKED ISSUE #4216: `sct_run_batch` - include date in the output log
The `sct_run_batch` wrapper outputs the `log/sct_run_batch_log.txt` file with information when the script started and when the script finished, for example:

```
Started: 13h37m43s | Ended: 14h52m07s | Duration: 01h14m24s
```

However, the information about the date is missing. The date might be useful, for example, when the same analysis is rerun on different days with different configurations/datasets.

mguaypaq:
Fixes #4216, with only 3 extra characters :wink:

From the manual page for `strftime`, the added date information is:
```
       %F     Equivalent to %Y-%m-%d (the ISO 8601 date format). (C99)

       %Y     The  year  as a decimal number including the century.  (The %EY con‐
              version specification corresponds to the full alternative year  rep‐
              resentation.)  (Calculated from tm_year)

       %m     The month as a decimal number (range 01 to  12).   (Calculated  from
              tm_mon.)

       %d     The  day of the month as a decimal number (range 01 to 31).  (Calcu‐
              lated from tm_mday.)
```

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -569,7 +569,7 @@ def main(argv: Sequence[str]):
     # Display timing
     duration = end - start
     timing_message = ('Started: {} | Ended: {} | Duration: {}\n'.format(
-        start.strftime('%Hh%Mm%Ss'),
+        start.strftime('%F %Hh%Mm%Ss'),
         end.strftime('%Hh%Mm%Ss'),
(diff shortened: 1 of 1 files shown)


valosekj (approved):
Thank you for the quick fix. Tested and it is working!

```
Started: 2023-09-14 18h27m25s | Ended: 18h28m40s | Duration: 00h01m15s
```

PR #4219, opened 2023-09-14
Simplify lesion volume computation

### LINKED ISSUE #4214: Simplify total lesion volume calculation
After an in-person discussion with @mguaypaq, we concluded that there is no need for the looping across slices when computing the total lesion volume:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2037cdc135ee891d634b17469a9f0f45d8734624/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L249-L255

Instead, the volume can be simply computed as the sum of non-zero voxels multiplied by the voxel dimensions.

Also note that there is no need for angle correction since we compute the total volume, not area across slices.

mguaypaq:
Fixes #4214.

The `self.volumes` array was only written, never read by anyone else, so there's no need to have it.

And once it's gone, there's no need to sum each slice first, then sum over the slices; we can just sum the whole array at once.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -99,7 +99,7 @@ def get_parser():
     return parser
 
 
-class AnalyzeLeion:
+class AnalyzeLesion:
     def __init__(self, fname_mask, fname_sc, fname_ref, path_template, path_ofolder, verbose):
         self.fname_mask = fname_mask
 
@@ -617,12 +617,12 @@ def main(argv: Sequence[str]):
         rm_tmp = True
 
     # create the Lesion constructor
-    lesion_obj = AnalyzeLeion(fname_mask=fname_mask,
-                              fname_sc=fname_sc,
-                              fname_ref=fname_ref,
-                              path_template=path_template,
-                              path_ofolder=path_results,
-                              verbose=verbose)
+    lesion_obj = AnalyzeLesion(fname_mask=fname_mask,
+                               fname_sc=fname_sc,
+                               fname_ref=fname_ref,
+                               path_template=path_template,
+                               path_ofolder=path_results,
+                               verbose=verbose)
 
     # run the analyze
     lesion_obj.analyze()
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -267,12 +264,9 @@ def _measure_volume(self, im_data, p_lst, idx):
         """
         Measure the volume of the lesion
         """
-        for zz in range(im_data.shape[2]):
-            self.volumes[zz, idx - 1] = np.sum(im_data[:, :, zz]) * p_lst[0] * p_lst[1] * p_lst[2]
-
-        vol_tot_cur = np.sum(self.volumes[:, idx - 1])
-        self.measure_pd.loc[idx, 'volume [mm3]'] = vol_tot_cur
-        printv('  Volume : ' + str(np.round(vol_tot_cur, 2)) + ' mm^3', self.verbose, type='info')
```
joshuacwnewton, line 275: The only reason I can think of for why the refactor might not be a 1:1 analog is if f `_measure_volume` was called multiple times, causing `self.volumes` to be repeatedly mutated, with the total volume sum changing across calls. (The variable was called `vol_tot_cur`, too, implying that the "current" sum may evolve over time, too.)

However, while it's true that `_measure_volume` does seem to be called [within a loop](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/02d775aabed88c5d871cf087ab7cfe8f06578acb/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L457-L468), all of the changes to the input arguments at the caller site are dependent on `idx`, and the volume sum is dependent on `idx` too. 

So, I think for the old code, each individual sum ended up completely isolated between calls, despite multiple calls sharing the same object? Meaning that eliminating `self.volumes` is definitely the way to go. :)

joshuacwnewton (approved):
LGTM! Thank you for this simplification. :hearts:

spinalcordtoolbox/scripts/sct_analyze_lesion.py
mguaypaq: > So, I think for the old code, each individual sum ended up completely isolated between calls, despite multiple calls sharing the same object?

That's my understanding, yes. The `vol_tot_cur` is only the sum for for a specific `idx`, namely the one that the function has just completely overwritten.

PR #4221, opened 2023-09-18
Fix `ValueError` by passing a tuple to `color=` instead of a numpy array

### LINKED ISSUE #4220: New test failures: `The truth value of an array with more than one element is ambiguous`
I'm seeing a few test failures for tests on PRs + Scheduled `master` runs. The errors affect all OSs, and started happening 3 days ago. 

```
____________________ test_sct_dmri_display_bvecs_png_exists ____________________

  [...]

  File "/home/runner/sct_6.1.dev0/testing/cli/test_cli_sct_dmri_display_bvecs.py", line 17, in test_sct_dmri_display_bvecs_png_exists
    sct_dmri_display_bvecs.main(argv=['-bvec', sct_test_path('dmri', 'bvecs.txt')])
  File "/home/runner/sct_6.1.dev0/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py", line 176, in main
    ax.scatter(x[i], y[i], z[i], color=shell_colors[bvals[i]], alpha=0.7)
  File "/home/runner/sct_6.1.dev0/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/__init__.py", line 1465, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/home/runner/sct_6.1.dev0/python/envs/venv_sct/lib/python3.9/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 2377, in scatter
    if kwargs.get('color', None):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

I don't think this is due to any sort of change in the codebase, so I'm assuming it's due to an upstream change of some kind. I'll try to pin it down and fix it. :)

joshuacwnewton:
## Description

Previously, we would create a dict of [1,4] numpy arrays representing RGB colors + A (transparency) for each unique bval. Due to a recent matplotlib update, though, using a numpy array causes an unexpected ValueError. (https://github.com/matplotlib/matplotlib/issues/26821)

> This error is what's causing the current CI failures across recent PRs.

Since a tuple of size [4] results in the exact same functionality, we can bypass the error by storing tuple-based colors instead.

## Linked issues

Fixes #4220.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
+++ b/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
@@ -132,7 +132,7 @@ def main(argv: Sequence[str]):
         # skip b=0
         if unique_bval < BZERO_THRESH:
             continue
-        shell_colors[unique_bval] = list(next(colors))
+        shell_colors[unique_bval] = tuple(next(colors))
 
     # Get total number of directions
     n_dir = len(x)
--- a/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
+++ b/spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
@@ -132,7 +132,7 @@ def main(argv: Sequence[str]):
         # skip b=0
         if unique_bval < BZERO_THRESH:
             continue
-        shell_colors[unique_bval] = next(colors)
+        shell_colors[unique_bval] = list(next(colors))
 
     # Get total number of directions
     n_dir = len(x)


spinalcordtoolbox/scripts/sct_dmri_display_bvecs.py
```diff
@@ -132,7 +132,7 @@ def main(argv: Sequence[str]):
         # skip b=0
         if unique_bval < BZERO_THRESH:
             continue
-        shell_colors[unique_bval] = next(colors)
+        shell_colors[unique_bval] = list(next(colors))
```
tacaswell, line 135: A very picky comment would be that `tuple` is better than `list`, but that is getting to be a philosophical point.

Sorry for the regression upstream.

joshuacwnewton: Not a worry! And, thank you for taking the time to look at this downstream PR. :hearts:

mguaypaq (approved):
Thanks for the quick fix!

PR #4201, opened 2023-08-18
Raise error if `sct_run_batch` config file has wrong suffix

valosekj:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [x] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

When working with the `sct_run_batch` script, I incorrectly defined the suffix of the config file to `.sh`. The error raised by the `sct_run_batch` script was not very specific. Thus, I added an error raise if the config file is not YML or JSON.

`master` branch:

```
$ sct_run_batch -config config.sh

--
Spinal Cord Toolbox (git-master-d6aeaef52466129b5f24a56fa250403839bd0166)

sct_run_batch -config config.sh
--

configuring
Traceback (most recent call last):
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 597, in <module>
    main(sys.argv[1:])
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_run_batch.py", line 359, in main
    config_keys = set(config.keys())
UnboundLocalError: local variable 'config' referenced before assignment
```

`jv/raise_error_if_sct_run_batch_config_has_wrong_suffix` branch:

```
$ sct_run_batch -config config.sh

--
[...]

Issue #4166, opened 2023-07-21
`sct_qc` - tweak commands to generate currenly unsuported QC reports


valosekj:
This issue collects several tweak commands for the `sct_qc` function to generate QC reports, which are not implemented yet.

1. Sagittal centerline QC: `sct_qc -p sct_label_vertebrae` - see [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4011#issuecomment-1403828459)
2. Sagittal spinal cord segmentation QC: 
- single slice:`sct_qc -p sct_label_vertebrae` (see [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4011#issuecomment-1561736362))
-  all sagittal slices: `sct_qc -p sct_deepseg_lesion -plane sagittal` (see [here](https://github.com/ivadomed/canproco/issues/37#issuecomment-1644497220))
3. Cropped image QC (`sct_crop_image`): `sct_qc -p sct_image_stitch`, see [here](https://github.com/ivadomed/canproco/issues/37#issuecomment-1644614442)
4. Compression labels QC: `sct_qc -p sct_label_utils` - see [here](https://github.com/spinalcordtoolbox/manual-correction/blob/52d3a75a2a7fb4581aa686e06caae8869bfd5739/manual_correction.py#L290-L292)

valosekj:
Adding one more tweak command for nerve rootlets segmentation (relevant for [ivadomed/model-spinal-rootlets](https://github.com/ivadomed/model-spinal-rootlets) project).

```
sct_qc -i sub-brnoUhb01_T2w.nii.gz -s sub-brnoUhb01_T2w_seg-manual.nii.gz -d sub-brnoUhb01_T2w_label-rootlet.nii.gz -p sct_deepseg_lesion -plane axial -qc qc -qc-subject sub-brnoUhb01
```

![Kapture 2023-07-27 at 14 12 57](https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/39456460/57f658eb-f157-4f6a-a297-fa051a7cb67d)

I also tried to dilate the SC mask to include more CSF (because nerve rootlets are located in CSF), but it seems that it does not have any effect on the QC.

```
sct_maths -i sub-brnoUhb01_T2w_seg-manual.nii.gz -dilate 30 -o sub-brnoUhb01_T2w_seg-manual_dil.nii.gz
[...]

Issue #4122, opened 2023-05-25
Add orientation labels to sagittal mosaic QC


joshuacwnewton:
I would love to add these labels! Looking into it, though, the current implementation makes this really tricky. :sweat:

For Axial mosaics, the [`_add_orientation_label`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a058fc2937cdf9a36a555e79af4420dc52f9e3a7/spinalcordtoolbox/reports/qc.py#L414-L431) method assumes a square size of 30x30, and so it _hardcodes the label locations_ (yikes!):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a058fc2937cdf9a36a555e79af4420dc52f9e3a7/spinalcordtoolbox/reports/qc.py#L423-L426

To be able to generalize this to Sagittal images of varying shapes, we would need to have access to the bounding box dimensions. But... we lose that context once we enter the `_add_orientation_label` method, which is called way after generating the mosaic image.

So, I think in order to properly position the orientation labels, we would need to move the orientation label step inside of the `mosaic` method, which would require a bigger refactor. :sweat:

_Originally posted by @joshuacwnewton in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4102#discussion_r1190017266_

PR #4224, opened 2023-09-21
Restructure QC report code to use `create_qc_entry` context manager (plus 1 PoC: `multimodal`)

mguaypaq:
This is a draft PR, to get early feedback on a proposed code structure for the QC report generation code.

Currently, it only contains a proof of concept for `sct_register_multimodal` (picked at random).

I manually tested that it produces an identical QC report entry for [a command from the tutorial](https://spinalcordtoolbox.com/user_section/tutorials/multimodal-registration/mtr-computation/coregistering-mt0-mt1.html), but in some corner cases the images may be different.

I'll do a self-review to point out a few things first, but after that I'd love to get @joshuacwnewton's feedback, either as comments or through a slack discussion, whatever's easier.

Fixes #4201.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -135,8 +135,8 @@ def create_qc_entry(
                 'contrast': contrast,
                 'fname_in': path_input.name,
                 'plane': plane,
-                'background_img': str(path_background_img.relative_to(path_qc)),
-                'overlay_img': str(path_overlay_img.relative_to(path_qc)),
+                'background_img': str(imgs_to_generate['path_background_img'].relative_to(path_qc)),
+                'overlay_img': str(imgs_to_generate['path_overlay_img'].relative_to(path_qc)),
                 'moddate': mod_date.strftime("%Y-%m-%d %H:%M:%S"),
                 'qc': '',
             }, file_result, indent=1)
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -62,7 +62,7 @@ def locked_file(path: Path):
 
 
 @contextmanager
-def qc_entry(
+def create_qc_entry(
     path_input: Path,
     path_qc: Path,
     command: str,
@@ -100,11 +100,17 @@ def qc_entry(
     # Make sure the image directory exists
     path_img = path_qc / dataset / subject / contrast / command / timestamp
     path_img.mkdir(parents=True, exist_ok=True)
-    path_background_img = path_img / 'background_img.png'
-    path_overlay_img = path_img / 'overlay_img.png'
 
-    # Let the caller generate the image files for the entry
-    yield (path_background_img, path_overlay_img)
+    # Ask the caller to generate the image files for the entry
+    imgs_to_generate = {
+        'path_background_img': path_img / 'background_img.png',
+        'path_overlay_img':  path_img / 'overlay_img.png'
+    }
+    yield imgs_to_generate
+    # Double-check that the images were generated during the 'with:' block
+    for img_type, file_img in imgs_to_generate.items():
+        if not Path.exists(file_img):
+            raise FileNotFoundError(f"Required QC image '{img_type}' was not found at the expected path: '{file_img}')")
 
     # We lock `index.html` so that we halt any other processes *before*
     # they have a chance to generate or read any .json files. This ensures
@@ -191,7 +197,7 @@ def sct_register_multimodal(
     cmdline.extend(argv)
 
     # Axial orientation, switch between two input images
-    with qc_entry(
+    with create_qc_entry(
         path_input=Path(fname_input).absolute(),
         path_qc=Path(path_qc),
         command=command,
@@ -199,7 +205,7 @@ def sct_register_multimodal(
         plane='Axial',
         dataset=dataset,
         subject=subject,
-    ) as (path_background_img, path_overlay_img):
+    ) as imgs_to_generate:
 
         # Resample images slice by slice
         p_resample = 0.6
@@ -245,7 +251,7 @@ def sct_register_multimodal(
         add_orientation_labels(ax)
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
-        img_path = str(path_background_img)
+        img_path = str(imgs_to_generate['path_background_img'])
         logger.debug('Save image %s', img_path)
         fig.savefig(img_path, format='png', transparent=True, dpi=300)
 
@@ -259,7 +265,7 @@ def sct_register_multimodal(
         add_orientation_labels(ax)
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
-        img_path = str(path_overlay_img)
+        img_path = str(imgs_to_generate['path_overlay_img'])
         logger.debug('Save image %s', img_path)
         fig.savefig(img_path, format='png', transparent=True, dpi=300)
 
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -1,0 +1,352 @@
+"""
+Quality Control report generator
+
+Copyright (c) 2023 Polytechnique Montreal <www.neuro.polymtl.ca>
+License: see the file LICENSE
+"""
+
+from contextlib import contextmanager
+import datetime
+import importlib.resources
+from importlib.abc import Traversable
+import json
+import logging
+import math
+import os
+from pathlib import Path
+import string
+from typing import Optional, Sequence
+
+from matplotlib.axes import Axes
+from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
+from matplotlib.figure import Figure
+import matplotlib.patheffects as path_effects
+import numpy as np
+import portalocker
+from scipy.ndimage import center_of_mass
+import skimage.exposure
+
+from spinalcordtoolbox.image import Image
+import spinalcordtoolbox.reports
+from spinalcordtoolbox.resampling import resample_nib
+from spinalcordtoolbox.utils import __version__, display_open, list2cmdline
+
+logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def locked_file(path: Path):
+    """
+    Open and lock a file for reading and/or writing.
+
+    Any other process that tries to lock the file will wait until this lock is released.
+    """
+    # Make sure the file exists before trying to lock it
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.touch()
+
+    # NB: We use 'r+' because it allows us to open an existing file for
+    # locking *without* immediately truncating the existing contents prior
+    # to opening. We can then use this file to overwrite the contents later.
+    file = path.open('r+', encoding='utf-8')
+    portalocker.lock(file, portalocker.LOCK_EX)
+    try:
+        # Let the caller use the open, locked file
+        yield file
+    finally:
+        # Safely release the lock
+        file.flush()
+        os.fsync(file.fileno())
+        portalocker.unlock(file)
+        file.close()
+
+
+@contextmanager
+def qc_entry(
+    path_input: Path,
+    path_qc: Path,
+    command: str,
+    cmdline: str,
+    plane: str,
+    dataset: Optional[str],
+    subject: Optional[str],
+):
+    """
+    Generate a new QC report entry.
+
+    This context manager yields a tuple of two paths, to be used for the QC report images:
+    1. the path to `background_img.png`, and
+    2. the path to `overlay_img.png`.
+
+    The body of the `with` block should create these two image files.
+    When the `with` block exits, the QC report is updated, with proper file synchronization.
+    """
+    if plane not in ['Axial', 'Sagittal']:
+        raise ValueError(f'Invalid plane: {plane!r}')
+
+    logger.info('\n*** Generating Quality Control (QC) html report ***')
+    mod_date = datetime.datetime.now()
+
+    # Not quite following BIDS convention, we derive the value of
+    # dataset, subject, contrast from the input file by splitting it as
+    # {dataset}/{subject}/{contrast}/filename
+    if dataset is None:
+        dataset = path_input.parent.parent.parent.name
+    if subject is None:
+        subject = path_input.parent.parent.name
+    contrast = path_input.parent.name
+    timestamp = mod_date.strftime('%Y_%m_%d_%H%M%S.%f')
+
+    # Make sure the image directory exists
+    path_img = path_qc / dataset / subject / contrast / command / timestamp
+    path_img.mkdir(parents=True, exist_ok=True)
+    path_background_img = path_img / 'background_img.png'
+    path_overlay_img = path_img / 'overlay_img.png'
+
+    # Let the caller generate the image files for the entry
+    yield (path_background_img, path_overlay_img)
+
+    # We lock `index.html` so that we halt any other processes *before*
+    # they have a chance to generate or read any .json files. This ensures
+    # that the last process to write to `index.html` has read in all of the
+    # available .json files, preventing:
+    # https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3701#discussion_r816300380
+    path_index_html = path_qc / 'index.html'
+    with locked_file(path_index_html) as file_index_html:
+
+        # Create a json file for the new QC report entry
+        path_json = path_qc / '_json'
+        path_json.mkdir(parents=True, exist_ok=True)
+        path_result = path_json / f'qc_{timestamp}.json'
+        with path_result.open('w') as file_result:
+            json.dump({
+                'cwd': str(Path.cwd()),
+                'cmdline': cmdline,
+                'command': command,
+                'sct_version': __version__,
+                'dataset': dataset,
+                'subject': subject,
+                'contrast': contrast,
+                'fname_in': path_input.name,
+                'plane': plane,
+                'background_img': str(path_background_img.relative_to(path_qc)),
+                'overlay_img': str(path_overlay_img.relative_to(path_qc)),
+                'moddate': mod_date.strftime("%Y-%m-%d %H:%M:%S"),
+                'qc': '',
+            }, file_result, indent=1)
+
+        # Collect all existing QC report entries
+        json_data = []
+        for path in sorted(path_json.glob('*.json')):
+            with path.open() as file:
+                json_data.append(json.load(file))
+
+        # Insert the QC report entries into index.html
+        path_assets = importlib.resources.files(spinalcordtoolbox.reports) / 'assets'
+        template = string.Template((path_assets / 'index.html').read_text(encoding='utf-8'))
+        # Empty the HTML file before writing, to make sure there's no leftover junk at the end
+        file_index_html.truncate()
+        file_index_html.write(template.substitute(sct_json_data=json.dumps(json_data)))
+
+        # Copy any missing assets
+        path_qc.mkdir(parents=True, exist_ok=True)
+        copy_missing(path_assets / '_assets', path_qc)
+
+    logger.info('Successfully generated the QC results in %s', str(path_result))
+    display_open(file=str(path_index_html), message="To see the results in a browser")
+
+
+def copy_missing(resource: Traversable, destination: Path):
+    """
+    Make sure that a copy of `resource` exists at `destination`,
+    by creating any missing files and directories recursively.
+    """
+    path = destination / resource.name
+    if resource.is_dir():
+        path.mkdir(exist_ok=True)
+        for sub_resource in resource.iterdir():
+            copy_missing(sub_resource, path)
+    elif resource.is_file():
+        if not path.is_file():
+            path.write_bytes(resource.read_bytes())
+    else:
+        # Some weird kind of resource? Ignore it
+        pass
+
+
+def sct_register_multimodal(
+    fname_input: str,
+    fname_output: str,
+    fname_seg: str,
+    argv: Sequence[str],
+    path_qc: str,
+    dataset: Optional[str],
+    subject: Optional[str],
+):
+    """
+    Generate a QC report for sct_register_multimodal.
+    """
+    command = 'sct_register_multimodal'
+    cmdline = [command]
+    cmdline.extend(argv)
+
+    # Axial orientation, switch between two input images
+    with qc_entry(
+        path_input=Path(fname_input).absolute(),
+        path_qc=Path(path_qc),
+        command=command,
+        cmdline=list2cmdline(cmdline),
+        plane='Axial',
+        dataset=dataset,
+        subject=subject,
+    ) as (path_background_img, path_overlay_img):
+
+        # Resample images slice by slice
+        p_resample = 0.6
+        logger.info('Resample images to %fx%f mm', p_resample, p_resample)
+        img_input = Image(fname_input).change_orientation('SAL')
+        img_input = resample_nib(
+            image=img_input,
+            new_size=[img_input.dim[4], p_resample, p_resample],
+            new_size_type='mm',
+            interpolation='spline',
+        )
+        img_output = resample_nib(
+            image=Image(fname_output).change_orientation('SAL'),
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # Copy any missing assets
+        path_qc.mkdir(parents=True, exist_ok=True)
+        copy_missing(path_assets / '_assets', path_qc)
+
+    logger.info('Successfully generated the QC results in %s', str(path_result))
+    display_open(file=str(path_index_html), message="To see the results in a browser")
+
+
+def copy_missing(resource: Traversable, destination: Path):
```
mguaypaq, line 159: `Traversable` and `Path` both present almost the same interface, the difference being that a `Traversable` could even represent something that lives inside a ZIP archive, etc., rather than something that actually exists in the filesystem directly.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    by creating any missing files and directories recursively.
+    """
+    path = destination / resource.name
+    if resource.is_dir():
+        path.mkdir(exist_ok=True)
+        for sub_resource in resource.iterdir():
+            copy_missing(sub_resource, path)
+    elif resource.is_file():
+        if not path.is_file():
+            path.write_bytes(resource.read_bytes())
```
mguaypaq, line 171: This doesn't try to check whether the target has the right _contents_, just whether it exists. This was already the existing behaviour, probably for (reasonable) performance reasons.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    path_qc: str,
+    dataset: Optional[str],
+    subject: Optional[str],
+):
+    """
+    Generate a QC report for sct_register_multimodal.
+    """
+    command = 'sct_register_multimodal'
+    cmdline = [command]
+    cmdline.extend(argv)
```
mguaypaq, line 191: I wrote it this way because `argv` is only assumed to be a `Sequence`, and writing `[command] + arg` would only have worked with an actual `list`. But `list.extend` accepts any iterable as an argument.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+            path.write_bytes(resource.read_bytes())
+    else:
+        # Some weird kind of resource? Ignore it
+        pass
+
+
+def sct_register_multimodal(
+    fname_input: str,
+    fname_output: str,
+    fname_seg: str,
```
mguaypaq, line 180: Since this is a function intended to be imported from elsewhere in SCT, I chose to keep the convention of representing paths as strings. They get converted to `Path`s in the function body.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+            image=Image(fname_output).change_orientation('SAL'),
+            image_dest=img_input,
+            interpolation='spline',
+        )
+        img_seg = resample_nib(
+            image=Image(fname_seg).change_orientation('SAL'),
+            image_dest=img_input,
+            interpolation='linear',
+        )
+        img_seg.data = (img_seg.data > 0.5) * 1
```
mguaypaq, line 224: I think a lot of the contortions in the existing QC report code came from trying to fit these three calls to `resample_nib` into a loop. But the arguments (and post-processing) are slightly different each time, so I think it's fine to unroll the loop.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # Each slice is centered on the segmentation
+        logger.info('Find the center of each slice')
+        centers = np.array([center_of_mass(slice) for slice in img_seg.data])
+        inf_nan_fill(centers[:, 0])
+        inf_nan_fill(centers[:, 1])
```
mguaypaq, line 230: The previous mosaic code also generated (and threw away) a mosaic for the segmentation image, unnecessarily.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+        # For QC reports, axial mosaics will often have smaller height than width
+        # (e.g. WxH = 20x3 slice images). So, we want to reduce the fig height to match this.
+        # `size_fig` is in inches. So, dpi=300 --> 1500px, dpi=100 --> 500px, etc.
+        size_fig = [5, 5 * img.shape[0] / img.shape[1]]
+
+        fig = Figure()
+        fig.set_size_inches(*size_fig, forward=True)
+        FigureCanvas(fig)
+        ax = fig.add_axes((0, 0, 1, 1))
+        ax.imshow(img, cmap='gray', interpolation='none', aspect=1.0)
```
mguaypaq, line 244: A possible future change: the mosaic code makes a grid of images in a single numpy array. But arguably this kind of layout operation would be easier in matplotlib? For example, by using [`Figure.subplots`](https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.subplots)?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+        add_orientation_labels(ax)
+        ax.get_xaxis().set_visible(False)
+        ax.get_yaxis().set_visible(False)
+        img_path = str(path_overlay_img)
+        logger.debug('Save image %s', img_path)
+        fig.savefig(img_path, format='png', transparent=True, dpi=300)
+
+
+def inf_nan_fill(A: np.ndarray):
```
mguaypaq, line 267: [Copied from `slice.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/aaa1dc71b0502fb8db87a1a85cc5a5ac583557cf/spinalcordtoolbox/reports/slice.py#L185-L198), so that this proof-of-concept is self-contained.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    num_col = math.floor(600 / (2*radius[0]))
+
+    # Center and crop each axial slice
+    cropped = []
+    for center, slice in zip(centers.astype(int), img.data):
+        # Add a margin before cropping, in case the center is too close to the edge
+        cropped.append(np.pad(slice, radius)[
+            center[0]:center[0] + 2*radius[0],
+            center[1]:center[1] + 2*radius[1],
+        ])
```
mguaypaq, line 302: This is actually a change in behaviour: the existing code returns a smaller patch instead of adding padding, when the center is too close to the edge. But this interacts poorly with the existing mosaic code, and the too-small patch ends up in the wrong location (which is a bug).

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+    # Pad the list with empty arrays, to get complete rows of num_col
+    empty = np.zeros((2*radius[0], 2*radius[1]))
+    cropped.extend([empty] * (-len(cropped) % num_col))
+
+    # Arrange the images into a grid
+    return np.block([cropped[i:i+num_col] for i in range(0, len(cropped), num_col)])
```
mguaypaq, line 309: I love it when numpy already has a function to do something finicky like this!

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+def add_orientation_labels(ax: Axes):
+    """
+    Add orientation labels (A, P, L, R) to a figure, yellow with a black outline.
+    """
+    for x, y, letter in [
+        (12, 6, 'A'),
+        (12, 28, 'P'),
+        (0, 18, 'L'),
+        (24, 18, 'R'),
```
mguaypaq, line 320: Clearly this should be customizable. And in this proof-of-concept, the caller of this function would actually have enough context to know which labels to put where (unlike in the existing code structure).

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    ]:
+        ax.text(x, y, letter, color='yellow', size=4).set_path_effects([
+            path_effects.Stroke(linewidth=1, foreground='black'),
+            path_effects.Normal(),
+        ])
+
+
+def equalize_histogram(img: np.ndarray):
```
mguaypaq, line 328: [Copied from `qc.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/aaa1dc71b0502fb8db87a1a85cc5a5ac583557cf/spinalcordtoolbox/reports/qc.py#L320-L345), so that this proof-of-concept is self-contained.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+        subject=subject,
+    ) as (path_background_img, path_overlay_img):
+
+        # Resample images slice by slice
+        p_resample = 0.6
+        logger.info('Resample images to %fx%f mm', p_resample, p_resample)
+        img_input = Image(fname_input).change_orientation('SAL')
+        img_input = resample_nib(
+            image=img_input,
+            new_size=[img_input.dim[4], p_resample, p_resample],
```
mguaypaq, line 210: I think this line is one of the main differences between `Axial` and `Sagittal`.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    and cropped to `radius` pixels in each direction (horizontal, vertical).
+
+    If `img` has N slices, then `centers` should have shape (N, 2).
+    """
+    # Fit as many slices as possible in each row of 600 pixels
+    num_col = math.floor(600 / (2*radius[0]))
+
+    # Center and crop each axial slice
+    cropped = []
+    for center, slice in zip(centers.astype(int), img.data):
```
mguaypaq, line 297: Subtly, this line is the other main difference between `Axial` and `Sagittal`: which axis of `img.data` is iterated over.

mguaypaq (commented):
Here's a self-review, to point out some things about the large body of code. I hope it helps!

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    # Axial orientation, switch between two input images
+    with qc_entry(
+        path_input=Path(fname_input).absolute(),
+        path_qc=Path(path_qc),
+        command=command,
+        cmdline=list2cmdline(cmdline),
+        plane='Axial',
+        dataset=dataset,
+        subject=subject,
+    ) as (path_background_img, path_overlay_img):
```
joshuacwnewton, line 202:
```suggestion
    ) as imgs_to_generate:
```

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+        fig = Figure()
+        fig.set_size_inches(*size_fig, forward=True)
+        FigureCanvas(fig)
+        ax = fig.add_axes((0, 0, 1, 1))
+        ax.imshow(img, cmap='gray', interpolation='none', aspect=1.0)
+        add_orientation_labels(ax)
+        ax.get_xaxis().set_visible(False)
+        ax.get_yaxis().set_visible(False)
+        img_path = str(path_background_img)
```
joshuacwnewton, line 248:
```suggestion
        img_path = str(imgs_to_generate['path_background_img'])
```

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+        img = equalize_histogram(mosaic(img_output, centers))
+        fig = Figure()
+        fig.set_size_inches(*size_fig, forward=True)
+        FigureCanvas(fig)
+        ax = fig.add_axes((0, 0, 1, 1), label='0')
+        ax.imshow(img, cmap='gray', interpolation='none', aspect=1.0)
+        add_orientation_labels(ax)
+        ax.get_xaxis().set_visible(False)
+        ax.get_yaxis().set_visible(False)
+        img_path = str(path_overlay_img)
```
joshuacwnewton, line 262:
```suggestion
        img_path = str(imgs_to_generate['path_overlay_img'])
```

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+    timestamp = mod_date.strftime('%Y_%m_%d_%H%M%S.%f')
+
+    # Make sure the image directory exists
+    path_img = path_qc / dataset / subject / contrast / command / timestamp
+    path_img.mkdir(parents=True, exist_ok=True)
+    path_background_img = path_img / 'background_img.png'
+    path_overlay_img = path_img / 'overlay_img.png'
+
+    # Let the caller generate the image files for the entry
+    yield (path_background_img, path_overlay_img)
```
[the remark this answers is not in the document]
joshuacwnewton: We may also want to include a quick check immediately after the yield, which further communicates the requirement to create these images:

```suggestion

    # Ask the caller to generate the image files for the entry
    imgs_to_generate = {
        'path_background_img': path_img / 'background_img.png',
        'path_overlay_img':  path_img / 'overlay_img.png'
    }
    yield imgs_to_generate
    # Double-check that the images were generated during the 'with:' block
    for img_type, path_img in imgs_to_generate.items():
        if not Path.exists(path_img):
            raise FileNotFoundError(f"Required QC image '{img_type}' was not found at the expected path: '{path_img}')")
```

joshuacwnewton (commented):
@mguaypaq  and I discussed this briefly over Slack on Friday. 

My main thought was this: "I like the [justification for using context managers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4224#discussion_r1333637422), though if we do use a context manager, we should consider aiming for similarity to the `with open(filepath) as file:` context manager". I figure that `open()` is most Python developers' first (and often only?) exposure to context managers, and so folks already have a certain usage pattern in mind.

As far as concrete suggestions go, I wanted to give stronger hints for how `qc_entry` is supposed to be used:

1. Renaming the context manager from a noun phrase (`qc_entry`) to a verb phrase (`create_qc_entry`), in a similar vein to how `open()` is named.
2. Making the purpose of the filepaths explicit, by requiring the user to query a dictionary called `imgs_to_generate`, thereby giving the user the cue that they need to generate these images. (Also, returning a single object -- to be operated on -- is more closely in line with the behavior of `open()`.)

----

I had one additional suggestion: Use a class-based context manager, which provides two separate methods `__enter__()` and `__exit__()` to make the "before"/"after" logic a little more clear. 

However, this introduces a lot of boilerplate code in order to preserve state between the "before" and "after" portions (e.g. the various metadata variables such as `dataset`, `subject`, etc.), whereas using a single function keeps everything in a single context. So, I think it ends up being simpler to not use a class after all.

One way to better indicate the separation between before/after code could be to use a `try:`/`finally:` block (as shown by the [`contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) docs)? Perhaps this is the best of both worlds.

----

(I'm also going to mark a good portion of your @mguaypaq's self-review as "Resolved" -- mainly, the uncontroversial points that were made that don't really need further discussion.)

spinalcordtoolbox/scripts/sct_register_multimodal.py
```diff
@@ -425,13 +421,16 @@ def main(argv: Sequence[str]):
...
 
-    if path_qc is not None:
-        if fname_dest_seg:
-            generate_qc(fname_src2dest, fname_in2=fname_dest, fname_seg=fname_dest_seg, args=argv,
-                        path_qc=os.path.abspath(path_qc), dataset=qc_dataset, subject=qc_subject,
-                        process='sct_register_multimodal')
-        else:
-            printv('WARNING: Cannot generate QC because it requires destination segmentation.', 1, 'warning')
+    if arguments.qc is not None:
+        qc2.sct_register_multimodal(
```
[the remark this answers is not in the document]
joshuacwnewton: I really like this change.

As you've pointed out, the logic for  _"mapping positional inputs (`path_1`, `path_2`) to keyword inputs (`path_seg`, `path_anat`)"_ is only really needed by the `sct_qc` script, since its interface must be uniform regardless of the given SCT process.

Yet, despite the fact that it's only needed for `sct_qc`, we are currently using this same interface _everywhere_. So, I very much like the idea of confining the "positional input" logic to `sct_qc`, and leaving the "keyword input" for each individual SCT script. It removes the need to scan through a bunch of boilerplate when dealing with QC code outside of `sct_qc`.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Specifically, this ties into an earlier issue we encountered:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4122

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Re: This comment + https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4224#discussion_r1333659259:

In other words, once we convert some additional QC methods (both `Axial` and `Sagittal`), we should be able to re-use this code by tweaking the order of `new_size` according to `plane`?

Or, is "trying to reuse the resampling code" part of the reason why the QC code got so hard to work with in the first place? (Since, we enforced a very inflexible structure on all of the various image-generating methods.)

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # Collect all existing QC report entries
+        json_data = []
+        for path in sorted(path_json.glob('*.json')):
+            with path.open() as file:
+                json_data.append(json.load(file))
+
+        # Insert the QC report entries into index.html
+        path_assets = importlib.resources.files(spinalcordtoolbox.reports) / 'assets'
+        template = string.Template((path_assets / 'index.html').read_text(encoding='utf-8'))
```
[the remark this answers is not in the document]
mguaypaq: On second thought, I guess `string.Template` is fine. It's just that shell-style substitution tends to be so full of subtle pitfalls that it immediately raises my hackles. But the rules for `string.Template` are simple enough that I don't think it's a problem.

(It didn't help that the `safe_substitute` method is what I would really consider to be the **unsafe** version of the `substitute` method, [as acknowledged in the docs](https://docs.python.org/3/library/string.html#string.Template.safe_substitute). The kind of thinking behind that name is exactly what leads to subtle pitfalls!)

Still, our usage is so simple that it could be replaced with a call to the builtin `str.replace` method.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Exactly. Thanks for the cross-reference.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Both! I very much think the singular resampling pipeline was a big source of inflexibility, but of course it's tempting to parametrize by `plane`. We should be mindful if/when/how we do it. I can think of a few different ways, for example:
- Having a `resample_slices` function which literally takes a stringly-typed `plane` argument. This might be convenient if we have several places where we only know whether we want axial or sagittal at runtime. But I suspect we'll only need this once, in `sct_qc.py`, in which case we don't gain much by abstracting essentially just an `if` statement.
- Having a pair of functions `resample_axial` and `resample_sagittal` which are very simple wrappers around `resample_nib`. Possibly also combined with the call to `change_orientation`? Something to think about. (Could be informed by the discussion in this [interesting talk on API design](https://caseymuratori.com/blog_0024).)

One thing to note is that only the first image needs this `new_size` parameter! All other images use the first resized image as a template, and don't depend on the axial/sagittal distinction. (For resampling. Later on, when iterating over slices, the axial/sagittal distinction re-appears for all images.)

mguaypaq:
@joshuacwnewton says:
> One way to better indicate the separation between before/after code could be to use a `try:`/`finally:` block (as shown by the [`contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) docs)? Perhaps this is the best of both worlds.

I considered this (and used this pattern for the `locked_file` utility), but decided against it, because in this case the `except` and `finally` blocks would all be just `pass`.

In the `locked_file` case, we're holding a resource (the filesystem lock) that we should release when we exit the block even if the `with` body raises an exception, otherwise there might be a deadlock scenario.

In the `qc_entry` case, if there's a exception in the `with` body (that is, an error while generating the QC report images), I don't think we can usefully catch it, and I think we don't want to create/update the QC report at all.

But, I really love your suggestion to check that the required files exist when returning from the `with` body, and hopefully that can serve as sufficient bracketting?

spinalcordtoolbox/reports/qc2.py
mguaypaq: Also, for this refactoring I'm trying to first port some more kinds of reports, and only create functions and function parameters once they're actually used in a few places. That way we'll have some real examples of usage to guide the API design.

mguaypaq:
I love the new broken link checker, but it's proving to be a little flaky (and maybe doing hundreds of web requests in the space of two minutes is a little rude?). Could we move it to run daily instead of as part of every PR?

Or if we want to get fancy, maybe it could check only links that are part of the PR's diff when it's run for a PR? (There would still be value in periodically checking that all links are still alive.)

joshuacwnewton:
> Could we move it to run daily instead of as part of every PR?

I think this is a very good idea. I've been running into failed links every now and then, and have had to repeatedly re-run the checker just to eliminate random spurious warnings due to some sort of flaky network/connection issue. 

Doing that process once per day when I start my work day seems very OK to me. :)

spinalcordtoolbox/reports/qc2.py
```diff
@@ -0,0 +1,359 @@
...
+    path_qc: str,
+    dataset: Optional[str],
+    subject: Optional[str],
+):
+    """
+    Generate a QC report for sct_register_multimodal.
+    """
+    command = 'sct_register_multimodal'
+    cmdline = [command]
+    cmdline.extend(argv)
```
joshuacwnewton, line 198: Minor nitpick regarding naming conventions:

Do we still want to call each `qc2` function name by its corresponding CLI script (e.g. `sct_register_multimodal`)? Or, do we want to create more "general" QC functions, in line with https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4166, and how we wish to remove the `-p`/`-process` flag from `sct_qc`? 

I guess it's a little tricky, because to some degree we need to specify the `command`/`cmdline` argument. But, still, I'm wondering if this is the "final" organization we want to settle on?

joshuacwnewton (commented):
I'm very happy with the bulk of the code changes here (function-based QCs, `create_qc_entry` context manager).

That said, on reflection, I had some minor organizational thoughts? Just, thinking about the entry points to the QC functions (for both the individual CLI scripts and the `sct_qc` standalone function).

spinalcordtoolbox/reports/qc2.py
mguaypaq: I don't think `qc2.sct_register_multimodal` is the last word, no! And I do want to support these "more general" QC functions down the line. But, I'm wary about premature abstraction, because I think that was the cause of the previous inflexibility in this code.

So, my main goal with this PR is to get the `with create_qc_entry` structure in place, which I think is pretty solid. After that, I envision a sequence of PRs where we "port" each existing QC report to this new structure, while de-abstracting the code as much as possible. As we do this, I hope that we can properly re-abstract the code, but only once 2 or (ideally) 3 instances of a given pattern appear. This should make it clear what functions to write, what their boundaries are, what their parameters are, and what their parameters *aren't*.

(This method is a little experimental for me, [inspired by a blog post](https://caseymuratori.com/blog_0015), and I think it will be worthwhile, but I'd like to have your approval before doing it. And if you find that the experiment is failing later on, I hope you'll tell me so that we can drop it.)

joshuacwnewton (approved):

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: For posterity: It was indeed a bug! You predicted it here months before it was reported: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4386

Issue #4225, opened 2023-09-21
Upstream bug: `onnxruntime==1.16.0` requires a parameter meant for GPUs (even though SCT uses the CPU version)


mguaypaq:
See for example this output, taken from [this testing run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/6237705089/job/17019439874):

```
FAILED testing/api/test_deepseg_gm.py::TestCore::test_segment_volume - ValueError: This ORT build has ['AzureExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['AzureExecutionProvider', 'CPUExecutionProvider'], ...)
FAILED testing/api/test_deepseg_lesion.py::test_segment - ValueError: This ORT build has ['AzureExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['AzureExecutionProvider', 'CPUExecutionProvider'], ...)
FAILED testing/api/test_deepseg_sc.py::test_deep_segmentation_spinalcord[params0] - ValueError: This ORT build has ['AzureExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['AzureExecutionProvider', 'CPUExecutionProvider'], ...)
FAILED testing/api/test_deepseg_sc.py::test_deep_segmentation_spinalcord[params1] - ValueError: This ORT build has ['AzureExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['AzureExecutionProvider', 'CPUExecutionProvider'], ...)
[...]

PR #4226, opened 2023-09-22
Temporarily pin `onnxruntime` to <1.16

mguaypaq:
Due to https://github.com/microsoft/onnxruntime/issues/17631, version 1.16.0 is causing CI to fail on master. We can change the pin to `!=1.16.0` (or to however many versions are buggy) once upstream releases a fix.

Part of #4225.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,7 +23,8 @@ nilearn
 numpy<1.24
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
-onnxruntime>=1.7.0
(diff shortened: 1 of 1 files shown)


mguaypaq:
But ultimately, I'm fine with either fix, up to you.

joshuacwnewton:
I think my thinking here was that we might want to keep the versions as open as possible? But, that said, I have no guarantee whether the `providers` fix is backwards-compatible with older verisons. So, on second thought, I think I like your fix a little more -- we can always go back and specify `onnxruntime!=1.16.0` once the patch comes out.

joshuacwnewton (approved):

mguaypaq:
Looking at [the changelog for ORT 1.9](https://github.com/Microsoft/onnxruntime/releases/tag/v1.9.0), which is mentioned in the error message. Looking at the first point under _Packages_, it looks like this was meant for the GPU packages only, not for the basic CPU package.

Issue #3295, opened 2021-03-16
'nib.nifti1.Nifti1Image' calls that don't specify header correctly return images with default s/qform codes


joshuacwnewton:
This is similar to the issue in https://github.com/ivadomed/ivadomed/pull/714. There are two situations to watch out for:

### 1 Passing affine only

In SCT, there is at least one call to `nib.nifti1.Nifti1Image` that only specifies the affine, _not_ the header:

https://github.com/neuropoly/spinalcordtoolbox/blob/60ea08c9d664cf403f2982f8f5c8ae1cc5d704a3/spinalcordtoolbox/image.py#L1417

The relevant [documentation](https://nipy.org/nibabel/nifti_images.html#default-sform-and-qform-codes) states:

> If you create a new image, the sform and qform codes will be initialised to 2 (aligned) and 0 (unknown) respectively

### 2 Passing affine + header that doesn't match

In SCT, there is at least one call to `nib.nifti1.Nifti1Image` that specifies the affine and the header:

https://github.com/neuropoly/spinalcordtoolbox/blob/60ea08c9d664cf403f2982f8f5c8ae1cc5d704a3/spinalcordtoolbox/image.py#L1330

From the same [documentation section](https://nipy.org/nibabel/nifti_images.html#default-sform-and-qform-codes):

> If you create a new image and specify an existing header, then the newly created image will inherit the same sform and qform codes that are in the provided header. However, if you create a new image with both an affine and a header specified, then the sform and qform codes will only be preserved if the provided affine is the same as the affine in the provided header. If the affines do not match, the sform and qform codes will be set to their default values of 2 and 0 respectively.

joshuacwnewton:
I wanted to update this issue to reference https://github.com/nipy/nibabel/issues/1012#issuecomment-831980457.

The guidance from `nibabel` here is to set both `header` and `affine`, but either make sure the affines match, or copy the codes after the fact.

PR #4230, opened 2023-09-27
Convert integer images to floating point when resampling with linear interpolation

### LINKED ISSUE #4210: When resampling a binary (UINT8) mask using linear interpolation, the output type should change to FLOAT
### Description

While debugging https://forum.spinalcordmri.org/t/pam50-template-registration-for-highly-anisotropic-axial-scans/1117, I noticed that when running `sct_resample` on a binary mask (eg a segmentation) and asking for linear interpolation, the output was binary. This happened because my input image was encoded in UINT8, and the output resampled image uses the same dtype.

I suggest that, if linear or spline is asked for in `sct_resample`, the script checks if the input dtype is integer, and if it is, it should be changed for FLOAT32, while also sending a warning message about it.

### Steps to Reproduce

version: Spinal Cord Toolbox (git-master-76c27e16f13974de3ff8f81e25c92abe8f6d80ec)

Image: [sub-m023917_ses-20160302_acq-ax_T2w_seg.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/12488227/sub-m023917_ses-20160302_acq-ax_T2w_seg.nii.gz)

```
sct_resample -i sub-m023917_ses-20160302_acq-ax_T2w_seg.nii.gz -mm 1 -x linear
```

Notice the output is binary.

If changing to FLOAT32 and re-running, it works as expected:
~~~
sct_image -i sub-m023917_ses-20160302_acq-ax_seg_T2w.nii.gz -type float32 -o sub-m023917_ses-20160302_acq-ax_seg_T2w_float32.nii.gz
sct_resample -i sub-m023917_ses-20160302_acq-ax_seg_T2w_float32.nii.gz -mm 1 -x linear
~~~

mguaypaq:
It can be surprising (and a big loss of precision) to do linear resampling on an integer-valued image and get back integer-valued results. This PR changes the behaviour of `resample_nib` (which is widely used in SCT for resampling images) so that resampling an integer-valued image with linear or spline interpolation first converts the image to a floating-point type.

If a type conversion happens, a warning message is produced.

Note that there is no type conversion when:
- the input image already has a floating-point type, even if it's float32 (instead of the default float64); nor when
- the interpolation type is "nearest neighbour", since this doesn't require any arithmetic on the voxel values.

The PR also includes two tests, to check that:
- a type conversion happens for linear interpolation of an integer image, and
- there is no type conversion for nearest neighbour interpolation of an integer image.

Only the first test fails before the code change, but I think the second test is a useful safety.

Note that this shouldn't affect QC reporting code much, since that code does its own thresholding on binary segmentations:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/792a155fdb55aaae8096a2d6b32874d1a98fcc13/spinalcordtoolbox/reports/slice.py#L299-L308

Fixes #4210.


joshuacwnewton:
Re: Test failures

Given the ubiquity of the resampling code, as well as the change in precision, I'm not terribly surprised that the `batch_processing.sh` results have changed? That said... woof!

spinalcordtoolbox/resampling.py
```diff
@@ -49,13 +49,21 @@ def resample_nib(image, new_size=None, new_size_type=None, image_dest=None, inte
...
     else:
         raise TypeError(f'Invalid image type: {type(image)}')
 
+    # convert to floating point if we're doing arithmetic interpolation
+    if interpolation != 'nn' and img.get_data_dtype().kind in 'biu':
+        original_dtype = img.get_data_dtype()
+        img = nib.nifti1.Nifti1Image(img.get_fdata(), img.affine, img.header)
+        img.set_data_dtype(img.dataobj.dtype)
+        logger.warning("Converting image from type '%s' to type '%s' for %s interpolation",
+                       original_dtype, img.get_data_dtype(), interpolation)
```
joshuacwnewton, line 65: Just out of curiosity, did you by chance look into whether there's a way to avoid this on the nibabel side of things? (i.e. passing an integer image into nibabel's resampling function, but somehow automatically returning a float data array?)

I ask because I'm curious where the actual type coercion is happening, and (if it's happening inside of nibabel), whether there's a more "official" way to avoid the coercion, so to speak.

jcohenadad: I remember spending time looking at this. Explorations brought me to scikit-image's rationale for converting output dtype when _it made sense_, see notably [this doc](https://scikit-image.org/docs/stable/user_guide/data_types.html#output-types). I'm not sure if nibabel is relying on scikit library though, but the rationale still holds, I think.

joshuacwnewton: I'll try and dig a bit right now using the data you provided in the issue. :)

joshuacwnewton: Quick investigation:

- `nibabel` is using [`scipy.ndimage.affine_transform`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html) under the hood
- `affine_transform` has an `output` parameter with the following description:
    >   output : array or dtype, optional
    > 
    >   The array in which to place the output, or the dtype of the returned array. By default an array of the same dtype as input will be created.
- `nibabel` does not specify this parameter, nor expose it to the `resample_from_to` function.

I could petition `nibabel` to expose this parameter, but otherwise I think @mguaypaq's solution (changing the dtype of the input array) is the best we can do right now given the default behavior of `affine_transform`.

spinalcordtoolbox/resampling.py
```diff
@@ -56,6 +56,14 @@ def resample_nib(image, new_size=None, new_size_type=None, image_dest=None, inte
     else:
         raise TypeError(f'Invalid image type: {type(image)}')
 
+    # convert to floating point if we're doing arithmetic interpolation
+    if interpolation != 'nn' and img.get_data_dtype().kind in 'biu':
+        original_dtype = img.get_data_dtype()
+        img = nib.nifti1.Nifti1Image(img.get_fdata(), img.affine, img.header)
```
joshuacwnewton, line 62: I think we may need to use `img.header.get_best_affine()` rather than `img.affine`? There have been some code issues in the past related to this: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3295#issuecomment-833966224

> Assuming that `img.affine == img.header.get_best_affine()` (which will be true if you're loading `img` from file), then `Nifti1Image(data, img.affine, img.header)` will get you the same thing, but also allow you go modify the file. If you don't trust your affine but do trust your header, then `Nifti1Image(data, img.header.get_best_affine(), img.header)` should work fine.

Since this change is happening inside `resample_nib`, rather than `resample_file`, I'm not sure if we can safely assume that `img.affine` produces the same result as `img.hdr.get_best_affine()`? So, using the latter I think should help us preserve the codes upon initializing a new image.

spinalcordtoolbox/resampling.py
mguaypaq: Thanks for the investigation!

spinalcordtoolbox/resampling.py
mguaypaq: Thanks for the suggestions, you're entirely right. I think I was relying on the call to `img.header.get_best_affine()` a few lines above, but of course that only covers the case of `isinstance(image, Image)`, not the case of `isinstance(image, nib.nifti1.Nifti1Image)`.
```suggestion
        img = nib.nifti1.Nifti1Image(img.get_fdata(), img.header.get_best_affine(), img.header)
```

joshuacwnewton (approved):

mguaypaq:
I want to look into the `batch_processing.sh` changes a bit before merging this.

mguaypaq:
After looking at the results of `batch_processing.sh`, I think this PR causes significant changes, so we probably don't want to merge it right before the 6.1 release. I've re-assigned it to the 6.2 release.

Looking only at the t2 images:
- The output of `sct_deepseg_sc` is almost the same: only 10 voxels are different, out of 64x320x320. (When going from "before" to "after", 3 voxels are removed and 7 voxels are added.)
- The output of `sct_label_vertebrae` is almost the same, too: only 18 voxels are different, out of 64x320x320. (While comparing to the previous point, I was surprised to see that ~230 voxels that are nonzero in `t2_seg.nii.gz` are labeled zero in `t2_seg_labeled.nii.gz`, even though they are much too low along the spine. Maybe this is a bug?)
- But the output of `sct_register_to_template` is visibly different. For example, here is a before/after screenshot:
  ```bash
  fsleyes before/t2/t2.nii.gz before/t2/template2anat.nii.gz after/t2/template2anat.nii.gz
  ```
  out
- The other commands in `batch_processing.sh` rely on the results of `sct_register_to_template`, so I didn't look at them.

jcohenadad:
thank you for the thorough investigation @mguaypaq. Yes, I am not surprised that very little differences in the segmentation produces visible changes in the registration and subsequent pipeline functions. Is that problematic? It depends. In this case I suspect changes are quite minimal. We just want to make sure that we are introducing changes that 'go in the right direction'

joshuacwnewton:
Given that the conclusion on this PR is that small discrepancies in batch processing results are OK, I'm going to re-run batch processing locally and update the results accordingly.

----

Side note: I've had to do this a few times recently, and it's a little cumbersome. I wonder if it might be good to refactor the CI workflow to upload the metric files to GitHub as artifacts, then test the results in a separate step.

That way, on failure, if we decide to keep the new results, we can simply download the csv files from GitHub, which would save quite a bit of hassle.

joshuacwnewton (approved):

PR #4232, opened 2023-09-28
Avoid `dipy` versions 1.6.0 + 1.7.0 that contain a `method=restore` bug

### LINKED ISSUE #4209: `ValueError: could not broadcast array (7) into shape (12)` during `-method restore`
### Description

`sct_dmri_compute_dti -method restore` fails with the following error:

```console
$ cd $SCT_DIR/data/course_london20/single_subject/data/dmri
$ sct_dmri_compute_dti -i dmri.nii.gz -bval bvals.txt -bvec bvecs.txt -method restore 

--
Spinal Cord Toolbox (git-master-76c27e16f13974de3ff8f81e25c92abe8f6d80ec)

sct_dmri_compute_dti -i dmri.nii.gz -bval bvals.txt -bvec bvecs.txt -method restore
--

data.shape (96, 36, 15, 35)
Computing tensor using "restore" method...
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_dmri_compute_dti.py", line 183, in <module>
    main(sys.argv[1:])
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_dmri_compute_dti.py", line 100, in main
    if not compute_dti(fname_in, fname_bvals, fname_bvecs, prefix, method, evecs, file_mask, verbose):
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_dmri_compute_dti.py", line 150, in compute_dti
    tenfit = dti_restore.fit(data)
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.8/site-packages/dipy/reconst/dti.py", line 813, in fit
    params_in_mask = self.fit_method(
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.8/site-packages/dipy/reconst/dti.py", line 1859, in restore_fit_tensor
    D = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.8/site-packages/dipy/reconst/dti.py", line 1327, in wrapped_fit_tensor
    dtiparams[i:i + step] = fit_tensor(design_matrix,
ValueError: could not broadcast input array from shape (10000,7) into shape (10000,12)
```

### Steps to Reproduce

```console
$ cd $SCT_DIR/data/course_london20/single_subject/data/dmri
$ sct_dmri_compute_dti -i dmri.nii.gz -bval bvals.txt -bvec bvecs.txt -method restore 
```

joshuacwnewton:
## Description

`sct_dmri_compute_dti` fails on `sct_example_data` images when using `-method restore`:

```
sct_dmri_compute_dti -i dmri.nii.gz -bval bvals.txt -bvec bvecs.txt -method restore
```

I dug into this on the `dipy` side of things, and found that the issue only exists for `dipy==1.6.0,1.7.0`, and has already been fixed upstream. `dipy==1.8.0` is releasing soon, so to avoid freezing a buggy version for SCT v6.1, we skip these two versions in our requirements.txt.

Note: We have an existing test for this. The reason that our test suite didn't catch this is because `dipy` chunks the processing, and the test image is smaller than a single chunk, but the bug only occurs when there are >1 chunks. I'm not quite sure how to test for this given our available testing data? But maybe given that this is a one-off issue with an upstream library, we don't necessarily need to test for this.

## Linked issues

Fixes #4209.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,8 @@
 #     stable releases. We do this because `pip freeze` will not capture options (e.g. --extra-index-url) or
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
-dipy
+# Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
+dipy<1.6.0,>1.7.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
 # Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
-dipy<1.6.0,>1.7.0
+dipy!=1.6.*,!=1.7.*
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze


requirements.txt
```diff
@@ -2,7 +2,8 @@
 #     stable releases. We do this because `pip freeze` will not capture options (e.g. --extra-index-url) or
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
-dipy
+# Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
+dipy<1.6.0,>1.7.0
```
joshuacwnewton, line 6: Whoops! This is not valid pip version range syntax. (pip treats this as an `AND` rather than an `OR`, so there's no way to satisfy both conditions.)

```suggestion
dipy!=1.6.*,!=1.7.*

mguaypaq (approved):
Thanks for the investigation and the upstream report! And avoiding buggy versions (like you did) seems nicer than pinning.

PR #4233, opened 2023-09-28
Avoid duplicate orthogonal labels during registration by checking existing labels first

### LINKED ISSUE #4228: Fix conflict between internal dummy label `99` and user-supplied label `99`
context: https://forum.spinalcordmri.org/t/registration-of-lumbar-data-to-pam50/1109/5

suggestions:
- we check for input label 99 in custom label, we update the doc to make it clear
- we use an alternative approach for creating our dummy label internally (to make sure no conflict is possible).

joshuacwnewton:
## Description

In `sct_register_to_template`, we add a dummy label (`99`) to ensure we have 3 labels during registration. The value `99` is arbitrary -- `100`, `101`, etc. would save the same purpose, because whichever label we choose gets added to both the input label image and the template label image.

In #4228, we encountered a bug where a user chose `99` for their own manually-defined lumbar data, which caused a clash with our hardcoded dummy label value (and hence poor registration results).

So, this PR makes sure that a unique value is always used (typically 99, but perhaps 100/101/etc. if 99 is taken).

## Linked issues

Fixes #4228.


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -777,9 +777,12 @@ def main(argv: Sequence[str]):
         printv('\nRemove unused label on template. Keep only label present in the input label image...', verbose)
         sct_labels.remove_missing_labels(Image(ftmp_template_label), Image(ftmp_label)).save(path=ftmp_template_label)
 
-        # Add one label because at least 3 orthogonal labels are required to estimate an affine transformation.
-        add_orthogonal_label(ftmp_label)
-        add_orthogonal_label(ftmp_template_label)
+        # Add a dummy label, because at least 3 orthogonal labels are required to estimate an affine transformation.
+        # -> Pick a dummy label >=99 that doesn't clash with the existing label values.
+        existing_label_vals = {coord.value for coord in img_tmp_label.getCoordinatesAveragedByValue()}
+        dummy_label = 99 if 99 not in existing_label_vals else max(existing_label_vals) + 1
```
joshuacwnewton, line 783: Another reason for moving the dummy label value determination to the outer `sct_register_to_template` context is that it ensures we use the same value for both images. _(It felt a little risky to programmatically determine the label twice in two separate calls to `add_orthogonal_label`.)_

(I think it should be safe to use the labels from `img_tmp_label`? Since, any labels in the template label image that aren't present in `img_tmp_label` get removed earlier on.)

jcohenadad:
Thank you for looking into this @joshuacwnewton. I _think_ the reason I picked 99 in the past is because FSLview (parent of FSLeyes) did not allow to create labels above 99 (maybe it is still the case with FSLeyes?), and I needed to easily create such labels for debugging purpose, thinking that a clash with a 99 label would have low probability (we now know it was a bad idea).

At the same time, I don't think we want to use label value 999999 because of the UINT8 encoding. So maybe we should take a 'very high' value, while keeping the same dtype?

joshuacwnewton:
Re: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4228#issuecomment-1741467792

> and of course re-doing the check in case the other label is also included in the user's custom label 😅

Exactly. That's why I went for the `max() + 1` of the existing labels in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4233. That way, we can keep the existing `99` label behavior in 99.9% of cases, but use `100`, `101`, `102` if necessary.

>  I think the reason I picked 99 in the past is because FSLview (parent of FSLeyes) did not allow to create labels above 99 (maybe it is still the case with FSLeyes?)

Ah! This is fascinating! Thank you for the insight. Assuming this is still a limitation, then tn that case, I can modify the code to do the following:

- Start with the set `{1:99}`
- Subtract the set of existing labels (e.g. in this case the user chose `{98,99}`)
- Use the max of the remaining labels

That way, in this user's case, the label `97` would be used for the dummy label, while `99` would continue be used in 99.9% of other cases.

> At the same time, I don't think we want to use label value 999999 because of the UINT8 encoding. So maybe we should take a 'very high' value, while keeping the same dtype?

If I'm understanding correctly, this suggestion is essentially "pick an even more obscure large label value (other than `99`), so that we decrease the chances that the user just so happens to choose that number for their own labels (i.e. 99.9% safe --> 99.999% safe)"?

If my understanding is correct, then is the goal of this approach to ensure that we always use a consistent orthogonal label value (rather than dynamically changing the label at runtime in special cases)?

jcohenadad:
> If I'm understanding correctly, this suggestion is essentially "pick an even more obscure large label value (other than 99), so that we decrease the chances that the user just so happens to choose that number for their own labels (i.e. 99.9% safe --> 99.999% safe)"?

yup! your understanding is correct

joshuacwnewton:
I spoke with @mguaypaq on Slack, and he suggested to use a value of either `127` (i.e. the upper limit for `INT8` images) or `255` (i.e. the upper limit for `UINT8` images). I think I like `127`, just to avoid any strange issues from the user supplying an `INT8` label image. (Unlikely, but I don't think there's any harm in avoiding `[128, 255]`?)

To double-check the FSLview/FSLeyes concern, I tried creating a new label in FSLeyes' edit mode on `PAM50_label_disc.nii.gz` (a `UINT8` image), and I can confirm that it can handle values above `99` just fine. (It could create labels `127` and `255`, but could not create label `256`, automatically snapping back to 255 in the "Fill value" dialogue. Makes sense for a `UINT8` image!)

spinalcordtoolbox/registration/labeling.py
```diff
@@ -15,9 +15,9 @@
 import spinalcordtoolbox.labels as sct_labels
 
 
-def add_orthogonal_label(fname_label):
+def add_orthogonal_label(fname_label, new_label_value):
     """
-    Add one label of value=99 at the axial slice that contains the label with the lowest value, 10 pixels to the right.
+    Add one label at the axial slice that contains the label with the lowest value, 10 pixels to the right.
```
mguaypaq, line 20: I know there's a TODO comment to make this code do "10 pixels to the right", but right now it does "5mm to the left", so the docstring should match the behaviour:
```suggestion
    Add one label at the axial slice that contains the label with the lowest value, 5mm to the left.
```

spinalcordtoolbox/registration/labeling.py
```diff
@@ -28,16 +28,10 @@ def add_orthogonal_label(fname_label):
     path_label = im_label.absolutepath
     im_label.change_orientation('RPI')
     coord_label = im_label.getCoordinatesAveragedByValue()  # N.B. landmarks are sorted by value
-    # Create new label
-    from copy import deepcopy
-    new_label = deepcopy(coord_label[0])
-    # move it 5mm to the left (orientation is RAS)
+    # Add new label 5mm to the left (orientation is RPI) to existing image
```
mguaypaq, line 31: Thanks for correcting the comment `RAS -> RPI`!

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -777,9 +777,13 @@ def main(argv: Sequence[str]):
         printv('\nRemove unused label on template. Keep only label present in the input label image...', verbose)
         sct_labels.remove_missing_labels(Image(ftmp_template_label), Image(ftmp_label)).save(path=ftmp_template_label)
 
-        # Add one label because at least 3 orthogonal labels are required to estimate an affine transformation.
-        add_orthogonal_label(ftmp_label)
-        add_orthogonal_label(ftmp_template_label)
+        # Add a dummy label, because at least 3 orthogonal labels are required to estimate an affine transformation.
+        # -> Pick a dummy label between [1, 127] that doesn't clash with the existing label values.
+        existing_label_vals = {coord.value for coord in img_tmp_label.getCoordinatesAveragedByValue()}
```
mguaypaq, line 782: There's no need to go through the trouble of averaging the coordinates by value, since we immediately throw away all that work, and since the set builder notation will de-duplicate the values just fine:
```suggestion
        existing_label_vals = {coord.value for coord in img_tmp_label.getNonZeroCoordinates()}
```

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -777,9 +777,13 @@ def main(argv: Sequence[str]):
...
+        positive_int8_vals = set(range(1, 128))
+        dummy_label = max(positive_int8_vals - existing_label_vals)  # Should be 127 in 99.99% of cases
```
mguaypaq, line 784: It's extremely unlikely that all 127 potential dummy labels would be used, so this call to `max` should be fine in practice. And if all candidates really are taken, this will safely throw a `ValueError`.

mguaypaq (approved):
Thanks for the fix! I agree that passing `new_label_value` to `add_orthogonal_label` is the right design. Feel free to merge this.

PR #4252, opened 2023-10-10
Remove `-s` functionality and add a deprecation warning

### LINKED ISSUE #4246: Remove option `-s` with new spinal level files
https://github.com/spinalcordtoolbox/PAM50/pull/18 will implement new spinal levels. These levels are integer (vs. the previous spinal levels which are probabilistic), and they are contained within a single file located under `PAM50/template` (previously `PAM50/spinal_levels`). It is therefore necessary to do the following change:
- when running `sct_warp_template`, remove the option -s because the spinal levels will be warped by default alongside all files under PAM50/template

joshuacwnewton:
## Description

Given the change from probabilistic spinal levels (multiple files) to integer spinal levels (one file) in https://github.com/spinalcordtoolbox/PAM50/pull/18, there is no longer a need for the `-s` option. 

Issue #4246 suggests removing the `-s` option outright. However, my concern is with our existing users -- they may want to know about the past issue with the existing spinal levels (and how to migrate to the new spinal levels). So, I wanted a more graceful error for any existing scripts that specify `-s 1` (since this was an opt-in option, given the default was `0` beforehand).

I'm not sure if we want to keep this deprecation notice indefinitely, or remove it at some point. (It is tricky with SCT, since many of our entrenched users may stick with older versions of SCT, and there is no telling when they may upgrade to a new version. So, I worry that if we remove it after, say, the 6.2 or 6.3 release, the warning will be missed by people who never used 6.1/6.2.)

## Linked issues

Fixes #4246.


spinalcordtoolbox/scripts/sct_warp_template.py
```diff
@@ -245,18 +235,31 @@ def get_parser():
...
+
+
 def main(argv: Sequence[str]):
     parser = get_parser()
     arguments = parser.parse_args(argv)
     verbose = arguments.v
     set_loglevel(verbose=verbose)
 
+    if arguments.s:
+        raise parser.error(S_DEPRECATION_STRING)
```
mguaypaq, line 256: Nitpick: [`parser.error`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) doesn't return an exception that can be raised; it exits the program.
```suggestion
        parser.error(S_DEPRECATION_STRING)
```

mguaypaq (approved):
This is a good deprecation message, and this approach is much nicer to users than just removing the option. Thanks!

I wondered whether we should give the exact filename (`PAM50_spinal_levels.nii.gz`), but I guess that might change and get out of date, and/or it might not be present if someone is using a different template than the builtin PAM50.

PR #4255, opened 2023-10-14
Dataset Table appended to sct_download_data -h

co-decode:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
Issue #4007 requested that the list of datasets for the `sct_download_data -h` command be more readable.

I have turned this list into a table, following the style of the table in `sct_deepseg -h`.
As suggested in the issue discussion, I have added a 'download_type' property to the DATASET_DICT items in `download.py` and sorted the datasets according to their type.

I have removed the duplicate lists for dataset choices and replaced them with the metavar `<dataset>`.

<details>
<summary> Image </summary>

image

</details>

## Linked issues
Fixes #4007

Issue #4007, opened 2023-01-21
`sct_download_data` - make a list of available binaries/data more readable


valosekj:
**Feature justification**

The list of binaries/data available for download by `sct_download_data` is a bit hard to read (because it is a long list without spaces).

```console
$ sct_download_data -h

--
Spinal Cord Toolbox (git-master-a15070f635264ad05aae41db83ce31160600e930)

sct_download_data -h
--

usage: sct_download_data -d
                         {binaries_linux,binaries_osx,binaries_win,c2c3_disc_models,course_beijing19,course_harvard21,course_hawaii17,course_london19,course_london20,course_paris18,deepreg_models,deepseg_gm_models,deepseg_lesion_models,deepseg_sc_models,exvivo_template,gm_model,MNI-Poly-AMU,optic_models,PAM50,pmj_models,sct_example_data,sct_testing_data}
```

Maybe we could consider more "human-readable" format.

joshuacwnewton:
Very good suggestion!

Especially because the long list of datasets is basically duplicated with the `choices` section for the actual `-d` argument:

```
usage: sct_download_data -d
                         {binaries_linux,binaries_osx,binaries_win,c2c3_disc_models,course_beijing19,course_harvard21,course_hawaii17,course_london19,course_london20,course_paris18,deepreg_models,deepseg_gm_models,deepseg_lesion_models,deepseg_sc_models,exvivo_template,MNI-Poly-AMU,optic_models,PAM50,pmj_models,sct_example_data,sct_testing_data}
                         [-h] [-o <folder>] [-k] [-v <int>]

Download binaries from the web.

MANDATORY ARGUMENTS:
  -d {binaries_linux,binaries_osx,binaries_win,c2c3_disc_models,course_beijing19,course_harvard21,course_hawaii17,course_london19,course_london20,course_paris18,deepreg_models,deepseg_gm_models,deepseg_lesion_models,deepseg_sc_models,exvivo_template,MNI-Poly-AMU,optic_models,PAM50,pmj_models,sct_example_data,sct_testing_data}
                        Name of the dataset.
```

Instead of just duplicating the list, we can definitely use the first section to provide a more human-readable list. :smile:

joshuacwnewton:
It would be nice to group things into sections, too? Something like:

```
- Binaries
   * binaries_linux
   * binaries_osx
   * binaries_win

- Templates
   * PAM50
   * MNI-Poly-AMU
[...]

Issue #4009, opened 2023-01-23
Replace multiple course dataset downloads ('17-'21) with a single download


joshuacwnewton:
_Originally posted by @jcohenadad in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4007#issuecomment-1400432556_

> Related comment: I think we should have a generic course dataset (eg: sct_course), because:
> - after 10 years we'll end up with 25 course_xx entries, which is confusing for users
> - given that a course is bound to a specific release, the URL to the release can simply be updated, while keeping the same `sct_course` entry point.
> 
> If we decide to do that, it will break backward compatibility, but we shouldn't care because ppl should train on the latest SCT version anyway

joshuacwnewton:
For context, we have already done the work to create a repo for the "generic course dataset": [`sct_tutorial_dataset`](https://github.com/spinalcordtoolbox/sct_tutorial_data/), which is used for both the SCT Course and the SCT Tutorials.

(The repo contains "tutorial releases" (e.g. [`r20220728`](https://github.com/spinalcordtoolbox/sct_tutorial_data/releases/tag/r20220728)) as well as "SCT course releases" (e.g. [`SCT-Course-2021116`](https://github.com/spinalcordtoolbox/sct_tutorial_data/releases/tag/SCT-Course-20211116)).

We just need to update `sct_download_data` to use only the most recent "SCT course release" of `sct_tutorial_data`.

Issue #3901, opened 2022-10-06
Ship testing data via pip


kousu:
From an idea brought up during https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2959:

Currently `sct_testing` will call `sct_download_data -d sct_example_data`. It puts the downloaded dataset in `$SCT_DIR/data/sct_example_data`, where it is likely to be prodded by users, or mangled by the test scripts themselves, and, in the case of a developer working on a long-term checkout of the code, we can't easily be sure if the test data's version matches the active branch.

I want to replace this with a `pip` package.

This would mean that end-users never see the testing data; instead the tests would do `import sct_example_data` and use its `.__path__` to orient themselves. If someone upgrades SCT in place and we have changed the tests + data, we simply bump the version number of `sct-example-data` required, and then the next test will install the updated data.

I have a prototype at https://github.com/kousu/sct_example_data and https://test.pypi.org/project/sct-example-data/. You can install it with

```
pip install -i https://test.pypi.org/simple/ sct-example-data
```

and find its contents with

```
python -m sct_example_data
```

<details><summary>example</summary>

```
p115628@joplin:~/src$ pip install -i https://test.pypi.org/simple/ sct-example-data
Defaulting to user installation because normal site-packages is not writeable
Looking in indexes: https://test.pypi.org/simple/
Collecting sct-example-data
  Downloading https://test-files.pythonhosted.org/packages/67/d0/2b3b7b5cb8f0aed9e439788c7e688087e1ab7e4328aa8c2eb1d045003549/sct_example_data-2018.5.25-py3-none-any.whl (44.3 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 44.3/44.3 MB 27.6 MB/s eta 0:00:00
Installing collected packages: sct-example-data
Successfully installed sct-example-data-2018.5.25
p115628@joplin:~/src$ python -m sct_example_data
SCT example data is in: /home/GRAMES.POLYMTL.CA/p115628/.local/lib/python3.10/site-packages/sct_example_data
p115628@joplin:~/src$ DATA=$(python -m sct_example_data | awk '{ print $NF }')
p115628@joplin:~/src$ echo $DATA
/home/GRAMES.POLYMTL.CA/p115628/.local/lib/python3.10/site-packages/sct_example_data
```

</details>

[...]

PR #4257, opened 2023-10-17
Color installed datasets for `sct_download_data -h`

co-decode:
#### GitHub

#### PR contents

## Description
In PR #4255 it was suggested that the table appended to the argparse help for `sct_download_data` could be colored to indicate whether a dataset was installed.

I have made a simple attempt to check for the existence of each dataset at its default location.

However, this is insufficient if a user has installed to a custom location.

Perhaps calling [`download.py::install_named_dataset()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/23776c374fa183308043f3d95f0017e0e63b51a3/spinalcordtoolbox/download.py#L359C16-L359C16) with a custom `dest_folder` parameter could create or add to a new file that will record custom install paths. This file would be created in a standard location and could then be referenced to accurately reflect whether each dataset has been installed.

<details>
<summary>Image</summary>

image

</details>

## Linked issues
#4007 
#4255

### DIFF
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -357,6 +357,23 @@ def install_data(url, dest_folder, keep=False):
     shutil.rmtree(extraction_folder)
 
 
+def record_custom_folders(dest_folder, dataset_name):
+    """
+    Record custom output folder to track installation status.
+    """
+    custom_paths = os.path.join(__sct_dir__, "data", ".custom_paths")
+    custom_path = dest_folder + "," + dataset_name + "\n"
+    if (os.path.exists(custom_paths)):
+        with open(custom_paths, 'r') as file:
+            lines = file.readlines()
+        if custom_path not in lines:
+            with open(custom_paths, 'a') as file:
+                file.write(custom_path)
+    else:
+        with open(custom_paths, 'w') as file:
+            file.write(custom_path)
+
+
 def install_named_dataset(dataset_name, dest_folder=None, keep=False):
     """
     A light wrapper for the 'install_data' function to allow downstream consumers to download
@@ -368,10 +385,53 @@ def install_named_dataset(dataset_name, dest_folder=None, keep=False):
                          f"{sorted(list(DATASET_DICT.keys()), key=str.casefold)}")
 
     urls = DATASET_DICT[dataset_name]["mirrors"]
+
+    custom = True
     if dest_folder is None:
+        custom = False
         dest_folder = DATASET_DICT[dataset_name]["default_location"]
 
     install_data(urls, dest_folder, keep)
+
+    if custom is True:
+        record_custom_folders(dest_folder, dataset_name)
+
+
+def is_installed(dataset_name, custom_paths):
+    """
+    :return: bool: Checks whether a dataset is installed
+    """
+    # Check for installation at default location
+    path_default = DATASET_DICT[dataset_name]['default_location']
+    if os.path.exists(path_default):
+        return True
+
+    if custom_paths is None:
+        return False
+
+    # Check if a custom installation pathway has been recorded,
+    # and whether it still exists.
+    lines_to_remove = []
+    for line in custom_paths:
+        custom_folder, name = line[:-1].split(",")
+        print("DEBUG", custom_folder,name, dataset_name)
+        if os.path.exists(custom_folder):
+            if name == dataset_name:
+                return True
+        else:
+            lines_to_remove.append(line)
+
+    # If a recorded custom pathway no longer exists,
+    # remove it from the record.
+    if len(lines_to_remove):
+        for line in lines_to_remove:
+            custom_paths.remove(line)
+        custom_paths_file = os.path.join(__sct_dir__, "data", ".custom_paths")
+        with open(custom_paths_file, 'w') as file:
+            for line in custom_paths:
+                file.write(line)
+
+    return False
 
 
 def list_datasets():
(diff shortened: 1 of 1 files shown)


co-decode:
I have included some code that will record downloads that are given a custom output directory to a file `__sct_dir__/data/.custom_paths`.

The command `sct_download_data -h` will now check that record if it cannot find a default installation folder.
In addition, if the directory a record points to is found to no longer exist, that record will be removed from the `.custom_paths` file.

spinalcordtoolbox/download.py
```diff
@@ -367,21 +385,80 @@ def install_named_dataset(dataset_name, dest_folder=None, keep=False):
...
+
+
+def is_installed(dataset_name, custom_paths):
+    """
+    :return: bool: Checks whether a dataset is installed
+    """
+    # Check for installation at default location
+    path_default = DATASET_DICT[dataset_name]['default_location']
+    if os.path.exists(path_default):
+        return True
```
joshuacwnewton, line 407: One additional wrinkle: Our binary files get installed directly to `bin/`, rather than a subfolder (like the other `data/` downloads). This is because `bin/` gets added to the `$PATH`, and so it's easiest to keep all executable files in a single directory.

This presents a problem, though, because `bin/` is used for other files, too. (Namely, the launcher entry points for SCT's CLI scripts). Because `bin/` is a multi-purpose directory for all executable files, it's possible for `bin/` to exist, yet not contain the necessary binaries. Meaning, a simple directory check doesn't quite work.

I'm not quite sure how to handle this without diving into a more complex solution (e.g. tracking the archive contents / hashing the files), which would expand the scope of this PR and get into previous discussions such as:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3380
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2959
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3901

joshuacwnewton (commented):
I have some initial observations, namely that adding this feature is more complicated than I had anticipated.

I'm not sure what the best way forward is between A) tackling a more complex implementation, B) sticking with a simpler implementation. I'm leaning towards B), but would love @mguaypaq's thoughts on my comments below.

co-decode:
Thank you for your thoughts @joshuacwnewton. 
It seems like tracking binary downloads and custom installation paths involves a lot of complexity for minimal gain.

For the scope of this PR, it might be easiest to:

- indicate whether `data/` downloads have been 'installed' to their default location, and 
- to leave the `bin/` downloads colored white, perhaps with a note in the legend explaining why.

This would involve rolling back my 'custom pathway' commit, and excluding datasets with the 'binaries' type from the coloring loop.

I want to be respectful of your time and the integrity of your project, so I fully expect to write some code that's only for the purposes of discussion.

joshuacwnewton:
> to leave the `bin/` downloads colored white, perhaps with a note in the legend explaining why.

I think this makes a lot of sense! For the note, we could say something like:

```
DATASET NAME                  TYPE
--------------------------------------------------
binaries_linux *              Binaries
binaries_osx *                Binaries
binaries_win *                Binaries
[...]
MNI-Poly-AMU                  Templates
exvivo_template               Templates
PAM50_normalized_metrics      Templates
sct_example_data              Testing Datasets
sct_testing_data              Testing Datasets

Legend: installed | not installed

* Binaries are automatically installed during 
  installation, depending on your operating system.
```

mguaypaq:
@co-decode thanks for this contribution! I'm sorry that my suggestion on #4255 turned out to be so full of corner cases.

@joshuacwnewton wrote:
> I'm leaning towards B), but would love @mguaypaq's thoughts on my comments below.

I agree with your analysis. I think the most useful (and simple) thing to do is to just track whether the non-binary datasets are installed in their default location, with a simple directory check. Hopefully it's not too verbose, but I would suggest a legend like:
```
Legend: installed | not installed (in the $SCT_DIR/data folder)
```
Maybe there's no need to explain the plain formatting for the binaries (as long as they're not red!).

co-decode:
As mentioned, I've reverted the changes relating to `.custom_paths` and left the Binary-type downloads white.

<details>
<summary>Image</summary>

image

</details>

spinalcordtoolbox/download.py
```diff
@@ -375,13 +376,25 @@ def install_named_dataset(dataset_name, dest_folder=None, keep=False):
...
-    :returns: A table containing listing the downloadable datasets
+    :returns: A table listing the downloadable datasets
     :rtype: str
     """
+    color = {True: 'LightGreen', False: 'LightRed'}
     table = f"{'DATASET NAME':<30s}{'TYPE':<20s}\n"
     table += f"{'-' * 50}\n"
-    sortedDatasets = sorted(DATASET_DICT, key=lambda k: DATASET_DICT[k]['download_type'])
+    sortedDatasets = sorted(DATASET_DICT,
+                            key=lambda k: DATASET_DICT[k]['download_type'])
```
joshuacwnewton, line 386: Slight nitpick: In the dataset ordering, currently we sort just by download type (and not name), so PAM50 and PAM50_normalized_metrics aren't beside one another.

At first, I thought to sort by name within categories:

```suggestion
    sortedDatasets = sorted(DATASET_DICT,
                            key=lambda k: DATASET_DICT[k]['download_type'] + k)
```

However, this un-sorts the list of course datasets, since their years (`17`, `18`, `19`, etc.) come after their names. But, maybe this is okay, given that we plan to tackle https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4009 anyway.

spinalcordtoolbox/download.py
```diff
@@ -375,13 +376,25 @@ def install_named_dataset(dataset_name, dest_folder=None, keep=False):
...
     for dataset_name in sortedDatasets:
         download_type = DATASET_DICT[dataset_name]['download_type']
-        table += f"{dataset_name:<30s}{download_type:<20s}\n"
+        dataset_status = dataset_name.ljust(30)
+        if download_type != "Binaries":
+            path_dataset = DATASET_DICT[dataset_name]['default_location']
+            installed = os.path.exists(path_dataset)
```
joshuacwnewton, line 392: Another nitpick: Perhaps we could also check if the folder is non-empty, by doing:

```suggestion
            installed = (os.path.exists(path_dataset)
                         and len(os.listdir(path_dataset)) > 0):
```

Still not a foolproof check, but it covers the case where the dataset folder was created but the actual downloading/unzipping process failed somehow.

spinalcordtoolbox/download.py
```diff
@@ -375,13 +376,25 @@ def install_named_dataset(dataset_name, dest_folder=None, keep=False):
...
+    :returns: A table listing the downloadable datasets
     :rtype: str
     """
+    color = {True: 'LightGreen', False: 'LightRed'}
     table = f"{'DATASET NAME':<30s}{'TYPE':<20s}\n"
     table += f"{'-' * 50}\n"
-    sortedDatasets = sorted(DATASET_DICT, key=lambda k: DATASET_DICT[k]['download_type'])
+    sortedDatasets = sorted(DATASET_DICT,
+                            key=lambda k: DATASET_DICT[k]['download_type'])
     for dataset_name in sortedDatasets:
```
joshuacwnewton, line 387: Tiniest style nitpick ever: Switch to snake_case? (`sorted_datasets`)

joshuacwnewton (approved):
This LGTM! I just had some tiny tiny nitpicks, but as soon as they're addressed, I think this is good to go. :)

co-decode:
Changes implemented as suggested. :+1:

joshuacwnewton:
Thank you so much for your continued contributions, @co-decode! As always, it's a pleasure to work with you. :relaxed:

Issue #4266, opened 2023-10-24
Registration PAM50 Template -> native subject space fails (silently)


jqmcginnis:
### Description

Instead of registering the subject to the PAM50 template, the [SCT - customizing registration tutorial](https://spinalcordtoolbox.com/user_section/tutorials/registration-to-template/template-registration/customizing-registration.html) recommends registering the template to the subject with the `-ref subject` flag instead. According to the documentation, this is ultimately beneficial for highly anisotropic data ("Use this approach if your image is acquired axially with highly anisotropic resolution (e.g. 0.7x0.7x5mm), because the straightening step can produce through-plane interpolation errors for thick slices."). 

When I use the default settings of `sct_register_to_template`, the command fails silently when I add the `-ref subject` flag. 
When I increase the debug capabilities (-v 2), I get some error messages.

### Steps to Reproduce
I tested everything on SCT v6.0. Take a whole-spine image (axial resolution e.g. 0.5x0.5x5), in my case e.g. sub-m030451 (I have sent this private subject to Neuropoly [before](https://forum.spinalcordmri.org/t/pam50-template-registration-for-highly-anisotropic-axial-scans/1117), let me know if I should resend it, in case you do not have it anymore.). 

`sct_register_to_template -ref subject -i sub-m030451_ses-20210708_acq-ax_T2w.nii.gz -s sub-m030451_ses-20210708_acq-ax_seg.nii.gz -ldisc sub-m030451_ses-20210708_acq-ax_disks.nii.gz -c t2 -ofolder sub-m030451_reversed -v 2`

I obtain the following debug output. Please note, that the registration works (with suboptimal performance though) if I omit the `-ref subject` command (which defaults to `-ref template`).  

**Expected behavior:** 

The registration of template -> subject finishes without any errors, not considering the registration performance here (i.e. how well aligned the images are.)

**Actual behavior:** 

My interpretation would be that the registration fails (not due to problems in the registration), but due to some bug in the routine. 
<details>
<summary>Debug Output</summary>

```
jmcginnis@pc:~/Desktop/PAM50_Template_Registration$ sct_register_to_template -ref subject -i sub-m030451_ses-20210708_acq-ax_T2w.nii.gz -s sub-m030451_ses-20210708_acq-ax_seg.nii.gz -ldisc sub-m030451_ses-20210708_acq-ax_disks.nii.gz -c t2 -ofolder sub-m030451_reversed -v 2 

--
[...]

PR #4267, opened 2023-10-24
Switch matmul syntax (`*` -> `@`) to follow-up previous `np.matrix` -> `np.array` change

joshuacwnewton:
## Description

This change is necessary is to accommodate the previous change of the `R` array from an `np.matrix` object to an `np.array` object. (See #4091, specifically https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/dee9e9fa90dedf44d6c7819d835996dd3abdd5ef.)

## Linked issues

This fixes the error in #4266, but does not address @jqmcginnis's failed registration, so we shouldn't close the issue quite yet.

### DIFF
--- a/spinalcordtoolbox/registration/algorithms.py
+++ b/spinalcordtoolbox/registration/algorithms.py
@@ -862,8 +862,8 @@ def register2d_centermassrot(fname_src, fname_dest, paramreg=None, fname_warp='w
         # display rotations
         if verbose == 2 and not angle_src_dest[iz] == 0 and not rot_method == 'hog':
             # compute new coordinates
-            coord_src_rot = coord_src[iz] * R
-            coord_dest_rot = coord_dest[iz] * R.T
+            coord_src_rot = coord_src[iz] @ R
+            coord_dest_rot = coord_dest[iz] @ R.T
             # generate figure
             plt.figure(figsize=(9, 9))
             # plt.ion()  # enables interactive mode (allows keyboard interruption)


mguaypaq (commented):
Thanks for tracking this down. The fix looks fine, but I'd like to take a second look at the files touched by dee9e9fa90dedf44d6c7819d835996dd3abdd5ef, to make sure that we didn't miss any other instances.

mguaypaq:
After looking at the rest of dee9e9fa90dedf44d6c7819d835996dd3abdd5ef, I did find another missed instance of `* -> @`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b33cc812d7f8153d45bf575688f0b4927a6a8ab1/spinalcordtoolbox/centerline/nurbs.py#L850-L852

But! Looking at `git log -p`, I can't find any evidence that the enclosing function `reconstructGlobalInterpolation` has _ever_ been called by SCT code. So maybe the correct fix is to delete this vestigial (and probably broken) function.

joshuacwnewton:
> But! Looking at `git log -p`, I can't find any evidence that the enclosing function `reconstructGlobalInterpolation` has _ever_ been called by SCT code. So maybe the correct fix is to delete this vestigial (and probably broken) function.

Wow! This method has been present for 9+ years, since the very first commit in the SCT repository (10f9a5d5), back in the Python 2.0 days when we had an `sct_nurbs` scripyt. 

It looks like the method used to be something that was intended to be called via SCT's Python API back then:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/10f9a5d5628669c7a8cdbe565a5b57b50c72b488/scripts/sct_nurbs.py#L31-L35

But, like you, I strongly suspect that this method was never used in internal SCT code, and was dutifully carried forward in time through multiple rewrites and refactors. So, I think it's very much safe to remove. :)

PR #4271, opened 2023-10-25
Add current `sct_tutorial_data` release as a new course dataset

### LINKED ISSUE #4009: Replace multiple course dataset downloads ('17-'21) with a single download

joshuacwnewton:
## Description

This PR allows us to write the SCT Course slides as follows:

```bash
# Go to your home folder (or another if you prefer)
cd ~

# Download course material (requires Internet and working SCT)
sct_download_data -d course_virtual23 -o sct_course_data

# Go inside course data folder
cd sct_course_data

# Inspect contents using `ls`
ls
  multi_subject  README.md  single_subject  tutorial-datasets.csv

# Go inside the ‘single_subject’ folder
cd single_subject
```

(Before, we had steps that explicitly used `curl`/`wget`/`unzip`/etc., which is not suitable for Windows users.)

> **Note**: This needs to be merged before 6.1 is released, otherwise users won't be able to use 6.1 to download the course dataset.
>
> However, if we were to merge all of the course datasets into a single download, then theoretically it should be viable to point to the "latest" sct_tutorial_data download, which would eliminate the need to repeatedly update the link in step with SCT releases.
> 
> That said, maybe we _want_ to freeze the course datasets, so that we avoid messing up old course instructions with future `sct_tutorial_data` changes.

## Linked issues

Fixes #4009. (Should we do this instead?)


jcohenadad:
> (Should we do this instead?)

I'm sorry I'm not sure I understand what you mean by "this"?

jcohenadad:
> We could take care of two birds with one stone by having a "general" sct_course_data entry in sct_download_data?

ahhh gotcha! yes, a general sct_course_data would be better, I think. That ensures users only download the data compatible/tested with the version of SCT

mguaypaq (approved):
LGTM, and the download link works.

jcohenadad (commented):
When unzipped, the folder is called: `sct_tutorial_data-r20231011`. Is this something we want, given the tutorial:

image

note: we can (and will) obviously update the course gslide, but i was wondering if we want to keep the release data (i see pros and cons...)

joshuacwnewton:
> When unzipped, the folder is called: `sct_tutorial_data-r20231011`. Is this something we want, given the tutorial:

It's true that directly downloading the zip from the URL results in this (because GitHub automatically generates this folder name from the repo + release names). 

However, `sct_download_data` automatically names the folder and extracts the contents. So, the current course instructions (e.g. `sct_download_data -d course_virtual23 -o sct_course_data && cd sct_course_data`) should work as-is? Unless I've missed something.

PR #4273, opened 2023-10-25
Fix typo in "Other shape metrics" tutorial

valosekj:
@joshuacwnewton, sorry for spamming, but I discovered a typo in the "Other shape metrics" tutorial.


joshuacwnewton:
No worries at all about spamming! Thanks for catching this. :)

PR #4250, opened 2023-10-09
Add tutorial for lumbar segmentation and registration

joshuacwnewton:
## Description

This tutorial compiles the recommendations from the forum posts into a web tutorial. A [corresponding section](https://docs.google.com/presentation/d/1t40Fd0fS0SwWR5FU_GWKEvHkB9d96LVddLQW6L3QAx0/edit#slide=id.g24c79a1b6f2_0_152) has also been added to the SCT Course slides.

> Note: You will need to reinstall the PAM50 template from this branch in order to get the necessary label `60` for registration.

You can find the tutorial for this PR here: https://spinalcordtoolbox--4250.org.readthedocs.build/user_section/tutorials/registration-to-template/lumbar-registration.html

## Linked issues

Fixes #4180.
Fixes #3599 (once https://github.com/spinalcordtoolbox/PAM50/pull/27 is merged, anyway!)

PR #4279, opened 2023-10-31
Fix broken link

mguaypaq:
See this page for the correct syntax:

https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref

Sorry for introducing this broken link in [my earlier suggestion](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4250#discussion_r1374998121)!


joshuacwnewton (approved):
Link works for me now! :)

PR #4281, opened 2023-11-01
Ensure that point label files are correctly interpolated during template warping

### LINKED ISSUE #4280: Warping of `PAM50_spinal_midpoint` is wrong
### Description

going through the SCT course-- i ran the warping of the template and got this:

~~~
julien-macbook:~/sct_course_data/single_subject/data/fmri $ sct_label_utils -i label/template/PAM50_spinal_midpoint.nii.gz -display

--
Spinal Cord Toolbox (git-master-2f402110657afb6c4266e65b852b1958a054c1b6)

sct_label_utils -i label/template/PAM50_spinal_midpoint.nii.gz -display
--

Position=(67,10,6) -- Value= 0.0019879301918479053
All labels (useful syntax):
67,10,6,0.0019879301918479053
~~~

which is unexpected-- i expected values 2, 3, 4, etc.

i think this is related to a wrong interpolation used when warping the template-- 

instructions: https://docs.google.com/presentation/d/1t40Fd0fS0SwWR5FU_GWKEvHkB9d96LVddLQW6L3QAx0/edit#slide=id.p115

This is HIGH PRIORITY-- sorry i have no time to look into this

joshuacwnewton:
## Description

Currently, either `linear` or `nn` interpolation is used when warping template files (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4280#issuecomment-1789177326). But, neither interpolation method is appropriate for label files _(`nn` may result in lost labels, while `linear` results in the label values themselves not being preserved (int `3.0` -> float `1.0234674`))_. This is why `sct_apply_transfo` has a specific interpolation setting called `label`, so we should use it here.

I've written a test that will fail for the current interpolation, but will pass when interpolation is switched to `label`, to demonstrate that my changes fix the issue.

## Linked issues

Fixes #4280.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_warp_template.py
+++ b/spinalcordtoolbox/scripts/sct_warp_template.py
@@ -128,11 +128,9 @@ def get_interp(file_label):
     # default interp
     interp = 'linear'
     # Nearest Neighbours interp
-    if any(substring in file_label for substring in [
-        '_level.nii.gz', '_levels.nii.gz', '_csf.nii.gz', '_CSF.nii.gz', '_cord.nii.gz'
-    ]):
+    if any(substring in file_label for substring in ['_levels.', '_csf.', '_cord.']):
         interp = 'nn'
-    elif any(substring in file_label for substring in ['_label', '_midpoint']):
+    elif any(substring in file_label for substring in ['_label_', '_midpoint.']):
         interp = 'label'
     # output
     return interp
--- a/spinalcordtoolbox/scripts/sct_warp_template.py
+++ b/spinalcordtoolbox/scripts/sct_warp_template.py
@@ -31,15 +31,13 @@ class Param:
         # self.warp_template = 1
         self.warp_atlas = 1
         self.warp_histo = 0
-        # list of files for which nn interpolation should be used. Default = linear.
-        self.list_labels_nn = ['_level.nii.gz', '_levels.nii.gz', '_csf.nii.gz', '_CSF.nii.gz', '_cord.nii.gz']
         self.verbose = 1  # verbose
         self.path_qc = None
 
 
 class WarpTemplate:
     def __init__(self, fname_src, fname_transfo, warp_atlas, folder_out, path_template,
-                 folder_template, folder_atlas, file_info_label, list_labels_nn,
+                 folder_template, folder_atlas, file_info_label,
                  verbose, warp_histo, folder_histo):
 
         # Initialization
@@ -53,7 +51,6 @@ class WarpTemplate:
         self.folder_atlas = folder_atlas
         self.folder_histo = folder_histo
         self.file_info_label = file_info_label
-        self.list_labels_nn = list_labels_nn
         self.verbose = verbose
 
         # printv(arguments)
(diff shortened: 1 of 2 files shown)


mguaypaq (commented):
Thanks for the fix, and for adding a test :heart:

I agree that this will catch the four relevant files from the PAM50 template:
```
PAM50_label_body.nii.gz
PAM50_label_disc.nii.gz
PAM50_label_discPosterior.nii.gz
PAM50_spinal_midpoint.nii.gz
```
I couldn't help but notice that `get_interp()` uses a different mechanism for the files which need nearest neighbor interpolation, namely having a `list_labels_nn` parameter. But this parameter has a single, non-modifiable source point, is really used only once (in this method), but has to be threaded through the entire call stack to reach `get_interp()`. For the sake of consistency, I suggest replacing `list_labels_nn` by an explicit list, like you did here. If you don't object, I can push an extra commit doing it.

I might also take this chance to review the strings in `list_labels_nn`, since it looks like there are some that are unused by the PAM50 template. Unless they're meant to magically work with other, custom templates?

mguaypaq (approved):
Alright, I added my suggestions as a pair of extra commits. Feel free to revert them if you'd rather stick to the initial, surgical fix. For now I'll pre-approve the PR.

joshuacwnewton:
> Unless they're meant to magically work with other, custom templates?

This was my guess at the intent, too, given that this list of substrings was added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/c99041e992ee5878120c12c25c9665bb12bc906a, which has the commit message "REF: warp_template: now handles any template (adapts interpolation type)". Implying that, if the user supplies a different template (with certain magic substrings), the interpolation behavior will also adapt.

I don't know if this logic is communicated clearly to the user, though. If I were supplying my own custom template, I would honestly be a bit mystified at how the interpolation methods were chosen unless I specifically looked at the source code.

spinalcordtoolbox/scripts/sct_warp_template.py
```diff
@@ -120,20 +116,22 @@ def warp_label(path_label, folder_label, file_label, fname_src, fname_transfo, p
...
 # ==========================================================================================
-def get_interp(file_label, list_labels_nn):
+def get_interp(file_label):
     # default interp
     interp = 'linear'
     # Nearest Neighbours interp
-    if any(substring in file_label for substring in list_labels_nn):
+    if any(substring in file_label for substring in ['_levels.', '_csf.', '_cord.']):
         interp = 'nn'
+    elif any(substring in file_label for substring in ['_label_', '_midpoint.']):
```
joshuacwnewton, line 133: Thank you for making this change! I agree with the reasoning outlined in commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4281/commits/c97eb5e355e23e01bcf3e868847d056d6328d505.

```suggestion
    # For safety and consistency, ensure strings are bracketed by `_` or `.` on both sides
    if any(substring in file_label for substring in ['_levels.', '_csf.', '_cord.']):
        interp = 'nn'
    elif any(substring in file_label for substring in ['_label_', '_midpoint.']):
```

joshuacwnewton: I also quickly manually double-checked this by setting breakpoints within the `elif` cases, then ensuring that they were triggered appropriately. This all works well!

Issue #4287, opened 2023-11-04
Add color to sct_run_batch output


NathanMolinier:
## Description

I just ran the command `sct_run_batch -script process_data.sh -path-data data/ -path-output output -jobs 3` when testing the course and I noticed that the output given was not as colorful as other sct functions:
<img width="569" alt="Screenshot 2023-11-04 at 5 03 36 PM" src="https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/68945192/1f0d042b-7e31-4d3e-b0a5-c35e7fe251e9">

I believe we could be more consistent and print the open command in green !

joshuacwnewton:
For other SCT functions, we call the `generate_sc()` function, which in turn calls the `display_open` function, which in turn calls `printv()` with `type='info'` (which triggers the LightGreen color).

Color will only get applied if `isatty()` is true:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7c1085297d7c53e3316c497b20e536d0ff182432/spinalcordtoolbox/utils/sys.py#L379-L382

So, presumably `isatty()` is returning false for `sct_run_batch`. Strange! (And, we can double-check this by removing the `isatty()` check.)

----

To debug this, I print `file` and `file.isatty()`, then compare:

```python
# Other SCT scripts (e.g. sct_deepseg_sc)
File: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
File is atty: True

# sct_run_batch
File: <spinalcordtoolbox.utils.fs.Tee object at 0x7f35e202bf70>
File is atty: False
```

Aha! There's our smoking gun:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7c1085297d7c53e3316c497b20e536d0ff182432/spinalcordtoolbox/scripts/sct_run_batch.py#L432-L437

----

There are a number of potential fixes here (restoring the original stdout, making sure `Tee` returns `True` for `isatty()` , etc.) I think I like modifying Tee, because it preserves the Tee functionality throughout all of sct_run_batch, while also having a minimal effect on SCT as a whole (Tee is currently only ever used in the context of sct_run_batch).

joshuacwnewton:
> Since this is a last minute (tiny) fix, and since it's a Sunday before the SCT v6.1 release, I think this is safe to commit directly to master, as to not block the release through reviews/revisions. But, I'd be happy to discuss alternate solutions for 6.2?

[...]

PR #4288, opened 2023-11-05
Filter out color codes from log files

mguaypaq:
With the last-minute fix from #4287, color codes will ~start to~ appear in the text log files created by `sct_run_batch`, as a compromise, so that they can also appear in the terminal window. This patch-upon-a-patch quick fix tries to have it both ways: output color codes to the terminal, and leave them out of the text log files.

~It's not elegant, and not the "right" fix (we should properly fix the Tee class for 6.2), but I think clean having clean text log files is important enough to warrant it.~ EDIT: After testing it manually, this PR doesn't prevent color codes from appearing in the text log files. But let's not hold up the 6.1 release just for this (very minor) point.

Ha! And after testing manually _before_ f5a46f328fe797b3d7c0e3844e17ad3f8add5ee1, I see that color codes have always appeared in the text log files! So they might as well appear in the terminal as well :sweat_smile:


spinalcordtoolbox/utils/fs.py
```diff
@@ -47,8 +52,18 @@ def close(self):
...
 
     def write(self, text):
-        self.fd1.write(text)
-        self.fd2.write(text)
+        # Quick and dirty solution to filter out ansi escape sequences (like color codes) from
+        # the text if one file is an interactive terminal but the other one is not.
+        # This will only filter out cases where the entire escape sequence is part of a single
+        # call to write(). See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4287
+        if self.fd2.isatty() and not self.fd1.isatty():
+            self.fd1.write(ansi_escape.sub('', text))
```
joshuacwnewton, line 60: From what I can tell, this _does_ work for me!

Terminal output:

image

Output to `sct_run_batch_log.txt` (`master`):

```
Your batch completed but some subjects may have not completed successfully, please consult the logs for:
sub-01
sub-03
sub-05

Started: 2023-11-07 15h07m13s | Ended: 15h07m13s | Duration: 00h00m00s

[0m
To open the Quality Control (QC) report in a web-browser, type:[0m
[92mxdg-open /home/joshua/repos/spinalcordtoolbox/data/sct_course_data/multi_subject/output/qc/index.html
[0m
```

Output to `sct_run_batch_log.txt` (`mgp/tee-quick-fix`):

```
Your batch completed but some subjects may have not completed successfully, please consult the logs for:
sub-01
sub-03
sub-05

Started: 2023-11-07 15h04m53s | Ended: 15h04m54s | Duration: 00h00m00s


To open the Quality Control (QC) report in a web-browser, type:
xdg-open /home/joshua/repos/spinalcordtoolbox/data/sct_course_data/multi_subject/output/qc/index.html
```

spinalcordtoolbox/utils/fs.py
```diff
@@ -28,6 +29,10 @@ def tmp_create(basename):
     return tmpdir
 
 
+# Taken from https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python/14693789#14693789
+ansi_escape = re.compile(r'\x1b(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])')
```
joshuacwnewton, line 33: Side note: The SO answer here contains the following snippet:

```python
# 7-bit C1 ANSI sequences
ansi_escape = re.compile(r'''
    \x1B           # ESC
    (?:            # Case 1. 7-bit C1 Fe escape sequences (except CSI)
        [@-Z\\-_]  #   - Byte in the range 0x40 to 0x5F
    |              # Case 2. Control Sequence Introducer (CSI) escape sequences
        \[         #   - [
        [0-?]*     #   - Parameter bytes
        [ -/]*     #   - Intermediate bytes
        [@-~]      #   - Final byte
    )
''', re.VERBOSE)
```

I didn't realize that regex could be written in this [verbose form](https://www.geeksforgeeks.org/verbose-in-python-regex/), and I like it a lot?

spinalcordtoolbox/utils/fs.py
```diff
@@ -47,8 +52,18 @@ def close(self):
...
+        # This will only filter out cases where the entire escape sequence is part of a single
+        # call to write(). See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4287
+        if self.fd2.isatty() and not self.fd1.isatty():
+            self.fd1.write(ansi_escape.sub('', text))
+        else:
+            self.fd1.write(text)
+        if self.fd1.isatty() and not self.fd2.isatty():
+            self.fd2.write(ansi_escape.sub('', text))
+        else:
+            self.fd2.write(text)
```
joshuacwnewton, line 66: If I understand correctly, the logic was written this way such that, if both FDs are _not_ ttys, then the control sequences must have been intentional, so print them to both?

It took me a bit of thinking to untangle the different cases here. Maybe we could make the cases explicit? It doesn't take too many extra lines:

```suggestion
        # If one FD is a TTY and the other is not, remove escape sequences from the non-TTY
        # Otherwise, if both/neither FDs are TTYs, then preserve escape sequences
        text_fd1, text_fd2 = text, text
        if not self.fd1.isatty() and self.fd2.isatty():
           text_fd1 = ansi_escape.sub('', text)
        elif not self.fd2.isatty() and self.fd1.isatty():
           text_fd2 = ansi_escape.sub('', text)
        self.fd1.write(text_fd1)
        self.fd2.write(text_fd2)
```

joshuacwnewton (approved):
Given that this works for me (see previous review), I think this could be merged? I just had one minor nitpick.

Though, I see this is still a draft PR, so perhaps you had plans to rework the strategy entirely?

mguaypaq:
Ok, I tried implementing what I thought might be a nicer fix: using Python's codecs mechanism to implement this kind of filtering as a file codec, that we could use when opening the log file. I felt like this would have been a cleaner separation of responsibilities than jamming some filtering code inside the `Tee` class, whose name indicates that it should just be a simple "one write in, two writes out" wrapper.

Here's the result: https://gist.github.com/mguaypaq/e0b70d874c9f81423ffd582bd6787438

This was trickier than I expected, and the codec interfaces seem like they just slightly don't fit nicely. (Very much so in the case of `StreamReader`, which I gave up implementing. It would have been entirely too large.)

So, I'm willing to go with the solution we already have in this draft PR as good enough and concise enough.

joshuacwnewton:
While the `codecs` module is new to me, this part:

> I felt like this would have been a cleaner separation of responsibilities than jamming some filtering code inside the `Tee` class, whose name indicates that it should just be a simple "one write in, two writes out" wrapper.

Is very convincing to me! So, I wanted to spend a little thought on how to avoid mixing filtering code with `Tee`:

<details>
<summary> Thoughts about how to make a `codec`-based solution work </summary>

For your implementation, we would then set `csi_filter` at the caller site, then pass the file objects to `Tee`. For our current usage in `sct_run_batch`, this would be something like:

```diff
+   import csi_filter
+   csi_filter.register()

-   batch_log = open(os.path.join(path_log, arguments.batch_log), 'w')
+   batch_log = open(os.path.join(path_log, arguments.batch_log), 'w', encoding='csi_filter')

    # Tee IO to batch_log and std(out/err)
    orig_stdout = sys.stdout
    orig_stderr = sys.stderr

    sys.stdout = Tee(batch_log, orig_stdout)
    sys.stderr = Tee(batch_log, orig_stderr)
```

This definitely feels like a cleaner solution? It does away with the weird `isatty` logic, since we instead leave the question of "whether to filter" to the caller, rather than guessing from within `Tee`.

----

I guess the tricky part, then, is the actual implementation of the codec. Looking at what you've done, it seems like it goes far beyond the simple regex filter we started with? There are all these aspects that I don't recognize, and I would have to read the `codecs` module to fully understand.

But, still, my immediate question is: "Could we partially implement a codec that does only the minimal filtering necessary for `sct_run_batch`, even if it couldn't be used as a "true" Python codec?" And, I came across [this SO answer](https://stackoverflow.com/a/52132999) which seems to focus on the "minimal" part.

</details>

<details>
<summary> Counter-thoughts about Tee being inherently tied to the color code behavior of printv </summary>

> I felt like this would have been a cleaner separation of responsibilities than jamming some filtering code inside the Tee class, whose name indicates that it should just be a simple "one write in, two writes out" wrapper.

I notice that `Tee` already inherently has some filter-based code, even prior to this PR, that goes beyond the scope of "simple one in two out":

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/9a7ecad0d3ac019f0f84ff239685a84381195daa/spinalcordtoolbox/utils/fs.py#L57-L64

I then looked at printv, and:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/9a7ecad0d3ac019f0f84ff239685a84381195daa/spinalcordtoolbox/utils/sys.py#L379-L384

Huh. This feels like it's doing the inverse of what our filtering behavior is doing. 

I tried to reason whether there's a way to combine all of this logic somehow to prevent doing this unnecessary "pretend to be a tty to add color codes, then remove them after the fact" since all of this logic seems to be coupled. But, as long as we're assigning `Tee` to `sys.stdout`, I don't see a way around it.

</details>

----

OK, how about this: Since we only use the `Tee` class in one place, and we only use it to log `stdout`/`stderr` to a file... what if we intentionally made the `Tee` class less general?

I was inspired by [this simpler `Logger` class](https://stackoverflow.com/a/616672), which is like `Tee`, but makes the assumption that one fd will always be a file, and one fd will always be terminal output. In that case, we could universally filter the text sent to the log file without it feeling _as_ out of place as it did for the `Tee` class? This would also remove the need for the complex "only filter if one `isatty()` and one isn't" conditions, too, since we wouldn't be dealing with general `fd1` and `fd2` objects anymore.

mguaypaq:
Ah! Thanks to the resources you found, I got a much shorter implementation of the codec:

https://gist.github.com/mguaypaq/e0b70d874c9f81423ffd582bd6787438

joshuacwnewton:
> Ah! Thanks to the resources you found, I got a much shorter implementation of the codec:

Lovely! I think I'd be on board for using a codec-based solution, then? I just had a few further questions:

- I notice that in the "minimal codec" link I shared, they don't bother at all to pass incremental/stream to `CodecInfo`. Are they still necessary for us?
- Likewise, is it necessary for us to have both encode and decode if we're only using `open(encoding=)`?
- Also, I notice that there's a new regex called `csi_suffix`, which seems to make the distinction of adding ` \x1b $` as a new case. Did you find that, in practice, there would be times where a CS introducer would be present without the required final byte?

mguaypaq:
> * I notice that in the "minimal codec" link I shared, they don't bother at all to pass incremental/stream to `CodecInfo`. Are they still necessary for us?

That example only uses the custom codec through the functions `codecs.encode` and `codecs.decode`, which don't need the incremental/stream implementations. But (I just tested this) using a registered codec when opening a file requires at least the stream implementation (and at that point, the incremental is just a few extra lines of boilerplate).

> * Likewise, is it necessary for us to have both encode and decode if we're only using `open(encoding=)`?

Technically, we're only using the `encode` direction, since we only ever open files with this codec _for writing_. But since we have the regexes and infrastructure handy, and it's only a few extra lines, I figured we might as well comply with the full spec?

> * Also, I notice that there's a new regex called `csi_suffix`, which seems to make the distinction of adding ` \x1b $` as a new case. Did you find that, in practice, there would be times where a CS introducer would be present without the required final byte?

I'm not sure in practice, but certainly in theory it's an important case. The idea with the incremental/stream interfaces is that you don't get to process the whole string all at once, only bits and pieces cut at arbitrary intervals (maybe because of buffering, or reading an entire file block at a time from disk, etc.). So, you have to recognize when to stop processing and say "the next part is ambiguous, it depends on the subsequent characters".

(Incidentally, this is why the basic encode/decode functions return a pair (encoded output, length of input consumed): it's so that the caller can shove the remaining part of the input into a buffer until the next call.)

So, the `csi_suffix` regex is what I used to detect this. It matches exactly the final characters of the string which may be the start of a CSI sequence, or not. This includes three cases:
1. The empty string (for convenience, so that the regex always matches).
2. Just the first character, ESC.
3. 1. Just the first 2 characters, ESC [.
   2. ESC [ followed by some parameter bytes.
   3. ESC [, some parameter bytes, some intermediate bytes (but _not_ the final byte!)

(That looks like 5 cases, but the last 3 cases are covered by just one branch of the `csi_suffix` regex.)

spinalcordtoolbox/utils/fs.py
```diff
@@ -28,23 +29,23 @@ def tmp_create(basename):
     return tmpdir
 
 
+# Taken from https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python/14693789#14693789
+ansi_escape = re.compile(r'\x1b(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])')
+
+
```
mguaypaq, line 35: I forgot to remove this, sorry:
```suggestion
```

spinalcordtoolbox/utils/csi_filter.py
```diff
@@ -0,0 +1,100 @@
...
+)
+
+# The last part of the input which *might* start a CSI escape sequence.
+# This matches every possible *incomplete* CSI escape sequence, including
+# the empty string for convenience (so that the regex always matches).
+csi_suffix = re.compile(
+    r"""
+    \Z
+    | \x1b \Z
+    | \x1b \[ [0-?]* [ -/]* \Z
```
mguaypaq, line 39: TIL (possibly not for the first time) that in Python regex, `$` can match at places other than the end of the string, even without `re.MULTILINE`! Specifically, it can match _one character before the end of string_ as well, if the last character is a newline. But in this case, I really wanted the end of string.

I noticed this because the little example I wrote at the top of the file didn't work at first: the newline character at the end got swallowed by the encoder.

spinalcordtoolbox/utils/csi_filter.py
```diff
@@ -0,0 +1,100 @@
...
+    def _buffer_decode(self, input: bytes, errors: str, final: bool) -> tuple[str, int]:
+        return CsiFilterCodec.decode(input, errors)
+
+
+class CsiFilterStreamReader(CsiFilterCodec, codecs.StreamReader):
+    pass
+
+
+class CsiFilterStreamWriter(CsiFilterCodec, codecs.StreamWriter):
+    pass
```
joshuacwnewton, line 81: If it's not too much to ask, could you add a small comment explaining the purpose of these classes?

I presume that subclassing both the codec and the base `Stream` classes is enough to meet the requirements of `CodecInfo`, but it's not clear to me on first glance why that is. (The docs for [`CodecInfo.streamreader`](https://docs.python.org/3/library/codecs.html#codecs.CodecInfo.streamwriter) aren't entirely clear to me, either?)

My confusion is basically: Are these classes non-functional, with the purpose of them being to trick `CodecInfo` into not throwing an error, with the assumption that we won't actually use the Stream parts of this codec? Or does subclassing CsiFilterCodec actually provide the necessary functionality?

joshuacwnewton (approved):
Sorry for the delay in reviewing this! 

The changes you've made seem very sensible to me, and the color-code filtering works for me on some sample `sct_run_batch` runs. :)

spinalcordtoolbox/utils/csi_filter.py
mguaypaq: Sure! I'll add a clarifying comment. But in short: these classes are completely functional (and in fact they're the ones used by `open(encoding=...)` behind the scenes). The `codecs.StreamReader/Writer` classes provide almost complete implementations, just missing `encode` and `decode` methods, which `CsiFilterCodec` provide.

PR #4292, opened 2023-11-10
Remove unused `Image.verbose` attribute

mguaypaq:
A tiny tiny PR, but I noticed it once and it bothered me: the attribute `Image.verbose` is saved in `__init__`, but is never actually read anywhere. Instead, the pattern is that individual methods of `Image` which can be chatty accept a `verbose` parameter.

(To be sure that no one at all uses `Image.verbose`, I searched all python files for the string `.verbose`. All instances are either: `self.verbose` in another class, `param.verbose` in various scripts, `SpinalCordStraightener.verbose`, `Transform.verbose`, or `ImageCropper.verbose`.)

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -274,8 +274,6 @@ class Image(object):
         if absolutepath is not None:
             self._path = os.path.abspath(absolutepath)
 
-        self.verbose = verbose
-
         # Case 1: load an image from file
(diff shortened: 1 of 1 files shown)


joshuacwnewton (approved):
I checked using PyCharm's "Find Usages" dialog, and it largely returned the same as your `.verbose` check. :)

PR #4294, opened 2023-11-10
Replace assertions with more appropriate errors

mguaypaq:
This PR replaces most instances of `assert` in SCT with `raise`.

Apart from unit tests, assertions are meant for internal invariants, and are not even guaranteed to be run (for example, if python is run with `-O`).

For bad script arguments, calling `parser.error` is better. For bad function arguments, usually raising `ValueError` or `TypeError` is better.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
+++ b/spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
@@ -274,7 +274,8 @@ class ComputeDistances:
         nx1, ny1, nz1, nt1, px1, py1, pz1, pt1 = self.im1.dim
         nx2, ny2, nz2, nt2, px2, py2, pz2, pt2 = self.im2.dim
 
-        assert np.isclose(px1, px2) and np.isclose(py1, py2) and np.isclose(px1, py1)
+        if not (np.isclose(px1, px2) and np.isclose(py1, py2) and np.isclose(px1, py1)):
+            raise ValueError("Resolutions are not close enough")
         self.dim_pix = py1
 
         if self.param.thinning:
@@ -308,7 +309,8 @@ class ComputeDistances:
         nx1, ny1, nz1, nt1, px1, py1, pz1, pt1 = self.im1.dim
         nx2, ny2, nz2, nt2, px2, py2, pz2, pt2 = self.im2.dim
         # assert np.round(pz1, 5) == np.round(pz2, 5) and np.round(py1, 5) == np.round(py2, 5)
-        assert nx1 == nx2
+        if nx1 != nx2:
+            raise ValueError(f"Mismatched image dimensions: {nx1} != {nx2}")
         self.dim_pix = py1
 
         if self.param.thinning:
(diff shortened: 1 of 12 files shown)


mguaypaq:
I was surprised to see the linting check fail, because locally it passed on my computer. I thought it might be because of different package versions, but from what I can tell, they're the same on my machine and in the CI run:
```
Successfully installed flake8-6.1.0 mccabe-0.7.0 pycodestyle-2.11.1 pyflakes-3.1.0
```
But! On my machine I use SCT's Python 3.9.18, and the CI run uses 3.12.0, and that makes all the difference.

I think the right fix is to change the linting check to use the same Python version as SCT? That's the version we can assume devs are working with.

spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
```diff
@@ -274,7 +274,8 @@ def compute_dist_2im_2d(self):
         nx1, ny1, nz1, nt1, px1, py1, pz1, pt1 = self.im1.dim
         nx2, ny2, nz2, nt2, px2, py2, pz2, pt2 = self.im2.dim
 
-        assert np.isclose(px1, px2) and np.isclose(py1, py2) and np.isclose(px1, py1)
+        if not (np.isclose(px1, px2) and np.isclose(py1, py2) and np.isclose(px1, py1)):
+            raise ValueError("Resolutions are not close enough")
```
joshuacwnewton, line 278: This is a bit of a weird one. It's checking the resolutions of two 2D images by comparing the X resolutions, the Y resolutions, but also the X1 and Y1 resolutions?

In other words, the images need to be A) isotropic and B) have similar resolutions? Given that there are two different conditions, maybe we should have two different error messages to disambiguate the error cases:

```suggestion
        if not all(np.isclose(px1, py1) and np.isclose(px2, py2)):
            raise ValueError(f"2D resolutions must both be isotropic, but got '{(px1, py1)}' and '{(px2, py2)}' instead.")
        if not np.all(np.isclose(px1, px2) and np.isclose(py1, py2)):
            raise ValueError(f"2D resolutions must be close to one another, but got '{(px1, py1)}' and '{(px2, py2)}' instead.")
```

spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -566,7 +566,8 @@ def multicomponent_split(im):
     :return:
     """
     data = im.data
-    assert len(data.shape) == 5
+    if len(data.shape) != 5:
+        raise ValueError(f"Wrong number of dimensions: {len(data.shape)}")
```
joshuacwnewton, line 570:
```suggestion
        raise ValueError(f"Expected 5D image but got '{len(data.shape)}' dimensions instead.")
```

joshuacwnewton (approved):
This mostly LGTM! I just had some minor clarity tweaks to the message wording (and one missed f-string).

spinalcordtoolbox/scripts/sct_compute_hausdorff_distance.py
joshuacwnewton: Technically there is a world where A == B == C == D but A != D, but that doesn't feel intentional on the part of the old check, given the symmetry in the logic.

mguaypaq: That's a much better set of checks, thank you for digging into it! I had noticed that there was a non-symmetric cross-comparison and thought it was curious, but decided to just go for the safe mechanical change. Your explanation makes a lot of sense!

mguaypaq:
@joshuacwnewton Thanks for the thoughtful review and the suggestions!

PR #4296, opened 2023-11-10
Use the same Python version as SCT for linting

mguaypaq:
This is the easiest way to make sure that developers' commit hooks generate the same warnings/errors as what runs in CI. Otherwise, we get annoyances like in #4294.

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.x'
+        python-version: '3.9'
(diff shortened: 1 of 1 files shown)


joshuacwnewton (commented):
Do you think now might be a good time to re-run linting on the project as a whole, then seeing what might have slipped through since the last time we did a full-project lint?

(I figure it might be good to group those changes with this PR, assuming there even are any.)

mguaypaq:
I just ran `flake8 spinalcordtoolbox/ testing/` locally, using SCT's Python (3.9.18) and flake8 (6.1.0), and I got no warnings at all :tada:

joshuacwnewton (approved):

PR #4300, opened 2023-11-15
Ensure reorientation is performed consistently before any resampling

### LINKED ISSUE #4299: Registration fails due to wrong cropping when seg/anat images have different orientations
A user on the forum reported a strange situation: 

- When registering using automatic segmentation, everything works fine.
- When registering using a manually-corrected segmentation, an `IndexError` occurs.

When comparing the segmentations, I noticed the following discrepancy:

- The automatic segmentation matches the orientation of the anat image/labels (ASR).
- The manually-corrected segmentation, on the other hand, was in a different orientation (RPI).

Reorienting the manually-corrected segmentation to ASR fixed the `IndexError`. But, in theory, the initial orientation should not matter, because all 3 input images are oriented to RPI before any processing occurs:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/69a4d2f1e62c44016f34ada7c1593447b3c1e971/spinalcordtoolbox/scripts/sct_register_to_template.py#L473-L486

So the question now is: Why does the initial input orientation still have an effect on the final outcome?

joshuacwnewton:
## Description

This PR fixes a bug where, when the `resample_labels()` function is called with `ftmp_labels` and `ftmp_data` not having the same orientation, the resampling is performed incorrectly. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/88798c51ca10952d25f289d8c8448634dfacbcd1/spinalcordtoolbox/scripts/sct_register_to_template.py#L483-L486

The funny thing is, immediately after this resampling, the images are reoriented to `RPI` anyway, so if we swap the order of these two steps (8bb6abe156db17b36bb9fd66cab89c9ebac55d25), we remove the buggy behavior.

Also, since we're now performing identical reorientation in both the `ref=template` and `ref=subject` cases, we can deduplicate the calls. (And, while we're at it, we might as well fix the if/elif pattern as well.)

## Linked issues

Fixes #4299.


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -456,6 +456,21 @@ def main(argv: Sequence[str]):
...
+    ftmp_data = add_suffix(img_tmp_data.absolutepath, "_rpi")
+    img_tmp_data.save(path=ftmp_data, mutable=True)
+
+    img_tmp_seg = Image(ftmp_seg).change_orientation("RPI")
+    ftmp_seg = add_suffix(img_tmp_seg.absolutepath, "_rpi")
+    img_tmp_seg.save(path=ftmp_seg, mutable=True)
+
+    img_tmp_label = Image(ftmp_label).change_orientation("RPI")
+    ftmp_label = add_suffix(img_tmp_label.absolutepath, "_rpi")
+    img_tmp_label.save(ftmp_label, mutable=True)
```
joshuacwnewton, line 472: It looks like these changes have caused deviations in the batch_processing.sh result:

```python
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[mt/mtr_in_wm.csv-0-MAP()] - assert 54.3539623020136 == 54.39462068731342 ± 5.4e-05
  comparison failed
  Obtained: 54.3539623020136
  Expected: 54.39462068731342 ± 5.4e-05
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[dmri/fa_in_cst.csv-0-WA()] - assert 0.7746611002178296 == 0.7793857115174943 ± 7.8e-07
  comparison failed
  Obtained: 0.7746611002178296
  Expected: 0.7793857115174943 ± 7.8e-07
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[dmri/fa_in_cst.csv-1-WA()] - assert 0.7727014657926683 == 0.7756883298906552 ± 7.8e-07
  comparison failed
  Obtained: 0.7727014657926683
  Expected: 0.7756883298906552 ± 7.8e-07
```

This is very curious to me! I didn't realize that there would be a meaningful difference between (Reorientation -> Resampling) vs. (Resampling -> Reorientation).

I could instead scope the changes to be even more narrow (e.g. adding a reorientation step to _just_ the `resample_labels` function) which may minimize the numerical impact of these changes.

But, from a code organization perspective, it does feel nice to deduplicate the reorientation...

mguaypaq: I'm not totally surprised to see that it would make a noticeable change. I guess the resampling grid is always aligned with one corner of the image box, and each dimension may be a fractional multiple of the grid step. In that case, reorienting the image means picking a different corner of the image box to align the resampling grid, and that can be a substantial change.

But because of this, I think it would be a good idea to _always_ reorient before resampling. That would mean that we have a consistent choice of resampling grid alignment, no matter what the input orientation is.

Since SCT works hard to make orientation usually not matter, I think users would be surprised to find out that this seemingly invisible property of the image file actually affects the result of an operation like resampling.

mguaypaq (approved):
We can discuss it at an SCT meeting, but I would suggest keeping your code fix and updating the batch processing results.

spinalcordtoolbox/scripts/sct_register_to_template.py
joshuacwnewton: As discussed in the most recent SCT meeting, we will keep the code changes here and update the batch processing results accordingly.

Issue #4298, opened 2023-11-12
Add function to compute flow


jcohenadad:
I'm working on a project where flow is to be computed from [VENC](https://mriquestions.com/what-is-venc.html) sequence. It would be nice to have a function to do that in SCT.

joshuacwnewton:
Just to keep the initial PR focused on velocity, I've removed the unused flow functions written by @jcohenadad and put them here:

```python
# function 1
def calculate_flow(velocity_data, pixel_area):
    """
    Calculate flow by integrating velocity over a given area.
    :param velocity_data: numpy array of velocity data
    :param pixel_area: area of each pixel in the image (in square meters)
    :return: flow rate in the vessel (in cubic meters per second)
    """
    flow_rate = np.sum(velocity_data) * pixel_area
    return flow_rate

# function 2
def compute_flow(nii_phase4d, venc):
    """
    Compute velocity from VENC technique. More details at: https://mriquestions.com/what-is-venc.html
    :param nii_phase4d: 4D phase image. The 4th dimension should be the velocity encoding (VENC) in cm/s.
    :param venc: float: Maximum velocity that can be encoded
    :return: nii_flow
    """
    # Convert input to avoid numerical errors from int16 data
    # Related issue: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3636
    nii_phase4d.change_type('float32')

    # Initialize Image object
    nii_flow = nii_phase4d.copy()

    # Compute flow
    nii_flow.data = np.true_divide(nii_phase4d.data, 2 * np.pi * venc)

    return nii_flow
```

If/when we want to re-add flow calculation, we can refer to these functions. But, from what I understand, these are lower priority than computing the velocity.

joshuacwnewton:
I've moved this off the 6.4 milestone because the main functionality (velocity calc) has been merged. And, the further features seem to be less urgent.

PR #4302, opened 2023-11-17
New function to compute velocity from VENC sequence

jcohenadad:
#### GitHub

#### PR contents

## Description

I'm working on a project where flow is to be computed from [VENC](https://mriquestions.com/what-is-venc.html) sequence. This PR implements that functionality.

- [x] Check input phase ranges between -pi and pi

Future features to consider/implement are listed in #4298.

## Linked issues

N/A (just implements velocity calculation)


jcohenadad:
Results from `3f301c20aa7337b4377aad09b1fe890516c0980a` using `CRL_1501A_3T_20231102_123128062_csf_flow_fl2d_tra_6through-plane-C2-3_20231102093852_36.nii` (private data)

<details><summary>Voxel in the spinal cord (expect minimum motion: [-0.2, 0.4] cm/s)</summary>

Screenshot 2023-12-15 at 4 18 10 PM

</details>

<details><summary>Voxel in the spinal cord (expect larger motion: [-0.5, 1.7 cm/s)</summary>

Screenshot 2023-12-15 at 4 18 46 PM

</details>

joshuacwnewton (commented):
I'd be happy to review this PR whenever it's ready! 

Just out of curiosity, though, do we have any sample VENC sequences that aren't private? (For the purposes of writing a test, etc.)

jcohenadad:
> Just out of curiosity, though, do we have any sample VENC sequences that aren't private? (For the purposes of writing a test, etc.)

I think we do have that, yes. I remember having acquired some data long time ago. Here you go: [36-csf_flow_fl2d_tra_20through-plane_P.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/13941238/36-csf_flow_fl2d_tra_20through-plane_P.zip)

joshuacwnewton:
> I think we do have that, yes. I remember having acquired some data long time ago. Here you go: [36-csf_flow_fl2d_tra_20through-plane_P.zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/13941238/36-csf_flow_fl2d_tra_20through-plane_P.zip)

Just taking a look at this now! My apologies in advance for the novice questions here:

Inside the .zip, there are 40 single-volume DICOM images. I converted them to a 4D `.nii` file using `dcm2niix`:

<details>

> ```console
> (venv_sct) jca/4298-flow|joshua@monarch:~/repos/spinalcordtoolbox/data$ dcm2niix 4302_sct_compute_flow/
> Chris Rorden's dcm2niiX version v1.0.20211006  (JP2:OpenJPEG) GCC11.2.0 x86-64 (64-bit Linux)
> Found 40 DICOM file(s)
> Warning: All images appear to be a single slice - please check slice/vector orientation
> Convert 40 DICOM as 4302_sct_compute_flow/4302_sct_compute_flow_csf_flow_fl2d_tra_20through-plane_20170525144102_36_ph (256x256x1x40)
> Warning: Check that 2D images are not mirrored.
> Conversion required 0.061963 seconds (0.060382 for core code).
> ```
> 
> I'm not sure if either of these warnings are something I should be concerned about? :thinking: 

</details>

After this, but prior to running `sct_compute_flow`, I wanted to inspect the 4D image (to better understand VENC). I noticed two things:

<details>
<summary> Image </summary>

> voxel-comparison

</details>

1. The min/max values appear to be approx `+/- 5000`
   - Compared to your test sequence (max 10cm/s) and the mriquestions page (max 50cm/s), this seems to be much too large to have units of cm/s. (Also, like... common sense! 5000cm/s seems very very fast, hehe.)
   - But, I'm not sure other SI unit prefixes make sense, either? 
      - 5000mm/s -> 500cm/s (too high?)
      - 5000um/s -> 0.5cm/s (too low? or maybe this is reasonable?)
   - So, I'm not sure what the units are for this sequence... :thinking: 
2. There are sharp intensity changes, where two adjacent voxels will have values of, say, `-4648 -> 4875`
   - I presume this is due to velocity aliasing, as described on the mriquestions page? 
   - And, if I understand correctly, since this aliasing is present even before running `sct_compute_flow`, it's more of an issue with the VENC parameter that was set during acquisition?

I have three main questions:

- Do you think this is an appropriate sequence to use in a test? 
- If so, should I scale the image before storing in in `sct_testing_data`?
- What should I specify for the `-venc` argument? (If I understand correctly, this is the parameter that is set during acquisition, so presumably it would be in the image metadata?)
   - `.json` sidecar generated by `dcm2niix`: [4302_sct_compute_flow_csf_flow_fl2d_tra_20through-plane_20170525144102_36_ph.json](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/13994869/4302_sct_compute_flow_csf_flow_fl2d_tra_20through-plane_20170525144102_36_ph.json)

Thanks for your patience! :hearts:

jcohenadad:
thank you for digging, yes, the images need to be scaled, this is partly dealt with by this function:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b85c6446e075d8b1e23ed6c2decba6799fc44a29/spinalcordtoolbox/qmri/flow.py#L70-L77

then, there are some assumptions about the phase being between -pi and pi. But i think the images i provided are fine (i ran then through my function and the output makes sense)

joshuacwnewton:
> But i think the images i provided are fine (i ran then through my function and the output makes sense)

What did you specify for the `-venc` argument when you ran the command?

jcohenadad:
> What did you specify for the -venc argument when you ran the command?

this is how we can quantify (in physics unit) the velocity, based on the gradient amplitude that were played out by the MRI. The relevant line is here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b85c6446e075d8b1e23ed6c2decba6799fc44a29/spinalcordtoolbox/qmri/flow.py#L28-L29

Ideally we would not have to enter that, and the info would be present in the metadata, but it is not always the case https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4335

joshuacwnewton:
Thank you for the background details!

I may have miscommunicated though. Since `-venc` is a mandatory parameter (`required=True`), I was wondering what specific value you chose for `-venc` when running `sct_compute_flow` on the [36-csf_flow_fl2d_tra_20through-plane_P.](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/13941238/36-csf_flow_fl2d_tra_20through-plane_P.zip) image from https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4302#issuecomment-1892499177.

jcohenadad:
ah! sorry, that would be venc=20 (that's the "20" in "tra_20through-plane_P")

spinalcordtoolbox/qmri/flow.py
```diff
@@ -0,0 +1,77 @@
...
+    :return: scaled phase data
+    """
+    # Check that phase data is encoded as int13
+    if np.any(data_phase > 4095) or np.any(data_phase < -4096):
+        logger.warning('Phase data is not encoded as INT13 (i.e., between -4096 and 4095). Please check your data.')
+    # Convert data to float
+    data_phase = data_phase.astype(np.float32)
+    # Scale the data from [-4096, 4095] to [-π, π]
+    scaled_data = (data_phase / 4096) * np.pi
+    return scaled_data
```
joshuacwnewton, line 77: To provide a more conclusive answer about the range of data, I found https://github.com/rordenlab/dcm2niix/issues/406:

> This sample follows the convention I have seen: the raw data is in the range 0..4095 (12-bit), with the [scl_slope = 2.0 and scl_inter = -4096](https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/scl_slopeinter.html) yielding values of -4096..4094. A common trick is to assume the full range is -pi..pi.

In other words, I think the current code might be slightly flawed:

- It mentions the range `[-4096, 4095]`, rather than the actual range of `[-4096, 4094]`
- It performs the scaling by dividing by `4096`, when I think we might need to perform a slightly more involved scaling operation (due to the mention of slope/inter == 2,-4096):

```suggestion
def scale_phase(data_phase):
    """
    Scale phase data between -pi and pi. We assume the data was originally encoded as int12 [0, 4095] (as this seems
    to be the default encoding for phase data from Siemens scanners), then scaled to [-4096, 4094] during conversion
    from DICOM to Nifti due to a slope/intercept of 2.0 and -4096.

    See also: https://github.com/rordenlab/dcm2niix/issues/406#issuecomment-654141193

    :param data_phase: numpy array of phase data
    :return: scaled phase data
    """
    # Check that phase data is encoded as scaled 12-bit integers
    slope = 2.0
    intercept = -4096
    bounds_unscaled = [0, 4095]
    bounds_scaled = [b * slope + intercept for b in bounds_unscaled]
    if np.any(data_phase < bounds_scaled[0]) or np.any(data_phase < bounds_scaled[1]):
        logger.warning(f'Phase data is not within the expected range for Siemens phase data (i.e., between '
                       f'{bounds_scaled[0]} and {bounds_scaled[1]}). Please check your data.')
    # Convert data to float
    data_phase = data_phase.astype(np.float32)
    # Scale the data from [-4096, 4094] back to [0, 4095]
    scaled_data = (data_phase - intercept) / slope
    # Scale the data from [0, 4095] to [0, 2]
    scaled_data = scaled_data / bounds_unscaled[1] * 2
    # Scale the data from [0, 2] to [-π, π]
    scaled_data = (scaled_data - 1) * np.pi
    return scaled_data
```

joshuacwnewton: Before: 

image

After:

image

spinalcordtoolbox/scripts/sct_compute_flow.py
```diff
@@ -0,0 +1,101 @@
...
+    printv(f"Load data: {arguments.i}", verbose)
+    nii_phase = Image(arguments.i)
+
+    # Scale phase data between -pi and pi
+    printv('Scale phase data between -pi and pi', verbose)
+    data_phase_scaled = flow.scale_phase(nii_phase.data)
+
+    # Calculate velocity
+    printv('Calculate velocity', verbose)
+    velocity = flow.calculate_velocity(data_phase_scaled, venc)
```
joshuacwnewton, line 81: I'm a bit confused here. The current steps are:

```python
scale_phase()         # [-4096, 4094] -> [-π, π]
calculate_velocity()  # [-π, π] -> [-venc, venc]
```

In other words, `compute_flow()` is not actually called right now, and the only operations that happen are scaling operations.

Is this intentional? (The reason I ask is because I see two versions of the `compute_flow()` function, one of which is commented out, so I'm wondering if this is still a WIP PR.)

joshuacwnewton (changes requested):
This is a relatively small CLI function, so I don't have too many comments here. :)

spinalcordtoolbox/download.py
```diff
@@ -40,7 +40,9 @@
     },
     "sct_testing_data": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/sct_testing_data/releases/download/r20230207/sct_testing_data-r20230207.zip",
+            # TODO: Replace with stable release generated by distriploy once PR is merged:
+            #       https://github.com/spinalcordtoolbox/sct_testing_data/pull/12
+            "https://github.com/spinalcordtoolbox/sct_testing_data/archive/jn/update-for-sct-4302-4345.zip",
```
joshuacwnewton, line 45: Side note: Please disregard the test failures! This is because the `sct_testing_data` changes also include changes meant for #4345.

joshuacwnewton: Resolved by merging master and updating the sct_testing_data link!

spinalcordtoolbox/scripts/sct_compute_flow.py
joshuacwnewton: JCA: The `compute_flow` functions are not called right now because people are not actually interested in the flow, they want the velocity.

My question is: If all this CLI script does is scale the data, is that necessary?

joshuacwnewton: JCA is currently using this function as-is, so it's OK to merge as-is without adding these new features.

spinalcordtoolbox/qmri/flow.py
joshuacwnewton: We looked over this revision and JCA gave the :+1:

joshuacwnewton (approved):
I've implemented the changes as discussed in the May 3rd SCT meeting, so I think this is good to merge. :)

jcohenadad:
oops tests failing..

jcohenadad:
i let you merge @joshuacwnewton

joshuacwnewton:
> oops tests failing..

The error is caused by this issue:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4471

I added a workaround in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4302/commits/bfe980b0be0ca0da1874a4f3e408ba7d5db8bec7, but I will fix the root cause in a separate PR.

PR #4306, opened 2023-11-22
Apply missing B1 correction to `r1map` and `a` calculations to fix MTsat bug

mathieuboudreau:
#### GitHub

#### PR contents

## Description

The MTsat equations in SCT were inherited from qMRLab. I've just discovered that the initial implementation of the B1 correction of MTsat in qMRLab was missing corrections for the PDw and T1w excitation flip angles, which for large B1 errors causes substantial differences in the T1 and MTsat maps. Other softwares have these B1 corrections ([QUIT](https://github.com/spinicist/QUIT/blob/5af31041a2194c81306826a729e6729f4d806f3d/Source/MT/MTSatModel.cpp#L12-L20), hMRI [R1](https://github.com/hMRI-group/hMRI-toolbox/blob/894a38be696e42a769765c313022f0bf1acafe64/hmri_calc_R1.m#L41-L47) and [A](https://github.com/hMRI-group/hMRI-toolbox/blob/894a38be696e42a769765c313022f0bf1acafe64/hmri_calc_A.m#L41-L47)). 

This PR seeks to fix this bug, and is designed closest to the QUIT implementation (with the B1 multiplications factored out of the main calculations, instead of multiplying each FA by B1).

Here's a before/after fix using SCT for a ultra high field mouse brain dataset that has sever B1 inhomogeneity.

# MTsat map
## Before
Screenshot 2023-11-22 at 1 29 13 PM

##After
Screenshot 2023-11-22 at 1 30 57 PM

# T1 map
## Before
Screenshot 2023-11-22 at 1 29 38 PM

## After
Screenshot 2023-11-22 at 1 31 20 PM

## Linked issues

https://github.com/qMRLab/qMRLab/pull/497

### DIFF
--- a/spinalcordtoolbox/qmri/mt.py
+++ b/spinalcordtoolbox/qmri/mt.py
@@ -8,6 +8,7 @@ License: see the file LICENSE
 """
 
 import logging
+
 import numpy as np
 
 logger = logging.getLogger(__name__)
@@ -102,7 +103,7 @@ def compute_mtsat(nii_mt, nii_pd, nii_t1,
         # compute R1
         logger.info("Compute T1 map...")
         r1map = 0.5 * np.multiply(np.multiply(nii_b1map.data, nii_b1map.data), np.true_divide((fa_t1_rad / tr_t1) * nii_t1.data - (fa_pd_rad / tr_pd) * nii_pd.data,
-                                     nii_pd.data / fa_pd_rad - nii_t1.data / fa_t1_rad))
+                                                                                              nii_pd.data / fa_pd_rad - nii_t1.data / fa_t1_rad))
         # remove nans and clip unrelistic values
         r1map = np.nan_to_num(r1map)
         ind_unrealistic = np.where(r1map < r1_threshold)
@@ -119,10 +120,10 @@ def compute_mtsat(nii_mt, nii_pd, nii_t1,
 
     # Compute A
     logger.info("Compute A...")
-    a = np.divide((tr_pd * fa_t1_rad / fa_pd_rad - tr_t1 * fa_pd_rad / fa_t1_rad) * \
-        np.true_divide(np.multiply(nii_pd.data, nii_t1.data, dtype=float),
-                       tr_pd * fa_t1_rad * nii_t1.data - tr_t1 * fa_pd_rad * nii_pd.data), \
-                       nii_b1map.data)
+    a = np.divide((tr_pd * fa_t1_rad / fa_pd_rad - tr_t1 * fa_pd_rad / fa_t1_rad) *
+                  np.true_divide(np.multiply(nii_pd.data, nii_t1.data, dtype=float),
+                                 tr_pd * fa_t1_rad * nii_t1.data - tr_t1 * fa_pd_rad * nii_pd.data),
+                  nii_b1map.data)
 
     # Compute MTsat
     logger.info("Compute MTsat...")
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/qmri/mt.py
```diff
@@ -101,8 +101,8 @@ def compute_mtsat(nii_mt, nii_pd, nii_t1,
     if nii_t1map is None:
         # compute R1
         logger.info("Compute T1 map...")
-        r1map = 0.5 * np.true_divide((fa_t1_rad / tr_t1) * nii_t1.data - (fa_pd_rad / tr_pd) * nii_pd.data,
-                                     nii_pd.data / fa_pd_rad - nii_t1.data / fa_t1_rad)
+        r1map = 0.5 * np.multiply(np.multiply(nii_b1map.data, nii_b1map.data), np.true_divide((fa_t1_rad / tr_t1) * nii_t1.data - (fa_pd_rad / tr_pd) * nii_pd.data,
+                                     nii_pd.data / fa_pd_rad - nii_t1.data / fa_t1_rad))
```
joshuacwnewton, line 105: Thank you for the fix!! Admittedly, I don't know if I'm the best person to review the substance of the PR, but I'm happy to take care of the failing CI tests.

jcohenadad (approved):
Amazing! Thank you for the fix @mathieuboudreau . So from what I gather, this ad hoc value is still relevant (and its justification still valid):
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4306/files#diff-aeeefe09db7eb324c78a41da6300add125370ba2ebb9b65fe9a00ee30604b2ddR85-L86

spinalcordtoolbox/qmri/mt.py
joshuacwnewton: It seems that currently `nii_b1map` is an optional argument to our CLI script:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/27e659857333a77790ea861d979a26569ae4b7d8/spinalcordtoolbox/scripts/sct_compute_mtsat.py#L168-L171

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/27e659857333a77790ea861d979a26569ae4b7d8/spinalcordtoolbox/qmri/mt.py#L62-L65

Because it is optional, there are cases where the B1 map won't be provided; the current PR fails in such cases.

So, I've refactored the PR such that the old computation will be preserved, and the new B1 correction will be applied on a separate line (i.e. only if `nii_b1map` is present).

joshuacwnewton (approved):
Assuming that my small refactor produces the results you expect (could you double-check on your mouse data?), I think this is good to merge @mathieuboudreau. :)

spinalcordtoolbox/qmri/mt.py
mathieuboudreau: >>Assuming that my small refactor produces the results you expect (could you double-check on your mouse data?)

Done - results reproduce exactly! Thanks for the test fix!

mathieuboudreau:
>Amazing! Thank you for the fix @mathieuboudreau . So from what I gather, this ad hoc value is still relevant (and its justification still valid):
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4306/files#diff-aeeefe09db7eb324c78a41da6300add125370ba2ebb9b65fe9a00ee30604b2ddR85-L86

Yes - somewhat. From the discussion with @spinicist and insights from Gunther Helms in another thread, basically this value was measured for a specific MT pulse protocol at 3T, but will change a bit at different field strengths / different pulses. It's not easily feasible to ask users to do new measurements to calculate the new constant for whatever their protocol is, but B1 correction with a 0.4 constant will give better results than no B1 correction at all. I'm going to look into if we could find the ideal constant for another given protocol using simulations qMRLab instead of having to measure it, but for now here I think it's sufficient to just keep it at 0.4.

PR #4231, opened 2023-09-27
Test installing `ivadomed`, `torch>=2.0.0`, `nnunet`, and `monai` in the same `venv_sct`

joshuacwnewton:
## Description

This PR is a quick test to see if `ivadomed` will install correctly alongside SCT when `torch>=2.0.0` is specified. The corresponding ivadomed PR is: https://github.com/ivadomed/ivadomed/pull/1304

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4182

Issue #4182, opened 2023-08-04
Integrate MONAI/nnUNet model inference into `sct_deepseg`


mguaypaq:
This is a feature request that came out of the SCT dev meeting on 2023-07-27 ([minutes](https://docs.google.com/document/d/1ItApJQfajO2lRzOU2yenWbeRg6alfsdut3J4AVVdo78/edit#heading=h.3jazj3wbpm6z)).

We should figure out how to integrate with MONAI ([link](https://monai.io/core.html)) for deepseg models. Currently we integrate with [ivadomed](https://github.com/ivadomed/ivadomed), but the long-term intention is to migrate to MONAI instead (see meeting minutes for 2023-02-07 and subsequent meetings [here](https://docs.google.com/document/d/19mEHKOZKFdprb3UPQYxmfO46mn0bPQfOgE-WPdeF7AQ/edit#heading=h.mh06cc7ad3ul)).

Action items:
- Need an update of the `sct_deepseg` API so that it can be run under MONAI (in addition to the current ivadomed-based workflow).
- Need to coordinate with the contrast-agnostic project (@rohanbanerjee, @naga-karthik)
- Also need to discuss packaging nnUNet ([publication](https://www.nature.com/articles/s41592-020-01008-z#code-availability), [code repo](https://github.com/MIC-DKFZ/nnUNet)) – used as a part of the fMRI EPI SC segmentation project.

Questions:
- Do we import MONAI, or does MONAI import us? Or do we just need to have a compatible API?

naga-karthik:
Thanks for opening this issue! I can briefly comment about packaging aspects of MONAI and nnUNet

> Do we import MONAI, or does MONAI import us? 

We import MONAI. I don't know if need an API or something. I created an [inference script](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/blob/nk/monai/monai/run_inference.py) that uses a model trained using MONAI. Note, that "training with MONAI" only means that we're mainly using its augmentations and some other functions. The model at the end is indeed a PyTorch model (just like ivadomed's). This is also the reason why I said we might not need an API.

Some specifics: 
1. [These lines](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/blob/nk/monai/monai/run_inference.py#L10-L13) are where I import what I need from MONAI. 
[...]

PR #4314, opened 2023-12-05
Update `numpy` upper bound from `<1.24` to `<1.25`

joshuacwnewton:
## Description

Previously, we pinned `numpy<1.24` due to an incompatibility with pystrum:

> ```bash
> # This pin is due to an incompatibility with pystrum==0.2, which is installed via voxelmorph -> neurite -> pystrum
> # See also: https://github.com/adalca/pystrum/issues/9
> ```

Pystrum has since been updated, so the `numpy<1.24` upper bound can be removed.

However, this runs into _another_ incompatibility, this time with `dipy`. (We're forced to avoid the newest versions of dipy due to an upstream bug https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328. And, the older version of dipy is incompatible with `numpy>=1.25`, hence the new upper bound.)

This will be relevant for PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4231, since the MONAI contrast-agnostic code uses a newer version of numpy. So, if we want to test those new additions, it would be good to have numpy unpinned.

## Linked issues

Related to #4182.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,9 @@ matplotlib
 msvc-runtime==14.29.30133; sys.platform == 'win32' # append_to_freeze
 nibabel
 nilearn
-numpy
+# Currently, we pin dipy such that 1.5.0 gets installed. However, dipy 1.5.0 is incompatible with
+# numpy>=1.25.0. So, we have to pin numpy on their behalf.
+numpy<1.25
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,9 +2,8 @@
 #     stable releases. We do this because `pip freeze` will not capture options (e.g. --extra-index-url) or
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
-# v1.5 and below are incompatible with newer versions of numpy
-# We also avoid v1.6 and v1.7 due to method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
-dipy>1.7.0
+# Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -18,10 +18,7 @@ matplotlib
 msvc-runtime==14.29.30133; sys.platform == 'win32' # append_to_freeze
 nibabel
 nilearn
-# This pin is due to an incompatibility with pystrum==0.2, which is installed via voxelmorph -> neurite -> pystrum
-# See also: https://github.com/adalca/pystrum/issues/9
-# If pystrum is updated to 0.3, then we should replace this version pinning with `pystrum>=0.3`
-numpy<1.24
+numpy
```
joshuacwnewton, line 21: It looks like unpinning has introduced a new incompatibility with `dipy`:

```python
E   ImportError: cannot import name 'Tester' from 'numpy.testing' (/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/numpy/testing/__init__.py)
```

This is a bit strange, as `dipy` is effectively unpinned right now (besides versions 1.6 and 1.7, which contain a bug). Presumably, with both versions at their newest, there would be no compatibility issues?

Installed versions:

- numpy 1.26.2
- dipy 1.5.0

Aha! The newest version of `dipy` _didn't_ get installed. I'll try to work out why this might be.

joshuacwnewton: Ah, wait. There is no newer version than `dipy==1.7.0`!

In that case, I'll just need to find the latest version of numpy that is compatible with dipy==1.5.0.

(Then, when dipy==1.8.0 is released, and the method=restore bug is fixed, we can once again unpin numpy.)

joshuacwnewton: It looks like, for dipy==1.5.0, we're locked to numpy<1.25:

- https://github.com/dipy/dipy/issues/2729
- https://github.com/numpy/numpy/pull/23041

joshuacwnewton:
> This will be relevant for PR #4231, since the MONAI contrast-agnostic code uses a newer version of numpy. So, if we want to test those new additions, it would be good to have numpy unpinned.

Hrm. It seems like `numpy~=1.24` isn't actually necessary for #4231; the inference script runs just fine with our current restriction. 

In that case, maybe it's just fine to wait until `dipy==1.8.0` releases to make this change? (That way, we can remove the pin entirely, rather than making such a minor upgrade to it.)

PR #4326, opened 2023-12-29
`sct_compute_compression` - fix slice numbering in output CSV file

valosekj:
#### GitHub

#### PR contents

## Description
When playing with the `sct_compute_compression` function, I discovered a bug in slice numbering in the output CSV files. It seems that the `slice(I->S)` column is reversed relative to the `compression_level` column:

`master` branch:

image

this branch:

image

The fix in this branch corresponds also with nii files in FSLeyes:

image

## Steps to reproduce

1. download `data_normalizing-morphometrics-compression.zip` from [our tutorial](https://spinalcordtoolbox.com/user_section/tutorials/shape-analysis/normalizing-morphometrics-compressions/generate-necessary-inputs.html)
2. run `sct_compute_compression`:

```console
sct_compute_compression -i t2_compressed_seg.nii.gz -vertfile t2_compressed_seg_labeled.nii.gz  -l t2_compressed_labels-compression.nii.gz  -normalize-hc 1 -metric area
```

## TODO
 - [x] [Fix tutorial table](https://spinalcordtoolbox.com/user_section/tutorials/shape-analysis/normalizing-morphometrics-compressions/normalize-morphometrics.html)


spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -643,7 +645,7 @@ def main(argv: Sequence[str]):
             metric_ratio_PAM50_result = None
             metric_ratio_norm_result = None
 
-        rows.append([arguments.i, level, slice_compressed[idx],
+        rows.append([arguments.i, level, slice,
```
joshuacwnewton, line 648: Just so I understand the source of the error:

- `slice_compressed` is a list of slices where the compression exists:
  https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c50bc2380f2b953e8b60b96233b557b83fb6bbe/spinalcordtoolbox/scripts/sct_compute_compression.py#L576-L577
- `idx` comes from `compressed_levels_dict`:
  https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c50bc2380f2b953e8b60b96233b557b83fb6bbe/spinalcordtoolbox/scripts/sct_compute_compression.py#L621
- And `compressed_levels_dict` comes from `slice_compressed`:
  https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c50bc2380f2b953e8b60b96233b557b83fb6bbe/spinalcordtoolbox/scripts/sct_compute_compression.py#L578
  This is weird. If both `idx` and `slice_compressed` come from the same source, then why does `slice_compressed[idx]` produce the wrong slice numbers?

----

Answer: `slice_compressed` is sorted by `z` slice index (I->S):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c50bc2380f2b953e8b60b96233b557b83fb6bbe/spinalcordtoolbox/scripts/sct_compute_compression.py#L174

While `compressed_levels_dict` is sorted by vertebral level order (S->I):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3c50bc2380f2b953e8b60b96233b557b83fb6bbe/spinalcordtoolbox/scripts/sct_compute_compression.py#L212

joshuacwnewton: It's a bit weird that `compressed_levels_dict` is sorted S->I while everything else is sorted I->S, so maybe that's worth changing?

But, regardless, I like your solution, because you explicitly get the slice number from the same source as the level, which prevents order from having any effect at all.

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -619,8 +619,10 @@ def main(argv: Sequence[str]):
     # Loop through all compressed levels (compute one MSCC per compressed level)
     rows = []
     for idx in compressed_levels_dict.keys():
+        # Get compressed level and slice
         level = list(compressed_levels_dict[idx].keys())[0]  # TODO change if more than one level
-        printv(f'\nCompression at level {int(level)}', verbose=verbose, type='info')
+        slice = list(compressed_levels_dict[idx].values())[0][0]
```
joshuacwnewton, line 624: Nitpick: We already access `list(compressed_levels_dict[idx].values())[0]` later on in this code block (`slice_avg`), so we're sort of duplicating that a little?

joshuacwnewton: Side note: This whole call is a little weird, what with the `[0][0]` and the access of `.keys()` and `.values()`. I think the weirdness is due to the structure of `compressed_levels_dict`:

image

It's a dict of dicts of lists, which is a bit of a strange structure to use... I presume this is because A) The compression site could theoretically span multiple levels, and B) The compression site could theoretically span multiple slices within the same level?

But, right now we ignore all of that and simply take the first level + first slice...

I think this could maybe be refactored to be a bit simpler? But, tbh, it's outside the scope of this PR, so I don't think there's really too much to worry about.

joshuacwnewton: I've made some very nitpicky changes to simplify and deduplicate the access of the slice number.

joshuacwnewton (approved):

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton: > It's a bit weird that `compressed_levels_dict` is sorted S->I while everything else is sorted I->S, so maybe that's worth changing?

One reason to keep the order `I->S` is that the CSV table would be more accurate. Right now, the CSV is sorted S->I, yet the header says I->S, which might be a bit confusing?

valosekj: Thanks for digging! 

> Right now, the CSV is sorted S->I, yet the header says I->S, which might be a bit confusing?

Yeah, this is confusing! An easy fix would be to update the header from `slice(I->S)` to `slice(S->I)`. But, by this, we would introduce an inconsistency with a CSV file produced by `sct_process_segmentation`, which uses `Slice (I->S)`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/264760c139548190650ecce8271b19bd5425a46e/spinalcordtoolbox/aggregate_slicewise.py#L559

So maybe a solution might be to reverse the table as all:

filename | compression_level | slice(I->S) | diameter_AP_ratio |   |  
-- | -- | -- | -- | -- | --
t2_compressed_seg.nii.gz | 6.0 | 99 | 21.88085837700506 |   |  
t2_compressed_seg.nii.gz | 5.0 | 118 | 10.780650030491945 |   |  
t2_compressed_seg.nii.gz | 4.0 | 140 | 11.314949666366614 |   |  
t2_compressed_seg.nii.gz | 3.0 | 159 | 3.3493565531077407 |   |

valosekj: nitpick: we can rename `slice(I->S)` to `Slice (I->S)` to be consistent with https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/264760c139548190650ecce8271b19bd5425a46e/spinalcordtoolbox/aggregate_slicewise.py#L559

spinalcordtoolbox/scripts/sct_compute_compression.py
valosekj: > A) Each compression site (0, 1, 2...) could theoretically span multiple levels (3.0/4.0, 4.0/5.0, etc.), and B) The compression site could theoretically span multiple slices within the same level (e.g. [159, 160])?

Correct for both. Currently, we simplify this by labeling only the maximum compressed slice for each compression (step 4 in [our tutorial](https://spinalcordtoolbox.com/user_section/tutorials/shape-analysis/normalizing-morphometrics-compressions/generate-necessary-inputs.html#label-spinal-cord-compression-sites)). In other words, even though compression spans several slices, we only put the label on one.

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton: > So maybe a solution might be to reverse the table as all:

I think so too! I think we can accomplish this by not inverting the order (as done above)?

valosekj: - `slice(I->S)` renamed to `Slice (I->S)` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4326/commits/fbb1b70ee61b8e562c0bdb6bbcefeb9ac70d23bf
- reverse rows in the output CSV table to match `Slice (I->S)` ordering in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4326/commits/a1df9bd2c26b81ca391721c8fda03d1d68b2e8ee
image

sandrinebedard: hmm, I don't remember if it was to have compression levels in order, or if there was more!

sandrinebedard (approved):
Everything looks good, I agree with the latest changes!!

joshuacwnewton:
Thanks much for the fix, @valosekj! :hearts:

PR #4332, opened 2024-01-10
Pull upstream `ivadomed` changes that let us upgrade previously-pinned versions of `dipy`/`numpy`/`nibabel`

### LINKED ISSUE #4319: `dipy==1.8.0` causes `trx-python` dependency conflict with `ivadomed`, resulting in failing SCT installations
### Description
I tried installing SCT and it is failing at the step when it's installing nilearn-0.6.2. I noticed the tests are also failing in the CI for all platforms in the repo.

### Steps to Reproduce
The [tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/workflows/tests.yml?query=branch%3Amaster) seem to have been failing for the pas 4 days. See [this test](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/7260966482/job/19781204087#step:4:538) for an example.

joshuacwnewton:
## Description

Upstream ivadomed changes: https://github.com/ivadomed/ivadomed/pull/1308

The upstream changes enabled a cascade of version upgrades (compared to `master`):

- Explicit dependency upgrades (`requirements.txt`):
  - `nibabel`: `3.2.2` -> `5.2.1`
  - `dipy`: `1.5.0` -> `1.8.0`
  - `numpy`: `1.23.5` -> `1.26.4`
- Second-order dependency upgrades (held back by previous packages):
  - `pandas`: `1.4.4` -> `1.5.3`
  - `pyparsing`: `2.4.7` -> `3.1.2`

Given that we've been held back on these versions for so long, some additional changes are needed to address upstream deprecation warnings and errors, mainly due to `numpy`.

## Linked issues

Fixes #4319.
Mostly addresses https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4408, but doesn't address https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4408#issuecomment-2007396098.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
 # Avoid 1.6 and 1.7 due to method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
-dipy!=1.6.*,!=1.7.*
+dipy!=1.6.*,!=1.7.*,<1.9.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ dipy!=1.6.*,!=1.7.*,<1.9.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
-ivadomed @ git+https://github.com/ivadomed/ivadomed@master # append_to_freeze
+ivadomed
 matplotlib
 monai[nibabel]
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,8 +3,7 @@
 #     platform-specific requirements (e.g. sys.platform == 'win32')
 
 # Avoid 1.6 and 1.7 due to method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
-# dipy 1.8 and above cause conflicts, see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319#issuecomment-1864670675
-dipy<1.6
+dipy!=1.6.*,!=1.7.*
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ dipy<1.6
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
-ivadomed
+ivadomed @ git+https://github.com/ivadomed/ivadomed@master # append_to_freeze
 matplotlib
 monai[nibabel]
 # Fresh Windows installations may be missing the C++ runtime library required by scikit-image.
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,10 +20,7 @@ msvc-runtime==14.29.30133; sys.platform == 'win32' # append_to_freeze
 nibabel
 nilearn
 nnunetv2
-# This pin is due to an incompatibility with pystrum==0.2, which is installed via voxelmorph -> neurite -> pystrum
-# See also: https://github.com/adalca/pystrum/issues/9
-# If pystrum is updated to 0.3, then we should replace this version pinning with `pystrum>=0.3`
-numpy<1.24
+numpy
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
@@ -38,6 +35,9 @@ psutil
 pyqt5==5.12.3
 # PyQt5-sip should be updated when PyQt5 is updated: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4354
 pyqt5-sip<12.13.0
+# pystrum==0.2 is incompatible with newer versions of numpy (installed via voxelmorph -> neurite -> pystrum)
+# See also: https://github.com/adalca/pystrum/issues/9
+pystrum>=0.3
 pytest
 pytest-cov
 requests
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,8 +44,8 @@ filterwarnings =
     # Source: test_sct_register_multimodal_mto_image_data --> `algorithms.py :: vxm.networks.VxmDense()`
     ignore:.*in an upcoming release, it will be required to pass the indexing argument.*:UserWarning:torch.*:
     # This warning is due to a deprecation for a call that is still used by many upstream packages. In our case:
-    # - wandb: https://github.com/wandb/wandb/issues/6711
-    ignore:pkg_resources is deprecated as an API:DeprecationWarning
+    # - monai: https://github.com/Project-MONAI/MONAI/issues/7559
+    ignore:pkg_resources is deprecated as an API:DeprecationWarning:monai.*:
     # These warnings are for non-GUI backends on headless machines (i.e. CI), so it's safe to ignore.
     ignore:Matplotlib is currently using agg.*:UserWarning:
     ignore:.*is non-interactive, and thus cannot be shown:UserWarning:
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,6 +11,7 @@ filterwarnings =
     #    seriously, but are time-consuming to debug. These should definitely be explored in future issues.
     #    1a. This is the most frequent error. Happens in at least 5 separate LOCs.
     ignore:invalid value encountered in divide.*:RuntimeWarning:
+    ignore:invalid value encountered in scalar divide.*:RuntimeWarning:
     #    1b. This is the second-most frequent error. Happens in at least 2 separate LOCs.
     ignore:invalid value encountered in double_scalars.*:RuntimeWarning:
     #    1c. This happens once: `test_cli_register_to_template` --> `compute_pca()` --> `coordsrc.mean(0)`
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,6 +43,9 @@ filterwarnings =
     #     I'm not sure if we can do much here, but it will only become an issue if and when we upgrade torch.
     # Source: test_sct_register_multimodal_mto_image_data --> `algorithms.py :: vxm.networks.VxmDense()`
     ignore:.*in an upcoming release, it will be required to pass the indexing argument.*:UserWarning:torch.*:
+    # This warning is due to a deprecation for a call that is still used by many upstream packages. In our case:
+    # - wandb: https://github.com/wandb/wandb/issues/6711
+    ignore:pkg_resources is deprecated as an API:DeprecationWarning
     # These warnings are for non-GUI backends on headless machines (i.e. CI), so it's safe to ignore.
     ignore:Matplotlib is currently using agg.*:UserWarning:
     ignore:.*is non-interactive, and thus cannot be shown:UserWarning:
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,6 +51,10 @@ filterwarnings =
     # This is a warning we ourselves generate, and needs discussion
     # Revisit as part of: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4357
     ignore:.*with additional arguments is discouraged:UserWarning:
+    # NB: `pandas` is being held to <2.0.0 by ivadomed, but numpy is free to upgrade, resulting in several upstream
+    # deprecation warnings by mixing an out-of-date pandas with an up-to-date numpy.
+    ignore:np.find_common_type is deprecated:DeprecationWarning:pandas.*:
+    ignore:NumPy will stop allowing conversion:DeprecationWarning:pandas.*:
 
 [flake8]
 max-line-length = 179
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,9 +43,6 @@ filterwarnings =
     #     I'm not sure if we can do much here, but it will only become an issue if and when we upgrade torch.
     # Source: test_sct_register_multimodal_mto_image_data --> `algorithms.py :: vxm.networks.VxmDense()`
     ignore:.*in an upcoming release, it will be required to pass the indexing argument.*:UserWarning:torch.*:
-    # This warning is due to a deprecation for a call that is still used by many upstream packages. In our case:
-    # - wandb: https://github.com/wandb/wandb/issues/6711
-    ignore:pkg_resources is deprecated as an API:DeprecationWarning
     # These warnings are for non-GUI backends on headless machines (i.e. CI), so it's safe to ignore.
     ignore:Matplotlib is currently using agg.*:UserWarning:
     ignore:.*is non-interactive, and thus cannot be shown:UserWarning:
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,9 @@ filterwarnings =
     # deprecation warnings by mixing an out-of-date pandas with an up-to-date numpy.
     ignore:np.find_common_type is deprecated:DeprecationWarning:pandas.*:
     ignore:NumPy will stop allowing conversion:DeprecationWarning:pandas.*:
+    # MONAI/nnUNet warnings: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4404
+    ignore:Detected old nnU-Net plans format:UserWarning:nnunetv2.*:
+    ignore:.*Our suggested max number of worker.*:UserWarning:torch.*:
 
 [flake8]
 max-line-length = 179
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -196,28 +196,25 @@ class AnalyzeLesion:
             copy(os.path.join(self.tmp_dir, file_), os.path.join(self.path_ofolder, file_))
 
     def pack_measures(self):
-        writer = pd.ExcelWriter(self.excel_name, engine='xlsxwriter')
-        self.measure_pd.to_excel(writer, sheet_name='measures', index=False, engine='xlsxwriter')
+        with pd.ExcelWriter(self.excel_name, engine='xlsxwriter') as writer:
+            self.measure_pd.to_excel(writer, sheet_name='measures', index=False, engine='xlsxwriter')
 
-        # Add the total column and row
-        if self.path_template is not None:
-            for sheet_name in self.distrib_matrix_dct:
-                if '#' in sheet_name:
-                    df = self.distrib_matrix_dct[sheet_name].copy()
-                    df = df.append(df.sum(numeric_only=True, axis=0), ignore_index=True)
-                    df['total'] = df.sum(numeric_only=True, axis=1)
-                    df.iloc[-1, df.columns.get_loc('vert')] = 'total'
-                    df.to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
-                else:
-                    self.distrib_matrix_dct[sheet_name].to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
+            # Add the total column and row
+            if self.path_template is not None:
+                for sheet_name in self.distrib_matrix_dct:
+                    if '#' in sheet_name:
+                        df = self.distrib_matrix_dct[sheet_name].copy()
+                        df = df.append(df.sum(numeric_only=True, axis=0), ignore_index=True)
+                        df['total'] = df.sum(numeric_only=True, axis=1)
+                        df.iloc[-1, df.columns.get_loc('vert')] = 'total'
+                        df.to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
+                    else:
+                        self.distrib_matrix_dct[sheet_name].to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
 
-        # Save pickle
-        self.distrib_matrix_dct['measures'] = self.measure_pd
-        with open(self.pickle_name, 'wb') as handle:
-            pickle.dump(self.distrib_matrix_dct, handle)
-
-        # Save Excel
-        writer.close()
+            # Save pickle
+            self.distrib_matrix_dct['measures'] = self.measure_pd
+            with open(self.pickle_name, 'wb') as handle:
+                pickle.dump(self.distrib_matrix_dct, handle)
 
     def show_total_results(self):
         """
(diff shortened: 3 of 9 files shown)


requirements.txt
```diff
@@ -4,11 +4,11 @@
 
 # Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
 # dipy 1.8.x causes conflicts, see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319#issuecomment-1864670675
-dipy!=1.6.*,!=1.7.*, !=1.8.*
+dipy!=1.6.*,!=1.7.*
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
-ivadomed~=2.9.6
+ivadomed @ git+https://github.com/ivadomed/ivadomed@jn/update-pinned-dependencies # append_to_freeze
```
joshuacwnewton, line 11: Because we've been limited to `nibabel==3.x` for so long, the jump to `5.x` introduced some test failures due to API changes.

I'll try and fix those up so that SCT is compatible with modern `nibabel`.

requirements.txt
```diff
@@ -4,11 +4,11 @@
 
 # Avoid method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
 # dipy 1.8.x causes conflicts, see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319#issuecomment-1864670675
-dipy!=1.6.*,!=1.7.*, !=1.8.*
```
joshuacwnewton, line 7: Note that when installing `dipy==1.8.0`, we can revisit this older PR of mine as well: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4314#issuecomment-1841570140

joshuacwnewton: Hrm... I've done what I think is correct in both https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332/commits/c13446b5a54f077c001d279446bcbf4138bf22e6 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332/commits/4db65f8f9e53d9675a92ed3aabdb5a416d69de81, but now our test suite is stalling specifically on one test, specifically on the macOS runners. The last log output we see is...

```
 testing/cli/test_cli_sct_register_multimodal.py::test_sct_register_multimodal_mt0_image_data_within_threshold[False-step=1,algo=slicereg,type=im,iter=5,smooth=0,metric=MeanSquares-mt/mt0_reg_slicereg_goldstandard.nii.gz] PASSED [ 91%]
Error: The operation was canceled
```

The test that comes immediately afterwards is:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8248383221/job/22558595898#step:5:431

Ah! It's `algo=dl`. So... some change I've implemented in this PR is incompatible with `algo=dl` on macOS 12 and 14 runners, but not macOS 11 runners?

I'm going to temporarily add a skip for the `algo=dl` test just to see if my theory is right.

joshuacwnewton: Looks like the tests are stalling (or, maybe not stalling, but taking a *very long time*) at one additional point: 

12:

```
testing/cli/test_sct_deepseg.py::test_install_model PASSED               [ 96%]
testing/cli/test_sct_deepseg.py::test_model_dict PASSED                  [ 97%]
```

14:

```
testing/cli/test_sct_deepseg.py::test_install_model PASSED               [ 96%]
testing/cli/test_sct_deepseg.py::test_model_dict PASSED                  [ 97%]
testing/cli/test_sct_deepseg.py::test_segment_nifti[/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2s/t2s.nii.gz-/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2s/t2s_seg-deepseg.nii.gz-t2s_seg_deepseg.nii.gz-seg_sc_t2star-0.9] PASSED [ 97%]
```

`algo=dl` and `sct_deepseg` both share `torch` as a common denominator, so I'll try to dig a little deeper into interactions between `torch`, `numpy`, and the macOS runners.

joshuacwnewton: In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332/commits/80419804947250978aae48f8ebb98924b956d71f, I think I confirmed that the stalls happen for both `macos-11` and `macos-12`. Meaning, I don't _think_ the updates from the `master` branch were the cause of the stalls.

Instead, I'm suspicious of the package updates that naturally occurred between the last passing run (Jan 12) and the first failing run (Mar 15). Given that `torch` seems to be the culprit here, I'm going to start by limiting the versions to see if that makes a difference.

joshuacwnewton: If my next set of tests don't work, then I think I might go nuclear and rebase this whole branch onto `master`, then replay each of my old commits one by one until something breaks.

joshuacwnewton: Aha! The tests [passed on macos-12](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8288556120/job/22686674485?pr=4332) after I downgraded from `dipy==1.9.0` to `dipy==1.8.0`!

I don't... have any clue why `dipy==1.9.0` would interact with the `torch` tests? But, I'll try to unpin all of the testing changes I made to see if I can isolate the changes.

----

Notably, though the macos-11 test still failed? And, it failed at the same test that was hanging on previous runs:

```
 testing/cli/test_sct_deepseg.py::test_segment_nifti[/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2.nii.gz-/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2_seg-manual.nii.gz-t2_seg_deepseg.nii.gz-seg_sc_contrast_agnostic-None] FAILED [ 97%]

 FAILED testing/cli/test_sct_deepseg.py::test_segment_nifti[/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2.nii.gz-/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2_seg-manual.nii.gz-t2_seg_deepseg.nii.gz-seg_sc_contrast_agnostic-None] - Exception: ('Download error', [ChunkedEncodingError(ProtocolError("Connection broken: TimeoutError(60, 'Operation timed out')", TimeoutError(60, 'Operation timed out')))])
```

A timeout error on downloading the contrast agnostic model? That would explain the apparent stalling, and it would perhaps explain intermittent behavior (i.e. maybe the macos-12 passing test was a fluke?).

joshuacwnewton: Ah, retrying it resulted in the `macos-11` tests passing. So, I guess it is a `dipy==1.9.0` issue, and that one timeout was just a really weird coincidence?

joshuacwnewton: tl;dr of the minimized comments:

- `dipy==1.8.0` works fine, as does upgrading `numpy` to the newest version.
- However, in the span of Jan-Mar, `dipy==1.9.0` was release as well. And, this version causes our macOS runners to hang on specific `torch` tests for reasons I have yet to identify.

requirements.txt
```diff
@@ -3,12 +3,11 @@
...
 
 # Avoid 1.6 and 1.7 due to method=restore bug: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4209#issuecomment-1739644328
-# dipy 1.8 and above cause conflicts, see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4319#issuecomment-1864670675
-dipy<1.6
+dipy!=1.6.*,!=1.7.*,<1.9.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
-ivadomed
+ivadomed @ git+https://github.com/ivadomed/ivadomed@master # append_to_freeze
```
joshuacwnewton, line 10: Note that the failing code quality check is due to the link above, which isn't actually a link that can be accessed via `curl`.

mguaypaq: I'm thinking `requirements.txt` and `requirements-freeze.txt` can be excluded from the link checker like `CHANGES.md`, since their non-comment links get tested by the installation checks anyway.

joshuacwnewton: Maybe there's still some value in checking the links found in the comments of `requirements.txt`? (e.g. in case an context-providing link for a version pin becomes inaccessible.) 

But, given that this is a minor corner case, I think I agree that we could be excluding the requirements files.

requirements.txt
joshuacwnewton: I've reported this upstream here:

- https://github.com/dipy/dipy/issues/3125

Just in case the dipy devs have insight into an upstream change that may be causing this behavior.

joshuacwnewton: I think for now I'm going to stick with `dipy==1.8.0`, and leave the troubleshooting for `1.9.0` to another PR: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332/commits/1f57611ec42fc2bcc19aaa940659aa5adadab3ed.

requirements.txt
```diff
@@ -21,10 +21,7 @@ msvc-runtime==14.29.30133; sys.platform == 'win32' # append_to_freeze
 nibabel
 nilearn
 nnunetv2
-# This pin is due to an incompatibility with pystrum==0.2, which is installed via voxelmorph -> neurite -> pystrum
-# See also: https://github.com/adalca/pystrum/issues/9
-# If pystrum is updated to 0.3, then we should replace this version pinning with `pystrum>=0.3`
-numpy<1.24
```
joshuacwnewton, line 27: There's one last thing I'd like to do for this PR: Troubleshoot the new warnings that have arisen due to upgrading various packages (`numpy`, `pandas`, etc.)

On our Ubuntu 20.04 runner, this looks like:

```
============================== warnings summary ===============================
testing/api/test_deepseg_gm.py::TestCore::test_segment_volume
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/deepseg_/gm.py:231: RuntimeWarning: invalid value encountered in cast
    pred_slices = np.asarray(pred_slices, dtype=np.uint8)

testing/api/test_labels.py::test_continuous_vertebral_levels[test_image0]
testing/api/test_labels.py::test_continuous_vertebral_levels[test_image1]
testing/api/test_labels.py::test_continuous_vertebral_levels[test_image2]
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/labels.py:425: RuntimeWarning: invalid value encountered in scalar divide
    continuous_values[iz] = level + 2.0 * distance_from_level / float(length_levels[level])

testing/api/test_process_seg.py: 190 warnings
testing/cli/test_cli_sct_compute_compression.py: 567 warnings
testing/cli/test_cli_sct_process_segmentation.py: 170 warnings
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/process_seg.py:91: DeprecationWarning: `np.math` is a deprecated alias for the standard library `math` module (Deprecated Numpy 1.25). Replace usages of `np.math` with `math`
    angle_AP_rad = np.math.atan2(np.linalg.det([v0, v1]), np.dot(v0, v1))

testing/api/test_process_seg.py: 190 warnings
testing/cli/test_cli_sct_compute_compression.py: 567 warnings
testing/cli/test_cli_sct_process_segmentation.py: 170 warnings
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/process_seg.py:95: DeprecationWarning: `np.math` is a deprecated alias for the standard library `math` module (Deprecated Numpy 1.25). Replace usages of `np.math` with `math`
    angle_RL_rad = np.math.atan2(np.linalg.det([v0, v1]), np.dot(v0, v1))

testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements[dummy_lesion0-0.001]
testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements[dummy_lesion1-0.001]
testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements[dummy_lesion2-0.01]
testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements_without_segmentation[dummy_lesion0-0.001]
testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements_without_segmentation[dummy_lesion1-0.001]
testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements_without_segmentation[dummy_lesion2-0.01]
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_analyze_lesion.py:220: FutureWarning: save is not part of the public API, usage can give unexpected results and will be removed in a future version
    writer.save()

testing/cli/test_cli_sct_compute_compression.py: 36 warnings
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/pandas/core/algorithms.py:522: DeprecationWarning: np.find_common_type is deprecated.  Please use `np.result_type` or `np.promote_types`.
  See https://numpy.org/devdocs/release/1.25.0-notes.html and the docs for more information.  (Deprecated NumPy 1.25)
    common = np.find_common_type([values.dtype, comps_array.dtype], [])

testing/cli/test_cli_sct_compute_compression.py::test_sct_compute_compression
testing/cli/test_cli_sct_compute_compression.py::test_sct_compute_compression
testing/cli/test_cli_sct_compute_compression.py::test_sct_compute_compression_sex_F
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/pandas/core/dtypes/cast.py:1641: DeprecationWarning: np.find_common_type is deprecated.  Please use `np.result_type` or `np.promote_types`.
  See https://numpy.org/devdocs/release/1.25.0-notes.html and the docs for more information.  (Deprecated NumPy 1.25)
    return np.find_common_type(types, [])

testing/cli/test_cli_sct_label_vertebrae.py::test_sct_label_vertebrae_initial_disc_too_low
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/labels.py:133: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of -1 to uint8 will fail in the future.
  For the old behavior, usually:
      np.array(value).astype(dtype)
  will give the desired result (the cast overflows).
    out.data[x, y, z_rpi] = value

testing/cli/test_cli_sct_register_multimodal.py: 5 warnings
testing/cli/test_cli_sct_register_to_template.py: 152 warnings
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/numpy/core/_methods.py:198: RuntimeWarning: invalid value encountered in scalar divide
    ret = ret.dtype.type(ret / rcount)

testing/cli/test_cli_sct_warp_template.py::test_sct_warp_template_warp_full_PAM50
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/scipy/ndimage/_measurements.py:1541: RuntimeWarning: invalid value encountered in scalar divide
    results = [sum(input * grids[dir].astype(float), labels, index) / normalizer

testing/cli/test_sct_deepseg.py::test_segment_nifti[/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2.nii.gz-/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2_seg-manual.nii.gz-t2_seg_deepseg.nii.gz-seg_sc_contrast_agnostic-None]
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/torch/utils/data/dataloader.py:558: UserWarning: This DataLoader will create 8 worker processes in total. Our suggested max number of worker in current system is 4, which is smaller than what this DataLoader is going to create. Please be aware that excessive worker creation might get DataLoader running slow or even freeze, lower the worker number to avoid potential slowness/freeze if necessary.
    warnings.warn(_create_warning_msg(

testing/cli/test_sct_deepseg.py::test_segment_nifti[/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2.nii.gz-/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2_seg-deepseg_rootlets.nii.gz-t2_seg_deepseg.nii.gz-seg_spinal_rootlets_t2w-None]
testing/cli/test_sct_deepseg.py::test_segment_nifti_multiclass[/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/data/sct_testing_data/t2/t2_fake_lesion.nii.gz-fnames_seg_manual0-t2_deepseg.nii.gz-suffixes0-seg_sc_lesion_t2w_sci-0.5]
  /home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/utilities/plans_handling/plans_handler.py:37: UserWarning: Detected old nnU-Net plans format. Attempting to reconstruct network architecture parameters. If this fails, rerun nnUNetv2_plan_experiment for your dataset. If you use a custom architecture, please downgrade nnU-Net to the version you implemented this or update your implementation + plans.
    warnings.warn("Detected old nnU-Net plans format. Attempting to reconstruct network architecture "

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========== 440 passed, 13 skipped, 2065 warnings in 424.42s (0:07:04) ==========
```

I'll troubleshoot locally, then push some fixes where possible. :)

joshuacwnewton: Hrm. Several of these deprecation warnings stem from `numpy` -> `pandas` (i.e. `pandas` is out of date, but `numpy` is up to date).

These issues have been presumably fixed upstream by `pandas`, but something must be holding `pandas` back. Why? :thinking: 

Ah. We're on `pandas==1.5.3` right now, which is the last version prior to `2.0.0`. And, we're not on `pandas==2.0.0` [because of `ivadomed`](https://github.com/ivadomed/ivadomed/blob/1fccf77239985fc3be99161f9eb18c9470d65206/requirements.txt#L12-L13).

```python
# `pandas==2.0` removed the `DataFrame.append` call, which we currently use in approx. 13 places (per PyCharm search)
pandas>=1.1,<2.0
```

To fix this, we'll have to go through another cycle of "update ivadomed" -> "make a new ivadomed release" -> "update SCT", which is a bit cumbersome :sweat:. 

For now, I can just add some ignores to our `setup.cfg`?

mguaypaq: If you do add some temporary ignores to `setup.cfg`, can you also open an issue about it (and link it in the comments, ideally)?

joshuacwnewton: Done! :)

testing/api/test_qmri.py
```diff
@@ -20,7 +20,7 @@ def make_sct_image(data):
     data: scalar
     """
     affine = np.eye(4)
-    nii = nibabel.nifti1.Nifti1Image(np.array([data, data]), affine)
+    nii = nibabel.nifti1.Nifti1Image(np.array([data, data], dtype=np.uint32), affine)
```
joshuacwnewton, line 23: Re: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4408

> I encountered this exact error on https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332, too. (See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4332/commits/cb7157ff4c202dd87b3ce59efdd4826897bdd413) So, the errors will be fixed when #4332 is merged.
> 
> However, given how "general" this error message is, I am concerned that there will be instances of this issue elsewhere in our code that aren't caught by the test suite. So, I'd like to consider adding a fix to the `Image` constructor (either by passing `dtype=np.int64` when the input array is of type `np.int64` to preserve the old behavior, or by converting input arrays to `np.int32` before loading them to fix the potential incompatibilities hinted at by the code).

I realize now that both of these tests (`test_qmri.py`, `test_deepseg_sc.py`) call the `Nifti1Image` constructor directly, bypassing any fixes we might add to `Image`. So, in order to test a fix that gets added to the `Image` constructor, I will also need to rewrite these dummy test fixtures.

joshuacwnewton: Ah, wait! If I rewrite these tests to pass the problematic `np.int64` array directly to the `Image` constructor, then there are no failures, even without an explicit `np.int64` fix!

This is because, when we initialize an `Image` object from a numpy array, we never actually call `nib.Nifti1Image`. Instead, we simply store the numpy array and a separate `Nifti1Header` object.

We do call `nib.Nifti1Image` when _saving_ an `Image` object. But, we explicitly call it with the `header` argument, always. And, by specifying `header`, we bypass the check (as mentioned by the error message):

> To use this type, pass an explicit header or dtype argument to Nifti1Image()

So, as long as we're specifying the header for all uses of `Nifti1Image`, we can trust that there won't be any unexpected failures down the line.

(Though, we should probably still heed nibabel's compatibility warning anyway!)

joshuacwnewton: I think I'm going to mark this as resolved in the scope of this PR, but perhaps leave #4408 open to consider https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4408#issuecomment-2007396098.

setup.cfg
```diff
@@ -44,8 +44,8 @@ filterwarnings =
     # Source: test_sct_register_multimodal_mto_image_data --> `algorithms.py :: vxm.networks.VxmDense()`
     ignore:.*in an upcoming release, it will be required to pass the indexing argument.*:UserWarning:torch.*:
     # This warning is due to a deprecation for a call that is still used by many upstream packages. In our case:
-    # - wandb: https://github.com/wandb/wandb/issues/6711
-    ignore:pkg_resources is deprecated as an API:DeprecationWarning
+    # - monai: https://github.com/Project-MONAI/MONAI/issues/7559
```
mguaypaq, line 47: Thanks for the attempted cleanup!! Even though the result is just a comment change, it's very important for future understandability. Plus, it led you to report it upstream, which is great open source citizenship.

requirements.txt
```diff
@@ -39,6 +35,9 @@ psutil
 pyqt5==5.12.3
 # PyQt5-sip should be updated when PyQt5 is updated: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4354
 pyqt5-sip<12.13.0
+# pystrum==0.2 is incompatible with newer versions of numpy (installed via voxelmorph -> neurite -> pystrum)
+# See also: https://github.com/adalca/pystrum/issues/9
+pystrum>=0.3
```
mguaypaq, line 40: :heart:

testing/api/test_qmri.py
mguaypaq: Thanks for the investigation!

mguaypaq (approved):
Great work, thanks. It almost feels like spring cleaning! I'll add a commit in a few minutes (see comment below), then it should be good to merge.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -217,7 +217,7 @@ def pack_measures(self):
             pickle.dump(self.distrib_matrix_dct, handle)
 
         # Save Excel
-        writer.save()
+        writer.close()
```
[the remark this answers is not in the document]
mguaypaq: Done in e059f79635912837b9778413f34fe4a72af98d49.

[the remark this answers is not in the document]
joshuacwnewton: Ahh, good eye! 

In hindsight, `close()` was definitely a smell for a context manager. Smart thinking for catching this! :tada:

PR #4346, opened 2024-01-19
Mitigate scaling issues (`0.999`) in both `sct_deepseg` and `sct_label_vertebrae`

### LINKED ISSUE #3232: Scaling issue with nibabel.save() when header datatype is `int`, but array datatype is `float`

joshuacwnewton:
## Description

In https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3232#issuecomment-1883229811, I ran into a two-pronged bug:

1. `sct_deepseg` uses `nib.save()` rather than `Image.save()`, meaning that it was susceptible to the float/int scaling issues (1.0 -> 0.999) as previously documented in #3232.
2. When using a poorly-scaled segmentation with `sct_label_vertebrae`, the multiplication part of the "clean labels" step would then transfer the scaling issues to the integer labels themselves.

This PR fixes both issues by:

- Working with `spinalcordtoolbox.image.Image` objects in `sct_deepseg` (since `Image.save()` contains a mitigation for the scaling issues.)
- Converting the segmentation to a boolean mask before multiplying it with the label image.

Note: I'm not sure whether it's worth writing a test for this, given that the scaling behavior is something that is unique to `nibabel`. Once the `nib.save()` function is removed from usage, there's practically no chance that this behavior will come up again.

## Linked issues

Partially fixes  #3232. (Should this be closed? Or should we keep it open until we replace all existing usages of `nib.save()` with `Image.save()`?)

### DIFF
--- a/spinalcordtoolbox/vertebrae/core.py
+++ b/spinalcordtoolbox/vertebrae/core.py
@@ -309,10 +309,12 @@ def crop_labels(im_labeled_seg, im_seg):
     modifying on the labeled segmentation in-place.
 
     :param Image im_labeled_seg: labeled segmentation
-    :param Image im_seg: segmentation (with values in [0, 1])
+    :param Image im_seg: segmentation (can be softseg or binary)
     :return: None
     """
-    im_labeled_seg.data *= im_seg.data
+    boolean_mask = (im_seg.data != 0)    # Nonzero -> True | Zero -> False
+    im_labeled_seg.data *= boolean_mask  # Keep only voxels that were nonzero in the segmentation image
+    # Note: NumPy will cast the bool type to the integer type, with False and True converted to 0 and 1 respectively.
 
 
 def expand_labels(im_labeled_seg):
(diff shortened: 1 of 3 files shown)


mguaypaq:
> Partially fixes  #3232. (Should this be closed? Or should we keep it open until we replace all existing usages of `nib.save()` with `Image.save()`?)

I'd like to keep the issue open while we still have calls to `nib.save()`, if you don't mind. I've edited the PR description to (hopefully) prevent auto-closing.

spinalcordtoolbox/vertebrae/core.py
```diff
@@ -309,10 +309,12 @@ def crop_labels(im_labeled_seg, im_seg):
...
 
     :param Image im_labeled_seg: labeled segmentation
-    :param Image im_seg: segmentation (with values in [0, 1])
+    :param Image im_seg: segmentation (can be softseg or binary)
     :return: None
     """
-    im_labeled_seg.data *= im_seg.data
+    boolean_mask = (im_seg.data != 0)    # Nonzero -> True | Zero -> False
+    im_labeled_seg.data *= boolean_mask  # Keep only voxels that were nonzero in the segmentation image
+    # Note: NumPy will cast the bool type to the integer type, with False and True converted to 0 and 1 respectively.
```
mguaypaq, line 317: I like this fix! It's very clean.

mguaypaq (approved):
Thanks for the fix, it looks good. I added a tiny documentation fix as an extra commit.

Issue #4352, opened 2024-02-01
`macos` CI runners fail to install `onnxruntime==1.17.0` wheels with `macosx_11_0` version specifiers


joshuacwnewton:
Notes:

- Most of our CI is passing (incl. batch_processing.sh, normal test suite, etc.)
- However, the "Generate Release" CI is failing with the following exception:

  ```
  ERROR: Ignored the following versions that require a different python version: 2.5.2 Requires-Python !=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,<3.9dev,>=2.7; 2.7.0 Requires-Python >=3.6, <3.9; 2.7.1 Requires-Python >=3.6, <3.9; 2.7.2 Requires-Python >=3.6, <3.9; 2.7.4 Requires-Python >=3.6, <3.9; 2.8.0 Requires-Python >=3.6, <3.9
  ERROR: Could not find a version that satisfies the requirement onnxruntime==1.17.0 (from versions: 1.7.0, 1.8.0, 1.8.1, 1.9.0, 1.10.0, 1.11.0, 1.11.1, 1.12.0, 1.12.1, 1.13.1, 1.14.1, 1.15.0, 1.15.1, 1.16.0, 1.16.1, 1.16.2, 1.16.3)
  ERROR: No matching distribution found for onnxruntime==1.17.0
  ```

- The release workflow works by installing SCT, freezing the dependencies, then re-installing SCT.

It is strange that the initial installation works without error, but the frozen installation fails. I'll have to investigate why this might be.

joshuacwnewton:
So, it seems like this is an OS-specific issue:

- The requirements are frozen using an Ubuntu runner.
- Then, these frozen requirements are used to test Ubuntu/macOS/Windows.
- It is the macos-11 CI that consistently fails.

If I compare the requirements from yesterday's passing run and today's failing run, I see that it is only [`onnxruntime`](https://www.diffchecker.com/Nim5lcZP/) that got updated. So, maybe there's an upstream issue where the wheel for macOS didn't get uploaded?

joshuacwnewton:
Reported upstream here: https://github.com/microsoft/onnxruntime/issues/19371

joshuacwnewton:
I got to the bottom of the issue!! :tada:

The cause of the failure is _very_ curious:

- When macOS 11 was in beta, the version number "10.16" was briefly used.
[...]

PR #4353, opened 2024-02-01
Update GitHub Actions versions (macOS runners, Windows runners, Node.js 16 actions -> 20)

joshuacwnewton:
## Description

This is a pretty straightforward maintenance PR! 

Relevant GitHub blog posts:

- Node.js 16 deprecated: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
- macos-11 deprecated: https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
- windows-2022 released: https://github.blog/changelog/2021-11-16-github-actions-windows-server-2022-with-visual-studio-2022-is-now-generally-available-on-github-hosted-runners/

## Linked issues

N/A


.github/workflows/tests.yml
```diff
@@ -264,18 +264,18 @@ jobs:
        run: |
          ./.ci.sh -t
 
-  macos-12:
-    name: macOS 12.0 (Monterey)
+  macos-14:
+    name: macOS 14.0 (Sonoma)
```
joshuacwnewton, line 268: Given that the macOS 14 image is the version that runs on M1 Macs, is it possible that we want to promote this to one of the actions that runs on PRs?

.github/workflows/tests.yml
```diff
@@ -396,12 +396,12 @@ jobs:
 
   windows-native:
     name: Windows (Native)
-    runs-on: windows-2019
+    runs-on: windows-2022
```
joshuacwnewton, line 399: It's possible that we want to selectively keep the old `windows-2019` actions, too, given [this quote](https://github.blog/changelog/2021-11-16-github-actions-windows-server-2022-with-visual-studio-2022-is-now-generally-available-on-github-hosted-runners/):

> The Windows Server 2022 runner image has [different tools and tool versions](https://github.com/actions/virtual-environments/issues/3949) than Windows Server 2019.

joshuacwnewton:
It seems as though all of the tests pass (phew!) -- the current CI failure is due to the upstream issue:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4352

.github/workflows/tests.yml
mguaypaq: Sure! I expect macOS14 to pick up either installation failures (which don't have to run on every PR) or numerical stability failures (which would be better to have on every PR). For the latter reason, we may also want to have results of `batch_processing.sh` for both macOS 12 and macOS 14 in the next release.

.github/workflows/tests.yml
mguaypaq: Sure. Unlike macOS 12 vs macOS 14, which has different CPU architecture, I expect Windows 2019 vs 2022 to only be relevant for installation failures. So, it doesn't have to run on every PR, but once in a while wouldn't hurt.

mguaypaq (approved):
Just one thing: from the commit message of 958cc639410df43c9e92cb6bf7f466aaa4dfbd05:
> setup-wsl@v2 -> not released yet (https://github.com/Vampire/setup-wsl/issues/52)

It looks like `setup-wsl@v3` is now released, so this one can be updated as well.

PR #4345, opened 2024-01-17
Integrate NNUnet/MONAI models into DeepSeg CLI (contrast-agnostic, SCI, rootlets, Zurich mouse)

joshuacwnewton:
## Description

This PR adds 4 new models (contrast agnostic softseg, SCI lesion/SC seg, rootlets seg, Zurich mouse GM/WM) trained using 2 separate frameworks (`monai` and `nnunetv2`).

I've tried my best to leave the original `sct_deepseg` (ivadomed) infrastructure untouched, siloing the new code the best I can. That way, it can be reviewed in isolation, without worrying too much about interfering with the existing codebase.

This PR is quite large. However, I've tried to organize the commits into groups to make the changes easier to understand:

- `Commits 01-03`: Update `requirements.txt`
- `Commits 04-05`: Add model files to `deepseg/models.py`
- `Commits 06-11`: Update `sct_deepseg` test with cases for the new models
- `Commits 12-13`: Prepare CLI -> API entry points (`sct_deepseg` -> `deepseg/inference.py`)
- `Commit 14`: Add the existing MONAI inference script (as-is) from the project repo
    - `Commits 15-31`: Refactor the MONAI inference script (module organization, code style, etc.)
- `Commit 32`: Add the existing NNUnet inference script (as-is) from the project repo
    - `Commits 33-41`: Refactor the NNUnet inference script (module organization, code style, etc.)
- `Commits 42-45`: Unifying the MONAI/NNUnet interfaces (deduplication, readability)

As you can see, most of these commits are refactoring, and could potentially be squashed for ease of review. But, I wanted to be transparent about the changes I've made to the original scripts (see f364c242c7d2b14e038a39070d0b08540af2d646).

Please let me know if there is anything else I can do to make this PR easier to review.

## TODOs

- [X] Update `sct_testing_data` to fix discovered T2s issue
- [x] Update `sct_testing_data` with sample input and output for new models:
  - [X] Softseg (using existing `t2.nii.gz` + manual seg ground truth)
  - [X] SCI - SC (using existing `t2.nii.gz` + manual seg ground truth)
  - [x] SCI - Lesion
  - [X] Rootlets (using existing `t2.nii.gz` + Added new rootlets ground truth file)
  - [x] Zurich mouse
- [X] Create official releases for the SCI and rootlets models.
  - [X] Rootlets
  - [X] SCI
- [x] Write long_descriptions for nnUNet models
  - [x] Rootlets
  - [x] SCI 
  - [x] Zurich Mouse

[...]

PR #4355, opened 2024-02-01
Address test warnings by pinning `pyqt5-sip<12.13.0` and updating `setup.cfg` ignore entries (`pkg_resources`)

### LINKED ISSUE #4354: Address new test warnings (and double-check the existing ignore list)
Our test suite is reporting new warnings, so I wanted to add a maintenance TODO for myself.

joshuacwnewton:
## Description

This is a small PR to revisit the warnings caught by our test suite:

- pyqt5-sip: See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4354
- pkg_resources deprecation --> ~~fixed upstream~~ Not true! `wandb` still uses `pkg_resources`

## Linked issues

Fixes #4354.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,6 +35,8 @@ psutil
 # Additionally, for versions between 5.12-5.14, the only release to have Python 3.9 wheels is 5.12.3. So, we pin it.
 # See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1533181462
 pyqt5==5.12.3
+# PyQt5-sip should be updated when PyQt5 is updated: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4354
+pyqt5-sip<12.13.0
 pytest
 pytest-cov
 requests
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -43,11 +43,6 @@ filterwarnings =
     ignore:.*in an upcoming release, it will be required to pass the indexing argument.*:UserWarning:torch.*:
```
joshuacwnewton, line 43: I think we could do a little more work with the `torch` related warnings, but I think we might be able to tackle those as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4345, given that we're upgrading torch in that PR.

joshuacwnewton:
Removing this from the 6.2 milestone since it is not critical to review this prior to 6.2.

mguaypaq (approved):
Looks good, thanks for this bit of maintenance!

requirements.txt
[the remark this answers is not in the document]
joshuacwnewton: Addressed in [82b23fc](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4355/commits/82b23fc66ed7f216f9372c20821b8a51a17e7b1b).

setup.cfg
joshuacwnewton: Updated in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4355/commits/e8c2cee755f995d1aaf3781329357dfae7610dea.

PR #4368, opened 2024-02-13
Document Rosetta 2 as a requirement for installation on Apple silicon (M1, M2, M3)

### LINKED ISSUE #4342: Document better the need for Rosetta 2 on Apple Silicon (M1, M2, M3)
Yesterday @plbenveniste was trying to install SCT on his new Apple M3 laptop, and ran into the error `Bad CPU type in executable` for the miniconda executable.

This problem is known (see [this forum post](https://forum.spinalcordmri.org/t/installing-sct-on-mac-m1-os-version-13-0/997) from last year), but I didn't remember, so we spent some time debugging it. The immediate fix is very simple: make sure that [Rosetta 2](https://en.wikipedia.org/wiki/Rosetta_(software)) is installed before running the SCT installer. But we should document it better:
- [ ] We should mention Rosetta 2 as a pre-requisite on the [MacOS installation page](https://spinalcordtoolbox.com/user_section/installation/mac.html).
- The installer should:
  - [ ] auto-detect whether it's running on Apple Silicon (that is, M1, M2, M3)
  - [ ] auto-detect whether Rosetta 2 is installed
  - [ ] print an appropriate message
  - [ ] maybe run `softwareupdate –install-rosetta`?

The good news is that we can test out the installer changes on @plbenveniste's laptop.

As an aside, Apple Silicon gives us another reason to work on [unpinning our dependency on `PyQt5<5.15`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3925#issuecomment-1532116693): from what I can see, it looks like PyQt5 has binary wheels for this platform [starting from 5.15.10](https://pypi.org/project/PyQt5/5.15.10/#files). And I'm unsure about this, but I think the fact that the wheels all use `abi3` as their [ABI tag](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#abi-tag) instead of `none` (as some of them did previously), means that `cp37` indicates a _minimum_ version of CPython 3.7, instead of an _exact_ version of CPython 3.7. So, we might have fewer installation issues on Windows, too.

mguaypaq:
[Rosetta 2](https://en.wikipedia.org/wiki/Rosetta_(software)#Rosetta_2) is needed to translate `x86_64` executables (which we rely on) so they can run on the `arm64` architecture of Apple silicon chips. This PR documents this requirement on the [Mac installation page for SCT](https://spinalcordtoolbox.com/user_section/installation/mac.html) and in the SCT installer, which:
* detects whether the computer is running macOS;
* if so, it tries to run a builtin command (`true`) while requesting the `x86_64` architecture;
* if this fails, it offers to install Rosetta 2.

The new code mirrors the check for `gcc` which appears right below in the installer.

The test uses the `arch` command, which has different functionality on Linux than on macOS; see [the linux manual page for `arch`](https://www.man7.org/linux/man-pages/man1/arch.1.html) and [an old macOS manual page for `arch`](https://www.unix.com/man-page/osx/1/arch/). I checked on a recent M2 laptop, and the `arch` command now also recognizes `arm64`, `arm64e`, and `x86_64h` as architecture names, but otherwise seems to still work the same way.

Fixes #4342.

### DIFF
--- a/documentation/source/user_section/installation/mac.rst
+++ b/documentation/source/user_section/installation/mac.rst
@@ -13,6 +13,16 @@ Supported Operating Systems
 * MacOS >= 10.13
 
 
+Mac computers with Apple silicon
+--------------------------------
+
+If your computer uses Apple silicon (M1, M2, M3, etc.), you need to make sure `Rosetta 2 <https://en.wikipedia.org/wiki/Rosetta_(software)#Rosetta_2>`__ is installed before installing or running SCT. To do this, you can open a Terminal and run:
+
+.. code:: sh
+
+    softwareupdate --install-rosetta
+
+
 Gnu Compiler Collection (gcc)
 -----------------------------
 
@@ -23,7 +33,7 @@ You need to have ``gcc`` installed. Check to see if ``gcc`` is installed by open
     gcc --version
 
 
-If it isn't installed, we recommend installing `Homebrew <https://brew.sh/>`_ and then run:
+If it isn't installed, we recommend installing `Homebrew <https://brew.sh/>`__ and then run:
 
 .. code:: sh
 
@@ -37,7 +47,7 @@ Installation Options
 Option 1: Install from Package (recommended)
 --------------------------------------------
 
-The simplest way to install SCT is to do it via a stable release. First, navigate to the `latest release <https://github.com/spinalcordtoolbox/spinalcordtoolbox/releases>`_, then download the install script for SCT (``install_sct-<version>_macos.sh``). Major changes to each release are listed in the :doc:`/dev_section/CHANGES`.
+The simplest way to install SCT is to do it via a stable release. First, navigate to the `latest release <https://github.com/spinalcordtoolbox/spinalcordtoolbox/releases>`__, then download the install script for SCT (``install_sct-<version>_macos.sh``). Major changes to each release are listed in the :doc:`/dev_section/CHANGES`.
 
 Once you have downloaded SCT, open a new Terminal in the location of the downloaded script, then launch the installer the ``bash`` command. For example, if the script was downloaded to `Downloads/`, then you would run:
 
@@ -124,7 +134,7 @@ Procedure:
 Option 4: Install with Docker
 -----------------------------
 
-`Docker <https://www.docker.com/what-container>`_ is a portable (Linux, macOS, Windows) container platform.
+`Docker <https://www.docker.com/what-container>`__ is a portable (Linux, macOS, Windows) container platform.
 
 In the context of SCT, it can be used:
 
@@ -136,7 +146,7 @@ In the context of SCT, it can be used:
 Basic Installation (No GUI)
 ***************************
 
-First, `install Docker Desktop <https://docs.docker.com/desktop/install/mac-install/>`_. Then, follow the examples below to create an OS-specific SCT installation.
+First, `install Docker Desktop <https://docs.docker.com/desktop/install/mac-install/>`__. Then, follow the examples below to create an OS-specific SCT installation.
 
 
 Docker Image: Ubuntu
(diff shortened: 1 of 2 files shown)


documentation/source/user_section/installation/mac.rst
```diff
@@ -23,7 +33,7 @@ You need to have ``gcc`` installed. Check to see if ``gcc`` is installed by open
     gcc --version
 
 
-If it isn't installed, we recommend installing `Homebrew <https://brew.sh/>`_ and then run:
+If it isn't installed, we recommend installing `Homebrew <https://brew.sh/>`__ and then run:
```
joshuacwnewton, line 36: Is there a reason that these external hyperlinks were changed to use anonymous targets? ([`_` vs `__`](https://stackoverflow.com/a/74258281))

e.g. Had you run into the error "Duplicate explicit target name"?

install_sct
```diff
@@ -183,6 +183,28 @@ function check_requirements() {
...
+        softwareupdate --install-rosetta
+        # check if we can now run x86_64 executables
+        if ! arch -x86_64 true >/dev/null 2>&1; then
+          die "ERROR: still cannot run x86_64 executables. Please contact SCT team for assistance."
+        fi
+      else
+        die "Please install \"Rosetta 2\" by running \"softwareupdate --install-rosetta\" and restart SCT installation."
+      fi
+    fi
+  fi
```
joshuacwnewton, line 207: I don't have an Apple Silicon Mac to check, but I do notice that the structure of the addition is [more or less identical](https://www.diffchecker.com/M4Lb416P/) to our existing `gcc` check, with the main functional difference being `arch -x86_64 true` over `gcc --version`.

As you've explained, `arch` on macOS can be used to ["run selected architecture of a universal binary"](https://www.unix.com/man-page/osx/1/arch/). But, the use of [`true`](https://man7.org/linux/man-pages/man1/true.1.html) ("do nothing, successfully") was new to me! Neat. :)

So, `arch -x86_64 true` will fail on Macs without Rosetta 2? :thinking:

joshuacwnewton: While researching this, I came across [this SE answer](https://apple.stackexchange.com/a/464165), which suggests both:

```bash
    if ! (arch -arch x86_64 uname -m > /dev/null) ; then
```

Which I guess is more or less identical to `true` anyway, since we're just testing success/failure.

And: 

```bash
softwareupdate --install-rosetta --agree-to-license
```

The behavior comparing "`--agree-to-license` vs. nothing" seems to be documented [here](https://discussions.apple.com/thread/253378816?sortBy=best). tl;dr it looks like it's just a single CLI prompt. 

Given that:

- We're already skipping the Rosetta installation for non-interactive installations anyway
- For interactive installations, the prompt would happen at the start of the installation (in much the same way our other interactive prompts are)
 
I think it should be fine to exclude the `--agree-to-license` as you've done here, since I don't expect it to be any sort of additional nuisance.

joshuacwnewton (approved):
I had some small comments, but the bulk of this PR LGTM!

documentation/source/user_section/installation/mac.rst
mguaypaq: I think anonymous targets are easier to think about, since they don't involve any global state (that is, the document-wide set of all targets, case-insensitive and whitespace-insensitive), so that's what I default to.

I think we should reserve named hyperlink references to cases where we actually want to refer to a URL multiple times, and in that case we should use the explicit format:
```rst
Here is some text mentioning a `carefully chosen link text`_.
And once again, this is a `Carefully Chosen Link Text`_ that is descriptive.

.. _Carefully chosen link text: https://www.example.com/
```

joshuacwnewton: That makes sense to me! I can open an issue to follow up on this for the rest of the documentation. :)

install_sct
mguaypaq: > So, `arch -x86_64 true` will fail on Macs without Rosetta 2? 🤔

Yes, it should. I wasn't able to test this exactly, since I was testing this on a laptop which already had Rosetta 2 (and I have no idea how to uninstall it), but instead I tried a different architecture and got:

```
~$ arch -x86_64h true
arch: posix-spawnp: true: Bad CPU type in executable
```

> I think it should be fine to exclude the `--agree-to-license` as you've done here, since I don't expect it to be any sort of additional nuisance.

Right. I think the spirit of `--agree-to-license` is that it's for sysadmins running deployment scripts on machines they control, so that they're empowered to agree to the license. In the case of the SCT installation script, I think we (the developers) can't agree to the license on behalf the the user, so we shouldn't hide it.

joshuacwnewton: Perfect!

PR #4369, opened 2024-02-14
Reverse spacings to match nnUNet's conventions

mguaypaq:
As per [this comment on #4345](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4345#discussion_r1489625543).

See: https://github.com/MIC-DKFZ/nnUNet/blob/997804c7510634dc8fd83f1194b434c60815a93e/nnunetv2/imageio/nibabel_reader_writer.py#L49-L52

### DIFF
--- a/spinalcordtoolbox/deepseg/inference.py
+++ b/spinalcordtoolbox/deepseg/inference.py
@@ -227,7 +227,8 @@ def segment_nnunet(path_img, tmpdir, predictor):
     data = np.expand_dims(data, axis=0).astype(np.float32)
     pred = predictor.predict_single_npy_array(
         input_image=data,
-        image_properties={'spacing': img_in.dim[4:7]},
+        # The spacings also have to be reversed to match nnUNet's conventions.
+        image_properties={'spacing': img_in.dim[6:3:-1]},
     )
     # Lastly, we undo the transpose to return the image from [z,y,x] (SimpleITK) to [x,y,z] (nibabel)
     pred = pred.transpose([2, 1, 0])


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -227,7 +227,8 @@ def segment_nnunet(path_img, tmpdir, predictor):
     data = np.expand_dims(data, axis=0).astype(np.float32)
     pred = predictor.predict_single_npy_array(
         input_image=data,
-        image_properties={'spacing': img_in.dim[4:7]},
+        # The spacings also have to be reversed to match nnUNet's conventions.
+        image_properties={'spacing': img_in.dim[6:3:-1]},
```
joshuacwnewton, line 231: Python's indexing syntax always catches me off guard when I first see it, hehe.

- `[4:7]`: Start at 4 (inclusive), stop at 7 (exclusive), step by 1 (implicit), result: `[4, 5, 6]`
- `[6:3:-1]`: Start at 6 (inclusive), stop at 3 (exclusive), step by -1 (explicit), result: `[6, 5, 4]`

I'd never really thought about it before, but I guess the indexing syntax basically matches the syntax of `range`? (Start, stop, step)

```python
>>> int_list = [0, 1, 2, 3, 4, 5, 6, 7, 8]
>>> int_list[4:7]
[4, 5, 6]
>>> int_list[6:3:-1]
[6, 5, 4]
>>> list(range(4, 7))
[4, 5, 6]
>>> list(range(6, 3, -1))
[6, 5, 4]
```

joshuacwnewton (approved):
Thanks for catching this!! This would have been quite a subtle bug to have to investigate down the road had we missed this. (It would have worked fine for isotropic images, but the results would be off for anisotropic images.) :sweat:

spinalcordtoolbox/deepseg/inference.py
mguaypaq: > I'd never really thought about it before, but I guess the indexing syntax basically matches the syntax of `range`? (Start, stop, step)

Exactly!

PR #4381, opened 2024-02-23
Rename `_gpu` input param to `_device` to match upstream breaking API change

### LINKED ISSUE #4380: `nnunetv2` 2.3.1 renamed keyword `perform_everything_on_gpu`, triggering test failures in CI
Context: https://github.com/MIC-DKFZ/nnUNet/commit/1315d2f05a2797438b19be6ceae71de48f2b8452

(Was a little tricky to find, as there was no issue, PR, or changelog entry describing this change.)

joshuacwnewton:
## Description

Context: https://github.com/MIC-DKFZ/nnUNet/commit/1315d2f05a2797438b19be6ceae71de48f2b8452

## Linked issues

Fixes #4380.

### DIFF
--- a/spinalcordtoolbox/deepseg/nnunet.py
+++ b/spinalcordtoolbox/deepseg/nnunet.py
@@ -34,7 +34,7 @@ def create_nnunet_from_plans(path_model):
         tile_step_size=tile_step_size,  # changing it from 0.5 to 0.9 makes inference faster
         use_gaussian=True,  # applies gaussian noise and gaussian blur
         use_mirroring=False,  # test time augmentation by mirroring on all axes
-        perform_everything_on_gpu=False,
+        perform_everything_on_device=False,
         device=torch.device('cpu'),
         verbose=False,
         verbose_preprocessing=False,


valosekj (approved):
LGTM! Tested with `sct_deepseg -task seg_spinal_rootlets_t2w` and working fine! Thanks!

valosekj:
btw. does this issue originate from the fact that we [do not specify](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/requirements.txt#L23) any particular `nnunetv2` version when installing SCT? Meaning that similar issues might happen in the future when `nnunetv2` is updated again? Is there a specific reason we do not fix to a specific version? Is it to prevent dependency conflicts? For example, in [model_seg_sci](https://github.com/ivadomed/model_seg_sci), we used `nnunetv2==2.2.1` (see line [here](https://github.com/ivadomed/model_seg_sci/blob/9832c82f25d2a7803c94cb53fb29303202cd1b31/packaging/requirements.txt#L3C1-L3C16)).

joshuacwnewton:
> btw. does this issue originate from the fact that we [do not specify](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/requirements.txt#L23) any particular `nnunetv2` version when installing SCT? Meaning that similar issues might happen in the future when `nnunetv2` is updated again?

Exactly!

Naga had a similar question here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4345#discussion_r1457508812. So, just to reshare my thoughts from that PR: 

> Note that for stable releases of SCT, the versions will still get pinned (via a separate `requirements-freeze.txt` file), so the question of "whether or not to pin dependencies" is mainly a question for non-stable releases -- people who install SCT via `master` (i.e. developers).
> 
> There are definitely pros and cons to pinning vs. not pinning on `master`, though!
> 
> <details>
> <summary> Lots of thoughts! </summary>
> 
> - Leaving things pinned, as you suggest, minimizes the amount of urgent errors for developers when things change upstream. (We have had a lot of open issues that say something like "The tests are failing now due to an update!!")
> - However, in many cases, leaving a package pinned is just postponing dealing with the problems until some later time. 
>   - Upgrading Python versions often necessitates upgrading dependencies to be compatible with the new Python version. 
>   - The various [maintenance PRs](https://github.com/ivadomed/ivadomed/pull/1308) we have had to work on in `ivadomed` lately also necessitated us upgrading packages, too.
> - So, IMO, proactively pinning dependencies is a question of "do we want to deal with problems incrementally as they come up" vs. "do we want to deal with problems all at once, often when it's critically needed due to external pressure."
> - Personally, I find "deal with issues as they come up" to be a little easier to work with? It makes me feel on top of things well in advance, as we're upgrading immediately (vs. being forced to upgrade later on). It also tends to isolate any issues to specific libraries, while leaving things pinned might result in interactions between multiple outdated libraries.
> - There is also the benefit of automatic security upgrades / bugfixes that could theoretically mitigate future issues? But then again, there's also the risk of subtle behavior changes between versions, so maybe this point is a bit of a wash.
> 
> Also, some comments specific to `nnunet`:
> 
> - The rewrite seems to have been a one-off event, with [very understandable justification](https://github.com/MIC-DKFZ/nnUNet#what-happened-to-the-old-nnu-net). 
> - Also, during the rewrite, they opted for a new package name altogether (`nnunet` vs. `nnunetv2`). Meaning, pinning the version wouldn't have had any effect in that specific case.
> 
> </details>
> 
> Ultimately, I think that either pinning or not pinning is OK. But, just to stick with what SCT has been doing so far, I think I'm leaning towards keeping it unpinned. :)

PR #4383, opened 2024-02-27
Fix `-p sct_label_vertebrae` QC report when providing TotalSegmentator labels

### LINKED ISSUE #4379: Separate labels with different values are being merged together in the QC report for `-p sct_label_vertebrae`
Hello SCT team,

I'm reaching out to suggest an enhancement for the sct_qc that could greatly improve the visualization of multi-label segmentations.

When leveraging `-p sct_label_vertebrae` for multiple labeled segmentations in the sct_qc, there seems to be an issue where labels are merged together, resulting in a visual representation that lacks the necessary distinction between different segmented areas. This presents a challenge when evaluating the precision of vertebral labeling, where clear and distinct colors for each label are essential to accurately differentiate adjacent anatomical structures.

image

I propose the implementation of a feature that allows sct_qc to automatically assign distinct colors to each label in multi-label segmentations. Ideally, this could be enabled through a command like `sct_qc -i <nii_image> -s <nii_seg> -p multi_label -plane sagittal`, which would generate a color-coded overlay where each label is clearly distinguishable.

image

Thank you so much!

Best,

Yehuda

joshuacwnewton:
## Description

The existing QC report for `-p sct_label_vertebrae` assumes a single continuous set of labels with low values (e.g. `[2, 3, 4, 5, 6]`). However, if the provided labels go beyond 105, and/or  contain multiple distinct subsets of labels, then the generated color mapping will be incorrect, causing segmented labels to be seemingly merged together.

This PR provides two fixes: 

- A simple fix (to ensure that we properly assign a unique color to each label)
- A more complex fix (to try to visually distinguish the different label subgroups)

<details>
<summary> Comparison screenshots </summary>

Before:

image

After (simple fix):

image

After (complex fix):

image

</details>

## Discussion

There was some back and forth in the original issue, but in short: My complex "automatic" approach won't necessarily result in our typical colormaps for the cord and vertebrae (red for cord, rainbow for vertebrae). 

But, I'm not sure how to parse whether we even _have_ vertebrae/cord labels unless we can confidently assume that specific values have semantic meaning (e.g. 200 = cord, 201 = CSF, etc.).

So, for now, I've kept my implementation general (i.e. unspecific to hardcoded label values), such that the code can more easily be ported to a future QC revamp (e.g. if `-p sct_label_vertebrae` becomes something like `-type labels`). But, my approach can definitely be revised!

## Linked issues

Fixes #4379.

### DIFF
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -187,12 +187,11 @@ class QcImage:
 
     def label_vertebrae(self, mask, ax):
         """Draw vertebrae areas, then add text showing the vertebrae names"""
-        from matplotlib import colors
         img = np.rint(np.ma.masked_where(mask < 1, mask))
         labels = np.unique(img[np.where(~img.mask)]).astype(int)  # get available labels
         color_list = self._assign_label_colors_by_groups(labels)
         ax.imshow(img,
-                  cmap=colors.ListedColormap(color_list),
+                  cmap=color.ListedColormap(color_list),
                   interpolation=self.interpolation,
                   alpha=1,
                   aspect=float(self.aspect_mask))
@@ -209,12 +208,12 @@ class QcImage:
                     index = int(val)
                     if index in self._labels_regions.values():
                         # NB: We need to subtract `min` to convert the label value into an index for the color list
-                        color = color_list[index - labels.min()]
+                        label_color = color_list[index - labels.min()]
                         y, x = center_of_mass(np.where(data == val, data, 0))
                         # Draw text with a shadow
                         x += data.shape[1] / 25
                         label = list(self._labels_regions.keys())[list(self._labels_regions.values()).index(index)]
-                        label_text = ax.text(x, y, label, color=color, clip_on=True)
+                        label_text = ax.text(x, y, label, color=label_color, clip_on=True)
                         label_text.set_path_effects([path_effects.Stroke(linewidth=2, foreground='black'),
                                                      path_effects.Normal()])
 
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -108,7 +108,7 @@ class QcImage:
         return color_list
 
     def __init__(self, qc_report, interpolation, action_list, process, stretch_contrast=True,
-                 stretch_contrast_method='contrast_stretching', fps=None):
+                 stretch_contrast_method='contrast_stretching', fps=None, draw_text=True):
         """
         :param qc_report: QcReport: The QC report object
         :param interpolation: str: Type of interpolation used in matplotlib
@@ -131,6 +131,7 @@ class QcImage:
             raise ValueError("Unrecognized stretch_contrast_method: {}.".format(stretch_contrast_method),
                              "Try 'equalized' or 'contrast_stretching'")
         self._fps = fps
+        self._draw_text = draw_text
         self._centermass = None  # center of mass returned by slice.Axial.get_center()
 
     def listed_seg(self, mask, ax):
@@ -197,22 +198,25 @@ class QcImage:
                   aspect=float(self.aspect_mask))
         ax.get_xaxis().set_visible(False)
         ax.get_yaxis().set_visible(False)
-        a = [0.0]
-        data = mask
-        for index, val in np.ndenumerate(data):
-            if val not in a:
-                a.append(val)
-                index = int(val)
-                if index in self._labels_regions.values():
-                    # NB: We need to subtract `min` to convert the label value into an index for the color list
-                    color = color_list[index - labels.min()]
-                    y, x = center_of_mass(np.where(data == val, data, 0))
-                    # Draw text with a shadow
-                    x += data.shape[1] / 25
-                    label = list(self._labels_regions.keys())[list(self._labels_regions.values()).index(index)]
-                    label_text = ax.text(x, y, label, color=color, clip_on=True)
-                    label_text.set_path_effects([path_effects.Stroke(linewidth=2, foreground='black'),
-                                                 path_effects.Normal()])
+
+        # Use the existing colormap to draw colored text for any vertebral labels belonging to `self._labels_regions`
+        if self._draw_text:
+            a = [0.0]
+            data = mask
+            for index, val in np.ndenumerate(data):
+                if val not in a:
+                    a.append(val)
+                    index = int(val)
+                    if index in self._labels_regions.values():
+                        # NB: We need to subtract `min` to convert the label value into an index for the color list
+                        color = color_list[index - labels.min()]
+                        y, x = center_of_mass(np.where(data == val, data, 0))
+                        # Draw text with a shadow
+                        x += data.shape[1] / 25
+                        label = list(self._labels_regions.keys())[list(self._labels_regions.values()).index(index)]
+                        label_text = ax.text(x, y, label, color=color, clip_on=True)
+                        label_text.set_path_effects([path_effects.Stroke(linewidth=2, foreground='black'),
+                                                     path_effects.Normal()])
 
     def highlight_pmj(self, mask, ax):
         """Hook to show a rectangle where PMJ is on the slice"""
@@ -637,7 +641,7 @@ def get_json_data_from_path(path_json):
 
 
 def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None, path_qc=None, dataset=None,
-                subject=None, process=None, fps=None):
+                subject=None, process=None, fps=None, draw_text=True):
     """
     Generate a QC entry allowing to quickly review results. This function is the entry point and is called by SCT
     scripts (e.g. sct_propseg).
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/reports/qc.py
```diff
@@ -61,6 +62,50 @@ class QcImage:
...
 
+    def _assign_label_colors_by_groups(self, labels):
+        """
+        This function handles label colors when labels may not be in a single continuous group:
+
+            - Normal vertebral labels:  [2, 3, 4, 5, 6, ...]
+            - Edge case, TotalSegmentator labels: [31, 32, 33, 200, 201, 217, 218, 219]
+
+        We assume that the subgroups of labesl (1: [31, 32, 33, ...], 2: [200, 201], 3: [217, 218, 219, ...])
+        should each be assigned their own distinct colormap, as to
```
joshuacwnewton, line 73: Oops, I must have gotten distracted mid-sentence.

```suggestion
        should each be assigned their own distinct colormap, as to group them semantically.

**Comment by yw7** [comment]
Work great for me (: Thank you!!!!

image

About this, will it be possible to remove the labels?

> Another issue is the vertebrae names (C1, C2, etc.), will there be an option to disable those for general multi-label segmentation? or using `-p <some new option>` instead of `-p sct_label_vertebrae` to disable it?

File: spinalcordtoolbox/reports/qc.py [hunk 2]
```diff
@@ -57,10 +58,54 @@ class QcImage:
...
+        """
+        # Arrange colormaps for max contrast between colormaps, and max contrast between colors in colormaps
+        distinct_colormaps = ['Blues', 'Reds', 'Greens', 'Oranges', 'Purples']
+        colormap_sampling = [0.25, 0.5, 0.75, 0.5]  # light -> medium -> dark -> medium -> (repeat)
+
+        # Split labels into subgroups --> we split the groups wherever the difference between labels is > 1
+        start_end = [0, len(labels)]
+        for idx, (prev, curr) in enumerate(zip(labels, labels[1:]), start=1):
+            if curr - prev > 1:
+                start_end.insert(len(start_end)-1, idx)
```
mguaypaq, line 83: If you wait until after the loop to put `len(labels)` at the end of `start_end`, then this `start_end.insert` could be a slightly simpler `start_end.append`. (This is a super optional suggestion.)

spinalcordtoolbox/reports/qc.py
```diff
@@ -57,10 +58,54 @@ class QcImage:
...
+            n_colors = labels.max() - labels.min() + 1        # get label range from min label and max label
+            q, r = divmod(n_colors, len(self._labels_color))  # repeat the base list of colors to cover label range
+            color_list = (q * self._labels_color) + self._labels_color[:r]
+
+        # Handle the edge case: Multiple continuous groups
+        else:
+            # Initialize a list by repeating the color black (#000000) to fill in the gaps between colors.
+            # We do this because matplotlib applies colormaps by scaling both the data and the colormap to [0, 1].
+            # Without filling in the gaps between groups, the colormap would be scaled incorrectly relative to the data.
+            # ((Note that, if done right, the #000000 color should never be assigned to our label values.))
```
mguaypaq, line 97: Thanks for these clear explanatory comments :heart:

mguaypaq (approved):
The "After (complex fix)" screenshot is beautiful! Thanks for this amazing improvement.

spinalcordtoolbox/reports/qc.py
```diff
@@ -157,7 +203,8 @@ def label_vertebrae(self, mask, ax):
                 a.append(val)
                 index = int(val)
                 if index in self._labels_regions.values():
-                    color = self._labels_color[index]
+                    # NB: We need to subtract `min` to convert the label value into an index for the color list
+                    color = color_list[index - labels.min()]
                     y, x = center_of_mass(np.where(data == val, data, 0))
                     # Draw text with a shadow
                     x += data.shape[1] / 25
```
joshuacwnewton, line 210: Originally asked by @yw7:

> Work great for me (: Thank you!!!!
> 
> image
> 
> About this, will it be possible to remove the labels?

I think that the way you suggest (adding a different option to `-p`) would be very doable. :)
   -

joshuacwnewton: I was also thinking of perhaps some sort of automatic way of disabling the label text under certain circumstances. We currently add text for these labels:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e3d00c58cbe98e615e469c63edcd6693f3a0004d/spinalcordtoolbox/reports/qc.py#L42-L48

I was thinking of something like:

- If there are extra labels beyond these hardcoded labels, then don't add the text.
- If the text would overlap onto a different segment (e.g. the vertebrae segments), then don't add the text.

But, the more I think about it, the more I think that this wouldn't be foolproof (and may conflict with user desires). So, I think I agree with the suggestion of making it an explicit option.

spinalcordtoolbox/reports/qc.py
joshuacwnewton: I think I originally had it as you suggest? But, when I was looking at the "append" version, it felt... unclear? So, I rewrote it this way in the hopes of making the "`start_end`" variable behavior more clear, since it's a bit of a weird variable name to grok, hehe. 

(With the insert version, by default, we start with one full "group" with `start = 0` and `end = len(labels)`. Then, whenever we encounter a new "break" in the group, we insert that index, such that we have two groups `(0, i)` and `(i, len(labels))`.)

spinalcordtoolbox/reports/qc.py
```diff
@@ -143,8 +143,11 @@ def label_vertebrae(self, mask, ax):
         from matplotlib import colors
         img = np.rint(np.ma.masked_where(mask < 1, mask))
         labels = np.unique(img[np.where(~img.mask)]).astype(int)  # get available labels
+        n_colors_needed = labels.max() - labels.min() + 1         # get label range from min label and max label
+        q, r = divmod(n_colors_needed, len(self._labels_color))   # repeat the base list of colors to cover label range
+        color_list = (q * self._labels_color) + self._labels_color[:r]
```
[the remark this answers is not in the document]
joshuacwnewton: Thank you for the suggestion! For some reason, I always shy away from `itertools` and try to stick with basic Python patterns? I don't even know if I have a rational reason for this!! It's like `os.path` vs `Pathlib` to me -- for some reason I try to stick with `os.path`, even though `Pathlib` adds a lot of useful functionality. 

This is definitely a weird bias/blind spot I have. It's almost like... to my brain, `Pathlib`/`itertools` feel like adding a new dependency to my code? Sort of like when you're browsing StackOverflow and someone chimes in "I've created a simple little utility library for this niche usecase!" and my reaction is "I don't want to add this package to my dependencies just for a one-off usage." (I get that same feeling for `itertools`/`Pathlib`, even though they're both a part of the standard library!)

On reflection, I think I really should properly take the time to learn the different facets of the Python stdlib and utilize them more!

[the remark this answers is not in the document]
mguaypaq: I have very similar feelings about adding dependencies! For what it's worth, in my mind `itertools` has been in the Python standard library for a very long time, so I don't think twice about using it. But `Pathlib` is a much more recent addition, and still evolving (so for each function, I always check the minimum Python version of it), so I _do_ think before using it. But it's really quite nice, and keeps improving :sweat_smile: so my thought process tends to be, "in this context, _how much_ of `Pathlib` can I safely use?"

spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -47,6 +47,11 @@ def get_parser():
                         help='Plane of the output QC. Only relevant for -p sct_deepseg_lesion.',
                         choices=('axial', 'sagittal'),
                         required=False)
+    parser.add_argument('-exclude-label-text',
```
mguaypaq, line 50: I like to avoid negative options, because they tend to cause double negatives down the line, which are confusing (as you noted in your commit description). Could I suggest `-text-labels 0/1` instead, with a default of `1`? This would also be consistent with the rest of SCT, for example `sct_convert -squeeze` and `sct_compute_compression -normalize-hc` and the ubiquitous `sct_* -r` for removing temporary files.

mguaypaq (approved):
I'm marking this as "Approve" so that we can merge it once the last discussion is resolved.

spinalcordtoolbox/scripts/sct_qc.py
joshuacwnewton: Ahhh, good thinking! That sounds good. :)

spinalcordtoolbox/reports/qc.py
joshuacwnewton: > Work perfect for me!

Glad to hear!

> Also, not sure if for now and here.. but I thought you can add something like this to overwrite _labels_regions:
-text-labels L5:18,L4:19,L3:20,L2:21,L1:22,T12:23,T11:24,T10:25,T9:26,T8:27,T7:28,T6:29,T5:30,T4:31,T3:32,T2:33,T1:34,C7:35,C6:36,C5:37,C4:38,C3:39,C2:40,C1:41,S:92,SC:200,CSF:201,L5S:202,L4L5:203,L3L4:204,L2L3:205,L1L2:206,T12L1:207,T11T12:208,T10T11:209,T9T10:210,T8T9:211,T7T8:212,T6T7:213,T5T6:214,T4T5:215,T3T4:216,T2T3:217,T1T2:218,C7T1:219,C6C7:220,C5C6:221,C4C5:222,C3C4:223,C2C3:224

Just to clarify -- do you mean that you would like this feature? Or, do you mean that the name "-text-labels" is a bit confusing, since the name suggests that you might be able to pass the "text labels" that you want to be displayed?

Would something like "-show-text-labels" be clearer?

yw7: Actually Both. But I thought we can do both in one param so if it will be pass with 0 it will remove the labels and if pass with a dictionary it will overwrite existing labels.
Also we can leave it as is for now and open another issue for later to add functionality to this param to overwrite the default labels.

Issue #4382, opened 2024-02-23
Newly-updated `checkout@v4` action requires additional dependencies for (Debian 9/CentOS 7/WSL) runners


joshuacwnewton:
PR #4353 passed for our PR runners, but failed for some of the runners that only run on schedule/pushes to `master`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8024513196/job/21924760735

mguaypaq:
Given that Debian 9 (stretch) is [no longer under LTS](https://www.debian.org/releases/stretch/) as of July 2022, and only gets [paid support](https://wiki.debian.org/LTS/Extended) now, maybe it's time to let it go. The oldest release that's even listed on packages.debian.org is Debian 10 (buster), aka [oldoldstable](https://packages.debian.org/oldoldstable/).

As for CentOS 7, it seems to only have a few more months of maintenance updates (until end of June 2024, [according to Wikipedia](https://en.wikipedia.org/wiki/CentOS#End-of-support_schedule)). And it looks like it only ships with glibc-2.17, according to [this package list](http://mirror.centos.org/centos/7/os/x86_64/Packages/), whereas node20 seems to require glibc-2.28 or above. Maybe it's also time to let it go.

For WSL, the problem seems different, but I'm not sure what it is, because I don't see any error messages.

mguaypaq:
Slightly more info for WSL: I think it's unrelated to the node20 dependencies, because the error happens after the SCT installer has started running.

The last few lines of output from [this test run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8101090495/job/22142993227#step:7:182) are from conda:
```
Extracting : zstandard-0.19.0-py312h5eee18b_0.conda:  95%|█████████▍| 69/73 [00:27<00:01,  2.32it/s]
Extracting : zstd-1.5.5-hc292b87_0.conda:  96%|█████████▌| 70/73 [00:27<00:01,  2.32it/s]           
Extracting : _libgcc_mutex-0.1-main.conda:  97%|█████████▋| 71/73 [00:27<00:00,  2.32it/s]
Extracting : _openmp_mutex-5.1-1_gnu.conda:  99%|█████████▊| 72/73 [00:27<00:00,  2.32it/s]
                                                                                           
Installing base environment...
Downloading and Extracting Packages: ...working... done
Downloading and Extracting Packages: ...working... done
Preparing transaction: ...working... done
```
[...]

PR #4390, opened 2024-03-04
Debug WSL CI errors

joshuacwnewton:
See #4382.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -289,7 +289,7 @@ jobs:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
     # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
-    runs-on: windows-2022
+    runs-on: windows-2019
     defaults:
       run:
         shell: wsl-bash {0} # https://github.com/marketplace/actions/setup-wsl#default-shell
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -288,7 +288,7 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
```
joshuacwnewton, line 291: Hrm. That didn't quite go according to plan. I expected that uncommenting the `if` statements would cause the WSL runners to be triggered, but that didn't work here.

I suppose I can try manually triggering the run? :thinking:

joshuacwnewton: I manually ran the tests for both:

- `windows-2019` + `setup-wsl@v3`
- `windows-2022` + `setup-wsl@v2`

The former passes but the latter fails, so it looks like `windows-2022` is the culprit. 

At least this give us a very straightforward path to fixing out CI in the short-term. Though, at some point we'll need to think about how to support WSL testing if/when GH deprecates `windows-2019`.

PR #4395, opened 2024-03-07
Add CI workflow step to detect broken links in repository files

### LINKED ISSUE #4393: Broken link in `sct_register_to_template.py` help
The link in `sct_register_to_template.py` help points to a no longer existing page:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/44587d4cb293e04e635e17fe56137a0c4aa28f8f/spinalcordtoolbox/scripts/sct_register_to_template.py#L122

The `vertebral-labeling` section was moved out from `registration-to-template` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/5b9f2451f40f5e45a3c04d2a9826f34b452d19e8.

joshuacwnewton:
## Description

This PR adds a CI workflow to check the repository for broken links.

Then, this PR fixes each broken link so that the test passes:

image

## Linked Issues/PRs

Fixes #4393.

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -38,7 +38,7 @@ jobs:
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \( \! -name 'CHANGES.md' \) | 
           xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" | 
           sed 's/:/;/' | 
-          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
+          xargs -n 1 ./.github/workflows/check-url.sh
 
     - name: Summarize broken link check
       run: |
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -34,6 +34,7 @@ jobs:
 
     - name: Check for broken links
       run: | 
+        touch valid_urls.txt redirected_urls.txt invalid_urls.txt
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
           xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" | 
           grep -v 'drive.google.com' |    # Returns 401 (needs google auth)
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -31,3 +31,10 @@ jobs:
       uses: ludeeus/action-shellcheck@master
       with:
         severity: style
+
+    - name: Check for broken links
+      run: | 
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
+          xargs grep -Eo "(http|https)://[a-zA-Z0-9./?=_%:-]*" | 
+          sed 's/:/;/' | 
+          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -36,7 +36,7 @@ jobs:
       run: | 
         touch valid_urls.txt redirected_urls.txt invalid_urls.txt
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \( \! -name 'CHANGES.md' \) | 
-          xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" | 
+          xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" | 
           sed 's/:/;/' | 
           xargs -n 1 ./.github/workflows/check-url.sh
 
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -38,6 +38,7 @@ jobs:
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
           xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" | 
           grep -v 'drive.google.com' |    # Returns 401 (needs google auth)
+          grep -v 'pipeline-hemis' |      # Returns 403 (private GH repo)
           grep -v 'sciencedirect.com' |   # Returns 403
           grep -v 'wiley.com' |           # Returns 403
           grep -v 'sagepub.com' |         # Returns 403
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -38,7 +38,7 @@ jobs:
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \( \! -name 'CHANGES.md' \) | 
           xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" | 
           sed 's/:/;/' | 
-          xargs -n 1 ./.github/workflows/check-url.sh
+          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
 
     - name: Summarize broken link check
       run: |
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -35,19 +35,8 @@ jobs:
     - name: Check for broken links
       run: | 
         touch valid_urls.txt redirected_urls.txt invalid_urls.txt
-        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \( \! -name 'CHANGES.md' \) | 
           xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" | 
-          grep -v 'drive.google.com' |    # Returns 401 (needs google auth)
-          grep -v 'pipeline-hemis' |      # Returns 403 (private GH repo)
-          grep -v 'sciencedirect.com' |   # Returns 403
-          grep -v 'wiley.com' |           # Returns 403
-          grep -v 'sagepub.com' |         # Returns 403
-          grep -v 'ncbi.nlm.nih.gov' |    # Returns 403          
-          grep -v 'oxfordjournals.org' |  # Returns 403
-          grep -v 'docker.com' |          # Returns 403
-          grep -v 'ieeexplore.ieee.org' | # Returns 418
-          grep -v 'CHANGES.md' |          # Has too many GitHub links, triggers "Too many requests"
-          grep -v '%s' |                  # conf.py extlinks URL
           sed 's/:/;/' | 
           xargs -n 1 -P 8 ./.github/workflows/check-url.sh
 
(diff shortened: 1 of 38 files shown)


.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,10 @@ jobs:
       uses: ludeeus/action-shellcheck@master
       with:
         severity: style
+
+    - name: Check for broken links
+      run: | 
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
+          xargs grep -Eo "(http|https)://[a-zA-Z0-9./?=_%:-]*" | 
+          sed 's/:/;/' | 
+          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
```
joshuacwnewton, line 40: Neat, it works! Though, some observations:

- The grep pattern I used will include trailing periods, which are somewhat common for links that are included as part of a sentence.
- We should intentionally exclude the `CHANGES.md` file, since it has a _ton_ of GitHub PR links, and this will trigger an Error 429 - Too Many Requests.
- The output is really noisy when including links that are OK.

mguaypaq: For the noisy output: maybe the _numbers_ of good/bad/total links, and the _list_ of bad links (with locations?) would be a good amount of output?

joshuacwnewton: By numbers, do you mean a summary similar to pytest output?

image

mguaypaq: Exactly!

.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,32 @@ jobs:
...
+          grep -v 'drive.google.com' |    # Returns 401 (needs google auth)
+          grep -v 'pipeline-hemis' |      # Returns 403 (private GH repo)
+          grep -v 'sciencedirect.com' |   # Returns 403
+          grep -v 'wiley.com' |           # Returns 403
+          grep -v 'sagepub.com' |         # Returns 403
+          grep -v 'ncbi.nlm.nih.gov' |    # Returns 403          
+          grep -v 'oxfordjournals.org' |  # Returns 403
+          grep -v 'docker.com' |          # Returns 403
+          grep -v 'ieeexplore.ieee.org' | # Returns 418
+          grep -v 'CHANGES.md' |          # Has too many GitHub links, triggers "Too many requests"
```
joshuacwnewton, line 49: Note: I noticed that some common domains in our `studies.rst` page consistently return 403. I could update these to DOI links (which simply redirect), but given the time it would take to do that, I thought it would be okay to just skip checking links from these domains.

.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,32 @@ jobs:
       uses: ludeeus/action-shellcheck@master
       with:
         severity: style
+
+    - name: Check for broken links
+      run: | 
+        touch valid_urls.txt redirected_urls.txt invalid_urls.txt
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) | 
+          xargs grep -Eo "\b(https?|ftp):\/\/[-A-Za-z0-9+&@#\/%?=~_|$!:,.;]*[A-Za-z0-9+&@#\/%=~_|$]" |
```
mguaypaq, line 39: I don't think forward slash needs to be escaped? According to `man grep` on my machine:
```
   Basic vs Extended Regular Expressions
       In  basic  regular expressions the meta‐characters ?, +, {, |, (, and )
       lose their special meaning; instead use the  backslashed  versions  \?,
       \+, \{, \|, \(, and \).
```

```suggestion
          xargs grep -Eo "\b(https?|ftp)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" | 
```

mguaypaq: Separately: do we actually have any ftp links, and do we want to check them? Might be overkill.

.github/workflows/check-code-quality.yml
mguaypaq, line 49: This will match both links in the file `CHANGES.md`, and links where this appears in the URL, is this intended?

If it's just the file that we want to avoid, it could be filtered out at the `find` stage, with an extra `\( \! -name 'CHANGES.md' \)`.

.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,32 @@ jobs:
...
+          grep -v 'sciencedirect.com' |   # Returns 403
+          grep -v 'wiley.com' |           # Returns 403
+          grep -v 'sagepub.com' |         # Returns 403
+          grep -v 'ncbi.nlm.nih.gov' |    # Returns 403          
+          grep -v 'oxfordjournals.org' |  # Returns 403
+          grep -v 'docker.com' |          # Returns 403
+          grep -v 'ieeexplore.ieee.org' | # Returns 418
+          grep -v 'CHANGES.md' |          # Has too many GitHub links, triggers "Too many requests"
+          grep -v '%s' |                  # conf.py extlinks URL
+          sed 's/:/;/' |
```
mguaypaq, line 51: Ah, this doesn't have `g` at the end, so it only matches the first `:` delimiting the file name in the `grep` output, clever!

.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,32 @@ jobs:
...
+          grep -v 'wiley.com' |           # Returns 403
+          grep -v 'sagepub.com' |         # Returns 403
+          grep -v 'ncbi.nlm.nih.gov' |    # Returns 403          
+          grep -v 'oxfordjournals.org' |  # Returns 403
+          grep -v 'docker.com' |          # Returns 403
+          grep -v 'ieeexplore.ieee.org' | # Returns 418
+          grep -v 'CHANGES.md' |          # Has too many GitHub links, triggers "Too many requests"
+          grep -v '%s' |                  # conf.py extlinks URL
+          sed 's/:/;/' | 
+          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
```
mguaypaq, line 52: Is the `-P 8` really important for saving time? I ask because the manual page for `xargs -P` warns against running processes in parallel that write to `stdout` and/or to the same file, because the output can be garbled in that case.

But I can also imagine that some of the links are slow to fetch or time out, in which case it may be important

.github/workflows/check-code-quality.yml
```diff
@@ -31,3 +31,32 @@ jobs:
...
+
+    - name: Summarize broken link check
+      run: |
+        export NUM_OK=$(wc -l valid_urls.txt | cut -d " " -f 1)
+        export NUM_REDIRECT=$(wc -l redirected_urls.txt | cut -d " " -f 1)
+        export NUM_BAD=$(wc -l invalid_urls.txt | cut -d " " -f 1)
```
mguaypaq, line 58: I think the `export` here is unneeded? These variables are only used by the shell.

.github/workflows/check-url.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [ $# -ne 1 ]; then
+    exit 1;
+fi
+
+filename=$(cut -d ";" -f 1 <<< "$1")
+URL=$(cut -d ";" -f 2 <<< "$1")
+
+status_code=$(curl --write-out '%{http_code}' --silent  --output /dev/null "$URL")
```
mguaypaq, line 10: I don't see it documented in `man curl`, but I think it supports `--` for marking the end of options (so, the beginning of the url), which is theoretically safer.
```suggestion
status_code=$(curl --write-out '%{http_code}' --silent  --output /dev/null -- "$URL")
```

.github/workflows/check-url.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [[ $status_code -ge 200 && $status_code -le 299 ]];then
+    echo "($status_code) $URL ($filename)" >> valid_urls.txt
+    echo -e "$filename: \x1B[32m✅ OK status code: $status_code for domain $URL  \x1B[0m"
+    exit 0
+fi
+
+if [[ $status_code -ge 300 && $status_code -le 399 ]];then
```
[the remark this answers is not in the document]
mguaypaq: Yes, `curl -L` seems like a good idea.

.github/workflows/check-url.sh
mguaypaq, line 1: The script passes shellcheck :+1:

documentation/source/conf.py
```diff
@@ -96,7 +96,7 @@
 
 extlinks = {
     # e.g. :sct_tutorial_data:`data_template-registration.zip` gets expanded into:
-    # 'https://github.com/spinalcordtoolbox/sct_tutorial_data/releases/download/<tag>/data_template-registration.zip'
+    # 'github.com/spinalcordtoolbox/sct_tutorial_data/releases/download/<tag>/data_template-registration.zip'
```
mguaypaq, line 99: I'm curious about this change? Why?

testing/cli/test_cli_sct_register_multimodal.py
```diff
@@ -92,7 +92,7 @@ def test_sct_register_multimodal_with_softmask(tmp_path):
 
     NB: For 'gaussian', ANTs binaries can't handle softmasks natively, so SCT should be applying
         the mask directly to the image. Related links:
-            * https://github.com/ivadomed/pipeline-hemis/issues/3.
+            * github.com/ivadomed/pipeline-hemis/issues/3.
```
mguaypaq, line 95: Same question as above, why remove the https?

mguaypaq (approved):
This is great! And thanks for fixing so many broken links!

testing/cli/test_cli_sct_register_multimodal.py
joshuacwnewton: Originally, I modified this to also exclude it from the link-checking, since it's a private GH repo. But, then I switched over to excluding the actual `pipeline-hemis` string... and then forgot to reinstate the `https://`. :facepalm:

.github/workflows/check-code-quality.yml
joshuacwnewton: Hmm... without `-P 8`, the workflow takes [about 13 minutes](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8194952477), compared to [~2m with -P 8](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8193300941). 

On the one hand, it's still less than our slowest tests. Plus, the linting checks will fail first (meaning that we'll still get pretty rapid feedback for everything but the broken links). 

I don't know if I feel strongly either way? Hmmm... What do you prefer between sensible stdout output vs. ~6x faster runtime?

joshuacwnewton: I think I might be leaning towards keeping `-P 8`? I've written the test such that the results are [summarized in a separate step](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8206297070/job/22445069619?pr=4395#step:8:17), anyway. Meaning that most of the time, we won't actually be looking at the in-process stdout of the `-P 8` step?

mguaypaq: I guess my main concern isn't `stdout`, but the three files `redirected_urls.txt`, `invalid_urls.txt`, `valid_urls.txt`, which get written to concurrently without synchronization. But looking at it again, even if they get garbled, I don't think it will affect the pass/fail nature of the check, so I guess it doesn't matter too much. The manual entry for `find -P` suggests having each concurrent process write to its own file, then merging them, but maybe that's too heavyweight for this little check.

joshuacwnewton: > I guess my main concern isn't `stdout`, but the three files `redirected_urls.txt`, `invalid_urls.txt`, `valid_urls.txt`, which get written to concurrently without synchronization.

Ahhh, I understand now! Thank you for clarifying, and my apologies for the misunderstanding. I hadn't thought about that, but you're absolutely right.

I guess if we wanted to protect the output while also running things in parallel, we would need to look at something like GNU parallel's [`sem`](https://stackoverflow.com/questions/42075841/bash-append-file-from-multiple-thread), then?

> But looking at it again, even if they get garbled, I don't think it will affect the pass/fail nature of the check, so I guess it doesn't matter too much.

Yes, that's true. The check will only fail if `invalid_urls.txt` has any contents at all.

In fact, these text files may be a red herring? I think I mainly used them for local debugging purposes, but they don't really affect how the CI check runs, as you noticed.

spinalcordtoolbox/scripts/sct_dmri_denoise_patch2self.py
joshuacwnewton, line 1: Oh wow! The newly-added link checking workflow has already [caught a new broken link](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8254667228/job/22579715572?pr=4395) in the span of the development of this PR:

```
 (404) http://medrxiv.org/lookup/doi/10.1101/2021.10.04.21264389
 (./spinalcordtoolbox/scripts/sct_dmri_denoise_patch2self.py)
```

Indeed, when I open said link in a personal browser, I see:

image

Showing it's use already! :tada:

joshuacwnewton: The medrxiv link stems from this link: https://doi.org/10.1101/2021.10.04.21264389, which redirects then 404s.

Here is a non-404ing link to the same publication: https://www.medrxiv.org/content/10.1101/2021.10.04.21264389v2. Note, though, that this very page contains the DOI url that 404s! Fascinating! I wonder if something's up with medrxiv right now.

joshuacwnewton: Link replaced!

valosekj: > The medrxiv link stems from this link: https://doi.org/10.1101/2021.10.04.21264389, which redirects then 404s.

Interesting! https://doi.org/10.1101/2021.10.04.21264389 now works for me!

I'm mentioning this because we also have other papers or medrxiv, such as SCIseg (https://doi.org/10.1101/2024.01.03.24300794).

Issue #4374, opened 2024-02-15
Consider disabling duplicate tests for merged PRs on `master`


joshuacwnewton:
When merging a few features in quick succession, new CI runs will be triggered on `master` for the merged PR, even if those exact same tests were just completed on the PR itself.

Because we have a limit for concurrent runners, these duplicate tests will delay future tests from being run, which can slow down development. (I'm currently waiting for the changelog PR to re-generate because it's being held up by `master` CI runs.)

(We should still keep the cron jobs, though! Those are useful for catching upstream changes.)

joshuacwnewton:
This felt like an annoyance at the time, but in hindsight, I'm not sure this is really all that big of a deal. I don't mind the redundancy?

Issue #4034, opened 2023-02-09
Update GitHub Actions workflow files to support Merge Queues


joshuacwnewton:
GitHub has released their new Merge Queue feature: https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/

Earlier today, I tried enabling this feature, and we tried enqueuing some open PRs. However, the required checks never fired. 

Digging further, I found https://github.com/orgs/community/discussions/46757#discussioncomment-4908050, which suggests that workflow files must be updated with new trigger conditions.

> If your CI workflow currently only uses `on.pull_request`, you will need to update it so that it gets triggered when the queue creates temporary branches (that include the changes from one or more queued pull requests).
> 
> When it's working correctly ( smile ), the recommendation is to use `on.merge_group`:
> 
> ```yaml
> on:
>   pull_request:
>     branches: [ main ]
>   merge_group:
> ```
> 
> But, as you've probably seen in this thread, there is an active issue, so your safest option right now is to update your workflow to use `on.push`:
> 
> ```yaml
> on:
>   pull_request:
>     branches: [ main ]
>   push:
>     branches:
>       - gh-readonly-queue/main/*
> ```

So, we should implement the above changes, while monitoring the discussion thread for updates to see when `merge_group` functionality begins working.

joshuacwnewton:
The more I read about this feature, the less confident I am about its stability. So, for now, I'm going to remove this issue from the 6.0 milestone, just to give it a little more time to mature (i.e. leave public beta?).

mguaypaq:
FYI, it looks like [merge queues left public beta in July 2023](https://github.blog/changelog/2023-07-12-pull-request-merge-queue-is-now-generally-available/), if we ever feel like trying it again.

PR #4402, opened 2024-03-14
Reduce concurrent macOS runners on PRs from 6 to 2 (and remove push-to-master triggers)

### LINKED ISSUE #4400: Limit of 5 concurrent macOS runners is causing development delays due to queued runners
In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4353, we expanded our macOS testing to include both 12 and 14 in our test suite. We did this because `macos-12` is an Intel runner, while `macos-14` is an M1 runner, and thus theoretically running both would provide better coverage.

However, what I didn't realize at the time is that GitHub limits concurrent macOS runners (of any kind) to [5 on free plans](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits). (For some reason, I was under the impression that macOS 12 and macOS 14 would be treated as separate, rather than drawing from the same pool of concurrent runners).

Combined with https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4374, this really creates a bottleneck, since we fill up our macOS quota real quick. Merging a pull request, then immediately syncing another open PR with master will automatically fill up the queue and then some. Note the 3 separate "Queued" runs after 10 minutes:

image

I've found myself manually cancelling runs that I suspect won't reveal anything new, just to free up macOS runners for the jobs I _do_ care about.

It doesn't look like we can pay our way out of this limitation, either -- "Pro" and "Team" plans are limited to 5 concurrent macOS runners as well. So, we need to make sure we're careful about how we're allocating our 5 macOS runners at any given time.

----

On PRs, pushing a new commit will trigger 6 macOS runners (2 each for both Tests, batch_processing.sh, and Create Release), already exceeding the quota.

So, some possible suggestions:

1. Don't test both macOS 12 and 14 on PRs. 
   - On reflection, given that SCT requires Rosetta, I'm not sure that regularly testing on both Intel and Apple Silicon Macs actually provides much benefit. SCT's environment is sandboxed, so, apart from some obscure Rosetta-specific bugs, the testing will largely be the same, I think? Meaning, we should be okay to save, say, the macOS 12 test for `master` runs.
   - Not to mention, what will happen when GitHub deprecates the `macos-12` runners, anyway? Will macOS runners be Apple Silicon only? In such a future, we won't need to run two sets of runners anyway.
3. Somehow smartly combine existing workflows such that some tests are executed sequentially on a single macOS runner. This might result in a longer time spent on a single runner, but would more effectively use our quota for concurrent runners.
   - Probably more trouble than it's worth? I imagine it's just as good to have fast concurrent runners that free up quickly?
5. Speed up duplicated CI runs and/or reduce duplication of effort between "Create Release" runs and "Tests" runs
   - e.g. For the "Create Release" runs, only test the installation + `sct_check_dependencies` (rather than running the full test suite) on PRs, since the "Create Release" runs are mainly intended to test weird package/installation quirks that stem from freezing our requirements?
   - e.g. For the "Create Release" runs, only run the test suite if the installed requirements actually differ from the expected requirements (for the aforementioned reasons). Maybe we can use some sort of caching here, since we're already doing the work of saving the `requirements-freeze.txt` artifact between runs?
6. Skip running tests on `master` immediately after merge, and leave `master` tests for cron jobs only, to minimize the bottleneck stemming from cascades of merge/syncs.
7. ~~Reschedule the cron jobs to run at a time that's unlikely to clash with normal development. (I think right now they run midday, but they could be changed to run at, say, 6am/7am EST so that we get feedback when we clock in.~~ They already run at 7AM ET.

joshuacwnewton:
## Description

This PR solves the queuing issues we've faced due to the limit of 5 concurrent macOS runners.

Here, I've added the lowest-hanging-fruit suggestions from the issue description of https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4400. I've tried to preserve the highest-impact CI workflows (tests.yml, test-batch-processing.yml) while relegating the lower-impact CI workflows (create-release.yml, macos-14) to cron-only. I've also removed the trigger for merges/pushes to master. 

(See extended commit messages for further justification of each change.)

With these changes, we can now:

- Update PRs twice concurrently without running into queuing issues. (2x2 < 5)
- Merge and update PRs without running into queuing issues.

In the short term, this should allow us to quickly work through our backlog of approved PRs. (This need is also why I've mainly focused on uncontroversial, easy-to-review changes.) Long term, however, we may want to consider a more sophisticated approach to testing the create-release.yml workflow.

## Linked issues

Fixes #4400.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4374.
Maybe even addresses https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4034, since now that push-to-master triggers are gone, trains of PR merges are less of a hassle?

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -56,7 +56,7 @@ jobs:
     needs: generate-requirements-freeze
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-12, windows-2022]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-12, macos-14, windows-2019, windows-2022]
     runs-on: ${{ matrix.os }}
     # NB: Setting `bash` as the default shell is a bit of a hack workaround for Windows to allow it to run .sh scripts.
     #     Normally, we would ask the user to install a Windows-compatible bash via either Cygwin or Git for Windows.
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -6,9 +6,6 @@ on:
       milestone_title:
         description: 'Milestone title (this release)'
         required: true
-  push:
-    branches:
-      - master
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -6,9 +6,6 @@ on:
       milestone_title:
         description: 'Milestone title (this release)'
         required: true
-  pull_request:
-    branches:
-      - '*'
   schedule:
     # https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onschedule
     # > Scheduled workflows run on the latest commit on the default or base branch
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -62,7 +62,7 @@ jobs:
     needs: generate-requirements-freeze
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-12, macos-14, windows-2022]
+        os: [ubuntu-20.04, macos-12, windows-2022]
     runs-on: ${{ matrix.os }}
     # NB: Setting `bash` as the default shell is a bit of a hack workaround for Windows to allow it to run .sh scripts.
     #     Normally, we would ask the user to install a Windows-compatible bash via either Cygwin or Git for Windows.
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -27,7 +27,7 @@ jobs:
     name: Test batch_processing.sh
     strategy:
       matrix:
-        os: [ ubuntu-20.04, macos-12, macos-14, windows-2022 ]
+        os: [ ubuntu-20.04, macos-12, windows-2022 ]
     runs-on: ${{ matrix.os }}
     # NB: Setting `bash` as the default shell is a bit of a hack workaround for Windows to allow it to run .sh scripts.
     #     Normally, we would ask the user to install a Windows-compatible bash via either Cygwin or Git for Windows.
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -2,10 +2,6 @@ name: Test batch_processing.sh
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - master
-      - release
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ on:
 # - you cannot refer to the matrix parameters in a job's `if:`, so it is impossible to decide whether to run a job or not; it is possible to do so in a *step*'s `if:`. See https://github.community/t/conditional-matrices/17206/2.
 #   ideally we would have a build matrix with all the platforms covered
 #   ```
-#   if: ${{ !matrix.nightly || (github.event_name == 'schedule' || github.event_name == 'push') }}
+#   if: ${{ !matrix.nightly || (github.event_name == 'schedule') }}
 #   ```
 #   but it's not allowed; that second half is legal but `matrix.nightly` is not.
 # - there's no way to mix the docker-based platforms with the windows ones in the same build matrix anyway, because specifying `container:`, even with an empty/undefined string, tries blindly to use the docker scripts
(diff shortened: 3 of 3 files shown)


.github/workflows/create-release.yml
```diff
@@ -6,12 +6,6 @@ on:
       milestone_title:
         description: 'Milestone title (this release)'
         required: true
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - '*'
```
joshuacwnewton, line 14: There is perhaps a more nuanced approach we can take that would allow us to run the create-release.yml workflow on PRs without using additional macOS runners. For example, we could choose to only run the workflow if `requirements.txt` changes. We could even compare the `requirements-freeze.txt` artifact to a cached version, and only run the test suite if this file has changed. But, given the care required to implement such a solution, and given the need to unblock PRs ASAP, I figure we could try this in a follow-up PR.

mguaypaq: I think it's fine if we only run the create-release workflow as a daily cron job (or on-demand, if it's in the days leading up to a release).

.github/workflows/test-batch-processing.yml
mguaypaq, line 1: Same as for create-release, I think it's fine to run test-batch-processing only on a daily schedule.

.github/workflows/tests.yml
```diff
@@ -2,9 +2,6 @@ name: Tests
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - master
```
mguaypaq, line 7: I agree with the logic here: if we just ran the test suite on an up-to-date PR before merging into master, there's almost certainly no need to re-run it on master. The very unlikely exceptions will get caught by the daily run anyway.

However, should we update all the `if: ${{ ... || github.event_name == 'push') }}` clauses to remove the references to push events, since they no longer exist?

.github/workflows/tests.yml
```diff
@@ -262,6 +259,10 @@ jobs:
 
   macos-14:
     name: macOS 14.0 (Sonoma)
+    # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
+    #if: ${{ env.NIGHTLY }}
+    # in the meantime, copy-paste this:
+    if: ${{ github.event_name == 'schedule' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
```
mguaypaq, line 265: Just to double-check my understanding: the commit description for this change says that it's **disabling** the macos-14 runner, but these lines actually **enable** the macos-14 runner, right? Specifically, it becomes enabled for pull_request events?

I think it's still true that merging this PR would reduce the number of macos runners for pull_request events from 3 to 2.

mguaypaq (approved):
Thanks for this simplification of the CI runs, it should solve the immediate bottleneck problems.

.github/workflows/tests.yml
joshuacwnewton: Ah! Sorry for the confusion... 

Without this line, the macos-14 runner would run unconditionally. Whereas, with this line, the macos-14 runner only runs for the triggers listed. And, since PRs aren't listed, it gets disabled for PRs.

.github/workflows/tests.yml
joshuacwnewton: Ah, of course! Good catch.

.github/workflows/test-batch-processing.yml
joshuacwnewton: The main reason I kept the batch_processing test on PRs is because in the past, it's helped us to know when PR changes would affect the results of our processing pipelines.

joshuacwnewton: That said, batch_processing.sh still isn't the best at detecting cross-platform numerical discrepancies, due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3194.

.github/workflows/tests.yml
joshuacwnewton: It's almost like a double-negative? It might be a little more clear if it was instead worded like this:

```
if ${{ github.event_name != 'pull_request' }}
```

That way it's clearer that it runs on every trigger _but_ PRs.

mguaypaq: Oh wow, I read the diff backwards, sorry! For some reason I was convinced that these lines got removed instead of added :facepalm:

PR #4409, opened 2024-03-19
Add improvements to CSA calculation for GM/WM masks (`-angle-corr-centerline`, `float32` precision fix, doc warnings)

### LINKED ISSUE #4367: Angle correction is inconsistent when running `sct_process_segmentation` across GM/WM/SC segmentations
Dear all,

I am currently analysing mouse spinal cord areas in different segments and extracting values from the sct_process_segmentation function as described here: https://github.com/ivadomed/model_seg_mouse-sc_wm-gm_t1/issues/45

However, the sum of the GM and WM is not equivalent to the area of the total spinal cord (called sc in the _Metrics measurement based on model output #45_ mentioned above).

Does someone has an idea for this issue?

Thank you very much in advance for your time and help,
Kind regards,
Benjamin F. Combes

### LINKED ISSUE #4378: Multiplying by `cos` angle correction factors results in slightly different values (~1e-3) compared to `-angle-corr 1`
While investigating #4367, I tried validating the suggestion by Julien (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4367#issuecomment-1948521324) by trying to apply angle correction 2 different ways (on just the spinal cord segmentation, i.e. no GM/WM involved in scope of this issue):

1. Automatically, using SCT's usual `-angle-corr 1` behavior.
2. Manually, by following @jcohenadad's [suggestion](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4367#issuecomment-1948521324) of multiplying by `cos(angle_RP/180*PI())` and `cos(angle_AP/180*PI())` using Google Sheets' `COS` and `PI` functions.

Green is the output of `-angle-corr 0`, Orange is manual multiplication of angle values, Blue is `-angle-corr 1`:

image

Both the orange and blue values ~_should_ be the same (since we're just multiplying the uncorrected area by `cos(angle)*cos(angle)` in both cases)~. However, they are not equal.

I will try to investigate why this happens and make sure we are outputting the "correct" angle-corrected values.

joshuacwnewton:
## Description

This PR is a follow up to my work helping @BFCombes on https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4367. During my investigation, I discovered a few small issues and areas of improvement. And, I wanted to make sure those observations were addressed before moving on, lest the context of the issue be lost to time.

This PR contains 3 main changes:

- A small precision fix for float32 zooms, which removes spurious digits for uncorrected CSA values.
    - We use [`np.spacing`](https://numpy.org/doc/stable/reference/generated/numpy.spacing.html) to figure out what decimal place we can reliably round to -- we can't just rely on `float32` resolution because there are some complexities there. I found [this SO thread](https://stackoverflow.com/questions/32465481/what-exactly-is-the-resolution-parameter-of-numpy-float) to be a helpful resource on the topic.
- New warnings in the tutorials + help description, communicating a lack of precision for CSA values.
- A new option `-angle-corr-centerline`, which enables angle correction to be performed for irregularly-shaped segmentations such as GM/WM masks. This option is then patched into the tutorials, to avoid needing to resort to `-angle-corr 0`.

## Linked issues

Fixes #4367.
Fixes #4378.


sandrinebedard (commented):
Maybe we should also specify that the curvature of the spinal cord affects this (if lrge FOV, it is impossible to be orthoginal through the entire FOV)

documentation/source/user_section/tutorials/gray-matter-segmentation/gm-wm-metric-computation/csa-gm-wm.rst
```diff
@@ -3,20 +3,39 @@ Using binary masks to compute CSA for gray and white matter
...
-   - **Note:** Turning off angle correction is only safe to do if you know that your axial slices were acquired roughly orthogonal to the cord.
+   - ``-angle-corr-centerline``: Normally, angle correction will be applied during ``sct_process_segmentation`` to account for scans where the spinal cord is positioned at an angle with respect to the superior-inferior axis. While this works well when ``-i`` is a full spinal cord segmentation, in this case we are instead providing GM/WM segmentations to ``-i``. The irregular cross-sectional shape of these segmentations can have a negative effect on so the estimation of the cord centerline, which in turn may cause the estimated angle to be incorrect. So, here we explicitly provide a full spinal cord segmentation for angle correction purposes, to ensure accurate and consistent angle correction.
+   - **Note:** Alternatively, if you don't have a full-cord segmentation, you could turn off angle correction entirely using ``-angle-corr 0``, but this will only provide accurate results if you know that your axial slices were acquired roughly orthogonal to the cord.
```
sandrinebedard, line 38: Maybe we should also specify that the curvature of the spinal cord affects this (if lrge FOV, it is impossible to be orthoginal through the entire FOV)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -181,6 +186,14 @@ def get_parser():
              "centerline and the axial plane. If the cord is already quasi-orthogonal to the slab, you can set "
              "-angle-corr to 0."
     )
+    optional.add_argument(
+        '-angle-corr-centerline',
+        metavar=Metavar.str,
+        help="Centerline to be used for angle correction. This is optional; if not provided, the centerline will be "
```
sandrinebedard, line 192: Should it be the spinal cord centerline or spinal cord segmentation file that we should provide here? Can it be either?

sandrinebedard (commented):
Thank you Joshua! I had minor comments

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: It can be either! Good point, yes, that isn't clear. :)

batch_processing.sh
```diff
@@ -117,8 +117,8 @@ sct_warp_template -d t2s.nii.gz -w warp_template2t2s.nii.gz
 # Subtract GM segmentation from cord segmentation to obtain WM segmentation
 sct_maths -i t2s_seg.nii.gz -sub t2s_gmseg.nii.gz -o t2s_wmseg.nii.gz
 # Compute cross-sectional area of the gray and white matter between C2 and C5
-sct_process_segmentation -i t2s_wmseg.nii.gz -vert 2:5 -perlevel 1 -o csa_wm.csv
-sct_process_segmentation -i t2s_gmseg.nii.gz -vert 2:5 -perlevel 1 -o csa_gm.csv
+sct_process_segmentation -i t2s_wmseg.nii.gz -vert 2:5 -perlevel 1 -o csa_wm.csv -angle-corr-centerline t2s_seg.nii.gz
+sct_process_segmentation -i t2s_gmseg.nii.gz -vert 2:5 -perlevel 1 -o csa_gm.csv -angle-corr-centerline t2s_seg.nii.gz
```
joshuacwnewton, line 121: Naturally, adding angle correction here changed the values for batch_processing.sh:

```
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[t2s/csa_gm.csv-3-MEAN(area)] - assert 12.515230379832953 == 12.487533885581323 ± 1.2e-05
  
  comparison failed
  Obtained: 12.515230379832953
  Expected: 12.487533885581323 ± 1.2e-05
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[t2s/csa_wm.csv-3-MEAN(area)] - assert 64.93314479093739 == 64.93464311280606 ± 6.5e-05
  
  comparison failed
  Obtained: 64.93314479093739
  Expected: 64.93464311280606 ± 6.5e-05
```

spinalcordtoolbox/image.py
```diff
@@ -973,7 +973,17 @@ def get_dimension(im_file, verbose=1):
...
+    #     as rounded 64-bit float values, according to the spacing of the value.
+    zooms = []
+    for zoom_f32 in im_file.header.get_zooms():
+        # extract the exponent from the spacing used by the float32 values (i.e. 1e-9 -> '-9')
+        spacing = np.spacing(zoom_f32)
+        spacing_exp = np.floor(np.log10(np.abs(spacing))).astype(int)
+        # convert to python float, then round to remove any spurious digits
+        zoom_f64 = float(zoom_f32)
+        zoom_f64_rounded = round(zoom_f64, abs(spacing_exp) - 1)
+        zooms.append(zoom_f64_rounded)
```
joshuacwnewton, line 986: This change had a very tiny effect on the resampling test values, too:

```
_ test_sct_resample_output_has_expected_dimensions[dmri/dmri.nii.gz--f-0.5x0.5x1-expected_dim1] _

AssertionError: assert 1.682692 == 1.6826923 ± 1.0e-07
  
  comparison failed
  Obtained: 1.682692
  Expected: 1.6826923 ± 1.0e-07
```

I wonder if this old value had a spurious significant figure? I can double check to see exactly how it was originally computed.

spinalcordtoolbox/image.py
```diff
@@ -973,7 +973,17 @@ def get_dimension(im_file, verbose=1):
...
+    # NB: Nibabel stores dim info as float32. Arithmetic using float32 values can cause issues. So, store the values
+    #     as rounded 64-bit float values, according to the spacing of the value.
+    zooms = []
+    for zoom_f32 in im_file.header.get_zooms():
+        # extract the exponent from the spacing used by the float32 values (i.e. 1e-9 -> '-9')
+        spacing = np.spacing(zoom_f32)
+        spacing_exp = np.floor(np.log10(np.abs(spacing))).astype(int)
+        # convert to python float, then round to remove any spurious digits
+        zoom_f64 = float(zoom_f32)
+        zoom_f64_rounded = round(zoom_f64, abs(spacing_exp) - 1)
```
mguaypaq, line 985: The `abs(spacing_exp)` should actually be `-spacing_exp`. You can check this with an example that has a positive `e` exponent:
- `zoom_f32 = np.float32(1e+30)`
- then `spacing_exp == 22`
- the incorrect result is `round(float(zoom_f32), 21) == 1.0000000150474662e+30`
- the correct result is `round(float(zoom_f32), -23) == 1e+30`
- and this isn't off-by-one, because `round(float(zoom_f32)), -22) == 1.00000002e+30`.

(If you're wondering what rounding to a negative number of digits means: `round(12345, -2) == 12300`)
```suggestion
        zoom_f64_rounded = round(zoom_f64, -spacing_exp - 1)
```

spinalcordtoolbox/process_seg.py
```diff
@@ -22,7 +22,8 @@
 NEAR_ZERO_THRESHOLD = 1e-6
 
 
-def compute_shape(segmentation, angle_correction=True, param_centerline=None, verbose=1, remove_temp_files=1):
+def compute_shape(segmentation, angle_correction=True, centerline_path=None, param_centerline=None,
+                  verbose=1, remove_temp_files=1):
```
mguaypaq, line 26: The new `centerline_path` argument should also be added to the docstring below.

documentation/source/user_section/tutorials/gray-matter-segmentation/gm-wm-metric-computation/csa-gm-wm.rst
```diff
@@ -3,20 +3,39 @@ Using binary masks to compute CSA for gray and white matter
...
-   - **Note:** Turning off angle correction is only safe to do if you know that your axial slices were acquired roughly orthogonal to the cord.
+   - ``-angle-corr-centerline``: Normally, angle correction will be applied during ``sct_process_segmentation`` to account for scans where the spinal cord is positioned at an angle with respect to the superior-inferior axis. While this works well when ``-i`` is a full spinal cord segmentation, in this case we are instead providing GM/WM segmentations to ``-i``. The irregular cross-sectional shape of these segmentations can have a negative effect on so the estimation of the cord centerline, which in turn may cause the estimated angle to be incorrect. So, here we explicitly provide a full spinal cord segmentation for angle correction purposes, to ensure accurate and consistent angle correction.
```
mguaypaq, line 37: Tiny typo:
```suggestion
   - ``-angle-corr-centerline``: Normally, angle correction will be applied during ``sct_process_segmentation`` to account for scans where the spinal cord is positioned at an angle with respect to the superior-inferior axis. While this works well when ``-i`` is a full spinal cord segmentation, in this case we are instead providing GM/WM segmentations to ``-i``. The irregular cross-sectional shape of these segmentations can have a negative effect on the estimation of the cord centerline, which in turn may cause the estimated angle to be incorrect. So, here we explicitly provide a full spinal cord segmentation for angle correction purposes, to ensure accurate and consistent angle correction.
```

mguaypaq (approved):
Thanks for these precision fixes, and for the good documentation of what's going on.

For the tutorial changes, should those be merged later, once SCT v6.3 is (nearly) out?

joshuacwnewton:
> For the tutorial changes, should those be merged later, once SCT v6.3 is (nearly) out?

Ahhh, yes! Absolutely, good call.

spinalcordtoolbox/image.py
joshuacwnewton: Ahhh, good catch! 

I think the reason I got mixed around here is that, because zooms are typically only ever in fractions of mm, I've only ever seen negative exponents? (It would be a bit strange to have the resolution be in meters/kilometers/etc, hehe.)

But, I appreciate the thoroughness and correctness -- you never know what sort of inputs we could get!

batch_processing.sh
joshuacwnewton: Currently running `batch_processing.sh` locally -- will update once it's finished.

joshuacwnewton: Done!

PR #4414, opened 2024-03-25
Fix detection of PSIR images for QC resampling to avoid unnecessary thresholding

### LINKED ISSUE #4389: Image contrast not appropriately set for some contrast
### Description

Here is how a QC report with PSIR looks like:

ezgif-4-3ce917b49a

Context/data: https://github.com/neuropoly/data-management/issues/303

joshuacwnewton:
## Description

The primary goal of this PR is to avoid treating PSIR anat images as labeled segmentation images when resampling for QC reports. Segmentations are thresholded after resampling to ensure that their background remains transparent. But, PSIR images can have positive _and_ negative values, so thresholding will make half the image dark:

Before (thresholded as seg image):

image

After (no thresholding as anat image):

image

> Note: There is less of a risk of the inverse issue where label images treated as anat images, since `sct_qc -p sct_label_vertebrae` doesn't perform resampling. (Though, `check_image_kind` is also used for picking the colormap to use for `fsleyes` commands, so we do still need to consider labeled segmentations.)

## Linked issues

Fixes #4389.

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1854,28 +1854,30 @@ def check_image_kind(img):
 
         - 'seg': Binary segmentation (0/1)
         - 'softseg': Nonbinary segmentation in the range [0, 1], where 0 and 1 are the majority of values
-        - 'labeled_seg': Nonbinary, whole values (0, 1, 2...) or (0.0, 1.0, 2.0...) only
-        - 'im': Any other float-valued image where 0 and 1 are not the majority of values.
+        - 'labeled_seg': Nonbinary, whole values (0, 1, 2...) where 0 is the majority of values
+        - 'im': Any other image where 0 is not the majority of values.
 
     Useful for determining A) which colormap should be applied to an image, or
                            B) which interpolation can be safely used on a given image
 
     Output strings should be compatible with display_viewer_syntax() function.
     """
-    # Check if image is a segmentation (binary or soft) by making sure:
-    # - 0/1 are the two most common voxel values
-    # - 0/1 account for >95% of voxels (to allow for some soft voxels)
+    # Count the unique values in the image
     unique, counts = np.unique(np.round(img.data, decimals=1), return_counts=True)
     unique, counts = unique[np.argsort(counts)[::-1]], counts[np.argsort(counts)[::-1]]  # Sort by counts
+    # This heuristic helps to detect binary and soft segmentations
     binary_most_common = set(unique[0:2].astype(float)) == {0.0, 1.0}
     binary_percentage = np.sum(counts[0:2]) / np.sum(counts)
+    # This heuristic helps to distinguish between PSIR images and label images (2-10% zero vs. 99% zero)
     is_whole_only = np.equal(np.mod(unique, 1), 0).all()
+    zero_most_common = float(unique[0]) == 0.0
+    zero_percentage = np.sum(counts[0]) / np.sum(counts)
     if binary_most_common:
         if binary_percentage == 1.0:
             return 'seg'
         elif binary_percentage > 0.95:
             return 'softseg'
-    if is_whole_only:
+    elif is_whole_only and zero_most_common and zero_percentage > 0.50:
         return 'seg-labeled'
     else:
         return 'anat'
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/image.py
```diff
@@ -1854,28 +1854,30 @@ def check_image_kind(img):
 
         - 'seg': Binary segmentation (0/1)
         - 'softseg': Nonbinary segmentation in the range [0, 1], where 0 and 1 are the majority of values
-        - 'labeled_seg': Nonbinary, whole values (0, 1, 2...) or (0.0, 1.0, 2.0...) only
-        - 'im': Any other float-valued image where 0 and 1 are not the majority of values.
+        - 'labeled_seg': Nonbinary, whole values (0, 1, 2...) where 0 is the majority of values
+        - 'im': Any other image where 0 is not the majority of values.
```
mguaypaq, line 1858: If you want to cover all the cases, I don't think you can say that the "default" case of `anat` has any guarantees about majority values. (Also, it's `anat` rather than `im`).
```suggestion
        - 'anat': Any other image.
```

spinalcordtoolbox/image.py
```diff
@@ -1854,28 +1854,30 @@ def check_image_kind(img):
...
     is_whole_only = np.equal(np.mod(unique, 1), 0).all()
+    zero_most_common = float(unique[0]) == 0.0
+    zero_percentage = np.sum(counts[0]) / np.sum(counts)
     if binary_most_common:
         if binary_percentage == 1.0:
             return 'seg'
         elif binary_percentage > 0.95:
             return 'softseg'
-    if is_whole_only:
+    elif is_whole_only and zero_most_common and zero_percentage > 0.50:
```
mguaypaq, line 1880: The more I think about it, the more I'm convinced that this should be an `if`, rather than `elif`. The heuristics we're using have complicated logical relationships, and we can't hope for a neat `if/elif/else` separation. It's more like a decision tree.

Also, `zero_percentage > 0.50` should imply `zero_most_common == True`, so we don't have to compute `zero_most_common` at all.

mguaypaq (changes requested):
I think you identified a good set of heuristics, but the code doesn't currently reflect the intention.

spinalcordtoolbox/image.py
joshuacwnewton: Sorry about the unnecessary change! I had looked at the second `if` and thought "huh, this is strange, why had I originally written it this way?" and changed it. But, it looks like you've answered the question for me.

spinalcordtoolbox/image.py
```diff
@@ -1850,32 +1850,36 @@ def generate_stitched_qc_images(ims_in: Sequence[Image], im_out: Image) -> Tuple
...
 def check_image_kind(img):
     """
-    Identify the image as one of 4 image types:
+    Identify the image as one of 4 image types (represented as one of 4 strings):
 
         - 'seg': Binary segmentation (0/1)
         - 'softseg': Nonbinary segmentation in the range [0, 1], where 0 and 1 are the majority of values
-        - 'labeled_seg': Nonbinary, whole values (0, 1, 2...) or (0.0, 1.0, 2.0...) only
-        - 'im': Any other float-valued image where 0 and 1 are not the majority of values.
+        - 'seg-labeled': Nonbinary, whole values (0, 1, 2...) where 0 is the majority of values
```
mguaypaq, line 1857: I hadn't noticed that `labeled_seg` was wrong too, good catch.

spinalcordtoolbox/image.py
```diff
@@ -1850,32 +1850,36 @@ def generate_stitched_qc_images(ims_in: Sequence[Image], im_out: Image) -> Tuple
...
+    zero_most_common = float(unique[0]) == 0.0
+    zero_percentage = np.sum(counts[0]) / np.sum(counts)
     if binary_most_common:
         if binary_percentage == 1.0:
             return 'seg'
         elif binary_percentage > 0.95:
             return 'softseg'
-    if is_whole_only:
+        else:  # binary_percentage <= 0.95
+            pass  # may be 'seg-labeled' or 'anat'
```
mguaypaq, line 1881: I think this is a good way to remind future-us why the line below is `if` rather than `elif`.

mguaypaq (approved):
Thanks for the fixes!

PR #4416, opened 2024-03-26
Replace remaining usages of `nib.save()` with `Image.save()` to mitigate scaling issues

### LINKED ISSUE #3232: Scaling issue with nibabel.save() when header datatype is `int`, but array datatype is `float`

joshuacwnewton:
## Description

This PR is a follow up to previous PRs:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3944
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4346

We have the fix for the scaling issues in `Image.save()` -- we just need to make sure `Image.save()` is actually being used instead of `nib.save()`. So, this PR takes care of the remaining usages of `nib.save()`, _except_:

- `image.py`: This the 1 necessary usage for `Image.save()`
- `sct_detect_pmj.py`: Here, we use `nib.save()` for a `.hdr` file, which I don't think SCT's `Image` class was built to handle.
- `testing/`: These usages are entirely for dummy data creation, where datatype mismatches wouldn't be an issue.

## Linked issues

Fixes #3232.

### DIFF
--- a/spinalcordtoolbox/deepseg_/gm.py
+++ b/spinalcordtoolbox/deepseg_/gm.py
@@ -18,6 +18,7 @@ import numpy as np
 from spinalcordtoolbox import resampling
 from spinalcordtoolbox.utils.sys import __data_dir__
 from spinalcordtoolbox.deepseg_.onnx import onnx_inference
+from spinalcordtoolbox.image import Image
 
 # Models
 # Tuple of (model, metadata)
@@ -275,5 +276,6 @@ def segment_file(input_filename, output_filename,
     if threshold is not None:
         res_data = threshold_predictions(res_data, 0.5)
 
-    nib.save(nii_resampled_original, output_filename)
+    img = Image(param=res_data, hdr=nii_resampled_original.header)
+    img.save(output_filename)
     return output_filename
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -377,7 +377,7 @@ class Image(object):
         else:
             return deepcopy(self)
 
-    def copy_qform_from_ref(self, im_ref):
+    def copy_affine_from_ref(self, im_ref):
         """
         Copy qform and sform and associated codes from a reference Image object
 
@@ -1407,7 +1407,7 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
     # save new image
     im_dest = Image(fname_dest)
     im_warp3d = Image(warp3d)
-    im_warp3d.copy_qform_from_ref(im_dest)
+    im_warp3d.copy_affine_from_ref(im_dest)
 
     # set "intent" code to vector, to be interpreted as warping field
     im_warp3d.header.set_intent('vector', (), '')
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1399,7 +1399,7 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
 
     for iz, fname in enumerate(fname_list):
         img = Image(fname)
-        warp2d = np.asanyarray(img.data)
+        warp2d = img.data
         warp3d[:, :, iz, 0, 0] = warp2d[:, :, 0, 0, 0]
         warp3d[:, :, iz, 0, 1] = warp2d[:, :, 0, 0, 1]
         del warp2d
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1398,20 +1398,20 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
     warp3d = np.zeros([nx, ny, nz, 1, 3])
 
     for iz, fname in enumerate(fname_list):
-        img = nib.load(fname)
-        warp2d = np.asanyarray(img.dataobj)
+        img = Image(fname)
+        warp2d = np.asanyarray(img.data)
         warp3d[:, :, iz, 0, 0] = warp2d[:, :, 0, 0, 0]
         warp3d[:, :, iz, 0, 1] = warp2d[:, :, 0, 0, 1]
         del warp2d
 
     # save new image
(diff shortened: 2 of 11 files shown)


spinalcordtoolbox/image.py
```diff
@@ -1398,20 +1398,20 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
...
         warp3d[:, :, iz, 0, 1] = warp2d[:, :, 0, 0, 1]
         del warp2d
 
     # save new image
-    im_dest = nib.load(fname_dest)
-    affine_dest = im_dest.affine
-    im_warp3d = nib.nifti1.Nifti1Image(warp3d, affine_dest)
+    im_dest = Image(fname_dest)
+    im_warp3d = Image(warp3d)
+    im_warp3d.copy_qform_from_ref(im_dest)
```
joshuacwnewton, line 1410: This change was a little unique, in that it's meant to copy an affine from one image into another image.

The method `copy_qform_from_ref` is pretty misleading -- the actual method copies both the qform and sform. So, perhaps it should instead be called `copy_affine_from_ref`?

spinalcordtoolbox/deepseg_/gm.py
```diff
@@ -275,5 +276,6 @@ def segment_file(input_filename, output_filename,
     if threshold is not None:
         res_data = threshold_predictions(res_data, 0.5)
 
-    nib.save(nii_resampled_original, output_filename)
+    img = Image(param=res_data, hdr=nii_resampled_original.header)
+    img.save(output_filename)
```
mguaypaq, line 280: Wait, is this actually equivalent? In the old code, it looks like `res_data` is just thrown away after being computed? Or does it get mutated in some way that survives in `nii_resampled_original`? But that seems unlikely after getting passed through `np.asanyarray()`.

EDIT: Ah! looking at `help(np.asanyarray)`, it confirms that some cases get passed through as-is (so that mutations flow upstream), and some cases get turned into copies (so that mutations **don't** flow upstream). I'm not sure which case `.dataobj` falls under, or whether Nibabel makes any promises. Seems dangerous!

spinalcordtoolbox/image.py
```diff
@@ -1398,20 +1398,20 @@ def concat_warp2d(fname_list, fname_warp3d, fname_dest):
     warp3d = np.zeros([nx, ny, nz, 1, 3])
 
     for iz, fname in enumerate(fname_list):
-        img = nib.load(fname)
-        warp2d = np.asanyarray(img.dataobj)
+        img = Image(fname)
+        warp2d = np.asanyarray(img.data)
```
mguaypaq, line 1402: I think `Image` guarantees that `img.data` is an actual numpy array, so this should be a no-op. Let's remove it for clarity.
```suggestion
        warp2d = img.data
```

spinalcordtoolbox/image.py
mguaypaq: Yes, that would be a good rename! I'm in favor of it.

mguaypaq (approved):
Thanks for this fix which harmonizes the way we write image files :heart: :heart: :heart:

And also for the very informative commit messages.

spinalcordtoolbox/deepseg_/gm.py
joshuacwnewton: Just to compare the before and after with the full context (since GitHub is doing a poor job showing the full diff):

```diff
nii_resampled_original = resampling.resample_nib(nii_segmentation, new_size=original_res, new_size_type='mm', interpolation='linear')
res_data = np.asanyarray(nii_resampled_original.dataobj)

# Threshold after resampling, only if specified
if threshold is not None:
    res_data = threshold_predictions(res_data, 0.5)

-nib.save(nii_resampled_original, output_filename)
+img = Image(param=res_data, hdr=nii_resampled_original.header)
+img.save(output_filename)
```

You're absolutely correct that `res_data` was thrown away previously. Though, I think there are actually 2 big differences that this change runs into:

1. The use of `np.asanyarray(img.dataobj)` (as you pointed out).
2. The call to `threshold_predictions`, which was completely ignored previously. (!!!)

To address the first point re: `np.asanyarray`, from what I understand, this is actually `nibabel`'s recommended method for extracting the data array from `Nifti1Image` objects. I think the reason I recognize this pattern is because of nibabel's previous deprecation of `get_data` and switch to `get_fdata` (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3854). In that issue, we [came to the conclusion](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3854#issuecomment-1313976579) that `np.asanyarray` was the way to go to preserve the old `get_data` functionality. (And, just to clarify why we use `np.asanyarray` rather than just using `img.dataobj` directly -- I think this is because of Nibabel's memory-mapped array behavior?)

As for the second point -- I'm realizing this is a behavior change! (Though, arguably the fact that we weren't using the thresholded `res_data` smells like a bug to me? So, maybe this is a welcome change...)

mguaypaq: Ah! It's not a behavior change. I just checked, and all the operations here, including the ones which _look_ like they make copies, actually affect the original array in-place:
```py
>>> import numpy as np
>>> x = np.arange(5); x
array([0, 1, 2, 3, 4])
>>> y = np.asanyarray(x); y
array([0, 1, 2, 3, 4])
>>> z = y[:]; z
array([0, 1, 2, 3, 4])
>>> z[0] = 99
>>> z
array([99,  1,  2,  3,  4])
>>> y
array([99,  1,  2,  3,  4])
>>> x
array([99,  1,  2,  3,  4])
```
So, the original code _did_ threshold, although maybe it only succeeded accidentally.

joshuacwnewton: Huzzah! Thank you for checking. :heart:

PR #4421, opened 2024-04-04
Add (internal, semi-private) support for `sct_deepseg` GPU inference on Linux

### LINKED ISSUE #4360: Investigate ways of installing SCT with GPU support for internal dev usage
SCT contains scripts built on deep learning frameworks (mainly PyTorch) that can be sped up using GPUs. Historically, however, SCT has chosen to be CPU-only to avoid the maintenance burden of managing GPU dependencies (e.g. CUDA) and to minimize installation size. 

In most cases, CPU-only works fine for our users. But, for internal R&D, long inference times can sometimes impede research (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4345#issuecomment-1922466223). And, internally, our students have more expertise for installing and troubleshooting GPU dependencies than the average SCT user. So, ideally we would have some sort of process for enabling GPU inference that students of the lab can follow when using SCT in their research.

This issue is for testing GPU inference, and for documenting a set of steps needed to install the necessary dependencies.

----

@valosekj notes:

> Maybe we could optimize `requirements_gpu.txt` to make SCT working on `romane` and `rosenberg`.

So, our internal workstations are probably a good testing ground for any changes we make to SCT.

joshuacwnewton:
## Description

This PR adds support for GPU inference, controlled by:

1. A new option to `install_sct` to install GPU torch.
2. An environment variable called `SCT_USE_GPU` which, if set, will trigger SCT to use `'cuda'` as the device of choice.

I've tried to set the `device` for all of the relevant inference packages (`ivadomed`, `monai`, `nnunet`) but I may have missed something -- it's a bit tricky because the methods for setting `cuda` vary from package to package.

The steps to use this PR are as follows:

> - Clone SCT to your home folder.
> - Checkout `jn/4360-torch-gpu-sct_deepseg`
> - Install SCT using the `-g` flag (e.g. `./install_sct -iyg`)
> - Make sure that `sct_check_dependencies` displays the available GPUs.
> - Set `CUDA_VISIBLE_DEVICES` to one of the GPU IDs (e.g. I booked GPU 2 on rosenberg, so I set `CUDA_VISIBLE_DEVICES=2`). Then, re-run `sct_check_dependencies` and make sure that the the "number of GPUs available to PyTorch" drops to 1.
>    - Note: Since PyTorch can only see the GPUs that are visible, they will start with ID 0 internally to PyTorch, even though the "real" ID for this GPU was 2. There's [not much we can do about that?](https://discuss.pytorch.org/t/how-to-change-the-default-device-of-gpu-device-ids-0/1041/10)  
> - Set `SCT_USE_GPU=1` to turn on GPU inference.
>    - For now, this acts as a "hidden" switch to turn on GPU usage, only for internal lab folks to know about. ;)  
> - Run a sample command (e.g. `SCT_USE_GPU=1 sct_deepseg -task seg_sc_lesion_t2w_sci -i t2.nii.gz`)

## Follow-up

- The steps above should be documented in the SCT dev wiki + the intranet.
- We can't test this PR on GitHub Actions due to a lack of GPU runners, but we might be able to test this on NeuroPoly workstations. 
- This PR only adds support to `sct_deepseg`, and only for Linux. 
    - We may want to extend support to Windows as well.
    - We may also want to consider installing `onnxruntime-gpu` as well, however this currently conflicts with the `ivadomed` package, which explicitly specifies the CPU version of `onnxruntime`.
- As it stands, this is a "hidden" option. We may want to make this a visible option in `sct_deepseg` if we feel ready to accept the maintenance burden that comes with this.

## Linked issues

Fixes #4360.


joshuacwnewton:
> However, it seems that it is unable to properly switch to CPU when `SCT_USE_GPU=0`. It seems to use `cuda` by default (not sure if this is the intended behavior in this branch?)

Ahh, I think I know the issue. Currently I have the check set to see if `SCT_USE_GPU` is set at all (to any value). In other words, to turn off GPU usage, you would have to run `unset SCT_USE_GPU`. But, that seems unintuitive, so I'm happy to change the check to test against the value of `SCT_USE_GPU` as well (0, false, etc.).

naga-karthik:
Yeah, I think that would be great! More so since `SCT_USE_GPU` is a binary variable -- it's straightforward to use `0` when someone doesn't want/cannot to use GPU! 

But yeah, apart from this issue, I think that GPU utilization part was super smooth! Thank you for implementing this ! The speed-ups we get with GPUs _cannot_ be understated !

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -28,7 +28,7 @@
 logger = logging.getLogger(__name__)
 
 
-def segment_and_average_volumes(model_paths, input_filenames, options):
+def segment_and_average_volumes(model_paths, input_filenames, options, use_gpu=False):
```
mguaypaq, line 31: The new argument should be added to the docstring below.

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -59,7 +59,13 @@ def segment_and_average_volumes(model_paths, input_filenames, options):
         if len(model_paths) > 1:  # We have an ensemble, so output messages to distinguish between seeds
             name_seed = Path(path_model).parts[-2]
             logger.info(f"\nUsing '{name_seed}'...")
-        nii_lst, target_lst = imed_inference.segment_volume(path_model, input_filenames, options=options)
+        # NB: ivadomed turns on GPU inference via specifying a single GPU ID. This isn't the recommended way to do
+        #     things, since it prevents us from running multi-GPU jobs. I think ivadomed did things this way to limit
+        #     which GPU is used, but we can already accomplish this using the more universal 'CUDA_VISIBLE_DEVICES'
+        #     environment variable). For now, the best we can do on our end is to select the first GPU from the list
+        #     of available GPUs.
+        gpu_id = 0 if use_gpu else None  # NB: If e.g. 'CUDA_VISIBLE_DEVICES=2,3,4', then 0 will refer to GPU 2.
```
mguaypaq, line 67: Thank you for this comment! It will be very useful for whoever reads the code next.

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -126,7 +134,7 @@ def segment_non_ivadomed(path_model, model_type, input_filenames, threshold, rem
     return im_lst, target_lst
 
 
-def segment_monai(path_img, tmpdir, predictor):
+def segment_monai(path_img, tmpdir, predictor, device):
```
mguaypaq, line 137: Since there's no other indication of the expected type for `device`, would an annotation be good here?
```suggestion
def segment_monai(path_img, tmpdir, predictor, device: torch.device):
```
(I'm also not sure about the expected type of `predictor`, but that one looks more polymorphic so maybe we don't want to pin it down.)

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -178,7 +187,7 @@ def segment_monai(path_img, tmpdir, predictor):
     return [fname_out], [target]
 
 
-def segment_nnunet(path_img, tmpdir, predictor):
+def segment_nnunet(path_img, tmpdir, predictor, device):
```
mguaypaq, line 190: Ditto for the expected type of `device` here, especially since it's unused, so it's hard to infer.

Actually, maybe add a note that `device` is unused, and only present so that `segment_nnunet` has the same function signature as `segment_monai`?

I guess Monai has no mechanism to specify the device?
```suggestion
def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
```

spinalcordtoolbox/deepseg/monai.py
```diff
@@ -42,13 +42,12 @@
 }
 
 
-def create_nnunet_from_plans(path_model):
+def create_nnunet_from_plans(path_model, device):
```
mguaypaq, line 45: Ditto.
```suggestion
def create_nnunet_from_plans(path_model, device: torch.device):
```

spinalcordtoolbox/deepseg/nnunet.py
```diff
@@ -14,7 +12,7 @@
 from nnunetv2.inference.predict_from_raw_data import nnUNetPredictor   # noqa: E402
 
 
-def create_nnunet_from_plans(path_model):
+def create_nnunet_from_plans(path_model, device):
```
mguaypaq, line 15: I was going to suggest putting another `torch.device` annotation here, but `torch` is no longer imported so we can't. I see that there's [a standard pattern for this situation](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING), but maybe that's too heavy.

mguaypaq (approved):
Since @naga-karthik already tested the PR, here's a formal "Approve" review. I still did a quick skim of the code, so I put in a few suggestions.

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -236,17 +238,43 @@ def main(argv: Sequence[str]):
...
+        try:
+            status, n_gpus = run_proc([
+                "nvidia-smi",
+                "--query-gpu=count",
+                "--format=csv,noheader",
+                "-i", "0",
+            ], verbose=0, raise_exception=False)
+        except FileNotFoundError as e:
+            e.strerror = "GPU version of torch is installed, but could not find NVIDIA's GPU software"
+            raise e
```
joshuacwnewton, line 251: Thanks to macOS, I realize now that `+cpu` isn't solely enough of a check (during `sct_check_dependencies`).

macOS doesn't actually have distinct versions -- `torch` _is_ the CPU version, because there is no Mac GPU version of torch.

mguaypaq (approved):
Thanks for the final fixes!

joshuacwnewton:
I'm trying to remember syntax... It might be possible that it all needs to be one line? (CUDA_AVAILABLE_DEVICES=1 sct_check_dependencies)

Or, alternatively, to have the setting stick around, you might need to run "export CUDA_AVAILABLE_DEVICED=1"

valosekj:
> It might be possible that it all needs to be one line?

Yes! You are right, thank you!

```console
p118175@romane:~$ CUDA_VISIBLE_DEVICES=1 sct_check_dependencies

--
Spinal Cord Toolbox (git-master-4868a2cdd79a7109bce1de8747f52cbed9d18b0f)

sct_check_dependencies
--


SYSTEM INFORMATION
------------------
SCT info:
- version: git-master-4868a2cdd79a7109bce1de8747f52cbed9d18b0f
- path: /home/GRAMES.POLYMTL.CA/p118175/code/spinalcordtoolbox
OS: linux (Linux-6.8.0-39-generic-x86_64-with-glibc2.39)
CPU cores: Available: 64, Used by ITK functions: 64
RAM: Total: 515637MB, Used: 25099MB, Available: 484450MB
GPU 0: NVIDIA RTX A6000 (Driver: 535.183.01, vBIOS: 94.02.5C.00.02) [VRAM Free: 48675 MiB/49140 MiB]
GPU 1: NVIDIA RTX A6000 (Driver: 535.183.01, vBIOS: 94.02.5C.00.02) [VRAM Free: 32397 MiB/49140 MiB]
GPU 2: NVIDIA RTX A6000 (Driver: 535.183.01, vBIOS: 94.02.5C.00.02) [VRAM Free: 31333 MiB/49140 MiB]
GPU 3: NVIDIA RTX A6000 (Driver: 535.183.01, vBIOS: 94.02.5C.00.02) [VRAM Free: 48675 MiB/49140 MiB]

OPTIONAL DEPENDENCIES
---------------------
Check FSLeyes version...............................[OK] (1.10.1)
Check if CUDA is available in PyTorch...............[OK]
Check CUDA version used by PyTorch..................[OK] (12.1)
Check number of GPUs available to PyTorch...........[OK] (Device count: 1)
```

valosekj:
> The steps above should be documented in the SCT dev wiki + the intranet.

Done now: https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Installing-SCT-for-GPU. Please let me know what you think :-) 
FYI: I've used the `Contributing:` category as it already contained a similar page, [Contributing: Installing SCT for development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Installing-SCT-for-development).

joshuacwnewton:
This is wonderful! Thank you so much for taking the time to do this, @valosekj. :hearts:

Issue #4423, opened 2024-04-05
QC reports require locking `index.html` which doesn't function correctly on NFS-mounted drives


joshuacwnewton:
## Background

Our QC reports contain a table of QC entries:

<details>

![image](https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/16181459/ca9d89c9-3828-4182-ac1c-89edbb80e6ee)

</details>

Each entry in the table is stored as simple JSON data listing the subject, date, CLI script, etc. We generate individual `.json` files for each entry via our `qc.py` module. 

However, when it comes to loading the data into a table in `index.html`, we store this JSON data _directly within a JavaScript variable inside the `index.html` file_. This means that every time we create a new QC entry using our Python script, we need to completely rewrite the `index.html` file, substituting the JSON data into a placeholder in the HTML template:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1af24e30cdae0b465f6abd33c5633b6bcb71de39/spinalcordtoolbox/reports/qc.py#L613

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1af24e30cdae0b465f6abd33c5633b6bcb71de39/spinalcordtoolbox/reports/assets/index.html#L159

![untitled(2)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/16181459/8bedf273-1ebd-4008-bd9d-a417abccbb2e)

## Problem

Because all new QC entries have the same destination file, concurrent writes can cause issues with corrupted/missing entries. So, we explicitly lock the `index.html` before writing to it.

However, in https://github.com/sct-pipeline/spine-park/issues/15#issuecomment-2038269172, we ran into some system errors when trying to lock and write to the `index.html` file while it was on a mounted drive.

## Proposed solution

Initially, I can think of two ways we could fix this issue:

- Lock and write to the `index.html` file in a tmpdir [on a non-mounted drive](https://github.com/sct-pipeline/spine-park/issues/15#issuecomment-2039987276), then copy the file to the destination directory when `sct_run_batch` is complete.
   - This may make the QC report folder difficult to access during processing, as pointed out by @jcohenadad in [this comment](https://github.com/sct-pipeline/spine-park/issues/15#issuecomment-2039987276). 
[...]

PR #4427, opened 2024-04-09
Keep CSS/JS/etc. assets up to date when QC report is regenerated with a newer version of SCT

### LINKED ISSUE #4422: `git fetch` then re-running `sct_qc` may result in broken report due to mismatched asset versions
> I cannot 'save all' (@joshuacwnewton would you mind looking into this?)
>
> _Originally posted by @jcohenadad in https://github.com/sct-pipeline/fmri-segmentation/issues/35#issuecomment-2037846826_

image

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1af24e30cdae0b465f6abd33c5633b6bcb71de39/spinalcordtoolbox/reports/assets/index.html#L313

joshuacwnewton:
## Description

When we generate QC reports, we universally update `index.html` with the copy stored in the repo, due to the behavior described in #4423.

This can cause problems when, for example, `index.html` and `main.js` are both updated simultaneously between SCT versions (e.g. #4317). In such a case, if QC reports are regenerated using the newer SCT version, `index.html` will be updated with the changes, but `main.js` will not. And, since `index.html` and `main.js` are so interdependent, this can often cause broken behavior in the QC report.

So, this PR ensures that the assets are updated alongside `index.html`, but only if the checksums differ, which is similar to @mguaypaq's versioning suggestion in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4422#issuecomment-2037901197.

## Testing

I tested this PR by generating a QC report, then adding a comment to `main.js`, then re-generating the QC report:

image

## Linked issues

Fixes #4422.


mguaypaq:
I'm not sure about doing checksums in this case. I would do either more or less, depending on the performance trade-off we're aiming for:
* Status quo: we assume that asset files have the right content if they have the right file name (with no embedded version number). Bad idea, because the assumption doesn't hold. The cost of this is:
  * Reading the directory entries (which contain the file names) always.
  * Reading the source asset file contents rarely (to do the copy).
  * Writing the target asset file contents rarely (to do the copy).
* Current solution: checksum the asset files. The cost of this is:
  * Reading the directory entries always.
  * Reading the source asset file contents always (plus computing the checksum).
  * Reading the target asset file contents always (plus computing the checksum).
  * Writing the target asset file contents rarely.
* Doing more 1: compare the actual asset file contents, instead of checksums. We're already paying the cost of reading the source and target asset file contents, so doing checksums doesn't save anything, but it _does_ introduce the (very unlikely) problem of hash collisions. The cost:
  * Reading the directory entries always.
  * Reading the source asset file contents always.
  * Reading the target asset file contents always.
  * Writing the target asset file contents rarely.
* Doing more 2: unconditionally write the asset file contents. This saves on reading, but some file systems may not like the constant writing, so it's probably a bad idea. Cost:
  * Reading the directory entries always.
  * Reading the source asset file contents always.
  * Writing the target asset file contents always.
* Doing less 1: pre-compute the source asset checksums at release time, and save them somewhere. This doesn't cover install-from-git cases, and would be a logistical nightmare, so it would be a bad idea. But it's one reason people sometimes use checksums, to have a tiny savings:
  * Reading the directory entries always.
  * Reading the source asset file contents **rarely** (only to copy, because the checksums are pre-computed).
  * Reading the target asset file contents always (still have to compute their checksums).
  * Writing the target asset file contents rarely.
* Doing less 2: include a content checksum in the asset filenames, similar to [the cache-busting techniques commonly used on the web](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#cache_busting). The cost:
  * Reading the directory entries always, which now contain the checksum values.
  * Reading the source asset file contents rarely (only to copy).
  * Reading the target asset file contents never.
  * Writing the target asset file contents rarely.

Sorry, this was long-winded, but I wanted to explicitly illustrate the trade-offs, because they're kind of subtle. We're dealing with small numbers of small files, so I think _Doing more 1_ and _Doing less 2_ are both reasonable.

joshuacwnewton:
Ahh, thank you so much for this deep dive. Just to give a tl;dr summary (to make sure I fully understand):

- If we're reading the file contents of both files every single time (to re-compute the checksum each and every time), then the checksum itself serves no purpose!
- If we want to actually use the checksum purposefully, it would be better to cache the checksums to avoid needing to read the files at all.

This was really informative! And, it feels obvious in hindsight, given that download hosts provide the checksums for validation purposes (the whole point being that you can check against a short string, rather than the entire copy of the file).

----

As far as _Doing more 1_ vs. _Doing less 2_, I'm leaning towards the former? While "Doing less" is "less" in terms of computation, it's still "more" in terms of logistics, since we'd have to fuss with A) re-committing the files with hashed names, and B) making sure those files can be read by one another (e.g. `index.html` needing to point to `main-{hash}.js`).

"Doing more 1" at least simplifies this PR greatly, since I can revert the checksum-based changes and whatnot.

mguaypaq (approved):
Looks good, thanks! I guess I should port this change over to `reports/qc2.py` as well.

PR #4428, opened 2024-04-09
Move most broken link checking to a daily job

### LINKED ISSUE #4426: Only enable broken link checker on `master`, and not PRs
- Running `curl` on 300+ links for every PR update is a bit rude
- Network checks are flaky, and can result in intermittent failures, which is noisy on PRs. (This is easily fixed 99% of the time by just re-running the check. But, it's a bit annoying to have to do that on PRs, since it can mislead us into thinking there are errors with code changes.)
- The broken link checker is most useful for checking when the link itself changes (e.g. a page moves/a site closes down/etc.). This happens irrespective of SCT's own changes. So, making it a cron job makes sense to me?

mguaypaq:
Fixes #4426.

I still kept a limited form of the link checker for PRs. It should only check added/modified lines, so that should catch typos in new/changed links. The daily job should catch websites that stop working.

For the daily run, I picked a scheduled time that's an hour before all the other tests (either 5am or 6am Montreal time, depending on daylight savings), to avoid resource bottlenecks.

### DIFF
--- a/.github/workflows/check-broken-links.yml
+++ b/.github/workflows/check-broken-links.yml
@@ -18,7 +18,7 @@ jobs:
           \( \! -name 'CHANGES.md' \) \( \! -name 'requirements.txt' \) \( \! -name 'requirements-freeze.txt' \) |
           xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" |
           sed 's/:/;/' |
-          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
+          xargs -n 1 ./.github/workflows/check-url.sh
 
     - name: Summarize
       run: |
--- a/.github/workflows/check-broken-links.yml
+++ b/.github/workflows/check-broken-links.yml
@@ -1,0 +1,30 @@
+name: Check for broken links
+
+on:
+  schedule:
+    - cron:  '0 10 * * *'
+
+jobs:
+  check_broken_links:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Check for broken links
+      run: |
+        touch valid_urls.txt redirected_urls.txt invalid_urls.txt
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \
+          \( \! -name 'CHANGES.md' \) \( \! -name 'requirements.txt' \) \( \! -name 'requirements-freeze.txt' \) |
+          xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" |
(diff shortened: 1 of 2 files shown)


.github/workflows/check-broken-links.yml
```diff
@@ -0,0 +1,30 @@
...
+      uses: actions/checkout@v4
+
+    - name: Check for broken links
+      run: |
+        touch valid_urls.txt redirected_urls.txt invalid_urls.txt
+        find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \
+          \( \! -name 'CHANGES.md' \) \( \! -name 'requirements.txt' \) \( \! -name 'requirements-freeze.txt' \) |
+          xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" |
+          sed 's/:/;/' |
+          xargs -n 1 -P 8 ./.github/workflows/check-url.sh
```
joshuacwnewton, line 21: Since we're now only running the full link checker on daily jobs, perhaps we want to get rid of `-P 8` to have more coherent logging in this step? (The previous discussion of "3m vs. 13m" is kind of moot now.)

```suggestion
          xargs -n 1 ./.github/workflows/check-url.sh
```

.github/workflows/check-code-quality.yml
```diff
@@ -32,21 +32,14 @@ jobs:
...
-        echo -en "========== \033[0;32m$NUM_OK passed\033[0;0m, \033[0;33m$NUM_REDIRECT redirected\033[0;0m, \033[0;31m$NUM_BAD failed\033[0;0m ==========\n"
-        exit $NUM_BAD
-
+        set -eu -o pipefail
+        git diff origin/master... |
+          (grep '^+' || (($? == 1)) ) |
+          (grep -Eo '\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]' || (($? == 1)) ) |
+          sed 's/:/;/' |
+          xargs -rn 1 .github/workflows/check-url.sh
+        test ! -s invalid_urls.txt
```
joshuacwnewton, line 45: Comparing this to the flake8 check (which also uses `git diff`)

- `flake8`: Diffs filenames only, filters for `.py` files, runs flake8 on filenames.
- `check-url`: Shows full diffs, filters for additions (`+`), filters added lines for URLs, then passes URLs to `check-url.sh`.
  - `-r` is used because the diff may not always contain URLs, `-P 8` is absent for reasons mentioned in my other comment.
  - `test ! -s` passes (returns `0`) if `NOT` "file size greater than 0", meaning... if the file size _is_ greater than 0, then test will fail and return code `1`. (And, `set -e` causes this to result in an error.)

Makes sense to me!

joshuacwnewton (approved):
LGTM! Thank you for going beyond and adding a modified check to the PRs. :)

.github/workflows/check-code-quality.yml
mguaypaq: Exactly!

PR #4430, opened 2024-04-09
Fix broken link checker

mguaypaq:
Since `grep` no longer receives filenames, it no longer puts `$FILENAME:` at the start of each line. But `check-url.sh` _does_ expect a semicolon before the url to check. So we need to insert it, rather than changing the colon after `https` into a semicolon.

Sorry!

Follow-up for #4428.

### DIFF
--- a/.github/workflows/check-broken-links.yml
+++ b/.github/workflows/check-broken-links.yml
@@ -16,7 +16,7 @@ jobs:
         touch valid_urls.txt redirected_urls.txt invalid_urls.txt
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \
           \( \! -name 'CHANGES.md' \) \( \! -name 'requirements.txt' \) \( \! -name 'requirements-freeze.txt' \) |
-          xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" |
+          xargs grep -Eio "\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]" |
           sed 's/:/;/' |
           xargs -n 1 ./.github/workflows/check-url.sh
 
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -39,7 +39,7 @@ jobs:
         set -eu -o pipefail
         git diff origin/master... |
           (grep '^+' || (($? == 1)) ) |
-          (grep -Eo '\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]' || (($? == 1)) ) |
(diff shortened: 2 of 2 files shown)


.github/workflows/check-code-quality.yml
```diff
@@ -40,6 +40,6 @@ jobs:
         git diff origin/master... |
           (grep '^+' || (($? == 1)) ) |
           (grep -Eo '\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]' || (($? == 1)) ) |
-          sed 's/:/;/' |
```
joshuacwnewton, line 43: > But check-url.sh does expect a semicolon before the url to check.

Ahh, of course! 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f3e79f86dcd262c12664d862df78f38e8335242b/.github/workflows/check-url.sh#L7-L8

I guess for PRs, the filename will be `''`, then?

```console
dot|joshua@monarch:~$ filename=$(cut -d ";" -f 1 <<< ";test")
dot|joshua@monarch:~$ echo $filename
```

That's not too big a deal, I suppose, since most PRs will only ever add one or two URLs at most, meaning they won't be too hard to identify (vs. combing through 300+ links?)

joshuacwnewton (approved):

.github/workflows/check-broken-links.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         touch valid_urls.txt redirected_urls.txt invalid_urls.txt
         find . -type f \( -name "*.txt" -o -name "*.md" -o -name "*.py" -o -name "*.rst" \) \
           \( \! -name 'CHANGES.md' \) \( \! -name 'requirements.txt' \) \( \! -name 'requirements-freeze.txt' \) |
-          xargs grep -Eo "\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]" |
+          xargs grep -Eio "\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]" |
```
joshuacwnewton, line 19: :hearts: !!!

Nice thinking with `-i`. :)

.github/workflows/check-code-quality.yml
mguaypaq: Oh! Good point, I didn't think that the filename was useful. We can add one!

.github/workflows/check-code-quality.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         set -eu -o pipefail
         git diff origin/master... |
           (grep '^+' || (($? == 1)) ) |
-          (grep -Eo '\b(https?)://[-A-Za-z0-9+&@#/%?=~_|$!:,.;]*[A-Za-z0-9+&@#/%=~_|$]' || (($? == 1)) ) |
-          sed 's/:/;/' |
+          (grep -Eio '\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]' || (($? == 1)) ) |
+          sed 's/^/;/' |
```
mguaypaq, line 43:
```suggestion
          sed 's/^/diff;/' |
```

PR #4432, opened 2024-04-10
Update QC assets when saving the QC report

mguaypaq:
Follow-up for #4427 in `reports/qc2.py`.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -172,7 +172,7 @@ def update_files(resource: Traversable, destination: Path):
     if resource.is_dir():
         path.mkdir(exist_ok=True)
         for sub_resource in resource.iterdir():
-            copy_missing(sub_resource, path)
+            update_files(sub_resource, path)
     elif resource.is_file():
         new_content = resource.read_bytes()
         old_content = path.read_bytes() if path.is_file() else None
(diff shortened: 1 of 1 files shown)


joshuacwnewton (changes requested):
It looks like there's still one existing instance of `copy_missing`, hence the error:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/67e0a5e7d9dab9207011f387822be26cbeadd731/spinalcordtoolbox/reports/qc2.py#L175

spinalcordtoolbox/reports/qc2.py
```diff
@@ -157,25 +157,27 @@ def create_qc_entry(
...
-    Make sure that a copy of `resource` exists at `destination`,
-    by creating any missing files and directories recursively.
+    Make sure that an up-to-date copy of `resource` exists at `destination`,
+    by creating or updating files and directories recursively.
     """
     path = destination / resource.name
     if resource.is_dir():
         path.mkdir(exist_ok=True)
         for sub_resource in resource.iterdir():
             copy_missing(sub_resource, path)
```
joshuacwnewton, line 175:
```suggestion
            update_files(sub_resource, path)
```

joshuacwnewton (approved):
Actually, I'll pre-approve since, apart from the name issue (which can be fixed via the suggestion), this LGTM!

PR #4439, opened 2024-04-15
Isolate `index.html`-writing code into reusable script + replace `Lock` with Semaphore-based `mutex`

### LINKED ISSUE #4423: QC reports require locking `index.html` which doesn't function correctly on NFS-mounted drives

### LINKED ISSUE #4438: Make it possible to remove lines in a QC report
That would be a very useful feature to facilitate sharing of experiments.

joshuacwnewton:
## Description

This PR contains 2 key changes, both with the goal of better isolating the part of the QC code that writes the report files to the QC directory:

1. Separate the "import-JSON-into-HTML" steps into a standalone script that can be run both as part of the QC-generation, but also by the user to remove unwanted entries.
   - Fixes #4438.
   - Another benefit of isolating this code is that it gives me a place to document the limitations of client-side JavaScript, which I spent quite a bit of time researching. :sweat_smile: 
2. Use a mutex instead of locking `index.html`.
   - The mutex uses `portalocker.BoundedSemaphore`, which still uses `portalocker.Lock` under the hood, but with some additional helper code (retries, timeouts, semaphore creation in a tempdir on the local filesystem), which helps us to avoid reinventing the wheel.
   - This allows us to limit the number of parallel processes independently of the location of `index.html`, which avoids issues with writing the QC reports to NFS-mounted drives. (Fixes #4423.)
   - (I'm presuming that SCT itself won't be installed on an NFS-mounted drive, only the data, which was the case for the above issue.)

## Linked issues

Fixes #4423.
Fixes #4438.


spinalcordtoolbox/reports/assets/__init__.py
joshuacwnewton, line 1: These `__init__.py` files are a bit bothersome, cluttering both the local assets folder and the output `_assets` folder.

We could instead remove these, and run the script in a subprocess using `run_proc` or something along those lines?

mguaypaq (changes requested):
I like the insight that the locking mechanism doesn't have to live in the same place as the QC report itself, freeing us from weird networked filesystem issues. And gaining the ability to refresh QC reports is cool!

I think this is good, but I'd like to double-check the final scoping of the lock (see my other comment below) before this PR gets merged.

spinalcordtoolbox/utils/fs.py
```diff
@@ -259,3 +262,24 @@ def relpath_or_abspath(child_path, parent_path):
...
+
+    portalocker.BoundedSemaphore is very similar to threading.BoundedSemaphore,
+    but works across multiple processes and across multiple operating systems. So,
+    we can spawn multiple processes using `sct_run_batch`, while still ensuring
+    that we create QC reports sequentially (without mangling the output files).
+
+    We use a mutex over a lock because the mutex doesn't depend on the destination
+    of the locked files, which allows us to avoid locking on e.g. NFS-mounted drives.
+    """
+    semaphore = portalocker.BoundedSemaphore(maximum=1, name=name)
```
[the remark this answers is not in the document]
joshuacwnewton: > Could we include something like a hash of the QC report path in the lock name? In which case, this would justify the existence of this context manager wrapper 😅

Ahhh, very good idea!! Good thinking. This makes a lot of sense given the whole purpose of the lock in the first place (concurrent writes to the same QC report path).

spinalcordtoolbox/utils/fs.py
```diff
@@ -283,3 +283,4 @@ def mutex(name):
         yield semaphore
     finally:
         semaphore.release()
+        os.remove(semaphore.get_filenames()[0])  # nb: BoundedSemaphore doesn't clean up after itself
```
mguaypaq, line 286: I think this interferes with the functioning of the lock. As a small test, I did this:
1. In one python instance, acquire a `BoundedSemaphore`. There was no wait.
2. In a second python instance, acquire a `BoundedSemaphore` with the same name. There was a wait, then a timeout. So, this scenario works.

In contrast, deleting the file broke the synchronization:
1. In one python instance, acquire a `BoundedSemaphore`. There was no wait.
2. Manually delete the file.
3. In a second python instance, acquire a `BoundedSemaphore` with the same name. There was no wait, but there should have been.

It's unfortunate to be leaving cruft around, but I think it's necessary in this case. At least for unix-likes, I think `/tmp` gets cleared at every reboot, usually. (I have no idea for Windows.)

mguaypaq: So, I think `portalocker`'s file locks are associated with an inode rather than a filename (at least on Unix-likes, who knows about Windows). To test this, I did:
1. In process A, acquire a `portalocker.Lock` on `somefile`.
2. In process B, try to acquire a lock on `somefile`: this blocks and times out, as expected.
3. Manually, I moved (renamed) the file from `somefile` to `otherfile`. This keeps the same inode, but now a different path refers to it.
4. In process B, try to acquire the lock on `somefile`: this immediately re-creates `somefile` and acquires the lock (as expected, because even though it's the same file name, it's a new inode).
5. In process B, try to acquire a lock on `otherfile`: this blocks and times out, because process A still holds the lock on that inode, from when it was called `somefile`.

So, for `portalocker.BoundedSemaphore` to work properly, we can't delete the associated tmp files, because that breaks the link between the file name (which we use to acquire the lock) and its inode (which is the thing that's actually locked).

Issue #4442, opened 2024-04-17
Add canproco nnUNet model to sct_deepseg models


plbenveniste:
Opening this issue as I will create a new branch and PR to add the canproco nnUNet model to the sct_deepseg model method. 

The model used can be found in this [release](https://github.com/ivadomed/canproco/releases/tag/r20240125). 
I will use the 2D model for SCT as it is faster than the 3D model and showed similar performances. 

The branch in use is `plb/4442_add_canproco_nnunet`

PR #4443, opened 2024-04-17
Add the canproco nnUNet model for MS lesion segmentation to sct_deepseg

plbenveniste:
## Description

Opening this PR to add the 2D nnUNet model to segment MS lesions in the spinal cord. The model is a region-based model which means that there are 2 classes (spinal cord and lesion) and the lesions are always located in the spinal cord. The model was trained on STIR and inverted PSIR (multiplied by -1). The trained model can be found in this release: https://github.com/ivadomed/canproco/releases/tag/r20240125. 

I used the 2D model as it is faster than the 3D model and showed similar performances.

## Testing the model

To test this PR, you will need to install SCT from this branch. Here are the steps:

1. Open a terminal, then run `git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git` to get a copy of the source code.
2. Run `cd spinalcordtoolbox` to enter the SCT source code folder.
3. Run `git checkout plb/4442_add_canproco_nnunet` to switch to this branch.
4. Run `./install_sct -iy` to install SCT (the installation will exist in this same folder). Make sure that you see all green `[OK]` at the end of the installation.
5. Run `sct_deepseg -install-task seg_sc_ms_lesion_stir_psir` to download the model.
6. Run `sct_deepseg -task seg_sc_ms_lesion_stir_psir -i img.nii.gz` to segment an image.

## Linked issues

This resolves issue #4442


plbenveniste:
The code is "done" but not yet tested. 
It raises the following [issue](https://github.com/MIC-DKFZ/nnUNet/issues/2112) which @joshuacwnewton opened on the nnUNet repo.

plbenveniste:
The code is "done" but not yet tested. 
It raises the following issue which @joshuacwnewton opened on the nnUNet repo: 
`ValueError: not enough values to unpack (expected 4, got 1)`

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: One small bit of early feedback:

We have a check that tests for "code quality", and it has given some feedback here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8740724210/job/23985054239?pr=4443

Could you please check these messages? (You can google the code (e.g. "E261 python") and it will lead you to a "[flake8rules.com](https://www.flake8rules.com/rules/E261.html)" page which explains in detail what changes are needed and why.)

plbenveniste:
I changed the `contrast` field of the model from `["stir", "psir"]` to `["stir or inverted psir"]` to avoid the following error returned by command line `sct_deepseg -i IMAGE -o OUTPUT -task seg_sc_ms_lesion_stir_psir`: 

```console
sct_deepseg: error: 1 input files found. Please provide all required input files for the task ['seg_sc_ms_lesion_stir_psir'], i.e. contrasts: stir, psir.
```

spinalcordtoolbox/deepseg/models.py
plbenveniste: Thanks for the feedback :) 
Corrections done !

spinalcordtoolbox/deepseg/models.py
```diff
@@ -321,6 +331,17 @@
                              'the University of Zurich.',
          'url': 'https://github.com/ivadomed/model_seg_mouse-sc_wm-gm_t1',
          'models': ['model_seg_gm_wm_mouse_nnunet']},
+    'seg_sc_ms_lesion_stir_psir':
+        {'description': 'Segmentation of spinal cord and MS lesions for STIR and inverted PSIR (multiplied by -1) contrasts ',
```
joshuacwnewton, line 335: Minor, minor nitpick, but the "description" is a bit longer than the other ones. On a fresh terminal (80 pixels wide), the description wraps quite a lot:

image

So, I would suggest a slight rewording just to make it fit:

```suggestion
        {'description': 'MS lesion/SC seg for STIR/inverted PSIR contrasts',
```

I think the extended description does a good job already of explaining the "multiplied by 1" part?

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton, line 1: It seems like your model was trained [using the AIL orientation](https://github.com/ivadomed/canproco/blob/da39ff9b3c3886b3ad931f5d8baf4963e45a9a2c/packaging/run_inference_single_subject.py#L91-L97)? If that's the case, then we will need to modify the `sct_deepseg` code, since it currently assumes that nnunetv2 models were trained using LPI orientation:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8f2b0e0eb956cbd12d1cabdfbee34e1ca8d99c34/spinalcordtoolbox/deepseg/inference.py#L198-L202

joshuacwnewton, line 1: For the suffix, it seems like `'Spinal cord'` (with a space) was used. Spaces can cause issues with terminal commands, so I would recommend changing the `dataset.json` configuration file to use a different suffix. (I discovered this because I got an error when copying and pasting the `fsleyes` command that is shown at the end of `sct_deepseg`.)

image

I would recommend `lesion` and `sc` (lowercase), since that's what was used previously with the SCIseg model:

image

joshuacwnewton (changes requested):
I tested this using the data you had previously shared for debugging that one QC issue with PSIR images :)

```
sct_deepseg -install-task seg_sc_ms_lesion_stir_psir
sct_deepseg -task seg_sc_ms_lesion_stir_psir -i sub-edm005_ses-M0_PSIR.nii.gz 
```

I ran into two main issues, plus one minor nitpick? But, apart from that, the changes mostly LGTM! :)

spinalcordtoolbox/deepseg/models.py
jcohenadad: i would not mention the multiplied by -1 in the description of the task

jcohenadad: To clarify: I thought that the code already multiplies by -1 if the string 'psir' is identified-- chatting with @plbenveniste I understand this is not the case, and the user is expected to do it. Problem: they won't understand/ will make mistake, etc.

bottomline: we should do it ourselves, probably in the wrapper (good idea?) or somewhere else...

spinalcordtoolbox/scripts/sct_deepseg.py
plbenveniste: Yes ! I think that's a good idea since we're going to have more models added and more specificity added.

plbenveniste: The `dataset.json` file was corrected in the release.
Done !

spinalcordtoolbox/deepseg/models.py
plbenveniste: Yes I think that's a good idea. 
It just requires to do : `sct_maths -i IMAGE -mul -1 -o OUTPUT`

Should the user then specify in the command if the image is PSIR or STIR ? Or should the code look for "PSIR" in the file name ?

jcohenadad: i though there was already a dedicated use of the flag ‘-c stir/psir’ (otherwise there is no use for this flag since it’s the same model for both contrasts)

joshuacwnewton: > otherwise there is no use for this flag since it’s the same model for both contrasts

That is exactly the reason I opened the issue 

- #4445

joshuacwnewton: I have some ideas for how we can update the `-c` argument to support switching between PSIR/STIR contrasts, without an error occurring if `-c` is not passed.

I will try to get to this tonight, or early Monday at the latest.

plbenveniste: Nice ! But I guess that we always want `-c` to be passed for the model, to avoid cases' where people forget to multiply PSIR images by -1. 
However, I agree with you that we don't want `-c` to be required for the majority of `sct_deepseg` models

jcohenadad: > Nice ! But I guess that we always want -c to be passed for the model, to avoid cases' where people forget to multiply PSIR images by -1.

The model won't be used by a lot of people in the next few months, so I think it's OK to not enforce this flag. At some point we will likely have a model that won't require to multiply by -1, so this is a temporary hack anyway

joshuacwnewton: I chose to enforce `-c` flag, but only for `-task seg_sc_ms_lesion_stir_psir`. So, this behavior will still be invisible to 99% of users, but for people who are using `seg_sc_ms_lesion_stir_psir`, they will explicitly choose whether to invert or not.

joshuacwnewton (approved):
I am personally happy with the changes here, but I will wait until @jcohenadad provides feedback for the open discussion thread.

(I will say that adding preprocessing (inversion) for a specific model was surprisingly tricky -- the multi-model, multi-task, multi-image behavior in `sct_deepseg` is quite convoluted, so it might be worth a secondary review.)

jcohenadad:
I tried 3b9edacff2b86ad503654e4d0389f6f032a64399

If -c missing, do we want to make the error if more intuitive for the user? Can we use conditional argparse?
```console
^[[Ajulien-macbook:~/data.neuro/canproco/sub-van222/ses-M0/anat $ sct_deepseg -i sub-van222_ses-M0_PSIR.nii.gz -task seg_sc_ms_lesion_stir_psir 

--
Spinal Cord Toolbox (git-plb/4442_add_canproco_nnunet-3b9edacff2b86ad503654e4d0389f6f032a64399)

sct_deepseg -i sub-van222_ses-M0_PSIR.nii.gz -task seg_sc_ms_lesion_stir_psir
--

  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 291, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 215, in main
    contrast = arguments.c[0]
TypeError: 'NoneType' object is not subscriptable
```

joshuacwnewton:
Sorry about that! You should see this now:

image

jcohenadad:
the help shows:
```
  -c <file> [<file> ...]
```
but "-c" is a string instead of a file, no?

jcohenadad:
3b976d40d9b5ebee8aa075fecf52769800e94298 fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4443#issuecomment-2070394623, thank you! Another thing I noticed: the result is not saved in the current directory (which is SCT's typical behaviour):
~~~
Image header specifies datatype 'float64', but array is of type 'uint8'. Header metadata will be overwritten to use 'uint8'.
Saving results to: /var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-04-22_15-58-45_sct_deepseg_nju85e6m/nnUNet_prediction/sub-van222_ses-M0_PSIR_pred_sc_seg.nii.gz
~~~

joshuacwnewton:
> Another thing I noticed: the result is not saved in the current directory (which is SCT's typical behaviour):

Ahhh, good catch! This is a result of me adding the inversion logic. (The output directory is inferred from the input file, however because I replaced the input file with the temporary inverted image, then the output directory also became the temporary directory. Oops.)

This should be a very quick fix. :hearts:

jcohenadad (changes requested):
Oops! I don't know if this is related to this PR, but i'm using the contrast-agnostic method for another project, and my output segmentation suffix is screwed up:

```bash
sct_deepseg -i sub-CRMBM1_UNIT1.nii.gz -task seg_sc_contrast_agnostic

--
Spinal Cord Toolbox (git-plb/4442_add_canproco_nnunet-b050b93e080378971964e17b687d66d1ecdf591e)

sct_deepseg -i sub-CRMBM1_UNIT1.nii.gz -task seg_sc_contrast_agnostic
--

Creating temporary folder (/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-04-23_12-49-07_sct_deepseg_sis9a22i)
Copied sub-CRMBM1_UNIT1.nii.gz to /var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-04-23_12-49-07_sct_deepseg_sis9a22i/sub-CRMBM1_UNIT1.nii.gz
Starting inference...
Inference done.
Total inference time: 0 minute(s) 13 seconds
Saving results to: /var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-04-23_12-49-07_sct_deepseg_sis9a22i/sub-CRMBM1_UNIT1_seg.nii.gz
Image header specifies datatype 'float32', but array is of type '<class 'numpy.uint8'>'. Header metadata will be overwritten to use '<class 'numpy.uint8'>'.
File /Users/julien/code/coil-qc-code/data/sub-CRMBM1/anat/sub-CRMBM1_UNIT1.nii.gz_seg.nii.gz already exists. Will overwrite it.

Done! To view results, type:
fsleyes sub-CRMBM1_UNIT1.nii.gz -cm greyscale sub-CRMBM1_UNIT1.nii.gz_seg.nii.gz -cm red -a 70.0 &
```


**EDIT 20240423_125236**: Trying the master version 5c5dd752b0fb5f01bf4bdc9f58ca7d3e3df4402d solves the issue:
~~~
[...]
fsleyes sub-CRMBM1_UNIT1.nii.gz -cm greyscale sub-CRMBM1_UNIT1_seg.nii.gz -cm red -a 70.0 &
~~~

joshuacwnewton:
Oh dear! This is definitely due to a change I made in this PR. My deepest apologies, I'll fix it asap.

jcohenadad:
> Oh dear! This is definitely due to a change I made in this PR. My deepest apologies, I'll fix it asap.

no worries 😊  I'm surprised it was not caught by our testing framework. Maybe something to add?

plbenveniste:
Code tested successfully ! `seg_sc_ms_lesion_stir_psir` works great ! 🚀 
Thanks @joshuacwnewton for your work ! 😃

joshuacwnewton:
> Maybe something to add?

Definitely. Right now, a lot of tests specify `-ofolder` (in order to output the files to a tmpdir). But, since `sct_deepseg` doesn't have an `-ofolder` (only an `-o`), the tests have to manually specify the output name, which overrides the "fallback output filename" behavior that was buggy.

Issue #3325, opened 2021-04-01
Introduce QC report for sct_deepseg


jcohenadad:
As done for sct_deepseg_sc

joshuacwnewton:
Duplicate of #4436.

Issue #4204, opened 2023-08-24
QC code revamp


mguaypaq:
This issue is to keep track of some ongoing work for refactoring the QC report generation code in SCT.

The goal is to make it much easier to fix bugs and add new features. Recent changes (#4102, #4129, #4126, #4133, #4151, #4189, #4193) and proposed changes (#4011, #4122, #4166, [PAM50-normalized-metrics#10](https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/pull/10)) seem like they are harder than they should be because of the current code structure.

Some of the points that I want to preserve during this revamp are detailed under the following section headings.

# Directory structure (no changes)

To help fix the scope of this revamp, I'm going to say that I don't want to change the directory structure of the output inside `path_qc`. In particular, this means:
- the front-end assets go under `_assets/`
- the json files go under `_json/`
- the QC images go under `{dataset}/{subject}/{contrast}/{process}/{timestamp}/`.

Abbreviated example:
```
path_qc
├── _assets
│   ├── css
│   ├── fonts
│   ├── imgs
│   └── js
├── dataset
│   └── subject
│       └── contrast
│           └── sct_deepseg_sc
│               └── 2023_08_24_143426.629219
│                   ├── background_img.png
│                   └── overlay_img.png
├── index.html
├── _json
│   ├── qc_2023_08_24_143337.744815.json
│   └── qc_2023_08_24_143426.629219.json
```

# Front-end (no changes)

Also to reduce scope, I don't want to change the front-end code at all (namely, the html, js, css files under `_assets/`). This sets the API boundary at:
- generating a pair of files `background_img.png` and `overlay_img.png`, and
- generating a json file with the following fields:
  - `moddate`
  - `dataset`
  - `subject`
  - `cwd`
  - `fname_in`
  - `contrast`
  - `command`
  - `cmdline`
  - `background_img`
  - `overlay_img`
  - `plane`
  - `sct_version`
  - `qc` (empty)

# Locking (no changes)

It's important to preserve the fact that many QC-generating commands can be run in parallel without corrupting the QC report. The current mechanism is that we open the `index.html` file and lock it with `portalocker`, then we do the following operations **while holding the lock**:
- Generating a new json file with the current report.
- Reading all the existing json files from previous reports.
- Re-generating the `index.html` file itself.
- Re-copying all the files under `_assets/`.

[...]

PR #4446, opened 2024-04-19
Add QC report for `sct_deepseg`

### LINKED ISSUE #4436: Add QC feature for `sct_deepseg`
As more and more people are going to use the contrast-agnostic model (among other models), we should probably start adding the QC feature

joshuacwnewton:
## Description

This works, but I've marked it as draft because it's in a bit of a messy state right now. (I just wanted to get a prototype going before bed!)

softseg (e.g. contrast agnostic):

image

SC/lesion seg (e.g. SCIseg) -- note the blue lesion in the bottom-left:

image

Notes:

- The colormaps probably need a bit of work (the matplotlib colormaps have a large contrast + very bright light colors)
- I wanted to focus on the axial mosaic to begin with, because that's what's likely most pressing for ISMRM/showing off the contrast agnostic seg.

## Linked issues

Fixes #4436.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3325.


jcohenadad:
This is fantastic! Thank you @joshuacwnewton 

Would it be possible to binarize the segmentations before QCing them maybe?

joshuacwnewton:
> Would it be possible to binarize the segmentations before QCing them maybe?

Absolutely! 

That said, I'm realizing that we previously solved the colormap problem for softsegs in QCs by using only 2 colors of red in a custom colormap:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8f2b0e0eb956cbd12d1cabdfbee34e1ca8d99c34/spinalcordtoolbox/reports/qc.py#L63

Since these two colors are closer together (`#4D0000`, `#FF0000`), it basically looks like a binarized segmentation, but with a bit of color change for the "soft" regions:

image

(If this is too stark a difference, we can even adjust the colors to be closer together to make the "soft" effect even more subtle.)

spinalcordtoolbox/reports/qc2.py
```diff
@@ -273,6 +274,116 @@ def sct_register_multimodal(
...
+        path_input=Path(fname_input).absolute(),
+        path_qc=Path(path_qc),
+        command=command,
+        cmdline=list2cmdline(cmdline),
+        plane='Axial',
+        dataset=dataset,
+        subject=subject,
+    ) as imgs_to_generate:
+        # FIXME: This code is more or less duplicated with the 'sct_register_multimodal' report, because both reports
+        #        use the old qc.py method "_make_QC_image_for_3d_volumes" for generating the background img.
```
joshuacwnewton, line 307: The amount of boilerplate and duplication between this `qc2.py` module and @mguaypaq's `sct_register_multimodal` addition is a little concerning to me. But, I'm hopeful that now we have a second data point, we can start to identify places where we can factor out the boilerplate.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -273,6 +274,116 @@ def sct_register_multimodal(
...
+        # Resample images slice by slice
+        p_resample = {
+            'human': 0.6, 'mouse': 0.1,
+        }[species]
```
joshuacwnewton, line 311: Just to demonstrate what happens when I try to create a QC report for mouse data using the human-centric 0.6mm resampling...

Before:

image

After:

image

(Ignore the poor QC quality, that has to do with the input data.)

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -253,6 +270,29 @@ def main(argv: Sequence[str]):
...
+        # Models can have multiple input images -- create 1 QC report per input image.
+        # Models can also either have 1 OR 2 outputs per input, so we may need to split output_filenames into 2 lists
+        if len(output_filenames) == len(input_filenames):
+            iterator = zip(input_filenames, output_filenames, [None] * len(input_filenames))
+        else:
+            assert len(output_filenames) == 2 * len(input_filenames)
+            iterator = zip(input_filenames, output_filenames[0::2], output_filenames[1::2])
+
+        # Create one QC report per input image, with one or two segs per image
+        species = 'mouse' if any(s in arguments.task[0] for s in ['mouse', 'mice']) else 'human'  # used for resampling
```
joshuacwnewton, line 283: This made me realize that we use both `mouse` and `mice` across model names. :sweat_smile:

joshuacwnewton (commented):
OK, I think this is ready for review!

Gosh, though. Transferring reports from `qc.py` to `qc2.py` is _tedious_. With the old QC, there are so many different places that you have to look for various bits of image processing (`generate_qc()`, `QcImage`, `Slice`, etc.) and the trace bounces from module to module.

I'm very happy that `qc2.py` has very straightforward, linear, single-function approach to generating the required images.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton, line 1: NB: I didn't add this new QC report to `sct_qc`, because:

1. I didn't want to mix `qc2.py` and `qc.py` in our QC-specific script.
3. I figure that if folks are using `sct_qc` to generate reports (rather than the built-in per-CLI QC), then they're skilled enough to be able to re-use the existing `-p sct_deepseg_sc` or `-p sct_deepseg_lesion` to generate whichever reports they desire.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -260,6 +277,29 @@ def main(argv: Sequence[str]):
...
         fname_prior = fname_seg
 
+    if arguments.qc is not None:
+        # Models can have multiple input images -- create 1 QC report per input image.
+        # Models can also either have 1 OR 2 outputs per input, so we may need to split output_filenames into 2 lists
+        if len(output_filenames) == len(input_filenames):
+            iterator = zip(input_filenames, output_filenames, [None] * len(input_filenames))
+        else:
+            assert len(output_filenames) == 2 * len(input_filenames)
+            iterator = zip(input_filenames, output_filenames[0::2], output_filenames[1::2])
```
mguaypaq, line 287: Neat!

mguaypaq (approved):
Thanks for this new QC report! I'm glad that the `qc2.py` structure was straightforward to use.

jcohenadad:
Nitpick: It would be great to also add the entry `sct_deepseg` in `sct_qc`

joshuacwnewton:
> Nitpick: It would be great to also add the entry `sct_deepseg` in `sct_qc`

The self-review comment was hidden, but I added a justification for this [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4446#discussion_r1577979078).

My thinking was that:

- If someone wants to display an SC seg, they would use the existing `-p sct_deepseg_sc` report.
- If someone wants to display lesion seg, they would use the existing `-p sct_deepseg_lesion` report.
- Long term (probably for v6.4/v6.5), all of the "`-p`" QC reports would be replaced with more flexible options (e.g. `-type {seg,softseg}` `-plane {Axial/Sagittal}`, etc.) as planned in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4204.

Issue #4453, opened 2024-04-24
Integration of EPI segmentation nnUNet model in SCT


rohanbanerjee:
I am opening this issue to inform the integration of the EPI spinal cord segmentation 3D nnUNetv2 model into SCT. The latest trained model can be found in the `v0.2` release here: https://github.com/sct-pipeline/fmri-segmentation/releases/download/v0.2/model-fmri-segmentation-v0.2.zip. I will opening up a PR for the integration. 

I am also uploading a sample image from my dataset for testing:

[spinefmri_sub-hamburgP23_task-pain_bold_0000.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/15095700/spinefmri_sub-hamburgP23_task-pain_bold_0000.nii.gz)

mguaypaq:
Were you hoping to include this in the SCT v6.3 release, which we're doing today/tomorrow? Or can this wait until SCT v6.4?

rohanbanerjee:
I'm in the process of pushing the PR now. If it's possible, we can go ahead with the v6.3 release itself otherwise v6.4 is okay too. Some people at ISMRM would find this use too (though I am not going) and we could start with the alpha testing then.

mguaypaq:
Ok, if it's ready today, and if it's useful for ISMRM, then we can include it.

PR #4454, opened 2024-04-24
Adding the EPI spinal cord segmentation model to SCT

rohanbanerjee:
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4453
## Description

Opening this PR to add the 3D nnUNet model to segment spinal cord on EPI data. The trained model can be found in this release: https://github.com/sct-pipeline/fmri-segmentation/releases/tag/v0.2

## Testing the model

To test this PR, you will need to install SCT from this branch. Here are
the steps:

1. Open a terminal, then run `git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git` to get a
copy of the source code.
2. Run `cd spinalcordtoolbox` to enter the SCT source code folder.
3. Run `git checkout rb/4453-seg-sc-epi-nnunet` to switch to this
branch.
4. Run `./install_sct` to install SCT (the installation will exist
in this same folder). Make sure that you see all green `[OK]` at the end
of the installation.
5. Run `sct_deepseg -install-task seg_sc_epi` to
download the model.

Sample image:
[spinefmri_sub-hamburgP23_task-pain_bold_0000.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/15097396/spinefmri_sub-hamburgP23_task-pain_bold_0000.nii.gz)

6. Download the sample image and run run `sct_deepseg -task seg_sc_epi -i <IMAGE_PATH> -c bold` to
segment an image.


spinalcordtoolbox/deepseg/models.py
```diff
@@ -342,6 +351,15 @@
                              'the Canadian Prospective Cohort Study (CanProCo).',
          'url': 'https://github.com/ivadomed/canproco',
          'models': ['model_seg_sc_lesion_canproco_nnunet']},
+    'seg_sc_epi':
+        {'description': 'Spinal cord segmentation for EPI data',
```
joshuacwnewton, line 355: I'm not 100% sure about the terminology, but I notice that the contrast is `BOLD`, and the repo also mentions "fMRI". I also see the term "EPI-BOLD fMRI" in some papers like [this](https://pubmed.ncbi.nlm.nih.gov/14696216/).

So, maybe it's also worth mentioning this in the task name and short description?

```suggestion
    'seg_sc_epi-bold_fmri':
        {'description': 'Spinal cord segmentation for EPI-BOLD fMRI data',
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -342,6 +351,15 @@
...
+         'long_description': 'This segmentation model for spinal cord on EPI data uses a 3D nnUNet model from '
+                             'the nnUNetv2 framework. The training data consists of 3D images (n=192) spanning numerous resolutions '
+                             'from multiple sites like Max Planck Institute for Human Cognitive and Brain Sciences - Leipzig, '
+                             ' University of Geneva, Stanford University, Kings College London, Universitätsklinikum Hamburg. '
+                             'The model has been trained in a human-in-the-loop active learning fashion. ',
```
joshuacwnewton, line 360: Slight spacing nitpick:

```suggestion
         'long_description': 'This segmentation model for spinal cord on EPI data uses a 3D nnUNet model from '
                             'the nnUNetv2 framework. The training data consists of 3D images (n=192) spanning numerous resolutions '
                             'from multiple sites like Max Planck Institute for Human Cognitive and Brain Sciences - Leipzig, '
                             'University of Geneva, Stanford University, Kings College London, Universitätsklinikum Hamburg. '
                             'The model has been trained in a human-in-the-loop active learning fashion.',
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -342,6 +351,15 @@
                              'the Canadian Prospective Cohort Study (CanProCo).',
          'url': 'https://github.com/ivadomed/canproco',
          'models': ['model_seg_sc_lesion_canproco_nnunet']},
+    'seg_sc_epi':
+        {'description': 'Spinal cord segmentation for EPI data',
+         'long_description': 'This segmentation model for spinal cord on EPI data uses a 3D nnUNet model from '
+                             'the nnUNetv2 framework. The training data consists of 3D images (n=192) spanning numerous resolutions '
```
joshuacwnewton, line 357: Is it worth mentioning whether these are healthy controls or pathological?

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: We do have some fMRI data in our `sct_testing_data` repo, so I was curious if it would be suitable to use in a test. We already have a test set up for `sct_deepseg` models -- all that's needed is specifying the task name and input data, basically. For example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/953727265ad490bdda976d6c60afa60ec826cc87/testing/cli/test_sct_deepseg.py#L51-L58

(Given that this is shortly before a new SCT release, if the test data isn't suitable, then I don't think it's worth rushing to try and add new testing data and create a new `sct_testing_data` release.)

I'm just reinstalling SCT right now, but I'll try it out ASAP (along with the test image you provided) and follow up here.

joshuacwnewton (changes requested):
Looking at the code additions, this mostly LGTM! My feedback here is pretty minor?

I still need to actually test the model, but I wanted to get some quick feedback out while reinstalling SCT.

spinalcordtoolbox/deepseg/models.py
jcohenadad: I'm not a huge fan of "-" in task name, because the way i install them is:
- run help
- identify the task
- double click on task and copy
- paste in syntax with flag `-install`

with the "-" the double click will faile

jcohenadad: the comment about the BOLD contrast if valid-- however in this case, i don't think we should alter the task name

moreover, the distinction is really about gre (gradient echo, used for fMRI) vs. spin echo EPI (used for diffusion)-- 

this comment can go in the long description

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Ah! Our fMRI test data is 4D, which results in the following failure:

<details>

```
rb/4453-seg-sc-epi-nnunet|joshua@monarch:~/repos/spinalcordtoolbox/data/sct_testing_data/fmri$ sct_deepseg -task seg_sc_epi -i fmri.nii.gz 

--
Spinal Cord Toolbox (git-rb/4453-seg-sc-epi-nnunet-31ba9a68c3a59543718a947c9dbae1e9ea063cde)

sct_deepseg -task seg_sc_epi -i fmri.nii.gz
--

perform_everything_on_device=True is only supported for cuda devices! Setting this to False
Running inference on device: cpu
/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/utilities/plans_handling/plans_handler.py:37: UserWarning: Detected old nnU-Net plans format. Attempting to reconstruct network architecture parameters. If this fails, rerun nnUNetv2_plan_experiment for your dataset. If you use a custom architecture, please downgrade nnU-Net to the version you implemented this or update your implementation + plans.
  warnings.warn("Detected old nnU-Net plans format. Attempting to reconstruct network architecture "
Model loaded successfully. Fetching test data...
Creating temporary folder (/tmp/sct_2024-04-24_12-58-58_sct_deepseg_czyve0vb)
Copied fmri.nii.gz to /tmp/sct_2024-04-24_12-58-58_sct_deepseg_czyve0vb/fmri.nii.gz
Changing orientation of the input to the model orientation (LPI)...
Starting inference...
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 272, in <module>
    main(sys.argv[1:])
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 230, in main
    im_lst, target_lst = inference.segment_non_ivadomed(path_model, model_type, input_filenames, thr,
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 125, in segment_non_ivadomed
    fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device)
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 240, in segment_nnunet
    data = img_in.data.transpose([2, 1, 0])
ValueError: axes don't match array
```

</details>

I notice that the sample image you provided is 3D. Is EPI-BOLD data typically 3D, or is this a single 3D volume of a 4D sequence? 

(If 4D sequences are typical, then we may want to consider adapting `sct_deepseg` to run on 4D volumes so that users won't have to split their data, run `sct_deepseg` a bunch, then merge the data. But, that can be done in a follow-up PR if necessary.)

Anyway, if I use `sct_image -i fmri.nii.gz -split t` first on the fMRI image in `sct_testing_data, then run `sct_deepseg` on a single volume, I get:

image

So, it looks like writing a test should be pretty easy! (There is no ground truth seg in `sct_testing_data` right now, though.)

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Thank you for the feedback! Very good point about `-`, and I appreciate the insight re: dMRI/fMRI terminology.

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: Oh yes, I missed on mentioning that these are mean images (and therefore 3D) that we can run the inference on.

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: Reverted now

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: Thanks for the feedback. Made the required changes.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -186,7 +186,7 @@
          "url": [
              "https://github.com/sct-pipeline/fmri-segmentation/releases/download/v0.2/model-fmri-segmentation-v0.2.zip"
          ],
-         "description": "Spinal cord segmentation for EPI data using nnUNetv2 model",
+         "description": "Spinal cord segmentation for mean EPI data using nnUNetv2 model",
```
jcohenadad, line 189:
```suggestion
         "description": "Spinal cord segmentation for EPI data (single 3D volume)",
```

data don't necessarily need to be average-- eg model works on single time point

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: But as it stands, some might miss out on the documentation and end up running the inference on the 4D images. Do you think mentioning this `sct_image -i fmri.nii.gz -split t` would be a better option for now?

spinalcordtoolbox/deepseg/models.py
```diff
@@ -353,7 +353,7 @@
          'models': ['model_seg_sc_lesion_canproco_nnunet']},
     'seg_sc_epi':
         {'description': 'Spinal cord segmentation for EPI-BOLD fMRI data',
-         'long_description': 'This segmentation model for spinal cord on EPI data uses a 3D nnUNet model from '
+         'long_description': 'This segmentation model for spinal cord on mean EPI data uses a 3D nnUNet model from '
```
jcohenadad, line 356:
```suggestion
         'long_description': 'This segmentation model for spinal cord on EPI data (single 3D volume) uses a 3D UNet model built from '
```

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: Indeed, thank you for pointing that out

spinalcordtoolbox/deepseg/models.py
rohanbanerjee: @joshuacwnewton -- This commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4454/commits/3e87480324025cce76dd56042fe2946321c49cf4 adds the above requirement.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -232,6 +233,16 @@ def main(argv: Sequence[str]):
                 if contrast != "stir":
                     parser.error("Task 'seg_sc_ms_lesion_stir_psir' requires the flag `-c` to be either psir or stir.")
 
+        if 'seg_sc_epi' in arguments.task[0]:
+            for image in arguments.i:
+                image_load = nb.load(image)
```
joshuacwnewton, line 238: I don't think it's necessary to use `nibabel` to load the image, given that `sct_deepseg` already imports SCT's own image-loading utility (the `Image` class).

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -232,6 +233,16 @@ def main(argv: Sequence[str]):
...
+                image_shape = image_load.shape
+                if len(image_shape) == 4:
+                    parser.error("Only 3D vloumes supported for this task. You can either provide a mean volume or a single time point")
+                else:
+                    input_filenames = arguments.i.copy()
```
joshuacwnewton, line 243: Is there a reason for duplicating this line, since `input_filenames` is already created earlier? I'm wondering, if we exclude it, will the processing be basically the same?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -232,6 +233,16 @@ def main(argv: Sequence[str]):
                 if contrast != "stir":
                     parser.error("Task 'seg_sc_ms_lesion_stir_psir' requires the flag `-c` to be either psir or stir.")
 
+        if 'seg_sc_epi' in arguments.task[0]:
+            for image in arguments.i:
+                image_load = nb.load(image)
+                image_shape = image_load.shape
+                if len(image_shape) == 4:
+                    parser.error("Only 3D vloumes supported for this task. You can either provide a mean volume or a single time point")
```
joshuacwnewton, line 241: It might be worth specifying the exact SCT commands that users can use to create these images:

```suggestion
                    parser.error("Only 3D volumes are supported for this task. You can either provide a mean volume "
                                 "(using 'sct_maths -mean') or a single time point (using 'sct_image -split t'.")
```

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -232,6 +233,16 @@ def main(argv: Sequence[str]):
...
+                else:
+                    input_filenames = arguments.i.copy()
+
+
```
joshuacwnewton, line 245: The automated "Code style quality" check has caught some linting issues, so I'd recommend checking the failing test and applying its suggestions. :)

joshuacwnewton (changes requested):
Thank you so much for adding the 4D check! I had a bit of feedback on the changes. :relaxed:

spinalcordtoolbox/scripts/sct_deepseg.py
rohanbanerjee: Let me check on this

spinalcordtoolbox/scripts/sct_deepseg.py
rohanbanerjee: You're right, fixed this!

spinalcordtoolbox/scripts/sct_deepseg.py
rohanbanerjee: Again, that sounds perfect. Removed this!

joshuacwnewton: Ah, oops! It looks like the wrong one might have been removed, as we have a new error now: 

```
UnboundLocalError: local variable 'input_filenames' referenced before assignment
```

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -283,7 +290,6 @@ def main(argv: Sequence[str]):
             # If output folder does not exist, create it
             if not (path_out == '' or os.path.exists(path_out)):
                 os.makedirs(path_out)
-
```
joshuacwnewton, line 286: This is just a nitpick, but ideally there would be no "empty diffs" like this, that way we only change the lines that are absolutely necessary for the PR.

(The reason for this is so that the "blame view" (see [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blame/b80ab5374ca6778d5f816672a67802ee2caa7cbb/spinalcordtoolbox/scripts/sct_deepseg.py#L241)) remains clean, so that we can properly identify the history of each line of code.)

rohanbanerjee: Oh I did this because this was getting caught in the `code quality` check. And it passed after the removal😅

joshuacwnewton: I think what happened is that... before, there was 1 empty line, and the code quality check found 2 empty lines, but the "fix" was 0 empty lines. :sweat_smile: 

Keeping the same amount of empty lines should make the "diff" go away. :blush:

spinalcordtoolbox/scripts/sct_deepseg.py
rohanbanerjee: Hmm, it works on my end. Do you still get the error?

joshuacwnewton: The error occurs in the test suite for non-EPI models, hehe: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/8822601995/job/24221178754?pr=4454#step:5:728

rohanbanerjee: Oh yeah, found it!

joshuacwnewton (approved):
Thank you so much for the prompt response to feedback!! LGTM! :hearts:

PR #4460, opened 2024-04-25
Switch from two-tone colormap to transparency for `sct_deepseg` QC

### LINKED ISSUE #4457: What colormap should be used for soft segmentation QCs? (single color vs. gradient)
While playing with the new QC sct_deepseg feature (yay!), I noticed the segmentations are soft:

image

Is this desirable?

joshuacwnewton:
## Description

> [!NOTE]
> Terribly sorry for the last minute PR, right before the SCT release! But, it's important that we have a good-looking QC when showing off the new contrast agnostic model. :)

This PR is a small amendment to #4446. Using a dark color for values <1.0 was making it much more difficult to QC the soft perimeter of softsegs.

This solution scales the transparency to the image values. If the segmentation == 1.0, then the image will be opaque red (as it already was). But, as the values get closer to 0, the voxels will get more and more transparent. This will prevent low values (e.g. 0.01) from showing up in the QC report.

Sample output for the SCI SC+lesion seg:

image

## Linked issues

Fixes #4457.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -316,21 +316,18 @@ def sct_deepseg(
         fig.set_size_inches(*size_fig, forward=True)
         FigureCanvas(fig)
         ax = fig.add_axes((0, 0, 1, 1))
-        colormaps = {
-            # Single color (Red, Cyan)
-            'seg': ["#ff0000", "#00ffff"],
-            # Two-tone colormap (Red-darkred, Cyan-darkcyan)
-            'softseg': [color.ListedColormap(["#4d0000", "#ff0000"]),
-                        color.ListedColormap(["#004d4d", "#00ffff"])],
-        }
+        colormaps = [color.ListedColormap(["#ff0000"]),  # Red for first image
+                     color.ListedColormap(["#00ffff"])]  # Cyan for second
         for i, image in enumerate([img_seg_sc, img_seg_lesion]):
             if not image:
                 continue
             img = mosaic(image, centers)
             img = np.ma.masked_equal(img, 0)
             ax.imshow(img,
-                      cmap=(colormaps['seg'] if check_image_kind(image.data) == 'seg' else colormaps['softseg'])[i],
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/reports/qc2.py
```diff
@@ -316,21 +316,18 @@ def sct_deepseg(
...
             if not image:
                 continue
             img = mosaic(image, centers)
             img = np.ma.masked_equal(img, 0)
             ax.imshow(img,
-                      cmap=(colormaps['seg'] if check_image_kind(image.data) == 'seg' else colormaps['softseg'])[i],
+                      cmap=colormaps[i],
                       norm=color.Normalize(vmin=0.5, vmax=1),
+                      # img==1 -> opaque, but soft regions -> more transparent as value decreases
+                      alpha=(img / img.max()),  # scale to [0, 1]
```
mguaypaq, line 330: If we really want to make sure we've scaled to [0, 1], maybe line 325 should mask out negative values as well (even though we don't expect them):
```py
img = np.ma.masked_less_equal(img, 0)
img.set_fill_value(0)
```
(I'm not sure what impact the `fill_value` has, but `masked_equal` sets it differently from `masked_less_equal`.)

mguaypaq (approved):
This looks great (both the code and the images)!

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: > (I'm not sure what impact the `fill_value` has, but `masked_equal` sets it differently from `masked_less_equal`.)

Wow, good catch!

```python
>>> img.fill_value
1e+20
```

That said, the `set_fill_value` doesn't seem to change the QC report itself? So, I too am left wondering what impact `fill_value` has, and it seems like we're not alone:

- https://github.com/numpy/numpy/issues/20850

I added my 2 cents in a follow-up comment, since it seems like multiple folks are getting tripped up by this behavior!!

joshuacwnewton:
I've gotten the OK from Julien, so we're good to go here. :)

PR #4466, opened 2024-05-02
Track `sct_deepseg` model provenance with `source.json` (in model folder) and JSON sidecar (in output)

### LINKED ISSUE #3394: Output JSON sidecar files after processing a la BIDS

### LINKED ISSUE #4465: Indicate versions of `sct_deepseg` models
## Related issues

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1572
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2489
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2788
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2959

## Context

As we are more actively training and updating models for `sct_deepseg`, the question of versioning becomes important.

Example: When listing a task, there is a link to the GitHub repository of the model, but no information about the model version:

```console
sct_deepseg -list-tasks

--
Spinal Cord Toolbox (git-master-72ff27030efda92431ee831b45f31e2858b8552d)

sct_deepseg -list-tasks
--

[...]

TASK:             seg_sc_contrast_agnostic
CONTRAST:         any            
MODELS:           model_seg_sc_contrast_agnostic_softseg_monai
LONG_DESCRIPTION: This model for contrast agnostic spinal cord segmentation uses an nnUNet architecture, and was created with
                  the MONAI package. Training data consists of healthy controls from the open-source Spine Generic Multi Subject
                  database and pathologies from private datasets including DCM and MS patients. Segmentation has been tested
                  with the following contrasts: [T1w, T2w, T2star, MTon_MTS, GRE_T1w, DWI, mp2rage_UNIT1], but other contrasts
                  that are close visual matches may also work well with this model.
URL:              https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/
INSTALLED:        Yes
```

Even more confusing: from SCT's git SHA (72ff27030efda92431ee831b45f31e2858b8552d), I see that v2.3 of the model is supposed to be installed:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/72ff27030efda92431ee831b45f31e2858b8552d/spinalcordtoolbox/deepseg/models.py#L141-L143

However, because I updated SCT without re-installing the model, the model could in fact be from an older version.

I verified I had the latest version by comparing the training logs:
```console
julien-macbook:~/code/spinalcordtoolbox/data/deepseg_models/model_seg_sc_contrast_agnostic_softseg_monai $ cat logs.txt 
[...]
2024-04-12 04:51:55.513 | INFO     | __main__:on_test_epoch_end:477 - Test (Soft) Dice: 0.8189144134521484
2024-04-12 04:51:55.513 | INFO     | __main__:on_test_epoch_end:478 - Test (Hard) Dice: 0.9547366771735345
2024-04-12 04:51:56.887 | INFO     | __main__:main:787 - TESTING DONE!
```

vs. freshly downloaded `model_soft_bin_20240410-1136`:
```console
julien-macbook:~/code/spinalcordtoolbox/data/deepseg_models/model_seg_sc_contrast_agnostic_softseg_monai $ cat /Users/julien/Desktop/model_soft_bin_20240410-1136/logs.txt 
[...]
2024-04-12 04:51:55.513 | INFO     | __main__:on_test_epoch_end:477 - Test (Soft) Dice: 0.8189144134521484
2024-04-12 04:51:55.513 | INFO     | __main__:on_test_epoch_end:478 - Test (Hard) Dice: 0.9547366771735345
2024-04-12 04:51:56.887 | INFO     | __main__:main:787 - TESTING DONE!
```

But we cannot ask our users to go through this verification step 😅.

## Solution

Display the model version in the list (and with https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3394, also output that info in the JSON)

joshuacwnewton:
## Description

`source.json` sample:

```json
{
    "model_urls": {
        "t2star_sc": [
            "https://github.com/ivadomed/t2star_sc/releases/download/r20231004/r20231004_t2star_sc.zip",
        ]
    }
}
```

JSON sidecar sample (for `test_segment_nifti`):

```json
{
    "GeneratedBy": [
        {
            "Name": "spinalcordtoolbox: sct_deepseg -i t2s.nii.gz -task seg_sc_t2star -o t2s_seg_deepseg.nii.gz -qc qc -thr 0.9",
            "Version": "git-jn/4465-track-model-provenance-80171e6d249befbb23e4bdd41c759b870c08fac5*",
            "CodeURL": "https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/80171e6d249befbb23e4bdd41c759b870c08fac5/spinalcordtoolbox/scripts/sct_deepseg.py",
            "ModelURL": "https://github.com/ivadomed/t2star_sc/releases/download/r20231004/r20231004_t2star_sc.zip"
        }
    ]
}
```

## Linked issues

Fixes #4465.
Fixes #3394.

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -451,9 +451,10 @@ def is_up_to_date(path_model):
     # Single-seed models
     if isinstance(expected_model_urls, list) and isinstance(actual_model_urls, list):
         # if 'actual_model_urls' is up-to-date, then subtracting the current URLs should result in an empty set
-        return not (set(actual_model_urls) - set(expected_model_urls))
+        if set(actual_model_urls) - set(expected_model_urls):
+            return False
     # Multi-seed, ensemble models
-    if isinstance(expected_model_urls, dict) and isinstance(actual_model_urls, dict):
+    elif isinstance(expected_model_urls, dict) and isinstance(actual_model_urls, dict):
         for seed, url in actual_model_urls.items():
             if seed not in expected_model_urls:
                 logger.warning(f"unexpected seed: {seed}")
@@ -464,7 +465,11 @@ def is_up_to_date(path_model):
         if expected_model_urls:
             logger.warning(f"missing seeds: {list(expected_model_urls.keys())}")
             return False
-        return True
+    else:
+        logger.warning("Mismatch between 'source.json' URL format and SCT source code URLs")
+        return False
+    logger.info(f"Model '{model_name}' is up to date")
+    return True
 
 
 def is_valid(path_models):
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -431,17 +431,24 @@ def install_default_models():
 
 
 def is_up_to_date(path_model, name_model):
+    """
+    Determine whether an on-disk model is up-to-date by comparing
+    the URL used to download the model with the latest mirrors.
+
+    :return: bool: whether the model is up-to-date
+    """
     source_path = os.path.join(path_model, "source.json")
     if os.path.isfile(source_path):
         with open(source_path, "r") as fp:
             source_dict = json.load(fp)
     else:
         logger.warning("Provenance file 'source.json' missing!")
-        source_dict = {'model_urls': {name_model: []}}  # empty list will force reinstall
+        source_dict = {'model_urls': {name_model: ['sample_url']}}  # dummy list will force reinstall
     source_urls = source_dict["model_urls"][name_model]
     newest_urls = MODELS[name_model]['url']
-    missing_urls = set(newest_urls) - set(source_urls)
-    return not missing_urls
+    # if 'source_urls' is up-to-date, then subtracting the current URLs should result in an empty set
+    outdated_urls = set(source_urls) - set(newest_urls)
+    return not outdated_urls
 
 
 def is_valid(path_models):
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -407,6 +407,16 @@ def install_model(name_model):
         for seed_name, model_urls in url_field.items():
             logger.info(f"\nInstalling '{seed_name}'...")
             download.install_data(model_urls, folder(os.path.join(name_model, seed_name)), keep=True)
+        model_urls = sum(url_field.values(), [])  # combine seed models into 1 list of urls
+
+    # Write `source.json` (for model provenance / updating)
+    source_dict = {
+        'model_urls': {
+            name_model: model_urls
+        }
+    }
+    with open(os.path.join(folder(name_model), "source.json"), "w") as fp:
+        json.dump(source_dict, fp, indent=4)
 
 
 def install_default_models():
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -446,7 +446,7 @@ def is_up_to_date(path_model):
     if model_name not in MODELS:
         logger.warning(f"Model name '{model_name}' from source.json does not match model names in SCT source code.")
         return False
-    expected_model_urls = MODELS[model_name]['url']
+    expected_model_urls = MODELS[model_name]['url'].copy()
     actual_model_urls = source_dict["model_urls"]
     # Single-seed models
     if isinstance(expected_model_urls, list) and isinstance(actual_model_urls, str):
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -404,11 +404,11 @@ def install_model(name_model):
     else:
         if not isinstance(url_field, dict):
             raise ValueError("Invalid url field in MODELS")
+        urls_used = {}
         for seed_name, model_urls in url_field.items():
             logger.info(f"\nInstalling '{seed_name}'...")
-            download.install_data(model_urls, folder(os.path.join(name_model, seed_name)), keep=True)
-        urls_used = sum(url_field.values(), [])  # combine seed models into 1 list of urls
-
+            urls_used[seed_name] = download.install_data(model_urls,
+                                                         folder(os.path.join(name_model, seed_name)), keep=True)
     # Write `source.json` (for model provenance / updating)
     source_dict = {
         'model_name': name_model,
@@ -446,11 +446,25 @@ def is_up_to_date(path_model):
     if model_name not in MODELS:
         logger.warning(f"Model name '{model_name}' from source.json does not match model names in SCT source code.")
         return False
-    newest_urls = MODELS[model_name]['url']
-    source_urls = source_dict["model_urls"]
-    # if 'source_urls' is up-to-date, then subtracting the current URLs should result in an empty set
-    outdated_urls = set(source_urls) - set(newest_urls)
-    return not outdated_urls
+    expected_model_urls = MODELS[model_name]['url']
+    actual_model_urls = source_dict["model_urls"]
+    # Single-seed models
+    if isinstance(expected_model_urls, list) and isinstance(actual_model_urls, list):
+        # if 'actual_model_urls' is up-to-date, then subtracting the current URLs should result in an empty set
+        return not (set(actual_model_urls) - set(expected_model_urls))
+    # Multi-seed, ensemble models
+    if isinstance(expected_model_urls, dict) and isinstance(actual_model_urls, dict):
+        for seed, url in actual_model_urls.items():
+            if seed not in expected_model_urls:
+                logger.warning(f"unexpected seed: {seed}")
+                return False
+            if url not in expected_model_urls.pop(seed):
+                logger.warning(f"wrong version for {seed}: {url}")
+                return False
+        if expected_model_urls:
+            logger.warning(f"missing seeds: {list(expected_model_urls.keys())}")
+            return False
+        return True
 
 
 def is_valid(path_models):
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -430,6 +430,20 @@ def install_default_models():
             install_model(name_model)
 
 
+def is_up_to_date(path_model, name_model):
+    source_path = os.path.join(path_model, "source.json")
+    if os.path.isfile(source_path):
+        with open(source_path, "r") as fp:
+            source_dict = json.load(fp)
+    else:
+        logger.warning("Provenance file 'source.json' missing!")
+        source_dict = {'model_urls': {name_model: []}}  # empty list will force reinstall
+    source_urls = source_dict["model_urls"][name_model]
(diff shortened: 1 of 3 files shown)


spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -314,6 +320,29 @@ def main(argv: Sequence[str]):
...
+        source_path = os.path.join(path_model, "source.json")
+        if os.path.isfile(source_path):
+            with open(source_path, "r") as fp:
+                source_dict = json.load(fp)
+        sidecar_json['GeneratedBy'].append(
+            {
+                'Name': f'sct_deepseg -task {arguments.task}',
+                "Version": source_dict["model_urls"][name_model][0]
+            }
+        )
```
joshuacwnewton, line 342: This shouldn't be a second `GeneratedBy` entry. Instead, the BIDS spec allows custom fields, so we can differentiate between `CodeURL` vs. `ModelURL`, and `SCTVersion` vs. `ModelVersion`.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -407,6 +407,16 @@ def install_model(name_model):
         for seed_name, model_urls in url_field.items():
             logger.info(f"\nInstalling '{seed_name}'...")
             download.install_data(model_urls, folder(os.path.join(name_model, seed_name)), keep=True)
+        model_urls = sum(url_field.values(), [])  # combine seed models into 1 list of urls
+
+    # Write `source.json` (for model provenance / updating)
+    source_dict = {
+        'model_urls': {
+            name_model: model_urls
```
joshuacwnewton, line 415: Rather than storing all `model_urls`, we should only store the ones that were actually used when downloading.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -420,6 +430,20 @@ def install_default_models():
...
+    if os.path.isfile(source_path):
+        with open(source_path, "r") as fp:
+            source_dict = json.load(fp)
+    else:
+        logger.warning("Provenance file 'source.json' missing!")
+        source_dict = {'model_urls': {name_model: []}}  # empty list will force reinstall
+    source_urls = source_dict["model_urls"][name_model]
+    newest_urls = MODELS[name_model]['url']
+    missing_urls = set(newest_urls) - set(source_urls)
+    return not missing_urls
```
joshuacwnewton, line 444: Instead of checking for missing URLs (then redownloading if there is a missing URL), we should instead check if the model URL that was used (see prev comment) is present in the model dict, and redownload if the URL that was used is not one of the current urls.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -314,6 +320,29 @@ def main(argv: Sequence[str]):
         # Use the result of the current model as additional input of the next model
         fname_prior = fname_seg
 
+        # write JSON sidecar file
+        sidecar_json = {
+            'GeneratedBy': [
+                {
+                    'Name': 'spinalcordtoolbox',
+                    "Version": __version__,
```
joshuacwnewton, line 328: `Name` should be the "Pipeline used", so it was suggested to do:

```json
    "Name": "spinalcordtoolbox: sct_deepseg -i <> -task <> [...]",
```

joshuacwnewton: We may also want to consider scrubbing the full paths from the input for privacy reasons (e.g. by calling `basename`)

joshuacwnewton (commented):
Small self-review to summarize discussion from SCT meeting

joshuacwnewton:
I think I've implemented all of the suggestions from the SCT meeting. :)

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -314,6 +320,25 @@ def main(argv: Sequence[str]):
...
 
+        # write JSON sidecar file
+        source_path = os.path.join(path_model, "source.json")
+        if os.path.isfile(source_path):
+            with open(source_path, "r") as fp:
+                source_dict = json.load(fp)
+        sidecar_json = {
+            'GeneratedBy': [
+                {
+                    "Name": f"spinalcordtoolbox: sct_deepseg {' '.join(os.path.basename(arg) for arg in argv)}",
```
mguaypaq, line 331: It seems strange to me that we're applying `basename` even to non-path arguments, but as long as they don't contain `/` or `\` then it should be fine? Especially since the alternative would be to duplicate much of the argument parsing logic.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -314,6 +320,25 @@ def main(argv: Sequence[str]):
...
+                    "Version": __version__,
+                    "CodeURL": f"https://github.com/spinalcordtoolbox/spinalcordtoolbox/"
+                               f"blob/{_git_info()[1].strip('*')}/spinalcordtoolbox/scripts/sct_deepseg.py",
+                    "ModelURL": source_dict["model_urls"][name_model][0],
```
mguaypaq, line 335: If you're ok with my other suggestions about `model_urls`, then it's also fine if this field is either-a-string-or-a-dict.

mguaypaq (changes requested):
Thanks for this PR, it's a great feature! I do have a few structural suggestions, though. (Not mandatory! Certainly up for discussion.)

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: I had the same feeling of weirdness? But, exactly as you pointed out, I figured it would be more straightforward than trying to parse for specific arguments, etc.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -399,14 +399,24 @@ def install_model(name_model):
...
-        download.install_data(model_urls, folder(name_model))
+        urls_used = [download.install_data(model_urls, folder(name_model))]
     # Dict of lists, with each list corresponding to a different model seed for ensembling
     else:
         if not isinstance(url_field, dict):
             raise ValueError("Invalid url field in MODELS")
         for seed_name, model_urls in url_field.items():
             logger.info(f"\nInstalling '{seed_name}'...")
             download.install_data(model_urls, folder(os.path.join(name_model, seed_name)), keep=True)
+        urls_used = sum(url_field.values(), [])  # combine seed models into 1 list of urls
```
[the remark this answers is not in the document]
joshuacwnewton: Thank you so much for this feedback!

- Great suggestion re: moving `model_name` to a separate key:value entry. Makes perfect sense!
- I love the suggestion re: ensemble models + checking seeds, too. 
  - At first I was a little hesitant about forcing a redownload if there is a mismatch in the number of seeds? (I was thinking that maybe the user would want to modify the on-disk seeds, e.g. using only a subset of them.) But, then I realized that `source.json` is separate from the actual on-disk files, and only tracks the versions that were initially downloaded, which allows the user to make modifications to the files without necessarily forcing redownloads. So, no problems there, I think!

spinalcordtoolbox/deepseg/models.py
```diff
@@ -399,14 +399,23 @@ def install_model(name_model):
     # List of mirror URLs corresponding to a single model
     if isinstance(url_field, list):
         model_urls = url_field
-        download.install_data(model_urls, folder(name_model))
+        urls_used = [download.install_data(model_urls, folder(name_model))]
```
mguaypaq, line 402: I think this can be a string, rather than a list with a single string in it.
```suggestion
        urls_used = download.install_data(model_urls, folder(name_model))
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -420,6 +429,49 @@ def install_default_models():
...
+    with open(source_path, "r") as fp:
+        source_dict = json.load(fp)
+    model_name = source_dict["model_name"]
+    if model_name not in MODELS:
+        logger.warning(f"Model name '{model_name}' from source.json does not match model names in SCT source code.")
+        return False
+    expected_model_urls = MODELS[model_name]['url']
+    actual_model_urls = source_dict["model_urls"]
+    # Single-seed models
+    if isinstance(expected_model_urls, list) and isinstance(actual_model_urls, list):
```
mguaypaq, line 452: If you take my suggestion above, this should also change:
```suggestion
    if isinstance(expected_model_urls, list) and isinstance(actual_model_urls, str):
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -420,6 +429,49 @@ def install_default_models():
...
+        # if 'actual_model_urls' is up-to-date, then subtracting the current URLs should result in an empty set
+        if set(actual_model_urls) - set(expected_model_urls):
```
mguaypaq, line 454: Ditto.
```suggestion
        if actual_model_urls not in expected_model_urls:
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -420,6 +429,49 @@ def install_default_models():
...
+    if not os.path.isfile(source_path):
+        logger.warning("Provenance file 'source.json' missing!")
+        return False  # NB: This will force a reinstall
+    with open(source_path, "r") as fp:
+        source_dict = json.load(fp)
+    model_name = source_dict["model_name"]
+    if model_name not in MODELS:
+        logger.warning(f"Model name '{model_name}' from source.json does not match model names in SCT source code.")
+        return False
+    expected_model_urls = MODELS[model_name]['url']
```
mguaypaq, line 449: Since we mutate this later with `expected_model_urls.pop`, it should be a copy (which should work whether it's a list or a dict):
```suggestion
    expected_model_urls = MODELS[model_name]['url'].copy()
```

mguaypaq (approved):
Looks good, thanks! Just a few minor fixes.

PR #4472, opened 2024-05-07
Fix small argparse bug when wrapping help text with whitespace-only lines

### LINKED ISSUE #4471: Custom argparse formatter fails for blank lines in multiline strings
Normally, for argparse help descriptions, we use multiple single-line strings with explicit newlines:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2c0649b1cb97b4f11bab9f506b68633fa967907b/spinalcordtoolbox/scripts/sct_extract_metric.py#L63-L68

However, in PR #4302, a multi-line string was instead used:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2c0649b1cb97b4f11bab9f506b68633fa967907b/spinalcordtoolbox/scripts/sct_compute_flow.py#L19-L25

Because there is a whitespace-only line, we get an error here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/069fe79472f3e7a664e0e38c4f123b149814aa59/spinalcordtoolbox/utils/shell.py#L331-L337

The issue is as follows:

- The line contains only whitespace (`'       '`).
- The line is non-empty, so it passes the `if len(li) > 0` check.
- But, because the line is only whitespace, `textwrap.fill(li, width_adjusted)` returns `''`, and `splitlines()` returns `[]`.
- Hence, `init_wrap[0]` is indexing an empty list.

I think this can be fixed easily by changing the check to `if len(li.strip()) > 0`. That way, the text wrapping will be skipped for whitespace-only lines.

joshuacwnewton:
## Description

This PR ensures that our line-wrapping code doesn't fail when passed a whitespace-only line (e.g. `'     '`). This can occur when using a multiline string with a blank line in between.

As a test, I've reverted the help description string for `sct_compute_flow` back to its multiline version.

## Linked issues

Fixes #4471.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_compute_flow.py
+++ b/spinalcordtoolbox/scripts/sct_compute_flow.py
@@ -16,13 +16,13 @@ from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 
 def get_parser():
     parser = SCTArgumentParser(
-        description=(
-            "Compute velocity from the MRI phase image for velocity encoding (VENC) sequences."
-            "\nMore details in: https://mriquestions.com/what-is-venc.html"
-            "\n"
-            "\nFurther features are planned for this script. Please refer to this issue for more info:"
-            "\n  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4298"
-        )
+        description=("""
+Compute velocity from the MRI phase image for velocity encoding (VENC) sequences.
+More details in: https://mriquestions.com/what-is-venc.html
+
+Further features are planned for this script. Please refer to this issue for more info:
+  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4298
+        """)
     )
 
     mandatoryArguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
--- a/spinalcordtoolbox/scripts/sct_compute_flow.py
+++ b/spinalcordtoolbox/scripts/sct_compute_flow.py
@@ -16,7 +16,7 @@ from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 
 def get_parser():
     parser = SCTArgumentParser(
-        description=("""
+        description=("""\
 Compute velocity from the MRI phase image for velocity encoding (VENC) sequences.
 More details in: https://mriquestions.com/what-is-venc.html
 
--- a/spinalcordtoolbox/utils/shell.py
+++ b/spinalcordtoolbox/utils/shell.py
@@ -329,6 +329,7 @@ class SmartFormatter(argparse.ArgumentDefaultsHelpFormatter):
         offsets = [re.match("^[ \t]*", line).group(0) for line in lines]
         wrapped = []
         for i, li in enumerate(lines):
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_compute_flow.py
```diff
@@ -16,13 +16,13 @@
...
 def get_parser():
     parser = SCTArgumentParser(
-        description=(
-            "Compute velocity from the MRI phase image for velocity encoding (VENC) sequences."
-            "\nMore details in: https://mriquestions.com/what-is-venc.html"
-            "\n"
-            "\nFurther features are planned for this script. Please refer to this issue for more info:"
-            "\n  - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4298"
-        )
+        description=("""
```
mguaypaq, line 19: Consider escaping the initial newline:
```suggestion
        description=("""\
```

spinalcordtoolbox/utils/shell.py
```diff
@@ -329,6 +329,7 @@ def _split_lines(self, text, width):
         offsets = [re.match("^[ \t]*", line).group(0) for line in lines]
         wrapped = []
         for i, li in enumerate(lines):
+            li = li.rstrip()  # strip trailing whitespace
```
mguaypaq, line 332: I was curious why `rstrip` rather than `lstrip` or even `strip`, but I see that this matches the behaviour of `textwrap.fill`: it preserves leading whitespace, but discards trailing whitespace. I guess leading whitespace is likely to be intentional formatting.

mguaypaq (approved):
Thanks for the fix! I love the simplicity.

BTW, it's not directly relevant here, since we want to align the quoted text to the start of the line, but did you know about [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent)?

spinalcordtoolbox/utils/shell.py
joshuacwnewton: That was exactly my thinking! Leading whitespace is frequently used when indenting lists, and in fact our formatter explicitly picks up on this:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c515397ec53616cbe3b834be24f53016537cccec/spinalcordtoolbox/utils/shell.py#L341-L342

joshuacwnewton:
> BTW, it's not directly relevant here, since we want to align the quoted text to the start of the line, but did you know about [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent)?

Ah!! This falls into the same class of "stdlib libraries I was aware of but subconsciously avoided because it involves another import" that we talked about previously with itertools/pathlib/etc.

I wonder... what would happen if we universally applied `textwrap.dedent` to the argparse formatter as a whole? This would auto-correct any multiline strings, while also (presumably) having no effect in the majority of cases?

Maybe still unnecessary, given that we don't actually use multiline strings across our documentation (outside of this one-off contribution). But, I want to keep it in my back pocket for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3095? I think it might be relevant there... (I'll type up a little write-up after I merge this PR.)

PR #4474, opened 2024-05-07
Run all tests in temporary directories (to preserve a clean copy of `sct_testing_data`)

### LINKED ISSUE #2911: Tests don't cleanup after themselves

joshuacwnewton:
## Description

**Context:** SCT used to manage its own testing framework. This framework was ported to `pytest` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373. However, many of the old tests used relative paths to input data, which would fail unless the new pytest tests were run in `sct_testing_data` (or the tests were rewritten).

At the time, in order to keep the diffs for https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3373 a little more manageable, I opted to keep the relative paths but add a fixture to run the tests in `sct_testing_data`. This was never meant to be a long-term solution, as it meant that outputs to the PWD would clutter `sct_testing_data`.

This PR is that final follow-up to properly rewrite all relative paths, so that we can run the tests themselves in tempdirs, rather than `sct_testing_data`. (This has the added benefit of rendering the data integrity check actually useful now, since it shouldn't throw spurious errors anymore.)

## Reviewing this PR

This PR is very large, but most of the changes are uncomplicated. I was able to find+replace most relative paths by searching for e.g. `'t2/`, `'dmri/`, etc. 

## Linked issues

Fixes #2911.

### DIFF
--- a/testing/conftest.py
+++ b/testing/conftest.py
@@ -33,12 +33,17 @@ def pytest_sessionstart():
         install_named_dataset('sct_testing_data', dest_folder=sct_test_path())
 
 
-@pytest.fixture
-def run_in_sct_testing_data_dir():
-    """Temporarily change the working directory to 'sct_testing_data'. This replicates the behavior of the old
-    `sct_testing`, and is needed to prevent tests from cluttering the working directory with output files."""
+@pytest.fixture(autouse=True)
+def run_in_tmp_path(tmp_path):
+    """
+    Temporarily change the working directory to a pytest temporary dir. This is needed to prevent tests from
+    cluttering the working directory *or* the sct_testing_data directory with output files.
+
+    Note: Because this uses the `tmp_path` fixture, which generates one temporary directory per test, this allows
+          tests to _also_ call `tmp_path` and access the same directory (e.g. for checking outputs).
+    """
     cwd = os.getcwd()
-    os.chdir(sct_test_path())
+    os.chdir(tmp_path)
     yield
     os.chdir(cwd)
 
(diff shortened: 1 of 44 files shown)


testing/conftest.py
```diff
@@ -33,12 +33,17 @@ def pytest_sessionstart():
         install_named_dataset('sct_testing_data', dest_folder=sct_test_path())
 
 
-@pytest.fixture
-def run_in_sct_testing_data_dir():
-    """Temporarily change the working directory to 'sct_testing_data'. This replicates the behavior of the old
-    `sct_testing`, and is needed to prevent tests from cluttering the working directory with output files."""
+@pytest.fixture(autouse=True)
```
mguaypaq, line 36: Ah! I was wondering why the commit description said "replace", even though it looked more like "delete", but I see that this says `autouse=True`. I didn't know this was a possibility. Oh, and I see this is also how we run the integrity check! Good to know all around.

testing/conftest.py
```diff
@@ -33,12 +33,17 @@ def pytest_sessionstart():
...
+def run_in_tmp_path(tmp_path):
+    """
+    Temporarily change the working directory to a pytest temporary dir. This is needed to prevent tests from
+    cluttering the working directory *or* the sct_testing_data directory with output files.
+
+    Note: Because this uses the `tmp_path` fixture, which generates one temporary directory per test, this allows
+          tests to _also_ call `tmp_path` and access the same directory (e.g. for checking outputs).
```
mguaypaq, line 43: Good explanation, thanks for including it.

mguaypaq (approved):
Thanks for this massive effort!!! And I always found the previous behaviour to be a little annoying, so this feels like a relief :tada:

Since the tests pass, I trust that you found and fixed all instances of relative paths :smile:

PR #4479, opened 2024-05-14
Fix tiny sagittal mosaic QC Report for large images with many slices

### LINKED ISSUE #4458: Sagittal mosaic QC report is tiny for 1.0mm istropic image with 75 slices
## Description

I was experimenting with QC generation on spine segmentation and figured that I could use the flag `-p sct_deepseg_lesion` to quality control my data.

<details>
<summary>Example</summary>
drawing

```python
subprocess.check_call([
                            "sct_qc", 
                            "-i", img_path,
                            "-s", spine_path,
                            "-d", spine_path,
                            "-p", "sct_deepseg_lesion",
                            "-plane", "sagittal",
                            "-qc", qc_path
                        ])
```

</details>

But then I wanted to use the same trick on another image with a bigger field of view and got this results:

Screenshot 2024-04-24 at 16 08 30

Would it be possible to optimize the image disposition used in the QC ? Or at least allow the user to specify a different matrix ?

In this case I would like to have more than 3 columns

joshuacwnewton:
## Description

Sagittal mosaics are a (relatively) new type of QC report, added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4102. They are based off of mosaics of Axial slices. However, there is a key difference: Axial slice mosaics are fixed at 30x30vox per axial slice, whereas sagittal slice mosaics vary depending on the extents of the sagittal segmentation.

In #4458, @NathanMolinier discovered an edge case: If the resolution of the image is sufficiently large, and if the segmentation spans many sagittal slices, then the mosaic can grow way too big in size. At that point, it hits our height limitations for QC reports (both in matplotlib canvas size, and CSS max-height), and thus the whole mosaic image gets scaled down to fit:

image

So, this PR adjusts the height limitations to allow sagittal mosaics to display in full size:

image

However, this causes a new issue: Now the QC report is so large that it becomes unwieldy to actually view, thanks to the resampling step. (1.0mm -> 0.6mm causes the voxel size of the image to balloon.) So, to counteract that, this PR adds a `-resample` option to allow the user to vary the resampling value to change the zoom of the QC report.

- `-resample 0` (keeping original resolution of 1.0mm):

image

- `-resample 2.0` (doubling mm resolution -> halving the voxel resolution)

image

## Testing this PR

Currently, the `batch_processing.sh` test uploads the generated QC report as a workflow artifact. So, we should be able to quickly manually check that these changes don't modify any of the existing QC reports. 

## Linked issues

Fixes #4458.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_qc.py
+++ b/spinalcordtoolbox/scripts/sct_qc.py
@@ -48,7 +48,9 @@ def get_parser():
                         choices=('axial', 'sagittal'),
                         required=False)
     parser.add_argument('-resample',
-                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling.',
+                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling. You can'
+                             'use this option to control the zoom of the QC report: higher values will result in '
+                             'smaller images, and lower values will result in larger images.',
                         type=float,
                         required=False)
     parser.add_argument('-text-labels',
--- a/spinalcordtoolbox/scripts/sct_qc.py
+++ b/spinalcordtoolbox/scripts/sct_qc.py
@@ -48,7 +48,7 @@ def get_parser():
                         choices=('axial', 'sagittal'),
                         required=False)
     parser.add_argument('-resample',
-                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling. You can'
+                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling. You can '
                              'use this option to control the zoom of the QC report: higher values will result in '
                              'smaller images, and lower values will result in larger images.',
                         type=float,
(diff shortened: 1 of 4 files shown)


spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -47,6 +47,12 @@ def get_parser():
                         help='Plane of the output QC. Only relevant for -p sct_deepseg_lesion.',
                         choices=('axial', 'sagittal'),
                         required=False)
+    parser.add_argument('-resample',
+                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling. You can'
```
mguaypaq, line 51: Missing space:
```suggestion
                        help='Millimeter resolution to resample the image to. Set to 0 to turn off resampling. You can '
```

mguaypaq (approved):
The code looks good, thanks for the fix!

PR #4480, opened 2024-05-14
Make sure that `fname_out` is set properly for `fsleyes` command when `-o` is not passed

### LINKED ISSUE #4478: viewer syntax at the end does not provide correct output file name
while working on #4477 I noticed that the output syntax of `sct_apply_transfo` is missing the output file name:

```console
julien-macbook:~/temp/Lydia/results_20240430_225215/data_processed/sub-BB277/anat $ sct_apply_transfo -i sub-BB277_T2_straight.nii.gz -d sub-BB277_T2.nii.gz -w warp_straight2curve.nii.gz 

--
Spinal Cord Toolbox (git-master-4e3b9ebbbd6672169a8bdfd9c5fde4b81716eaa7)

sct_apply_transfo -i sub-BB277_T2_straight.nii.gz -d sub-BB277_T2.nii.gz -w warp_straight2curve.nii.gz
--


Parse list of warping fields...

Get dimensions of data...
  63 x 89 x 318 x 1

Apply transformation...

Apply transformation and resample to destination space...
/Users/julien/code/spinalcordtoolbox/bin/isct_antsApplyTransforms -d 3 -i sub-BB277_T2_straight.nii.gz -o sub-BB277_T2_straight_reg.nii.gz -t warp_straight2curve.nii.gz -r sub-BB277_T2.nii.gz -n 'BSpline[3]' # in /Users/julien/temp/Lydia/results_20240430_225215/data_processed/sub-BB277/anat
Copy affine matrix from destination space to make sure qform/sform are the same.

Done! To view results, type:
fsleyes sub-BB277_T2.nii.gz . &
```

Notice the last line. I would expect to see this instead:
~~~
fsleyes sub-BB277_T2.nii.gz sub-BB277_T2_straight_reg.nii.gz &
~~~

joshuacwnewton:
## Description

The fallback filename (`{fname_in}_reg`) was only set inside the context of the `Transform` class. Meaning, if `-o` wasn't passed, then the `fsleyes` command would display a blank filename. So, I moved setting `fname_out` to before we call `Transform`, which is more in line with our API/CLI separation conventions anyway.

(Side note: I notice that our variable naming conventions are all over the place for this module (`fname`, `file`, `filename`, etc.). Maybe that's worth tidying up in this PR, too?)

## Linked issues

Fixes #4478.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -13,7 +13,7 @@ import os
 import functools
 from typing import Sequence
 
-from spinalcordtoolbox.image import Image, generate_output_file
+from spinalcordtoolbox.image import Image, generate_output_file, add_suffix
 from spinalcordtoolbox.cropping import ImageCropper
 from spinalcordtoolbox.math import dilate
 from spinalcordtoolbox.labels import cubic_to_point
@@ -78,8 +78,7 @@ def get_parser():
         "-o",
         help='Registered source. Example: dest.nii.gz',
         required=False,
-        metavar=Metavar.file,
-        default='')
+        metavar=Metavar.file)
     optional.add_argument(
         "-x",
         help=""" Interpolation method. The 'label' method is to be used if you would like to apply a transformation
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_apply_transfo.py
mguaypaq, line 1: I see that `fname_out` is really not optional in `Transform.apply()` (for example, line 209), so could you also remove the default value for `output_filename` in the constructor `Transform.__init__()` (line 112)?

(I would have commented directly on one of those lines, but neither of them is in the diff, so I can't, sorry.)

mguaypaq (approved):
This is definitely an improvement to the logic here, thanks!

PR #4482, opened 2024-05-14
Add postprocessing functionality for non-ivadomed models

### LINKED ISSUE #4481: Add postprocessing functionality for the new monai/nnunet models
The contrast-agnostic segmentation model segments other objects in the image in addition to the spinal cord as mentioned by a collaborator in [this issue](https://github.com/sct-pipeline/spine-park/issues/31#issuecomment-2104200138). @joshuacwnewton then [clarified](https://github.com/sct-pipeline/spine-park/issues/31#issuecomment-2104935738) that the postprocessing (such as removing objects) only exists within the ivadomed models. 

As the new monai/nnunet models are being rapidly being used by the collaborators, it is important to bring the postprocessing functionality to these new models.

naga-karthik:
#### GitHub

## Description

This PR adds postprocessing functionalities for our latest monai/nnunet (i.e. non-ivadomed) models. Specifically, the `keep_largest_object` function is now called after inference. 

The updates brought by this PR are based specifically on SC segmentation (hence any small objects outside the SC segmentation are removed). But, since this is applied by default in `segment_non_ivadomed` function, it might also be applied in the case of lesion segmentation models (we have to be careful about this). 

Any suggestions on improving the PR for robustness across models are welcome!  

## Linked issues

Resolves #4481


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -125,6 +125,10 @@ def segment_non_ivadomed(path_model, model_type, input_filenames, threshold, use
         fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device)
         for fname_out, target in zip(fnames_out, targets):
             im_out = Image(fname_out)
+            # Apply postprocessing (replicates existing functionality from ivadomed package)
+            # 1. Keep the largest connected object
+            im_out.data = keep_largest_object(im_out.data)
```
joshuacwnewton, line 130: > But, since this is applied by default in segment_non_ivadomed function, it might also be applied in the case of lesion segmentation models (we have to be careful about this).

I had the exact same thought, especially in the context of lesion segmentation. So, I was presuming that we would make this behavior opt-in, and re-use the existing `-remove-small` argument that already exists in `sct_deepseg`.

I'm wondering, since `-remove-small` is just an option to turn on ivadomed postprocessing, and `keep_largest_object` is based on the ivadomed code... were they both based on the same ivadomed code? Can we re-use the existing ivadomed code, and thus be able to re-use `-remove-small` too?

naga-karthik: Thanks for the suggestion! One of the reasons I didn't call the ivadomed function directly is that I/we don't know if ivadomed will be supported in the future! There are active projects that are trying to port ivadomed models into monai so I wasn't sure if importing the postprocessing functions from ivadomed would be future compatible? what do you think?

joshuacwnewton: Ah, sorry! I mean like, copying the function entirely? (If we're going to make a new function inside of `math.py` anyway, we might as well make it identical to the existing logic?)

joshuacwnewton: Oh! My mistake. It looks like the function you added is identical to the [ivadomed function](https://github.com/ivadomed/ivadomed/blob/1fccf77239985fc3be99161f9eb18c9470d65206/ivadomed/postprocessing.py#L99-L116) already!

In that case, all we have to do is tap into the value of `sct_deepseg -largest` to control this option.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -287,6 +287,7 @@ def main(argv: Sequence[str]):
             thr = (arguments.binarize_prediction if arguments.binarize_prediction
                    else models.MODELS[name_model]['thr'])  # Default `thr` value stored in model dict
             im_lst, target_lst = inference.segment_non_ivadomed(path_model, model_type, input_filenames, thr,
+                                                                keep_largest=arguments.keep_largest,
                                                                 use_gpu=use_gpu, remove_temp_files=arguments.r)
```
joshuacwnewton, line 291: @naga-karthik If you have some time, would you be able to add functionality for the other `sct_deepseg` postprocessing arguments, too? 

There are only two other arguments (`-fill-holes` and `-remove-small`), and you should be able to copy the ivadomed functions from [`postprocessing.py`](https://github.com/ivadomed/ivadomed/blob/1fccf77239985fc3be99161f9eb18c9470d65206/ivadomed/postprocessing.py) in much the same way as was done for `-largest`.

All you need to do is pass the argument value to `segment_non_ivadomed`, then call the necessary function if the value `is not None`.

naga-karthik: Sure! I added both these postprocessing options in the following commits: 
1. `-fill-holes` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/2d43faa1c0f08332abd9b4368e9841bac3024c84  
2. `-remove-small` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/07ec6d9ae60b37f758e37046408d56caf81c48f5

the fill-holes part took some time as the original logic was a bit complicated (it was calling a function tied to a class), the commit tries to include the whole logic into a single function `remove_small_objects`

Please let me know your thoughts on these changes!

joshuacwnewton: Thank you so much for taking the time to do this! Taking a look now. :hearts:

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -125,8 +126,23 @@ def segment_non_ivadomed(path_model, model_type, input_filenames, threshold, use
         fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device)
         for fname_out, target in zip(fnames_out, targets):
             im_out = Image(fname_out)
+            # Apply postprocessing (replicates existing functionality from ivadomed package)
+            # 1. Binarize predictions based on the threshold value
+            # NOTE: Any post-processing is done only after thresholding (as post-processing operations 
+            # expect binary masks)
```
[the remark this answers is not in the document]
joshuacwnewton: I think `remove_small_objects` and `keep_largest` might work with softsegs? (But, `binary_fill_holes` definitely doesn't, of course.)

spinalcordtoolbox/math.py
```diff
@@ -345,3 +345,99 @@ def denoise_patch2self(data_in, bvals_in, patch_radius=0, model='ols'):
...
+
+    Args:
+        predictions (ndarray or nibabel object): Input binary segmentation. Image could be 2D or 3D.
+        structure (tuple of integers): Structuring element, number of ints equals
+            number of dimensions in the input array.
+
+    Returns:
+        ndrray or nibabel (same object as the input). Output type is int.
+    """
+    assert np.array_equal(predictions, predictions.astype(bool))
```
joshuacwnewton, line 386: This is the check that will fail if the input is not binary. However, it doesn't really fail with a useful error message? 

Since people may try to use `-fill-holes` with softseg, we should probably turn this into a useful error message that explains that they need to use `-thr` to binarize their data.

joshuacwnewton (changes requested):
This review is a bit all over the place (some comments, some suggestions, some observations) but feel free to mark the non-suggestions as "Resolved" once you've read them. :)

spinalcordtoolbox/math.py
```diff
@@ -345,3 +345,99 @@ def denoise_patch2self(data_in, bvals_in, patch_radius=0, model='ols'):
...
+    Args:
+        data (ndarray): Input data.
+        dim_lst (list): Dimensions of a voxel in mm.
+        unit (str): Indicates the units of the objects: "mm3" or "vox"
+        thr (int or list): Minimal object size to keep in input data.
+    
+    Attributes:
+        bin_structure (ndarray): Structuring element that defines feature connections.
+        size_min (int): Minimal object size to keep in input data.
+
```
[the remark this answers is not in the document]
naga-karthik: ah I just thought of defining what those variables `bin_structure` and `size_min` mean as they are used below. but probably docstring is not the right place for this though. I will move it below as a comment

spinalcordtoolbox/math.py
```diff
@@ -345,3 +345,99 @@ def denoise_patch2self(data_in, bvals_in, patch_radius=0, model='ols'):
     denoised = dipy.denoise.patch2self.patch2self(data_in, bvals_in, patch_radius=patch_radius, model=model)
 
     return denoised
+
+
+def keep_largest_object(predictions):
```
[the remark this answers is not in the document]
naga-karthik: Hmm, good observation! on the other hand, I think we should do [@binarize_with_low_threshold](https://github.com/ivadomed/ivadomed/blob/1fccf77239985fc3be99161f9eb18c9470d65206/ivadomed/postprocessing.py#L36-L52) by default (irrespective of whether the users want to use `-thr` or not). The model predictions tend to have a lot of small values (i.e. in the range of `1e-4/5/6` and we have to remove them to get an output that's "presentable"). 

Note that this is only for the monai models, the nnunet models don't support soft outputs

spinalcordtoolbox/math.py
naga-karthik: good point! added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/3599cb8d40990bffdb0b0a6d35118dc82a18ed2e

spinalcordtoolbox/math.py
```diff
@@ -345,3 +345,99 @@ def denoise_patch2self(data_in, bvals_in, patch_radius=0, model='ols'):
...
+    # for spinal cord (SC) segmentation, num_classes=1, 
+    # for region-based models with both SC and lesion segmentations, num_classes=2
+    num_classes = data.shape[0] if len(data.shape) == 4 else 1
+
+    bin_structure = generate_binary_structure(3, 2)
+    data_label, n = label(data, structure=bin_structure)
+
+    if isinstance(thr, list) and (num_classes != len(thr)):
+        raise ValueError("Length mismatch for remove small object postprocessing step: threshold length of {} "
+                            "while the number of predicted class is {}.".format(len(thr), num_classes))
```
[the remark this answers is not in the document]
naga-karthik: thanks for pointing this out! I fixed these issues and now i see [no failing tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9295863674/job/25583457423?pr=4482)! 

commits: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/3eeca3385dd77d9cacc2e43d962095124a0ef760 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/79f29a70063816ee864f2e25e94b410180c225be

spinalcordtoolbox/math.py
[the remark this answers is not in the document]
naga-karthik: fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4482/commits/3599cb8d40990bffdb0b0a6d35118dc82a18ed2e !

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -126,12 +126,14 @@ def segment_non_ivadomed(path_model, model_type, input_filenames, threshold, kee
...
+            # add soft threshold (not binarization)
+            im_out.data[im_out.data < 0.1] = 0
             # Apply postprocessing (replicates existing functionality from ivadomed package)
             # 1. Binarize predictions based on the threshold value
             # NOTE: Any post-processing is done only after thresholding (as post-processing operations
             # expect binary masks)
-            if threshold is not None:
-                im_out.data = binarize(im_out.data, threshold)
+            # if threshold is not None:
+            #     im_out.data = binarize(im_out.data, threshold)
```
joshuacwnewton, line 136: Note: The "hacky" fix in aadf57aa587f6cd921f0946514abb2c3a19e9d71 is superseded by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4508 (which adds a soft thresholding for just the MONAI models).

Once that PR is merged, we should be able to remove the commit in this PR. :)

joshuacwnewton (approved):
Thank you for this PR! I've tested this with binary segmentations and everything works well. 

Once this is merged, I'll follow up on the other PR for softsegs and make sure everything works well with `-thr 0`, too.

spinalcordtoolbox/deepseg/inference.py
joshuacwnewton: Actually, scratch that. I'll merge this one first, then test softsegs in the other PR. :)

naga-karthik:
Thank you @joshuacwnewton  for taking care of this! I have tight deadlines this month and couldn't make progress on this PR in time! Really appreciate you taking over and completing it! :)

PR #4484, opened 2024-05-15
Make `sct_download_data` errors non-fatal during installations of SCT

### LINKED ISSUE #4410: File permission error during SCT installation
@sandrinebedard reports a Windows installation bug, when installing from `master` under WSL. The installation crashes in one of the `sct_download_data` commands in a non-crucial step, but it stops the further installation steps. Specifically, there's a file permission error in one of these lines, where temporary files that were just created are being cleaned up:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/56c7f6cdb0f763f1e287a3255b25ab532e257a00/spinalcordtoolbox/download.py#L323-L324

We tested, and `rm -r $THE_FILES` works fine in the same terminal that was used for the installation attempt. Possibly this is because Windows refuses to delete files that are still open? I'll have to wait until I'm at an actual Windows machine to reproduce the bug and investigate.

In the meantime, we just commented out the two `shutil.rmtree` lines from Sandrine's local copy of SCT to unblock the installation.

joshuacwnewton:
## Description

This PR follows the reasoning in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4410#issuecomment-2112883978. By allowing `sct_download_data` failures, we mitigate crashing the entire SCT installation if a single dependency fails to install (e.g. due to network/disk issues).

That said, it is still important for users to know that the installation of dependencies has failed. So, we also add a new check to `sct_check_dependencies` that checks if the default data dependencies are installed. As a bonus, this means that the list of defaults no longer has to be hardcoded inside of the individual install scripts.

## Linked issues

Fixes #4410.


install_sct.bat
```diff
@@ -187,7 +187,7 @@ python\envs\venv_sct\Scripts\pip install -e . || goto error
 rem Install external dependencies
 echo:
 echo ### Downloading model files and binaries...
-FOR %%D IN (PAM50 deepseg_sc_models deepseg_gm_models deepseg_lesion_models binaries_win deepreg_models PAM50_normalized_metrics) DO sct_download_data -d %%D -k || goto error
+FOR %%D IN (PAM50 deepseg_sc_models deepseg_gm_models deepseg_lesion_models binaries_win deepreg_models PAM50_normalized_metrics) DO sct_download_data -d %%D -k || echo Failed to install '%%D'. Please re-run 'sct_download_data -d %%D' after installation.
```
mguaypaq, line 190: I want to say "ditto" here, but honestly I have no idea how to do even this much string manipulation in `.bat` scripts.

mguaypaq (commented):
Possibly this would be good to bring up at today's SCT meeting!

install_sct
```diff
@@ -687,9 +687,12 @@ if [[ -n "$NO_SCT_BIN_INSTALL" ]]; then
   print warning "WARNING: SCT binaries will not be (re)-installed"
 else
   print info "Installing binaries..."
-  run sct_download_data -d "binaries_${OS}" -k
+  # shellcheck disable=SC2015  # it would be very strange if print (echo) failed and triggered the || case
+  run sct_download_data -d "binaries_${OS}" -k \
+    && print info "All requirements installed!" \
+    || print error "Failed to install 'binaries_${OS}'. Please re-run 'sct_download_data -d binaries_${OS}' after installation."
```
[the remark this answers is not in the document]
joshuacwnewton: > If this is just a line buried in the middle of the install log, with no bold red text at the end of the install process, I think people will ignore it.

I think `print error` is what adds the bold red text? But, I on reflection I agree that this would get buried.

install_sct.bat
mguaypaq: I guess if we push this check to `sct_check_dependencies`, then that takes care of the problem for us!

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -442,6 +436,14 @@ def main(argv: Sequence[str]):
         print_fail()
         print(err)
 
+    # check if data folder contains the required default depedencies
+    for dataset_name in default_datasets() + [_which_sct_binaries()]:
+        print_line(f"Check data dependency '{dataset_name}'")
+        if is_installed(dataset_name):
+            print_ok()
+        else:
+            print_fail(f" Run 'sct_download_data -d {dataset_name}' to reinstall")
```
joshuacwnewton, line 445: New check added!

Success logs:

image

I just need to fixup my batch script syntax to actually install the dependencies on Windows :sweat_smile: 

Failing logs:

image

install_sct
```diff
@@ -698,9 +698,7 @@ else
   # Download data
   print info "Installing data..."
   run mkdir -p "$SCT_DIR/$DATA_DIR"
-  for data in PAM50 deepseg_sc_models deepseg_gm_models deepseg_lesion_models deepreg_models PAM50_normalized_metrics; do
-    run sct_download_data -d "$data"
-  done
+  python/envs/venv_sct/bin/python -c 'import spinalcordtoolbox.download; spinalcordtoolbox.download.install_default_datasets()'
```
joshuacwnewton, line 701: Hrm... this doesn't seem to have the same `logging` output as `sct_download_data`. I'll have to look into that...

joshuacwnewton: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4484/commits/6401119a0f71c5fba20e576031ca0ea82a463c4c.

mguaypaq (approved):
This is great, thanks!

PR #4485, opened 2024-05-15
Allow multiple `sct_maths` operations to be run sequentially on an image in one command

### LINKED ISSUE #1447: Allow `sct_maths` to run multiple sequential operations in one command
Example:
~~~
sct_maths -i target.nii -mul 7 -add b0_moco_T0006.nii -div 7 -o target.nii
~~~

joshuacwnewton:
## Description

This PR allows users to provide multiple arguments and have them be run sequentially, in the order provided. Example:

```
sct_maths -i im_0.nii.gz -add im_1.nii.gz -mul im_2.nii.gz
```

This PR adds a test, then updates `sct_maths` to iterate over the input arguments, running one operation at a time.

I decided to go with the "parse `argv` to get argument order" method as opposed to the "add a custom action to collect the arguments" method because I figured it was the most straightforward, and it allows us to use other actions for our argparse options.

Given that this is a large refactor, this probably needs some thorough testing for edge cases, but I wanted to get a PR open prior to the meeting.

## Linked issues

Fixes #1447.


spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -417,69 +402,72 @@ def main(argv: Sequence[str]):
                 b = int(i.split('=')[1])
         data_out = sct_math.denoise_nlmeans(data, patch_radius=p, block_radius=b)
 
-    elif arguments.symmetrize is not None:
+    else:
+        assert arg_name == "symmetrize"
         data_out = sct_math.symmetrize(data, arguments.symmetrize)
```
joshuacwnewton, line 407: Whoops! Looks like I'm getting a failure for `batch_processing.sh`

```
    assert arg_name == "symmetrize"
AssertionError
```

mguaypaq (changes requested):
This is a nice attempt: it tries to make minimal changes to the existing code, and it respects the order of operations on the command line.

But, I'm not a fan of this design, because it loses the connection between the operations and their arguments. For example, when I ran the following test, I got the wrong result:
```
sct_maths -i t2.nii -add 1 -add 2 -o result.nii
```
The problem is that `arguments.add` only stores the last value, so both operations added 2, instead of 1 and 2.

I think we'll need the argument parser to actually build a list of operations (and their arguments). But we should carefully think about the possible interactions between arguments, because this change turns the `sct_maths` command-line into a mini programming language. In particular, the current "mathematical morphology" options `-shape` and `-dim` sort of work like global variables, which is tricky. And the "similarity metric" options don't respect the "image in -> image out" function signature.

joshuacwnewton:
Ahhh, very true -- relying on the [default behavior of argparse](https://stackoverflow.com/questions/36166225/using-the-same-option-multiple-times-in-pythons-argparse) (via `parser.parse_args(argv)`) will lose information from multiple calls to a single argument.

Truthfully, though, I'm still a bit hesitant about using a custom action to implement this behavior? I feel like the complexity increases quite a bit when we try to account for behavior beyond the simplest cases.

But, I think I have an idea of how to resolve the conflict between a new custom action and the existing custom action. Right now, the current "`ParseDataOrScalarArgument`" behaves very similarly to a type function. So, I think we can shift some of that behavior over to free up the action slot. I'm willing to give it a shot! :)

mguaypaq:
Since this change is really a design change rather than a simple bug fix, here's an attempt at a short design doc, explore the constraints of the problem and a proposed solution.

If you don't have any objections, I can add a few commits to implement it.

## Design doc: command-line language for `sct_maths`

### Problem statement

The goal is to allow the user to give a sequence of multiple operations to be performed by `sct_maths` in a single command-line invocation.

Conceptually, the script should have some notion of "the current value", and do something like:
```
current_value = (get from -i)
for operation, other_args in parsed_command_line:
    current_value = operation(current_value, other_args)
(save to -o)
```

Some difficulties:
- The order of operations matters. But the usual parsing approach for SCT commands loses the order of arguments.
- Some operations may be repeated. But the usual parsing approach for SCT commands only keeps a single copy of each flag (and associated arguments).
- Not all flags represent operations to be done (for example, `-i`, `-o`, `-h`, `-v`).
- Some flags act like global variables (`-shape` and `-dim`).
- Some operations have a different function signature than "one or more images in, one image out" (the similarity metrics `-mi`, `-minorm`, `-corr`).
- Most of the inputs can be either an image file or a plain number, and we should all type combinations sensibly. But the final output can't be a plain number (because, what would be the image dimensions, affine, etc.?).

Two design constraints I would like to impose:
- The script should parse the entire command-line before opening any files or starting any operations, because there might be an `-h` or a syntax error later in the command-line.
- The script should check whether the entire sequence of input and output types makes sense before opening any files or starting any operations, and output a sensible error message with the usage string.

One non-goal:
- We could consider also checking that each input image has the right number of dimensions, and that the dimensions match up for image-image operations, before actually performing the operations. But this would require loading each image twice (once to check, once to operate) or loading all images in memory at the same time. I think the cost is bigger than the potential benefit.

### Proposed solution

I suggest a 3-stage approach:
1. We use `argparse` to turn the entire command-line into:
   - a list of (operation, arguments) to perform, and
   - the non-operation arguments.
2. When `parser.parse_args` returns, we do a separate type-checking pass, and call `parser.error` if needed.
3. We run the actual operations, opening any needed files as we go, and raise a (non-parser) error if we encounter any problems.

**Q:** Why not do the type-checking incrementally during Stage 1?
**A1:** This might be possible if we kept track of "the current type" during parsing. But then we would need add the requirement that `-i` appears before any operations, which we currently don't require.
**A2:** We would still need a separate check after `parser.parse_args` that the final type is not a plain number. But this could be folded into Stage 3, so it's not a strong argument.
**A3:** Having Stage 2 also makes it possible to report the error that a similarity metric (`-mi`, `-minorm`, `-corr`) is not the final operation at that argument, rather than at the next operation.

**Q:** Why include the arguments in the list of (operation, arguments)?
**A1:** This is in part to make it easy to deal with the "global variable" flags `-shape` and `-dim`. The parser can keep track of the "current value" for each one, and include it in the arguments for operations that need them.
**A2:** Also, this makes a cleaner separation, where `argv` is only given to the Stage 1 parser, and the resulting `arguments` is what's used for Stages 2 and 3.

joshuacwnewton:
What you've written all reads as very sensible to me! Many of the constraints you've detailed were things that I had in mind as well during my changes on this PR, but it's nice to see it all laid out in a very concrete, centralized form.

----

> Q: Why not do the type-checking incrementally during Stage 1?
> A1: This might be possible if we kept track of "the current type" during parsing. But then we would need add the requirement that -i appears before any operations, which we currently don't require.

Note: We actually kind of sort of have this as a pseudo-requirement, due to `ParseDataOrScalarArgument`?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/dd2656ec77b66048bed7b54b086a17a179e5196f/spinalcordtoolbox/scripts/sct_maths.py#L41-L42

It looks like the author of this comment might have a misconception about how `argparse` parses arguments in order. But, I'll come back to this in one sec.

----

Just to make sure I'm understanding correctly the differences between what I've done so far, and what you've proposed:

- Right now, we use `ParseDataOrScalarArgument`, which awkwardly tries to create arrays from integer arguments and store them in `arguments`, which in turn requires knowing the shape of the input, and hence requires `-i` to be parsed first. But, you're suggesting to do this kind of data/scalar parsing in a secondary stage, rather than partway through argparsing. 
- Right now, we construct `ordered_args` in `parse_args`, then filter out the "non-array-operation args" afterwards. If "non-array-operation args" are needed, we get them directly `arguments`. But, you're proposing to construct both lists at the same time within `parse_args`?
- Right now, the PR is treating the "similarity metric operations" as entirely separate from the other array-based operations. In other words, it is not possible with this PR to perform array operations first, and _then_ compute a similarity metric at the end -- instead, array operations are only ever done together (meaning we can safely assume types as array in -> array out). But, your proposal is accounting for mixing these operations (due to mentions of input/output types)? 
   - I wonder how common this use-case is? Would it make our lives easier to not try and mix the two types of operations? (In fact, I was almost thinking that the similarity metrics belong in their own CLI script, similar to how we have quite a few existing `sct_compute_<thing>` scripts already that are separate from `sct_maths`.) 

Apart from these three changes, I think my PR is pretty well-aligned with your design doc? (Constructing a list of (operation, argument) pairs to iterate through, fetching non-operation values from `arguments`, taking into account multiple instances of the same argument, etc.) But, maybe I'm missing some key differences. :thinking:

mguaypaq:
> > But then we would need add the requirement that -i appears before any operations, which we currently don't require.
> 
> Note: We actually kind of sort of have this as a pseudo-requirement, due to `ParseDataOrScalarArgument`?

You're absolutely right! I overlooked that we already have this requirement. Although right now we get an obscure `TypeError: Image constructor takes at least one argument.` traceback rather than a clear usage message if we don't put `-i` before any operations.

Also, the fact that `ParseDataOrScalarArgument` tries to open the `-i` image during command-line parsing means we also get an ugly traceback if there's a typo in the name of the `-i` file.

> Just to make sure I'm understanding correctly the differences between what I've done so far, and what you've proposed:
> 
>     * Right now, we use `ParseDataOrScalarArgument`, which awkwardly tries to create arrays from integer arguments and store them in `arguments`, which in turn requires knowing the shape of the input, and hence requires `-i` to be parsed first. But, you're suggesting to do this kind of data/scalar parsing in a secondary stage, rather than partway through argparsing.

Yes, I'd like to change `ParseDataOrScalarArgument` so that it doesn't open any files.

>     * Right now, we construct `ordered_args` in `parse_args`, then filter out the "non-array-operation args" afterwards. If "non-array-operation args" are needed, we get them directly `arguments`. But, you're proposing to construct both lists at the same time within `parse_args`?

I guess I'm proposing that the `ordered_args` should only contain the operation args in the first place (and the operation args shouldn't be also stored in `arguments`). I think augmenting `argparse._StoreAction` is cute and clever, but a little too intrusive for my taste. And the filtering-out would have to be updated if we ever add other non-operation args, which feels like a weird coupling to me. (Granted, that's unlikely, but weird coupling also makes code harder to read and understand.)

>     * Right now, the PR is treating the "similarity metric operations" as entirely separate from the other array-based operations. In other words, it is not possible with this PR to perform array operations first, and _then_ compute a similarity metric at the end -- instead, array operations are only ever done together (meaning we can safely assume types as array in -> array out). But, your proposal is accounting for mixing these operations (due to mentions of input/output types)?

Yes, being able to get a similarity metric for a computed image would be a (very minor) bonus of what I'm proposing.

>       * I wonder how common this use-case is? Would it make our lives easier to not try and mix the two types of operations? (In fact, I was almost thinking that the similarity metrics belong in their own CLI script, similar to how we have quite a few existing `sct_compute_<thing>` scripts already that are separate from `sct_maths`.)

I would be in favour of spinning off these operations! Probably as a separate PR though.

> Apart from these three changes, I think my PR is pretty well-aligned with your design doc? (Constructing a list of (operation, argument) pairs to iterate through, fetching non-operation values from `arguments`, taking into account multiple instances of the same argument, etc.) But, maybe I'm missing some key differences. 🤔

I think the PR is pretty good actually! But I'd really like to tackle two extra points:
* The script should parse the entire command-line before opening any files or starting any operations, because there might be an -h or a syntax error later in the command-line. (So, that means I would significantly re-work `ParseDataOrScalarArgument`.)
* If `shape` and `dim` appear more than once (maybe intended for different `-dilate` operations), we shouldn't just use the last value while ignoring the previous ones.

joshuacwnewton:
Ahhhh, thank you for clarifying! That definitely clears up some misconceptions I had!

Going back to previous discussions from the issue several weeks ago, I remember `ParseDataOrScalar` being a blocker for your idea of having `ordered_args` be tied to a per-argument `CustomAction`. Luckily, if we move the data/scalar logic to a secondary stage (and not a custom action), then that would indeed free up the custom action slot for those arguments, allowing us to construct `ordered_args` without needing to hack up `_StoreAction`.

So, I think all of your proposals end up being pretty harmonious? I'm down for them! :)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -300,187 +375,189 @@ def main(argv: Sequence[str]):
...
+            try:
+                list_data = get_data_arrays(data.shape, arg_value)
+            except ValueError as e:
+                printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
+            data -= np.sum(list_data, axis=0)
+
+        elif arg_name == "mul":
+            if data.ndim == 4 and not arg_value:
+                # special case to multiply a 4D volume across the t axis
+                data = np.sum(data, axis=3)
```
mguaypaq, line 436: Oops, forgot to fix this up after copy-pasting:
```suggestion
                data = np.prod(data, axis=3)
```

mguaypaq (approved):

mguaypaq:
(I restored the branch in case it's needed for rebasing #4499.)

PR #4486, opened 2024-05-16
Update url to slicer documentation

mguaypaq:
The old url has a big banner at the top of the page, claiming that this new url contains the latest version. A quick look confirms that the content is essentially the same.

I noticed this when checking the daily broken url results:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9050554894/job/24866074154


mguaypaq:
Ah, so that's why I suddenly started getting these notifications!

I don't think we both need to receive them. If you'd prefer to receive them, I guess the simplest way would be for you to disable and re-enable the workflow. (It wasn't intuitive to me how to do this, but [this help page](https://docs.github.com/en/actions/using-workflows/disabling-and-enabling-a-workflow) shows where the relevant menu is hidden.)

PR #4487, opened 2024-05-17
Skip empty slicegroups to avoid creating blank rows in output metric CSV file

joshuacwnewton:
## Description

For data where an image is [split into multiple chunks](https://github.com/sct-pipeline/spine-park/issues/10#issuecomment-2113482362), you may want to call the same `sct_extract_metric` command on multiple chunks spanning a range of vertlevels (e.g. `-vert 1:12`). But, not all chunks will contain all of the specified levels.

For this case, there is a bug where an `agg_metric` entry will be created for the slicegroup "`()`", i.e. a group containing no slices. This will result in an empty row containing no data.

This PR adds a test that reproduces the condition and fails. As soon as the empty slicegroup row is removed, the test will pass.

## Linked issues

Fixes https://github.com/sct-pipeline/spine-park/issues/35.


jcohenadad (approved):
I have not tested the code, but the change makes sense, there is now a test, so i'm approving. Thank you @joshuacwnewton

mguaypaq (approved):
Looks good, and thanks for the little cleanup.

PR #4489, opened 2024-05-22
Automatically compute midsagittal tissue bridges

valosekj:
#### GitHub

#### PR contents

## Description

This PR adds computation of the _midsagittal tissue bridges_ into the `sct_analyze_lesion` function. The tissue bridges are currently measured manually; the idea to automate the process was raised during the meeting with the Zurich group at 2024 ISMRM.

Briefly, the midsagittal tissue bridges are used to quantify the width of spared tissue at the minimum distance from cerebrospinal fluid to the tSCI (traumatic spinal cord injury) lesion:

image

Relevant articles [here](https://pubmed.ncbi.nlm.nih.gov/28393423/), [here](https://pubmed.ncbi.nlm.nih.gov/33190619/), and [here](https://pubmed.ncbi.nlm.nih.gov/35521064/).

## Implementation

After a discussion with the Zurich team (2024-06-06), I have updated the code to compute the tissue bridges from all sagittal slices containing lesion. In other words, the script computes the bridges from the midsagittal slice and also from parasagittal slices.

The script can now also handle multiple lesions. In such cases, the tissue bridges are computed for each lesion separately. Then, the minimum bridges for the **_midsagittal slice_** are computed across the lesions and printed into CLI (see this [commit message](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/8a994ef7e776aa7ce7371cf405bfb2c49c4fb999) for details).

## How to test the code

Download dummy [test data](https://www.dropbox.com/scl/fo/r76whcpwidcxhjymdgyqb/AOdKwznb6NBKAJFC8yykpoY?rlkey=6fgp900vjqr0hbbw2ffn99uvf&dl=0) and run: 

```console
cd $SCT_DIR
git fetch
git checkout jv/compute_midsagittal_tissue_bridges
sct_analyze_lesion -m sub-01_T2w_lesion.nii.gz -s sub-01_T2w_seg.nii.gz
```

The script output is:

```console
Measures on lesion #1...
...
  Midsagittal slice of the spinal cord: 8
  Sagittal slice 7, Minimum dorsal tissue bridge width: 1.02 mm (axial slice 300)
  Sagittal slice 7, Minimum ventral tissue bridge width: 1.37 mm (axial slice 294)
  Sagittal slice 7, Total tissue bridge width: 2.39 mm
  Sagittal slice 8, Minimum dorsal tissue bridge width: 1.67 mm (axial slice 302)
  Sagittal slice 8, Minimum ventral tissue bridge width: 2.26 mm (axial slice 296)
  Sagittal slice 8, Total tissue bridge width: 3.93 mm
  Sagittal slice 9, Minimum dorsal tissue bridge width: 1.31 mm (axial slice 317)
  Sagittal slice 9, Minimum ventral tissue bridge width: 1.21 mm (axial slice 306)
  Sagittal slice 9, Total tissue bridge width: 2.52 mm
...
Minimum tissue bridges across all lesions for the midsagittal slice...
  Minimum dorsal bridge width [mm]: 1.67
  Minimum ventral bridge width [mm]: 2.26
...
```

<details><summary>The script also generates a PNG figure for debug:</summary>

image

</details>


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +304,142 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        tissue_bridges_df = pd.DataFrame.from_dict(tissue_bridges_dict, orient='index')
+
+        # Get minimum dorsal and ventral tissue bridges together with their corresponding slices (slices are indices)
+        # NOTE: we get minimum because tissue bridges are quantified as the width of spared tissue at the minimum
+        # distance from cerebrospinal fluid to the lesion boundary
+        min_dorsal_bridge_width_slice = tissue_bridges_df['dorsal_bridge_width'].idxmin()
+        min_ventral_bridge_width_slice = tissue_bridges_df['ventral_bridge_width'].idxmin()
+
+        # Get the width of the tissue bridges in mm
+        # TODO: verify whether to use [0] or [1] for the pixel size
```
naga-karthik, line 390: Can we know which index corresponds to the height and width of the pixel? Ideally, because we're traversing from P-A (A-P) to calculate the tissue bridge for each axial slice, I think we have to multiply with the _height_ of the pixel (to get the values in mm). 

Don't you think it should be [0] then in that case?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +304,142 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        dorsal_bridge_width_mm = tissue_bridges_df.loc[min_dorsal_bridge_width_slice, 'dorsal_bridge_width'] * p_lst[1]
+        ventral_bridge_width_mm = tissue_bridges_df.loc[min_ventral_bridge_width_slice, 'ventral_bridge_width'] * p_lst[1]
+
+        # Save the maximum axial damage ratio
```
naga-karthik, line 394:
```suggestion
        # Save the minimum mid-sagittal tissue bridges
```

valosekj: Good catch! Thank you!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Thanks, Naga! 
I debugged the code again, and I believe that `p_lst[1]` is okay. I clarified it in [this commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/93e7953d0e50313153adc47076d824058a93f6dc):
```
# Get the width of the tissue bridges in mm
# NOTE: the orientation is RPI (because we reoriented the image to RPI using orient2rpi()); therefore p_lst[0]
# is the pixel size in the R-L direction, p_lst[1] is the pixel size in the A-P direction, and p_lst[2] is the
# pixel size in the S-I direction.
# Since we are computing dorsal and ventral tissue bridges, we use p_lst[1] (A-P direction)
```

naga-karthik: wonderful! this clarification note was important! I agree, `p_lst[1]` should be the correct option!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # axial_lesion_slices = np.unique(np.where(im_data)[2])
+
+        # 2) Get the midsagittal slice based on the sagittal slice with the largest lesion area
+        lesion_area_per_slice = [np.sum(im_data[slice, :, :]) for slice in np.unique(np.where(im_data)[0])]
+        mid_sagittal_slice = np.unique(np.where(im_data)[0])[np.argmax(lesion_area_per_slice)]
+        # Get axial slices (S-I direction) with lesion for the midsagittal slice
+        axial_lesion_slices = np.unique(np.where(im_data[mid_sagittal_slice, :, :])[1])
+
+        if self.verbose == 2:
+            printv('  Mid-sagittal slice : ' + str(mid_sagittal_slice), self.verbose, type='info')
```
jcohenadad, line 349:
```suggestion
            printv('  Mid-sagittal slice: ' + str(mid_sagittal_slice), self.verbose, type='info')
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+
+        # --------------------------------------
+        # Compute tissue bridges for each axial slice
+        # --------------------------------------
+        tissue_bridges_dict = {}
+        # Iterate across axial slices to compute tissue bridges
+        for axial_slice in axial_lesion_slices:
+            # Get the column of the current slice
```
jcohenadad, line 357:
```suggestion
            # Get the data of the selected 2D axial slice
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            slice_data = im_data[mid_sagittal_slice, :, axial_slice]
+            # Get the spinal cord column of the current slice
```
jcohenadad, line 359:
```suggestion
            # Get the spinal cord mask of the selected 2D axial slice, from the input spinal cord segmentation
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            slice_sc_data = im_sc_data[mid_sagittal_slice, :, axial_slice]
+            # Get the indices of the lesion in the column
```
jcohenadad, line 361:
```suggestion
            # Get the indices of the lesion for the selected axial slice
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            lesion_indices = np.where(slice_data)[0]
+            # Get the indices of the spinal cord in the column
```
jcohenadad, line 363:
```suggestion
            # Get the indices of the spinal cord mask for the selected axial slice
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
 
+    def _measure_tissue_bridges(self, im_data, p_lst, idx):
+        """
+        Measure the midsagittal tissue bridges (widths of spared tissue ventral and dorsal to the spinal cord lesion).
+        Tissue bridges are quantified as the width of spared tissue at the **minimum** distance from cerebrospinal fluid
+        (i.e., the spinal cord boundary) to the lesion boundary.
+        REF: Huber E, Lachappelle P, Sutter R, Curt A, Freund P. Are midsagittal tissue bridges predictive of outcome
+        after cervical spinal cord injury? Ann Neurol. 2017 May;81(5):740-748. doi: 10.1002/ana.24932.
+
+        :param im_data: 3D numpy array, binary mask of the lesion
```
jcohenadad, line 318:
```suggestion
        :param im_data: 3D numpy array, binary mask of the lesion. The orientation is assumed to be RPI.
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -302,6 +307,146 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        :param p_lst: list, pixel size of the lesion
+        :param idx: int, index of the lesion
+        """
+
+        # Load the spinal cord image
```
jcohenadad, line 323:
```suggestion
        # Load the spinal cord segmentation
```

valosekj:
Thanks for the initial review, @naga-karthik and @jcohenadad! I updated the script in the recent commits to compute the tissue bridges from _**all**_ sagittal slices containing lesion. I also updated the [PR description](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#issue-2311427775) accordingly. 

I believe the PR is now ready for the first formal review from @joshuacwnewton and/or @mguaypaq.

TODO for me: implement angle correction

joshuacwnewton:
For now, I've just quickly resolved the conflicts with the `master` branch (using LazyLoader for the `matplotlib.pyplot` import to save time). 

I'll give this a proper review tomorrow, as soon as I'm done with the recent high-priority template registration issue that was just opened up. :)

valosekj (commented):
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/c6d7c741ea5909a801ff09e0f52a7b99adc3c15a

Good catch! Thanks for fixing it! 👍🏻

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -41,12 +41,38 @@ def dummy_lesion(request, tmp_path):
...
 
 def compute_expected_measurements(dim, starting_coord=None, path_seg=None):
-    # Find the minimum SC area surrounding the lesion
     if path_seg:
+        # Find the minimum SC area surrounding the lesion
         data_seg = Image(path_seg).data
         min_area = min(np.sum(data_seg[:, n_slice, :])
                        for n_slice in range(starting_coord[1], starting_coord[1] + dim[1]))
 
+        # Find the minimum mid-sagittal tissue bridge width for each LR slice in the lesion
```
joshuacwnewton, line 50: I've added the bridge widths to the "dummy lesion measurements" test.

My goal with this was to re-implement the calculations, to ensure that I fully understanding what's being proposed here.

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -41,12 +41,38 @@ def dummy_lesion(request, tmp_path):
...
+                dorsal_bridge_widths.append(max(0.0, dorsal_sc_width - dorsal_lesion_width))
+
+            # find minimum widths
+            # FIXME: Because t2.nii.gz starts as AIL but is reoriented to RPI, there is a RL flip, meaning the
+            #        z slice number of the AIL image won't correspond to the z slice number used for the column name.
+            z_flipped = data_seg.shape[2] - (z + 1)  # account for 0-based index
+            tissue_bridges[f"slice_{z_flipped}_dorsal_bridge_width [mm]"] = min(dorsal_bridge_widths)
+            tissue_bridges[f"slice_{z_flipped}_ventral_bridge_width [mm]"] = min(ventral_bridge_widths)
+            tissue_bridges[f"slice_{z_flipped}_total_bridge_width [mm]"] = (min(dorsal_bridge_widths) +
+                                                                            min(ventral_bridge_widths))
```
joshuacwnewton, line 74: While writing the test, I noticed that the slice numbers in the column names correspond to the RPI slice numbers, rather than the slice numbers for the original orientation.

However, for this test, the input image is AIL, meaning there is a RL flip. So, the slice numbers in the column name do not match the slice numbers if we were to view the original input images in, say, FSLeyes.

joshuacwnewton, line 74: I also noticed that the "total" value is the sum of the two (separate) ventral/dorsal minimums.

However, if the ventral minimum and the dorsal minimum are on different axial slices, then the computed value will be a different value than "slicewise minimum total width" (i.e. computing, for each slice, the full SC width minus the full lesion width).

Is this intentional?

valosekj:
I implemented the tissue bridges angle correction in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/bbecd8eb5039500d0a3e3257511516562c558898. I'm reusing already [existing angles](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bbecd8eb5039500d0a3e3257511516562c558898/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L722-L740) available in `.self.angles` and multiplying the tissue bridges width by `cos` of the angle between the slice corresponding to the ventral and dorsal tissue bridges and the centerline.

I created a GIF to illustrate the angle correction (GIF refreshes every 5s):

tissue_bridges

joshuacwnewton (commented):
Here is a first draft review, focusing on just the "high level" functionality and results.

I'm happy to give a follow-up review for more nitty-gritty, line-by-line analysis of the `_measure_tissue_bridges` function afterwards. :)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +325,210 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+                                      np.cos(self.angles[min_dorsal_bridge_width_slice]))
+            ventral_bridge_width_mm = (float(df_temp[df_temp['axial_slice'] ==
+                                                     min_ventral_bridge_width_slice]['ventral_bridge_width'] * p_lst[1]) *
+                                       np.cos(self.angles[min_ventral_bridge_width_slice]))
+            total_bridge_width_mm = dorsal_bridge_width_mm + ventral_bridge_width_mm
+
+            # Save the minimum tissue bridges
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_dorsal_bridge_width [mm]'] = dorsal_bridge_width_mm
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_ventral_bridge_width [mm]'] = ventral_bridge_width_mm
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_total_bridge_width [mm]'] = total_bridge_width_mm
```
joshuacwnewton, line 450: Since we create a new column for every sagittal slice, what happens if we have multiple lesions spanning different sagittal slice ranges?

I guess in that case, there will be quite a lot of columns (3 per sagittal slice), but some of the cells will be blank for different lesions (since not all lesions span all slices).

I'm wondering if there is a cleaner way to organize the sheet. I am reminded of the `-f` option of `sct_analyze_lesion`, which produces 1 sheet page per lesion. In that case, we could have "sagittal slice" be one axis, and "dorsal/ventral/total" be the other axis.

(Though, this would mean that the bridge width data would be separate from the other measurements.)

testing/cli/test_cli_sct_analyze_lesion.py
valosekj: Good point!

> Is summing the two separate minimums intentional?

Yes, this is intentional. I got inspiration for this from a SOP for manual measurements (shown below). The ventral bridge (VB) can be located in a different axial slice than the dorsal bridge (DB). `total_bridge` is than just a simple sum of VB and DB.

image

valosekj: Excellent point! 👏🏻  So you basically mean that I'm currently printing slices for RPI (because `sct_analyze_lesion` does RPI reorientation internally) regardless of the user's image orientation (e.g., AIL), right?

>  I am wondering if this might also confuse users?

Yes, this might be definitely confusing!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: > Since we create a new column for every sagittal slice, what happens if we have multiple lesions spanning different sagittal slice ranges?

Very good point! Tbh, I kind of simplified the problem now only for a single lesion because tissue bridges are currently computed exclusively in spinal cord injury (SCI), which is characterized by a single lesion. But the truth is that `sct_analyze_lesion` can also be used for MS patients with multiple lesions... The question is, how clinically relevant are tissue bridges in MS?

joshuacwnewton: > tissue bridges are currently computed exclusively in spinal cord injury (SCI), which is characterized by a single lesion.

This is very important context! Thank you for clarifying.

Maybe since we only want to compute this measure for one specific case, it could be opt-in? (As to not overcomplicate the output if the user has no use for the measure.) And, if we had a CLI argument, then the help description would be a good place where we can document the option (by explaining the relevance for SCI + linking the relevant citation).

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +325,210 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # --------------------------------------
+        # Get minimal tissue bridges
+        # --------------------------------------
+        # Convert the dictionary to a DataFrame (for easier manipulation)
+        # 1. Create a MultiIndex from the dictionary keys
+        index = pd.MultiIndex.from_tuples(tissue_bridges_dict.keys(), names=['sagittal_slice', 'axial_slice'])
+        # 2. Create the DataFrame using the MultiIndex and the dictionary values
+        tissue_bridges_df = pd.DataFrame(list(tissue_bridges_dict.values()), index=index)
+        # 3. Reset the index to make 'sagittal_slice' and 'axial_slice' as columns
+        tissue_bridges_df.reset_index(inplace=True)
```
joshuacwnewton, line 405: I think that you could also keep the `MultiIndex` to simplify indexing here too?

For example:

```python
# before
df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice]
# after
df_temp = tissue_bridges_df[slice(sagittal_slice)]
```

Slicing MultiIndexes is [a bit more advanced](https://pandas.pydata.org/docs/user_guide/advanced.html#using-slicers), but it might be useful to work through as a way of learning `pandas` a little better?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +325,210 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # Get slices of minimum dorsal and ventral tissue bridges for each sagittal slice
+        # NOTE: we get minimum because tissue bridges are quantified as the width of spared tissue at the minimum
+        # distance from cerebrospinal fluid to the lesion boundary
+        for sagittal_slice in sagittal_lesion_slices:
+            # Get df for the selected sagittal slice
+            df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice]
+            # Get the axial slice with the minimum dorsal tissue bridge for the selected sagittal slice
+            min_dorsal_bridge_width_slice = df_temp.loc[df_temp['dorsal_bridge_width'].idxmin(), 'axial_slice']
+            # Get the axial slice with the minimum ventral tissue bridge for the selected sagittal slice
+            min_ventral_bridge_width_slice = df_temp.loc[df_temp['ventral_bridge_width'].idxmin(), 'axial_slice']
```
joshuacwnewton, line 418: I absolutely love your use of comments in most of this method. They really helped me understand the functionality!

I think you may be able to trim down on some of the comments, though? I like to have 1 comment per "block" of related lines. For example, in this case, you're finding the minimum for both ventral/dorsal. So, you could do this instead:

```suggestion
            # Get the axial slices corresponding to the minimum bridge widths
            min_dorsal_bridge_width_slice = df_temp.loc[df_temp['dorsal_bridge_width'].idxmin(), 'axial_slice']
            min_ventral_bridge_width_slice = df_temp.loc[df_temp['ventral_bridge_width'].idxmin(), 'axial_slice']
```

That way, you still get the gist of your comments, and also the single comment helps to visually group the related chunks of code.

(There are a lot of parts of the code that have 1 comment per line -- I think you could probably simplify most of these into 1 comment per "block of lines".)

joshuacwnewton (commented):
Thank you so much for writing such a clear and well-documented method! My comments are mainly focused on clarity + readability, as I think the overall logic is sound. :hearts:

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -34,6 +35,8 @@ def get_parser():
                     'If the proportion of lesion in each region (e.g. WM and GM) does not sum up to 100%, it means '
                     'that the registered template does not fully cover the lesion. In that case you might want to '
                     'check the registration results.'
+                    '- dorsal_bridge_width [mm]: width of spared tissue dorsal to the spinal cord lesion\n'
+                    '- ventral_bridge_width [mm]: width of spared tissue ventral to the spinal cord lesion\n\n'
```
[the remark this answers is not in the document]
valosekj: Good point! Sorry for not documenting this properly. As you correctly pointed out, ventral = anterior and dorsal = posterior.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Thank you! Absolutely agree!

valosekj:
Feedback from collaborators: 

image

--> update the algorithm to restrict the lesion to the cord

image

--> idea: plot anatomical image (i.e., T2w) as background, overlaid by semi-transparent (or just contours) lesion and cord masks

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +325,210 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        tissue_bridges_df.reset_index(inplace=True)
+
+        # Initialize auxiliary dict to store info for plotting
+        plot_dict = {}
+        # Get slices of minimum dorsal and ventral tissue bridges for each sagittal slice
+        # NOTE: we get minimum because tissue bridges are quantified as the width of spared tissue at the minimum
+        # distance from cerebrospinal fluid to the lesion boundary
+        for sagittal_slice in sagittal_lesion_slices:
+            # Get df for the selected sagittal slice
+            df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice]
```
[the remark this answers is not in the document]
valosekj: Very good suggestion! Thank you!
Code updated in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/3d1da0afe9b42c31600d4a27a96b47275d70521f.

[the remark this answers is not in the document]
valosekj: Hmm, looking deeper, this change introduced issues with `axial slices`:

<details><summary>before 3d1da0a </summary>

Screenshot 2024-06-17 at 4 57 09 PM

</details>

<details><summary>after 3d1da0a </summary>

Screenshot 2024-06-17 at 4 57 40 PM

</details>

This can be easily solved by adding `inplace=True` to `df_temp.set_index('axial_slice', inplace=True)`. But then, there are issues with `['axial_slice']` indexing at the following lines:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3d1da0afe9b42c31600d4a27a96b47275d70521f/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L509
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3d1da0afe9b42c31600d4a27a96b47275d70521f/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L517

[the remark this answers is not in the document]
valosekj: Since I wanted to continue working on [suggestions from collaborators](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#issuecomment-2168479780), I am thinking about reverting https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/3d1da0afe9b42c31600d4a27a96b47275d70521f

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Interesting! I tried keeping the `MultiIndex`, but the code became a little bit less readable, tbh. Maybe I would keep indexing as it is for now.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Okay! Even tSCI patients might have multiple lesions (see the last screenshot in [this comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#issuecomment-2168479780)) --> update the script to handle multiple lesions properly.

valosekj: Okay, it seems that the code is already capable of handling multiple lesions! In such a case, statistics (including tissue bridges) for each lesion are stored as individual rows in the output XLS file, for example:

image

I just added this [small commit](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/f030fbad8b13327eef2bcef1ab44d1e273acf23f) to replace `nan` with `n/a` (`nan` results in empty cells in Excel).

Just the [plotting of tissue bridges](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f030fbad8b13327eef2bcef1ab44d1e273acf23f/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L479-L549) is not ready for multiple lesions --> I will update it!

valosekj:
@joshuacwnewton, @mguaypaq, I believe the PR is now ready for a re-review. I also updated the [first comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#issue-2311427775) of this PR to reflect the current behaviour. The link for the [test data](https://www.dropbox.com/scl/fo/r76whcpwidcxhjymdgyqb/AOdKwznb6NBKAJFC8yykpoY?rlkey=6fgp900vjqr0hbbw2ffn99uvf&dl=0) now contains both single and multiple lesion examples to allow PR testing on both single lesion and multiple lesions.
Since I pushed a bunch of new commits from your last review, I tried to make the commit messages clear and helpful. In these new commits, I mainly addressed the feedback from collaborators (tracked in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#issuecomment-2168479780)

joshuacwnewton (changes requested):
Thank you for the update to this PR! I had a few suggestions addressing the test failures, the changes, and the TODOs, while also proposing a structure for a potential QC report.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -734,6 +800,10 @@ def measure(self):
                                                       im_lesion=im_lesion_data_cur,
                                                       p_lst=p_lst)
 
+        # Plot tissue bridges
+        # Note: we do plotting here as we plot all lesions into a single figure, so we need to be outside the loop
+        self._plot_tissue_bridges(im_lesion_data, p_lst, label_lst)
```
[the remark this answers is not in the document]
valosekj: Good catch!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -433,8 +829,14 @@ def measure(self):
         im_lesion_data = im_lesion.data
         p_lst = im_lesion.dim[4:7]  # voxel size
 
+        # TODO: label_lst could become self.label_lst, then we would not need to pass it as an argument to
+        #  _measure_within_im
```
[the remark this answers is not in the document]
valosekj: Thanks for your thoughts. Agree --> removing the comment for now in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/ea3a6cba84808a021931cc02f6a90359d3c50b5e :-)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            # # Calculate the change in x and y based on the dorsal bridge width and the angle
+            # delta_x = dorsal_bridge_width / np.sqrt(1 + slope ** 2)
+            # delta_y = slope * delta_x
+            # # Define start and end points of the line
+            # start_x = x_dorsal
+            # start_y = y_dorsal
+            # end_x = start_x - delta_x
+            # end_y = start_y - delta_y
+            # # Plot the line segment
+            # axes[idx_row, idx_col].plot([start_x, end_x], [start_y, end_y], 'g--', linewidth=1)
```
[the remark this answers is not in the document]
valosekj: Agree :-) Removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/7a363b1b95030d1b6e84951d4d7a186f04ca9b0e

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            # # Calculate the change in x and y based on the dorsal bridge width and the angle
+            # delta_x = ventral_bridge_width / np.sqrt(1 + slope ** 2)
+            # delta_y = slope * delta_x
+            # # Define start and end points of the line
+            # start_x = x_ventral
+            # start_y = y_ventral
+            # end_x = start_x + delta_x
+            # end_y = start_y + delta_y
+            # # Plot the line segment
+            # axes[idx_row, idx_col].plot([start_x, end_x], [start_y, end_y], 'g--', linewidth=1)
```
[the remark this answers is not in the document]
valosekj: Agree :-) Removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/7a363b1b95030d1b6e84951d4d7a186f04ca9b0e

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+
+                # Crop the slice around the lesion (to zoom in)
+                axes[idx_row, idx_col].set_xlim(np.min(np.where(im_label_data_cur)[1]) - 20,
+                                                np.max(np.where(im_label_data_cur)[1]) + 20)
+                axes[idx_row, idx_col].set_ylim(np.min(np.where(im_label_data_cur)[2]) - 20,
+                                                np.max(np.where(im_label_data_cur)[2]) + 20)
+
+                # --------------------------------------
+                # Add line segments and text for tissue bridges
+                # --------------------------------------
```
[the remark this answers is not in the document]
valosekj: Good catch! Comments fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/83f2990e10fd7bd7d7a66efb3efa9a956ca7295b

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            printv(f'  Sagittal slice {sagittal_slice}, Minimum ventral tissue bridge width: '
+                   f'{np.round(ventral_bridge_width_mm, 2)} mm (axial slice {min_ventral_bridge_width_slice})',
+                   self.verbose, type='info')
+            printv(f'  Sagittal slice {sagittal_slice}, Total tissue bridge width: '
+                   f'{np.round(total_bridge_width_mm, 2)} mm', self.verbose, type='info')
+
+    def _plot_tissue_bridges(self, im_lesion_data, p_lst, label_lst):
+        """
+        Plot the tissue bridges.
+        The figure has one row per lesion and one column per sagittal slice containing the lesion.
```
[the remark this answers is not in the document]
valosekj: > would it be possible to create the QC report using only the values found in lesion_analysis.xls

This is an excellent point, actually! After removing the plotting of the line segments (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/7a363b1b95030d1b6e84951d4d7a186f04ca9b0e), we indeed need just sagittal slice number and tissue bridges widths. All of this is already outputted into `lesion_analysis.xls`.

(For context: `axial_slice` and `lesion_indices` were used mainly to draw the line segments. After removing the line segments, these two variables are now used only to [place the text](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/83f2990e10fd7bd7d7a66efb3efa9a956ca7295b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L537-L554). Which can be done easily in another way, though. For example, based on the lesion center of mass.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+
+        # Loop across lesions
+        for idx_row, lesion_label in enumerate(label_lst):
+            # NOTE: As the 'label_lesion()' function has been called at the beginning of the script, im_lesion_data is
+            # now "labeled" meaning that different lesions have different values, e.g., 1, 2, 3
+            # As we are looping across lesions, we get the lesion mask for the current lesion label
+            im_label_data_cur = im_lesion_data == lesion_label
+            # Restrict the lesion mask to the spinal cord mask (from anatomical level, it does not make sense to have
+            # lesion outside the spinal cord mask)
+            im_label_data_cur = im_label_data_cur * im_sc_data
```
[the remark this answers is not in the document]
valosekj: > Is there any chance that the SC could be soft?

Good question!

Right now, SCIseg outputs binary SC and binary lesion. But theoretically, users can use the contrast-agnostic model to get SC and then it would be soft. To be ready for this case, I updated the code based on your suggestion in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/9bc734a56e10e3552e5f974aa14bed2838fc907c.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -219,6 +219,10 @@ def tmp2ofolder(self):
                  os.path.join(self.path_ofolder, self.tissue_bridges_png_name))
 
     def pack_measures(self):
+
+        # Replace 'nan' with 'n/a' ('nan' results in empty cells in Excel)
+        self.measure_pd = self.measure_pd.replace(np.nan, 'n/a')
```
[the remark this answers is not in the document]
valosekj: > was there a specific issue with having nan/empty cells, rather than 'n/a'?

Good question! I think I just wanted to avoid empty cells in Excel:

image

I tried to comment the line in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/1d4641408b269d71e27c59177fe64b2aac38ec29 to see whether the tests pass

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +359,354 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # outside the spinal cord mask)
+        im_lesion_data = im_lesion_data * im_sc_data
+
+        # Get and print the midsagittal slice of the spinal cord (count all slices with the spinal cord mask and get
+        # the middle slice)
+        mid_sagittal_sc_slice = int(np.mean([np.min(np.unique(np.where(im_sc_data)[0])),
+                                             np.max(np.unique(np.where(im_sc_data)[0]))]))
+        # Note: as the midsagittal slice is computed from the spinal cord mask, it is the same for all lesions (all idx)
+        # TODO: consider whether it is necessary to print the midsagittal slice for each lesion as it is the same for
+        #  all lesions. This would require moving the print statement outside the loop across lesions.
```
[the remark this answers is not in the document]
valosekj: > Or... maybe at that point, it's getting too complicated, so we should just leave the duplicated column, aha.

Agree, haha.

valosekj:
> Thank you for the update to this PR! I had a few suggestions addressing the test failures, the changes, and the TODOs, while also proposing a structure for a potential QC report.

Thank you for the review! I addressed/replied to your comments and fixed the test failures in my recent commits.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
joshuacwnewton: It looks like the tests pass! 

To me, empty cells seem clear enough to me? But, maybe there is something I'm missing.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
joshuacwnewton: Before we go ahead and make the changes, I'm just curious: Why does the figure use the rounded values (3.0mm, 2.0mm) but the xls prints the angle-corrected values (2.997796, 1.998531).

We should probably display the angle-corrected values, right?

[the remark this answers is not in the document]
valosekj: Good point! 
The figure uses `np.round`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/42c77503619dced05be5af770276b560740fe5f7/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L554

while `self.measure_pd` stores (and saves to XLS) floats:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/42c77503619dced05be5af770276b560740fe5f7/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L230 

image

I checked publications (for example Table 1 in [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6815206/)) that use tissue bridges, and I believe we can stick with two decimals for both the figure and XLS file

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
valosekj: I think empty cells are not ideal for CSV, for example, when printing CSV in CLI using `cat.` But here, as we use XLS, we can go with empty cells, I hope.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
joshuacwnewton: Perfect! I'll go ahead and implement the QC report, then. :)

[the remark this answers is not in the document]
joshuacwnewton: The QC report has been added! It now accepts `measure_pd` as the only input (besides the images themselves), which removes the need to access `tissue_bridges_plot_dict`, `p_lst`, `angles`, etc.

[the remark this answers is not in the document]
valosekj: Awesome! Thank you, @joshuacwnewton! 💯 I also love your commit messages, making it easy to understand the changes.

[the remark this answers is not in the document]
valosekj: I tested the updated PR, and I could successfully generate the QC for the single lesion. When testing on an image with multiple lesions, I get the following error: 

```python
  File "/Users/valosek/code/sct_nnunet/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 873, in <module>
    main(sys.argv[1:])
  File "/Users/valosek/code/sct_nnunet/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 840, in main
    sct_analyze_lesion(
  File "/Users/valosek/code/sct_nnunet/spinalcordtoolbox/reports/qc2.py", line 468, in sct_analyze_lesion
    axes[idx_row, idx_col].text(min(np.where(slice_lesion)[0]) - 3,
ValueError: min() arg is an empty sequence
```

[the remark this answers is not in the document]
valosekj: Okay, this was an easy fix :-) Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/2df1fa019282fe12064dd820022533c963d7bb33; see the commit message for details.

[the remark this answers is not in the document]
valosekj: Two last things:
1. we should probably add a test for multiple lesions
2. do we want to generate the QC in default (as done now)? Or maybe it would be better to generate the QC only when `sct_analyze_lesion.py -qc` is used.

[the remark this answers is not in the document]
joshuacwnewton: Good idea! Let's go with 2., since that seems to be the default behavior of most scripts. (I'm not quite sure why I did it this way, aha.)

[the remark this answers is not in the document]
joshuacwnewton: Fixes added!

joshuacwnewton (approved):
I think this might be good to go? (Note: The test failures are due to #4570, I believe.)

But, maybe another review from Mathieu of my additions could be good, too.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -343,6 +345,146 @@ def sct_deepseg(
...
+
+        # Loop across lesions
+        for idx_row, lesion_label in enumerate(label_lst):
+            # NOTE: As the 'label_lesion()' function has been called at the beginning of the script, im_lesion_data is
+            # now "labeled" meaning that different lesions have different values, e.g., 1, 2, 3
+            # As we are looping across lesions, we get the lesion mask for the current lesion label
+            im_label_data_cur = im_lesion_data == lesion_label
+            # Restrict the lesion mask to the spinal cord mask (from anatomical level, it does not make sense to have
+            # lesion outside the spinal cord mask)
+            boolean_mask = (im_sc_data.data != 0)  # Nonzero -> True | Zero -> False; we use this in case of soft SC
```
mguaypaq, line 420: I'm pretty sure this is a typo:
```suggestion
            boolean_mask = (im_sc_data != 0)  # Nonzero -> True | Zero -> False; we use this in case of soft SC
```

I was surprised that the code didn't crash, but it turns out that numpy arrays happen to have a `data` attribute:
```py3
>>> from spinalcordtoolbox.image import Image
>>> im = Image('data/sct_example_data/t2/t2.nii.gz')
>>> type(im.data)
<class 'numpy.ndarray'>
>>> type(im.data.data)
<class 'memoryview'>
```

However trying to get a boolean mask from it gives a constant, not a mask:
```py3
>>> type(im.data != 0)
<class 'numpy.ndarray'>
>>> type(im.data.data != 0)
<class 'bool'>
```

spinalcordtoolbox/reports/qc2.py
```diff
@@ -343,6 +345,146 @@ def sct_deepseg(
...
+            im_label_data_cur = im_label_data_cur * boolean_mask
```
mguaypaq, line 421: How about defining `im_label_data_cur` directly as the `np.logical_and()` of the two boolean conditions?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -343,6 +345,146 @@ def sct_deepseg(
...
+                    if col_name_ventral in measure_pd.columns and not pd.isna(ventral_bridge_width_mm):
+                        axes[idx_row, idx_col].text(max(np.where(slice_lesion)[0]) + 3,
+                                                    min(np.where(slice_lesion)[1]),
+                                                    f'Ventral bridge\n{np.round(ventral_bridge_width_mm, 2)} mm',
+                                                    color='red', fontsize=12, ha='left', va='bottom')
+        # tight layout
+        mpl_plt.tight_layout()
+        for fname in imgs_to_generate.values():
+            mpl_plt.savefig(fname)
+        mpl_plt.close()
```
mguaypaq, line 485: What does this `close()` do? I don't think we ever open a figure window? (Since I saw a `close()` call I instinctively tried to look up whether there was a pyplot context manager, but I only got more confused.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -205,6 +226,10 @@ def tmp2ofolder(self):
             copy(os.path.join(self.tmp_dir, file_), os.path.join(self.path_ofolder, file_))
 
     def pack_measures(self):
+
+        # Replace 'nan' with 'n/a' ('nan' results in empty cells in Excel)
+        # self.measure_pd = self.measure_pd.replace(np.nan, 'n/a')
+
```
mguaypaq, line 232: Why is there some commented-out code here? Should it be removed?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -230,14 +255,42 @@ def show_total_results(self):
...
+            # Iterate across lesions to get the bridges for the midsagittal slice
+            for idx, row in self.measure_pd.iterrows():
+                if row['midsagittal_spinal_cord_slice'] is not None:        # just safety check
+                    # Get the midsagittal slice number for the selected lesion
+                    # Note that the midsagittal slice is the same for all lesions as it is based on the spinal cord
+                    # segmentation
+                    midsagittal_slice = str(int(row['midsagittal_spinal_cord_slice']))
+                    # Get dorsal and ventral tissue bridges for the mid-sagittal slice
+                    dorsal_tissue_bridge = row['slice_' + midsagittal_slice + '_dorsal_bridge_width [mm]']
+                    ventral_tissue_bridge = row['slice_' + midsagittal_slice + '_ventral_bridge_width [mm]']
```
mguaypaq, line 280: Nitpick: these could be f-strings.
```suggestion
                    dorsal_tissue_bridge = row[f'slice_{midsagittal_slice}_dorsal_bridge_width [mm]']
                    ventral_tissue_bridge = row[f'slice_{midsagittal_slice}_ventral_bridge_width [mm]']
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +364,153 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+
+        # Get slices of minimum dorsal and ventral tissue bridges for each sagittal slice
+        # NOTE: we get minimum because tissue bridges are quantified as the width of spared tissue at the minimum
+        # distance from cerebrospinal fluid to the lesion boundary
+        for sagittal_slice in sagittal_lesion_slices:
+            # Get df for the selected sagittal slice
+            df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice]
+            # Get the axial slices corresponding to the minimum bridge widths
+            min_dorsal_bridge_width_slice = df_temp.loc[df_temp['dorsal_bridge_width'].idxmin(), 'axial_slice']
+            min_ventral_bridge_width_slice = df_temp.loc[df_temp['ventral_bridge_width'].idxmin(), 'axial_slice']
```
mguaypaq, line 474: It looks like you're finding the minimum _before_ doing angle correction, is that ok?

I guess either:
- the angle is almost the same for the whole lesion, in which case this is approximately correct; or
- the angle changes significantly over the length of the lesion, in which case angle correction is flawed anyway.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +364,153 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+                                      np.cos(self.angles[min_dorsal_bridge_width_slice]))
+            ventral_bridge_width_mm = (float(df_temp[df_temp['axial_slice'] == min_ventral_bridge_width_slice]
+                                             ['ventral_bridge_width'] * p_lst[1]) *
+                                       np.cos(self.angles[min_ventral_bridge_width_slice]))
+            total_bridge_width_mm = dorsal_bridge_width_mm + ventral_bridge_width_mm
+
+            # Save the minimum tissue bridges
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_dorsal_bridge_width [mm]'] = dorsal_bridge_width_mm
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_ventral_bridge_width [mm]'] = ventral_bridge_width_mm
+            self.measure_pd.loc[idx, f'slice_{str(sagittal_slice)}_total_bridge_width [mm]'] = total_bridge_width_mm
```
mguaypaq, line 504: I think these explicit calls to `str()` are redundant?
```suggestion
            self.measure_pd.loc[idx, f'slice_{sagittal_slice}_dorsal_bridge_width [mm]'] = dorsal_bridge_width_mm
            self.measure_pd.loc[idx, f'slice_{sagittal_slice}_ventral_bridge_width [mm]'] = ventral_bridge_width_mm
            self.measure_pd.loc[idx, f'slice_{sagittal_slice}_total_bridge_width [mm]'] = total_bridge_width_mm
```

mguaypaq (approved):
I didn't look very deeply, but this looks reasonable. Thanks for the new feature!

spinalcordtoolbox/reports/qc2.py
```diff
@@ -343,6 +345,146 @@ def sct_deepseg(
...
+
+        #  Create a figure
+        #  The figure has one row per lesion and one column per sagittal slice containing the lesion
+        fig, axes = mpl_plt.subplots(num_of_lesions,
+                                     num_of_sag_slices,
+                                     figsize=(num_of_sag_slices * 5, num_of_lesions * 5))
+
+        # Force axes to be a 2-dimensional array (to avoid indexing issues if we have only a single lesion or a single
+        # sagittal slice)
+        axes = np.atleast_2d(axes)
```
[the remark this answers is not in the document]
valosekj: Hmm, interesting. I tried `np.atleast_2d(axes)` with `num_of_sag_slices == 1`, nad it worked fine.
Now I tried `np.asanyarray(axes).reshape((num_of_lesions, num_of_sag_slices))` with both `num_of_sag_slices == 1` and `num_of_sag_slices > 1` and it also works. So I believe it is save to commit this change.

spinalcordtoolbox/reports/qc2.py
valosekj: good catch! thanks!

spinalcordtoolbox/reports/qc2.py
valosekj: something like this: `im_label_data_cur = np.logical_and(im_label_data_cur, boolean_mask)`?

spinalcordtoolbox/reports/qc2.py
valosekj: good point. `close()` indeed seems be unnecessary. Maybe it stayed here from the developing phase when I opened the figure for debugging.

--> removing `close()`

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
valosekj: Thank you! Tested, and the code works fine with the `-qc` flag.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: good catch!
this is legacy of https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#discussion_r1670859216 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/1d4641408b269d71e27c59177fe64b2aac38ec29
We can remove these lines

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: oops! thank you for catching this! we should actually do angle correction first and then find the minimum tissue width!
I fixed the code and described the change in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/031339784d1a3a2d58df7669ba824a9368c83f1d

spinalcordtoolbox/reports/qc2.py
valosekj: Okay, I see! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/d5a66f48254bcb9e0b6d1465728fcd35f8b98dbb

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: you are right, thanks! `str()` removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489/commits/1f50b3b05f6c90351857da8ef0df69c087169719

valosekj:
Thanks for the reviews! 

If you agree with the change made in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4489#discussion_r1690263286, I believe it is good to go for now. And we can potentially continue with further improvements/tweaking in a new branch.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -466,47 +466,50 @@ def _measure_tissue_bridges(self, im_lesion_data, p_lst, idx):
...
-                                            ['dorsal_bridge_width'] * p_lst[1]) *
-                                      np.cos(self.angles[min_dorsal_bridge_width_slice]))
-            ventral_bridge_width_mm = (float(df_temp[df_temp['axial_slice'] == min_ventral_bridge_width_slice]
-                                             ['ventral_bridge_width'] * p_lst[1]) *
-                                       np.cos(self.angles[min_ventral_bridge_width_slice]))
-            total_bridge_width_mm = dorsal_bridge_width_mm + ventral_bridge_width_mm
+            dorsal_bridge_width_mm = df_temp.apply(lambda row: row['dorsal_bridge_width'] * p_lst[1] *
+                                                               np.cos(self.angles[row['axial_slice']]), axis=1)
+            ventral_bridge_width_mm = df_temp.apply(lambda row: row['ventral_bridge_width'] * p_lst[1] *
+                                                                np.cos(self.angles[row['axial_slice']]), axis=1)
```
mguaypaq: From in-person discussion: I'm concerned that we may be using the wrong angle here for this purpose. `self.angles[i]` is the 3D angle between vertical (the z-direction) and the spinal cord centerline. But here, we want to be using the 2D angle in the sagittal plane. (The difference would occur mostly when there is some scoliosis, I think.)

I'll suggest a fix for this in an extra commit.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -311,6 +361,158 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+            # Get the axial slices corresponding to the minimum bridge widths
+            # This information is printed to terminal
+            min_dorsal_bridge_width_slice = df_temp.loc[df_temp['dorsal_bridge_width_mm'].idxmin(), 'axial_slice']
+            min_ventral_bridge_width_slice = df_temp.loc[df_temp['ventral_bridge_width_mm'].idxmin(), 'axial_slice']
+
+            # Get the minimum dorsal and ventral bridge widths
+            min_dorsal_bridge_width_mm = float(df_temp[df_temp['axial_slice'] == min_dorsal_bridge_width_slice]
+                                               ['dorsal_bridge_width_mm'])
+            min_ventral_bridge_width_mm = float(df_temp[df_temp['axial_slice'] == min_ventral_bridge_width_slice]
+                                                ['ventral_bridge_width_mm'])
```
mguaypaq, line 500: You can get the minimum value more directly:
```suggestion
            min_dorsal_bridge_width_mm = float(df_temp['dorsal_bridge_width_mm'].min())
            min_ventral_bridge_width_mm = float(df_temp['ventral_bridge_width_mm'].min())
```

valosekj: Good point! Agree. We can commit this suggestion

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Thank you for the quick fix! Fortunately, the impact of this is fix is very subtle (second decimal and only for some slices -- see cursor):

Kapture 2024-07-25 at 14 27 23

mguaypaq (approved):

spinalcordtoolbox/scripts/sct_analyze_lesion.py
naga-karthik: Nice, thanks for this subtle fix, Mathieu! ❤️

Issue #3077, opened 2020-12-04
Lazy-load CLI scripts to reduce script execution time


joshuacwnewton:
>  **Note**: This issue has been created to document an old WIP pull request that had become stale. No issue had been opened for that PR, so this issue is to discuss whether it's worth starting fresh and continuing that work.

Before it was closed, PR #1913 had aimed to reduce script execution time by lazy-loading modules (i.e. by importing them in the middle of the script, rather than at the top). 

_(That PR also included some refactoring that turned some scripts into API modules, but I don't quite understand why those changes were included given the purpose of that PR.)_

Is this an idea worth pursuing again?

jcohenadad:
> Is this an idea worth pursuing again?

I don't know... @zougloub @kousu @Drulex any insights?

kousu:
It sounds like a nice idea. But would you be up to running profiling on it before diving in?

https://docs.python.org/3/library/profile.html#instant-user-s-manual

```
python -m cProfile -s cumulative scripts/sct_run_whatever.py
```

This line might not work as-is because of the weird way we install and import certain parts but I think you can probably get around that.

My rule of thumb is that python startup time is dominated by

1. the interpreter itself
2. large extension modules like matplotlib and scipy

Drulex:
Relevant: https://github.com/ivadomed/ivadomed/issues/510

PR #4490, opened 2024-05-22
Avoid unconditionally importing expensive packages to reduce load times for CLI scripts 

### LINKED ISSUE #4370: Slow imports
This is a non-urgent issue that I noticed during #4345 -- running `sct_deepseg -h` takes a noticeably long time. 

I'm curious as to why this is, and I would love to try to reduce the "startup" time for scripts if I can.

joshuacwnewton:
## Description

Here is some important background information from my initial research in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4370#issuecomment-2087620810:

<details>
<summary> PEP 690 - Lazy Imports </summary> 

----

I had the question: "Python style guides generally recommend keeping imports at the top of each module. Isn't lazy-loading incompatible with this? What is the "Pythonic" way to reduce import times?"

It turns out that [**PEP 690 – Lazy Imports**](https://peps.python.org/pep-0690/) (as well as [its rejection](https://discuss.python.org/t/pep-690-lazy-imports-again/19661/26)) explicitly references this dilemma:

> Common Python code style [prefers](https://peps.python.org/pep-0008/#imports) imports at module level, so they don’t have to be repeated within each scope the imported object is used in, and to avoid the inefficiency of repeated execution of the import system at runtime. This means that importing the main module of a program typically results in an immediate cascade of imports of most or all of the modules that may ever be needed by the program.
>
> Consider the example of a Python command line program (CLI) with a number of subcommands. Each subcommand may perform different tasks, requiring the import of different dependencies. But a given invocation of the program will only execute a single subcommand, or possibly none (i.e. if just --help usage info is requested). Top-level eager imports in such a program will result in the import of many modules that will never be used at all; the time spent (possibly compiling and) executing these modules is pure waste.

I couldn't have written a more accurate summary of our current issue.

----

</details>

tl;dr: CLI scripts are uniquely susceptible to import slowdowns, because some options may use expensive packages, but other options may not. In the extreme case, a command line `script --help` shouldn't ever need to load packages like `torch`. Yet, because we don't lazy load our expensive packages, our CLI scripts take quite a long time to load!

----

I wrote a quick test that will check the execution time for `script -h`, then fail if it's over a certain threshold (e.g. 2 seconds). Here are all of the scripts that take longer than 2s on `master`, sorted by duration:

<details>

image

</details>

Here are the times for `script -h` after implementing the fixes in this PR:

<details>

image

</details>

We're really close to getting every script under 2s, and I think we could get there by lazy importing `matplotlib` in `qc.py`/`qc2.py`, but because our imports are spread out across several submodules, this is a bit tedious to do. I want to do it, but I'll follow up in another commit later on this PR. (Just wanted to get the ball rolling on this PR.)

----

Much of the slowdown actually came from `inspect.stack()`, so a refactor to remove calls to this function reduced times across our entire CLI, without needing to do any lazy importing at all. (But, lazy importing took care of much of the remaining slowdowns.)

## Linked issues

Fixes #4370.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3077.


spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -18,14 +18,17 @@
...
-from spinalcordtoolbox.deepseg import models, inference
 from spinalcordtoolbox.image import splitext, Image, check_image_kind
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, display_viewer_syntax, ActionCreateFolder
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel, __version__, _git_info
 from spinalcordtoolbox.utils.fs import tmp_create
+from spinalcordtoolbox.utils.sys import lazy_import
+
+cuda = lazy_import('torch.cuda')
+
+qc2 = lazy_import('spinalcordtoolbox.reports.qc2')
```
joshuacwnewton, line 29: TODO: Replace this lazy import with `matplotlib` lazy imports inside both `qc2.py` and `qc.py`.

joshuacwnewton:
It looks like I seriously misunderstood how lazy importing affects import _submodules_. I am in the process of reworking this PR, but it will involve a big rebase, so please disregard the current state of this PR at the moment.

PR #4493, opened 2024-05-24
Use Tensorflow's simple `LazyLoader` class to minimize import times of expensive packages

### LINKED ISSUE #4370: Slow imports

joshuacwnewton:
> [!NOTE]
> This PR is largely the same as #4490, however it swaps the use of `importlib`'s [`lazy_import`](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports) for a copy of Tensorflow's `LazyLoader`, which I've found to perform much more predictably. I had been struggling with `lazy_import` for much of the day (https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/57e3bf272e7be5eccf03322f63bedcaa8f7df86c), trying to track down obscure bugs and weird performance issues, until I finally decided to eschew it for something that's a little more battle-tested (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4493/commits/b265cbf64af41b8d7d673b746c0a725fbaa37887).

## Description

Here is some important background information from my initial research in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4370#issuecomment-2087620810:

<details>
<summary> PEP 690 - Lazy Imports </summary> 

----

I had the question: "Python style guides generally recommend keeping imports at the top of each module. Isn't lazy-loading incompatible with this? What is the "Pythonic" way to reduce import times?"

It turns out that [**PEP 690 – Lazy Imports**](https://peps.python.org/pep-0690/) (as well as [its rejection](https://discuss.python.org/t/pep-690-lazy-imports-again/19661/26)) explicitly references this dilemma:

> Common Python code style [prefers](https://peps.python.org/pep-0008/#imports) imports at module level, so they don’t have to be repeated within each scope the imported object is used in, and to avoid the inefficiency of repeated execution of the import system at runtime. This means that importing the main module of a program typically results in an immediate cascade of imports of most or all of the modules that may ever be needed by the program.
>
> Consider the example of a Python command line program (CLI) with a number of subcommands. Each subcommand may perform different tasks, requiring the import of different dependencies. But a given invocation of the program will only execute a single subcommand, or possibly none (i.e. if just --help usage info is requested). Top-level eager imports in such a program will result in the import of many modules that will never be used at all; the time spent (possibly compiling and) executing these modules is pure waste.

I couldn't have written a more accurate summary of our current issue.

----

</details>

tl;dr: CLI scripts are uniquely susceptible to import slowdowns, because some options may use expensive packages, but other options may not. In the extreme case, a command like `script --help` shouldn't ever need to load packages like `torch`. Yet, because we don't lazy load our expensive packages, our CLI scripts take quite a long time to load!

----

I wrote a quick test that will check the execution time for `script -h`, then fail if it's over a certain threshold (e.g. 2 seconds). Here are all of the scripts that take longer than 2s on `master`, sorted by duration:

<details>

image

</details>

Here are the times for `script -h` after implementing the fixes in this PR:

<details>

image

</details>

There are probably more gains to be had (by lazy-loading e.g. `nibabel` and `numpy`), but at this point, with script times at ~1s, I think any changes are likely to be diminishing returns. Plus, our usage of `nibabel` and `numpy` are widespread, which would require a lot of tedious updating for every single usage.

----

Much of the slowdown actually came from `inspect.stack()`, so a refactor to remove calls to this function reduced times across our entire CLI, without needing to do any lazy importing at all. (But, lazy importing took care of much of the remaining slowdowns.)

## Linked issues

Fixes #4370.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3077.

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -48,7 +48,8 @@ check_dependencies() {
 
 run_tests() {
   activate_venv_sct
-  pytest testing/api testing/cli
+  # re-run failed tests, just in case something flaky has happened. if no failures, do nothing.
+  pytest testing/api testing/cli || pytest --last-failed --last-failed-no-failures none
   # NB: 'testing/batch_processing' is run by a separate CI workflow
 }
 
--- a/contrib/tensorflow/lazy_loader.py
+++ b/contrib/tensorflow/lazy_loader.py
@@ -1,0 +1,86 @@
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""A LazyLoader class."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import importlib
+import types
+import logging
+
+
+# Note: Prior to adopting Tensorflow's LazyLoader class, we tried instead to use the `lazy_import` magic function,
+# as detailed in https://docs.python.org/3/library/importlib.html#implementing-lazy-imports.
+#
+# However, this function was a little too "magical": There were quite a few behind-the-scenes side effects that made
+# the lazy behavior unpredictable:
+#   - Importing a parent module would throw an `AttributeNotFoundError` when trying to access submodules/classes/etc.
+#   - Importing a submodule instead (e.g. `lazy_import("sklearn.metrics")` would inadvertently trigger imports in the
+#     parent module during the call to `loader.exec_module(module)`.
+#   - There seemed to be complex side effects for our deep learning packages, especially those that depended on `torch`.
+#     Lazy loading `torch` seemed to cause obscure errors for other packages such as `monai`.
+#
+# Because we are trying to lazy load deep learning packages, it made sense to look to the Python DL ecosystem for
+# lazy loading implementations. And, luckily, Tensorflow has an incredibly straightforward implementation that removes
+# much of the magic that comes with the `ModuleSpec` and `LazyLoader` classes from importlib.
+#
+# This class is also in use as-is in many downstream DL packages, signifying its reliability. For example:
+#  - https://github.com/bentoml/BentoML/blob/df2c7b86decbdff053002bddf7beb228d012a895/src/bentoml/_internal/utils/lazy_loader.py
+
+
+class LazyLoader(types.ModuleType):
+    """Lazily import a module, mainly to avoid pulling in large dependencies.
+
+    `contrib`, and `ffmpeg` are examples of modules that are large and not always
+    needed, and this allows them to only be loaded when they are used.
+    """
+
+    # The lint error here is incorrect.
+    def __init__(self, local_name, parent_module_globals, name, warning=None):  # pylint: disable=super-on-old-class
+        self._local_name = local_name
+        self._parent_module_globals = parent_module_globals
+        self._warning = warning
+
+        super(LazyLoader, self).__init__(name)
+
+    def _load(self):
+        """Load the module and insert it into the parent's globals."""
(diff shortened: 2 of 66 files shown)


contrib/tensorflow/lazy_loader.py
mguaypaq, line 1: Thankfully this `LazyLoader` class is pretty short, relatively straightforward, and has some good comments. So even though it uses some deeper features (`globals()`, `__getattr__`, `__dir__`, `__dict__`), we should be able to debug hypothetical future problems.

testing/cli/test_cli.py
```diff
@@ -18,12 +19,13 @@
...
+    """
+    Test that SCT's scripts all return error code 2 and show usage descriptions when called with no arguments.
+    Also, ensure that calling the help takes under 2.0 seconds per script.
+    """
+    start_time = time.time()
+    completed_process = subprocess.run([script], capture_output=True)
+    duration = time.time() - start_time
+    assert completed_process.returncode == 2
+    assert b'usage' in completed_process.stderr
+    assert duration < 2.0, f"Expected '{script} -h' to execute in under 2.0s, but took {duration}"
```
mguaypaq, line 31: If this test becomes flaky in the future, we can probably just raise the threshold, so I'm not too worried about it.

spinalcordtoolbox/utils/sys.py
```diff
@@ -98,16 +100,7 @@ class ANSIColors16(object):
...
-    """Return the first non-`utils.sys` module in the stack (to see where `utils.sys` is being called from)."""
-    for frame in inspect.stack():
-        mod = inspect.getmodule(frame[0])
-        if mod.__file__ != __file__:
-            break
-    return mod
-
-
-def set_loglevel(verbose):
+def set_loglevel(verbose, caller_module_name):
```
mguaypaq, line 103: I like that this is now explicit, rather than magic.

spinalcordtoolbox/utils/shell.py
mguaypaq, line 1: Given that we only ever call `SCTArgumentParser` with `description` and, occasionally, `epilog`, the whole `update_parent_default` things seems unnecessary. Why not replace `__init__(self, *args, **kwargs)` with `__init__(self, description, epilog=None)`? It's not like we're writing a library and need to be super flexible; we control both the callers and the callee.

testing/api/test_utils_shell.py
```diff
@@ -23,7 +23,7 @@ def test_sct_argument_parser(capsys):
 
     # Check that new defaults are set properly
     parser3 = shell.SCTArgumentParser()
-    assert parser3.prog == "test_utils_shell"
+    # assert parser3.prog == "test_utils_shell"  this may actually be `pytest`, `sct_testing`, etc. -- hard to test
```
mguaypaq, line 26: Some parts of this test are clearly written to match the code, rather than being independent requirements/specs. We could probably get rid of those low-value parts.

mguaypaq (approved):
Thanks for this massive effort, and the research that went into it! It'll be a great quality-of-life improvement.

joshuacwnewton:
Since the RTD failures are upstream (see above issue), this should be good to merge. :)

.ci.sh
```diff
@@ -48,7 +48,8 @@ check_dependencies() {
 
 run_tests() {
   activate_venv_sct
-  pytest testing/api testing/cli
+  # re-run failed tests, just in case something flaky has happened. if no failures, do nothing.
+  pytest testing/api testing/cli || pytest --last-failed --last-failed-no-failures none
```
joshuacwnewton, line 52: I've added a command here to re-run failed tests, that also [plays nicely with parametrized tests](https://docs.pytest.org/en/latest/how-to/cache.html#rerunning-only-failures-or-failures-first), too. That way, the `2.0s` threshold will only trigger an error if a script takes longer than 2 seconds in back-to-back runs.

testing/cli/test_cli.py
joshuacwnewton: I've added a command here to re-run failed tests, that also [plays nicely with parametrized tests](https://docs.pytest.org/en/latest/how-to/cache.html#rerunning-only-failures-or-failures-first), too. That way, the `2.0s` threshold will only trigger an error if a script takes longer than 2 seconds in back-to-back runs.

I figure that this is a little more useful than increasing the threshold? Since it keeps a more strict goal for our scripts. (I still want the test to trip if a contributor adds a non-LazyLoaded component that slows down a script.)

PR #4496, opened 2024-05-27
Improve documentation of the `-f` argument to make the percentages clearer

### LINKED ISSUE #4394: Usage of the `sct_analyze_lesion -f` option is unclear (percentage distributions)
Hi all, 

I am currently working on a large-scale MS lesion analysis project using whole spine scans for which I (1) register all images to the PAM50 (2) warp PAM50 atlases to the patient space (3) run sct_analyze_lesions using the lesions in patient space and the warped template. After taking a thorough look at the analyze_lesion_results, I suspect a bug in the analysis pipeline.

### Steps to Reproduce

1. Using the newest SCT v6.2 release with (updated) PAM50 Template, please run the following command:

`sct_analyze_lesion -m lesion_mask.nii.gz -f /path/to/warped/PAM50 -i /path/to/image`

2. Having a look at the results, specifically the xls file generated by the command, we have noticed some unexpected things:

The xls file typically consists of multiple sheets, i.e. measures, lesion1#_distribution, lesion2#_distribution, ..., lesionN#_distribution, and finally ROI_occupied_by_lesion.

- Given the lesion volume statistics in the "measure" sheets, we notice that they do not correspond to the total lesion volumes in the corresponding lesion#N_distribution sheet. This may be the case as the individual atlas labels might be overlapping, i.e. the intersection is not exclusive to one region, and thus may result in higher intersection summation results? However, we also notice that the total volume sum is wrong (even within a sheet), specifically it is always "100".

- Given the miscalculated lesion volume sheets, the "ROI_occupied_by_lesion" seems off as well (especially if compared to the measures stats.) We are unsure if this is related to the preceding analysis, and thus an effect of a bug, or is a separate issue.

2. To reproduce results of all pipeline stages I have shared slected images (and the entire pipeline script) via a link privately shared via slack. Otherwise, have a look at the xls file I have attached here :+1: 
[tum_acq-ax_T2w_lesion_analysis.xls](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/14526657/tum_acq-ax_T2w_lesion_analysis.xls)

Please let me know, how I can assist :slightly_smiling_face: 

-Julian

joshuacwnewton:
## Description

This PR improves the documentation for the `-f` argument to hopefully make interpreting the `lesion_statistics.xls` spreadsheet easier.

Also, this PR adds a very basic test (to make running `sct_analyze_lesion -f` easier for devs).

## Linked issues

Fixes #4394.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -67,9 +67,17 @@ def get_parser():
     optional.add_argument(
         "-f",
         help="Path to folder containing the atlas/template registered to the anatomical image. If provided, the "
-             "function computes: (i) the distribution of each lesion depending on each vertebral level and on each "
-             "region of the template (e.g. GM, WM, WM tracts) and (ii) the proportion of ROI (e.g. vertebral level, "
-             "GM, WM) occupied by lesion.",
+             "function computes:\n"
+             "\n"
+             "  a. for each lesion, the proportion of that lesion within each vertebral level and each region "
+             "of the template (e.g. GM, WM, WM tracts). Each cell contains a percentage value representing how much of "
+             "the lesion volume exists within the region indicated by the row/column (rows represent vertebral levels, "
+             "columns represent ROIs). The percentage values are summed to totals in both the bottom row and the right "
+             "column, and the sum of all cells is 100 (i.e. 100 percent of the lesion), found in the bottom-right.\n"
+             "  b. the proportions of each ROI (e.g. vertebral level, GM, WM) occupied by lesions.\n"
+             "\n"
+             "These percentage values are stored in different pages of the output `lesion_analysis.xls` spreadsheet;"
+             "one page for each lesion (a.) plus a final page summarizing the total ROI occupation of all lesions (b.)",
         metavar=Metavar.str,
         default=None,
         required=False)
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -67,9 +67,17 @@ def get_parser():
...
+             "\n"
+             "  a. for each lesion, the proportion of that lesion within each vertebral level and each region "
+             "of the template (e.g. GM, WM, WM tracts). Each cell contains a percentage value representing how much of "
+             "the lesion volume exists within the region indicated by the row/column (rows represent vertebral levels, "
+             "columns represent ROIs). The percentage values are summed to totals in both the bottom row and the right "
+             "column, and the sum of all cells is 100 (i.e. 100 percent of the lesion), found in the bottom-right.\n"
+             "  b. the proportions of each ROI (e.g. vertebral level, GM, WM) occupied by lesions.\n"
+             "\n"
+             "These percentage values are stored in different pages of the output `lesion_analysis.xls` spreadsheet;"
+             "one page for each lesion (a.) plus a final page summarizing the total ROI occupation of all lesions (b.)",
```
joshuacwnewton, line 80: @jqmcginnis: Do you think this description would have helped? Is there anything that might be missing/unclear?

jqmcginnis: I think this is very helpful and clarifies the misunderstandings I initially had. Thank you for adding this!

joshuacwnewton:
Given that @jqmcginnis has given approval, and given that this is a simple documentation change, I'm going to go ahead and merge. :)

PR #4499, opened 2024-05-30
Add `-volumewise` to process the individual 3D volumes of 4D input images

### LINKED ISSUE #4181: Support volume-wise 3D operations on 4D images
This is a feature request that came out of the SCT dev meeting on 2023-07-27 ([minutes](https://docs.google.com/document/d/1ItApJQfajO2lRzOU2yenWbeRg6alfsdut3J4AVVdo78/edit#heading=h.3jazj3wbpm6z)).

For someone who has a time-series of 3D images saved as a single 4D image, it's common to want to apply the same `sct_maths` operation to each volume in the series (for example, applying a convolution kernel). Currently, the only supported workflow is something like:
1. Run `sct_image -split t my-image.nii.gz` to generate a separate file for each 3D volume.
2. Run the actual `sct_maths` transformation in a bash loop over each file.
3. Run `sct_image -concat t my-image_t*.nii.gz` to recombine the 3D volumes into a single 4D file.

This is pretty annoying, and we should figure out a better way.

Some questions:
- What would be a good set of command-line flags/parameters for this feature?
- Would this conflict with doing actual 4D operations on 4D images?
- Would it possibly cause out-of-memory problems with large images?
  (If we're careful not to make extra copies of the data in memory it should be fine; after all, we're able to load the entire image in the `sct_image -split` step.)

Possibly related: #4024.

joshuacwnewton:
## Description

This PR adds a `-volumewise` argument to automate the steps described in issue #4181:

> 1. Run `sct_image -split t my-image.nii.gz` to generate a separate file for each 3D volume.
> 2. Run the actual `sct_maths` transformation in a bash loop over each file.
> 3. Run `sct_image -concat t my-image_t*.nii.gz` to recombine the 3D volumes into a single 4D file.

Instead, in this PR, we split the input data array, run the requested operations on each 3D volume, then recombine the data array.

## Linked issues

Fixes #4181.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -423,15 +423,11 @@ def main(argv: Sequence[str]):
     # If volumewise is specified, treat 4D image as a set of 3D volumes
     if arguments.volumewise:
         data_list = [np.squeeze(arr, axis=3) for arr in np.array_split(im.data, im.data.shape[3], 3)]
-        dim = list(im.dim)
-        dim[3] = 1  # Replace 4D dim with 1 (indicating single 3D volume)
-        dim_list = [tuple(dim)] * len(data_list)
     else:
         data_list = [im.data]
-        dim_list = [im.dim]
 
     data_out_list = []
-    for data, dim in zip(data_list.copy(), dim_list.copy()):
+    for data in data_list:
         for arg_name, arg_value in arguments.todo:
             assert data is not None
 
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -101,6 +101,17 @@ def get_parser():
 
     optional = parser.add_argument_group("OPTIONAL ARGUMENTS")
     optional.add_argument(
+        '-volumewise',
+        type=int,
+        help='Specifying this option will process a 4D image in a "volumewise" manner:\n'
+             '  - Split the 4D input into individual 3D volumes\n'
+             '  - Apply the maths operations to each 3D volume\n'
+             '  - Merge the processed 3D volumes back into a single 4D output image',
+        required=False,
+        choices=(0, 1),
+        default=0
+    )
+    optional.add_argument(
         "-h",
         "--help",
         action="help",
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -45,7 +45,8 @@ def is_float(element):
         for val in values:
             # Case 1: Argument is a float
             if is_float(val):
-                data_out.append(np.full_like(data_in, float(val)))
+                shape = data_in.shape[:-1] if namespace.volumewise else data_in.shape
+                data_out.append(np.ones(shape) * float(val))
```
joshuacwnewton, line 49: Since the `DataOrScalarArgument` custom action parses `-i` to get the shape of the input, this gets a bit awkward with `-volumewise`...

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -426,9 +438,24 @@ def apply_array_operation(data, dim, arg_name, arg_value, parser):
...
+            return i, i+len(small)
+    return False
+
+
 # MAIN
 # ==========================================================================================
 def main(argv: Sequence[str]):
+    # ensure '-volumewise 1' is parsed before any of the 'ParseDataOrScalarArgument' args
+    if idx := contains(['-volumewise', '1'], argv):
+        argv = ['-volumewise', '1'] + list(argv)[:idx[0]] + list(argv)[idx[1]:]
```
joshuacwnewton, line 457: ...because `-volumewise` must be parsed before any of the array operation arguments.

joshuacwnewton: It looks like this constraint may be fixed by the proposal from @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4485#issuecomment-2195052430, so my hacky fix here will no longer be needed. :tada:

joshuacwnewton:
It looks like there are some significant conflicts now that https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4485 has been refactored a bit. I'll try to clean these up and get this PR ready for review. :)

spinalcordtoolbox/scripts/sct_maths.py
joshuacwnewton: This is no longer an issue, since we no longer use `DataOrScalarArgument`.

spinalcordtoolbox/scripts/sct_maths.py
joshuacwnewton: This is no longer an issue, since we no longer need to parse `-volumewise` first (thanks to `ActionTodo`).

joshuacwnewton:
It looks like @mguaypaq's changes in #4485 has made this much nicer; we no longer need any of the weird hacks (parsing -volumewise first for the sake of `DataOrScalarArgument`), and we can just focus on the core logic (splitting the 4D array, iterating through the 3D arrays, then recombining).

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -408,152 +419,169 @@ def main(argv: Sequence[str]):
...
+        dim[3] = 1  # Replace 4D dim with 1 (indicating single 3D volume)
+        dim_list = [tuple(dim)] * len(data_list)
+    else:
+        data_list = [im.data]
+        dim_list = [im.dim]
+
+    data_out_list = []
+    for data, dim in zip(data_list.copy(), dim_list.copy()):
+        for arg_name, arg_value in arguments.todo:
+            assert data is not None
```
joshuacwnewton, line 436: Note that we're basically just indenting the entire if/else; GitHub is just showing the diff strangely.

(Encapsulating the if/else into a function might have made the diff a bit neater? But, it's nbd.)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -408,152 +419,169 @@ def main(argv: Sequence[str]):
...
-
-        if arg_name == "add":
-            if data.ndim == 4 and not arg_value:
-                # special case to sum a 4D volume across the t axis
-                data = np.sum(data, axis=3)
-            else:
+
+    # If volumewise is specified, treat 4D image as a set of 3D volumes
+    if arguments.volumewise:
+        data_list = [np.squeeze(arr, axis=3) for arr in np.array_split(im.data, im.data.shape[3], 3)]
```
mguaypaq, line 425: I'm going to rename `data_list` to `data_in_list` if you don't mind, since:
* there's a corresponding `data_out_list` variable below, and
* it's very easy to confuse with the existing `list_data` variable.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -408,152 +419,169 @@ def main(argv: Sequence[str]):
...
+        dim = list(im.dim)
```
mguaypaq, line 426: This actually clobbers the `dim` value computed from `arguments.dim` above the loop. Thankfully, the current version of the operations loop uses `data.shape` and `data.ndim` directly, so there's no need to track the array shape separately.

mguaypaq (approved):
Looks good, and thanks for the extra test! I'm just going to add a couple of commits for the two little problems I point out below.

joshuacwnewton:
Feel free to make the fixes! And, thank you for catching them. :hearts:

PR #4508, opened 2024-06-03
Ensure that using `-thr 0` with the contrast agnostic model generates a useful softseg

### LINKED ISSUE #4503: The threshold parameter of `sct_deepseg` doesn't work with the model `seg_sc_contrast_agnostic` for soft segmentation.
There is an issue with the threshold parameter of `sct_deepseg`.
Setting it to 0 is supposed to perform soft-segmentation but it doesn't seem to work with the model `seg_sc_contrast_agnostic`.

Here is an example with a canproco volume : 

`sct_deepseg -task seg_sc_contrast_agnostic -thr 0 -i sub-mon003_ses-M0_PSIR.nii.gz`

What I get is a binary segmentation : 

Capture d’écran du 2024-05-31 10-45-48

instead of a soft segmentation.

### LINKED ISSUE #4507: Output softseg from MONAI inference is not in the range `[0, 1]`
When testing https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4503, I noticed that the output colormap used for the softseg was `grayscale`, indicating that `check_image_kind` did not see it as a softseg.

The cause was values outside of the range `[0, 1]`, which are present from the postprocessed MONAI input.

image

However, interestingly, when I inspect the `batch['pred']` output from `sliding_window_inference`, I get the expected results:

image

This implies that something is going wrong in the MONAI postprocessing:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/20be5e53d05d82b09b8027e59809356bda694b3b/spinalcordtoolbox/deepseg/inference.py#L165-L168

joshuacwnewton:
## Description

There were a number of bugs preventing `-thr 0` from outputting a softseg:

- Conditions with `is not None`
  - `-thr 0` would not be passed to the inference function.
  - `-thr 0` would not be applied even if it were passed.
- Order 2 spline interpolation
  - When inverting the preprocessing transforms, the interpolation would cause the softseg to have values outside of the range `[0, 1]`.

I've added fixes for each of these bugs, and now a softseg is outputted when `-thr 0` is used:

```
Done! To view results, type:
fsleyes t2.nii.gz -cm greyscale t2_seg.nii.gz -cm red-yellow -a 70.0 &
```

However, as observed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4503#issuecomment-2144115486, `-thr 0` doesn't actually produce great results:

<details>

image

</details>

So, we apply a threshold of 0.1 (_without_ binarization) to the output softseg for MONAI models, regardless of what the value of `-thr` is, so that `-thr 0` will produce a nice looking softseg.

<details>

image

</details>

Side-note: I realize that the Red-Yellow colormap doesn't look as I expected -- I was thinking it would be Yellow-Red instead? It is possible to invert the colormap in FSLeyes settings, but I don't know if this can be triggered via the command line. However, you _can_ specify matplotlib colormaps, and there is an appropriate colormap we can use here (`YlOrRd`):

<details>

image

</details>

## Linked issues

Fixes #4503.
Fixes #4507.


joshuacwnewton:
TODO: Figure out how to apply `bilinear` to the inverse transform while still preserving the original `mode=2` (order 2 spline) initial interpolation.

joshuacwnewton:
I have been unable to figure out a way to use `mode=2` for the forward interpolation method, but `bilinear` for the inverse interpolation method. I've summarized my investigation in this feature request:

- https://github.com/Project-MONAI/MONAI/issues/7836

naga-karthik:
Whoa, super cool investigations, @joshuacwnewton ! I didn't think that it would be as straightforward as changing their (hard-coded) interpolation method to the value that we want! 

> however this creates an unnatural ring of "1.0" values around the perimeter.

I agree with your concern if we go with clipping to `[0, 1]` for the time being. Also, the change you suggested seems like it could break a lot of their other tutorials/functions, etc. that rely on the `nearest_interp` param, not to mention that most likely every user who has used `Invertd` transform will be affected. I'm curious to see how the monai devs would respond!

joshuacwnewton:
I've added tests to ensure the output of `-thr 0` is actually a softseg, and I also tested the postprocessing options from #4482 with softsegs to ensure they work. This should be ready for review now. :)

> Also, the change you suggested seems like it could break a lot of their other tutorials/functions, etc. that rely on the nearest_interp param, not to mention that most likely every user who has used Invertd transform will be affected.

Thanks for the feedback! I very much agree, so I edited my suggestion to be a little softer (including a deprecation period to allow people to transition their API, or perhaps making `nearest_interp` a synonym for `interp='nearest'`).

mguaypaq (approved):
The code looks fine to me, and thanks for adding some tests!

PR #4511, opened 2024-06-04
Increase timeout of `mutex` to avoid `AlreadyLocked` errors with many concurrent processes

### LINKED ISSUE #4501: `portalocker` error when running `sct_qc` with a large number of jobs in parallel
I ran into a `portalocker.exceptions.AlreadyLocked` error when I running `sct_qc` in a [`sct_run_batch` script](https://github.com/jqmcginnis/miccai_amai_lesions_spine/blob/main/dataset-conversion/generate_qc_datasets_gt.sh) with a large number of jobs in parallel (`-jobs 32`). @mguaypaq suggested to use lesser number of jobs (ideally 1 to never run into this) but I was able to run the script successfully with `-jobs 8`. The full error trace is shown below.

<details>
<summary> Error trace </summary>

```
*** Generate Quality Control (QC) html report ***
Resampling image "sub-m018732_ses-20180107_acq-ax_T2w_desc-straightened.nii.gz" to 0.6x0.6 mm
Resampling image "sub-m018732_ses-20180107_acq-ax_T2w_seg-manual_desc-straightened.nii.gz" to 0.6x0.6 mm
[33mConverting image from type 'uint8' to type 'float64' for linear interpolation[0m
[33mImage header specifies datatype 'float64', but array is of type 'int64'. Header metadata will be overwritten to use 'int64'.[0m
QcImage: layout with Axial slice
Compute center of mass at each slice
/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/scipy/ndimage/_measurements.py:1542: RuntimeWarning: invalid value encountered in scalar divide
  results = [sum(input * grids[dir].astype(float), labels, index) / normalizer
/home/GRAMES.POLYMTL.CA/u114716/tum-poly/qc_straight_stitched/qc/sub-m018732/sub-m018732/ses-20180107/anat/sct_deepseg_sc/2024_05_30_130243.408480/background_img.png
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_qc.py", line 121, in <module>
    main(sys.argv[1:])
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_qc.py", line 104, in main
    generate_qc(fname_in1=arguments.i,
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 771, in generate_qc
    QcImage(
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 296, in layout
    self._make_QC_image_for_3d_volumes(img, mask, plane=self.qc_report.plane)
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 341, in _make_QC_image_for_3d_volumes
    self.qc_report.update_description_file()
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 588, in update_description_file
    with mutex(f"sct_qc-{os.path.basename(path_qc)}-{md5(self.path_qc.encode('utf-8')).hexdigest()}"):
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/contextlib.py", line 119, in __enter__
    return next(self.gen)
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/spinalcordtoolbox/utils/fs.py", line 281, in mutex
    semaphore.acquire()
  File "/home/GRAMES.POLYMTL.CA/u114716/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/portalocker/utils.py", line 493, in acquire
    raise exceptions.AlreadyLocked()
portalocker.exceptions.AlreadyLocked

```

</details>

Command used:

```
sct_run_batch -path-data ~/tum-poly/straightened_sc_stitch_linear_bin_full/data_processed_clean -path-output ~/tum-poly/qc_straight_stitched_2 -jobs 32 -script ~/tum-poly/miccai_amai_lesions_spine/dataset-conversion/generate_qc_datasets_gt.sh -script-args "stitched"
```

joshuacwnewton:
## Description

When many processes are running (e.g. `sct_run_batch -jobs 32`), we may have many different subjects trying to write to the QC report at one time. To avoid corruption/file overwriting, we limit the QC report to 1 process at a time, while other processes have to wait their turn. By default, the timeout is 5s, after which an `AlreadyLocked` error will occur.

To fix this, I first amended our parallel QC test to try to better catch this issue, since this error only occurs when there are many jobs running, but the test only checked against `jobs=2`. (Anecdotally, 2-8 jobs will be able to acquire the lock on first try, but 16+ jobs will start to back up, causing waits and re-attempts to acquire the lock.) Also, I modified portalocker's logger so that we can see `portalocker`'s internal behavior when run to its limits. To see the new `DEBUG` output locally, run:

```
sct_testing -o log_cli=true testing/api/test_qc_parallel.py::test_many_qc
```

Then, I increased the timeout to allow more time for older processes to eventually acquire the lock.

(Note: I also sped up the check interval a little (`0.25` -> `0.1`) in the hopes of increasing the probability that "waiting" locks will acquire the lock ahead of new processes. That way, we lower the possibility of new processes cutting in line and delaying a single old process that just can't seem to acquire the lock.)

I think ideally, we would try to optimize the code that occurs under the lock (so that processes release the lock quicker). But, I think increasing the timeout is a much easier short-term solution. 

## Linked issues

Fixes #4501.

### DIFF
--- a/testing/api/test_qc_parallel.py
+++ b/testing/api/test_qc_parallel.py
@@ -20,13 +20,13 @@ def test_many_qc():
     if multiprocessing.cpu_count() < 2:
         pytest.skip("Can't test parallel behaviour")
 
-    p = multiprocessing.Pool(2)
+    p = multiprocessing.Pool(multiprocessing.cpu_count())
 
     with TemporaryDirectory(prefix="sct-qc-") as tmpdir:
         # This `try, finally` pattern mitigates hanging with pytest-cov
         # See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3661#issuecomment-1029057900
         try:
-            p.map(gen_qc, [tmpdir] * 5)
+            p.map(gen_qc, [tmpdir] * (multiprocessing.cpu_count() * 2))
         finally:
             p.close()  # Marks the pool as closed.
             p.join()   # Waits for workers to exit.
--- a/testing/api/test_qc_parallel.py
+++ b/testing/api/test_qc_parallel.py
@@ -2,6 +2,8 @@
 
 from tempfile import TemporaryDirectory
 import pytest
+import logging
+import sys
 
 import multiprocessing
 
@@ -17,6 +19,13 @@ def gen_qc(path_qc):
 
 def test_many_qc():
     """Test many qc images can be made in parallel"""
+    # Turn on debug logging for fs.py to check mutex acquire/release times
+    # To check this output locally, run: `sct_testing -o log_cli=true testing/api/test_qc_parallel.py::test_many_qc`
+    screen_handler = logging.StreamHandler(stream=sys.stderr)
+    logger = logging.getLogger("portalocker.utils")
+    logger.setLevel(logging.DEBUG)
+    logger.addHandler(screen_handler)
+
     if multiprocessing.cpu_count() < 2:
         pytest.skip("Can't test parallel behaviour")
 
--- a/spinalcordtoolbox/utils/fs.py
+++ b/spinalcordtoolbox/utils/fs.py
@@ -277,7 +277,7 @@ def mutex(name):
     We use a mutex over a lock because the mutex doesn't depend on the destination
     of the locked files, which allows us to avoid locking on e.g. NFS-mounted drives.
     """
-    semaphore = portalocker.BoundedSemaphore(maximum=1, name=name)
+    semaphore = portalocker.BoundedSemaphore(maximum=1, name=name, timeout=60, check_interval=0.1)
     semaphore.acquire()
     try:
         yield semaphore


testing/api/test_qc_parallel.py
```diff
@@ -17,16 +19,23 @@ def gen_qc(path_qc):
 
 def test_many_qc():
     """Test many qc images can be made in parallel"""
+    # Turn on debug logging for fs.py to check mutex acquire/release times
+    # To check this output locally, run: `sct_testing -o log_cli=true testing/api/test_qc_parallel.py::test_many_qc`
+    screen_handler = logging.StreamHandler(stream=sys.stderr)
```
joshuacwnewton, line 24: Note: Setting `stream=sys.stderr` has the helpful effect of logging all `portalocker` DEBUG messages together at the end of the test output.

However... huh. I'm realizing that this behavior only occurs inside of PyCharm, and not more generally in terminals. So, this might be unnecessary?

image

joshuacwnewton: I'm also realizing that this log output doesn't do a great job at identifying which process the log messages belong to, nor does it give any sort of timestamp... :thinking:

mguaypaq: At least it confirms that the test processes are actually trying to lock the file at the same time?

PR #4517, opened 2024-06-06
Update CentOS Stream 8 runner to work past EOL

### LINKED ISSUE #3675: CentOS 8 runner - `Error: Failed to download metadata for repo 'appstream'`
This issue is hitting multiple of our PRs (e.g. [this one](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/5011777965?check_suite_focus=true)). Will dig deeper soon, but wanted to quickly open an issue for this so that we have a place to discuss. It's probably an upstream thing, rather than an issue on our end!

The relevant line in the workflow file is: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fc34ee93cf0073a2abaca9f786aa0f794101e0c4/.github/workflows/tests.yml#L165-L167

joshuacwnewton:
## Description

CentOS Stream 8 hit EOL on May 31st, 2024 after which its packages were moved to vault.centos.org. To continue installing the necessary dependencies, then, it is necessary to point yum's mirrorlist to this new location.

I've tested the changes here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9405745110

We could instead consider retiring CentOS Stream 8 entirely? But, this seems like a relatively easy way to keep testing things (at least until something else changes in the future).

## Linked issues

Fixes #3675.


mguaypaq (approved):
Looks good, thanks! And let's bring up the issue of whether we still want to test CentOS Stream 8 at the dev meeting.

PR #4519, opened 2024-06-06
Indicate model version in "model is up to date" log message

### LINKED ISSUE #4518: Indicate deepseg model
Thanks to a recent PR, the model version is now checked, and the user sees this at inference:

~~~
Model 'model_seg_sc_contrast_agnostic_softseg_monai' is up to date
~~~

This is great, but what would be even greater, would be to display the model version, eg:
~~~
Model 'model_seg_sc_contrast_agnostic_softseg_monai' is up to date (vX.Y)
~~~

joshuacwnewton:
## Description

"Model version" is a bit tricky because we use URLs as versions (because multiple assets can be attached to a single GitHub release, and also because of OSF mirrors).

Using the URLs looks like this:

```
Model 'model_seg_sc_contrast_agnostic_softseg_monai' is up to date (Source: https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v2.4/model_soft_bin_20240425-170840.zip
```

It's a bit verbose, but it is very explicit.

## Linked issues

Fixes #4518.

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -467,7 +467,7 @@ def is_up_to_date(path_model):
     else:
         logger.warning("Mismatch between 'source.json' URL format and SCT source code URLs")
         return False
-    logger.info(f"Model '{model_name}' is up to date")
+    logger.info(f"Model '{model_name}' is up to date (Source: {actual_model_urls})")
     return True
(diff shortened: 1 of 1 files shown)


jcohenadad (approved):
Excellent fix, the URL is actually more informative than just the version

PR #4529, opened 2024-06-13
Simplify calls to `np.einsum`

mguaypaq:
This is a tiny refactor to simplify two lines of code in `types.py`, since I recently had to dig through this file for debugging.

The reason is that:
- `np.transpose` is generally a little confusing to the reader, so it's best to set up proper conventions up-front and avoid transposes if possible. (It's not always possible, though.)
- `np.einsum` is a powerful tool, but it's a bit hard to read. Let's not make it harder than it needs to be.
- `np.rollaxis` is massively confusing, so it's really best to avoid it.

And in the case of `np.einsum`, we have complete control over the ordering of axes for the inputs and the outputs, so we really can avoid transposes and rollaxes. (This is the equivalent of getting rid of double negatives: it makes things easier to read!)

### DIFF
--- a/spinalcordtoolbox/types.py
+++ b/spinalcordtoolbox/types.py
@@ -327,10 +327,10 @@ class Centerline:
         return coordinates - unit_derivatives * dot_products
 
     def get_in_plans_coordinates(self, coordinates, indexes):
-        return einsum('mnr,nr->rm', rollaxis(self.inverse_matrices[indexes], 0, 3), (coordinates - self.points[indexes]).transpose())
+        return einsum('mnr,rn->rm', rollaxis(self.inverse_matrices[indexes], 0, 3), coordinates - self.points[indexes])
 
     def get_inverse_plans_coordinates(self, coordinates, indexes):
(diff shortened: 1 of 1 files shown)


joshuacwnewton (approved):
This very much demystified `einsum` for me, so thank you.

I see now that we're basically doing what we did in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4514#discussion_r1629897664 (having to multiply row vectors by an affine matrix). But, instead of having to do `.T` before and after like I suggested, `einsum` takes care of all of the axis manipulation internally so that it just... works? And all you need to do is think about axis labels.

I'm still not entirely sure what these methods are meant to do in the grand scheme of the straightening algorithm, but that's probably out of scope for this PR (since the straightening code could use a _lot_ of clarifying comment IMO). I fully understand the functional aspects of the methods, so that alleviates my concerns from the previous PR.

I just had a (very nitpicky) little suggestion to help make the functionality of `einsum` even more clear, since your commit message for https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4529/commits/a05a2728a764a60b306b1a03114b345d501c98f7 is what made things click for me, and I wanted to capture the spirit of it in comments.

PR #4533, opened 2024-06-17
Add rotation to landmark-based preregistration (`step=0`) during template registration

### LINKED ISSUE #4527: Label-based transformation during template registration is wrong
### Description

Issue originally described here: https://github.com/sct-pipeline/spine-park/issues/33, but followed up here since it seems to be an SCT-specific issue.

Quick tl;dr:

- This issue is pre-existing, and was thankfully not caused by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4514.
- The first landmark-based rigid registration did not work, leading to a strong cropping of the straightened image:

image

jcohenadad:
#### GitHub

#### PR contents

## Description

Added a label, 5mm to the left, of the smallest label, both in the src_straight and dest (PAM50), to accommodated for tilted qform. The degrees of freedom for the step=0 registration was also changed to accommodate more rotations.

- [x] Address case of [only one label present](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65ffa7524dc145a23086127be6d9f03ccc6b0778/spinalcordtoolbox/scripts/sct_register_to_template.py#L441) 👉 Hopefully covered by [testing](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a3182e5171aeef84e1aa6aebc4a6e9caeaeba81f/testing/cli/test_cli_sct_register_to_template.py#L31-L61)
- [x] [Address both `ref=template` and `ref=subject` cases](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65ffa7524dc145a23086127be6d9f03ccc6b0778/spinalcordtoolbox/scripts/sct_register_to_template.py#L475)
- [ ] Address [`level_alignment` case if number of labels >2](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65ffa7524dc145a23086127be6d9f03ccc6b0778/spinalcordtoolbox/scripts/sct_register_to_template.py#L489)

## Linked issues
Fixes #4527


jcohenadad:
Using current implementation, a tilted qform cannot properly register the labels to the PAM50 template because the transformations are `TxTyTzSz` and there are only two label pairs:

image

jcohenadad:
When adding one label 5mm to the left and setting `dof=`:

image

> [!WARNING]
> On the straightened source image, the added label (yellow) is close to the existing label blob (red), which could (or could not) be an issue if they overlap:
image

jcohenadad:
For fun, I tried all 9 dofs (`Tx_Ty_Tz_Rx_Ry_Rz_Sx_Sy_Sz`) and the registration is now perfect:

image

I _think_ this is because of the order of transformations applied. Ie: in theory, there should be only Sz _after_ the labels are aligned with that of the template, but this is not how it is optimized during the estimation of the label-based transform, hence the need to specify 9 dofs. 

As long as this registration is 100% robust, I guess this is fine to specify 12 dofs.

jcohenadad:
Interestingly, if `reference=subject`, step=0 is overwritten with these dof:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/65ffa7524dc145a23086127be6d9f03ccc6b0778/spinalcordtoolbox/scripts/sct_register_to_template.py#L334-L346

Which seems to be also enough for 3-label registration with `reference=template`:

getRigidTransformFromLandmarks_plot_Tx_Ty_Tz_Rx_Ry_Rz_Sz

So I'll probably implement this set of DOFs for both scenarios

jcohenadad:
The proposed fix implemented in 634433f9572e451eaae9a7d8bc94d9d52ff17461 seems to have solved the issue.

Before the fix (the cord is cropped at the edge and the slicewise registration is imperfect due to the rotated straight cord in the PAM50 space):
ezgif-4-71f1d81e7b

After the fix:
ezgif-4-434265ec90

jcohenadad:
While adding a test for >2 labels in 42aea7e33575cd98b4e7fd2308ff3497f4bb19ca, I noticed strange deformation of the spinal cord. Below is a toggle between 
- the registered anat (see the weird deformation of the two vertebrae in the middle of the FOV, the abnormal stretch)
- the template (the associated disc labels are in red)
- the registered disc labels (in blue) for the two central discs (the two other labels at the top/bottom are appropriately registered)

ezgif-6-6024b83af2


things to watch out for: 
- projection of the disc labels onto the centerline

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -332,18 +332,13 @@ def main(argv: Sequence[str]):
...
-        if ref == 'subject':
-            paramregmulti.steps['0'].dof = 'Tx_Ty_Tz_Rx_Ry_Rz_Sz'
         # add user parameters
         for paramStep in arguments.param:
             paramregmulti.addStep(paramStep)
     else:
         paramregmulti = ParamregMultiStep([step0, step1, step2])
-        # if ref=subject, initialize registration using different affine parameters
-        if ref == 'subject':
-            paramregmulti.steps['0'].dof = 'Tx_Ty_Tz_Rx_Ry_Rz_Sz'
```
joshuacwnewton, line 346: Thanks for this de-duplication! This code will be more straightforward to understand now. :hearts:

testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -28,6 +29,24 @@ def template_lpi(tmp_path_factory):
...
+    im_labels = Image(sct_test_path('t2', 'labels.nii.gz'))
+    im_labels.data = np.zeros(im_labels.data.shape)
+    # im_labels.data[23,53,26] = 3
+    # im_labels.data[24,24,26] = 4
+    # im_labels.data[25,17,26] = 5
+    # im_labels.data[26,1,26] = 6
+    im_labels.data[30,53,26] = 3
+    im_labels.data[31,34,26] = 4
+    im_labels.data[31,17,26] = 5
+    im_labels.data[32,1,26] = 6
```
joshuacwnewton, line 45: Since the old label coordinates were wrong, is there any reason to keep them? :stuck_out_tongue_closed_eyes: 

```suggestion
    im_labels.data[30,53,26] = 3
    im_labels.data[31,34,26] = 4
    im_labels.data[31,17,26] = 5
    im_labels.data[32,1,26] = 6
```

testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -125,3 +144,29 @@ def test_sct_register_to_template_mismatched_xforms(tmp_path, capsys):
                                             '-s', sct_test_path('t2', 't2_seg-manual.nii.gz'),
                                             '-l', sct_test_path('t2', 'labels.nii.gz')])
     assert "Image sform does not match qform" in capsys.readouterr().out
+
+
+def test_sct_register_to_template_3_labels(tmp_path, labels_discs):
+    """Test registration with 3 labels."""
```
joshuacwnewton, line 150: I'm a bit confused by the name/description of this test. It says "3 labels", but `labels_discs` contains 4 disc labels? 

From the context in this PR, it seems like this is meant to test the ">2 labels" case, so perhaps we could update it like this:

```suggestion
def test_sct_register_to_template_more_than_2_labels(tmp_path, labels_discs):
    """
    Test registration with >2 labels. This test (and the custom disc label file) are needed because the existing
    `t2/labels.nii.gz` file only contains 2 labels. But, registration will be performed differently depending on
    whether there are 1, 2, or >2 labels. See also:

    https://spinalcordtoolbox.com/user_section/tutorials/vertebral-labeling/how-many-labels-for-registration.html
    """
```

testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -125,3 +144,29 @@ def test_sct_register_to_template_mismatched_xforms(tmp_path, capsys):
...
+    # Compute pairwise distance between the template label and the registered label, ie: compute distance between dest and src_reg
+    # for label of value '2', then '3', etc. and compute the mean square of all distances. The labels should be touching, hence
+    # the mean square should be 0.
+    im_label_dest = Image(sct_test_path('template', 'template', 'PAM50_small_label_disc.nii.gz'))
+    im_label_src_reg = Image(str(tmp_path/'labels_discs_reg.nii.gz'))
+    assert compute_mean_squared_error(im_label_dest, im_label_src_reg) == 0
+
+
+
+
```
joshuacwnewton, line 172: There are some linting/style issues, but I can take care of those for you. :)

spinalcordtoolbox/scripts/sct_register_to_template.py
joshuacwnewton, line 1: While reading the file, I noticed this comment from @benjamindeleener:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/634433f9572e451eaae9a7d8bc94d9d52ff17461/spinalcordtoolbox/scripts/sct_register_to_template.py#L669-L679

This seems to be relevant to this PR. Does it need to be updated/removed?

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -612,8 +607,17 @@ def main(argv: Sequence[str]):
...
-            # Compute rigid transformation straight landmarks --> template landmarks
+            
+            # Add labels in the X-Z and Y-Z planes to accommodate for rotations about the Y and X axes
+            printv('\nAdd labels in the X-Z and Y-Z planes to accommodate for rotations about the Y and X axes...', verbose)
+            # -> Pick a dummy label between [1, 127] that doesn't clash with the existing label values.
+            existing_label_vals = {coord.value for coord in img_tmp_label.getNonZeroCoordinates()}
+            positive_int8_vals = set(range(1, 128))
+            dummy_label = max(positive_int8_vals - existing_label_vals)  # Should be 127 in 99.99% of cases
+            add_orthogonal_label(ftmp_label, new_label_value=dummy_label)
+            add_orthogonal_label(ftmp_template_label, new_label_value=dummy_label)
```
joshuacwnewton, line 618: This newly added code is duplicated from the `ref == 'subject'` case, right?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/634433f9572e451eaae9a7d8bc94d9d52ff17461/spinalcordtoolbox/scripts/sct_register_to_template.py#L770-L776

That makes sense, because now both cases compute an affine transformation (incl. rotation) for the pre-registration step. 

I'm wondering if maybe it's worth creating a function for this logic, then calling the function in both places (to avoid duplication)?

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -332,18 +332,13 @@ def main(argv: Sequence[str]):
         smooth=arguments.centerline_smooth)
     # registration parameters
     if arguments.param is not None:
-        # reset parameters but keep step=0 (might be overwritten if user specified step=0)
+        # reset parameters but keep step=0 (can be overwritten if user specified step=0)
```
joshuacwnewton, line 335: Re: This comment... In what cases would the user want to overwrite `step=0`? Do we want to test this case?

For example, what if the user specifies a different `dof`? Would that cause a crash? (Or, is this such a rare thing for the user to do that the user should basically expect crashes/unstable behavior from modifying step 0?)

joshuacwnewton (changes requested):
It seems like the "tl;dr" of this PR (after all the revisions) is that we are unifying the `ref='subject'` and `ref='template'` cases to both compute affine transformations (incl. rotations) for the pre-registration Step 0. (Before, `ref='template'` computed only translations/z-scaling due to a faulty assumption that we did not need to care about rotations.)

Most of the changes here LGTM! Most of my comments here are nitpicky  (organizational changes, style changes, naming changes) but the overall logic makes sense to me.

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -56,7 +56,7 @@ def __init__(self):
 
 # get default parameters
 # Note: step0 is used as pre-registration
-step0 = Paramreg(step='0', type='label', dof='Tx_Ty_Tz_Sz')  # if ref=template, we only need translations and z-scaling because the cord is already straight
+step0 = Paramreg(step='0', type='label', dof='Tx_Ty_Tz_Rx_Ry_Rz_Sz')  # if ref=template, we only need translations and z-scaling because the cord is already straight
```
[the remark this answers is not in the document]
jcohenadad: yes! good catch, do you mind fixing it since you're at it? many thanks!

testing/cli/test_cli_sct_register_to_template.py
jcohenadad: > Since the old label coordinates were wrong, is there any reason to keep them? 😝

no reason, sorry for not cleaning up 😅 
do you mind removing it since you,re at it? thanks!

testing/cli/test_cli_sct_register_to_template.py
jcohenadad: > I'm a bit confused by the name/description of this test. It says "3 labels", but labels_discs contains 4 disc labels?

sorry about that-- i started doing it with 3 labels, and then added one. Do you mind fixing the name of the function and the docstrings? thanks!!

spinalcordtoolbox/scripts/sct_register_to_template.py
jcohenadad: yeah, i also noticed this could be relevant to the current issue... but unfortunately no time to dig deeper. I would leave it as is for now

spinalcordtoolbox/scripts/sct_register_to_template.py
jcohenadad: > I'm wondering if maybe it's worth creating a function for this logic, then calling the function in both places (to avoid duplication)?

yup, i thought about it, but unfortunately no time to do it-- could be for subsequent PR maybe?

spinalcordtoolbox/scripts/sct_register_to_template.py
jcohenadad: > Re: This comment... In what cases would the user want to overwrite step=0? Do we want to test this case?

ideally we should. For subsequent PR though

> For example, what if the user specifies a different dof? Would that cause a crash? (Or, is this such a rare thing for the user to do that the user should basically expect crashes/unstable behavior from modifying step 0?)

hum... hopefully not 😬

joshuacwnewton (approved):

spinalcordtoolbox/scripts/sct_register_to_template.py
joshuacwnewton, line 1: Presumably these `batch_processing.sh` changes are to be expected?

```
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[mt/mtr_in_wm.csv-0-MAP()] - assert 54.43385545448296 == 54.409838390430586 ± 5.4e-05
  
  comparison failed
  Obtained: 54.43385545448296
  Expected: 54.409838390430586 ± 5.4e-05
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[dmri/fa_in_cst.csv-0-WA()] - assert 0.7819845030091397 == 0.778704497861937 ± 7.8e-07
  
  comparison failed
  Obtained: 0.7819845030091397
  Expected: 0.778704497861937 ± 7.8e-07
FAILED testing/batch_processing/test_batch_processing.py::test_batch_processing_results[dmri/fa_in_cst.csv-1-WA()] - assert 0.7745645889969645 == 0.7763774523351279 ± 7.8e-07
  
  comparison failed
  Obtained: 0.7745645889969645
  Expected: 0.7763774523351279 ± 7.8e-07
========================= 3 failed, 5 passed in 0.07s ==========================
```

joshuacwnewton: I wonder if we could create a more useful `batch_processing.sh` test...

Like, these value changes are from `sct_extract_metric` which uses a mask, right? So, I wonder if these changes are caused by, say, a single voxel change in the mask? If so, I feel like "number of changed voxels" might be a more useful way to communicate the changes from a PR. :thinking:

jcohenadad: yup-- could you please update the batch processing? thank you

PR #4554, opened 2024-07-03
Replace `seg_ms_lesion_mp2rage` model with a nnUnet based model on `sct_deepseg`

### LINKED ISSUE #4522: Replace deepseg MS lesion seg model
after https://github.com/ivadomed/model_seg_ms_mp2rage/pull/82 is merged, model URL should be updated

Nilser3:
#### GitHub

#### PR contents

## Description

PR for replace `seg_ms_lesion_mp2rage` `deepseg` model for a single-class nnUnet model
[Model repo](https://github.com/ivadomed/model_seg_ms_mp2rage/tree/nlm/nnunet_ms_lesion) was updated, explaining their use and Slicer implementation.

## Linked issues

Fixes https://github.com/ivadomed/model_seg_ms_mp2rage/issues/75
Fixes #4522 

Related to https://github.com/ivadomed/model_seg_ms_mp2rage/issues/81
Related to https://github.com/ivadomed/model_seg_ms_mp2rage/pull/82


spinalcordtoolbox/deepseg/models.py
```diff
@@ -360,6 +340,21 @@
...
+                        'segmentations. Training consisted of MP2RAGE data on UNIT1 contrast at 1.0 mm3 isotropic (322 subjects from '
+                        '3 sites: National Institutes of Health, Bethesda, USA, University Hospital Basel and University of Basel, Basel, '
+                        'Switzerland and Center for Magnetic Resonance in Biology and Medicine, CRMBM-CEMEREM, UMR 7339, CNRS,  '
+                        'Aix-Marseille University, Marseille, France). '
+                        'To crop the data you can first segment the spinal cord using the contrast agnostic model, This could be '
+                        'done using: "sct_deepseg -i IMAGE_UNIT1 -task seg_sc_contrast_agnostic -o IMAGE_UNIT1_sc", then crop the '
+                        'IMAGE_UNIT1 image with 30 mm of dilation on axial orientation around the spinal cord. This could be done using: '
+                        '"sct_crop_image -i IMAGE_UNIT1 -m IMAGE_seg -dilate 30x30x5" . Note that 30 is only for 1mm isotropic '
+                        'resolution, for images with another resolution divide 30/your_axial_resolution.',
```
joshuacwnewton, line 355: Minor nitpick -- Indentation alignment:

```suggestion
         'long_description': 'This segmentation model for multiple sclerosis lesion segmentation on cropped MP2RAGE-UNIT1 spinal cord data. '
                             'Uses a 3D U-Net, trained with the nnUNetV2 framework. It is a single-class model outputting binary MS lesions '
                             'segmentations. Training consisted of MP2RAGE data on UNIT1 contrast at 1.0 mm3 isotropic (322 subjects from '
                             '3 sites: National Institutes of Health, Bethesda, USA, University Hospital Basel and University of Basel, Basel, '
                             'Switzerland and Center for Magnetic Resonance in Biology and Medicine, CRMBM-CEMEREM, UMR 7339, CNRS,  '
                             'Aix-Marseille University, Marseille, France). '
                             'To crop the data you can first segment the spinal cord using the contrast agnostic model, This could be '
                             'done using: "sct_deepseg -i IMAGE_UNIT1 -task seg_sc_contrast_agnostic -o IMAGE_UNIT1_sc", then crop the '
                             'IMAGE_UNIT1 image with 30 mm of dilation on axial orientation around the spinal cord. This could be done using: '
                             '"sct_crop_image -i IMAGE_UNIT1 -m IMAGE_seg -dilate 30x30x5" . Note that 30 is only for 1mm isotropic '
                             'resolution, for images with another resolution divide 30/your_axial_resolution.',
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -360,6 +340,21 @@
                              'dataset has healthy control subjects. The model has been trained in a human-in-the-loop active learning fashion.',
          'url': 'https://github.com/sct-pipeline/fmri-segmentation',
          'models': ['model_seg_sc_epi_nnunet']},
+    'seg_ms_lesion_mp2rage':
+        {'description': 'Segmentation of spinal cord MS lesions on MP2RAGE UNIT1 contrast',
```
joshuacwnewton, line 344: The cropping requirement is only mentioned in `long_description`, which is only shown when users type `sct_deepseg -list-tasks`. 

It might be good to also mention "cropped" in the regular description, too. We could keep the old description, maybe?

```suggestion
    'seg_ms_lesion_mp2rage':
        {'description': 'MS lesion segmentation on cropped MP2RAGE data',
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -191,6 +174,15 @@
          "thr": None,  # Images are already binarized
          "default": False,
      },
+    "model_seg_ms_lesion_mp2rage": {
+         "url": [
+             "https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20240610/nnUNetTrainer_seg_ms_lesion_mp2rage__nnUNetPlans__3d_fullres.zip"
+         ],
+         "description": "Segmentation of spinal cord MS lesions on MP2RAGE UNIT1 contrast",
+         "contrasts": ["UNIT1"],
```
joshuacwnewton, line 182: I notice that we used to have the `'mp2rage'` contrast for the old ivadomed model? And, actually, we still do use `mp2rage` for the old spinal cord model, too. 

> Note: The exact value for `contrasts` doesn't have any functional purpose (since `-c` is only considered for two models) -- its only use is in the documentation (`-list-tasks`) to guide users.

To keep the models consistent, should we use `mp2rage` or should we use `UNIT1`? If we want to use `UNIT1`, then maybe we should also update the `contrasts` field of the other MP2RAGE model?

spinalcordtoolbox/deepseg/models.py
```diff
@@ -95,23 +95,6 @@
...
         "default": False,
     },
-    "model_seg_ms_lesion_mp2rage": {
-        "url": {
-            "seed1": ["https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20230210/model_seg_lesion_mp2rage_r20230210_dil32_seed01.zip"],
-            "seed2": ["https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20230210/model_seg_lesion_mp2rage_r20230210_dil32_seed02.zip"],
-            "seed3": ["https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20230210/model_seg_lesion_mp2rage_r20230210_dil32_seed03.zip"],
-            "seed4": ["https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20230210/model_seg_lesion_mp2rage_r20230210_dil32_seed04.zip"],
-            "seed5": ["https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20230210/model_seg_lesion_mp2rage_r20230210_dil32_seed05.zip"],
-        },
```
joshuacwnewton, line 105: Note: Since the old model was the only one to use "ensemble" inference, and because "ensemble" inference was only ever implemented for the `ivadomed` models, then that means that our ensemble code is now "dead".

CC: @jcohenadad, @mguaypaq How likely is it that we create model ensembles in the future, and is the ensemble logic worth maintaining?

Personally, I think because we could easily implement ensembling in the future for non-ivadomed models, I think it might be worth leaving the ensembling code as-is for now? (Since, most of the ensembling code is "general", and it would be a bit dangerous IMO to remove the ensembling logic only to have to re-add it later on, since it involves finicky tweaks in quite a few places.)

joshuacwnewton: For posterity, here is a quick list of places where ensembling is considered:

- "General" code
  - `models.py :: find_model_folder_paths`
  - `models.py :: is_up_to_date`
  - `models.py :: is_valid`
  - `models.py :: install_model`
- ivadomed-specific code:
  - `inference.py :: segment_and_average_volumes`

spinalcordtoolbox/deepseg/models.py
```diff
@@ -360,6 +340,21 @@
                              'dataset has healthy control subjects. The model has been trained in a human-in-the-loop active learning fashion.',
          'url': 'https://github.com/sct-pipeline/fmri-segmentation',
          'models': ['model_seg_sc_epi_nnunet']},
+    'seg_ms_lesion_mp2rage':
```
joshuacwnewton, line 343: I notice that we don't have a test for this task. Presumably, this is because we don't have MP2RAGE data in our testing dataset?

However, it might be worth adding a test anyway using dummy T2 data, just to see whether it even runs (and if any errors are thrown).

This can be easily done by adding a new case to the existing `test_segment_nifti_binary_seg` test:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ecea2beaadc67ec7c02f6b539592db3bc01a1495/testing/cli/test_sct_deepseg.py#L41-L58

The new test case would look something like:

```python
     (sct_test_path('t2', 't2.nii.gz'),  # dummy image since no MP2RAGE test data
      None,                              # no ground truth, just test if it runs
      't2_seg_deepseg.nii.gz', 
      'seg_ms_lesion_mp2rage', 
      None), 
```

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 105: I was curious how the "check if model is up to date" code would handle the transition from a multi-seed ensemble task to a single-seed task. But, it looks like it handles things just fine:

```console
(venv_sct) nlm/update-_seg_ms_lesion_mp2rage-model|joshua@monarch:~/repos/spinalcordtoolbox/data/sct_testing_data/t2$ sct_deepseg -task seg_ms_lesion_mp2rage -i t2.nii.gz

--
Spinal Cord Toolbox (git-nlm/update-_seg_ms_lesion_mp2rage-model-a82c72c5990f604d69ecee1667692d9aaeccf11e)

sct_deepseg -task seg_ms_lesion_mp2rage -i t2.nii.gz
--

Mismatch between 'source.json' URL format and SCT source code URLs
Model model_seg_ms_lesion_mp2rage is out of date. Re-installing it now...

INSTALLING MODEL: model_seg_ms_lesion_mp2rage
Removing existing destination folder '/home/joshua/repos/spinalcordtoolbox/data/deepseg_models/model_seg_ms_lesion_mp2rage'
Trying URL: https://github.com/ivadomed/model_seg_ms_mp2rage/releases/download/r20240610/nnUNetTrainer_seg_ms_lesion_mp2rage__nnUNetPlans__3d_fullres.zip
```

Great job to @mguaypaq for writing robust code that takes into account a potential mismatch in URL formats:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a82c72c5990f604d69ecee1667692d9aaeccf11e/spinalcordtoolbox/deepseg/models.py#L462-L464

joshuacwnewton (changes requested):
This is a fairly simple PR and it mostly LGTM! I just had some small documentation comments + requesting a test (which is easy enough to add).

My other comments are mainly just double-checking, to ensure everything still works with the existing `sct_deepseg` infrastructure.

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Note that we also want to keep the description short, so that everything fits in the table when we type `sct_deepseg -h`.

spinalcordtoolbox/deepseg/models.py
Nilser3: I think we should use `UNIT1`, because we have datasets (ie. `nih-ms-mp2rage` , `hc-leipzig-7t-mp2rage`) with more than `UNIT1` contrast (ie. `T1map`, `desc-denoising_UNIT1`).

spinalcordtoolbox/deepseg/models.py
mguaypaq: I think the ensemble model logic is worth keeping around.

I typically like to remove dead code, because:
1. it's untested, and
2. it puts constraints on the rest of the code, which has to stay compatible.

Here point 1 still applies, but we know that it was successfully used at least once, so that mitigates it.

For point 2, extra constraints on the rest of the code are usually annoying, but here I think they're beneficial: the dead code actually serves as a precise, executable form of documentation for the kinds of things that we need to keep in mind in the future, if we want to stay compatible with ensemble models with our data formats, file and folder naming, etc.

joshuacwnewton (approved):

joshuacwnewton:
1 ivadomed model down! Great work! :tada:

Issue #4545, opened 2024-06-20
Add entry `-p sct_deepseg`


jcohenadad:
However, given that there are multiple models in deepseg, this feature will not be straightforward. It will be part of the major refactoring. 

I'm just opening the issue for prosperity, because of https://github.com/sct-pipeline/spine-park/issues/39

joshuacwnewton:
> However, given that there are multiple models in deepseg, this feature will not be straightforward.

Actually, the `sct_deepseg` QC that I added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4446 does indeed try to be "general" (supporting spinal cord AND lesion seg/softseg). It works for everything but rootlets QC (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4461). So, we are 90% there already for the `sct_deepseg` QC feature.

The [reasons](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4446/files#r1577979078) I didn't add `-p sct_deepseg` to `sct_qc` at the time were:

> 1. `sct_deepseg` QC was generated using the "new" `qc2.py` interface, and I didn't want to mix `qc2.py` and `qc.py` in our QC-specific script, as it would end up being quite messy/complex.
> 2. I figure that if folks are using `sct_qc` to generate reports (rather than the built-in per-CLI QC), then they're skilled enough to be able to re-use the existing `-p sct_deepseg_sc` or `-p sct_deepseg_lesion` to generate whichever reports they desire. 

I see now from context that you need `-p sct_qc` [for manual SC segmentations](https://github.com/sct-pipeline/spine-park/blob/8e66fb444aa1a887babaf2fc2d56b966acbe801e/batch_processing.sh#L80-L102). Since this is just a SC QC, are you able to re-use `sct_deepseg_sc` for now?

jcohenadad:
> Since this is just a SC QC, are you able to re-use sct_deepseg_sc for now?

yup, that's exactly what I did https://github.com/sct-pipeline/spine-park/commit/46499f4fa7e7ad2bf6663763736af9863461cc58
thank you for the detailed background information!

joshuacwnewton:
A user on the forum has requested the ability to generate the `sct_deepseg` report via `sct_qc`, so I'm increasing the priority of this issue, especially as it is long overdue.

PR #4555, opened 2024-07-04
Fix `sct_deepseg` rootlets QC by improving the cropping, centering, and colormap

### LINKED ISSUE #4461: Fix issues with `sct_deepseg` rootlets QC report
Most `sct_deepseg` models involve binary/soft segmentations within the spinal cord (full cord, gm/wm, lesions, etc.). So, the initial PR #4446 works well for most models.

However, the rootlets model is quite different:

- It outputs integer labels (2.0, 3.0, 4.0, etc.) instead of a binary/soft seg.
- The rootlets labels are outside the spinal cord, with variable center of mass.

This means that we get an inconsistent crop + wrong colormap with the current "universal" `sct_deepseg` QC.

### LINKED ISSUE #4553: `matplotlib` has deprecated `cm.get_cmap` (use `pyplot.get_cmap` instead)
Source:

- https://github.com/matplotlib/matplotlib/commit/7f23ee15f80c279100f5909ca1d681c575aeb23b

We use `cm.get_cmap` in one place, but I guess our tests don't traverse this bit of code. (It was only added in the context of the TotalSegmentator labels).

I'll fix this as part of #4461 since that's where I came across this issue.

joshuacwnewton:
## Description

This PR adds a number of fixes to create a more legible QC report for the rootlets segmentation.

Before:

image

- Centered on rootlets seg (not SC seg)
- Wrong colormap (red for all rootlet labels)
- Tight crop (would exclude rootlets if centered on SC)
- Resampled rootlets seg (not pixel-accurate)

After:

image

- Centered on SC seg
- Label-focused colormap
- Bigger crop
- Rescaling instead of resampling
- Text to distinguish the labels in addition to color
- Light outline around segmentation to improve color contrast

## Discussion

Right now, the `sct_deepseg` QC report is littered with `if "seg_spinal_rootlets_t2w" in argv:` statements. When there was only 1 or 2 tweaks, then maybe this was okay. But, since there are 4+ tweaks, I'm wondering if it would be more appropriate to create an entirely new QC report.

At what point does it stop making sense to re-use the overall QC report structure? :thinking: 

## Linked issues

Fixes #4461.
Fixes #4553.


valosekj:
Thank you @joshuacwnewton!

> After:
> 
> ezgif-4-be21564466 

This looks great! Two minor thoughts:
- I feel a bit that the change from the original image to the segmentation with the transparent black layer and back is slightly distracting because of the change in contrast between the two due to the transparent black layer. On the other hand, I fully understand the [rationale behind using the transparent black layer](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4461#issuecomment-2209456419). Thinking whether the contrast could be somehow fixed between the two layers.
- Would it be possible to add numbers corresponding to the rootlets to the corner of each image? Something like this:

Screenshot 2024-07-04 at 5 07 59 PM

---

I also wonder if the rootlets QC is available only through the `sct_deepseg` command. Or can I generate it also using `sct_qc`? I'm asking because sometimes we have manually created rootlets, and thus, we want to use `sct_qc` directly as there is no need to segment the rootlets automatically by `sct_deepseg`.

joshuacwnewton:
>  can I generate it also using sct_qc?

Very good point! There is an open issue for this (#4545), and I realize now that adding access to sct_qc will have an impact on how this is implemented (since argv will not contain the task name). I'll have to think on this a little more!

joshuacwnewton:
On second thought, I wonder if the "numbering the rootlets" idea actually fixes the "distracting transparency" problem. Since we know that red is a great contrasting color, we could always keep the red (or even YlOrRed) colormap used by normal segs, but use numbers to distinguish between the rootlets.

For example:

image

valosekj:
> On second thought, I wonder if the "numbering the rootlets" idea actually fixes the "distracting transparency" problem. Since we know that red is a great contrasting color, we could always keep the red (or even YlOrRed) colormap used by normal segs, but use numbers to distinguish between the rootlets.

Good point! I agree that, in this case, we could stick with the red color and numbers! The issue might come when a single slice contains multiple rootles (which can happen due to the rootlet's anatomy and their overlap). In other words, a single slice can have rootlets from two classes, e.g., both 7 and 8.

joshuacwnewton:
Noted!

I spent a bit of time trying to figure out colormaps, but there aren't many colors that contrast well against the fully grayscale spectrum. So, I tried instead to add contouring, which I think helps a lot to highlight light colors while still providing contrast against light backgrounds.

0.5 line width:

image

0.25 line width:

image

valosekj:
Great idea! Agree; contouring looks good! 0.25 could be fine. A thicker contour could "false positively" dilate the rootlets segmentation.

valosekj:
> Right now, none of our `qc2.py` are accessible via `sct_qc`. This is because `sct_qc` mainly uses the old `qc.py` framework (with function wrappers), and would require a refactor in order to merge the two frameworks.
 
I see!

> So, on second thought, I think it might be better to focus on `sct_qc -p sct_deepseg` in a separate PR?

Okay. This makes sense.

valosekj (commented):
I tried the QC on `sub-amu04` from spine-generic:

```console
sct_deepseg -i sub-amu04_T2w.nii.gz -task seg_spinal_rootlets_t2w -qc qc -qc-subject sub-amu04
```

It seems there are no rootlets in the QC:

Kapture 2024-07-25 at 15 59 39

joshuacwnewton:
> It seems there are no rootlets in the QC:

I am able to reproduce this issue. It looks like the issue is with the centerline detection, and therefore the parts of the cord containing the rootlets were cropped out. 

I call `get_centerline` inside the QC function with its default parameters. I assumed this would find the centerline with `optic`. However, the default parameters are actually `bspline`, meaning it only take the center of mass, resulting in this:

image

If I change the parameters to specifically use `optic`, then I get a proper centerline (however I run into another bug with the `optic` method). Once I fix _that_ bug, I get a proper QC:

 
image


I guess this didn't show up because for the test image, the cord is already roughly centered. Thank you so much for catching this!

valosekj:
@KaterinaKrejci231054, could you please also try this PR (T2w rootlets model via `sct_deepseg`) on some T2w images and let us know what you think about the QC?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -369,41 +402,75 @@ def mosaic(img: Image, centers: np.ndarray, radius: tuple[int, int] = (15, 15)):
...
-        (24, 18, 'R'),
+    # Ensure that letter locations are determined as a function of the bounding box. For a 15,15 radius (30x30):
+    #    A                    [12,  6]
+    # L     R   -->  [0, 17]            [24, 17]
+    #    P                    [12, 28]
+    for letter, x, y, in [
+        ('A', radius[0] - 3,   6),
+        ('P', radius[0] - 3,   radius[1]*2 - 2),
+        ('L', 0,               radius[1] + 2),
+        ('R', radius[0]*2 - 6, radius[1] + 2)
```
mguaypaq, line 437: This is lovely, both the code and the comment :heart:

mguaypaq (commented):
Here are a few quick comments. I'll add some commits to address them.

mguaypaq (approved):
Ok, here are my proposed changes! @joshuacwnewton feel free to merge after reviewing (unless you want further changes, of course).

mguaypaq:
Looks like there's a test failure specifically on Windows:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/10206792498/job/28240356757?pr=4555#step:6:1190

joshuacwnewton:
Looks like there was a download error during installation: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/10206792498/job/28240356757?pr=4555#step:3:626

I'll try restarting to see if it was a one-off... :thinking:

joshuacwnewton:
I manually checked the QC, and it LGTM! Thank you for the refactoring changes. :hearts:

PR #4557, opened 2024-07-08
Minor improvements

mguaypaq:
~~This is a large change, and probably controversial, so I'm opening this PR-upon-a-PR (for #4485) to give @joshuacwnewton a chance to give feedback on whether it's acceptable.~~

~~I haven't tested it yet, but this is the structure I had in mind. Is it overkill, or too much boilerplate?~~

EDIT: It was overkill, but still contained a few minor improvements.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -23,7 +23,7 @@ from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import extract_fname
 
 
-def number_or_image(arg: str) -> Union[float, str]:
+def number_or_fname(arg: str) -> Union[float, str]:
     """
     Parse an argument as either a number or an image file name.
 
@@ -109,7 +109,7 @@ def get_parser():
     mandatory.add_argument(
         "-i",
         metavar=Metavar.file,
-        type=number_or_image,
+        type=number_or_fname,
         help="Input file. Example: data.nii.gz",
         required=True)
     mandatory.add_argument(
@@ -130,7 +130,7 @@ def get_parser():
         "-add",
         metavar='',
         nargs="*",
-        type=number_or_image,
+        type=number_or_fname,
         action=StoreTodo,
         help='Add following input. Can be a number or one or more 3D/4D images (separated with space). Examples:'
              '\n  - sct_maths -i 3D.nii.gz -add 5                       (Result: 3D image with "5" added to each voxel)'
@@ -146,7 +146,7 @@ def get_parser():
         "-sub",
         metavar='',
         nargs="+",
-        type=number_or_image,
+        type=number_or_fname,
         action=StoreTodo,
         help='Subtract following input. Can be a number, or one or more 3D/4D images (separated with space).',
         required=False)
@@ -154,7 +154,7 @@ def get_parser():
         "-mul",
         metavar='',
         nargs="*",
-        type=number_or_image,
+        type=number_or_fname,
         action=StoreTodo,
         help='Multiply by following input. Can be a number, or one or more 3D/4D images (separated with space). '
              '(See -add for examples.)',
@@ -163,7 +163,7 @@ def get_parser():
         "-div",
         metavar='',
         nargs="+",
-        type=number_or_image,
+        type=number_or_fname,
         action=StoreTodo,
         help='Divide by following input. Can be a number, or one or more 3D/4D images (separated with space).',
         required=False)
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -63,7 +63,7 @@ class StoreTodo(argparse.Action):
     """
     def __call__(self, parser, namespace, values, option_string=None):
         # make sure values is a list, rather than a single bare value
-        if self.nargs in [None, '?']:
+        if not isinstance(values, list):
             values = [values]
 
         namespace.todo.append((self.dest, values, {}))
@@ -80,7 +80,7 @@ class StoreTodoWithShapeDim(argparse.Action):
     """
     def __call__(self, parser, namespace, values, option_string=None):
         # make sure values is a list, rather than a single bare value
-        if self.nargs in [None, '?']:
+        if not isinstance(values, list):
             values = [values]
 
         shape = namespace.shape
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -9,7 +9,7 @@ import sys
 import pickle
 import gzip
 import argparse
-from typing import Sequence
+from typing import Sequence, Union
 
 import numpy as np
 import matplotlib.pyplot as plt
@@ -19,6 +19,38 @@ from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, list_type, display_viewer_syntax
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import extract_fname
+
+
+def number_or_image(arg: str) -> Union[float, str]:
+    """
+    Parse an argument as either a number or an image file name.
+
+    Can be used as the `type` argument to `parser.add_argument`.
+
+    Examples:
+    - '-arg 5'             (int argument, converted to float)
+    - '-arg 5.0'           (float argument)
+    - '-arg image.nii.gz'  (image file name)
+    """
+    try:
+        return float(arg)
+    except ValueError:
+        return arg
+
+
+def denoise_params(arg: str) -> tuple[int, int]:
+    """
+    Parse the argument for `-denoise` into a pair (patch radius, block radius).
+
+    Any sub-arguments that don't start with "p=" or "b=" are silently ignored.
+    """
+    p, b = 1, 5  # defaults
+    for sub_arg in arg.split(","):
+        if sub_arg.startswith("p="):
+            p = int(sub_arg[2:])
+        elif sub_arg.startswith("b="):
+            b = int(sub_arg[2:])
+    return (p, b)
 
 
 class ParseDataOrScalarArgument(argparse.Action):
@@ -86,6 +118,7 @@ def get_parser():
     mandatory.add_argument(
         "-i",
         metavar=Metavar.file,
+        type=number_or_image,
         help="Input file. Example: data.nii.gz",
         required=True)
     mandatory.add_argument(
@@ -106,6 +139,7 @@ def get_parser():
         "-add",
         metavar='',
         nargs="*",
+        type=number_or_image,
         action=ParseDataOrScalarArgument,
         help='Add following input. Can be a number or one or more 3D/4D images (separated with space). Examples:'
              '\n  - sct_maths -i 3D.nii.gz -add 5                       (Result: 3D image with "5" added to each voxel)'
@@ -121,6 +155,7 @@ def get_parser():
         "-sub",
         metavar='',
         nargs="+",
+        type=number_or_image,
         action=ParseDataOrScalarArgument,
         help='Subtract following input. Can be a number, or one or more 3D/4D images (separated with space).',
         required=False)
@@ -128,6 +163,7 @@ def get_parser():
         "-mul",
         metavar='',
         nargs="*",
+        type=number_or_image,
         action=ParseDataOrScalarArgument,
         help='Multiply by following input. Can be a number, or one or more 3D/4D images (separated with space). '
              '(See -add for examples.)',
@@ -136,6 +172,7 @@ def get_parser():
         "-div",
         metavar='',
         nargs="+",
+        type=number_or_image,
         action=ParseDataOrScalarArgument,
         help='Divide by following input. Can be a number, or one or more 3D/4D images (separated with space).',
         required=False)
@@ -171,7 +208,7 @@ def get_parser():
     thresholding.add_argument(
         "-adap",
         metavar=Metavar.list,
-        type=list_type(',', int),
+        type=list_type(',', int, 2),
         help="Threshold image using Adaptive algorithm (from skimage). Provide 2 values separated by ',' that "
              "correspond to the parameters below. For example, '-adap 7,0' corresponds to a block size of 7 and an "
              "offset of 0.\n"
@@ -182,7 +219,7 @@ def get_parser():
     thresholding.add_argument(
         "-otsu-median",
         metavar=Metavar.list,
-        type=list_type(',', int),
+        type=list_type(',', int, 2),
         help="Threshold image using Median Otsu algorithm (from dipy). Provide 2 values separated by ',' that "
              "correspond to the parameters below. For example, '-otsu-median 3,5' corresponds to a filter size of 3 "
              "repeated over 5 iterations.\n"
@@ -259,7 +296,8 @@ def get_parser():
         required=False)
     filtering.add_argument(
         '-denoise',
-        help='Non-local means adaptative denoising from P. Coupe et al. as implemented in dipy. Separate with ". Example: p=1,b=3\n'
+        type=denoise_params,
+        help='Non-local means adaptative denoising from P. Coupe et al. as implemented in dipy. Separate with "," Example: p=1,b=3\n'
              ' p: (patch radius) similar patches in the non-local means are searched for locally, inside a cube of side 2*p+1 centered at each voxel of interest. Default: p=1\n'
              ' b: (block radius) the size of the block to be used (2*b+1) in the blockwise non-local means implementation. Default: b=5 '
              '    Note, block radius must be smaller than the smaller image dimension: default value is lowered for small images)\n'
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -683,6 +683,21 @@ def apply_denoise(current_value, patch_radius, block_radius):
     )
 
 
+def apply_mi(current_value, fname):
+    """Implementation for -mi."""
+    return (current_value, Image(fname), 'mi', 'Mutual information')
+
+
+def apply_minorm(current_value, fname):
+    """Implementation for -minorm."""
+    return (current_value, Image(fname), 'minorm', 'Normalized Mutual information')
+
+
+def apply_corr(current_value, fname):
+    """Implementation for -corr."""
+    return (current_value, Image(fname), 'corr', 'Pearson correlation coefficient')
+
+
 def apply_symmetrize(current_value, axis):
     """Implementation for -symmetrize."""
     return Image(
@@ -712,9 +727,9 @@ APPLY = {
     "smooth": apply_smooth,
     "laplacian": apply_laplacian,
     "denoise": apply_denoise,
-    "mi": None,
-    "minorm": None,
-    "corr": None,
+    "mi": apply_mi,
+    "minorm": apply_minorm,
+    "corr": apply_corr,
     "symmetrize": apply_symmetrize,
 }
 
@@ -760,31 +775,6 @@ def main(argv: Sequence[str]):
         compute_similarity(img1, img2, arguments.o, metric, metric_full, verbose)
         printv(f"\nDone! File created: {arguments.o}", verbose, 'info')
 
-    fname_in = arguments.i
-    fname_out = arguments.o
-    output_type = arguments.type
-
-    # Case 1: Compute similarity metrics
-    if any(getattr(arguments, similarity_arg) is not None for similarity_arg in ['mi', 'minorm', 'corr']):
-        if arguments.mi is not None:
-            # input 1 = from flag -i --> im
-            # input 2 = from flag -mi
-            im_2 = Image(arguments.mi)
-            compute_similarity(im, im_2, fname_out, metric='mi', metric_full='Mutual information', verbose=verbose)
-
-        elif arguments.minorm is not None:
-            im_2 = Image(arguments.minorm)
-            compute_similarity(im, im_2, fname_out, metric='minorm', metric_full='Normalized Mutual information',
-                               verbose=verbose)
-
-        else:
-            assert arguments.corr is not None
-            # input 1 = from flag -i --> im
-            # input 2 = from flag -mi
-            im_2 = Image(arguments.corr)
-            compute_similarity(im, im_2, fname_out, metric='corr', metric_full='Pearson correlation coefficient',
-                               verbose=verbose)
-
 
 def compute_similarity(img1: Image, img2: Image, fname_out: str, metric: str, metric_full: str, verbose):
     """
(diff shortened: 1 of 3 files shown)


spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -21,71 +23,93 @@
...
+
+        namespace.todo.append((self.dest, values, {}))
+
+
+class StoreTodoWithShapeDim(argparse.Action):
+    """
+    Store the arguments of `-dilate` or `-erode` in `arguments.todo`.
+
+    These two operations are special-cased because they need the "current"
+    values of `-shape` and `-dim` to also be saved.
```
mguaypaq, line 77: I also like your previous suggestion of putting `shape` and `dim` information directly into the argument for `-dilate` and `-erode`, but that's not backwards compatible, so I thought it was maybe out of scope.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -21,71 +23,93 @@
...
+
+    The format is: (operation name, positional args, keyword args).
```
mguaypaq, line 79: I chose this format because it should be easily readable in the debugger.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -253,13 +295,16 @@ def __call__(self, _, namespace, values, option_string=None):
...
         help='Laplacian filtering. Supply values for standard deviations in mm. If a single value is provided, it will '
              'be applied to each axis of the image. If multiple values are provided, there must be one value per '
              'image axis. (Examples: "-laplacian 2.0,3.0,2.0" (3D image), "-laplacian 2.0" (any-D image)).',
         required=False)
     filtering.add_argument(
         '-denoise',
-        help='Non-local means adaptative denoising from P. Coupe et al. as implemented in dipy. Separate with ". Example: p=1,b=3\n'
+        type=denoise_params,
+        action=StoreTodo,
+        help='Non-local means adaptative denoising from P. Coupe et al. as implemented in dipy. Separate with "," Example: p=1,b=3\n'
```
mguaypaq, line 307: Fixed a typo in the help text.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+        help="Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode",
+        required=False)
 
     return parser
 
 
-def apply_array_operation(data, dim, arg_name, arg_value, parser):
-    dim_list = ['x', 'y', 'z', 't']
+class SctMathsTypeError(Exception):
+    """Inappropriate argument type for an `sct_maths` operation."""
```
mguaypaq, line 365: I made custom exception classes so that we could catch them specifically, but maybe that's overkill.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+    "corr": check_metric,
+    "symmetrize": check_image_to_image,
+}
+
+
+class SctMathsValueError(Exception):
+    """Inappropriate argument value for an `sct_maths` operation."""
+
+
+def get_header_and_values(
```
mguaypaq, line 432: I took me a while to come up with this argument normalizer, but I'm happy with it. Making sure that it had a somewhat reasonable type signature really helped me figure it out.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+    header, values = get_header_and_values(current_value, args)
+    if header is None:
+        return sum(values)
     else:
-        assert arg_name == "symmetrize"
-        data_out = sct_math.symmetrize(data, arg_value)
+        return Image(
+            np.sum(values, axis=0),
+            hdr=header,
+        )
```
mguaypaq, line 495: Note that the `Image` constructor automatically adjusts the header to match the shape of the data array.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
-            # input 2 = from flag -mi
-            im_2 = Image(arguments.corr)
-            compute_similarity(im, im_2, fname_out, metric='corr', metric_full='Pearson correlation coefficient',
-                               verbose=verbose)
+def apply_mean(current_value, dim):
+    """Implementation for -mean."""
+    if current_value.data.ndim == 3 and dim == 't':
+        return current_value
+    return Image(
+        np.mean(current_value.data, ('x', 'y', 'z', 't').index(dim)),
```
mguaypaq, line 546: I think this operation makes no sense when `dim != 't'`? Something to discuss at the SCT meeting.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+
+def apply_otsu(current_value, nbins):
+    """Implementation for -otsu."""
+    return Image(
+        sct_math.otsu(current_value.data, nbins),
+        hdr=current_value.hdr,
+    )
+
+
+def apply_adap(current_value, block_size, offset):
```
mguaypaq, line 586: One thing I like about this structure is that we can give meaningful names to the arguments (like `block_size` and `offset`), rather than always using `arg[0]` and `arg[1]`.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+
+def apply_uthr(current_value, threshold):
+    """Implementation for -uthr."""
+    return Image(
+        sct_math.threshold(current_value.data, uthr=threshold),
+        hdr=current_value.hdr,
+    )
+
+
+def apply_dilate(current_value, size, *, shape, dim):
```
mguaypaq, line 626: And keyword arguments are also possible, like `shape` and `dim` here.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+    """Implementation for -smooth."""
+    ndims = current_value.data.ndims
+    if len(sigmas) == 1:
+        sigmas *= ndims
+    elif len(sigmas) != ndims:
+        raise SctMathsValueError(f"expected 1 or {ndims} values, got {len(sigmas)} values")
+
+    # adjust sigma based on voxel size
+    pdims = current_value.dim[4:8]
+    sigmas = [s/d for s, d in zip(sigmas, pdims)]
```
mguaypaq, line 652: We should discuss at the SCT meeting whether this operation makes sense for 4D images.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+    """Implementation for -denoise."""
+    return Image(
+        sct_math.denoise_nlmeans(current_value.data, patch_radius, block_radius),
+        hdr=current_value.hdr,
+    )
+
+
+def apply_mi(current_value, fname):
+    """Implementation for -mi."""
+    return (current_value, Image(fname), 'mi', 'Mutual information')
```
mguaypaq, line 688: I'm still on-board for turning the similarity metrics into their own SCT command outside of `sct_maths`, but I think this design is livable until then.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -270,223 +315,465 @@ def __call__(self, _, namespace, values, option_string=None):
...
+    parser = get_parser()
+    arguments = parser.parse_args(argv)
+    verbose = arguments.v
+    set_loglevel(verbose=verbose)
+
+    # Check data types
+    current_type = 'number' if isinstance(arguments.i, float) else 'image'
+    for operation, args, kwargs in arguments.todo:
+        try:
+            current_type = CHECK[operation](current_type, *args, **kwargs)
```
mguaypaq, line 749: It turns out that simple type checking and simple program-running have basically the same structure: we just have to replace the `CHECK` dictionary with `APPLY`.

mguaypaq (commented):
Some self-review comments.

mguaypaq:
We're able to manually trigger workflows, so I kicked off a few test runs:
* [Tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9859202001)
* [Test batch_processing.sh](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9859204140)

(I think they're just attached to the branch, not the PR, so they probably won't show up under the "Checks" tab.)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -9,8 +9,10 @@
 import pickle
 import gzip
 import argparse
-from typing import Sequence
+import math
+from typing import Sequence, Union
 
+import nibabel as nib
```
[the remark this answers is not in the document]
joshuacwnewton: Nibabel is one that we currently aren't lazy-importing, yes!

It's probably the 6th or 7th most expensive import (and 1-6 e.g. torch/monai/etc. have been handled already), so I'd say it's next on the chopping block. But, I agree that it's out of scope for this PR.

spinalcordtoolbox/scripts/sct_maths.py
joshuacwnewton: That's my intuition as well, but definitely worth bringing up.

joshuacwnewton (changes requested):
As mentioned on Slack, I really love the bulk of the changes here! The `CHECK`/`APPLY` structure makes a lot of sense to me. :)

My main feedback is regarding the input/output typing for both the arithmetic and metric operations? As well as a handful of nitpicks here and there.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -21,6 +21,38 @@
 from spinalcordtoolbox.utils.fs import extract_fname
 
 
+def number_or_image(arg: str) -> Union[float, str]:
```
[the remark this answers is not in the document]
mguaypaq: Absolutely! Thanks for the suggestion :heart:

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -53,66 +53,55 @@ def denoise_params(arg: str) -> tuple[int, int]:
...
-            getattr(namespace, "ordered_args", []) + [(dest, values)])
+    def ___call___(self, parser, namespace, values, option_string=None):
+        # make sure values is a list, rather than a single bare value
+        if self.nargs in [None, '?']:
+            values = [values]
+
+        shape = namespace.shape
+        if shape in ['disk', 'square']:
+            # 2D kernels need the value of `-dim`
+            dim = namespace.dim
```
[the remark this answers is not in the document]
mguaypaq: True, I hadn't thought of that. If we want to support `-shape` and `-dim` coming _after_ the corresponding `-dilate` or `-erode` (which we currently do), then I guess there's no way to have different values for different operations. Probably the friendliest thing to do is detect multiple instances of `-shape` or `-dim` and give a parse error.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -483,6 +547,21 @@ def main(argv: Sequence[str]):
     verbose = arguments.v
     set_loglevel(verbose=verbose)
 
+    # Check data types
+    current_type = 'number' if isinstance(arguments.i, float) else 'image'
```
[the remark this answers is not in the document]
mguaypaq: I think it doesn't support it currently, but I wanted to add support specifically for two use cases:
- negating an image with `sct_maths -i 0 -sub img.nii.gz`
- inverting an image with `sct_maths -i 1 -div img.nii.gz`

I'm willing to accept that it's too much added complexity, though. (And it could probably be addressed with new `-neg` and `-inv` operators.)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -483,6 +547,21 @@ def main(argv: Sequence[str]):
...
+    for operation, args, kwargs in arguments.todo:
+        try:
+            current_type = CHECK[operation](current_type, *args, **kwargs)
+        except SctMathsTypeError as e:
+            parser.error(f"wrong type for -{operation}: {e}")
+    final_type = current_type
+    if final_type == 'number':
+        parser.error('there must be at least one image in the list of inputs')
```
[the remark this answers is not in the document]
mguaypaq: Yes, assuming that `-i` is always an image would simplify several things. I'll add some commits to do this.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -359,6 +359,70 @@ def get_parser():
...
+def check_arithmetic(current_type: str, *args) -> str:
+    """Type checker for `-add`, `-sub`, etc."""
+    # special case for -add, -mul to operate on a 4D volume across the t axis
+    if not args:
+        return check_image_to_image(current_type)
+
+    # in the usual case, numbers can get promoted to images
+    if current_type == 'number':
+        if all(isinstance(arg, float) for arg in args):
+            return 'number'
```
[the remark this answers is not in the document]
mguaypaq: I'll remove this ability and simplify. Haha but I'm still happy to have done the exercise, even if it gets scrapped.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -562,19 +594,18 @@ def main(argv: Sequence[str]):
             printv("WARNING: Output type conversion has no effect for similarity metrics, "
                    "-type argument will be ignored.", verbose=verbose, type='warning')
 
+    # Actually do the computations
+    current_value = arguments.i if isinstance(arguments.i, float) else Image(arguments.i)
+    for operation, args, kwargs in arguments.todo:
```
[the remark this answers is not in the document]
mguaypaq: I was thinking that the no-operation case was useful for type conversion (which happens in post-processing), but I guess we already have `sct_image -type` for that. I'll add an error.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -538,15 +538,44 @@ def apply_div(current_value, *args):
...
+
+
+def apply_rms(current_value, dim):
+    """Implementation for -rms."""
+    data = current_value.data.astype(float)
+    if current_value.data.ndim == 3 and dim == 't':
+        rms = np.abs(data)
+    else:
+        axis = ('x', 'y', 'z', 't').index(dim)
+        rms = np.sqrt(np.mean(np.square(data), axis))
```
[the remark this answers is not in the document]
mguaypaq: Yes, I'll add a comment, thanks for the suggestion. Given that there's an `if` statement either way (to add a dummy 4th dimention in some cases), I'm not convinced that the old code was more elegant.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -538,15 +538,44 @@ def apply_div(current_value, *args):
...
+    return Image(rms, hdr=current_value.hdr)
+
+
+def apply_std(current_value, dim):
+    """Implementation for -std."""
+    return Image(
+        np.std(current_value.data, ('x', 'y', 'z', 't').index(dim), ddof=1),
+        hdr=current_value.hdr,
+    )
```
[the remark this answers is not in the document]
mguaypaq: I should add a comment here too! I looked into it and determined that the "3D+T" case makes no sense. The reason is `ddof=1`:
```
    ddof : int, optional
        Means Delta Degrees of Freedom.  The divisor used in calculations
        is ``N - ddof``, where ``N`` represents the number of elements.
        By default `ddof` is zero.
```

In the "3D+T" case, we would have `N = 1` for the dummy 4th dimension, which would give a division by zero.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -623,6 +623,22 @@ def apply_uthr(current_value, threshold):
     )
 
 
+def apply_dilate(current_value, size, *, shape, dim):
```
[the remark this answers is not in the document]
mguaypaq: It's not necessary, actually. But just like `--` in bash commands, it's used to differentiate between different kinds or arguments. If we ignore `*args` and `**kwargs`, the function signature syntax for python functions is:
```py
def f(positional_only, /, positional_or_keyword, *, keyword_only):
```

I guess I wanted to be explicit that `shape` and `dim` are given as keywords, but on second thought that's a bit too fancy for our codebase.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -639,6 +639,42 @@ def apply_erode(current_value, size, *, shape, dim):
...
+    """Implementation for -laplacian."""
+    ndims = current_value.data.ndims
+    if len(sigmas) == 1:
+        sigmas *= ndims
+    elif len(sigmas) != ndims:
+        raise SctMathsValueError(f"expected 1 or {ndims} values, got {len(sigmas)} values")
+
+    # adjust sigma based on voxel size
+    pdims = current_value.dim[4:8]
+    sigmas = [s/d for s, d in zip(sigmas, pdims)]
```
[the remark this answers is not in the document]
mguaypaq: Yes, to both comments.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -750,6 +750,16 @@ def main(argv: Sequence[str]):
...
 
+    # Post-processing of the results
+    if final_type == 'image':
+        current_value.save(arguments.o, dtype=arguments.type)
+        display_viewer_syntax([arguments.o], verbose=verbose)
+    else:
+        assert final_type == 'metric'
+        img1, img2, metric, metric_full = current_value
+        compute_similarity(img1, img2, arguments.o, metric, metric_full, verbose)
+        printv(f"\nDone! File created: {arguments.o}", verbose, 'info')
```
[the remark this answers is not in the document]
mguaypaq: Sure, I'll add a check for extra safety. (And I'll rename it `final_value` instead of `current_value`, to mirror `final_type`.)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -750,6 +750,16 @@ def main(argv: Sequence[str]):
...
             printv(f"ERROR: -{operation}: {e}", 1, 'error')
 
+    # Post-processing of the results
+    if final_type == 'image':
+        current_value.save(arguments.o, dtype=arguments.type)
+        display_viewer_syntax([arguments.o], verbose=verbose)
+    else:
+        assert final_type == 'metric'
+        img1, img2, metric, metric_full = current_value
+        compute_similarity(img1, img2, arguments.o, metric, metric_full, verbose)
```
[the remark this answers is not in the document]
mguaypaq: How about `metric_todo` instead of `metric` for the name of the tuple? I think it pretty much _has_ to happen in the post-processing part of the script, but it would be more explicit. Thanks for your comments about naming, it's important to get it right.

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -683,6 +683,21 @@ def apply_denoise(current_value, patch_radius, block_radius):
     )
 
 
+def apply_mi(current_value, fname):
+    """Implementation for -mi."""
+    return (current_value, Image(fname), 'mi', 'Mutual information')
```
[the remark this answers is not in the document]
mguaypaq: I'm hesitant to touch the inside of `compute_similarity` in this PR, especially if the plan is to spin them off into their own script soon-ish.

spinalcordtoolbox/scripts/sct_maths.py
mguaypaq: I brought it up at the last SCT meeting, and I wish I had taken better notes. But I think we settled on:
- If we take a mean across one of the spatial dimensions (x, y, z), then we should make that axis survive as a length=1 dimension in the output shape. (Otherwise, the affine matrix becomes meaningless.)
- If we take a mean across the time dimension, then the result should be 3-dimensional.

spinalcordtoolbox/scripts/sct_maths.py
mguaypaq: Again, I wish I had taken better notes at the SCT meeting, but from what I recall:
- These operations are indeed only meant for 3D images.
- We may eventually want them to be applied volume-wise to 4D images, but that will be another PR.

mguaypaq (commented):
I simplified various things according to your comments @joshuacwnewton. Since the `CHECK` dictionary is no longer useful, I think I'll revert the `APPLY` dictionary to the previous structure of a simple loop over a big `if...elif` which updates a plain numpy array (rather than an `Image`). Hopefully this will result in a much smaller diff!

mguaypaq:
@joshuacwnewton how do you feel about this version? Much closer to your original PR.

joshuacwnewton:
It looks really good to me! I'm glad you still kept the argparse helper functions/actions.

Though, I'm curious about the thought process behind moving the big if/else back into `main`? I think the reason I had originally encapsulated the if/else into a separate function was to make `main` easier to read through, since the "pre-operation" and "post-operation" steps fit on a single screen (minimizing the need to scroll when reading). But, I'm curious to hear your thoughts!

mguaypaq:
I guess it comes down to the similarity metrics:
- From your previous comments, I thought you wanted their "implementation" to actually do the call to `compute_similarity`, rather than creating a tuple of arguments.
- But that means they need to create an actual image, from the original image header. And they need to know the output file name.
- And once they're done, they need to replace the normal post-processing (dtype conversion, save image, `display_viewer_syntax`) with their own post-processing (save a non-image file, display a different message).

The last two points break the natural encapsulation of the normal-looking loop:
```py
for arg_name, arg_value in arguments.todo:
    data = apply_operation(data, arg_name, arg_value, shape, dim)
```

If you really prefer having the pre- and post-processing on the same screen, I can make it a separate function anyway, though.

joshuacwnewton (approved):
Ahhhhh of course, right! I had forgotten that the old approach was no longer valid once we parse everything into a single TODO arg (similarity metrics included). 

All good, then! :)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -434,59 +374,190 @@ def main(argv: Sequence[str]):
...
+
+        elif arg_name in ["smooth", "laplacian"]:
+            # Adjust sigmas from millimeters to voxels
+            # This uses the resolution of the starting value `im`
+            sigmas = [mm/pixdim for mm, pixdim in zip(arg_value, im.dim[4:7])]
+            if arg_name == "smooth":
+                data = sct_math.smooth(data, sigmas)
+            else:
+                assert arg_name == "laplacian"
+                data = sct_math.laplacian(data, sigmas)
```
mguaypaq, line 530: Let's use a dict here:
```suggestion
            data = {
                "smooth": sct_math.smooth,
                "laplacian": sct_math.laplacian,
            }[arg_name](data, sigmas)
```

PR #4559, opened 2024-07-10
Update redirected citation link

mguaypaq:
Although the current URL works in the browser, every day it shows up as the only broken link in the link checker workflow. This ~~DOI~~ direct link should be more stable.


jcohenadad (approved):
thanks!

PR #4573, opened 2024-07-24
Rescale flattened values from `[-1.0, 1.0]` back to their original range and datatype

### LINKED ISSUE #3737: Volume values rescaled by sagittal flattening
### Description

I tried using `sct_flatten_sagittal` on a UNI volume acquired as part of a MP2RAGE sequence. The values stored in the original UNI volume  range from 0 to ~4000 (see below).

image

However once flattened, the values range from -1 to 1 (see below)

image

Shouldn't the values remain the same ?

I use the following line of code:

```bash
sct_flatten_sagittal -i UNI.nii.gz -s UNI_seg.nii.gz
```
The files are linked here:
[UNI_seg.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/8257482/UNI_seg.nii.gz)
[UNI.nii.gz](https://github.com/spinalcordtoolbox/spinalcordtoolbox/files/8257481/UNI.nii.gz)

joshuacwnewton:
## Description

This is a very simple PR that restores the flattened image back to its original state to avoid user confusion.

I was uncertain whether to keep the scaled data as `float` or whether to convert back to original datatype (e.g. `int`). Converting back to `int` can cause some small rounding error (e.g. 350.6473 -> 351), but since this image is meant for visualization only, I figured that the rounding errors are not critical.

## Linked issues

Fixes #3737.

### DIFF
--- a/spinalcordtoolbox/flattening.py
+++ b/spinalcordtoolbox/flattening.py
@@ -43,11 +43,12 @@ def flatten_sagittal(im_anat, im_centerline, verbose):
 
     # change type to float32 and scale between -1 and 1 as requested by img_as_float(). See #1790, #2069
     im_anat_flattened = change_type(im_anat, np.float32)
-    min_data, max_data = np.min(im_anat_flattened.data), np.max(im_anat_flattened.data)
-    im_anat_flattened.data -= min_data               # [min, max]   -> [ 0, max-min]
-    im_anat_flattened.data /= (max_data - min_data)  # [0, max-min] -> [ 0, 1]
-    im_anat_flattened.data *= 2                      # [0, 1]       -> [ 0, 2]
-    im_anat_flattened.data -= 1                      # [0, 2]       -> [-1, 1]
+    flattened_data = im_anat_flattened.data.copy()
+    min_data, max_data = np.min(flattened_data), np.max(flattened_data)
+    flattened_data -= min_data               # [min, max]   -> [ 0, max-min]
+    flattened_data /= (max_data - min_data)  # [0, max-min] -> [ 0, 1]
+    flattened_data *= 2                      # [0, 1]       -> [ 0, 2]
+    flattened_data -= 1                      # [0, 2]       -> [-1, 1]
 
     # loop and translate each axial slice, such that the flattened centerline is centered in the medial plane (R-L)
     for iz in range(nz):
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/flattening.py
```diff
@@ -58,7 +58,11 @@ def flatten_sagittal(im_anat, im_centerline, verbose):
         img_reg = transform.warp(img, tform)
         im_anat_flattened.data[:, :, iz] = img_reg
 
-    # change back to native orientation
+    # Change [-1, 1] values back to the original range ([min, max])
+    im_anat_flattened.data = (im_anat_flattened.data + 1) * (max_data - min_data) / 2
```
mguaypaq, line 62: Good news and bad news:
- This formula _is_ the inverse transformation of the original transformation.
- But the original transformation _doesn't_ map `[min, max]` to `[-1, 1]`.

I think the original should be:
```py3
im_anat_flattened.data = 2/(max_data - min_data) * (im_anat_flattened.data - (max_data + min_data)/2)
```

And then the reverse would be:
```py3
im_anat_flattened.data = (max_data - min_data)/2 * im_anat_flattened.data + (max_data + min_data)/2
```

mguaypaq (changes requested):
I think there's a problem with your formula, but only because there was already a problem with the existing formula.

spinalcordtoolbox/flattening.py
joshuacwnewton: > But the original transformation doesn't map [min, max] to [-1, 1].

Ahhh, good catch! I think the current formula is only valid when `min == 0`, in which case it becomes:

```python
# original LOC
im_anat_flattened.data = 2 * im_anat_flattened.data / (max_data - min_data) - 1

# when min == 0
im_anat_flattened.data = 2 * im_anat_flattened.data / max_data - 1

# substeps:
im_anat_flattened.data /= max_data  # [0, max] -> [ 0, 1]
im_anat_flattened.data *= 2         # [0, 1]   -> [ 0, 2]
im_anat_flattened.data -= 1         # [0, 2]   -> [-1, 1]
```

That might explain how this formula was originally determined? However, if `min != 0`, then a shift is needed prior to the division step:

```python
# substeps:
im_anat_flattened.data -= min_data               # [min, max]   -> [ 0, max-min]
im_anat_flattened.data /= (max_data - min_data)  # [0, max-min] -> [ 0, 1]
im_anat_flattened.data *= 2                      # [0, 1]       -> [ 0, 2]
im_anat_flattened.data -= 1                      # [0, 2]       -> [-1, 1]
```

So, I think the original could be simplified to:

```python
# [min, max] -> [0, max-min] -> [0, 1] -> [0, 2] -> [-1, 1]
im_anat_flattened.data = ((im_anat_flattened.data - min_data) / (max_data - min_data) * 2) - 1
```

spinalcordtoolbox/flattening.py
```diff
@@ -44,7 +44,10 @@ def flatten_sagittal(im_anat, im_centerline, verbose):
     # change type to float32 and scale between -1 and 1 as requested by img_as_float(). See #1790, #2069
     im_anat_flattened = change_type(im_anat, np.float32)
     min_data, max_data = np.min(im_anat_flattened.data), np.max(im_anat_flattened.data)
-    im_anat_flattened.data = 2 * im_anat_flattened.data / (max_data - min_data) - 1
+    im_anat_flattened.data -= min_data               # [min, max]   -> [ 0, max-min]
```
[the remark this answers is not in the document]
joshuacwnewton: Ah, thank you for pointing this out! I can make sure that the steps operate on a copy of the data?

PR #4574, opened 2024-07-25
Refactor `qc.py` to use `create_qc_entry`

mguaypaq:
I'm not finished with porting the existing QC reports to the new structure, but I've completed a significant step, and it's coherent enough to be reviewed, so here's a pull request!

My approach is to erode away the old structure by inlining as many function calls as possible, especially when they're only ever called once.

I tried to make a separate commit for each logical change, but the last one was necessarily bigger.

Part of #4204.

### DIFF
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -304,8 +304,6 @@ class QcImage:
             action(self, mask[i], ax)
         self._save(fig, self.qc_report.abs_overlay_img_path(), dpi=self.qc_report.dpi)
 
-        self.qc_report.update_description_file()
-
     def _make_QC_image_for_4d_volumes(self, images_after_moco, images_before_moco):
         """
         Generate background and overlay gifs for sct_fmri_moco and sct_dmri_moco
@@ -323,8 +321,6 @@ class QcImage:
 
         self._generate_and_save_gif(images_before_moco, images_after_moco, size_fig)
         self._generate_and_save_gif(images_before_moco, images_after_moco, size_fig, is_mask=True)
-
-        self.qc_report.update_description_file()
 
     def _func_stretch_contrast(self, img):
         if self._stretch_contrast_method == "equalized":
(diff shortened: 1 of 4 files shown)


mguaypaq:
I'll have to adjust the tests for these changes.

joshuacwnewton (commented):
On first read-through, this all seems very sensible to me!

To summarize, we're basically deconstructing the main `qc.py` classes (`QcReport`, `QcImage`, and both the `Sagittal`/`Axial` slice classes).

- `QcReport`: Previously responsible for generating the QC files (images, json, assets). Fully replaced by `create_qc_entry()`, which covers the same responsibility.
- `QcImage`: Previously responsible for QC image data, specifically the layouts defined by `-process` (plotting images, colormaps, etc.). It still does this, but it's now disentangled from `QcReport`, and instead writes its plots directly to the filenames requested by `create_qc_entry()`,
- `Slice`: Previously responsible for QC image data, specifically the Sagittal/Axial layout code (mosaic, single) that is agnostic of `-process`. It still does this, but has been refactored to make it simpler (e.g. inlined methods).

So, I would guess that the next steps are:

- `QcImage`: Convert these methods into functions similar to the existing `qc2.py` functions (`sct_deepseg()`, `sct_register_multimodal()`, etc.)
- `Slice`: Convert these methods into functions similar to the existing `qc2.py` function `mosaic()`.

I'll wait to approve until I see the test changes, but so far this LGTM!

mguaypaq:
Ok! The tests are now updated and pass @joshuacwnewton

spinalcordtoolbox/reports/qc.py
```diff
@@ -776,18 +776,24 @@ def qcslice_layout(x): return x.mosaic()[:2]
         p_resample = p_resample_default
     elif p_resample == 0:   # If user specified `-resample 0`, turn off resampling
         p_resample = None
-    qcslice = SliceSubtype(im_list, p_resample=p_resample)
-    qc_report = QcReport(fname_in1, process, args, plane, path_qc, dpi, dataset, subject)
 
-    QcImage(
+    qcslice = object.__new__(SliceSubtype)
```
joshuacwnewton, line 596: For my own understanding... is there a difference between the `object.__new__.Slice` syntax (used here) and the `Slice()` syntax (used in the test refactors)?

joshuacwnewton (approved):
This LGTM! I ran through each of the commits a second time, and your changes look sensible to me.

As a further check, I downloaded the QC from the `batch_processing.sh` test before and after the changes here, and visually inspected each of the reports. They appear visually identical to me (and have the same filesizes), so I'm giving this my approval. :)

spinalcordtoolbox/reports/qc.py
mguaypaq: Oh, good catch!

In general, yes, there's an important difference. The `Slice()` syntax implicitly calls `Slice.__init__`, whereas `object.__new__(Slice)` doesn't.

In this case, the `Sagittal.__init__` and `Slice.__init__` no longer exist, so they have the same effect. But during the refactor, I hadn't gotten rid of `Slice.__init__` yet, so I had to split the object creation into `object.__new__` and `x.__init__`.

joshuacwnewton: Aha, gotcha!

Issue #4568, opened 2024-07-22
For `sct_analyze_lesion -f`, the atlas ROI spreadsheet columns are unsorted (e.g. 17, 21, 07, 11, etc.)


joshuacwnewton:
![image](https://github.com/user-attachments/assets/abcc3875-5e47-47a5-9653-25967b724403)

I suspect that this is due to the inherent randomness in loading images by globbing:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/625f62e449000f8822b3a59dd9b6bb8dd96f4a7d/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L572-L578

I bet that this is a one-line fix (adding a call to `sorted()` to the filename list).

Issue #4567, opened 2024-07-22
Add support for combined atlas labels when computing lesion distribution percentages


joshuacwnewton:
Currently, `sct_analyze_lesion` processes atlas image files as follows:

- It scans the atlas folder for `.nii.gz` files:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/625f62e449000f8822b3a59dd9b6bb8dd96f4a7d/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L572-L578
- Then, it loads all of the atlas data into a dictionary:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/625f62e449000f8822b3a59dd9b6bb8dd96f4a7d/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L453-L460
- Finally, there are specialized functions that compute percentages for each tract (`__relative_ROIvol_in_mask`, `_measure_eachLesion_distribution`)
- Because `.nii.gz` files are scanned directly, there is no processing of "meta" `CombinedLabels`, which are collections of labels defined in the atlas's `info_label.txt`.

----

By comparison, `sct_extract_metric` _does_ consider the `CombinedLabels`:

- First, it loads all of the individual atlas files together as a single `[N, M, L, 37]` array:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0481a037166257c15e5cdcbd43bfafde1784c9ea/spinalcordtoolbox/scripts/sct_extract_metric.py#L352-L357
- Then, it `np.sum`s the labels together for combined labels:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a7a3f8dd021b6b7f8d2b9fa2680002ed4b1c5f2c/spinalcordtoolbox/aggregate_slicewise.py#L448-L450

----

If we wish to adopt the same behavior for `sct_analyze_lesion`, we will need to implement:

- [ ] Parsing of the `info_label.txt` file. (This is currently not done for `sct_analyze_lesion`. As shown above, the atlas files are parsed by their presence on-disk.)
- [ ] Summing the percentage label values. 
- [ ] Adding the `CombinedLabels` into new columns in the sheet.

Note that there are some caveats to be aware of:

- Since there is overlap with some of the existing functionality from the `aggregate_slicewise.py` API, we should keep in mind https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3378.
[...]

PR #4584, opened 2024-07-30
Support `CombinedLabels` from `info_label.txt` when computing lesion distributions with `-f`

### LINKED ISSUE #2677: Output lesion statistics per slice

### LINKED ISSUE #4587: Improve performance of `sct_analyze_lesion -f` by avoiding redundant calls to `np.where`
This is an issue that was previously brought up in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4394#issuecomment-1986482032 and re-encountered during the implementation of:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2677

> By switching from per-level aggregation (usually ~10-13 vert levels) to per-slice aggregation (as many as 500 slices), the optimization issues get much worse. (As in, couldn't even finish processing worse.) So, to properly implement this feature, it would require some sort of optimization fix to improve the run time.

I've run some profiling code and determined that the slowness is almost entirely due to repeated calls to `np.where`:

image

I think it should be straightforward enough to de-duplicate these calls. I have some ideas about how to refactor the methods to avoid the slow-down. After this is finished, I can continue working to implement #2677.

joshuacwnewton:
## Description

This PR amends two methods responsible for computing the distribution of a lesion within a warped atlas (i.e. the `-f` option of `sct_analyze_lesion`):

- `_measure_eachLesion_distribution`: Responsible for the `lesion#Ndistribution` pages in the sheet. The values are "percentage of the lesion that exists in a given ROI". Since the CombinedLabels are just combined ROIs, we can simply sum the percentages.
- `_measure_totLesion_distribution`: Responsible for the `ROI_occupied_by_lesion` page in the sheet. The values are "percentage of the ROI taken up by _any_ lesion". This is a bit trickier, since the ROI area is in the denominator (rather than the numerator). Thankfully, it turns out that this functionality was already implemented! 
    - Before: Hardcoded "combined labels" that are identical to the entries in `info_label.txt`.
    - After: Read the "combined labels" directly from `info_label.txt`.

The end result is that the user can now modify `info_label.txt` to add their own custom labels, and they will get results for the pages generated by `-f`.

## Testing this PR

Testing this is a bit tricky, because it requires lengthy preprocessing steps: A) lesion segmentation B) registering the image to the template and C) warping the template (including the atlas).

In our `sct_testing_data` dataset, we do have a "mini" version of the template, but it does _not_ contain the atlas files. So, it's necessary to work with full-sized images (to access the existing atlas of the PAM50 template).

I was able to find a suitable subject in the SCI Zurich dataset for testing purposes. However, `sct_label_vertebrae` failed initially, so I need to manually identify the C2-C3 disc. In total, the full testing steps were:

```
sct_deepseg -task seg_sc_lesion_t2w_sci -i sub-zh01_ses-01_acq-sag_T2w.nii.gz 
sct_label_utils -i sub-zh01_ses-01_acq-sag_T2w.nii.gz -create-viewer 3
sct_label_vertebrae -i sub-zh01_ses-01_acq-sag_T2w.nii.gz -s sub-zh01_ses-01_acq-sag_T2w_sc_seg.nii.gz -initlabel labels.nii.gz -c t2
sct_label_utils -i sub-zh01_ses-01_acq-sag_T2w_sc_seg_labeled.nii.gz -vert-body 3,9
sct_register_to_template -i sub-zh01_ses-01_acq-sag_T2w.nii.gz -s sub-zh01_ses-01_acq-sag_T2w_sc_seg.nii.gz -l labels.nii.gz 
sct_warp_template -d sub-zh01_ses-01_acq-sag_T2w.nii.gz -w warp_template2anat.nii.gz
sct_analyze_lesion -m sub-zh01_ses-01_acq-sag_T2w_lesion_seg.nii.gz -s sub-zh01_ses-01_acq-sag_T2w_sc_seg.nii.gz -f label/
```

I would be happy to share the files privately on slack if you want to test (or you could try yourself on a different subject).

## Note about other features

I also took the time to solve a small issue re: column ordering. All it took was a single call to `sorted`.

I wanted to solve #2677 while I was at it, since this was also requested by the user. However, I ran into an issue that I previously observed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4394#issuecomment-1986482032:

> > [!NOTE]
> > As a side observation, I noticed that adding the `-f` option makes `sct_analyze_lesion` _very_ slow on my personal laptop, even for a single subject. There is no feedback in the output while the template-related processing is running, too. This isn't a great user experience, and I would love to look further into improving this after I debug the numerical issues.

By switching from per-level aggregation (usually ~10-13 vert levels) to per-slice aggregation (as many as 500 slices), the optimization issues get _much_ worse. (As in, couldn't even finish processing worse.) So, to properly implement this feature, it would require some sort of optimization fix to improve the run time. At that point, I figured it might be better to get this initial PR going, and follow-up with a second fix afterwards.

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4567
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4568
Fixes #4587 (When #4588 is merged)
Fixes #2677 (When #4588 is merged)

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -231,17 +231,11 @@ class AnalyzeLesion:
         with pd.ExcelWriter(self.excel_name, engine='xlsxwriter') as writer:
             self.measure_pd.to_excel(writer, sheet_name='measures', index=False, engine='xlsxwriter')
 
-            # Add the total column and row
+            # Save spreadsheet (for -f option)
             if self.path_template is not None:
                 for sheet_name in self.distrib_matrix_dct:
-                    if '#' in sheet_name:
-                        df = self.distrib_matrix_dct[sheet_name].copy()
-                        df = df.append(df.sum(numeric_only=True, axis=0), ignore_index=True)
-                        df['total % (all tracts)'] = df.sum(numeric_only=True, axis=1)
-                        df.iloc[-1, df.columns.get_loc('vert')] = 'total % (all vert)'
-                        df.to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
-                    else:
-                        self.distrib_matrix_dct[sheet_name].to_excel(writer, sheet_name=sheet_name, index=False, engine='xlsxwriter')
+                    self.distrib_matrix_dct[sheet_name].to_excel(writer, sheet_name=sheet_name, index=False,
+                                                                 engine='xlsxwriter')
 
             # Save pickle
             self.distrib_matrix_dct['measures'] = self.measure_pd
@@ -587,6 +581,13 @@ class AnalyzeLesion:
             for tract_id in atlas_data:
                 val = self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_' + str(tract_id).zfill(2)].values[0]
                 self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_' + str(tract_id).zfill(2)] = val * 100.0 / vol_mask_tot
+
+        # Add the total column and row
+        df = self.distrib_matrix_dct[sheet_name]
+        df = df.append(df.sum(numeric_only=True, axis=0), ignore_index=True)
+        df['total % (all tracts)'] = df.sum(numeric_only=True, axis=1)
+        df.iloc[-1, df.columns.get_loc('vert')] = 'total % (all vert)'
+        self.distrib_matrix_dct[sheet_name] = df
 
     def __regroup_per_tracts(self, vol_dct, tract_limit):
         res_mask = [vol_dct[t][0] for t in vol_dct if t >= tract_limit[0] and t <= tract_limit[1]]
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -588,6 +588,12 @@ class AnalyzeLesion:
         self.distrib_matrix_dct[sheet_name]['total % (all tracts)'] = \
             self.distrib_matrix_dct[sheet_name].sum(numeric_only=True, axis=1)
 
+        # Add additional columns for the "CombinedLabels" defined by info_label.txt
+        for label_name, sublabels in self.atlas_combinedlabels.items():
+            column_names_to_sum = [f"PAM50_{str(subid).zfill(2)}" for subid in sublabels]
+            self.distrib_matrix_dct[sheet_name][label_name] = \
+                self.distrib_matrix_dct[sheet_name][column_names_to_sum].sum(axis=1)
+
         # Add the total row
         self.distrib_matrix_dct[sheet_name] = self.distrib_matrix_dct[sheet_name].append(
             self.distrib_matrix_dct[sheet_name].sum(numeric_only=True, axis=0),
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -603,9 +603,9 @@ class AnalyzeLesion:
             -1, self.distrib_matrix_dct[sheet_name].columns.get_loc('vert')
         ] = 'total % (all vert)'
 
-    def __regroup_per_tracts(self, vol_dct, tract_limit):
-        res_mask = [vol_dct[t][0] for t in vol_dct if t >= tract_limit[0] and t <= tract_limit[1]]
-        res_tot = [vol_dct[t][1] for t in vol_dct if t >= tract_limit[0] and t <= tract_limit[1]]
+    def __regroup_per_tracts(self, vol_dct, tracts):
+        res_mask = [vol_dct[t][0] for t in vol_dct if t in tracts]
+        res_tot = [vol_dct[t][1] for t in vol_dct if t in tracts]
         return np.sum(res_mask) * 100.0 / np.sum(res_tot)
 
     def _measure_totLesion_distribution(self, im_lesion, atlas_data, im_vert, p_lst):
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -584,70 +595,91 @@ def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, im_t
...
-    def __regroup_per_tracts(self, vol_dct, tract_limit):
-        res_mask = [vol_dct[t][0] for t in vol_dct if t >= tract_limit[0] and t <= tract_limit[1]]
-        res_tot = [vol_dct[t][1] for t in vol_dct if t >= tract_limit[0] and t <= tract_limit[1]]
+        # Add the total column
+        self.distrib_matrix_dct[sheet_name]['total % (all tracts)'] = \
+            self.distrib_matrix_dct[sheet_name].sum(numeric_only=True, axis=1)
+
+        # Add additional columns for the "CombinedLabels" defined by info_label.txt
+        for label_name, sublabels in self.atlas_combinedlabels.items():
+            column_names_to_sum = [f"PAM50_{str(subid).zfill(2)}" for subid in sublabels]
```
mguaypaq, line 633: This format spec should be equivalent for numbers:
```suggestion
            column_names_to_sum = [f"PAM50_{subid:02}" for subid in sublabels]
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -584,70 +595,91 @@ def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, im_t
...
+            **self.rows,
+            # numpy array index equivalent to [:, :, :]
+            total_row: (slice(None), slice(None), slice(None)),
+        }
+        self.distrib_matrix_dct[sheet_name] = pd.DataFrame.from_dict({'row': [str(r) for r in rows_with_total]})
 
         # initialized to 0 for each vertebral level and each PAM50 tract
         for tract_id in atlas_data:
-            self.distrib_matrix_dct[sheet_name]['PAM50_' + str(tract_id).zfill(2)] = [0] * len(self.vert_lst + ['total % (all vert)'])
+            self.distrib_matrix_dct[sheet_name]['PAM50_' + str(tract_id).zfill(2)] = [0] * len(rows_with_total)
```
mguaypaq, line 664: To align with the code above:
```suggestion
            self.distrib_matrix_dct[sheet_name][f"PAM50_{tract_id:02}"] = [0] * len(rows_with_total)
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -584,70 +595,91 @@ def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, im_t
...
+                                                                                indices_to_keep=indices_to_keep)
 
-                # group tracts to compute involvement in GM, WM, DC, VF, LF
-                self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_GM'] = self.__regroup_per_tracts(vol_dct=res_perTract_dct, tract_limit=[30, 35])
-                self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_WM'] = self.__regroup_per_tracts(vol_dct=res_perTract_dct, tract_limit=[0, 29])
-                self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_DC'] = self.__regroup_per_tracts(vol_dct=res_perTract_dct, tract_limit=[0, 3])
-                self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_VF'] = self.__regroup_per_tracts(vol_dct=res_perTract_dct, tract_limit=[14, 29])
-                self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_LF'] = self.__regroup_per_tracts(vol_dct=res_perTract_dct, tract_limit=[4, 13])
+                # group tracts to compute involvement in CombinedLabels (GM, WM, DC, VF, LF)
+                for label_name, sublabels in self.atlas_combinedlabels.items():
```
mguaypaq, line 680: This will add a 6th one for the whole spinal cord, and change the label names, presumably that's ok?

From `data/PAM50/atlas/info_label.txt`:
```
# Combined labels
# Keyword=CombinedLabels (Please DO NOT change this line)
# ID, name, IDgroup
50, spinal cord, 0:35
51, white matter, 0:29
52, gray matter, 30:35
53, dorsal columns, 0:3
54, lateral funiculi, 4:13
55, ventral funiculi, 14:29
```

mguaypaq (approved):
Seems reasonable, thanks for the new feature!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: This is a really good point, and something I had been thinking about too, but for a different reason: The "whole spinal cord" is identical to the existing "total" column, meaning that the column is duplicated.

This could be a reason to ignore the spinal cord column, but it also feels like a good sanity check (if the columns don't match, something has gone wrong). It also kind of clarifies the meaning of the "total" column a bit: the "total" is equivalent to the lesion distribution across all tracts i.e. the full spinal cord.

joshuacwnewton: I think https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4584/commits/89881dccc0e91c11b3257d36904bb524fa80c847 might be a good compromise? In that, we exclude `spinal cord` but _only_ if it will match the total column.

If anything doesn't match (for example if the `info_label.txt` file is modified to include the CSF file for some reason), then the `spinal cord` label will be preserved.

mguaypaq: That's an elegant solution, I like it!

PR #4588, opened 2024-07-31
Add `-perslice` option (plus performance improvements to make it feasible)

### LINKED ISSUE #2677: Output lesion statistics per slice

### LINKED ISSUE #4567: Add support for combined atlas labels when computing lesion distribution percentages

joshuacwnewton:
## Description

This was a tricky refactor, because `-perslice` goes against the default behavior of `sct_analyze_lesion` (aggregating by vertebral levels), and there are a lot of hardcoded `vert` variable names. It took quite a few changes of (vert -> row) to avoid a complete rewrite.

That said, big-picture "aggregating by vertebral levels" basically boils down to this:

- Iterate over vertebral levels.
- For each vertebral level, set all other voxels equal to 0, then compute percentages.

The general structure of this code still works if you substitute "vertebral level" for slice; You can instead iterate over slices and still set all other voxels equal to 0. To do this, I used "rows" to represent either vertebral levels *or* slices, and kept a separate variable "row_name" to track whether we have vertlevels or slices. This mostly gets the job done.

Note: While the previous performance improvements help, this is still a slow process for any image with a large number of slices:

- `-perslice 0`, no optimization: ~300s (~5m)
- `-perslice 0`, with optimization: ~50s (~1m)
- `-perslice 1`, no optimization: Did not finish (est. 3600s, or 1hr)
- `-perslice 1`, with optimization: ~600s (10m)

So, perhaps there is more optimization work to be done. Here is the `tuna` profile for `-perslice 1` on the same image I used to test #4584:

image

It's a lot of calls to `np.sum`, `np.count_nonzero`, etc. I think this is because we've got naive `for` loops over both slices (~500) and tracts (~36). I'm sure there's some sort of smart numpy array operations we could be doing, but it's hard to implement that without a full rewrite.

It may also be worth just... skipping the slices that don't have lesion values? With large images, this results in a large output spreadsheet with many empty rows. But, if we exclude the empty rows, then perhaps we can save on processing time as well.

## Linked issues

Fixes #4567 
Fixes #2677

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -627,10 +627,11 @@ class AnalyzeLesion:
     def _measure_totLesion_distribution(self, im_lesion, atlas_data, im_vert, p_lst):
 
         sheet_name = 'ROI_occupied_by_lesion'
+        total_row = f'total % (all {self.row_name})'
         rows_with_total = {
             **self.rows,
             # numpy array index equivalent to [:, :, :]
-            f'total % (all {self.row_name})': (slice(None), slice(None), slice(None)),
+            total_row: (slice(None), slice(None), slice(None)),
         }
         self.distrib_matrix_dct[sheet_name] = pd.DataFrame.from_dict({'row': [str(r) for r in rows_with_total]})
 
@@ -642,7 +643,7 @@ class AnalyzeLesion:
         # loop over slices/vertlevels
         for row, indices_to_keep in sct_progress_bar(rows_with_total.items(), unit=self.row_name,
                                                      desc="  Computing ROI distribution for all lesions"):
-            if row.startswith('total') or np.count_nonzero(im_vert_and_lesion[indices_to_keep]):
+            if row == total_row or np.count_nonzero(im_vert_and_lesion[indices_to_keep]):
                 res_perTract_dct = {}  # for each tract compute the volume occupied by lesion and the volume of the tract
                 idx = self.distrib_matrix_dct[sheet_name][self.distrib_matrix_dct[sheet_name].row == str(row)].index
                 for tract_id in atlas_data:  # loop over the tracts
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -548,8 +548,7 @@ class AnalyzeLesion:
             image_out[:, :, indices] = image[:, :, indices]
         return image_out
 
-    def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, im_template_vert_data,
-                                  indices_to_keep):
+    def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, indices_to_keep):
         #
         #   Goal:
         #         This function computes the percentage of ROI occupied by binary mask
@@ -560,11 +559,11 @@ class AnalyzeLesion:
         #   Inputs:
         #           - im_mask_data - type=NumPyArray - binary mask (eg lesions)
         #           - im_atlas_roi_data - type=NumPyArray - ROI in the same space as im_mask
-        #                        - p_lst - type=list of float
-        #           - im_template_vert_data - type=NumPyArray - vertebral template in the same space as im_mask
-        #           - vert_level - type=int - vertebral level ID to restrict the ROI
+        #           - p_lst - type=list of float
+        #           - indices_to_keep - type=(anything that can be used to index numpy arrays)
+        #                               anything outside this mask will be set to 0
         #
-        if im_template_vert_data is not None and indices_to_keep:
+        if indices_to_keep:
             im_atlas_roi_data = self.__keep_only_indices(im_atlas_roi_data, indices_to_keep)
             im_mask_data = self.__keep_only_indices(im_mask_data, indices_to_keep)
 
@@ -594,7 +593,6 @@ class AnalyzeLesion:
                     res_lst = self.__relative_ROIvol_in_mask(im_mask_data=np.copy(im_lesion),
                                                              im_atlas_roi_data=np.copy(atlas_data[tract_id]),
                                                              p_lst=p_lst,
-                                                             im_template_vert_data=np.copy(im_vert_cur),
                                                              indices_to_keep=indices_to_keep)
                     self.distrib_matrix_dct[sheet_name].loc[idx, 'PAM50_' + str(tract_id).zfill(2)] = res_lst[0]
                     vol_mask_tot += res_lst[0]
@@ -656,7 +654,6 @@ class AnalyzeLesion:
                     res_perTract_dct[tract_id] = self.__relative_ROIvol_in_mask(im_mask_data=np.copy(im_lesion),
                                                                                 im_atlas_roi_data=np.copy(atlas_data[tract_id]),
                                                                                 p_lst=p_lst,
-                                                                                im_template_vert_data=np.copy(im_vert_cur),
                                                                                 indices_to_keep=indices_to_keep)
 
                 # group tracts to compute involvement in CombinedLabels (GM, WM, DC, VF, LF)
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -187,7 +187,7 @@ class AnalyzeLesion:
             self.path_levels = os.path.join(self.path_template, "template", "PAM50_levels.nii.gz")
         else:
             self.path_atlas, self.path_levels = None, None
-        self.rows = None
+        self.rows = {}
         self.row_name = "slice" if perslice else "vert"
         self.atlas_roi_lst = None
         self.atlas_combinedlabels = {}
@@ -542,10 +542,7 @@ class AnalyzeLesion:
     def __keep_only_indices(self, image, indices):
         """Keep values defined by indices, and set all other coordinates to zero."""
         image_out = np.zeros_like(image)
-        if isinstance(indices, list) and len(indices) == 3:  # Lists of [x], [y], and [z] coords produced by np.where()
-            image_out[indices] = image[indices]
-        else:
-            image_out[:, :, indices] = image[:, :, indices]
+        image_out[indices] = image[indices]
         return image_out
 
     def __relative_ROIvol_in_mask(self, im_mask_data, im_atlas_roi_data, p_lst, indices_to_keep):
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -531,11 +542,14 @@ def ___pve_weighted_avg(self, im_mask_data, im_atlas_data):
     def __keep_only_indices(self, image, indices):
         """Keep values defined by indices, and set all other coordinates to zero."""
         image_out = np.zeros_like(image)
-        image_out[indices] = image[indices]
+        if isinstance(indices, list) and len(indices) == 3:  # Lists of [x], [y], and [z] coords produced by np.where()
```
mguaypaq, line 545: But `np.where` returns a `tuple`, not a `list`?

I'm concerned that this might silently return the wrong result, because `image[indices]` and `image[:, :, indices]` both "work", but with extremely different results:
```py3
>>> data = np.arange(24).reshape((2, 3, 4)); data
array([[[ 0,  1,  2,  3],
        [ 4,  5,  6,  7],
        [ 8,  9, 10, 11]],

       [[12, 13, 14, 15],
        [16, 17, 18, 19],
        [20, 21, 22, 23]]])

>>> data[np.where(data < 5)]
array([0, 1, 2, 3, 4])

>>> data[:, :, np.where(data < 5)]
array([[[[ 0,  0,  0,  0,  0],
         [ 0,  0,  0,  0,  1],
         [ 0,  1,  2,  3,  0]],

        [[ 4,  4,  4,  4,  4],
         [ 4,  4,  4,  4,  5],
         [ 4,  5,  6,  7,  4]],

        [[ 8,  8,  8,  8,  8],
         [ 8,  8,  8,  8,  9],
         [ 8,  9, 10, 11,  8]]],


       [[[12, 12, 12, 12, 12],
         [12, 12, 12, 12, 13],
         [12, 13, 14, 15, 12]],

        [[16, 16, 16, 16, 16],
         [16, 16, 16, 16, 17],
         [16, 17, 18, 19, 16]],

        [[20, 20, 20, 20, 20],
         [20, 20, 20, 20, 21],
         [20, 21, 22, 23, 20]]]])
```

And yes, the type is wrong:
```py3
>>> isinstance(np.where(data < 5), list)
False
>>> isinstance(np.where(data < 5), tuple)
True
```

mguaypaq (changes requested):
The optimization is very significant, and yes reducing calls to `np.where` (which iterates over the whole array each time) is very good.

I'm concerned that the numpy array indexing is a bit wrong though. I'll propose a fix as an extra commit.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Re: list vs. tuple, thank you for catching this :hearts:

Re: `np.where` vs. `[:, :, slice]`... I was under the impression that, we're only using the indexing/slicing as a roundabout way to zero out all other voxels, the shape of the the intermediate array (1D vs. 3D) doesn't matter? But maybe I'm wrong here... :thinking:

joshuacwnewton: I guess it would be a lot clearer if `indices_to_keep` had a single form (e.g. tuple of x,y,z coordinates) rather than multiple forms (tuple *or* slice number). So, I can definitely refactor things to avoid this inconsistency.

mguaypaq: In the commits I added, I made it so that `indices_to_keep` is always something that can be used directly to index a numpy array. Hopefully this works? Let me know what you think.

mguaypaq: > Re: np.where vs. [:, :, slice]... I was under the impression that, because we're only using the indexing/slicing as a roundabout way to zero out all other voxels, the shape of the the intermediate array (1D vs. 3D) doesn't matter? But maybe I'm wrong here... 🤔

In my example above with `np.arange`, I think it would actually make a difference. In particular:
* Most of them would redundantly get set to zero several times (not a real problem).
* All the entries would get set to zero, not just the first five (yes a problem).

```py3
>>> import numpy as np
>>> data = np.arange(24).reshape((2, 3, 4)); data
array([[[ 0,  1,  2,  3],
        [ 4,  5,  6,  7],
        [ 8,  9, 10, 11]],

       [[12, 13, 14, 15],
        [16, 17, 18, 19],
        [20, 21, 22, 23]]])

>>> data[np.where(data < 5)] = 0; data
array([[[ 0,  0,  0,  0],
        [ 0,  5,  6,  7],
        [ 8,  9, 10, 11]],

       [[12, 13, 14, 15],
        [16, 17, 18, 19],
        [20, 21, 22, 23]]])

>>> data[:, :, np.where(data < 5)] = 0; data
array([[[0, 0, 0, 0],
        [0, 0, 0, 0],
        [0, 0, 0, 0]],

       [[0, 0, 0, 0],
        [0, 0, 0, 0],
        [0, 0, 0, 0]]])
```

joshuacwnewton: I very much like the new structure you've added with `slice(None)` representing `:`, and I have no complaints! :tada:

(As an aside, I'm not 100% sure the `[:, :, np.where(data < 5)]` example given is 1:1 with the old code (since before, it was more like `[:, :, n_slice]`), but that's kind of moot now as you've undoubtedly improved `indices_to_keep`, so I have nothing much to add, hehe.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -611,27 +627,29 @@ def __regroup_per_tracts(self, vol_dct, tracts):
...
-                im_vert_cur = np.copy(im_vert)
-                im_vert_cur[np.where(im_vert_cur != vert)] = 0
-            else:
-                im_vert_cur = None
-            if im_vert_cur is None or np.count_nonzero(im_vert_cur * np.copy(im_lesion)):
+        im_vert_and_lesion = im_vert * im_lesion  # to check which vertebral levels have lesions
+        # loop over slices/vertlevels
+        for row, indices_to_keep in sct_progress_bar(rows_with_total.items(), unit=self.row_name,
+                                                     desc="  Computing ROI distribution for all lesions"):
+            if row.startswith('total') or np.count_nonzero(im_vert_and_lesion[indices_to_keep]):
```
joshuacwnewton, line 645: I think since `row` is an integer otherwise (slice/vert number), this needs to be `str(row)`?

joshuacwnewton (commented):
I reran `-perslice 1` and ran into a little bug, but after fixing it, the same identical spreadsheet was produced. :tada:

spinalcordtoolbox/scripts/sct_analyze_lesion.py
mguaypaq: Fixed, thanks for catching this! I changed it back to an equality test, because that really captures the intent better.

PR #4596, opened 2024-08-06
Silence `pandas` warning by explicitly using `copy` to remove view/copy ambiguity

### LINKED ISSUE #4591: `sct_analyze_lesion` -  pandas warnings are mixing with the function output
### Description

When running `sct_analyze_lesion` on `sub-zh102` from `sci-zurich` using:

```console
sct_analyze_lesion -m derivatives/labels/sub-zh102/ses-01/anat/sub-zh102_ses-01_acq-sag_T2w_lesion-manual.nii.gz -s derivatives/labels/sub-zh102/ses-01/anat/sub-zh102_ses-01_acq-sag_T2w_seg-manual.nii.gz -qc qc
```

The CLI output contains `SettingWithCopyWarning` warnings, making the output difficult to follow:

image

joshuacwnewton:
## Description

As per https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4591#issuecomment-2271973627:

> But we are not explicit about whether we want a "view" (i.e. changes the original dataframe) or a "copy" (independent of the original dataframe). Because we are not explicit, there is potential ambiguity, hence why pandas throws this warning. To fix this warning, we just need to be explicit about what we want, and there are multiple ways to do this (not just `.loc`).

I think in this case, because we only ever access the newly-added columns from df_temp and not from the original `tissue_bridges_df` DataFrame, we don't need a "view", and so I think it should be fine to denote df_temp as a copy:

## Linked issues

Fixes #4591.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -468,7 +468,7 @@ class AnalyzeLesion:
         # distance from cerebrospinal fluid to the lesion boundary
         for sagittal_slice in sagittal_lesion_slices:
             # Get df for the selected sagittal slice
-            df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice]
+            df_temp = tissue_bridges_df[tissue_bridges_df['sagittal_slice'] == sagittal_slice].copy()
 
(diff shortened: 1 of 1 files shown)


valosekj (approved):
Thank you for the fix! Tested and works fine (`SettingWithCopyWarning` are no longer printed to CLI)!

PR #4598, opened 2024-08-08
Convert `install_sct` into `install_ads` as proof of concept for AxonDeepSeg install script

joshuacwnewton:
WIP, will update with proper description once I discuss further with Mathieu Boudreau

only changes on ADS end: creating [`requirements.txt`](https://github.com/axondeepseg/axondeepseg/commit/6c5220fe9a61f4325f3e662985ebede92ee06593) for pip-only dependency install


joshuacwnewton:
Moving future development to ADS repo. (I only used SCT repo because we had everything set up already.)

PR #4600, opened 2024-08-08
Add `--use-pep517` to install scripts to silence warning

### LINKED ISSUE #4599: `DEPRECATION: Legacy editable install` during `pip install -e .` step of `install_sct`
The full error message is:

```
Installing collected packages: spinalcordtoolbox
  DEPRECATION: Legacy editable install of spinalcordtoolbox==6.4.dev0 from file:///home/runner/sct_6.4.dev0 (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for spinalcordtoolbox
Successfully installed spinalcordtoolbox
```

This error has made me realize that, while we pin our _pip dependencies_ via `pip freeze`, we don't pin the dependencies that conda provides (e.g. `pip` itself), meaning that all past SCT versions may be susceptible to breakage once `pip==25.0` is released. (It's not clear to me if pip 25.0 will fail, or if it will automatically use the `--use-pep517` option.)

Anyway, testing `--use-pep517` locally shows no issue:

```
(venv_sct) jn/ads-install-script|joshua@monarch:~/repos/spinalcordtoolbox$ pip install -e . --use-pep517
Obtaining file:///home/joshua/repos/spinalcordtoolbox
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Building wheels for collected packages: spinalcordtoolbox
  Building editable for spinalcordtoolbox (pyproject.toml) ... done
  Created wheel for spinalcordtoolbox: filename=spinalcordtoolbox-6.4.dev0-0.editable-py3-none-any.whl size=9784 sha256=a395c601e67e1ff13fb3db19825d68c81503207613e6e967a4fcbb20ad0e87b8
  Stored in directory: /tmp/pip-ephem-wheel-cache-k5304kt1/wheels/12/d4/75/b75a0b0fa690f5d1477b7d2b91f862af1d4ece58ca4c5ba5a5
Successfully built spinalcordtoolbox
Installing collected packages: spinalcordtoolbox
  Attempting uninstall: spinalcordtoolbox
    Found existing installation: spinalcordtoolbox 6.4.dev0
    Uninstalling spinalcordtoolbox-6.4.dev0:
      Successfully uninstalled spinalcordtoolbox-6.4.dev0
Successfully installed spinalcordtoolbox-6.4.dev0
```

So, it should at the very least be an easy fix.

joshuacwnewton:
## Description

Pip has started throwing a deprecation warning for `setup.py develop` behavior that gets invoked when we run `pip install -e .`. From what I can tell, though, the upcoming change in `pip==25.0` is mainly a backend change that will be largely invisible.

Since the core Python part (`spinalcordtoolbox`) is a very straightforward package, with no C++ extensions or other complicated build requirements, I think that the switch from `setup.py develop` to the new wheel-building behavior should have no effect. And in fact, it's probably fine to leave `pip install -e .` as-is and let the upgrade to `pip==25.0` happen silently in the background (meaning that I don't think this PR is urgent?). But, I'm seeking clarity on this here: https://github.com/pypa/pip/issues/11457#issuecomment-2276205988

But, I'm manually invoking here to A) silence the warnings and B) test to see if it has any effect. In case `pip==25.0` causes an error, we'll need to backport this change, but I would be very surprised if `pip` did something so destructive, so I think we should be fine.

## Linked issues

Fixes #4599.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -664,7 +664,7 @@ fi
   # https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4352
   SYSTEM_VERSION_COMPAT=0 python/envs/venv_sct/bin/pip install -r "$REQUIREMENTS_FILE" --ignore-installed certifi &&
   print info "Installing spinalcordtoolbox..." &&
-  python/envs/venv_sct/bin/pip install -e .
+  python/envs/venv_sct/bin/pip install -e . --use-pep517
 ) ||
   die "Failed running pip install: $?"
 
--- a/install_sct.bat
+++ b/install_sct.bat
@@ -182,7 +182,7 @@ echo ### Installing SCT and its dependencies from %requirements_file%...
 rem Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
 python\envs\venv_sct\python -m pip install -U "pip^!=21.2.*" || goto error
 python\envs\venv_sct\Scripts\pip install -r %requirements_file% || goto error
-python\envs\venv_sct\Scripts\pip install -e . || goto error
+python\envs\venv_sct\Scripts\pip install -e . --use-pep517 || goto error
 
 rem Install external dependencies
 echo:


mguaypaq (approved):
It would take a very long time for me to get up to speed on the current state of python packaging, so I'll just approve this to unblock the 6.4 release.

PR #4602, opened 2024-08-08
Add `-m` and `-qc` options to fMRI TSNR script

### LINKED ISSUE #4290: Add TSNR maps after motion correction for fMRI in qc report
## Description

Typically,  after motion correction of fMRI timeseries, we compare the TSNR maps to ensure that the TSNR is not lower than before motion correction. We could add this feature in the `sct_qc`.

An idea would be to overlay the axial TSNR map before motion correction and after:

* the color map goes from blue --> orange

anim
 

Also tagging @dpfyffer

sandrinebedard:
#### GitHub

#### PR contents

## Description
This PR adds a qc for `sct_compute_tsnr `and adds option to compute tsnr in a mask per slice.

How to test:
With sct_fmri_compute_tsnr:
~~~
sct_fmri_compute_tsnr -i fmri.nii.gz -m fmri_mean_seg.nii.gz -qc .qc 
~~~

With sct_qc: 
~~~
# Compare before and after motion correction
sct_qc -i fmri_tsnr.nii.gz -d fmri_moco_tsnr.nii.gz -s fmri_mean_seg.nii.gz -p sct_fmri_compute_tsnr -qc .qc
~~~

Data: 
[fmri_data.zip](https://github.com/user-attachments/files/16591279/fmri_data.zip)

## Linked issues
Fixes #4290


joshuacwnewton (changes requested):
My first round of reviews is mostly just focused on de-duplicating repeated functionality, and keeping the contribution simple/straightforward.

spinalcordtoolbox/reports/qc.py
```diff
@@ -82,6 +82,20 @@ def no_seg_seg(self, mask, ax):
...
+        norm = mpl_colors.Normalize(vmin=vmin, vmax=vmax)
+        ax = fig.add_axes((0, 0, 0.93, 1))
+        fig_ax = ax.imshow(mask, cmap=cmap, norm=norm, interpolation=self.interpolation, aspect=self.aspect_mask)
+        cax = ax.inset_axes([1.005, 0.07, 0.011, 0.86])
+        cbar = fig.colorbar(fig_ax, cax=cax, orientation='vertical', pad=0.01, shrink=0.5, aspect=1, ticks=[vmin, vmax])
+        cbar.ax.tick_params(labelsize=5, length=2, pad=1.7)
+        ax.text(1.5, 6, '2', color='white', size=3.25)
+        self._add_orientation_label(ax)
+        ax.get_xaxis().set_visible(False)
+        ax.get_yaxis().set_visible(False)
```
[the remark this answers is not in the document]
sandrinebedard: I agree we can add a new action list for the colorbar! I had a bit of trouble trying to place the code! Thanks for all the help!

spinalcordtoolbox/scripts/sct_fmri_compute_tsnr.py
```diff
@@ -22,19 +24,35 @@ def __init__(self):
...
+    def orient2rpi(self):
+        # save input image orientation
+        if self.mask is not None:
+            self.orientation_mask = Image(self.mask).orientation
+            if not self.orientation_mask == 'RPI':
+                printv('\nOrient input image(s) to RPI orientation...', self.verbose, 'normal')
+                self._orient(self.mask, 'RPI')
+        self.orientation_fmri = Image(self.fmri).orientation
+        if not self.orientation_fmri == 'RPI':
+            self._orient(self.fmri, 'RPI')
```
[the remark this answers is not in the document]
sandrinebedard: Agreed!

spinalcordtoolbox/reports/qc.py
```diff
@@ -223,11 +229,26 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
...
+            ax = fig.add_axes((0, 0, 0.93, 1))
+            cmap = 'seismic'
+            vmin = min(int(np.amin(img)), int(np.amin(mask)))
+            vmax = max(int(np.amax(img)), int(np.amax(mask[0]))) - 2
+            norm = mpl_colors.Normalize(vmin=vmin, vmax=vmax)
+        else:
+            ax = fig.add_axes((0, 0, 1, 1))
+            cmap = 'gray'
+            norm = None
+            colorbar = False
```
joshuacwnewton, line 243: I think we might be able to simplify these default values a little more by instead making them the default values inside of `no_seg_seg`:

```python
    def no_seg_seg(self, mask, ax, cmap='gray', norm=None, fig=None, colorbar=False):
```

By using keywords, we can then construct a dictionary for the keyword arguments:

```python
kwargs = {}
if self.process == 'sct_fmri_compute_tsnr': 
    kwargs['cmap'] = 'seismic'
    kwargs['colorbar'] = True
    kwargs['fig'] = fig
    vmin = min(int(np.amin(img)), int(np.amin(mask)))
    vmax = max(int(np.amax(img)), int(np.amax(mask[0]))) - 2
    kwargs['norm'] = mpl_colors.Normalize(vmin=vmin, vmax=vmax)
```

Which lets us use a single `action()` function signature as before, by using [`**kwargs`](https://www.digitalocean.com/community/tutorials/how-to-use-args-and-kwargs-in-python-3#using-args-and-kwargs-in-function-calls) expansion to fill in the remaining (optional) parameters:

```python
   # here
   QcImage.no_seg_seg(self, mask, ax, **kwargs)
   
   # later on
   action(self, mask[i], ax, **kwargs)
```

(Saving us the need for an additional `if/else` block when calling `action`.)

spinalcordtoolbox/reports/qc.py
```diff
@@ -75,9 +75,14 @@ def template(self, mask, ax):
...
         ax.get_yaxis().set_visible(False)
 
-    def no_seg_seg(self, mask, ax):
+    def no_seg_seg(self, mask, ax, cmap, norm, fig=None, colorbar=False):
         """Create figure with image overlay. Notably used by sct_registration_to_template"""
-        ax.imshow(mask, cmap='gray', interpolation=self.interpolation, aspect=self.aspect_mask)
+        fig_ax = ax.imshow(mask, cmap=cmap, norm=norm, interpolation=self.interpolation, aspect=self.aspect_mask)
+        if colorbar:
+            cax = ax.inset_axes([1.005, 0.07, 0.011, 0.86])
+            cbar = fig.colorbar(fig_ax, cax=cax, orientation='vertical', pad=0.01, shrink=0.5, aspect=1, ticks=[norm.vmin, norm.vmax])
```
joshuacwnewton, line 83: Actually, wait, is `fig` even necessary to make a colorbar?

Since we're already passing `fig_ax` and `cax`, we should already be able to identify the correct axes, right? So, would it be possible to just call [`pyplot.colorbar`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.colorbar.html) directly? If this is possible, then that would reduce the number of parameters needed for `no_seg_seg`...

joshuacwnewton (changes requested):
Thank you for promptly responding to feedback! :hearts:

My changes here are just some more suggestions for simplifying/deduplicating the code. :relaxed:

spinalcordtoolbox/reports/qc.py
joshuacwnewton: Ah! Also, reading through the details of `pyplot.colorbar`, I notice it mentions:

> cax: Axes, optional
> Axes into which the colorbar will be drawn. If [None](https://docs.python.org/3/library/constants.html#None), then a new Axes is created and the space for it will be stolen from the Axes(s) specified in ax.

I'm curious... if the "space for it will be stolen", then does that mean we _don't_ need to manually set dimensions for the axes? (If so, that would simplify things even further!)

sandrinebedard: > Ah! Also, reading through the details of `pyplot.colorbar`, I notice it mentions:
> 
> > cax: Axes, optional
> > Axes into which the colorbar will be drawn. If [None](https://docs.python.org/3/library/constants.html#None), then a new Axes is created and the space for it will be stolen from the Axes(s) specified in ax.
> 
> I'm curious... if the "space for it will be stolen", then does that mean we _don't_ need to manually set dimensions for the axes? (If so, that would simplify things even further!)

This is what I did before, but it made the images reaaly small, so that is why I wanted to specify the size of th axis instead!

sandrinebedard: it works when I call matplolib_pypplot, but I get this warining: 
```
/home/sabeda/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py:83: UserWarning: Adding colorbar to a different Figure <Figure size 500x25 with 1 Axes> than <Figure size 640x480 with 0 Axes> which fig.colorbar is called on.
  cbar = mpl_plt.colorbar(fig_ax, cax=cax, orientation='vertical', pad=0.01, shrink=0.5, aspect=1, ticks=[norm.vmin, norm.vmax])
```

spinalcordtoolbox/reports/qc.py
sandrinebedard: Thanks for the suggetsion! Now done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/6fd1a2f9d231394d4611533cfd8f4d49a721e461

spinalcordtoolbox/reports/qc.py
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/7ed632d20bee49cade8dd9b0b8fb161cfea951c5

spinalcordtoolbox/reports/qc.py
```diff
@@ -223,11 +229,26 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
...
-        ax.get_yaxis().set_visible(False)
+        if self.process == 'sct_fmri_compute_tsnr':
+            colorbar = True
+            ax = fig.add_axes((0, 0, 0.93, 1))
+            cmap = 'seismic'
+            vmin = min(int(np.amin(img)), int(np.amin(mask)))
+            vmax = max(int(np.amax(img)), int(np.amax(mask[0]))) - 2
+            norm = mpl_colors.Normalize(vmin=vmin, vmax=vmax)
+        else:
+            ax = fig.add_axes((0, 0, 1, 1))
```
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/0f05813f2fce7b1d356b6799d715337cb7040aad

spinalcordtoolbox/reports/qc.py
```diff
@@ -236,11 +257,22 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
         mpl_backend_agg.FigureCanvasAgg(fig)
         for i, action in enumerate(self.action_list):
             logger.debug('Action List %s', action.__name__)
-            if self._stretch_contrast and action.__name__ in ("no_seg_seg",):
+            if self._stretch_contrast and action.__name__ in ("no_seg_seg") and not colorbar:
```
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/0f05813f2fce7b1d356b6799d715337cb7040aad

spinalcordtoolbox/reports/qc.py
```diff
@@ -202,7 +207,8 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
         """
 
         if self._stretch_contrast:
-            img = self._func_stretch_contrast(img)
+            if self.process != 'sct_fmri_compute_tsnr':
+                img = self._func_stretch_contrast(img)
```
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/0f05813f2fce7b1d356b6799d715337cb7040aad

sandrinebedard:
Thank you @joshuacwnewton for your suggetsions!! I had a bit of trouble organizing this code !I think I addressed all your comments

spinalcordtoolbox/reports/qc.py
joshuacwnewton: > it works when I call matplolib_pypplot, but I get this warining:

I'm able to reproduce the warning. I've figured out the reason why this warning is thrown, too:

- My suggestion above (creating a colorbar using `pyplot`) requires that our figures be managed by `pyplot`, so that `pyplot` can create a colorbar on the correct figure.
- However, currently `qc.py` is written in the "[explicit (object oriented) style](https://matplotlib.org/stable/users/explain/quick_start.html#the-explicit-and-the-implicit-interfaces)". We create new figures via calls to the class `matplotlib.figure.Figure()`, which _doesn't_ register the figure with `pyplot`.
- To demonstrate further, if we try to manually [set this figure as the current figure](https://stackoverflow.com/a/7987462) with `pyplot`, matplotlib will throw an error: "`ValueError: The passed figure is not managed by pyplot`"
- To fix this, if we call `matplotlib.pyplot.figure()` instead of `figure.Figure()`, the figure will be managed by pyplot and set as the "current figure", and thus `pyplot.colorbar` will be correctly associated with the figure, and no warning will be thrown.

I was initially worried about the risks of mixing "OO-style" matplotlib code with implicit, pyplot-managed code. However, the good news is that even in `matplotlib`'s own example code, they still use the "implicit" `matplotlib.pyplot.figure()` to create the figure.

> So one can use the OO-style
>
> ```python
> # Note that even in the OO-style, we use `.pyplot.figure` to create the Figure.
> fig, ax = plt.subplots(figsize=(5, 2.7), layout='constrained')
> ```

So, I think that means we should use pyplot-managed figures, even if we continue to use the explicit style in most of our code.

((Tagging @mguaypaq as I think "implicit vs explicit" is something we should be aware of when we do our QC refactors. I wasn't really familiar with the nuances here until I did this little deep dive.))

joshuacwnewton (changes requested):
I wrote a test for `-m` and `-qc`. Here's what the QC looks like for our testing data:

image

In the process of writing the test, I noticed a few more possible changes. However, most of the PR LGTM, so this is probably going to be my last round of changes before approval. :)

spinalcordtoolbox/scripts/sct_fmri_compute_tsnr.py
```diff
@@ -102,17 +143,31 @@ def main(argv: Sequence[str]):
...
-    tsnr = Tsnr(param=param, fmri=fname_src, out=fname_dst)
+    tsnr = Tsnr(param=param, fmri=fname_src, mask=fname_mask, out=fname_dst, verbose=verbose)
     tsnr.compute()
 
     display_viewer_syntax([fname_dst], verbose=verbose)
 
+    if path_qc is not None:
+        generate_qc(fname_dst, fname_in2=fname_dst, fname_seg=fname_mask, args=argv, path_qc=os.path.abspath(path_qc),
+                    dataset=qc_dataset, subject=qc_subject, process='sct_fmri_compute_tsnr')
+
```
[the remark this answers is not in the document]
sandrinebedard: I think we have the same requirement for sct_fmri_moco, it would make sense to have the same thing!

spinalcordtoolbox/reports/qc.py
```diff
@@ -75,9 +75,14 @@ def template(self, mask, ax):
...
-    def no_seg_seg(self, mask, ax):
+    def no_seg_seg(self, mask, ax, cmap='gray', norm=None, colorbar=False):
         """Create figure with image overlay. Notably used by sct_registration_to_template"""
-        ax.imshow(mask, cmap='gray', interpolation=self.interpolation, aspect=self.aspect_mask)
+        fig_ax = ax.imshow(mask, cmap=cmap, norm=norm, interpolation=self.interpolation, aspect=self.aspect_mask)
+        if colorbar:
+            cax = ax.inset_axes([1.005, 0.07, 0.011, 0.86])
+            cbar = mpl_plt.colorbar(fig_ax, cax=cax, orientation='vertical', pad=0.01, shrink=0.5, aspect=1, ticks=[norm.vmin, norm.vmax])
+            cbar.ax.tick_params(labelsize=5, length=2, pad=1.7)
+            ax.text(1.5, 6, '2', color='white', size=3.25)
```
sandrinebedard, line 85: TODO: change the number 2 as an argument to have 1 and 2

sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/106bff867a6b4c145a6c51c8b0be2f5c9abecadb

spinalcordtoolbox/reports/qc.py
```diff
@@ -236,11 +249,11 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
...
         for i, action in enumerate(self.action_list):
             logger.debug('Action List %s', action.__name__)
-            if self._stretch_contrast and action.__name__ in ("no_seg_seg",):
+            if self._stretch_contrast and action.__name__ in ("no_seg_seg"):
                 logger.debug("Mask type %s" % mask[i].dtype)
                 mask[i] = self._func_stretch_contrast(mask[i])
-            ax = fig.add_axes((0, 0, 1, 1), label=str(i))
-            action(self, mask[i], ax)
+            ax = fig.add_axes(ax_dim)
+            action(self, mask[i], ax, **kwargs)
```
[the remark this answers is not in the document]
sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/129358c11e151854c7b25ca5742e50e34d461b05

spinalcordtoolbox/scripts/sct_fmri_compute_tsnr.py
```diff
@@ -45,11 +51,28 @@ def compute(self):
...
+        if self.mask is not None:
+            mask = Image(self.mask)
+            orientation_mask = mask.orientation
+            if not orientation_mask == 'RPI':
+                mask = mask.change_orientation('RPI')
+            data_tsnr_masked = data_tsnr * mask.data
+            for z in range(data_tsnr_masked.shape[-1]):
+                # Display result
+                tsnr_roi = (data_tsnr_masked[:, :, z])[data_tsnr_masked[:, :, z] != 0].mean()
+                printv(f'\nSlice {z},  tSNR = {tsnr_roi:.2f}', type='info')  # TODO: display slice number in original space
```
sandrinebedard, line 66: @joshuacwnewton maybe we should address this todo before merging?

spinalcordtoolbox/reports/qc.py
joshuacwnewton: Since I updated the code for this PR, I think I'm going to mark this as resolved, but create a new issue to address the remaining examples.

spinalcordtoolbox/scripts/sct_fmri_compute_tsnr.py
joshuacwnewton: Yes, definitely! It should be fairly easy to do, too, since we have that new function written by @valosekj that converts RPI slice numbers back to original slice numbers.

sandrinebedard: @valosekj would you mind pointing me to that function?

sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4602/commits/8a248040776a169db9b653b5ec81301a4dbbfe36

joshuacwnewton (approved):

PR #4603, opened 2024-08-14
Properly maintain verbose status for internal ANTs `run_proc` call (to silence logging for `-v 0`)

### LINKED ISSUE #4581: Verbose CLI arg (`sct_apply_transfo -v`) is not passed to the subprocess call to `antsApplyTransforms`
My suggestion would be to _also_ change the verbose logic of the "Apply moco" section, and when verbose=1 is specified, silence the call to `isct_antsApplyTransforms`:

~~~
-------------------------------------------------------------------------------
  Apply moco
-------------------------------------------------------------------------------

Input parameters:
  Input file ............ data.nii
  Reference file ........ datasub_0_mean.nii.gz
  Polynomial degree ..... 0
  Smoothing kernel ...... 0
  Gradient step ......... 1
  Metric ................ MeanSquares
  Sampling .............. None
  Todo .................. apply
  Mask  ................. mask.nii
  Output mat folder ..... mat_final

Data dimensions:
  96 x 50 x 25 x 60

Copy file_target to a temporary file...

Register. Loop across Z (note: there is only one Z if orientation is axial)
Z=0/0:   0%|                                           | 0/60 [00:00<?, ?iter/s]/Users/julien/code/spinalcordtoolbox/bin/isct_antsApplyTransforms -d 3 -i /private/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-07-29_15-46-43_moco-wrapper_v0_qq83w/data_T0000.nii -o /private/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-07-29_15-46-43_moco-wrapper_v0_qq83w/data_T0000_moco.nii -t mat_final/mat.Z0000T0000Warp.nii.gz -r datasub_0_mean.nii.gz -n Linear # in /private/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-07-29_15-46-43_moco-wrapper_v0_qq83w
Z=0/0:   2%|▌                                  | 1/60 [00:00<00:15,  3.78iter/s]/Users/julien/code/spinalcordtoolbox/bin/isct_antsApplyTransforms -d 3 -i /private/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/s
~~~

Interestingly the output should already be silenced:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/682a6dcfa72010406f8f1a2cf8485f3d758eadb6/spinalcordtoolbox/moco.py#L755-L761

I see, the problem is that verbose is never passed to `run_proc()` within the class `Transform`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/682a6dcfa72010406f8f1a2cf8485f3d758eadb6/spinalcordtoolbox/scripts/sct_apply_transfo.py#L209-L214

This is out of scope (large repercussion if I change the logic here) so I will open another issue.

_Originally posted by @jcohenadad in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4580#issuecomment-2256791486_

joshuacwnewton:
## Description

In SCT CLI scripts, it is very common to want to apply a warping field to an image mid-processing (e.g. during registration, motion correction, etc.). SCT has a separate CLI script for this (`sct_apply_transfo`), which means that we're often calling one CLI script from another CLI script. 

In every single case of this, the sub-call specifies `-v 0` (to silence the logging messages, presumably because the intermediate output would be confusing/distracting from the "main" processing). However, this verbose status is not actually propagated to the call from `sct_apply_transfo` (the SCT wrapper script) to `antsApplyTransfo` (the ANTs binary that does the actual warping).

This means that, for scripts such as MOCO that apply warping fields repeatedly to a 4D image, we get a lot of noisy output that interrupts the pretty TQDM progress bar:

image

By propagating the `-v 0` status, we get this instead:

image

## Concerns

This change will also silence the logging output for registration, straightening, etc. However, I don't think this will have a negative effect, because there is typically always logging prior to the call to `sct_apply_transfo`, and so the blue text is extraneous. For example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d151cc390a088bbc3251cc1a3086abd602b24dfd/spinalcordtoolbox/registration/core.py#L169-L178

## Linked issues

Fixes #4581.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -211,7 +211,9 @@ class Transform:
                       '-i', fname_src,
                       '-o', fname_out,
                       '-t'
-                      ] + fname_warp_list_invert + ['-r', fname_dest] + interp, is_sct_binary=True)
+                      ] + fname_warp_list_invert + ['-r', fname_dest] + interp,
+                     verbose=verbose,
+                     is_sct_binary=True)
 
         # if 4d, loop across the T dimension
         else:


mguaypaq (approved):
Thanks for the fix! It'll be nice having uninterrupted progress bars.

PR #4605, opened 2024-08-15
Skip `-h` duration test on macOS CI runners due to sporadic runtimes

### LINKED ISSUE #4515: macOS/Windows runners take an inconsistent amount of time to run, sporadically triggering the `-h` profiling test
I got a failure notification on master the other day:

- [38m 46s](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9382948813/job/25835620060) - many failures
- [21m 40s](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/9382948813/job/25845676211#step:9:509) - no failures

I don't know how frequent this will be. I'm OK with restarting the jobs if it's once in a blue moon, but if it's frequent, then we might want to consider modifying the test.

(Maybe we could choose to only run it on PRs? Since, we mainly want to identify when a change will add a non-lazyloaded, slow import.)

joshuacwnewton:
## Description

The problem described in #4515 turned out to not be a "one in a blue moon" kind of slowdown. I'd figure it's more a 50/50 chance to get a slow macOS runner, and so I've been having to repeatedly restart the cron jobs to ensure passing runs.

We could go with a fancier solution (checking the runner hardware, having a different "passing" time for macOS, etc.) but given that our other platforms (Ubuntu, Windows) have _never_ failed this check, I think it should be OK to rely on them as a signal for whether a PR is adding a slow import.

## Linked issues

Fixes #4515.

### DIFF
--- a/testing/cli/test_cli.py
+++ b/testing/cli/test_cli.py
@@ -1,5 +1,7 @@
 # pytest unit tests for all cli scripts
 
+import os
+import sys
 import pytest
 from importlib.metadata import entry_points
 import subprocess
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
Yeah, it seems fine to skip the macOS speed check if it's inconsistent and we test it on other platforms.

PR #4610, opened 2024-08-27
Allow error code 406 in broken link checker

joshuacwnewton:
## Description

Our "broken link checker" CI workflow started failing a few days ago due to the Radiopaedia.org domain. The links open fine in browser, but they can't be fetched by curl/wget. I also tried using https://http.app/, but got the same error.

I spent an hour or so deep-diving into the error code, but I couldn't identify the source of the error, so my best guess is that this is a configuration issue in Radiopaedia.org's CloudFlare settings, or an attempt to stop crawlers from scraping Radiopaedia's articles (due to recent LLM controversies re: ignoring `robots.txt`).

We could have also filtered the Radiopaedia domain (as we had done in the past with other domains), but due to the limited scope of the 406 error code, I figure that filtering it would allow us to continue checking Radiopaedia for other return codes (in case their config changes again in the future).

> [!NOTE]
> I also took ownership of the notifications for this workflow (by disabling then re-enabling the workflow) that way there are fewer notifications for @mguaypaq to have to manage.

## Linked issues

N/A.

### DIFF
--- a/.github/workflows/check-url.sh
+++ b/.github/workflows/check-url.sh
@@ -33,6 +33,23 @@ if [[ $status_code -ge 200 && $status_code -le 299 ]];then
     exit 0
 fi
 
+# Check for "406 Not Acceptable" error code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406)
+# From https://http.dev/406: "In practice, this error is rarely used because the server supplies a default
+#                             representation instead."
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
The code looks fine (I like that the exact status code gets written to `valid_urls.txt`). Thanks for including all the relevant context in the code comment!

PR #4611, opened 2024-08-27
Convert softmask to binary mask before passing to ANTs, rather than applying softmask to input data

### LINKED ISSUE #4579: Suspicious application of a mask in moco
While working on #4578, I noticed suspicious things about the use of the input mask. 

## Inconsistent input type

Here is the line that calls the application of the mask:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/aa4aa7b038745e86b3058c5f8819de999d5c01ed/spinalcordtoolbox/moco.py#L605-L606

As can be seen, the second input argument to `apply_mask_if_soft()` is an Image Class `im_maskz_list[iz]`. Proof from the debugger:
```python
print(im_maskz_list[iz])
<spinalcordtoolbox.image.Image object at 0x7f8523e4fac0>
```

However, the function expects a file name:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/aa4aa7b038745e86b3058c5f8819de999d5c01ed/spinalcordtoolbox/image.py#L1657

So this is the first weird thing I observed.

## Strange logic to multiply the mask with the image

~The second weird thing is that the function is supposed to multiply the mask with the image, and then output the masked image. In order to do so, the mask and the image should have the same shape in order to be multiplied. So I assume this is what this check is for:~

SCRAP THAT: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4579#issuecomment-2256715468

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/aa4aa7b038745e86b3058c5f8819de999d5c01ed/spinalcordtoolbox/image.py#L1668

However, it performs the multiplication only if the shape of the mask and of the image are NOT the same. Shouldn't we expect something like this instead?
```python
    if np.array_equal(im_mask.data, im_mask.data.astype(bool)):
```

### Multiplying mask with image

Another issue, even more concerning, is about the rationale for multiplying the image with the mask. When doing so, the output image will have zero outside the mask, and image content inside. However, the registration algorithm will treat the boundary of the zeroed region as a strong contrast for aligning the source and target image, resulting in poor registration performance. I am wondering why we implemented this logic in the first place?

joshuacwnewton:
## Description

Currently, we use an `apply_mask_if_soft` function (originally added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2634) to sidestep the limitation of the ANTs binaries only accepting binary masks. However, this method has caveats:

> When doing so, the output image will have zero outside the mask, and image content inside. However, the registration algorithm will treat the boundary of the zeroed region as a strong contrast for aligning the source and target image, resulting in poor registration performance. 

So, this PR applies the suggestions from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4579#issuecomment-2259002303, namely removing `apply_mask_if_soft` and binarizing the input mask wherever we would use `apply_mask_if_soft`.

## Testing this PR

Note: This is a behavior change, and so I expect the moco test results to change. For example, in the `test_sct_dmri_moco_with_mask_check_params` test, we were using a mask that looked like this:

image

Binarizing this mask at 0.5 will result in a much smaller mask, so I've increased the size to compensate.

Note: I also noticed that we weren't using the `-m` argument in `batch_processing.sh`, and instead directly cropping the data using the mask. So, I expect the batch_processing.sh results to change as well.

## Linked issues

Fixes #4579.


mguaypaq (approved):
The code change seems fine. Thanks for cleaning up the now-unused function, and for documenting in the SCT usage strings that masks are automatically binarized.

PR #4615, opened 2024-08-28
Test EPI model to ensure that it has the correct `nnUNetTrainer` model structure

### LINKED ISSUE #4614: EPI spinal cord seg model is missing `nnUNetTrainer` folder, causing it to crash on v6.4
In PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4583, we added the requirement that all `nnUNetv2` models must retain their original `nnUNetTrainer` folders, so that the model folders can be easily re-used with 3DSlicer.

In that PR, we made sure that all of the `nnUNetV2` models had the correct structure, and that all of the tests passed. Yet, somehow, we missed the EPI model (#4454), leading to the following forum post: https://forum.spinalcordmri.org/t/problem-with-seg-sc-epi-nnunettrainer/1300/1

On closer inspection, we're not actually testing the EPI model at all! So, this slipped through the cracks. This is my mistake, as I should have requested a test on the original PR adding the EPI model.

TODO:

- [x] Write a test ensuring that the EPI model is run (even non-EPI data is OK)
- [x] Create a new release of the EPI model containing the correct folder structure.
- [x] Update the URL in `models.py`
- [x] Backport the fix to version 6.4 to ensure `sct_deepseg` doesn't crash

joshuacwnewton:
## Description

This PR adds a failing test for the EPI model. Then, it updates the model release to one with the correct `nnUNetTrainer` structure, which will cause the test to pass.

## Linked issues

Fixes #4614.

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -167,7 +167,7 @@ MODELS = {
     },
     "model_seg_sc_epi_nnunet": {
          "url": [
-             "https://github.com/sct-pipeline/fmri-segmentation/releases/download/v0.2/model-fmri-segmentation-v0.2.zip"
+             "https://github.com/sct-pipeline/fmri-segmentation/releases/download/v0.2/model-fmri-segmentation-v0.2_nnUNetTrainer.zip"
          ],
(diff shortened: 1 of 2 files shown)


mguaypaq (approved):
Perfect

PR #4620, opened 2024-09-03
Expose `space` API parameter in centerline CLI to allow user to specify `-space phys`

### LINKED ISSUE #4619: Output physical coordinates in the ".csv" file
**Feature justification**
I need to use the centerline coordinates from the ".csv" file produced when using `sct_get_centerline` for further processing, but they need to be in a physical coordinates system. Currently, they are in terms of number of pixels.

I believe a simple command line option that specifies whether the ".csv" file outputs in "pix" or in "phys" coordinates. I looked around at the code and [an API](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b2a6393c842ed0aae0a84083b3683d720a48b511/spinalcordtoolbox/centerline/core.py#L95) seems to have it already implemented, however the user is not able to select it since there is no option for it on the command line.

joshuacwnewton:
## Description

For sample T2* data...

`pix`:

image

`phys`:

image

> [!NOTE]
> The affine for the test `t2s.nii.gz` has non-diagonal values (i.e. rotation), hence the different `phys` coordinates for the same `x` `pix` coordinates.
> 
> image

## Linked issues

Fixes #4619.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_get_centerline.py
+++ b/spinalcordtoolbox/scripts/sct_get_centerline.py
@@ -85,6 +85,15 @@ def get_parser():
         help="Binary or soft centerline. 0 = binarized, 1 = soft. Only relevant with -method fitseg."
     )
     optional.add_argument(
+        "-space",
+        metavar=Metavar.str,
+        type=str,
+        choices=["pix", "phys"],
+        default="pix",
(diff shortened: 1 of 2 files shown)


mguaypaq (approved):
Looks good, thanks for the quick and straightforward fix, and for adding a test!

PR #4622, opened 2024-09-04
Update `get_centerline` to ensure that output arrays match the orientation of the output centerline image

### LINKED ISSUE #4621: CSV centerline output file contains invalid coordinates if input is not in `RPI-` orientation
**Feature justification:**
`sct_get_centerline` currently outputs the coordinates of the centerline in the ".csv" file in RPI- coordinates. It would be great if it could output it in other coordinate systems. This is useful because RPI- coordinates is not necessarily the coordinate system I need and it is not in the coordinate system as the input image.

- Best case scenario would probably to have the ability to select the coordinate system from the command line. 
- Otherwise, the coordinate system could be the same as the coordinate system of the NIfTI file. I believe this is a possible solution but requires the user to run `sct_image -setorient` beforehand.

**Describe alternatives you've considered**
I tried changing the coordinate system of the NIfTI to something other than LPI- with `sct_image -setorient` then using `sct_get_centerline` to force the coordinate system I wanted. It seems possible to get the coordinate system I want (i.e.: LPI-) by running:
`sct_image -setorient RPI`
`sct_get_centerline -space phys`
Note that I had to use **RPI**. This therefore very confusing.

joshuacwnewton:
This PR attempts to fix an orientation-related bug within the `get_centerline()` function in `centerline/core.py`.

### I/O

The function outputs multiple representations of the centerline: 

1. The typical `Image` object output (`im_centerline`), but also 
2. Objects called `arr_ctl`/`arr_ctl_der` (`3xN` arrays of coordinates).

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/38793f1ace92ae22ddb8237d4d56d4827f736171/spinalcordtoolbox/centerline/core.py#L283-L286

> Note that `fit_results` only represents some statistics about the intermediate curve fitting procedure, and is mainly used for debug visualizations, so it's not as relevant for our purposes.

### Reorientation

At the start of this function, the image is reoriented to RPI- (LAS+), as is standard in SCT, to preserve the assumption in code that x=LR, y=AP, and z=SI.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/38793f1ace92ae22ddb8237d4d56d4827f736171/spinalcordtoolbox/centerline/core.py#L106-L108

The original orientation is stored so that the resulting output centerline can be changed back to match the input.

## The issue

While the `im_centerline` object is correctly reoriented back to the original orientation of the input image, there are a number of bugs with the corresponding `arr_ctl` coordinate output:

- When `space=pix`, the `arr_ctl` stays in RPI-/LAS+ orientation, and is never reoriented back to the original orientation.
- When `space=phys`, the issue gets even worse, as those same RPI-/LAS+ pixel coordinates get matmul'd by the affine _of the original image_. This means that the matmul is only valid if the original image was in RPI-/LAS+ orientation (RPI- * RPI-); all other input orientations will result in nonsense physical coordinates.

## Scope of the issue

The `arr_ctl` object gets used in the following ways:

- **CLI**: When calling `sct_get_centerline`, it is saved directly as an output `.csv` file, but not used directly in any processing. _(**Note**: This was the context for [the original issue](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4621), and has a [hacky workaround](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4622#issuecomment-2334345341).)_
- **API**: Outside of `sct_get_centerline`, `arr_ctl` is used in the following places:
   image
   If we take a look at one of these usages (e.g. `csa_pmj.py`), we can see that it is often used immediately to construct a `Centerline()` object, in order to access a number of unique analysis methods:
   https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d151cc390a088bbc3251cc1a3086abd602b24dfd/spinalcordtoolbox/csa_pmj.py#L61

Some open questions:

- How does `Centerline()` actually use the `arr_ctl` object?
- If `Centerline()` was built around this buggy behavior, will "fixing" the `arr_ctl` coordinates have unintended consequences?
- Does the answer to this question depend on whether we choose `space='pix'` or `space='phys'`?

EDIT: The answer to the above questions is addressed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4622#discussion_r2036049931.

## Linked issues

Fixes #4621.


spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -214,11 +215,21 @@ def main(argv: Sequence[str]):
...
+                                            # space=arguments.space,
+                                            verbose=verbose,
+                                            remove_temp_files=arguments.r)
+
+    # Fetch `arr_centerline` from `im_centerline`
+    # (This is the easiest way to get coords that are in their original orientation. This is largely equivalent to how
+    #  arr_centerline is determined inside `get_centerline`, but without the RPI orientation.)
+    arr_centerline = np.array(find_and_sort_coord(im_centerline))
+    # Convert from pixel coordinates to physical coordinates (in original orientation)
+    if arguments.space is "phys":
```
po09i, line 231: Screenshot 2024-09-04 at 8 57 59 PM


```suggestion
    if arguments.space == "phys":
```

po09i: The output coordinates in the ".csv" are in pixels even though I have `-space phys`. This is probably why.

joshuacwnewton:
I think that I have fundamentally misunderstood affine matrices in this PR. I'm temporarily closing this PR so that I can discuss further in the issue.

spinalcordtoolbox/scripts/sct_get_centerline.py
joshuacwnewton, line 1: I thought I had made the correct changes, but on further testing, I am actually quite confused. I'm not sure that I fully understand physical coordinates.

For example, take the [t2.nii.gz](https://github.com/user-attachments/files/16895390/t2.nii.gz) file that currently exists in `sct_testing_data`. When I check its header, I see:

```
sform_xorient	Anterior-to-Posterior
sform_yorient	Inferior-to-Superior
sform_zorient	Left-to-Right
```

However, I notice that the q/sforms are not diagonal:

```
sto_xyz:1	0.000000 0.000000 1.000000 -32.219284 
sto_xyz:2	-1.000000 0.000000 0.000000 31.786606 
sto_xyz:3	0.000000 1.000000 0.000000 -19.321671 
```

Meaning that even though the image is in AIL orientation, applying the affine matrix will actually result in LPI physical coordinates. ((We can confirm this by reorienting the image to LPI, then checking the header once more to confirm that it is diagonal and positive.))

```
sto_xyz:1	1.000000 0.000000 0.000000 -32.219284 
sto_xyz:2	0.000000 1.000000 0.000000 -27.213394 
sto_xyz:3	0.000000 0.000000 1.000000 -19.321671 
sform_xorient	Left-to-Right
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
```

In other words, no matter which orientation we choose with `sct_image -setorient`, the affine matrix will always be rearranged in tandem so that the resulting physical coordinates stay the same.

In that case, is my PR useless? Well... not exactly. With this PR, now the `.csv` file correctly matches the physical coordinates given by FSLeyes. e.g.:

image

image

((On `master`, the .csv coordinates do not match what is seen in FSLeyes!))

However, I'm not sure if this matches @po09i's expectations. (i.e. `-setorient` can't be used to change the format of the `.csv` file, because it is fixed regardless of `-setorient`.)

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -214,11 +215,21 @@ def main(argv: Sequence[str]):
...
+                                            # NB: We intentionally skip the pix->phys conversion here, since we want to
+                                            #     convert to phys AFTER the image has been reoriented away from RPI.
+                                            # space=arguments.space,
+                                            verbose=verbose,
+                                            remove_temp_files=arguments.r)
+
+    # Fetch `arr_centerline` from `im_centerline`
+    # (This is the easiest way to get coords that are in their original orientation. This is largely equivalent to how
+    #  arr_centerline is determined inside `get_centerline`, but without the RPI orientation.)
+    arr_centerline = np.array(find_and_sort_coord(im_centerline))
```
mguaypaq, line 229: This is true for the `optic` method (the default), or when `param.soft == 1` (_not_ the default), but in other cases will result in centerline coordinates being rounded to integer values. Is this loss of precision ok?

I guess the CSV is informative only, and this is a quick fix before looking at whether `get_centerline` has a deeper bug.

spinalcordtoolbox/scripts/sct_get_centerline.py
po09i: Thanks for looking into this! Coordinate systems are a pain!

> For example, take the [t2.nii.gz](https://github.com/user-attachments/files/16895390/t2.nii.gz) file that currently exists in sct_testing_data. When I check its header, I see:

```
sform_xorient	Anterior-to-Posterior
sform_yorient	Inferior-to-Superior
sform_zorient	Left-to-Right
```

However, I notice that the q/sforms are not diagonal:

```
sto_xyz:1	0.000000 0.000000 1.000000 -32.219284 
sto_xyz:2	-1.000000 0.000000 0.000000 31.786606 
sto_xyz:3	0.000000 1.000000 0.000000 -19.321671 
```

> Meaning that even though the image is in AIL orientation, applying the affine matrix will actually result in LPI physical coordinates. ((We can confirm this by reorienting the image to LPI, then checking the header once more to confirm that it is diagonal and positive.))

I do not believe this is true. I believe this means that it will result in LAI physical coordinates. Here is why I think that. I believe AIL means that the axis of the voxel data in the first dimension varies from A-->P, 2nd axis I->S, 3rd axis L->R. However, the x, y and z axis of the world coordinates have some implicit associations (x: L<->R, y: A<->P, z: I<->S). Notice that in the second affine matrix, the values not only moved but also changed (31 to -27). Also notice that this is the y coordinate (AP) which is the one different between LAI and LPI).

So if we use `sct_image -setorient LAI` on the t2 image, we now expect it to be diagonal, but with the same values just reorganized as your first affine you posted in your comment.

```
10 0 -32.22
0 -1 0 31.79
0 0 1 -19.32
0 0 0 1
```

Note: I thinks Philips have their x and y flipped but I'm not sure how NIfTIs handle that. I believe converters just store the Philips "y" as "x" and vice versa but I'm not sure.

> In other words, no matter which orientation we choose with sct_image -setorient, the affine matrix will always be rearranged in tandem so that the resulting physical coordinates are in LPI order.

This does not hold anymore

po09i: I just noticed that you explicitly wanted a positive diagonal which is why you chose to reorient to LPI.

joshuacwnewton: Sorry, I think that was a communication error on my part. When I say LPI I mean "Left to right, posterior to anterior, inferior to superior" but this convention is different elsewhere.

spinalcordtoolbox/scripts/sct_get_centerline.py
```diff
@@ -214,11 +215,21 @@ def main(argv: Sequence[str]):
...
+    # Convert from pixel coordinates to physical coordinates (in original orientation)
+    if arguments.space == "phys":
+        arr_centerline = im_centerline.transfo_pix2phys(arr_centerline.T, mode="absolute").T  # same as get_centerline
```
mguaypaq, line 232: I really suspect that the corresponding lines in `get_centerline` are buggy, if there's a `im_seg.change_orientation` in between generating the voxel coordinates of the centerline and transforming them to physical space. But this will require careful thinking and testing.

mguaypaq (approved):
Since your manual testing shows that the output values are reasonable, I'm OK approving this quick fix because of the ISMRM deadline, but we should really figure out the correct fix (which should address my point below about loss of precision) as a follow-up.

(But mostly: thanks for the quick fix! I appreciate it.)

mguaypaq (changes requested):
After a very clarifying zoom call, the conclusion was that we should probably fix the bug differently.

mguaypaq:
### Summary from a zoom discussion with @joshuacwnewton and @po09i

The first point of confusion was that there are two notions of orientation:

* World (aka physical) coordinates
* Voxel coordinates

#### World orientation

The nifti standard says that world coordinates are always RAS+ (aka LPI-).
The world coordinates are $(x, y, z)$, floating point numbers, measured in mm (or some other physical unit), and RAS+ for nifti means:

* $+x$ is to the subject's Right (and $-x$ to the Left)
* $+y$ is to the subject's Anterior/front (and $-y$ to the Posterior/back)
* $+z$ is to the subject's Superior/top (and $-z$ to the Inferior/bottom)

Because the subject may not be perfectly aligned to the scanner, $+x$ is _mostly_ to the Right, but maybe also a little bit A/P/S/I, etc.

#### Voxel orientation

In contrast, the voxel coordinates $(i, j, k)$ are positive integers, indexing into the 3D numpy array stored in the nifti file. They have no units. You can think of the voxel coordinates as describing an evenly spaced (but possibly not axis-aligned and possibly not isotropic) sampling grid in physical space. The affine matrix is a 4x4 matrix which specifies this sampling grid:

```
[xi xj xk x0]
[yi yj yk y0]
[zi zj zk z0]
[ 0  0  0  1]
```

* The last column gives physical, floating point coordinates of the center of the $(0, 0, 0)$ voxel.
* The first column gives the $(x, y, z)$ vector which corresponds to taking one step in the $i$ direction.
* The second column is the same for the $j$ direction.
* The third column is the same for the $k$ direction.

The "voxel orientation" describes this mapping. For example, "RPI-" means:

* The $i$ direction points mostly from R to L (so, `xi` is large and negative)
* The $j$ direction points mostly from P to A (so, `yj` is large and positive)
* The $k$ direction points mostly from I to S (so, `zk` is large and positive)

#### Desired result

What Alexandre wants is to get centerline coordinates (floating point) in the scanner's coordinate system.

* This is a physical coordinate system (floating point values with a physical unit like mm), not voxel coordinates
* But depending on the manufacturer, this may be in either RAI+ or ARI+ world orientation, not nifti's mandated RAS+ world orientation

#### SCT's behaviour

This kind of world coordinate conversion is probably more the domain of DICOM/NIFTI conversion tools, rather than SCT. Because SCT focuses on nifti files, we follow nifti's RAS+ world orientation convention. What functions like `sct_image -set-orient` and `Image.change_orientation()` do is:

* Transpose and/or reverse the axes of the numpy array stored in the nifty file
* Also permute and/or negate the columns of the affine matrix (and possibly fix the origin column) so that the sampling grid in physical space stays the same

That means that `sct_get_centerline -space phys` should output the same physical coordinates regardless of any `sct_image -set-orient` commands. Currently it doesn't because of a bug, but it should.

Also, `sct_get_centerline -space pix` should output different voxel coordinates based on `sct_image -set-orient` (so that they can be used directly to index into the data array), but it doesn't, because of the same bug.

#### Proposed solution

Alexandre and I discussed the possibility of using `sct_image -flip` and `sct_image -transpose` to trick SCT into outputting the world coordinates that he wants as a quick fix. This would mostly work (even if it's a hack), except that the origin value might be wrong. So our conclusion was that he should:

* Use `sct_image -set-orient RPI` on the segmentation file first, to avoid the bugs mentioned above. (Also on the image file, if he wants a QC report?)
* Use `sct_get_centerline -space phys` to get the floating point coordinates in RAS+ world orientation
* Write two very simple python scripts to convert the resulting CSV file to either RAI+ or ARI+ (depending on the scanner manufacturer) after the fact

Meanwhile, for Joshua and I:
* The "quick fix" PR is a lot less urgent
* We should probably just fix the underlying bug

joshuacwnewton:
> We should probably just fix the underlying bug

As per your recommendations, I've updated the title and description of this PR to describe the buggy behavior for both the `space=pix` case and the `space=phys` case.

---

Also, I'm hopeful that the underlying changes to `get_centerline` will actually be quite simple:

- Make sure that both the pixel coordinates and the affine matrix are in the same pixel coordinate space prior to calling `pix2phys`.
   - Either by using the RPI- affine to match the RPI- coordinates.
   - Or, by reorienting the RPI- coordinates to match the orientation original affine. 
- This will ensure that the output of `pix2phys` is consistently RAS+/LPI-, in line with the nifti standard.

However, the tricky part here is what the callers of `get_centerline` actually expect from the coordinates:

- Is each instance of `space=phys` in our codebase written to work with RAS+/LPI- world coordinates?
- Is each instance of `space=pix` in our codebase written to work only with the (current) LAS+/RPI- output? Or, does it expect (/can it be rewritten to expect) original-orientation pixel coordinates, too?

In other words, I'm not sure if the current implementation is broken, or if changing the output of `get_centerline` will break the callers in new ways, or both. :sweat:

spinalcordtoolbox/scripts/sct_get_centerline.py
joshuacwnewton: Your suspicions ended up being 100% correct! :blush:

spinalcordtoolbox/centerline/core.py
```diff
@@ -274,9 +272,12 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
     arr_ctl_der = np.array([x_centerline_deriv, y_centerline_deriv, np.ones_like(z_ref)])
     # If 'phys' is specified, adjust centerline coordinates (`Centerline.points`) and
-    # derivatives (`Centerline.derivatives`) to physical ("phys") space and native (`im_seg`) orientation.
+    # derivatives (`Centerline.derivatives`) to physical ("phys") space, i.e. world coordinates.
+    # Note: According to the NIFTI spec, physical coordinates should always be in RAS+/LPI- orientation, no matter
+    #       what orientation the voxel coordinates are originally in. (This is because changing the image orientation
+    #       will also change the affine in tandem, causing RAS+/LPI- physical coordinates to result every time.)
     if space == 'phys':
-        # Transform centerline to physical coordinate system
+        # Note: im_seg is still in RPI, and the coordinates are in RPI, resulting in RAS+/LPI- physical coordinates
```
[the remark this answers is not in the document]
joshuacwnewton: ...All of the tests have passed (including `batch_processing.sh`, which is notoriously finicky).

I am really surprised! I thought that fixing this bug would have changed `Centerline` behavior drastically.

joshuacwnewton:
I've moved this off of the 6.5 milestone since it is a complicated numerical change, and I don't want to rush it (given that we're already rushing a bit for the new models + SCT Course).

Post SCT Course 7.0 release seems like a better time to make a change such as this.

joshuacwnewton:
Quick status update: In preparation for tackling this PR, I have done some refactoring work to better highlight the problematic lines of code:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ef7fa8b123c8b423143427f450f294b932e54ca1/spinalcordtoolbox/centerline/core.py#L269-L277

That way it's clearer where to insert the fix (since it wasn't entirely clear prior to the refactor).

spinalcordtoolbox/centerline/core.py
joshuacwnewton, line 1: I wanted to test my change, and found the most curious (existing) test in the process:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0481a037166257c15e5cdcbd43bfafde1784c9ea/testing/api/test_centerline.py#L178-L180

What a smoking gun! The test is explicitly only checking `arr_out` if the input image is RPI, which is exactly the behavior we encountered in the source issue (#4621).

However, it gets worse:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0481a037166257c15e5cdcbd43bfafde1784c9ea/testing/api/test_centerline.py#L168-L172

Because of the `if 'exclude_polyfit':` condition, the actual checks are unreachable! (This really should be a `pytest.skip` instead.)

When I re-enable the test, it fails... meaning we haven't been testing this behavior this whole time. :facepalm:

joshuacwnewton (commented):
Alright, I have a first draft for a fix, but I ran into a bit of a snag when trying to test my change.

Still, I think this is enough of a start to mark this as ready for review.

SomeoneInParticular (changes requested):
This is just a first (quick and dirty) review with some minor changes I'd like seen too, nothing major. 

As well, as you mentioned in your comment, tests to confirm this is working correctly also need to be implemented as well (or, rather, our existing tests need to actually do their job); in the meantime, I'll work on a more in-depth review focusing on the functionality which was changed. I don't see anything immediately problematic, but want to be sure before approving formally.

spinalcordtoolbox/types.py
```diff
@@ -66,33 +66,40 @@ def __eq__(self, other):
...
         dim_permute = np.array([dim_arr[perm[0]], dim_arr[perm[1]], dim_arr[perm[2]]])
         # invert indices based on maximum dimension for each axis
         for i in range(3):
             if inversion[i] == -1:
-                coord_permute[i] = dim_permute[i] - coord_permute[i]
+                if mode == 'absolute':
+                    coord_permute[i] = dim_permute[i] - coord_permute[i]
+                else:
+                    assert mode == 'relative'
+                    coord_permute[i] = -1 * coord_permute[i]
```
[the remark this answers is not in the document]
joshuacwnewton: Ah! This is actually a pattern I picked up from @mguaypaq. 

I think it mainly serves the same purpose of your pattern (making sure there isn't a bare 'else:' case).

But, the way I understand it (correct me if I'm wrong @mguaypaq!), we use this pattern because our users don't tend to access SCT's API? Since `mode` is an internal parameter (and not something controlled by any CLI arguments), the only people who should run into this check are us devs. So, the assertion acts more like a small reminder of the valid options (i.e. it just verifies assumed state), rather than acting as actual input validation.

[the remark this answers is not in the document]
mguaypaq: A slight nuance: I like to reserve the `if ... else assert` pattern for invariants that are guaranteed within a single function. A typical example would be:

```py
def do_something(mode):
    # early check that the function is called correctly
    if mode not in ['absolute', 'relative']:
        raise ValueError(f"bad {mode=}")

    # now do other stuff
    ...

    # conditional behaviour
    if mode == 'absolute':
        ...
    else:
        assert mode == 'relative'
        ...

    # maybe more common code
    ...

    # more conditional behaviour
    if mode == 'absolute':
        ...
    else:
        assert mode == 'relative'
        ...
```

The reasons are:
- After the initial check, it's hard to think of a good error message here:

  ```py
  if mode == 'absolute':
      ...
  elif mode == 'relative':
      ...
  else:
      raise ValueError("literally impossible")
  ```

- But, the code still documents close to the `else` body that we're specifically expecting `mode == 'relative'`.

- Technically, `assert` statements can be skipped (if run with `python -O`), and we wouldn't want the code to have different observable behaviour just because of optimizations (assuming no internal bugs, but allowing bugs outside the API boundary of the function call).

But, if there's no initial consistency check inside the function, I think `if ... elif ... else raise` is the right pattern to use.

spinalcordtoolbox/centerline/core.py
```diff
@@ -103,6 +104,15 @@ def get_centerline(im_seg, param=ParamCenterline(), verbose=1, remove_temp_files
...
 
+    # Handle whether we should keep the centerline as a temporary output
+    if remove_temp_files:
+        fname_ctr_temp = None  # Rather than creating then deleting the temp output, we can just skip creating it at all
+    else:
+        temp_path = tmp_create(basename="get-centerline")
+        fname_ctr = (add_suffix(os.path.basename(im_seg.absolutepath), "_ctr") if im_seg.absolutepath
+                     else "centerline.nii.gz")
+        fname_ctr_temp = os.path.join(temp_path, fname_ctr)
+
```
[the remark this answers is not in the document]
joshuacwnewton: I actually originally had it exactly as you describe? But, when everything was put together, it felt strangely... awkward:

```diff
    # Save outputs
-   if fname_ctr_temp:  # Preserve centerline image in tempdir for debugging purposes
-       im_centerline.save(fname_ctr_temp, mutable=True)
+   if remove_temp_files:
+       pass  # Skip creating the file, rather than wasting effort creating then deleting the temp output
+   else:
+       fname_ctr_temp = os.path.join(
+           tmp_create(basename="get-centerline"),
+           (add_suffix(os.path.basename(im_seg.absolutepath), "_ctr") if im_seg.absolutepath
+            else "centerline.nii.gz")
+       )
+       im_centerline.save(fname_ctr_temp, mutable=True)
    return (im_centerline,
            arr_ctl,
            arr_ctl_der,
            fit_results)
```

Something about seeing the above pattern rubbed my brain the wrong way, and it's kind of hard for me to place why that is. I think it comes down to readability/cognitive load?

- Because we interpret the `remove_temp_files` argument as a way of deciding *not* to do something, we're forced to use weird patterns like `if remove_temp_files: pass` (wasteful), `if not remove_temp_files:` (unintuitive and hard to grok). I even considered something like `write_tempfile = (not remove_temp_files); if write_tempfile:` to convert the argument, but that felt clunky too.
- There's a lot of space dedicated to constructing the temp path, which makes the "post-processing" block of the function busier that it needs to be.
- By determining the path elsewhere, the `if fname_ctr_temp:` condition is much simpler + easy to read.

I figured it was okay to split up the block (separated by ~100 lines), because I look at the function mentally like this?

```python
def get_centerline():
    # 1. Parameter parsing (im_seg, space, remove_temp_files)
    # 2. Big switch statement (`algo_fitting={optic,linear,nurbs,etc.}`)
    # 3. Post-processing + output saving
```

And so it's easier to mentally collapse that big "2." block, and put the parts of `remove_temp_files` into their requisite parts of the function. 

----

That said, this is a whole lot of words over something that I don't actually feel too strongly about, so I'm happy to change it back if you found the new pattern unintuitive as well.

spinalcordtoolbox/centerline/core.py
joshuacwnewton: The good news is that the `if 'exclude_polyfit'` was actually just a syntax error. Changing it to `params.get('exclude_polyfit', False)` works fine, and the test fully passes! 

(Only one test case is skipped, with 8/9 polyfit cases passing as expected.)

joshuacwnewton: Setting a breakpoint inside of `permute`, it seems like the code accidentally tries to permute (`orig_orientation`+`orig_orientation), leading to no change.

This is an easy fix. I just needed to specify `im_centerline_rpi` instead of `im_centerline`, and make sure that `im_centerline_rpi` doesn't get mutated by the reorientation.

However, even after fixing this bug, I notice that the reoriented coordinates no longer match, having either it's L/R or S/I values swapped. :sweat: 

image

image

I'll have to dig deeper into the coordinate reorientation now, and see if I made a math error.

joshuacwnewton: Huzzah!! I've gotten the `test_get_centerline_polyfit` to work on non-RPI images, eliminating the hacky `# check arr_out only if input orientation is RPI (because the output array is always in RPI)` caveat. Now the output array is **not** always in RPI.

It involved fixing both an bug in my own code (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4622/commits/267c156d6ac3618c25f4d7b1548298de196db44a) and an off-by-one bug in the existing `permute` function (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4622/commits/a76875f7039120a750773df570c553a32b6ea809). But after doing that, the `arr_out` coordinates match `im_out` in the original orientation! :D

spinalcordtoolbox/types.py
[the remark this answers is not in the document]
joshuacwnewton: Thanks for the clarification! It looks like I was applying the pattern too liberally, then. :stuck_out_tongue_closed_eyes:

testing/api/test_centerline.py
```diff
@@ -116,9 +116,9 @@ def dummy_centerline(size_arr=(9, 9, 9), pixdim=(1, 1, 1), subsampling=1, dilate
      {}),
     (dummy_centerline(size_arr=(30, 20, 50), subsampling=1, outlier=[20]),
      {'median': 0, 'rmse': 0.8, 'laplacian': 70, 'norm': 14},
-     {'exclude_nurbs': True}),
```
joshuacwnewton, line 119: It looks like the `exclude_nurbs` check that I thought was unnecessary still seems to be necessary. Even though it passed for me locally on Ubuntu 24.04, it still seems to be failing for CentOS, Ubuntu 22.04, etc. so I'll re-add it.

joshuacwnewton: Rebased!

spinalcordtoolbox/centerline/core.py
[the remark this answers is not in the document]
SomeoneInParticular: # Save outputs
-   if fname_ctr_temp:  # Preserve centerline image in tempdir for debugging purposes
-       im_centerline.save(fname_ctr_temp, mutable=True)
+   if not remove_temp_files:
+       fname_ctr_temp = os.path.join(
+           tmp_create(basename="get-centerline"),
+           (add_suffix(os.path.basename(im_seg.absolutepath), "_ctr") if im_seg.absolutepath
+            else "centerline.nii.gz")
+       )
+       im_centerline.save(fname_ctr_temp, mutable=True)
    return (im_centerline,
            arr_ctl,
            arr_ctl_der,
            fit_results)
            
Why not this then?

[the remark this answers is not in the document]
joshuacwnewton: Yeah, I think that's the approach that makes the most sense, even if `if not remove_temp_files` is a bit hard to grok, aha.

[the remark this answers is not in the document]
SomeoneInParticular: Fair enough; honestly it was a bit of a nit, and I'd rather not spend any further time bikshedding on something that will likely be stripped out entirely if a future PR. Specifically, one which standardizes how Temp Folders are handled in SCT; I'll make a corresponding issue for that discussion momentarily.

spinalcordtoolbox/types.py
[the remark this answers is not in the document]
SomeoneInParticular: Yeah, that was my sentiment. Related to this, its probably worth going through SCT and converting these strings to proper `enum`s at some point, though that is definitely out of scope for this PR.

testing/api/test_centerline.py
joshuacwnewton, line 1: I've dug around a bit, and I think the answer to the question "how do the callers consume the resulting `arr_ctl` object?" has turned out to be really straightforward:

1. When `space='phys'`, there is no change in output. 
   - _This should make some intuitive sense: coordinates in physical space *should* be invariant to pixel space orientation. This is because, to convert coordinates from `pix2phys`, we need to multiply by the corresponding affine. And we ensure that for RPI coordinates, we use the RPI affine, and likewise for any other orientations._
2. When `space='pix'`, there **would** be a difference... if we didn't already include multiple layers of "reorient to RPI" steps prior to calling `get_centerline`. 
   - Meaning, `get_centerline` never actually sees the "original orientation" image in most cases.

Out of the 12 cases where `arr_ctl`/`arr_ctl_der` are used:

- 4/12 were `space='phys'`: `csa_pmj.py`, `sct_compute_compression`, and `straightening.py` x2
- 8/12 were pre-reoriented to RPI: `flattening.py`, `csa_pmj.py`, `labels.py` x2, `labeling.py`, `process_seg.py`, `sct_compute_compression`, and `sct_analyze_lesion`.

So... it looks like the `sct_get_centerline` output CSV file was actually the *only* place where the "original orientation" coordinates were actually accessed, and thus the scope of the bug was very limited.

I think we can remove the `reorient_array` parameter, then, and just universally reorient (as well as add a big warning that if you want output coords to be in RPI, then you must reorient the image prior to calling `get_centerline`).

SomeoneInParticular (approved):
Thank you for adding the canary test; LGTM to me!

mguaypaq (approved):
Thank you for this massive bug fixing effort. The code looks good, and I like the structure of the fix.

Thanks for checking all callers to check that they're either using `space='phys'` or already in RPI, so that they're unaffected by the fix.

Thanks for spotting the off-by-one in `Coordinate.permute` (and I agree that your fix is correct).

And thanks for fixing/adding tests!

PR #4625, opened 2024-09-10
Split "Command Line Tools" page into multiple individual pages (with markdown formatting for `argparse`)

### LINKED ISSUE #3095: Split "Command-line Tools" RTD page into 1 page per CLI script

joshuacwnewton:
## Description

The purpose of this PR is to go from this:

> image

To [**this**](https://spinalcordtoolbox--4625.org.readthedocs.build/user_section/command-line/sct_analyze_lesion.html#sct-analyze-lesion):

> image

This way, each SCT script will have its own dedicated page with nice formatting to put additional information (such as tips and tricks, links to relevant tutorials, etc.)

As an added benefit, this will also let us link _to_ each page from the tutorials. Meaning, dedicated information for each script will be one click away at all times.

This is a draft, and definitely needs to be proofread, but please check it out here: https://spinalcordtoolbox--4625.org.readthedocs.build/user_section/command-line.html

> [!NOTE]
> This PR is gigantic, but much of the effort was done via script-based editing. I didn't do all of this by hand, heh. :sweat: 

## Linked issues

Fixes #3095.


spinalcordtoolbox/deepseg/models.py
```diff
@@ -535,7 +535,7 @@ def list_tasks_string():
     # Display coloured output
     color = {True: 'LightGreen', False: 'LightRed'}
     table = f"{'TASK':<30s}{'DESCRIPTION':<50s}\n"
-    table += f"{'-' * 80}\n"
+    table += f"- {'-' * 76}- \n"
```
joshuacwnewton, line 538: TODO: Find a more aesthetically pleasing way of tricking Sphinx into thinking that this table header is not an rST header.

documentation/source/requirements.txt
```diff
@@ -4,3 +4,4 @@ sphinx_rtd_theme
 sphinx-copybutton
 furo==2023.09.10
 sphinx
+sphinx_argparse
```
joshuacwnewton, line 7: In this week's SCT meeting, feedback was raised to see if we can get `sphinx_argparse` to create linkable anchors for each of the arguments in a script's argparse help.

mguaypaq (approved):
Thank you for this massive effort, and for such a nice breakdown into individual commits, and for the code comments explaining how to update the manual TOC tree! It's an amazing improvement.

spinalcordtoolbox/scripts/sct_crop_image.py
```diff
@@ -18,19 +19,30 @@ def get_parser():
...
+              sct_crop_image -i t2.nii.gz -xmin 5 -xmax 60 -zmin 5 -zmax -2
+              ```
+            - To crop an image using a binary mask, and keep a margin of 5 voxels on each side in the x and y directions only:
+              ```
+              sct_crop_image -i t2.nii.gz -m mask.nii.gz -dilate 5x5x0
+              ```
+        """),  # noqa: E501 (line too long)
     )
 
     mandatoryArguments = parser.add_argument_group("\nMANDATORY ARGUMENTS")
```
joshuacwnewton, line 48: I've found a slight bug. The fact that we start every argument group with `\n` causes the anchor link for these sections to be rendered with an extra space (`' '`), which breaks `sphinx-argparse`.

Sidebar link (working):

image

Anchor link from `¶` hover button (not working):

image

----

The good news, though, is... the `\n` actually have no effect on the terminal output! I'm not sure if Sphinx strips whitespace or what, but even after removing the `\n`, the output of `sct_script -h` renders identically (i.e. with a blank line).

So, I think I can go ahead and mass replace all instances of `parser.add_argument_group("\n` with `parser.add_argument_group("`.

documentation/source/requirements.txt
joshuacwnewton: I've tried, but I was unable to make this work. :(

- `sphinx-argparse` uses `docutils.nodes` under the hood, and creates a tree structure where each element of the documentation has a specific node type. 
- The anchor links for argument section titles (`MANDATORY ARGUMENTS`, etc.) are generated via the `nodes.title` type.
  - Title nodes get converted into this `headerlink` HTML: 
     ```
     <h2>MANDATORY ARGUMENTS<a class="headerlink" href="[#mandatory-arguments](view-source:http://localhost:63342/spinalcordtoolbox/documentation/build/html/user_section/command-line/sct_deepseg.html#mandatory-arguments)" title="Link to this heading">¶</a></h2>
     ```
  - This transformation from "docutils node" to "formatted html" is _not_ handled by the extension itself. Instead, it's handled by Sphinx in the [`add_permalink_ref`](https://github.com/sphinx-doc/sphinx/blob/914f3f4e541309378d506662d2dd4b392f3691f2/sphinx/writers/html5.py#L413-L418) method, which gets called by the `depart_title` method for titles.
- By comparison, each individual argparse argument is given the `nodes.option_list_item` type instead (which makes sense!). 
  - Option list nodes get converted into this html:
     ```
     <dl class="option-list">
     <dt><kbd>-i</kbd></dt>
     <dd><p>Image to segment. Can be multiple images (separated with space).</p>
     </dd> 
     [...]
     </dl>
     ```
  - The `<kbd>` tag corresponds to the [`depart_literal`](https://github.com/sphinx-doc/sphinx/blob/914f3f4e541309378d506662d2dd4b392f3691f2/sphinx/writers/html5.py#L619-L621) method, which has no such options for headerlinks.

So, unless we somehow modify the `sphinx_argparse` extension to _not_ generate option lists, or somehow modify `sphinx` to change how it handles option lists, I think we're out of luck here.

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Easy peasy! Both of the problematic tables were in argparse `epilogs`, so if we just add `:noepilog:` to both `sct_deepseg` and `sct_download_data`, we avoid the mis-rendered tables and prevent the error.

Issue #4637, opened 2024-09-27
Improve testing coverage of CLI entry points using basic test template


joshuacwnewton:
Recently, issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4636 was reported by a user. Frankly, I'm embarrassed that I let this bug slip through, as it reveals that areas of `sct_maths` are entirely untested. Catching this bug would have been very simple, not requiring checking of results at all.

----

In hindsight, I think SCT has been a bit lax with testing, as lately we've been focusing our efforts on new features/refactors. I understand why; personally I find tests to be taxing to write, since they involve both tedious boilerplate + careful thought about validating results (ground truth files? dummy data? etc.).

But, even if we don't shoot for 100% coverage of the entire SCT project, I think we should make it a goal to test all of the CLI options for each script. That way, we can at least eliminate low-hanging-fruit bugs for typical usage like the ones from #4636.

----

To aid in this effort, I think it might be helpful to come up with a test template that can be used as a starting point. For example:

```python
@pytest.mark.parametrize("ofolder", ["tmp_path", None])
@pytest.mark.parametrize("o", ["img_out.nii.gz", None])
@pytest.mark.parametrize("args_to_test, fname_in, fname_gt", [
    (["ARGUMENTS"], sct_test_path("", ""), None),
], ids=["-test_name"])
def test_SCT_SCRIPT_basic(args_to_test, fname_in, fname_gt, o, ofolder, tmp_path):
    """Basic test to ensure core functionality runs without error."""
    # base args to test
    args = ["-i", fname_in]
    args.extend(args_to_test)

    # test all combinations of both (user-supplied and default) output filenames and folders
    fname_out = o if o else "DEFAULT_FILENAME"
    path_out = (str(tmp_path) if ofolder == "tmp_path"
                else os.getcwd())  # Default ofolder may be CWD or it may be `os.path.dirname(fname_in)`
    if ofolder:
        args.extend(["-ofolder", path_out])
    if o:
        args.extend(["-o", fname_out])
    # note: if script doesn't have -ofolder, use this pattern instead:
    # if o:
    #     args.extend(["-o", os.path.join(path_out, fname_out) if ofolder else fname_out])

    # run script
    SCT_SCRIPT.main(argv=args)

    # test all output files exist
    outfiles = [fname_out]
    for outfile in outfiles:
        assert os.path.exists(os.path.join(path_out, outfile))

[...]

PR #4638, opened 2024-09-27
Fix incorrectly lazy-loaded `dipy` imports

### LINKED ISSUE #4636: `dipy` is lazy-loaded incorrectly, causing an `AttributeError` when running `sct_maths -denoise`
### Description

LazyLoader seems to cause issues in my system, previously working with SCT version 6.1.

### Steps to Reproduce

0. Install SCT <release or git revision and platform, you can use the
   first lines of output of sct_check_dependencies>
```
SCT info:
- version: 6.4
- path: /scratch/01/user/workspace/src/projects/sct
OS: linux (Linux-5.15.0-122-generic-x86_64-with-glibc2.35)
```

1. <First Step>
sct_maths -i image.nii.gz -denoise 2 -o denoised_image.nii.gz

Command executes without error

```
  File "/scratch/01/user/workspace/src/projects/sct/spinalcordtoolbox/scripts/sct_maths.py", line 621, in <module>
    main(sys.argv[1:])
  File "/scratch/01/user/workspace/src/projects/sct/spinalcordtoolbox/scripts/sct_maths.py", line 553, in main
    data = sct_math.denoise_nlmeans(data, patch_radius, block_radius)
  File "/scratch/01/user/workspace/src/projects/sct/spinalcordtoolbox/math.py", line 316, in denoise_nlmeans
    sigma = dipy.denoise.noise_estimate.estimate_sigma(data_in)
  File "/scratch/01/user/workspace/src/projects/sct/contrib/tensorflow/lazy_loader.py", line 82, in __getattr__
    return getattr(module, item)
AttributeError: module 'dipy' has no attribute 'denoise'
```

joshuacwnewton:
## Description

This PR:

- Tests the "basic" test template proposed in #4637 (the tests will initially fail in order to demonstrate the bug)
- Fixes the lazy-loading to correctly import the lowest-most submodule.

Note: I did try to look for other instances of this kind of faulty lazy-loading, and I did notice this one:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/96eaaaa9739859df2584868a4d07455766420a52/spinalcordtoolbox/registration/algorithms.py#L33

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/96eaaaa9739859df2584868a4d07455766420a52/spinalcordtoolbox/registration/algorithms.py#L526-L527

However, surprisingly, `algo='dl'` (which uses voxelmorph) is already being tested, and there are no errors. I'm not sure why this doesn't fail with a "voxelmorph has no attribute 'networks'" error...

## Linked issues

Fixes #4636.


mguaypaq (approved):
Thanks for the fix, and for documenting this corner case of `LazyLoader`.

testing/cli/test_cli_sct_maths.py
```diff
@@ -1,18 +1,57 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.parametrize("ofolder", ["tmp_path", None])
+@pytest.mark.parametrize("args_to_test, fname_in, fname_gt", [
+    (["-denoise", "1"],       sct_test_path("t2", "t2.nii.gz"), None),  # TODO: Add gt
+    (["-otsu-median", "5,5"], sct_test_path("t2", "t2.nii.gz"), None),  # TODO: Add gt
+], ids=["-denoise", "-otsu-median"])
+def test_sct_maths_basic(args_to_test, fname_in, fname_gt, ofolder, tmp_path):
```
joshuacwnewton, line 22: Before I merge, I wanted to just quickly ask you @mguaypaq: How do you feel about the test structure proposed in #4637?

- Before: Individual tests for each argument, even if the overall structure of the tests is more or less the same.
- After: Shared structure between (some) tests, with the arguments being parametrized, leading to (hopefully) quicker tests to fill in missing gaps.

I feel like this works decently for `sct_maths`, and in fact I imagine it could even absorb some of the existing sct_maths tests (e.g. `-percent 95`, `-add 1`, `-symmetrize {0,1,2}`, etc.). I tried this now for kicks and found that `-symmetrize` works:

<details>

```python
def symmetrize_gt(fname_in, dim):
    im_in = Image(fname_in)
    im_out = im_in.copy()
    im_out.data = (im_in.data + np.flip(im_in.data, axis=int(dim))) / 2.0
    fname_out = os.path.join(os.getcwd(), f"symmetrize_{dim}.nii.gz")
    im_out.save(fname_out)
    return fname_out


@pytest.mark.parametrize("ofolder", ["tmp_path", None])
@pytest.mark.parametrize("args_to_test, fname_in, fname_gt", [
    (["-denoise", "1"],       sct_test_path("t2", "t2.nii.gz"), None),  # TODO: Add gt
    (["-otsu-median", "5,5"], sct_test_path("t2", "t2.nii.gz"), None),  # TODO: Add gt
    (["-symmetrize", "0"],    sct_test_path("t2", "t2.nii.gz"), symmetrize_gt(sct_test_path("t2", "t2.nii.gz"), 0)),
    (["-symmetrize", "1"],    sct_test_path("t2", "t2.nii.gz"), symmetrize_gt(sct_test_path("t2", "t2.nii.gz"), 1)),
    (["-symmetrize", "2"],    sct_test_path("t2", "t2.nii.gz"), symmetrize_gt(sct_test_path("t2", "t2.nii.gz"), 2)),
], ids=["-denoise", "-otsu-median", "-symmetrize 0", "-symmetrize 1", "-symmetrize 2"])
```
</details>

testing/cli/test_cli_sct_maths.py
```diff
@@ -1,18 +1,57 @@
...
+    # test all combinations of both (user-supplied and default) output filenames/folders
+    fname_out = "img_out.nii.gz"
+    path_out = (str(tmp_path) if ofolder == "tmp_path" else os.getcwd())
+    args.extend(["-o", os.path.join(path_out, fname_out) if ofolder else fname_out])
+
+    # run script
+    sct_maths.main(argv=args)
+
+    # test all output files exist
+    outfiles = [fname_out]
```
mguaypaq, line 37: I get that the template needs to account for multiple output files, but maybe when there's a single one we could inline `outfiles` and unroll the resulting single-iteration loops? Possibly also give a nice name to `os.path.join(path_out, fname_out)` which gets repeated 4 times?

(I'm saying this because this hardcoded value is part of the test body itself, rather than one of the parameters. By contrast, I think it's fine to keep the `if fname_gt` bit, even though `fname_gt` is currently always `None`, because that's a more contingent fact.)

testing/cli/test_cli_sct_maths.py
mguaypaq: Sure! I like this structure. And feel free to move existing tests into it if it's easy. (It's fine to have other tests too, if they don't easily fit the mould!)

mguaypaq (approved):

testing/cli/test_cli_sct_maths.py
joshuacwnewton: Absolutely!

PR #4640, opened 2024-10-01
Move "Labeling Conventions" page to "Concepts", while preserving tutorial using `.. include::` directive

### LINKED ISSUE #4632: Move "Labeling conventions" documentation page to a more visible location
Currently, it's buried in the tutorials. But, it could be moved to the "Key Concepts" section of the documentation instead. 

We would have to think about the flow of the tutorials, and making sure that we aren't losing information while people are completing the vertebral labeling tutorial. But, it should be a fairly quick fix nonetheless.

joshuacwnewton:
## Description

By using the [`.. include::`](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/reuse.html#include-a-shared-file) directive, we can have the content live in the "SCT Concepts" section, while still being able to insert the page into our vertebral labeling tutorial. :)

Rendered both:
  - [add link when docs render]

## Linked issues

Fixes #4632.


documentation/source/overview/concepts/labeling-conventions.rst
```diff
@@ -0,0 +1,19 @@
+Point labeling conventions
+##########################
+
+SCT accepts two different conventions for single-voxel (point) vertebral labels:
+
+1. **Vertebral levels**: For this type of label file, labels should be placed as though the vertebrae were projected onto the spinal cord, with the label centered in the middle of each vertebral level.
+2. **Intervertebral discs**: For this type of label file, labels should placed on the posterio  r tip of each disc.
```
mguaypaq, line 7: Typo:
```suggestion
2. **Intervertebral discs**: For this type of label file, labels should placed on the posterior tip of each disc.
```

mguaypaq (approved):
Thanks for putting this useful reference in a more obvious place!

PR #4646, opened 2024-10-09
Add clearer instructions to fix issue when $SCT_DIR isn't writeable during installation

### LINKED ISSUE #4623: Installer fails on macOS with write permission error for `$SCT_DIR`
### Description

I have a macOS; with zsh Terminal; when installing using the procedure described [here](https://spinalcordtoolbox.com/user_section/installation/mac.html#option-1-install-from-package-recommended), I downloaded the file under my Download/ folder; and then after running the install script I got permission issue: 

```console
thomasdagonneau@MacBook-Pro-de-Thomas-3 Downloads % bash install_sct-6.4_macos.sh


Checking OS type and version...

Darwin MacBook-Pro-de-Thomas-3.local 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:28:58 PST 2023; root:xnu-10002.81.5~7/RELEASE_X86_64 x86_64
ProductVersion:		14.3.1

Checking requirements...


OK!

Source files not present. Downloading source files (@ 6.4) to /var/folders/ny/ww_nlk897_x59csrbfysz19w0000gn/T/tmp.FiWu7c9SGJ/spinalcordtoolbox.
Cloning into '/var/folders/ny/ww_nlk897_x59csrbfysz19w0000gn/T/tmp.FiWu7c9SGJ/spinalcordtoolbox'...
Note: switching to '1f12b63e930354e018d637b9717c00ff35c6d6e0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false


SCT version ......... 6.4
Installation type ... package
Operating system .... osx (	14.3.1)
Shell config ........ /Users/thomasdagonneau/.zshrc

SCT will be installed here: [/Users/thomasdagonneau/sct_6.4]



Do you agree? [y]es/[n]o: y

ERROR: /Users/thomasdagonneau/sct_6.4 exists but does not have write permission.


Installation failed!

Please find the file "/Users/thomasdagonneau/Downloads/install_sct_log.txt",
--> https://forum.spinalcordmri.org/c/sct
```

This is surprising, given that the installation folder is under my home folder.

So; to circumvent,vent the issue I used sudo:

However; SCT cannot be launched; which is likely because the SCT PATH was added to `.bashrc`, which is not called in my ZSH terminal.

joshuacwnewton:
## Description

Minimizes the time spent debugging issues if users run into this issue again.

## Linked issues

Fixes #4623.

### DIFF
--- a/install_sct
+++ b/install_sct
@@ -516,7 +516,7 @@ mkdir -p "$SCT_DIR"
 # check if directory was created
 PERMISSION_ADVICE="\n\nMake sure that you own the folder you're writing to, and that the folder has write permissions.
 \nTo ensure this, either change the installation directory, or consider running the following commands:
-    sudo chown $USER \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\"
+    sudo chown \"$USER\" \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\"
     sudo chmod u+w \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\"
 if [[ -d "$SCT_DIR" ]]; then
   # check write permission
--- a/install_sct
+++ b/install_sct
@@ -514,13 +514,19 @@ done
 # Create directory
 mkdir -p "$SCT_DIR"
 # check if directory was created
+PERMISSION_ADVICE="\n\nMake sure that you own the folder you're writing to, and that the folder has write permissions.
+\nTo ensure this, either change the installation directory, or consider running the following commands:
+    sudo chown $USER $(dirname "$SCT_DIR")
+    sudo chown $USER $SCT_DIR
+    sudo chmod +w $(dirname "$SCT_DIR")
+    sudo chmod +w $SCT_DIR"
 if [[ -d "$SCT_DIR" ]]; then
   # check write permission
   if [[ ! -w "$SCT_DIR" ]]; then
-    die "ERROR: $SCT_DIR exists but does not have write permission."
+    die "ERROR: $SCT_DIR exists but does not have write permission.$PERMISSION_ADVICE"
   fi
 else
-  die "ERROR: $SCT_DIR cannot be created. Make sure you have write permission."
+  die "ERROR: $SCT_DIR cannot be created.$PERMISSION_ADVICE"
 fi
 
 # Copy files to destination directory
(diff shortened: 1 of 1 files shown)


install_sct
```diff
@@ -514,13 +514,19 @@ done
 # Create directory
 mkdir -p "$SCT_DIR"
 # check if directory was created
+PERMISSION_ADVICE="\n\nMake sure that you own the folder you're writing to, and that the folder has write permissions.
+\nTo ensure this, either change the installation directory, or consider running the following commands:
+    sudo chown $USER $(dirname "$SCT_DIR")
+    sudo chown $USER $SCT_DIR
+    sudo chmod +w $(dirname "$SCT_DIR")
+    sudo chmod +w $SCT_DIR"
```
mguaypaq, line 522: Both `chown` and `chmod` accept a list of files, so the commands can be combined. And it would be much safer to scope the write permissions change to only the owning user:
```suggestion
    sudo chown $USER \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\"
    sudo chmod u+w \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\""
```

mguaypaq (approved):
I find it a bit scary to suggest running `sudo` commands, especially when the circumstances suggest something very weird is going on, but these particular commands seem ok. It really seems like the best we can do.

install_sct
joshuacwnewton: I'm not sure if we need to escape the double-quotes? That seems to just result in the literal `"` character being part of the path, which ends up as an error for me:

```console
dot|joshua@mosaic:~$     sudo chmod u+w \"$(dirname "$SCT_DIR")\" \"$SCT_DIR\"
[sudo] password for joshua: 
chmod: cannot access '"/home/joshua/repos"': No such file or directory
chmod: cannot access '"/home/joshua/repos/spinalcordtoolbox"': No such file or directory
```

Apart from that, thank you for the tweaks to the command! <3

PR #4647, opened 2024-10-09
Rootlets-informed registration to the template

### LINKED ISSUE #4645: Register to template using rootlets segmentation
## Description

Now that we have a model to segment the dorsal nerve rootlets, and we have those rootlets in the PAM50 template, an idea would be to use those rootlets for registration to template.

See https://github.com/ivadomed/model-spinal-rootlets/issues/78

sandrinebedard:
#### GitHub

#### PR contents

## Description
This PR adresses #4645. Briefly, it used the dorsal spinal rootlets to improve registration to the template pipeline.

More precisely:
1. Uses spinal rootlets mid-points of native and PAM50 to perform straightening
2. Refines the Z alignement with BSplineSyn Z reg of masked PAM50 and masked straightened images with dilated rootlets.
3. Averages Z warping field slicewise to make sure it is symmetric in R-L direction.

Example command to use: 

**Note**: get the following PAM50 version for  PAM50 rootlets midpoints, please git checkout to: `sb/add-midpoints` in PAM50 [repository](https://github.com/spinalcordtoolbox/PAM50)

~~~
sct_register_to_template -i sub-amu01_space-other_T2w.nii.gz -s sub-amu01_space-other_T2w_seg.nii.gz -lrootlet sub-amu01_space-other_T2w_rootlets.nii.gz -ofolder reg_rootelts_testing
~~~

### Testing data: 
[testing_data.zip](https://github.com/user-attachments/files/17720655/testing_data.zip)

## Linked issues
Resolves #4645


spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -60,6 +60,7 @@ def __init__(self):
 step1 = Paramreg(step='1', type='imseg', algo='centermassrot', rot_method='pcahog')
 step2 = Paramreg(step='2', type='seg', algo='bsplinesyn', metric='MeanSquares', iter='3', smooth='1', slicewise='0')
 paramregmulti = ParamregMultiStep([step0, step1, step2])
+step_rootlets = Paramreg(step='1', algo='bsplinesyn', metric='MeanSquares', iter='30x20',shrink='4x2', smooth='1x0', slicewise='0', deformation='0x0x1', gradStep='0.1')
```
sandrinebedard, line 63: enlever smoothing

sandrinebedard: 8x4x2 --> baisser le nb d'itération

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +680,90 @@ def main(argv: Sequence[str]):
...
+                '-w', 'warp_curve2straightAffine.nii.gz',
+                '-x', 'nn',  #TODO to validate
+                '-v', '0',
+            ])
+            ftmp_rootlets= add_suffix(ftmp_rootlets, '_straightAffine')
+
+            # Dilate rootlets masks:
+            src_mask = Image(dilate(Image(ftmp_rootlets), size=2, shape='ball'), hdr=Image(ftmp_rootlets).hdr).save(add_suffix(ftmp_rootlets, '_dil'))
+            src_mask = add_suffix(ftmp_rootlets, '_dil')
+            dest_mask = Image(dilate(Image(ftmp_template_rootlets), size=2, shape='ball'), hdr=Image(ftmp_template_rootlets).hdr).save(add_suffix(ftmp_template_rootlets, '_dil'))
```
sandrinebedard, line 700: try other dilation factors maybe
or go back to rootlets

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +680,90 @@ def main(argv: Sequence[str]):
...
+            src_im = ftmp_data
+            dest_im = ftmp_template
+            scr_regStep = add_suffix(src_im, '_regStep' + str(step_rootlets.step))
+            metricSize = '4'  # TODO: maybe try 0
+            # TODO: condsider cropping before reg --> will maybe be faster
+            
+            cmd_rootlets = ['isct_antsRegistration',
+                '--dimensionality', '3',
+                '--transform', step_rootlets.algo + '[' + step_rootlets.gradStep
+                + ',26,0,3' + ']',
```
sandrinebedard, line 711: retester par défault de SCT

sandrinebedard (commented):
comments discussed with Julien

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -293,6 +303,13 @@ def main(argv: Sequence[str]):
     if arguments.l is not None:
         fname_landmarks = arguments.l
         label_type = 'body'
+    elif arguments.lrootlet is not None:
+        fname_rootlets = arguments.lrootlet
+        label_type = 'rootlet'
+        if arguments.ldisc is not None:
+            fname_landmarks = arguments.ldisc
+        else:
+            raise ValueError('Single disc label should be provided with rootlets labels.')
```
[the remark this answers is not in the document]
sandrinebedard: Sorry if this was not clear, yes the flag `-ldisc`, but I ma using it for rootlets midpoint as a bypass, but it may be confusing for the user. I am maybe thinking of computing the rootlets midpoints inside `sct_register_to_template` insetead, and the user would only need to provide `-lrootlet` does this seam okay?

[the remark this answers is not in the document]
joshuacwnewton: Yes, that makes perfect sense to me!

Though, I am curious, because for vertebral labeling + registration, [we extract the midpoint](https://spinalcordtoolbox.com/stable/user_section/tutorials/vertebral-labeling/extracting-specific-labels-for-registration.html) and [use the midpoints for the `-l` argument](https://spinalcordtoolbox.com/stable/user_section/tutorials/registration-to-template/template-registration/applying-the-registration-algorithm.html).

So, to make sure I understand: Rootlets-based registration requires rootlets seg + rootlets midpoint for all rootlets (hence automatic extraction), while vertebral level-based registration requires just the midpoints of 2 (or more) specific levels (hence using `sct_label_utils` to pick specific levels).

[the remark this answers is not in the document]
sandrinebedard: > So, to make sure I understand: Rootlets-based registration requires rootlets seg + rootlets midpoint for all rootlets (which is why automatic extraction would be beneficial), while vertebral level-based registration requires just the midpoints of 2 (or more) specific levels (hence using `sct_label_utils` to pick specific levels).

yes exactly!

sandrinebedard:
Hey @joshuacwnewton , the PR is ready for you to review whenever you have time!

sandrinebedard:
yes no problem @joshuacwnewton !

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -60,6 +60,7 @@ def __init__(self):
 step1 = Paramreg(step='1', type='imseg', algo='centermassrot', rot_method='pcahog')
 step2 = Paramreg(step='2', type='seg', algo='bsplinesyn', metric='MeanSquares', iter='3', smooth='1', slicewise='0')
 paramregmulti = ParamregMultiStep([step0, step1, step2])
+step_rootlets = Paramreg(step='1', metric='CC', iter='6x6x3', shrink='8x4x2', smooth='0x0x0', slicewise='0', deformation='0x0x1', gradStep='0.1')
```
sandrinebedard, line 63: maybe add a flag -param-rootlets to modify this

jcohenadad, line 63: consider adding a flag `-param-rootlets` in the CLI

sandrinebedard, line 63:
```suggestion
step_rootlet = Paramreg(step='1', metric='CC', iter='6x6x3', shrink='8x4x2', smooth='0x0x0', slicewise='0', deformation='0x0x1', gradStep='0.1')
```

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -630,6 +665,7 @@ def main(argv: Sequence[str]):
             '-o', add_suffix(ftmp_data, '_straightAffine'),
             '-d', ftmp_template,
             '-w', 'warp_curve2straightAffine.nii.gz',
+            '-x', 'linear',
```
joshuacwnewton, line 668: I notice the batch_processing.sh results have changed. Skimming the PR, I think 99% of your changes would have no effect on the other registration types, but I noticed the addition of `-x linear` here. 

The default value is `spline`... is there a reason for this change? I couldn't find any reasoning in the source commit from #4681: https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/16428885c1d4005ec9f66cf82b5f17070ee7ad20

sandrinebedard: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/46c682c7a86cc65154afa8ecf06979cfb4cee226

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -170,6 +171,15 @@ def get_parser():
             Each label is a single voxel, which value corresponds to the spinal level (e.g.: 2 for spinal level 2). If you are using more than 2 labels, all spinal levels covering the region of interest should be provided (e.g., if you are interested in levels C2 to C7, then you should provide spinal level labels 2,3,4,5,6,7)."
         """)  # noqa: E501 (line too long)
     )
+    optional.add_argument(
+        '-lrootlet',
+        metavar=Metavar.file,
+        help=textwrap.dedent("""
+            Dorsal nerve rootlets segmentation from C2 to C8  Example: `anat_rootlets.nii.gz`
```
joshuacwnewton, line 178: I am a bit confused by the specification of "C2-C8". Is this because the PAM50 template only has those rootlets?

If so, maybe we should mention that in more detail. ("Only labels within the range C2-C8 are supported. If labels outside this range are provided, they will be ignored.") or something like that.

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -717,10 +889,6 @@ def main(argv: Sequence[str]):
         if level_alignment:
             dimensionality = len(Image("data.nii").hdr.get_data_shape())
             cmd = ['isct_ComposeMultiTransform', f"{dimensionality}", 'warp_template2anat.nii.gz', '-R', 'data.nii', 'warp_straight2curve.nii.gz', warp_inverse]
-            status, output = run_proc(cmd, verbose=verbose, is_sct_binary=True)
-            if status != 0:
-                raise RuntimeError(f"Subprocess call {cmd} returned non-zero: {output}")
-
```
joshuacwnewton, line 723: Thanks for this simplification!

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +679,142 @@ def main(argv: Sequence[str]):
         ])
         ftmp_seg = add_suffix(ftmp_seg, '_straightAffine')
 
+        # Register spinal rootlets to template:
+        if label_type == 'rootlet':
+
```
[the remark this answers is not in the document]
joshuacwnewton: I am also a bit confused. I thought that this PR would use the rootlets labels to inform the registration of the anat image to the template. But after the "if" block, we continue with normal registration to template (lines 872-874).

So, is rootlets registration actually entirely separate from the normal registration? (i.e. we don't use the rootlets as `ftmp_label` in `register_wrapper` or anything?)

Maybe it will be more clear to me once this is turned into a function, since the input and output variables will be clearer.

spinalcordtoolbox/scripts/sct_register_to_template.py
joshuacwnewton, line 1: Would you be able to add a test for this? (i.e. calling `sct_register_to_template` with the `-lrootlets` argument?)

joshuacwnewton (changes requested):
The first pass of my review is mainly focused on helping make the PR a bit easier to understand for me. :sweat_smile:

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: Yes it is because the automatic rootlets segmentation is currently limited to C2-C8 (as the labels in the PAM50) (in sct_deepseg)

spinalcordtoolbox/scripts/sct_register_to_template.py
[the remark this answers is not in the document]
sandrinebedard: Maybe this figure helps: 
figure_methods

So the main difference is that the center-of-mass of the rootlets is aligned with the center-of mass of the rootlets instead of aligning the discs and a refinement is added after (steps 4 & 5) and then the last steps of the registration to template are the same than the normal pipeline (step 6), is it clearer?

`ftmp_label` here corresponds to the center-of-mass of the rootlet segmetation (that is created in the script, which uses the rootlet segmentation)

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: added in : https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/01721239c686904d3c99ded1d22759f9623077a1

spinalcordtoolbox/math.py
```diff
@@ -402,6 +402,42 @@ def symmetrize(data, dim):
     return data_out
 
 
+def slicewise_mean(data, dim, exclude_zeros=False):
+    """
+    Symmetrize data along specified dimension. Does not count zeros in the mean.
+    :param data: numpy.array 3D data.
+    :param dim: dimension of array to symmetrize along.
```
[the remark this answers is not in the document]
sandrinebedard: we can keep only slicewise mean if it is clearer

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: yes will do!, Note that I provided example data and command in the PR body too!

sandrinebedard: The test is now added: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/1296390bdadf91580ed1a7e750f3176911a3125d

`sct_testing_data` needs to be updated --> https://github.com/spinalcordtoolbox/sct_testing_data/pull/14

spinalcordtoolbox/math.py
```diff
@@ -402,6 +402,29 @@ def symmetrize(data, dim):
...
+
+    :return data_out: slicewise averaged data
+    """
+    if exclude_zeros:
+        data[data == 0] = np.nan
+    data_out = np.zeros_like(data)
+    for slices in range(data.shape[dim]):
+        idx_to_slice = [slice(None)] * data.ndim
+        idx_to_slice[dim] = slices
```
joshuacwnewton, line 418: I'm sorry, I think I led you wrong here! I'm getting the following error when I run the test:

```python
    if np.isnan(data[idx_to_slice]).all():
# IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) 
# and integer or boolean arrays are valid indices
```

It looks like the syntax has changed, and numpy now requires a tuple of slice objects rather than a list. So we need to add:

```suggestion
        idx_to_slice = [slice(None)] * data.ndim
        idx_to_slice[dim] = slices
        idx_to_slice = tuple(idx_to_slice)  # requirement of numpy indexing
```

joshuacwnewton (commented):
Once I added the fix below, I was able to test your PR along with https://github.com/spinalcordtoolbox/PAM50/pull/31 and https://github.com/spinalcordtoolbox/sct_testing_data/pull/14, and the rootlets midpoints are perfectly aligned!

<details>

ezgif-7-13fad47835
</details>

Based on this, I'm happy to approve and merge the other two PRs, and create new PAM50/sct_testing_data releases. :)

Then I'll move onto reviewing this PR a little further!

spinalcordtoolbox/scripts/sct_register_to_template.py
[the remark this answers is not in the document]
joshuacwnewton: > `ftmp_label` here corresponds to the center-of-mass of the rootlet segmetation (that is created in the script, which uses the rootlet segmentation)

Ahhh, I totally missed that `-lrootlet` gets assigned to `fname_landmarks`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/01721239c686904d3c99ded1d22759f9623077a1/spinalcordtoolbox/scripts/sct_register_to_template.py#L306-L309

Which then becomes `ftmp_label`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/01721239c686904d3c99ded1d22759f9623077a1/spinalcordtoolbox/scripts/sct_register_to_template.py#L421

That makes much more sense now. Thank you!!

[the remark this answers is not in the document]
joshuacwnewton: Just to clarify one more thing:

- By using `-lrootlets` for `ftmp_label`, we are using the rootlets midpoints as the landmarks for the "coarse slicewise (SI-axis) alignment" step of the usual cord-based registration to template (as you described above).
- However, in order to register the _rootlets segmentations_ to the template, we need an entirely separate registration procedure (200 lines of new registration code starting with `if label_type == 'rootlet':`)
    - The reason that this needs to be a separate step (compared to the similar "GM-informed registration") is that we do not get the benefit of a segmentation like the WM, which encodes the shape of both GM and SC.
    - And, because we essentially perform 2 different registrations, that means that we end up using two separate warping fields, too? ("rootlets -> PAM50_rootlets" and "anat -> PAM50")? 
    - Looking at the outputs, it seems like the "rootlets -> PAM50_rootlets" warping fields (`step10Warp_zmean.nii.gz`) are not output to the user, or used in regular regisration at all...

I wonder if that will be confusing to the user, as the rootlets will get registered in a kind of "opaque" way, without access to these "rootlets-specific" warping fields.

Is there no way to create a single warping field that will accurately register both the cord and the rootlets (similar to the GM-informed registration)? This would simplify the contribution greatly... :thinking:

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +679,142 @@ def main(argv: Sequence[str]):
...
+                'isct_ComposeMultiTransform',
+                str(dimensionality),
+                'warp_curve2straightAffine.nii.gz',
+                '-R', 'template.nii',
+                'step10Warp_zmean.nii.gz',
+                'warp_curve2straightAffine.nii.gz',
+            ]
+            status, output = run_proc(cmd, verbose=verbose, is_sct_binary=True)
+            if status != 0:
+                raise RuntimeError(f"Subprocess call {cmd} returned non-zero: {output}")
```
sandrinebedard, line 784: @joshuacwnewton Here I concatenate the warping field `step10Warp_zmean.nii.gz` to `warp_curve2straightAffine.nii.gz` but the output name is also `warp_curve2straightAffine.nii.gz` so that it is concatenanted with the last 2 steps of regostration to template and include in `warp_anat2template.nii.gz` at the end!

spinalcordtoolbox/scripts/sct_register_to_template.py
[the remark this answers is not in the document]
sandrinebedard: The rootlets registration part is only to further align the SI axis using the rootlets, but then the warping fields are concatenated to the last 2 steps of the registration, the goal of using the rootlets segmentation is only to better align the z axis, but this doesn't affect the cord at all

> And, because we essentially perform 2 different registrations, that means that we end up using two separate warping fields, too? ("rootlets -> PAM50_rootlets" and "anat -> PAM50")?

the warping field `warp_anat2template` and `warp_template2anat` will include the rootlet step

The lines under `if label_type == 'rootlet'` only replace disc alignement or vertebrae alignement step

`Is there no way to create a single warping field that will accurately register both the cord and the rootlets (similar to the GM-informed registration)? This would simplify the contribution greatly, and reduce potential variability between the two separate registrations... 🤔`

This is exactly what you get with the warp_anat2template no? 

(see line 772 and with comment)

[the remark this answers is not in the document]
joshuacwnewton: I totally missed the concatenation that you highlighted in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647#discussion_r1890748283.

My bad!! Thank you so much for explaining.

spinalcordtoolbox/scripts/sct_register_to_template.py
joshuacwnewton: Since landmark registration is called "step #0", and we perform rootlets segmentation immediately after, it's like we're essentially inserting a new step in-between step 0 and step 1. 

So, maybe we should make it a requirement that if `-lrootlets` is specified, we only apply the rootlets-based reg if the user specifies `-param step=1,type=rootlets,[...]`. That way, `-param` can be re-used instead of `-param-rootlet`, and it also makes it 100% explicit to the user that the rootlets-based registration will be performed immediately after step=0.

(Right now, it's kind of ambiguous, and the user has no control over when or how the rootlets reg will be inserted into the substeps.)

spinalcordtoolbox/math.py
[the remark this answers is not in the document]
joshuacwnewton: I think I just don't understand the how a slicewise_mean == symmetrize... I'll take a moment to trace the function and make sure I understand how the array changes. Ignore this for now, I'll fix my own understanding. :sweat_smile:

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +679,142 @@ def main(argv: Sequence[str]):
...
+            printv(output, verbose)
+            if status != 0:
+                raise RuntimeError(f"Subprocess call {cmd_rootlet} returned non-zero: {output}")
+            printv('\nApply transformation after rootlets adjustment...', verbose)
+            # Average perslice warping field
+            cmd_split = ['sct_image', '-i', 'step10Warp.nii.gz', '-mcs']
+            status, output = run_proc(cmd_split, verbose, is_sct_binary=True)
+            printv(output, verbose)
+            # Compute slicewise mean in Z to ensure symmetry
+            img = Image('step10Warp_Z.nii.gz')
```
valosekj, line 731: Maybe we could an extra print to inform user that we're computing slicewise mean in Z. It took me some time to understand why we do `sct_image -mcs` and `sct_image -omc` as I did not imidietily noticed the difference `step10Warp_Z.nii.gz` vs `step10Warp_Z_mean.nii.gz`

image


```suggestion
            # Compute slicewise mean in Z to ensure symmetry
            printv('\nComputing slicewise mean in Z to ensure symmetry....', verbose)
            img = Image('step10Warp_Z.nii.gz')
```

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -643,6 +679,142 @@ def main(argv: Sequence[str]):
...
+            if status != 0:
+                raise RuntimeError(f"Subprocess call {cmd} returned non-zero: {output}")
+
+            # INVERSE WARPING FIELD
+            # Average perslice warping field
+            cmd_split = ['sct_image', '-i', 'step10InverseWarp.nii.gz', '-mcs']
+            status, output = run_proc(cmd_split, verbose, is_sct_binary=True)
+            printv(output, verbose)
+            # Compute slicewise mean in Z to ensure symmetry
+            img = Image('step10InverseWarp_Z.nii.gz')
```
valosekj, line 792: idem; see https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/files#r1890792804

```suggestion
            # Compute slicewise mean in Z to ensure symmetry
            printv('\nComputing slicewise mean in Z to ensure symmetry....', verbose)
            img = Image('step10InverseWarp_Z.nii.gz')
```

spinalcordtoolbox/math.py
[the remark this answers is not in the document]
valosekj: maybe we can be consistent with the `sct_maths.py` arg help:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/03fec7cf92cdb873a687e491ce87a24500fa4108/spinalcordtoolbox/scripts/sct_maths.py#L203

```suggestion
def slicewise_mean(data, dim, exclude_zeros=False):
    """
    Compute slicewise mean the specified dimension. Zeros are not inlcuded in the mean.
    :param data: numpy.array 3D data.
    :param dim: dimension of array to symmetrize along.
```

spinalcordtoolbox/scripts/sct_register_to_template.py
valosekj, line 178: maybe we can now drop "C2 to C8" from this line as the following line details the C2-C8 range
```suggestion
            Dorsal nerve rootlets segmentation. Example: `anat_rootlets.nii.gz`
```

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -369,13 +389,17 @@ def main(argv: Sequence[str]):
     printv('  Path template:        ' + path_template, verbose)
     printv('  Remove temp files:    ' + str(param.remove_temp_files), verbose)
 
+    # Compute center-of-mass of the rootlets segmentation:
+    if label_type == 'rootlet':
+        fname_rootlets_points = 'rootlets_mid.nii.gz'
+        fname_landmarks = sct_labels.cubic_to_point(Image(fname_rootlets)).save(fname_rootlets_points)
```
valosekj, line 395: Finding the rootlets center of mass might be strongly influenced by the quality of the rootlets segmentation. If we had some missing rootlets, for example, due to lower contrast, artifact, or canal stenosis, the center of mass might even be shifted outside the cord. Consider the following example where the left C2 rootlets are entirely missing (I removed it manually in FSLeyes), then the center of mass (the blue point is outside the cord).

image

Debugging:

```console
sct_label_utils -i sub-amu01_space-other_T2w_rootlets.nii.gz -cubic-to-point -o sub-amu01_space-other_T2w_rootlets_midpoints.nii.gz
sct_label_utils -i sub-amu01_space-other_T2w_rootlets_missing_left_C2.nii.gz -cubic-to-point -o sub-amu01_space-other_T2w_rootlets_missing_left_C2_midpoints.nii.gz
```

[sub-amu01_space-other_T2w_rootlets_missing_left_C2.nii.gz](https://github.com/user-attachments/files/18188861/sub-amu01_space-other_T2w_rootlets_missing_left_C2.nii.gz)
[sub-amu01_space-other_T2w_rootlets.nii.gz](https://github.com/user-attachments/files/18188862/sub-amu01_space-other_T2w_rootlets.nii.gz)

valosekj: To avoid this, perhaps we could consider centering the rootlets center of mass to the spinal cord centerline?

sandrinebedard: hmm, I though I was projecting on the centerline, let me check!

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: yes we can do that!

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: Good idea!

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: Agreed!

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: Ahhh, actually, all the landmarks are projected on the spinal cord segmentation:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b4d8faeb6bfd79199a371dae70faf145c33618c5/spinalcordtoolbox/scripts/sct_register_to_template.py#L459-L462

valosekj: I see! I got probably confused by the `rootlets_mid.nii.gz` temp file:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b4d8faeb6bfd79199a371dae70faf145c33618c5/spinalcordtoolbox/scripts/sct_register_to_template.py#L394-L396

which does _**not**_ have the landmarks projected on the spinal cord: 

image

Command I used:

```console
sct_register_to_template -i sub-amu01_space-other_T2w.nii.gz -s sub-amu01_space-other_T2w_seg.nii.gz -lrootlet sub-amu01_space-other_T2w_rootlets_missing_left_C2.nii.gz  -ofolder reg_rootelts_testing
```

sandrinebedard: yes exaclty! I have a todo to remove this file to avoid confusion! or move to tmp

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: done in [183b69b](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/183b69b4164c323dbabbcb3dd8f29d539a10501e)

joshuacwnewton (commented):
Apart from getting the test working (I need to [merge the PRs and create the releases](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647#pullrequestreview-2512579086), doing that now!), my only real feedback is regarding the code organization.

I am curious if we can re-use the existing registration functions (`register_wrapper`, `register`). If so, it might help us de-duplicate some of the code, and also allow the user to customize the step using the existing `-param`. 

See this discussion: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647#discussion_r1890788682

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: I will dig into this!

> This would allow the user to customize not only the parameters, but the step order too

I would not want the user to change the order of the step in the case of registration to the template I think!

sandrinebedard:
> I am curious if we can re-use the existing registration functions (`register_wrapper`, `register`). If so, it might help us de-duplicate some of the code, and also allow the user to customize the step using the existing `-param`.

@joshuacwnewton I have moved everything to call `register_wrapper` and `register` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/e25cc1fa4619c194efcb278925bf7c1975d3b791 ,  https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/804f2ed7c7ef3a5acd35031520e6d3e4c4464e58 , https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/c2f01f446944b39a19f44aa49362b976656db1ac, https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/1d77302dc3ecea3a6fe0a968bc92bb45160790bb

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -341,6 +339,68 @@ def register_step_label(src, dest, step, verbose=1):
...
+    out.data = slicewise_mean(out.data, 2)
+    out.save(image.add_suffix(warp_inverse_out, '_Z_mean'))
+    # Merge warp back together
+    cmd_split = ['sct_image', '-i',
+                 image.add_suffix(warp_inverse_out, '_X'), image.add_suffix(warp_inverse_out, '_Y'), image.add_suffix(warp_inverse_out, '_Z_mean'),
+                 '-omc', '-o', image.add_suffix(warp_inverse_out, '_zmean')]
+    status, output = run_proc(cmd_split, verbose, is_sct_binary=True)
+    warp_inverse_out = image.add_suffix(warp_inverse_out, '_zmean')
+
+    return warp_forward_out, warp_inverse_out
```
joshuacwnewton, line 403: I guess technically since we're repeating identical steps twice, we could just loop over the filenames instead?

```suggestion
    output_warping_fields = register_step_ants_registration(src=src_im,
                                                            dest=dest_im,
                                                            step=step,
                                                            masking=['-x', '[' + dest_mask + ',' + src_mask + ']'],
                                                            ants_registration_params=ants_registration_params,
                                                            padding=padding,
                                                            metricSize=metricSize,
                                                            verbose=1)

    # FORWARD WARPING FIELD
    logger.info('\nApply transformation after rootlets adjustment...')
    for i in range(len(output_warping_fields)):
        fname_warp = output_warping_fields[i]
        # Average perslice warping field
        cmd_split = ['sct_image', '-i', fname_warp, '-mcs']  # Split warping field in x, y, z
        status, output = run_proc(cmd_split, verbose, is_sct_binary=True)
    
        # Compute slicewise mean in Z to ensure symmetry
        logger.info('\nComputing slicewise mean in Z to ensure symmetry....')
        img = image.Image(image.add_suffix(fname_warp, '_Z'))
        out = img.copy()
        out.data = slicewise_mean(out.data, 2)
        out.save(image.add_suffix(fname_warp, '_Z_mean'))
        # Merge warp back together
        cmd_split = ['sct_image', '-i',
                     image.add_suffix(fname_warp, '_X'), image.add_suffix(fname_warp, '_Y'), image.add_suffix(fname_warp, '_Z_mean'),
                     '-omc', '-o', image.add_suffix(warp_forward_out, '_zmean')]
        status, output = run_proc(cmd_split, verbose, is_sct_binary=True)
        output_warping_fields[i] = image.add_suffix(fname_warp, '_zmean')

    return output_warping_fields
```

spinalcordtoolbox/registration/core.py
```diff
@@ -161,9 +161,12 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
             src = ['src_label.nii']
             dest = ['dest_label_RPI.nii']
             interp_step = ['nn']
+        elif step.type == 'rootlet':
+            src = ['src.nii', 'src_label.nii']
+            dest = ['dest_RPI.nii', 'dest_label_RPI.nii']
+            interp_step = ['spline']  # Maybe do or linear?
```
joshuacwnewton, line 167: What are the pros and cons of `spline` vs. `linear`? o:

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -341,6 +339,68 @@ def register_step_label(src, dest, step, verbose=1):
     return warp_forward_out, warp_inverse_out
 
 
+def register_rootlet(src, dest, step, ants_registration_params, metricSize, padding, verbose=1):
+    """
+    """
```
joshuacwnewton, line 346: I was going to suggest writing a docstring, but it looks like none of the functions in this file have docstrings. :sweat_smile: 

Perhaps we can do them all at once in a follow-up PR.

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -170,6 +171,15 @@ def get_parser():
...
     )
+    optional.add_argument(
+        '-lrootlet',
+        metavar=Metavar.file,
+        help=textwrap.dedent("""
+            Dorsal nerve rootlets segmentation. Example: `anat_rootlets.nii.gz`
+            Only labels within the range C2-C8 are supported. If labels outside this range are provided, they will be ignored.
+            Each value corresponds to the spinal level (e.g.: 2 for spinal level 2). If you are using more than 2 labels, all spinal levels covering the region of interest should be provided (e.g., if you are interested in levels C2 to C7, then you should provide spinal level labels 2,3,4,5,6,7)."
+        """)  # noqa: E501 (line too long)
+    )
```
joshuacwnewton, line 182: We might want to add a summary of how the current `type=rootlets` behavior works:

```suggestion
    optional.add_argument(
        '-lrootlet',
        metavar=Metavar.file,
        help=textwrap.dedent("""
            Dorsal nerve rootlets segmentation. Example: `anat_rootlets.nii.gz`
            Only labels within the range C2-C8 are supported. If labels outside this range are provided, they will be ignored.
            Each value corresponds to the spinal level (e.g.: 2 for spinal level 2). If you are using more than 2 labels, all spinal levels covering the region of interest should be provided (e.g., if you are interested in levels C2 to C7, then you should provide spinal level labels 2,3,4,5,6,7).
            By default, the rootlets labels will be used in 2 places:
            
              - step=0: The center of mass of each label will be used as landmarks for SI-axis slice alignment with the rootlets of the template. 
              - step=1: ANTs BSplineSyn adjustment using the axial shape of the rootlets combined with the SC segmentation.
            
            If you are customizing the `-param` argument, you can choose to omit `step=1`, or choose to manually specify `step=1` yourself. We recommend starting with the following `step=1` parameters and adjusting as needed:
            
                step=1,type=rootlet,algo=bsplinesyn,metric=CC,iter=6x6x3,shrink=8x4x2,smooth=0x0x0,slicewise=0,deformation=0x0x1,gradStep=0.1
        """)  # noqa: E501 (line too long)
    )
```

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -324,6 +338,11 @@ def main(argv: Sequence[str]):
         # add user parameters
         for paramStep in arguments.param:
             paramregmulti.addStep(paramStep)
+    elif label_type == 'rootlet':
+        # Add rootlet steps before step 1 and 2
+        step1.step = '2'
+        step2.step = '3'
+        paramregmulti = ParamregMultiStep([step0, step_rootlet, step1, step2])
```
joshuacwnewton, line 345: At first I was a little wary of manually inserting this step on behalf of the user, but I think it makes sense as long as we include the summary of what happens when `-lrootlets` is specified.

spinalcordtoolbox/scripts/sct_register_to_template.py
```diff
@@ -699,7 +755,7 @@ def main(argv: Sequence[str]):
             paramregmulti.steps[key].rot_dest = 0
         fname_src2dest, fname_dest2src, warp_forward, warp_inverse = register_wrapper(
             ftmp_data, ftmp_template, param, paramregmulti, fname_src_seg=ftmp_seg, fname_dest_seg=ftmp_template_seg,
-            same_space=True)
+            fname_src_label=ftmp_rootlet, fname_dest_label=ftmp_template_rootlets, same_space=True)
```
joshuacwnewton, line 763: I understand that `ftmp_rootlet` and `ftmp_template_rootlets` will be `None` if `-lrootlets` isn't passed. But, still, if someone is just skimming through the code, it might be a bit confusing to see these variables being passed unconditionally. ("I didn't use rootlets! Why is this being passed?" etc.)

So, maybe to better convey that the rootlets are completely optional, we could change it to this?

```suggestion
            fname_src_label=(ftmp_rootlet if label_type == 'rootlet' else None), 
            fname_dest_label=(ftmp_template_rootlets if label_type == 'rootlet' else None), 
            same_space=True)
```

It is technically redundant, but it at least self-documents that these options will only be used if `label_type == 'rootlet'`.

joshuacwnewton (changes requested):
This PR is looking great!! I'm so glad that we were able to re-use the existing `register_wrapper` and deduplicate some of the steps involving the warping fields.

I think the overall structure of the PR is ready, and so my last few comments are just small suggestions for individual code blocks.

I also cleaned up some of the small whitespace changes that weren't critical to the PR (just to keep the diff small and focused on your changes).

spinalcordtoolbox/registration/core.py
```diff
@@ -358,6 +362,19 @@ def register(src, dest, step, param):
             verbose=param.verbose,
         )
 
+    elif step.type in ['rootlet']:
+        if step.algo:
+            printv(f"Parameter 'algo={step.algo}' has no effect for 'type=label' registration.", type='warning')
```
joshuacwnewton, line 367: Ah! I think this copy-pasted warning is not relevant, since technically the user could customize the `algo` for now.

spinalcordtoolbox/registration/algorithms.py
sandrinebedard: agreed!

spinalcordtoolbox/registration/core.py
sandrinebedard: I think usually for images we use spline interpolation, which I would not do for segmentations or labels

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: sounds good!

spinalcordtoolbox/registration/core.py
sandrinebedard: now removed in [1e4d052](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/1e4d052fe67c54f7cc3b95670ef333803bd639f6)

spinalcordtoolbox/scripts/sct_register_to_template.py
sandrinebedard: I modified a bit the description in [2c0bd0e](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4647/commits/2c0bd0ea901d8b538f537537d570b33c7c04886a)

sandrinebedard:
> This PR is looking great!! I'm so glad that we were able to re-use the existing `register_wrapper` and deduplicate some of the steps involving the warping fields.

Thank you @joshuacwnewton for all your feedback! I struggled a bit with this PR, sorry!

I addressed your last comments I think!

joshuacwnewton (approved):
Thank you so much for taking the time to respond to all my feedback! I'm very happy with how it turned out, and happy that we got to re-use the existing `-param` argument. :)

Apart from the one testing issue, this LGTM!

testing/cli/test_cli_sct_register_to_template.py
```diff
@@ -174,3 +174,29 @@ def test_sct_register_to_template_more_than_2_labels(tmp_path, labels_discs):
...
+def test_sct_register_to_template_rootlets(tmp_path):
+    """
+    Test registration with rootlets segmentation.
+    """
+    sct_register_to_template.main(argv=['-i', sct_test_path('t2', 't2.nii.gz'),
+                                        '-s', sct_test_path('t2', 't2_seg-manual.nii.gz'),
+                                        '-lrootlet', sct_test_path('t2', 't2_seg-deepseg_rootlets-manual.nii.gz'),
+                                        '-t', sct_test_path('template'),
+                                        '-param', 'step=1,type=seg,algo=centermassrot,metric=MeanSquares:'
+                                                  'step=2,type=seg,algo=bsplinesyn,iter=0,metric=MeanSquares',
```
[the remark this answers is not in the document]
sandrinebedard: ahh I see

sandrinebedard:
It would be nice to also save the midpoints file projected on the centerline when `-v 2` is specified

PR #4650, opened 2024-10-18
Add an interactive check before deleting output directory when downloading data

### LINKED ISSUE #4562: should sct_download_data remove output folder without asking permission?
While testing #4561, I ran this command from my SCT directory:

~~~
julien-macbook:~/code/spinalcordtoolbox $ sct_download_data -d template-dog -o data/

--
Spinal Cord Toolbox (git-rb/4558-add-dog-template-b61e237ba38f48d71aab7f6bf3fb006df09962f9)

sct_download_data -d template-dog -o data/
--

Removing existing destination folder 'data'
Trying URL: https://github.com/spinalcordtoolbox/template-dog/releases/download/r20240709/templatedog_r20240709.zip
Downloading: templatedog_r20240709.zip
Status: 100%|██████████████████████████████████████████████████| 20.6M/20.6M [00:01<00:00, 14.1MB/s]
Creating temporary folder (/var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-07-15_16-21-01_install-data_6xh1tyip)
Unzip data to: /var/folders/5f/6n99hhyj5_g5cv_1bbx1q7cr0000gn/T/sct_2024-07-15_16-21-01_install-data_6xh1tyip
Copying data to: data
Removing temporary folders...
Done!
~~~

I thought that the dog template would be created _as a new folder inside_ `$SCT_DIR/data`, but instead, the entire folder was removed. This is not a great behaviour and could lead to frustrated users.

Shouldn't we at least be asking for a yes/no prompt before removing the folder?

joshuacwnewton:
## Description

I've followed the logical suggestions from @mguaypaq in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4562#issuecomment-2229691874, however I've added an extra check for non-emptiness, since an empty directory seems OK to remove by default (seeing as it's just going to get remade anyway).

- ~~Note that my changes only provide protection for callers of `install_named_dataset`, meaning that `install_data` by itself is not protected. But, `install_named_dataset` is the only place we do the "default vs. non-default location" check, so I think it's the only place we *can* add this safety valve.~~ EDIT: See below.
   - ~~The only other usage of `install_data` is `install_model` (`sct_deepseg`), however, and sct_deepseg does not allow specifying a custom output directory when installing. So, I think this is suitable?~~
- One other note is that I find it a bit weird to reference CLI args from API functions. Is that a hint that this check should actually be performed in the `sct_download_data` CLI script? (i.e. checking `arguments.o` and `arguments.k` in the layer above)
   - One argument for this is that we probably don't want interactive prompts to occur at an API level. EDIT: In writing this out, I've actually convinced myself, aha. Revising now!  

## Testing

image

## Linked issues

Fixes #4562.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_download_data.py
+++ b/spinalcordtoolbox/scripts/sct_download_data.py
@@ -5,6 +5,7 @@
 # Copyright (c) 2015 Polytechnique Montreal <www.neuro.polymtl.ca>
 # License: see the file LICENSE
 
+import os
 import sys
 from typing import Sequence
 import textwrap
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_download_data.py
```diff
@@ -72,7 +73,28 @@ def main(argv: Sequence[str]):
...
+
+        # Make sure we don't accidentally overwrite a critical user folder
+        if dest_folder is not None and os.path.isdir(dest_folder) and os.listdir(dest_folder) and not keep:
+            printv(f"Output directory '{dest_folder}' exists and is non-empty. Contents will be erased.",
+                   type="warning")
+
+            while True:
+                answer = input("Do you wish to overwrite this directory? ([Y]es/[N]o): ")
+                if answer.lower() in ["y", "yes"]:
+                    break  # keep = False
```
mguaypaq, line 87: Thanks for the comment, it really helps readability.

mguaypaq (approved):
Looks good, thanks for making this dangerous behaviour safer!

PR #4651, opened 2024-10-18
Remove `/en/latest/` slugs from `spinalcordtoolbox.com` URLs

joshuacwnewton:
## Description

The broken link checker caught these!

SCT's RTD config is _not_ set to the "Multiple versions with translations" option, because we don't have any non-English translations.

image

Because of this, the presence of `/en/` breaks our URLs, and so we rely on a custom redirect just to keep these links alive. 

image

However, just for future-proofing, it would be better to just use our "base" url, then let RTD's built-in redirects do their magic (i.e. to stable instead of latest). This way, when we change from having "latest" as our default to "stable" (like we've just done recently), the links update appropriately without needing to change the actual slugs.

(Note: Most of the URLs in our project are the "base" urls, but a few had `/en/latest/`, so this PR unifies our docs.)

## Linked issues

N/A

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ For macOS and Linux users, the simplest way to install SCT is to download `the l
 
    ./install_sct
 
-For more complex installation setups (Windows users, Docker, FSLeyes integration), see the `Installation <https://spinalcordtoolbox.com/en/latest/user_section/installation.html>`_ page.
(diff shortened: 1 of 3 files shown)


mguaypaq (approved):
Great attention to detail, thanks!

PR #4631, opened 2024-09-18
Change how midsagiital slice is determined for even-width cord mask

valosekj:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/onboarding/software-development.html#pr-labels) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR changes how `sct_analyze_lesion` obtains the midsagittal slice based on the spinal cord segmentation mask.
 
### `master` branch
The original code under the `master` branch uses `int()` effectively rounding any number down. This results in the same midsagittal slice regardless of whether the number of sagittal slices is odd or even.

<details><summary>examples</summary>

3 vs 4 slices:

`int(np.mean([np.min([8, 9, 10]), np.max([8, 9, 10])]))` --> 9
`int(np.mean([np.min([8, 9, 10, 11]), np.max([8, 9, 10,11])]))` --> 9

5 vs 6 slices:

`int(np.mean([np.min([8, 9, 10, 11, 12]), np.max([8, 9, 10, 11, 12])]))` --> 10
`int(np.mean([np.min([8, 9, 10, 11, 12, 13]), np.max([8, 9, 10, 11, 12, 13])]))` --> 10

</details>

### This PR

The current implementation works as follows:
  - If the list is odd-length, use the middle slice
  - If the list is even-length, select the slice with the larger cord area from two middle slices

[...]

PR #4656, opened 2024-10-23
Interpolate midsagittal slice

valosekj:
## Description

This PR implements a new method for obtaining the midsagittal slice in order to measure midsagittal lesion metrics (length, width, tissue bridges). This method was proposed during the 2024-10-21 meeting (@jcohenadad, @naga-karthik, @dpfyffer, Patrick). 

The proposed method uses the following logic (A: pseudocode; B: practical example):

image

## How to test the PR

```console
cd $SCT_DIR
git fetch
git checkout jv/interpolate_the_lesion
```

Download the attached dummy files below and run:

```console
sct_analyze_lesion -m sub-001_lesion.nii.gz -s sub-001_seg.nii.gz
```

[sub-001_seg.nii.gz](https://github.com/user-attachments/files/17491070/sub-001_seg.nii.gz)
[sub-001_lesion.nii.gz](https://github.com/user-attachments/files/17491071/sub-001_lesion.nii.gz)

## Context

This PR is replacing https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4631.

## TODO

- [x] `sct_analyze_lesion -qc` now raises an error as the midsagittal slice is now float (due to the interpolation) - fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656/commits/dc528a74bd00bcd55ea52f7b431f663824a349c2
- [x] remove the Jupyter Notebook (and figures) used for debugging 
- [x] update metrics in the script description
- [x] deal with multiple lesions


valosekj:
> 5a. Find the two closest slices

After some in-person discussion with @jcohenadad, it might be better to consider three slices; for example, for the mean spinal cord center of mass of 111.2, I would consider 111, 112, 113.

jcohenadad:
> for example, for the mean spinal cord center of mass of 111.2, I would consider 111, 112, 113.

actually, for 111.2 you would consider 110, 111 and 112

jcohenadad:
Relevant links:
- https://www.slicer.org/wiki/Coordinate_systems
- https://afni.nimh.nih.gov/pub/dist/edu/data/CD.expanded/afni_handouts/afni30_vol_dset_basics.pdf (check last slides)
- https://discovery.ucl.ac.uk/id/eprint/10146893/1/geometry_medim.pdf

valosekj:
> > for example, for the mean spinal cord center of mass of 111.2, I would consider 111, 112, 113.
> 
> actually, for 111.2 you would consider 110, 111 and 112

Yes, you are right. (I meant 112.2 but wrote 111.2).

However, @joshuacwnewton pointed me to this section from [NiBabel documentation](https://nipy.org/nibabel/coordinate_systems.html), which mentions "center of the voxel":

> We have not yet made this explicit, but voxel coordinates are also in a space. In this case the space is defined by the three voxel axes (first axis, second axis, third axis), where 0, 0, 0 is the **center of the first voxel**__ in the array and the units on the axes are voxels. Voxel coordinates are therefore defined in a reference space called voxel space.

If I understand this correctly, it would mean that 2 slices should be enough (as currently implemented in this PR); see the following sketch:

image

jcohenadad:
> If I understand this correctly, it would mean that 2 slices (or even 1 slice in the case of int) should be eventually enough (as currently implemented in this PR); see the following sketch:

If voxel origin is at the center of the voxel, then with 2 slices you would effectively cover [111.5, 113.5], which is centered around 112.5 (hence your partial volume estimation would be biased towards higher slices), whereas if you include 111, 112 and 113, the center is 112 (which is what you want).

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -984,15 +975,37 @@ def get_midsagittal_slice(self):
...
-        interpolation_factor = mean_spinal_cord_center_of_mass_x - int(mean_spinal_cord_center_of_mass_x)   # e.g., 0.2
+        # 5a. Find the three closest slices to mean_spinal_cord_center_of_mass_x
+        center_slice = round(mean_spinal_cord_center_of_mass_x)    # e.g., 112.2 --> 112; 112.9 --> 113
+        slice1 = center_slice - 1       # e.g., for 112.2 --> 111; for 112.9 --> 112
+        slice2 = center_slice           # e.g., for 112.2 --> 112; for 112.9 --> 113
+        slice3 = center_slice + 1       # e.g., for 112.2 --> 113; for 112.9 --> 114
+        # 5b. Calculate weights for each slice based on distance to the mean
+        # Convert positions to centered coordinates (slice centers)
+        pos1, pos2, pos3 = slice1 + 0.5, slice2 + 0.5, slice3 + 0.5
+        target_pos = mean_spinal_cord_center_of_mass_x + 0.5
```
valosekj, line 986: Okay. Claude is suggesting the following logic with 0.5 offset. This would mean that the example float numbers in [my sketch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656#issuecomment-2436578278) were wrong (and should be actually shifted by 0.5).

Screenshot 2024-10-25 at 11 15 10 AM

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -984,15 +975,37 @@ def get_midsagittal_slice(self):
...
+            # Calculate weights using inverse distance to ensure:
+            #   - Slices closer to our target position should have more influence
+            #   - Slices farther away should have less influence
+            #   - The influence decreases smoothly with distance
+            w1 = 1.0 / d1
+            w2 = 1.0 / d2
+            w3 = 1.0 / d3
+            # Normalize weights to sum to 1
+            total = w1 + w2 + w3
+            weights = [w1 / total, w2 / total, w3 / total]
```
valosekj, line 1005: Explanation:
image

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Claude is directly contradicted by quotes from both nibabel and NIfTI themselves that state that integer indices refer to the center of the voxel. ((In other words, the center of the voxel is ***not*** offset by 0.5 at all. AI misinformation strikes again D:))

Here's a graphic I made in MS Paint: 

image

valosekj: Yeah :-D 

Also, as the shifted slices

```python
pos1, pos2, pos3 = slice1 + 0.5, slice2 + 0.5, slice3 + 0.5
target_pos = mean_spinal_cord_center_of_mass_x + 0.5
```
are used in the following code only to compute the difference, the 0.5 shift actually seems to have no effect at all...

```python
d1 = abs(target_pos - pos1)
d2 = abs(target_pos - pos2)
d3 = abs(target_pos - pos3)
```

joshuacwnewton: I would really hesitate to refer to Claude in situations where factual information is required. It's good for ideation/brainstorming and subjective/stylistic tasks, but anything where truth/certainty is required is a big :warning:  :warning:  :warning:.)

valosekj: Thanks for your comments, Joshua!
I also agreed with @jcohenadad during a meeting a few minutes ago that there is no need to add 0.5. I'll remove the shift in the next commit.

valosekj:
I believe the PR is ready to be reviewed. 

Apologies for a lot of commits; I think the relevant commits are mainly after commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656/commits/9554cdb19d878553e76b69c5204c315ed970a97e, when we switched the interpolation logic to three slices (previous commits work with 2 slices).

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -914,19 +937,123 @@ def angle_correction(self):
...
+            #   - Slices farther away should have less influence
+            #   - The influence decreases smoothly with distance
+            w1 = 1.0 / d1       # e.g., 1 / 0.7 = 1.429
+            w2 = 1.0 / d2       # e.g., 1 / 0.3 = 3.333
+            w3 = 1.0 / d3       # e.g., 1 / 1.3 = 0.769
+            # Normalize weights to sum to 1
+            total = w1 + w2 + w3       # e.g., 1.429 + 3.333 + 0.769 = 5.531
+            weights = [w1 / total, w2 / total, w3 / total]      # e.g., 1.429/5.531, 3.333/5.531, 0.769/5.531 --> 0.258, 0.603, 0.139
+            # As seen above, the middle slice (9) is the closest to the target position (8.7) and has the highest weight
+        self.interpolation_weights = weights    # store it to be used by the `_interpolate_values` function
```
joshuacwnewton, line 1056: I notice that these weights are derived from a single float value (`x_target`). This means that `x_target` contains all of the "information" necessary for further computations, and thus the weights are merely a more complex way of storing the information we already have. 

Because everything comes from a single slice value, I would really try to create a clear separation between "midsagittal slice determination" and "applying the midsagittal slice to measurements via weights". For example:

- Keep the old `get_midsagiital_slice` function, with that function saving a single value (`x_target`) exactly as before, just with the new method instead ("5 axial slice mean of centers of mass").
- Put any weights-related code into a separate function, then call that function as part of the measurement methods.

joshuacwnewton: Side note: I find it strange that we are using 3 values across 3 slices to represent a single (float) slice value... I can't put my finger on it, but it just feels off? Like... overly convoluted? 

Like, via the example in the code comments, if "8.7" is the midsagittal slice, then isn't that kind of self-explanatory, as 70% of the way between slices 8 and 9? Why is slice 10 involved at all?

(cc: @mguaypaq)

joshuacwnewton (changes requested):
I am finding this PR to be a little difficult to review, because the diff is very large. It seems like in addition to the functional changes, there are also some non-functional changes (i.e. code being moved to new locations, renaming variables, etc.). 

My expectation was that some of the existing structure would stay the same (i.e. even though the method we use to determine the midsagittal slice is different, we would just "swap out" the method in-place). 

So, my broad feedback would be to try to emulate the old structure, that way it is easier to do a 1:1 comparison of old vs. new. (And, my specific feedback below will help to work towards that goal.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Thanks for the comments, Joshua!

> Like, via the example in the code comments, if "8.7" is the midsagittal slice, then isn't that kind of self-explanatory, as 70% of the way between slices 8 and 9? Why is slice 10 involved at all?

Good point! Yeah, this is what I tried to illustrate in the sketch in [this resolved comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656#issuecomment-2436578278).
After discussing it in person with Mathieu, the current implementation might indeed be a little bit overcomplicated. Also, it actually results in measures very similar to the "2 slice" approach. --> Further investigation is in process.

joshuacwnewton: Perfect! I can mark the "3 slice -> 2 slice" feedback as addressed, and just leave the "code organization" comment open for now.

valosekj: Just FYI, I updated the code to consider only two slices for the interpolation in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656/commits/4a4f0ecb48adac7d2a64ee1437a88d5e24951f35. I also updated the pseudocode diagram in the [PR description](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656#issue-2608422512).

valosekj:
Merging this PR into its base branch `jv/change_get_midsagittal_slice_logic`

PR #4660, opened 2024-10-23
Fix dtype mismatch error when mixing integers and floating point numbers

### LINKED ISSUE #4652: `sct_maths -add` fails when masks have different dtype
Our collaborator @dpfyffer ran into the following issue with the `sct_maths -add` command for SCT 6.4 when two masks (attached) have different dtypes:

```console
/home/dpfyffer/sct_6.4/bin/sct_maths -i sub-374_ses-01spinalcord_task-MP_bold_mean_seg.nii.gz -add sub-374_ses-01spinalcord_task-MP_bold_mean_CSF_seg.nii.gz -o sub-374_ses-01spinalcord_task-MP_bold_mean_SC_canal_seg_test.nii.gz

--
Spinal Cord Toolbox (git-HEAD-06cceb585baacccff0385324f69a637b2dde65ab)

sct_maths -i sub-374_ses-01spinalcord_task-MP_bold_mean_seg.nii.gz -add sub-374_ses-01spinalcord_task-MP_bold_mean_CSF_seg.nii.gz -o sub-374_ses-01spinalcord_task-MP_bold_mean_SC_canal_seg_test.nii.gz
--

  File "/home/dpfyffer/sct_6.4/spinalcordtoolbox/scripts/sct_maths.py", line 621, in <module>
    main(sys.argv[1:])
  File "/home/dpfyffer/sct_6.4/spinalcordtoolbox/scripts/sct_maths.py", line 443, in main
    data += np.sum(list_data, axis=0)
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int8') with casting rule 'same_kind'
```

I can reproduce the bug with the SCT master. Interestingly, the command works fine for v6.3. 

[sub-374_ses-01spinalcord_task-MP_bold_mean_CSF_seg.nii.gz](https://github.com/user-attachments/files/17439681/sub-374_ses-01spinalcord_task-MP_bold_mean_CSF_seg.nii.gz)
[sub-374_ses-01spinalcord_task-MP_bold_mean_seg.nii.gz](https://github.com/user-attachments/files/17439682/sub-374_ses-01spinalcord_task-MP_bold_mean_seg.nii.gz)

---

A side question: Dario and I installed SCT v6.4 using the following commands:

```console
git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox sct_6.4
cd sct_6.4
git checkout 6.4
./install_sct
```

Why the SCT version shown in the terminal when running `sct_maths` is `git-HEAD-06cceb585baacccff0385324f69a637b2dde65ab` and not `6.4`?

mguaypaq:
We can't use in-place operators (like `+=`, `*=`) on numpy arrays for operations that would need to promote the array's dtype (for example, integer += float needs to be stored as float, but the existing array can only store integer).

I added a test which fails with "Cannot cast ufunc output dtype" on `master`, and passes with the fix in this branch.

Fixes #4652.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -459,14 +459,16 @@ def main(argv: Sequence[str]):
                         list_data = get_data_arrays(data.shape, arg_value)
                     except ValueError as e:
                         printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                    data += np.sum(list_data, axis=0)
+                    # the addition can't use += because the dtypes could be different
+                    data = data + np.sum(list_data, axis=0)
 
             elif arg_name == "sub":
                 try:
                     list_data = get_data_arrays(data.shape, arg_value)
                 except ValueError as e:
                     printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                data -= np.sum(list_data, axis=0)
+                # the subtraction can't use -= because the dtypes could be different
+                data = data - np.sum(list_data, axis=0)
 
             elif arg_name == "mul":
                 if data.ndim == 4 and not arg_value:
@@ -477,14 +479,16 @@ def main(argv: Sequence[str]):
                         list_data = get_data_arrays(data.shape, arg_value)
                     except ValueError as e:
                         printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                    data *= np.prod(list_data, axis=0)
+                    # the multiplication can't use *= because the dtypes could be different
+                    data = data * np.prod(list_data, axis=0)
 
             elif arg_name == "div":
                 try:
                     list_data = get_data_arrays(data.shape, arg_value)
                 except ValueError as e:
                     printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                data /= np.prod(list_data, axis=0)
+                # the division can't use /= because the dtypes could be different
+                data = data / np.prod(list_data, axis=0)
 
             elif arg_name == "mean":
                 axis = ('x', 'y', 'z', 't').index(arg_value)
--- a/testing/cli/test_cli_sct_maths.py
+++ b/testing/cli/test_cli_sct_maths.py
@@ -79,6 +79,38 @@ def test_sct_maths_symmetrize(dim, tmp_path):
                           (im_in.data + np.flip(im_in.data, axis=int(dim))) / 2.0)
 
 
+def test_arithmetic_mixed_dtype(tmp_path):
+    """Test whether arithmetic operations can be performed on images with different dtypes without crashing."""
+    # make a dummy integer image
+    path_im_int = str(tmp_path / "im_int.nii.gz")
+    Image(np.ones((2, 3, 4), dtype=int)).save(path_im_int)
+
+    # make a dummy floating point image
+    path_im_float = str(tmp_path / "im_float.nii.gz")
+    Image(np.ones((2, 3, 4), dtype=float)).save(path_im_float)
+
+    # output filename, will be overwritten several times
+    path_im_out = str(tmp_path / "im_out.nii.gz")
+
+    # for each operation, test various combinations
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -459,14 +459,16 @@ def main(argv: Sequence[str]):
                         list_data = get_data_arrays(data.shape, arg_value)
                     except ValueError as e:
                         printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                    data += np.sum(list_data, axis=0)
+                    # the addition can't use += because the dtypes could be different
+                    data = data + np.sum(list_data, axis=0)
```
joshuacwnewton, line 463: The error message mentions `casting='same_kind'`, but I'm struggling to figure out how `v1 += v2` and `v1 = v1 + v2` map to numpy's 5 casting rules ([`{‘no’, ‘equiv’, ‘safe’, ‘same_kind’, ‘unsafe’}`](https://numpy.org/devdocs/reference/generated/numpy.can_cast.html)). It's hard to find information on the underlying mechanics?

But, maybe I'm overthinking things by trying to understand the exact mechanics. A quick test shows that... practically, this change fixes the issue, and `ints` will be promoted to `floats`.

joshuacwnewton, line 463: I was curious to see if `np.sum` is susceptible, but it isn't. I guess a good rule of thumb, then, is that we should only be concerned about in-place operations (where something gets added to something else), rather than operations whether the order doesn't matter.

joshuacwnewton (approved):
LGTM! Thanks for the fix. :hearts:

spinalcordtoolbox/scripts/sct_maths.py
mguaypaq: I think you can look at `np.ufunc.resolve_dtypes` to see the difference between regular addition, and in-place addition.

For regular addition (the triple of dtypes is (left input, right input, output), and `None` means "unconstrained"):
```py
>>> np.add.resolve_dtypes((np.dtype('int8'), np.dtype('float64'), None))
(dtype('float64'), dtype('float64'), dtype('float64'))
```
I'm interpreting this to mean:
- The left input gets cast from `int8` to `float64`.
- The right input stays `float64`.
- The output (which is a newly allocated array) will be `float64`.

For in-place addition (meaning, it tries to re-use/modify the existing allocation of the left input for the output):
```py
>>> np.add.resolve_dtypes((np.dtype('int8'), np.dtype('float64'), np.dtype('int8')))
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int8') with casting rule 'same_kind'
```
I'm interpreting this to mean:
- The output should be a `float64` array, but this can't be stored into an existing `int8` array.

mguaypaq: It's kind of tricky! But the numpy docs do specify a lot of it. For example, `help(np.add)` shows that:
- It's possible to specify an existing array to write the output.
- It's possible to specify the casting, which defaults to `same_kind`.

For example, it's possible to trigger the casting error with `np.add`:
```py
>>> left = np.arange(3, dtype=int)
>>> right = np.arange(3, dtype=float)
>>> out = np.zeros(3, dtype=int)
>>> # regular addition, newly allocated array for the output
>>> np.add(left, right)
array([0., 2., 4.])
>>> # existing array for the output
>>> np.add(left, right, out=out)
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

Whereas `help(np.sum)` shows that:
- It's possible to specify an existing array to write the output.
- But, there's no way to specify the casting rule to use.
- However there's an implicit typecast (which can truncate values!)
  ```
    out : ndarray, optional
        Alternative output array in which to place the result. It must have
        the same shape as the expected output, but the type of the output
        values will be cast if necessary.
  ```

PR #4668, opened 2024-10-29
Use faster dilation algorithm when dilating "mostly zero" point label images (`-x label`)

### LINKED ISSUE #4663: Investigate slow template warping after label dilation
From https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4303:

> - [ ] for `sct_warp_template`, I noticed that after the modification to warp the label files appropriately (using the dilation step), the function is muuuuuuch slower. It disrupted the course quite a bit. Given that these registered label files are not really used, maybe we can re-consider warping them by default. Alternatively, we could find a way to dilate them much faster.

joshuacwnewton:
## Description

Since 99.99% of voxels are 0 when `islabel=True`, we don't *need* to run dilation on every single voxel, which is what skimage will do by default. Instead, we can fetch the nonzero voxels, extract an area the size of the footprint, run dilation on that area, and then put it back.

I was concerned that this might behave differently in cases where we have two labels very close together. But, I think it ends up working identically, with the larger label dominating the smaller label (thanks to the use of `maximum` internally).

There's an even faster approach (just multiplying `footprint` by the label value, then setting the area centered on the nonzero voxel equal to this). I've implemented that as well in a follow-up approach.

## Testing

On my (fast) laptop, dilation time for a typical image went from 7s to 0.15s to <0.01s, thanks to minimizing the area we call `skimage.dilation` on, and ultimately removing `skimage.dilation` entirely.

## Linked issues

Fixes #4663.

### DIFF
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -149,7 +149,7 @@ def _dilate_point_labels(data, footprint):
         data_start = data_coords - pre_distances
         fp_start = fp_center - pre_distances
 
-        post_distances = np.min(data_shape - data_coords, fp_shape - fp_center)
+        post_distances = np.minimum(data_shape - data_coords, fp_shape - fp_center)
         data_stop = data_coords + post_distances
         fp_stop = fp_center + post_distances
 
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -132,8 +132,8 @@ def _dilate_point_labels(data, footprint):
     fp_shape = np.array(footprint.shape)
 
     # For odd-length dimensions, the center of the footprint is unique.
-    # For even-length dimensions, we round down.
-    fp_center = (fp_shape - 1) // 2
+    # For even-length dimensions, we round up.
+    fp_center = (fp_shape - 1 + 1) // 2
 
     data_out = np.zeros_like(data)
     for data_coords in np.argwhere(data):
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -118,49 +118,54 @@ def dilate(data, size, shape, dim=None, islabel=False):
     else:
         footprint = _get_footprint(shape, size, dim)
         if islabel:
-            return _dilate_point_labels(data, shape=shape, size=size, footprint=footprint)
+            return _dilate_point_labels(data, footprint=footprint)
         else:
             if data.dtype in ['uint8', 'uint16']:
                 return rank.maximum(data, footprint=footprint)
             else:
-                return dilation(data, footprint=footprint, out=None)
+                return dilation(data, footprint=footprint)
 
 
-def _dilate_point_labels(data, shape, size, footprint):
+def _dilate_point_labels(data, footprint):
     """
-    A more efficient dilation algorithm when we know the image is mostly nonzero (i.e. point label image)
+    A more efficient dilation algorithm when we know the image is mostly zero (i.e. point label image)
     """
-    # NB1: "size" means different things for different shapes, hence we compute the radius on a per-shape basis
-    # NB2: For even-length square/cubes, the radius will be n+0.5, but this gets handled by later calls to `int()`
-    radius = {'square': (size - 1) / 2, 'cube': (size - 1) / 2, 'disk': size, 'ball': size}[shape]
-    data_out = data.copy()
-    for (x, y, z) in np.argwhere(data != 0):
-        # Compute the (relative) extents of the bounding box, i.e. the left-right offsets: | <-> [x,y,z] <-> |
-        # NB: If the footprint would go outside the image boundaries, then we make bbox offsets < radius using `min()`
-        bbox = (
-            (min(radius, x), min(radius, data.shape[0] - 1 - x)),  # x offsets: | <-> x <-> |
-            (min(radius, y), min(radius, data.shape[1] - 1 - y)),  # y offsets: | <-> y <-> |
-            (min(radius, z), min(radius, data.shape[2] - 1 - z)),  # z offsets: | <-> z <-> |
-        )
+    if len(data.shape) != len(footprint.shape):
+        raise ValueError(f"incompatible shapes: {data.shape=} {footprint.shape=}")
+    if footprint.size == 0:
+        raise ValueError(f"cannot handle empty footprint: {footprint.shape=}")
 
-        # Extract a patch out of the image with [x,y,z] at the center
-        x_min, x_max = int(x - bbox[0][0]), int(x + bbox[0][1] + 1)
-        y_min, y_max = int(y - bbox[1][0]), int(y + bbox[1][1] + 1)
-        z_min, z_max = int(z - bbox[2][0]), int(z + bbox[2][1] + 1)
-        data_patch = data_out[x_min:x_max, y_min:y_max, z_min:z_max]
+    data_shape = np.array(data.shape)
+    fp_shape = np.array(footprint.shape)
 
-        # Extract an identically-sized patch out of the dilation footprint, with [c1,c2,c3] at the center
-        c = [(s-1)/2 for s in footprint.shape]
-        fp_x_min, fp_x_max = int(c[0] - bbox[0][0]), int(c[0] + bbox[0][1] + 1)
-        fp_y_min, fp_y_max = int(c[1] - bbox[1][0]), int(c[1] + bbox[1][1] + 1)
-        fp_z_min, fp_z_max = int(c[2] - bbox[2][0]), int(c[2] + bbox[2][1] + 1)
-        footprint_patch = footprint[fp_x_min:fp_x_max, fp_y_min:fp_y_max, fp_z_min:fp_z_max]
+    # For odd-length dimensions, the center of the footprint is unique.
+    # For even-length dimensions, we round down.
+    fp_center = (fp_shape - 1) // 2
 
-        # Multiply the footprint by the [x,y,z] voxel value to simulate dilation of the voxel
-        dilated_voxel = footprint_patch * data_out[x, y, z]
+    data_out = np.zeros_like(data)
+    for data_coords in np.argwhere(data):
+        # data_coords is a numpy array, which behaves differently from tuples
+        # when indexing another numpy array. We want the tuple behaviour.
+        data_value = data[tuple(data_coords)]
 
-        # Take the maximum of both patches. This will dilate the center voxel while also preserving close-by voxels.
-        data_out[x_min:x_max, y_min:y_max, z_min:z_max] = np.maximum(data_patch, dilated_voxel)
+        # If a data coordinate is too close to the edge, we have to use a
+        # truncated version of the footprint. These formulas compute the
+        # adjusted slice start and stop coordinates in the data array, and in
+        # the footprint array.
+        pre_distances = np.min(data_coords, fp_center)
+        data_start = data_coords - pre_distances
+        fp_start = fp_center - pre_distances
+
+        post_distances = np.min(data_shape - data_coords, fp_shape - fp_center)
+        data_stop = data_coords + post_distances
+        fp_stop = fp_center + post_distances
+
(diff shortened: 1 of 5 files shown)


mguaypaq:
If you _did_ want to multiply the footprint by the label value, you could use `np.maximum` to fix the stability problem for close labels.

Actually, is it possible that the current code could double-dilate some labels (albeit confined to a bounded box)?

mguaypaq:
Ooooh, there's even [`np.maximum.at()`](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.at.html) for efficiently folding new values into a (multi-)slice of an existing array using the maximum.

joshuacwnewton:
I went with `np.maximum` because the slicing logic was complicated enough as-is, but if you have an elegant way to simplify what I've done, please feel free give it a go! :hearts:

spinalcordtoolbox/math.py
```diff
@@ -113,14 +113,56 @@ def dilate(data, size, shape, dim=None):
...
+            if data.dtype in ['uint8', 'uint16']:
+                return rank.maximum(data, footprint=footprint)
+            else:
+                return dilation(data, footprint=footprint, out=None)
+
+
+def _dilate_point_labels(data, shape, size, footprint):
+    """
+    A more efficient dilation algorithm when we know the image is mostly nonzero (i.e. point label image)
+    """
```
joshuacwnewton, line 132: I might have done a lot of work for nothing? After some digging, it seems like scipy's `grey_dilation` may be significantly faster than skimage's dilation: https://github.com/scikit-image/scikit-image/issues/1190#issuecomment-572384985

Meaning, we don't have to reinvent the wheel here...

joshuacwnewton: EDIT: Nope, `grey_dilation` is still as slow as numpy's `dilation` in my tests. I wonder why?

mguaypaq:
@joshuacwnewton, how do you feel about this version of the algorithm (haha, assuming that the tests pass)? Feel free to revert my commits if the code is too tricky.

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,61 @@ def dilate(data, size, shape, dim=None):
...
+    for data_coords in np.argwhere(data):
+        # data_coords is a numpy array, which behaves differently from tuples
+        # when indexing another numpy array. We want the tuple behaviour.
+        data_value = data[tuple(data_coords)]
+
+        # If a data coordinate is too close to the edge, we have to use a
+        # truncated version of the footprint. These formulas compute the
+        # adjusted slice start and stop coordinates in the data array, and in
+        # the footprint array.
+        pre_distances = np.min(data_coords, fp_center)
```
joshuacwnewton, line 148: The error seems to be: 

```
   File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/math.py", line 148, in _dilate_point_labels
    pre_distances = np.min(data_coords, fp_center)
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 2953, in min
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 88, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
TypeError: only integer scalar arrays can be converted to a scalar index
```

From SO:

> I ran into the problem when venturing to use numpy.concatenate to emulate a C++ like pushback for 2D-vectors; If A and B are two 2D numpy.arrays, then numpy.concatenate(A,B) yields the error.
> The fix was to simply to add the missing brackets: numpy.concatenate( ( A,B ) ), which are required because the arrays to be concatenated constitute to a single argument

So it looks like the fix here is just:

```suggestion
        pre_distances = np.min((data_coords, fp_center))
```

mguaypaq, line 148: D'oh:
```suggestion
        pre_distances = np.minimum(data_coords, fp_center)
```

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,61 @@ def dilate(data, size, shape, dim=None):
...
+        data_start = data_coords - pre_distances
+        fp_start = fp_center - pre_distances
+
+        post_distances = np.min(data_shape - data_coords, fp_shape - fp_center)
```
mguaypaq, line 152:
```suggestion
        post_distances = np.minimum(data_shape - data_coords, fp_shape - fp_center)
```

mguaypaq (commented):
Oops, `min` is for minimum _inside_ an array (or along an axis). Here I wanted `minimum`, which is the element-wise minimum _between_ two arrays.

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,61 @@ def dilate(data, size, shape, dim=None):
...
+        raise ValueError(f"incompatible shapes: {data.shape=} {footprint.shape=}")
+    if footprint.size == 0:
+        raise ValueError(f"cannot handle empty footprint: {footprint.shape=}")
+
+    data_shape = np.array(data.shape)
+    fp_shape = np.array(footprint.shape)
+
+    # For odd-length dimensions, the center of the footprint is unique.
+    # For even-length dimensions, we round down.
+    fp_center = (fp_shape - 1) // 2
```
mguaypaq, line 136: It seems silly to say `x - 1 + 1` instead of just `x`, but `x // 2` really looks like rounding _down_, which would be confusing.
```suggestion
    # For even-length dimensions, we round up.
    fp_center = (fp_shape - 1 + 1) // 2
```

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,61 @@ def dilate(data, size, shape, dim=None):
...
+        pre_distances = np.min(data_coords, fp_center)
+        data_start = data_coords - pre_distances
+        fp_start = fp_center - pre_distances
+
+        post_distances = np.min(data_shape - data_coords, fp_shape - fp_center)
+        data_stop = data_coords + post_distances
+        fp_stop = fp_center + post_distances
+
+        data_loc = tuple(slice(start, stop) for start, stop in zip(data_start, data_stop))
+        fp_loc = tuple(slice(start, stop) for start, stop in zip(fp_start, fp_stop))
```
joshuacwnewton: I'll admit that I do find this specific organization of things a little unclear (it's hard for me to conceptualize what the variables are supposed to represent when just scanning through).

That said, once I get to the `np.maximum.at` part, it becomes a lot more clear. So, I think I'd maybe like to refactor it just a little (for clarity purposes only)?

I'd start with a comment that clarifies what our end goal is:

```python
        # To dilate each voxel, we multiply the footprint by the voxel's value
        # 
        # But, to preserve overlapping dilated voxels, we'd like to use `np.maximum` 
        # to take the maximum (so that the '0' voxels in the footprint don't overwrite
        # any existing nonzero voxels in the image).
        # 
        # To perform `np.maximum` in place, we use the `np.ufunc.at` syntax, which
        # requires an `indices` argument to specify *where* to perform the maximum at.
        # So, we just need to compute the bounding box surrounding the voxel, 
        # specified as [(x1,x2), (y1,y2), (z1,z2)]
```

Then, for the `np.maximum.at` call, we might explicitly specify the argument names to tie back into the earlier comment:

```python
        np.maximum.at(a=data_out, indices=data_loc, b=data_value*footprint[fp_loc])
```

And finally, I might change the `pre_`/`post_distances` blocks to better convey what values are being computed? These two lines in particular are really tripping me up for some reason, I think because of the implied "`- 0`" that's missing from the first statement:

```python
        pre_distances = np.min(data_coords, fp_center)
        post_distances = np.min(data_shape - data_coords, fp_shape - fp_center)
```

I find it really hard to parse what's happening here? Looking it over, I *think* this is what's happening:

```python
        # Make sure that the footprint doesn't exceed the image boundaries
        # If the footprint fits inside the image, then:
        #    - (width of footprint) <= (distance from voxel to image boundary) 
        # However, if the footprint would go outside the boundaries, then:
        #    - (distance from voxel to image boundary) < (width of footprint)
        # So we take the minimum to always use the shorter of the two distances.
```

One last thing that could be done to make it a little easier to grok is to nix the "pre" and "post" variable names? I totally get why you did it... there's not really an elegant word for "left/right" (since we're covering x/y/z), and "plus/minus" doesn't quite work either, so "pre/post" feel the best? 

But maybe something like this could make more sense (along with adding the redundant `- 0` to align the subtractions, and make it clear we're getting the distance from 0 to the center).

All together, maybe this would look like:

```python
        # Make sure that the footprint doesn't exceed the image boundaries
        # If the footprint fits inside the image, then:
        #    - (width of footprint) <= (distance from voxel to image boundary) 
        # However, if the footprint would go outside the boundaries, then:
        #    - (distance from voxel to image boundary) < (width of footprint)
        # So we take the minimum to always use the shorter of the two distances.
        distances = (
            # Distance from 0 --> voxel
            np.min(data_coords - (0, 0, 0),
                   fp_center - (0, 0, 0)),
            # Distance from voxel --> data boundaries
            np.min(data_shape - data_coords,
                   fp_shape - fp_center)
        )
        data_start = data_coords - distances[0]
        data_stop  = data_coords + distances[1]
        fp_start = fp_center - distances[0]
        fp_stop  = fp_center + distances[1]
```

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,90 @@ def dilate(data, size, shape, dim=None):
...
+        data_indices = tuple(slice(start, stop) for start, stop in zip(data_start, data_stop))
+        fp_indices = tuple(slice(start, stop) for start, stop in zip(fp_start, fp_stop))
+
+        # Now we can apply the (possibly truncated) footprint to the right part
+        # of the data array, blending in the data value with the "max" function
+        np.maximum.at(
+            a=data_out,
+            indices=data_indices,
+            b=data_value*footprint[fp_indices],
+        )
```
mguaypaq, line 190: It turns out that `np.ufunc.at` does not support keyword arguments, only positional :shrug: 
```suggestion
        np.maximum.at(
            data_out, data_indices,
            data_value*footprint[fp_indices],
        )
```

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,89 @@ def dilate(data, size, shape, dim=None):
...
+    # specified as a tuple of `slice` objects, which acts like
+    # data_out[x1:x2, y1:y2, z1:z2] (in 3D) or data_out[x1:x2, y1:y2] (in 2D).
+    #
+    # As an added complication, we may also need to crop the footprint, if the
+    # pixel to be dilated is close to an edge of the array. So, we also need to
+    # compute a bounding box surrounding the center of the footprint, also as
+    # a tuple of `slice` objects.
+
+    data_corners = np.array([[0]*dim, data.shape], dtype=int)
+    fp_corners = np.array([[0]*dim, footprint.shape], dtype=int)
```
joshuacwnewton, line 150: I love this change! Thank you for making it :heart:

spinalcordtoolbox/math.py
```diff
@@ -113,14 +106,89 @@ def dilate(data, size, shape, dim=None):
...
+            np.minimum(data_corners[1] - data_coords,
+                       fp_corners[1] - fp_center),
+        ]
+        data_start = data_coords - distances[0]
+        data_stop = data_coords + distances[1]
+        fp_start = fp_center - distances[0]
+        fp_stop = fp_center + distances[1]
+
+        data_indices = tuple(slice(start, stop) for start, stop in zip(data_start, data_stop))
+        fp_indices = tuple(slice(start, stop) for start, stop in zip(fp_start, fp_stop))
```
joshuacwnewton, line 182: Thanks for renaming these variables from `loc` to `indices`! This helps make up for the fact that the `ufunc.at` syntax doesn't accept keyword arguments.

joshuacwnewton (commented):
LGTM! Thanks so much for your help in designing this function. :heart:

PR #4675, opened 2024-11-04
Bump oldest macOS version used by test runners (12 -> 13)

mguaypaq:
The macOS 12 image is in the process of being removed from github actions: https://github.com/actions/runner-images/issues/10721

For reference, the last time we bumped the macOS version was in #4353.

As an aside, I wonder whether we should update our uses of `macos-14` to `macos-latest`, so that we automatically pick up new versions when they appear. (Whereas I think it's reasonable to need an explicit action on our part to drop old versions when they disappear. And besides, [there's no convenient label like `macos-oldest`](https://github.com/actions/runner-images).)

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -56,7 +56,7 @@ jobs:
     needs: generate-requirements-freeze
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, macos-12, macos-14, windows-2019, windows-2022]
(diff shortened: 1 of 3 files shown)


joshuacwnewton (approved):
> As an aside, I wonder whether we should update our uses of macos-14 to macos-latest, so that we automatically pick up new versions when they appear. 

I do kind of like the idea of having distinct commits showing when we upgrade our upper end? `latest` can be a little opaque. 

But, at the same time, the auto-updating would result in fewer finicky commits, so maybe it's worth it? I don't have strong feelings either way, I think.

Issue #4679, opened 2024-11-07
nnUNet tests are failing with `ModuleNotFoundError: No module named 'blosc2'`


joshuacwnewton:
It looks like `nnunetv2` broke something once again. :sweat_smile: 

```
  File "/home/runner/sct_0.0/spinalcordtoolbox/deepseg/inference.py", line 27, in <module>
    import spinalcordtoolbox.deepseg.nnunet as ds_nnunet
  File "/home/runner/sct_0.0/spinalcordtoolbox/deepseg/nnunet.py", line 13, in <module>
    from nnunetv2.inference.predict_from_raw_data import nnUNetPredictor   # noqa: E402
  File "/home/runner/sct_0.0/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/inference/predict_from_raw_data.py", line 22, in <module>
    from nnunetv2.inference.data_iterators import PreprocessAdapterFromNpy, preprocessing_iterator_fromfiles, \
  File "/home/runner/sct_0.0/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/inference/data_iterators.py", line 12, in <module>
    from nnunetv2.preprocessing.preprocessors.default_preprocessor import DefaultPreprocessor
  File "/home/runner/sct_0.0/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/preprocessing/preprocessors/default_preprocessor.py", line 23, in <module>
    from nnunetv2.preprocessing.cropping.cropping import crop_to_nonzero
  File "/home/runner/sct_0.0/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/preprocessing/cropping/cropping.py", line 5, in <module>
    from acvl_utils.cropping_and_padding.bounding_boxes import get_bbox_from_mask, crop_to_bbox, bounding_box_to_slice
  File "/home/runner/sct_0.0/python/envs/venv_sct/lib/python3.9/site-packages/acvl_utils/cropping_and_padding/bounding_boxes.py", line 5, in <module>
    import blosc2
ModuleNotFoundError: No module named 'blosc2'
```

joshuacwnewton:
- First failing run: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/11721678652
- Last passing run: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/11702234029
[...]

PR #4680, opened 2024-11-07
Preserve integer dtype when warping an image using `NearestNeighbour` interpolation

### LINKED ISSUE #4572: `sct_register_multimodal` is changing the data type when using nearest neighbours interpolation
## Description

I needed to use `sct_register_multimodal -identity 1 -x nn` to put segmentation masks in the same orientation/space as my images and I noticed that the function was changing the type of my output label. Is it normal ?

## Example

The type of my input label is `INT16` before using the function and the type of my image is `FLOAT64`.

After running
```
sct_register_multimodal -i labelsTr/sub-amuAL_T1w_label.nii.gz -d imagesTr/sub-amuAL_T1w.nii.gz -o reg/sub-amuAL_T1w.nii.gz -identity 1 -x nn
```

The type of my output label is `FLOAT64` even though I used nearest neighbour interpolation. Is it what I should expect ?

joshuacwnewton:
## Description

This will preserve integer images when `-x nn` is used (which I've explicitly referenced in my test).

## Linked issues

Fixes #4572.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -109,6 +109,33 @@ def get_parser():
     return parser
 
 
+def isct_antsApplyTransforms(dimensionality,
+                             fname_input,
+                             fname_output,
+                             fname_reference,
+                             transforms,
+                             interpolation_args,
+                             verbose):
+    """
+    Wrapper for CLI binary that allows for pre/postprocessing steps.
+
+    Note: Most logic should still go in `sct_apply_transfo`, because our SCT
+    script is essentially one big wrapper script for the ANTs call. Only use
+    this wrapper to fix bugs in the binary itself, or to compensate for
+    missing behavior/options in the ANTs source code. This wrapper should more
+    or less function like the original binary (with only minor tweaks added).
+    """
+    run_proc(['isct_antsApplyTransforms',
+              '-d', dimensionality,
+              '-i', fname_input,
+              '-o', fname_output,
+              '-t'] + transforms +  # list of transforms
+             ['-r', fname_reference] +
+             interpolation_args,  # list of ['-n', interp_type]
+             verbose=verbose,
+             is_sct_binary=True)
+
+
 class Transform:
     def __init__(self, input_filename, fname_dest, output_filename, list_warp, list_warpinv=[], verbose=0, crop=0,
                  interp='spline', remove_temp_files=1, debug=0):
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
It looks like the tests here are only failing because of #4679.

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -109,6 +111,44 @@ def get_parser():
...
+             ['-r', fname_reference] +
+             interpolation_args,  # list of ['-n', interp_type]
+             verbose=verbose,
+             is_sct_binary=True)
+    # Preserve input datatype if output data can be safely cast to it
+    # Note: This is mostly relevant for integer input images + NearestNeighbour
+    #       interpolation, since ANTs will automatically cast to float, but
+    #       we usually want to preserve the integer datatype.
+    im_out = Image(fname_output)
+    if np.array_equal(im_out.data, im_out.data.astype(dtype_in)):
```
mguaypaq, line 145: I would have expected this `if` statement to do a static check for certain arguments in the call to `isct_antsAppleTransforms` (something like `'nn' in interpolation_args`), rather than a dynamic check of whether the actual data values are affected by the casting operation. Is there a case that's better handled this way?

I ask because my experience is that having the output _type_ change based on the output _values_ can lead to very hard to debug corner cases.

mguaypaq (commented):
Thanks for adding the test, and congrats on adding the long-desired wrapper for `isct_antsApplyTransforms`! I do have one question about the approach, but otherwise this looks good.

spinalcordtoolbox/scripts/sct_apply_transfo.py
joshuacwnewton: Thank you for the feedback! It definitely makes sense to be explicit about the cases we wish to apply this logic too.

I think the reason I wrote it this way was because I figured it was a safe operation to apply blindly (since we still ensure array equality), and we're applying this operation to a file that we generate (meaning the file is kind of still within "our domain")? And I figured it might help us catch other instances of type changing that we can't think of on the spot (e.g. the user supplies a float32 file or something because they're keen on saving disk space, but ANTs outputs a float64 file by default, etc.)

But, I'm alright with reverting my logic and explicitly defining the cases we want to apply this logic to (as to match the scope of the original issue).

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -109,6 +111,43 @@ def get_parser():
...
+             ['-r', fname_reference] +
+             interpolation_args,  # list of ['-n', interp_type]
+             verbose=verbose,
+             is_sct_binary=True)
+    # Preserve integer datatype when interpolation is NearestNeighour to
+    # counter ANTs behavior of always outputting float64 images
+    im_out = Image(fname_output)
+    dtype_in = Image(fname_input).data.dtype
+    input_is_int = issubclass(dtype_in.type, np.integer)
+    if input_is_int and 'NearestNeighbor' in interpolation_args:
```
mguaypaq, line 144: It might be even simpler to just have the rule that NearestNeighbor interpolation outputs the same dtype as the input, regardless of integer/float/etc.? But I'm happy either way.
```suggestion
    if 'NearestNeighbor' in interpolation_args:
```

mguaypaq (approved):
Thanks for the update! And I do think your reasoning was very reasonable. But I still prefer this more static approach to dtypes.

PR #4630, opened 2024-09-17
Add subcommands for each DeepSeg model and display them in a gallery of models

joshuacwnewton:
### 1. Subcommands

This PR adds subparsers for each task. This changes the syntax for `sct_deepseg`:

```bash
# Before
sct_deepseg -task TASK_NAME -i input.nii.gz [OTHER_ARGS]

# After
sct_deepseg TASK_NAME -i input.nii.gz [OTHER_ARGS]
```

By adding subparsers, it allows us to add custom options on a per-model basis. This means that we can, for example, supply the `-c` argument for the two models that need it, without it being present for all other models (#4445). Additionally, this opens the door to moving `sct_deepseg_sc` et al. into `sct_deepseg` (#4532), since we can have an `sct_deepseg sc` subcommand with all of the original arguments of `sct_deepseg_sc`.

### 2. Model gallery

Additionally, adding subparsers allows us to clean up the `sct_deepseg -h` command, while storing the more complex options in `sct_deepseg TASK_NAME -h`. This frees up the `sct_deepseg.rst` page for information beyond just `-h`. 

Also, because each task has its own dedicated help, it means we can now create 1 documentation page per task. Meaning that, all together, we can use `sct_deepseg.rst` as a [**Model Gallery**](https://spinalcordtoolbox--4630.org.readthedocs.build/user_section/command-line/sct_deepseg.html#sct-deepseg) that points to each of the dedicated pages for each task:

image

Each image is clickable, and will open up the argparse help for each specific model:

image

This allows us to create dedicated hubs with information and usage for each model.

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4318.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3767.
Better solution for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4445.
Paves the way for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4532.

PR #4684, opened 2024-11-07
Add model gallery for current version of DeepSeg (`-task` syntax)

### LINKED ISSUE #3767: Improve our advertising for our `sct_deepseg` suite of models and tasks
### Background

New `ivadomed` models are often quietly integrated into SCT with just a small pull request (#3766, #3637, #3527, #3224). These models are quite valuable from a developer's perspective, as they represent a (relatively) low-effort way to introduce new features for our users, without the need for large infrastructure changes. These models are also important from a lab student's perspective; they're often the culmination of an internship or research project with Ivadomed.

However, despite the value of these models, we don't really show them off as much as we could be doing:

* There usually isn't really any fanfare or advertisement alongside these PRs. We might include a small note in the changelog and the release announcement, but other than that, we don't really _do anything_ to get our users to try out the new models. 
* The biggest concern to me is that `sct_deepseg` is completely absent from our tutorials and the SCT course. (Our "Segmentation" section focuses entirely on `sct_propseg`, `sct_deepseg_sc`, and `sct_deepseg_gm`, and these functions are mostly highlighted as a preprocessing step for registration. In other words, there is no focus on our diverse range of models available for _other_ kinds of segmentation tasks.)

Right now, the main way that users would discover a given model is by reading the publication associated with the research project (if it even exists), then finding their way to SCT via the "this model is available for use in SCT" note attached to it. But, I think we could be doing better.

### Proposed solution

- [ ] Include a short section in the SCT course slides for basic usage of `sct_deepseg` (viewing tasks, selecting a task to run segmentation).
- [ ] Then, mirror these tutorial steps in our SCT Tutorials.

This would be a good start, but I think we could go even further: 

- [ ] Create a gallery for `sct_deepseg` models (similar to the example galleries from [`sklearn`](https://scikit-learn.org/stable/auto_examples/index.html) and [`skimage`](https://scikit-image.org/docs/stable/auto_examples/index.html)) that would show off the capabilities of each model. (Then, we could link to this gallery as a "Read more" at the end of the aforementioned "basic tutorial".)

However, I think that this last task would be a large undertaking, especially because of the effort required to integrate these pages with our existing documentation. (We would need to put a lot of thought into making sure the pages are discoverable, while also making sure that our website is not bloated, i.e. that it is still easy to understand and navigate).

So, if possible, I would like to see #3095 (i.e. https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3340) handled first, so that we have better infrastructure in place for navigating between "CLI Script" pages and "Tutorial" pages (see https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3095#issuecomment-958061803). That way, users could start at the `sct_deepseg` page, then find their way to usage examples from that page.

joshuacwnewton:
## Description

This PR is roughly equivalent to half of PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4630. It is meant to grant us the benefits of the `sct_deepseg` model gallery right away (for v6.5), while also allowing us to update the gallery when we adopt `sct_deepseg` subparsers for v7.0.

This PR contains no functional changes; only documentation changes. Therefore, it should be safe to merge right away, especially with our recent change to only update the documentation for stable releases.

Note: One new change I added that differs from PR #4630 was switching from a 4-column table to a 3-column table to accommodate the longer task names:

image

Perhaps this is too large, though? Feedback is appreciated!

## Linked issues

Fixes #3767.

### DIFF
--- a/documentation/source/user_section/command-line/sct_deepseg.rst
+++ b/documentation/source/user_section/command-line/sct_deepseg.rst
@@ -8,19 +8,19 @@ Here we provide a gallery of each model available in the ``sct_deepseg`` CLI too
 Spinal cord
 -----------
 
-.. |seg_sc_contrast_agnostic| image:: https://i.imgur.com/1QXBTWf.png
+.. |seg_sc_contrast_agnostic| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/contrast_agnostic.png
    :target: deepseg/seg_sc_contrast_agnostic.html
 
-.. |seg_lumbar_sc_t2w| image:: https://i.imgur.com/aVJNqp1.png
+.. |seg_lumbar_sc_t2w| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/lumbar_t2.png
    :target: deepseg/seg_lumbar_sc_t2w.html
 
-.. |seg_sc_epi| image:: https://i.imgur.com/zj9idbJ.png
+.. |seg_sc_epi| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/epi_bold.png
    :target: deepseg/seg_sc_epi.html
 
-.. |ms_sc_mp2rage| image:: https://i.imgur.com/cqsrFWF.png
+.. |ms_sc_mp2rage| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/ms_sc_mp2rage.png
    :target: deepseg/seg_ms_sc_mp2rage.html
 
-.. |mice_sc| image:: https://i.imgur.com/MaajGC2.png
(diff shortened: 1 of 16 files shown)


documentation/source/user_section/command-line/sct_deepseg.rst
```diff
@@ -1,10 +1,160 @@
...
-   :prog: sct_deepseg
-   :markdownhelp:
-   :noepilog:
+Here we provide a gallery of each model available in the ``sct_deepseg`` CLI tool.
+
+Spinal cord
+-----------
+
+.. |seg_sc_contrast_agnostic| image:: https://i.imgur.com/1QXBTWf.png
+   :target: deepseg/seg_sc_contrast_agnostic.html
```
joshuacwnewton, line 12: TOOD: Replace imgur links with links to `doc-figures` repo.

mguaypaq:
Also: even after #4685, the non-task-specific help message from `sct_deepseg -h` will still be available on the "all tools" page, which is good.

Issue #4626, opened 2024-09-11
Integrate TotalSpineSeg project into SCT


joshuacwnewton:
This issue is meant to track the necessary steps required to integrate [`neuropoly/TotalSpineSeg`](https://github.com/neuropoly/totalspineseg) into SCT. 

> Context: https://github.com/neuropoly/totalspineseg/pull/46#issuecomment-2342387047
> 
>> [U]ltimately this method will be callable via SCT. [W]e should already think of an appropriate API to plug it inside SCT. Tagging @joshuacwnewton @mguaypaq so we can start thinking about it

TotalSpineSeg uses a two-stage approach, with both stages based on `nnUNetv2`. `sct_deepseg` already supports multi-model inference (see [`-task seg_tumor-edema-cavity_t1-t2`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fd89b9e240720c19c8f1d368a1329b6e123d4c21/spinalcordtoolbox/deepseg/models.py#L241-L252)). But, I'm not sure if `sct_deepseg`'s infrastructure is built to handle the non-`nnUNet` steps that occur [between the Step 1 and Step 2 models](https://github.com/neuropoly/totalspineseg/blob/d3318be8c84bb10ce494d2fb3b1cf7e52c86b4b7/totalspineseg/inference.py#L240-L342).

----

I think some initial steps could be:

- [ ] Try installing [TotalSpineSeg's dependencies](https://github.com/neuropoly/totalspineseg/blob/d3318be8c84bb10ce494d2fb3b1cf7e52c86b4b7/pyproject.toml#L42-L54) into the SCT venv.
- [ ] Try running inference using each of TotalSpineSeg's models individually.

If those work OK, then we can begin brainstorming how to adopt the postprocessing steps used by TotalSpineSeg.

----

Alternatively, we could just `pip install` the entire TotalSpineSeg project, and simply call a designated entry point function for inference (a similar pattern to how we use ivadomed). This would take care of handling the implementation of the postprocessing steps.

[...]

Issue #4556, opened 2024-07-08
Empty prediction for `-task seg_sc_lesion_t2w_sci` when using `nnunetv2==2.4.2`


valosekj:
### Description

Running `sct_deepseg -task seg_sc_lesion_t2w_sci` on a sample T2w image (Dropbox [link](https://www.dropbox.com/scl/fi/czg2mgc07ytxpf6g086m1/sub-2416B6027B_ses-2416B_T2w.nii.gz?rlkey=ucdxljro3hn7tlbq0bti40uab&dl=0)) results in an empty spinal cord prediction:

```bash
$ sct_deepseg -i sub-2416B6027B_ses-2416B_T2w.nii.gz -task seg_sc_lesion_t2w_sci


--
Spinal Cord Toolbox (git-master-65e44b5068fe5d8a094e3e0571566c01d2d460ab)

sct_deepseg -i sub-2416B6027B_ses-2416B_T2w.nii.gz -task seg_sc_lesion_t2w_sci
--

Model 'model_seg_sci_multiclass_sc_lesion_nnunet' is up to date (Source: https://github.com/ivadomed/model_seg_sci/releases/download/r20240130/model-sci-multisite_r20240130.zip)
...
```

<details><summary>screenshot</summary>

<img width="700" alt="image" src="https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/39456460/676d64bf-a51f-448f-9219-c4be9e90d4f7">

</details>

However, when downloading the model from [its source](https://github.com/ivadomed/model_seg_sci/releases/tag/r20240130) and running the prediction using the [run_inference_single_subject.py](https://github.com/ivadomed/model_seg_sci/blob/main/packaging/run_inference_single_subject.py) script, the spinal cord segmentation is fine!

```bash
$ conda activate nnunet
$ python ~/code/model_seg_sci/packaging/run_inference_single_subject.py -i sub-2416B6027B_ses-2416B_T2w.nii.gz -path-model ~/Downloads/model-sci-multisite_r20240130 -pred-type sc -o sub-2416B6027B_ses-2416B_T2w_seg.nii.gz
```

<details><summary>screenshot</summary>

<img width="700" alt="image" src="https://github.com/spinalcordtoolbox/spinalcordtoolbox/assets/39456460/4acfc7c5-889f-4acc-8274-3e517c2c7cda">

</details>

valosekj:
[...]

Issue #3661, opened 2022-01-26
Ubuntu 20.04 runners are intermittently stalling out during the runs of the test suite


joshuacwnewton:
I've noticed this on a few PRs, most notably https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3660#discussion_r792733908. But, it seems to be intermittent, as I've seen it pass without issue after restarting a job, and I've also seen it pass fine on master recently.

So, this probably requires further investigation. :thinking:

mguaypaq:
On #3660, I noticed that the Ubuntu 20.04 runner seems to either succeed in under 20 minutes (like the other runners), or else it goes for hours, stuck at the "Run pytest test suite (with code coverage)" step without any output.

mguaypaq:
The step which stalls is different on Ubuntu 20.04 than the other runners, changed recently in #3463 ([the specific lines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3463/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR213-R215)). Maybe related?

On [the Actions tab](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/workflows/tests.yml) I see there are currently two stalled runs (spinning yellow bullet points), but there's no output to see.

joshuacwnewton:
Ah! That's a really good insight, and makes perfect sense as to why the Ubuntu runners specifically are stalling. I'm willing to bet that it very much has to do with the addition of coverage.

mguaypaq:
[This cancelled run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4944025569?check_suite_focus=true) has output logs for the "Run pytest test suite (with code coverage)" step, shows it was 81% done after 6 hours. I don't remember whether I cancelled it manually, or else it timed out by itself.

[This currently stuck run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4954284311?check_suite_focus=true) has been running for almost 5 hours. I'll check again tomorrow to see its fate.

joshuacwnewton:
> [This cancelled run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/runs/4944025569?check_suite_focus=true) has output logs for the "Run pytest test suite (with code coverage)" step, shows it was 81% done after 6 hours. I don't remember whether I cancelled it manually, or else it timed out by itself.

[...]

PR #4690, opened 2024-11-08
Add totalspineseg to SCT

NathanMolinier:
#### GitHub

#### PR contents

## Description

The aim of this PR is to implement totalspineseg in SCT. Ideally, if possible, we should be able to call totalspineseg for: 
- disc labeling (normal call)
- spine segmentation (add a specific flag to the command) 

## Concerns

- From what I see, if I install totalspineseg weights in the associated package, it will be difficult for sct to say if weights are installed or not. So maybe I can modify totalspineseg code to allow the installation of weights where we want. Just need to modify this:
https://github.com/neuropoly/totalspineseg/blob/e6019bbbfff5ae7e9500262949576cf5a02b6f2a/totalspineseg/init_inference.py#L35-L48
> I updated the package accordingly to facilitate the use in sct see https://github.com/neuropoly/totalspineseg/blob/4a18a71e517043993e41157c31d1970704943483/totalspineseg/init_inference.py#L38-L47
- Currently a different nnunetv2 version is being installed by totalspineseg but I believe we can stick to the current version used by SCT

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4626.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4693.


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -316,3 +322,45 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
...
+            min(os.cpu_count(),
+                psutil.virtual_memory().total / 2**30 // 8),
+            1
+        )),
+        quiet=False,
+    )
+    fnames_out, targets = [], []
+    for output_dirname in ["step1_canal", "step1_cord", "step1_levels", "step1_output", "step2_output"]:
+        fnames_out.append(os.path.join(tmpdir_nnunet, output_dirname, os.path.basename(path_img)))
+        targets.append(f"_{output_dirname}")
```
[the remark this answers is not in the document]
NathanMolinier: Regarding the `torch.device` request, we need to update totalspineseg to use nnunet api instead of cli. Is it ok if we keep it as it is for now ?

Regarding outputting a list of the final paths, I am not really sure it is the role of the inference function. However, I printed the paths at the end with a quick description to make sure that people understand the different outputs.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -183,6 +185,16 @@
          "thr": None,  # Images are already binarized
          "default": False,
      },
+    "totalspineseg": {
+         "url": {
+             "step1": ["https://github.com/neuropoly/totalspineseg/releases/download/r20241005/Dataset101_TotalSpineSeg_step1_r20241005.zip"],
+             "step2": ["https://github.com/neuropoly/totalspineseg/releases/download/r20241005/Dataset102_TotalSpineSeg_step2_r20241005.zip"],
+         },
```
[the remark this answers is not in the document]
NathanMolinier: The `url` are not really hardcoded, they are accessible inside the pyproject.toml. But I'm wondering if it is possible to access these URLs after a pip installation

requirements.txt
```diff
@@ -54,6 +54,7 @@ requirements-parser
 scipy
 scikit-image
 scikit-learn
+totalspineseg
```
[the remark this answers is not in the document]
NathanMolinier: Maybe I could downgrade the version that totalspineseg uses ?

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: TODO: Add new entry to `sct_deepseg` model gallery:

- [ ] Export the figure from [`sct_deepseg.rst model figures`](https://docs.google.com/presentation/d/1y38emw3TB6TAtkcVDIokl2iPBLWfMClER2IvkZbg5DA/edit#slide=id.g3142f5bca78_1_0) as .png, then upload to `doc-figures` master branch.
- [ ] Add a new page to the `documentation/source/user_section/command-line/sct_deepseg/` folder that matches the other model pages. 
- [ ] Edit `documentation/source/user_section/command-line/sct_deepseg.rst` to add:
   - A new image link (with `:target:`)
   - A new entry in the appropriate `list-table` table.
   - A new entry in the `toctree` at the very bottom of the page.

NathanMolinier: I think I did all the steps you asked for

spinalcordtoolbox/deepseg/models.py
```diff
@@ -401,11 +420,17 @@ def install_model(name_model, custom_url=None):
...
-        urls_used = {}
-        for seed_name, model_urls in url_field.items():
-            logger.info(f"\nInstalling '{seed_name}'...")
-            urls_used[seed_name] = download.install_data(model_urls,
-                                                         folder(os.path.join(name_model, seed_name)), keep=True)
+        # totalspineseg handles data downloading itself, so just pass the urls along as if they were used
+        # TODO: update `init_inference` to return the urls used instead
+        if name_model == 'totalspineseg':
+            init_inference(data_path=Path(folder(name_model)), quiet=False)
+            urls_used = {step: urls[0] for step, urls in url_field.items()}
```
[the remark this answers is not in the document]
NathanMolinier: I added the `ZIP_URLS` and removed this return

spinalcordtoolbox/deepseg/models.py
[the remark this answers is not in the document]
NathanMolinier: I did it in the version [20241115](https://pypi.org/project/totalspineseg/)

requirements.txt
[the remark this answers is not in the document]
joshuacwnewton: After talking with @NathanMolinier on Slack, it appears that TotalSpineSeg is unable to downgrade to 2.3.1, so I will try to figure out what is keeping SCT from upgrading to 2.4.2:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4556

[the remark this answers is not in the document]
joshuacwnewton: It turns out that the newest version of `nnunetv2` (`2.5.1`) may fix the problems of both SCT and TotalSpineSeg. :tada:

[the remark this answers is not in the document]
joshuacwnewton: Hrm. It seems like `nnunetv2==2.5.1` fixes the above problem, however it is incompatible with TotalSpineSeg due to https://github.com/MIC-DKFZ/nnUNet/issues/2480.

How unforunate... there is not a single version between 2.4.0 and 2.5.1 that is compatible with *both* SCT and TotalSpineSeg.

Maybe 2.5.2 will be our lucky day. :sweat_smile:

documentation/source/user_section/command-line/deepseg/totalspineseg.rst
```diff
@@ -0,0 +1,11 @@
+
+                
+totalspineseg
+============
```
joshuacwnewton, line 4: The ReadTheDocs build failed with the error:

```
/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/checkouts/4690/documentation/source/user_section/command-line/deepseg/totalspineseg.rst:4: 
WARNING: Title underline too short.
```

```suggestion
totalspineseg
=============
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -183,7 +185,7 @@
          "thr": None,  # Images are already binarized
          "default": False,
      },
-    "model_seg_ms_lesion": {
+      "model_seg_ms_lesion": {
```
joshuacwnewton, line 188: The Code Quality check failed with the following message:

```
spinalcordtoolbox/deepseg/models.py:188:7: E131 continuation line unaligned for hanging indent
```

This one is my fault, I added this in my merge commit. My bad! :sweat: 

```suggestion
    "model_seg_ms_lesion": {
```

joshuacwnewton (changes requested):
Just some comments addressing the automated test failures...

joshuacwnewton (changes requested):
Looks like there are some more test failures from `pytest` itself:

requirements.txt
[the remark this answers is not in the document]
joshuacwnewton: LGTM! Thanks for this! :hearts:

spinalcordtoolbox/deepseg/models.py
```diff
@@ -18,6 +18,8 @@
 from spinalcordtoolbox import download
 from spinalcordtoolbox.utils.sys import stylize, __deepseg_dir__
 
+from totalspineseg.init_inference import init_inference
```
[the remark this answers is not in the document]
joshuacwnewton: It looks like the slowdown is caused by the import of `init_inference`.

Before (0.706s): 

image

After (1.937s):

image

I will update this to use a call to LazyLoader.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -443,11 +461,16 @@ def install_model(name_model, custom_url=None):
...
         if not isinstance(url_field, dict):
             raise ValueError("Invalid url field in MODELS")
-        urls_used = {}
-        for seed_name, model_urls in url_field.items():
-            logger.info(f"\nInstalling '{seed_name}'...")
-            urls_used[seed_name] = download.install_data(model_urls,
-                                                         folder(os.path.join(name_model, seed_name)), keep=True)
+        # totalspineseg handles data downloading itself, so just pass the urls along
+        if name_model == 'totalspineseg':
+            tss_init.init_inference(data_path=Path(folder(name_model)), quiet=False, dict_urls=url_field)
```
joshuacwnewton, line 466: I noticed that the `init_inference` function behaves in a specific way:

- It [keeps the model `.zip` files on disk](https://github.com/neuropoly/totalspineseg/blob/88377f2d8bb60d73736496e30126c75511f830d9/totalspineseg/utils/install_weights.py#L96-L97) in the `exports/` folder (to avoid redownloading them)
- It will skip reinstalling the model [if the output folder already exists](https://github.com/neuropoly/totalspineseg/blob/88377f2d8bb60d73736496e30126c75511f830d9/totalspineseg/utils/install_weights.py#L91).

This is a bit troublesome for SCT because:

- [ ] It essentially stores 2 copies of the models on disk (taking up ~800MB of space instead of ~400MB)
- [ ] Users can't reinstall the models if something has happened to the files on disk, or if the versions have updated. (`sct_deepseg` has logic that will check the model urls used in the download, then reinstall the model if the URLs are not the latest version.)

Would you be open to changing `totalspineseg` to allow reinstallation and not storing the .zips on disk?

joshuacwnewton, line 466: It looks like the tests are failing on Windows due to a FileNotFound error for this call:

```
    subprocess.run(['nnUNetv2_install_pretrained_model_from_zip', str(zip_file)])
```

I think maybe this needs to be `nnUNetv2_install_pretrained_model_from_zip.exe` on Windows? 

To avoid cross-compatibility issues, I would double-check to see if there is a way to run the `nnunetv2` commands via the Python API, rather than using `subprocess`. Otherwise, you could maybe just check if the OS is Windows, then append `.exe` to all of your subprocess calls.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -201,6 +202,16 @@
         "thr": None,  # Images are already binarized
         "default": False,
     },
+    "totalspineseg": {
+         # NB: Rather than hardcoding the URLs ourselves, use the URLs from the totalspineseg package.
+         # This means that when the totalspineseg package is updated, the URLs will be too, thus triggering
+         # a re-installation of the model URLs
+         "url": tss_init.ZIP_URLS,
```
joshuacwnewton, line 209: There is a new problem with the import time.

Because we need to access `init_inference` in order to get the `ZIP_URLS` dictionary, we also have to import the heavy `torchio` package at the same time.

@NathanMolinier would you be able to place the `ZIP_URLS` object in a location where there are no other unnecessary imports? (That way, accessing `ZIP_URLS` won't incur a big slowdown.)

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: It looks like totalspineseg breaks the following assumption by the QC report, and so the assertion fails, because there are 5 outputs for every input:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/f8796ec1d3b329f9b5b1611bf67bfdade2627cac/spinalcordtoolbox/scripts/sct_deepseg.py#L374-L380

Which outputs should we use for QC? Should we just do the Step 2 output, maybe? Or should we try to do QC for the spinal cord/canal too?

joshuacwnewton (changes requested):
OK! Now that we are able to install totalspineseg and run inference, I have a new round of reviews to address the up-to-date test failures.

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Yes I can try that

spinalcordtoolbox/deepseg/models.py
NathanMolinier: I'm not sure we are currently able to qc the output of totalspineseg. But we also output a preview folder that could serve as qc ?

joshuacwnewton: I am working on a QC right now. :)

joshuacwnewton: How does this look?

image

If it looks good, I can push some commits that add the required changes.

jcohenadad: i'm wondering wether we should prefer ax or sagittal for this model? @NathanMolinier

joshuacwnewton: Semi-related:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4705

I'm not sure how much effort a sagittal QC will be for `sct_deepseg`, so I mostly added the above QC as a quick placeholder to get this PR merged (since we have collaborators asking by email about this feature).

Maybe we could make further adjustments to the QC in a follow-up PR?

NathanMolinier: > i'm wondering wether we should prefer ax or sagittal for this model? @NathanMolinier

For a quick look, I personally prefer sagittal but axial makes a lot of sense especially to look at the spinal cord segmentation

spinalcordtoolbox/deepseg/models.py
NathanMolinier: I moved `ZIP_URLS` to `totalspineseg.utils.utils`

spinalcordtoolbox/deepseg/models.py
NathanMolinier: I added the `.exe` as requested on the subprocess calls

NathanMolinier: I added a variable in the function `init_inference` to keep or not the zip after installation:
```python
def init_inference(
        data_path,
        dict_urls,
        store_export=True,
        quiet=False
```

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: I'm very sorry, but because `utils.py` is a subset of `utils`, and `utils` contains `augment`, the slowdown is the exact same as before. :sweat: 

image

joshuacwnewton: I am not sure what the solution is, actually... the import graph above shows `totalspineseg.utils` -> `totalspineseg` -> `totalspineseg.utils.augment`.

Let me think about how to best approach this. Give me one moment.

joshuacwnewton: Turns out that SCT can fetch the URLs directly from totalspineseg: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4690/commits/56f2ae4906c5959df39ec9f3e9eeb3a5c82ed0b4

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Huh! I'm surprised... the Windows tests are still failing: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/11959633540/job/33341854040?pr=4690#step:6:916

Maybe my hunch was wrong, then? I'll try to debug on a Windows machine, maybe...

joshuacwnewton: I mean, the script is listed:

image

But it's not on the PATH... hmm... Oh!! I know what's happening here.

- On Linux/macOS, the folder with the CLI scripts is called `bin/`
- On Windows, the folder is called `Scripts/`

When I check `os.environ['PATH']` on Linux, I see:

```
/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/bin
```

However, when I check `os.environ['PATH']` on Windows, I see:

```console
PS C:\Users\joshu\sct_6.4> python/envs/venv_sct/python.exe -c "import os; print(os.environ['PATH'])"
C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Windows\System32\OpenSSH\;C:\Program Files\Git\cmd;C:\Program Files\Git\mingw64\bin;C:\Program Files\Git\usr\bin;C:\Users\joshu\AppData\Local\Microsoft\WindowsApps;C:\Users\joshu\sct_6.4\bin\;
```

So, for some reason, on Linux, the path containing the nnUNet executable is on the PATH. But, on Windows, the corresponding directory is _not_ on the path. I'm not sure why, but that certainly explains it.

The reason that TotalSpineSeg has worked in the past is because the README.md asks users to activate their venv. But, for SCT, we don't have that luxury. So, right now `totalspineseg` is broken as a PyPI package due to its use of subprocess...

joshuacwnewton: I would recommend swapping out the subprocess calls for calls to Python API calls, [for example](https://github.com/MIC-DKFZ/nnUNet/blob/ac79a612ae696af224169cbefde7519354782e64/nnunetv2/model_sharing/model_import.py#L6):

```python
from nnunetv2.model_sharing import install_model_from_zip_file
```

This will work for all platforms, and will save us from having to worry about the PATH.

joshuacwnewton: Thank you!!

NathanMolinier: I just talked with @jcohenadad and maybe we should consider not being compatible with windows users yet as more problems may occurs and because it may take too much time solving all of them before the release. The original goal was to have a temporary implementation before having a complete integration inside SCT. So maybe we can raise a warning for windows users.

joshuacwnewton: Unfortunately it seems like the "FileNotFound" error is not Windows-specific... I think this will actually block the PR for Linux, too, unless it is fixed: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4690#discussion_r1854013930

joshuacwnewton: Maybe we can add a quick-fix to add the `bin/` (UNIX) and `Scripts/` (Windows) folders to the PATH?

I don't like this solution, but it will help us avoid having to rewrite the TotalSpineSeg package. I'll submit the fix now to test it out.

spinalcordtoolbox/deepseg/inference.py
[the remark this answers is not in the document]
joshuacwnewton: Nope, the tests still seem to be stalling out. :/

NathanMolinier:
Hey @joshuacwnewton, I updated the [package](https://pypi.org/project/totalspineseg/) to replace all CLI calls with API calls ! It is also possible now to specify the device you want to use !

spinalcordtoolbox/deepseg/inference.py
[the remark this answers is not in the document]
joshuacwnewton: I have good news and bad news! 

- The good news is, @NathanMolinier's changes from CLI to API have fixed the FileNotFound error.
- The bad news is, inference is still stalling out...
- But the good news is, I was able to finally reproduce the stall on my local machine! Meaning that, we should be able to debug it now:

```
Copying the input images to the input folder for processing:
100%|███████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 553.34it/s]

Converting 4D images to 3D:
100%|████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 51.29it/s]

Transforming images to canonical space:
100%|████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 54.79it/s]

Resampling images to 1x1x1mm:
  0%|                                                                                                        | 0/1 [00:00<?, ?it/s]
```

I will try my best to figure out why this is occurring, and once I can address this, we should be finally OK to merge... (famous last words)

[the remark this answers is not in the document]
joshuacwnewton: I temporarily edited my local copy of `totalspineseg/resample.py` to get rid of the `process_map` call and just run resampling as-is. And it works!

But, then it immediately stalls on the _next_ copy of `process_map`...

image

So, it seems like it has something to do with the sequential use of `process_map` (i.e. it works for the first three, then something happens, and then the next instance gets stuck).

----

When I Ctrl+C during the stall, I get the following:

```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/threading.py:1080: KeyboardInterrupt
```

Line 1080 corresponds to this:

```diff
    def _wait_for_tstate_lock(self, block=True, timeout=-1):
        # Issue #18808: wait for the thread state to be gone.
        # At the end of the thread's life, after all knowledge of the thread
        # is removed from C data structures, C code releases our _tstate_lock.
        # This method passes its arguments to _tstate_lock.acquire().
        # If the lock is acquired, the C code is done, and self._stop() is
        # called.  That sets ._is_stopped to True, and ._tstate_lock to None.
        lock = self._tstate_lock
        if lock is None:
            # already determined that the C code is done
            assert self._is_stopped
            return

        try:
+           if lock.acquire(block, timeout):
                lock.release()
                self._stop()
        except:
            if lock.locked():
                # bpo-45274: lock.acquire() acquired the lock, but the function
                # was interrupted with an exception before reaching the
                # lock.release(). It can happen if a signal handler raises an
                # exception, like CTRL+C which raises KeyboardInterrupt.
                lock.release()
                self._stop()
            raise
```

Huh! Okay. So, it seems like maybe this is a [thread deadlock](https://stackoverflow.com/questions/34512/what-is-a-deadlock) ([source](https://stackoverflow.com/a/64617814))?


<details>
<summary> Red herring: pytest-cov </summary>

Given that the error only occurs on Linux (macOS and Windows now both pass), I am reminded of some older issues we fixed in SCT:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3661

That issue mentions the problem stems from `pytest-cov`, but we don't use code coverage anymore... do we?

> plugins: cov-6.0.0, console-scripts-1.4.1

Aha! The `pytest-cov` plugin is still being loaded, even though we've stopped using code coverage. And in fact, it's still listed in our `requirements.txt`!!!

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/db894a3c49871397a6b4c81ec29ba0b32f9dbc86/requirements.txt#L56

If I uninstall it from my venv... the resampling step still stalls. :(

</details>

My gut says this is an issue with `tqdm.contrib.process_map`, which you are using for a multiprocessing-compatible progress bar... But there are issues describing similar behavior, such as:

- https://github.com/tqdm/tqdm/issues/1228

I don't know what to suggest apart from trying to use multiprocessing in a different way... :(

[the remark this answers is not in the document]
joshuacwnewton: It looks like this may be an interaction specifically with `pytest`, since the CLI still works on Linux just fine when running outside of pytest...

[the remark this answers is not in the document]
joshuacwnewton: I tried setting the multiprocessing start method to `spawn` instead of `fork` (https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). I figured that because `fork` is the default on Linux, while `spawn` is the default on Windows and macOS, this might explain the discrepancy between passing/failing platforms.

And... it works!

joshuacwnewton (approved):
Huzzah!! Everything passes! :tada:

PR #4685, opened 2024-11-07
Use argparse subparsers for `sct_deepseg`

joshuacwnewton:
## Description

This PR updates the `sct_deepseg` script to use subparsers (i.e. 1 parser per task). This has a number of benefits:

- Shorter command syntax
- Per-task arguments (i.e. if a single task needs a specific argument, we no longer have to specify something like "this argument is only used for XXX task" in the help description)
- Better documentation:
   - Dedicated argparse help output (descriptions, citation, etc.) for each task
   - Echoing the argparse improvements to the `sct_deepseg` model gallery, too (allowing for more useful linkable per-task doc pages)

> [!NOTE]
> There was a previous version of this PR (#4630). However, that PR was closed and split into two. The first PR (adding the SCT DeepSeg Model Gallery -- Old syntax) was merged for SCT v6.5 in #4684.
>
> This is the second PR, which adds subparsers to `sct_deepseg` (i.e. individual help and docs pages for each model), and updates the model gallery to match the new changes. It should be merged for SCT v7.0.

## Linked issues

Fixes #4318.

PR #4695, opened 2024-11-13
Update contrast-agnostic model to `r2024113`

naga-karthik:
## Description

This PR updates the contrast-agnostic model to its latest version `r20241113`. In addition to updating the URL of the model, I also added the citation info which is printed to the terminal when using the model (see commit fd853e3148176636489b81879e0ccc8fb8e9008a). I feel like users aren't aware of what specific paper the model refers to most of them only cite SCT as a whole. I believe printing the citation info on the terminal (just as how nnUNet commands do) would be one step towards mitigating the issue. 

### Usage instructions for training the new model:

* Assuming you already have a working installation of SCT, checkout to my branch by running `git checkout nk/update-contrast-agnostic-model`
* Run `sct_deepseg -install-task seg_sc_contrast_agnostic` to download the latest model.
* Run `sct_deepseg -task seg_sc_contrast_agnostic -i <path-to-image.nii.gz` to get the segmentations.

<details>

<summary> Output </summary>

```bash

--
Spinal Cord Toolbox (git-nk/update-contrast-agnostic-model-fd853e3148176636489b81879e0ccc8fb8e9008a)

sct_deepseg -task seg_sc_contrast_agnostic -i temp_data/sub-amuAL_T2w.nii.gz
--


#######################################################################
Please cite the following paper when using the contrast-agnostic model:

Bédard*, S., Karthik*, E.N., Tsagkas, C., Pravatà, E., Granziera, C., Smith, A.C., Weber, K.A., & Cohen-Adad, J. (2023). 
Towards contrast-agnostic soft segmentation of the spinal cord. ArXiv, abs/2310.15402. https://arxiv.org/abs/2310.15402
#######################################################################

Model 'model_seg_sc_contrast_agnostic_softseg_monai' is up to date (Source: https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/r20241113/model_contrast_agnostic_20241018-0508.zip)
Creating temporary folder (/tmp/sct_2024-11-13_09-02-48_sct_deepseg_9adsc8ig)
Copied temp_data/sub-amuAL_T2w.nii.gz to /tmp/sct_2024-11-13_09-02-48_sct_deepseg_9adsc8ig/sub-amuAL_T2w.nii.gz
Starting inference...
Inference done.
Total inference time: 0 minute(s) 15 seconds
Saving results to: /tmp/sct_2024-11-13_09-02-48_sct_deepseg_9adsc8ig/sub-amuAL_T2w_seg.nii.gz

```

</details>

Thoughts on this, @joshuacwnewton, @valosekj ?

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -122,7 +122,7 @@ MODELS = {
     #       - Binarization is applied within SCT code
     "model_seg_sc_contrast_agnostic_softseg_monai": {
         "url": [
-            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v2.5/model_contrast-agnostic_20240930-1002.zip"
+            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/r20241113/model_contrast_agnostic_20241018-0508.zip"
         ],
         "description": "Spinal cord segmentation agnostic to MRI contrasts using MONAI-based nnUNet model",
         "contrasts": ["any"],
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -132,11 +132,10 @@ MODELS = {
                                     #######################################################################
                                     Please cite the following paper when using the contrast-agnostic model:
 
-                                    Bédard*, S., Karthik*, E.N., Tsagkas, C., Pravatà, E., Granziera, C., Smith, A.C., Weber, K.A., & Cohen-Adad, J. (2023). 
+                                    Bédard*, S., Karthik*, E.N., Tsagkas, C., Pravatà, E., Granziera, C., Smith, A.C., Weber, K.A., & Cohen-Adad, J. (2023).
                                     Towards contrast-agnostic soft segmentation of the spinal cord. ArXiv, abs/2310.15402. https://arxiv.org/abs/2310.15402
                                     #######################################################################
                                 """),
-                                    
     },
     "model_seg_sci_multiclass_sc_lesion_nnunet": {
         "url": [
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -128,6 +128,15 @@ MODELS = {
         "contrasts": ["any"],
         "thr": 0.5,  # Softseg model -> threshold at 0.5
         "default": False,
+        "citation": textwrap.dedent("""
+                                    #######################################################################
+                                    Please cite the following paper when using the contrast-agnostic model:
+
+                                    Bédard*, S., Karthik*, E.N., Tsagkas, C., Pravatà, E., Granziera, C., Smith, A.C., Weber, K.A., & Cohen-Adad, J. (2023). 
+                                    Towards contrast-agnostic soft segmentation of the spinal cord. ArXiv, abs/2310.15402. https://arxiv.org/abs/2310.15402
+                                    #######################################################################
+                                """),
+                                    
     },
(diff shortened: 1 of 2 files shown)


valosekj:
> I also added the citation info which is printed to the terminal when using the model (see commit [fd853e3](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/fd853e3148176636489b81879e0ccc8fb8e9008a)).

I like this idea!

spinalcordtoolbox/deepseg/models.py
```diff
@@ -122,12 +122,20 @@
...
         "thr": 0.5,  # Softseg model -> threshold at 0.5
         "default": False,
+        "citation": textwrap.dedent("""
+                                    #######################################################################
+                                    Please cite the following paper when using the contrast-agnostic model:
+
+                                    Bédard*, S., Karthik*, E.N., Tsagkas, C., Pravatà, E., Granziera, C., Smith, A.C., Weber, K.A., & Cohen-Adad, J. (2023).
+                                    Towards contrast-agnostic soft segmentation of the spinal cord. ArXiv, abs/2310.15402. https://arxiv.org/abs/2310.15402
+                                    #######################################################################
+                                """),
```
joshuacwnewton, line 138: It turns out that we had similar thoughts! I had prototyped a `citation` field in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4685, too!! I think it is a very good idea, and we should strive to add this for all our models. :hearts:

How do you feel about using the bibtex format instead? Similar to what was done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4685/commits/1167734cc07fdb25b06646d7a6b162ce717ea507.

Also, I think perhaps the "Please cite the following paper" text should live inside of `sct_deepseg`, so that we don't duplicate it for every single model. That way, `"citation"` contains only the citation and no other extraneous text.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -477,6 +485,18 @@ def is_up_to_date(path_model):
...
+def get_citation_info(name_model):
+    """
+    Get citation information for a given model.
+
+    :param name_model: str: Name of model.
+    :return: str: Citation information.
+    """
+    if name_model not in MODELS:
+        raise ValueError(f"Model '{name_model}' not found in MODELS")
+    return MODELS[name_model].get('citation', None)
```
joshuacwnewton, line 497: I don't think we need a whole new function for this? The code is simple enough that we can just put it directly inside `sct_deepseg`.

Also, since `sct_deepseg` already contains the conditional `if name_model in list(models.MODELS.keys()):`, we probably don't need the `ValueError` check since we can safely assume that the model will be present.

joshuacwnewton (changes requested):
Thanks for this!! It mostly LGTM, I just had some small feedback about the citation feature.

joshuacwnewton:
Closing due to discussion on Slack re: poorer results due to investigation from @naga-karthik:

> I QC’d for the v2.5 model on the spider-challenge-2023 dataset along with the GT predictions. 
> 
> The predictions are quite bad on these images and the GT masks themselves are questionable. I am considering removing this dataset entirely and keep the our existing v2.5 model which was doing well until now. We already have a few lumbar images in the training set (i.e. lumbar-epfl, lumbar-vanderbilt and a few lumbar images from SCI datasets)
> 
> I also feel that there is a limit until which we can make the model “contrast-agnostic” after which we get diminishing returns (i.e. the upwards trend of the CSA error that I showed in the last meeting). Also, the features learned by the model to segment cervical/thoracic spine might be different from lumbar spine and hence adding more and more lumbar data is affecting the performance of the other spinal regions.

So, we will stick with the existing model for now, and instead use other models to target the lumbar region (e.g. TotalSpineSeg).

PR #4697, opened 2024-11-13
Fix bug due to discrepancy introduced by `set_qform` and `set_sform` methods

### LINKED ISSUE #4689: `change_orientation` causes sform/qform mismatch (`set_qform` seems to have no effect)
A [user on the forum](https://forum.spinalcordmri.org/t/dog-template-registration-qform-sform-error/1324/1) encountered an sform/qform mismatch error during `sct_register_to_template`, despite their input data having matching sform/qforms.

The mismatch was introduced here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ff0945d455521598474e99f31e4679f0015985b0/spinalcordtoolbox/scripts/sct_register_to_template.py#L448-L450

When I set a breakpoint before and after reorientation, I see that `im_src` has roughly the same affines (within our `check_sform` tolerance, at least) before:

```python
>>> Image(ftmp_seg).header.get_sform() - Image(ftmp_seg).header.get_qform()
array([[ 1.24708333e-08, -1.99139080e-10,  3.86632998e-11, 0.00000000e+00],
       [ 1.36003458e-10,  1.79459173e-08, -2.86510378e-10, 0.00000000e+00],
       [-5.80472270e-11,  3.40933358e-12,  1.17948034e-08, 0.00000000e+00],
       [ 0.00000000e+00,  0.00000000e+00,  0.00000000e+00, 0.00000000e+00]])
```

However, after reorientation, the difference is much larger:

```python
>>> Image(ftmp_seg).header.get_sform() - Image(ftmp_seg).header.get_qform()
array([[-1.23808742e-06, -8.25844864e-06, -1.41198199e-03, 0.00000000e+00],
       [ 8.76812599e-06, -1.79952582e-08, -4.15943024e-06, 0.00000000e+00],
       [-1.50030867e-03, -4.14610448e-06,  1.16511079e-06, 0.00000000e+00],
       [ 0.00000000e+00,  0.00000000e+00,  0.00000000e+00, 0.00000000e+00]])
```

Checking inside of the `change_orientation` function, I see that we're supposed to have a step that should eliminate any differences:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/96eaaaa9739859df2584868a4d07455766420a52/spinalcordtoolbox/image.py#L1165-L1166

However, when I manually run the `set_qform()` call, nothing seems to happen! There is no change in the result of `get_qform()` before and after setting it. 

I will try to find out why this is happening, and possibly report upstream to `nibabel`.

joshuacwnewton:
## Description

This PR ensures that the qform will always match the sform when reorienting the image.

Note that with the old method:

- sform == im_dst_aff
- qform == [orthogonal(im_dst_aff)](https://github.com/nipy/nibabel/blob/d9c479a03df4119fcac0ffd3bd1b2b0fabf650e7/nibabel/nifti1.py#L1251-L1260)

And because orthogonal(im_dst_affine) may not always equal im_dst_aff, the two `set_xform` commands may result in different matrices.

Depending on how we utilize the `sform`, this change may result in numerical differences? So, I am opening this PR also as a way of quickly checking our test suite.

## Linked issues

Fixes #4689.

### DIFF
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1162,8 +1162,12 @@ def change_orientation(im_src, orientation, im_dst=None, inverse=False):
         im_src_data.shape)
     im_dst_aff = np.matmul(im_src_aff, aff)
 
+    # NB: When setting the xforms, the qform will be made orthogonal (to meet NIfTI1 requirements),
+    #     while the sform won't. Since im_dst_aff doesn't always equal orthogonal(im_dst_aff), this
+    #     could introduce a discrepancy. So, we first set the qform, then we set the sform to the
+    #     qform (i.e. orthogonal(im_dst_aff)), which ensures that the two matrices are identical.
     im_dst.header.set_qform(im_dst_aff)
-    im_dst.header.set_sform(im_dst_aff)
+    im_dst.header.set_sform(im_dst.header.get_qform())
     im_dst.header.set_data_shape(data.shape)
     im_dst.data = data
 


mguaypaq (approved):
Thank you for digging and fixing this! I was vaguely aware that the qform matrix was more restricted than the sform matrix, and I was a bit surprised that it didn't seem to cause any problems (but this here now is exactly the kind of problem that would happen!).

I believe you're 100% right that copying from the qform to the sform (rather than the other way around) is the safe thing to do.

PR #4564, opened 2024-07-18
Avoid buggy version of `nnunetv2` that produces empty output predictions

joshuacwnewton:
## Description

Given the comments in `requirements.txt`, it seems like 2.3.1 is the only safe version of `nnunetv2` we can rely on for now.

Action item for myself: 

- [ ] Report this upstream and help debug why this issue is occurring in the first place. (Maybe after SCT v6.4, since I have a backlog I need to work through before I can spend time digging into a tricky upstream issue.)

## Testing

Testing this is a bit tricky, because our test (on cropped input) still works fine on 2.4.2? The problem only reveals itself on a full test image. 

Maybe we could write a test similar to the mouse test, and stick the large image in the new testing dataset?

## Linked issues

Fixes #4556.

PR #4699, opened 2024-11-18
Allow `nnunetv2>=2.5.1` due to bugfix for previous issue with `2.4.2`

### LINKED ISSUE #4556: Empty prediction for `-task seg_sc_lesion_t2w_sci` when using `nnunetv2==2.4.2`

joshuacwnewton:
## Description

Previously, we encountered a bug with `nnunetv2==2.4.2` where an empty segmentation would be produced by our SCI model. We temporarily limited `nnunetv2==2.3.1`, since that was seemingly the only version that worked for us. See:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4564

However, there is a newer version of `nnunetv2` (`2.5.1`) that seems to fix the "empty segmentation" problem. 

Note that TotalSpineSeg is incompatible with 2.5.1, so we're probably going to revert back to 2.3.1 immediately. But, this PR will prepare for an eventual 2.5.2 release that will finally work for both SCT and TotalSpineSeg.

## Linked issues

Fixes #4556.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,11 +27,12 @@ monai[nibabel]
 msvc-runtime==14.29.30133; sys.platform == 'win32' # append_to_freeze
 nibabel
 nilearn
-# Only allow nnunetv2==2.3.1 or 2.5.1 and above (NB: there is no 2.5.0)
-# 2.4.2 causes empty predictions (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4556)
-# 2.4.0/2.4.1 fail during inference due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4444
+# Only allow nnunetv2==2.3.1 or 2.5.1 (NB: there is no 2.5.0)
 # 2.3.0 and below are incompatible due to breaking API change (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4380)
-nnunetv2>2.3.0,!=2.4.*
+# 2.4.0/2.4.1 fail during inference due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4444
+# 2.4.2 causes empty predictions (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4556)
+# 2.5.2 and above will be tested manually when they are released.
+nnunetv2>2.3.0,!=2.4.*,<=2.5.1
 # SCT is compatible, however the upgrade to numpy==2.0.0 broke *many* upstream packages. Let's try again later. See https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4535
 numpy<2
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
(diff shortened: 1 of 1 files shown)


mguaypaq:
Given that `nnunetv2` has had a lot of buggy or backwards incompatible releases, do we actually want to switch back to a permissive version spec (that is, one with a `>=`)? On the other hand, I'm not sure what kind of workflow we would use for upgrading this dependency if we always kept a strict version pin.

joshuacwnewton:
> On the other hand, I'm not sure what kind of workflow we would use for upgrading this dependency if we always kept a strict version pin.

Maybe what we could do is set a permissive upper bound (e.g. `<=2.5.1`), then I could turn on notifications for new `nnunetv2` releases, and whenever a new version is released, we manually test it. Then, after we've verified that it works, we update the upper bound (e.g. `<=2.5.2`).

It would require a little manual work, but I'm alright with that, because as you say... new `nnunetv2` releases can't really be trusted. :sweat:

mguaypaq:
This would probably be a lot of up-front work, but: what if we had at least one regression test for each `nnunetv2` model/task? (I mean something like the `batch_processing.sh` tests, where we have an input/saved output pair that we run and check for output equality.) Since it's likely that we need some private test data, it could live in https://data.neuro.polymtl.ca/datasets/sct-testing-release.

That way, we could have a permissive version spec in dev, and still have some trust that we would catch any regressions before pinning each release to a specific version.

But maybe that's too pie-in-the-sky.

mguaypaq (approved):

Issue #4607, opened 2024-08-22
`sct_analyze_lesion` - output lesion length and width for the midsagittal slice 


valosekj:
`sct_analyze_lesion` currently outputs the lesion length [considering all lesion slices](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L569). For example, if the lesion spans axial slices 10 to 20 in the midsagittal slice and axial slices 6 to 12 in the parasagittal slice, the length is measured from slices 6 to 20. Making it a kind of 3D measurement.

One of our collaborators, @dpfyffer, asked me by email whether it would also be possible to output the lesion length for the midsagittal slice only (in addition to the current length measurement). I believe it should be definitely doable. I will take a look at it in the next few days!

UPDATE 2024-08-28: Dario asked whether it would be possible to implement also lesion width (defined as maximum lesion length in the A-P axis) on the midsagittal slice. I will add both features (lesion length and lesion width) in the following PR.

joshuacwnewton:
Reopening this PR until https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4631 is merged, due to the change in method to determine the midsagittal slice for even-numbered slice ranges.

PR #4702, opened 2024-11-20
Interpolate mid-sagittal slices using two slices

valosekj:
#### GitHub

#### PR contents

## Description

This PR follows https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656 (#4656 base branch was this `jv/change_get_midsagittal_slice_logic` branch).
This PR is open into `master`.

TL;DR: we now interpolate the midsagittal slice using the two closest sagittal slices using the following logic:

image

See https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4656 description for details.

Additionally, besides computing metrics for the midsagittal slice (such as `length_interpolated_midsagittal_slice`, `width_interpolated_midsagittal_slice`, ...), this PR adds computation of a new 3D metric (i.e., considering all slices covered by the lesion): lesion `width` defined as maximal width along the Anterior-Posterior (AP) axis across all slices. 

## Related Issues/PRs

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4607.


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -143,16 +146,18 @@ def get_parser():
...
         self.path_ofolder = path_ofolder
         self.verbose = verbose
         self.wrk_dir = os.getcwd()
         # NOTE: the tissue bridges are NOT included in self.measure_keys because we do not want to average them
-        self.measure_keys = ['volume [mm3]', 'length [mm]', 'max_equivalent_diameter [mm]', 'max_axial_damage_ratio []']
+        self.measure_keys = ['volume [mm3]', 'length [mm]', 'width [mm]',
+                             'max_equivalent_diameter [mm]', 'max_axial_damage_ratio []']
+        self.interpolated_midsagittal_slice = None      # target float sagittal slice number used for the interpolation. This number is based on the spinal cord center of mass.
+        self.interpolation_slices = None                # sagittal slices used for the interpolation
+        self.interpolation_factor = None                # factor used for the interpolation of sagittal slices
```
joshuacwnewton, line 160: This is just a nitpick, but I like making changes that allow the PR diff to be easy to read.

Here, instead of adding the new attributes to the end of the list, we could put them in the same location the old attributes used to be. That way, it is clear that we went from `self.midsagiital_sc_slice` to `self.interpolated_midsagittal_slice`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -214,7 +219,6 @@ def analyze(self):
         #     no angle correction is needed
         if self.fname_sc is not None:
             self.angle_correction()
-            self.get_midsagittal_slice()
```
joshuacwnewton, line 217: I notice that the call to `self.get_midsagittal_slice()` is removed here, but it isn't replaced with the equivalent call to `self.midsagittal_slice_interpolation()`. Instead, the call is placed inside the `measure` method, with a bunch of extra logic:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4cfb3673da6e40e283f0a48992945c99ac1fe01e/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L851-L862

Maybe what we could do is take the "get largest lesion" logic and put it inside of the midsagittal slice method, keep the old name `get_midsagittal_slice`, and put it in the same location it used to live at.

That way, it's very clear in the diff that all we're doing is updating the logic that we use to get the midsagittal slice, with the overall structure staying the same.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -817,6 +848,22 @@ def measure(self):
...
+        largest_lesion_idx = self.measure_pd.loc[pd.to_numeric(self.measure_pd['volume [mm3]']).idxmax()]['label']
+        printv(f'Largest lesion index: {largest_lesion_idx}', self.verbose, 'info')
+        # Get sagittal slices and compute weights to be used for the midsagittal slice interpolation
+        # NOTE: the interpolation will be done separately for each lesion in the for loop below
+        im_lesion_data_largest_lesion = np.copy(im_lesion_data == largest_lesion_idx)
+        self.midsagittal_slice_interpolation(im_lesion_data_largest_lesion)
+        # NOTE: the interpolated midsagittal slice is the same for all lesions
+        printv(f'Interpolated midsagittal slice (same across lesions) = '
+               f'{round(self.interpolated_midsagittal_slice, 2)}', self.verbose, 'info')
+
```
joshuacwnewton, line 866: See: Previous comment...

All of this extra logic could live in the same location as the old `get_midsagittal_slice` function (to keep the diff neat and tidy).

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -33,13 +33,16 @@ def get_parser():
 
               - volume `[mm^3]`: volume of the lesion
               - length `[mm]`: maximal length along the Superior-Inferior (SI) axis across all slices
+              - width `[mm]`: maximal width along the Anterior-Posterior (AP) axis across all slices
               - max_equivalent_diameter `[mm]`: maximum diameter of the lesion, when approximating the lesion as a circle in the axial plane
               - max_axial_damage_ratio `[]`: maximum ratio of the lesion area divided by the spinal cord area
-              - midsagittal_spinal_cord_slice: midsagittal slice number of the spinal cord defined based on the spinal cord segmentation
```
joshuacwnewton, line 38: It looks like [some of the tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/11939714655/job/33280820096?pr=4702#step:5:1941) are failing because we got rid of this key:

```
 KeyError: 'midsagittal_spinal_cord_slice'
```

If I Ctrl+F the project for this key, I find a number of instances that will need to be updated:

image

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -888,19 +937,93 @@ def angle_correction(self):
...
+                        factor = 8.7 – int(8.7)
+                        factor = 8.7 – 8.0
+                        factor = 0.7
+                c. Calculate the interpolated slice:
+                        interp_slice = (1 – 0.7) * slice_1 + 0.7 * slice_2
+                        (slice_1 and slice_2 are 2D arrays)
+        :param im_lesion_data: 3D numpy array, RPI-oriented mask of the largest lesion
+        """
+        # Get the RPI-oriented spinal cord mask
         im_sc_data = Image(self.fname_sc).data
```
joshuacwnewton, line 1003: [Some of the tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/11939714655/job/33280820096?pr=4702#step:5:1945) are failing because the `midsagittal_slice_interpolation` method requires the spinal cord segmentation, yet we call it regardless of whether an SC seg was passed:

```
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 862, in measure
    self.midsagittal_slice_interpolation(im_lesion_data_largest_lesion)
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 1003, in midsagittal_slice_interpolation
    im_sc_data = Image(self.fname_sc).data
  File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/image.py", line 322, in __init__
TypeError: Image constructor takes at least one argument.
```

Previously, we had the conditional to prevent us from running into the error. But, because the location of the `midsagittal_slice` method changed, we no longer have that guarantee.

This error should be fixed if we move the call to the `midsagittal_slice` method back to its old location.

joshuacwnewton (changes requested):
Now that the tests are running, I have quite a bit of feedback! (Sorry for the big review!)

Maybe it is just enough to address the final two comments (about the failing tests), depending on whether your goal is to make these changes live for v6.5, or whether this PR can wait until v7.0.

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -53,7 +53,12 @@ def compute_expected_measurements(dim, starting_coord=None, path_seg=None):
 
         # Find the midsagittal slice of the spinal cord (assuming AIL input image)
         nonzero_slices = np.unique(np.where(data_seg)[2])  # AIL image: [2] -> LR (sagittal)
-        mid_sagittal_slice = int(np.mean([np.min(nonzero_slices), np.max(nonzero_slices)]))
+        mid_idx = len(nonzero_slices) // 2
+        if len(nonzero_slices) % 2:
+            mid_sagittal_slice = nonzero_slices[mid_idx]
+        else:
+            slice1, slice2 = nonzero_slices[mid_idx - 1:mid_idx + 1]
+            mid_sagittal_slice = slice1 if np.sum(data_seg[:, :, slice1]) > np.sum(data_seg[:, :, slice2]) else slice2
```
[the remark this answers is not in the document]
valosekj: Good question! The `midsagittal_slice_interpolation` has been updated (to the logic in the diagram in the [PR description](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702#issue-2676704262)). 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4cfb3673da6e40e283f0a48992945c99ac1fe01e/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L1005-L1026

I need to update also the tests.

valosekj:
> Now that the tests are running, I have quite a bit of feedback! (Sorry for the big review!)
> 
> Maybe it is just enough to address the final two comments (about the failing tests), depending on whether your goal is to make these changes live for v6.5, or whether this PR can wait until v7.0.

Thank you for your review! I will try to address your comments on Friday. Depending on my success, we can decide then whether to proceed with v6.5 or v7.0.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: good catch. fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/46fa51a4015bd019cfd8a21a19b44e6db778ccaa

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: excelent point. done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/11cdd6567853945cd189a02a38da4c5b1838fc45

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/11cdd6567853945cd189a02a38da4c5b1838fc45

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: fair point. thank you. reordered in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/943c83a5aca2203b431741e6f7143022b4f4e612

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -116,9 +116,9 @@ def compute_expected_measurements(dim, starting_coord=None, path_seg=None, mid_s
     ([(29, 45, 25), (3, 10, 2)], 0.001),
     ([(29, 27, 25), (1, 4, 1)], 0.001),  # NB: Example from #3633
     # Curved region of `t2.nii.gz` -> lots of curvature -> larger tolerance
-    ([(29, 0, 25), (4, 15, 3)], 0.01),
+    ([(31, 0, 25), (4, 15, 3)], 0.01),
     # Multiple lesions
-    ([[(29, 0, 25), (4, 15, 3)],
+    ([[(31, 0, 25), (4, 15, 3)],
```
valosekj, line 123: I took the liberty to slightly shift the lesion to be entirely in the spinal cord.

<details><summary>old: (29, 0, 25)</summary>

Screenshot 2024-12-13 at 3 39 34 PM

</details>

<details><summary>new (31, 0, 25)</summary>

Screenshot 2024-12-13 at 3 40 51 PM

</details>

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Good catch. Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/8750f7c6bbb05940e9aeacd60706ac8eead67e46

testing/cli/test_cli_sct_analyze_lesion.py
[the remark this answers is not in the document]
valosekj: I updated the tests to be compatible with the new logic interpolating two slices in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/e05b992bbac979baee59c7aef4b45a5687d282bf

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -888,19 +937,93 @@ def angle_correction(self):
...
+        # 4. Calculate target position in right-left axis (x direction) for the interpolation (mean of spinal cord
+        # center of mass (in the x-axis (R-L direction))
+        x_target = np.mean(stored_x_coordinates)    # e.g., for [8.6, 8.6, 8.9, 8.8, 9.6] --> 8.7
+        self.interpolated_midsagittal_slice = x_target     # store it to output in the output XLS file
+        # 5. Interpolate the lesion
+        slice1 = int(np.floor(x_target))     # e.g., 8
+        slice2 = int(np.ceil(x_target))      # e.g., 9
+        interpolation_factor = x_target - int(x_target)    # e.g., 8.7 - 8 = 0.7
+        self.interpolation_slices = [slice1, slice2]     # store it to be used for tissue bridges
+        self.interpolation_factor = interpolation_factor    # store it to be used by the `_interpolate_values` function
```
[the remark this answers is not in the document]
valosekj: Good points! Dropping `self.interpolation_factor` was pretty easy, indeed. Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/37beb424cec5373f3fd6e8a2e180ed4fdf64094b.

`self.interpolation_slices` is used in `_interpolate_lesion` on [L952-L953](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/675fba79a820e83e29ceafa225b1efadc8b20025/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L952-L953) but also in `_measure_interpolated_tissue_bridges` on [L371](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/675fba79a820e83e29ceafa225b1efadc8b20025/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L371) and  [L389-L390](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/675fba79a820e83e29ceafa225b1efadc8b20025/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L389-L390). So calculation to get these two values (`8/9` in the dummy example above) would need to be done several times. So maybe it's better to keep the `self.interpolation_slices` variable?

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -89,8 +89,8 @@ def compute_expected_measurements(dim, starting_coord=None, path_seg=None, mid_s
         # NB: we can compute length_midsagittal_slice and width_midsagittal_slice here from dim for the purposes of
         #  testing, but in the actual script, we need the spinal cord segmentation to compute these values based on
         #  the midsagittal slice
-        'length_midsagittal_slice [mm]': dim[1],
-        'width_midsagittal_slice [mm]': dim[0],
+        # 'length_midsagittal_slice [mm]': dim[1],
+        # 'width_midsagittal_slice [mm]': dim[0],
```
valosekj, line 94: Because the midsagittal length and width are due to the new logic computed from the interpolated midsagittal slices (see the functions below), I commented out these two lines to avoid test crashing. I'm not sure how we can now easily approximate the midsagittal length and width for testing purposes.  

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/37beb424cec5373f3fd6e8a2e180ed4fdf64094b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L575

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/37beb424cec5373f3fd6e8a2e180ed4fdf64094b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L634

joshuacwnewton (changes requested):
I just have one more review related to code organization + fixing the GitHub diff. (Specifically, I am using the "Side by side" view, so that I can compare equivalent lines between the old and the new.)

I'm sorry to be so nitpicky about the GitHub diff, hehe. My goal here is just to be able to keep this PR neat and tidy for posterity, so if we ever have to go back and analyze the changes we can easily see:

1. Which methods were modified (e.g. `get_midsagittal_slice`, `_measure_width_midsagittal_slice`, `_measure_length_midsagittal_slice`, etc.)
2. Which methods are newly added (e.g. `_measure_width`, `_interpolate_values`, `_interpolate_lesion`)

The more we can make the diff straightforward to read, the easier it will be to understand when reviewing.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -564,74 +574,99 @@ def _measure_length(self, im_data, p_lst, idx):
...
 
-        # Check if the lesion exists in the midsagittal slice, if not, set the length to 'np.nan' and return
-        if not np.any(im_data_midsagittal):
-            length_cur = np.nan
-        else:
-            # Compute the length of the lesion along the superior-inferior axis in the midsagittal slice
-            # The length is computed as the sum of the angle corrected axial slice thicknesses
-            length_cur = np.sum([p_lst[2] / np.cos(self.angles_sagittal[s])  # p_lst[2] -> pixel size of SI axis
-                                 for s in nonzero_axial_slices])
+    def _measure_width(self, im_data, p_lst, idx):
```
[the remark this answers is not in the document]
valosekj: Good point! Agree, methods reordered in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/14f5473386796efc4dd9a1fd142b6f8bdaab8c1e

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -143,16 +146,17 @@ def get_parser():
 class AnalyzeLesion:
     def __init__(self, fname_mask, fname_sc, fname_ref, path_template, path_ofolder, perslice, verbose):
         self.fname_mask = fname_mask
-        self.midsagittal_sc_slice = None
-        self.midsagittal_sc_slice_rpi = None
+        self.interpolated_midsagittal_slice = None  # target float sagittal slice number used for the interpolation. This number is based on the spinal cord center of mass.
+        self.interpolation_slices = None                # sagittal slices used for the interpolation
```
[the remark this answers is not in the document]
valosekj: 👍🏻 done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/53389c5f3f40fa5fb95522ecbc07848d1ca73331

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -888,19 +928,112 @@ def angle_correction(self):
...
+        """
+        # Interpolate the two sagittal slices; 3D --> 2D, dim=[AP, SI]
+        im_lesion_interpolated = self._interpolate_values(im_lesion_data[self.interpolation_slices[0], :, :],
+                                                          im_lesion_data[self.interpolation_slices[1], :, :])
+        # Fetch a list of axial slice numbers that are nonzero in the interpolated midsagittal slice
+        nonzero_axial_slices = np.unique(np.where(im_lesion_interpolated)[1])  # [1] -> SI
+
+        return im_lesion_interpolated, nonzero_axial_slices
+
+    def get_midsagittal_slice(self, im_lesion_data, label_lst, p_lst):
```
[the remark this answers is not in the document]
valosekj: Good catch! Thanks! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/705cb70e71a95719169edecc9ed679132a889cdb

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -214,7 +218,6 @@ def analyze(self):
         #     no angle correction is needed
         if self.fname_sc is not None:
             self.angle_correction()
-            self.get_midsagittal_slice()
```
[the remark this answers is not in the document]
valosekj: Good question! I moved `self.get_midsagittal_slice()` from `analyze()` to `measure()` because `self.get_midsagittal_slice()` now uses both spinal cord and lesion masks to determine the midsagittal slices. Previously, the function used only the spinal cord segmentation so it could be called from `analyze()`. But now as also the lesion mask is needed, I found it better to move the `self.get_midsagittal_slice()` call into `measure()` because lesion mask is already available in this method:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/14f5473386796efc4dd9a1fd142b6f8bdaab8c1e/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L806-L808

valosekj:
> I just have one more review related to code organization + fixing the GitHub diff. (Specifically, I am using the "Side by side" view, so that I can compare equivalent lines between the old and the new.)
> ...
> The more we can make the diff straightforward to read, the easier it will be to understand when reviewing.

Thank you, @joshuacwnewton! Completely agree! The diff is much cleaner after reorganizing the code according to your comments.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
joshuacwnewton: Ahhh, gotcha!! Alright then, that makes perfect sense. :)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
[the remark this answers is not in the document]
joshuacwnewton: Agreed!

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton: I think since we have such simple dummy lesions (rectangles), we don't need to be exact with our measurement logic?

In fact, I think we can actually still re-use `dim[1]` and `dim[0]` exactly as before, as there's no difference between the length/width for the slices +/- 1 and the interpolated slice itself.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -33,13 +33,16 @@ def get_parser():
 
               - volume `[mm^3]`: volume of the lesion
               - length `[mm]`: maximal length along the Superior-Inferior (SI) axis across all slices
+              - width `[mm]`: maximal width along the Anterior-Posterior (AP) axis across all slices
```
joshuacwnewton, line 36: Just for my own understanding: We use SI and AP for length/width because we're using a sagittal FOV (i.e. LR slices)?

If so, maybe this could be a helpful clarification:

```suggestion
              - length `[mm]`: maximal length along the Superior-Inferior (SI) axis across all sagittal slices of the lesion
              - width `[mm]`: maximal width along the Anterior-Posterior (AP) axis across all sagittal slices of the lesion
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -888,19 +928,112 @@ def angle_correction(self):
...
+        :param data1: 2D numpy array (slice 1) or single int64 (tissue bridge for slice 1)
+        :param data2: 2D numpy array (slice 2) or single int64 (tissue bridge for slice 2)
+        :return: 2D numpy array (interpolated slice) or single float64 (interpolated tissue bridge)
+        """
+        interpolation_factor = self.interpolated_midsagittal_slice - int(self.interpolated_midsagittal_slice)   # e.g., 8.7 - 8 = 0.7
+        return (1 - interpolation_factor) * data1 + interpolation_factor * data2
+
+    def _interpolate_lesion(self, im_lesion_data):
+        """
+        Interpolate the currently processed lesion
```
joshuacwnewton, line 1025: I found this method name to be a bit confusing? To me, it's not clear what "interpolate_lesion" actually means...

Maybe a name like `_get_lesion_midsagittal_slice` could be a bit more descriptive? (That way, it matches the naming convention of `_compute_width_midsagittal_slice` and `_compute_length_midsagittal_slice`.)


```suggestion
    def _get_lesion_midsagittal_slice(self, im_lesion_data):
        """
        Get the lesion data from the midsagittal slice by interpolating between the
        the slices above and below.
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -553,7 +563,7 @@ def _measure_tissue_bridges(self, im_lesion_data, p_lst, idx):
 
     def _measure_length(self, im_data, p_lst, idx):
```
joshuacwnewton, line 564: I notice that the `_measure_length` method is quite short in comparison to the brand new `_measure_width` function, which computes `width` for all nonzero axial slices, then takes the max.

That made me wonder: Is the `_measure_length` logic wrong? Should we be computing `length` for all nonzero sagittal slices, then taking the max?

Instead, it seems like length is computed using any and all nonzero SI slices, regardless of whether those SI slices are all present in a single sagittal slice, which seems to be a different "maximal" computation compared to `_measure_width`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Ah, but we actually use `axial` slices in `_measure_width` for computing the max AP width... same result, but minor inconsistency, hehe.

joshuacwnewton (changes requested):
Thank you so so much for the cleanup! This PR is so much easier to review now. :hearts:

Most of my comments are just me trying to understand the PR further, but I did have some changes for the test as well.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: > maybe this could be a helpful clarification

sounds good 👍🏻

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Sounds good; the method renamed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/e131faaa8ee230021508fd81a432fd5ab7df057c

> interpolating between the slices above and below.

just a minor clarification here: we interpolate two sagittal slices (R-L axis), not the slices above and below. I clarified this in the method docstring

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: > I notice that the _measure_length method is quite short in comparison to the brand new _measure_width function, which computes width for all nonzero axial slices, then takes the max.

Yeah, good point; I simplified the `for` cycle below into dict comprehension in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/10ea01e92fbc4c9356c70148219f129cb687297f

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/91985398ed990465adee857b2c58eb021212673d/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L582-L588

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -144,14 +143,25 @@ def test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements(dummy_les
...
+    # Get center of mass in S-I axis of the largest lesion
+    # (we can simply use the lesion coordinates as the SC curvature is not too big)
+    # AIL --> [1]
+    largest_lesion = lesion_params[0]   # largest lesion is always the first one in the list
+    z_center = int(round(np.mean(list(range(largest_lesion[0][1], largest_lesion[0][1] + largest_lesion[1][1])))))
+    z_range = np.arange(z_center - 2, z_center + 3)     # two slices above and below the lesion center of mass
+    # For each of these slices, compute the spinal cord center of mass in the R-L direction
+    # Note: as path_seg has the AIL orientation, we need to reverse the last axis from "L-R" to "R-L" using [::-1]
+    sc_com = [center_of_mass(Image(path_seg).data[:, z, ::-1])[1] for z in z_range]     # 2D slice from AIL: [1] --> L-R
+    mid_sagittal_slice = np.mean(sc_com)     # target slice in right-left axis (x direction) for the interpolation
```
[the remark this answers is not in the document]
valosekj: Very good point! Thank you! Implemented in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/5c8b470f799f87c02c6e1b5b8eab63bd79b91eff

testing/cli/test_cli_sct_analyze_lesion.py
valosekj: Yeah, just uncommenting (and fixing variable names (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/f7fc55397a6c682795a6e387593e00bcf65e5255)) seems not to be enough. Several tests are now failing when I try to run them locally...

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: I think that maybe it's not the comments that were the problem, but instead the indentation of the `p_lst` line.

The flake8 message mentions "over-indentation", and I think this is because it doesn't recognize that you're trying to align the `p_lst` line with the `: ` part of `axial_slice: `.

To fix this warning, you can do one of two things.

1: Align with the `{` bracket:

```diff
        # Get the width in mm and apply the angle correction
        width_cur_dict = {axial_slice: (np.max(np.where(im_data[:, :, axial_slice])[1]) -  # [1] -> AP; np.max returns the most ventral (anterior) element
                                        np.min(np.where(im_data[:, :, axial_slice])[1]) + 1) *  # np.min returns the most dorsal (posterior) element
-                                      p_lst[1] * np.cos(self.angles_sagittal[axial_slice])  # p_lst[1] -> pixel size of AP axis
+                         p_lst[1] * np.cos(self.angles_sagittal[axial_slice])  # p_lst[1] -> pixel size of AP axis
                          for axial_slice in np.unique(np.where(im_data)[2])}  # [2] -> SI
```

2: Add another set of `()` brackets to group the multi-line math expression, then indent by an additional ` `:

```diff
        # Get the width in mm and apply the angle correction
        # [1] -> AP axis; [2] -> SI axis; p_lst[1] -> pixel size of AP axis
-       width_cur_dict = {axial_slice: (np.max(np.where(im_data[:, :, axial_slice])[1]) -  # np.max returns the most ventral (anterior) element
-                                       np.min(np.where(im_data[:, :, axial_slice])[1]) + 1) *  # np.min returns the most dorsal (posterior) element
-                                      p_lst[1] * np.cos(self.angles_sagittal[axial_slice]) 
+       width_cur_dict = {axial_slice: ((np.max(np.where(im_data[:, :, axial_slice])[1]) -     # np.max returns the most ventral (anterior) element
+                                        np.min(np.where(im_data[:, :, axial_slice])[1]) + 1)  # np.min returns the most dorsal (posterior) element
+                                       * p_lst[1] * np.cos(self.angles_sagittal[axial_slice]))
                          for axial_slice in np.unique(np.where(im_data)[2])}
```

(I also moved the `*` to the next line to make it more obvious that we're multiplying the following terms.)

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton: When I run the tests locally now, it fails for both `length` and `width` on the interpolated slice. In all cases, it's because the length and width are smaller than the expected value.

The reason for the shorter length/width is because we compute the midsagittal slice using the **spinal cord center of mass**. This means that if the lesion is not 100% centered on the spinal cord (or if the lesion is only 1 voxel wide), then the interpolation will be done with the adjacent empty slices next to the lesion (since our dummy lesions are rectangles).

<details>

For example, in the (1, 4, 1) dummy lesion test, the midsagittal slice is `~26.0099`. However, because the lesion is only 1-voxel wide, the mask has value `1` in slice 26, and `0` in slice 27. This causes the `interpolated_lesion` to be computed as: `(0 * .0099) + (1 * (1 - .0099))` and thus the interpolated lesion is returned as this:

image

Conceptually, these values represent the proportion of the lesion that is present in a hypothetical "26.0099" LR voxel. Kind of like, the lesion is only "99% present" in that location. And thus, the length is treated as 99% of the original length (`3.96`).

</details>

This is hard to write a test for, since we would need different checks depending on whether the lesion is centered or not (i.e. if the lesion is present in both interpolation slices).

This is a hacky fix, though it's not very precise:

```
                # For interpolated measures, because the midsagittal slice is computed using the **spinal cord** center of mass, 
                # there is no guarantee that this slice will fully contain the lesion. Meaning, the averaged value may be
                # computed using empty slices, and thus it should be either close to or less than the expected value.
                if key in ['length_interpolated_midsagittal_slice [mm]', 'width_interpolated_midsagittal_slice [mm]']:
                    try:
                        np.testing.assert_allclose(measurements.at[idx, key],
                                                   expected_value, rtol=rtol)
                    except AssertionError as e:
                        if not (measurements.at[idx, key] <= expected_value):
                            raise e  # Only raise exception if the value is greater than expected
                else:
```

valosekj: Yes, exactly! Thanks for the hacky fix. I added it in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/be622cbe0ba68e1997398c6ba9c93ab508ca619e, and the tests are now passing!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Thank you, Joshua! I added the second option in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/816d3688c0ed221550071a41637e3901bef14b81 and flake8 is now happy!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -393,14 +356,62 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+
+        # If there are no axial slices, it means that the current lesion is not captured on the sagittal slices used
+        # for the interpolation. In this case, we will not compute the interpolated tissue bridges.
+        if len(axial_slices) == 0:
+            self.measure_pd.loc[idx, 'interpolated_dorsal_bridge_width [mm]'] = np.nan
+            self.measure_pd.loc[idx, 'interpolated_ventral_bridge_width [mm]'] = np.nan
+            return
+
+        for axial_slice in axial_slices:
+            # Get bridges for the 3 sagittal slices used for the interpolation
```
joshuacwnewton, line 381: Nit: 3 -> 2?

joshuacwnewton (approved):
I think this LGTM! Thank you so much for all of your hard work, and for your quick responses to the PR feedback! :hearts: 

Feel free to merge once the comment nitpick is updated. :)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: good catch 👍🏻 fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4702/commits/4727c354f03f1b63b807132d09f68bc45c009ca8

valosekj:
Thank you for your help and insightful comments, @joshuacwnewton! 🙏🏻

PR #4708, opened 2024-11-21
Add `-qc-plane` flag to allow switching the QC view to `'Sagittal'`

### LINKED ISSUE #4705: Allow `sct_deepseg` to output sagittal QC reports (for use with `seg_ms_lesion`)
Hi ! 
I am opening this issue to add some flags to `sct_deepseg` which specify which QC is wanted. 
This is motivated by the fact that when I do `sct_deepseg -task seg_ms_lesion`, I often want the QC to be in the sagittal axis. This requires to add the spinal cord segmentation as well. 
Currently, I have to do it in 2 steps:

1. First, I segment the lesions `sct_deepseg -task seg_ms_lesion`
2. Run the QC: `sct_qc -i {file} -p sct_deepseg_lesion -d {lesion_seg} -s {sc_seg} -qc {path_qc} -plane sagittal`

The flags I will add are: 
-qc-plane: the plane chosen for the QC
~~-s: the segmentation to use as the boundary for the QC~~

Tagging @Nilser3 who might be interested in this as well

plbenveniste:
#### GitHub

#### PR contents

## Description

This PR adds a possibility to perform sagittal QC for `sct_deepseg`. It displays the entire field of the view without any cropping.

## Linked issues

Fixes #4705.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -530,7 +530,7 @@ def sct_deepseg_sagittal(
     mpl_backend_agg.FigureCanvasAgg(fig)
     ax = fig.add_axes((0, 0, 1, 1))
     ax.imshow(img, cmap='gray', interpolation='none', aspect=1.0)
-    add_orientation_labels(ax, radius=radius, letters = ['S', 'I', 'P', 'A'])
+    add_orientation_labels(ax, radius=radius, letters=['S', 'I', 'P', 'A'])
     ax.get_xaxis().set_visible(False)
     ax.get_yaxis().set_visible(False)
     img_path = str(imgs_to_generate['path_background_img'])
@@ -762,7 +762,7 @@ def mosaic(img: Image, centers: np.ndarray, radius: tuple[int, int] = (15, 15), 
     return np.block([cropped[i:i+num_col] for i in range(0, len(cropped), num_col)])
 
 
-def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15), letters: tuple[str, str, str ,str] = ['A', 'P', 'L', 'R']):
+def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15), letters: tuple[str, str, str ,str] = ('A', 'P', 'L', 'R')):
     """
     Add orientation labels (A, P, L, R) to a figure, yellow with a black outline.
     """
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -762,7 +762,7 @@ def mosaic(img: Image, centers: np.ndarray, radius: tuple[int, int] = (15, 15), 
     return np.block([cropped[i:i+num_col] for i in range(0, len(cropped), num_col)])
 
 
-def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15), orientation = 'SAL'):
+def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15), orientation: str = 'SAL'):
     """
     Add orientation labels (A, P, L, R) to a figure, yellow with a black outline.
     """
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -770,7 +770,6 @@ def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15)
     #    A                    [12,  6]
     # L     R   -->  [0, 17]            [24, 17]
     #    P                    [12, 28]
-    
     letters = ['A', 'P', 'L', 'R']
     if orientation == 'RSP':
         letters = ['S', 'I', 'P', 'A']
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -254,6 +254,7 @@ def sct_deepseg(
     path_qc: str,
     dataset: Optional[str],
     subject: Optional[str],
+    plane: Optional[str],
 ):
     """
     Generate a QC report for sct_deepseg, based on which task was used.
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/reports/qc2.py
```diff
@@ -464,6 +468,103 @@ def sct_deepseg_spinal_rootlets_t2w(
...
+    # (e.g. WxH = 20x3 slice images). So, we want to reduce the fig height to match this.
+    # `size_fig` is in inches. So, dpi=300 --> 1500px, dpi=100 --> 500px, etc.
+    size_fig = [5, 5 * img.shape[0] / img.shape[1]]
+
+    fig = mpl_figure.Figure()
+    fig.set_size_inches(*size_fig, forward=True)
+    mpl_backend_agg.FigureCanvasAgg(fig)
+    ax = fig.add_axes((0, 0, 1, 1))
+    ax.imshow(img, cmap='gray', interpolation='none', aspect=1.0)
+    add_orientation_labels(ax)
```
plbenveniste, line 533: This needs to be fixed since the image has been re-oriented.

plbenveniste: Fixed!

plbenveniste:
This is a rather small change. Could this be added in SCT v6.5 since we added several models in sct_deepseg ?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -661,19 +762,22 @@ def mosaic(img: Image, centers: np.ndarray, radius: tuple[int, int] = (15, 15),
...
     """
     Add orientation labels (A, P, L, R) to a figure, yellow with a black outline.
     """
     # Ensure that letter locations are determined as a function of the bounding box. For a 15,15 radius (30x30):
     #    A                    [12,  6]
     # L     R   -->  [0, 17]            [24, 17]
     #    P                    [12, 28]
+    letters = ['A', 'P', 'L', 'R']
+    if orientation == 'RSP':
+        letters = ['S', 'I', 'P', 'A']
```
joshuacwnewton, line 775: I think instead of mapping `orientation` to `letters`, it might make more sense to instead just put `letters` as the parameter.

That way, it's clear at the caller which letters are being used.

joshuacwnewton (dismissed):
Since this change is opt-in (i.e. we still use `Axial` by default), there should be 0 difference for regular usage, so I agree that this should be safe to merge for 6.5.

Also, I love your use of the new `qc2.py` API. It makes the changes very easy to review. (Thanks again to @mguaypaq for the backend refactor.)

I only have a very small change request for now, so I'm pre-approving. (There's maybe a conversation to be had about de-duplicating `qc2.py` functionality, but I think we can save that for v7.0.)

spinalcordtoolbox/reports/qc2.py
plbenveniste: Okay ! Makes sense ! On it now !

spinalcordtoolbox/reports/qc2.py
```diff
@@ -267,19 +268,22 @@ def sct_deepseg(
...
     ) as imgs_to_generate:
         if "seg_spinal_rootlets_t2w" in argv:
             sct_deepseg_spinal_rootlets_t2w(
                 imgs_to_generate, fname_input, fname_seg, fname_seg2, species)
-        else:
-            sct_deepseg_default(
+        if plane == 'Axial':
+            sct_deepseg_axial(
+                imgs_to_generate, fname_input, fname_seg, fname_seg2, species)
+        elif plane == 'Sagittal':
```
joshuacwnewton, line 281: Minor nitpick, but by using `if/elif` with no `else`, there's theoretically a chance that this could be called with a `plane` value other than Axial/Sagittal, and then the behavior would be unclear.

So, it tends to be safer to use the following pattern (even if you're 99% sure that you'll only ever get Axial/Sagittal). Like, even though -qc-plane only has 2 options, maybe in the future the `sct_deepseg()` QC function gets used somewhere else, and someone makes a typo... who knows! :stuck_out_tongue_closed_eyes:

```suggestion
        else:
            assert plane == 'Sagittal', (f"`plane` must be either 'Axial' "
                                         f"or 'Sagittal', but got {plane}")
```

That way, all of the non-Axial cases will go into the "else", and we'll do an internal check to make sure that `plane` is supposed to be 'Sagittal'.

joshuacwnewton: Ah, I don't want to hold this PR up... It's really not a big deal.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton, line 1: I just gave the feature a try on some different data. It worked well on the cropped data from `sct_testing_data/t2/t2.nii.gz`, but when I tried on `sct_example_data` (non-cropped), I ran into the following error:

```
ValueError: Image size of 1500x96000 pixels is too large. It must be less than 2^16 in each direction.
```

The image I tried it on was [64, 320, 320]. What size images have you tried this feature on, @plbenveniste?

plbenveniste: I tried on some examples such as some canproco images (size [320,320,9]) which are sagittal. I guess it doesn't work on axial images however because then the number of slices is too high.
One way to solve that would be to resample the image in the L-R direction it the number of slices is too big (let's say 30 slices). 
On it now !

joshuacwnewton: Ah, of course!! That makes sense... for your use-case, `-qc-plane Sagittal` is meant for sagittally-acquired images. 

In that case, we could even leave it as-is for now and just add a disclaimer to `-qc-plane` until we come up with a more sophisticated approach? (For example, I know for the sagittal `sct_deepseg_lesion` QC, we crop the image around the segmentation (+/- a few extra buffer slices).

plbenveniste: I didn't see your message and added the resampling to 30 slices to avoid memory issues.
Should I revert my changes ? Or should we keep it like this for now ? 

Here is an example of what the output looks like on large field of views for 2D axial data [320 x 250 x 30] (RPI) for spinal cord segmentation with contrasts agnostic model. 

image

As we can see, because the FOV is so large, the spinal cord segmentation is only visible on one slice. 

I am referencing this problem in the PR related issue to keep it in mind.

joshuacwnewton: I think it is better to have resampling than to have a crash! Although maybe it is not the most useful, it is at least not an error.

The important thing is that this feature is available for your own work with sagittal images. :)

joshuacwnewton (approved):
Approving for now, but we can discuss further methods in a follow-up PR. :)

PR #4720, opened 2024-11-27
Update warnings filter to properly filter `Private repos` warning from `requirements.parse`

### LINKED ISSUE #4712: UserWarning: Private repos not supported. Upon checking SCT install
After installing SCT 6.5, the last 'checking' step shows me this:

````
Check if dipy is installed........................../Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_check_dependencies.py:164: UserWarning: Private repos not supported. Skipping.
  for req in requirements.parse(requirements_txt):
````

Not a big deal, but we should remove is not too complicated

joshuacwnewton:
## Description

We can no longer rely on `module` to filter this warning due to an upstream addition of `stacklevel=2`. So, we must instead filter based on message, rather than module.

Before:

image

After:

image

## Linked issues

Fixes #4712.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -159,7 +159,7 @@ def get_dependencies(requirements_txt=None):
     requirements_txt = open(requirements_txt, "r", encoding="utf-8")
 
     # workaround for https://github.com/davidfischer/requirements-parser/issues/39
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
Thanks for figuring out this fix, and communicating with upstream. I think the new filter is still appropriately specific to the warning we're trying to mask :+1:

PR #4723, opened 2024-11-28
Implement QC outlines for multi-valued segmentations

### LINKED ISSUE #4704: Rootlets QC outlining algorithm glitches out when re-used for TotalSpineSeg QC
For the rootlets QC, we use an algorithm to draw an outline around the rootlets to improve contrast and make the regions easier to see ([`plot_outlines()`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b2a6393c842ed0aae0a84083b3683d720a48b511/spinalcordtoolbox/reports/qc2.py#L740-L800)).

However, when I try to re-use this QC for TotalSpineSeg, the lines start glitching out:

image

I am assuming that this is because the connected components touch, breaking the assumption of the algorithm? I will maybe try to turn this off for TotalSpineSeg's QC.

mguaypaq:
I haven't had time to test it yet, but here's a reasonably efficient algorithm for drawing a black outline around multi-valued segmentations in QC reports. (The previous algorithm only handled binary segmentations.)

This should fix #4704.

I imagine it won't get merged into SCTv6.5, so it's really not urgent.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -865,9 +865,9 @@ def equalize_histogram(img: np.ndarray):
     return np.array(c * (max_ - min_) + min_, dtype=img.dtype)
 
 
-def plot_outlines(bool_img: np.ndarray, ax: mpl_axes.Axes, **kwargs):
+def plot_outlines(img: np.ndarray, ax: mpl_axes.Axes, **kwargs):
     """
-    Draw the outlines of a 2D binary Numpy array with Matplotlib.
+    Draw the outlines of every equal-value area of a 2D Numpy array with Matplotlib.
 
     kwargs are forwarded to `matplotlib.collections.PolyCollection` for styling.
     """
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/reports/qc2.py
```diff
@@ -280,7 +280,7 @@ def sct_deepseg(
         elif "totalspineseg" in argv:
             sct_deepseg_spinal_rootlets_t2w(
                 imgs_to_generate, fname_input, fname_seg, fname_seg2, species,
-                radius=(40, 40), outline=False)
```
joshuacwnewton, line 283: Testing `outline=True` with `totalspineseg`, the QC now looks like this:

image

Splendid!!

mguaypaq: Woot! Thanks for testing this! I'll mark the PR as non-draft, then.

joshuacwnewton (commented):
I would love to take the time to fully understand the algorithm here, but I'm not sure I have the time before our SCT meeting?

I can approve this PR purely based on the QC report it generates for `totalspineseg` and `seg_rootlets_t2`, though I'm not sure that's sufficient.

I don't think it's so urgent that we need it? The totalspineseg qc is almost as legible without the outline, so maybe it's fine to leave this for now and merge it for v7.0.

mguaypaq:
There's really no rush, and we have plenty of other stuff to work on right now :+1:

joshuacwnewton (approved):
Your comments are great, but nonetheless I still felt I had to visualize it (alongside variable inspection) to truly understand it... in Google Sheets? (Completely overkill!)

I've attached my visualization here just so that we can always go back and look at it:

image

My example array only contained an open polygon, but closed polygons are a more straightforward case of the traversal behavior anyway, so I'm satisfied.

Issue #333, opened 2015-04-17
sct_change_image_type doesn't work with type float16


SaraDupont:
![screen shot 2015-04-17 at 4 40 50 pm](https://cloud.githubusercontent.com/assets/5857909/7210530/8d16995e-e520-11e4-9c94-30cc4bffed40.png)

tanguymagnan:
It seems that nibabel nifti1 header does not support float16 type as numpy does... Not a code problem

PR #4730, opened 2024-11-29
Make sure subtractions are signed and divisions are floating point

### LINKED ISSUE #4725: Suspicious extreme value when performing subtraction operation
### Description

While testing the SCT course, during the subtraction of the T2\*w cord seg to the T2\*w GM seg, the output shows an abnormal intensity range:

Screenshot 2024-11-29 at 9 43 20 AM

Whereas the cord seg and GM seg don't show such extreme values:

Screenshot 2024-11-29 at 9 43 15 AM

Screenshot 2024-11-29 at 9 43 12 AM

This might have caused the "apparent" black output mentioned by @tomDag25 in [his comment](https://docs.google.com/presentation/d/1uUOpgshwnyC2p8r2GalXlUczQLpX6PfJbtzNELxbqdI/edit?disco=AAABZfjAT1c), whereas in fact, the viewer simply 'squeezed' the dynamic of the signal intensity to accommodate such extreme value, resulting in pixels having value=1 appearing as _almost_ black (hence invisible in contrast with the background).

I suspect this is a precision issue during type conversion, that was introduced in 'fairly' recent versions, given that this was not an issue during the last course.

Possibly related to #4724 (investigating further)

mguaypaq:
Numpy arithmetic can be tricky, because subtractions using unsigned integers can result in wraparound to huge positive values. Also, divisions of integer arrays default to `float64`, which may be unnecessarily big.

Rather than relying on these defaults, we should be careful about the expected dtype of arithmetic operations in `sct_maths`. (For uniformity, I also updated the code for addition and multiplication, so that all arithmetic uses the same pattern.)

Fixes #4725.

(I can add a test case on Monday if @joshuacwnewton doesn't beat me to it :wink: but I wanted to get this PR open asap.)

### DIFF
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -459,16 +459,18 @@ def main(argv: Sequence[str]):
                         list_data = get_data_arrays(data.shape, arg_value)
                     except ValueError as e:
                         printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                    # the addition can't use += because the dtypes could be different
-                    data = data + np.sum(list_data, axis=0)
+                    # for addition, this dtype is usually ok
+                    safe_dtype = np.result_type(data, *list_data)
+                    data = np.add(data, np.sum(list_data, axis=0, dtype=safe_dtype), dtype=safe_dtype)
 
             elif arg_name == "sub":
                 try:
                     list_data = get_data_arrays(data.shape, arg_value)
                 except ValueError as e:
                     printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                # the subtraction can't use -= because the dtypes could be different
-                data = data - np.sum(list_data, axis=0)
+                # for subtraction, make sure the dtype is at least signed by including int8
+                safe_dtype = np.result_type(data, *list_data, np.int8)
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -479,16 +481,18 @@ def main(argv: Sequence[str]):
...
 
             elif arg_name == "div":
                 try:
                     list_data = get_data_arrays(data.shape, arg_value)
                 except ValueError as e:
                     printv(f"ERROR: -{arg_name}: {e}", 1, 'error')
-                # the division can't use /= because the dtypes could be different
-                data = data / np.prod(list_data, axis=0)
+                # for division, make sure the dtype is at least floating point by including float16
+                safe_dtype = np.result_type(data, *list_data, np.float16)
```
joshuacwnewton, line 494: Tested this (pushing a test soon), and found it runs into: 

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/333

joshuacwnewton (approved):
`np.result_type` is neat! Thanks for this :hearts:

mguaypaq:
Thanks for the tests and the fixes! Looks great! I'll merge this now.

Issue #4734, opened 2024-12-02
Orientation labels (`L`, `R`, `A`, `P`) are misplaced for some axial views (but not all)


joshuacwnewton:
When running through `batch_single_subject.sh` on my local computer, I noticed the following:

![image](https://github.com/user-attachments/assets/0aaa43f6-2e03-444e-b661-986c269362f0)

Yet for other very similar QCs, the labels are fine:

![image](https://github.com/user-attachments/assets/04cb60b4-b036-46f3-b22b-2e5cd8aeb4ef)

The faulty commands that produce the weird labels are:

```
sct_deepseg -task seg_sc_contrast_agnostic -i t2.nii.gz -qc /home/joshua/qc_singleSubj

sct_deepseg -task seg_sc_contrast_agnostic -i t2_compressed.nii.gz -qc /home/joshua/qc_singleSubj

sct_deepseg -task seg_sc_contrast_agnostic -i mt1.nii.gz -qc /home/joshua/qc_singleSubj

sct_register_multimodal -i /home/joshua/repos/spinalcordtoolbox/data/PAM50/template/PAM50_t2.nii.gz -iseg /home/joshua/repos/spinalcordtoolbox/data/PAM50/template/PAM50_cord.nii.gz -d mt1.nii.gz -dseg mt1_seg.nii.gz -m mask_mt1.nii.gz -initwarp ../t2/warp_template2anat.nii.gz -param step=1,type=seg,algo=centermass:step=2,type=seg,algo=bsplinesyn,slicewise=1,iter=3 -owarp warp_template2mt.nii.gz -qc /home/joshua/qc_singleSubj

sct_register_multimodal -i mt0.nii.gz -d mt1.nii.gz -dseg mt1_seg.nii.gz -m mask_mt1.nii.gz -param step=1,type=im,algo=slicereg,metric=CC -x spline -qc /home/joshua/qc_singleSubj

sct_deepseg -task seg_sc_contrast_agnostic -i t2s.nii.gz -qc /home/joshua/qc_singleSubj

[...]

PR #4722, opened 2024-11-27
Fix rootlets QC for anisotropic images

joshuacwnewton:
## Description

Typically, for our QC functions, anisotropic images (i.e. images with resolutions that are not equal for all three x/y/z axes) don't really cause any issues when plotting axial mosaics. This is because typically we will [resample the images to an isotropic resolution](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/89b1ffcbc649cb0f90573cb37aa378f1a3cd3b08/spinalcordtoolbox/reports/qc2.py#L316-L323), which allows us to proceed under the assumption that the slicewise X/Y resolutions are equal.

However, for the rootlets QC, we skip this resampling step and instead modify the crop radius to fit the native resolution of the data. This helps to preserve the thin, often 1-voxel wide rootlets labels when displaying them in the QC report. (See 4c0c1f4180883d800036a11054f237609095d12b for more context.)

But because we skip the resampling, the assumption I described earlier (that the slicewise X/Y resolutions are equal) is no longer valid. Yet, I mistakenly developed the rootlets QC by only testing isotropic images. So, there are a number of steps that use faulty logic, causing anisotropic images to produce a poor QC report.

This PR adds a number of fixes that cause anisotropic images to display correctly:

- d7ca325e1848ff089b8941f15637d2f554a09d95: Choose a crop radius that respects both X and Y resolutions.
- 3bf1eaba7839f1f89040537c7e67a8229779504f: Apply `aspect` to scale the figure so that the voxel size matches the respective X/Y resolutions.
- Fix 2 other bugs that were invisible when tested using only isotropic images:
    - 13006f441ea0600309834750e4d5812272b2de49: Fix bug with how we specific the `pad_width` argument to `np.pad` in the `mosaic` function.
    - 36cd25a2160122397a9bffdcf980d40ce6e746dc: Fix bug with how we specify the x/y coords when writing text.

Before:

image

After:

image

## Linked issues

Fixes #4716.

PR #4736, opened 2024-12-02
Fix orientation labels by properly swapping x/y radii rather than x/y coords

joshuacwnewton:
## Description

In PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4722, I fixed some issues that affected anisotropic QC reports that we had been blind to for isotropic images (i.e. the case where `radius[0] == radius[1]`). 

One of those changes swapped the x/y coordinates for orientation labels, because numpy arrays use "zyx" axis ordering, thus radius[0] corresponds to y and radius[1] corresponds to x.

However, in doing so, I not only swapped radius[0]/radius[1], but I also swapped the _axis-specific offsets_, causing the orientation labels to be misplaced. To fix this, I need to just swap radii.

Before (isotropic):

image

After (isotropic):

image

After (anisotropic):

image

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4734.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -795,7 +795,7 @@ def add_orientation_labels(ax: mpl_axes.Axes, radius: tuple[int, int] = (15, 15)
     #    A                    [12,  6]
     # L     R   -->  [0, 17]            [24, 17]
     #    P                    [12, 28]
-    for letter, y, x, in [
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
Thanks for including those before/after screenshots, they're enough to convince me that you've nailed it.

PR #4737, opened 2024-12-02
Fix softseg QC by using only a single transparency value (`alpha=1.0`)

### LINKED ISSUE #4735: QC of sct_deepseg displays "soft" segmentations
## Description

When running the Qc report in `sct_deepseg` function, the segmentation ouput in the QC report is soft, but when rerunning the command with `sct_qc -p sct_deepseg_sc`, the ouput in the QC is binarized. I know it can be useful to have it soft if the flag `-thr 0` is used, but the visual is a bit weird.

I would suggest to binarize also the `sct_deepseg` QC if the flag `-thr 0` is not there? 

### `sct_deepseg` QC
Capture d’écran, le 2024-12-02 à 11 34 47

### `sct_deepseg_sc` QC
Capture d’écran, le 2024-12-02 à 11 35 09

joshuacwnewton:
## Description

Request by @sandrinebedard.

Before: 

image

After: 

image

## Linked issues

Fixes #4735.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -375,7 +375,7 @@ def sct_deepseg_axial(
                   cmap=colormaps[i],
                   norm=mpl_colors.Normalize(vmin=0.5, vmax=1),
                   # img==1 -> opaque, but soft regions -> more transparent as value decreases
-                  alpha=(img / img.max()),  # scale to [0, 1]
+                  alpha=1.0,
                   interpolation='none',
                   aspect=1.0)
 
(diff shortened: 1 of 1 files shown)


sandrinebedard (approved):
Thanks for this @joshuacwnewton ! I am curious thta the results looks very different:

This is contrast-agnostic on the T2star of the course material:

Before:
image

After: 

image

So I think with the alpha, we were missing some part of the segmentation, curious!

mguaypaq:
It looks a bit weird that the ALRP labels are "under" the segmentation. But I guess that's because they're drawn on the background image, and not on the foreground image. Not a big deal.

sandrinebedard:
I also noticed that!

sandrinebedard:
and on other images: 
image
 maybe we should fix too

joshuacwnewton:
> So I think with the alpha, we were missing some part of the segmentation, curious!

Maybe there were values very close to 0, making them pretty much transparent?

> and on other images, they are not at the right spot:

Just fixed now in:

- #4736

PR #4739, opened 2024-12-02
Set header datatype when applying dtype fix for NearestNeighbours interpolation to avoid warnings in log

### LINKED ISSUE #4738: Unnecessary verbose during template warping
When running warping, there is a lot of warning about type conversion, which might confuse/scare some users:

image

joshuacwnewton:
## Description

Since we're intentionally changing the datatype here, we can also safely change the header datatype (to avoid warnings about the header dtype not being int).

This makes `sct_warp_template` a lot less noisy, with almost all of the warnings coming from uint8 images in the template.

## Linked issues

Fixes #4738.


mguaypaq (approved):
Yes, this is a much better fix than ignoring/silencing the warnings!

Issue #3250, opened 2021-02-25
Validate additional functionality in 'batch_processing.sh' (using QC, metric values)


joshuacwnewton:
> **Note:** I'm creating this issue to summarize the ideas from #1929, #2034, and #2487 so that they're all in one place.

### Context

`batch_processing.sh` is a script we use to demonstrate typical workflows for SCT tools. At the end of the script, we let the user check the results in 2 ways:

* **17** QC reports generated by the various functions. 
* **6** calculated metrics, taken from `.csv` result files. (#2888 is how we formally test these values.)

### Proposal

We could be testing more. Some ideas:

- [ ] QC on T1 segmentation
- [ ] CSA on T1
- [ ] `sct_analyze_lesion`
- [ ] `sct_process_segmentation -p shape` (already run but not appended to final results)
- [ ] tSNR on fMRI
    * `sct_fmri_compute_tsnr -i fmri_crop_moco.nii.gz`
    * `sct_extract_metric -i fmri_crop_moco_tsnr.nii.gz -f fmri_crop_moco_mean_seg_manual.nii.gz -o tsnr.txt `

Feel free to edit the list above with any additional ideas.

PR #4745, opened 2024-12-03
Revert breaking xform change, and instead temporarily skip `check_affines_match`-related crashes

### LINKED ISSUE #4744: Reorienting to RPI (then back) will subtly change the affine matrices by ~1e-4 due to NIFTI orthogonal requirement x2
### Description

@sandrinebedard discovered that spinal cord and lesion segmentation from the SCIsegV2 model (`sct_deepseg seg_sc_lesion_t2w_sci`) have different qform and sform matrixes than the original image. I can confirm the behaviour by running:

https://github.com/spinalcordtoolbox/sct_tutorial_data/blob/4034ad14637290c915497d8f4ff3bf96a15841a6/single_subject/batch_single_subject.sh#L392:

```console
sct_deepseg -i t2.nii.gz -task seg_sc_lesion_t2w_sci -qc ~/qc_singleSubj
```

Headers diff: https://www.diffchecker.com/2J9FXJou/

Headers:

<details><summary>t2.nii.gz</summary>

```console
sct_image -i t2.nii.gz -header

--
Spinal Cord Toolbox (git-master-64f55fdf022659db5d5c779f01c605d91ca7d052)

sct_image -i t2.nii.gz -header
--

sizeof_hdr	348
data_type	INT16
dim		[3, 64, 320, 320, 1, 0, 0, 0]
vox_units	mm
time_units	s
datatype	4
nbyper		2
bitpix		16
pixdim		[1.0, 0.8, 0.8, 0.8, 1.5, 0.0, 0.0, 0.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	2
freq_dim	1
slice_dim	3
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Unknown
intent_code	0
intent_name
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	0.799726 0.000000 -0.020942 -19.452049
qto_xyz:2	0.000000 0.800000 0.000000 -115.151794
qto_xyz:3	0.020942 -0.000000 0.799726 -122.996811
qform_xorient	Left-to-Right
sform_name	Scanner Anat
sform_code	1
sto_xyz:1	0.799726 0.000000 -0.020942 -19.452049
sto_xyz:2	0.000000 0.800000 0.000000 -115.151794
sto_xyz:3	0.020942 -0.000000 0.799726 -122.996811
sform_xorient	Left-to-Right
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
file_type	NIFTI-1+
file_code	1
descrip		TE=1.2e+02;Time=110001.045;phase=1
aux_file	imgComments
```

</details>

<details><summary>t2_lesion_seg.nii.gz </summary>

```console
sct_image -i t2_lesion_seg.nii.gz -header

--
Spinal Cord Toolbox (git-master-64f55fdf022659db5d5c779f01c605d91ca7d052)

sct_image -i t2_lesion_seg.nii.gz -header
--

sizeof_hdr	348
data_type	UINT8
dim		[3, 64, 320, 320, 1, 1, 1, 1]
vox_units	mm
time_units	s
datatype	2
nbyper		1
bitpix		8
pixdim		[1.0, 0.8, 0.8, 0.8, 1.0, 1.0, 1.0, 1.0]
vox_offset	352
cal_max		0.000000
cal_min		0.000000
scl_slope	1.000000
scl_inter	0.000000
phase_dim	2
freq_dim	1
slice_dim	3
slice_name	Unknown
slice_code	0
slice_start	0
slice_end	0
slice_duration	0.000000
toffset		0.000000
intent		Unknown
intent_code	0
intent_name
intent_p1	0.000000
intent_p2	0.000000
intent_p3	0.000000
qform_name	Scanner Anat
qform_code	1
qto_xyz:1	0.799726 0.000000 -0.020939 -19.452053
qto_xyz:2	0.000000 0.800000 0.000000 -115.151794
qto_xyz:3	0.020939 -0.000000 0.799726 -122.996651
qform_xorient	Left-to-Right
sform_name	Scanner Anat
sform_code	1
sto_xyz:1	0.799726 0.000000 -0.020939 -19.452053
sto_xyz:2	0.000000 0.800000 0.000000 -115.151794
sto_xyz:3	0.020939 -0.000000 0.799726 -122.996651
sform_xorient	Left-to-Right
sform_yorient	Posterior-to-Anterior
sform_zorient	Inferior-to-Superior
file_type	NIFTI-1+
file_code	1
descrip		TE=1.2e+02;Time=110001.045;phase=1
aux_file	imgComments
```

</details>

(I tried `sct_deepseg seg_sc_lesion_t2w_sci` also on another subject (`sub-zh01_ses-01_acq-sag_T2w.nii.gz`), and the qform and sform matrices are also different!)

joshuacwnewton:
## Description

This is an emergency hotfix to revert the breaking change introduced by #4697. While that PR did eliminate the crash found in the original issue's [forum post](https://forum.spinalcordmri.org/t/dog-template-registration-qform-sform-error/1324/1), it introduced an even larger issue in #4744.

I think in the short term, we should prioritize numerical consistency between versions (v6.4 -> v6.5) -- reverting #4697 will keep the `batch_processing.sh` results the same. Then, to still address the crash from the forum post, this PR instead allows processing to continue in situations where it previously wouldn't due to sform/qform mismatches. This means that the original issue (#4689) will get reopened, but I think that this is the [lesser of two evils](https://en.wikipedia.org/wiki/Lesser_of_two_evils_principle) in the short-term.

But, still, these changes deserve discussion, hence tagging a few different people for reviews.

## Linked issues

Reverts #4697.
Reopens #4689.
Closes #4744.


testing/batch_processing/cached_results/dmri/fa_in_cst.csv
joshuacwnewton, line 1: Side note: I think I have become a bit complacent about `batch_processing.sh` changes.

Right now, our integration test for processing pipelines (i.e. `batch_processing.sh`) is very very sensitive to any sort of change. By comparison, our unit tests will typically only check e.g. DICE score > 0.95, which is useful, but also kind of insensitive to small changes. (We use "> 0.95" because if we checked the exact voxel output, we'd get spurious failures when making 'safe' changes, which would block e.g. updates to `sct_deepseg` models.)

Still, there is a lot of gray area between "minute 1e-6mm change in CSA" and "DICE score >0.95". (e.g. Did the CSA change because of a single voxel difference, or multiple voxels? Which part of the pipeline changed? Was it a single script, or multiple scripts? Did one small change at the start of the pipeline cascade across steps, or was there a breaking change across multiple steps?) 

Because of this, we semi-often run into in cases where the `batch_processing.sh` results change, but *no other tests fail*. So, it can be hard to truly get a sense of the scope of the changes -- we lack the ability to fully contextualize `batch_processing.sh` failures. Meaning that when the `batch_processing.sh` results change, our dev team typically reacts by saying "yep, this was expected" and updating the cached results, without fully probing the cause. 

So, what I would like to do (following this release) is prioritize catching minute numerical differences to provide context when `batch_processing.sh` fails. That way, we can say "oh, batch_processing.sh results changed _and_ this is why, and we're okay with that".

joshuacwnewton: Somewhat related:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3250
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3194
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3597

(There are ways that `batch_processing.sh` results differ between SCT versions and platforms that we don't fully understand due to a lack of context.)

joshuacwnewton: I'm spitballing here, but I could imagine a post-pytest step where we take the entirety of the `pytest` tmpdir output, and summarize all of the images on disk in a tidy way (e.g. we extract the dtype, xforms, a hash of the array contents, etc. and stick the contents in a csv file alongside the filename and pytest tmpdir name (which encodes the test name, too!)). Then, we would be able diff the csv file between versions and platforms, and see _exactly_ where and when an output image changes in any way. This would be overly sensitive, but it would provide so much context that we're currently missing.

spinalcordtoolbox/image.py
```diff
@@ -331,16 +331,23 @@ def __init__(self, param=None, hdr=None, orientation=None, absolutepath=None, di
...
+                         f"results. Please use 'sct_image -i {self._path} -header' to check that both affine "
+                         f"matrices are valid. Then, consider running either 'sct_image -set-sform-to-qform' or "
+                         f"'sct_image -set-qform-to-sform' to fix any discrepancies you may find.")
+            logger.error("If internal SCT code has produced an intermediate/temporary file with this issue, please report this on GitHub at "
+                         "https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues or on the SCT forum https://forum.spinalcordmri.org/.")
+
+            # Temporarily skip raising an error, because we now know that "orthogonal qform matrices" from reorientation can cause sform/qform
+            # discrepancies that trigger this error.
+            #
+            # raise ValueError("Image sform does not match qform")
```
mguaypaq, line 344: At first glance, I was going to say that disabling the `ValueError` at this very general level (the `Image` constructor) may be too drastic. But, it's at least replaced by the (now unconditional) `logger` messages above, so that's acceptable to me.

mguaypaq (approved):
I agree with your analysis that our common approach of reorient-then-process-then-reorient-back means that #4697 applies more often than we initially thought. And we don't want to ignore the orientation warning in FSLeyes.

If we want to fix both issues, we may need some big architectural changes in SCT, which we can't do right in the middle of a release. So yes, reverting to the previous behaviour seems like the only sensible option for now.

testing/batch_processing/cached_results/dmri/fa_in_cst.csv
joshuacwnewton: I'll move this brainstorming to a new issue, then mark this comment as resolved, since this is out of scope for this PR.

valosekj:
Thank you for the prompt fix! 🙏🏻

Issue #4376, opened 2024-02-15
Drop ANTs build


kousu:
**Feature justification: Is your feature request related to a problem? Please describe.**

Updating ANTs to new versions is difficult due to the tricky [build process](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Packaging%3A-How-to-compile-C---binaries-(ANTs,-isct)) we use. We wrote a [separate build script](https://github.com/spinalcordtoolbox/build_ANTs/blob/c7824ace54ac3d1f4a5283136d197f0184681c03/.github/workflows/build-ants.yml) that handles updating them for us, and then we have the installer download the builds from https://github.com/spinalcordtoolbox/build_ANTs/releases/.

**Describe the solution you'd like**
Since we wrote that, ANTs has themselves migrated to using GitHub CI, and they [produce their own binaries](https://github.com/ANTsX/ANTs/blob/4a0fb48012ab9ce0f8babb3980c5c8de356aae3d/.github/workflows/release-binaries.yml#L97) and publish them to https://github.com/ANTsX/ANTs/releases.

Please investigate if it is possible to migrate to using these upstream builds over using our own in-house builds.

joshuacwnewton:
Gosh, thank you!

It looks like we set up `build_ANTs` in May 2022, and ANTs started publishing built binaries in... August 2022. :facepalm: Quite the timing, woof! 

I think that this should allow us to revisit some older issues related to upstream ANTs features, though:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/742
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/1178

PR #4759, opened 2024-12-23
Improve how we track changes to SCT's outputs by exporting a summary of all output files

joshuacwnewton:
## Description

During the lead up to the SCT course, we discovered many subtle behavior changes that slipped through our testing, and required manual examination to catch. Even worse, I almost accidentally introduced a big change right before the SCT course, and it wasn't caught by the tests (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4744#issuecomment-2514633760). Because of this, I felt especially motivated to try to make some changes to improve SCT's testing coverage over the holiday break.

IMO, we've got a big gap in our testing. Right now, our tests cover the following areas:

- **Low sensitivity, high specificity**: Our unit tests for CLI/API are good at measuring the outputs of specific commands, but often use measures such as `DICE > 0.95` to avoid spurious failures from small, acceptable changes over time.
- **High sensitivity, low specificity**: Our integration test (`batch_processing.sh`) is more  sensitive to changes in the processing pipeline, with small changes to one input often cascading throughout labeling/registration/CSA steps. However, when it does fail, it can be hard to pinpoint the exact cause of the change, and so we often blindly update the cached results because we're limited in how well we can understand the change.
- We don't really have any **high sensitivity, high specificity** tests because A) the specificity would make the tests themselves time-consuming to write and B) the sensitivity would likely cause many spurious failures, which would again be time-consuming to debug and update.

To fill this void, I tried writing a test-agnostic function that will summarize the files in a folder based not only on (size, hash) but also image properties. That way, we can export the summary as a whole for `tmp/pytest`, `sct_testing_data`, etc. and compare it between versions and between OSs.

I'm hoping that diffing these files will provide some insight into _why_ the `batch_processing.sh` output changes, and minimize the time spent manually comparing images locally, guessing where the change might have first occurred. And, I'm hoping we can come up with a workflow to diff the files in a way that won't be disruptive to our PRs.


.github/workflows/test-batch-processing.yml
```diff
@@ -84,3 +84,10 @@ jobs:
           else
                sct_testing testing/batch_processing/test_batch_processing.py
           fi
+
+      - name: Upload folder summaries
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: Data Folder Summary (${{ matrix.os }})
+          path: "testing/sct_example_data.csv"
```
joshuacwnewton, line 93: Comparing the outputs inside `sct_example_data` based for [**macOS vs. Ubuntu**](https://www.diffchecker.com/DGZB2NNP/), we can see that discrepancies only begin to occur when we get to the registration steps.

This makes me think that the ANTs binaries are the source of issue:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3194

And this makes me especially interested in trying out the following fix:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4376

It's possible that our custom build environments (e.g. ITK versions) differ between platforms, thus causing the discrepancy.

I'm going to try zipping up the official binaries, then comparing the results of the outputs using this PR.

joshuacwnewton: On the whole, the cross-platform registration discrepancies didn't really improve by updating ANTs to the officially built binaries. 

There is one thing that did improve, though: un-registered DTI metrics.

- `master`: https://www.diffchecker.com/DGZB2NNP/

image

- v2.5.4: https://www.diffchecker.com/sTT8bMDP/

image

That's only a small change, though, so upgrading the ANTs binaries wasn't exactly fruitful for solving #3194.

Still, I hope that the "folder summary" idea is still useful for tracking regressions from PRs?

contrib/fslhd.py
```diff
@@ -257,8 +260,39 @@ def generate_nifti_fields(header):
...
+        'arr_mode': mode(arr, axis=None, keepdims=False).mode,
+        'arr_mean': arr.mean(),
+        'arr_std': arr.std(),
+        'arr_sum': arr.sum(),
+        'arr_p1': np.percentile(arr, 1),
+        'arr_p10': np.percentile(arr, 10),
+        'arr_p90': np.percentile(arr, 90),
+        'arr_p99': np.percentile(arr, 99),
+        'arr_iqr': np.percentile(arr, 75) - np.percentile(arr, 25),
+        'arr_mad': np.median(np.abs(arr - np.median(arr))),
```
joshuacwnewton, line 297: We probably don't need all of these to tell if the array has changed, hah...

.github/workflows/test-batch-processing.yml
```diff
@@ -84,3 +85,44 @@ jobs:
...
+          path: artifacts
+          merge-multiple: true
+
+      - name: Diff summaries (Ubuntu vs macOS)
+        continue-on-error: true
+        run: diff -u artifacts/sct_example_data_ubuntu-20.04.json artifacts/sct_example_data_macos-13.json | ydiff
+
+      - name: Diff summaries (Ubuntu vs. Windows)
+        continue-on-error: true
+        run: diff -u artifacts/sct_example_data_ubuntu-20.04.json artifacts/sct_example_data_windows-2022.json | ydiff
```
joshuacwnewton, line 128: These diffs aren't exactly necessary, and are more to act as a test case for whichever diffing tool we go with.

Ideally I would want to compare against the latest run on `master` from the same OS, potentially with something like https://github.com/marketplace/actions/download-workflow-artifact, which should be pretty straightforward to add once this runs on `master`.

.github/workflows/test-batch-processing.yml
```diff
@@ -22,6 +22,7 @@ jobs:
   batch_processing:
     name: Test batch_processing.sh
     strategy:
+      fail-fast: false
```
joshuacwnewton, line 25: I'm surprised we hadn't added this before -- now at least runs will continue across the matrix even if one fails.

testing/conftest.py
```diff
@@ -33,6 +39,51 @@ def pytest_sessionstart():
...
+    """Perform actions that must be done after the test session."""
+    # get the newest temporary path created by pytest
+    tmp_paths = glob(os.path.join(tempfile.gettempdir(), "pytest-of-*", "pytest-current"))
+    ctimes = [os.path.getctime(p) for p in tmp_paths]
+    tmp_path = tmp_paths[ctimes.index(max(ctimes))]
+
+    # generate directory summaries for both sct_testing_data and the temporary directory
+    for (folder, fname_out) in [(tmp_path, "pytest-tmp.json"),
+                                (sct_test_path(), "sct_testing_data.json"),
+                                (sct_test_path().replace("testing", "example"), "sct_example_data.json")]:
```
joshuacwnewton, line 52: Ideally I would like to add this sort of non-blocking json comparison to `tests.yml` too, so that we can monitor our pytest outputs as well. (I've kept it to `batch_processing.yml` for now to make it more straightforward to review from the outset.)

I like the thought that by analyzing the folder as a whole, we get this analysis for free for any existing and new tests we write. It feels like a very low-maintenance way to see how our outputs change from PR to PR. Even if it doesn't necessarily check if the outputs are sensible from a semantic POV, we can rest assured that we aren't silently changing the results when we don't mean to (assuming we check the diffs at least once before merging each PR).

This should make us feel a lot more confident when prepping for e.g. SCT Courses year after year.

mguaypaq, line 52: I think `sct_test_path()` is subtly the wrong function to use to locate `sct_example_data`, because [it takes into account the `SCT_TESTING_DATA` environment variable](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d218b9e6d62e46e8d78bd768130a4188554455fd/spinalcordtoolbox/utils/sys.py#L390-L407) (although possibly this is a vestigial feature that we should polish away?). Whereas I think `sct_example_data` is always installed at [the default location](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d218b9e6d62e46e8d78bd768130a4188554455fd/spinalcordtoolbox/utils/sys.py#L529):
```suggestion
                                (os.path.join(__data_dir__, "sct_example_data"), "sct_example_data.json")]:
```
(This also needs an extra import of `__data_dir__` from `utils.sys` at the top.)

contrib/fslhd.py
```diff
@@ -257,8 +260,39 @@ def generate_nifti_fields(header):
...
+    Generate useful summary statistics for any arbitrary numpy array.
+    """
+    arr = arr.astype(np.float64)
+    return {
+        'arr_nbytes': arr.nbytes,
+        'arr_size': arr.size,
+        'arr_num_zeros': arr[arr == 0.0].size,
+        'arr_num_nonzeros': arr[arr != 0.0].size,
+        'arr_num_positives': arr[arr > 0.0].size,
+        'arr_num_negatives': arr[arr < 0.0].size,
```
mguaypaq, line 281: I think you can avoid some redundant array allocations by using `np.count_nonzero` instead:
```suggestion
        'arr_num_zeros': np.count_nonzero(arr == 0.0),
        'arr_num_nonzeros': np.count_nonzero(arr != 0.0),
        'arr_num_positives': np.count_nonzero(arr > 0.0),
        'arr_num_negatives': np.count_nonzero(arr < 0.0),
```

contrib/fslhd.py
```diff
@@ -257,8 +260,39 @@ def generate_nifti_fields(header):
...
+        'arr_num_nans': arr[arr == np.nan].size,
+        'arr_num_infs': arr[arr == np.inf].size,
```
mguaypaq, line 283: One weird corner of IEEE floating point numbers is that `np.nan != np.nan`, so `arr_num_nans` will always be zero. There's a specific function to test whether a floating point value is `nan` (and also one for (possibly negative) infinity):
```suggestion
        'arr_num_nans': np.count_nonzero(np.isnan(arr)),
        'arr_num_infs': np.count_nonzero(np.isinf(arr)),
```

contrib/fslhd.py
mguaypaq: TIL about [median absolute deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation), I'd never heard of it before.

mguaypaq (approved):
Thanks for developing this bit of infrastructure! I'm sure it will be useful for tracking down changes. We can always tweak it after merging, once we see it working in practice.

contrib/fslhd.py
joshuacwnewton: It's really funny you came to this conclusion! I swear I had authored a commit that did exactly this. (I had stumbled across [this SO answer](https://stackoverflow.com/a/42916378) which is where I learned that `count_nonzero` count be used to efficiently count `== 0` elements!)

I don't know where that commit went, so thank you for bringing it up again. :stuck_out_tongue_closed_eyes:

PR #4760, opened 2024-12-28
Add new CLI script to predict compression probability

valosekj:
#### GitHub

#### PR contents

## Description

This PR adds a script to predict compression probability in a spinal cord MRI image using spinal cord shape metrics. The script requires spinal cord segmentation (used to compute the shape metrics) and disc labels.

Briefly, the script computes shape metrics for 3 axial slices at intervertebral disc levels C3/C4 to C6/7 (3 slices are used to compensate for potential disc label shift in the superior-inferior (S-I) axis; details: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760#discussion_r1903118334 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/90cbfa6654154987b0f542a6b544505fcaf8f076). Then, it predicts compression probability for each of these axial slices using a logistic regression model. The compression probability is printed to CLI and saved to a CSV file; see the next section for examples. More details in: https://pubmed.ncbi.nlm.nih.gov/35371944/. Particularly, equation 4.

## How to test this PR

```bash
# Download the script
cd $SCT_DIR
git fetch
git checkout jv/add_sct_detect_compression

# Activate SCT python
cd $SCT_DIR
source ./python/etc/profile.d/conda.sh
conda activate venv_sct

# Run the script
python $SCT_DIR/spinalcordtoolbox/scripts/sct_detect_compression.py -s T2w_seg.nii.gz -discfile T2w_discs.nii.gz
```

## Testing on a few spine-generic subjects with mild compression

<details><summary>sub-tokyoIngenia02</summary>

image

```bash
cd ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-tokyoIngenia02/anat
python $SCT_DIR/spinalcordtoolbox/scripts/sct_detect_compression.py -s sub-tokyoIngenia02_space-other_T2w_label-SC_seg.nii.gz -discfile sub-tokyoIngenia02_T2w_label-discs_dlabel.nii.gz
...
Results saved to: ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-tokyoIngenia02/anat/sub-tokyoIngenia02_space-other_T2w_label-SC_seg_compression_results.csv
Disc C3/C4, compression probability: no
Disc C4/C5, compression probability: yes
Disc C5/C6, compression probability: yes
Disc C6/C7, compression probability: yes
```

Output CSV:
image

</details>

<details><summary>sub-ucl05</summary>

image

```bash
cd ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-ucl05/anat
python $SCT_DIR/spinalcordtoolbox/scripts/sct_detect_compression.py -s sub-ucl05_space-other_T2w_label-SC_seg.nii.gz -discfile sub-ucl05_T2w_label-discs_dlabel.nii.gz
...
Results saved to: ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-ucl05/anat/sub-ucl05_space-other_T2w_label-SC_seg_compression_results.csv
Disc C3/C4, compression probability: no
Disc C4/C5, compression probability: no
Disc C5/C6, compression probability: yes
Disc C6/C7, compression probability: possible
```

Output CSV:
image

</details>

<details><summary>sub-vuiisAchieva01</summary>

image

```bash
cd  ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-vuiisAchieva01/anat
python $SCT_DIR/spinalcordtoolbox/scripts/sct_detect_compression.py -s sub-vuiisAchieva01_space-other_T2w_label-SC_seg.nii.gz -discfile sub-vuiisAchieva01_T2w_label-discs_dlabel.nii.gz
...
Results saved to: ~/data/data.neuro.polymtl.ca/data-multi-subject/derivatives/labels/sub-vuiisAchieva01/anat/sub-vuiisAchieva01_space-other_T2w_label-SC_seg_compression_results.csv
Disc C3/C4, compression probability: yes
Disc C4/C5, compression probability: no
Disc C5/C6, compression probability: no
Disc C6/C7, compression probability: no
```

Output CSV:
image

</details>

> [!NOTE]  
> Note that the predictive model was trained using axial T2star images from a private dataset. The example spine-generic images shown above are T2w 0.8mm iso. I used these images because they are [open-access](https://github.com/spine-generic/data-multi-subject/tree/master) to verify that the script works properly.
(I did not choose spine-generic T2star images as they cover only C2-C5, while the model detects compression also at discs C5/6 and C6/7).

## Linked issues

Follow-up of: https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/32153ad9e8136141dcff22d1ddc25291fa065b9a

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ setup(
                 'sct_deepseg_lesion',
                 'sct_deepseg_sc',
                 'sct_denoising_onlm',
+                'sct_detect_compression',
                 'sct_detect_pmj',
                 'sct_dice_coefficient',
                 'sct_dmri_compute_bvalue',
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -115,7 +115,7 @@ def predict_compression_probability(cr, csa, solidity, torsion, disc):
         + coefficients["cr"] * cr
         + coefficients["csa"] * csa
         + coefficients["solidity"] * solidity
-        + coefficients["torsion"] * torsion
+        + (coefficients["torsion"] * torsion if torsion is not None else 0)     # edge-case when torsion is None
         + (coefficients["level_c6c7"] if disc == 7 else 0)
     )
     return math.exp(model) / (1 + math.exp(model))
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -56,7 +56,7 @@ def get_parser():
             (the sum of sensitivity and specificity). More info in the reference below.
 
             Reference:
-            
+
                 - Horáková M, Horák T, Valošek J, Rohan T, Koriťáková E, Dostál M, Kočica J, Skutil T, Keřkovský M, Kadaňka Z Jr, Bednařík P, Svátková A, Hluštík P, Bednařík J. Semi-automated detection of cervical spinal cord compression with the Spinal Cord Toolbox. Quant Imaging Med Surg 2022; 12:2261–2279.
                   https://doi.org/10.21037/qims-21-782
         """),  # noqa: E501 (line too long)
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -232,23 +232,23 @@ def process_compression(metrics_agg_merged, disc_slices):
     """
     compression_df = pd.DataFrame()
     # Loop across discs
-    for disc, slices in disc_slices.items():
+    for disc, disc_slice in disc_slices.items():
         # Add one slice above and below to process multiple slices around the disc to compensate for potential
         # disc label shift in the superior-inferior (S-I) axis
-        slices = [slices + 1, slices, slices - 1]
-        for slice in slices:
+        slices = [disc_slice + 1, disc_slice, disc_slice - 1]
+        for slc in slices:
             if disc in SUPPORTED_DISCS.keys():
                 # Note: [slice,] is used to convert int to tuple
-                cr = metrics_agg_merged[slice,]['MEAN(compression_ratio)'] * 100    # to convert to %
-                csa = metrics_agg_merged[slice,]['MEAN(area)']
-                solidity = metrics_agg_merged[slice,]['MEAN(solidity)'] * 100     # to convert to %
-                torsion = metrics_agg_merged[slice,]['Torsion']
+                cr = metrics_agg_merged[slc,]['MEAN(compression_ratio)'] * 100    # to convert to %
+                csa = metrics_agg_merged[slc,]['MEAN(area)']
+                solidity = metrics_agg_merged[slc,]['MEAN(solidity)'] * 100     # to convert to %
+                torsion = metrics_agg_merged[slc,]['Torsion']
                 # Compute compression probability
                 probability = predict_compression_probability(cr, csa, solidity, torsion, disc)
                 compression_category = 'yes' if probability > CUT_OFF_HIGH else \
                     'possible' if probability > CUT_OFF_MODERATE else 'no'
                 compression_df = pd.concat([compression_df,
-                                            pd.DataFrame([{'Disc': SUPPORTED_DISCS[disc], 'Axial slice #': slice,
+                                            pd.DataFrame([{'Disc': SUPPORTED_DISCS[disc], 'Axial slice #': slc,
                                                            'Compression probability': probability,
                                                            'Compression probability category': compression_category,
                                                            'Compression ratio (%)': cr, 'CSA (mm2)': csa,
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -100,10 +100,8 @@ def get_parser():
         choices=[0, 1],
         default=1,
         help=textwrap.dedent("""
-            Angle correction for computing morphometric measures. When angle correction is used, the cord within the
-            slice is stretched/expanded by a factor corresponding to the cosine of the angle between the centerline
-            and the axial plane. If the cord is already quasi-orthogonal to the slab, you can set  -angle-corr to 0.
-        """)
+            Angle correction for computing morphometric measures. When angle correction is used, the cord within the slice is stretched/expanded by a factor corresponding to the cosine of the angle between the centerline and the axial plane. If the cord is already quasi-orthogonal to the slab, you can set  -angle-corr to 0.
+        """)  # noqa: E501 (line too long)
     )
     optional.add_argument(
         '-o',
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -315,9 +315,11 @@ def main(argv: Sequence[str]):
     for disc in compression_df['Disc'].unique():
         # If any axial slice around the disc is compressed, the disc is considered compressed
         if 'yes' in compression_df.loc[compression_df['Disc'] == disc, 'Compression probability category'].values:
-            printv(f"Disc {disc}: compressed", verbose)
+            printv(f"Disc {disc}, compression probability: yes", verbose)
         elif 'possible' in compression_df.loc[compression_df['Disc'] == disc, 'Compression probability category'].values:
-            printv(f"Disc {disc}: possibly compressed", verbose)
+            printv(f"Disc {disc}, compression probability: possible", verbose)
+        elif 'no' in compression_df.loc[compression_df['Disc'] == disc, 'Compression probability category'].values:
+            printv(f"Disc {disc}, compression probability: no", verbose)
 
     if verbose == 2:
         for index, row in compression_df.iterrows():
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -65,9 +65,9 @@ def get_parser():
         metavar=Metavar.file,
         required=True,
         help=textwrap.dedent("""
-            File with disc labels, which will be used to identify the axial slices for processing.
+            File with disc labels. Each label must be a single voxel. Only label values 4, 5, 6, and 7 (C3/C4 to C6/C7) are supported for now; all other labels will be ignored.
             Labels can be located either at the posterior edge of the intervertebral discs, or at the orthogonal projection of each disc onto the spinal cord.
-            Such label file can be manually created using: sct_label_utils -i IMAGE_REF -create-viewer 4:7 or
+            Such a label file can be manually created using: sct_label_utils -i IMAGE_REF -create-viewer 4:7 or
             obtained automatically using the sct_label_vertebrae function (the file with the \'labeled_discs.nii.gz\' suffix).
             Example: t2s_discs.nii.gz.
         """),  # noqa: E501 (line too long)
--- a/spinalcordtoolbox/scripts/sct_detect_compression.py
+++ b/spinalcordtoolbox/scripts/sct_detect_compression.py
@@ -1,0 +1,296 @@
+#!/usr/bin/env python
+#
+# Predict compression probability in a spinal cord MRI image using spinal cord shape metrics.
+# IMPORTANT NOTE: The script process *only* axial slices at the level of intervertebral discs C3/C4 (value: 4) to C6/7
+# (value: 7). In other words, the script does *not* process all axial slices of the spinal cord segmentation.
+#
+# The script requires spinal cord segmentation (used to compute the shape metrics) and disc labels.
+#
+# More details in: https://pubmed.ncbi.nlm.nih.gov/35371944/
+#
+# Copyright (c) 2024 Polytechnique Montreal <www.neuro.polymtl.ca>
+# License: see the file LICENSE
+
+import sys
+import math
+from typing import Sequence
+
+import numpy as np
+import pandas as pd
+
+from spinalcordtoolbox.process_seg import compute_shape
+from spinalcordtoolbox.aggregate_slicewise import aggregate_per_slice_or_level, func_wa, func_std, merge_dict
+from spinalcordtoolbox.centerline.core import ParamCenterline
+from spinalcordtoolbox.template import get_slices_from_vertebral_levels
+from spinalcordtoolbox.image import Image, add_suffix, splitext
+from spinalcordtoolbox.utils.fs import get_absolute_path
+from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar
+from spinalcordtoolbox.utils.sys import init_sct, set_loglevel, printv
+
+
+# Currently, the model works only for discs C3/C4 (4) to C6/7 (7)
+SUPPORTED_DISCS = [4, 5, 6, 7]
+# Cut-off to determine the presence of the compression
+CUT_OFF = 0.451
+
+def get_parser():
+    parser = SCTArgumentParser(
+        description=("""\
+Predict compression probability in a spinal cord MRI image using spinal cord shape metrics.
+IMPORTANT NOTE: The script process *only* axial slices at the level of intervertebral discs C3/C4 (value: 4) to C6/7 (value: 7).
+More details in: https://pubmed.ncbi.nlm.nih.gov/35371944/
+        """)
+    )
+
+    mandatoryArguments = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatoryArguments.add_argument(
+        '-s',
+        metavar=Metavar.file,
+        required=True,
+        help="Segmentation of the spinal cord. Example: t2s_seg.nii.gz"
+    )
+    mandatoryArguments.add_argument(
+        '-discfile',
+        metavar=Metavar.file,
+        required=True,
+        help="File with disc labels. Example: t2s_discs.nii.gz "
+             "The convention for disc labels is the following: value=3 -> disc C2/C3, value=4 -> disc C3/C4, etc."
+             "Such a label file can be manually created using: sct_label_utils -i IMAGE_REF -create-viewer 4:7 "
+             "or obtained automatically using the sct_label_vertebrae function "
+             "(the file with the \'labeled_discs.nii.gz\' suffix)."
+    )
+
+    optional = parser.add_argument_group('OPTIONAL ARGUMENTS')
+    optional.add_argument(
+        '-angle-corr',
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=1,
+        help="Angle correction for computing morphometric measures used within "
+             "\'spinalcordtoolbox.process_seg.compute_shape\'. When angle correction is used, the cord within "
+             "the slice is stretched/expanded by a factor corresponding to the cosine of the angle between the "
+             "centerline and the axial plane. If the cord is already quasi-orthogonal to the slab, you can set "
+             "-angle-corr to 0."
+    )
+    optional.add_argument(
+        "-h",
+        "--help",
+        action="help",
+        help="Show this help message and exit")
+    optional.add_argument(
+        '-v',
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1, 2],
+        default=1,
+        # Values [0, 1, 2] map to logging levels [WARNING, INFO, DEBUG], but are also used as "if verbose == #" in API
+        help="Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode")
+
+    return parser
+
+
+def predict_compression_probability(cr, csa, solidity, torsion, disc):
+    """
+    Predict compression probability using the spinal cord shape metrics for currently processed axial slice.
+    See Equation 4 and Table 3 for details: https://pubmed.ncbi.nlm.nih.gov/35371944/
+    :param cr: compression ratio (CR) in %
+    :param csa: cross-sectional area (CSA) in mm2
+    :param solidity: solidity in %
+    :param torsion: torsion in degrees
+    :param disc: intervertebral disc level (e.g., value=4 -> disc C3/C4)
+    :return: probability
+    """
+    coefficients = {
+        "constant": 57.501,
+        "cr": -0.273,
+        "csa": -0.102,
+        "solidity": -0.408,
+        "torsion": 2.168,
+        "level_c6c7": -2.729,
+    }
+    model = (
+        coefficients["constant"]
+        + coefficients["cr"] * cr
+        + coefficients["csa"] * csa
+        + coefficients["solidity"] * solidity
+        + coefficients["torsion"] * torsion
+        + (coefficients["level_c6c7"] if disc == 7 else 0)
+    )
+    return math.exp(model) / (1 + math.exp(model))
+
+def compute_shape_metrics(fname_seg, fname_disc, angle_correction, verbose):
+    """
+    Compute shape metrics per slice.
+    Then, compute compression ratio and torsion.
+    :param fname_seg: file with spinal cord segmentation
+    :param fname_disc: file with disc labels
+    :param angle_correction: bool: angle correction for computing morphometric measures
+    :param verbose: verbosity
+    :return: dictionary with aggregated metrics
+    """
+    metrics, fit_results = compute_shape(fname_seg,
+                                         angle_correction=angle_correction,
+                                         param_centerline=ParamCenterline(),
+                                         verbose=verbose)
+
+    # Compute the average and standard deviation across axial slices
+    group_funcs = (('MEAN', func_wa), ('STD', func_std))
+    metrics_agg = {}
+    for key in ['area', 'diameter_AP', 'diameter_RL', 'solidity', 'orientation']:
+        # Note: we do not need to calculate all the metrics, we need just:
+        #   - area (will be CSA)
+        #   - diameter_AP and diameter_RL (used to calculate compression ratio)
+        #   - solidity
+        #   - orientation (used to calculate torsion)
+        # Note: we have to calculate metrics across all slices (perslice) to be able to compute orientation
+        metrics_agg[key] = aggregate_per_slice_or_level(metrics[key],
+                                                        perslice=True,
+                                                        perlevel=False, fname_vert_level=fname_disc,
+                                                        group_funcs=group_funcs
+                                                        )
+    metrics_agg_merged = merge_dict(metrics_agg)
+    # Compute compression ratio
+    metrics_agg_merged = compute_compression_ratio(metrics_agg_merged)
+    # Compute torsion
+    metrics_agg_merged = compute_torsion(metrics_agg_merged, verbose)
+
+    return metrics_agg_merged
+
+def compute_compression_ratio(metrics_agg_merged):
+    """
+    Compute compression ratio (CR) as 'diameter_AP' / 'diameter_RL'
+    :param metrics_agg_merged: dictionary with aggregated metrics
+    :return: dictionary with compression ratio added
+    """
+    for metrics in metrics_agg_merged.values():  # Loop across slices
+        ap = metrics['MEAN(diameter_AP)']
+        rl = metrics['MEAN(diameter_RL)']
+        metrics['MEAN(compression_ratio)'] = None if ap is None or rl is None else ap / rl
+
+    return metrics_agg_merged
+
+def compute_torsion(metrics_agg_merged, verbose):
+    """
+    Compute torsion as the average of absolute differences in orientation between the given slice and 3 slices
+    above and below. For details see Equation 3 and Supplementary Material in https://pubmed.ncbi.nlm.nih.gov/35371944/.
+    NOTE: As the torsion is computed from 3 slices above and below, it cannot be computed for the 3 first and last
+    3 slices --> `torsion = None` is used for the 3 first and 3 last slices.
+    :param metrics_agg_merged: dictionary with aggregated metrics
+    :param verbose: verbosity
+    :return: dictionary with torsion added
+    """
+    slices = list(metrics_agg_merged.keys())[3:-3]
+
+    for key in metrics_agg_merged.keys():  # Loop across slices
+        if key in slices:
+            try:
+                # Note: the key is a tuple (e.g. `1,`), not an int (e.g., 1), thus key[0] is used to convert tuple to int
+                # and `,` is used to convert int back to tuple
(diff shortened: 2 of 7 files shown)


spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,296 @@
...
+    :return: dataframe with compression probability
+    """
+    compression_df = pd.DataFrame()
+    # Loop across discs
+    for disc, slice in disc_slices.items():
+    # for disc, slices in disc_slices.items():
+    #     # Add one slice above and below to process multiple slices around the disc to compensate for potential
+    #     # disc label shift in the superior-inferior direction
+    #     slices = [slices - 1, slices, slices + 1]
+    #     for slice in slices:
```
valosekj, line 229: During debugging, I tried to include more than one slice for each disc.

Because as mentioned in the script description:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ff2fde2eb99a55d3d887578200631dda4e14859b/spinalcordtoolbox/scripts/sct_detect_compression.py#L4-L5

The output then looks like this (i.e., the compression probability is predicted for three slices for each disc):
```bash
Disc 6 at axial slice 118 has a high probability of compression: 99.37%. 
Disc 6 at axial slice 119 has a high probability of compression: 99.05%. 
Disc 6 at axial slice 120 has a high probability of compression: 93.22%. 
```

image

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,302 @@
...
+    # Save the results to a CSV file
+    fname_out = splitext(add_suffix(fname_seg, '_compression_results'))[0] + '.csv'
+    compression_df.to_csv(fname_out, index=False)
+    printv(f"Results saved to: {fname_out}", verbose)
+
+    # Loop across compressions and print the results if the probability is above the cut-off
+    for index, row in compression_df.iterrows():
+        if row['Compression probability'] > CUT_OFF:
+            printv(f"Disc {int(row['Disc'])} at axial slice {int(row['Axial slice #'])} has a high probability "
+                   f"of compression: {row['Compression probability'] * 100:.2f}%. ", verbose)
```
valosekj, line 297: Besides outputting CSV and printing to the terminal, the function could also output a nii file with compression labels (voxel(s) with a value ‘1’ at the point of compression(s)). This file could be used for downstream tasks such as [MSCC computation](https://spinalcordtoolbox.com/stable/user_section/tutorials/shape-analysis/normalizing-morphometrics-compressions/generate-necessary-inputs.html#label-spinal-cord-compression-sites).

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,309 @@
...
+
+    # Get discs corresponding to the compression
+    disc_slices = get_disc_slices(fname_disc)
+
+    # Process axial slices for each disc to compute compression probability
+    compression_df = process_compression(metrics_agg_merged, disc_slices)
+
+    # Save the results to a CSV file
+    fname_out = splitext(add_suffix(fname_seg, '_compression_results'))[0] + '.csv'
+    compression_df.to_csv(fname_out, index=False)
```
valosekj, line 297: After adding cut-offs also for moderate and low probability of compression in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/48dbc5fec3b78810343c55d0a2695cd3ef83aab8, the output CSV now looks like:

image

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,309 @@
...
+            cr = metrics_agg_merged[slice,]['MEAN(compression_ratio)'] * 100    # to convert to %
+            csa = metrics_agg_merged[slice,]['MEAN(area)']
+            solidity = metrics_agg_merged[slice,]['MEAN(solidity)'] * 100     # to convert to %
+            torsion = metrics_agg_merged[slice,]['Torsion']
+            # Compute compression probability
+            probability = predict_compression_probability(cr, csa, solidity, torsion, disc)
+            compression_category = 'high' if probability > CUT_OFF_HIGH else \
+                'moderate' if probability > CUT_OFF_MODERATE else 'low'
+            compression_df = pd.concat([compression_df,
+                                        pd.DataFrame([{'Disc': disc, 'Axial slice #': slice,
```
valosekj, line 250: We might reconsider the name of this column, `Axial slice #`. For example, `sct_process_segmentation` outputs `Slice (I->S)` column.

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: Compression probability category values were renamed from `high/moderate/low` to `yes/possible/no` in
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/391577d0d69d769423715d77eefe9f0cf2c26404 after a discussion with https://pubmed.ncbi.nlm.nih.gov/35371944/ paper authors.

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: After a discussion with https://pubmed.ncbi.nlm.nih.gov/35371944/ paper authors, we decided on this approach (i.e., using the disc slice and one slice above and below). Implemented in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/90cbfa6654154987b0f542a6b544505fcaf8f076.

valosekj: Eventually, I made it possible to specify the number of slices to process as an input arg in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/b7af562ba1d0518b3e2d8fef99b8f01c0d7ef9ea

sandrinebedard:
I tested on 2 subjects from stanford:
Works well I think!

<details>
  <summary>sub-NSDCM001</summary> 

image

csv result:
image

C3/C4:
image

C4/C5: 
image

C5/C6:
image

C6/C7
image

</details>

<details>
  <summary>sub-NSDCM002</summary> 

image

CSV file: 
image


C3/C4
image

C4/C5:
image

C5/C6:
image

C6/C7:
image


</details>

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,333 @@
...
+    mandatoryArguments.add_argument(
+        '-discfile',
+        metavar=Metavar.file,
+        required=True,
+        help=textwrap.dedent("""
+            File with disc labels. Example: t2s_discs.nii.gz.
+            The convention for disc labels is the following: value=3 -> disc C2/C3, value=4 -> disc C3/C4, etc.
+            Such a label file can be manually created using: sct_label_utils -i IMAGE_REF -create-viewer 4:7 or
+            obtained automatically using the sct_label_vertebrae function (the file with the \'labeled_discs.nii.gz\' suffix).
+        """),  # noqa: E501 (line too long)
```
sandrinebedard, line 67: You could specify that it can be at the posterior tip of the disc or centered on the spinal cord maybe? To be consistent with other functions like `sct_register_to_template`)

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,333 @@
...
+        elif 'possible' in compression_df.loc[compression_df['Disc'] == disc, 'Compression probability category'].values:
+            printv(f"Disc {disc}, compression probability: possible", verbose)
+        elif 'no' in compression_df.loc[compression_df['Disc'] == disc, 'Compression probability category'].values:
+            printv(f"Disc {disc}, compression probability: no", verbose)
+
+    if verbose == 2:
+        for index, row in compression_df.iterrows():
+            printv(f"Disc {(row['Disc'])}, axial slice {int(row['Axial slice #'])}: "
+                   f"probability of compression: {row['Compression probability'] * 100:.2f}% "
+                   f"(compression probability category: {row['Compression probability category']})", verbose)
```
sandrinebedard, line 328: I think it would be nice to always have the percentage printed too!

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,333 @@
...
+                cr = metrics_agg_merged[slc,]['MEAN(compression_ratio)'] * 100    # to convert to %
+                csa = metrics_agg_merged[slc,]['MEAN(area)']
+                solidity = metrics_agg_merged[slc,]['MEAN(solidity)'] * 100     # to convert to %
+                torsion = metrics_agg_merged[slc,]['Torsion']
+                # Compute compression probability
+                probability = predict_compression_probability(cr, csa, solidity, torsion, disc)
+                compression_category = 'yes' if probability > CUT_OFF_HIGH else \
+                    'possible' if probability > CUT_OFF_MODERATE else 'no'
+                compression_df = pd.concat([compression_df,
+                                            pd.DataFrame([{'Disc': SUPPORTED_DISCS[disc], 'Axial slice #': slc,
```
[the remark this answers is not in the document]
valosekj: I'm trying renaming the column from `Axial slice #` to `Slice (I->S)`:

image

However, the rows are now not ordered `I->S` but `S->I`, i.e., not matching the column name `I->S`. 
I can simply reorder the rows using `df.sort_values` to get:

image

This convention is now in line with the `sct_process_segmentation` output:

image

valosekj:
Just writing it here not to forget: currently, the function outputs a separate CSV file for each subject. However, we might consider the append option, similarly as done by `sct_process_segmetnation`. In other words, compression probabilities across subjects would be aggregated into a single CSV.
If we decide on that, we would need to add `Timestamp`, `SCT Version`, `Filename` columns to this script to distinguish files across subjects (as done by `sct_process_segmetnation`):

image

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: Good point! Thanks! Done as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/ea9f82125c3aeaff0b574e619c65e4904d0e89fd!

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: Good point! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/d0eb78eedf45f71fa289194faa4cf314e86684fd.

valosekj:
Thank you for your review, @sandrinebedard! I've addressed your comments in recent commits.

I think the PR is now ready for review by @joshuacwnewton.

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,348 @@
...
+                                                            abs(metrics_agg_merged[key]['MEAN(orientation)'] -
+                                                                metrics_agg_merged[key[0] + 1,]['MEAN(orientation)']) +
+                                                            abs(metrics_agg_merged[key[0] - 1,]['MEAN(orientation)'] -
+                                                                metrics_agg_merged[key[0] - 2,]['MEAN(orientation)']) +
+                                                            abs(metrics_agg_merged[key[0] + 1,]['MEAN(orientation)'] -
+                                                                metrics_agg_merged[key[0] + 2,]['MEAN(orientation)']) +
+                                                            abs(metrics_agg_merged[key[0] - 2,]['MEAN(orientation)'] -
+                                                                metrics_agg_merged[key[0] - 3,]['MEAN(orientation)']) +
+                                                            abs(metrics_agg_merged[key[0] + 2,]['MEAN(orientation)'] -
+                                                                metrics_agg_merged[key[0] + 3,]['MEAN(orientation)']))
```
joshuacwnewton, line 237: This computation is a little difficult to parse, but I think there's a way that we can make this easier to understand.

Basically, we are calculating the same expression 6 different times, each for different slice pairs:

- `[(0,-1), (0,+1), (-1,-2), (+1,+2), (-2,-3), (+2,+3)]`
    - Though since we use `abs(difference)`, we can do this instead:
- `[(-3,-2), (-2,-1), (-1,0), (0,+1), (+1,+2), (+2,+3)]`
- Or, even simpler: `for n in range(-3, 3); (n, n+1)`

This means we can use the following expression instead:

```python
# Note: the key is a tuple (e.g. `(1,)`), thus key[0] converts to int, and `(s,)` converts back
# Here we use pairs of adjacent slices, with each slice pair offset from the initial slice
slice_pairs = [(key[0]+n, key[0]+n+1) for n in range(-3, 3)]  
metrics_agg_merged[key]['Torsion'] = 1/6 * sum([abs(metrics_agg_merged[(s1,)]['MEAN(orientation)'] - 
                                                    metrics_agg_merged[(s2,)]['MEAN(orientation)'])
                                                for s1, s2 in slice_pairs])

File: spinalcordtoolbox/scripts/sct_detect_compression.py [hunk 9]
```diff
@@ -0,0 +1,348 @@
...
+            except Exception as e:
+                # TODO: the warning below is raise mainly in cases of no SC segmentation, for example above the C1 level
+                #  --> consider processing only slices with SC segmentation
+                printv(f"WARNING: Failed to compute torsion for slice {key[0]}. Use `-v 2` flag to see details",
+                       verbose, type='warning')
+                if verbose == 2:
+                    printv(f"\t{e}", verbose)
+                metrics_agg_merged[key]['Torsion'] = None
+        else:
+            metrics_agg_merged[key]['Torsion'] = None
```
joshuacwnewton, line 247: One way that you can simplify this a little bit is to start with `= None` (as a "default" value), and this value will only be overwritten if the computation is successful:

```python
torsion = None
if key in slices:
    try:
       # [...]
    except:
       # [...]
metrics_agg_merged[key]['Torsion'] = torsion
```

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,348 @@
...
+    optional = parser.add_argument_group('OPTIONAL ARGUMENTS')
+    optional.add_argument(
+        '-num-of-slices',
+        metavar=Metavar.int,
+        type=int,
+        default=0,
+        help=textwrap.dedent("""
+            Number of slices above and below the intervertebral disc to process.
+            If 0 is provided, only the axial slice corresponding to the disc is processed.
+            If 1 is provided, the axial slice corresponding to the disc and one slice above and below are processed (i.e., 3 slices in total).
```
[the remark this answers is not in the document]
joshuacwnewton: Oh, wait, now that I look at the code, it seems like this is _not_ the behavior after all:

```python
     # Add one slice above and below to process multiple slices around the disc to compensate for potential
        # disc label shift in the superior-inferior (S-I) axis
        slices = [disc_slice] if num_of_slices == 0 else \
            [disc_slice + num_of_slices, disc_slice, disc_slice - num_of_slices]
```

So it seems like the options are actually to either use 1 slice or 3 slices, and `-num-of-slices` actually computes the... offset? So, if I put `-num-of-slices 2`, then it will use slices `(n-2, n, n+2)`? 

If this is a bug, and we don't expect users to choose a number other than `1`, then maybe we should put `choices=(0, 1)` to limit this argument.

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,348 @@
...
+    mandatoryArguments.add_argument(
+        '-discfile',
+        metavar=Metavar.file,
+        required=True,
+        help=textwrap.dedent("""
+            File with disc labels, which will be used to identify the axial slices for processing.
+            Labels can be located either at the posterior edge of the intervertebral discs, or at the orthogonal projection of each disc onto the spinal cord.
+            Such label file can be manually created using: sct_label_utils -i IMAGE_REF -create-viewer 4:7 or
+            obtained automatically using the sct_label_vertebrae function (the file with the \'labeled_discs.nii.gz\' suffix).
+            Example: t2s_discs.nii.gz.
```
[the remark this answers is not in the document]
joshuacwnewton: > Will the extra labels be silently ignored? Will an error be thrown?

We might want to filter out the extra labels explicitly (and throw a warning in the log), and then maybe add a test for this case that looks for the warning in the log, using a file containing extraneous labels?

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+# Currently, the model works only for discs C3/C4 (4) to C6/C7 (7)
+SUPPORTED_DISCS = {4: 'C3/C4', 5: 'C4/C5', 6: 'C5/C6', 7: 'C6/C7'}
+# Compression cut-offs --> compression probability
+#   >0.451 high probability of compression --> yes
+#   0.345-0.451 moderate probability of compression --> possible
+#   <0.345 low probability of compression --> no
+CUT_OFF_HIGH = 0.451
+CUT_OFF_MODERATE = 0.345
```
joshuacwnewton, line 40: Maybe we could expose these values in the argparse help and explain how they were determined? Something like:

```python
f"""
    Compression categories are determined based on the following cut-offs:
    
        - 'no'      : p <= {CUT_OFF_MODERATE}
        - 'possible': {CUT_OFF_MODERATE} < p <= {CUT_OFF_HIGH}
        - 'yes'     : p > {CUT_OFF_HIGH}

    These cut-off values were determined by [insert summary of methodology here]. More info in [REF].
"""
```

Right now, we are just outputting "yes"/"possible"/"no" in the CSV without much explanation of how those categories were determined.

joshuacwnewton (changes requested):
This is a fantastic PR! The CLI script structure looks great, the `main` function is very easy to follow, I love the re-use of existing APIs, and I appreciate all of the documentation provided. This was a very easy PR to review.

I do have some feedback, but this really is an excellent starting point. Bravo. :hearts:

spinalcordtoolbox/scripts/sct_detect_compression.py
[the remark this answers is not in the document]
valosekj: > Since we only support labels 4:7, it's not 100% clear to me what the behavior will be if the user passes in the full output of sct_label_vertebrae. Will the extra labels be silently ignored?

Good question! Yes, the script processed only labels 4:7:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fc9128e9b422148d153224ab4bb0a291f750538f/spinalcordtoolbox/scripts/sct_detect_compression.py#L268

If the input label file does not contain these labels, an error is raised:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/fc9128e9b422148d153224ab4bb0a291f750538f/spinalcordtoolbox/scripts/sct_detect_compression.py#L301-L302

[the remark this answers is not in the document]
valosekj: btw, thanks for the suggestion, I accepted it

spinalcordtoolbox/scripts/sct_detect_compression.py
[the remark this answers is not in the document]
valosekj: Good catch! Indeed, there is a bug! 

> I expected -num 2 to produce a 5 slice analysis range, hehe.

Exactly! This is supposed to be the behaviour! I fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/a2ec910fcde28cdf2a9cff2a86798d1594de0294 and added an extra test!

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,348 @@
...
+
+def get_parser():
+    parser = SCTArgumentParser(
+        description=textwrap.dedent("""
+            Predict compression probability in a spinal cord MRI image using spinal cord shape metrics.
+            The script process axial slices at the level of intervertebral discs C3/C4 (value: 4) to C6/C7 (value: 7).
+
+            Reference:
+                - Horáková M, Horák T, Valošek J, Rohan T, Koriťáková E, Dostál M, Kočica J, Skutil T, Keřkovský M, Kadaňka Z Jr, Bednařík P, Svátková A, Hluštík P, Bednařík J. Semi-automated detection of cervical spinal cord compression with the Spinal Cord Toolbox. Quant Imaging Med Surg 2022; 12:2261–2279.
+                  https://doi.org/10.21037/qims-21-782
```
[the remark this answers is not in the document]
valosekj: Good catch! Thanks! Unified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/2e21c762511504987e5c2feab5059e2245c405d3 to "doi.org" format to be consistent with other functions, e.g., [sct_compute_compression.py](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_compute_compression.py#L54-L61)

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: great suggestion! thanks! added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/b05b6f9883e98dc88aa4a49c8cbe1baf456bcd91

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: thanks 👍🏻 agree, done as part of https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/b05b6f9883e98dc88aa4a49c8cbe1baf456bcd91

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: Good idea! 👍🏻 agree. done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/36fc7da992d6e3330abff125586b349480a49c79

joshuacwnewton (changes requested):
This PR is really close to being merged! Great work again, thank you. :hearts:

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,346 @@
...
+        '-angle-corr',
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=1,
+        help=textwrap.dedent("""
+            Angle correction for computing morphometric measures. When angle correction is used, the cord within the
+            slice is stretched/expanded by a factor corresponding to the cosine of the angle between the centerline
+            and the axial plane. If the cord is already quasi-orthogonal to the slab, you can set  -angle-corr to 0.
+        """)
```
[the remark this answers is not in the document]
valosekj: Good catch! You'are right 👍🏻  Accepting.

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,346 @@
...
+        rl = metrics['MEAN(diameter_RL)']
+        metrics['MEAN(compression_ratio)'] = None if ap is None or rl is None else ap / rl
+
+    return metrics_agg_merged
+
+
+def compute_torsion(metrics_agg_merged, verbose):
+    """
+    Compute torsion as the average of absolute differences in orientation between the given slice and 3 slices
+    above and below. For details see Equation 3 and Supplementary Material in https://doi.org/10.21037/qims-21-782.
```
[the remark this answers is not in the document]
valosekj: Good question! We want to use "torsion (v3)": 
image

I believe [the current code](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d7795d871fea45c46562223ce5231ce6e3d5de6d/spinalcordtoolbox/scripts/sct_detect_compression.py#L233) is correct. So we can just update the doctoring, maybe like this:

```python
"""
Compute torsion as the average of absolute differences in orientation taking into account 3 slices above and 
3 slices below. For details, see Equation 3 in https://doi.org/10.21037/qims-21-782.
For a comparison between torsion computed using 1, 2, and 3 slices, see https://cdn.amegroups.cn/static/public/QIMS-21-782-supplementary.pdf
"""
```

(I also included the link to the supplementary material; it can be found by clicking on any cross-ref with "S" in [the article](https://qims.amegroups.org/article/view/88416/html), e.g., "Table S1")

testing/cli/test_cli_sct_detect_compression.py
```diff
@@ -0,0 +1,115 @@
...
+    # Test presence of output CSV file
+    assert os.path.isfile(filename)
+
+    df = pd.read_csv(filename)
+    assert len(df) == 5     # '-num-of-slices 2' --> 5 slices in total --> 5 rows in the CSV file
+    assert float(df[df['Axial slice #'] == 7]['Compression probability'].values) == pytest.approx(0.04174064023173820)
+    assert float(df[df['Axial slice #'] == 8]['Compression probability'].values) == pytest.approx(0.011063310989869700)
+    assert float(df[df['Axial slice #'] == 9]['Compression probability'].values) == pytest.approx(0.009026609893705780)
+    assert float(df[df['Axial slice #'] == 10]['Compression probability'].values) == pytest.approx(0.00944641208649796)
+    assert float(df[df['Axial slice #'] == 11]['Compression probability'].values) == pytest.approx(0.03042584296488690)
```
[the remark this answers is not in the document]
valosekj: Great suggestion! Thank you! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/5223ca6c51e39daaee3b7eb146c5373ab3072c5b

spinalcordtoolbox/scripts/sct_detect_compression.py
```diff
@@ -0,0 +1,346 @@
...
+    :return: dataframe with compression probability
+    """
+    compression_df = pd.DataFrame()
+    # Loop across discs
+    for disc, disc_slice in disc_slices.items():
+        # Add one slice above and below to process multiple slices around the disc to compensate for potential
+        # disc label shift in the superior-inferior (S-I) axis
+        slices = [disc_slice] if num_of_slices == 0 \
+            else range(disc_slice - num_of_slices, disc_slice + num_of_slices + 1)
+        for slc in slices:
```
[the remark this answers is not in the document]
valosekj: good point! done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/cca5be8250966e83190ab5747316bf4e5d5ad590

setup.py
```diff
@@ -77,6 +77,7 @@
                 'sct_deepseg_lesion',
                 'sct_deepseg_sc',
                 'sct_denoising_onlm',
+                'sct_detect_compression',
```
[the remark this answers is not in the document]
valosekj: Good point! Thanks! I added the script to our doc in recent commits. Current doc [here](https://spinalcordtoolbox--4760.org.readthedocs.build/4760/user_section/command-line/sct_detect_compression.html). It looks pretty good! :-)

I'm just thinking about the ref formatting. I tried to add a blank line in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/3a4d8de3ad36269c981a22842b933be96a6e6dab to format the ref as a list (as done for [sct_compute_compression](https://spinalcordtoolbox--4760.org.readthedocs.build/4760/user_section/command-line/sct_compute_compression.html)) but it formats the ref into a code block (which is actually not too bad): 

image


Before adding the blank line, the ref was formatted as a single sentence:

```
These cut-off values and compression categories were determined by ROC analysis and the Youden’s index (the sum of sensitivity and specificity). More info in the reference below.

Reference: - Horáková M, Horák T, Valošek J, Rohan T, Koriťáková E, Dostál M, Kočica J, Skutil T, Keřkovský M, Kadaňka Z Jr, Bednařík P, Svátková A, Hluštík P, Bednařík J. Semi-automated detection of cervical spinal cord compression with the Spinal Cord Toolbox. Quant Imaging Med Surg 2022; 12:2261–2279.
  https://doi.org/10.21037/qims-21-782
```

spinalcordtoolbox/scripts/sct_detect_compression.py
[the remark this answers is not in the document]
joshuacwnewton: Your proposed docstring makes perfect sense! :)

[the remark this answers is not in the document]
valosekj: Thanks! done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4760/commits/d29c00cef21fdd46de219b98f29f600d3c1651f8

setup.py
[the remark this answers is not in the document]
joshuacwnewton: > I'm just thinking about the ref formatting. I tried to add a blank line in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/3a4d8de3ad36269c981a22842b933be96a6e6dab to format the ref as a list (as done for [sct_compute_compression](https://spinalcordtoolbox--4760.org.readthedocs.build/4760/user_section/command-line/sct_compute_compression.html)) but it formats the ref into a code block (which is actually not too bad):

I think the issue here is very subtle: `sct_compute_compression` has a 2 space indent, while `sct_detect_compression` has a 4 space indentation. This is also why the "yes", "possible", "no" is also a code block.

After: 

> image

Alternatively, we could add formatting so that it is similar to SCT's studies.rst page:

> image

joshuacwnewton (approved):
Once the reference formatting nitpick is complete, I think this is good to go!

spinalcordtoolbox/scripts/sct_detect_compression.py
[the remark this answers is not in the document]
joshuacwnewton: Created an issue for the greater discussion about S->I ordering.

setup.py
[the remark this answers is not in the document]
valosekj: Excellent point, Joshua! Thanks! Looks good now, I think!

image

> Alternatively, we could add formatting so that it is similar to SCT's studies.rst page:

Yeah, this is a good idea. I'm opening a separate issue about this because we could reformat refs also in other functions, such as [sct_compute_compression.html](https://spinalcordtoolbox--4760.org.readthedocs.build/4760/user_section/command-line/sct_compute_compression.html).

spinalcordtoolbox/scripts/sct_detect_compression.py
valosekj: Moved into https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4769 to save this idea for a follow-up PR.

PR #4764, opened 2025-01-17
Add new column to QC report to rank images numerically from 0-9

### LINKED ISSUE #4758: Add new column in QC to allow ranking of segmentations
In the context of ms lesion segmentation, we want to have the possibility of ranking lesion segmentation. This means, assigning each QC line a value between 1 and 5. 

To do so, we want to create a new column where these values would be stored. This column will be named "QC ranking". The values would not be entered, but the person QCing would have to select some of the possible values (similarly to selecting "pass", "fail"...). The possible values will be: None, 1, 2, 3, 4 and 5. 

**Alternatives:** 

We had discussed adding these values to the existing QC column, where we store pass or fail. However, because in this context we don't want people filling "fail", I would suggest creating a separate column which could only store 1 to 5. 

Other discussion: Should we limit the values to None, 1, 2 and 3 (instead of None, 1, 2, 3, 4 and 5). 

Other discussion: we can also discuss the name of this new column.

**Help wanted**

I struggle to understand the organization of the QC construction with the multiple coding language involved. For that, I would appreciate if some people could guide me in what should be done. 

Tagging @jcohenadad @joshuacwnewton @mguaypaq for additional feedback on this feature request

joshuacwnewton:
## Description

Details:

- Using name ~~"Grade"~~ "Rank" as suggested by Julien
- Column is not shown by default (should it be?)
- Keys from 0-9 should work
- Grades are stored in LocalStorage and can also be exported/restored, just like QC
- Looks like this:

image

## Linked issues

Fixes #4758.


spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -147,35 +147,53 @@ $(document).ready(function(){
...
+          // Save QC state to local storage
+          localStorage.setItem(uniqueId+"_qc", sct_data[index].qc);
+          // Update table display with updated sct_data
+          set_download_yml_btn_state(heavy_excl_mark);
+          set_download_yml_btn_state(heavy_ballot_x);
+
+        }
+        if ((evt.which >= 48 && evt.which <= 57) || (evt.which >= 96 && evt.which <= 105)) {
+          // 0-9 keys (number row, keypad) => store the value directly
+          sct_data[index].grade = String.fromCharCode(evt.which);
```
joshuacwnewton, line 182: I wasn't sure about the value "None" -- do you want the actual word "None" or does "0" suffice? Otherwise, I can change it so that 0 puts "None" instead, but that seemed a little strange.

spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -147,35 +147,53 @@ $(document).ready(function(){
...
-      $("#table").bootstrapTable({data: sct_data});
-      $("#table").bootstrapTable("load", sct_data);
-      hideColumns();
-      // Set focus on the row that was just updated
-      document.getElementById("table").rows[0].classList.remove("active");
-      document.getElementById("table").rows[parseInt(rel_index)+1].className="active";
-      selected_row = document.getElementById("table").rows[parseInt(rel_index)+1].innerText;
-      document.getElementById("table").rows[parseInt(rel_index)+1].scrollIntoViewIfNeeded(false);
+    // Keys that update table state (f key, number keys)
+    if ((evt.which == 70) || (evt.which >= 48 && evt.which <= 57) || (evt.which >= 96 && evt.which <= 105)) {
```
joshuacwnewton, line 151: If someone has a spare keypad/numpad (large keyboard), could you try using it with Firefox? 

The codes should be the same as the number line in theory, but apparently for Mozilla the codes may be in a different range (96-105), which is shared by other keys, so I'm not sure how it works in practice.

jcohenadad:
> Using name "Grade" as suggested by Julien

how about "Score"? Grade could be interpreted differently (eg grade of a disease). Open to discussion.

> Column is not shown by default (should it be?)

no

jcohenadad:
Thinking more about this, I am wondering if we shouldn't call it "Rank" instead. The first usage of this new feature is to actually ask radiologist to do ranking, not grading/scoring. So Seeing the term "Rank" will be less prone to confusion. @plbenveniste what do you think?

spinalcordtoolbox/reports/assets/_assets/js/main.js
plbenveniste: I prefer "None" since it means that the ranking hasn't been assigned. Using "0" can be confusing: some people might understand it as a ranking of its own. Does that make sense?

spinalcordtoolbox/reports/assets/_assets/js/main.js
plbenveniste: I can't test this sorry. Anyone else can?

plbenveniste:
I agree with you @jcohenadad. "Rank" makes more sense and so far, I cannot imagine a situation in which a "Grade" would be given.

mguaypaq (approved):
The code looks fine (although I'm far from an expert on this part of our codebase). I guess all that remains is changing from "Grade" to "Rank" everywhere, and from "0" to "None" (null? or deleting a key?). And @plbenveniste can check in the browser that the interactions and downloading the yml file works in the way he wants.

plbenveniste (commented):
I don't have any comment on the code (I don't understand much of it). I tested the QC interface to assign a grade. It worked great. I don't have anything to change except for the name "Rank" instead of "Grade" and the ranking "None" instead of "0" (as we discussed previously. 

Thanks a lot for the help 😃

spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -147,35 +147,64 @@ $(document).ready(function(){
...
+        }
+        if ((evt.which >= 48 && evt.which <= 57) || (evt.which >= 96 && evt.which <= 105)) {
+          // Normalize numpad event codes (to avoid writing non-numeric characters to the table)
+          let code = evt.which;
+          if (evt.which >= 96 && evt.which <= 105) {
+            code -= 48;
+          }
+          // 0 key, store "None"
+          if (code == 48) {
+            sct_data[index].rank = "None"
```
mguaypaq, line 188: We should probably store an empty string, rather than `None`, to match the behaviour of an untouched column (both visually in the QC, and in the JSON file generated by "Save All").
```suggestion
            sct_data[index].rank = ""
```

spinalcordtoolbox/reports/assets/_assets/js/main.js
mguaypaq: I think it's fine to support OS/browser/keyboard combinations on a best effort basis for something like keypad behaviour.

spinalcordtoolbox/reports/assets/_assets/js/main.js
joshuacwnewton: This makes sense to me, and I like the suggestion, but as @plbenveniste is requesting this feature specifically for his own research, and has requested "None" specifically, I'm not sure if there is a reason for that text that he needs?

I think maybe it could be reasonable to merge this PR as-is, let @plbenveniste try it out for his data, while keeping a specific eye on the None/"" fields, and if there is a need, maybe change it in a follow-up PR?

Alternatively we could add event handling for the backspace/delete key to "clear" the row (thus gaining the same "" behavior while still preserving "none"). But I wouldn't want to hold up this PR any further if it is urgent.

jcohenadad: about the None vs. empty: I think the behaviour should be the same as for the check/fail/warning column, ie: there is no "None", but there is the possibility to go back to the empty value (eg: if we click on 'check' accidentallly and we want to go back to the pristine state)

plbenveniste: I agree with the last solution. I like the idea of being able to go back to an empty entry.

Issue #4670, opened 2024-10-31
Anaconda licensing changes may require changes to installer 


SomeoneInParticular:
## Justification

Anaconda quietly changed their licensing agreement earlier this year (March 2024) to clarify that they no longer allow the use of `miniconda` or of packages from the `default` channel for any organization with more than 200 employees and contractors without a paid software license. The sole exceptions to this are schools (and only if they use it as part of a class curriculum) and open-source code (but it must be completely open-source). More recently, they have also recently started pursuing legal action against some of these organizations, most recently suing Intel for their use of packages from the Anaconda channel, so this change is clearly being enforced.

While this does not effect us directly, as NeuroPoly has less than 200 employees, our use of Miniconda does potentially put SCT users at risk of breaching this new license. Right now its a legal grey area, as while they are technically using both `miniconda` and the `default` channel, whether a court would see it that way has yet to be shown. Nevertheless, its probably worth moving away from `miniconda` and the `default` channel to be on the safe side.

## Proposed Solution

I propose that, we change the install script to use `miniforge`  instead of `miniconda` . This is a conda distribution which does not use the default channel, instead using `conda-forge` . As an added bonus, `miniforge`  would also give us access to `mamba` as our package installer, which is much faster at installing packages than the default `conda` install script. 

`pip` installed packages should not be effected at all either, as they use their own channel system which does not run into the Anaconda licensing issues as far as I am aware.

## Alternatives

`mamba` (and `micromamba`) can be used as standalone package managers as well, though their initial setup differs from `conda` enough that a more substantial change to the code-base would be required than with `miniforge`

`pip` is also theoretically possible, but would require a much more substantial re-write of the installation script and associated files. Similar things can be said about the other python-based package managers available (`poetry`, `virtualenv` etc.)

`docker` also exists; using it would take far too much effort, and their license agreement is just as likely to change to put us in the exact same situation later. `moby` avoids the latter issue somewhat, but would still be overkill in my opinion.

## Additional Context

* A summary of how the Anaconda License changes may effect us can be found [here](https://www.datacamp.com/blog/navigating-anaconda-licensing)
* An article describing Anaconda suing Intel over breaching these new changes can be found [here](https://www.cdotrends.com/story/4173/anaconda-threatens-legal-action-over-licensing-terms)
* Anaconda's official statement on the License change can be found [here](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)
[...]

PR #4761, opened 2025-01-15
Add new GitHub Actions workflow to test if old releases install without error

joshuacwnewton:
## Description

This PR is a helper PR, so that I can see where the issues lie with our old releases. We can then take these fixes (or alternative fixes depending on the review) and backport them to the corresponding tags.

This is an expensive PR (huge matrix of OSs and versions), so ideally my hope is that we run this as a cron job maybe... weekly? So that we can feel safe knowing that older versions still work, without using up a ton of runners on every PR.

We can also use this workflow to test backported changes before we merge them. (e.g. if we make a branch off of `5.4` called e.g. `jn/5.4-backport-fix`, we can also temporarily update this workflow and add `jn/5.4-backport-fix` to easily install and test that version, without having to fuss around with installing the older versions locally.

## Linked issues

Partially addresses #4696.
Partially addresses https://forum.spinalcordmri.org/t/problem-installing-sct-v5-8-on-linux-mint-21-1/1367.

Issue #4696, opened 2024-11-13
Pip update (`24.1`) incompatible with `dipy==1.5.0`, causing errors for old SCT releases


joshuacwnewton:
Forum post: https://forum.spinalcordmri.org/t/problem-installing-sct-v6-1/1337

Traceback:

```python
WARNING: Ignoring version 1.5.0 of dipy since it has invalid metadata:
# Requested dipy==1.5.0 from https://files.pythonhosted.org/packages/4d/ff/933d40f73187c3e1996e833d8fe20c0863130e8d881e09bba14449a42e85/dipy-1.5.0-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl (from -r requirements-freeze.txt (line 17)) 
# has invalid metadata: Expected matching RIGHT_PARENTHESIS for LEFT_PARENTHESIS, after version specifier
    fury (>=0.8.0scikit-learn) ; extra == 'all'
         ~~~~~~~~^
Please use pip<24.1 if you need to use this version.
ERROR: Ignored the following versions that require a different python version: 1.3.1 Requires-Python >=3.10
ERROR: Could not find a version that satisfies the requirement dipy==1.5.0 (from versions: 0.5.0, 0.6.0, 0.7.0, 0.7.1, 0.8.0, 0.9.1, 0.9.2, 0.10.1, 0.11.0, 0.12.0, 0.13.0, 0.14.0, 0.15.0, 0.16.0, 1.0.0, 1.1.0, 1.2.0, 1.3.0, 1.4.0, 1.4.1, 1.5.0, 1.6.0, 1.7.0, 1.8.0, 1.9.0)
ERROR: No matching distribution found for dipy==1.5.0
```

The underlying issue here is that we freeze specific versions of our dependencies, but we _don't_ freeze `pip` itself, meaning that future `pip` upgrades can retroactively break old frozen versions of packages. 

So, ideally, in addition to limiting `pip` here, we could also consider freezing our `conda` dependencies as part of the stable release procedure. :thinking:

PR #4768, opened 2025-01-21
Change over to MiniForge as the package manager (from MiniConda prior)

SomeoneInParticular:
#### GitHub

#### PR contents

- [x] ~~I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution.~~ As this is a modification at-install, current tests should suffice.
- [x] ~~I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages.~~ Not needed, how you use the installed tool should not have changed at all.

## Description

As described in #4670 (which details this problem much more in depth than will be done here), Anaconda's recent licensing change may put our users at risk of legal trouble from the Anaconda foundation. As such, this commit changes our install script to no longer use their channel to avoid this, by swapping from `miniconda` to `miniforge` as our python environment manager.

Incidentally, doing this also gives us access to `mamba` as well, a superset of `conda` re-written to be much faster and less computationally intensive. Even with preliminary tests, this sped up my SCT installation noticeably, but further testing needs to be done before we swap over in-full.

## Linked issues

Addresses #4670, may need to be back-ported to previous versions though.


SomeoneInParticular:
From preliminary debugging, it seems that the Windows MiniForge installer errors out when it tries to allocate the install directory, but does so with a pop-up (rather than a log), hence it "hanging" for 6 hours. 

I would test further, but unfortunately (due to Microsoft's incompetence) there are no longer downloads for the Windows ISO to create a virtual machine for further testing. If anyone _has_ a windows machine they can test on, I would greatly appreciate some assistance here.

SomeoneInParticular:
Fix didn't work, hung at the same point. Going to try some other stuff before making another commit

SomeoneInParticular:
Reviewing the failed check, it seems that the errors were caused by 3 timing failures, namely:
- `sct_merge_images -h` took 0.0007s longer than the 2s threshold to run
- `sct_register_multimodal -h` took 0.00095s longer than the 2s threshold to run
- `sct_register_to_template -h` took 0.04749s longer than the 2s threshold to run.

I'll look into seeing if there's a way to improve on this, but its up to the reviewer whether this is tolerable or not as-is

joshuacwnewton:
> I'll look into seeing if there's a way to improve on this, but its up to the reviewer whether this is tolerable or not as-is

Definitely not a big blocker for the review! It might just be a one-off issue, so I've restarted the workflow run.

(I would be surprised if this PR was responsible for the change, as the relevant scientific python libraries should still be using `pip` under the hood.)

install_sct.bat
```diff
@@ -156,15 +157,15 @@ if exist %SCT_DIR%\spinalcordtoolbox.egg-info\ (
...
 echo:
-echo ### Downloading Miniconda installer...
-curl -o %TMP_DIR%\miniconda.exe https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe
+echo ### Downloading Miniforge installer...
+curl -o %TMP_DIR%\miniconda.exe -L https://github.com/conda-forge/miniforge/releases/download/24.11.2-1/Miniforge3-Windows-x86_64.exe
 echo:
-echo ### Installing portable copy of Miniconda...
-start /wait "" %TMP_DIR%\miniconda.exe /InstallationType=JustMe /AddToPath=0 /RegisterPython=0 /NoRegistry=1 /S /D=%cd%\python
+echo ### Installing portable copy of Miniforge...
+start /wait "" %TMP_DIR%\miniconda.exe /InstallationType=JustMe /RegisterPython=0 /NoRegistry=1 /S /D=%cd%\python
```
joshuacwnewton, line 166: Since commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4768/commits/5d6e2a09a89934d55dba58a0a8f4fbfdc2a9dcd0 didn't seem to have any effect on the 6h hang, does that mean that we can re-add `/AddToPath=0`? 

(This option seemed to be necessary [as of 2016](https://github.com/conda/conda/issues/1977#issuecomment-227797275) to create a "portable" installation, isolated from the rest of the system. But, on closer look, the default value now seems to be [`0`](https://docs.conda.io/projects/conda/en/stable/user-guide/install/windows.html)? Maybe still worth it to be explicit about our preferences, though, rather than relying on defaults.)

install_sct.bat
```diff
@@ -156,15 +157,15 @@ if exist %SCT_DIR%\spinalcordtoolbox.egg-info\ (
 rem Move into the SCT installation directory
 pushd %SCT_DIR% || goto error
 
-rem Install portable miniconda instance. (Command source: https://github.com/conda/conda/issues/1977)
+rem Install portable miniforge instance. (Command source: https://github.com/conda/conda/issues/1977)
 echo:
-echo ### Downloading Miniconda installer...
-curl -o %TMP_DIR%\miniconda.exe https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe
+echo ### Downloading Miniforge installer...
+curl -o %TMP_DIR%\miniconda.exe -L https://github.com/conda-forge/miniforge/releases/download/24.11.2-1/Miniforge3-Windows-x86_64.exe
```
joshuacwnewton, line 163: The change from `latest` to a fixed version is something worth highlighting... but I think it might be a welcome change?

We've run into trouble in the past due to auto-updating dependencies (see `pip==24.1` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4696). While we currently freeze our Python dependencies, we *don't* freeze our toolchain, and I'd like to fix that.

joshuacwnewton: Side note: As far as fixing the aforementioned pip issue, I started PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4761. That PR should help us more easily test backported changes in general, too.

And, since I already plan to backport a `pip` version fix ASAP, I imagine we could use this time to also test backporting `miniforge` changes as well. It will probably be a bit tedious, but at least we can guarantee that our old versions are up to date and working well. :)

install_sct
joshuacwnewton, line 1: Since `miniforge` is a drop-in replacement for `miniconda` (but with `conda-forge` defaults), we don't have to worry about any instances of the `conda` command, right? It should all just work the same?

And, it's only if we transition to `mamba` that we'd have to take a closer look at all of our conda-related scripts?

joshuacwnewton (commented):
This seems like a very easy PR to approve as-is. Perhaps we could merge this in the short-term as a transition step (to handle the immediate `anaconda` licensing issues), but also create a secondary issue to weigh the long-term pros and cons of transitioning from `conda` to `mamba`?

install_sct.bat
SomeoneInParticular: Regarding the change from latest, that was a mistake on my part; MiniForge just hides its "latest" download link for any given OS, and given the recent splurge of updates its been releasing, I wanted a set version to test with anyways. If we want to swap to an auto-updating format, we can do so with the links on [this website](https://conda-forge.org/download/).

I agree, though, that moving to freeze our toolchain is probably smart, especially given Conda-Forge sees many more updates than Anaconda, being open to public contribution and all.

install_sct
SomeoneInParticular: Correct: under the hoods, its the same `conda` being called for any commands which use it (though the version used differs between miniconda and miniforge occasionally).

Re: `mamba`, I believe it also auto-delegates to `conda` if you try to call it with a `conda` command it has not yet implemented, so it should be safe to just swap everything over. Take that with a grain of salt though; I haven't tested every single `conda` command in `mamba` yet.

install_sct.bat
SomeoneInParticular: Will commit with the change to test, must have gotten clipped by mistake when I was testing.

SomeoneInParticular:
As suspected by @joshuacwnewton, the slow Window's test failures were a fluke (Window's is such a wonder of modern engineering, ain't it?)

Just pushing the `/AddToPath=0` change quick to confirm it doesn't break anything; if it doesn't, we just need to hammer out some implementation details and should be good to go!

install_sct.bat
SomeoneInParticular: Done, CI is running now

SomeoneInParticular:
> This seems like a very easy PR to approve as-is. Perhaps we could merge this in the short-term as a transition step (to handle the immediate `anaconda` licensing issues), but also create a secondary issue to weigh the long-term pros and cons of transitioning from `conda` to `mamba`?

I don't mind having it merged pre-v7. Definitely agree an issue to discuss `mamba` vs. `conda` should be made before actually implementing it though (though that makes the PR name misleading; oh well). Let's move the conversation about that there too, so we don't bloat this PR

SomeoneInParticular:
All tests passed. Going to merge in changes from `master` to be super duper extra sure, but I think we're good to go.

I've postponed the `mamba` swap for now; see #4770 for discussion re: that

joshuacwnewton (approved):
LGTM!

I plan to test these changes further in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4761 as preparation for the backport to our older versions. But, I think it's alright to merge these changes as-is for now on our main branch. :)

SomeoneInParticular:
Sounds good. Just debugging something else quick, then will give it a final once-over and merge

PR #4776, opened 2025-01-28
Temporarily set `torch<2.6` due to upstream `weights_only=True` change

### LINKED ISSUE #4775: PyTorch 2.6 defaults to `weights_only=True` causing an unpicking error when loading model checkpoints
Failing run: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13017301707/job/36309535920

Error message:

```
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
	WeightsUnpickler error: Unsupported global: GLOBAL torch.optim.adam.Adam was not an allowed global by default. Please use `torch.serialization.add_safe_globals([Adam])` or the `torch.serialization.safe_globals([Adam])` context manager to allowlist this global if you trust this class/function.
```

Interestingly, the GPU PyTorch 2.6 doesn't seem to be out on PyPI yet, as `torch` is still on [2.5.1](https://pypi.org/project/torch/). I think that's why macOS still succeeds.

It looks like it's available from our "extras" URL, though: http://download.pytorch.org/whl/cpu/torch/

joshuacwnewton:
## Description

PyTorch 2.6 changed the default of `torch.load()` to `weights_only=True`, breaking the loading of some of our models. We can set `weights_only=False` to retain the old behavior, however there are additional calls to `torch.load()` within our dependencies as well, meaning that we need to wait until they update themselves (or become compatible with `weights_only=True`).

Until then, we can pin torch to a version prior to this breaking change.

There is a bigger conversation to be had about the attack vector of someone taking over one of the URLs to one of our models, causing users to download malicious model weights. But, I'm not sure what we would need to do to make our models become compatible with this `weights=True` option. We should watch how the upstream `nnunetv2` package handles it, since I imagine they will probably get some good discussion from downstream users.

## Linked issues

Fixes #4775.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,11 @@ dipy!=1.6.*,!=1.7.*,<1.9.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
+# In January 2025, PyTorch changed the default of `torch.load()` to `weights_only=True`, breaking the loading of some of our models.
+# We can set `weights_only=False` to retain the old behavior, however there are additional calls to `torch.load()` within our 
+# dependencies as well, meaning that we need to wait until they update themselves (or become compatible with `weights_only=True`).
+# Until then, we can pin torch to a version prior to this breaking change.
+torch<2.6
 ivadomed
 matplotlib
 # `matplotlib-inline` is an optional package that comes with `jupyter` for use with the '%matplotlib inline' directive.


mguaypaq (approved):
Thanks for the quick investigation and the context!

And yes, python's pickles are terrible for both cross-version compatibility, and security, so I'm always sad when a dependency uses them. If we can migrate to `weights_only=True` eventually, it sounds like we would avoid some potential security problems, which is always good. But as you point out, we'll have to wait and see how things unfold with our dependencies.

PR #3383, opened 2021-05-09
Add GitHub Actions workflow to publish a new release

joshuacwnewton:
#### GitHub

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [X] I've applied a [release milestone](https://github.com/neuropoly/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/neuropoly/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Tests) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/neuropoly/spinalcordtoolbox/wiki/Programming%3A-Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

In a previous PR (#3264), [**Step 1: Updating CHANGES.md with newest PRs**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#step-1-updating-changesmd-with-newest-prs) of the release process was automated.

Now, this PR automates [**Step 2: Updating the release branch**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#step-2-updating-the-release-branch) and [**Step 3: Creating the release on GitHub**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#step-3-creating-the-release-on-github). When this PR is merged, we'll have automated just about every part of the release process.

Some things I chose _not_ to automate: 

* Automatically triggering this workflow once the [`automate-changelog-PR.yml`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/workflows/automate-changelog-PR.yml) PR is merged.
[...]

PR #4779, opened 2025-01-30
Ensure that `torch<2.6` requirement gets copied over to `requirements-freeze.txt`

### LINKED ISSUE #4775: PyTorch 2.6 defaults to `weights_only=True` causing an unpicking error when loading model checkpoints

joshuacwnewton:
## Description

This commit is a follow-up to the previous fix of pinning `torch<2.6` (#4776).

The `grep -v torch` line is preventing the `torch<2.6` fix from being propagated to the `requirements-freeze.txt` file, causing the Create Release nightly workflow to fail (https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13050987631).

The `grep -v torch` line was originally added nearly 4 years ago in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3383. At that time, the state of the repository included several platform-specific frozen requirements that we wanted to explicitly copy over, so that we would override the actually-installed version of torch:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/634670e4e0e282d27abaf457def5feb49da9a1f6/requirements.txt#L33-L36

However, we now have a better way of copying over specific lines in `requirements-freeze.txt` (namely, the string `# append_to_freeze`)

## Linked issues

Fixes #4775.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,6 +11,12 @@ on:
     # > Scheduled workflows run on the latest commit on the default or base branch
     # i.e. this can only run on master
     - cron: '0 11 * * *'
+  pull_request:
+    # temporarily enable for testing, note that this refers to the base branch (i.e. master)
+    # so specifying 'jn/4775-ensure-torch-gets-frozen' wouldn't work here.
+    # See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request
+    branches:
+      - '*'
 
 env:
   MAIN_BRANCH: "master"
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -48,7 +48,6 @@ jobs:
         pip freeze | grep -v "-e git+" | grep -v "torch" | grep -v "certifi @ file" > requirements-freeze.txt
         conda deactivate
         grep "# append_to_freeze" requirements.txt >> requirements-freeze.txt
-        grep "torch" requirements.txt >> requirements-freeze.txt
 
     - name: Upload requirements-freeze.txt
       uses: actions/upload-artifact@v4
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -11,12 +11,6 @@ on:
     # > Scheduled workflows run on the latest commit on the default or base branch
     # i.e. this can only run on master
     - cron: '0 11 * * *'
-  pull_request:
-    # temporarily enable for testing, note that this refers to the base branch (i.e. master)
-    # so specifying 'jn/4775-ensure-torch-gets-frozen' wouldn't work here.
-    # See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request
-    branches:
-      - '*'
 
 env:
   MAIN_BRANCH: "master"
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release.yml
```diff
@@ -39,7 +45,7 @@ jobs:
       run: |
         source python/etc/profile.d/conda.sh
         conda activate venv_sct
-        pip freeze | grep -v "-e git+" | grep -v "torch" | grep -v "certifi @ file" > requirements-freeze.txt
+        pip freeze | grep -v "-e git+" | grep -v "certifi @ file" > requirements-freeze.txt
```
joshuacwnewton, line 48: Ah shoot, I guess `grep -v torch` is necessary after all: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13058658736/job/36436133378?pr=4779#step:5:840

I guess we can go ahead and adopt the old approach we used to have, namely copying `torch<2.6` into the frozen requirements (but still excluding it from the `pip freeze` output).

.github/workflows/create-release.yml
```diff
@@ -42,6 +48,7 @@ jobs:
         pip freeze | grep -v "-e git+" | grep -v "torch" | grep -v "certifi @ file" > requirements-freeze.txt
         conda deactivate
         grep "# append_to_freeze" requirements.txt >> requirements-freeze.txt
+        grep "torch" requirements.txt >> requirements-freeze.txt
```
mguaypaq, line 51: As it is, I think this solution would copy several more lines than intended; specifically, lines 12, 13, 14, 16 from this snippet:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7a62febc30a5b3abb376e8a64ab543dfbf6fade2/requirements.txt#L12-L17

In this case it happens to be fairly harmless lines (some comments, and an `--extra-index-url` that already has `# append_to_freeze`), but I don't like the potential for accidents in the future.

Why not just use the existing `# append_to_freeze` mechanism for the specific line(s) we want to keep?

mguaypaq (changes requested):
As it is, I

.github/workflows/create-release.yml
joshuacwnewton: ...that is a much simpler solution. :facepalm: 

I think I got flustered enough from the tests failing that I didn't see the forest for the trees here. :)

mguaypaq: That's ok! It's why we have reviews :heart:

mguaypaq (approved):
Thanks for the very quick turnaround time on this one!

PR #4783, opened 2025-02-05
Replace monai-based contrast agnostic model to nnunet-based r20250204 release

naga-karthik:
#### GitHub

## Description

This PR aims to update the `contrast-agnostic` model to the nnunet based one (which was monai-based until now). Experiments showed that the nnunet model was more robust across difficult subjects than the current monai-based model. It also fixed issues we were having with shifted predictions on a few test cases, and works _much more_ better on compressed cord and whole-spine images.  

Notable change is that I set the `keep_largest` argument as the default for this model as we have seen some cases where the model was predicting a few pixels outside the cord (quite rarely though). 

## Testing the PR

1. Check out to the branch `nk/update-contrast-agnostic-to-v3.1`
2. Run `sct_deepseg -install-task seg_sc_contrast_agnostic`
3. Run `sct_deepseg -task seg_sc_contrast_agnostic -i <path-to-the-image>.nii.gz` on your test image. 

@valosekj or @sandrinebedard could you please checkout to this branch and try the latest model? Thank you!

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -116,11 +116,11 @@ MODELS = {
     #       - Binarization is applied within SCT code
     "model_seg_sc_contrast_agnostic_softseg_monai": {
         "url": [
-            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v2.5/model_contrast-agnostic_20240930-1002.zip"
+            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v3.1/model_contrast_agnostic_20250123.zip"
         ],
-        "description": "Spinal cord segmentation agnostic to MRI contrasts using MONAI-based nnUNet model",
+        "description": "Spinal cord segmentation agnostic to MRI contrasts",
         "contrasts": ["any"],
-        "thr": 0.5,  # Softseg model -> threshold at 0.5
+        "thr": None,  # not using Softseg model anymore since switching to the more robust nnunet model
         "default": False,
     },
     "model_seg_sci_multiclass_sc_lesion_nnunet": {
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -306,11 +306,12 @@ TASKS = {
          'models': ['model_seg_epfl_t2w_lumbar_sc']},
     'seg_sc_contrast_agnostic':
         {'description': 'Spinal cord segmentation agnostic to MRI contrasts',
-         'long_description': 'This model for contrast agnostic spinal cord segmentation uses an nnUNet '
-                             'architecture, and was created with the MONAI package. Training data consists of healthy '
-                             'controls from the open-source Spine Generic Multi Subject database and pathologies from '
-                             'private datasets including DCM and MS patients. Segmentation has been tested with the '
-                             'following contrasts: [T1w, T2w, T2star, MTon_MTS, GRE_T1w, DWI, mp2rage_UNIT1], but '
+         'long_description': 'The contrast agnostic spinal cord segmentation uses a 3D CNN model based on the nnUNet '
+                             'framework. Training data consists of healthy controls from the open-source Spine Generic '
+                             'Multi Subject database and pathologies from private datasets including DCM, MS, '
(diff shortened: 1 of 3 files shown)


spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: One thing to note with this change is that the contrast agnostic model is the only model that uses the MONAI framework. (And, I'm not sure if anyone else is using MONAI in the lab, vs. nnUNet?) If we remove this model, then all of our MONAI-specific code will be unused.

So, we could consider removing `monai` as one of our requirements, and purging the MONAI-specific code, since it would no longer be needed. (Especially since much of [`monai.py`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/monai.py) is specific to the architecture we used for the contrast-agnostic model. Meaning, if we were to ever return to using MONAI, we'd probably have to write a lot of new code from scratch, as opposed to re-using the older code, so keeping the older functions may not make much sense.)

joshuacwnewton (commented):
The initial changes LGTM, but this is a big change that has large implications on SCT's codebase, so I wanted to start discussion on MONAI as a whole.

spinalcordtoolbox/deepseg/models.py
naga-karthik: > we could consider removing monai as one of our requirements, and purging the MONAI-specific code, since it would no longer be needed. 

Is monai a heavy dependency?  I am think of a scenario where a user could try one of the previous versions of the model (if they lock the SCT version) which is monai-based.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -116,11 +116,11 @@
...
     "model_seg_sc_contrast_agnostic_softseg_monai": {
         "url": [
-            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v2.5/model_contrast-agnostic_20240930-1002.zip"
+            "https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v3.1/model_contrast_agnostic_20250123.zip"
         ],
-        "description": "Spinal cord segmentation agnostic to MRI contrasts using MONAI-based nnUNet model",
+        "description": "Spinal cord segmentation agnostic to MRI contrasts",
         "contrasts": ["any"],
-        "thr": 0.5,  # Softseg model -> threshold at 0.5
+        "thr": None,  # not using Softseg model anymore since switching to the more robust nnunet model
```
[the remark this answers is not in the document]
naga-karthik: ah man, I always forget about the tests! I will get to it

spinalcordtoolbox/deepseg/models.py
jcohenadad: I would maybe wait a liiiiiiittle bit because it is possible that we will have other MONAI models in the future. But definitely something to keep on our radar.

joshuacwnewton: In that case, this PR should be pretty straightforward to merge, and we can open up a follow-up "Refactoring"-type issue. :)

naga-karthik: > If we remove the monai code, it will only affect SCT v7.0 and beyond 

You're right. I don't know what I was thinking when writing that. Previous versions of SCT would still have monai in hte requirements so it shouldn't be a problem. Sorry for the confusion

spinalcordtoolbox/deepseg/models.py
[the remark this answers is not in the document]
naga-karthik: @joshuacwnewton I was wondering what your suggestion would be here -- because we have switched to an nnunet model that doesn't produce soft masks, should we remove this test from `test_sct_deepseg.py` script? None of the other models produce soft masks ... 

Also, we still have the binary seg test for `seg_sc_contrast_agnostic` which I believe is working and should be kept.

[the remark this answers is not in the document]
joshuacwnewton: I think that makes sense! Thanks for checking here. :)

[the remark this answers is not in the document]
naga-karthik: I believe the tests have passed so I think we're good to go?

joshuacwnewton (approved):

Issue #4756, opened 2024-12-17
`totalspineseg` fails for uncompressed images (`.nii`)


joshuacwnewton:
While helping out @sriosq, we encountered an error for the 4D image file `cervical_exp_concat_mag.nii`:

```
Running step 1 model:
perform_everything_on_device=True is only supported for cuda devices! Setting this to False
Traceback (most recent call last):
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 398, in <module>
    main(sys.argv[1:])
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 298, in main
    im_lst, target_lst = inference.segment_non_ivadomed(path_model, model_type, input_filenames, thr,
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 134, in segment_non_ivadomed
    fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device)
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 361, in segment_totalspineseg
    tss_inference(
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/totalspineseg/inference.py", line 388, in inference
    predict_nnunet(
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/totalspineseg/utils/predict_nnunet.py", line 241, in predict_nnunet
    predictor.predict_from_files(str(images_dir), str(output_dir), save_probabilities=save_probabilities,
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/inference/predict_from_raw_data.py", line 245, in predict_from_files
    self._manage_input_and_output_lists(list_of_lists_or_source_folder,
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/nnunetv2/inference/predict_from_raw_data.py", line 166, in _manage_input_and_output_lists
    list_of_lists_or_source_folder = create_lists_from_splitted_dataset_folder(list_of_lists_or_source_folder,
[...]

PR #4784, opened 2025-02-05
Allow ANTs binaries to access `msvc-runtime` DLLs to avoid DLL not found errors

### LINKED ISSUE #4655: Missing `VCRUNTIME140_1.dll` error even though `msvc-runtime` is installed in `venv_sct`
### Description
When running `sct_register_multimodal -identity 1`, SCT complains about a missing ".dll" when running `sct_antsRegistration.exe`. The dll missing is: `VCRUNTIME140_1.dll`. After talking with @mguaypaq, we think that this .dll gets installed when visual studio gets installed, however, this is not an option on this machine.

Note:
I am running a weird installation of SCT which is probably why this error does not get flagged in the installation (the installation is done on another computer and the SCT folder is copied on the other computer)

### Possible solution
A solution I see and suggested by @jcohenadad is to remove the Ants dependency, and therefore the .dll missing, in `sct_register_multimodal`, maybe with a flag?

### Related
This might be related to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3740

joshuacwnewton:
## Description

As discussed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4655#issuecomment-2430178901, we install `msvc-runtime` on Windows to avoid missing DLL errors. However, we completely missed that subprocesses can't see these DLLs, meaning ANTs binaries will fail if these DLLs aren't present via other means.

We've subtly disguised this missing DLL error by asking users to [install the necessary dependency themselves](https://spinalcordtoolbox.com/stable/user_section/installation/windows.html#visual-c-2019-runtime). However, for workstations with limited installation privileges, this may not be possible. And, this second dependency installation is largely redundant with our own installation of `msvc-runtime`, anyway. 

Since we already have the necessary DLLs in the Scripts folder, why not (temporarily) add Scripts to the PATH when specifically running "SCT binaries" (i.e. the `isct_XXX` binaries)?

## Further discussion

It's possible that we may want to remove the requirement of Visual Studio Runtime from our installation docs. But, it also doesn't seem like it's necessarily harmful, either, as it's the sort of thing that users will probably want to install at some point. And, I wouldn't want to cause errors just in case there is another part of SCT that isn't covered by this path fix.

## Linked issues

Fixes #4655.

### DIFF
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -316,7 +316,7 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
         # also, for windows, add the bin directory to the path (to allow ANTs to access 'msvc-runtime' DLLs)
         # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4655#issuecomment-2430178901
         if sys.platform.startswith("win32"):
-            env['PATH'] = env['PATH'] + os.pathsep + __bin_dir__
+            env['PATH'] = __bin_dir__ + os.pathsep + env['PATH']
 
     if isinstance(cmd, str):
         cmdline = cmd
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -316,7 +316,9 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
         # also, for windows, add the bin directory to the path (to allow ANTs to access 'msvc-runtime' DLLs)
         # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4655#issuecomment-2430178901
         if sys.platform.startswith("win32"):
-            env['PATH'] = env['PATH'] + os.pathsep + __bin_dir__
+            # sanity check to make sure this is really the issue
+            pass
+            # env['PATH'] = env['PATH'] + os.pathsep + __bin_dir__
 
     if isinstance(cmd, str):
         cmdline = cmd
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -301,7 +301,7 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
         cwd = os.getcwd()
 
     if env is None:
-        env = os.environ
+        env = os.environ.copy()
 
     if is_sct_binary:
         # create an absolute path to the binary inside of SCT's bin directory
--- a/spinalcordtoolbox/utils/sys.py
+++ b/spinalcordtoolbox/utils/sys.py
@@ -316,9 +316,7 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
         # also, for windows, add the bin directory to the path (to allow ANTs to access 'msvc-runtime' DLLs)
         # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4655#issuecomment-2430178901
         if sys.platform.startswith("win32"):
-            # sanity check to make sure this is really the issue
-            pass
-            # env['PATH'] = env['PATH'] + os.pathsep + __bin_dir__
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/utils/sys.py
```diff
@@ -304,15 +304,20 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
...
         if isinstance(cmd, list):
             cmd[0] = path
         elif isinstance(cmd, str):
             rem = cmd.split(" ", 1)[1:]
             cmd = path if len(rem) == 0 else "{} {}".format(path, rem[0])
 
+        # also, for windows, add the bin directory to the path (to allow ANTs to access 'msvc-runtime' DLLs)
+        # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4655#issuecomment-2430178901
+        if sys.platform.startswith("win32"):
+            env['PATH'] = env['PATH'] + os.pathsep + __bin_dir__
```
joshuacwnewton, line 319: I'm not entirely sure why yet, but this change seems to be affecting [one of our tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13163814367/job/36740764081#step:6:997):

```
FAILED testing/cli/test_sct_deepseg.py::test_segment_nifti_binary_seg[D:\\a\\spinalcordtoolbox\\spinalcordtoolbox\\data\\sct_testing_data\\t2s\\t2s.nii.gz-D:\\a\\spinalcordtoolbox\\spinalcordtoolbox\\data\\sct_testing_data\\t2s\\t2s_seg-deepseg.nii.gz-t2s_seg_deepseg.nii.gz-seg_sc_t2star-0.9] - assert 0.34709066305818675 > 0.95
```

What's strange is that it seems to be non-deterministic:

- Run 1: `dice = 0.20235467255334805`
- Run 2: `dice = 0.34709066305818675`

I'll try running it a third time just to see, but I'm surprised that we're getting such dramatic differences vs. without the PATH modifications.

joshuacwnewton: Ahhhh, wait a minute. It's possible that this change is actually due to @NathanMolinier's parallel update of `totalspineseg` earlier today: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4756#issuecomment-2637898572

mguaypaq: Your approach seems fine, but I'm puzzled by the non-deterministic test too.

joshuacwnewton: It looks like the non-deterministic test failure was due to the `PATH` changes leaking outside of the `run_proc` commands. I switched to using `os.environ.copy` and the test now passes.

(I think this makes sense? The `run_proc` calls should only apply to calls to `isct_` binaries, but the test is related to `sct_deepseg`, so there should be no relation if the `PATH` changes are sufficiently isolated.)

spinalcordtoolbox/utils/sys.py
```diff
@@ -301,18 +301,23 @@ def run_proc(cmd, verbose=1, raise_exception=True, cwd=None, env=None, is_sct_bi
...
```
mguaypaq, line 319: Maybe put our `__bin_dir__` at the front, to give it priority? This would be consistent with a few lines above, where we force the command to be from `__bin_dir__` rather than relying on the `PATH`.
```suggestion
            env['PATH'] = __bin_dir__ + os.pathsep + env['PATH']
```

mguaypaq (approved):
Making a copy of `os.environ` rather than modifying it in-place is a good idea. Thanks!

spinalcordtoolbox/utils/sys.py
joshuacwnewton: I think I originally put our bin dir last because I wanted to minimize the effects of this change, and only supply what was missing (rather than overriding DLLs that were previously available on the PATH).

But, I guess if our goal is to isolate the SCT env from the rest of the system, then relying on files from our own venv is preferable to relying on system files anyway? 

I imagine it's the sort of thing that will be hard to predict the impact of until we see it in action. But, regardless, I think this is an okay change to make given the limited conditions where this applies.

PR #4788, opened 2025-02-10
Pin `torch` to `<2.3` to keep versions in sync with Intel Mac platforms

### LINKED ISSUE #4781: macOS vs. Linux/Windows install different versions of `torch` (2.2.2 vs. latest)
As part of my investigation into https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4775#issuecomment-2631658081, I tried freezing `torch==2.5.1`, only to run into the following error on macOS:

Image

This explains why macOS was unaffected by the torch 2.6 bug, but worryingly means that our torch versions have been out of sync for some time.

I can temporarily change the backported fix to `torch<2.6` to ensure that all versions install without error, but long-term, I want to find out why this is occurring and how to keep these versions in sync moving forward.

joshuacwnewton:
## Description

Intel Macs are only supported up until v2.2.2. Meaning, non-ARM runners are installing 2.2.2, and all other platforms are installing 2.5.1/2.6.0.

To keep our platforms in sync, I've added another restriction beyond `2.6`. (Apologies for the repeated PRs for the same dependency; this issue was only discovered during the investigation into the first issue.)

## Linked issues

Partially fixes #4781, though we may also wish to backport this fix to v6.2-6.5, which are also affected.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,11 +10,6 @@ dipy!=1.6.*,!=1.7.*,<1.9.0
 # PyTorch's Linux distribution is very large due to its GPU support,
 # but we only need that for training models. Our users only need CPU.
 --extra-index-url https://download.pytorch.org/whl/cpu # append_to_freeze
-# In January 2025, PyTorch changed the default of `torch.load()` to `weights_only=True`, breaking the loading of some of our models.
-# We can set `weights_only=False` to retain the old behavior, however there are additional calls to `torch.load()` within our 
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
This seems reasonable: until we drop support for Intel Macs, we should try to keep the same `torch` version for all platforms. And I don't _think_ any bugs or features we care about happened between 2.3 and 2.6?

PR #4793, opened 2025-02-12
Upgrade Ubuntu 20.04 -> 22.04/24.04 in test suite

### LINKED ISSUE #4792: Upgrade deprecated `ubuntu-20.04` runners in the test suite
Email went out from GitHub today: https://github.com/actions/runner-images/issues/11101

Deadline is April 1st, 2025.

joshuacwnewton:
## Description

This PR addresses the upcoming deprecation of Ubuntu 20.04 runners.

(Side note: This PR might be a cool excuse to test that "compare cross-platform results" feature I made. We could see if the batch_processing.yml results change from Ubuntu 22.04 to 24.04?)

## Linked issues

Fixes #4792.

### DIFF
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, macos-13, windows-2022 ]
+        os: [ ubuntu-22.04, macos-13, windows-2022 ]
     runs-on: ${{ matrix.os }}
     # NB: Setting `bash` as the default shell is a bit of a hack workaround for Windows to allow it to run .sh scripts.
     #     Normally, we would ask the user to install a Windows-compatible bash via either Cygwin or Git for Windows.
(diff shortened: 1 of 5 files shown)


.github/workflows/test-batch-processing.yml
```diff
@@ -121,8 +121,12 @@ jobs:
...
-        run: diff -u artifacts/sct_example_data_ubuntu-20.04.json artifacts/sct_example_data_macos-13.json | ydiff
+        run: diff -u artifacts/sct_example_data_ubuntu-22.04.json artifacts/sct_example_data_macos-13.json | ydiff
 
       - name: Diff summaries (Ubuntu vs. Windows)
         continue-on-error: true
-        run: diff -u artifacts/sct_example_data_ubuntu-20.04.json artifacts/sct_example_data_windows-2022.json | ydiff
\ No newline at end of file
+        run: diff -u artifacts/sct_example_data_ubuntu-22.04.json artifacts/sct_example_data_windows-2022.json | ydiff
+
+      - name: Diff summaries (Ubuntu 22 vs. Ubuntu 24)
```
mguaypaq, line 130: Cool!

mguaypaq (approved):
Thanks!

.github/workflows/test-batch-processing.yml
joshuacwnewton: Thanks for reminding me about this! 

It looks like there's no difference between 22 and 24? The only difference is in the [hash of the output CSV files](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13290160574/job/37111319384#step:6:16), but I think this is only because the CSV files contain timestamps?

I'd be curious to maybe remove those timestamps to see if we can minimize the diff (and thus only highlight important diffs).

PR #4803, opened 2025-02-25
Compare new `batch_processing.sh` file summaries between PRs and `master` branch

joshuacwnewton:
## Description

This PR is a follow-up to:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4759

In that PR, because the workflow hadn't been merged yet, comparisons could only be done between platforms. But, now that it's been running on `master` for a few weeks, we can compare the results of PRs against the cached results from master.

## Linked issues

Follow-up to #4759.

### DIFF
--- a/.github/workflows/test-batch-processing.yml
+++ b/.github/workflows/test-batch-processing.yml
@@ -142,7 +142,8 @@ jobs:
           do
             # Full log output
             echo "::group::Diff for $os"
-            diff -u "artifacts_master/sct_example_data_$os.json/sct_example_data_$os.json" "artifacts/sct_example_data_$os.json" > $DIFF_FILE
+            # diff will throw exit code 1 if diff is non-empty, so use '|| true' to prevent the job from failing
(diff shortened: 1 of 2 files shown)


.github/workflows/test-batch-processing.yml
joshuacwnewton, line 1: Huzzah! I think I've worked out all the kinks for this PR, and gotten warning annotations to work as well. (See this [older run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/13791588305/job/38575140767), new run currently in progress)

mguaypaq (approved):
I didn't know about [workflow commands](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions), that's pretty cool!

joshuacwnewton:
On reflection, I think I might want to add one more change to this PR, to filter the date/branch/etc. from the CSV files prior to generating their stats json. 

Until we do that, the `diff` output will put warnings on every PR, since the metadata will never match between `master` and the PRs. And that's noisy enough that I want to avoid that, hehe.

PR #4804, opened 2025-02-28
Update web tutorials to match the changes made for the 2024 SCT Course

joshuacwnewton:
## Description

This PR mimics the following PR: 

- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25

The above PR contains all of the changes we made to the `sct_tutorial_data` repo to get it to match the slides. Because of the fact that `sct_tutorial_data` was previously in sync with the web tutorials, I figured that any updates to that repo would just need to be applied to the web tutorials on a 1:1 basis to get everything back in sync.

Therefore, the commits in this PR directly reference other commits. However, a handful of `sct_tutorial_data` commits were not applicable:

<details>

> TODO for self, organize this list a bit better

Skipped commits: 

- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/70e1482782aa706364af3ca1c00f0e8a517af33f (script-specific typo fix)
- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/7708ee252ec2c491f23fe01fda126152bf8087e6 (adding missing flag that's already in tutorials)
- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/3c7323d777a98a2ac40714ae2c9b81cd0e2a66c3 (adding missing command that's already in tutorials)
   - https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/dea569bd4680cca66cf0e91fa5083c11ef39c52a (addendum to above commit)
   - https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/ce7466d035e9773ce02f3348114bca6ec5d518f1 (addendum o above commit)
   - https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/1a104c43ce86297d1e1c519fb3cda1fefb03b737 (addendum)
- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/04234a86e6dc7e12dc1dad82a0cdbfa2050ecf19 (totalspineseg, "new and upcoming features")
- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/576828a751ab3e14af1ee66e3069044e6a0584af (canal seg, "new and upcoming features")
- https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/25/commits/bb812ed1c5c3f095a70dd8b8427bf9ea09ae5b3e (specific to the multi-subject script)
[...]

PR #4809, opened 2025-03-07
Preserve the `_seg` suffix for the contrast-agnostic model

### LINKED ISSUE #4805: Newest version of contrast agnostic model outputs different default suffix
Related to:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4783

Change:
 
- Before: outputs e.g. `t2_seg.nii.gz` 
- After: outputs e.g. `t2_sc_seg.nii.gz`

This is because our nnUNet code was written to use `dataset.json` to get the output suffix:

```json
    "labels": {
        "background": 0,
        "sc": 1
    },
```

While our old MONAI code hardcoded the `_seg` suffix:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/64f55fdf022659db5d5c779f01c605d91ca7d052/spinalcordtoolbox/deepseg/inference.py#L205-L207

This change breaks our tutorials, the SCT course, etc. So we either have to update our tutorials and the SCT course to use `t2_sc_seg.nii.gz`, or we need to hardcode an exception for the contrast agnostic segmentation tool to override the output suffix.

> [!NOTE]
> This wasn't caught by the tests because we use `-o` in our tests, while this change only affects the non-`-o` usage.
>
> Testing `-o` vs. non-`-o` is a concern related to:
>
> - https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4637

joshuacwnewton:
## Description

When the contrast-agnostic model was changed from MONAI to nnUNet, the method for determining the output suffix also changed, thus the output switched from `_seg.nii.gz` to `_sc_seg.nii.gz`. I think adding the "sc" part makes sense for multiclass models (SC/lesion, SC/GM, etc.). But, for our single-class SC tools (`sct_deepseg_sc`, `sct_propseg`, etc.) we typically just output `_seg.nii.gz`. 

So, this PR adds a special case to check for single-output SC models, then choose the "basic" suffix accordingly.

This PR is being tested in https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/27/files#r1985521419, because that PR is getting blocked by this issue (SCT course was written for `_seg.nii.gz suffix`).

## Linked issues

Fixes #4805.
Needed for https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/27, which is in turn needed for https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4804.

### DIFF
--- a/spinalcordtoolbox/deepseg/inference.py
+++ b/spinalcordtoolbox/deepseg/inference.py
@@ -309,7 +309,7 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
     # for spinal cord segmentation models with only 1 output label, save the image directly with specific "_seg" suffix
     # this is added to preserve the typical expected output for SC models (`sct_propseg`, `sct_deepseg_sc`, etc.)
     # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4805
-    elif list(labels.keys()) == ['sc']:
+    elif sorted(labels.keys()) == ['sc']:
         targets = ["_seg"]
         outputs = [img_out]
     # for the other multiclass models (SCI lesion/SC, mouse GM/WM, etc.), save 1 image per label
--- a/spinalcordtoolbox/deepseg/inference.py
+++ b/spinalcordtoolbox/deepseg/inference.py
@@ -306,6 +306,12 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
     if is_rootlet_model:
         targets = ["_rootlets"]
         outputs = [img_out]
+    # for spinal cord segmentation models with only 1 output label, save the image directly with specific "_seg" suffix
+    # this is added to preserve the typical expected output for SC models (`sct_propseg`, `sct_deepseg_sc`, etc.)
+    # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4805
+    elif list(labels.keys()) == ['sc']:
+        targets = ["_seg"]
+        outputs = [img_out]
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -306,6 +306,12 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
     if is_rootlet_model:
         targets = ["_rootlets"]
         outputs = [img_out]
+    # for spinal cord segmentation models with only 1 output label, save the image directly with specific "_seg" suffix
+    # this is added to preserve the typical expected output for SC models (`sct_propseg`, `sct_deepseg_sc`, etc.)
+    # see also: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4805
+    elif list(labels.keys()) == ['sc']:
```
mguaypaq, line 312: It's a bit redundant here, because we're comparing to a list of length 1, but I generally like to make these kinds of comparisons deterministic by using `sorted`:
```suggestion
    elif sorted(labels.keys()) == ['sc']:
```

mguaypaq (approved):
Looks fine, thanks for the fix! I hope this chain of dependent fixes doesn't get any longer, hehe.

spinalcordtoolbox/deepseg/inference.py
joshuacwnewton: Neat! I didn't realize that `sorted` on a `dict_keys` object returns a `list`. Definitely internalizing that, thanks!

```python
>>> labels = {'sc': 1}
>>> labels.keys()
dict_keys(['sc'])
>>> sorted(labels.keys())
['sc']

PR #4812, opened 2025-03-11
Gracefully handle permission issues on RC files during installation

### LINKED ISSUE #3622: SCT installation fails without read permissions on `~/.bash_profile`
### Description

When there is no read permission on the .bash_profile (or equivalent), the installer throws:
```console
julien-macbook:~/code/sct $ ./install_sct 


Checking OS type and version...

Darwin julien-macbook.local 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64
ProductVersion:	10.15.7
./install_sct: line 250: /Users/julien/.bash_profile: Permission denied

Installation failed!

Please find the file /Users/julien/code/sct/install_sct_log.txt, 
```

This issue was reported by a user: https://forum.spinalcordmri.org/t/installation-failed-macbook/788

I was able to reproduce that issue by copying my `~/.bash_profile` into the file `~/.bash_profile_nopermission`, and then changing the permission on the later file:
~~~
julien-macbook:~ $ chmod 000 .bash_profile_nopermission 
julien-macbook:~ $ ll .bash_profile_nopermission 
----------  1 julien  staff  3131 11 Dec 21:01 .bash_profile_nopermission
~~~

I also substituted `.bash_profile` for `.bash_profile_nopermission` in the `install_sct` file, then I ran it and observed the error described above.

When setting read-only permission (see below), there is no error.
~~~
-r--------@ 1 julien  staff  3131 11 Dec 21:01 .bash_profile_nopermission
~~~

### Possible workaround

Should we verify the permissions of the *profile files before starting the script?

joshuacwnewton:
## Description

We've been getting more and more reports of issues related to permissions on `~/.bash_profile` and `~/.bashrc`. Ideally, these files would be owned by the user running the script (with full permissions), since they exist in the user's home directory. However, for some reason or another, this isn't the case (e.g. maybe a university/company-managed user account?).

The fix we suggest is always the same (`chown`+`chmod`), so we might as well check permissions then recommend these exact instructions in the installer. Hopefully this will reduce the number of reports we get?

As for `.bash_profile` specifically, I think that our current approach to this file is wholly unnecessary and in fact quite damaging. But, thankfully I think we can ditch it entirely, since it's a niche, invasive workaround for an already niche situation (.bashrc on macOS).

## Linked issues

Fixes #3622.


install_sct
```diff
@@ -259,28 +257,53 @@ function get_shell_rc_path() {
...
-    echo "$sourceblock" >>~/.bash_profile
+  # Note: Ideally we would properly color each individual message, however `print` adds a *lot* of
+  #       unnecessary spacing, so chaining multiple `print` messages looks very bad. We could remove
+  #       the spacing from `print`, but this would drastically change the output of our installer.
+  MSG_TO_WRITE="$MSG_EXIST\n$MSG_OWNED\n$MSG_READS\n$MSG_WRITE"
+  if [ $WARNING == "True" ]; then
+    print warning "$MSG_TO_WRITE"
+  else
+    print info "$MSG_TO_WRITE"
   fi
```
joshuacwnewton, line 307: Note that this is how the message currently looks:

image

And for e.g. an unreadable .bashrc:

image

I would prefer to have more granular coloring, but the `print` function we created for `install_sct` is a bit finicky (see above comment).

install_sct
```diff
@@ -450,6 +473,10 @@ echo -e "Installation type ... $SCT_INSTALL_TYPE"
 echo -e "Operating system .... $OS ($OSver)"
 echo -e "Shell config ........ $RC_FILE_PATH"
 
+# check permissions on shell config file (and fail if necessary)
+echo ""
+check_permissions "$RC_FILE_PATH"
```
joshuacwnewton, line 478: I was going back and forth on whether we should fail early or just warn. 

Failing early will truly force the user to act, but our current error-handling emits a big scary "Installation failed" message. We want our users to change the permissions first, not go straight to the forum.

(I also worry about the user being locked out from continuing the installation if they *can't* change the permissions on the RC file. But, maybe there's a situation where they would want to proceed and just not edit the RC file.)

Hence using just a warning.

install_sct
```diff
@@ -259,28 +257,53 @@ function get_shell_rc_path() {
...
+  if [ -w "$FILE_TO_CHECK" ]; then WRITE="True"; else WRITE="False"; fi
+
+  MSG_EXIST="  - File exists: $EXIST"
+  MSG_OWNED="  - File ownership: $OWNED"
+  MSG_READS="  - Read permissions: $READS"
+  MSG_WRITE="  - Write permissions: $WRITE"
+
+  WARNING="False"
+  if [ $EXIST == "False" ]; then
+    MSG_EXIST+="\n    NOTE: The SCT installer will create this file."
```
mguaypaq, line 281: I appreciate this non-warning note.

mguaypaq (approved):
Thank you for the screenshots! I think the warning looks great. And I'm glad we're getting rid of `force_bashrc_reloading`.

I think just printing the warning without hard-failing is the right approach. The information comes right before the "shall we?" prompt, which exactly the right time for the user to either abort and fix, or plow ahead (and hopefully answer "no" to the path-setting question).

Issue #4801, opened 2025-02-18
Indicate total processing time at the end of sct_deepseg


jcohenadad:
since some of the models are pretty long, this is useful to know--

we can display the same message that we display for sct_run_batch

SomeoneInParticular:
Self-assigned this, as I want this functionality anyways for debugging purposes.

Might also extend this to track RAM usage (given sufficiently high logging verbosity), as that seems to be a source of errors as well (particularly with `sct_run_batch`); thoughts?

SomeoneInParticular:
Just wanted to note before making a PR which allows for this functionality (and more) in a more intuitive way; this is already technically possible, but requires setting a global variable in the OS (namely, `SCT_TIMER`).

PR #4817, opened 2025-03-18
Consolidate Common Arguments into SCTArgumentParser

SomeoneInParticular:
#### GitHub

#### PR contents

- [x] ~~I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution~~ Not relevant
- [x] ~~I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages~~ See 'Other Considerations'

## Description

Currently, the `-h`, `-v`, and `-r` arguments are copy-pasted into every CLI script manually. This is very prone to error, while also preventing the easy unification of associated functionality. This PR will move their declaration into a single place, standardizing their use.

### Changes Made

This PR resolves this by adding two new functions to the SCTArgumentParser class:

* `add_common_args`: Adds the `-h` and `-v` (help and verbose) arguments in a standardized way. These are present in every CLI interface, so having them consolidated in one place will make them much easier to manage and extend
* `add_tempfile_args`: Adds the `-r` argument (remove_tempfiles) in a standardized way. Not as universal as the prior, but is a very common argument for any CLI script which generated tempfiles as part of its runtime

As well, SCTArgumentParser now has an argument group managed internally as well; "MISC". This group is used as a fallback for when the user does not explicitly define an argument group themselves in the prior functions.

### Why Sub-Class Functions?

`argparse` provides a means of argument inheritance in a similar manner via the `parents` keyword. However, this will always add 'inherited' arguments in place immediately (that is, right at the front, and in the argument groups defined by the parent), whereas we often want these common arguments placed into a specific argument group. This approach also results in difficult-to-debug behavior when adding argument groups, deleting existing groups (and their arguments) in some cases, or crashing outright in others. Working around the latter is possible, but adds overhead and complexity which is likely not worth it.

These reasons are also why it is not done automatically during init/argument parsing. Such an approach would either force us to revise the `ArgumentParser` signature (i.e. to specify the argument groups to place these arguments within), or force these common arguments to always be placed in a predefined group and position. If the latter is desirable, it is trivial to rewrite this PR to do so.

### Other Considerations

`sct_run_batch` has been exempted, as its argument parsing structure and use of -v deviates substantially from the rest of the CLI arguments; discussion on how to group its arguments needs to be made before folding it in.

This PR is a intermediary one, laying the groundwork for future standardized CLI-related capabilities, namely #4801 and #2676. On its own, it will just clean up the code base; no functionality has been added, removed, or changed.

The documentation will change slightly, as the change to `argparse` signatures created by this unification will result in the auto-generated documentation which uses them will also changing. 

This PR is likely to directly conflict with PR #4685, and as such should not be merged/reviewed until its changes have been reviewed and completed. 

## Linked issues

None directly resolved, see "Other Considerations"


SomeoneInParticular:
Closing this PR, as #4819 now exists (which is just these changes, but with the duplicated commits I accidentally created by preemptively merging #4685 purged to keep the `git` repo clean)

Issue #2892, opened 2020-09-11
Redundant information in argparse help descriptions (between "Example:" and default value)


joshuacwnewton:
Often, we manually add `Example: <value>` or `(e.g. <value>)` text in our argparse help descriptions. Sometimes this is just a repeat of the default value, for example:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/120ce12612333efd69c50bb00d27c815e981965d/scripts/sct_analyze_lesion.py#L82-L88

Because default values are now automatically shown, thanks to #2886, the following output is produced:

```
  -ofolder <folder>  Output folder (e.g. "./") (default: ./)
```

I imagine this is happening in other places, too, but I haven't done a thorough investigation yet.

We could just manually edit each description to remove redundant text, `(e.g. "./")`. Although, there might also be a better way to display "Example" vales for argparse rather than manually appending them to the description.

Drulex:
I agree. There is the [`epilog`](https://docs.python.org/3/library/argparse.html#epilog) attribute that can be used to display arbitrary text after the description of arguments. We could aggregate examples in that field?

jcohenadad:
> I agree. There is the epilog attribute that can be used to display arbitrary text after the description of arguments. We could aggregate examples in that field?

epilog is really useful indeed. We have started using it, see eg `sct_qc`

cfhammill:
So is the idea we'd refactor all instances of `Example: ` so that they appear in the epilogue and while we do that refactor remove cases where the example is the same as the default? I see these as two somewhat separate issues, although it might be a good opportunity to make two fixes at once.

I don't immediately see why using `epilog` would be preferable to putting examples in the help, although maybe it makes it more amenable to programmatic manipulation.

Drulex:
> So is the idea we'd refactor all instances of `Example: ` so that they appear in the epilogue and while we do that refactor remove cases where the example is the same as the default?

I think we would all agree on removing duplicate examples as a primary objective.

> I don't immediately see why using epilog would be preferable to putting examples in the help, although maybe it makes it more amenable to programmatic manipulation.

Personally I think that having a dedicated section for examples would encourage us to add more comprehensive examples in that section as opposed to the current (sometimes not so helpful) scoped examples that appear in the `help` section of some arguments. This adds value for users.

Drulex:
For example: https://spinalcordtoolbox.com/en/latest/user_section/command-line.html#sct-image

[...]

PR #4819, opened 2025-03-19
Consolidate Common Arguments into SCTArgumentParser (QUASHED COMMIT DUPLICATES)

SomeoneInParticular:
**THIS IS A DUPLICATE OF #4817, WITH COMMIT DUPLICATES QUASHED. IT IS OTHERWISE IDENTICAL**

#### GitHub

#### PR contents

- [x] ~~I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution~~ Not relevant
- [x] ~~I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages~~ See 'Other Considerations'

## Description

Currently, the `-h`, `-v`, and `-r` arguments are copy-pasted into every CLI script manually. This is very prone to error, while also preventing the easy unification of associated functionality. This PR will move their declaration into a single place, standardizing their use.

### Changes Made

This PR resolves this by adding two new functions to the SCTArgumentParser class:

* `add_common_args`: Adds the `-h` and `-v` (help and verbose) arguments in a standardized way. These are present in every CLI interface, so having them consolidated in one place will make them much easier to manage and extend
* `add_tempfile_args`: Adds the `-r` argument (remove_tempfiles) in a standardized way. Not as universal as the prior, but is a very common argument for any CLI script which generated tempfiles as part of its runtime

As well, SCTArgumentParser now has an argument group managed internally as well; "MISC". This group is used as a fallback for when the user does not explicitly define an argument group themselves in the prior functions.

### Why Sub-Class Functions?

`argparse` provides a means of argument inheritance in a similar manner via the `parents` keyword. However, this will always add 'inherited' arguments in place immediately (that is, right at the front, and in the argument groups defined by the parent), whereas we often want these common arguments placed into a specific argument group. This approach also results in difficult-to-debug behavior when adding argument groups, deleting existing groups (and their arguments) in some cases, or crashing outright in others. Working around the latter is possible, but adds overhead and complexity which is likely not worth it.

These reasons are also why it is not done automatically during init/argument parsing. Such an approach would either force us to revise the `ArgumentParser` signature (i.e. to specify the argument groups to place these arguments within), or force these common arguments to always be placed in a predefined group and position. If the latter is desirable, it is trivial to rewrite this PR to do so.

### Other Considerations

`sct_run_batch` has been exempted, as its argument parsing structure and use of -v deviates substantially from the rest of the CLI arguments; discussion on how to group its arguments needs to be made before folding it in.

This PR is a intermediary one, laying the groundwork for future standardized CLI-related capabilities, namely #4801 and #2676. On its own, it will just clean up the code base; no functionality has been added, removed, or changed.

The documentation will change slightly, as the change to `argparse` signatures created by this unification will result in the auto-generated documentation which uses them will also changing. 

This PR is likely to directly conflict with PR #4685, and as such should not be merged/reviewed until its changes have been reviewed and completed. 

## Linked issues

None directly resolved, see "Other Considerations"


spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -48,20 +48,15 @@ def get_parser():
         """),  # noqa: E501 (line too long)
     )
 
-    mandatory_arguments = parser.add_argument_group("MANDATORY ARGUMENTS")
-    mandatory_arguments.add_argument(
```
joshuacwnewton, line 52: Thanks for cleaning up these naming inconsistencies!

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -54,7 +54,7 @@ def get_parser():
         metavar=Metavar.file,
     )
 
-    optional = parser.add_argument_group("OPTIONAL ARGUMENTS")
+    optional = parser.add_argument_group("OPTIONAL_ARGUMENTS")
```
joshuacwnewton, line 57: Just out of curiosity, why did we add a `_` here?

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -198,7 +194,8 @@ def get_parser():
 def main(argv: Sequence[str]):
     parser = get_parser()
     arguments = parser.parse_args(argv)
-    verbose = complete_test = arguments.complete
+    complete_test = arguments.complete
+    verbose = arguments.v
```
joshuacwnewton, line 198: At first I thought this behavior change was a concern? Since, setting `-complete` will no longer have an effect on `-v`.But, I kind of prefer the new approach? By adding a separate `-v` argument, it's a lot clearer to the user what the arguments are actually doing.

That said, now that we have an explicit `-v` argument for `sct_check_dependencies`, I notice that there is specific behavior controlled by `-complete` that should honestly fall under `-v` instead? Namely, printing the output of the test commands:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7b35920d1c5b9761b155d58ad8c99ad70b12a5c6/spinalcordtoolbox/scripts/sct_check_dependencies.py#L395-L397

Though maybe there's an argument that the entire `-complete` and `-short` arguments are just _de facto_ aliases for `-v 2` and `-v 0` respectively. :laughing:

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -40,24 +40,13 @@ def get_parser(subparser_to_return=None):
         description="Segment an anatomical structure or pathologies according to the specified deep learning model.",
         epilog=models.list_tasks_string()
     )
-    parser.add_argument(
+    optional = parser.add_argument_group("OPTIONAL COMMANDS")
```
joshuacwnewton, line 43: Nit: Commands -> Arguments

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -194,49 +181,33 @@ def get_parser():
...
-
     optional.add_argument(
         '-qc-subject',
         metavar=Metavar.str,
         help="If provided, this string will be mentioned in the QC report as the subject the process was run on."
     )
 
+    # Arguments which implement shared functionality
+    parser.add_common_args()
+    parser.add_tempfile_args()
```
joshuacwnewton, line 209: I notice that `tempfile` was added to `sct_label_utils` even though a corresponding `-r` argument wasn't removed. Was this intentional?

spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -28,70 +28,75 @@ def get_parser():
...
-    parser.add_argument('-v',
-                        action='store_true',
-                        help="Verbose")
-    parser.add_argument('-h',
-                        '--help',
-                        action="help",
-                        help="show this message and exit")
+
+    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatory.add_argument(
```
joshuacwnewton, line 33: Nice catch! I hadn't realized that `sct_qc` lacked our standardized argument groups... :eyes:

joshuacwnewton (changes requested):
Thank you SO much for this colossal effort. :hearts: 

It must have been very tedious to put together! But the effort absolutely seems worth it -- the amount of reduction is quite significant. :)

spinalcordtoolbox/scripts/sct_apply_transfo.py
SomeoneInParticular: Good question, I have no idea; will push a fix momentarilly

spinalcordtoolbox/scripts/sct_check_dependencies.py
SomeoneInParticular: So, I was going to address this in a future PR (whose proposal I'm still writing); the TLDR relevant to this is as follows:

* Logging verbosity defaults to "INFO" level (`verbosity == 1` in the current standard)
* `-v` will now just be a flag (as detailed in #2676 ), which sets the logger to "DEBUG" (`verbosity == 2`) if present
* New `-q` flag (for "quiet') which is mutually exclusive with the prior, and does the opposite (setting the logger to "WARNING" (`verbosity == 0`). This is mainly to preserve the old `-v 0` functionality, but can be dropped if needed

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: Every CLI argument group we add is full uppercase; argparse's are default lower case. Moving everything to the latter wouldn't be too hard (just tedious), and would save us having to override/overload `ArgumentParser`s default behavior. I think we should bring it up at the next SCT dev meeting first though...

spinalcordtoolbox/scripts/sct_convert.py
```diff
@@ -22,41 +22,31 @@ def get_parser():
...
-        '-v',
-        metavar=Metavar.int,
-        type=int,
-        choices=[0, 1, 2],
-        default=1,
-        # Values [0, 1, 2] map to logging levels [WARNING, INFO, DEBUG], but are also used as "if verbose == #" in API
-        help="Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode")
+
+    # Arguments which implement shared functionality
+    parser.add_common_args(optional)
```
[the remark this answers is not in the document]
SomeoneInParticular: I think that would be a good idea (and streamline the function a bit), but wanted to pass it by someone else to ensure I wasn't breaking some unspoken convention somewhere. If you think its safe to do, I can make a commit to do this explicitly easy enough

spinalcordtoolbox/scripts/sct_qc.py
SomeoneInParticular: Good catch; pushing a fix now

spinalcordtoolbox/scripts/sct_apply_transfo.py
SomeoneInParticular: Resolved with [b38156f](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/b38156f90f3dd85dd94edaba84dfe639d7f7418f)

spinalcordtoolbox/scripts/sct_check_dependencies.py
SomeoneInParticular: Regarding this case specifically, I agree this needs a review pass, but that doing so is also outside of the scope of this PR; I'll open an issue to discuss this once this PR is finalized.

spinalcordtoolbox/scripts/sct_label_utils.py
SomeoneInParticular: Nope; this has been fixed as of [cee2fd3](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/cee2fd3b7e1fc726fe540b6e07360099f0e13ef3)

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: Re: the nit, this has been resolved as of [903c172](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/903c17225225a07ccfaf63fd775ccde8c1e00411)

joshuacwnewton:
((Tiny nit: I've renamed this PR to keep its simpler name, and renamed https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4817 as well to disambiguate it. I figure the noisy notes are better suited for the closed PR, rather than the open PR. :smile:))

spinalcordtoolbox/scripts/sct_convert.py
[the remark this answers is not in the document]
SomeoneInParticular: Resolved as of [bb136de](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/bb136de22bcbb7f2f11dc9d90cade1cb726bbfaf)

SomeoneInParticular:
Recent push added a few more features to aid in future unification of arguments. Namely:

* The `MANDATORY`, `OPTIONAL` and `MISC` argument groups are now pre-defined by the parser, and can be accessed to ensure uniformity
    * The `MANDATORY` group will also force added arguments to be required, without us needing to declare it!
* Passed through all CLIs to insure uniform formatting and structure (mostly example syntax and argument groups)

joshuacwnewton:
> * The `MANDATORY`, `OPTIONAL` and `MISC` argument groups are now pre-defined by the parser, and can be accessed to ensure uniformity

Splendid! I've tweaked the title a little more just to cover the (very slightly) expanded scope of the PR.

> Passed through all CLIs to insure uniform formatting and structure (mostly example syntax and argument groups)

Wonderful! Does that mean we can potentially attach this older issue as being closed by this PR?

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2892

spinalcordtoolbox/utils/shell.py
```diff
@@ -243,24 +233,59 @@ def error(self, message):
...
+        """
+        The "MANDATORY" argument group.
+
+        This should include arguments which are required for the CLI task to run.
+        """
+        mandatory = self.add_argument_group("MANDATORY ARGUMENTS")
+
+        # Force arguments in this group to be required via in-place function override
+        _f = partial(mandatory.add_argument, required=True)
+        mandatory.add_argument = _f
```
joshuacwnewton, line 248: This is really cool use of `functools.partial`! :hearts:

I was curious what would happen if a dev were to supply a value for `required` anyway. (Would it error? Which value would win out?) 

Right now, the manually-specified value will override the default, which almost feels like a bit of a shame (as it would be nice to have a built-in mechanism for ensuring that optional args aren't put in mandatory and vice versa). But I think this is good regardless!

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -32,29 +32,26 @@ def get_parser():
         description='Apply transformations. This function is a wrapper for antsApplyTransforms (ANTs).'
     )
 
-    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatory = parser.mandatory_arggroup
     mandatory.add_argument(
         "-i",
-        required=True,
-        help='Input image. Example: t2.nii.gz',
+        help='Input image. Example: `t2.nii.gz`',
```
joshuacwnewton, line 38: Thank you so much for making these changes! All of the syntax here gets parsed as Markdown when displayed in the docs (e.g. [`sct_apply_transfo`](https://spinalcordtoolbox--4819.org.readthedocs.build/4819/user_section/command-line/sct_apply_transfo.html#mandatory-arguments)), so this formatting work goes beyond just CLI convention!

spinalcordtoolbox/scripts/sct_compute_flow.py
```diff
@@ -43,10 +44,10 @@ def get_parser():
     optional.add_argument(
```
joshuacwnewton, line 44: Ctrl+F'ing for `parser.add_argument_group`, I notice 2 instances of `OPTIONAL ARGUMENTS` that may have gotten missed.

- `sct_compute_flow`
- `sct_dmri_display_bvecs`

spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -187,6 +186,7 @@ def get_parser():
         "-short",
         help="Short test. Only shows SCT version, CPU cores and RAM available.",
         action="store_true")
+    parser.add_common_args(optional)
```
joshuacwnewton, line 189: Likewise, I notice that we're adding the common args to `optional` for `sct_check_dependencies`? I'm not sure if that's a stylistic choice or if we still want to keep things consistent in `MISC`?

spinalcordtoolbox/scripts/sct_compute_compression.py
```diff
@@ -86,33 +84,14 @@ def get_parser():
...
-        type=int,
-        choices=[0, 1],
-        help=textwrap.dedent("""
-            Set to 1 to normalize the metrics using a database of healthy controls. Set to 0 to not normalize.
 
-            Note: This flag should not be set to 1 when computing the MCC (i.e. using spinal canal segmentation). It should only be used when computing the MSCC (i.e. using spinal cord segmentation).
-        """),  # noqa: E501 (line too long)
-    )
-
-    optional = parser.add_argument_group("OPTIONAL ARGUMENTS")
```
joshuacwnewton: Good catch, finding these non-mandatory arguments that were under `MANDATORY`!

I also completely agree with moving them into their own `NORMALIZATION` category, feels very clean. :)

spinalcordtoolbox/scripts/sct_compute_flow.py
```diff
@@ -7,6 +7,7 @@
 
 import sys
 from typing import Sequence
+from textwrap import dedent
```
joshuacwnewton, line 10: This is the tiniest nitpick, but what are your feelings on `from textwrap import dedent` vs. `import textwrap; textwrap.dedent`?

(Asking only because I think other scripts are using the latter form, and I'm wondering if that's something worth keeping consistent, or changing elsewhere if there's a significant difference between the two.)

spinalcordtoolbox/scripts/sct_compute_mtsat.py
```diff
@@ -28,63 +28,68 @@ def get_parser():
...
-    optional = parser.add_argument_group('OPTIONAL ARGUMENTS')
-    optional.add_argument(
+    sidecar_overrides = parser.add_argument_group(
+        "JSON SIDECAR OVERRIDES",
+        description="Any arguments within this group will override their corresponding values in the JSON side-car files "
+                    "of the MRI sequences you have provided. Otherwise, all values will be fetched from the "
+                    "associated JSON sidecar, assuming it exists. If not, you should provide the corresponding metrics "
+                    " using these arguments."
+    )
+    sidecar_overrides.add_argument(
```
joshuacwnewton, line 55: Another very sensible change! Thank you for de-duplicating the old message (and providing a very clear representation of the behavior + user guidance).

spinalcordtoolbox/scripts/sct_compute_snr.py
```diff
@@ -71,6 +67,11 @@ def get_parser():
              "compute SNR on is specified by `-vol`.",
         choices=('diff', 'mult', 'single'),
         default='diff')
+    optional.add_argument(
+        '-m-noise',
+        help="Binary (or weighted) mask within which noise will be calculated. Only valid for `-method single`.",
+        metavar=Metavar.file,
+        default='')
```
joshuacwnewton, line 74: Just out of curiosity, why did this argument get re-ordered? Was it just for alphabetical sake?

(I think the purpose of its old location was to group it thematically with the `-m` argument?)

spinalcordtoolbox/scripts/sct_deepseg_gm.py
```diff
@@ -69,20 +53,36 @@ def get_parser():
...
         choices=('large', 'challenge'),
         default='large')
-    misc.add_argument(
+    optional.add_argument(
         "-thr",
         type=float,
         help='Threshold to apply in the segmentation predictions, use 0 (zero) to disable it. Example: `0.999`',
         metavar=Metavar.float,
         default=0.999)
-    misc.add_argument(
```
joshuacwnewton, line 82: Nice!! I really like the idea of standardizing on QC == MISC, but all other (functional) optional args in OPTIONAL.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -148,7 +149,7 @@ def get_parser(subparser_to_return=None):
                  "smaller than 10 voxels for class 1 and 3, and smaller than 20 voxels for class 2).",
             default=None)
 
-        misc = subparser.add_argument_group('\nMISC')
```
joshuacwnewton, line 162: On that note, it seems like there might still be some places where `MISC` is used for "non-QC/non-common" functional args? (e.g. `sct_deepseg` with `-thr`, `-remove-small`, `-keep-largest`, etc.)

I did a quick Ctrl+Shift+F for `misc.add_argument(` to find these instances.

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton, line 1: Also in `sct_deepseg`, I notice that an `-r` argument is still present! 

(I did a project-wide regex search for `add_argument\(\n*\s*['"]-r['"],`, and it seems like only the single instance was missed.)

spinalcordtoolbox/scripts/sct_deepseg_lesion.py
```diff
@@ -53,26 +51,22 @@ def get_parser():
...
         "-brain",
         type=int,
         help='Indicate if the input image contains brain sections (to speed up segmentation). This flag is only '
              'effective with `-centerline cnn`.',
-        required=False,
         choices=(0, 1),
         default=1)
     optional.add_argument(
         "-ofolder",
-        help='Output folder. Example: My_Output_Folder',
```
joshuacwnewton, line 80: Nit, but here's a case of an `Example:` argument potentially clashing with a default value.

spinalcordtoolbox/scripts/sct_denoising_onlm.py
```diff
@@ -41,30 +41,26 @@ def get_parser():
                     'Engineering and Technology, 2011). The implementation is based on Dipy (https://dipy.org/).'
     )
 
-    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatory = parser.mandatory_arggroup
     mandatory.add_argument(
         "-i",
-        default=None,
```
joshuacwnewton, line 47: Nice! `default=None` is indeed redundant.

spinalcordtoolbox/scripts/sct_qc.py
```diff
@@ -29,22 +29,20 @@ def get_parser():
...
     )
 
-    mandatory = parser.add_argument_group("MANDATORY ARGUMENTS")
+    mandatory = parser.mandatory_arggroup
     mandatory.add_argument(
         '-i',
         metavar='IMAGE',
-        help='Input image #1 (mandatory)',
-        required=True)
+        help='Input image #1 (mandatory)')
```
joshuacwnewton, line 36: This extra `(mandatory)` seems a bit redundant, eh? :laughing:

joshuacwnewton (changes requested):
This new commit is lovely! Thank you so much for going through all of the scripts and cleaning things up. :hearts:

spinalcordtoolbox/utils/shell.py
SomeoneInParticular: > I was curious what would happen if a dev were to supply a value for required anyway. (Would it error? Which value would win out?)

It would error, yes (due to "duplicated keyword arguments"); in theory, though, we should never need to specify "required" with this setup at all, for any arguments going forward.

> Right now, the manually-specified value will override the default, which almost feels like a bit of a shame (as it would be nice to have a built-in mechanism for ensuring that optional args aren't put in mandatory and vice versa). But I think this is good regardless!

I considered implementing that actually, but decided against it for 3 reasons:

1. It would make the use of "MANDATORY" group distinct from the other two ("MISC" and "OPTIONAL"), and indeed any other custom groups. 
2. Theirs no way to realistically automate similar functionality for deciding whether new arguments should be in "MISC" or "OPTIONAL" (aside from just dumping everything into "OPTIONAL" outright, which seems a bit odd)
3. Its actually nearly impossible to implement this efficiently, as argument groups are actually sub-classes of the argument parser they are created from. This would result in trying to add arguments to explicit sub-groups being "stolen", unless we add some really convoluted logic to validate every new argument.

spinalcordtoolbox/scripts/sct_compute_flow.py
SomeoneInParticular: Good catch, fixed with [bf2a711](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/bf2a711fcf8bb9f2eba1ed085c0e3e42f0483a3e)

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -333,36 +320,36 @@ def get_parser():
...
         '-type',
         choices=('uint8', 'int16', 'int32', 'float32', 'complex64', 'float64',
                  'int8', 'uint16', 'uint32', 'int64', 'uint64'),
         default=None,  # needed because argument_default=argparse.SUPPRESS
         help='Output type.',
-        required=False)
+        )
 
     # Arguments which implement shared functionality
     parser.add_common_args(misc)
```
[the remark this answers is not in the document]
SomeoneInParticular: Good catch; I did catch that, but just made it so it adds arguments to our "standard" MISC group instead of using its own in that case. I can't think of a much better solution too this, honestly, without setting up a formal way to decide when "custom" argument groups should be used.

spinalcordtoolbox/scripts/sct_check_dependencies.py
SomeoneInParticular: Nope, thats just me missing it. Fixed with [aeb95b8](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/aeb95b8374a59289f3028866514a532909ac57a1)

spinalcordtoolbox/scripts/sct_compute_flow.py
SomeoneInParticular: I always lean to importing the explicit call personally (`from ... dedent`), as it helps reduce typing in this case without being likely to be mistaken for something else in the namespace. 

It is also, technically, _slightly_ more efficient computationally, only requiring one name look-up each call rather than 2. Honestly, though, this is such a minor change that if the lab prefers the former, then I don't mind swapping over.

spinalcordtoolbox/scripts/sct_compute_snr.py
SomeoneInParticular: Oh, I moved it so that it was placed below the flag it was referencing (so that the user doesn't wonder what the hell its referring to)

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: See prior remark

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: Good catch, resolved with [b15eacb](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/b15eacbf7add58e2ec8a9af0a6c6101b6e9cdb40)

spinalcordtoolbox/scripts/sct_compute_snr.py
joshuacwnewton: Oh!! Good catch! I missed that. I'll mark this as resolved.

spinalcordtoolbox/scripts/sct_qc.py
SomeoneInParticular: No idea how I missed that; resolved with [db9489b](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/db9489b72cae3025ed85a4fd215efddb7c79479a)

spinalcordtoolbox/scripts/sct_deepseg_gm.py
joshuacwnewton: That's very fair. I'm A-OK treating it as out of scope for now.

spinalcordtoolbox/scripts/sct_maths.py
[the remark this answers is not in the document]
joshuacwnewton: Ah, so just doing `parser.add_common_args()` won't work in this case?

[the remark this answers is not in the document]
SomeoneInParticular: OH, thats what you meant; sorry, missed that entirely. I thought you were referring to it using the "MISC" group directly. 

Resolved with [ca04394](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/ca04394e2aec341de0478801c41429b8d30d58fd)

spinalcordtoolbox/utils/shell.py
SomeoneInParticular: Slight correction; it actually doesn't error (at least on Python 3.9); functools seems to be smarter than I gave it credit for! Evidence `sct_compute_snr` was working despite its use in `-i`; this has been fixed in [abf9fa8](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/abf9fa860997f58f604dadc47fd01cc7966a1249)

spinalcordtoolbox/scripts/sct_deepseg_lesion.py
SomeoneInParticular: As discussed, see commit [929a46b](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/929a46ba9a878bd8962659671598fbf8becdc673) to resolve (hopefully)

joshuacwnewton (approved):
It's a shame that the default "`default:` with formatting" commit didn't work out, but it's a bit tangential to this PR anyway, so I'm happy to approve!

spinalcordtoolbox/scripts/sct_deepseg_gm.py
```diff
@@ -42,8 +42,7 @@ def get_parser():
     optional.add_argument(
         "-o",
         help="Output segmentation file name. Example: `sc_gm_seg.nii.gz`",
-        metavar=Metavar.file,
-        default=None)
```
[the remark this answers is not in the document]
SomeoneInParticular: Ah, good catch. I'll review them all to validate their isn't some unusual use case, and remove those which don't. Shouldn't take too much time (knocks on wood).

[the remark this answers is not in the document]
SomeoneInParticular: Should be resolved with [685b2ee](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4819/commits/685b2ee26e97ddefc595c29056ea2eea73c3f7c3); assuming it doesn't cause any errors in the checks, this should be the final commit!

PR #4824, opened 2025-03-24
Pin SetupTools to Prevent Crashing at Documentation

### LINKED ISSUE #4822: Malformed `batchgenerators` Config Results in Failed Documentation Build
### Description

Documentation fails to be built, as one of our dependencies (`batchgenerators`) uses a now-invalid tag in their setup-tools configuration (with a dash, when it should be an underscore).

This is a bug out of our control (see [the issue on `batchgenerators`](https://github.com/MIC-DKFZ/batchgenerators/issues/135)), but should be kept in mind until it is resolved.

### Steps to Reproduce

0. Try to build the docs on ReadTheDocs (i.e. through the GitHub CI action on PR update)

**Expected behavior:** The documentation is build correctly

**Actual behavior:** Errors out, complaining about "Invalid dash-separated key 'description-file' in 'metadata' (setup.cfg)", resulting in no documentation being built. Example available [here](https://app.readthedocs.org/api/v2/build/27616579.txt)

SomeoneInParticular:
#### GitHub

#### PR contents

## Description
See issue #4822; this PR aims to try and fix it (by pinning SetupTools to a version where the issue does not occur)

## Linked issues
Resolves #4822

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,6 +10,22 @@ build:
   tools:
       # This python version needs to be compatible with SCT, see setup.py
       python: "3.9"
+  jobs:
+    # NOTES:
+    # - Add fix for mismatch between `batchgenerators` and `setuptools==78.0`
+    # - We choose 75.8.2 to match the version installed in the conda 3.9 venv_sct
+    # - This step *should* run after the environment is created, but before
+    #   SCT+dependencies are installed (i.e. the `python: install:` config below)
+    # - The reason we can't just pin this in `requirements.txt` is because
+    #   RTD automatically upgrades `setuptools` beforehand. So we need to
+    #   downgrade it after the environment is created + upgraded, but *before*
+    #   installing everything in `requirements.txt`.
+    #
+    # FURTHER READING:
+    #  - https://docs.readthedocs.com/platform/stable/builds.html#understanding-the-build-process
+    #  - https://docs.readthedocs.com/platform/stable/build-customization.html#extend-or-override-the-build-process
+    post_create_environment:
+      - python -m pip install setuptools==75.8.2
 
 python:
   install:
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -24,7 +24,7 @@ build:
     # FURTHER READING:
     #  - https://docs.readthedocs.com/platform/stable/builds.html#understanding-the-build-process
     #  - https://docs.readthedocs.com/platform/stable/build-customization.html#extend-or-override-the-build-process
-    pre_install:
+    install:
       - python -m pip install setuptools==75.8.2
 
 python:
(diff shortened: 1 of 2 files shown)


SomeoneInParticular:
Failed as well; seems that SetupTools is pinned on ReadTheDoc's end, so we cannot do anything about this until one or the other changes! Yay!

joshuacwnewton:
> Failed as well; seems that SetupTools is pinned on ReadTheDoc's end, so we cannot do anything about this until one or the other changes! Yay!

I wouldn't rule out your solution quite yet? I think we're close!

I notice that in the failing run, we see the newest version get installed [first](https://app.readthedocs.org/projects/spinalcordtoolbox/builds/27619189/#265082565--18), and then we [attempt the downgrade](https://app.readthedocs.org/projects/spinalcordtoolbox/builds/27619189/#265082626--81), but because it's being installed simultaneously with the faulty version of `batchgenerators`, pip fails before it can downgrade.

So, I'm wondering if there's a way to downgrade `setuptools` *before* installing the set of requirements. I'm going to look into this and see if I can revive the PR. :)

.readthedocs.yaml
```diff
@@ -10,6 +10,22 @@ build:
...
+    # - The reason we can't just pin this in `requirements.txt` is because
+    #   RTD automatically upgrades `setuptools` beforehand. So we need to
+    #   downgrade it after the environment is created + upgraded, but *before*
+    #   installing everything in `requirements.txt`.
+    #
+    # FURTHER READING:
+    #  - https://docs.readthedocs.com/platform/stable/builds.html#understanding-the-build-process
+    #  - https://docs.readthedocs.com/platform/stable/build-customization.html#extend-or-override-the-build-process
+    post_create_environment:
+      - python -m pip install setuptools==75.8.2
```
joshuacwnewton, line 28: It looks like my first attempt (`post_create_environment`) inserted the `setuptools` downgrade at the wrong location:

image

Going to try `pre_install` instead, next.

joshuacwnewton: No dice, same issue. :thinking: 

image

joshuacwnewton: Aha! It looks like the solution is to override the `install` step, adding my manual install while also replicating the existing steps (sphinx, SCT install).

This should be easy enough!

joshuacwnewton: Huzzah, it worked! :)

SomeoneInParticular:
This issue seems to have resolved itself on ReadTheDocs side; my recent push to PR #4825 ran fine, even without its branch being branched off of this PR's... still might be worth doing this though to be on the safe side

joshuacwnewton:
Yep, we can close this & alert the `batchgenerators` package of the non-issue.

PR #4827, opened 2025-03-25
Rewrite & extract QC report UI

elhil:
#### GitHub

#### PR contents

## Description
To allow for easier updates and bug fixes on QC reports, the javascript/HTML has been extracted to [spinalcordtoolbox/report-ui](https://github.com/spinalcordtoolbox/report-ui), where it can take advantage of a build system meant for JavaScript, as well as other modern tooling, without adding noise to this repo. Instead, version releases from there will be copied to this repo when appropriate (see 85237653a876a71106c2d2ea05c62367d47300f6 for an example update, with instructions).

The current implementation aims for full backward compatibility with the current report system. That said, since generated reports are self-contained HTML directories, some minor changes were convenient to make, namely:
- the JSON generated to aid rendering the report uses camelCase for better compatibility with JS naming standards, and some fields are renamed for clarity
- the directory structure of the generated reports has changed
- the YAML created by saving fails/artifacts has been slightly streamlined

The new UI comes with many performance and layout improvements while still maintaining the structure & feel of the legacy UI. Users with muscle memory for the reports should find it effectively the same.

An upshot of this rewrite is that new features, refactors, and fixes should be much easier: the modularity and pure-function orientation of React makes this so!


elhil:
Probably makes sense to rebase once https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4825 is merged.

SomeoneInParticular:
> Probably makes sense to rebase once #4825 is merged.

I think it should be safe to merge this PR before or after #4825; they don't have overlap currently, and I will make a future QR to standardized how QC is mediated via CLI later (once we've discussed how best to approach matters with the entire lab)

spinalcordtoolbox/reports/qc2.py
joshuacwnewton, line 1: Navigating to https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14066366003?pr=4827 and downloading [**Batch Processing QC (ubuntu-24.04)**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14066366003/artifacts/2817968026), I'm not seeing any issues whatsoever! Everything works exactly like we've been testing it. :hearts:

image

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton, line 1: The one tiny thing I notice is that the tab title reads "Vite + React + TS". If this is changeable, maybe something like "SCT QC Report" would be better?

If we're feeling really fancy, and it's easy to change the tab title dynamically based on the highlighted row, maybe even "SCT QC Report - {subject_name}" could be worthwhile. But that's probably overkill.

joshuacwnewton (commented):
This is wonderful!!! I've been so happy to see this come together.

I would love to ask @valosekj and @sandrinebedard if they notice any QC report layouts that are missing from the Batch Processing test that we should verify? (See comment below.)

Otherwise, we can chat about this more at Thursday's meeting. :)

valosekj:
Thank you @elhil!

> I would love to ask @valosekj and @sandrinebedard if they notice any QC report layouts that are missing from the Batch Processing test that we should verify? (See comment below.)

I'm comparing [this PR #4827 Batch Processing QC (ubuntu-24.04)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14066366003/artifacts/2817968026) (left) with [master's Batch Processing QC (ubuntu-24.04)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14058042379/artifacts/2815000834) (right):

image

A few minor findings:
- the shadowing of the currently selected file/row could be more visible. I can barely see it on the left side (`#4827`). (it's worth saying that even on the right side (`master`) is hard to see it 😄)
- for longer commands, such as `sct_register_multimodal`, the `master`'s QC shows the whole command making it very easy to read. While `#4827` renders the command in a box with a slider (requiring scrolling to see the entire command).
- the overall layout is slightly different: instead of having everything (image and command) aligned to the top right, now the image is centred, and the command is aligned to the bottom right. Is there a specific reason for this? I guess it's not a big deal and it's just a matter of habit. That said, when all components are aligned to the top right corner (`master`), it is easy to make a small screenshot capturing both the image and the command (which is sometimes useful for sharing).

elhil:
Thanks for the feedback @valosekj
> * the shadowing of the currently selected file/row could be more visible. I can barely see it on the left side (`#4827`). (it's worth saying that even on the right side (`master`) is hard to see it 😄)

I'll add this for sure.
> * for longer commands, such as `sct_register_multimodal`, the `master`'s QC shows the whole command making it very easy to read. While `#4827` renders the command in a box with a slider (requiring scrolling to see the entire command).
> * the overall layout is slightly different: instead of having everything (image and command) aligned to the top right, now the image is centred, and the command is aligned to the bottom right. Is there a specific reason for this? I guess it's not a big deal and it's just a matter of habit. That said, when all components are aligned to the top right corner (`master`), it is easy to make a small screenshot capturing both the image and the command (which is sometimes useful for sharing).

The reasoning behind vertically centering the image and constraining the size of the command display was to reduce the visual jitter of moving quickly through the rows. I hadn't thought of the screenshot use case, which is also presumably a good reason to show the full command.

These updates are both complete in the latest commit. @valosekj can you tell me if this behaves correctly?

mguaypaq (approved):
I haven't tested it locally, but code-wise everything looks fine. Thanks for this big improvement!

valosekj:
Thank you, @elhil!

The row shadowing now looks good! The long commands are also rendered fine now! The only thing is that there are now images in the QC now (I downloaded [this artifacts](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14091467928/artifacts/2826529572)) 😅 

Screenshot 2025-03-27 at 1 29 40 PM

valosekj:
Hmm, this is very interesting! I just opened the QC report in a different browser (Chrome instead of Safari), and now the images display correctly. Thanks for the tip!

Out of curiosity: I often have multiple reports open simultaneously when comparing different analyses, and I’ve never run into this issue before. Could this be related to the refactoring introduced in this PR?

spinalcordtoolbox/reports/assets/_assets/js/main.js
```diff
@@ -1,327 +0,0 @@
...
-      }
-  };
-  reader.readAsText(file);
-}
-
-function updateQcStates() {
-    // Load and set QC state from local storage
-    // TODO: create a function for this code block
-    sct_data.forEach((item, index) => {
-      var uniqueId = sct_data[index].moddate + '_' + sct_data[index].fname_in + '_' + sct_data[index].command;
```
joshuacwnewton, line 313: It looks like the old QC report used the `moddate`+`fname_in`+`command` as part of a way to try to guarantee a unique ID for the entry. Which I guess makes sense, given that it's path-independent, and it's unlikely for a QC report to be generated to the exact second with the exact filename and command? :thinking:

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton, line 1: My apologies for the delay in testing this feature, but, I finally got around to:

- https://github.com/spinalcordtoolbox/report-ui/issues/9

(This PR made it easier -- just run the test `testing/cli/test_sct_deepseg.py :: test_deepseg_with_cropped_qc`.)

Anyway, here's the QC ([qc.zip](https://github.com/user-attachments/files/19539085/qc.zip)), and here's how it looks:

image

Everything fits ( :tada: ) but unfortunately the "zoom out" effect makes it hard to actually QC the image. This is with cropped data, too -- with larger images, I could see the mosaic being shrunken down even more.

For reference:

- Old QC: Display the image in full, even if it means scrolling the table outside the viewport.
- New QC Proposal 1: Allow for scrolling within the image container (while still keeping everything else on screen)
- New QC Proposal 2: Allow for zooming in and out

I'm hoping that the new backend makes implementing either of these suggestions easier? Or if you have an alternate suggestion, that would be lovely too!

spinalcordtoolbox/reports/assets/_assets/js/main.js
joshuacwnewton: Tried the new UUID approach on Firefox, and everything seems to be working splendidly! I like that we're keeping the whole table now (as well as the metadata), rather than doing these sorts of ad-hoc per-row unique IDs.

Samples (note the old entries for QCs that were cached from other reports I put on my Desktop, aha):

image

image

spinalcordtoolbox/reports/assets/index.html
elhil: For Proposal 1: I assume the image would take up the full width of the column, and then whatever height it needs? "Landscape" oriented pictures already get downsized to fit the width of the column. Or should all images be displayed at original resolution, with scrollbars? I *think* a zoom feature makes more sense—what's your intuition?

Zoom options: click to zoom, hover to zoom, key that makes the image expand to take up the entire column, eclipsing the command info box? @valosekj any thoughts as a user?

joshuacwnewton: > I assume the image would take up the full width of the column, and then whatever height it needs?

Exactly. Right now, the way the mosaic generation code works, it will display each image in the grid at a 1:1 pixel ratio (or upscaled 2x if the resolution is especially tiny), with the goal being to display the segmentations faithfully.

To avoid creating mosaics that are too large (width-wise), I think we have a hardcoded pixel value (something like ~1000 pixels) where mosaics will wrap to the next row when the limit is exceeded.

Meaning, it should always be possible to display pixels faithfully without requiring horizontal scrolling, i.e. overflow should only ever occur vertically (famous last words).

valosekj: Some sort of zooming feature might be indeed relevant! Currently, when we need to zoom (e.g., to see some subtle details on an axial segmentation), we simply zoom within our browsers.

> @valosekj any thoughts as a user?

How would the zoom feature work? Only for the selected image or globally? I'm asking because if we have hundreds of images, currently with the "Old QC" layout, we can scroll across images (rows in the QC report) pretty fast using the previous and next arrow keys because the images are shown in full-size (no need to zoom).

@sandrinebedard, @naga-karthik, any thoughts?

elhil: Maybe it would be helpful to express the specification as some "user stories". For example, from your comment, it sounds like "as a user, I need to easily and quickly scroll through full-resolution images".

The question, then, is whether, as a user, you need to always be able to see the command information window (the info below the image). We can always show full-size images, at the expense of the command window, or we can show the command window, and scale the image to use the space that's left.

It sounds like the old behaviour is to show the image as large as possible, and who cares about the command window.

joshuacwnewton: > We can always show full-size images, at the expense of the command window, or we can show the command window, and scale the image to use the space that's left.

I'm a bit ignorant to web technologies, so apologies if this suggestion isn't possible, but... Would it be possible to have our cake and eat it too by displaying the description *and* the full size image (cut off vertically) with the scrollbar being inside the image container (indicated in red), rather than the scrollbar being for the whole page?

I've made a quick (hacky) mockup here:

mockup

The idea being that the user can still view the mosaic 1 row at a time (albeit with a slight height constraint), while also keeping the command on-screen at all times (for screenshots).

elhil: Okay, in a potentially overkill move... I made a button which switches between "fit to height" (current behaviour) and "full size" (proposed behaviour).

@joshuacwnewton you can try it out at the [v0.1.0-rc4](https://github.com/spinalcordtoolbox/report-ui/releases/tag/v0.1.0-rc4) tag!

joshuacwnewton: Following the instructions you helpfully gave in the commit message of https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4827/commits/8dc98c910a438084c0a1fa9613779fd412b9d88a (thanks!), I generated the sagittal mosaic report, and... this is so slick!!

[Screencast from 2025-04-03 13-03-49.webm](https://github.com/user-attachments/assets/f80e065f-9677-4c8a-9756-5a688272fd01)

The only nitpick I have is that at first, the greyed out nature of the "toggle button" design gave me the impression that the "full size" button was disabled at first, since "greyed out" is how our other buttons used to behave when disabled. Also, with the grey being low contrast against the white (plus the lack of outline), I couldn't tell that the two buttons were actually a single toggle button.

image

It was only after clicking (and seeing the sliding animation) that I realized it wasn't disabled. 

I was curious about the UI design of toggle buttons and found this super in-depth case study ([Part 1](https://www.smashingmagazine.com/2022/08/toggle-button-case-study-part-1/), [Part 2](https://www.smashingmagazine.com/2022/08/toggle-button-case-study-part-1/)). tl;dr from [Part 2 results](https://www.smashingmagazine.com/2022/08/toggle-button-case-study-part-2/#best-toggles): 

- Instead of white+black (active) and grey+grey (inactive), it might be better to use white as the background, and use bold text (or color) to denote active status. 
- It also might be good to have a singular black outline around the pair of buttons (matching the 'toggle overlay' button) to denote that the two buttons are linked in state.

joshuacwnewton, line 1: On a related note, I just realized now that the 'Save Fails' and 'Save Artifacts' buttons are enabled always (and clicking them will export an empty `.yml` file with the text "`# No rows matched`"

Is there any way we can have these buttons be greyed out if no QC status is set? (This would preserve the old behavior + make it clearer how the buttons correspond to the QC behavior by having them come to life when the user QCs the subjects.)

elhil: This is much more than a nitpick; that visual discrepancy was bothering me. The [latest rc](https://github.com/spinalcordtoolbox/report-ui/releases/tag/v0.1.0-rc5) implements your suggestion, I think, and I'm much happier with it. Possibly the background on the toggle makes this visually stand out too much for how minor a feature it is?

elhil: [Implemented](https://github.com/spinalcordtoolbox/report-ui/releases/tag/v0.1.0-rc5) :)

valosekj:
Testing the [QC artifact](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14267313041/job/39992186613?pr=4827) from  https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4827/commits/27af050903dd40d481060904215317d987a34523. 

It looks great! 

Small nitpick: when I click on the `Full size` button and then use `Right arrow` to toggle overlay, I'm switching between `Full size` and `Fit to height` instead of toggling overlay. Personally, I would prefer if the `Right arrow` key was always reserved only for toggling overlay (and not for switching between zoom).

Kapture 2025-04-04 at 10 48 39

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: I agree that the grey background stands out a lot! It's a sort of "cool"-hued grey, which stands out among the monochrome nature of the rest of the report.

image

If bold + background is feeling like overkill, I was curious whether just bold would be sufficient. So, I tried replacing `bg-slate-300` with `bg-background`, as well as removing `muted` from `text-muted-foreground`.

image

> Side note: I also fussed a bit with the spacing? The height of the radio group felt slightly smaller than the 'Toggle overlay' button, so I changed the height ever so slightly to `h-[2.55rem]`. Also, it felt like there was just a little too much horizontal whitespace, so I changed the width of each inner label from 36 to 30, and removed `justify-center` (since the center justification felt like it resulted in doubled whitespace between the buttons, especially with the BG removed). 

The only issue with this is that, without a visually distinct background, the sliding animation bits don't really serve a purpose anymore. And in fact, the adaptive text fitting becomes a bit distracting, almost looking like the text is translating up and to the right a bit when bolded.

[Screencast from 2025-04-04 11-22-54.webm](https://github.com/user-attachments/assets/955c0c76-181b-4346-ba7f-a50517a19636)

I'd be curious on your thoughts here! :)

joshuacwnewton:
> Small nitpick: when I click on the `Full size` button and then use `Right arrow` to toggle overlay, I'm switching between `Full size` and `Fit to height` instead of toggling overlay.

Quick add-on to @valosekj's comment: This also happens with the up/down arrow keys as well, toggling the state rather than changing rows.

elhil:
> Small nitpick: when I click on the Full size button and then use Right arrow to toggle overlay, I'm switching between Full size and Fit to height instead of toggling overlay. Personally, I would prefer if the Right arrow key was always reserved only for toggling overlay (and not for switching between zoom).

Yeah, from a usability/accessibility perspective, intercepting the right arrow key globally was already a bit of a hack. It would add even more hackiness to start overriding browser behaviours for switching inputs, focus, etc. I think a better course of action would be to give a hotkey you could use _while scrolling the table_ to toggle the fit mode. If you click on the button, it's expected that to use the table again you'll click back on the table, but I want to give ergonomic ways of interacting as well.

We already expect a user to have a finger on the F key. What about D for toggle fit mode?

spinalcordtoolbox/reports/assets/index.html
elhil: Okay, I think you're right about the colour thing, so I did away with all the slidey parts (I didn't build those anyway, don't worry). I think I hit a middle ground with your adjustments and still being able to tell that it's a toggle button if you haven't used it before, you can try it out at [b8bd2e9](https://github.com/spinalcordtoolbox/report-ui/commit/b8bd2e909ee05f8cfb94e42cc59eb1a88dba0ff9)

valosekj:
> We already expect a user to have a finger on the F key. What about D for toggle fit mode?

Good idea! This could work!

spinalcordtoolbox/reports/assets/index.html
joshuacwnewton: Marvelous. No notes. :100:

joshuacwnewton:
> We already expect a user to have a finger on the F key. What about D for toggle fit mode?

Given that the right arrow key is used for toggling the overlay, and the left arrow key is unused, is there any merit to using that key instead? (Since it keeps the idea of "left hand on `F` for managing the table, right hand on arrow keys for managing what's displayed in the viewer").

elhil:
@joshuacwnewton @valosekj you can test the keyboard shortcut version [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14312553790/artifacts/2895837965). It's using the D key but it's trivial to change it to another.

> Given that the right arrow key is used for toggling the overlay, and the left arrow key is unused, is there any merit to using that key instead? (Since it keeps the idea of "left hand on `F` for managing the table, right hand on arrow keys for managing what's displayed in the viewer").

My only concern is that this could add to the confusion when the fit mode toggle is focused, and left arrow would always select "fit to height" rather than toggling. Otherwise it may more intuitive from a UX perspective.

joshuacwnewton:
The 'D' key works perfectly! I see the upsides of that key being used too, as it allows the user to scroll the image using a right-handed mouse wheel/2 figure gesture, and still toggle the view with their left hand. :)

----

The only other feedback I have is that the "Legend" is now wrapping, so maybe we want to insert a newline ourselves to group related buttons?

Before:

image

After (quick mockup):

legend

valosekj:
Thanks! `D` works fine! 
But for some reason, the `Rank` (`0` to `9` keys) does not work on my laptop keyboard now. When I try to type numbers into the "Search" box, they are written fine.

elhil:
> Thanks! `D` works fine! But for some reason, the `Rank` (`0` to `9` keys) does not work on my laptop keyboard now. When I try to type numbers into the "Search" box, they are written fine.

I can't reproduce this unfortunately. Is a row selected when you're typing the numbers?

joshuacwnewton:
I can reproduce the behavior that @valosekj is describing. Some details:

- I have a row selected.
- I am specifically using the number row on a laptop keyboard (and not a numpad).
- I am using Firefox 136.0.3 on Ubuntu 24.04.2.

elhil:
Upcoming RC will allow us to debug keypresses :)

elhil:
@valosekj in [this version](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14406808316/artifacts/2928782500) could you add `?debug_keys=1` at the end of the URL, open the developer console, then interact with the table and press some number keys? Debug information should show up in the console.

valosekj:
> @valosekj Are you manually enabling the 'Rank' column?
>
> (Personally, just now I made a very silly error where I was looking for changes in the default QC column ...

Oh, yes, sorry for this. I was looking at the QC column 🤦‍♂️ and I had the `Rank` column hidden.

Everything works fine now! Thank you for these improvements! 💯

valosekj (approved):

joshuacwnewton:
Splendid! This should be good to merge then, I think? (I'm currently about to merge https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4622, then I'll update this PR against master and merge it too.)

elhil:
> Splendid! This should be good to merge then, I think? (I'm currently about to merge #4622, then I'll update this PR against master and merge it too.)

Seems ready to me!

joshuacwnewton:
This is a very momentous PR! :hearts: 

Thank you so much for taking the time to author this rewrite @elhil, and I look forward to working with the `report-ui` codebase in the future for new features and fixes! :tada:

PR #4828, opened 2025-03-26
Amend argparse help of transform tool to make `-d` and `-o` arguments clearer

joshuacwnewton:
## Description

Direct response to https://github.com/ivadomed/ms-multi-spine-challenge-2024/pull/13#discussion_r2013788948.

## Linked issues

N/A

### DIFF
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -40,7 +40,7 @@ def get_parser():
     )
     mandatory.add_argument(
         "-d",
-        help='Destination image. Example: `out.nii.gz`',
+        help='Destination image. Example: `dest.nii.gz`',
         metavar=Metavar.file,
     )
     mandatory.add_argument(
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -68,7 +69,7 @@ def get_parser():
         choices=(0, 1, 2))
     optional.add_argument(
         "-o",
-        help='Registered source. Example: `dest.nii.gz`',
+        help='Registered source. Example: `out.nii.gz`',
```
valosekj, line 72: maybe something like this might be more informative 

```suggestion
        help='Output image (i.e., input image registered to the output image). Example: `out.nii.gz`',
```

joshuacwnewton: Yes, absolutely! We could even just say "filename to use for the output image" since it sounds like people should be providing an image when it's just a filename?

valosekj: or even "absolute path"?

joshuacwnewton: Does [4aeaa19](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4828/commits/4aeaa195b1c6f289b1f9d0a7014b3d2e5394ff1b) sound reasonable?

valosekj: Yes, thank you!

plbenveniste (approved):
Much clearer from my point of view! 👍

Issue #4674, opened 2024-11-02
ModuleNotFoundError: No module named 'IPython' when running within jupyter notebook


jcohenadad:
When running SCT via a jupyter notebook, the QC report sometimes fail, throwing an "IPython missing" message:

<details><summary>Terminal output</summary>

```console
--
Spinal Cord Toolbox (git-master-a458fa37fd22179dd2a6c13d33b7d191552abee7)

sct_qc -i sub-CRMBM1_UNIT1.nii.gz -s sub-CRMBM1_UNIT1_seg_labeled.nii.gz -p sct_label_vertebrae -qc /Users/julien/code/coil-qc-code/data-human/qc -qc-subject sub-CRMBM1
--


*** Generating Quality Control (QC) html report ***
Traceback (most recent call last):
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_qc.py", line 125, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_qc.py", line 108, in main
    generate_qc(fname_in1=arguments.i,
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 632, in generate_qc
    qc_image._make_QC_image_for_3d_volumes(img, mask, plane, imgs_to_generate)
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/reports/qc.py", line 228, in _make_QC_image_for_3d_volumes
    fig = mpl_plt.figure()
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/pyplot.py", line 1027, in figure
    manager = new_figure_manager(
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/pyplot.py", line 549, in new_figure_manager
    _warn_if_gui_out_of_main_thread()
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/pyplot.py", line 526, in _warn_if_gui_out_of_main_thread
    canvas_class = cast(type[FigureCanvasBase], _get_backend_mod().FigureCanvas)
[...]

PR #4838, opened 2025-03-28
Replace `pyplot` usage in `qc.py` to address crashing in Jupyter notebooks

SomeoneInParticular:
#### GitHub

#### PR contents

- [x] ~~I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution~~ See "Other Remarks"
- [x] ~~I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages~~ See "Other Remarks"

## Description

`sct_qc` would crash if called from within a Jupyter Notebook, but QC files would be generated just fine if they were made at the same time as the script (i.e. `sct_deepseg spinalcord ... -qc qc_out`). This should fix the former, making it behave more like the latter.

### Other Remarks

Right now, we have neither tests nor documentation on how SCT runs with Jupyter. Is this worth pursuing, or should be just place it under a "caveat emptor" and squash related bugs as they arise? See #4839 for further discussion.

As well, a whole QC unification should be considered, to ensure bugs like this are less prominent and easier to resolve. Some amazing effort to that end has already been done by @elhil and @joshuacwnewton, but much remains to be done. This would require a substantial rewrite of the documentation as well.

## Linked issues

First step to fixing issue #4674


spinalcordtoolbox/reports/qc.py
```diff
@@ -225,7 +225,7 @@ def _make_QC_image_for_3d_volumes(self, img, mask, plane, imgs_to_generate):
                 * float(self.aspect_img))
         ]
 
-        fig = mpl_plt.figure()
+        fig = mpl_figure.Figure()
```
joshuacwnewton, line 228: Just to summarize the change (so that I can be sure I fully understand it):

- **Before:** [`matplotlib.pyplot.figure()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html)
- **After:** [`matplotlib.figure.Figure()`](https://matplotlib.org/3.3.4/api/_as_gen/matplotlib.figure.Figure.html)

This sounded familiar, and it turns out we have a old (2020) SCT Developer Wiki page on this called "[Moving away from pyplot](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Moving-away-from-pyplot)" that says:

> Matplotlib has two interfaces: pyplot (based on MATLAB) or object-oriented (OO) interface. Pyplot inteface is causing problems with backend on some headless stations. The purpose of this wiki is to explain how to refactor the code to a all OO interface. With this, figures are not displayed to the frontend, but saved. In this case, we utilize an instance of axes.Axes in order to render visualizations on an instance of figure.Figure.

Our Wiki page isn't clear on what "causing problems" means, but it seems that the "not displayed to the frontend" part lines up exactly with what you wrote in the [underlying issue](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4674#issuecomment-2762337603):

> * The `figure` function of `matplotlib.pyplot` checks against the back-end to see whether the figures should be drawn interactively (specifically [here](https://github.com/matplotlib/matplotlib/blob/c38f70a57277f661c7426c7d464c07c5ad0c2318/lib/matplotlib/pyplot.py#L1061))
> * In contrast, requesting a `Figure` instance directly does _not_ check against the back-end, thus side-stepping the problem entirely.

So, it seems like we should proceed with moving away from `pyplot` to avoid crashes on setups where we don't intend to interactively render plots (headless machines, Jupyter notebooks, etc.)

joshuacwnewton: It looks like there's actually only a single remaining instance of `pyplot` in this file (for a colorbar)?

SomeoneInParticular: > So, it seems like we should proceed with moving away from pyplot to avoid crashes on setups where we don't intend to interactively render plots (headless machines, Jupyter notebooks, etc.)

Agreed, just didn't want to any changes that weren't strictly required to fix the bug until it had been discussed. Will remove any remaining `pyplot` calls in the next push.

joshuacwnewton: I think it might still be required? 

Perhaps the `colorbar` issue hadn't come up because the Jupyter notebook QC report that was tested in the issue didn't use the colorbar. But, I have a hunch that if we try the tSNR QC (which I think uses the colorbar), it will cause a similar crash. :sweat_smile:

SomeoneInParticular: Yeah, I'll dig into it some more. 

Frankly, though, I think it might be best that we just "ignore" this for the time being, focusing instead on moving everything over to the newer `qc2.py` functions. We already have some de-synchronization between the two.  

`lesion_sci_t2` is an excellent example of this; implict QC generation within the `deepseg` call (which uses `qc2`) will plot both the segmentation and lesions, while post-hoc qc using `sct_qc` (which uses `qc`) will only plot the lesions (even though we give it a segmentation file).

SomeoneInParticular: > lesion_sci_t2 is an excellent example of this; implict QC generation within the deepseg call (which uses qc2) will plot both the segmentation and lesions, while post-hoc qc using sct_qc (which uses qc) will only plot the lesions (even though we give it a segmentation file; far as I can tell it just uses it for cropping).

An example:
example

SomeoneInParticular: > My apologies, but has this patch been pushed? I'm only seeing master merge commits over the past few days.

Not yet, got side-tracked with some other work. I'm just finalizing some testing on it, should have a push ready momentarily.

SomeoneInParticular: Just pushed the patch (and some other minor fixes I noticed when implementing it). Still believe its worth moving everything over to `qc2.py` long term, but this should hold things stable for the 7.0 release.

joshuacwnewton: Absolutely agreed. Thank you so much!

joshuacwnewton (commented):
I ran the test `test_sct_sct_fmri_compute_tsnr_with_mask` locally (which includes QC), and the colorbar looks to be generated just fine! :tada: 

image

joshuacwnewton:
Note that I've emergency hotfixed the failing test in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/434690e65c976d71c7a5ee2f495a9461985ed393, so syncing `master` should fix the issue.

Issue #4795, opened 2025-02-13
Benchmark inference times for DeepSeg models (CPU vs. GPU)


joshuacwnewton:
During our last SCT meeting (Jan 29th, 2025) we discussed the inference times of `sct_deepseg` models, and how for some models, GPU helps a lot, while for others, GPU barely speeds up the model at all. 

So, @valosekj had the idea to document some sample times in the [**Inference Time**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/DL-models%3A-Inference-time) SCT Wiki page. This will help students see whether it's worth the time to set up a GPU installation of SCT, and also judge whether their models are too big/slow. Ideally GPUs wouldn't be needed for inference at all, as many of our users are limited to CPU-only.

This issue is for adding additional data points for other tasks to the comparison table.

SomeoneInParticular:
To be discussed at next meeting (March 13th, 2024); if the plan I propose is acceptable, will assign myself to this and proceed with developing these benchmarks.

SomeoneInParticular:
After the aforementioned meeting, I will wait on implementing this until the rest of the lab has a chance to propose any other aspects of a standard SCT run they want bench-marked. Specifically, until the sub-issue #4801 is done. Cheers!

SomeoneInParticular:
Update on this front; the repository [here](https://github.com/SomeoneInParticular/sct_timings) has been made specifically to address this. I'm just adding the final touches (namely CI, plot generation, and moving it under the SCT organization), then this issues should be resolved!

PR #4844, opened 2025-04-03
Add detailed time and memory profiling for CLI scripts

SomeoneInParticular:
#### GitHub

#### PR contents

## Description

As discussed in #4795, we've been meaning to add methods of empirically evaluated SCT's runtime performance for a while. This PR adds two new faculties to accomplish this implicitly:

### Time-Based Profiling

When activated through the command-line (by adding the `-profile-time` argument), the total runtime inside each call made in the program is tracked. Once the program terminates, the total results are placed in an output file for later review. By default, this file is named `time_profiler_results.txt`, but the user can provide an alternative path after the argument if they wish to save it elsewhere.

### Memory-Based Tracing

When activated through the command-line (by adding the `-trace-memory` argument), memory tracing functionality is activated. This does two things by default:

1. The ability to "snapshot" the memory state of the program becomes available via the `profiling.snapshot_memory` function. As the name implies, calling this function will snapshot the state of memory at that moment, and save it to a results file. 
1. As the program runs, the peak memory use (in bits) is tracked, and upon program termination saved to a the same file for reference.

Similarly to the `-profile-time` function, the output file will be saved to `memory_trace_results.txt` unless otherwise specified by the user.

### Other Notes:

The `snapshot_memory` call needs to be made explicitly within the code-base, **by the programmer**, either during debugging or assessment. To account for this, it acts similar to a "debug" logging call; if memory tracing is not active, the call is ignored entirely (with a negligible effect on runtime).

Both profilers _will_ impact the runtime and memory traces slightly in and of themselves, though this impact is very minimal. Nevertheless, this should be kept in mind when making conclusions about the program's efficiency! 

## Linked issues

Step towards resolving #4795

### DIFF
--- a/spinalcordtoolbox/utils/profiling.py
+++ b/spinalcordtoolbox/utils/profiling.py
@@ -153,25 +153,29 @@ class RepeatCallTimer(threading.Timer):
 
 class MemoryTracingManager:
 
-    default_output = Path('memory_tracer_results.txt')
+    time_file = Path('memory_across_time.tsv')
+    snapshot_file = Path('memory_snapshots.txt')
     default_interval = .1
 
-    def __init__(self, out_file=None, interval=default_interval):
+    def __init__(self, out_path=Path('.'), interval=default_interval):
         # Initialize and enable the profiler
         tracemalloc.start()
 
-        # Track the output file, if it was provided
-        if out_file is not None:
-            self._output_file = out_file
-        else:
-            self._output_file = self.default_output
+        # Generate the requested output directory
+        out_path.mkdir(parents=True, exist_ok=True)
 
-        # Reset the file if it already exists
-        if out_file.exists():
-            out_file.unlink()
+        # Track the output file destinations
+        self._timed_outputs = out_path / self.time_file
+        self._snapshot_outputs = out_path / self.snapshot_file
 
-        # Open a file stream, rather than cacheing, to reduce the impact the memory profiling will have on memory
-        self._open_output = open(out_file, 'w')
+        # Reset the files if they already exist
+        if self._timed_outputs.exists():
+            self._timed_outputs.unlink()
+        if self._snapshot_outputs.exists():
+            self._snapshot_outputs.unlink()
+
+        # Open a file stream for timings, rather than cacheing, to reduce the impact the profiling will have on memory
+        self._open_output = open(self._timed_outputs, 'x')
 
         # Write a header for easy of use
         self._open_output.write("Time (s)\tMemory (KiB)\n")
@@ -199,18 +203,23 @@ class MemoryTracingManager:
         self._mem_timer.cancel()
         self._open_output.close()
 
-    def snapshot_memory(self, label: str):
+    def snapshot_memory(self, caller) -> Path:
         # Snapshot the current memory state
         mem_snapshot = tracemalloc.take_snapshot()
         mem_stats = mem_snapshot.statistics('lineno', cumulative=True)
 
         # Generate the header for this output
-        header = f"=== {label} ===\n"
+        header = f"=== {caller.function} ({Path(caller.filename).name}, line {caller.lineno}) ===\n"
 
-        # Save the stats to file with this header
-        self._open_output.write(header)
-        self._open_output.writelines([str(x) + '\n' for x in mem_stats])
-        self._open_output.write('\n')
+        # Creat the initial file if it doesn't already exist
+        if not self._snapshot_outputs.exists():
+            self._snapshot_outputs.touch()
+
+        # Save the snapshot
+        with open(self._snapshot_outputs, 'a') as fp:
+            fp.write(header)
+            fp.writelines([str(x) + '\n' for x in mem_stats])
+            fp.write('\n')
 
     def _finish_tracing(self):
         # End the non-blocking timer to prevent any race conditions
@@ -229,7 +238,7 @@ class MemoryTracingManager:
         self._open_output.close()
 
         # Report that the file was written, and where to
-        logging.info(f"Saved memory tracing results to '{self._output_file.resolve()}'.")
+        logging.info(f"Saved memory tracing results to '{self._timed_outputs.resolve()}'.")
 
 
 def begin_tracing_memory(out_path: Path = None):
@@ -254,11 +263,10 @@ def snapshot_memory():
 
     # Get the calling function
     caller = inspect.stack()[1]
-    calling_id = f"{caller.function} ({Path(caller.filename).name}, line {caller.lineno})"
 
-    # Snapshot the current memory state, saving it to file immediately
+    # Snapshot the current memory state, saving it to file immediately and
     # noinspection PyUnresolvedReferences
-    MEMORY_TRACER.snapshot_memory(calling_id)
+    MEMORY_TRACER.snapshot_memory(caller)
 
 
 class MemoryTracingAction(Action):
@@ -275,7 +283,7 @@ class MemoryTracingAction(Action):
 
         # If no value as provided, set the output path to be in the current directory
         if values is None:
-            out_path = MemoryTracingManager.default_output
+            out_path = MemoryTracingManager.time_file
         else:
             # Otherwise, try to get the path associated with this parameter, and confirm its root exists
             out_path = Path(values)
@@ -283,7 +291,7 @@ class MemoryTracingAction(Action):
         # Check to see if the file exists and is a directory
         if out_path.exists() and out_path.is_dir():
             # If so, set the result to be saved to our default file output
-            out_path /= MemoryTracingManager.default_output
+            out_path /= MemoryTracingManager.time_file
 
         # Finally, initiate the memory tracer, designating it's output file to be the user specified one
         begin_tracing_memory(out_path)
--- a/spinalcordtoolbox/utils/profiling.py
+++ b/spinalcordtoolbox/utils/profiling.py
@@ -1,5 +1,6 @@
 import atexit
 import cProfile
+import inspect
 import io
 import logging
 import pstats
@@ -144,11 +145,18 @@ class MemoryTracingManager:
     def __init__(self, out_file=None):
         # Initialize and enable the profiler
         tracemalloc.start()
+
         # Track the output file, if it was provided
         if out_file is not None:
             self._output_file = out_file
         else:
             self._output_file = self.default_output
+
+        # Reset the file if it already exists
+        if out_file.exists():
+            out_file.unlink()
+            out_file.touch()
+
         # Ensure the profiler completes its runtime at program exit
         atexit.register(self._finish_tracing)
 
@@ -156,6 +164,21 @@ class MemoryTracingManager:
         # If the profiler is deleted explicitly, just clean up without reporting anything
         atexit.unregister(self._finish_tracing)
         tracemalloc.stop()
+
+    def snapshot_memory(self, label: str):
+        # Snapshot the current memory state
+        mem_snapshot = tracemalloc.take_snapshot()
+        mem_stats = mem_snapshot.statistics('lineno', cumulative=True)
+
+        # Generate the header for this output
+        header = f"=== {label} ===\n"
+
+        # Save the stats to file with this header
+        with open(self._output_file, 'a') as fp:
+            fp.write(header)
+            fp.writelines([str(x) + '\n' for x in mem_stats])
+            fp.write('\n')
+
 
     def _finish_tracing(self):
         # Get the peak memory consumed during the program, and save it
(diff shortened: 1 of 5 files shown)


spinalcordtoolbox/utils/profiling.py
```diff
@@ -24,13 +38,216 @@ def stop(self):
...
+
+    def _finish_profiling(self):
+        # Finish profiling
+        self._profiler.disable()
+
+        # Grab the stats, and organize them from most to least time-consuming (cumulatively)
+        io_stream = io.StringIO()
+        profiling_stats = pstats.Stats(self._profiler, stream=io_stream)
+        profiling_stats = profiling_stats.sort_stats(pstats.SortKey.CUMULATIVE)
+        profiling_stats.print_stats()
```
joshuacwnewton, line 82: It would be really neat if `-profile-time` could be used alongside a profiling visualizer such as `snakeviz` or `tuna`. 

Right now, `-profile-time` outputs a human-readable log, whereas the aforementioned utilities expect a `.prof` file in a binary format. Thankfully, gnerating the `.prof` binary format is simple enough here, since we can use the `dump_stats()` method of the existing `_profiler` object. 

Maybe we could continue to dump the `.txt` output by default, but also allow users to opt-in to the `.prof` format by checking the extension of the requested output file, and output different formats depending on whether the file extension is `.txt` vs `.prof`?

spinalcordtoolbox/utils/profiling.py
```diff
@@ -24,13 +38,216 @@ def stop(self):
...
+    def _finish_tracing(self):
+        # Get the peak memory consumed during the program, and save it
+        _, traced_peak = tracemalloc.get_traced_memory()
+
+        # Finish profiling after sampler (as otherwise the peak is reset to 0)
+        tracemalloc.stop()
+
+        # Save the peak results to the output file
+        with open(self._output_file, 'a') as fp:
+            fp.write(f"PEAK MEMORY USE; {traced_peak / 1024:.3f} KiB")
```
joshuacwnewton, line 191: Likewise, it would be neat if we could have more in-depth memory usage than just the peak?

I tried to dig into "tracemalloc visualization", expecting there to be something akin to Task Manager/System Monitor's memory vs. time graphs:

image

But I didn't get any immediate results like I expected (re: snakeviz/tuna/etc.).

joshuacwnewton: I did come across some alternatives to `tracemalloc`, however?

[`Fil`](https://pythonspeed.com/articles/memory-profiler-data-scientists/) in particular seems aimed at our use-case, and maybe is worth exploring. It creates visualizations (flame graphs) but also automatically does so at peak memory usage, letting us skip the trial and error of manually adding snapshot calls in our code.

joshuacwnewton (commented):
So far everything works as described without much issue!

My main consideration is re: how useful the outputs are as-is, vs. whether we can potentially make them more useful through visualizations.

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: Sounds good, will make a demo commit implementing this (and a corresponding test) in a bit.

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: I agree that a better visualization would be useful, though I have a few concerns with implementing `Fil` (and similar profilers) natively. Specifically:

* It's another dependency we would need to add to our tool that the end user wouldn't actually be using 99% of the time, making testing for future Python versions more tricky (I also feel similarly about PyTest, incidentally).
  * In fact, installing `Fil` just to give it a spin required setting up a custom 'testing' `mamba` environment, and that was just to ensure it worked alongside `flake8` and `pytest`.
  * We could set up an "alternate" install for this, similar to how we do for GPU-enabled SCT, but this still adds maintenance overhead.
* Users who want/need this can just install the profiler in the `venv` themselves and run it; I did this with `memray` (solely out of preference, this works with `Fil` too) and it worked just peachy:

Screenshot from 2025-04-04 14-28-53

That all being said, there _might_ be a way to have it evaluate the memory use periodically to give us that "Task Manager" like view, but I need to do more digging before I can say for sure.

joshuacwnewton: Thank you for sharing thoughts on `Fil`/`memray`! I hadn't actually tried `Fil` nor had I heard of `memray`, so I appreciate your insight here. :) 

> Users who want/need this can just install the profiler in the venv themselves and run it;

I tried out `Fil` just now, only to realize that its output was complicated by the fact that we use a launcher file which then calls our scripts with a subprocess (for historical reasons I don't entirely remember off the top of my head). Because `Fil` can't see into subprocesses, the output of `Fil` only covered the `launcher.py` scope, which is useless to us:

image

> ### [How Fil works](https://pythonspeed.com/fil/docs/how-it-works.html#how-fil-works)
> 
> Fil uses the LD_PRELOAD/DYLD_INSERT_LIBRARIES mechanism to preload a shared library at process startup. This is why Fil can’t be used as regular library and needs to be started in a special way: it requires setting up the correct environment before Python starts.

Looking at your screenshot of `memray`, I'm curious if the same restriction applies. 354 KiB is suspiciously low for a deepseg call (and looks similar to my `Fil` attempt). 

I'm tempted to open a separate issue for this, since as long as we have the `launcher.py` + `subprocess` pattern, these tools will be useless to us.

joshuacwnewton: > That all being said, there might be a way to have it evaluate the memory use periodically to give us that "Task Manager" like view, but I need to do more digging before I can say for sure.

I think it's probably fine to just stick with peak usage for now, then follow-up with fancier output if we find that peak usage isn't sufficient? I wouldn't want to delay a perfectly sensible PR. :)

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: Should be resolved with [edf3415](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/edf34153d27d67af40f3e1a21609d4d4bccff98c)

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: > Looking at your screenshot of memray, I'm curious if the same restriction applies. 354 KiB is suspiciously low for a deepseg call (and looks similar to my 0.4 MiB Fil attempt).

Good catch! I re-ran the `memray` trace to check, and you're correct; it it not catching the sub-processes content, and `memray` explicitly states as much as well: https://github.com/bloomberg/memray/discussions/157.

I will just focus on implementing the tracking-over-time for now, but we definitely should dig more into this after 7.0...

SomeoneInParticular: Added the memory-over-time tracing with [97e2a1e](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/97e2a1e70b1e18fe45d1137e9be0c4a2837b858f)

SomeoneInParticular: Further refined with [d535a5e](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/d535a5e3f25bfb65022ee2964492ee0d28916bee)

spinalcordtoolbox/utils/shell.py
```diff
@@ -298,13 +299,38 @@ def add_common_args(self, arg_group=None):
             help="Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode."
         )
 
+        # Add a flag which allows time-based profiling the profiler
+        arg_group.add_argument(
+            '-profile-time',
+            nargs='?',
+            metavar=Metavar.file,
```
joshuacwnewton, line 306: I noticed that if the path provided contains a non-existent folder (e.g. `-profile-time test/test.prof`), it won't be created and instead will fail:

```
Error in atexit._run_exitfuncs:
  File "/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/utils/profiling.py", line 84, in _finish_profiling
    profiling_stats.dump_stats(self._output_file)
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/pstats.py", line 196, in dump_stats
    with open(filename, 'wb') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'test/test.prof'
```

joshuacwnewton (commented):
I don't really have many comments on the time profiler, since it's very simple! It's really just the input validation.

I haven't yet looked deeply at the memory profiling, largely because of its existing hanging issues.

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: Yeah, I think I'll excise it for now, and we can wrap back to this once I have more time to dig into resolving this hang post-7.0

spinalcordtoolbox/utils/profiling.py
```diff
@@ -24,13 +39,261 @@ def stop(self):
...
+
+    def __init__(self, out_path=Path('.'), interval=default_interval):
+        # Initialize and enable the profiler
+        tracemalloc.start()
+
+        # Generate the requested output directory
+        out_path.mkdir(parents=True, exist_ok=True)
+
+        # Track the output file destinations
+        self._timed_outputs = out_path / self.time_file
```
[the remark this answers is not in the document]
SomeoneInParticular: You're absolutely correct; this should be fixed with [4194be9](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/4194be928f98bcc1a9e0b49b6094491c912ca89f)

spinalcordtoolbox/utils/shell.py
SomeoneInParticular: I sidestepped the error entirely by just having the profile generate the directory path if it didn't already exist. This was already what was being done within the memory profiler, must have slipped my mind when I did it for the time profiler. See [3655335](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/36553350791f925f06a80ac4b840cb003e1df353)

spinalcordtoolbox/utils/profiling.py
```diff
@@ -24,13 +38,228 @@ def stop(self):
...
+        begin_profiling_time(out_path)
+
+        setattr(namespace, self.dest, out_path)
+
+
+# == Memory-based Profiling == #
+class MemoryTracingManager:
+
+    time_file = Path('memory_across_time.txt')
+    snapshot_file = Path('memory_snapshots.txt')
```
joshuacwnewton, line 152: Do we want to add these filenames (as well as `time_profiler_results.txt` and perhaps any `.prof` files) to SCT's `.gitignore`?

spinalcordtoolbox/utils/profiling.py
```diff
@@ -24,13 +38,228 @@ def stop(self):
...
+
+        # If no value as provided, set the output path to be in the current directory
+        if values is None:
+            out_path = Path('.')
+        else:
+            # Otherwise, try to get the path associated with this parameter, and confirm its root exists
+            out_path = Path(values)
+
+        # Finally, initiate the memory tracer, designating it's output file to be the user specified one
+        begin_tracing_memory(out_path)
```
joshuacwnewton, line 263: I'm still noticing a bit of bugginess with how the `out_path` is processed for `-trace-memory`. For example, specifying `-trace-memory test/test.txt` will result in the following output:

> `Saved memory tracing results to '/home/joshua/repos/spinalcordtoolbox/data/sct_testing_data/t2/test/test.txt/memory_across_time.txt'.`

This behavior only seems to happen with the memory tracer, and not the time profiler.

joshuacwnewton (commented):
Only had a few small comments, but the majority of this PR LGTM!

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: Good point, resolved with [cd3ea22](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/cd3ea229832530f60a9a2779c4409d7a0e7ce6fa)

spinalcordtoolbox/utils/profiling.py
SomeoneInParticular: Yes, that's because the time profiler outputs a single file (which the user can specify the path and name of), whereas the memory profiler produces two files (and thus the user can only specify the directory the files can be placed into). I don't mind unifying this functionality if its confusing, or stacking the file outputs of the memory profiler into a single file, but there was method to this madness.

SomeoneInParticular: Update: I revised the output message of the Memory Profiler to be much clearer on what its doing and where (see [9b4d9bc](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/9b4d9bcde40d1a8868f88b9eec17deec7da80eee)). Let me know your thoughts on it!

SomeoneInParticular: Good idea; extended this format to the TimeProfiler as well as part of commit [ef05deb](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4844/commits/ef05deb0eccecb190fa5157bc30d61ec6f676c06)

Issue #4412, opened 2024-03-22
Upload `batch_processing.sh` CSV files as artifacts during CI run


joshuacwnewton:
This is a very small improvement to the `batch_processing.sh` CI: Upload the CSV files it generates as artifacts.

This would make updating the cached results a lot easier, since we wouldn't have to run the entire `batch_processing.sh` script locally, which can take 15-30 minutes.

There is already an example of uploading artifacts in the `batch_processing.sh` workflow file, so it should be easy enough to duplicate the step and change some names.

PR #4850, opened 2025-04-08
Fix slice indexing for `-angle-corr-centerline`

### LINKED ISSUE #4807: Different input images (`-i`) result in different `angle-corr` angles even when using the same `-angle-corr-centerline` image
## Description

It was pointed out by a user in https://forum.spinalcordmri.org/t/calculate-csa-of-the-spinal-canal-and-cord/1406/5.

TL;DR: the user wants to compute cord and canal CSA using `sct_process_segmentation`. To make sure the same angle correction is applied to both segmentations, the user correctly uses the `-angle-corr-centerline` arg:

```bash
sct_process_segmentation -i canal.nii.gz -angle-corr-centerline canal.nii.gz -vertfile canal_labeled.nii.gz -perslice 1 -o canal.csv
sct_process_segmentation -i cord.nii.gz -angle-corr-centerline canal.nii.gz -vertfile canal_labeled.nii.gz -perslice 1 -o cord.csv
```

However, `MEAN(angle_AP)` and `MEAN(angle_RL)` differ between the cord and canal (see the second graph in [the user's post](https://forum.spinalcordmri.org/t/calculate-csa-of-the-spinal-canal-and-cord/1406/5)).

(the user shared the files with me by email; I can forward them)

## Investigation

I put a breakpoint to L101 in `process_seg.py` to compare `angle_AP_rad` between the cord and canal segmentations:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7af12e107e14e5a4bc5b2dd7b09d2d87cc448a30/spinalcordtoolbox/process_seg.py#L98-L101

And there is indeed a difference in the angle between the cord and canal masks! This arises from a difference in the `tangent_vect`, which is calculated using a different `min_z_index`. The canal's `min_z_index=178`, while the cord's `min_z_index=501` (this is expected because the canal segmentation covers more slices than the cord segmentation).

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7af12e107e14e5a4bc5b2dd7b09d2d87cc448a30/spinalcordtoolbox/process_seg.py#L92-L94

Details (open the screenshots in two separate tabs and toggle between them to see the differences):

<details>
<summary>canal</summary>

Image

</details>

<details>
<summary>cord</summary>

Image

</details>

mguaypaq:
Fixes #4807.

If the input mask supplied in `-i` and the custom centerline supplied in `-angle-corr-centerline` did not start at the same lowest SI slice, then the wrong slice index was used for angle correction.

With the proposed code, everything is indexed directly with SI slice numbers, rather than trying to do relative indexing. A quick manual test using the data from #4807 shows that this fixes the problem: for the slices which appear in both CSVs, they have the same `angle_AP` and `angle_RL`.

If the `-angle-corr-centerline` is too short, there's now an error message to explain the situation:

image

### DIFF
--- a/spinalcordtoolbox/process_seg.py
+++ b/spinalcordtoolbox/process_seg.py
@@ -15,6 +15,7 @@ from spinalcordtoolbox.image import Image
 from spinalcordtoolbox.aggregate_slicewise import Metric
 from spinalcordtoolbox.centerline.core import get_centerline
 from spinalcordtoolbox.resampling import resample_nib
+from spinalcordtoolbox.utils.shell import parse_num_list_inv
 from spinalcordtoolbox.utils.sys import sct_progress_bar
 
 # NB: We use a threshold to check if an array is empty, instead of checking if it's exactly 0. This is because
@@ -84,20 +85,24 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
         # columns of `arr_ctl_der` contain the x and y components of the centerline derivative
         deriv = {int(z_ref): arr_ctl_der[:2, index] for index, z_ref in enumerate(arr_ctl[2])}
 
+        # check for slices in the input mask not covered by the centerline
+        missing_slices = sorted(set(range(min_z_index, max_z_index + 1)).difference(deriv.keys()))
+        if missing_slices:
+            raise ValueError(
+                "The provided angle correction centerline does not cover slice(s) "
+                f"{parse_num_list_inv(missing_slices)} of the input mask. Please "
+                "supply a more extensive '-angle-corr-centerline', or disable angle "
+                "correction ('-angle-corr 0')."
+            ) from None
+
     # Loop across z and compute shape analysis
     for iz in sct_progress_bar(range(min_z_index, max_z_index + 1), unit='iter', unit_scale=False, desc="Compute shape analysis",
                                ncols=80):
         # Extract 2D patch
         current_patch = im_segr.data[:, :, iz]
         if angle_correction:
-            try:
-                dx, dy = deriv[iz]
-            except KeyError:
-                raise ValueError(f"The provided angle correction centerline does not cover slice {iz} of "
-                                 "the input mask. Please supply a more extensive '-angle-corr-centerline', "
-                                 "or disable angle correction ('-angle-corr 0').") from None
             # Extract tangent vector to the centerline (i.e. its derivative)
-            tangent_vect = np.array([dx * px, dy * py, pz])
+            tangent_vect = np.array([deriv[iz][0] * px, deriv[iz][1] * py, pz])
             # Compute the angle about AP axis between the centerline and the normal vector to the slice
             angle_AP_rad = math.atan2(tangent_vect[0], tangent_vect[2])
             # Compute the angle about RL axis between the centerline and the normal vector to the slice
--- a/spinalcordtoolbox/process_seg.py
+++ b/spinalcordtoolbox/process_seg.py
@@ -94,8 +94,8 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
                 dx, dy = deriv[iz]
             except KeyError:
                 raise ValueError(f"The provided angle correction centerline does not cover slice {iz} of "
-                                 "the input mask. Please supply a more extensive -angle-corr-centerline, "
-                                 "or set -angle-corr to 0.") from None
+                                 "the input mask. Please supply a more extensive '-angle-corr-centerline', "
+                                 "or disable angle correction ('-angle-corr 0').") from None
             # Extract tangent vector to the centerline (i.e. its derivative)
             tangent_vect = np.array([dx * px, dy * py, pz])
             # Compute the angle about AP axis between the centerline and the normal vector to the slice
--- a/spinalcordtoolbox/process_seg.py
+++ b/spinalcordtoolbox/process_seg.py
@@ -101,13 +101,9 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
             # Normalize vector by its L2 norm
             tangent_vect = tangent_vect / np.linalg.norm(tangent_vect)
             # Compute the angle about AP axis between the centerline and the normal vector to the slice
-            v0 = [tangent_vect[0], tangent_vect[2]]
-            v1 = [0, 1]
-            angle_AP_rad = math.atan2(np.linalg.det([v0, v1]), np.dot(v0, v1))
+            angle_AP_rad = math.atan2(tangent_vect[0], tangent_vect[2])
             # Compute the angle about RL axis between the centerline and the normal vector to the slice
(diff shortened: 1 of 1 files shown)


spinalcordtoolbox/process_seg.py
```diff
@@ -65,7 +65,7 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
     min_z_index, max_z_index = min(Z), max(Z)
 
     # Initialize dictionary of property_list, with 1d array of nan (default value if no property for a given slice).
-    shape_properties = {key: np.full_like(np.empty(nz), np.nan, dtype=np.double) for key in property_list}
+    shape_properties = {key: np.full(nz, np.nan, dtype=np.double) for key in property_list}
```
SomeoneInParticular, line 68: Good catch!

spinalcordtoolbox/process_seg.py
```diff
@@ -78,30 +78,30 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
+                raise ValueError(f"The provided angle correction centerline does not cover slice {iz} of "
+                                 "the input mask. Please supply a more extensive -angle-corr-centerline, "
+                                 "or set -angle-corr to 0.") from None
             # Extract tangent vector to the centerline (i.e. its derivative)
-            tangent_vect = np.array([arr_ctl_der[0][iz - min_z_index] * px,
-                                     arr_ctl_der[1][iz - min_z_index] * py,
-                                     pz])
-            # Normalize vector by its L2 norm
-            tangent_vect = tangent_vect / np.linalg.norm(tangent_vect)
+            tangent_vect = np.array([dx * px, dy * py, pz])
```
SomeoneInParticular, line 100: Very clean; solves our problem and is much more slick too boot!

spinalcordtoolbox/process_seg.py
```diff
@@ -78,30 +78,30 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
                                ncols=80):
         # Extract 2D patch
         current_patch = im_segr.data[:, :, iz]
         if angle_correction:
+            try:
+                dx, dy = deriv[iz]
+            except KeyError:
+                raise ValueError(f"The provided angle correction centerline does not cover slice {iz} of "
+                                 "the input mask. Please supply a more extensive -angle-corr-centerline, "
+                                 "or set -angle-corr to 0.") from None
```
SomeoneInParticular, line 98: Very minor nit: 

```suggestion
                raise ValueError(f"The provided angle correction centerline does not cover slice {iz} of "
                                 "the input mask. Please supply a more extensive '-angle-corr-centerline', "
                                 "or disable angle correction ('-angle-corr 0').") from None
```

Just a tad clearer on the face of it IMO, though I'm not married to this change if you disagree.

(Also, TIL you can suppress the stack with `from None`; good to know, will be using that from now on)

spinalcordtoolbox/process_seg.py
```diff
@@ -78,30 +78,30 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
             # Compute the angle about AP axis between the centerline and the normal vector to the slice
-            v0 = [tangent_vect[0], tangent_vect[2]]
-            v1 = [0, 1]
-            angle_AP_rad = math.atan2(np.linalg.det([v0, v1]), np.dot(v0, v1))
+            angle_AP_rad = math.atan2(tangent_vect[0], tangent_vect[2])
             # Compute the angle about RL axis between the centerline and the normal vector to the slice
-            v0 = [tangent_vect[1], tangent_vect[2]]
-            v1 = [0, 1]
-            angle_RL_rad = math.atan2(np.linalg.det([v0, v1]), np.dot(v0, v1))
+            angle_RL_rad = math.atan2(tangent_vect[1], tangent_vect[2])
```
SomeoneInParticular, line 104: Very good code clean up, thank you.

SomeoneInParticular (changes requested):
One remark before discussing the code changes: we really should extend the `test_compute_shape` tests to test for cases like this in the future. However, as that would require a substantial re-write of the testing function, I think it is out of scope for this PR; I am open to discussion on this front, though.

Otherwise, having manually cross-referenced the code across the repo, and I don't see any major conflicts in these changes anywhere. Paired with your tests, I think this is safe to merge; just address the one nit, please.

spinalcordtoolbox/process_seg.py
joshuacwnewton, line 98: Is there any merit to not failing immediately, and instead keeping track of a list of missing slices, then if that list is non-empty, throwing an error emitting all missing slices?

(It would create more verbose output, but it might also save the user the hassle of, say, manually adding slice X to the centerline, re-running, getting a new single-slice error, and repeat.)

joshuacwnewton: Ah wait, nevermind, the way things are written, we have to fail here since further processing requires the result of this computation.

mguaypaq: Yes, that's a good idea. I implemented it in 41c1e1f416aa5232143b5a442154704d28c874ff. (And I added a line break on the `raise ValueError(` line so that the first line of the error message isn't repeated in the traceback.)

mguaypaq:
Thank you both for the reviews and suggestions! If/when tests pass, I think this is ready to merge.

spinalcordtoolbox/process_seg.py
joshuacwnewton, line 1: Ah, interesting! I noticed just now that this PR triggered the check warnings on the "detailed" batch_processing.sh test. (Go to "Files Changed" -> Scroll down -> 11 file properties changed.)

joshuacwnewton: Odd. It seems to just be the hashes and size of the CSV files:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14389434703/job/40355069065#step:5:91

I really need to implement https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4412 to make debugging this sort of thing a bit easier.

joshuacwnewton: Running the batch scripts locally on `master` and this branch ([batch_processing_csvs.zip](https://github.com/user-attachments/files/19709859/batch_processing_csvs.zip)), then diffing them, I see only the tiniest differences:

image

The date/version/etc. shouldn't matter here since I filter these out before doing the diff. But the last significant digits seem to change. Perhaps the refactor causes some float precision changes?

It doesn't matter all that much here, I don't think, since the results are basically equivalent. But, it's nice that the new test was even able to detect this at all!

joshuacwnewton (approved):
Given my last observation resulted in no needed changes, I'll approve. :)

SomeoneInParticular (approved):
I don't see any issues with the code either, and given the test Joshua mentioned seems primed to catch any issues in the future, I'm approving this PR as well.

Issue #4849, opened 2025-04-08
Re-order tasks and create sub-categories


jcohenadad:
Suggestion (just an example, following the web-based ordering and categories):

```console
TASK                          DESCRIPTION
--------------------------------------------------------------------------------
SPINAL CORD
spinalcord                    Spinal cord segmentation agnostic to MRI contrasts
sc_epi                        Spinal cord segmentation for EPI-BOLD fMRI data   
sc_lumbar_t2                  Lumbar cord segmentation with 3D UNet             
sc_mouse_t1                   Cord segmentation on mouse MRI                    
--------------------------------------------------------------------------------
GRAY MATTER
gm_sc_7t_t2star               SC/GM seg on T2*-weighted contrast at 7T          
gm_wm_exvivo_t2               Grey/white matter seg on exvivo human T2w         
gm_wm_mouse_t1                Exvivo mouse GM/WM segmentation for T1w contrast  
gm_mouse_t1                   Gray matter segmentation on mouse MRI             
--------------------------------------------------------------------------------
PATHOLOGIES
lesion_ms                     MS lesion segmentation on spinal cord MRI images  
lesion_ms_axial_t2            Intramedullary MS lesion and spinal cord segmentation in Axial T2w MRI
lesion_ms_mp2rage             MS lesion segmentation on cropped MP2RAGE data    
lesion_sci_t2                 Intramedullary SCI lesion and cord segmentation in T2w MRI
tumor_t2                      Cord tumor segmentation on T2-weighted contrast   
tumor_edema_cavity_t1_t2      Multiclass cord tumor/edema/cavity segmentation   
--------------------------------------------------------------------------------
OTHER
rootlets_t2                   Segmentation of spinal nerve rootlets for T2w contrast
totalspineseg                 Intervertebral discs labeling and vertebrae segmentation
sc_canal_t2                   Segmentation of spinal canal on T2w contrast      

Legend: installed | not installed
```

We should also re-order the gallery accordingly

SomeoneInParticular:
PR #4852 has now implemented this, feel free to give it a look and give some feedback!

PR #4832, opened 2025-03-27
Add new contrast-agnostic `graymatter` model

Nilser3:
#### GitHub

- [ ] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've linked relevant issues in the PR body
- [ ] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/onboarding/software-development.html#pr-labels) to this PR
- [ ] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [ ] I've assigned a reviewer

#### PR contents

- [ ] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [ ] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description
- This PR is for add the new model contrast and region agnostic : `graymatter` to `sct_deepseg`
- Model repo: [model-gm-contrast-region-agnostic](https://github.com/ivadomed/model-gm-contrast-region-agnostic) 
- The implementation was done based on the new `sct_deepseg` parser

## Linked issues
Resolves [#2](https://github.com/ivadomed/model-gm-contrast-region-agnostic/issues/2)

PR #4852, opened 2025-04-09
Improve `deepseg` usage and help in preparation for 7.0

### LINKED ISSUE #4848: Update `-custom-url` example command in argparse help
After introducing subparsers for `sct_deepseg` (i.e. 1 parser per task) in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4685, we now need to update the `-custom-url` usage. From

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/38793f1ace92ae22ddb8237d4d56d4827f736171/spinalcordtoolbox/scripts/sct_deepseg.py#L112-L113

to

```
sct_deepseg spinalcord -install -custom-url https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/download/v2.5/model_contrast-agnostic_20240930-1002.zip
```

i.e., the change is the swap of `TASK_NAME` and `-install`:

`sct_deepseg -install TASK_NAME` --> `sct_deepseg TASK_NAME -install`

### LINKED ISSUE #4851: General `deepseg` Improvements
As discussed in the recent internal meeting, a number of improvements to how `deepseg` is run and documented should be made before the 7.0 release. These include:

- [x] Add help for users used to the old `-task` syntax, so they can transition to our new `argparse` subtasks
- [x] Resolve the `-largest` argument missing a default value
- [x] Change the `TASK_NAME` metavar to be called `TASK` and move it to the front of the argument list
- [x] Change `-list-tasks` argument to be called `-task-details`, and update its help to be a clearer
- [x] Ensure the order of `tasks` available to the user are ordered in a more sensible manner (namely, so that it matches how the documentation orders them)
- [ ] Add deprecation warnings for methods which are now provided implicitly within `deepseg`, including
    - [x] `sct_deepseg_sc` -> `sct_deepseg spinalcord`
    - [ ] `sct_propseg` -> `sct_deepseg spinalcord` [**NOTE:** not a full deprecation, should make this clear]
    - [ ] `sct_label_vertebrae` -> `sct_deepseg totalspineseg` [**NOTE:** not a full deprecation, should make this clear]
    - [ ] `sct_deepseg_gm` -> `sct_deepseg graymatter` [**NOTE:** requires PR #4832 to be merged first]
    - [x] `sct_deepseg_lesion` -> `sct_deepseg ms_lesion`
- [x] Move `sc_canal_t2` into the "Other" models in documentation (rather than stranding it on its own)
- [x] Remove the `common_args` call for the top-level `deepseg` parser to clean it up a bit

I will extend and update this checklist as I work through it, and link to a corresponding PR when it is ready.

SomeoneInParticular:
#### GitHub

#### PR contents

## Description

This PR aims to implement the changes requested in our prior dev meeting, as detailed in #4851

## Linked issues

Resolves #4848, #4849, and #4851


SomeoneInParticular:
Something I discovered when improving the `test_sct_deepseg` test as part of [bc1e38e](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/bc1e38ea3a8546b9c13c28afc16e270082769fb7); we haven't updated the test in a while. As a result, the T2 rootlet segmentation is already detecting a mystery `4` label, which doesn't exist in the corresponding manual segmentation. 

Given the new rootlet segmentation model coming down the pipe (as discussed [here](https://github.com/ivadomed/model-spinal-rootlets/issues/90)), its very likely that adding this new model will further break the test. As such, its probably work re-creating a updated manual segmentation once the new model arrives, and updating our testing data with it. For now, though, I have just made it so "unexpected" labels raise a warning rather than failing the test.

SomeoneInParticular:
I have made it so deprecated calls now "pause" the runtime for 3 seconds to let the user read the warning raise (which details what scripts they can use instead). I'm open to other approaches on how to update the user though; this was just quick and easy to implement as a "proof of concept"

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -34,13 +35,60 @@
 logger = logging.getLogger(__name__)
 
 
+class _TaskDeprecationAction(Action):
+    """
+    ArgParse action which, similar to help, terminates the program early if the user tries to use the old `-task`
+    syntax and prompts them to update their command with the new `deepseg task_name` syntax instead
+    """
```
joshuacwnewton, line 42: This is splendid! It makes me wish we had used this pattern in previous instances where we had removed/modified CLI arguments.

I'd almost want to make a general `DeprecationAction` then have `_TaskDeprecationAction` be a child of it, though I'm sure since these will be one-off cases it might be better to just revisit that when we actually need a second one. :stuck_out_tongue_closed_eyes:

spinalcordtoolbox/deepseg/models.py
```diff
@@ -200,8 +200,14 @@
...
 
-# List of task. The convention for task names is: action_(animal)_region_(contrast)
-# Regions could be: sc, gm, lesion, tumor
+# List of tasks. The convention for task names is: action_(animal)_region_(contrast)
+#   Regions could be: sc, gm, lesion, tumor
+#   If the animal is human, omit it from the task label.
+# Some optional tags to help with groupin and sorting during help display are as follows:
+#   Group: The sub-header the tasks should be displayed under. If not specified, the model is placed under "OTHER"
+#   Priority: Higher priority tasks are displayed higher within their own group, and vice versa. Default priority is 0,
+#     and ties are sorted alphabetically.
```
joshuacwnewton, line 209: This `priority` idea is really smart! I love that it works like `nice` et al. in that we can have multiple high priority models while still fussing with the order (e.g. `100`, `99`, etc.).

It also feels a lot more intuitive than, say, a hardcoded `order` parameter, since we're indicating high/low priority intent.

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -253,6 +253,18 @@ def get_parser():
...
+    from warnings import warn
+    from spinalcordtoolbox.utils.sys import stylize
+    from time import sleep
+    warn(
+        stylize(
+            "`sct_label_vertebrae` is deprecated, and will be removed in a future version of SCT. Please use "
+            "`sct_deepseg totalspineseg` instead.", ["Red", "Bold"]
+        ), DeprecationWarning
+    )
+    sleep(3) # Give the user 3 seconds to read the message
```
joshuacwnewton: I would maybe suggest using a slightly softer deprecation message for `sct_label_vertebrae`? Something like "try out our new `totalspineseg` approach!", or something along those lines, rather than fully suggesting we plan to remove `sct_label_vertebrae`.

The reason I suggest that is because our `sct_label_vertebrae` CLI script is a little more involved than the other two deprecated scripts:

- `sct_deepseg_sc` and `sct_deepseg_lesion` are simply stanadlone TensorFlow models (that we've isolated by converting them to the ONNX model representation), so they're pretty painless to remove and sub out for drop-in replacements. 
- `sct_label_vertebrae`, on the other hand, is a rather novel [multi-step procedure](https://spinalcordtoolbox.com/stable/user_section/tutorials/vertebral-labeling/sct_label_vertebrae.html) based on older (albeit possibly outdated) traditional computer vision methods. It includes straightening, C2-C3 disc detection, and template matching using similarity measures. 
   - Similar to `sct_propseg`, this is also what make it more configurable than a black-box deep learning model. I think it provides a few levers the user can pull in case of a failed labeling?
   - Side note: It's also still heavily featured in our tutorials + the SCT course, whereas `sct_deepseg_sc` has already been swapped out for `sct_deepseg spinalcord`. 

Because of its novelty and complexity, I would be hesitant to remove it entirely, even if `totalspineseg` becomes our go-to approach? In much the same way we're keeping `sct_propseg` (while removing the newer `sct_deepseg_sc`), I think there's something to be said for preserving these older traditional approaches.

spinalcordtoolbox/scripts/sct_label_vertebrae.py
```diff
@@ -253,6 +253,18 @@ def get_parser():
 
 
 def main(argv: Sequence[str]):
+    # METHOD IS DEPRECATED, WARN THE USER AND UPDATE THEM ON WHAT TO USE INSTEAD
+    from warnings import warn
+    from spinalcordtoolbox.utils.sys import stylize
+    from time import sleep
+    warn(
+        stylize(
```
joshuacwnewton, line 261: I notice that the `warn(stylize(` pattern leaves a stray bit of `warn(` text in the console:

```console
dot|joshua@mosaic:~$ sct_label_vertebrae -h

--
Spinal Cord Toolbox (git-kjoa/deepseg_cleanup-0217bba2a29bd39c655b75bfaab07e7e5b3a678e)

sct_label_vertebrae -h
--

/home/joshua/repos/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_label_vertebr
ae.py:260: DeprecationWarning: `sct_label_vertebrae` is deprecated, and will be 
removed in a future version of SCT. Please use `sct_deepseg totalspineseg` 
instead.
  warn(
```

This is a little peculiar. Do you know why this might be?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -34,32 +35,107 @@
...
+        # Stop parsing immediately
+        parser.exit()
+
+
 def get_parser(subparser_to_return=None):
     # Initialize the top-level `sct_deepseg` argparser
     parser = SCTArgumentParser(
         description="Segment an anatomical structure or pathologies according to the specified deep learning model.",
         epilog=models.list_tasks_string()
     )
```
joshuacwnewton, line 111: I think one of the bits of feedback we got in the meeting was that the `usage:` string in the deepseg `-h` was a little unintuitive, since `TASK` came last in the list of arguments:

```console
> sct_deepseg -h

--
Spinal Cord Toolbox (git-kjoa/deepseg_cleanup-0217bba2a29bd39c655b75bfaab07e7e5b3a678e)

sct_deepseg -h
--

usage: sct_deepseg [-task-details] [-h] TASK ...
```

Luckily, though, it seems like `usage` can be overridden on a case-by-case basis. So, maybe we could update it to be something like:

```
usage: sct_deepseg TASK [-h] [-task-details]...
```

Or, we could even replace the `[-h] [-task-details]` text and put them alongside the usage examples, then pass everything to `usage`?


```
usage: sct_deepseg TASK ...
examples:
   sct_deepseg TASK -h
   sct_deepseg TASK -install
   sct_deepseg TASK -i input.nii.gz
view available tasks:
   sct_deepseg -h
   sct_deepseg -task-details

**Review by joshuacwnewton — COMMENTED** [review]
This is a splendid PR! So many small quality of life fixes and cleanups. Very satisfying to see. :)

File: spinalcordtoolbox/scripts/sct_deepseg.py (hunk 1 above)
Comment (SomeoneInParticular) [reply spinalcordtoolbox/scripts/sct_deepseg.py:42]: So, I was considering doing that, but as of Python 3.13 `argparse` just has that built-in. As such, I think our efforts would be better spent trying to port SCT up to the latest Python version and getting this benefit for free.

File: spinalcordtoolbox/scripts/sct_label_vertebrae.py (hunk 4 above)
Comment (SomeoneInParticular) [reply spinalcordtoolbox/scripts/sct_label_vertebrae.py:261]: I think its a side-effect of `warn` under the hood being an Exception, so the line that calls it is always reported to the user (as any exception is by default). Its this, or PyTest the co. not detecting the warning as a genuine warning; IMO this is the lesser evil (though I do hate how ugly it is).

File: spinalcordtoolbox/scripts/sct_deepseg.py (hunk 5 above)
Comment (SomeoneInParticular) [reply spinalcordtoolbox/scripts/sct_deepseg.py:111]: Oh yes! This totally slipped my mind, will add this in the next push.

**Comment by SomeoneInParticular** [comment]
Needs to be discussed; how should we handle "partial" deprecation, as is the case with `propseg` and `sct_label_vertebrae`? After all, they will stick around, but we still want to encourage most users to transition over to the newer tools. 

I believe there are a few viable options for this:

* Treat them like any other deprecation, w/ or w/o the 3 second pause.
* Display a warning message at runtime completion to notify the user that a newer method exists, and may be worth using.
* Mention it in the `help` of the command (and associated documentation) that another tool exists which may do this job better.

File: spinalcordtoolbox/scripts/sct_label_vertebrae.py (hunk 3 above)
Comment (SomeoneInParticular) [reply spinalcordtoolbox/scripts/sct_label_vertebrae.py]: I've removed the deprecation warning for now; will come back to this after we can discuss it as a group (as I'd rather not discourage users from using this until we are sure we want to).

File: spinalcordtoolbox/scripts/sct_deepseg.py (hunk 5 above)
Comment (SomeoneInParticular) [reply spinalcordtoolbox/scripts/sct_deepseg.py]: Resolved with [1c80943](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/1c8094308906c3e2b3397303a8d4abf6c2e12378)

File: spinalcordtoolbox/scripts/sct_deepseg.py [hunk 6]
```diff
@@ -106,6 +106,18 @@ def get_parser(subparser_to_return=None):
...
+        
+        Examples:
+            sct_deepseg spinalcord -h
+            sct_deepseg gm_mouse_t1 -install
+            sct_deepseg lesion_ms -i cMRI3712.nii.gz
+        
+        View available tasks:
+            sct_deepseg -h
+            sct_deepseg -task-details
+        """[1:]),
```
joshuacwnewton, line 120: This is a funny way of excluding that initial `\n` from putting `sct_deepseg` on its own line. :laughing:

joshuacwnewton (approved):
Thank you for this!! :hearts:

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: It was that or trigger my developer OCD; seems far cleaner this way.

SomeoneInParticular:
Just a note to myself (or other devs) for the future; due to the deprecation of `sct_deepseg_sc` and co., quite a few methods which relied on it will need to be updated soon as well. Of particular note is the FSLeyes plugin, which honestly has needed an update pass for quite a while anyways...

SomeoneInParticular:
> We also need to update `batch_processing.sh` tests to references results from the new (better) models, as the results have changed enough to fail them now.

This has been done as of commit [f871e96](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/f871e96cd7cb59da036ac378f1d346f9c9145eda)

batch_processing.sh
```diff
@@ -58,6 +58,13 @@ start=$(date +%s)
 if [[ "$SCT_BP_DOWNLOAD" == "1" ]]; then
   sct_download_data -d sct_example_data
 fi
+
+# If there is no SCT_DIR set (i.e. this is a git clone), try to use the current directory instead
+if [[ -z "$SCT_DIR" ]]; then
+  SCT_DIR="$PWD"
+fi
+
+# Enter our data directory
```
joshuacwnewton, line 67: Interesting! I'm surprised that you would have the SCT commands present on the `$PATH` _without_ `$SCT_DIR` set? Usually the two go hand-in-hand (e.g. [my `.bashrc`](https://github.com/joshuacwnewton/dotfiles/blob/8d988e4a0e68720eb413ba39b6144b94b0efa6f1/.bashrc#L51-L53)).

batch_processing.sh
```diff
@@ -58,14 +58,23 @@ start=$(date +%s)
...
 
+# install the sct_deepseg model this script will use later
+sct_deepseg spinalcord -install
 
 # t2
 # ===========================================================================================
 cd t2
 # Segment spinal cord
-sct_deepseg_sc -i t2.nii.gz -c t2 -qc "$SCT_BP_QC_FOLDER"
+sct_deepseg spinalcord -i t2.nii.gz -qc "$SCT_BP_QC_FOLDER"
```
joshuacwnewton, line 77: Thank you for this! It's quite a radical change (hence the significant difference in results), but given that we've already migrated everything else over to `spinalcord`, I'm glad to have finally cleaned things up here as well. :)

spinalcordtoolbox/scripts/sct_deepseg_sc.py
joshuacwnewton, line 1: I don't really know why I hadn't thought to do this before, but I tried Ctrl+Shift+F'ing for `sct_deepseg_sc` and `sct_deepseg_lesion`, and I notice a few stray mentions of `sct_deepseg_sc` in both `sct_propseg.py` and `sct_straighten_spinalcord.py`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/192c518c829a7d8ea7764a63c4b46748ff2ed836/spinalcordtoolbox/scripts/sct_propseg.py#L649-L654

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/192c518c829a7d8ea7764a63c4b46748ff2ed836/spinalcordtoolbox/scripts/sct_straighten_spinalcord.py#L36-L41

joshuacwnewton (commented):
Almost there!

README.rst
```diff
@@ -88,18 +88,29 @@ The primary way to invoke SCT is through terminal commands. For example:
...
-  Cropping the image around the spinal cord...
-  Normalizing the intensity...
-  Segmenting the spinal cord using deep learning on 2D patches...
-  Reassembling the image...
-  Resampling the segmentation to the native image resolution using linear interpolation...
-  Binarizing the resampled segmentation...
-  Compute shape analysis: 100%|################| 55/55 [00:00<00:00, 106.05iter/s]
-
-  Done! To view results, type:
-  fsleyes t2.nii.gz -cm greyscale t2_seg.nii.gz -cm red -a 70.0 &
```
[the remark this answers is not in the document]
SomeoneInParticular: Fair; just wanted to be on the safe side was all. Will trim it down next push.

batch_processing.sh
SomeoneInParticular: You're actually correct (I had activated the SCT environment before running the tests and forgot). There is a workaround though; we can explicitly activate the environment at this point do accomplish the same (as the environment implicitly adds any entrypoints it has to PATH). Thoughts?

spinalcordtoolbox/scripts/sct_deepseg_sc.py
SomeoneInParticular: Yup; still plenty more to do, just wanted to review and push the substantial changes to our cached results ASAP to save us some headache.

batch_processing.sh
joshuacwnewton: > I had activated the SCT environment before running the tests 

Ahhh, I see! If you're manually activating the venv, then, does that mean that your `.bashrc` contains neither the SCT_DIR nor the PATH modifications?

((Generally our users are not entirely technical, and so we try to save them the burden of having to activate any sort of environment -- instead, we modify their `.bashrc`/`.zshrc` files during the `install_sct` script so that SCT_DIR and PATH are perpetually set, that way everything just sort of "works". Generally speaking, the presence of SCT_DIR and PATH is an assumption that SCT tutorials, courses, etc. make, no environment activation needed.))

SomeoneInParticular: Sounds good, will remove that code snippet then (and work under the assumption the user has made a full SCT install). Might come back around to making a 'local' testing batch script later, which doesn't require a full install, but thats for a later PR.

joshuacwnewton: Just out of curiosity, what steps did you perform to get a so-called "non-full" install? Something akin to https://spinalcordtoolbox.com/latest/user_section/installation/linux.html#option-3-install-with-pip-experimental?

SomeoneInParticular: Sorry, just saw this.

1. `git clone sct`
2. `sh install_sct.sh`
3. When prompted whether you want to add the `sct_*` tools to `PATH`, type `n` to decline and proceed.

This is mostly to prevent SCT from polluting my `.bashrc` file every time I need to re-install it (which, given I am often breaking it during testing, is rather often), but also to ensure that `sct_*` utils are only accessible on the command line when I explicitly want them to be (useful for testing other scripts).

SomeoneInParticular:
Have we come to a conclusion on `spinalcord-smoothing.rst` now that `sct_deepseg_sc` is being deprecated? If so, I'll delete it from the docs as part of this PR.

README.rst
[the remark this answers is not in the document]
SomeoneInParticular: Resolved with [f4090af](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/f4090afdcfff79d6dbffd3431f99d1a4ba4e45ba)

documentation/source/user_section/command-line/all_tools.rst
joshuacwnewton, line 1: > Have we come to a conclusion on spinalcord-smoothing.rst now that sct_deepseg_sc is being deprecated? If so, I'll delete it from the docs as part of this PR.

I recall the meeting response being sort of... tepid? Inconclusive? And I think that says all we need to know about this page: `sct_smooth_spinalcord` is not a vital feature, and what we do with that page is of low consequence.

I think we can just remove the references to `sct_deepseg_sc` entirely, then rewrite the concluding section to use weasel words about the efficacy of the command? Something like:

```
.. figure:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/spinalcord-smoothing/io-sct_smooth_spinalcord.png
   :align: center

After smoothing, the apparent noise is reduced, while the cord edges are preserved. This may 
allow for the spinal cord to be segmented in difficult cases where segmentation tools initially 
produce poor results.

It is recommended, however, to always inspect your output and perform manual corrections when 
necessary. This is especially critical when performing quantitative analyses across a multi-
subject dataset (e.g. computing CSA). By performing smoothing on one image (but not others), 
you may introduce a per-subject bias that can lead to under/over-segmentation relative to 
other subjects within the same dataset.
```

The idea being that rather than giving a decisive conclusion ("smoothing improves the quality of the output" or "smoothing is not necessary now that we have better models"), we simply say that it _may_ help, but requires careful application.

joshuacwnewton, line 1: > Edit: Ditto w/ lumbar-segmentation.rst, which opens by saying that sct_deepseg_sc is better at non-lumbar sequences (which I don't think is the case anymore)

I think the reason that I had written it that way in the past was because I wanted to really hammer home the relative strengths and weaknesses of both `sct_deepseg_sc` and `sct_deepseg sc_lumbar_t2`.

But, as you say, `sct_deepseg spinalcord` probably fares a bit better now with the lumbar region. So, like with the other page, I think we could probably stick to vaguer, less decisive recommendations:

```
In previous registration tutorials, :ref:`sct_deepseg spinalcord <sct_deepseg_spinalcord>` 
was used to segment the spinal cord. This is because ``spinalcord`` was largedly trained on 
typical spinal cord scans spanning the cervical and thoracic regions. For some images, it 
may perform poorly on the lumbar region. In cases where the output of ``spinalcord`` is 
insufficient, we can rely on an alternate :ref:`sct_deepseg` model instead that was trained 
entirely on lumbar scans (:ref:`sct_deepseg_sc_lumber_t2`).
```

Then nix the parts about `sct_deepseg`'s various models and etc.

joshuacwnewton (commented):
I have no real comments on the most recent batch of changes! You still have my approval (though I'll wait for confirmation that you don't have any further changes to make before giving a final approval). :slightly_smiling_face:

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -21,6 +21,8 @@
 from textwrap import dedent
 import functools
 
+from Cython.Debugger.Cygdb import usage
+
```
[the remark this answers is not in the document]
SomeoneInParticular: I need to look into doing that, this has been a headache for a while now...

documentation/source/user_section/command-line/all_tools.rst
SomeoneInParticular: Resolved w/ [920c690](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/920c690f8a8dc8ec5a2b723aaff570ae6fa5ca55)

SomeoneInParticular: Resolved with [7b5a717](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/7b5a71718d74a38bd01949edb5a1fe9ea2dc0b7b)

SomeoneInParticular:
@joshuacwnewton please give it a last minute once over to confirm I didn't lose the plot with any of the final changes. If not, this is RTM!

joshuacwnewton (commented):
Just a heads up -- there's a sister pending PR right now:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4832

The changes from that PR will probably cause a merge conflict -- sorry in advance! 

> (I figure it's probably easiest to fix any conflicts here, rather than vice-versa, just because you have a deep understanding of `git` and the structure of `sct_deepseg` at this point!)

Once that gets resolved, I think this should be ready to go as the last task for v7.0. :tada:

documentation/source/user_section/command-line/sct_deepseg.rst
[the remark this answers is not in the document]
SomeoneInParticular: Oh yes, of course. Will add it in the next push.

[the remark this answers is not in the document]
SomeoneInParticular: Resolved with [d6da5e8](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/d6da5e8e915722fa919e8c723edfdf18a6dd2249)

joshuacwnewton (approved):
LGTM! Thank you so much for this colossal effort. I'm so glad we were able to get to everyone's feedback prior to the release.

PR #4854, opened 2025-04-14
Temporarily test branch `jv/33-remove_pixels_outside_of_cord`

joshuacwnewton:
> @joshuacwnewton, do we have any Github workflow in SCT which uses the atlas that we could use to quickly test whether we get the same results?

I think the `batch_processing.sh` test will help here, especially now that we have the "detailed analysis" secondary step that checks the exact properties of each file generated during `batch_processing.sh`. 

So, this PR runs all of our usual tests (including `batch_processing.sh`)

For context: @valosekj's PR https://github.com/spinalcordtoolbox/PAM50/pull/35

### DIFF
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -66,7 +66,7 @@ DATASET_DICT = {
     },
     "PAM50": {
         "mirrors": [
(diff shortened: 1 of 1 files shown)


joshuacwnewton:
Closing as this PR has done its job. :) 

We can always re-open to merge the new PAM50 release once the upstream PAM50 PR is merged and a new release is created.

PR #4856, opened 2025-04-15
Enable `macos-15` and `windows-2025` in all CI workflows

### LINKED ISSUE #4853: macOS 15 and Windows 2025 runners are now available
See this notice: https://github.blog/changelog/2025-04-10-github-actions-macos-15-and-windows-2025-images-are-now-generally-available/

I'd like to briefly test these platforms as part of my "pre-release" checklist for v7.0. :)

joshuacwnewton:
## Description

Draft PR for testing.

I've temporarily enabled them on PRs, even if we don't plan to *always* enable them on PRs right now, so that I can test the current `master` on these OSs.

## Linked issues

Fixes #4853.

### DIFF
--- a/.github/workflows/test-past-releases.yml
+++ b/.github/workflows/test-past-releases.yml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         sct_version: [ "6.5", "6.4", "6.3", "6.2", "6.1", "6.0", "5.8", "5.7", "5.6", "5.5", "5.4", "5.3.0", "5.2.0", "5.1.0" ]
-        os: [ ubuntu-latest, macos-latest, windows-2019 ]
+        os: [ ubuntu-latest, macos-latest, macos-15, windows-2019, windows-2025 ]
         exclude:
           # Windows support was only properly introduced in 5.7
           - sct_version: "5.6"
@@ -42,6 +42,18 @@ jobs:
             os: windows-2019
           - sct_version: "5.1.0"
             os: windows-2019
+          - sct_version: "5.6"
+            os: windows-2025
+          - sct_version: "5.5"
+            os: windows-2025
+          - sct_version: "5.4"
+            os: windows-2025
+          - sct_version: "5.3.0"
+            os: windows-2025
+          - sct_version: "5.2.0"
+            os: windows-2025
+          - sct_version: "5.1.0"
+            os: windows-2025
     runs-on: ${{ matrix.os }}
     defaults:
       run:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -284,6 +284,28 @@ jobs:
         run: |
           ./.ci.sh -t
 
+  macos-15:
+    name: macOS 15.0 (Sequoia)
+    # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
+    #if: ${{ env.NIGHTLY }}
+    # in the meantime, copy-paste this:
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+    runs-on: macos-15
+    steps:
+      - name: Dependencies
+        run: |
+          # github runners come with dev-tools pre-installed
+      - uses: actions/checkout@v4
+      - name: Install SCT
+        run: |
+          ./.ci.sh -i
+      - name: Check dependencies
+        run: |
+          ./.ci.sh -c
+      - name: Run pytest test suite
+        run: |
+          ./.ci.sh -t
+
   windows-wsl-ubuntu-24_04:
     # with the help of https://github.com/marketplace/actions/setup-wsl
     name: WSL [Ubuntu 24.04]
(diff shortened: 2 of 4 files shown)


.github/workflows/test-past-releases.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         sct_version: [ "6.5", "6.4", "6.3", "6.2", "6.1", "6.0", "5.8", "5.7", "5.6", "5.5", "5.4", "5.3.0", "5.2.0", "5.1.0" ]
-        os: [ ubuntu-latest, macos-latest, windows-2019 ]
+        os: [ ubuntu-latest, macos-latest, macos-15, windows-2019, windows-2025 ]
```
joshuacwnewton, line 30: Everything appears to be working OK on these new platforms! :tada:

The only catch is `windows-2025` on some older releases. There seems to be 1 failing test for the `seg_sc_t2star` model: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14473727158/job/40594082475#step:10:463

But, given that we've retired this model anyway, I don't think it's high-priority at all. Meaning, we can proceed as expected with the v7.0 release.

joshuacwnewton: I don't really have a solution for this failing model, so I've added a bit of a hack fix to skip the test for `task == 'seg_sc_t2star'` on Windows in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4856/commits/dc37c5eab2f4acfdc248cb2842091412f97df070.

Pretty sure this works for all versions of SCT, but it took a bit of trial and error: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15258537463 (famous last words, it's still a WIP right now)

.github/workflows/tests.yml
```diff
@@ -284,6 +284,28 @@ jobs:
         run: |
           ./.ci.sh -t
 
+  macos-15:
+    name: macOS 15.0 (Sequoia)
+    # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
+    #if: ${{ env.NIGHTLY }}
+    # in the meantime, copy-paste this:
+    # if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```
joshuacwnewton, line 292: I imagine we probably want to also keep macos-15 from running on PRs? (Or swap out some other OS?)

I think this one cuts into our macOS M1 allocation for runners, so might be best to save it for the nightly runs.

.github/workflows/test-past-releases.yml
```diff
@@ -52,6 +64,14 @@ jobs:
         with:
           ref: ${{ matrix.sct_version }}
 
+      # The model `seg_sc_t2star` is failing on Windows 2025, but not on other platforms.
+      # It seems to produce non-deterministic outputs with varying Dice scores between runs.
+      # However, this model has been retired in v7.0, and was not really used even when it
+      # was available. So, we can just skip the test when running tests on Windows 2025.
+      - name: Add `seg_sc_t2star` hotfix for Windows 2025
+        if: matrix.os == 'windows-2025'
+        run: sed -i '/fname_out = str(tmp_path\/fname_out)  \# tmp_path for automatic cleanup/i\    if task == "seg_sc_t2star" and os.name == '\''nt'\'':\n        pytest.skip("Model '\''seg_sc_t2star'\'' has non-deterministic output on Windows")' testing/cli/test_sct_deepseg.py
```
mguaypaq, line 73: I'm not very comfortable with this dynamic patching of the code with a `sed` script that matches on a fairly generic line of code.

I would have suggested using `patch` (which is explicitly made for this purpose, and includes much more context surrounding the patch), but I see that the patched code changes too much over the different versions of SCT (in particular, there's only one modified chunk in the earlier versions, and two modified chunks in the later versions).

Maybe this patch can just be actually applied and backported to the affected SCT versions? It's more work, but it would put the information closer to the expected source of truth. (I can imagine seeing a skipped test in the CI output while seeing no indication in the source code, or seeing all the other line numbers be off-by-two, and being quite puzzled.)

.github/workflows/tests.yml
mguaypaq: Yes, sounds like a good idea.

mguaypaq (changes requested):
Thanks for moving our testing infrastructure forward!

.github/workflows/test-past-releases.yml
joshuacwnewton: > Maybe this patch can just be actually applied and backported to the affected SCT versions?

I absolutely agree with your apprehension here -- this very much feels like a shaky hack, and `sed` feels like a very fragile way of applying the fix.

That said, the patch I'm applying is more like a silencer -- it's not actually fixing the underlying problem, but more so quieting a known error for dev sanity?

If we were to backport this "`pytest.skip`" fix, then we would be silencing the test error on the user's end too, which seems risky to do without the problem itself being fixed? I wanted it to still be possible for a hypothetical Windows user to run `sct_testing` on an old version of SCT, see the failing test, and be clued into the fact that `seg_sc_t2star` is broken for certain versions of Windows.

If we had an actual fix for the model itself, I would absolutely backport it. But, because it's just modifying our test suite, I figured it would be okay to localize the changes to the CI... :thinking:

mguaypaq (approved):
This looks fine now, thanks!

joshuacwnewton:
Thank you so much for the help @mguaypaq! I appreciate you finding the correct syntax to exclude tests in a granular manner. :hearts:

PR #4860, opened 2025-04-16
Add `-step1-only` argument to the `totalspineseg` task's subparser

### LINKED ISSUE #4794: Add ability to run only `--step1` for totalspineseg to avoid unnecessary processing
I wanted to use totalspineseg with sct_deepseg to obtain the labelling of the IVDs on an unlabelled dataset.
Totalspineseg works really well for this because the output of the step1 from the model gives extcaly the points of those disks. 

But since it's step1 output I don't need to run the other steps. The totalspineseg repo had anticipated this, providing a flag when running the model : --step1

So I was wandering if it was possible to add this to sct_deepseg on -task totalspineseg @NathanMolinier 

Thanks!

joshuacwnewton:
## Description

This PR adds a new argument `-step1-only` to `totalspineseg`'s subparser, which gets passed to the `step1_only` parameter of the `inference()` function for `totalspineseg`. This allows users to skip step2, which may be costly/time-consuming.

Additionally, this PR adds a test, with the resulting QC report shown below:

[Screencast from 2025-04-16 11-14-57.webm](https://github.com/user-attachments/assets/bd4a6f4e-f8a9-4273-96e8-183c3284179a)

> Kudos to @elhil who recently revamped the QC report! Now we have a nice zoom toggle, which makes it much easier to view large QC reports like this one. :hearts:

## Linked issues

Fixes #4794.

### DIFF
--- a/documentation/source/user_section/tutorials/lesion-analysis/lesion-metrics.rst
+++ b/documentation/source/user_section/tutorials/lesion-analysis/lesion-metrics.rst
@@ -4,9 +4,9 @@ Compute lesion morphometric measures
 Various morphometrics
 ---------------------
 
-Here, we will see how to compute various morphometric measures on segmented lesions. For example, the number of lesions, lesion length, lesion volume, etc. The statistics will be output in an Excel (XLS) file.
+Here, we will see how to compute various morphometric measures on segmented lesions. For example, the number of lesions, lesion length, lesion volume, etc. The statistics will be output in an Excel (XLSX) file.
 
-In the case of multiple lesions, the function assigns an ID value to each lesion (1, 2, 3, etc.) and then outputs morphometric measures for each lesion to an XLS file.
+In the case of multiple lesions, the function assigns an ID value to each lesion (1, 2, 3, etc.) and then outputs morphometric measures for each lesion to an XLSX file.
 
 The following morphometric measures are computed:
 
@@ -65,4 +65,4 @@ Details:
   * `Pfyffer, D., et al. Neurology (2019) <https://doi.org/10.1212/WNL.0000000000008318>`_
   * `O'Dell, D.R., et al. J Spinal Cord Med (2020) <https://doi.org/10.1080/10790268.2018.1527079>`_
   * `Pfyffer, D., et al. Lancet Neurol. (2024) <https://doi.org/10.1016/S1474-4422%2824%2900173-X>`_
-  * `Enamundram, N.K., Valošek, J., et al. arXiv (2024) <https://doi.org/10.48550/arXiv.2407.17265>`_
+  * `Enamundram, N.K., Valošek, J., et al. MICCAI-AMAI (2024) <https://doi.org/10.1007/978-3-031-82007-6_19>`_


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -131,7 +131,7 @@ def create_net(pm, _): return pm  # just echo `path_model` back as 'net'
     for fname_in in input_filenames:
         tmpdir = tmp_create(basename="sct_deepseg")
         # model may be multiclass, so the `inference` func should output a list of fnames and targets
-        fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device)
+        fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device, **extra_inference_kwargs)
```
SomeoneInParticular, line 134: I always love to see the unpacking operator in action. Good job!

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -188,6 +188,16 @@ def get_parser(subparser_to_return=None):
         help="Contrast of the input. Specifies the contrast order of input images (e.g. `-c t1 t2`)",
         choices=('t1', 't2', 't2star'),
         metavar=Metavar.str)
+    parser_dict['totalspineseg'].add_argument(
```
SomeoneInParticular, line 191: Just noticed this places the argument in the `optional arguments` argument group, which does not follow the ALL CAPS standard of the others. Something like this will resolve that:

```suggestion
    parser_dict['totalspineseg'].optional_arggroup.add_argument(
```

We might want to do the same with the `tumor_edema_cavity_t1_t2` argument too, come to think of it

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -322,7 +332,12 @@ def main(argv: Sequence[str]):
                 keep_largest=1 if arguments.task == 'spinalcord' else arguments.keep_largest,
                 fill_holes_in_pred=arguments.fill_holes,
                 remove_small=arguments.remove_small,
-                use_gpu=use_gpu, remove_temp_files=arguments.r)
+                use_gpu=use_gpu, remove_temp_files=arguments.r,
+                # Pass any "extra" kwargs defined in task-specific subparsers
+                extra_inference_kwargs={arg_name: getattr(arguments, arg_name)
+                                        for arg_name in ["step1_only"]  # Used only by totalspineseg
+                                        if hasattr(arguments, arg_name)}
```
SomeoneInParticular, line 339: Why is this a dict comprehension? `getattr` allows you to set a default return (as, under the hood, its just a dict query anyways):

```suggestion
                extra_inference_kwargs={"step1_only": getattr(arguments, "step1_only", None)}
```

*NOTE:* despite the default of `getattr`'s 'default' already being `None`, we need to specify it anyways due to how `getattr` works under the hood (raising an error on use if no default was explicitly define).

SomeoneInParticular (commented):
Only a few minor remarks; otherwise this is a very elegant PR, and LGTM! Thank you for adding tests already as well.

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: I think my reasoning was that specifying `{'step1_only': None}` by default would cause an error for MONAI/nnUNet models? (Since the `step1_only` parameter is only present in the `segment_totalspineseg` function.)

```python
>>> from spinalcordtoolbox.deepseg.inference import segment_totalspineseg
>>> segment_totalspineseg("test", "test", "test", "test", **{'step1_only': None})
AssertionError: No 'nnUNet/results/rYYYYMMDD' subdirectory found in test

>>> from spinalcordtoolbox.deepseg.inference import segment_nnunet
>>> segment_nnunet("test", "test", "test", "test", **{'step1_only': None})
  File "/opt/pycharm-2024.2.3/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
TypeError: segment_nnunet() got an unexpected keyword argument 'step1_only'
```

The dictionary comprehension was just to default to `{}` if non-totalspineseg tasks are used. 

(Also, I figured the pattern was general enough that we could in theory add other subparser-specific arguments to the same list to be able to pass those along too.)

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -188,6 +188,16 @@ def get_parser(subparser_to_return=None):
         help="Contrast of the input. Specifies the contrast order of input images (e.g. `-c t1 t2`)",
         choices=('t1', 't2', 't2star'),
         metavar=Metavar.str)
+    parser_dict['totalspineseg'].add_argument(
+        "-step1-only",
+        type=int,
+        help="If set to '1', only step one will be performed.\n"
+             "- Step 1: <>\n"
+             "- Step 2: <>\n"
```
joshuacwnewton, line 196: Whoops! This was meant to be a placeholder that I filled in once I dug into the outputs. Fixing this now, too. :)

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: The changes I made were a little more involved (in order to access the original `input_output` and `params` groups more easily), but it really was just an indent required. :)

https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4860/commits/dbbbd87d8fe9a82946ff41c237e0098551e07906

SomeoneInParticular: Oh, whoops; I usually catch that, but mixed it up. Thanks for catching that!.

spinalcordtoolbox/scripts/sct_deepseg.py
SomeoneInParticular: Fair enough; at that point, though, would it not be smarter to simply bundle everything in `kwargs` wholesale and pass it through?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -181,13 +182,24 @@ def get_parser(subparser_to_return=None):
...
+        if task_name == 'totalspineseg':
+            params.add_argument(
+                "-step1-only",
+                type=int,
+                help="If set to '1', only Step 1 will be performed. If not provided, both steps will be run.\n"
+                     "- Step 1: Segments the spinal cord, spinal canal, vertebrae, and intervertebral discs (IVDs). Labels the IVDs, but vertebrae are left unlabeled.\n"
+                     "- Step 2: Fine-tunes the segmentation, applies labels to vertebrae, and segments the sacrum if present.\n"
+                     "More details on TotalSpineSeg's two models can be found here: https://github.com/neuropoly/totalspineseg/?tab=readme-ov-file#model-description",
+                choices=(0, 1),
+                default=0)
```
SomeoneInParticular, line 202: Thank you! Looks great.

NathanMolinier (commented):
It worked well for me ! Just a minor point, regarding the flag `-step1-only` could we set it to `store_true` ? Instead of specifying 1 or 0 after ?

abelsalm:
I can't review since I don't have the developper mode on Windows... 
But from what I se this looks great !
Thanks Joshua

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: Good question! I touched on this a little in the extended commit message:

> To accomplish this, we use a dictionary of "extra" keyword arguments to smuggle in kwargs that may be present for some tasks/frameworks/subparsers, but not for others.
>
> It's possible that this is a smell that the "common function signature" has outgrown its usefulness, and that we need to refactor how arguments are parsed for `sct_deepseg`? But this works for a first draft, and I like that it clearly delineates between common arguments and unique arguments.

In other words, I wanted to still try and enforce a common function signature with similar args (i.e. every NN framework function should have an input image, a predictor, a `torch` device, and a `tmpdir` to write outputs to). And the "extra kwags" parameter was a nod to the fact that `step1_only` is an outlier, and specific to `totalspineseg`.

If we combined everything into `kwargs`, then it increases the risk that some parameter could be mis-specified, isn't as self-documenting, etc. I'm kind of treating `kwargs` like a last resort, aha.

joshuacwnewton:
> It worked well for me ! Just a minor point, regarding the flag `-step1-only` could we set it to `store_true` ? Instead of specifying 1 or 0 after ?

I mainly did this to keep it consistent with the existing flags (such as `-keep-largest`, etc.) since that's the format we standardized on in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4852/commits/6f8338eb6f3eecc33e5073ea7ebaaa0cd0c2b369?

NathanMolinier:
I see ! No problem if it's the standard way for SCT then !

SomeoneInParticular:
> It worked well for me ! Just a minor point, regarding the flag -step1-only could we set it to store_true ? Instead of specifying 1 or 0 after ?

Just to follow-up on this as well; now that I have some practice with `Argparse` Actions, I think there might be a way of having our cake and eating it too. That is, if we make a standard "flag-like" argument type, it could call an action which auto-assigns to a default (1 or 0, based on context) if none is specified. Definitely not for 7.0 though (and we should discuss it in an SCT meeting first regardless).

SomeoneInParticular:
Assuming all tests pass (and the others have no issue with it), this LGTM!

SomeoneInParticular (approved):

NathanMolinier (approved):

Issue #3925, opened 2022-10-31
`PyQt5>=5.15.x` crashes on Ubuntu due to missing `libxcb-xinerama.so.0` object file


joshuacwnewton:
Throughout SCT's history, we've encountered various quirks when specifying the range of allowable versions for PyQt5, which we use for some GUI-based scripts.

- Pinning to `PyQt5==5.11.3`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2524
- Unpinning `PyQt5`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3916#discussion_r997444960
- Pinning to `<5.15`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3928
- Pinning to `<5.14`: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1533181462

During this process, we've noticed that when trying to upgrade to `PyQt5>=5.15.*` on Ubuntu, our GUI-based scripts crash:

<details>

```
Check if figure can be opened with PyQt.............qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.
```

Re-running `sct_check_dependencies` with `export QT_DEBUG_PLUGINS=1` set yields the true source of the error:

<details>

```
Check if figure can be opened with PyQt.............QFactoryLoader::QFactoryLoader() checking directory path "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.8/site-packages/PyQt5/Qt5/plugins/platforms" ...
QFactoryLoader::QFactoryLoader() looking at "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.8/site-packages/PyQt5/Qt5/plugins/platforms/libqeglfs.so"
Found metadata in lib /home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.8/site-packages/PyQt5/Qt5/plugins/platforms/libqeglfs.so, metadata=
{
[...]

PR #4869, opened 2025-04-29
Update SCT's virtual environment to use Python 3.10 instead of 3.9

### LINKED ISSUE #4356: Add support for Python 3.10 by 2025-10 (Python 3.9 EOL)
In the same vein as previous issues:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3227

This issue is to ensure that we are aware of the challenges of moving to Python 3.10 well in advance of when we need to migrate off of Python 3.9.

joshuacwnewton:
## Description

This PR replaces the nearly EOL Python 3.9 with Python 3.10 in SCT's venv. (Python versions 3.11-3.13 were also tested, however some issues were encountered that must be solved in a separate PR, see below.)

To get Python 3.10 to install correctly, this PR also addresses a longstanding PyQt5 issue by bundling some new required libs + adopting a `LD_LIBRARY_PATH` fix to access those libs. (Clearing out `LD_LIBRARY_PATH` just happens to fix a recently-reported forum issue, too! But, perhaps that should be handled separately, since it is a bigger change.)

## Linked issues

- Fixes #4356.
- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3925.
- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4903.


requirements.txt
```diff
@@ -42,11 +42,9 @@ onnxruntime>=1.7.0,!=1.16.0
 pandas
 portalocker
 psutil
-# pyqt5=5.11.X causes https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3916#discussion_r997435037
-# pyqt5=5.15.X causes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3925
-# Additionally, for versions between 5.12-5.14, the only release to have Python 3.9 wheels is 5.12.3. So, we pin it.
-# See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3367#issuecomment-1533181462
-pyqt5==5.12.3
```
joshuacwnewton, line 49: Note: It looks like we might be forced to finally upgrade `pyqt5`. 

While `5.12.3` succeeds on Ubuntu/macOS for Python 3.10, it [fails for Windows](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14732764900/job/41351001912?pr=4869#step:3:613), since it maxes out at Python 3.9 (even though the other OSs say Python 3.9+):

> image

Unfortunately, upgrading to `5.15.x` will likely run us headfirst into:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3925

So, we might have to consider steps that will make it easier for users to install the necessary dependency.

requirements.txt
```diff
@@ -38,15 +38,14 @@ numpy<2
 # 1.7.0>onnxruntime>=1.5.1 required `brew install libomp` on macOS.
 # So, pin to >=1.7.0 to avoid having to ask users to install libomp.
 # Avoid version 1.16.0 due to: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4225
-onnxruntime>=1.7.0,!=1.16.0
+# 1.15.1 seems to include a special fix related to Windows DLL loading.
+onnxruntime>=1.7.0,!=1.16.0,==1.15.1
```
joshuacwnewton, line 42: Separately from the PyQt5 issues, we're now running into a different Windows-specific error:

```
Check if onnxruntime is installed...................[FAIL]
An error occured while importing module onnxruntime ->
    DLL load failed while importing onnxruntime_pybind11_state: A dynamic link library (DLL) initialization routine failed.
```

This looks to be related to:

- https://github.com/microsoft/onnxruntime/issues/16116

According to this issue, the bug was fixed in `onnxruntime==1.15.1`. And curiously, a much newer version is collected during package resolution:

```
 Collecting onnxruntime!=1.16.0,>=1.7.0 (from -r requirements.txt (line 41))
  Downloading onnxruntime-1.21.1-cp310-cp310-win_amd64.whl.metadata (4.9 kB)
```

You would think that the issue would still be fixed in newer versions of `onnxruntime`... but maybe the fix is specific to `1.15.1`? I've tried pinning this version just to see what will happen.

requirements.txt
joshuacwnewton: I tried installing SCT from this branch on an entirely fresh copy of Ubuntu 24.04 (in a VM), with the only steps taken beforehand being `sudo apt install git gcc`, and was able to reproduce the issue (with the exact same outputs as the original issue description).

I was curious why this didn't occur on my local machine, so I ran `ldd libqxcb.so | grep libxcb-xinerama.so.0` and got:

```
	libxcb-xinerama.so.0 => /lib/x86_64-linux-gnu/libxcb-xinerama.so.0 (0x00007d952ef22000)
```

I then tried to find out which package installed this for me, and got:

```console
> aptitude why libxcb-xinerama0
i   zoom Depends libxcb-xinerama0
```

Aha, Zoom. Ok! That explains it.

joshuacwnewton: OK, I was finally able to find a page that describes why `5.15` introduced this requirement: [Qt for X11 Requirements](https://doc.qt.io/qt-5/linux-requirements.html)

> Note: From Qt 5.15 onwards, Qt does require libxcb 1.11. Also, the `-qt-xcb` configure option got removed that was bundling some of the libs below. Anyhow, you can now configure with `-bundled-xcb-xinput` to avoid a dependency to system xcb-xinput.

The thing is, I'm not sure how this `-bundled-xcb-xinput` configuration flag will work with `PyQt5`, since this is a build configuration flag, and `PyQt5` is just a collection of Python bindings...

joshuacwnewton (commented):
## Python 3.13

@SomeoneInParticular shared a request to upgrade to Python 3.13 specifically, however I've encountered a big blocker: Upgrading `torch` to the necessary version would have the side effect of removing support for Intel Mac systems.

requirements.txt
joshuacwnewton, line 1: For Python 3.11/3.12 support, it seems that `ivadomed` is depending on some internal `torch` modules (`_six`) that were changed in future versions of torch:

```python
  File "/home/joshua/repos/spinalcordtoolbox/python/envs/venv_sct/lib/python3.12/site-packages/ivadomed/loader/utils.py", line 11, in <module>
    from torch._six import string_classes, int_classes
ModuleNotFoundError: No module named 'torch._six'
```

I'll be fixing these issues in a separate PR:

- https://github.com/ivadomed/ivadomed/pull/1319

joshuacwnewton, line 1: The biggest issue, however, with upgrading to Python 3.11/3.12 seems to be `voxelmorph`:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4899

I think until we address this issue, we are forced to use Python 3.10. :sob:

joshuacwnewton (commented):
## Python 3.11/3.12

This would be an intermediate step between Python 3.10 and Python 3.13. However, I've encountered another big blocker here: The unmaintained `voxelmorph` package.

requirements.txt
joshuacwnewton: Note: `LD_LIBRARY_PATH` actually came up in a different forum issue, where the fix for Qt issues might just be to [unset `LD_LIBRARY_PATH` entirely](https://forum.spinalcordmri.org/t/installing-sct-7-0-undefined-symbol-in-site-packages-pyqt5-qtcore-so/1436/9?u=joshuacwnewton).

If that's the case, then maybe bundling our own files and ignoring external libraries might just be the way to go.

I can try applying this environment setting in our `launcher.py` file just to see what breaks?

joshuacwnewton: My apologies, but while I was careful to begin with, I did end up doing the exact opposite in my most recent commits. :sweat_smile: 

Separately from the `libxcb` issue, a user on the forum had set their own `LD_LIBRARY_PATH` override that [interfered with the SCT installation](https://forum.spinalcordmri.org/t/installing-sct-7-0-undefined-symbol-in-site-packages-pyqt5-qtcore-so/1436/10?u=joshuacwnewton). So, I had thought that clearing out `LD_LIBRARY_PATH` in the context of `sct_` CLI commands specifically (i.e. only in the `launcher.py` context) would be beneficial. 

I was making the assumption that SCT would rather have an empty `LD_LIBRARY_PATH` to avoid conflicts, as opposed to allowing users to override via `LD_LIBRARY_PATH`. The latter felt very niche to me? But, I'm very open to walking back that change if you think that we should keep `LD_LIBRARY_PATH` untouched! :)

mguaypaq: I see! Well, wholesale replacement is probably fine in such a self-contained environment. If anyone is brave enough to mess with `LD_LIBRARY_PATH`, then surely they're brave enough to change the code where we do this override.

joshuacwnewton: That makes sense... :thinking: 

Maybe what we could do instead is check for the presence of `LD_LIBRARY_PATH` in `launcher.py` and emit a warning if it's set prior to running the CLI script? That way, users who have set `LD_LIBRARY_PATH` can unset it if it causes issues, but we otherwise leave it untouched as before.

joshuacwnewton:
I think that even though we were unable to get Python 3.11/3.12/3.13 working, this PR is in an okay state if we're alright with just switching to Python 3.10 only. :)

testing/cli/test_cli.py
```diff
@@ -36,9 +36,11 @@ def test_calling_scripts_with_no_args_shows_usage(capsys, script):
...
     assert b'usage' in completed_process.stderr
-    # NB: macOS GitHub Actions runners are inconsistent in their processing speed. Sometimes our requirement is met,
+    # NB: macOS/Windows GitHub Actions runners are inconsistent in their processing speed. Sometimes our requirement is met,
     #     and other times scripts take significantly longer. It's possible to just "retry" the GHA run, but to save
-    #     development headache, we just skip this check if we're on a macOS GitHub Actions runner.
-    if not (sys.platform.startswith("darwin") and "CI" in os.environ):
+    #     development headache, we just skip this check if we're on a macOS/Windows GitHub Actions runner.
+    # TODO: Revisit import profiling and see if there are new enhancements we can make to get scripts back under 2.0s
+    if not (sys.platform.startswith("darwin") or sys.platform.startswith("win32")
+            and "CI" in os.environ):
```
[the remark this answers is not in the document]
joshuacwnewton: Whoops! I thought I had done this, but I think the parentheses for continuing to the new line made it look like I had when I hadn't. Good catch. :hearts:

PR #4889, opened 2025-05-02
Improve how scaling is applied to QC reports for `rootlets`/`totalspineseg`

### LINKED ISSUE #4870: QC Report  for `totalspineseg` upscales the image too much causing a very long QC
### Description

QC generation crashed with totalspineseg 

### Steps to Reproduce

0. Install SCT ab2e0af5decd20c1819772c94965dc83ed19d480

1. get data from user [forum](https://forum.spinalcordmri.org/t/some-questions-about-the-registering-template-to-dmri-of-lumbar-enlargement-area/1426/8): `duke/temp/joshuacwnewton`

2. define: QC_FOLDER
 
3. run: sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1 -qc $QC_FOLDER**Expected behavior:** no crash

**Actual behavior:** crash:

```console
julien-macbook:~/Desktop/forum_data_dingdong971108/1/L/t2 $ sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1 -qc $QC_FOLDER
--
Spinal Cord Toolbox (git-master-ab2e0af5decd20c1819772c94965dc83ed19d480)

sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1 -qc ~/20250429_SCT_forum
--

Model 'totalspineseg' is up to date (Source: {'Dataset101_TotalSpineSeg_step1': 'https://github.com/neuropoly/totalspineseg/releases/download/r20241115/Dataset101_TotalSpineSeg_step1_r20241115.zip', 'Dataset102_TotalSpineSeg_step2': 'https://github.com/neuropoly/totalspineseg/releases/download/r20241115/Dataset102_TotalSpineSeg_step2_r20241115.zip'})
Creating temporary folder (/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7)
Copied t2.nii.gz to /var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/t2.nii.gz

Running TotalSpineSeg with the following parameters:
input = "/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/t2.nii.gz"
output = "/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/nnUNet_prediction"
iso = False
loc = "None"
suffix = ['']
loc_suffix = ""
step1_only = True
data_dir = "/Users/julien/code/spinalcordtoolbox/data/deepseg_models/totalspineseg"
max_workers = 1
max_workers_nnunet = 1
device = "cpu"


Making input dir with _0000 suffix:

Copying the input images to the input folder for processing:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.17s/it]

Preprocessing images:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.63s/it]

Reorienting images to LPI(-):
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.69s/it]

Resampling images to 1x1x1mm:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.69s/it]

Generating preview images for input:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.39s/it]

Running step 1 model:
perform_everything_on_device=True is only supported for cuda devices! Setting this to False
There are 1 cases in the source folder
I am process 0 out of 1 (max process ID is 0, we start counting with 0!)
There are 1 cases that I would like to predict

Predicting t2:
perform_everything_on_device: False
100%|██████████████████████████████████████████████████████████████████████████████████████████| 24/24 [00:50<00:00,  2.12s/it]
sending off prediction to background worker for resampling and export
done with t2

Generating preview images for step 1:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.56s/it]

Extracting the largest connected component:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.71s/it]

Using an iterative algorithm to label IVDs with the definite labels:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.68s/it]

Filling spinal canal label to include all non cord spinal canal:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.44s/it]

Transforming labels to input images space:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.42s/it]

Generating preview images for the step 1 labeled images:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.43s/it]

Generating preview images for the step 1 labeled images with tags:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.45s/it]

Extracting spinal canal soft segmentation from step 1 model output:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.67s/it]

Extracting spinal cord soft segmentation from step 1 model output:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.79s/it]

Removing the raw files from step 1 to save space...

Extracting the levels of the vertebrae and IVDs from the step 1 model output:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.78s/it]

Generating preview images for the step 1 levels:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.48s/it]

Resampling step1_output to the input images space:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.50s/it]

Resampling step1_cord to the input images space:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.59s/it]

Resampling step1_canal to the input images space:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.54s/it]

Resampling step1_levels to the input images space:
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:03<00:00,  3.26s/it]

Results of iterative labeling algorithm for step 1:
/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/nnUNet_prediction/step1_output

Spinal cord soft segmentations:
/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/nnUNet_prediction/step1_cord

Spinal canal soft segmentations:
/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/nnUNet_prediction/step1_canal

Single voxel in canal centerline at each intervertebral disc level:
/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-23-35_sct_deepseg_emrm7xz7/nnUNet_prediction/step1_levels

*** Generating Quality Control (QC) html report ***
Find the center of each slice
Creating temporary folder (/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-25-36_optic-detect-centerline_pkrh9g20)
Remove temporary files...
rm -rf /var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-04-29_10-25-36_optic-detect-centerline_pkrh9g20
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 537, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 512, in main
    qc2.sct_deepseg(
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/reports/qc2.py", line 260, in sct_deepseg
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/reports/qc2.py", line 454, in sct_deepseg_spinal_rootlets_t2w
    fig.savefig(img_path, format='png', transparent=True, dpi=300)
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/figure.py", line 3395, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backend_bases.py", line 2204, in print_figure
    result = print_method(
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backend_bases.py", line 2054, in <lambda>
    print_method = functools.wraps(meth)(lambda *args, **kwargs: meth(
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 496, in print_png
    self._print_pil(filename_or_obj, "png", pil_kwargs, metadata)
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 444, in _print_pil
    FigureCanvasAgg.draw(self)
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 382, in draw
    self.renderer = self.get_renderer()
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 397, in get_renderer
    self.renderer = RendererAgg(w, h, self.figure.dpi)
  File "/Users/julien/code/spinalcordtoolbox/python/envs/venv_sct/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 70, in __init__
    self._renderer = _RendererAgg(int(width), int(height), dpi)
ValueError: Image size of 1500x86466 pixels is too large. It must be less than 2^16 in each direction.
Total runtime; 128.404 seconds.
```

</details>

joshuacwnewton:
## Context

In PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4555, we modified the `mosaic()` function in the `qc2.py` file to add the ability to scale the mosaic up by an integer factor. In practice, we only used this for the "rootlets" QC, and it occurred whenever the image resolution was smaller than our "baseline" human resolution of 0.6mm (since a smaller resolution would use a smaller crop around the ROI, and thus we would compensate by scaling the whole thing up by 2.0x due to a call to `ceil()`).

However, what we didn't realize at the time was that our code was _already_ applying an implicit scale factor of 2.5x on top, because:

- The array size was capped at 600px.
- The output image size was fixed at (5" * 300DPI = 1500px).

This means that we were essentially duplicating the scaling logic, _and_ applying a much larger scaling than we realized. Because our new scaling was restricted to integer factors, small compensatory scaling values (~1.05x) would be rounded to 2.0x, which would then expand to _5.0x_!

As a consequence, we would end up with QC Reports with huge mosaics that could barely fit any columns in each row, and thus the QC would expand vertically to be _very_ long, sometimes even causing crashes (#4870).

## Description

This PR makes a number of small refactors to ensure that we no longer hardcode the magic number of 600px -- instead, the PR defines the 600px value as a function of target width (1500px) combined with the implicit "2.5x" scaling factor (now made explicit).

Then, it gets rid of the integer scale factor, merging that functionality with the existing 2.5x scaling. That way, we can scale up QCs more gradually (e.g. 2.6x, 2.7x, 3.0x, etc.), and avoid creating unnecessarily large (5.0x!) QC reports.

## Scope

> [!IMPORTANT]
> Assuming I've refactored things correctly, the QC reports for normal axial QCs should not change at all. Additionally, the QC reports for "1.0" scale factors won't change either. In fact, before I removed the call to `math.ceil`, even the `scale=5.0` QC report still worked exactly as it did before.
> 
> This PR should only have an impact on the QC reports that used `ceil()` to round up a ~1 integer scaling factor to ~2 (i.e. rootlets/totalspineseg QCs when the resolution was smaller than 0.6mm). Those QCs should now use a more reasonable scaling factor, as shown below.

Before (from source issue #4870):

https://github.com/user-attachments/assets/f81f00f8-59ba-40e7-9844-453e9cb870c2

After (on the same subject):

image

If we want to scale this down even more, it's as simple as changing the `scale` parameter (rather than messing around with inches, DPI, or the 600px magic number).

For example, for `scale=0.5` with labels/outlines removed, we get:

image

## Linked issues

Fixes #4870.


spinalcordtoolbox/reports/qc2.py
```diff
@@ -419,8 +429,10 @@ def sct_deepseg_spinal_rootlets_t2w(
     p_ratio = tuple(p_resample / p for p in p_original)
     radius = tuple(int(r * p) for r, p in zip(radius, p_ratio))
     # - One problem with this, however, is that if the crop radius ends up being smaller than the default, the QC will in turn be smaller as well.
-    #   So, to ensure that the QC is still readable, we scale up by an integer factor whenever the p_ratio is < 1
-    scale = int(math.ceil(1 / max(p_ratio)))  # e.g. 0.8mm human => p_ratio == 0.6/0.8 == 0.75; scale == 1/p_ratio == 1/0.75 == 1.33 => 2x scale
+    #   So, to ensure that the QC is still readable, we scale up whenever the p_ratio is < 1
+    scale = max((1 / ratio) for ratio in p_ratio)  # e.g. 0.8mm human => p_ratio == 0.6/0.8 == 0.75; scale == 1/p_ratio == 1/0.75 == 1.33
+    # - Note: `mosaic()` already has a default scaling factor of 2.5 (to help make the QC readable). So, we preserve it here.
+    scale *= 2.5
```
joshuacwnewton, line 435: @NathanMolinier I hope this `scale = 2.5` is a lot easier to understand and modify compared to the old code.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -779,8 +790,14 @@ def mosaic(img: Image, centers: np.ndarray, radius: tuple[int, int] = (15, 15),
 
     If `img` has N slices, then `centers` should have shape (N, 2).
     """
-    # Fit as many slices as possible in each row of 600 pixels
-    num_col = math.floor(600 / (2*radius[0]*scale))
+    # Note: This function used to hardcode a max row width of 600 pixels
+    #       In practice, because the canvas size is fixed to 1500 pixels, this
+    #       resulted in a permanent upscaling by 2.5x when saving the image.
+    #       To make things clearer, we now use a variable.
+    max_row_width = TARGET_WIDTH_PIXL / scale
```
joshuacwnewton, line 797: The main idea with making this change is that:

- Instead of scaling _up_ each mosaic image by an integer factor, we can
- Scale _down_ the array size using `scale`, keep the mosaic images a 1:1 pixel representation, and let `matplotlib` handle the upscaling.

NathanMolinier:
Hey @joshuacwnewton ! Thanks for taking care of this problem ! I did not test the PR yet, but looking at the code I have a few comments.

Is there a way to not include constants at all ? Find another way to have the most optimal shape ? (Wondering if we could save the individual slices individually and scale automatically in the web interface)

Also could we, not resample to 0.6mm ? Upsampling is just generating more slices that are not necessary. I would go with the native resolution or even downsampled...

spinalcordtoolbox/reports/qc2.py
```diff
@@ -44,6 +44,20 @@
...
+#   - This shouldn't ever need to be changed, since "target width" is related
+#     to the design of the QC report interface. If things need to be scaled,
+#     it should be done to the array itself before it is written to the .
+#   - `matplotlib` uses inches/DPI to define the canvas. But, given we want
+#     a fixed output size, we can choose some arbitrary values for both.
+#     Presumably, choosing a different DPI value (and thus a different canvas
+#     size in inches) wouldn't change the output image at all. (Is this true?)
+TARGET_WIDTH_PIXL = 1500
+DPI = 300
+TARGET_WIDTH_INCH = TARGET_WIDTH_PIXL / DPI
```
joshuacwnewton, line 59: > Is there a way to not include constants at all ? Find another way to have the most optimal shape ? (Wondering if we could save the slices individually and scale automatically in the web interface)

My use of constants here is just because the code was already using constants, but they were buried inside the code (and repeated in multiple places). So, I wanted to make it clearer that these values are the same across many QC functions.

I think it is definitely possible to switch to a more dynamic behavior, however would be a more radical change and would require a lot of testing and refactoring and etc.

Right now, my goal was just to allow us to update only the rootlets and totalspineseg `scale` factor to be something other than `2.5 * upscale`, which wasn't easily possible before.

joshuacwnewton: > Also could we, not resample to 0.6mm ? Upsampling is just generating bigger slices. I would go with the native resolution.

I think it is somewhat important that we keep the resampling fixed at **some** resolution, that way we have a reliable measure of mm-to-pixels for cropping purposes.

Truthfully, I don't know how the actual `0.6mm` value was chosen? All I know is that:

- Also, for axial mosaics, a resolution of 0.6mm and a crop radius of 15 (i.e. a 30x30 bounding box) has been used for many years.
- This implies that we probably decided at one point that 18mm x 18mm is a suitable bounding box around the spinal cord centerline for highlighting the cord.
- Since we also used to wrap at 600 pixels by default, maybe we decided that an Nx20 array was appropriate for axial mosaics.

It may be that this historical decision was just passed along as new QC reports were created, but now we have reached a point that 0.6mm is _not_ appropriate, since we are displaying much more than just the cord....

joshuacwnewton: Still, I think for the purpose of this PR, keeping most of the existing behavior is important, though discussing bigger changes could be an idea for the "QC Revamp" issue?

joshuacwnewton (commented):
@NathanMolinier I think your suggestions definitely sound reasonable? :slightly_smiling_face: 

Just to clarify, though: Because we wanted to backport this change to the v7.0 release (due to the crashing QC reports encountered by @jcohenadad when assisting a user on the forum), I wanted to make very minimal refactoring changes that would keep the behavior mostly the same as before, just avoiding the crashes.

spinalcordtoolbox/reports/qc2.py
NathanMolinier: Sounds reasonable !

NathanMolinier:
I got this when running on a cluster:
Screenshot 2025-05-07 at 14 34 42

And this on my laptop with macOS:
Screenshot 2025-05-07 at 14 35 52

NathanMolinier:
I played with some of you new parameters and got this QC when setting `TARGET_WIDTH_PIXL = 5000`.
Screenshot 2025-05-07 at 16 07 50

There are still some white borders, so ideally it would be cool to increase this value automatically. This looks definitely more like a QC that I would use.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Sorry. Just to be clear:

- Increasing `2.5` will increase the size of each individual square in the mosaic (fewer columns).
- _Decreasing_ `2.5` will decrease the size of the mosaic squares (meaning more columns can fit).

NathanMolinier: Oh yes, I discovered that when changing its value. Changing `TARGET_WIDTH_PIXL` allowed me to do what I wanted.

NathanMolinier:
> since we do not want to display 1:1 pixel accuracy, setting scale<1 is necessary

What you did also changed the size of each individual image (the discs and vertebrae segmentation are cropped too much here)

joshuacwnewton:
> What you did also changed the size of each individual image (the discs and vertebrae segmentation are cropped too much here)

I am pretty sure that changing the value of `scale` did not cause what you describe. I am working on a different subject, and for that subject, the crop radius has always been the same. See the images in: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4889#issue-3036482780

If the crop radius is an issue, that is a separate problem (and will require new changes to fix, e.g. increasing the crop radius for `totalspineseg`).

NathanMolinier:
About creating a QC for totalspineseg with sct_label_vertebrae to show a sagittal view of the spine. Because the method was designed to project all voxels in the same plane, some weird artifacts appear in the image when inputting a segmentation as seen in these images:
Screenshot 2025-05-08 at 11 20 31
Screenshot 2025-05-08 at 11 20 40

mguaypaq:
Conclusion from meeting: we should use the sagittal QC from sct_label_vertebrae (even though it has some glitches currently), but make sure to use the correct vertebra labels (make it a function parameter if possible, otherwise copy-paste-rename the whole function).

joshuacwnewton:
Thanks for the feedback! I guess we can close this PR for now, and instead try to replace the totalspineseg QC in a separate PR?

NathanMolinier:
Maybe you can merge it because your changes are stil relevant to help clarify the code ? And open another PR after ?

PR #4901, opened 2025-05-13
Limit `onnx<1.16.2` to fix CI failures on Windows runners (Python 3.9)

### LINKED ISSUE #4900: All CI runs are failing due to `onnx==1.18.0` causing installation issues on Windows
- May 12, 2025 (passing): https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14970539402/job/42050278905
- May 13, 2025 (failing): https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14994949475

Diff between dependency versions: https://www.diffchecker.com/7cD5tLSF/... looks like `onnx==1.18.0` is the culprit.

joshuacwnewton:
## Description

- `onnx` recently released version `1.18.0`, which breaks when imported by `monai` on Windows.
- However, versions `1.16.2` and `1.17.0` also had buggy behavior on Windows, but `1.17.0` kind of worked (as long as SCT itself never imported it), so it's been fine for a while now.
- Given the circumstances, though, I think now is a good time to also avoid these buggy versions, too.

Additional context: https://github.com/onnx/onnx/issues/6267#issuecomment-2877327002

## Linked issues

Fixes #4900.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_check_dependencies.py
+++ b/spinalcordtoolbox/scripts/sct_check_dependencies.py
@@ -73,7 +73,9 @@ def resolve_module(framework_name):
         # Importing `matplotlib_inline` requires IPython, but we don't install IPython (on purpose). This is because
         # `matplotlib_inline` is only needed to run SCT scripts in Jupyter notebooks, and IPython would already be
         # installed in the parent context. So, we map `matplotlib-inline` to None to skip import (which would fail).
-        'matplotlib-inline': (None, False)
+        'matplotlib-inline': (None, False),
+        # We specify the version of `onnx` on behalf of ivadomed -- we don't import it directly. So, skip checking its version.
+        'onnx': (None, False),
     }
 
     try:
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/scripts/sct_check_dependencies.py
```diff
@@ -73,7 +73,9 @@ def resolve_module(framework_name):
         # Importing `matplotlib_inline` requires IPython, but we don't install IPython (on purpose). This is because
         # `matplotlib_inline` is only needed to run SCT scripts in Jupyter notebooks, and IPython would already be
         # installed in the parent context. So, we map `matplotlib-inline` to None to skip import (which would fail).
-        'matplotlib-inline': (None, False)
+        'matplotlib-inline': (None, False),
+        # We specify the version of `onnx` on behalf of ivadomed -- we don't import it directly. So, skip checking its version.
+        'onnx': (None, False),
```
joshuacwnewton, line 78: It looks like `onnx==1.17.0` only works if we explicitly avoid importing `onnx` during `sct_check_dependencies` (see this skip here).

If we want `onnx` to work universally, we'd have to downgrade even further to `onnx==1.16.1` and below.

mguaypaq (approved):
Thanks for checking the different versions!

PR #4906, opened 2025-05-23
Add fixes for `PyQt5` and `onnxruntime` to amend our previous switch to Python 3.10

### LINKED ISSUE #4905: Recent upgrade to Python 3.10 is failing for Windows/Arch/Debian due to `PyQt5-Qt5` and `onnxruntime`
Looks like while the upgrade passed on PRs, it failed for Create Release ([Windows 2022](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15160532621/job/42652900978)) and Tests ([Arch/Debian](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15160471344/job/42652906998)).

I'll dig into this ASAP!

joshuacwnewton:
## Description

In PR #4869, we switched our conda environment from Python 3.9 to Python 3.10. While the tests passed on the PR, they failed for the `master`-only tests: #4905.

This PR tries to address 2 separate bugs from that issue:

- `PyQt5==5.15.x` has a finicky sister package `PyQt5-Qt5`, where only a single version is cross-platform.
- `onnxruntime` doesn't seem to have a single package version that works on both Arch/Debian _and_ Windows on Python 3.10.

To address the former, we pin to the single working version, and subsequently update our hacky check of PyQt5 versions.

To address the latter, well... we would either need to wait for a response to https://github.com/microsoft/onnxruntime/issues/24843#issuecomment-2904854046, or upgrade to Python 3.11. One (or both) are likely to happen within this release cycle, so for now, I'm prioritizing Windows and temporarily silencing the tests for Arch/Debian, see:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15214451356

## Linked issues

Fixes #4905.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,9 @@ jobs:
     # TODO: when actions supports using ${{env}} in job.*.if and not just job.*.steps.*.if, use this:
     #if: ${{ env.NIGHTLY }}
     # in the meantime, copy-paste this:
-    if: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+    if: false  # ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
+    # FIXME: Temporarily disabled due to issues with `onnxruntime==1.15.1`
+    #        See: https://github.com/microsoft/onnxruntime/issues/24843#issuecomment-2904854046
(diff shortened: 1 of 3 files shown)


mguaypaq (approved):
Thanks for the investigation, and for this sensible temporary solution. Ah, dependencies!

PR #4909, opened 2025-05-29
Filter out `Version` and `CodeURL` when diffing JSON sidecars

### LINKED ISSUE #4886: "Compare batch_processing results" CI job is throwing false positives due to JSON sidecar files
Sample failing run: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/14795954840/job/41545535770?pr=4862#step:5:41

This is distracting on PRs and takes away the utility of the comparison (highlighting when files actually change)

I assume that the JSON sidecar files encode the branch/SCT version, so I will need to filter out this information in much the same way we do for CSV files when computing the filesize/hash.

joshuacwnewton:
## Description

In https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4759, I added a highly sensitive test for `batch_processing.sh` that outputs warnings whenever file properties differ between `master` and PRs.

Because the test checks against filesize and md5 hashes, any sort of branch-specific metadata will throw false-positives. We already encountered this with output `.csv` files that report the SCT version (fixed in #4803), but we didn't catch the output `.json` sidecar files that also report the SCT version.

So, this PR mimics our CSV filtering to also filter JSON sidecar files, which should eliminate the remaining 12 false positives currently reported in PR warnings. From now on, if there are warnings on PRs, it will actually mean something. :)

Screenshot from 2025-05-29 15-15-07

## Linked issues

Fixes #4886.

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -37,6 +37,9 @@ jobs:
     # Add any file patterns you want ignored to 'check_url_blacklist.txt'
     - name: Check for broken links
       run: |
+        # Fail outright if any failure occurs, rather than attempting to proceed
+        set -eu -o pipefail
+
         # Get a list of files which have been changed from master on this branch
         git_diff_files=$(git diff origin/master... --name-only)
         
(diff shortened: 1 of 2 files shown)


SomeoneInParticular:
Merged in the (hopefully) fixed URL checking code to confirm it was just that breaking things. :crossed_fingers:

mguaypaq (approved):
Seems fine! Thanks for getting rid of these false positives.

PR #4857, opened 2025-04-15
Add Apptainer installation for use on HPC systems

SomeoneInParticular:
#### GitHub

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/onboarding/software-development.html#pr-labels) to this PR
- [x] I've applied a [release milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/milestones) (major, minor, patch) in line with [Semantic Versioning guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Misc%3A-Creating-a-new-release#convention-for-naming-releases) 
- [x] I've assigned a reviewer

#### PR contents

- [x] I've consulted [SCT's internal developer documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki) to ensure my contribution is in line with any relevant design decisions
- [ ] I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution
- [x] I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages

## Description

This PR adds some utility scripts which enable you to create an Apptainer instance with SCT inside of it, allowing SCT to be run on HPC systems that block/discourage the use of native installs. This was developed specifically to work with Alliance Canada, which disallows the native Conda installation SCT relies on to install itself natively.

## Linked issues
Resolves #4593

PR #4910, opened 2025-05-30
Fix `check-url` Github CI failing when no new URLs are present

SomeoneInParticular:
Mini-PR to fix a bug spotted by @joshuacwnewton, as described [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4857#pullrequestreview-2879399588).

Wasn't caught during initial testing, as the bug crops up only when a GitHub PR is made which happens to edit files without any URLs in it.

### DIFF
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -37,9 +37,6 @@ jobs:
     # Add any file patterns you want ignored to 'check_url_blacklist.txt'
     - name: Check for broken links
       run: |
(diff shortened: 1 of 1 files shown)


joshuacwnewton (changes requested):
Thanks for the clean up! I like that this has evolved away from "magical pipeline incantation" to a set of single-command, readable, debuggable steps. :hearts:

I only have the one small comment, but apart from that, testing it locally worked like a charm for me.

PR #4917, opened 2025-06-04
Use `Path.resolve()` instead of `Path.absolute()`

### LINKED ISSUE #4916: Error when using relative paths with `sct_qc`
### Description

I wanted to QC segmentations from the inside of a BIDS folder (current directory --> `whole-spine/derivatives`), so I used this following bash loop:

```
for sub in $(find labels_refined/ -type f -name '*.nii.gz' | grep -o 'sub-[^_/]*' | uniq);do file=../"$sub"/anat/"$sub"_T2w.nii.gz; seg=labels_refined/"$sub"/anat/"$sub"_T2w_label-spine_dseg.nii.gz; sct_qc -i "$file" -s "$seg" -p sct_label_vertebrae -qc qc_spine; done
```

The command ran fine but after, I discovered that folders for every subjects were created in addition to the `qc_spine` folder:

Example of a subject folder:
```
sub-unfErssm001
└── anat
    └── sct_label_vertebrae
        └── 2025_06_03_162930.809856
            ├── background_img.png
            └── overlay_img.png
```

qc_spine:
```
qc_spine/
├── _assets
│   └── py
├── images
│   └── key-icon.png
├── index.html
├── js
│   └── datasets.js
├── _json
│   ├── qc_2025_06_03_150042.473853.json
│   ├── qc_2025_06_03_150046.068278.json
│   ├── qc_2025_06_03_150049.838355.json
│   ├── qc_2025_06_03_150053.299265.json
│   ├── qc_2025_06_03_150057.454461.json
│   ├── qc_2025_06_03_150101.335200.json
│   ├── qc_2025_06_03_150105.283767.json
│   ├── qc_2025_06_03_150109.174518.json
│   ├── qc_2025_06_03_150113.327495.json
│   ├── qc_2025_06_03_150117.033206.json
│   ├── qc_2025_06_03_150120.979456.json
│   ├── qc_2025_06_03_150124.547725.json
│   ├── qc_2025_06_03_150128.326533.json
│   ├── qc_2025_06_03_150131.669011.json
│   ├── qc_2025_06_03_150135.176256.json
│   ├── qc_2025_06_03_150138.685950.json
│   ├── qc_2025_06_03_150142.143308.json
│   ├── qc_2025_06_03_150145.522652.json
│   ├── qc_2025_06_03_150149.181573.json
│   ├── qc_2025_06_03_150153.241299.json
│   ├── qc_2025_06_03_150156.692478.json
│   ├── qc_2025_06_03_150200.374356.json
│   ├── qc_2025_06_03_150204.162269.json
│   ├── qc_2025_06_03_150207.723525.json
│   └── qc_2025_06_03_150211.200211.json
└── py
    └── refresh_qc_entries.py
```

To solve the issue, I had to use an absolute path instead of the relative path with `../`

mguaypaq:
Unlike `os.path.abspath()`, `pathlib.Path.absolute()` does _not_ remove `..` components from the path. This is unexpected, and causes subtle problems.

In contrast, `pathlib.Path.resolve()` removes `..` components (and also de-references symlinks along the path, which may lead to different surprises, but less likely).

`git grep 'absolute('` only found these instances to fix, and some unrelated uses of `np.absolute()`.

Fixes #4916.

### DIFF
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -603,7 +603,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None
     # Get the aspect ratio (height/width) based on pixel size. Consider only the first 2 slices.
     qc_image.aspect_img, qc_image.aspect_mask = qcslice.aspect()[:2]
 
-    path_input = Path(fname_in1).absolute()
+    path_input = Path(fname_in1).resolve()
     path_qc = Path(path_qc)
     command = process
     cmdline = [command]
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -165,7 +165,7 @@ def sct_register_multimodal(
 
     # Axial orientation, switch between two input images
     with create_qc_entry(
-        path_input=Path(fname_input).absolute(),
+        path_input=Path(fname_input).resolve(),
         path_qc=Path(path_qc),
         command=command,
         cmdline=list2cmdline(cmdline),
@@ -255,7 +255,7 @@ def sct_deepseg(
     cmdline.extend(argv)
 
     with create_qc_entry(
-        path_input=Path(fname_input).absolute(),
+        path_input=Path(fname_input).resolve(),
         path_qc=Path(path_qc),
         command=command,
         cmdline=list2cmdline(cmdline),
@@ -646,7 +646,7 @@ def sct_analyze_lesion(
 
     # Axial orientation, switch between one anat image and 1-2 seg images
     with create_qc_entry(
-        path_input=Path(fname_input).absolute(),
+        path_input=Path(fname_input).resolve(),
         path_qc=Path(path_qc),
         command=command,
         cmdline=list2cmdline(cmdline),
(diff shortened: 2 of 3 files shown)


mguaypaq:
I don't think adding a test for this is particularly relevant, because it wouldn't catch any _new_ uses of `Path.absolute()` in the future. The right solution would be to have some static code analysis linter to prevent uses of `Path.absolute()` throughout SCT, but I don't know how we'd do that.

joshuacwnewton (approved):
W.r.t static linters, I imagine the same sort of logic could apply to all of the other "anti-patterns to avoid" we've had sitting around in our Wiki over the years, e.g. [Moving away from pyplot](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming%3A-Moving-away-from-pyplot)

I think I agree that automatically checking for this issue is a bit tricky, since it needs to be explicitly `pathlib.Path.absolute()` as opposed to any `absolutepath()`-esque methods. 

I think I'm OK just sticking a new page in the Wiki and hoping that the institutional knowledge sticks around for reviews on this project. Custom linting rules sound fun to build, but maybe not the best use of our time? It's hard to say though. :thinking:

mguaypaq:
I put a note about this in https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Programming:-Useful-file-tools-in-Python, and yes hopefully we remember this corner case in the future.

Issue #235, opened 2014-12-09
give the possibility to register only using segmentation (hence, do not use slicereg)


jcohenadad:

benjamindeleener:
already addressed in another issue

Issue #234, opened 2014-12-09
when using sliceReg, we should not allow to add padding (-z option)


benjamindeleener:
Reason: sliceReg does not work with padding

Issue #231, opened 2014-12-08
remove "using info image" from sct_propseg


benjamindeleener:

benjamindeleener:
Will be fixed when merging branch bdl_propseg

Issue #24, opened 2014-07-03
VM NeuroDebian


jcohenadad:
- cannot import appliance. Error: "Implementation of the USB 2.0 controller not found!"
  - made it working by unchecking "unable USB 2.0"
- webbrowser should have SF website in favs
- webbrowser: currently opens with google in German (???)
- current shared folder settings is: "C:\Users\benja_000\VM_folder_shared"

PR #4927, opened 2025-06-10
Mask output of `sct_apply_transfo` when applying `-initwarp` during multimodal registration

### LINKED ISSUE #4876: Duplicate image appears when applying a small FOV warping field to an image with a larger FOV
### Description

While working on #4871, I noticed that applying the warping field to another image produced an unexpected behaviour: two images appear on the field of view:

Image

It seems like the _wrong_ second image originates from other slices, that are not in the FOV of the destination image (in this case the dmri). I'm wondering if this is related to the extension of the warping field in the rostral direction as reported in:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4884

### Steps to Reproduce

```bash
# The syntax is compatible with SCT 7.0
# Define QC_FOLDER
QC_FOLDER="~/20250429_SCT_forum"
# Process t2 data
cd t2
# Segment the spinal cord
sct_deepseg spinalcord -i t2.nii.gz -qc $QC_FOLDER
# Label vertebrae
sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1
sct_qc -i t2.nii.gz -s t2_step1_levels.nii.gz -p sct_label_utils -qc $QC_FOLDER
# Register to template
# Note: for step=1 we use type=seg instead of type=imseg because of a bug (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4873)
sct_register_to_template -i t2.nii.gz -s t2_seg.nii.gz -ldisc t2_step1_levels.nii.gz -c t2 -qc $QC_FOLDER -param step=1,type=seg,algo=centermassrot:step=2,type=seg,metric=MeanSquares,algo=bsplinesyn,slicewise=1,iter=5,smooth=1:step=3,type=im,metric=CC,algo=syn,slicewise=1,iter=3
# Warp the PAM50 template
sct_warp_template -d t2.nii.gz -w warp_template2anat.nii.gz -a 0 -qc $QC_FOLDER
# Go to dmri folder
cd ../dmri
# generate mask for moco
sct_dmri_separate_b0_and_dwi -i dmri.nii.gz -bvec bvecs.txt
# Segment the spinal cord
sct_deepseg spinalcord -i dmri_dwi_mean.nii.gz -qc $QC_FOLDER
# Create spinal cord mask
sct_create_mask -i dmri_dwi_mean.nii.gz -p centerline,dmri_dwi_mean_seg.nii.gz -size 35mm
# Apply motion correction
sct_dmri_moco -i dmri.nii.gz -m mask_dmri_dwi_mean.nii.gz -bvec bvecs.txt -qc $QC_FOLDER -qc-seg dmri_dwi_mean_seg.nii.gz
# Create "good" spinal cord mask on moco'd image
sct_deepseg spinalcord -i dmri_moco_dwi_mean.nii.gz -qc $QC_FOLDER
# Register dmri data to the template (via the t2 data)
sct_register_multimodal -i "${SCT_DIR}/data/PAM50/template/PAM50_t1.nii.gz" -iseg "${SCT_DIR}/data/PAM50/template/PAM50_cord.nii.gz" -d dmri_moco_dwi_mean.nii.gz -dseg dmri_moco_dwi_mean_seg.nii.gz -initwarp ../t2/warp_template2anat.nii.gz -initwarpinv ../t2/warp_anat2template.nii.gz -owarp warp_template2dmri.nii.gz -owarpinv warp_dmri2template.nii.gz -param step=1,type=seg,algo=centermass:step=2,type=seg,algo=bsplinesyn,slicewise=1,iter=3 -qc $QC_FOLDER -r 0
```

joshuacwnewton:
## Checklist

## Context

Registration (e.g. `sct_register_to_template`) will produce warping fields that map the voxels in one coordinate space (e.g. the PAM50 template space) onto the space of another image. 

If a warping field is originally determined using an image with one FOV (e.g. a T2 image with dimensions (40, 515, 512)), but then the warping field is applied to an image with a different FOV (e.g. a dMRI image with dimensions (256, 256, 13)), then the ANTs tool `antsApplyTransforms` may hallucinate phantom copies of the image in areas where the initial warping field was not valid.

<details>
<summary> Warping field FOV (blue) vs. dMRI FOV (red) </summary>

image

</details>

<details>
<summary> Hallucinated copy outside of the warping field FOV </summary>

image

</details>

When would a warping field be applied to an image with a different FOV? Well, in fact, we do it all the time when we specify the `-initwarp` and `-initwarpinv` arguments [as a "pre-registration" step](https://spinalcordtoolbox.com/stable/user_section/tutorials/diffusion-weighted-mri/template-registration-for-dmri.html#registering-the-template-to-the-dti-space). 

(Side note: More often than not, this will affect `-initwarp` specifically, because the template -> anat will have different anat FOVs. `-initwarpinv` is spared, because the anat -> template uses the same destination space (the template) in both cases. However, in the case of e.g. multimodal non-template registration, `-ref template`, etc. this assumption may not hold.)

## Description

To mitigate this issue, because we have the warping field, it's possible to mask out any output voxels beyond the warping field FOV (considering them "invalid"), which removes this phantom copy:

<details>
<summary> Output warped image with phantom copy masked out </summary>

image

</details> 

And as it turns out, `sct_apply_transfo` has an option called `-crop` which was already designed to perform this exact masking step, added all the way back in 2015. However, `-crop` is broken in several ways:

- It uses the wrong mask when cropping the image (data dimensions, rather than warping field dimensions)
- It uses a rectangular bounding box derived from the min/max coords of the warping field. This acts as a sort of [convex hull](https://scikit-image.org/docs/0.24.x/auto_examples/edges/plot_convex_hull.html), covering an area that's a little larger than the warping field itself, which may be too large and not mask out enough voxels.

To address these two issues, this PR adds the following fixes:

- Use the correct dimensions when creating the mask to use for `-crop`.
- Adds a new option to the `-crop` flag (`-crop 3`) to mask out voxels instead of just cropping with a rectangular bbox.

While `-crop 3` will solve the issue for user calls to `sct_apply_transfo`, it won't actually fix the problem in any scripts that internally call `sct_apply_transfo`. Because this issue is commonly seen inside `sct_register_multimodal` with `-initwarp`, I decided to turn on this flag when `-initwarp` is used (e.g. `step=1`).

I don't think changing this default should have any effect unless these phantom copies were present, and even then, the effect should only ever be positive. This should hopefully be confirmed by the fact that the `batch_processing` results should not change despite the use of `-initwarp`.

## Alternative approaches

In theory, we could crop the destination image to the same extents as the warping field instead (_before_ we apply `antsApplyTransform`, rather than after). Then, we would update the dimensions of the cropped output image to match the full FOV of the destination image. This would be more thorough and likely prevent phantom copies from appearing at all, which might be relevant in situations where the phantom copy [overlaps](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4876#issuecomment-2959955246) the true copy.

Note: The destination image would have to be truly cropped (i.e. changing the dimensions) and not just masking the voxels to `0`, since `antsApplyTransform` only cares about the destination FOV and not the contents of the destination image.

However, this is a more complicated approach that requires some care, because cropping the image changes the origin of the array (and thus the affine). I couldn't quite wrap my head around how to do this properly. (Something like generating the warping field mask, passing it to `sct_crop_image -mask`, running `antsApplyTransforms` on the cropped image, then... well, I'm not sure how to fix the dimensions of the output image. I thought about combining it with a `zeros_like` image, but `sct_maths -add` imposes a "same dimensions" restriction.)

Because of this, I stuck with the simpler approach of masking the output instead, which seems to work [well enough](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4876#issuecomment-2966803339) on the "overlapping" phantom copy cases.

## Linked issues

Fixes #4876.

### DIFF
--- a/spinalcordtoolbox/registration/core.py
+++ b/spinalcordtoolbox/registration/core.py
@@ -177,6 +177,7 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
                     '-d', dest[ifile],
                     '-o', add_suffix(src[ifile], '_reg'),
                     '-x', interp_step[ifile],
+                    '-crop', '3' if (i_step == 1 and fname_initwarp) else '0',
                     '-v', '0',
                     '-w'] + warp_forward
                 )
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -65,10 +65,10 @@ def get_parser():
         default=[])
     optional.add_argument(
         "-crop",
-        help="Crop Reference. 0: no reference, 1: sets background to 0, 2: use normal background.",
+        help="Crop Reference. 0: no reference, 1: sets background to 0, 2: use normal background, 3: mask the image.",
         type=int,
         default=0,
-        choices=(0, 1, 2))
+        choices=(0, 1, 2, 3))
     optional.add_argument(
         "-o",
         help='Filename to use for the output image (i.e. the transformed image). Example: `out.nii.gz`',
@@ -316,24 +316,31 @@ class Transform:
         warping_field = fname_warp_list_invert[-1]
         # If the last transformation is not an affine transfo, we need to compute the matrix space of the concatenated
         # warping field
-        if not isLastAffine and crop_reference in [1, 2]:
+        if not isLastAffine and crop_reference in [1, 2, 3]:
             printv('Last transformation is not affine.')
+            img_out = Image(fname_out)
+            # Extract only the first n dims of the warping field by creating a dummy image with the correct shape
+            img_warp = Image(warping_field)
+            warp_shape = img_warp.data.shape[:int(dim)]  # dim = {'2', '3', '4'}
+            img_warp_ndim = Image(np.ones(warp_shape), hdr=img_warp.hdr)
             if crop_reference in [1, 2]:
-                # Extract only the first n dims of the warping field by creating a dummy image with the correct shape
-                img_warp = Image(warping_field)
-                warp_shape = img_warp.data.shape[:int(dim)]  # dim = {'2', '3', '4'}
-                img_warp_ndim = Image(np.ones(warp_shape), hdr=img_warp.hdr)
                 # Set zero to everything outside the warping field
-                cropper = ImageCropper(Image(fname_out))
+                cropper = ImageCropper(img_out)
                 cropper.get_bbox_from_ref(img_warp_ndim)
-                if crop_reference == 1:
+                if crop_reference in 1:
                     printv('Cropping strategy is: keep same matrix size, put 0 everywhere around warping field')
                     img_out = cropper.crop(background=0)
                 elif crop_reference == 2:
                     printv('Cropping strategy is: crop around warping field (the size of warping field will '
                            'change)')
                     img_out = cropper.crop()
-                img_out.save(fname_out)
+            elif crop_reference == 3:
+                # Resample the warping field mask (in reference coordinates) into the space of the image to be cropped
+                from spinalcordtoolbox.resampling import resample_nib
+                img_ref_r = resample_nib(img_warp_ndim, image_dest=img_out, interpolation='nn', mode='constant')
+                # Simply mask the output image instead of doing a bounding-box-based crop
+                img_out.data = img_out.data * img_ref_r.data
+            img_out.save(fname_out)
 
 
 # MAIN
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -335,7 +335,7 @@ class Transform:
                 # Set zero to everything outside the warping field
                 cropper = ImageCropper(img_out)
                 cropper.get_bbox_from_ref(img_warp_ndim)
-                if crop_reference in 1:
+                if crop_reference == 1:
                     printv('Cropping strategy is: keep same matrix size, put 0 everywhere around warping field')
                     img_out = cropper.crop(background=0)
                 elif crop_reference == 2:
--- a/spinalcordtoolbox/scripts/sct_apply_transfo.py
+++ b/spinalcordtoolbox/scripts/sct_apply_transfo.py
@@ -20,6 +20,7 @@ from spinalcordtoolbox.image import Image, generate_output_file, add_suffix
 from spinalcordtoolbox.cropping import ImageCropper
 from spinalcordtoolbox.math import dilate
 from spinalcordtoolbox.labels import cubic_to_point
+from spinalcordtoolbox.resampling import resample_nib
 from spinalcordtoolbox.utils.shell import SCTArgumentParser, Metavar, get_interpolation, display_viewer_syntax
 from spinalcordtoolbox.utils.sys import init_sct, run_proc, printv, set_loglevel
 from spinalcordtoolbox.utils.fs import tmp_create, rmtree, extract_fname, copy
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -314,17 +324,16 @@ def apply(self):
...
+            img_warp = Image(warping_field)
+            warp_shape = img_warp.data.shape[:int(dim)]  # dim = {'2', '3', '4'}
+            img_warp_ndim = Image(np.ones(warp_shape), hdr=img_warp.hdr)
             if crop_reference in [1, 2]:
-                # Extract only the first ndim of the warping field
-                img_warp = Image(warping_field)
-                if dim == '2':
-                    img_warp_ndim = Image(img_src.data[:, :], hdr=img_warp.hdr)
-                elif dim in ['3', '4']:
-                    img_warp_ndim = Image(img_src.data[:, :, :], hdr=img_warp.hdr)
```
joshuacwnewton, line 325: Note that this change here is a small refactor of the logic, since instead of trying to extract a relevant slice from `img` using `[:, :]` or `[:, :, :]` (which doesn't even work), we can just make a `ones` image with the dimensions of our choice, since all we're using `ImageCropper` for is masking.

spinalcordtoolbox/registration/core.py
```diff
@@ -177,6 +177,7 @@ def register_wrapper(fname_src, fname_dest, param, paramregmulti, fname_src_seg=
                     '-d', dest[ifile],
                     '-o', add_suffix(src[ifile], '_reg'),
                     '-x', interp_step[ifile],
+                    '-crop', '3' if (i_step == 1 and fname_initwarp) else '0',
```
mguaypaq, line 180: I was going to suggest `fname_initwarp is not None` for clarity, but I see that it could also default to `''`, so this is correct as written.

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -314,17 +324,16 @@ def apply(self):
         warping_field = fname_warp_list_invert[-1]
         # If the last transformation is not an affine transfo, we need to compute the matrix space of the concatenated
         # warping field
-        if not isLastAffine and crop_reference in [1, 2]:
+        if not isLastAffine and crop_reference in [1, 2, 3]:
```
mguaypaq, line 328: I'm a bit puzzled why cropping is silently ignored when `isLastAffine == True`, but I think this is outside the scope of this PR.

spinalcordtoolbox/scripts/sct_apply_transfo.py
mguaypaq: Yes, this seems much better.

spinalcordtoolbox/scripts/sct_apply_transfo.py
```diff
@@ -333,7 +342,13 @@ def apply(self):
                     printv('Cropping strategy is: crop around warping field (the size of warping field will '
                            'change)')
                     img_out = cropper.crop()
-                img_out.save(fname_out)
+            elif crop_reference == 3:
+                # Resample the warping field mask (in reference coordinates) into the space of the image to be cropped
+                from spinalcordtoolbox.resampling import resample_nib
```
mguaypaq: Is this imported here (rather than at the top) to prevent a circular import? Or for performance reasons?

joshuacwnewton: Ah shoot, thank you for catching this. Truthfully, for my prototype, I was just mimicking what already occurs inside of the cropping module:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d151cc390a088bbc3251cc1a3086abd602b24dfd/spinalcordtoolbox/cropping.py#L120-L137

But I had no personal reasons to do so, nor do I know why the cropping code does so. I can take a look to see if it's fine to move it up without taking a performance hit. :)

joshuacwnewton: Tada! All good.

image

Issue #4871, opened 2025-04-29
Investigate better pipeline for processing user's lumbar data (experimental)


jcohenadad:
Based on data from a [user](https://forum.spinalcordmri.org/t/some-questions-about-the-registering-template-to-dmri-of-lumbar-enlargement-area/1426/1), I observed strange behaviours of our regular registration pipeline. I'm opening this issue as I debug this case.

I'm using SCT ab2e0af5decd20c1819772c94965dc83ed19d480.
Data: `duke:temp/joshuacwnewton/forum_data_dingdong971108`

I ran the registration using the disc labels (more than 2) output by totalspineseg:

<img width="894" alt="Image" src="https://github.com/user-attachments/assets/3d812605-bbfe-402e-b151-4e663f4c578c" />

The first observation, is that the default parameters yield pretty bad registration:

![Image](https://github.com/user-attachments/assets/3042daad-db9a-45ea-8fe5-e79ec2eeadb5)

I checked the spinal cord segmentation, and it is _not_ the culprit.

next, I re-run the syntax with `-r 0` so I can check the intermediate steps. Interesting observations:

The `data_rpi_1mm_straightAffine.nii` and `data_rpi_1mm_straightAffine_crop.nii` have a strange replication of the top slice:

<img width="1275" alt="Image" src="https://github.com/user-attachments/assets/31ff74e1-d91f-48cb-af94-f4b4bd2e83ad" />

This could be due to the extrapolation associated with the use of more than 2 discs.

More problematic, is the label projection on the spinal cord centerline, which is clearly wrong. See these two labels, misplaced:

![Image](https://github.com/user-attachments/assets/582f0b3e-dda2-4ed2-8b3c-439e9186c57b)

Also note that the correct label value is 21 (so 21 is properly project, but not 25, which comes from nowhere):

<img width="930" alt="Image" src="https://github.com/user-attachments/assets/f391076b-44c9-4f73-88fe-ee65296ecceb" />

jcohenadad:
Attempting to remove the problematic label by stopping halfway into `sct_register_to_template`, manually removing the label in FSLeyes, saving the file, and continuing to run the script. Now only these tree labels remain:

```console
[...]

PR #4932, opened 2025-06-13
Prevent repeated slices during straightening by correctly zeroing out duplicate slices

### LINKED ISSUE #4881: Debug user's pipeline for registering dMRI lumbar region to PAM50 template
Earlier this week, we began debugging a user's [forum post](https://forum.spinalcordmri.org/t/some-questions-about-the-registering-template-to-dmri-of-lumbar-enlargement-area/1426) for a registration pipeline on T2 + dMRI lumbar data.

- Originally, I tried the user's commands verbatim (and identified a potential root cause of the issue -- [misalignment of the original T2 and dMRI images](https://forum.spinalcordmri.org/t/some-questions-about-the-registering-template-to-dmri-of-lumbar-enlargement-area/1426/8?u=joshuacwnewton)). 
- However, @jcohenadad [experimented](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4871) with some modified commands, and came up with [a working pipeline](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4881#issuecomment-2845086510).

Separately, Julien's experiments have revealed a cascade of issues related to the new commands. Because of this, I will try to use GitHub's "meta issues" feature to try to keep track of the various sub-tasks separately from the "working pipeline" we end up providing the user.

I've also included the QC Report issues that were discovered in the process, since these are a blocker for Julien sharing his solution with the user.

### LINKED ISSUE #4884: Lumbar registration with 3+ labels results in strange replication of top slice
As mentioned by @jcohenadad in his investigation in issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4871:

> I ran the registration using the disc labels (more than 2) output by totalspineseg [and] with `-r 0` so I can check the intermediate steps. Interesting observations:

> The `data_rpi_1mm_straightAffine.nii` and `data_rpi_1mm_straightAffine_crop.nii` have a strange replication of the top slice:
>
> Image
> 
> This could be due to the extrapolation associated with the use of more than 2 discs.

joshuacwnewton:
## Checklist

- [X] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)

## Description

This PR attempts to fix issue #4884, where the output of `sct_straighten_spinalcord` would contain repeated slices in regions that should be undefined instead.

<details>

image

</details>

The cause of this bug is due to our "slice deduplication" logic, found here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/48e2489dac81ba0c3d70c068d6f8df094c56da3b/spinalcordtoolbox/straightening.py#L392-L401

For reasons that are unknown to me, this code stops scanning halfway through the image (`// 2`). Meaning, if the duplicated slices go _past_ the halfway point of the destination image, they will not be deduplicated.

If I remove this stopping point, then all of the slices will be deduplicated, and the output will not contain any repeated slices:

image

## Alternative solution

> [!WARNING]
> Here be dragons. This part of the PR attempts to fix some older SCT code (~2019 or earlier) that has silently been doing nothing for many years.

The straightening code also contains the notion of a "safe zone", i.e. the range of slices covered by the spinal cord segmentation. In theory, if this safe zone is applied to the warping field, we can achieve the same result.

However, the old safe zone code was very broken (and not even applied at all). So, even though I believe I've fixed this code, applying it by default during registration would be a more radical change. So, I opted to keep this behavior off by default.

Even though I haven't applied the safe zone during registration, I've still exposed it as an "advanced parameter" option for `sct_straighten_spinalcord`, which can in theory be opted-into by users, even though this script isn't used all that often on its own.

## Testing this PR

This PR was structured to be understood commit by commit, so please feel free to read each one when reviewing.

This PR can be tested using the following commands:

```bash
# Common QC folder
PATH_QC=qc
# Segment the spinal cord
sct_deepseg spinalcord -i t2.nii.gz -qc $PATH_QC
# Label vertebrae
sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1
# Remove the labels that are not present in the PAM50 template (#4872)
sct_label_utils -i t2_step1_levels.nii.gz -remove 22,23,24,25
mv t2_step1_levels.nii.gz t2_step1_levels_22_25.nii.gz
mv labels.nii.gz t2_step1_levels.nii.gz
# Register to template
# Note: for step=1 we use type=seg instead of type=imseg because of a bug (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4873)
sct_register_to_template -i t2.nii.gz -s t2_seg.nii.gz -ldisc t2_step1_levels.nii.gz -c t2 -qc $PATH_QC -param step=1,type=seg,algo=centermassrot:step=2,type=seg,metric=MeanSquares,algo=bsplinesyn,slicewise=1,iter=5,smooth=1:step=3,type=im,metric=CC,algo=syn,slicewise=1,iter=3 -r 0
```

You can replace the template registration command with this `sct_straighten_spinalcord` command instead if you'd like to view the straightening results more directly, since it mimics the call that occurs inside of `sct_register_to_template`:

```
sct_straighten_spinalcord -i t2.nii.gz -s t2_seg.nii.gz \
                          -dest "${SCT_DIR}/data/PAM50/template/PAM50_cord.nii.gz" \
                          -ldisc-input t2_step1_levels.nii.gz \
                          -ldisc-dest "${SCT_DIR}/data/PAM50/template/PAM50_label_disc.nii.gz" \
                          -param safe_zone=1
```

The data for testing is the same found in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4871: `duke:temp/joshuacwnewton/forum_data_dingdong971108`

## Linked issues

Fixes #4884.
Fixes #4881. (Assuming this is the last of the sub-issues to be completed.)


spinalcordtoolbox/straightening.py
```diff
@@ -485,13 +484,34 @@ def straighten(self):
...
             [[0, 0, bound_straight[0]]]), image_centerline_straight.transfo_phys2pix([[0, 0, bound_straight[1]]])
 
-        if radius_safe > 0:
+        # note: we need to ensure that straight bounds don't go outside the warping field's z-dimension, since
+        #       `transfo_phys2pix` has no safeguards, and could return negative or too large indices.
+        coord_bound_straight_inf[0][2] = max(0, coord_bound_straight_inf[0][2])
+        coord_bound_straight_sup[0][2] = min(nz_s, coord_bound_straight_sup[0][2])
+
+        if self.safe_zone:
+            logger.info('Applying safe zone to warping fields')
```
joshuacwnewton, line 517: Due to the new `batch_processing.sh` "detailed check", we can see that this change did in fact affect the existing warping field produced during the T2 registration to template: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15640363898/job/44068537856#step:5:2154 

This in turn had a small effect on all of the files warped from the template to the T2 space. However, usefully, this was not enough to change the results of `batch_processing.sh`, nor was it enough to change the effect on the subsequent registration steps (where the warping fields are reused for `-initwarp`, etc.).

Still, I would like to double-check the differences between the warping fields produced by this PR vs. the warping fields from `master`, so that I can better understand exactly the effect these changes had.

joshuacwnewton: On closer look, it looks like my changes may have inadvertently applied a `100000` value instead of the `-100000` value that the [`threshold_distance`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4884#issuecomment-2968356463) code applies: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15640363898/job/44068537856#step:5:2170

This likely isn't a huge difference, but I'm curious if keeping the value consistent will have a smaller effect on the output warping field.

joshuacwnewton: Fixing that error did reduce the number of "changed properties" from 182 to ~100, so that's a step in the right direction, but ideally this change wouldn't have much of an effect at all on the warping field. :thinking:

joshuacwnewton: It looks like there may be an off-by-one error? When I check the `batch_processing.sh` results, the top slice of the warping field gets overwritten on my branch but is preserved on `master`, despite the spinal cord covering that slice.

That would explain the slight differences. I'll try to dig in deeper. :)

joshuacwnewton: Yep! The existing code had an off-by-one indexing error indeed. (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4932/commits/042f81659e824ec7e784e14e1179c165f054fc6a)

joshuacwnewton: Hrm. Fixing this off-by-one error reduced the number of changed properties to ~70. Now the `warp_curve2straight.nii.gz` warping field is identical (no changed properties), but the `warp_straight2curve.nii.gz` warping field differs.

I'll try to dig into the changes caused to that warping field, too. :)

joshuacwnewton:
During today's SCT meeting, Sandrine and Julien rightfully pointed out that we don't know enough about why this is occurring for this specific user's data (3+ labels, lumbar region) but not other cases (2 label lumbar, 3+ label cervical).

So, I'm going to mark this PR as draft for now. (I think there are still some useful fixes and probing of the "safe zone" code, but I think the cause needs to be better understood before we can feel confident in applying this fix.)

spinalcordtoolbox/straightening.py
```diff
@@ -469,6 +484,8 @@ def straighten(self):
                 distances_curved = centerline.get_distances_from_planes(physical_coordinates,
                                                                         nearest_indexes_curved)
                 lookup = lookup_curved2straight[nearest_indexes_curved]
+                # NB: Any indexes that were mapped to `0` will have their warping field value set to `-100000` (i.e. no warping),
+                #     regardless of the value of `self.threshold_distance`.
                 indexes_out_distance_curved = np.logical_or(
                     np.logical_or(distances_curved > self.threshold_distance,
                                   distances_curved < -self.threshold_distance), lookup == 0)
```
joshuacwnewton, line 491: Note: I'm kind of suspicious of using the value `0` to mark slice indexes that should be cleared in the warping field.

If there is a value `0` in the lookup table, it's not immediately apparent to me if this means "this index maps to slice index 0, the first z slice of the target image" or if it means "the value for this index has been set to 0 for the purposes of clearing the warping field at this index". _(In fact, does this logic mean that the 0th slice will always get cleared? Are we always ignoring `straight[:, :, 0]` and `curved[:, :, 0]` when constructing the warping field?)_

I'm wondering if it would make more sense to use some other value (`-1`, `None`, etc.) to denote "clear this slice in the warping field", that way we can preserve the actual 0th slice.

spinalcordtoolbox/scripts/sct_straighten_spinalcord.py
```diff
@@ -124,13 +125,16 @@ def get_parser():
     optional.add_argument(
         "-param",
         metavar=Metavar.list,
+        type=list_type(",", str),
```
mguaypaq, line 128: Wow, this bug made `-param` have no effect before this!

spinalcordtoolbox/straightening.py
```diff
@@ -163,7 +164,7 @@ def straighten(self):
 
         # Computation of the safe zone.
         # The safe zone is defined as the length of the spinal cord for which an axial segmentation will be complete
-        # The safe length (to remove) is computed using the safe radius (given as parameter) and the angle of the
+        # The safe length (to remove) is computed using the safe radius and the angle of the
```
mguaypaq, line 167: A lot of the computations below are moot and could be simplified, given that `radius_safe` is hard-coded to zero (and always has been, it seems?). But, let's not wake the dragons.

spinalcordtoolbox/straightening.py
```diff
@@ -389,12 +388,23 @@ def straighten(self):
                     lookup_curved2straight[index] = idx_closest
                 else:
                     lookup_curved2straight[index] = 0
-        for p in range(0, len(lookup_curved2straight) // 2):
```
mguaypaq, line 392: I see, this is why the bug only started showing up on lumbar images, where the halfway point isn't in the image, so this loop (among others) might stop too early.

spinalcordtoolbox/straightening.py
mguaypaq: I agree! In general I would suggest `None`, but it doesn't fit nicely into numpy arrays. I would avoid `-1`, because that's very likely to lead to even weirder indexing bugs (where it may mean the last index). But, it looks like there's absolutely no reason for `lookup_curved2straight` and `lookup_straight2curved` to be converted to numpy arrays, so maybe `None` would work.

I'll let you decide whether this is in-scope for this PR.

mguaypaq (approved):
This looks reasonable. Thanks for digging into this part of the code!

spinalcordtoolbox/straightening.py
joshuacwnewton: Huh. It looks like distinguishing between `None` and `0` may have caused a change in values in `batch_processing.sh`. (e.g. perhaps the 0th slice that was previously wiped out in the warping field is now preserved, and this is changing the results when averaging e.g. the CSA?)

I'm leaning towards thinking that this is fixing a bug, but I'll double-check the files on disk to make sure.

mguaypaq: :100:

joshuacwnewton: I'm realizing that this might be a bit time consuming to confirm, and that there are more urgent things I need to work on before the upcoming release (e.g. the PAM50 template issues).

I think I'd rather err on the side of not changing the behavior here, and stick with what has already worked thus far.

joshuacwnewton: Issue opened: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4965

mguaypaq (approved):

Issue #4928, opened 2025-06-12
Add "tissue bridge ratio" metrics to the lesion analysis output


valosekj:
### Description

A feature request from our collaborator @dpfyffer:

> I wondered whether we can include two more metrics within the sct_analyze_lesion output:
> _Ventral to total tissue bridge ratio_ and _dorsal to total tissue bridge ratio_ (ventral/dorsal midsagittal tissue bridge width divided by total midsagittal tissue bridge width)?
>  
> To get a feeling of how much ventral vs. dorsal tissue is spared, independent of the absolute width.

This shouldn't be difficult to implement. I'll work on it next week.

PR #4933, opened 2025-06-14
Implement "tissue bridge ratio" metrics

valosekj:
## Checklist

## Description

This PR adds two new metrics (and one old one) to the `sct_analyze_lesion` output:
```
- interpolated_total_bridge_width `[mm]`: total width of spared tissue in the **interpolated midsagittal slice** (i.e. dorsal + ventral)
- ventral_bridge_ratio: ventral midsagittal tissue bridge width divided by total midsagittal tissue bridge width
- dorsal_bridge_ratio: dorsal midsagittal tissue bridge width divided by total midsagittal tissue bridge width
```

<details>
<summary>example outputs on different lesions</summary>

Screenshot 2025-06-14 at 5 48 47 PM

```
...
  Midsagittal dorsal tissue bridge width: 0.77 mm
  Midsagittal ventral tissue bridge width: 0.76 mm
  Midsagittal dorsal bridge ratio: 50.14 %
  Midsagittal ventral bridge ratio: 49.86 %
...
```

Screenshot 2025-06-14 at 5 48 51 PM

```
...
  Midsagittal dorsal tissue bridge width: 0.74 mm
  Midsagittal ventral tissue bridge width: 2.95 mm
  Midsagittal dorsal bridge ratio: 20.02 %
  Midsagittal ventral bridge ratio: 79.98 %
...
```

Screenshot 2025-06-14 at 5 48 53 PM

```
...
  Midsagittal dorsal tissue bridge width: 0.0 mm
  Midsagittal ventral tissue bridge width: 4.34 mm
  Midsagittal dorsal bridge ratio: 0.0 %
  Midsagittal ventral bridge ratio: 100.0 %
...
```

</details>

> [!NOTE]
> Right now, I have implemented the tissue bridge ratio metrics only for the midsagittal tissue bridges. However, since `sct_analyze_lesion` computes tissue bridges for each sagittal slice (midsagittal and parasagittal), I can easily extend the ratio computation to include the parasagittal slices as well.

Additionally, the PR: 
 - adds a [new test](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/442cdc3a2762c68cfdd65003147439d1d000a526/testing/cli/test_cli_sct_analyze_lesion.py#L322-L361) to test that the script exits when no lesion is found in the input image
 - fixes bugs

## How to test the PR?

Download the attached zip archive with sample data and then run:

```bash
cd $SCT_DIR
git fetch
git checkout jv/4928-implement_tissue_bridge_ratio

sct_analyze_lesion -m t2_lesion.nii.gz -s t2_seg.nii.gz
sct_analyze_lesion -m t2_lesion2.nii.gz -s t2_seg.nii.gz
sct_analyze_lesion -m t2_lesion3.nii.gz -s t2_seg.nii.gz
```

[4933-sample_data_tissue_bridge_ratios.zip](https://github.com/user-attachments/files/20738570/4933-sample_data_tissue_bridge_ratios.zip)

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4928


valosekj:
Hey @dpfyffer! I've implemented the first version of tissue bridge ratio metrics. Please, let me know what you think 😊

dpfyffer:
@valosekj awesome, thank you!
Will test and let you know.

valosekj:
Tested on @dpfyffer's machine, and it works! 

@joshuacwnewton, the pull request is now ready for review 🙂

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -333,6 +335,32 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
         printv('  Maximum axial damage ratio: ' + str(np.round(maximum_axial_damage_ratio, 2)),
                self.verbose, type='info')
 
+    def _compute_issue_bridge_ratio(self, idx):
```
joshuacwnewton: nitpick, but: typo!

```suggestion
    def _compute_tissue_bridge_ratio(self, idx):
```

(will have to fix the caller site as well)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -333,6 +335,32 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # Compute the total tissue bridge width
+        total_bridge_width = dorsal_bridge_width + ventral_bridge_width
+
+        # Compute the dorsal and ventral bridge ratios
+        dorsal_bridge_ratio = dorsal_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
+        ventral_bridge_ratio = ventral_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
+
+        # Save the ratios to the measure DataFrame
+        self.measure_pd.loc[idx, 'dorsal_bridge_ratio [%]'] = dorsal_bridge_ratio
+        self.measure_pd.loc[idx, 'ventral_bridge_ratio [%]'] = ventral_bridge_ratio
```
joshuacwnewton, line 361: This should be something we can test fairly easily in our current test suite, right? Since we're already checking both bridge widths in the "compute_expected_measurements" function, the calculation of the "expected" ratio seems very straightforward too. :D

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -1027,6 +1063,10 @@ def label_lesion(self):
         self.measure_pd['label'] = [label for label in np.unique(im_2save.data) if label]
         printv('Lesion count = ' + str(len(self.measure_pd['label'])), self.verbose, 'info')
 
+        # Exit the script if no lesion is found
+        if len(self.measure_pd['label']) == 0:
+            printv('ERROR: No lesion found in the input image.', self.verbose, 'error')
```
joshuacwnewton: Maybe we could add a test for this as well? I'm curious what happen if an empty lesion file is passed to `sct_analyze_lesion`. 

Does it fail earlier? Or was it not failing at all, hence the need for the check?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -333,6 +335,32 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        dorsal_bridge_ratio = dorsal_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
+        ventral_bridge_ratio = ventral_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
+
+        # Save the ratios to the measure DataFrame
+        self.measure_pd.loc[idx, 'dorsal_bridge_ratio [%]'] = dorsal_bridge_ratio
+        self.measure_pd.loc[idx, 'ventral_bridge_ratio [%]'] = ventral_bridge_ratio
+        printv(f'  Midsagittal dorsal bridge ratio: {np.round(dorsal_bridge_ratio, 2)} %',
+               self.verbose, type='info')
+        printv(f'  Midsagittal ventral bridge ratio: {np.round(ventral_bridge_ratio, 2)} %',
+               self.verbose, type='info')
```
joshuacwnewton: Tiny nit, but you can use f-string formatting to get the desired output:

```suggestion
        printv(f'  Midsagittal dorsal bridge ratio: {dorsal_bridge_ratio:.2f} %',
               self.verbose, type='info')
        printv(f'  Midsagittal ventral bridge ratio: {ventral_bridge_ratio:.2f} %',
               self.verbose, type='info')
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -333,6 +335,32 @@ def _measure_axial_damage_ratio(self, im_data, p_lst, idx):
...
+        # Get the dorsal and ventral tissue bridges
+        dorsal_bridge_width = float(self.measure_pd.loc[idx, 'interpolated_dorsal_bridge_width [mm]'])
+        ventral_bridge_width = float(self.measure_pd.loc[idx, 'interpolated_ventral_bridge_width [mm]'])
+
+        # Compute the total tissue bridge width
+        total_bridge_width = dorsal_bridge_width + ventral_bridge_width
+
+        # Compute the dorsal and ventral bridge ratios
+        dorsal_bridge_ratio = dorsal_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
+        ventral_bridge_ratio = ventral_bridge_width / total_bridge_width * 100 if total_bridge_width > 0 else 0
```
joshuacwnewton: nit: I think maybe we want `0.0` instead of `0` just to keep the float type consistent?

Also, these one-liners are getting a little long, so it might help with readability to split them up:

```suggestion
    if total_bridge_width > 0:
        dorsal_bridge_ratio = (dorsal_bridge_width / total_bridge_width) * 100
        ventral_bridge_ratio = (ventral_bridge_width / total_bridge_width) * 100
    else:
        dorsal_bridge_ratio = 0.0
        ventral_bridge_ratio = 0.0
```

joshuacwnewton (changes requested):
This mostly looks good to me! The calculation is very straightforward, so my comments are mostly nitpicky re: code readability and testing.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: good catch! fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/ae34f190f329a8f6ed9877c23b9b97cd14e5084c

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: sounds good! accepting the suggestion

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Yes, in the case of an empty lesion, the script is terminated by exactly this line

```shell
Lesion count = 0
ERROR: No lesion found in the input image.

Total runtime; 0.486 seconds.
Process finished with exit code 1
``` 

Speaking about the script termination, as you've opened this question, I've been actually thinking for a while about changing this behaviour slightly. Because currently, if there's no lesion, the script ends with exit code 1. If `sct_analyze_lesion` is called from a bash script run by our `sct_run_batch` wrapper, this then, in turn, results in termination of the entire bash script. I've been using the following workaround to not terminate the bash script in the case of an empty lesion.

```
  # Check if the lesion is empty, if so, skip sct_analyze_lesion (because otherwise sct_analyze_lesion would stop the script)
  status=$($SCT_DIR/python/envs/venv_sct/bin/python -c "import nibabel as nib; import numpy as np; print('empty' if np.all(nib.load('${file}_lesion.nii.gz').get_fdata() == 0) else 'not empty')")
  if [ "$status" = "empty" ]; then
      echo "❌ [$(date '+%Y-%m-%d %H:%M:%S')] ${file}: No lesion found in the input image. Skipping lesion analysis." >> "${PATH_LOG}/T2w_lesion_analysis.log"
  else
      sct_analyze_lesion -m ${file}_lesion.nii.gz -s ${file}_seg.nii.gz -f label_T2w -qc ${PATH_QC} -qc-subject ${file}
      echo "✅ [$(date '+%Y-%m-%d %H:%M:%S')] ${file}: Lesion analysis completed successfully." >> "${PATH_LOG}/T2w_lesion_analysis.log"
  fi
```

But I think it should not be diffucult to change the script behaviour to end with exit code 0, e.g., by doing something like this:

```python
# Exit the script gracefully if no lesion is found
if len(self.measure_pd['label']) == 0:
    printv('WARNING: No lesion found in the input image. Exiting.', self.verbose, 'warning')
    print('Analysis completed with no results (no lesion found).')
    sys.exit(0)  # Exit with success code
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: That's an excellent point! I actually started writing a test for these new ratio metrics, but then I realized that we are actually not computing (and testing) `interpolated_dorsal_bridge_width` and `interpolated_ventral_bridge_width` in our test suite. These interpolated bridges are used to compute `dorsal_bridge_ratio` and `ventral_bridge_ratio`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0f97779b53b3f81d6698b7382920f9914fba290c/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L345-L347 

In other words, if we want to test `dorsal_bridge_ratio` and `ventral_bridge_ratio`, we need to add computation of `interpolated_dorsal_bridge_width` and `interpolated_ventral_bridge_width`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: I drafted a possible workaround and a new test in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/b51ae77e33004907272394c35c11081262d28e68. Let me know what you think :-)

joshuacwnewton: I am curious about your use-case? Normally, I would expect that if a lesion image is empty, you would _want_ to halt processing and not do anything else in the bash script! May I ask what commands you would hope to run in the script after skipping the empty lesion image?

----

Anyway, the reason I mention this is that I think the exit code `1` is a useful signal, and `bash` scripts can do a lot with this! 

- Make sure to turn off `set -e` if you want the "continue on failure" behavior. (I know most of our bash scripts include this by default as boilerplate, but it sounds like you _don't_ want this behavior.)
- Alternatively, if you want to preserve the "halt on failure" behavior and only work around `sct_analyze_lesion`, you can use the `||` syntax to absorb the failure and run a command on failure.
- To easily capture the exit code (for further processing), you can use `$?`.

So, for example, you could either remove `set -e`, or do something like:

```bash
status=0
sct_analyze_lesion || status=$?
# logging based on status goes here
```

valosekj: After a discussion during the 2025-07-31 SCT dev meeting, I switched back to exit code 1 in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/13b6283e1cf3f8f42d383e59070fc5e3060e7ced

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Maybe we can document this in an issue and address it in a follow-up PR to be able to proceed with this PR.

joshuacwnewton: It's okay! I spent a little time implementing the test, and it seems to work OK?

```python
            # Estimate the interpolated bridge widths at the mid-sagittal slice
            decimal = mid_sagittal_slice - math.floor(mid_sagittal_slice)
            z_ceil, z_floor = int(np.ceil(mid_sagittal_slice)), int(np.floor(mid_sagittal_slice))
            for key in ['dorsal_bridge_width', 'ventral_bridge_width', 'total_bridge_width']:
                tissue_bridges[f'interpolated_{key} [mm]'] = (decimal * tissue_bridges.get(f'slice_{z_ceil}_{key} [mm]', 0.0) +
                                                              (1 - decimal) * tissue_bridges.get(f'slice_{z_floor}_{key} [mm]', 0.0))

            # Compute the bridge ratio using the interpolated bridge widths
            for key in ['dorsal', 'ventral']:
                # avoid zero division
                if tissue_bridges[f'interpolated_total_bridge_width [mm]'] != 0.0:
                    tissue_bridges[f'{key}_bridge_ratio [%]'] = (tissue_bridges[f'interpolated_{key}_bridge_width [mm]'] /
                                                                 tissue_bridges[f'interpolated_total_bridge_width [mm]']) * 100
                else:
                    tissue_bridges[f'{key}_bridge_ratio [%]'] = 0.0

            # We don't actually store the total interpolated bridge width in the final output, so delete it
            del tissue_bridges['interpolated_total_bridge_width [mm]']
```

----

I also needed to add a check for `nan` values, as the interpolated values are output as `nan` for one of the test cases:

```python
            # Sometimes the actual value is NaN (e.g. for lesions outside the cord). Skip checking these.
            if np.isnan(measurements.at[idx, key]):
                continue
```

Is this expected behavior?

----

Lastly, I found that one of these checks fails:

```python
                # The values will be adjusted according to the cos of the angle
                # between the spinal cord centerline and the S-I axis, as per:
                # https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3681#discussion_r804822552
                if key == 'max_equivalent_diameter [mm]' or 'bridge' in key:
                    assert measurements.at[idx, key] <= expected_value
```

The values are the same, but one value has slightly more precision than the other, so it is greater than expected. I'm not quite sure the best fix for this? Maybe rounding both values to the same number of decimal points before comparing? :thinking:

valosekj: > ```python
>             # We don't actually store the total interpolated bridge width in the final output, so delete it
>             del tissue_bridges['interpolated_total_bridge_width [mm]']
> ```

That's a good point, actually! Total interpolated bridge width is also useful, so why not store it? I added it to the main script in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/69dc10d72000c1be4ea12a87371814ad5245b82f.

valosekj: > It's okay! I spent a little time implementing the test, and it seems to work OK?

This is awesome, @joshuacwnewton! Thank you a lot! I took the liberty of including your code in my recent commits.

valosekj: > The values are the same, but one value has slightly more precision than the other, so it is greater than expected. I'm not quite sure the best fix for this? Maybe rounding both values to the same number of decimal points before comparing? 🤔

Good point! I tried rounding to 10 decimal places, and it works, so I implemented it in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/f28111b5a40ac1ac8335815fad324dcb8ff11f02

valosekj:
To test for the case described in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933#issuecomment-3150529290, I added a new dummy lesion file with two lesions along the spinal cord RL axis in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/6f16882913b045e1efefc5e087e4d592cabe15a9:

image

When running our existing test suite on this new dummy lesion, I discovered an indentation bug in the dictionary assignment 
when computing expected measurements for multiple lesions. I fixed it in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4933/commits/442cdc3a2762c68cfdd65003147439d1d000a526

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -143,8 +158,7 @@ def compute_expected_measurements(lesion_params, path_seg=None):
             'max_axial_damage_ratio []': dim[0] * dim[2] / min_area if min_area != 0 else None,
             **tissue_bridges
         }
-
-    expected_measurements_dict[idx] = measurements
```
joshuacwnewton, line 147: Nice catch! :hearts:

joshuacwnewton (approved):
LGTM! Thanks so much for this contribution. :hearts:

PR #4935, opened 2025-06-17
Validate `$SCT_DIR` for path-length issues (even for the default directory + non-interactive mode)

### LINKED ISSUE #4813: SCT installation fails when the working directory's path length exceeds ~120 characters
This error report was relayed to me by @naga-karthik ~~and @valosekj~~ via existing workflow files ([1](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/blob/e51707fc249556f17fbf22db3f3bda3143df2d23/.github/workflows/download_dataset.yml), [2](https://github.com/spinalcordtoolbox/manual-correction/blob/main/.github/workflows/ci.yml)). (EDIT: The second workflow file is *not* reporting this error.)

The error is as follows:

```console
> ./install_sct -y

# [...]

Installing base environment...
Transaction
  Prefix: /home/runner/work/contrast-agnostic-softseg-spinalcord/contrast-agnostic-softseg-spinalcord/spinalcordtoolbox/python
  Updating specs:

# [...]

Transaction finished

# [...]

Creating conda environment...

  File "/home/runner/work/contrast-agnostic-softseg-spinalcord/contrast-agnostic-softseg-spinalcord/spinalcordtoolbox/python/bin/conda", line 12, in <module>
    from conda.cli import main
ModuleNotFoundError: No module named 'conda'
```

<details>
<summary> Incorrect guess, ignore </summary>

Looking at the following issue:

- https://github.com/conda/conda/issues/9672

It seems like the error occurs due to conflicting Python/conda versions? And specifically, having the variable `PYTHONPATH` set might mess things up. 

Using this clue, I notice that both of the affected workflow files set up separate Python and conda installations *before* installing SCT. I'm wondering if this may be impacting the installation...

</details>

joshuacwnewton:
## Checklist

- [X] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)

## Description

The SCT installation will fail in an unintuitive way on Ubuntu when the path length of `$SCT_DIR` exceeds 107:

image

This PR adds a check for this path length, and provides a more intuitive error instead:

image

However, while authoring this PR, I encountered a fatal flaw: The path checks will only be applied for user-entered directories. If the default directory has issues and the user accepts the default directory, or if the script is run in non-interactive mode, then the path will not be checked and the user will be given no chance to fix the issue, potentially resulting in destructive behavior on unsafe paths.

So, to address this loophole, I reworked the conditional structure to check the default directory. In addition, this check will escalate from warning to error in case the user is in non-interactive mode and cannot fix the issue:

image

## Linked issues

Fixes #4813.


mguaypaq (approved):
LGTM! I guess this is limited to Unix-likes, so it doesn't need to be ported to the Windows installer?

joshuacwnewton:
> LGTM! I guess this is limited to Unix-likes, so it doesn't need to be ported to the Windows installer?

Truthfully, I don't fully understand the cause of the path issues. The limit (107) is so much smaller than any noted path limits for conda? But this is a good question, and I really should dive deeper to learn why this error is occurring before feeling comfortable that it's truly resolved for all OSs.

joshuacwnewton:
Just did some quick tests, and of course there is a path length limit for Windows as well. The error given is different ("Insufficient memory" during the `xcopy` step) but nonetheless this is a confusing error that could be handled by us. I'll push a fix along the same lines for the Windows installer.

install_sct.bat
```diff
@@ -79,7 +79,9 @@ rem Allow user to set a custom installation directory
 :while_loop_sct_dir
   echo:
   echo ### SCT will be installed here: [%SCT_DIR%]
-  set keep_default_path=yes
```
joshuacwnewton, line 82: [Oops!](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/15761025498/job/44427528943?pr=4935#step:3:24) It looks like this default value of `yes` may have been what allowed the script to run in non-interactive mode.

I'll have to rethink the flow of logic to allow for non-interactive mode.

joshuacwnewton: The fix here is quite simple! If the `keep_default_path` value doesn't change from the `SET /P` command, then we can assume it's a noninteractive shell and auto-accept the default.

Likewise, we can apply the same logic to the `new_install` variable to prevent an infinite loop there, as well (in case the default install location is invalid and we're unable to update it).

mguaypaq:
Apparently Windows batch scripts can use `CALL` and `EXIT /B` instead of `GOTO` for slightly more normal function calls:
https://stackoverflow.com/questions/3713601/subroutines-in-batch-files
https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/call

install_sct.bat
```diff
@@ -75,41 +75,71 @@ if %SCT_INSTALL_TYPE%==in-place (
   set SCT_DIR=%USERPROFILE%\sct_%SCT_VERSION%
 )
 
-rem Allow user to set a custom installation directory
-:while_loop_sct_dir
+rem Validate the default installation directory 
+rem If it's not valid, don't propose it to the user
+call :validate_sct_dir || goto :while_prompt_custom_path
```
joshuacwnewton: It might be good to emit some sort of message prior to the first check?

Something like: 

```
echo ### Setting default installation directory to home folder: %USERPROFILE%
echo ### Checking default installation directory for potential issues...
```

That way, if the default directory has issues, and we immediately emit a validation warning, we can still provide some context for where that error came from.

install_sct.bat
```diff
@@ -75,41 +75,71 @@ if %SCT_INSTALL_TYPE%==in-place (
...
-    echo ### No new path provided. Exiting installation.
+
+  rem The non-interactive default is to fail the installation
+  set SCT_DIR=non-interactive
+  set /p SCT_DIR="### Warning^! Give full path ^(e.g. C:\Users\username\sct_v3.0^): "
+
+  if ["%SCT_DIR%"]==["non-interactive"] (
+    rem The installation is non-interactive, and the default path is invalid
+    echo ### ERROR: The default installation directory is invalid.
+    echo            Please run the installer in interactive mode instead.
```
joshuacwnewton: Hrm. I found a bug with my logic? If the user presses enter at this point, then the condition will flag it as a non-interactive installation. In fact, I'm not sure there's a way to differentiate between pressing enter and a non-interactive shell?

Maybe we could do something a bit silly, like counting how many times we get the "enter/no input" result, then breaking after N times where N is large enough that a human probably wouldn't press enter that many times in a row? (But, if they do, I've also expanded the error message to be explicit about the fact that no input was detected, that way we're not 100% assuming a non-interactive shell in our message.)

This seemed to do the trick for me (after setting `attempt=0` earlier in the script):

```cmd
  rem Avoid getting caught in a loop if non-interactive (i.e. no input)
  rem But, still allow some "non-input" before failing (e.g. if user presses enter)
  if ["%SCT_DIR%"]==["non-interactive"] (
    set /a attempts=attempts+1
    if !attempts! GTR 10 (
      rem The installation is non-interactive, and the default path is invalid
      echo ### ERROR: The default installation directory is invalid, and no input was detected.
      echo            Please make sure to enter an input, or run the installer in interactive mode instead.
      goto error
    )
    goto :while_prompt_custom_path
  )
```

joshuacwnewton (commented):
Tested this on Windows! It basically works as expected, with only some slight quirks, but nothing that wasn't already present in the old goto version. Your refactoring worked splendidly, and and the `exit /b` syntax is really nifty for allowing the callers to decide what they want to do with various return codes.

install_sct.bat
mguaypaq: Instead I made the default validation silent in 70884c12791f26843c7a036b2407c1328d67e267.

install_sct.bat
mguaypaq: Good point! Done in 35df40fad76bc10fad3ea2eeeed52eb77b619d95.

install_sct.bat
```diff
@@ -75,64 +75,93 @@ if %SCT_INSTALL_TYPE%==in-place (
...
+  rem or the user accepts the default path
+  if /i ["%keep_default_path%"]==["y"] goto :done_sct_dir
+
+  rem The user wants a non-default path
+  if /i ["%keep_default_path%"]==["n"] goto :while_prompt_custom_path
+goto :while_prompt_default_path
+
+:while_prompt_custom_path
+  set /a attempt=attempt+1
+  if !attempts! GTR 10 (
```
joshuacwnewton: Whoops! My bad.

```suggestion
  if !attempt! GTR 10 (
```

install_sct.bat
```diff
@@ -75,64 +75,93 @@ if %SCT_INSTALL_TYPE%==in-place (
...
+    rem The installation is probably non-interactive,
+    rem and the default path is invalid, so fail
+      echo ### ERROR: The default installation directory is invalid, and no input was detected.
+      echo            Please make sure to enter an input, or run the installer in interactive mode instead.
+      goto error
```
joshuacwnewton: I guess technically we can trip this condition by selecting `'n'` to decline the default path, then selecting an invalid path 10 times.

While very rare, I guess it makes the message invalid. (I had already rewritten this message once, but I guess my understanding was off?)

```suggestion
    rem The install path was invalid 10 times, so this is probably non-interactive. Halt.
      echo ### ERROR: The chosen installation directory is invalid, and no valid input was passed.
      echo            Please make sure to enter a valid input.
      goto error

```

joshuacwnewton (commented):
I can't approve my own PR, but after fixing the issue, I tested it a bit more and I think this is good to go. :)

install_sct.bat
joshuacwnewton: Hum. Since we're now silencing warnings for the default path, _and_ we're overwriting the default path with `''` during the custom path prompt loop, right now we have no way of telling why a default path might have failed in non-interactive mode, which was [sort of the whole reason this PR was opened](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4813#issuecomment-2741278653). :sweat_smile:

joshuacwnewton: Does this look OK? https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4935/commits/ba4b43ed8cc58addfa6064021f56149cd6d7ca30

If so, I think this PR is good to merge now. Tested again locally and the messages display as expected.

mguaypaq (approved):

install_sct.bat
mguaypaq: Perfect, thanks!

PR #3493, opened 2021-07-31
Add tutorials for the remaining material from the SCT Course

joshuacwnewton:
I'm opening this draft PR because I wanted to be transparent/communicative about what I have been working on in my off-time these past few weeks.

I've finished writing the bulk of the GM/WM material, which is about 50% of the entire PR, but the following still needs to be done:

- [ ] Rewrite the atlas/partial volume effect pages to be easier to read/parse.
- [ ] Finish filling out smaller tutorials (dMRI, fMRI, and other one-page tutorials specific to individual SCT functions)
- [ ] Add CSV files from metric calculation steps
- [ ] Add figures from slideshow
- [ ] Listen back to presentation in real time to make sure that tutorial pages flow smoothly, and contain all of the necessary information from the in-person course.

I'll rewrite this PR description to give a proper overview once this PR is ready for review.

PR #4948, opened 2025-06-27
Add flag `-discfile` to allow for determining `VertLevel` using projected disc labels

### LINKED ISSUE #4937: Allow discs to be input directly (bypassing `sct_label_vertebrae`) when computing metrics
Right now, functions like `sct_process_segmentation` and `sct_extract_metric` require a `-vertfile` argument in order to aggregate metrics by vertebral level. This vertfile must be a ["full body"](https://spinalcordtoolbox.com/stable/user_section/tutorials/vertebral-labeling/types-of-vertebral-labels.html) labeled segmentation, traditionally produced by `sct_label_vertebrae`.

<details>

Image

</details>

However, recently we have been exploring the output of `sct_deepseg totalspineseg` as a potential replacement for `sct_label_vertebrae`. Instead of a "full body" labeled segmentation, it provides disc "point labels".

Because we don't have the "full body" labels by default, we have to adapt our old workflows to this new output:

1. We can make it easier to generate the full body labels from the point labels.
    - Addressed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4896.
    - The downside of this is that it requires an intermediate `sct_label_vertebrae` step after running `totalspineseg`, which might be confusing/not very intuitive: Why run TSS at all if we just have to run `sct_label_vertebrae` afterwards anyway?
2. Alternatively, we could remove the requirement for full body labels, since having a segmentation + disc labels _should_ encode the same information (and should be enough to aggregate metrics by level).

The latter idea was brought up in the [2025-06-05 SCT Developer meeting](https://docs.google.com/document/d/10txnIVrX3roF-VYP3yYUVLk7KmeL-43-ppKMKJ1BCpE/edit?tab=t.5f0y18irce5a), hence opening this issue.

sandrinebedard:
## Checklist

## Description

This PR adresses #4937. It allows to directly input discs instead of a labeled segmentation in `sct_process_segmentation`. It adds the extra argument `-discfile`, but the `-vertfile` argument can also be used.

In brief:
- It projects on the centerline the discs
- Creates a labeled centerline (saves it too for validation)

### How to test
Here are the commands to test on the spine generic subject sub-amu01:

Here are the images to use: 
[sub-amu01_T2w_seg.nii.gz](https://github.com/user-attachments/files/20956038/sub-amu01_T2w_seg.nii.gz)
[sub-amu01_T2w_seg_labeled.nii.gz](https://github.com/user-attachments/files/20956040/sub-amu01_T2w_seg_labeled.nii.gz)
[sub-amu01_T2w_label-discs_dlabel.nii.gz](https://github.com/user-attachments/files/20956043/sub-amu01_T2w_label-discs_dlabel.nii.gz)

```python
# using flag -discfile
sct_process_segmentation -i sub-amu01_T2w_seg.nii.gz -vert 1:8 -perlevel 1 -discfile sub-amu01_T2w_label-discs_dlabel.nii.gz  -o csa_discs_centerline_projected.csv
```

```python
# using flag -vertfile
sct_process_segmentation -i sub-amu01_T2w_seg.nii.gz -vert 1:8 -perlevel 1 -vertfile sub-amu01_T2w_seg_labeled.nii.gz   -o csa_seg_labeled.csv
```

## Linked issues

Partially addresses #4937. (Remaining work will be done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4996.)


spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -425,6 +445,14 @@ def main(argv: Sequence[str]):
                                          param_centerline=param_centerline,
                                          verbose=verbose,
                                          remove_temp_files=arguments.r)
+    # Project discs labels to centerline for discfile
+    if levels_disc:
```
valosekj: Is the `levels_disc` variable necessary? Maybe we could use `if arguments.discfile`?

joshuacwnewton: It is a bit complicated because both `discfile` and `vertfile` have a default value, so as currently written, we cannot use `if arguments.discfile`.

For `-vertfile` specifically, this was done to allow the user to specify `-vert` without having to specify `-vertfile`.

That being said, I remember when I was developing the SCT course material a few years ago, I really disliked this (because you couldn't tell which vertlevels were being used by default based on the command alone). So, I intentionally added `-vertfile` to all of the course commands (even though it wasn't strictly necessary).

In fact, it was @sandrinebedard who noticed this 4 years ago :laughing: : https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3493#discussion_r687103187

Maybe we could remove the default value from `-discfile` (forcing users to manually specify it for clarity's sake, and thus allow the usage of `if arguments.discfile is not None`)?

valosekj: I see! Thanks for the details :-)

sandrinebedard: > Maybe we could remove the default value from -discfile (and thus allow the usage of if arguments.discfile is not None)?

Agreed!

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -161,6 +162,18 @@ def get_parser():
             Vertebral labeling file. Only use with flag `-vert`.
 
             The input and the vertebral labelling file must in the same voxel coordinate system and must match the dimensions between each other.
```
sandrinebedard: TODO add depreciation message

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -382,17 +395,24 @@ def main(argv: Sequence[str]):
...
+    elif normalize_pam50 and fname_vert_level is None:
+        parser.error("Option '-normalize-PAM50' requires option '-vertfile' or '-discfile'.")
+    elif levels:
+        parser.error("Option '-normalize-PAM50' requires option '-vertfile' or '-discfile'.")
+    if fname_vert_level is not None:
+        if not os.path.isfile(fname_vert_level):
+            raise FileNotFoundError(f"Vertebral level file {fname_vert_level} does not exist. Vert level information will "
+                                    f"not be displayed. To use vertebral level information, you may need to run "
+                                    f"`sct_warp_template` to generate the appropriate level file in your working directory.")
+            fname_vert_level = None  # Discard the default '-vertfile', so that we don't attempt to find vertebral levels
```
sandrinebedard: @joshuacwnewton since I removed the vertfile and discfile default values, I had to update these exceptions, please feel free to update if there is a better way to do it here!

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -382,17 +395,24 @@ def main(argv: Sequence[str]):
...
+        fname_vert_level = arguments.vertfile
+    elif normalize_pam50 and fname_vert_level is None:
+        parser.error("Option '-normalize-PAM50' requires option '-vertfile' or '-discfile'.")
+    elif levels:
+        parser.error("Option '-normalize-PAM50' requires option '-vertfile' or '-discfile'.")
+    if fname_vert_level is not None:
+        if not os.path.isfile(fname_vert_level):
+            raise FileNotFoundError(f"Vertebral level file {fname_vert_level} does not exist. Vert level information will "
+                                    f"not be displayed. To use vertebral level information, you may need to run "
+                                    f"`sct_warp_template` to generate the appropriate level file in your working directory.")
```
sandrinebedard: @joshuacwnewton @jcohenadad I would update this message to run like `sct_label_vertebrae` or totalspineseg instead of `sct_warp_template` here, what do you think?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -156,11 +157,23 @@ def get_parser():
...
+            The input and the vertebral labelling file must be in the same voxel coordinate system and must match the dimensions between each other.
+            Example: ./label/template/PAM50_levels.nii.gz
+            This flag will be deprecated in favor of -discfile in the future.
+        """),
+    )
+    optional.add_argument(
+        '-discfile',
+        metavar=Metavar.str,
+        help=textwrap.dedent("""
+            File with disc labels. Only use with flag `-vert`.
```
joshuacwnewton: Maybe we could update this description to say something like:

```
File with single-voxel labels identifying the intervertebral discs. Used with `-vert` to 
aggregate metrics within vertebral levels. Disc labels will be projected onto the spinal 
cord to identify vertebral levels.
```

That way it's clearer how the `-discfile` interacts with the `-vert` argument? (But maybe also it's too wordy.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -156,11 +157,23 @@ def get_parser():
     optional.add_argument(
         '-vertfile',
         metavar=Metavar.str,
-        default=os.path.join('.', 'label', 'template', 'PAM50_levels.nii.gz'),
```
joshuacwnewton, line 159: I just realized that removing the default value changes how we populate the `VertLevel` column when `-vert` isn't passed. Now we can't fall back to the PAM50 file like we used to be able to do: (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4004)

Maybe we fix this by **not** putting the local path in `default=`, but instead assigning the path after parsing the arguments. Something like:

```python
# if user hasn't specified their own vert/discfile, try to find one on their behalf
if arguments.vertfile is None and arguments.discfile is None:
    logger.info("No -vertfile/-discfile argument provided. Attempting to get VertLevel information from local PAM50 warped template file (if it exists).")
    fname_vert_level = os.path.join('.', 'label', 'template', 'PAM50_levels.nii.gz')
    # NB: If this file doesn't exist, we will need to change FileNotFoundError to be a warning instead
```  

That way, we still keep the default value, but we don't put it in `default=` (so we can still check user input), and we also emit a log message to make it clear that we're falling back to the warped PAM50 levels only because the user did not put `-vertfile` or `-discfile`. (So they are still encouraged to put the argument themselves.)

Best of both worlds?

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: I think that's a good idea.

Something like:

- To use vertebral level information, specify either `-vertfile` ("full-body labels" generated by `sct_label_vertebrae` or `sct_warp_template`) or `-discfile` (single-voxel "point labels" generated by `sct_deepseg totalspineseg` or `sct_label_utils`).

Maybe too much information though? Maybe instead we just say:

- To use vertebral level information, specify either `-vertfile` or `-discfile`. For generating the required files, please review `sct_process_segmentation -h` for these arguments.

Then, we put the CLI script details in the `-vertfile` and `-discfile` help.

joshuacwnewton (commented):
Thank you for the fix, and for adding a test! This is a great start. :hearts:

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: I think this works well as is, but I might do a bit of tidying up? (Combined with my other suggestion [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948#discussion_r2185685960)...)

```python
    # TODO: Remove `fname_vert_level = None` statement from earlier (not needed)
```

```python
    # make sure we have a valid VertLevel file (used for aggregation + VertLevel column)
    if arguments.vertfile is not None and arguments.discfile is not None:
        parser.error("Both '-vertfile' and '-discfile' were specified. Please only specify one of these options.")
    elif arguments.discfile is not None:
        fname_vert_level = arguments.discfile
    elif arguments.vertfile is not None:
        fname_vert_level = arguments.vertfile
    else:
        logger.info("No -vertfile/-discfile argument provided. Attempting to get VertLevel "
                    "information from local PAM50 warped template file (if it exists).")
        fname_vert_level = os.path.join('.', 'label', 'template', 'PAM50_levels.nii.gz')
    
    if not os.path.isfile(fname_vert_level):
        logger.warning(f"Vertebral level file {fname_vert_level} does not exist. Vert level information will "
                       f"not be displayed. To use vertebral level information, you may need to run "
                       f"`sct_warp_template` to generate the appropriate level file in your working directory.")
        # Discard the default '-vertfile', so that we don't attempt to find vertebral levels
        fname_vert_level = None
        # If `fname_vert_level` is invalid but vert levels are required, raise an error.
        if normalize_pam50:
            parser.error("Option '-normalize-PAM50' requires a valid vertebral level file ('-vertfile' or '-discfile').")
        elif levels:
            parser.error("Option '-vert' requires a valid vertebral level file ('-vertfile' or '-discfile').")
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: hmm, I still feel that this is not the right behaviour we want, you could still pass `-vertfile` without `-vert` just so the vertebral levels are displayed in the `output .csv`

As the `.label/` folder means that the user did register to the template, and probably has the discs labels in anycase, wich would be better to input than the warped ones from the template

Maybe @valosekj and @jcohenadad have an opinion on this!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/4d30a46af4c817cdce89619cbe726cd39a360025

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/403ef700a6a941965545a44b1058a1110aded92f

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/27dc9618a85b2c4e76c3c33db063f1c29e3747c3

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: > hmm, I still feel that this is not the right behaviour we want, you could still pass `-vertfile` without `-vert` just so the vertebral levels are displayed in the `output .csv`

Just to clarify -- my suggestion still allows for this? (See implementation in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948#discussion_r2185709756.)

The user can still pass `-vertfile` without `-vert`. My suggestion is just to keep the behavior where we fall back to the warped PAM50 template if `-vertfile` and `-discfile` are both not passed.

> As the .label/ folder means that the user did register to the template, and probably has the discs labels in anycase, wich would be better to input than the warped ones from the template

The user can still input the warped template files to `-vertfile`, right? This is something we do in the tutorials ([example](https://spinalcordtoolbox.com/stable/user_section/tutorials/diffusion-weighted-mri/extracting-dti-from-specific-regions.html)).

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -382,17 +394,30 @@ def main(argv: Sequence[str]):
...
+    else:
+        logger.info("No -vertfile/-discfile argument provided. Attempting to get VertLevel "
+                    "information from local PAM50 warped template file (if it exists).")
+        fname_vert_level = os.path.join('.', 'label', 'template', 'PAM50_levels.nii.gz')
     if not os.path.isfile(fname_vert_level):
         logger.warning(f"Vertebral level file {fname_vert_level} does not exist. Vert level information will "
-                       f"not be displayed. To use vertebral level information, you may need to run "
-                       f"`sct_warp_template` to generate the appropriate level file in your working directory.")
+                       f"not be displayed. To use vertebral level information, specify either -vertfile or -discfile."
+                       f"For generating the required files, please review sct_process_segmentation -h for these arguments.")
```
joshuacwnewton, line 424: Thanks for adding this message! Though, I think we may still need to update the `-discfile` and `-vertfile` arguments to explain which CLI scripts can be used to generate the files.

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: I am okay with that!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here [e57b12e](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/e57b12ebe521c61947970bd89cb93710a0892ec2)

joshuacwnewton (commented):
The current changes LGTM! Thank you so much for doing this. :hearts:

> ### Whats left?
>
> - [x] Add the same thing to `sct_extract_metric` if this is what we want
> - [ ] Add tests

Should I save my approval until after these TODOs are done?

sandrinebedard:
For `sct_extract_metric`, maybe projecting on the centerline is not the best strategy:

- Thick slices, not that precise in the vertebral labeling anyway, 
- Mostly rely on warped PAM50 for labels
- Segmentation is not provided as an input


Possible solutions:

1. Keep `-vertfile` as an input and use the value at the centerline instead of highest
2. Keep `-vertfile` as an input and use the majority level of the slice (if 20% level 2, 80% of level 3  --> level 3)

sandrinebedard:
@joshuacwnewton I removed the modifications from`sct_extract_metric` and will address in another PR

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton, line 1: _Originally posted by @jcohenadad in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4948#issuecomment-3073905765_

> @NathanMolinier In the code that looks for the closest point between the disc and the centerline, how is the origin of the disc label defined? There are several conventions for defining the origin of a voxel, eg: its geometric center, or one of its corner. This could have dramatic impact in the case of anisotropic resolution, as is usually the case for qMRI data. See illustration below:
> 
> image

joshuacwnewton: _Originally posted by @NathanMolinier in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4948#issuecomment-3149708043_

> I'm not sure I understand how to use different voxel conventions, because I only have access to a single coordinate when computing the distance. Here is the [code](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4647d04e6b8c28c59c84287c0b01f1c98113f99a/spinalcordtoolbox/labels.py#L510) if you want a closer look.

testing/cli/test_cli_sct_process_segmentation.py
```diff
@@ -169,6 +169,16 @@ def test_sct_process_segmentation_check_normalize_PAM50_missing_vertfile(tmp_pat
...
 
 
+def test_sct_process_segmentation_check_both_discfile_vertfile(tmp_path):
+    """ Run sct_process_segmentation with -vertfile and -discfile"""
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    with pytest.raises(SystemExit) as e:
+        sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'),
+                                            '-vertfile', sct_test_path('t2', 't2_seg-manual_labeled.nii.gz'),
+                                            '-discfile', sct_test_path('t2', 'labels.nii.gz'), '-o', filename])
+        assert e.value.code == 2
```
joshuacwnewton, line 179: Right now, there is only a test for ensuring that `-vertfile` + `-discfile` raise an error.

But, there aren't any tests for the actual functionality (i.e. passing `-discfile` then checking the outputs). Could you add a test for this?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -425,6 +450,14 @@ def main(argv: Sequence[str]):
...
                                          verbose=verbose,
                                          remove_temp_files=arguments.r)
+    # Project discs labels to centerline for discfile
+    if arguments.discfile is not None:
+        discs_projected = project_centerline(Image(fname_segmentation), Image(fname_vert_level))
+        if verbose == 2:
+            discs_projected.save(add_suffix(fname_vert_level, '_projected'))
+        ctl_projected = label_regions_from_reference(Image(fname_segmentation), discs_projected, centerline=True)
+        fname_vert_level = add_suffix(fname_vert_level, '_projected_centerline')
+        ctl_projected.save(fname_vert_level)
```
joshuacwnewton: Right now, we are saving the `_projected` files to the working directory. Maybe what we could do instead is save them to a temporary directory, and then if `-verbose 2`, we copy them from the tempdir to the output directory? That way the user "opts in" to these extra files.

I don't know if these are the kinds of files we should be saving by default, since it might be confusing when the expected output is the metric CSVs.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -156,11 +157,22 @@ def get_parser():
...
         metavar=Metavar.str,
-        default=os.path.join('.', 'label', 'template', 'PAM50_levels.nii.gz'),
         help=textwrap.dedent("""
-            Vertebral labeling file. Only use with flag `-vert`.
+            Vertebral labeling file generated by sct_label_vertebrae or sct_warp_template. Only use with flag `-vert`.
 
-            The input and the vertebral labelling file must in the same voxel coordinate system and must match the dimensions between each other.
+            The input and the vertebral labelling file must be in the same voxel coordinate system and must match the dimensions between each other.
+            Example: ./label/template/PAM50_levels.nii.gz
+            This flag will be deprecated in favor of -discfile in the future.
```
joshuacwnewton, line 167: Maybe instead of just adding a notice that it will be deprecated, we could add an additional message for people who supply `-vertfile` (e.g. in existing scripts) so that they are aware of the changes and can migrate accordingly.

We've done this in other PRs, see `sct_deepseg_lesion` and `sct_deepseg_sc`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/49ffe0c02c59d75348655c983e035a22a2655a75/spinalcordtoolbox/scripts/sct_deepseg_lesion.py#L83-L93

testing/cli/test_cli_sct_process_segmentation.py
sandrinebedard: Done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/fc15975579ae454121bb999e32a96ecaf1626bb5

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/f55edb0d2f94a8b1df90c4a81dfc118df2b16639

But I wasn't sure if this was the right way use temps folders in SCT (we should also delete afterwards no?)

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/fba960dc7257b4067fe228008e89e942c2e98ac5

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Yes, typically we would delete afterwards unless `-r 0` is specified.

It seems like other scripts use this pattern at the end of processing:

```python
    if remove_temp_files:
        logger.info("\nRemove temporary files...")
        tmp_folder.cleanup()
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -382,17 +394,40 @@ def main(argv: Sequence[str]):
...
+    # make sure we have a valid VertLevel file (used for aggregation + VertLevel column)
+    if arguments.vertfile is not None and arguments.discfile is not None:
+        parser.error("Both '-vertfile' and '-discfile' were specified. Please only specify one of these options.")
+    elif arguments.discfile is not None:
+        fname_vert_level = arguments.discfile
+    elif arguments.vertfile is not None:
+        fname_vert_level = arguments.vertfile
+        from warnings import warn
+        from spinalcordtoolbox.utils.sys import stylize
+        from time import sleep
```
joshuacwnewton: My apologies, my code was unclear! These imports should still go at the top of the file, I think. :sweat_smile:

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -425,6 +460,18 @@ def main(argv: Sequence[str]):
                                          param_centerline=param_centerline,
                                          verbose=verbose,
                                          remove_temp_files=arguments.r)
+    # Project discs labels to centerline for discfile
+    if arguments.discfile is not None:
```
joshuacwnewton: I think maybe there is a (tiny) potential for a crash here if `-discfile` is specified, but the file doesn't exist?

Right now, `fname_vert_level` gets overridden with `None` if the file doesn't exist. Then the script exits if `-vert` or `-normalize-PAM50` are specified. But if those neither of those arguments are specified, then we will try to load `Image(None)` in this block.

So I think maybe we could move the code up into an `else:` condition? That way, the full code looks like this:

```python

    # Make sure that the vertfile exists before processing it
    if not os.path.isfile(fname_vert_level):
        logger.warning(f"Vertebral level file {fname_vert_level} does not exist. Vert level information will "
                       f"not be displayed. To use vertebral level information, specify either -vertfile or -discfile."
                       f"For generating the required files, please review sct_process_segmentation -h for these arguments.")
        # Discard the default '-vertfile', so that we don't attempt to find vertebral levels
        fname_vert_level = None
        # If `fname_vert_level` is invalid but vert levels are required, raise an error.
        if normalize_pam50:
            parser.error("Option '-normalize-PAM50' requires a valid vertebral level file ('-vertfile' or '-discfile').")
        elif levels:
            parser.error("Option '-vert' requires a valid vertebral level file ('-vertfile' or '-discfile').")   
    
    # Vertfile exists, so pre-process it if it's a `-discfile`
    else:
        # Project discs labels to centerline for discfile
        if arguments.discfile is not None:
            discs_projected = project_centerline(Image(fname_segmentation), Image(fname_vert_level))
            temp_folder = TempFolder(basename="process-segmentation")
            path_tmp = temp_folder.get_path()
            discs_projected.save(os.path.join(path_tmp, add_suffix(fname_vert_level, '_projected')))
            ctl_projected = label_regions_from_reference(Image(fname_segmentation), discs_projected, centerline=True)
            fname_vert_level = os.path.join(path_tmp, add_suffix(fname_vert_level, '_projected_centerline'))
            ctl_projected.save(fname_vert_level)
            if verbose == 2:
                copy(fname_vert_level, os.getcwd())
```

(Note: `from spinalcordtoolbox.utils.fs import TempFolder` should be moved to the top as a top-level import too.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -382,17 +394,40 @@ def main(argv: Sequence[str]):
...
+        from spinalcordtoolbox.utils.sys import stylize
+        from time import sleep
+        warn(
+            stylize(
+                "`-vertfile flag` is deprecated, and will be removed in a future version of SCT. Please use "
+                "`-discfile` instead (single-voxel labels identifying the intervertebral discs).", ["Red", "Bold"]
+            ), DeprecationWarning
+        )
+        sleep(3)  # Give the user 3 seconds to read the message    else:
+    else:
```
joshuacwnewton: I think an extra `else:` was accidentally committed?

```suggestion
        sleep(3)  # Give the user 3 seconds to read the message
    else:
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Ah, I also realized that we should probably use the parent directory of the `file_out` (`-o`) file, rather than `os.getcwd()`?

sandrinebedard: Now fixed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/5eca8fbd0772a27e20fdf746d0bc4a37b5b39cb6 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/6d46e7608c61dd96d0f8994213082112f7acc0f1

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: now added here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/21c312f484672b14f9b1ef07dd87cb3421b4d0f3

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/9bb775a6d40ac26ea19ebe1a332b1205a4d31535

sandrinebedard:
Thanks @joshuacwnewton for the feedback, I think now it has all been addressed ! And regarding:

> The only remaining question is the one Julien raised, but maybe this is outside the scope of the PR?

I agree that this is out of scope of this PR

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -518,7 +568,10 @@ def main(argv: Sequence[str]):
                 parser.error('-qc-image is required to display QC report.')
         else:
             logger.warning('QC report only available for PMJ-based CSA. QC report not generated.')
-
+    # Clean up temp
+    if arguments.r:
+        logger.info("\nRemove temporary files...")
+        temp_folder.cleanup()
```
joshuacwnewton: I think the tests are failing because `temp_folder` only exists if `discfile` exists.

sandrinebedard: woops I noticed it at the same time, fixed here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/cb34b4621de7d24f601cf87a71ce4a63fda15155

joshuacwnewton: Actually, now that I think about it, a slightly more robust approach would be to initialize `temp_folder` to be `None`, then only call `.cleanup` if `temp_folder is not None`.

That way, the condition is directly linked to the thing we are accessing, and we avoid some weird bugs in the future where the state of `discfile` and `temp_folder` somehow get out of sync.

```diff
    # Make sure that the vertfile exists before processing it
+   temp_folder = None
    if not os.path.isfile(fname_vert_level):
```

sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4948/commits/1f46e7baccc1578df82c3f345958adfb96b9387f

joshuacwnewton (approved):
LGTM! I'll open a new issue for the disc discussion. :)

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Moved to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5004.

Issue #4939, opened 2025-06-23
`skimage.regionprops` does not produce reliable properties for heavily compressed cords (orientation, diameter)


joshuacwnewton:
## Expected behavior

Currently, in our "[Other shape metrics](https://spinalcordtoolbox.com/stable/user_section/tutorials/shape-analysis/compute-csa-and-other-shape-metrics/other-shape-metrics.html)" tutorial, we use the following figure:

![Image](https://github.com/user-attachments/assets/5cf4fcc4-12f6-4391-8e71-fc0f46c98aa4)

The "AP diameter" and "RL diameter" images suggest that we somehow compute these diameters using the spinal cord mask (because the red line only spans the spinal cord). 

## Actual behavior

However, this is not actually the case:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/03d344749b7e5e351dbf11f1c939fb024eaee8c5/spinalcordtoolbox/process_seg.py#L201-L204

Instead, we use `major_axis_length` and `minor_axis_length` directly, [without much further modification](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/03d344749b7e5e351dbf11f1c939fb024eaee8c5/spinalcordtoolbox/process_seg.py#L282-L301).

These lengths are computed by `skimage.regionprops` using an [elliptical approximation](https://github.com/scikit-image/scikit-image/blob/v0.25.2/skimage/measure/_regionprops.py#L1178-L1183), which may not accurately reflect the proper orientation or shape of the cord:

![Image](https://github.com/user-attachments/assets/260e61fc-9727-4f2e-a8aa-50e31980fa87)

(Figure generated using the [jv/debug_cord_morphometrics](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/master...jv/debug_cord_morphometrics) branch)

## Potential solutions

It seems that we were aware of this limitation back in 2019, when this comment was authored (#2200):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b9c511d01e05d6abb9c01a0b2ae283036275ed3d/spinalcordtoolbox/process_seg.py#L204

However, "summing weighted voxels" isn't entirely clear, and there are many possible ways to define the diameter of a blob.

## Action items

- [ ] Fix our documentation to be clear that we are basing our diameter value off an ellipsis, _not_ the SC mask itself.
[...]

PR #4943, opened 2025-06-25
Add spinal cord CSA quadrant measurements

valosekj:
## Checklist

- [ ] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)
- [ ] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [ ] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
[...]

Issue #4624, opened 2024-09-06
Create a morphometrics API


mguaypaq:
We sometimes tell users to use `sct_analyze_lesion` when they have a structure which is _not_ a lesion, but for which they want similar statistics, for example the length, volume.

Some of these statistics or analyses are also similar to `sct_process_segmentation`, so there is some code duplication (for example, angle correction).

We like having separate CLI scripts with descriptive names like "analyze lesion" and "process segmentation", but we could centralize some of the common functions into a new API module for morphometrics. This should make it relatively easy to create new CLI scripts for computing relevant statistics for other anatomical structures.

joshuacwnewton:
I independently had a similar thought in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#discussion_r2566058211:

> I'm wondering if it might make some sense to make a new API module to group all of the morphometrics-related modules? Something like:
> 
> ```
> spinalcordtoolbox/
> ├─ morphometrics/
> │  ├─ aggregate_slicewise.py
> │  ├─ csa_pmj.py
> │  ├─ metrics_to_PAM50.py
> │  ├─ plots.py
> │  ├─ process_seg.py
> │  └─ [...]
> └─ [...]
> ```

PR #4958, opened 2025-07-08
Update AP diameter to be based on rotated seg + add new `-anat` flag (for symmetry, quadrant area, and HOG angle properties)

valosekj:
## Description

This PR also updates the way we compute AP diameter and add symmetry metrics.

**Major changes:**

- Orientation (of ellipse) is now signed and in degrees
- Resampling to 0.1mm in plane instead of upscale
- Modification of AP diameter measures
- Adding symmetry metrics and quadrants areas based on  HOG angle

## Context

`sct_process_segmentation.py` currently computes orientation by `skimage.regionprops` using an [elliptical approximation](https://github.com/scikit-image/scikit-image/blob/v0.25.2/skimage/measure/_regionprops.py#L1294-L1297) of the cord segmentation ([L267-L268](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/d1b9e478405fdb9c6b932e2baa6e2dc89df5ab07/spinalcordtoolbox/process_seg.py#L267-L268)).

But, in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4943, we found that this might not be ideal because the ellipse sometimes does not approximate the cord as we would expect.

### Update AP diameter computation

***Before, the AP diameter was computed as the minor axis of the ellipse fit***

image

***Proposed method: We rotate the segmentation with the orientation of the ellipse:***

image

### Symmetry based on HOG angle

This PR also extends `sct_process_segmentation` functionally to compute the HOG angle of the image. To do so:
- The argument `-symmetry` now takes the image, in order to compute the HOG angle

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/4c29ef6f2dc50b306212001e5a3aeca80e0fa8b0: use `compute_pca` and  `find_angle_hog` from `spinalcordtoolbox.registration.algorithms`  to compute the HOG angle
- [rotate the segmentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4252b6eee8c8560eab1381c85caab98e95a5e86f/spinalcordtoolbox/process_seg.py#L321) by the HOG angle found from the image to align with AP/RL axes
- [added a new QC](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4252b6eee8c8560eab1381c85caab98e95a5e86f/spinalcordtoolbox/reports/qc2.py#L240)
- 3 types of symmetry are computed (dice, symmetric difference and hausdorff distance):

image

## How to test the PR

`sct_tetsing_data/t2/`:

```bash
# Without symmetry
sct_process_segmentation -i t2_seg-manual.nii.gz -perslice 1 -o ap.csv -qc ./qc -v 1
```

```bash
# With symmetry
sct_process_segmentation -symmetry t2.nii.gz -i t2_seg-manual.nii.gz  -perslice 1 -o symmetry.csv -qc ./qc -v 1
```

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4939


jcohenadad:
Useful cross-ref:
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2331
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2218
- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/836

jcohenadad:
To visually assess quality of angle estimation, you can use this QC code:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/7dfb515728f5ba702df1dabdd9a33f00c085e224/spinalcordtoolbox/reports/qc.py#L95-L127

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -106,11 +106,17 @@ def get_parser():
...
-        help="Mask to compute morphometrics from. Could be binary or weighted. E.g., spinal cord segmentation."
+        help="Segmentation mask to compute morphometrics from. Could be binary or weighted. E.g., spinal cord segmentation."
              "Example: seg.nii.gz"
     )
+    mandatory.add_argument(
+        '-i',
+        metavar=Metavar.file,
+        help="Input image used to compute spinal cord orientation (using HOG method)."
+             "Example: t2.nii.gz"
+    )
```
valosekj: consider if keep the `-i` arg mandatory or not. Making it mandatory would introduce backward incompatibility. There might be users who might have only the segmentation without the image.

valosekj: Okay, I've made `-i` optional in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/8e234237f969e99b9dbf89647c6aae5b56707f9d

spinalcordtoolbox/reports/qc2.py
```diff
@@ -307,12 +296,52 @@ def sct_process_segmentation(
...
+                    y_end = int(y + radius[1] * np.cos(angle_rad))
+                    # Create coordinates for the line
+                    rr, cc = draw_line(y_start, x_start, y_end, x_end)
+                    # Filter coordinates to ensure they're within image bounds
+                    valid_idx = (0 <= rr) & (rr < img_temp.dim[1]) & (0 <= cc) & (cc < img_temp.dim[2])
+                    # Set line pixels to 1
+                    if np.any(valid_idx):
+                        img_temp.data[img_temp.dim[0] - 1 - i, rr[valid_idx], cc[valid_idx]] = 1
+
+        img = mosaic(img_temp, centers, radius)
```
valosekj: Okay, this works, but the line is pixelated.

image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -301,62 +302,59 @@ def sct_process_segmentation(
...
+                        y_end = y_mosaic + radius[1]/2 * np.cos(angle_rad)
+
+                        # Plot the line
+                        ax.plot([x_start, x_end], [y_start, y_end], '-', color='red', linewidth=0.7)
+
+                        # Include the angle text in degrees
+                        angle_deg = -np.degrees(angle_rad)
+                        ax.text(x_mosaic + radius[0] * 0.2, y_mosaic - radius[1] * 0.3, # upper right corner
+                                f'{angle_deg:.1f}°', color='red', fontsize=3,
+                                path_effects=[mpl_patheffects.withStroke(linewidth=0.5, foreground='white')])
```
valosekj: Plotting the lines directly on axes (using vector graphics) looks much better!

image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -235,6 +237,132 @@ def sct_register_multimodal(
...
+        img_path = str(imgs_to_generate['path_overlay_img'])
+        logger.debug('Save image %s', img_path)
+        fig.savefig(img_path, format='png', transparent=True, dpi=DPI)
+
+
+def add_angle_lines(ax, num_slices, metrics, radius: tuple[int, int] = (15, 15), scale: float = 2.5):
+    """
+    Overlay HOG angle lines and add angle text onto an Axial mosaic.
+    """
+    if 'angle_hog' in metrics:
```
valosekj: I think this `if` statement can be moved up right after the `# Generate the second QC report image - overlay image with lines for HOG angles` comment -- as there is no need to generate anything if there are no angles available

valosekj: Okay, done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/aece9be90180e4a53ede6258cdb5692dc69247e3

spinalcordtoolbox/process_seg.py
valosekj, line 1: from spine-generic

```
sub-mniS05_T2w.nii.gz
derivatives/labels/sub-mniS05_T2w_label-SC_seg.nii.gz
```

Screenshot 2025-07-29 at 11 28 38 AM

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton, line 1: Hrm. I'm not able to reproduce the "plateau" behavior that you noted.

<details>
<summary> Plots </summary>

Here is the plot from the slides:

image

And here is the plot I generated for `unfErssm004`:

sub-unfErssm004_metrics_PAM50

</details>

Note that when checking out `whole-spine`, there was not a segmentation or manual labels. So, just to get up and running, I ran `sct_deepseg spinalcord` and `sct_label_vertebrae` then renamed the cord seg and disc labels to the appropriate names in your sample command:

> ```
> sct_process_segmentation -i sub-unfErssm004_T2w.nii.gz -s sub-unfErssm004_T2w_label-SC_seg.nii.gz -discfile sub-unfErssm004_T2w_labels-disc-manual.nii.gz -perslice 1 -normalize-PAM50 1 -o sub-unfErssm004_metrics_PAM50.csv -qc ./qc -v 1
> ```

Could you share the exact files you used so that I could debug further?

valosekj: Sorry for not being clear, I used cord segmentation and discs from `derivatives/labels/` (I processed the subject using a [bash script](https://github.com/sct-pipeline/dcm-metric-normalization/blob/d41ceee1fb20f3362df638ab87b3007c32fe251f/hog_angle/compute_diameters.sh#L119-L130)):

```
derivatives/labels/sub-unfErssm004/anat/sub-unfErssm004_T2w_label-SC_seg.nii.gz
derivatives/labels/sub-unfErssm004/anat/sub-unfErssm004_T2w_labels-disc-manual.nii.gz
sub-unfErssm004/anat/sub-unfErssm004_T2w.nii.gz
```

valosekj: It's interesting to see that your curves (segmented by `sct_deepseg spinalcord` using the [contrast-agnostic model v3](https://github.com/sct-pipeline/contrast-agnostic-softseg-spinalcord/releases/tag/v3.0)) differ from the curves from the slides (based on `derivatives/labels/sub-unfErssm004/anat/sub-unfErssm004_T2w_label-SC_seg.nii.gz`, which was, based on its JSON sidecar, segmented by the same model version: `v3.0/model_contrast_agnostic_20250123.zip`). But as the JSON sidecar says, some manual correction was done. But IIRC, manual corrections of the `whole-spine` dataset were supposed to be focused mainly on the spinal cord tip, right, @sandrinebedard?

sandrinebedard: > manual corrections of the whole-spine dataset were supposed to be focused mainly on the spinal cord tip, right, @sandrinebedard?

yes exactly!

spinalcordtoolbox/process_seg.py
```diff
@@ -254,3 +412,191 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
...
+    :return result: Dictionary containing the measured diameters and pixel counts along AP and RL axes.
+    """
+    # Get center of mass (which is computed in the PCA function); centermass_src: [RL, AP] (assuming RPI orientation)
+    # Note: I'm using [rl0, ap0] instead of [y0, x0] to make it easier to track the axes as numpy handle them in a bit unintuitive way :-D
+    rotated_bin = np.array(seg_crop_r_rotated > 0.5, dtype='uint8') # binarize the rotated segmentation for PCA
+    _, _, [rl0, ap0] = compute_pca(rotated_bin)    # same as `y0, x0 = region.centroid`
+    rl0_r, ap0_r = round(rl0), round(ap0)
+
+    # Threshold seg_crop_r_rotated to keep the softness of the segmentation but weight the pixel sum
+    seg_crop_r_rotated_thr = seg_crop_r_rotated * (seg_crop_r_rotated > 0.25)   # TODO: find the best threshold value, 0.5 seems to be too conservative
```
valosekj: As discusssed during the 2025-08-26 DCM meeting, we might not need to threhold the rotated segmentation at all

valosekj: though, we'll probably still need some sort of thresholding/binariztion for the [debug visualization](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/70faf0f2092732b0bb0bafeb89c278de402657f8/spinalcordtoolbox/process_seg.py#L564-L565) if we want to see how the cord is rotated.


<details>
<summary>binarized -- we can easily see the rotation</summary>

slice_118


</details>

<details>
<summary>soft (no threshold) -- we cannot see the rotation</summary>

slice_118 copy


</details>

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: Hum. Now I'm even more confused. After copying over the files from `derivatives/labels`, I still can't seem to reproduce the behavior. Here is a gif for my graph vs the one in the issue:

ezgif-6eedf651a2bb9a

Slight differences due to fonts, but otherwise the diameters are identical but the hog values are much different. :thinking:

spinalcordtoolbox/process_seg.py
valosekj: Okay, I switched to a non-thresholded mask in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/a556190b345665351ff4c751d71ea84d2407c674. (and I use a binarized mask for the debug plotting to show the rotation)

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: In the corresponding slide, you say "Notice that the HOG-based AP diameter (dashed red) is the same across slices" (for slices 443, 445, 447).

When I check the figures for those same slices, I see:

<details>
<summary> My figures </summary>

slice_443

slice_445

slice_447

</details>

The values recorded are `[8.85mm, 8.91mm, 8.91mm]`. But in the slides, you have `[9.00mm, 9.00mm, 9.00mm]`. 

It seems to me that there is a rounding error somewhere that is happening for you but not for me. That would explain the values "clamping" to specific whole numbers (9, 8, etc.). 

What confuses me is the timing... Wait. Ok. If I check the history of the slide deck, it looks like the "plateau" figure was actually uploaded on Aug 11. But commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/830cab7fe7b302c4e4ee9b3d7ce4afcfc7b18c8a was added on Aug 13. 

Is it possible the soft seg fixed the issue?

joshuacwnewton: Yep! If I `git checkout 4d80f8ab67e85afd052875ffea7f2442c482df84` (the commit immediately _before_ the Aug 13 fix ([830cab7](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/830cab7fe7b302c4e4ee9b3d7ce4afcfc7b18c8a))), I am able to reproduce the bug!

So, it looks like it's been resolved by [830cab7](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/830cab7fe7b302c4e4ee9b3d7ce4afcfc7b18c8a)? I am guessing that the binarization caused the cord to "lock" at specific pixel intervals?

sandrinebedard:
After some digging on the orientation of the ellipse, I figured out we were actually using the wrong orientation to generate the debug figures.

skimage gives the orientation between -pi/2 and pi-2 (counter-clockwise). However, in the script, the orientation is modified by the function `fix_orientation`


https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f74cb34419670c10edb121de55367d2ef7c01a1/spinalcordtoolbox/process_seg.py#L224-L234

that returns the absolute of the orientation (which was fine as it was only used to determine which diameter is AP or RL), but this would explain why the ellipse on the images sometimes looked flipped.

Here is the orientation definition: https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3f74cb34419670c10edb121de55367d2ef7c01a1/spinalcordtoolbox/process_seg.py#L199

I have now added an extra parameter with the original orientation to use in the plots (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/a8e9e13dcc8bd17fcc4413d0c6501ae966d0f958)

Example: Sub-321870_acq-axial (dcm-zurich)

Before:
slice_4



After: 

slice_4

valosekj:
Thank you for catching and fixing that, @sandrinebedard! 👏🏻 

Just a short note: the slight difference between the HOG-measured diameters in the figures in [your comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#issuecomment-3234836910) is caused because the "before" figure was generated using a thresholded mask. I now regenerated the figures using a soft mask, and the measured values are the same; the bug concerned the ellipse orientation. 


<details>
<summary>before fixing ellipse orientation (commit a556190)</summary>

slice_4

</details>

<details>
<summary>after (commit a8e9e13)</summary>

slice_4

</details>

spinalcordtoolbox/scripts/sct_process_segmentation.py
valosekj: Thanks for trying the PR, @joshuacwnewton!

> So, it looks like it's been resolved by [830cab7](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/830cab7fe7b302c4e4ee9b3d7ce4afcfc7b18c8a)? I am guessing that the binarization caused the cord to "lock" at specific pixel intervals?

Yes!

spinalcordtoolbox/process_seg.py
```diff
@@ -523,11 +525,11 @@ def _add_ellipse(ax, x0, y0):
             (x0, y0),
             width=properties['diameter_AP'] * upscale / dim[0],
             height=properties['diameter_RL'] * upscale / dim[1],
-            angle=properties['orientation'],
+            angle=properties['orientation_OG']*180.0/math.pi,
             edgecolor='orange',
             facecolor='none',
             linewidth=2,
-            label="Ellipse fitted using skimage.regionprops"
+            label="Ellipse fitted using skimage.regionprops, angle{:.2f}".format(properties['orientation_OG']*180.0/math.pi)
```
valosekj: In [your comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#issuecomment-3234836910), the "after" figure has a positive angle of `15.41`. While in [my comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#issuecomment-3235828998) (after using this commit), the angle is negative of `-15.41`. So maybe there's a line which has not been committed?


Note that we're also changing the sign of the HOG angle for the debug figure:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/a8e9e13dcc8bd17fcc4413d0c6501ae966d0f958/spinalcordtoolbox/process_seg.py#L571-L574

I believe a positive sign makes sense here to be consistent with our QC orientation, which is:

image

Rotated and flipped debug figure to match the QC orientation: 

image

In other words, if the angle is "towards" `R`, we use a positive sign, while if the angle is "towards" `L`, we use a negative sign. Hope it makes sense 😅

spinalcordtoolbox/process_seg.py
```diff
@@ -23,43 +30,82 @@
...
+        if im_seg.dim[:3] != im.dim[:3]:
+            raise ValueError(
+                f"The input segmentation image ({im_seg.path}) and the input image ({im.path}) do not have the same "
+                f"dimensions. Please provide images with the same dimensions."
+            )
+
     # Getting image dimensions. x, y and z respectively correspond to RL, PA and IS.
     nx, ny, nz, nt, px, py, pz, pt = im_seg.dim
-    pr = min([px, py])
+    pr = 0.1
```
valosekj: just to clarify why we use 0.1

```suggestion
    pr = 0.1   # we use a fixed value to be independent from the input image resolution 
```

spinalcordtoolbox/process_seg.py
```diff
@@ -254,3 +423,496 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
...
+    else:
+        _, _, [rl0, ap0] = compute_pca(rotated_bin)    # same as `y0, x0 = region.centroid`
+        rl0_r, ap0_r = round(rl0), round(ap0)
+
+        # Note: seg_crop_r_rotated is soft (due to the rotation) so we sum its values to account for softness
+        # Sum non-zero pixels along AP axis, i.e., the number of pixels in the row corresponding to the center of mass along the RL axis
+        # Compute AP diameter average acrross 3 mm extent centered at rl0_r
+
+        # Use centroid for AP diameter
+        extent_avg = 30  # extent used for averaging the minimum (to account for noise)
```
valosekj:
```suggestion
        extent_avg = 30  # extent of 3 mm (because seg was resampled to 0.1mm) used for averaging the minimum (to account for noise)
```

spinalcordtoolbox/process_seg.py
```diff
@@ -254,3 +423,496 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
...
+        indices = np.array([i for i in range(rl0_r - extent_avg//2, rl0_r + extent_avg//2 + 1)])
```
valosekj: nitpick: indices is a list of 31 items, so technically the extent is 3.1mm (not 3mm)

We can get a list of 30 items if we remove `+ 1`:

```suggestion
        indices = np.array([i for i in range(rl0_r - extent_avg//2, rl0_r + extent_avg//2)])
```

spinalcordtoolbox/process_seg.py
```diff
@@ -254,3 +423,496 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
...
+            _debug_plotting_hog(angle, ap0_r, ap_diameter, dim, iz, properties, rl0_r, properties["diameter_RL_ellipse"],
+                                seg_crop_r_rotated, seg_crop_r, coord_ap)
+
+    return result
+
+
+def _calculate_symmetry(seg_crop_r_rotated, centroid, dim, iz=None, verbose=1):
+    """
+    Compute symmetry metrics by flipping the segmentation along the RL and AP axes and comparing with the original.
+    Calculates Dice coefficient, symmetric difference, and Hausdorff distance for both axes.
```
valosekj, line 490: What about referring to the figure with the symmetry illustration from the PR description? 

```suggestion
    Calculates Dice coefficient, symmetric difference, and Hausdorff distance for both axes. See figure in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#issue-3212203682 for details.
```

spinalcordtoolbox/process_seg.py
```diff
@@ -254,3 +423,496 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
...
+                                         [coords[:, 0], coords[:, 1]],
+                                         order=1).reshape(seg_crop_r.shape)      # order of the spline interpolation --> order=1: linear interpolation
+
+    return seg_crop_r_rotated
+
+
+def _measure_rotated_diameters(seg_crop_r, seg_crop_r_rotated, dim, angle, iz, properties, verbose):
+    """
+    Measure the AP and RL diameters in the rotated segmentation.
+    This function counts the number of pixels along the AP and RL axes in the rotated segmentation and converts them
```
valosekj: I think we could rename this function and update its docstring, as it now measures only AP diameter as average across 3 mm extent. Maybe simply something like `_measure_ap_diameter`?

(I can do it if we agree on it)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -21,6 +21,7 @@
 import numpy as np
 from matplotlib.ticker import MaxNLocator
 
+from spinalcordtoolbox.reports import qc2
```
valosekj, line 24: Adding a comment here, as I am unable to comment on unchanged lines.

 We should update the `diameter_AP` description below (L81) as it's now measured as average across 3 mm extent

> " - diameter_AP, diameter_RL: Finds the major and minor axes of the cord and measure their length.\n"

maybe something like:

```
 - diameter_AP: Measured as average across 3 mm extent centred at cord mask center of mass.\n
 - diameter_RL: Measured as the major axis of the ellipse fitted to the cord.\n
```

spinalcordtoolbox/process_seg.py
```diff
@@ -155,69 +301,92 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
         solidity = np.nan
     else:
         solidity = region.solidity
+
     # Fill up dictionary
     properties = {
         'area': area,
-        'diameter_AP': diameter_AP,
+        'diameter_AP_ellipse': diameter_AP,
+        'diameter_RL_ellipse': diameter_RL,
```
valosekj:
```python
'diameter_RL_ellipse': diameter_RL,
'diameter_RL': diameter_RL,
```
As we decided to stick with the ellipse-based `diameter_RL` as the default, I think we can keep only `diameter_RL` and remove `diameter_RL_ellipse` (and replace its occurrences in the script with `diameter_RL`), right?

sandrinebedard: done here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/e5c893706e4fdd2384a2f7161da098b5db7a50b8

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/6442611408586771076b5637916ae5e9d1c33877

spinalcordtoolbox/process_seg.py
sandrinebedard: done here: 
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/40e0b69d4b8ca64b245933ddb46a3370058fde57

sandrinebedard:
@joshuacwnewton I think the PR is ready for you to review!

Maybe one thing left to discuss, right now the parameter `filter_size` is set by default, so there is always regularization regarding the HOG angle, is this the behaviour we want, or do we want to include this as a parameter the user can set? (as the code is made ro have the option of regularization or not), maybe @valosekj you have an insight on this as you implemented this part?

spinalcordtoolbox/process_seg.py
```diff
@@ -23,43 +30,82 @@
...
     """
-    # List of properties to output (in the right order)
+    # List of properties that are always available
     property_list = ['area',
                      'angle_AP',
                      'angle_RL',
                      'diameter_AP',
+                     'diameter_AP_ellipse',
                      'diameter_RL',
+                     'diameter_RL_ellipse',
```
joshuacwnewton: It looks like the tests are currently failing because of the following error:

```python
  File "/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 634, in <module>
    main(sys.argv[1:])
  File "/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 511, in main
    metrics, fit_results = compute_shape(fname_segmentation,
  File "/Users/runner/work/spinalcordtoolbox/spinalcordtoolbox/spinalcordtoolbox/process_seg.py", line 241, in compute_shape
    shape_properties[property_name][iz] = shape_property[property_name]
KeyError: 'diameter_RL_ellipse'
```

I think `'diameter_RL_ellipse'` is the right key to use. However, we're assigning it to the `'diameter_RL'` key instead:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2c833990b79cb8e6c57fe6dc9c4b7fe3cf9832bd/spinalcordtoolbox/process_seg.py#L359-L360

If I understand correctly, the fix should be to rename all instances of the `'diameter_RL'` key to `'diameter_RL_ellipse'` then remove `diameter_RL` from the initial property dictionary (since we don't compute the HOG-based RL diameter, right?).

sandrinebedard: oups, while addressing @valosekj comment (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#discussion_r2476742816) I did not test it after, will fix, sorry!

sandrinebedard: now fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/53e15092243fd56c0d904b219071e46fb95021a7

sandrinebedard: the tests passsed now!

joshuacwnewton (commented):
I'm just going to create a small commit that moves the debug plotting functions into their own file. (That way, it's a little easier for me to distinguish between the core changes and the debugging changes.)

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,103 @@
...
-def compute_shape(segmentation, angle_correction=True, centerline_path=None, param_centerline=None,
-                  verbose=1, remove_temp_files=1):
+def compute_shape(segmentation, image=None, angle_correction=True, centerline_path=None, param_centerline=None,
+                  verbose=1, remove_temp_files=1, filter_size=5):
     """
     Compute morphometric measures of the spinal cord in the transverse (axial) plane from the segmentation.
     The segmentation could be binary or weighted for partial volume [0,1].
 
     :param segmentation: input segmentation. Could be either an Image or a file name.
+    :param image: input image. Could be either an Image or a file name.
```
joshuacwnewton: It might be good to mention that `image` is necessary to turn on HOG/symmetry-based metrics.

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -257,6 +258,13 @@ def get_parser(ascor=False):
         help="Extent (in mm) for the mask used to compute morphometric measures. Each slice covered by the mask is "
              "included in the calculation. (To be used with flag `-pmj` and `-pmj-distance`.)"
     )
+    optional.add_argument(
+        '-symmetry',
+        metavar=Metavar.file,
+        default=None,
+        help="Input image used to compute spinal cord orientation (using HOG method)."
+             "Example: t2.nii.gz"
+    )
```
joshuacwnewton, line 273: I'm wondering if we should instead call this argument something like `-anat`, move it between `-i` and `-o` and say something like: "Anatomical image used to compute orientation-based metrics (quadrant area, symmetry, etc.)"

That way, the argument stays general, and so if we add additional anatomical image-based metrics in the future, the name "`-symmetry`" won't become outdated.

(Also, it would better match the API usage, because we use `fname_image` as the variable, and `image` as the parameter in the `compute_shape` function.)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -410,6 +418,7 @@ def main(argv: Sequence[str]):
     group_funcs = (('MEAN', func_wa), ('STD', func_std))  # functions to perform when aggregating metrics along S-I
 
     fname_segmentation = get_absolute_path(arguments.i)
+    fname_image = get_absolute_path(arguments.symmetry) if arguments.symmetry is not None else None
```
joshuacwnewton: I notice that we use `get_absolute_path` in the `type` of the `-pmj` argument:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6442611408586771076b5637916ae5e9d1c33877/spinalcordtoolbox/scripts/sct_process_segmentation.py#L239-L245

I'm wondering if it would be better to use it for `-i` and `-symmetry`, too? That would allow us to avoid the awkward "is not None else None" pattern. (Since if the path isn't provided, the value will still be `None`.)

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -559,12 +559,30 @@ def save_as_csv(agg_metric, fname_out, fname_in=None, append=False):
...
+                'MEAN(symmetry_hausdorff_AP)', 'STD(symmetry_hausdorff_AP)',
+                'MEAN(symmetry_difference_AP)', 'STD(symmetry_difference_AP)',
+                'MEAN(symmetry_anterior_RL)', 'STD(symmetry_anterior_RL)',
+                'MEAN(symmetry_posterior_RL)', 'STD(symmetry_posterior_RL)',
+                'MEAN(area_quadrant_anterior_left)', 'STD(area_quadrant_anterior_left)',
+                'MEAN(area_quadrant_anterior_right)', 'STD(area_quadrant_anterior_right)',
+                'MEAN(area_quadrant_posterior_left)', 'STD(area_quadrant_posterior_left)',
+                'MEAN(area_quadrant_posterior_right)', 'STD(area_quadrant_posterior_right)',
+                'SUM(length)', 'WA()', 'BIN()', 'ML()', 'MAP()', 'MEDIAN()', 'STD()', 'MAX()'
+    ]    # TODO: if append=True but file does not exist yet, raise warning and set append=False
```
joshuacwnewton: This list is getting... very very long! I'm wondering if we want to keep a "single source of truth" for all of these `sct_process_segmentation` fields.

For example, we could do something like this:

```python
list_item = ['Label', 'Size [vox]']
for m in PROCESS_SEGMENTATION_METRICS:
    list_item.extend([f"MEAN({m})", f"STD({m})"])
list_item.extend(['SUM(length)', 'WA()', 'BIN()', 'ML()', 'MAP()', 'MEDIAN()', 'STD()', 'MAX()']
```

Then, we define `PROCESS_SEGMENTATION_METRICS` inside of `process_seg.py`, but at the top of the file instead of inside of `compute_shape`.

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,103 @@
...
-    # List of properties to output (in the right order)
+    # List of properties that are always available
     property_list = ['area',
                      'angle_AP',
                      'angle_RL',
                      'diameter_AP',
+                     'diameter_AP_ellipse',
                      'diameter_RL',
                      'eccentricity',
+                     'orientation_abs',
```
joshuacwnewton: Neither of the new properties are listed inside of the `aggregate_metric.py` list of properties (`list_item`). Is this intentional?

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,103 @@
...
                      'orientation',
                      'solidity',
                      'length'
                      ]
```
joshuacwnewton: As per the previous suggestion, these could be a module-level constant (`KEYS_DEFAULT`) and then we could initialize the property list with `property_list = KEYS_DEFAULT`, that way they are available to other scripts.

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,103 @@
...
+        ]
+        # Add symmetry properties
+        symmetry_keys = [
+            'symmetry_dice_RL',
+            'symmetry_hausdorff_RL',
+            'symmetry_difference_RL',
+            'symmetry_dice_AP',
+            'symmetry_hausdorff_AP',
+            'symmetry_difference_AP',
+        ]
```
joshuacwnewton: We could do a similar thing here -- move these to be module-level constants (`KEYS_HOG`, `KEYS_QUADRANT`, `KEYS_SYMMETRY`) then construct the property list the same way.

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,103 @@
...
+        if im_seg.dim[:3] != im.dim[:3]:
+            raise ValueError(
+                f"The input segmentation image ({im_seg.path}) and the input image ({im.path}) do not have the same "
+                f"dimensions. Please provide images with the same dimensions."
+            )
+
     # Getting image dimensions. x, y and z respectively correspond to RL, PA and IS.
     nx, ny, nz, nt, px, py, pz, pt = im_seg.dim
-    pr = min([px, py])
+    pr = 0.1   # we use a fixed value to be independent from the input image resolution
```
joshuacwnewton, line 85: This seems like a pretty big change! Is there any reason you chose `0.1` in particular? 

I'm curious what effect this has on the actual metric computation / execution time / etc. (since `0.1` is relatively high-resolution).

spinalcordtoolbox/process_seg.py
```diff
@@ -99,6 +151,14 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
                                ncols=80):
         # Extract 2D patch
         current_patch = im_segr.data[:, :, iz]
+        # Special check for the edge case when segmentation has only a single pixel (e.g., in the lumbar region),
+        # in this case we skip the slice as we cannot compute PCA
+        # Change Nan for zeros to avoid problems in PCA computation
+        if np.count_nonzero(current_patch) <= 1/0.1:  # equivalent to np.sum(current_patch) <= 1 after resampling
```
joshuacwnewton: I think it's important to define a variable earlier called `p_resample = 0.1` (since we use the value "0.1" in multiple places, and we don't want to cause a bug by changing it in one place but not the other).

Also, I think rearranging the variables might be a bit clearer?

```python
if np.count_nonzero(current_patch) * p_resample <= 1:
```

If we do this, then I think we don't need the `# equivalent to` comment anymore, since it's basically self-explanatory.

spinalcordtoolbox/process_seg.py
```diff
@@ -116,34 +176,112 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
-            for property_name in property_list:
-                shape_properties[property_name][iz] = shape_property[property_name]
+
+        # Store basic properties and angles to be used later after regularization
+        if image is not None:
+            # compute PCA and get center or mass based on segmentation; centermass_src: [RL, AP] (assuming RPI orientation)
+            # Check for empty slice
+            if np.count_nonzero(current_patch_scaled) <= 1/0.1:  # equivalent to np.sum(current_patch_scaled) <= 1 after resampling
+                logging.warning(f'Skipping slice {iz} as the segmentation contains only a single pixel.')
+                continue
```
joshuacwnewton: This check is duplicated, I think? (And if the first check occurs, then it will be `continue`'d, so we'd never reach this point I think.)

spinalcordtoolbox/process_seg.py
```diff
@@ -99,6 +151,14 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
         # Extract 2D patch
         current_patch = im_segr.data[:, :, iz]
+        # Special check for the edge case when segmentation has only a single pixel (e.g., in the lumbar region),
+        # in this case we skip the slice as we cannot compute PCA
+        # Change Nan for zeros to avoid problems in PCA computation
+        if np.count_nonzero(current_patch) <= 1/0.1:  # equivalent to np.sum(current_patch) <= 1 after resampling
+            logging.warning(f'Skipping slice {iz} as the segmentation contains only a single pixel.')
+            continue
+
+        current_patch_im = im_r.data[:, :, iz] if image is not None else None
```
joshuacwnewton: I think it would be a little clearer if we assigned this variable right under `current_patch` (so that they line up)

spinalcordtoolbox/process_seg.py
```diff
@@ -116,34 +176,112 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
                                                   output_shape=current_patch.shape,
                                                   order=1,
                                                   )
+            if image is not None:
+                current_patch_im = current_patch_im.astype(np.float64)
+                current_patch_im_scaled = transform.warp(current_patch_im,
+                                                         tform.inverse,
+                                                         output_shape=current_patch_im.shape,
+                                                         order=1,
+                                                         )
```
joshuacwnewton: Is the `np.float64` only for the transformation? 

If so, then it's probably better to write:


```python
                current_patch_im_scaled = transform.warp(current_patch_im.astype(np.float64),
                                                         tform.inverse,
                                                         output_shape=current_patch_im.shape,
                                                         order=1,
                                                         )
```

However, if we do want it to be `float64` all the time, then we should probably do that at the very start.

spinalcordtoolbox/process_seg.py
```diff
@@ -116,34 +176,112 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
         else:
             current_patch_scaled = current_patch
+            current_patch_im_scaled = current_patch_im if current_patch_im is not None else None
```
joshuacwnewton: Luckily, we don't need the `is not None else None`, since if it's None then we can just assign that too:


```suggestion
            current_patch_im_scaled = current_patch_im
```

spinalcordtoolbox/process_seg.py
```diff
@@ -116,34 +176,112 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
+                shape_property['centermass_y'] = centermass_src[1]
+                shape_property['angle_hog'] = -angle_hog * 180.0 / math.pi     # degrees, and change sign to match negative if left rotation
+                shape_property['angle_AP'] = angle_AP_rad * 180.0 / math.pi     # convert to degrees
+                shape_property['angle_RL'] = angle_RL_rad * 180.0 / math.pi     # convert to degrees
+                shape_property['length'] = pz / (np.cos(angle_AP_rad) * np.cos(angle_RL_rad))
+                # Loop across properties and assign values for function output
+                for property_name in property_list:
+                    shape_properties[property_name][iz] = shape_property[property_name]
+            else:
+                logging.warning(f'\nNo properties for slice: {iz}')
```
joshuacwnewton: Is there any way that we can de-duplicate some of the shape property logic in these blocks of code?
We're performing the same calculations multiple times, and the control flow is quite complex:

- `if image is None or filter_size < 0`:
  - `if image is not None`:
- `if image is not None and filter_size > 0 and len(z_indices) > 0`:

I wonder if we can simplify this at all... For example, in the previous code, we would run the `# compute shape properties on 2D patch code unconditionally`. Could we leave this as-is, and simply add the image-specific properties afterwards?

Likewise, maybe we take the HOG properties out of `_properties2d`, and leave that function as-is. Then, we compute the HOG properties separately, adding them to `properties` later on? (That way also we don't need `iz` just for plotting in `_properties2d`.)

spinalcordtoolbox/process_seg.py
```diff
@@ -253,3 +413,254 @@ def _find_AP_and_RL_diameter(major_axis, minor_axis, orientation, dim):
     diameter_AP *= dim[0]
     diameter_RL *= dim[1]
     return diameter_AP, diameter_RL
+
+
+def _rotate_segmentation_by_angle(seg_crop_r, angle):
```
joshuacwnewton, line 392: The 4 functions at the end of the script are all really great! I love that they are straightforward in terms of the function signatures (a few variables in, 1 quantity out). They're very neat and tidy. :)

spinalcordtoolbox/process_seg.py
joshuacwnewton: Oh wait, I guess this could occur if post-transformation there's a single-voxel slice. Hmm... this is tricky.

I guess we can remove the first check instead? Since, if there's no angle correction, current_patch_scaled will be set to current_patch (and the check will be equivalent).

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: Maybe we should discuss this at an SCT meeting? we should maybe dig a bit more into the other functions we have, making sure it is consistent, maybe we have other cases like this?
@valosekj @jcohenadad what do you think about this?

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: thanks for the suggestion, I had a few questions on how to implement:

- what about the propreties that are used inside `compute_shape`, but not in the output csv? (like `centermass_x`)
- would we add this `PROCESS_SEGMENTATION_METRICS` as a new input to `save_as_csv` function? or export the variable to the aggrgate_slicewise?

spinalcordtoolbox/process_seg.py
sandrinebedard: `diameter_AP_ellipse` is used for the debug figures
we can remove  `orientation_abs`

spinalcordtoolbox/process_seg.py
sandrinebedard: This is to make sure we resample to the same resolution all the images instead of doing a resampling to the highest resolution and then upscale, 0.1 is pretty high, but we removed the upscale, so pretty similar to what we had

indeed the values of CSA changed a bit, I noticed in the tests
tagging @jcohenadad and @valosekj as this was decided with them

spinalcordtoolbox/scripts/sct_process_segmentation.py
jcohenadad: I like the '-anat' flag because it makes it explicit what we are asking for. Also, it's true that other features might come up in the future, where the anat will be required.

However it makes the 'symmetry' computation less explicit. One possibility, which is a big overhead, would be to also add a flag `-symmetry 0(default)/1` if ppl want it, and check that if symmetry=1, an anat is provided.

spinalcordtoolbox/process_seg.py
sandrinebedard: actually, there is already a variable for `p_resample` -->  `pr`, so going to use it

spinalcordtoolbox/process_seg.py
sandrinebedard: I think the complicated thing here is that if filter_size > 0 (so if we want regularization), we need to go through the slices twice

jan implemented this part before I chipped in, so I was also confused about the logic, I tried to simplify it, but still is complicated, so whatever you think is best!

spinalcordtoolbox/process_seg.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/8e1c088f19b421cb2e7de2112cd8e7b6e9c81f00

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: chnaged https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/02c21c16509a99f89e434327b313d52c0152b7d2

spinalcordtoolbox/process_seg.py
sandrinebedard: changed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/af505b5790e4ff584a4b54cf4ec88da0bcc8d9d3

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: changed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/af505b5790e4ff584a4b54cf4ec88da0bcc8d9d3

but no yet inside [aggregate_slicewise.py](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/files/8b4518bea73526fcf6afdce33d7c46dc54ce9826#diff-cdea0cf730b01ec60e2321dc470f9eb00379b9316145826fbb30faf7604e0a5c)

spinalcordtoolbox/process_seg.py
sandrinebedard: changed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/4bb2129c0880b1b545dc2cfff3eaade7bd6550f4

spinalcordtoolbox/process_seg.py
sandrinebedard: done https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/af505b5790e4ff584a4b54cf4ec88da0bcc8d9d3

spinalcordtoolbox/process_seg.py
sandrinebedard: fixed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/af505b5790e4ff584a4b54cf4ec88da0bcc8d9d3

spinalcordtoolbox/process_seg.py
sandrinebedard: fixed here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/1f90baf319b7ca08150452f7afdd73aeedbc76f5

spinalcordtoolbox/process_seg.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/ed6f024f1b43ebdcfc9e48357ff62755c41c1301

spinalcordtoolbox/process_seg.py
sandrinebedard: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/e6f94458c0f819459ed71e7fdffbdba3473e8d92

spinalcordtoolbox/process_seg.py
sandrinebedard: dne in here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/1f90baf319b7ca08150452f7afdd73aeedbc76f5

spinalcordtoolbox/scripts/sct_process_segmentation.py
valosekj: Good idea, @joshuacwnewton! I also like `-anat`. We could update its help to mention symmetry, e.g.:

```python
    optional.add_argument(
        '-anat',
        metavar=Metavar.file,
        default=None,
        help="Anatomical image used to estimate spinal cord orientation with the HOG method, "
            "required for symmetry computation. Example: t2.nii.gz"
    )
```

And we can (should) add symmetry also to the list of morphometrics in function description:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e6f94458c0f819459ed71e7fdffbdba3473e8d92/spinalcordtoolbox/scripts/sct_process_segmentation.py#L88-L92

spinalcordtoolbox/process_seg.py
valosekj: right, I implemented the second part of the code snippet (starting from L224) exactly as Sandrine says: we first need to iterate over slices to find HOG angles per slice:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/e6f94458c0f819459ed71e7fdffbdba3473e8d92/spinalcordtoolbox/process_seg.py#L182-L185

To be able to regularize (smooth the HOG angles):

```python
        angle_hog_array = np.array(angle_hog_values)
        # Apply Gaussian filter to regularize the angles
        angle_hog_regularized = gaussian_filter1d(angle_hog_array, filter_size)
```

but looking at it now, I guess it can be indeed simplified 😅

valosekj: Also, I think we can remove this comment:

```python
    # The code snippet below is taken from algorithms.register2d_centermassrot -- maybe it could be extracted into a
    # function and reused
```

It originally referred to [this section](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/bdab828fee9b26c14388669bda8c727715f14efa/spinalcordtoolbox/registration/algorithms.py#L885), but as you extracted plotting into `create_regularized_hog_angle_plot`, it's been addressed.

spinalcordtoolbox/scripts/sct_process_segmentation.py
joshuacwnewton: > However it makes the 'symmetry' computation less explicit.

You're right, this is tricky... 

That said, there is precedence for this pattern? I'm reminded of the `-s`/`-i`/`-f` arguments for `sct_analyze_lesion` (where each one contains a note of "if provided, XXX metric will be computed"). Maybe adding tutorials and notes will be enough to help with discoverability? :thinking:

spinalcordtoolbox/process_seg.py
```diff
@@ -29,6 +29,22 @@
...
+                'eccentricity',
+                'orientation',
+                'solidity',
+                'length'
+                ]
+KEYS_HOG = ['centermass_x', 'centermass_y', 'angle_hog']
+KEYS_QUADRANT = ['area_quadrant_anterior_left', 'area_quadrant_anterior_right', 'area_quadrant_posterior_left',
+                 'area_quadrant_posterior_right']
+KEYS_SYMMETRY = ['symmetry_dice_RL', 'symmetry_hausdorff_RL', 'symmetry_difference_RL', 'symmetry_dice_AP',
+                 'symmetry_hausdorff_AP', 'symmetry_difference_AP']
```
joshuacwnewton: Tiny tiny nit (because it makes it easier to see the symmetry between the two lines)

```suggestion
KEYS_QUADRANT = ['area_quadrant_anterior_left', 'area_quadrant_anterior_right', 
                 'area_quadrant_posterior_left', 'area_quadrant_posterior_right']
KEYS_SYMMETRY = ['symmetry_dice_RL', 'symmetry_hausdorff_RL', 'symmetry_difference_RL',
                 'symmetry_dice_AP', 'symmetry_hausdorff_AP', 'symmetry_difference_AP']
```

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,89 @@
...
-                     'length'
-                     ]
+    # List of properties that are always available
+    property_list = KEYS_DEFAULT
 
     im_seg = Image(segmentation).change_orientation('RPI')
+    # Check if the input image is provided (i.e., image is not None)
+    if image is not None:
+        # HOG-related properties that are only available when image (`sct_process_segmentation -i`) is provided
+        # TODO: consider whether to use this workaround or include the columns even when image is not provided and use NaN
```
joshuacwnewton: Reminder for TODO...

(Personally, I think I'm happy with the workaround, as I don't want to clutter the output with a bunch of blank columns.)

spinalcordtoolbox/process_seg_debug.py
joshuacwnewton, line 1: I'm wondering if it might make some sense to make a new API module to group all of the morphometrics-related modules? Something like:

```
spinalcordtoolbox/
├─ morphometrics/
│  ├─ aggregate_slicewise.py
│  ├─ csa_pmj.py
│  ├─ metrics_to_PAM50.py
│  ├─ plots.py
│  ├─ process_seg.py
│  └─ [...]
└─ [...]
```

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -77,7 +78,8 @@ def get_parser(ascor=False):
             "accounted for by inputing a mask comprising values within [0,1]. Can be normalized when specifying the flag `-normalize`\n"
             "  - angle_AP, angle_RL: Estimated angle between the cord centerline and the axial slice. This angle is "
             "used to correct for morphometric information.\n"
-            "  - diameter_AP, diameter_RL: Finds the major and minor axes of the cord and measure their length.\n"
+            "  - diameter_AP: Measured as average across 3 mm extent centred at cord mask center of mass.\n"
+            "  - diameter_RL: Measured as the major axis of the ellipse fitted to the cord.\n"
```
joshuacwnewton, line 82: TODO: Add `symmetry` and other metrics and make sure they mention the requirement of `-anat` (assuming we change to that name)

joshuacwnewton (changes requested):
Most of the changes look good to me so far, but my main concern is still the complexity of the `compute_shape` function (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#discussion_r2524757461).

I have ideas on how to make it simpler, but it is a bit tricky to implement. I'm going to try to implement myself on a separate branch (just to showcase the changes), but it may take me a little while. I hope to have it done by end of day.

spinalcordtoolbox/process_seg.py
joshuacwnewton: @sandrinebedard @valosekj I have finished a refactor of the `image`-based properties.

- The full branch containing my changes is here: [`jn/refactor_hog_logic`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/jv/sct_process_segmentation_use_hog...jn/refactor_hog_logic?expand=1)
- A commit that highlights the image-specific logic is here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/2e4112e12ac4deba9641c9e931def6b911d8e211

Basically, these changes move most of the `if image is not None:` logic into a function, allowing us to consider those properties separately from the "base" properties (when `image` is not present).

I followed a step-by-step approach for each individual commit to isolate the `if image is not None:` logic, and made sure the tests pass at every step, so I think this should be good? But, please feel free to take a look over my changes first commit by commit.

If everything looks good, I could merge the changes into this branch and then review the code further.

spinalcordtoolbox/process_seg_debug.py
joshuacwnewton: It looks like there's an open issue for this:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4624

In that case we can always do that later, no need to do it now.

spinalcordtoolbox/process_seg.py
sandrinebedard: amazing! thank you @joshuacwnewton !

spinalcordtoolbox/scripts/sct_process_segmentation.py
valosekj: okay, we agreed on `-anat` during the 2025-12-02 SCT dev meeting

spinalcordtoolbox/process_seg.py
joshuacwnewton: So sorry for the delay!! I've added all my changes from the above branch, then I simplified the "image properties" function even further. I'm really happy with how things look now!

> right, I implemented the second part of the code snippet (starting from L224) exactly as Sandrine says: we first need to iterate over slices to find HOG angles per slice:

I managed to de-duplicate the `properties2d` call, working around the angle regularization that you mention, too! :tada:

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Note: `angle_AP_ellipse` is also not in `aggregate_slicewise.py`, not sure if we want to include that or not.

sandrinebedard: I think it is only used for debugging figures, but to validate

spinalcordtoolbox/process_seg.py
sandrinebedard: agreed!

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: Hi @joshuacwnewton , I realized that you did not respond to these questions:

> thanks for the suggestion, I had a few questions on how to implement:
> * what about the propreties that are used inside compute_shape, but not in the output csv? (like centermass_x)
> * would we add this PROCESS_SEGMENTATION_METRICS as a new input to save_as_csv function? or export the variable to the aggrgate_slicewise?

It would me understand how to implement this suggestion, thanks!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/00a615cf62f6f27aa9f5450a72e14270f93e2275

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/73dd769c646f4fd8bb93aa741ec6e44276f508f3

sandrinebedard:
Hi @joshuacwnewton I think this is the last feedbask to address here(https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958#discussion_r2583153569) but I wasn't sure how to implement it

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: My apologies for the late reply!

What I would do is just make sure that all of the internal properties are part of their own constant. For example, you have already separated out `centermass_x` as part of the `KEYS_HOG` constant, so we can just exclude that constant.

If I understand correctly, the following should work?

```python
list_item = ['Label', 'Size [vox]']
for k in (KEYS_DEFAULT + KEYS_QUADRANT + KEYS_SYMMETRY):
    if k != 'length':  # `length` is added as `SUM()` later
       list_item.extend([f"MEAN({k})", f"STD({k})"])
list_item.extend(['SUM(length)', 'WA()', 'BIN()', 'ML()', 'MAP()', 'MEDIAN()', 'STD()', 'MAX()']
```

spinalcordtoolbox/process_seg.py
```diff
@@ -154,69 +188,174 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
+        shape_property = _properties2d(current_patch_scaled, [px, py], iz, angle_hog=angle_hog, verbose=verbose)
+
+        if shape_property is not None:
+            # Add custom fields
+            shape_property['centermass_x'] = centermass_src[0]
+            shape_property['centermass_y'] = centermass_src[1]
+            shape_property['angle_hog'] = -angle_hog * 180.0 / math.pi     # degrees, and change sign to match negative if left rotation
+            shape_property['angle_AP'] = angle_AP_rad * 180.0 / math.pi     # convert to degrees
+            shape_property['angle_RL'] = angle_RL_rad * 180.0 / math.pi     # convert to degrees
+            shape_property['length'] = pz / (np.cos(angle_AP_rad) * np.cos(angle_RL_rad))
```
joshuacwnewton: I missed this before, but these properties don't actually change when `angle_hog` is provided, right? 

If they are identical, then we can probably remove these lines from the `_properties_image` function (since they will have already been computed in `compute_shape`).

spinalcordtoolbox/process_seg.py
```diff
@@ -154,69 +188,174 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
...
+            )
+    # If filter_size <= 0, then just use the original angles
+    else:
+        angle_hog_regularized = angle_hog_values
+
+    # Now compute shape properties using the new angles
+    for iz, angle_hog, centermass_src in zip(z_indices, angle_hog_regularized, centermass_values):
+        current_patch_scaled = current_patches[iz]['patch']
+        angle_AP_rad = current_patches[iz]['angle_AP_rad']
+        angle_RL_rad = current_patches[iz]['angle_RL_rad']
```
joshuacwnewton: Ditto...

testing/api/test_process_seg.py
joshuacwnewton, line 1: Just a small reminder that we need to update the cached `batch_processing.sh` CSV files.

To make this easier, we can add a step to the batch processing GHA `.yml` workflow file that will upload the CSV files as an artifact:

```yml
    - name: "Upload CSV files for easier batch_processing.sh updating"
      uses: actions/upload-artifact@v4
      with:
        name: Batch Processing CSV files (${{ matrix.os }})
        path: data/sct_example_data/**/*.csv
```

I'm honestly really surprised I hadn't done that yet, since I did it for `sct_tutorial_data` in https://github.com/spinalcordtoolbox/sct_tutorial_data/pull/31. 🤦

testing/cli/test_cli_sct_process_segmentation.py
```diff
@@ -195,7 +210,26 @@ def test_sct_process_segmentation_check_discfile(tmp_path):
...
-        assert float(row['MEAN(area)']) == pytest.approx(78.58643257836141)
+        assert float(row['MEAN(area)']) == pytest.approx(78.17036713469571)
+
+
+def test_sct_process_segmentation_symmetry(tmp_path):
+    """ Run sct_process_segmentation with -symmetry and check the results"""
+    filename = str(tmp_path / 'tmp_file_out.csv')
+    sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'),
+                                        '-vert', '1:10', '-perslice', '1',
+                                        '-discfile', sct_test_path('t2', 'labels.nii.gz'), '-symmetry', sct_test_path('t2', 't2.nii.gz'), '-o', filename])
```
joshuacwnewton: oopsie


```suggestion
def test_sct_process_segmentation_anat_properties(tmp_path):
    """ Run sct_process_segmentation with -anat and check the symmetry results"""
    filename = str(tmp_path / 'tmp_file_out.csv')
    sct_process_segmentation.main(argv=['-i', sct_test_path('t2', 't2_seg-manual.nii.gz'),
                                        '-vert', '1:10', '-perslice', '1',
                                        '-discfile', sct_test_path('t2', 'labels.nii.gz'), '-anat', sct_test_path('t2', 't2.nii.gz'), '-o', filename])
```

spinalcordtoolbox/process_seg.py
```diff
@@ -9,57 +9,85 @@
...
+)
+
 # NB: We use a threshold to check if an array is empty, instead of checking if it's exactly 0. This is because
 # resampling can change 0 -> ~0 (e.g. 1e-16). See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3402
 NEAR_ZERO_THRESHOLD = 1e-6
+KEYS_DEFAULT = ['area',
+                'angle_AP',
+                'angle_RL',
+                'diameter_AP',
+                'diameter_AP_ellipse',
```
joshuacwnewton, line 36: Is there a reason we store `diameter_AP_ellipse`? Is it just for debugging purposes, or...

I'm wondering if we should just exclude this key entirely and stop storing the elliptical AP diameter (internally as well), since we now plan to always use the rotseg AP diameter instead... :thinking:

testing/cli/test_cli_sct_process_segmentation.py
joshuacwnewton: We should probably test the new `-qc` flag in this test too, right? (remember to use `tmp_path_qc` (a fixture that gets used similarly to `tmp_path`) so that all the qc reports get created in the same location)

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -598,10 +617,21 @@ def main(argv: Sequence[str]):
                 parser.error('-qc-image is required to display QC report.')
         else:
             logger.warning('QC report only available for PMJ-based CSA. QC report not generated.')
-    # Clean up temp
-    if arguments.r and temp_folder is not None:
-        logger.info("\nRemove temporary files...")
-        temp_folder.cleanup()
```
joshuacwnewton, line 604: Ah, shoot, we should probably merge the two `if path_qc is not None` blocks and also update the warning so that it mentions "only available for `-pmj` or `-anat`".

Right now, I think this warning will be thrown even for `-anat -qc`

spinalcordtoolbox/reports/qc2.py
```diff
@@ -277,6 +278,134 @@ def sct_register_multimodal(
...
+            y_end = y_mosaic + radius[1] / 2 * np.cos(-angle_deg * np.pi / 180)
+            # Plot the line
+            ax.plot([x_start, x_end], [y_start, y_end], '-', color='red', linewidth=0.7)
+            # Include the angle text in degrees
+            # Flip sign to match PCA convention
+            # See https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ba30577e80a4e7387498820f0ff30b8965fbf2a4/spinalcordtoolbox/registration/algorithms.py#L834
+            # TODO: figure out why the link below flip the angle sign only for src_hog but not for dest_hog
+            ax.text(x_mosaic + radius[0] * 0.2, y_mosaic - radius[1] * 0.3,  # upper right corner
+                    f'{angle_deg:.1f}°', color='red', fontsize=3,
+                    path_effects=[mpl_patheffects.withStroke(linewidth=0.5, foreground='white')])
```
joshuacwnewton: Personally I find the white outline to be a little bit hard to see. To stay consistent with the slice numbering in `sct_deepseg`, can we try a black outline instead with linewidth 0.75?


```suggestion
                    path_effects=[mpl_patheffects.withStroke(linewidth=0.75, foreground='black')])
```

Image

Image

spinalcordtoolbox/process_seg.py
sandrinebedard: Since the RL is still the ellipse, I think it is still relevant, the orientation is also taken from the ellipse fitting, and to plot the ellipse, we need both major and minor axis

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: but the constants `KEYS_DEFAULT` etc are defined in `proecess_seg.py` and not in `aggregate_slicewise.py` that was more my concern, how to bring them there?

joshuacwnewton: Oh, sorry! I wasn't clear about that. Just:

```python
from spinalcordtoolbox.process_seg import KEYS_DEFAULT, KEYS_QUADRANT, KEYS_SYMMETRY
```

at the top of `aggregate_slicewise` should work just fine. :)

spinalcordtoolbox/process_seg.py
joshuacwnewton: Ohhhh okay, that makes sense!

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: When adding this:

> from spinalcordtoolbox.process_seg import KEYS_DEFAULT, KEYS_QUADRANT, KEYS_SYMMETRY


I got the following error:

```
File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_process_segmentation.py", line 24, in <module>
    from spinalcordtoolbox.reports import qc2
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/reports/qc2.py", line 33, in <module>
    from spinalcordtoolbox.aggregate_slicewise import Metric
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/aggregate_slicewise.py", line 24, in <module>
    from spinalcordtoolbox.process_seg import KEYS_DEFAULT, KEYS_QUADRANT, KEYS_SYMMETRY
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/process_seg.py", line 17, in <module>
    from spinalcordtoolbox.aggregate_slicewise import Metric
ImportError: cannot import name 'Metric' from partially initialized module 'spinalcordtoolbox.aggregate_slicewise' (most likely due to a circular import) (/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/aggregate_slicewise.py)
```

not sure what to do...

spinalcordtoolbox/process_seg.py
sandrinebedard: done here https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/a8daf0d1dcbfeab57c7ab5ea5b4700b32be29f63

spinalcordtoolbox/process_seg.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/a8daf0d1dcbfeab57c7ab5ea5b4700b32be29f63

testing/cli/test_cli_sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/c45be25ccf9ee15cf3bd837d01ef24240cd21719

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/da87e3c788cc7f3d94729899e471b9180310ea03

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958/commits/b35a5e030553ab898374d624c0e2805a1272976d

PR #4961, opened 2025-07-09
`sct_analyze_lesion` QC restrict lesion to spinal cord

valosekj:
## Checklist

## Description

This is a tiny PR fixing an indexing error when generating tissue bridges QC for `sct_analyze_lesion`.

### Context:

I added the restriction of the lesion mask to the spinal cord mask -- the same operation is also done in the main `sct_analyze_lesion` script. In (rare) cases where the lesion mask contains slices outside spinal cord mask, the QC generation would fail because we are getting unique sag slices with lesions using `np.unique(np.where(im_lesion_data)[0])` and using the slice numbers to fetch tissue bridges from the XLSX file (which, however, was created from the restricted lesion and thus doesn't contain lesion slices outside the cord):

```
*** Generating Quality Control (QC) html report ***
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3802, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 138, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 165, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 5745, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 5753, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'slice_7_dorsal_bridge_width [mm]'

The above exception was the direct cause of the following exception:

  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 1216, in <module>
    main(sys.argv[1:])
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/scripts/sct_analyze_lesion.py", line 1179, in main
    sct_analyze_lesion(
  File "/Users/valosek/code/sct_latest/spinalcordtoolbox/reports/qc2.py", line 753, in sct_analyze_lesion
    dorsal_bridge_width_mm = measure_pd[col_name_dorsal][idx_row]
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.10/site-packages/pandas/core/frame.py", line 3807, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/Users/valosek/code/sct_latest/python/envs/venv_sct/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 3804, in get_loc
    raise KeyError(key) from err
KeyError: 'slice_7_dorsal_bridge_width [mm]'
Total runtime; 3.865 seconds.
```

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -670,6 +670,8 @@ def sct_analyze_lesion(
         # Get the total number of lesions; this will represent the number of rows in the figure. For example, if we have
         # 2 lesions, we will have two rows. One row per lesion.
         num_of_lesions = len(label_lst)
+        # Restrict the lesion mask to the spinal cord mask, as lesions should not occur outside the cord
+        im_lesion_data = im_lesion_data * im_sc_data
         # Get the sagittal lesion slices
         sagittal_lesion_slices = np.unique(np.where(im_lesion_data)[0])
         # Get the minimum sagittal slice with lesion. For example, if a lesion cover slices 7,8,9, get 7
--- a/spinalcordtoolbox/scripts/sct_analyze_lesion.py
+++ b/spinalcordtoolbox/scripts/sct_analyze_lesion.py
@@ -406,8 +406,7 @@ class AnalyzeLesion:
         im_sc = Image(self.fname_sc)
         im_sc_data = im_sc.data
 
-        # Restrict the lesion mask to the spinal cord mask (from anatomical level, it does not make sense to have lesion
-        # outside the spinal cord mask)
+        # Restrict the lesion mask to the spinal cord mask, as lesions should not occur outside the cord
         im_lesion_data = im_lesion_data * im_sc_data
 
         # Get the dimensions of the lesion mask


spinalcordtoolbox/reports/qc2.py
```diff
@@ -670,6 +670,8 @@ def sct_analyze_lesion(
         # Get the total number of lesions; this will represent the number of rows in the figure. For example, if we have
         # 2 lesions, we will have two rows. One row per lesion.
         num_of_lesions = len(label_lst)
+        # Restrict the lesion mask to the spinal cord mask, as lesions should not occur outside the cord
+        im_lesion_data = im_lesion_data * im_sc_data
```
joshuacwnewton: By adding this fix, I think it means we no longer need the other fix?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ae939b1c3be5f9def829562c82a45cb9dda8dabb/spinalcordtoolbox/reports/qc2.py#L701-L704

It seems that this code was meant for a similar purpose.

joshuacwnewton: I think that maybe this fix should be moved up a few lines (before we generate `label_lst`), just in case that there is a lesion that is outside the SC that gets included in `label_lst` but gets excluded from the masking operation.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton, line 1: Since our `sct_analyze_lesion` test uses dummy data, and we have control over where we define the lesion mask, can we add a test that uses a lesion that exists outside the cord? Just to see if the test would have caught the bug + if the changes fix the bug.

joshuacwnewton (changes requested):
Just to keep things proper, I opened an issue describing the problem so that this PR could be linked with an issue.

(Also, I think the failing test was just a one-off network error. I restarted the job and I expect it to pass.)

spinalcordtoolbox/reports/qc2.py
valosekj: Good catch! I completely missed this line! 
You're right, this was meant to prevent this exact edge case. But it comes too late as we get `min_sag_slice` and `max_sag_slice` before

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6e9236a2ce03a2d20a0f226294bed6b53b68b635/spinalcordtoolbox/reports/qc2.py#L677-L680

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -147,7 +148,9 @@ def compute_expected_measurements(lesion_params, path_seg=None):
     ([(31, 0, 25), (4, 15, 3)], 0.1),
     # Multiple lesions
     ([[(31, 0, 25), (4, 15, 3)],
-      [(29, 45, 25), (3, 10, 2)]], 0.1)
+      [(29, 45, 25), (3, 10, 2)]], 0.1),
+    # Lesion partly outside the spinal cord segmentation (z (RL) slice 19 is outside the SC seg)
+    ([(29, 45, 19), (3, 10, 2)], 0.001)
```
joshuacwnewton: :eyes:

```
   File "/home/runner/work/spinalcordtoolbox/spinalcordtoolbox/testing/cli/test_cli_sct_analyze_lesion.py", line 197, in test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements
    np.testing.assert_allclose(measurements.at[idx, key],

FAILED testing/cli/test_cli_sct_analyze_lesion.py::test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements[dummy_lesion4-0.001]
 - KeyError: 'slice_19_dorsal_bridge_width [mm]'
```

Looks like the KeyError is just in the test function, and thankfully not in the code. :sweat_smile:

spinalcordtoolbox/reports/qc2.py
valosekj: removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/02e07e4cc1c109841eda039a743dd46695497db8

spinalcordtoolbox/reports/qc2.py
valosekj: Yes! Test added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/235b97c8266cea27099358024582fc0330e61a60

spinalcordtoolbox/reports/qc2.py
valosekj: good point. moved up in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/95d53879c2a598402e39a2016ec43a4e5a001271

testing/cli/test_cli_sct_analyze_lesion.py
valosekj: Yeah, I think it comes from the `compute_expected_measurements` function:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/cdd912d3e5380a509a8edf2e8801db3462b3bdd5/testing/cli/test_cli_sct_analyze_lesion.py#L49

The thing is that in the function, we work with the lesion coordinates, not the lesion as an array:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/cdd912d3e5380a509a8edf2e8801db3462b3bdd5/testing/cli/test_cli_sct_analyze_lesion.py#L61

So, the masking will need to be done in a slighly different way

valosekj: Okay! When debugging this, I discovered that the lesion restriction was applied only when measuring tissue bridges. But we actually want to use the masked lesion for all metrics! So I moved the command up in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/d81c3e75d84ccff706d366e60ee5489764f835df!

I also tweaked the dummy lesion's postition and size to make it easier to crop it for the purposes of testing in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/45e9d4710cce6fe6c01e329657b6b07cad684f1f:

image

I then added the manual cropping in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/c121065bf07faa56a7b0dc3055bec8eb7dfce27e

valosekj: I'm thinking that maybe we can rename the PR as it is now not only about fixing the QC but also about a change in the actual `sct_analyze_lesion` code (which can impact users' metrics).

valosekj: Just a possible explanation why I didn't notice this earlier: in our analyses, I usually use an extra command to do the intersection of lesion and cord segmentation (`sct_maths -i $file_lesion -mul $file_sc -o $file_lesion`) before calling `sct_analyze_lesion`. But I believe it's better and more systematic (and less prone to errors) to restrict the lesion to the cord inside `sct_analyze_lesion`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -773,6 +769,11 @@ def _measure_totLesion_distribution(self, im_lesion, atlas_data, im_vert, p_lst)
     def measure(self):
         im_lesion = Image(self.fname_label)
         im_lesion_data = im_lesion.data
+
+        # Restrict the lesion mask to the spinal cord mask, as lesions should not occur outside the cord
+        if self.fname_sc is not None:
+            im_lesion_data = im_lesion_data * Image(self.fname_sc).data
+
```
joshuacwnewton, line 776: Hum. If we're using `fname_sc` more broadly, does that mean we need to change the kind of files we expect for `-s`?

Right now, the help string says that a centerline file is acceptable. But, wouldn't a centerline file zero out most of the lesion data now?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/192c518c829a7d8ea7764a63c4b46748ff2ed836/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L58-L65

In fact I wonder if using a centerline has been invalid for a while now... we even call the file "`fname_sc`" implying an assumption that it will be a cord segmentation.

joshuacwnewton (approved):
This mostly LGTM! I just had one concern with regards to the `-s` argument help...

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: This is an excellent point. Good catch! Tbh, I always provide segmentation.

We're using the same `if self.fname_sc is not None:` statement also here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/95e49ab26f7b9de0405341e6cc12110bd3405f10/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L195-L196

To call the `self.angle_correction` function, which fits the centerline using `get_centerline`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/95e49ab26f7b9de0405341e6cc12110bd3405f10/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L877-L879

As I think most users will have the segmentation available, I would suggest updating the help message to remove the mention of the centerline and refer only to the segmentation. 
Side note: do we have a robust way to automatically determine whether the input file is a centerline or a segmentation? (Perhaps we could simply check the number of nonzero voxels per axial slice, as a centerline typically has a single point per slice?)
What do you think?

joshuacwnewton: > Side note: do we have a robust way to automatically determine whether the input file is a centerline or a segmentation? (Perhaps we could simply check the number of nonzero voxels per axial slice, as a centerline typically has a single point per slice?)

This feels like a safe assumption (at least for binary centerlines -- soft centerlines wouldn't, since they sum to 1 in each slice implying multiple non-zero voxels per slice)? 

But I don't think we've put this into code yet. It would be interesting to at least write a test for this condition (e.g. run `sct_get_centerline` for all of the different types of centerline algorithms and see if they meet the assumption of max 1 voxel per axial slice). This would probably help us validate the function since we're not doing that too much yet.

Maybe we could just change the docstring for now but add a follow-up issue for making such a test.

valosekj: Sounds good! I updated the `-s` arg help in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4961/commits/c90eb6410bdc928c93174bbe6477426f96c46821 and opened issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4967

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -51,17 +51,17 @@ def get_parser():
...
     optional = parser.optional_arggroup
     optional.add_argument(
         "-s",
         help=textwrap.dedent("""
-            Spinal cord centerline or segmentation file, which will be used to correct morphometric measures with cord angle with respect to slice. (e.g. `t2_seg.nii.gz`)
-
-            If provided, then the lesion volume, length, diameter, axial damage ratio, and tissue bridges will be computed. Otherwise, if not provided, then only the lesion volume will be computed.
+            Spinal cord segmentation (e.g. `t2_seg.nii.gz`). If provided, it is used to restrict the lesion mask (provided with `-m` option) to the spinal cord segmentation.
+            Additionally, the spinal cord segmentation is used to obtain the spinal cord centerline, which is used to correct morphometric measures with cord angle with respect to slice.
+            If not provided, only the lesion volume is be computed.
```
joshuacwnewton: "Restrict the lesion mask" might be a bit unclear? Maybe we could be explicit about the masking operation that gets performed:

```suggestion
            Spinal cord segmentation (e.g. `t2_seg.nii.gz`). If provided, any lesion voxels outside the cord segmentation will be set to 0. 
            Additionally, the spinal cord segmentation is used to obtain the spinal cord centerline, which is used to correct morphometric measures with cord angle with respect to slice.
            If not provided, only the lesion volume is be computed.
```

Also, there is a slight typo (`is be` -> `is`):

```suggestion
            Spinal cord segmentation (e.g. `t2_seg.nii.gz`). If provided, any lesion voxels outside the cord segmentation will be set to 0. 
            Additionally, the spinal cord segmentation is used to obtain the spinal cord centerline, which is used to correct morphometric measures with cord angle with respect to slice.
            If not provided, only the lesion volume is computed.
```

valosekj: Agree! Your wording is much clearer! Thanks!

PR #4966, opened 2025-07-10
Implement anisotropic dilation and erosion

### LINKED ISSUE #4936: Allow mask dilation by distance with `sct_maths -dilate`
Currently, spinal cord mask dilation can be done using `sct_maths -dilate`. 
This dilation is done by a certain amounts of voxels. 

I am opening this issue to see if it would be possible to add dilation around the spinal cord mask by q certain distance (in mm) instead of a number of voxels. This is particularly useful in cases of anisotropic images. For example of a 2D sagittal MRI of the spinal cord, a dilation done on the axial plane using `sct_maths -i ..-dilate 1 -shape disk -o .. -dim 2` would end up dilating by 2 voxels on the R-L direction and only 2 voxels in the A-P direction. The result could like the following: 

Image

This issue is related to: https://github.com/ivadomed/ms-multi-spine-challenge-2024/issues/35

Currently, in this case, the workaround method is to perform this dilation twice using a disk shape. The first time with a disk set on the I-S direction and a second time with a disk set on the R-L direction.

mguaypaq:
This allows using the `sct_maths -dilate` and `-erode` operations with a single number of voxels (as before), or a list of 2 or 3 numbers separated by `x`, like `1x2x3`, which can be useful for anisotropic images.

While `skimage.morphology.footprint_rectangle` works for 2D and 3D box shapes, the `disk` and `ball` functions only take a single radius. The `ellipse` function could have worked for 2D, but not for 3D, and it confusingly turns (width, height) arguments into a (height, width) resulting array. I decided that it was simpler to just generate the correct numpy array directly.

Fixes #4936.

### DIFF
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -63,7 +63,7 @@ def _get_footprint(shape, size, dim):
         fp_x = ((r_y * np.arange(-r_x, r_x+1))**2).reshape((2*r_x+1, 1))
         fp_y = ((r_x * np.arange(-r_y, r_y+1))**2).reshape((1, 2*r_y+1))
         footprint = ((fp_x + fp_y) <= (r_x * r_y)**2).astype(np.uint8)
-    elif shape == 'cube':
+    elif shape == 'ball':
         # ellipsoid equation, after clearing denominators:
         # ((x - center_x)*radius_y*radius_z)**2
         # + ((y - center_y)*radius_x*radius_z)**2
--- a/spinalcordtoolbox/math.py
+++ b/spinalcordtoolbox/math.py
@@ -8,7 +8,7 @@ License: see the file LICENSE
 import logging
 
 import numpy as np
-from skimage.morphology import erosion, dilation, disk, ball, footprint_rectangle
+from skimage.morphology import erosion, dilation
 from skimage.filters import threshold_local, threshold_otsu, rank
 from scipy.ndimage import gaussian_filter, gaussian_laplace, label, generate_binary_structure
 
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/math.py
```diff
@@ -32,24 +32,51 @@ def _get_footprint(shape, size, dim):
...
+        raise ValueError(f"Invalid size for shape {shape}: {size}")
+
+    if shape == 'disk':
+        # ellipse equation is: ((x - center_x)/radius_x)**2 + ((y - center_y)/radius_y)**2 <= 1
+        # clearing denominators: ((x - center_x)*radius_y)**2 + ((y - center_y)*radius_x)**2 <= (radius_x*radius_y)**2
+        r_x, r_y = size
+        # the thin arrays fp_x and fp_y will be efficiently broadcast to a full rectangle by numpy
+        fp_x = ((r_y * np.arange(-r_x, r_x+1))**2).reshape((2*r_x+1, 1))
+        fp_y = ((r_x * np.arange(-r_y, r_y+1))**2).reshape((1, 2*r_y+1))
+        footprint = ((fp_x + fp_y) <= (r_x * r_y)**2).astype(np.uint8)
```
joshuacwnewton, line 65: I trust your implementation here, but because it's tricky for me to parse the math, I figured it might be worth writing a quick test on your branch to ensure that the output of our internal `_get_footprint` function matches the output of `skimage.morphology`.

Even though `ellipse` can't cover the 3D case (as you mentioned in the PR description), I wanted to use it to test the 2D case? And, in doing do, I found that `ellipse` actually produces a different shape by default compared to your function.

`_get_footprint` (e.g. shape 5x3):

image

`morphology.ellipse` (e.g. shape 5x3):

image

I went down a [big rabbit hole](https://github.com/scikit-image/scikit-image/issues/5470#issuecomment-3079622794) exploring why `ellipse` produced different results, and it turns out that there is already an open issue pointing out the differing shapes produced by different footprint functions, _and_ an open PR to perform the same refactoring to the `disk`/`ball`/`ellipse` functions as was done to the `square`/`cube` functions:

- https://github.com/scikit-image/scikit-image/issues/5470
- https://github.com/scikit-image/scikit-image/pull/7628

Maybe we could consider adopting their code (or at least exchanging implementation thoughts) to keep things in sync with upstream while we wait for the PR to be merged?

spinalcordtoolbox/scripts/sct_maths.py
```diff
@@ -243,20 +255,22 @@ def get_parser():
...
-        type=int,
+        type=one_two_three_ints,
         action=AppendTodo,
         help="Dilate binary or greyscale image with specified size. If shape={'square', 'cube'}: size corresponds to the length of "
              "an edge (size=1 has no effect). If shape={'disk', 'ball'}: size corresponds to the radius, not including "
-             "the center element (size=0 has no effect).",
+             "the center element (size=0 has no effect). "
+             "If shape={'square', 'disk'}: size can be a single number of voxels, "
+             "or 2 numbers separated by `x` to get a rectangle/ellipse footprint (useful for anisotropic images). "
+             "If shape={'cube', 'ball'}: size can be a single number, or 3 numbers separated by `x` to get a box/ellipsoid.",
```
joshuacwnewton: I'm finding this help to be a little overstuffed with info? Especially since we have to context switch between square/cube -> disk/ball -> square/disk -> cube/ball.

Maybe something like this would be a little clearer?

```python
        help="Dilate binary or greyscale image.\n"
             "You can customize the structural element by combining the arguments `-dilate`, `-shape`, and `-dim`. "
             "(The values passed to `-dilate` will control the side length or radius of whatever shape is chosen.)\n"
             "You can provide either a single number, or 2/3 numbers separated by `x` (depending on the shape).\n"
             "\n"
             "Examples:\n"
              "  - `-shape cube -dilate 3`            -> Side length 3   -> A 3x3x3 cube.\n"
              "  - `-shape ball -dilate 2x2x5`        -> Radius 2x2x5    -> A 5x5x11 ball.\n"
              "  - `-shape disk -dilate 3 -dim 2`     -> Radius 3        -> An 7x7 disk in the X-Y plane, applied to each Z slice.\n"
              "  - `-shape square -dilate 1x4 -dim 0` -> Side length 1x4 -> A 1x4 rectangle in the Y-Z plane, applied to each X slice.",
```

My thinking is that, rather than exhaustively explaining each possible 1, 2, or 3 valued combination, we could instead explain the general concept and then provide some examples (increasing in complexity). That way, the usage becomes a bit self-evident.

joshuacwnewton (commented):
Barring the docs fix (which should be pretty straightforward to implement), and maybe barring our involvement in the upstream PR (where we could consider implementing the whole `strict_radius={True,False}` options), this mostly LGTM!

I'll wait until you return from your break to give the final approval.

spinalcordtoolbox/math.py
mguaypaq: Thanks for digging and connecting with upstream!

Hmm, it looks like `skimage`'s `ellipse` function doesn't match their own `disk`/`ball` functions, in the case of equal radii. I agree that `ellipse` looks a little more rounded out, but `disk`/`ball` is probably what we should stay backwards compatible with.

Looking at their code, it looks mathematically equivalent to what I proposed (except that theirs is potentially less accurate because of floating point issues). I think it'll be easy to switch back to `skimage.morphology` if/when that PR gets merged, if we still want to at that point.

spinalcordtoolbox/scripts/sct_maths.py
mguaypaq: This is great, I'll update the documentation.

mguaypaq:
Let me know if you'd like other changes, otherwise I think this is good to go.

PR #5001, opened 2025-08-05
Update and reorganize Docker installation instructions

### LINKED ISSUE #4922: Moving SCT Docker installation
Currently the instructions to install SCT in Docker are located [here for linux](https://spinalcordtoolbox.com/stable/user_section/installation/linux.html#option-4-install-with-docker) and [here for mac](https://spinalcordtoolbox.com/stable/user_section/installation/mac.html#option-4-install-with-docker). 

Making these instructions more obvious could help users to install SCT inside Docker (example https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4918). 

One solution could be to add an additional point on this page: https://spinalcordtoolbox.com/stable/user_section/installation.html

Currently, this is what it looks like: 

Image

This functionality also depends on how often SCT is installed inside a Docker. Do we want to add more stuff to this page? We cannot add points for every possible option.

### LINKED ISSUE #4923: Adding Dockerfile instructions to show how to install SCT in Docker
I struggled to install SCT in Docker based on the current instructions.

The problem came from the fact that when installing SCT in Docker we need to point out where SCT is installed. 

Here is a proposed Dockerfile to add as an alternative or replace the current instructions: 

```Dockerfile
# Use a lightweight Python image
FROM python:3.10-slim

# Creates a directory for the application
WORKDIR /app

# Install Python dependencies
RUN apt-get update && apt-get install -y \
    build-essential \
    bash \
    curl \
    bzip2 \
    libglib2.0-0 \
    libgl1-mesa-glx \
    libxrender1 \
    libxkbcommon-x11-0 \
    libdbus-1-3 gcc \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/*

# Install SCT
RUN git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git /root/sct && \
    /root/sct/install_sct -y
# Add SCT to PATH
ENV PATH="/root/sct/bin:$PATH"
# Automatically activate SCT env in bash
RUN echo "source /root/sct/python/etc/profile.d/conda.sh && conda activate venv_sct" >> /root/.bashrc

# Point d’entrée par défaut (peut être redéfini au runtime)
ENTRYPOINT ["python", "scripts/main.py"]
``` 

This is just an example solution. 

Thanks :)

mguaypaq:
The point of this pull request is to:

- Make Docker installation instruction more visible, by making a top-level page for them (alongside MacOS, Linux, Windows, and Apptainer).
- Provide a Dockerfile for those who prefer to use that, rather than doing an interactive installation and saving the resulting Docker container.

It turns out that we already _had_ an example Dockerfile, but it wasn't linked from anywhere in the documentation. Also, the accompanying README listed a mismatched version of Ubuntu.

Along the way, I fixed a few other typos and issues I found in the files I was already touching.

I welcome further improvements to the wording, and I have **not** tested these instructions.

Fixes #4922, fixes #4923.


contrib/docker/README.md
```diff
@@ -7,6 +7,6 @@ This directory contains a minimal Dockerfile to install spinalcoordtoolbox.
 Inside this directory, run:
 
     docker build -t dockersct .
-    docker run --name sctcontainer -i -t dockersct
-    
-For more a more detailed set of instructions, refer to the documentation at https://spinalcordtoolbox.com/user_section/installation.html
+    docker run --name sctcontainer -it dockersct
+
+For more a more detailed set of instructions, refer to the documentation at <https://spinalcordtoolbox.com/latest/user_section/installation/docker.html>
```
joshuacwnewton: Do we want this to be `latest`? (Rather than `stable`?)

documentation/source/user_section/installation/docker.rst
joshuacwnewton, line 1: I am very much in favor of all of the changes prior to commit 33781ff276103b30739dc19dd4ead256582ba2d2, but the last commit lost me a little bit?

As far as I understand, this commit de-duplicates the bit about installing SCT, but leaves the remaining bits (OS-specific dependencies) in their respective OS pages. This makes sense from a de-duplication perspective, but I don't know if it's very user-friendly?


1. User arrives at the Docker installation page. 
2. User sees the steps for installing SCT, but is meant to ignore these steps (for now).
3. User navigates to the page for their specific OS.
4. User follows the steps to install the necessary dependencies (Docker, etc.)
5. User returns back to the previous page to install SCT.
6. User goes forward to their OS page once more to optionally install GUI script support.

I think this may be one of those cases where duplicating the info makes for a better user experience, since it would cut out steps 2, 5, and 6? So, maybe we could consider using the [`.. include::`](https://docutils.sourceforge.io/docs/ref/rst/directives.html#include) directive. That way, the steps for installing SCT can still live in a single place (and only need to be updated in one location), but when rendering the docs, they end up in all 3 pages.

contrib/docker/Dockerfile
```diff
@@ -1,7 +1,35 @@
+# Use a the default Ubuntu 22.04 image
 FROM ubuntu:22.04
 
-RUN apt-get update && apt-get install -y git curl bzip2 libglib2.0-0 libgl1-mesa-glx libxrender1 libxkbcommon-x11-0 libdbus-1-3 gcc
-RUN git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git sct
-RUN cd sct
-RUN ./install_sct -y
-RUN rource /root/.bashrc
```
joshuacwnewton, line 7: image < Ruh-roh!

contrib/docker/Dockerfile
```diff
@@ -1,7 +1,35 @@
...
+    && /root/sct/install_sct -y
+
+# Add SCT to PATH
+ENV PATH="/root/sct/bin:$PATH"
+
+# Automatically activate SCT env in bash
+RUN echo "source /root/sct/python/etc/profile.d/conda.sh && conda activate venv_sct" >> /root/.bashrc
+
+# Add a default entry point (can be redefined at runtime)
+ENTRYPOINT ["python", "scripts/main.py"]
```
joshuacwnewton: I don't fully understand the purpose of this line. Is it necessary to specify some sort of dummy value when creating the image?

documentation/source/user_section/installation/docker.rst
joshuacwnewton: Alternatively... couldn't we just include the installation commands as part of the Dockerfile itself? That way, SCT is installed as part of the process of building the image?

contrib/docker/Dockerfile
mguaypaq: I have no idea, but I suspect that @plbenveniste wanted to include this as an example of how to define an entrypoint when defining a Docker container to run a processing pipeline (rather than just for starting an interactive terminal with SCT installed) (if that's even what this line is doing).

contrib/docker/Dockerfile
mguaypaq: Haha yeah

documentation/source/user_section/installation/docker.rst
mguaypaq: Sure, I can re-duplicate the instructions. And link to the Dockerfile at the appropriate point as an alternative for people who already know how to use it.

contrib/docker/README.md
mguaypaq: Sure!

mguaypaq: Sorry, I just copy-pasted the link without looking carefully at the version. Fixed!

documentation/source/user_section/installation/docker.rst
mguaypaq: Done, hopefully this is better.

contrib/docker/Dockerfile
```diff
@@ -1,7 +1,35 @@
...
+        curl \
+        gcc \
+        git \
+        libdbus-1-3 \
+        libgl1-mesa-glx \
+        libglib2.0-0 \
+        libxkbcommon-x11-0 \
+        libxrender1 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
joshuacwnewton: Should we try to unify the dependencies between the Dockerfile and the manual instructions we provide?

```
git curl bzip2 libglib2.0-0 libgl1-mesa-glx libxrender1 libxkbcommon-x11-0 libdbus-1-3 gcc
```

It seems like there are a few extra dependencies here (like `bash` and `build-essential`)

joshuacwnewton: Likewise, do we want to unify the "ensure image size is minimal" steps? (the `clean` and `rm` steps?)

contrib/docker/Dockerfile
joshuacwnewton: That would make sense! If it's an example, then maybe we could comment it out, at least? :thinking:

contrib/docker/Dockerfile
joshuacwnewton, line 1: I'm realizing now that across all of our instructions, we clone the main branch (as well as doing a full clone).

Is there any merit to picking a stable version tag for demonstration purposes (e.g. 7.0) then adding a comment saying "replace with whichever version you want to install"?

Also, do we want to do any sort of shallow clone to minimize the image size? (I'm assuming our dependencies are much larger than the git repo, so maybe this isn't worth the pain of potentially not having the git history available.)

contrib/docker/Dockerfile
```diff
@@ -1,7 +1,35 @@
+# Use a the default Ubuntu 22.04 image
 FROM ubuntu:22.04
 
-RUN apt-get update && apt-get install -y git curl bzip2 libglib2.0-0 libgl1-mesa-glx libxrender1 libxkbcommon-x11-0 libdbus-1-3 gcc
-RUN git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git sct
-RUN cd sct
-RUN ./install_sct -y
-RUN rource /root/.bashrc
+# Create a directory for the application
+WORKDIR /app
```
joshuacwnewton: I see we make this `/app` folder the working directory, but then all of the steps afterwards use the `/root/sct/` folder. Is this `WORKDIR` command not really doing anything, then?

documentation/source/user_section/installation/linux.rst
```diff
@@ -129,8 +127,8 @@ Then, follow the example below to create an OS-specific SCT installation (in thi
     # Launch interactive mode (command-line inside container)
     sudo docker run -it ubuntu:22.04
     # Now inside Docker container, install SCT dependencies
-    apt-get update
-    apt install -y git curl bzip2 libglib2.0-0 libgl1-mesa-glx libxrender1 libxkbcommon-x11-0 libdbus-1-3 gcc
+    apt update
+    apt install git curl bzip2 libglib2.0-0 libgl1-mesa-glx libxrender1 libxkbcommon-x11-0 libdbus-1-3 gcc
```
joshuacwnewton: I guess because we are running an interactive shell (`-it`) we don't need the `-y` anymore, right? :thinking: 

Still, is it worth it to keep the `-y` to make the process easier on the user?

contrib/docker/Dockerfile
mguaypaq: I know very little about Docker, but I suspect that (along with the entrypoint) the purpose of this is for the hypothetical _processing pipeline_ (rather than SCT) to live under `/app`. (Or possibly, it comes from copy-pasting random lines from examples found online and the output of some generative AI until something works.) Since @plbenveniste specifically had trouble with paths within Docker, I would like to hear his opinion about whether WORKDIR and ENTRYPOINT are useful parts of the Dockerfile.

contrib/docker/Dockerfile
mguaypaq: For `bash` and `build-essential`, maybe the difference is because Pierre-Louis's example Dockerfile was based on `FROM python:3.10-slim` rather than `FROM ubuntu:22.04`, or because these dependencies were useful for his specific processing pipeline.

For the "minimal image size" steps, I think they might be appropriate in a Dockerfile (which is more geared towards reproducibility), but I think they would be irrelevant noise in the manual instructions. (And I don't think people who generate a Docker image by manually running commands in an interactive terminal care about reproducibility and image size?)

So, this raises a question: _do_ we want to harmonize the Dockerfile and the manual steps? I think they have mostly disjoint audiences, but I don't really mind either way.

contrib/docker/Dockerfile
mguaypaq: Sure, doing an example checkout of a specific version is probably a good idea. I don't think the manual steps are the place to complicate things with a shallow clone, though.

documentation/source/user_section/installation/linux.rst
mguaypaq: I kept the `-y` in `sct_install` below, because I think that's genuinely easier for the user. But in this case, I think `apt install` will just require an extra press of the enter key, so I figure let's embrace the interactive, duct-taped nature of this installation method.

mguaypaq: Note that this is for the same reason I put `apt` in the interactive instructions (which is meant for user-friendly interaction, and explicitly not meant for scripting), versus `apt-get` in the Dockerfile (which _is_ meant for scripting use).

mguaypaq:
@joshuacwnewton since Pierre-Louis is currently on vacation, I decided to mostly just harmonize the Dockerfile with the manual instructions, so that we can get this merged for 7.1. We can always improve the Dockerfile later. Let me know it this addresses your review comments.

contrib/docker/Dockerfile
```diff
@@ -1,7 +1,30 @@
...
+        libxrender1 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install version 7.0 (for example) of SCT
+RUN git clone \
+      --branch 7.0 \
+      https://github.com/spinalcordtoolbox/spinalcordtoolbox.git \
+      /root/sct \
+    && /root/sct/install_sct -y
```
joshuacwnewton: It looks like the switch from the main branch to `7.0` caused our installer to think that we wanted a "package" installation, and thus SCT gets installed to `/root/sct_7.0` instead of `/root/sct`. 

I suppose we need to explicitly pass `-i` to get an in-place installation for cloned stable releases? Happy to change this. :)

joshuacwnewton: Tested myself, and everything else in the Dockerfile/instructions LGTM!

Issue #4974, opened 2025-07-16
totalspineseg: -qc-plane Sagittal outputs axial orientation


jcohenadad:
> Funnily, when asking the QC in sagittal, I get it in axial 🤣 
> 
> ~~~
> sct_deepseg totalspineseg -i t2.nii.gz -step1-only 1 -qc qc -qc-plane Sagittal -qc-seg t2_seg.nii.gz 
> ~~~
> 
> <img width="1066" height="406" alt="Image" src="https://github.com/user-attachments/assets/85de4e8c-860b-4cea-bf41-02776ca54101" /> 

 _Originally posted by @jcohenadad in [#4973](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4973#issuecomment-3075086945)_

Issue #4888, opened 2025-05-02
sct_deepseg MS lesion MP2RAGE: incorrect reference


plbenveniste:
There is a problem in the documentation of this sct_deepseg function: 
https://spinalcordtoolbox.com/stable/user_section/command-line/deepseg/lesion_ms_mp2rage.html

The reference displayed is not the correct one.

Issue #4861, opened 2025-04-16
Task-specific subparsers contain irrelevant arguments/information that only apply to specific tasks


joshuacwnewton:
In much the same way we add arguments that are only relevant for specific models (e.g. https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4860/commits/dbbbd87d8fe9a82946ff41c237e0098551e07906), we could consider _removing_ arguments that have no meaningful effect on the output for certain models.

For example, for multiclass models (especially that of something like `totalspineseg`), options such as `-thr` and `-keep-largest` don't exactly make sense, do they?

That way, the interfaces for each model remain simple, and options are only provided if they are useful in some way or another.

valosekj:
Good idea!

> For example, for multiclass models (especially that of something like `totalspineseg`), options such as `-thr` and `-keep-largest` don't exactly make sense, do they?

Same for the rootlets model (`rootlets_t2).

SomeoneInParticular:
Agree; might it be worth making `SCTArgumentParser` have a subclass specifically for DeepSeg then?

PR #5008, opened 2025-08-12
Clean up argparse help to only display arguments and descriptions relevant to specific tasks

joshuacwnewton:
## Checklist

- [x] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)

## Description

This PR aims to improve the `sct_deepseg -h` (argparse) by limiting what information gets displayed for individual task help (e.g. `sct_deepseg totalspineseg -h`).

The goal is to try and remove arguments that are irrelevant for certain tasks, and also limit warnings and notes to only the tasks that they are relevant for.

Here is a sample before/after:

- Before: https://spinalcordtoolbox.com/stable/user_section/command-line/deepseg/totalspineseg.html
- After: https://spinalcordtoolbox--5008.org.readthedocs.build/5008/user_section/command-line/deepseg/totalspineseg.html

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4861.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4974.
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4888.

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -37,7 +37,7 @@ MODELS = {
         ],
         "description": "Cord segmentation model on mouse MRI. Data from University of Queensland.",
         "contrasts": ["t1"],
-        "thr": 0.0,  # postprocessing.binarize_prediction is not present in model json
+        "thr": 0.0,  # Only for display in argparse help (postprocessing.binarize_prediction is not present in model json)
         "default": False,
     },
     "mice_uqueensland_gm": {
@@ -47,7 +47,7 @@ MODELS = {
         ],
         "description": "Gray matter segmentation model on mouse MRI. Data from University of Queensland.",
         "contrasts": ["t1"],
-        "thr": 0.0,  # postprocessing.binarize_prediction is not present in model json
+        "thr": 0.0,  # Only for display in argparse help (postprocessing.binarize_prediction is not present in model json)
         "default": False,
     },
     "t2_tumor": {
@@ -56,7 +56,7 @@ MODELS = {
         ],
         "description": "Cord tumor segmentation model, trained on T2-weighted contrast.",
         "contrasts": ["t2"],
-        "thr": 0.5,
+        "thr": 0.5,  # Only for display in argparse help (mirrors postprocessing.binarize_prediction, which is 0.5 in model json)
         "default": False,
     },
     "findcord_tumor": {
@@ -65,7 +65,7 @@ MODELS = {
         ],
         "description": "Cord localisation model, trained on T2-weighted images with tumor.",
         "contrasts": ["t2"],
-        "thr": 0.5,
+        "thr": 0.5,  # Only for display in argparse help (mirrors postprocessing.binarize_prediction, which is 0.5 in model json)
         "default": False,
     },
     "model_seg_sctumor-edema-cavity_t2-t1_unet3d-multichannel": {
@@ -74,7 +74,7 @@ MODELS = {
         ],
         "description": "Multiclass cord tumor segmentation model.",
         "contrasts": ["t2", "t1"],
-        "thr": 0.5,
+        "thr": 0.5,  # Only for display in argparse help (mirrors postprocessing.binarize_prediction, which is 0.5 in model json)
         "default": False,
     },
     "model_seg_exvivo_gm-wm_t2_unet2d-multichannel-softseg": {
@@ -83,7 +83,7 @@ MODELS = {
         ],
         "description": "Grey/white matter seg on exvivo human T2w.",
         "contrasts": ["t2"],
-        "thr": 0.0,  # postprocessing.binarize_prediction is not present in model json
+        "thr": 0.0,  # Only for display in argparse help (postprocessing.binarize_prediction is not present in model json)
         "default": False,
     },
     "model_7t_multiclass_gm_sc_unet2d": {
@@ -93,7 +93,7 @@ MODELS = {
         "description": "SC/GM multiclass segmentation on T2*-w contrast at 7T. The model was created by N.J. Laines Medina, "
                        "V. Callot and A. Le Troter at CRMBM-CEMEREM Aix-Marseille University, France",
         "contrasts": ["t2star"],
-        "thr": 0.5,
+        "thr": 0.5,  # Only for display in argparse help (mirrors postprocessing.binarize_prediction, which is 0.5 in model json)
         "default": False,
     },
     "model_seg_epfl_t2w_lumbar_sc": {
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/deepseg/models.py
```diff
@@ -37,6 +37,7 @@
         ],
         "description": "Cord segmentation model on mouse MRI. Data from University of Queensland.",
         "contrasts": ["t1"],
+        "thr": 0.0,  # postprocessing.binarize_prediction is not present in model json
```
joshuacwnewton: I've added `thr` values to the ivadomed entries in the model dict purely as something to display in the `-h` so that users have a reference value to compare to when overriding it with their own `-thr` values.

Choosing a display value turned out to be a bit awkward for ivadomed models that don't even use have the thresholding JSON entry. But because this means that those models don't apply thresholding (i.e. soft segmentation), and because the help for `-thr` already mentions `0 == soft segmentation`, then choosing `0.0` felt like an appropriate value here as well.

joshuacwnewton: > [!NOTE]
> The value from the dict won't be processed because ivadomed models have their own means of fetching the threshold values from the model configuration JSON file.
>
> The dictionary threshold value is only fetched for non-ivadomed models.
>
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0060f1cfba76952623f559cfc676cb520a3e5580/spinalcordtoolbox/scripts/sct_deepseg.py#L409-L418

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -256,18 +264,26 @@ def get_parser(subparser_to_return=None):
             metavar=Metavar.str,
             choices=('Axial', 'Sagittal'),
             default='Axial',
-            help="Plane of the output QC. If Sagittal, it is highly recommended to provide the `-qc-seg` option, "
-                 "as it will ensure the output QC is cropped to a reasonable field of view. "
-                 "(Note: Sagittal view is not currently supported for rootlets/totalspineseg QC.)")
+            # Note: Sagittal view is not currently supported for rootlets/totalspineseg QC
+            help=(SUPPRESS if task_name in ('rootlets', 'totalspineseg') else
```
joshuacwnewton: I chose to use `SUPPRESS` for both `-thr` and `-qc-plane` because I figured it would be the least disruptive way to hide these arguments when they're not relevant. 

Technically this means that they can still be passed, but even if a user does pass a phantom argument, this will be no different than the current behavior (thus having minimal impact).

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: I should probably add comments explaining this, as it's not clear at all from the code alone.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -256,18 +264,26 @@ def get_parser(subparser_to_return=None):
...
+                  "Plane of the output QC. If Sagittal, it is highly recommended to provide the `-qc-seg` option, "
+                  "as it will ensure the output QC is cropped to a reasonable field of view."))
+
+        # This note is meaningless for rootlets/totalspineseg QC since sagittal mosaics aren't supported.
+        note_qc_seg = ""
+        if task_name not in ('rootlets', 'totalspineseg'):
```
SomeoneInParticular: Given you use this set pair previously as well, we probably should move it to a constant variable for ease of reference.

Alternatively, these iterative "special" changes could be done after the initial loop; thats a much bigger undertaking though, so I mention it more for discussion sake rather than for this PR.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -173,17 +173,22 @@ def get_parser(subparser_to_return=None):
...
+            data must be cropped around the spinal cord.
+            ({models.CROP_MESSAGE})
+        """)
         input_output.add_argument(
             "-i",
             nargs="+",
-            help=f"Image to segment. Can be multiple images (separated with space)."
-                 f"\n\nNote: If choosing `lesion_ms_mp2rage`, then the input "
-                 f"data must be cropped around the spinal cord. ({models.CROP_MESSAGE})",
+            help=("Image to segment. Can be multiple images (separated with space)."
```
SomeoneInParticular: [Nit] I think this can be re-written to be a bit cleaner. 

```suggestion
            help=("Image(s) to segment. If using multiple, separate them with a space."
```

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -194,23 +199,26 @@ def get_parser(subparser_to_return=None):
...
             "-thr",
             type=float,
             dest='binarize_prediction',
-            help="Binarize segmentation with specified threshold. Set to 0 for no thresholding (i.e., soft segmentation). "
-                 "Default value is model-specific and was set during optimization "
-                 "(more info at https://github.com/sct-pipeline/deepseg-threshold).",
+            help=(SUPPRESS if all(t is None for t in thr_values) else
+                  f"Binarize segmentation with specified threshold. Set to 0 for no thresholding (i.e., soft segmentation). "
+                  f"Default value is '{thr_values}', and was chosen by experimentation "
+                  f"(more info at https://github.com/sct-pipeline/deepseg-threshold)."),
```
SomeoneInParticular: Clever, though a small comment explaining this would be appreciated; you do that for every other `SUPPRESS` use in this PR, this one just sticks out.

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: > Alternatively, these iterative "special" changes could be done after the initial loop; thats a much bigger undertaking though, so I mention it more for discussion sake rather than for this PR.

I think this is a super reasonable suggestion, especially given that we already have a spot where these sorts of task-specific modifications are done (albeit still within the loop):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6ac625a01f1724a4814d7f1c580510b205eaba88/spinalcordtoolbox/scripts/sct_deepseg.py#L293-L310

The tricky thing is, the above is just for adding arguments -- I'm not immediately sure how to go about retroactively editing the help text of existing arguments. At first I thought we would have to reach into the internals of an `ArgumentParser` object (e.g. `_actions`)? 

image

But looking more closely, it does seem like the `add_argument` method returns the `_Action` objects? So, theoretically we could store the relevant ones and modify them later, all while using public APIs. :thinking:

I think I can whip up a quick commit for this?

joshuacwnewton: Is this similar to what you had in mind? 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5008/commits/6d22a627bd0a5a1b77895ffd2c82ddd67508d23e

SomeoneInParticular: Yup! Looks great to me.

Issue #5013, opened 2025-08-20
Update the reference to the `lesion_ms`  model


plbenveniste:
Since the model was accepted for presentation at ESMRMB 2025 (through an abstract), the reference should be added to SCT. 

I'm on it.

PR #5014, opened 2025-08-20
Added ref for `lesion_ms` ESMRMB 2025 model

plbenveniste:
Related to this issue: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5013

This is a small change with the addition of the reference to the abstract corresponding to the ESMRMB 2025 MS lesion model.


mguaypaq (approved):
This looks fine, thanks! (I was going to comment that the previous citation uses bibtex escapes for the accents in author names, but the rest of the file is not consistent about this.)

PR #5003, opened 2025-08-06
[WIP] Prototype fork of `nnunetv2` with `torch<2.3` fixes.

joshuacwnewton:
## Checklist

- [ ] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)
- [ ] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [ ] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
[...]

PR #5027, opened 2025-09-10
Switch to `neuropoly` fork of `nnunetv2` to support multi-fold "custom trainer" version of `lesion_ms`

### LINKED ISSUE #4977: Consider maintaining a full NeuroPoly fork of `nnunetv2`
## Context

Historically, SCT has struggled with the `nnunetv2` package due to upstream breaking changes and slow maintainer responses. New releases of `nnunetv2` frequently break compatibility and disrupt our development (5-6 breaking issues over the past year or two).

Presently, there are 2 issues with current versions of `nnunetv2` that we have not been able to easily solve:

- [nnUNet#2742](https://github.com/MIC-DKFZ/nnUNet/issues/2742#issuecomment-3009472493): The newest versions of `nnunetv2` (`>=2.6.0`) are not compatible with Intel Macs due to `torch>=2.3` requirement.
- [spinalcordtoolbox#4955](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4955): There is no easy way to perform inference for models trained with a custom trainer, since `nnunetv2` hardcodes a directory within the Python package files.

The actual fixes are honestly pretty simple for both issues (maybe 1-5 line changes). However, the problem is that stable PyPI versions of `nnunetv2` have already been released, so even if these issues are fixed, we have to be very careful with how we `pip install nnunetv2`. Even our current `requirements.txt` is a bit of a minefield:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8c0c89849ec914dbe164103c678544be9f8ecacc/requirements.txt#L29-L35

## Current solutions

Right now, in issue #4955, we have been discussing radical changes that are very complex (extracting `nnunetv2` code and maintaining a partial fork that we distribute with models). 

However, for this proposed solution, students who are training models with `nnunetv2` would still `pip install nnunetv2` (i.e. the newest version), and we would just have to hope that our extracted code would stay compatible with the main project. 

To be honest, my gut feeling is that dependency version conflicts would still occur and still cause us headaches. 😓 

## Proposed solution

But, what if there's a simpler solution: Maintaining our own _complete_ fork? Even though it sounds more complex, I think it might actually be simpler to maintain:

- GitHub provides simple mechanisms to keep forks in sync with upstream. (One-click button)
- The patches we want to apply are fairly simple, so we shouldn't encounter conflicts very often.
- The workflow could largely be automated (in terms of syncing with upstream, reapplying our patches, and creating new PyPI releases in lockstep with upstream).

Not to mention all of the benefits it would provide:

- We have the freedom to fix little bugs that the upstream developers are not quick to fix themselves. (Past breaking API changes have often been very simple/small as well.)
- We can add our own testing _for our fork of the nnUNet package_ to anticipate when upstream changes will break SCT. (It can act as the "first line of defense" before a potential new breaking release, and make sure that nnUNet's dependencies stay compatible with SCT.)
- We don't have to worry about whether our "custom fork changes" are compatible with upstream, because we can just... stop using upstream `nnunetv2`. (In other words, students would both train and do inference using our `nnunetv2-neuropoly` fork, and ignore the `nnunetv2` package).

The more I think about it, the more I'm convinced that maintaining a 1:1 copy of upstream (with minor tweaks) is much less maintenance effort than extracting inference code that we ship with our models (but still have to install all of the necessary dependencies for).

I could even fairly quickly whip up a proof of concept fork with both patches applied, just to see if it's possible to do?

### LINKED ISSUE #4995: New MS lesion version requires N-fold averaging and custom trainer support
Hi!

I want to add the new MS lesion model.
However, after discussion with @jcohenadad, we have decided that it would be better to add the 5-fold model to SCT to:
- get better segmentation (compared to a 1-fold model)
- allow creation of soft segmentation from the average output of the 5 folds. 

My questions are the following:
- How much refactoring will this cause?
- This model having a custom trainer, what has been decided for now regarding this?

Thanks 😃

plbenveniste:
## Checklist

- [x] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)

## Description

This adds the new MS model to SCT. It is a multi-fold model that outputs both binary and soft segmentation. Hence, this will require a bit of change in the code:

1. `nnunetv2` fork
   - Switching from `nnunetv2` to `nnunetv2-neuropoly==2.6.2`, which adds support for custom trainers (along with a few other fixes, such as returning per-fold logits and also supporting older versions of `torch`)
   - Updating `totalspineseg` (which allows for `nnunetv2-neuropoly` to be installed without conflict)
2. `lesion_ms` multi-fold update
   - Replacing the single-fold model URL with a dictionary of 5 fold URLs.
   - Updating the existing "multi-seed" model installation code to work with the `nnunetv2` directory structure.
   - Changing the call to the `nnunetv2` inference function to return multi-fold outputs.
   - Adding a function to average the predictions of multi-fold `nnunetv2` models.
   - Adding new model-specific arguments (to allow for single-fold inference and soft lesion output)

## Linked issues

Fixes #4995.
Fixes #4977.


plbenveniste:
I am working on adding the new MS model to SCT but I am facing the problem of having a specific trainer of nnunet so inference doesn't work. @joshuacwnewton  or @mguaypaq mentionned using our fork of nnunet. How can I use it? I would like to do this as soon as possible as a collaborator would like to try my model as soon as possible.

mguaypaq:
It looks like [this branch](https://github.com/joshuacwnewton/nnUNet/tree/neuropoly-fork-patches) is the current prototype fork of nnUNet that Joshua is working on. @plbenveniste you should probably base your own fork from there, to have better compatibility with SCT's dependencies.

requirements.txt
```diff
@@ -32,7 +32,8 @@ nilearn
 # 2.4.2 causes empty predictions (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4556)
 # 2.6 includes an API change that requires `torch>=2.3`. However, `torch==2.3` removes 
 # support for Intel Macs, and we don't want to remove support, so we use `<2.6`.
-nnunetv2>2.3.0,!=2.4.*,<2.6
+# nnunetv2>2.3.0,!=2.4.*,<2.6
+git+https://github.com/joshuacwnewton/nnUNet@neuropoly-fork-patches#egg=nnunetv2-neuropoly
```
plbenveniste: I am having some trouble with the installation now. It seems that the "new" `nnunetv2-neuropoly` package is installed correctly but then it is not detected when checking the dependencies.

Here is the output log: 
[install_sct_log.txt](https://github.com/user-attachments/files/22261643/install_sct_log.txt)


Any help is welcome @mguaypaq and @joshuacwnewton 😃

plbenveniste: I was able to fix this issue with the next commit. I am using a fork of Joshua's fork where the package name is nnunetv2 not nnunetv2-neuropoly.

joshuacwnewton: Sorry for the late reply, and for being away.

The name clash during `sct_check_dependencies` is fixable by mapping `nnunetv2-neuropoly` to `nnunetv2`. I did this in my draft PR: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5003

I think we should maybe continue to use `nnunetv2-neuropoly` as a name, just to avoid confusion with the main `nnunetv2` package and make sure that we have a unique name on PyPI. (Right now we've uploaded the package to Test PyPI, but eventually we'll want to upload it to the real PyPI to avoid the need for the `--extra-index-url`.)

plbenveniste: Okay ! Thanks for adding the link to your PR. It helps :)

plbenveniste: @joshuacwnewton How can I edit the nnunetv2-neuropoly code to add my trainer?

joshuacwnewton: I have transferred the ownership to the `spinalcordtoolbox` organization: https://github.com/spinalcordtoolbox/nnUNet-neuropoly. My branch (with the necessary custom trainer fixes) is here: [`neuropoly-fork-patches`](https://github.com/spinalcordtoolbox/nnUNet-neuropoly/commits/neuropoly-fork-patches/). Since it's now on the `spinalcordtoolbox` org, you should be able to edit the branch and push new changes if you'd like. :)

> How can I edit the nnunetv2-neuropoly code to add my trainer?

You don't have to! 

My idea was that, once we make [this very small change](https://github.com/spinalcordtoolbox/nnUNet-neuropoly/commit/511204ebed0a780cbdf0e14d011fc037f52f5796), `nnunetv2` can now accept classes here:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4ffde21949945c6a18ceb831ed48cc7c923172f3/spinalcordtoolbox/deepseg/nnunet.py#L54-L59

So, the trainer class could be packaged with your model `.zip`, then SCT can import it ourselves and pass it to this function.

----

The reason I did it this way was to keep the changes in the fork "light" and keep the bulk of SCT's specific trainer code on our side. That way, it makes syncing the fork easier, since we'd just be applying some small changes on top of it.

joshuacwnewton: In terms of prototyping/testing, what you can do is:

- Add trainer class to this PR temporarily (long-term it will live in the model `.zip`, short-term we can include it here just to test)
- Pass the trainer class as a parameter to the above function, something like:

```diff
 # initializes the network architecture, loads the checkpoint 
 predictor.initialize_from_trained_model_folder( 
     join(path_model), 
     use_folds=folds_avail, 
     checkpoint_name=checkpoint_name, 
+    trainer_class=yourCustomNnunetTrainer if "model_seg_ms_lesion" in path_model else None
 ) 
```

Just a rough sketch, but this is the general idea I had.

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -282,7 +289,17 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
...
+        save_or_return_probabilities=soft_ms_lesion,
     )
+    # If we saved the probabilities, `pred` is a tuple of (binary pred, prob map)
+    if soft_ms_lesion:
+        # keep the soft segmentation
+        prob_map = pred[1]
+        # The shape of the prob_map is (num_classes, z, y, x), so we keep only the non-background class (1)
+        pred = prob_map[1]
+        # We threshold the soft segmentation at 1e-3 to avoid having very small values in the output
+        pred[pred < 1e-3] = 0
```
plbenveniste: Instead of doing this, I should do like nnUnet as it proves to perform better. Value for class 1 (=lesion) should be kept if it is higher than value for class 0 (=background).

plbenveniste: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5027/commits/e88b2d69c48bbb455c557327c9f89ed6176eca7f

spinalcordtoolbox/deepseg/nnunet.py
```diff
@@ -51,12 +58,29 @@ def create_nnunet_from_plans(path_model, device: 'torch.device'):
...
+    This functions load a custom nnUNet trainer class if available in the model folder.
+    """
+    trainer_file = os.path.join(path_model, "trainer_class.py")
+    if os.path.exists(trainer_file):
+        spec = importlib.util.spec_from_file_location("trainer_class", trainer_file)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        if hasattr(module, "get_trainer_class"):
+            return module.get_trainer_class()
+    return None
```
plbenveniste: Out of transparency, I want to precise that this function was written by ChatGPT since I never worked with `importlib`. Particular attention should be given when reviewing this function.

joshuacwnewton: It looks like this function is very similar to what I would have recommended anyway: https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly. So it is possible that ChatGPT has come across the above example in its web scraping and internalized it.

The only difference is the fact that it does not include `    sys.modules[module_name] = module`. I think this is okay though, because we do not necessarily need the module to be available to everyone (using `import module_name`)? We only need it for the purposes of fetching the trainer class, which you have done wonderfully.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -450,10 +467,15 @@ def main(argv: Sequence[str]):
...
+                single_fold=getattr(arguments, "single_fold", False),
                 # Pass any "extra" kwargs defined in task-specific subparsers
                 extra_inference_kwargs={arg_name: getattr(arguments, arg_name)
-                                        for arg_name in ["step1_only"]  # Used only by totalspineseg
-                                        if hasattr(arguments, arg_name)}
+                                        for arg_name in ["step1_only", "soft_ms_lesion", "task"]
+                                        if hasattr(arguments, arg_name)
+                                        # "step1_only" -> used only by totalspineseg
+                                        # "soft_ms_lesion" and "task" -> used only by lesion_ms
+                                        },
```
joshuacwnewton: The failing tests are thankfully unrelated to the `nnunetv2-neuropoly` fork changes.

However, I think what is happening is that `task` is getting passed as an extra inference kwarg in a case where it is not valid (i.e. for `totalspineseg`).

So, I think we need a different way to tell `segment_nnunet` whether or not we are running the `lesion_ms` task. 🤔

plbenveniste: Ah damn! Do we want to create a specific function for `segment_lesion_ms`? Or that is not ideal?

joshuacwnewton: Yeah, I think if we can avoid having to create a new function, that would be great... hrm.

Oh! I think I might have an idea? I'll try authoring a commit and pushing it. :)

plbenveniste: I am not sure I understand what differences your changes made, but it seems to have solved the issue!

joshuacwnewton: Sorry for the confusion, let me explain:

- The old code used `hasattr("task")`:
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0a523b68d8fb50e3c159cb15a136c9fe31e3dc37/spinalcordtoolbox/scripts/sct_deepseg.py#L473-L475
- But, every single task has the argument `task`! So, `task` would _always_ get passed along.
- However, the function `segment_totalspineseg` didn't have a `task` parameter, so an error occured:
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/0a523b68d8fb50e3c159cb15a136c9fe31e3dc37/spinalcordtoolbox/deepseg/inference.py#L383

----

My changes fix this by:

- Not using `task` anymore.
- Using `ensemble` instead (only for `lesion_ms`):
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6aea832fe7f0e1188a112ade9294685f899a1b05/spinalcordtoolbox/scripts/sct_deepseg.py#L476-L478
- `ensemble` gets passed to `segment_nnunet`:
    https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6aea832fe7f0e1188a112ade9294685f899a1b05/spinalcordtoolbox/deepseg/inference.py#L223
- However, if `arguments.task == "totalspineseg"`, nothing will happen, so no error will occur for `segment_totalspineseg`.

plbenveniste: Oh okay! Makes sense! Thanks for the explanation

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -282,7 +284,33 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
...
+            # Divide by number of folds to get the average
+            pred = ensembled_pred / len(prob_maps)
+            # We do not binarize the output, since we want the soft segmentation
+        else:
+            # We sum the elements of the list to get the ensembled output
+            ensembled_pred = np.sum(pred, axis=0)
+            # Divide by number of folds to get the average
+            pred = ensembled_pred / len(pred)
+            # Binarize the ensembled output at a low threshold to get the final segmentation
+            pred = binarize(pred, 0.5)
```
joshuacwnewton: To make the code a little more readable, I think it would make sense to make a separate function just for averaging the predictions?

Something like:

```python
   if ensemble:
      pred = average_nnunet_predictions(pred, probabilities=soft_ms_lesion)
```

Then the function could be like:

```python
def average_nnunet_predictions(pred, probabilities=False):
    """
    Compute an average prediction for multi-fold output of `nnunetv2`.
    
    The averaging process will differ depending on whether `pred` contains probability maps or per-fold logits.
    """
    if probabilities:
        # [...]
    else:
        # [...]
```

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -342,6 +374,7 @@ def segment_nnunet(path_img, tmpdir, predictor, device: torch.device):
 
 
 def segment_totalspineseg(path_img, tmpdir, predictor, device, step1_only=False):
+    # Save probabilities is not used in totalspineseg, but we include it in the function signature to match the other segmentation functions
```
joshuacwnewton: I think we can remove this comment? Since, we're not actually modifying the `segment_totalspineseg` function signature (thankfully!).

spinalcordtoolbox/deepseg/models.py
```diff
@@ -516,32 +509,10 @@
...
     'lesion_ms':
         {'description': 'MS lesion segmentation on spinal cord MRI images',
-         'long_description': 'This segmentation model for spinal cord MS lesion segmentation uses a 3D U-Net architecture. It outputs a binary '
-                             'segmentation of MS lesions. We used a foundational model, pretrained on over 10,000 CT scans. The model was fine-tuned '
-                             'and tested on datasets including 20 sites, 1850 patients and 4430 volumes. The datasets, mostly coming from “real world” '
-                             'clinical scans at 1.5T and 3T (on GE, Siemens and Philips), included: T1w (n=23), T2w (n=3061), T2*w (n=548), '
-                             'PSIR (n=363), STIR (n=92), MP2RAGE-UNI (n=343). The field-of-view coverage varied across sites (some included the brain '
-                             'and the upper cord, while others only included the spinal cord), and acquisitions were either 2D (axial: 2895, sagittal: 1169) '
-                             'or 3D (n=366), with voxel dimensions ranging from 0.2x0.2x5 mm3 to 0.8x0.8x9 mm3. ',
+         'long_description': 'TODO',
```
joshuacwnewton: Reminder for TODO.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -516,32 +509,10 @@
...
-                 Mainero, Caterina and Treaba, Constantina Andrada and Kerbrat,  Anne and Bannier, Elise and Edan, Gilles and Labauge, Pierre and
-                 O'Grady, Kristin P and Smith, Seth A and Shepherd, Timothy M and Charlson, Erik and Brisset, Jean-Christophe and Talbott, Jason
-                 and Liu, Yaou and Lombaert, Hervé and Cohen-Adad, Julien},
-                 title={Reinforcing the generalizability of spinal cord multiple sclerosis lesion segmentation models},
-                 booktitle={Proceedings of the 41st Annual Scientific Meeting of the ESMRMB},
-                 year={2025},
-                 address={Marseille, France},
-             }
-             ```"""),  # noqa E501 (line too long),
+         'citation': "TODO",  # noqa E501 (line too long),
```
joshuacwnewton: Reminder for TODO.

spinalcordtoolbox/deepseg/models.py
```diff
@@ -646,6 +617,27 @@ def install_model(name_model, custom_url=None):
...
+                    dst = os.path.join(folder(name_model), next(d for d in os.listdir(folder(name_model)) if d.startswith("nnUNetTrainer")))
+                    if not os.path.exists(dst):
+                        raise FileNotFoundError(f"Could not find destination folder {dst}")
+                    else:
+                        # Move contents of src inside dst
+                        shutil.move(src, os.path.join(dst, f"fold_{i}"))
+            # Then we remove all folders which are not nnUNetTrainer... (leftover from downloading each fold separately)
+            for d in os.listdir(folder(name_model)):
+                if not d.startswith("nnUNetTrainer"):
+                    shutil.rmtree(os.path.join(folder(name_model), d))
```
joshuacwnewton: I don't actually think we need all of this logic? I think we can actually re-use the existing multi-seed download code below.

Before:

```python
        else:
            urls_used = {}
            for seed_name, model_urls in url_field.items():
                logger.info(f"\nInstalling '{seed_name}'...")
                urls_used[seed_name] = download.install_data(model_urls,
                                                             folder(os.path.join(name_model, seed_name)), keep=True)
```

After:

```python
        else:
            urls_used = {}
            for seed_name, model_urls in url_field.items():
                # For lesion_ms, we can extract all the folds to the same `nnunetTrainer` directory 
                # For other multi-seed models, create subfolders for each seed
                if name_model == "model_seg_ms_lesion":
                    target_directory = folder(name_model)
                    dirs_to_preserve = ("nnUNetTrainer",)
                else:
                    target_directory = folder(os.path.join(name_model, seed_name))
                    dirs_to_preserve = ()
                logger.info(f"\nInstalling '{seed_name}'...")
                urls_used[seed_name] = download.install_data(model_urls, target_directory, keep=True,
                                                             dirs_to_preserve=dirs_to_preserve)
```

I tried this and it seemed to work OK for me? :)

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -307,6 +307,23 @@ def get_parser(subparser_to_return=None):
...
+            )
+
+        # Add possibility of segmenting on only 1 fold for quicker inference
+        if task_name == 'lesion_ms':
+            params.add_argument(
+                "-single-fold",
+                action="store_true",
+                help="If set, only 1 fold will be used for inference instead of the full 5-fold ensemble. This will speed up inference, but may reduce segmentation quality."
+            )
+
```
joshuacwnewton: I think we can move these arguments up so that they are next to the other "task-specific" args?

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton, line 282: Right here!

requirements.txt
```diff
@@ -75,7 +67,7 @@ requirements-parser
 scipy
 scikit-image
 scikit-learn
-totalspineseg
+totalspineseg[neuropoly] @ git+https://github.com/neuropoly/totalspineseg@jn/switch-to-nnunetv2-neuropoly-fork
```
joshuacwnewton: Once a new version of `totalspineseg` is released, we can remove this `git` specifier.

spinalcordtoolbox/deepseg/inference.py
joshuacwnewton: There is another instance for for `segment_monai` we can remove as well.

spinalcordtoolbox/deepseg/inference.py
plbenveniste: Done!

spinalcordtoolbox/deepseg/inference.py
plbenveniste: Good catch!

spinalcordtoolbox/deepseg/models.py
plbenveniste: Done! Thanks for the reminder!

spinalcordtoolbox/deepseg/models.py
plbenveniste: Good catch! We are in the process on submitting the manuscript to a journal so we don't have a reference yet! I am opening an issue to keep this in mind: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5062

spinalcordtoolbox/deepseg/models.py
plbenveniste: Much nicer indeed! Done!

spinalcordtoolbox/scripts/sct_deepseg.py
plbenveniste: Done!

spinalcordtoolbox/scripts/sct_deepseg.py
plbenveniste: Gotcha!

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: Perfect! I think we should probably switch this to "" for now then, since the text "TODO" will show up in the page for the `lesion_ms` model?

spinalcordtoolbox/deepseg/nnunet.py
```diff
@@ -29,6 +34,12 @@ def create_nnunet_from_plans(path_model, device: 'torch.device'):
     if not fold_dirs:
         raise FileNotFoundError(f"No 'fold_*' directories found in model path: {path_model}")
     folds_avail = 'all' if fold_dirs == ['fold_all'] else [int(f.split('_')[-1]) for f in fold_dirs]
+    if single_fold:
+        folds_avail_temp = folds_avail
+        folds_avail = [1]  # use only fold 1 it exists for all models (as it was the best for the lesion_ms model)
+        if 1 not in folds_avail:
+            folds_avail = [folds_avail[0]]  # otherwise use the first fold available
+        print(f'Using single fold: {folds_avail} for inference instead of the full ensemble of {sorted(folds_avail_temp)}')
```
joshuacwnewton, line 43: I think there might be a bug here due to the use of `folds_avail_temp` and `folds_avail`? Since we assign `[1]` to `folds_avail`, then the condition `1 not in folds_avail` will always evaluate to `False`.

I think maybe we could simplify this by doing something like:


```suggestion
    if single_fold:
        # use only fold 1 it exists for all models (as it was the best for the lesion_ms model)
        # otherwise use the first fold available
        folds_avail = [1] if 1 in folds_avail else [folds_avail[0]]  
        print(f'Using single fold: {folds_avail} for inference instead of the full ensemble of {sorted(folds_avail_temp)}')
```

plbenveniste: Commited! Much nicer indeed!

plbenveniste:
Amazing! Thanks @joshuacwnewton 🚀

PR #5030, opened 2025-09-18
Add new `-exclude-file` argument to support exclusion of subjects/files based on YAML list

### LINKED ISSUE #3971: `sct_run_batch` - exclude subjects using yml/json file
## Feature justification

If users want to exclude subjects from the analysis run by the `sct_run_batch`, they have to list these subjects under the `'-exclude-list'` flag: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/812a8e00bd8862c3c9f2d099e24e0143a5337976/spinalcordtoolbox/scripts/sct_run_batch.py#L124-L127

It would also be helpful to have the option to provide the yml/json file listing subjects to exclude. For example, by a flag `-exclude-config`. 
The justification for providing the yml/json exclude file instead of `'-exclude-list'` is the fact that the yml/json exclude file can be easily stored within a repository, allowing higher reproducibility of analysis.

The same could also be applied to the `'-include-list'` flag.

The user discussion and example `exclude.yml` file [here](https://github.com/ivadomed/canproco/pull/10).

joshuacwnewton:
## Checklist

## Description

This PR contains a new argument `-exclude-file` that allows a YAML file to be passed. The YAML file must contain a list of either subjects or files. I based this off of the [`exclude.yml`](https://github.com/sct-pipeline/ukbiobank-spinalcord-csa/blob/04b9dd20bf2ad646e734e3e3effae174d4e451ca/exclude.yml) file in `ukbiobank`.

The list will be treated differently depending on whether subjects or files are listed:
 - Subjects (e.g. `sub-01` or `sub-01/ses-01`): The subject will not be processed.
 - Files (e.g. `sub-01_ses-01_T1w.nii.gz`): The script will still be run, however an environment variable called `EXCLUDE_FILES` will be created containing the list of files. That way, the user can choose to skip specific processing steps within their script.

This follows the conclusions we came to in a previous SCT meeting: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3971#issuecomment-3140354514

I've tried to provide example `bash` syntax for using `EXCLUDE_FILES` to skip processing. I hope it is clear.

## Discussion

- Is checking for `.nii` and `.nii.gz` a robust enough way of determining whether the list is excluding files or folders?
- When testing out the "Save Fails" button on the QC report, I noticed that the format actually does not match the expected format:
    image
    Not only is `t2.nii.gz` the same file every time, but there are category headers, plus the full path gets saved too:
    ```yml
    FILES_SEG:
        - /home/joshua/data/forum_qq_reg/clean/data_processed/1/L/t2/t2.nii.gz
    FILES_REG:
        - /home/joshua/data/forum_qq_reg/clean/data_processed/1/L/t2/t2.nii.gz
        - /home/joshua/data/forum_qq_reg/clean/data_processed/1/L/t2/t2.nii.gz
    ```
    I'm wondering how best to handle this file format. I could collapse all of the different categories into a single list, and only take the base filename (`t2.nii.gz`)?
    This would lose a lot of context, but would be the simplest form for `EXCLUDE_FILES`.
    This would also require users to make sure their filenames are distinct (e.g. `{sub}_t2.nii.gz`), but maybe that's a good thing?

Any thoughts would be appreciated!

## Linked issues

Fixes #3971.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_run_batch.py
+++ b/spinalcordtoolbox/scripts/sct_run_batch.py
@@ -26,6 +26,7 @@ import shutil
 from typing import Sequence
 from types import SimpleNamespace
 import textwrap
+import argparse
 
 import yaml
 import psutil
@@ -38,6 +39,20 @@ from spinalcordtoolbox.utils.fs import Tee
 from stat import S_IEXEC
 
 csi_filter.register_codec()
+
+
+class ParseExcludeFileAction(argparse.Action):
+    """Reads in a YML file and determines if exclusion list contains subjects and/or files."""
+    def __call__(self, parser, namespace, values, option_string=None):
+        with open(values, 'r') as yaml_file:
+            yaml_contents = yaml.safe_load(yaml_file)
+        if not isinstance(yaml_contents, list) and all(isinstance(item, str) for item in yaml_contents):
+            raise ValueError(f"The exclude file {values} should contain a list of subjects or files to exclude.")
+        yaml_data = {
+            'files': [item for item in yaml_contents if (item.endswith('.nii')) or item.endswith('.nii.gz')],
+            'subjects': [item for item in yaml_contents if not (item.endswith('.nii')) or item.endswith('.nii.gz')]
+        }
+        setattr(namespace, self.dest, yaml_data)
 
 
 def get_parser():
@@ -135,6 +150,28 @@ def get_parser():
                             Examples: `-exclude-list sub-003 sub-004` or `-exclude-list sub-003/ses-01 ses-02`
         """),  # noqa: E501 (line too long)
                         nargs='+')
+    parser.add_argument('-exclude-file',
+                        action=ParseExcludeFileAction,
+                        help=textwrap.dedent("""
+                            Path to a YAML file (e.g. `exclude.yml`) containing a bullet list (`-`) of subjects or files to exclude.
+
+                            The list will be treated differently depending on whether subjects or files are listed:
+                              - Subjects (e.g. `sub-01` or `sub-01/ses-01`): The subject will not be processed (script will not be run).
+                              - Files (e.g. `sub-01_ses-01_T1w.nii.gz`): The script will still be run, however an environment variable called `EXCLUDE_FILES` will be created containing the list of files. That way, you can choose to skip specific processing steps within your script.
+
+                            You can use the following `bash` syntax to skip files:
+
+                                # check if FILE_T2 is in the list of excluded files
+                                FILE_T2="$SUBJECT"_T2w.nii.gz
+                                if [[ ! " $EXCLUDE_FILES " =~ " $FILE_T2 " ]]; then
+                                    # process file only if it is not excluded
+                                    FILE_T2_SEG="$SUBJECT"_T2w_seg.nii.gz
+                                    sct_deepseg spinalcord $FILE_T2 -o $FILE_T2_SEG
+                                fi
+
+                            Note: You can use QC reports to generate a YAML file with excluded files (via the "Save Fails" button).
+                            """)   # noqa: E501 (line too long)
+                        ),
     parser.add_argument('-ignore-ses', action='store_true',
                         help="By default, if 'ses' subfolders are present, then 'sct_run_batch' will run the script "
                              "within each individual 'ses' subfolder. Passing `-ignore-ses` will change the behavior "
@@ -257,7 +294,7 @@ def _filter_directories(dir_list, include=None, include_list=None, exclude=None,
 
 
 def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_data_processed, path_results, path_log,
-               path_qc, itk_threads, continue_on_error=False):
+               path_qc, itk_threads, continue_on_error=False, exclude_files=None):
     """
     Job function for mapping with multiprocessing
     :param subj_dir:
@@ -306,6 +343,8 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
     })
     if 'SCT_DIR' not in envir:
         envir['SCT_DIR'] = __sct_dir__
+    if exclude_files:
+        envir['EXCLUDE_FILES'] = ' '.join(exclude_files)
 
     cmd = [script_full, subj_dir] + script_args.split(' ')
 
@@ -524,12 +563,17 @@ def main(argv: Sequence[str]):
     if (arguments.include is not None) and (arguments.include_list is not None):
         parser.error('Only one of `include` and `include-list` can be used')
 
-    if (arguments.exclude is not None) and (arguments.exclude_list is not None):
-        parser.error('Only one of `exclude` and `exclude-list` can be used')
+    if sum([arg is not None for arg in [arguments.exclude, arguments.exclude_list, arguments.exclude_file]]) > 1:
+        parser.error('Only one of `exclude`, `exclude-list`, and `exclude-file` can be used')
+
+    # Check `-exclude-list` (or if `-exclude-file` contains subjects)
+    exclude_list = arguments.exclude_list
+    if arguments.exclude_file['subjects']:
+        exclude_list = arguments.exclude_file['subjects']
 
     subject_dirs = _filter_directories(subject_dirs,
                                        include=arguments.include, include_list=arguments.include_list,
-                                       exclude=arguments.exclude, exclude_list=arguments.exclude_list)
+                                       exclude=arguments.exclude, exclude_list=exclude_list)
 
     # Determine the number of jobs we can run simultaneously
     if arguments.jobs < 1:
@@ -557,7 +601,8 @@ def main(argv: Sequence[str]):
                                                path_log=path_log,
                                                path_qc=path_qc,
                                                itk_threads=arguments.itk_threads,
-                                               continue_on_error=arguments.continue_on_error)
+                                               continue_on_error=arguments.continue_on_error,
+                                               exclude_files=arguments.exclude_file['files'])
             results = list(p.imap(run_single_dir, subject_dirs))
     except Exception as e:
         if do_email:
--- a/testing/cli/test_sct_run_batch.py
+++ b/testing/cli/test_sct_run_batch.py
@@ -3,6 +3,7 @@
 import glob
 import os
 import json
+import yaml
 
 import pytest
 from stat import S_IEXEC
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -40,6 +41,20 @@
 csi_filter.register_codec()
 
 
+class ParseExcludeFileAction(argparse.Action):
+    """Reads in a YML file and determines if exclusion list contains subjects and/or files."""
+    def __call__(self, parser, namespace, values, option_string=None):
+        with open(values, 'r') as yaml_file:
+            yaml_contents = yaml.safe_load(yaml_file)
+        if not isinstance(yaml_contents, list) and all(isinstance(item, str) for item in yaml_contents):
+            raise ValueError(f"The exclude file {values} should contain a list of subjects or files to exclude.")
```
joshuacwnewton: This is the part that is incompatible with the "QC Fails" yaml format, but it can easily be updated to parse that format.

testing/cli/test_sct_run_batch.py
```diff
@@ -90,6 +112,50 @@ def test_only_one_include_exclude(tmp_path, dummy_script):
...
+        (data / sub / 'anat').mkdir(parents=True)
+    out = tmp_path / 'out'
+
+    # generate `exclude.yml`
+    sub_exclude_list = sub_dir_list[::2]  # exclude every other subject
+    exclude_list = (sub_exclude_list if exclusion_type == 'subject'
+                    else [f"{sub}_T2w.nii.gz" for sub in sub_exclude_list])
+    fname_exclude = tmp_path / 'exclude.yml'
+    with open(fname_exclude, 'w') as fp:
+        yaml.dump(exclude_list, fp)
```
joshuacwnewton: Here is a sample `exclude.yml`:

```
- sub-001
- sub-003
- sub-011
```

When parsed, here is what will show in the log for `sct_run_batch`:

```
INPUT ARGUMENTS
---------------

# [...]
exclude: null
exclude_file:
  files: []
  subjects:
  - sub001
  - sub003
  - sub011
```

On the other hand, if files are provided (ending in `.nii.gz`), then this will be the log instead:

```
INPUT ARGUMENTS
---------------

# [...]
exclude: null
exclude_file:
  files:
  - sub-001_T2w.nii.gz
  - sub-003_T2w.nii.gz
  - sub-011_T2w.nii.gz
  subjects: []
```

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -524,12 +563,17 @@ def reset_streams():
...
 
-    if (arguments.exclude is not None) and (arguments.exclude_list is not None):
-        parser.error('Only one of `exclude` and `exclude-list` can be used')
+    if sum([arg is not None for arg in [arguments.exclude, arguments.exclude_list, arguments.exclude_file]]) > 1:
+        parser.error('Only one of `exclude`, `exclude-list`, and `exclude-file` can be used')
+
+    # Check `-exclude-list` (or if `-exclude-file` contains subjects)
+    exclude_list = arguments.exclude_list
+    if arguments.exclude_file['subjects']:
+        exclude_list = arguments.exclude_file['subjects']
```
joshuacwnewton: Whoops, this line is causing an issue:

```
TypeError: 'NoneType' object is not subscriptable
```

Should be an easy fix:

```suggestion
    if arguments.exclude_file is not None and arguments.exclude_file['subjects']:
        exclude_list = arguments.exclude_file['subjects']
```

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -557,7 +601,8 @@ def reset_streams():
                                                path_log=path_log,
                                                path_qc=path_qc,
                                                itk_threads=arguments.itk_threads,
-                                               continue_on_error=arguments.continue_on_error)
+                                               continue_on_error=arguments.continue_on_error,
+                                               exclude_files=arguments.exclude_file['files'])
```
joshuacwnewton:
```suggestion
                                               exclude_files=(arguments.exclude_file['files'] if arguments.exclude_file is not None else None)
```

joshuacwnewton (commented):
Whoops, tests are failing!

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -40,6 +41,44 @@
 csi_filter.register_codec()
 
 
+class ParseYMLAction(argparse.Action):
+    """Reads in a YML file and determines if it contains subjects and/or files."""
+    def __call__(self, parser, namespace, values, option_string=None):
+        with open(values, 'r') as yaml_file:
+            yaml_contents = yaml.safe_load(yaml_file)
```
mguaypaq: Note that `parser.parse_args()` only handles exceptions of type `argparse.ArgumentError` (for which it displays the script usage string, and the error message itself). All other exception types (file not found, invalid YAML, ValueErrors, etc.) instead result in a big traceback. Is this what we want?

If we do want to use `argparse.ArgumentError`, note that it should be instantiated as `raise argparse.ArgumentError(self, "some message")` from inside the `__call__(self, ...)` of an `argparse.Action` subclass. Unlike most exception types, which only need `"some message"` and no `self`.

This comment also applies to the 3 instances of `raise ValueError` below.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -40,6 +41,44 @@
 csi_filter.register_codec()
 
 
+class ParseYMLAction(argparse.Action):
```
mguaypaq, line 90: Thanks for generalizing this so it can be used for both `-include-yml` and `-exclude-yml`.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -124,6 +163,27 @@ def get_parser():
...
+                        action=ParseYMLAction,
+                        help=textwrap.dedent("""
+                            Path to a YAML file (e.g. `include.yml`) containing a bullet list (`-`) of subjects or files to include.
+
+                            The list will be treated differently depending on whether subjects or files are listed:
+                              - Subject/session folders (e.g. `sub-01` or `sub-01/ses-01`): The same as `-include-list` (subject/session inclusion).
+                              - Files (e.g. `sub-01_ses-01_T1w.nii.gz`): The script will be run, and an environment variable called `INCLUDE_FILES` will be created
+                              containing the list of files. That way, you can choose to run specific processing steps within your script for only those files.
+
+                            You can use the following `bash` syntax to process included files:
```
mguaypaq, line 190: This seems like a very useful snippet!

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -306,6 +388,10 @@ def run_single(subj_dir, script, script_args, path_segmanual, path_data, path_da
     })
     if 'SCT_DIR' not in envir:
         envir['SCT_DIR'] = __sct_dir__
+    if include_files:
+        envir['INCLUDE_FILES'] = ' '.join(include_files)
+    if exclude_files:
+        envir['EXCLUDE_FILES'] = ' '.join(exclude_files)
```
mguaypaq, line 408: A shell purist would tell you that this is unsafe for filenames which could contain whitespace, but:
- such filenames are widely known to cause problems, and
- the usual advice to delimit filenames with null bytes doesn't apply, because environment variables cannot contain null bytes.

So, this is fine.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -40,6 +41,44 @@
...
+        set_entries = set()
+        for entry_list in lists_to_merge:
+            for entry in entry_list:
+                if not isinstance(entry, str):
+                    raise ValueError(f"The YML file {values} should contain filename strings, but encountered {entry}.")
+                set_entries.add(entry)
+
+        # sort entries into files or subjects
+        yaml_data = {
+            'files': [item for item in set_entries if (item.endswith('.nii') or item.endswith('.nii.gz'))],
```
mguaypaq: Maybe the help strings should mention that our definition of "file" is "ends with .nii or .nii.gz"? But maybe the text is already long enough.

spinalcordtoolbox/scripts/sct_run_batch.py
```diff
@@ -521,15 +607,25 @@ def reset_streams():
 
     subject_dirs = _parse_dataset_directory(path_data, arguments.subject_prefix, arguments.ignore_ses)
 
-    if (arguments.include is not None) and (arguments.include_list is not None):
-        parser.error('Only one of `include` and `include-list` can be used')
+    if sum([arg is not None for arg in [arguments.include, arguments.include_list, arguments.include_yml]]) > 1:
+        parser.error('Only one of `include`, `include-list`, and `include-yml` can be used')
+
+    if sum([arg is not None for arg in [arguments.exclude, arguments.exclude_list, arguments.exclude_yml]]) > 1:
```
mguaypaq: `sum` can take a generator rather than a list, so you can save some brackets:
```suggestion
    if sum(arg is not None for arg in [arguments.include, arguments.include_list, arguments.include_yml]) > 1:
        parser.error('Only one of `include`, `include-list`, and `include-yml` can be used')

    if sum(arg is not None for arg in [arguments.exclude, arguments.exclude_list, arguments.exclude_yml]) > 1:
```

testing/cli/test_sct_run_batch.py
```diff
@@ -90,6 +120,60 @@ def test_only_one_include_exclude(tmp_path, dummy_script):
...
+
+    # generate `include.yml` or `exclude.yml` depending on test case
+    sub_exclude_list = sub_dir_list[::2]   # exclude every other subject
+    sub_include_list = sub_dir_list[1::2]  # include the non-excluded subjects
+    yml_to_write = sub_exclude_list if __clude == "exclude" else sub_include_list
+    yml_to_write = (yml_to_write if entry_type == 'subject'
+                    else [f"{sub}_T2w.nii.gz" for sub in yml_to_write])
+    # mimic the YML formatting of the QC report (dict of lists)
+    if yml_format == 'dict':
+        yml_to_write = {f"FILES_{value}": [value] for value in yml_to_write}
```
mguaypaq: This is maybe pushing the line of how much logic there should be in a test case, but ok.

mguaypaq (changes requested):
While testing this out, I ran into the bug that `-include-yml` and `-exclude-yml` don't work when specified in a `-config`. This is because `-config` is implemented by sticking the config values in `parser.set_defaults`, and this bypasses any `action=` processing defined in `parser.add_argument()`.

Arguably this is also a bug for the arguments which use `action='store_true'`, but python's truthy value handling (mostly?) makes it work in that case.

spinalcordtoolbox/scripts/sct_run_batch.py
joshuacwnewton: Not to mention that we're only using this for checking "if filename in this list of filenames" (via string comparison specifically) and not for any sort of ingestion of the filenames. :sweat_smile:

testing/cli/test_sct_run_batch.py
joshuacwnewton: Since most of this logic is really just tweaking different variations of `include.yml` and `exclude.yml` (and not actually test logic itself), I've tried moving this logic into input test fixtures: [0af683a](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5030/commits/0af683abb4ebba72f2c9896d25162b28d06afb33)

I'm hoping that this makes the actual test logic simpler (though it does involve a lot of parametrization boilerplate that might end up being overkill).

PR #5047, opened 2025-10-01
Refactor the broken link checker

### LINKED ISSUE #5045: Fix broken link checker for changed lines in PR
I mistakenly merged a fix for the broken link checker (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5044/commits/d8d67e4b37e0af8c32a5232d0c2d0451b6b42db3) in #5044:

> This is a tricky corner case: `grep -v` exits with status 1 when two conditions are met:
> 1. at least one matching line was excluded; and
> 2. every line was excluded.

On this other PR (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5040), I'm now seeing this failure:

```
 grep: spinalcordtoolbox/scripts/sct_dmri_moco.py
spinalcordtoolbox/scripts/sct_fmri_moco.py
spinalcordtoolbox/utils/shell.py
testing/cli/test_cli_sct_dmri_moco.py
testing/cli/test_cli_sct_fmri_moco.py: No such file or directory
```

Testing locally, this error comes from the following command:

```console
> echo $git_diff_files
spinalcordtoolbox/scripts/sct_dmri_moco.py spinalcordtoolbox/scripts/sct_fmri_moco.py spinalcordtoolbox/utils/shell.py testing/cli/test_cli_sct_dmri_moco.py testing/cli/test_cli_sct_fmri_moco.py
> git_diff_urls=$(grep -HEio '\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]' "$git_diff_files" || (($? == 1)) )
grep: spinalcordtoolbox/scripts/sct_dmri_moco.py
spinalcordtoolbox/scripts/sct_fmri_moco.py
spinalcordtoolbox/utils/shell.py
testing/cli/test_cli_sct_dmri_moco.py
testing/cli/test_cli_sct_fmri_moco.py: No such file or directory
```

I think the issue is that `$git_diff_files` is a space-delimited string, rather than an array, so [`grep` doesn't parse it properly](https://stackoverflow.com/questions/62856618/grep-multiple-files-whose-file-names-are-stored-in-a-variable). (Potential fixes: https://stackoverflow.com/questions/9293887/how-to-read-a-space-delimited-string-into-an-array-in-bash)

I think this issue has actually been present for a while now, but masked by the old `echo ""` command, and @mguaypaq's fix revealed it. Perhaps we haven't been checking URLs on PRs for some time now.

And, the reason it wasn't revealed on #5044 is because that PR had only 1 changed file, and hence the space-delimiting was not present.

mguaypaq:
Here's a rewrite of the broken link checker, as a standalone bash script, so that:
- we don't have to repeat mostly the same logic twice; and
- it's checkable by shellcheck.

The new `check-broken-links.sh` script accepts a list of files to check as its arguments, which can come from either:
- a `find` command for the daily run which checks everything; or
- a `git diff` command for the files touched by any given pull request.

In either case, the exclude list which used to be in `check_url_blacklist.txt` is now included inline in the `check-broken-links.sh` script. It could be restored as a file by using `grep --exclude-from` instead of `grep --exclude`, but I think that would make a future reader less likely to check the documentation of `grep --exclude` to see how the exclusions are interpreted (it's somewhat subtle).

I took the liberty of changing `check-url.sh` so that it takes two real arguments, rather than having to split around a semicolon.

Fixes #5045.

### DIFF
--- a/.github/workflows/check-broken-links.sh
+++ b/.github/workflows/check-broken-links.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+# Find and check each URL in the files provided as arguments to the script.
+# The (($# == 0)) test is to prevent grep from using standard input if no files were given.
+# The grep -H option is to output the filename even if a single file was given.
+# The grep -Zz options are to make the filename and matching URL terminated by null bytes,
+# which is the safest way to delimit arbitrary text.
+# The (($? == 1)) test is to prevent the script from exiting if no URLs were found.
+# Setting IFS= (the empty string) prevents `read` from trimming whitespace.
+# The </dev/null redirection is to prevent check-url.sh from stealing the output of grep.
+{ (($# == 0)) || grep -EioHZz \
+  --exclude 'sct.def' \
+  --exclude 'CHANGES.md' \
+  --exclude 'requirements.txt' \
+  --exclude 'requirements-freeze.txt' \
+  '\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]' \
+  -- "$@" || (($? == 1)) ; } |
+  while IFS= read -rd '' filename && IFS= read -rd '' url; do
+    .github/workflows/check-url.sh "$filename" "$url" </dev/null
+  done
+
+# Summarize the results
+touch valid_urls.txt redirected_urls.txt invalid_urls.txt
+NUM_OK=$(wc -l valid_urls.txt | cut -d " " -f 1)
+NUM_REDIRECT=$(wc -l redirected_urls.txt | cut -d " " -f 1)
+NUM_BAD=$(wc -l invalid_urls.txt | cut -d " " -f 1)
+cat invalid_urls.txt
+echo -en "========== \033[0;32m$NUM_OK passed\033[0;0m, \033[0;33m$NUM_REDIRECT redirected\033[0;0m, \033[0;31m$NUM_BAD failed\033[0;0m ==========\n"
+
+# Exit with failure if there are any bad URLs
+((NUM_BAD == 0))
--- a/.github/workflows/check-code-quality.yml
+++ b/.github/workflows/check-code-quality.yml
@@ -24,7 +24,7 @@ jobs:
         set -eu -o pipefail
         git diff --name-only --diff-filter=d \
           --merge-base refs/remotes/origin/master -- \
-          | (grep '\.py$' || (($? == 1)) ) \
+          | { grep '\.py$' || (($? == 1)) ; } \
           | xargs --delimiter='\n' --no-run-if-empty flake8
 
     - name: Check shell scripts with shellcheck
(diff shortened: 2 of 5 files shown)


.github/workflows/check-url.sh
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
 
-if [ $# -ne 1 ]; then
+if [ $# -ne 2 ]; then
     exit 1;
 fi
 
-filename=$(cut -d ";" -f 1 <<< "$1")
-URL=$(cut -d ";" -f 2 <<< "$1")
```
joshuacwnewton, line 8: Thanks for this! I recall the `:`->`;` logic being a bit of a pain, so this seems much more sensible. <3

.github/workflows/check-broken-links.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Find and check each URL in the files provided as arguments to the script.
+# The (($# == 0)) test is to prevent grep from using standard input if no files were given.
+# The grep -H option is to output the filename even if a single file was given.
+# The grep -Zz options are to make the filename and matching URL terminated by null bytes,
+# which is the safest way to delimit arbitrary text.
+# The (($? == 1)) test is to prevent the script from exiting if no URLs were found.
+# Setting IFS= (the empty string) prevents `read` from trimming whitespace.
+# The </dev/null redirection is to prevent check-url.sh from stealing the output of grep.
+{ (($# == 0)) || grep -EioHZz \
```
joshuacwnewton, line 14: For the sake of logging, is there any way we can echo what gets passed to this `check-broken-links.sh` script? (Like, do the `(($# == 0))` test separately, and if the input args aren't empty, then assign the input to a bash array, echo the length of the array + the contents, and _then_ iterate over it? And, if the input args _are_ empty, maybe we could emit a "No files to check" message?)

It would be nice to have a way to confirm which files are actually being tested (i.e. which files were fetched from `git diff`/`find`)? It might be helpful for debugging in case no URLs are found, or if URLs are found in some files but not others. In fact, right now on this PR, it's not possible to see if the workflow is passing because it didn't find URLs, if it's passing because there were no valid files to check, or if it's passing because it's actually silently faulty. 😝 

It might also be a little more self documenting since we would have a space to outline what the expected input args are for the script.

.github/workflows/check-code-quality.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         set -eu -o pipefail
         git diff --name-only --diff-filter=d \
           --merge-base refs/remotes/origin/master -- \
-          | (grep '\.py$' || (($? == 1)) ) \
+          | { grep '\.py$' || (($? == 1)) ; } \
```
joshuacwnewton, line 27: Presumably this change came about because you were creating a similar pipe in the same file, noticed the use of a subshell, and changed it to a command group to avoid the unnecessary subshell?

.github/workflows/check-broken-links.sh
```diff
@@ -0,0 +1,32 @@
...
+  --exclude 'sct.def' \
+  --exclude 'CHANGES.md' \
+  --exclude 'requirements.txt' \
+  --exclude 'requirements-freeze.txt' \
+  '\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]' \
+  -- "$@" || (($? == 1)) ; } |
+  while IFS= read -rd '' filename && IFS= read -rd '' url; do
+    .github/workflows/check-url.sh "$filename" "$url" </dev/null
+  done
```
joshuacwnewton, line 23: This is really curious! I didn't realize you could pipe directly into a `while` loop? 🤔

.github/workflows/check-broken-links.sh
```diff
@@ -0,0 +1,32 @@
...
+# The </dev/null redirection is to prevent check-url.sh from stealing the output of grep.
+{ (($# == 0)) || grep -EioHZz \
+  --exclude 'sct.def' \
+  --exclude 'CHANGES.md' \
+  --exclude 'requirements.txt' \
+  --exclude 'requirements-freeze.txt' \
+  '\b(https?)://[-a-z0-9+&@#/%?=~_|$!:,.;]*[a-z0-9+&@#/%=~_|$]' \
+  -- "$@" || (($? == 1)) ; } |
+  while IFS= read -rd '' filename && IFS= read -rd '' url; do
+    .github/workflows/check-url.sh "$filename" "$url" </dev/null
```
joshuacwnewton, line 22: Do we want to do any error handling for `check-url.sh` here? Like, if `curl` fails for some unexpected reason? I guess in such a case we would re-run the workflow anyway, so it's better to just fail early.

.github/workflows/check-broken-links.sh
joshuacwnewton: Actually, for the sake of logging, would it be possible to not directly read the filename/URL pairs from the pipe, and instead do the same sort of no-urls-checking + logging as the file check earlier? (e.g. output "No URLs found" if the output of grep is empty?)

.github/workflows/check-code-quality.yml
mguaypaq: Exactly! By using the same idiom in both places, we don't have to worry about whether there was an intentional difference.

.github/workflows/check-broken-links.sh
mguaypaq: Good idea. I'll add more logging here.

.github/workflows/check-broken-links.sh
mguaypaq: Re: piping to a while loop, the previous code already had a redirection into a while loop (with `while ... done <<< $(...)`), just of a different flavour. Redirections in bash are really mind-bending!

Re: logging in-between `grep` and `check-url.sh`, this would be a little tricky (because of handling the null bytes) but possible. However, each call to `check-url.sh` ends up reflected in the existing summary at the end already, no?

.github/workflows/check-broken-links.sh
mguaypaq: I think that's the responsibility of the `check-url.sh` script itself (which already does some number of retries, etc.), and out of scope for this refactoring. I'm also not sure what we would do differently in case of unexpected errors anyway.

.github/workflows/check-broken-links.sh
mguaypaq: Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5047/commits/d34e75f7acce347d51b9cebdf8aea171f869c0a6.

.github/workflows/check-broken-links.sh
joshuacwnewton: Good call! Yeah, I think it's totally fine then. :)

.github/workflows/check-broken-links.sh
joshuacwnewton: Aha!

image

> ```
> '\.(txt|md|py|rst)$'
> ```

Checks out! This PR only modifies .sh and .yml files (though it deletes a `.txt` file).

Issue #5026, opened 2025-09-08
Inconsistent slice numbering between SCT QC and FSLeyes


valosekj:
### Description

Slice numbering in SCT QC does not match the slice numbering in FSLeyes:
 
<img width="1100" height="325" alt="Image" src="https://github.com/user-attachments/assets/e9851cd6-024d-4187-a121-f7cf5ae11d1f" />

<img width="1100" height="567" alt="Image" src="https://github.com/user-attachments/assets/7d94cf84-6362-49fe-9e9a-703aec55a377" />

### Steps to Reproduce

RPI-oriented cord seg (attached):

```console
$ orientation sub-001_acq-ax_T2w_seg-manual.nii.gz
RPI
```

```console
sct_qc -i sub-001_acq-ax_T2w_seg-manual.nii.gz  -s sub-001_acq-ax_T2w_seg-manual.nii.gz  -p sct_deepseg_sc -qc ./qc
```

(seg only to not share the images)

[sub-001_acq-ax_T2w_seg-manual.nii.gz](https://github.com/user-attachments/files/22203785/sub-001_acq-ax_T2w_seg-manual.nii.gz)

mguaypaq:
Possibly the slices are in reverse order because the SCT QC convention is SAL (rather than RPI like the rest of SCT). I guess what we should do is:
- Still display the slices in the same order (so that the superior-most slice is at the top left).
- Adjust the overlaid numbering so that it matches the original image orientation (that is, increasing for 0 to n-1 if the original image orientation contained "S", and decreasing from n-1 to 0 if the original image orientation contained "I").

Is that what you would like?

valosekj:
> Possibly the slices are in reverse order because the SCT QC convention is SAL (rather than RPI like the rest of SCT).

Good point. Indeed, that might actually be the case.

> Adjust the overlaid numbering so that it matches the original image orientation

I would lean toward this option. In `sct_analyze_lesion` QC, we use a similar approach and report slices in the original orientation using [spinalcordtoolbox.image.rpi_slice_to_orig_orientation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/5188547b5f426fbb8eb05f75c9c27c01dee6b12a/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L18C44-L18C73). But it's just my personal opinion, curious what others think.
[...]

Issue #4984, opened 2025-07-23
Use more granular timestamp when generating QCs to prevent duplicate entries


joshuacwnewton:
**Context:** SCT's `pytest` tests often run their own `-qc` tests, but that we don't often check over manually (because it's tedious to do so). I'm working to fix this, so for the first time, I'm looking over all of the reports generated by our test suite:

- [Sample run](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/16453719566): [Pytest QC (ubuntu 24.04).zip](https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/16453719566/artifacts/3590915016)

One thing I notice is that the first few entries (which are dummy entries generated by the test [`test_qc_parallel.py :: test_many_qc`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/43ab1a1297960f6ce3722565050e57cff9b39bf2/testing/api/test_qc_parallel.py#L19-L39) cause some weird glitches when using the arrow key.

I think this is because of how we uniquely determine row IDs... the entries end up with the same timestamp (down to the second), meaning they are considered identical in the eyes of the QC report, and thus we can't uniquely fetch the row ID.

If we added even a little more precision, then perhaps we could uniquely identify the rows. Or, alternatively, we could consider this a WONTFIX due to the improbable conditions (identical timestamp + filename + folder) which would not occur in real world usage (even for identical timestamp reports, there would be different subject IDs and etc.)

PR #4921, opened 2025-06-04
Allow the use of custom mapping for `sct_label_vertebrae`

NathanMolinier:
## Checklist

- [x] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)
- [ ] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [ ] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
[...]

PR #5050, opened 2025-10-03
WIP: Proof of concept for qc refactor

mguaypaq:
Here is a proof-of-concept re-implementation of the QC report for `sct_register_multimodal`. @joshuacwnewton I'd love to get your opinion, and I'm sorry for the large diff.

I tried to make it modular, so that parts of it could be re-used for other QC reports with minimal copy-pasting. In particular, I wanted to cleanly separate two steps that were previously intermixed:
- The step of determining the exact orientation and resolution and position of the slices we want to show (the "slicing spec").
- The step of extracting these slices from each relevant image.

I implemented this "slicing spec" for axial slices, but I think it will work well with sagittal slices too.

Two incidental benefits are that:
- I was able to center each axial slice individually, with possibly-fractional coordinates. (The previous code could only do integer coordinates, but this was *after* resampling to a fixed resolution, so this wasn't any truer to the original image voxels.) I think this is what [this old TODO comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/9c499409970de40db50db6d49cc572242e348c32/spinalcordtoolbox/reports/slice.py#L264) was referring to.
- The code runs a bit faster (on my laptop, 1.6 seconds, compared to 2.2 seconds for the corresponding `sct_qc` command), possibly because it resamples fewer total voxels (intuitively, it doesn't resample any cropped voxels).

I was also careful to take #5026 into account when adding slice labels.

For ease of comparison, I implemented this POC as an extra option for `sct_qc -p`, named `scratch`. Here is a QC report with three rows:
1. The result of running `sct_register_multimodal -qc` from [this tutorial](https://spinalcordtoolbox.com/stable/user_section/tutorials/multimodal-registration/contrast-agnostic-registration.html).
2. The result of running `sct_qc -p sct_register_multimodal` on the same images. It looks very different, because the code paths in `qc.py` and `qc2.py` have diverged for this report. This is something I want to fix with this POC.
3. The result of running `sct_qc -p scratch` on the same images (this POC). It looks very similar to 1, with the addition of slice labels.

[qc-2025-10-03.zip](https://github.com/user-attachments/files/22688207/qc-2025-10-03.zip)

### Implementation notes

SCT normally does image resampling through `nibabel`, which wraps `scipy.ndimage`. Here, for more control and better performance, I decided to use `scipy.ndimage` directly. (In particular, `nibabel` doesn't expose the crucial `prefilter=False` option for resampling.)

I also decided to use `nibabel`'s utilities for image orientations directly, instead of SCT's `Image` class, because I wanted to manipulate an abstract image FOV (described by just an affine matrix and a shape tuple) without doing costly manipulations of the data array.

### TODO

Before merging, I would:
- make both `sct_qc` and `sct_register_multimodal` use the new code,
- make `sct_register_to_template` use the new code as well,
- remove the two old code paths.

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -5,7 +5,7 @@ Copyright (c) 2023 Polytechnique Montreal <www.neuro.polymtl.ca>
 License: see the file LICENSE
 """
 
-from contextlib import contextmanager
+from contextlib import AbstractContextManager, contextmanager
 import datetime
 from hashlib import md5
 import importlib.resources
@@ -18,7 +18,6 @@ import shutil
 from typing import Optional, Sequence
 
 import numpy as np
-from scipy.ndimage import center_of_mass
 import skimage.exposure
 
 from spinalcordtoolbox.centerline.core import get_centerline, ParamCenterline
@@ -40,7 +39,8 @@ mpl_colors = LazyLoader("mpl_colors", globals(), "matplotlib.colors")
 mpl_backend_agg = LazyLoader("mpl_backend_agg", globals(), "matplotlib.backends.backend_agg")
 mpl_patheffects = LazyLoader("mpl_patheffects", globals(), "matplotlib.patheffects")
 mpl_collections = LazyLoader("mpl_collections", globals(), "matplotlib.collections")
-
+nib_orientations = LazyLoader("nib_orientations", globals(), "nibabel.orientations")
+ndimage = LazyLoader("ndimage", globals(), "scipy.ndimage")
 
 logger = logging.getLogger(__name__)
 
@@ -72,7 +72,7 @@ def create_qc_entry(
     dataset: Optional[str],
     subject: Optional[str],
     image_extension: str = 'png',
-):
+) -> AbstractContextManager[dict[str, Path]]:
     """
     Generate a new QC report entry.
 
@@ -148,6 +148,281 @@ def create_qc_entry(
 
     logger.info('Successfully generated the QC results in %s', str(path_result))
     display_open(file=str(path_index_html), message="To see the results in a browser")
+
+
+def reorient_grid(
+    shape: tuple[int, int, int],  # image data array shape
+    affine: np.ndarray,  # image affine, of shape (4, 4)
+    orientation: str,  # for example, "SAL"
+) -> tuple[
+    tuple[int, int, int],  # reoriented shape
+    np.ndarray,  # reoriented affine, of shape (4, 4)
+]:
+    """
+    Reorient a sampling grid to the given orientation.
+    """
+    opposite = dict(zip("RASLPI", "LPIRAS"))  # SCT and nibabel use opposite conventions
+    ornt = nib_orientations.ornt_transform(
+        nib_orientations.io_orientation(affine),
+        nib_orientations.axcodes2ornt(tuple(opposite[char] for char in orientation))
+    )
+
+    new_shape = [0, 0, 0]
+    for axis, new_axis in enumerate(ornt[:, 0]):
+        new_shape[int(new_axis)] = shape[axis]
+
+    new_affine = np.dot(affine, nib_orientations.inv_ornt_aff(ornt, shape))
+
+    return tuple(new_shape), new_affine
+
+
+def rescale_grid(
+    shape: tuple[int, int, int],  # image data array shape
+    affine: np.ndarray,  # image affine, of shape (4, 4)
+    resolution: list[float | None],  # desired resolution for each axis
+) -> tuple[
+    tuple[int, int, int],  # rescaled shape
+    np.ndarray,  # rescaled affine, of shape (4, 4)
+]:
+    """
+    Rescale a sampling grid to the given resolution for each axis.
+
+    No rescaling is done for an axis if the requested resolution is `None`.
+    """
+    # Make a copy of the input before modifying it
+    shape = list(shape)
+    affine = np.array(affine)
+
+    old_resolution = np.linalg.norm(affine[:, :3], axis=0)
+
+    for axis, (old_res, new_res) in enumerate(zip(old_resolution, resolution)):
+        if new_res is not None:
+            affine[:, axis] *= new_res / old_res
+            # Round up to make sure we cover the original FOV entirely
+            shape[axis] = math.ceil(shape[axis] * old_res / new_res)
+
+    return tuple(shape), affine
+
+
+def get_slices(
+    img: Image,  # 3D image to sample from
+    shape: tuple[int, int],  # output slice shape
+    affine: np.ndarray,  # output affine of shape (4, 4)
+    offsets: np.ndarray,  # list of offsets, of shape (N, 3)
+    order: int,  # interpolation order, can be [0, 1, 2, 3, 4, 5]
+) -> list[np.ndarray]:
+    """
+    Get several slices of the same shape from a single image by resampling.
+
+    The offsets are measured in output voxels, and can be fractional.
+    """
+    # To fill regions outside of the image with zeros,
+    # while avoiding sharp cutoffs around the edges.
+    mode = "grid-constant"
+
+    voxel_to_voxel = np.linalg.inv(img.affine).dot(affine)
+    matrix = voxel_to_voxel[:3, :3]
+    origin = voxel_to_voxel[:3, 3]
+
+    if order > 1:
+        # Doing this prefilter step outside of the loop is crucial for performance
+        data_filtered = ndimage.spline_filter(img.data, order=order, mode=mode)
+    else:
+        data_filtered = img.data.astype(np.float64)
+    return [
+        ndimage.affine_transform(
+            data_filtered,
+            matrix=matrix,
+            offset=(origin + matrix.dot(offset)),
+            output_shape=(1, *shape),
+            order=order,
+            mode=mode,
+            prefilter=False,
+        )[0]
+        for offset in offsets
+    ]
+
+
+def get_axial_slicing_specs(
+    img_input: Image,
+    img_seg: Image,
+    p_resample: float = 0.6,
+    radius: int = 15,
+) -> tuple[
+    tuple[str, str, str, str],  # orientation_labels
+    list[str],  # slice_labels
+    tuple[int, int],  # shape
+    np.ndarray,  # affine, of shape (4, 4)
+    np.ndarray,  # offsets, of shape (N, 3)
+]:
+    """
+    Generate a resampling grid for each axial slice of the input image.
+
+    The input image is first re-oriented to SAL, then each S-I slice is:
+    - resampled isotropically to `p_resample` millimeters,
+    - individually centered around its center of mass in the segmentation,
+    - cropped to a rectangle of `2*radius` voxels.
+
+    The resulting `shape`, `affine`, `offsets` can be passed to `get_slices`
+    to extract the same slices from any nifti image.
+
+    The slices are returned in S-I order (index 0 is Superior), regardless of
+    the original input image orientation, for consistency when generating
+    mosaics. The returned `slice_labels` can be displayed as slice numbers
+    from the original image orientation.
+
+    Returns:
+    1. The appropriate orientation labels for a mosaic.
+    2. The list of slice numbers from the original input image orientation.
+    3. The single slice shape `(2*radius, 2*radius)`.
+    4. The resampling affine, of shape (4, 4).
+    5. The resampling offsets, as a numpy array of shape (N, 3).
+    """
+    # Standardize to SAL with a fixed resolution for each AL slice
+    shape, affine = img_input.data.shape, img_input.affine
+    shape, affine = reorient_grid(shape, affine, "SAL")
+    shape, affine = rescale_grid(shape, affine, [None, p_resample, p_resample])
+
+    # Resample the segmentation to the current voxel grid,
+    # so that we can compute each slice's center of mass
+    nx, ny, nz = shape
+    shape = (ny, nz)
+    offsets = np.array([[x, 0, 0] for x in range(nx)], dtype=np.float64)
+    slices_seg = get_slices(img_seg, shape, affine, offsets, order=1)
+
+    # Recenter each slice around its center of mass, and interpolate missing
+    # slices. Note that the coordinates may be fractional, which is fine.
+    for x, slice_seg in enumerate(slices_seg):
+        offsets[x, 1:3] = ndimage.center_of_mass(slice_seg)
+    inf_nan_fill(offsets[:, 1])
+    inf_nan_fill(offsets[:, 2])
+
+    # Crop to the desired radius
+    shape = (2*radius, 2*radius)
+    offsets[:, 1:3] -= radius
+
+    # The slice labels in the original orientation of the input image
+    slice_labels = [str(x) for x in range(nx)]
+    if "I" in img_input.orientation:
+        slice_labels.reverse()
+
+    orientation_labels = ("A", "P", "L", "R")
+    return orientation_labels, slice_labels, shape, affine, offsets
+
+
+@contextmanager
+def ax_to_save(path: Path) -> 'AbstractContextManager[mpl_axes.Axes]':
+    """
+    Generate a new figure.
+
+    This context manager yields an `mpl_axes.Axes` object, which can be used
+    to add various graphical elements to the figure. When the `with` block
+    exits, the resulting figure is rendered and saved to the path as a png.
+    """
+    # Initialize the figure
+    fig = mpl_figure.Figure(
+        figsize=(TARGET_WIDTH_INCH, TARGET_WIDTH_INCH),
+        dpi=DPI,
+    )
+    mpl_backend_agg.FigureCanvasAgg(fig)
+    ax = fig.add_axes((0, 0, 1, 1))
+    ax.xaxis.set_visible(False)
+    ax.yaxis.set_visible(False)
+
+    # Give ax to the body of the `with` block
+    yield ax
+
+    # Finalize by saving the figure as a png
+    fig.savefig(str(path), format='png', transparent=True)
+
+
+def generate_mosaic(
+    ax: 'mpl_axes.Axes',
+    img: Image,
+    orientation_labels: tuple[str, str, str, str],
+    slice_labels: list[str],
+    shape: tuple[int, int],
+    affine: np.ndarray,
+    offsets: np.ndarray,
+):
+    """
+    Extract slices from an image and draw them as a mosaic, with labels.
+    """
+    # Make a numpy array of the right shape to hold the mosaic
+    num_col = max(math.floor(424 / shape[1]), 1)  # TODO: TARGET_WIDTH_PIXL / scale
+    num_row = math.ceil(len(slice_labels) / num_col)
+    canvas = np.zeros((num_row * shape[0], num_col * shape[1]))
+    patches = [
+        (slice(x, x + shape[0]), slice(y, y + shape[1]))
+        for x in range(0, canvas.shape[0], shape[0])
+        for y in range(0, canvas.shape[1], shape[1])
+    ]
+
+    # Extract image slices and insert them in the mosaic
+    img_slices = get_slices(img, shape, affine, offsets, order=2)
+    for patch, data in zip(patches, img_slices):
+        canvas[patch] = data
+    ax.imshow(equalize_histogram(canvas), cmap='gray', interpolation='none', aspect=1.0)
+
+    # Add orientation labels in the first tile,
+    # and slice labels in the other tiles.
+    text_args = dict(
+        color='yellow',
+        fontsize=4,
+        path_effects=[
+            mpl_patheffects.Stroke(linewidth=1, foreground='black'),
+            mpl_patheffects.Normal(),
+        ],
+    )
+    for i, (patch, label) in enumerate(zip(patches, slice_labels)):
+        top, bottom, _ = patch[0].indices(canvas.shape[0])
+        left, right, _ = patch[1].indices(canvas.shape[1])
+        if i == 0:
+            ax.text((left + right) / 2, top, orientation_labels[0], ha='center', va='top', **text_args)
+            ax.text((left + right) / 2, bottom, orientation_labels[1], ha='center', va='bottom', **text_args)
+            ax.text(left, (top + bottom) / 2, orientation_labels[2], ha='left', va='center', **text_args)
+            ax.text(right, (top + bottom) / 2, orientation_labels[3], ha='right', va='center', **text_args)
+        else:
+            ax.text(left, top, label, ha='left', va='top', **text_args)
+
+
+def scratch(
+    fname_input: str,
+    fname_output: str,
+    fname_seg: str,
+    argv: Sequence[str],
+    path_qc: str,
+    dataset: Optional[str],
+    subject: Optional[str],
+    p_resample: float,
+):
+    """Scratch"""
+    command = 'sct_register_multimodal'
+    cmdline = [command]
+    cmdline.extend(argv)
+
+    # Axial orientation, switch between two input images
+    with create_qc_entry(
+        path_input=Path(fname_input).resolve(),
+        path_qc=Path(path_qc),
+        command=command,
+        cmdline=list2cmdline(cmdline),
+        plane='Axial',
+        dataset=dataset,
+        subject=subject,
+    ) as imgs_to_generate:
+
+        img_input = Image(fname_input)
+        img_output = Image(fname_output)
+        img_seg = Image(fname_seg)
+
+        slicing_specs = get_axial_slicing_specs(img_input, img_seg, p_resample)
+
+        with ax_to_save(imgs_to_generate['path_background_img']) as ax:
+            generate_mosaic(ax, img_input, *slicing_specs)
+
+        with ax_to_save(imgs_to_generate['path_overlay_img']) as ax:
+            generate_mosaic(ax, img_output, *slicing_specs)
 
 
 def add_slice_numbers(ax, num_slices, patch_size, margin: int = 2):
@@ -232,7 +507,7 @@ def sct_register_multimodal(
 
         # Each slice is centered on the segmentation
         logger.info('Find the center of each slice')
-        centers = np.array([center_of_mass(slice) for slice in img_seg.data])
+        centers = np.array([ndimage.center_of_mass(slice) for slice in img_seg.data])
         inf_nan_fill(centers[:, 0])
         inf_nan_fill(centers[:, 1])
 
@@ -378,7 +653,7 @@ def sct_deepseg_axial(
     logger.info('Find the center of each slice')
     # Use the -qc-seg mask if available, otherwise use the spinal cord mask
     img_centers = img_qc_seg if fname_qc_seg else img_seg_sc
-    centers = np.array([center_of_mass(slice) for slice in img_centers.data])
+    centers = np.array([ndimage.center_of_mass(slice) for slice in img_centers.data])
     inf_nan_fill(centers[:, 0])
     inf_nan_fill(centers[:, 1])
 
@@ -482,7 +757,7 @@ def sct_deepseg_spinal_rootlets(
     logger.info('Find the center of each slice')
     centerline_param = ParamCenterline(algo_fitting="optic", contrast="t2")
     img_centerline, _, _, _ = get_centerline(img_input, param=centerline_param)
-    centers = np.array([center_of_mass(slice) for slice in img_centerline.data])
+    centers = np.array([ndimage.center_of_mass(slice) for slice in img_centerline.data])
     inf_nan_fill(centers[:, 0])
     inf_nan_fill(centers[:, 1])
 
@@ -609,7 +884,7 @@ def sct_deepseg_sagittal(
     # Use the -qc-seg mask if available to get crop radius (as well as the center of mass) for each slice
     if fname_qc_seg:
         radius = get_max_sagittal_radius(img_qc_seg)
-        centers = np.array([center_of_mass(slice) for slice in img_qc_seg.data])
+        centers = np.array([ndimage.center_of_mass(slice) for slice in img_qc_seg.data])
         inf_nan_fill(centers[:, 0])
         inf_nan_fill(centers[:, 1])
     # otherwise, if -qc-seg isn't provided, display the full sagittal slice and use the center of the uncropped image
--- a/spinalcordtoolbox/scripts/sct_qc.py
+++ b/spinalcordtoolbox/scripts/sct_qc.py
@@ -40,7 +40,8 @@ def get_parser():
         choices=('sct_propseg', 'sct_deepseg_sc', 'sct_deepseg_gm', 'sct_deepseg_lesion',
                  'sct_register_multimodal', 'sct_register_to_template', 'sct_warp_template',
                  'sct_label_vertebrae', 'sct_detect_pmj', 'sct_label_utils', 'sct_get_centerline',
-                 'sct_fmri_moco', 'sct_dmri_moco', 'sct_image_stitch', 'sct_fmri_compute_tsnr'))
+                 'sct_fmri_moco', 'sct_dmri_moco', 'sct_image_stitch', 'sct_fmri_compute_tsnr',
+                 'scratch'))
 
     optional = parser.optional_arggroup
     optional.add_argument(
@@ -110,20 +111,33 @@ def main(argv: Sequence[str]):
     if arguments.p == 'sct_deepseg_lesion' and arguments.plane is None:
         parser.error('Please provide the plane of the output QC with `-plane`')
 
-    generate_qc(fname_in1=arguments.i,
-                fname_in2=arguments.d,
-                fname_seg=arguments.s,
-                # Internal functions use capitalized strings ('Axial'/'Sagittal')
-                plane=arguments.plane.capitalize() if isinstance(arguments.plane, str) else arguments.plane,
-                args=f'("sct_qc {list2cmdline(argv)}")',
-                path_qc=arguments.qc,
-                dataset=arguments.qc_dataset,
-                subject=arguments.qc_subject,
-                process=arguments.p,
-                fps=arguments.fps,
-                p_resample=arguments.resample,
-                draw_text=bool(arguments.text_labels),
-                path_custom_labels=arguments.custom_labels)
+    if arguments.p == 'scratch':
+        from spinalcordtoolbox.reports.qc2 import scratch
+        scratch(
+            fname_input=arguments.i,
+            fname_output=arguments.d,
+            fname_seg=arguments.s,
+            argv=argv,
+            path_qc=arguments.qc,
+            dataset=arguments.qc_dataset,
+            subject=arguments.qc_subject,
+            p_resample=(0.6 if arguments.resample is None else arguments.resample),
+        )
+    else:
+        generate_qc(fname_in1=arguments.i,
+                    fname_in2=arguments.d,
+                    fname_seg=arguments.s,
+                    # Internal functions use capitalized strings ('Axial'/'Sagittal')
+                    plane=arguments.plane.capitalize() if isinstance(arguments.plane, str) else arguments.plane,
+                    args=f'("sct_qc {list2cmdline(argv)}")',
+                    path_qc=arguments.qc,
+                    dataset=arguments.qc_dataset,
+                    subject=arguments.qc_subject,
+                    process=arguments.p,
+                    fps=arguments.fps,
+                    p_resample=arguments.resample,
+                    draw_text=bool(arguments.text_labels),
+                    path_custom_labels=arguments.custom_labels)
 
 
 if __name__ == "__main__":


mguaypaq:
While looking more carefully at the QC report, I noticed that:
- I incorrectly used `t2_crop_reg.nii.gz` instead of `t2_crop.nii.gz` as the "destination" image for rows 2 and 3 of the sample QC report, which explains why the row 1 image looks much sharper than the other two, has better centering, and more non-black slices.
- We resample each slice to 0.6mm isotropic, then cut out a 30x30 square (that is, 18.0mm), but then this gets rendered to 75x75 pixels in the png, which is a second layer of resampling. Wouldn't it be more accurate to resample just once to 0.24mm, then cut out a 75x75 square?

  But one problem with this is that we expose the `0.6mm` as the default value of the argument `sct_qc -resample`. So, changing this may be backwards incompatible. But because of the double resampling, the current value is kind of a lie.
- In row 3, I have a big margin of transparent pixels on the left and right of the image, because my proposed code uses a square figure, and renders a non-square numpy array into it. This means that the data dependency is:
  1. Figure out the size and number of slices/tiles in the mosaic.
  2. Figure out the number of rows/columns in the mosaic based on this.
  3. Use these dimensions (divided by the DPI) in the `Figure` constructor before calling `Figure.add_axes`.

  This is incompatible with my proposed `ax_to_save` context manager, so I'll have to come up with a different code structure. But I'm not worried about this, compared to the bigger change of structure I'm proposing with the resampling grids.

spinalcordtoolbox/scripts/sct_qc.py
joshuacwnewton: If I'm understanding correctly, I can basically just ignore the changes in this `sct_qc.py` file for now, since all you really did here was make a CLI entry point for the POC?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +150,359 @@ def create_qc_entry(
...
+        img_output = Image(fname_output)
+        img_seg = Image(fname_seg)
+
+        slicing_specs = {
+            'Axial':    get_axial_slicing_specs,
+            'Sagittal': get_sagittal_slicing_specs,
+        }[plane](img_input, img_seg, p_resample)
+
+        generate_mosaic(img_input, *slicing_specs, imgs_to_generate['path_background_img'])
+        generate_mosaic(img_output, *slicing_specs, imgs_to_generate['path_overlay_img'])
```
joshuacwnewton: ## 1. `slicing_spec` is opaque

My initial reservation comes from the opaque nature of what a "slicing spec" actually is? It's not very clear what's actually going on when looking at the highlighted section of the `scratch` function above.

> _`slicing_spec` represents a collection of 5 related properties! 3 properties for image dimensions, and 2 properties for labels. However, most of the documentation for these properties is placed in the "return value" section of the docstrings of each spec function..._

I'm thinking about alternate ways to organize this. For example, it would be nice if there was a single location where each of these 5 properties was defined and validated? (e.g. we would describe `resampling affine`, ensure it's always a (4, 4) numpy array, and make it mandatory to create for anyone making a new QC report).

| Pattern                                                                                                   | Pros/cons                                                                                                                                                           |
|-----------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Simply have 5 separate variables, each passed explicitly to corresponding arguments of `generate_mosaic`. | _(More transparent, but does less to enforce an expected interface between A) "constructing the slicing spec" and B) "using the slicing spec to generate the images".)_ |
| Create a class (e.g. `SlicingSpec`), return `spec` objects from slicing spec functions, and pass `spec` objects directly to `generate_mosaic`.               | _(Enforces a shared interface by providing an place for validation + documentation for each of the 5 attributes (e.g. the class constructor), but has the same "opaque" problem.)_          |
| Create a class (e.g. `SlicingSpec`), but pass `spec.ATTRIBUTE` to each argument of `generate_mosaic`.     | _(More transparent, can still enforce validation/documentation, but does feel a bit strange to create an object only to immediately discard it.)_                   |

My only reservation about creating a class is that I'm wary of basically reinventing what we had with the old `Slice` objects back in `qc.py`. 😆

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +150,359 @@ def create_qc_entry(
...
+        data_filtered = img.data.astype(np.float64)
+    return [
+        ndimage.affine_transform(
+            data_filtered,
+            matrix=matrix,
+            offset=(origin + matrix.dot(offset)),
+            output_shape=(1, *shape),
+            order=order,
+            mode=mode,
+            prefilter=False,
```
joshuacwnewton: ## 2. Separation between `slicing_spec` and `get_slices`

One thing I notice about the current implementation is that there's quite a bit of distance between:

1. The place where the slicing spec is computed.
2. The place where the slicing spec actually gets used (`get_slices`).

In other words: If we were to ask one of our students to create a new QC report, they would need to write code to accomplish 1. while simultaneously holding in their heads how each of the 5 properties would be used for 2. later on (potentially before 2. is even written). It feels like a bit of a chicken-and-egg: we need 1. to write 2., but we need the knowledge of 2. to know how to properly construct 1.?

We could try to close the gap by say, making `get_slices` a method of `SlicingSpec`? I think this makes a bit of sense because `generate_mosaic` only uses `affine` and `offsets` directly in `get_slices` and nowhere else, meaning `get_slices` is kind of already coupled with the slicing spec? So maybe that relationship should be made explicit by grouping everything under a class umbrella.

With this idea, maybe the structure would look like this instead:

1. Construct the slicing spec, passing each property to `SlicingSpec`.
2. Use the `SlicingSpec.get_slices` method to extract slices.
3. Pass the extracted slices to the plotting function (e.g. `generate_mosaic`)

spinalcordtoolbox/reports/qc2.py
mguaypaq: Those are some very valid points.

How would you feel about a [dataclass](https://docs.python.org/3/library/dataclasses.html)? (I think of it like a slightly more modern take on [namedtuples](https://docs.python.org/3/library/collections.html#collections.namedtuple), with better affordances for declaring types and adding validations in `__init__`.)

I'm also very wary of recreating the `Slice` classes, but I think this is different:
- The `Slice` classes used inheritance (which is usually hard to follow), a somewhat weird set of methods, and didn't actually deal with consistent re-orientation and re-sampling of images. So each different way of slicing needed both a new `Slice` subclass, and different calling code to handle it.
- The proposed slicing spec idea is a single class of just immutable data (computed once and then bundled), and it can be used in a uniform way by the mosaic-generating code on several images, regardless of original orientation and resolution.

joshuacwnewton: I was vaguely considering proposing a dataclass in my original comment, in fact! I used them [in a personal project](https://github.com/vivaria/tja2fumen/blob/main/src/tja2fumen/classes.py) and enjoyed the syntatic sugar quite a bit. It makes everything so clean! :blush:

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: I spoke with @mguaypaq on Slack, and we came to the following conclusions:

- People generally won't be writing their own "`get_slicing_sepc`" and "`generate_mosaic`" functions from scratch, and will instead be basing new functions off of the existing examples (tweaking as needed).
- `get_slices()` is used not only in generating the QC mosaics, but also in determining the initial axial slicing spec. Meaning, it can't really be a _method_ of the slicing spec.
- A lot of the difficulty in understanding the patterns can be fixed via better docstrings/comments/etc.

So, I think I'm alright marking this comment as "resolved".

mguaypaq: After interactive discussion elsewhere: thank you for the feedback and design ideas! I will do another round of prototyping, where `SlicingSpec` becomes a documented `dataclass`, and `offsets` becomes an ordered `dict` keyed by `slice_labels`.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -494,13 +532,15 @@ def scratch(
...
-        slicing_specs = {
-            'Axial':    get_axial_slicing_specs,
-            'Sagittal': get_sagittal_slicing_specs,
-        }[plane](img_input, img_seg, p_resample)
+        if plane == 'Axial':
+            slicing_spec = SlicingSpec.full_axial(img_input, p_resample).center_patches(img_seg)
+        elif plane == 'Sagittal':
+            slicing_spec = SlicingSpec.full_sagittal(img_input, p_resample).center_columns(img_seg)
+        else:
+            raise ValueError(f"Invalid {plane=}")
```
joshuacwnewton: Woah woah woah. I've never seen this syntactic sugar! This is so neat. :hearts:

```python
>>> plane = "Axail"
>>> print(f"Invalid {plane=}")
# Invalid plane='Axail'
```

spinalcordtoolbox/reports/qc2.py
```diff
@@ -204,263 +205,300 @@ def rescale_grid(
...
-            prefilter=False,
-        )[0]
-        for offset in offsets
-    ]
+        The first column gives the step size and direction (in physical space)
+        for taking a 1-voxel step from i to i+1. The second and third columns
+        are the same for j to j+1, or k to k+1. The last column gives the
+        physical coordinates of the (0, 0, 0) voxel. The last row is fixed.
 
+        The 2D slices to be extracted are all parallel to the (j, k) plane.
```
joshuacwnewton, line 231: I'm sure there's a clear reason for this, but "`parallel to the (j, k) plane`" is a bit confusing to me just because I'm used to slices corresponding to `[:, :, n]` (i.e. SI being the last axis).

I think we talked about this and decided on `[n, :, :]` convention here, but I can't recall why? I'll have to come back to this point when I read the PR more.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -204,263 +205,300 @@ def rescale_grid(
...
+    def full_axial(img: Image, p_resample: float = 0.6) -> 'SlicingSpec':
+        """
+        A slicing spec to extract full axial slices.
+
+        The field of view is given by `img`.
+        The number of slices is the same as `img` in the S-I axis,
+        but the other two axes are resampled to `p_resample` mm.
+        The slice labels correspond to the original orientation of `img`,
+        but the output slices are always ordered and oriented in "SAL".
+        """
```
joshuacwnewton, line 284: Given how similar the `full_axial` and `full_sagittal` methods are, I could envision a world where we merge them into a `full_slices` method with a `plane=('Axial', 'Sagittal')` parameter?

The only differences are:

- `'SAL'` vs. `'RSP'`
- `'I'` vs. `'L'`
- `("A", "P"), ("L", "R")` vs. `("S", "I"), ("P", "A")`

I could even imagine a world where the latter two properties are derived from the first?

- Take `orientation[0]` and flip it (e.g. SAL = S -> I)
- Take `orientation[1]` and `orientation[2]` and get the associated pairs (e.g. SAL = A,L -> AP,LR)

Meaning the only plane-specific logic we'd need in the method is `{"Axial": "SAL", "Sagittal": "RSP"}`? I kind of like this because the act of deriving the properties kind of encodes _why_ we're using those specific strings in the first place. But at the same time, maybe this is "too cute" just to save some hardcoded strings. :stuck_out_tongue_closed_eyes: 

Still, I think even with hardcoded strings, I think it would save a little bit of code.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -204,263 +205,300 @@ def rescale_grid(
...
+            affine=self.affine,
+            offsets=offsets_dict,  # new offsets
+            shape=shape,  # new shape
+            axis_labels=self.axis_labels,
+        )
+
+    def generate_mosaic(self, img: Image, path: Path, scale: float = 2.5):
+        """
+        Extract slices from an image and save the result as a png.
+        """
```
joshuacwnewton: Do we want to add some input validation here to ensure that `slicing_spec` has its properties set before proceeding?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -204,263 +205,300 @@ def rescale_grid(
...
+                label: np.array([i, 0, 0], dtype=np.float64)
+                for i, label in enumerate(slice_labels)
+            },
+            shape=shape,
+            axis_labels=(("S", "I"), ("P", "A")),
+        )
+
+    def center_columns(self, seg: Image) -> 'SlicingSpec':
+        """
+        A re-centered and cropped slicing spec, meant for sagittal views.
```
joshuacwnewton: It looks like the `center_` methods also seem to have a requirement that a `full_` method needs to have been run first, so maybe we could add validation for that too?

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Actually, because the `center_*` and `generate_mosaic` methods are both dependent on running a `full_` method first, maybe we could make `full_` the de facto `__init__` method for the class, such that in order to even initialize the SlicingSpec class, an image + `p_resample` + `plane` must be passed first? 

That way, all of the other methods will be guaranteed to work for free (without needing error checking as I mention in my other comments).

joshuacwnewton (changes requested):
I really really love this change. I only had some minor organizational ideas, but the core of it feels fantastic.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Well, I guess that latter idea is more of a "use a regular, mutable class instead of a frozen dataclass" suggestion... hrm.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Yep! It's intended for easier debugging and logging messages. Also, if you prefer a different spacing around the 'equals' sign, I believe Python uses whatever sequence of spaces and tabs are to the left and to the right of the 'equals' sign in the format string.

```py3
>>> plane = "Axail"
>>> f"Invalid {plane  = }"
"Invalid plane  = 'Axail'"
>>> f"Invalid {plane    =}"
"Invalid plane\t='Axail'"
```

spinalcordtoolbox/reports/qc2.py
mguaypaq: I think most of SCT uses RPI as the standard orientation, so that S-I slices are `[:, :, n]`. But historically, QC reports use SAL (for axial), so that S-I slices are instead `[n, :, :]`.

Here I just decided to follow the previous QC report convention, but we could change it.

Edit: also, the default when iterating over a numpy 3D array is to produce `[n, :, :]` slices, like the historical QC convention. But I view this justification as very thin, more like a tie-breaker if there are no other considerations.

spinalcordtoolbox/reports/qc2.py
mguaypaq: I'll note that all the properties of `SlicingSpec` are required arguments to its `__init__` method, because none of them has a default value assigned at the class level. So, this is already somewhat validated.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Re: the similarity of `full_axial` and `full_sagittal`, it's true that there's a very natural generalization that covers both cases. My only worry is about clearly documenting the general case, but it should be doable.

Re: `full_*` and `center_*` and `__init__`, a few points:
- I should put this in the docstring, but I consider `__init__` to be a kind of internal method: it has all the `SlicingSpec` fields as required arguments. This is somewhat unavoidable, because there should generally be _some_ fast way to create instances of any class (for example, for copy constructors, or for internal methods).
- Note that the `full_*` methods are static methods (they don't have a `self`), so they're meant as user-friendly factory functions, and will create a `SlicingSpec` instance from scratch (rather than modifying an existing one).
- In contrast the `center_*` methods _do_ have a `self`, so the (admittedly pretty basic) validation of `self` is already done for free. They then return an adjusted `SlicingSpec` as a fresh instance just so we don't have to worry about mutability.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: Ahhh, of course, good point.

I guess what I really meant instead was a way to distinguish between "a method that you only run once you've already initialized a SlicingSpec" and "a method used to a return a SlicingSpec that has been initialized" beyond just the presence of `self`.

Since the `full_` methods are ones that can be run straight from `SlicingSpec`, it might follow to the reader that the `center_` methods are also like that (but they aren't AFAIU, and will presumably hit some sort of error if you try to run them without initializing the spec first?).

mguaypaq: Yes, since the `center_*` methods are not `@staticmethod`s, really the only way to call them is to already have a valid `self`.

I should make the distinction more clear in the docstrings between the two factory functions `full_*` and the two transformer methods `center_*`.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,432 @@ def create_qc_entry(
...
+
+        Used for the top-left tile in a mosaic.
+    """
+    # ----------------------------------------------------------------
+    # Mandatory fields
+    # ----------------------------------------------------------------
+    # This is enforced by the default @dataclass __init__ method,
+    # which should be considered an "internal" method. Instead of
+    # creating instances directly with SlicingSpec(...), users of this
+    # class should use one of the factory functions below.
```
joshuacwnewton: Tiny nit, but calling this section "mandatory fields" then going on to say "don't call SlicingSpec(), use a factory function" might be a bit confusing?

It might be better to explicitly discourage any notion of mandatory/requirement by saying something like "There is no need to modify these fields directly. Instead, these fields will be automatically set when calling one of the factory functions below"

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,432 @@ def create_qc_entry(
...
+    affine: np.ndarray  # shape=(4, 4)
+    offsets: dict[str, np.ndarray]  # each shape=(3,), insertion order is important
+    shape: tuple[int, int]  # 2D shape of all output slices
+    axis_labels: tuple[tuple[str, str], tuple[str, str]]  # ((top, bottom), (left, right))
+
+    # ----------------------------------------------------------------
+    # Factory functions
+    # ----------------------------------------------------------------
+    # This is the normal way for users of this class to create
+    # instances, instead of calling SlicingSpec(...) directly.
```
joshuacwnewton: Ditto as above -- rather than highlighting what not to do, it might be better to just jump directly into describing the expected usage and nix the possibility of a direct call from readers' minds. 

Something like "These factory functions are used to create new SlicingSpec objects, using syntax such as:
slicing_spec = SlicingSpec.full_axial(...)"

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,432 @@ def create_qc_entry(
...
+    # After using a factory function, the transformer methods below
+    # can be used to adjust the resulting SlicingSpec.
```
joshuacwnewton: This is a useful note but "below" might be a bit confusing/ambiguous since the factory functions are also "below"? It might be better to just remove this note, and allow users to discover the transformer methods by scrolling down a bit more.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,432 @@ def create_qc_entry(
...
+            axis_labels=tuple(
+                (letter, opposite[letter])
+                for letter in orientation[1:3]
+            ),
+        )
+
+    # ----------------------------------------------------------------
+    # Transformer methods
+    # ----------------------------------------------------------------
+    # These methods take an existing SlicingSpec, and produce a new
```
joshuacwnewton: "Take an existing SlicingSpec" makes me think you would pass a SlicingSpec as a parameter, but you don't have to provide the object -- instead you call the method on the already-initialized object.

It might be better to say something like "These methods can only be called after the spec has already been initialized. For example:
spec = SlicingSpec.full_axial(...)  # initialize the spec
spec = spec.center_patches(...)  # transform the spec

joshuacwnewton (dismissed):
I had only some nitpicky comments for clarity, but here is my explicit approval for the pattern itself! :tada:

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+        # Take a single mid-sagittal line from each axial slice to compose a 2D image.
+        slicing_spec = SlicingSpec.full_axial(img_input, p_resample).center_lines(img_seg)
+
+        # Quadratic resampling for the actual image.
+        slices_input = slicing_spec.get_slices(img_input, order=2)
+        data_input = equalize_histogram(np.array([s[:, 0] for s in slices_input.values()]))
+
+        # Nearest-neighbour resampling for the segmentation labels.
+        slices_labels = slicing_spec.get_slices(img_labels, order=0)
+        data_labels = np.array([s[:, 0] for s in slices_labels.values()])
```
joshuacwnewton, line 896: I'm noticing that the generated QC report for `sct_label_utils` seems to be higher resolution / is inventing voxels that aren't there.

For example, on this `(3, 10, 2)` generated fake lesion:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/98161abf01be1d552e7e36ae1786fc8a8bdbe90b/testing/cli/test_cli_sct_analyze_lesion.py#L171

`master` is `3` voxels wide:

Image

`mgp/qc-poc` is `5` voxels wide:

Image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+        ]:
+            mosaic = slicing_spec.get_mosaic()
+            mosaic.insert_slices(slicing_spec.get_slices(img, order=2))
+            mosaic.canvas = equalize_histogram(mosaic.canvas)
+            mosaic.ax.imshow(mosaic.canvas, cmap='gray', interpolation='none')
+            mosaic.add_labels()
+            mosaic.save(path)
+
+
+def sct_fmri_compute_tsnr(
```
joshuacwnewton, line 649: tSNR report looks great!

Image

Image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+        ]:
+            mosaic = slicing_spec.get_mosaic(rect=(0, 0, 0.93, 1))
+            mosaic.insert_slices(slices)
+            axes_image = mosaic.ax.imshow(
+                mosaic.canvas,
+                cmap='seismic',
+                norm=mpl_colors.Normalize(vmin, vmax),
+                interpolation='none',
+            )
+            colorbar = mosaic.fig.colorbar(
```
joshuacwnewton, line 710: Nit: Is it possible to shift the position of the colorbar slightly rightwards? It's a tad squished up against the mosaic:

Image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+        # A version of img_labels with only 0-1 values, for center-of-mass computations.
+        img_seg = img_labels.copy()
+        img_seg.data = (img_seg.data != 0)
+
+        # Take a single mid-sagittal line from each axial slice to compose a 2D image.
+        slicing_spec = SlicingSpec.full_axial(img_input, p_resample).center_lines(img_seg)
+
+        # Quadratic resampling for the actual image.
+        slices_input = slicing_spec.get_slices(img_input, order=2)
+        data_input = equalize_histogram(np.array([s[:, 0] for s in slices_input.values()]))
```
joshuacwnewton, line 892: Something wild seems to have happened with the histogram equalization for one of the `sct_label_utils` tests:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ae918a7ed5f27c167186d2d975d1370ea3925e9e/testing/cli/test_cli_sct_label_utils.py#L26-L31

It's possible the test on `master` was actually faulty as well, or the input data is bad, but the change is striking:

`master`:

Image

`mgp/qc-poc`:

Image

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: This behavior is demonstrated very clearly for the `sct_label_utils` test, which I believe shows interpolation errors due to the quadratic resampling?

Image

The previous report was just a binary segmentation:

Image

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+
+        # Nearest-neighbour resampling for the segmentation labels.
+        slices_labels = slicing_spec.get_slices(img_labels, order=0)
+        data_labels = np.array([s[:, 0] for s in slices_labels.values()])
```
joshuacwnewton, line 772: I believe the same sampling issues affect `sct_label_vertebrae` as well:

Image

That said, the `center_lines` slicing spec method seems to work like a charm apart from that!

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+    fname_input: str,
+    fname_output: str,
+    fname_seg: str,
+    command: str,
+    argv: Sequence[str],
+    path_qc: str,
+    dataset: str | None,
+    subject: str | None,
+    p_resample: float | None = 0.6,
+):
```
joshuacwnewton, line 612: Wow! Not only has `qc2.sct_register` been greatly simplified, but the output QC report is much more legible, too!

`master`:

Image

`mgp/qc-poc`:

Image

joshuacwnewton: My one (very minor) complaint is that, because `sct_warp_template` hasn't been ported yet, the "flow" of the course is slightly awkward since you go from the "zoomed in" registration results to the "small grid" warped template results.

Not a big deal though, I think, since it'll be cleared up once we finish porting the remaining reports.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -553,36 +553,59 @@ def main(argv: Sequence[str]):
...
+                    fname_input=fname_input,
+                    fname_seg=fname_discs,
+                    **common_args,
+                )
+                qc2.sct_label_vertebrae(
+                    fname_input=fname_input,
+                    fname_seg=fname_all,
+                    path_custom_labels=path_custom_labels,
+                    **common_args,
+                )
```
joshuacwnewton, line 582: I'm noticing that these back-to-back calls are running into a bug both on this branch and on master:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4984

Adding microseconds ([`%f`](https://strftime.org/)) to this line does fix the issue:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8e7709f521dbaebf1f8b666faf4a2a70896700ca/spinalcordtoolbox/reports/qc2.py#L138

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+                example = Path(__sct_dir__) / 'spinalcordtoolbox' / 'reports' / 'sct_label_vertebrae_regions.json'
+                raise ValueError(f"Invalid format for custom labels, see {example} for an example. ({e})")
+
+            # Add the text labels
+            for label_num in labels:
+                if label_num in dict_labels:
+                    # NB: We need to subtract `min` to convert the label value into an index for the color list
+                    label_color = color_list[label_num - labels.min()]
+                    # Position the label text
+                    y, x = ndimage.center_of_mass(img == label_num)
```
joshuacwnewton, line 840: I'm noticing that the center of mass-based method can lead to some not-so-ideal locations for totalspineseg:

Image

Image

Some brainstorming on how to improve this:

- I wonder if we could do some [`skimage.measure.label`](https://scikit-image.org/docs/0.25.x/api/skimage.measure.html#skimage.measure.label) shenanigans to apply the label to each sub-region within the label?
- I'm kind of second-guessing the decision to offset the labels from the center of masses -- I think this was originally intended for `sct_label_vertebrae` with the small, vertically-progressing regions, but now we're working with larger regions that exist horizontally, and so the offset kinda just makes the labels look misaligned. I'm wondering if centered labels would be better now?
- I'm wondering if we could try to prevent label overlap by keeping track of the centers of mass used for each label, then offsetting e.g. the y position to ensure that no labels are within X amount of voxels from each other?

spinalcordtoolbox/reports/qc2.py
```diff
@@ -150,6 +151,777 @@ def create_qc_entry(
...
+            shape=shape,  # new shape
+            axis_labels=self.axis_labels,
+        )
+
+    def center_lines(self: 'SlicingSpec', seg: Image) -> 'SlicingSpec':
+        """
+        A re-centered and cropped slicing spec, meant to transform a full axial
+        slicing spec into a collection of single-line slices which can be
+        arranged into a single wavy mid-sagittal image.
+        """
```
joshuacwnewton: Nit, but the name and description come across to me as a little difficult to grok. (Specifically, the notion of `center_lines` != `centerlines`, and the terms "recentered and cropped" as well as "wavy".)

I absolutely get what you mean by the name + terms, but I wonder if something like this might be a little more intuitive?

```suggestion
    def composite_sagittal(self: 'SlicingSpec', seg: Image) -> 'SlicingSpec':
        """
        Construct a composite 2D sagittal image out of axial slices (one 
        row from each axial slice), similarly to `sct_flatten_sagittal`.

        (For each axial slice, the center of mass is used to identify the 
        midsagittal row. The individual rows can then be stitched together 
        to build a composite midsagittal 2D image.)
        """
```

spinalcordtoolbox/reports/totalspineseg_step1_regions.json
joshuacwnewton, line 1: Thanks for following up on this older request of mine!

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4921#pullrequestreview-2900943794

spinalcordtoolbox/reports/qc2.py
mguaypaq: Ah, you're right! `sct_label_vertebrae` and `sct_label_utils` both had `p_resample_default = None` previously, and this got lost in the refactor. I'll add it back.

spinalcordtoolbox/reports/qc2.py
mguaypaq: It's also weird that the scale started at 0, and now at -1. I'm not sure why that changed.

spinalcordtoolbox/reports/qc2.py
mguaypaq: These are great points, but I think out of scope for right now. I'd be willing to add a temporary toggle to the code so that `sct_deepseg spine` doesn't offset the labels from the centers of mass.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Yes, I expect to port the remaining QC reports after the course.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Should be fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5050/commits/cfe872de9eed8bde122c9f339516b8f9d6184a0e.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Should also be fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5050/commits/cfe872de9eed8bde122c9f339516b8f9d6184a0e.

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: That makes sense! Maybe I can sneak in a quick follow-up PR then, maybe updating v7.2 post-course (so that there isn't any additional pressure on the pre-course release).

(I was mainly just thinking that it might be a bad look to have mis-aligned labels until v7.3 next April, but there's no reason it needs to be merged specifically for the course.)

spinalcordtoolbox/scripts/sct_deepseg.py
joshuacwnewton: If we want to hackily fix this within scope without making this timestamp modification, I think just adding a 1s sleep would also fix this. :stuck_out_tongue_closed_eyes:

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: LGTM! 

image

joshuacwnewton: Ah, though there is a strange darkening at the top and bottom of the segmentation?

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: I checked this test again post-resampling fix, and the results are equally as curious.

image

What's especially interesting to me is that the call to `qc2` shouldn't be significantly different than https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5050#discussion_r2589628835 (same input image), so I don't know why even the `master` test would be different. Maybe `create` mutates the image?

I'll try to dig into this, it might be out of scope for this PR though.

joshuacwnewton: Ah, of course. `sct_label_utils` uses `center_lines` based on the labels:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/cfe872de9eed8bde122c9f339516b8f9d6184a0e/spinalcordtoolbox/reports/qc2.py#L871-L878

And, since we create the labels 1+2 in the nonsensical coordinates 1,1,1 and 2,2,2, the midsagittal slices that get selected are outside of the segmentation, hence no segmentation is displayed.

Changing the coordinates to something more sensible causes the QC report to get fixed as well.

image

mguaypaq: I think we can ignore this.

The problem is that the "centerline" for this composite mid-sagittal image is computed from those two labels at voxel positions (1, 1, 1) and (2, 2, 2), and nothing else. For more "normal" labels, I think the new and old implementations should match more closely.

(Also, this was affected by the doubling of the two label voxels due to the accidental resampling, fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/cfe872de9eed8bde122c9f339516b8f9d6184a0e.)

spinalcordtoolbox/reports/qc2.py
mguaypaq: Fixed in 1e9d06026b704e116b7f615bce56783ac6728121 and c5b0b901116fae7e587fb73a29c5fc50b87a1558. It turns out that neither version had correct min/max values for the colorbar!

spinalcordtoolbox/scripts/sct_deepseg.py
mguaypaq: Fixed in a88ae02ec7d0a0dd81bd8aa6badaa37279a36c05.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Fixed in d3c4b574e1d7ad2965ae700cf896b0fb298294d0.

spinalcordtoolbox/reports/qc2.py
mguaypaq: Composite sagittal is a great name!! Fixed in 66f44fedd1133f474e1212a1e22cbed9f28d7e37.

jcohenadad:
tested b575e12ccd85243e8760a9653f5650e0c6ac5f62

few remarks:
- the sagittal slice is not the same between the two qc outputs
- the disc segmentation is masked by the "Cx-Cy" letters-- I suggest moving these letters such that the "y" is located at a fixed distance from the left most edge of the disc
- the colormap is counter intuitive: why choosing the same colors for [c2 and c6], and [c3 and c5]?
- there is a visible discontinuity on the spinal cord segmentation around C1 level, likely caused by a weird right-left displacement of the slide

ezgif-3fc4956e978a6268

If these are too cumbersome to change now, I suggest opening an issue for subsequent releases

joshuacwnewton:
Just to expand on this point specifically:

> The colormap is currently auto-generated.

The logic I wrote tries to assign related colors (blues, greens, reds) to adjacent groups of labels (30,31,32..., 50,51,52... Etc.). I think I assumed that there would only be a few distinct groups of labels, so I only included a few color groups. 

However, for TSS, the various vertebral groups (cervical, thoracic, lumbar) are separated into distinct label groups. Meaning, the available color sets get exhausted too quickly, and the colors loop back around to the already-used 'reds' (which was done to avoid index errors). 

I think we can fix this temporarily for now by simply adding more color groups to the list of possible sets, so that red avoids getting reused for vertebrae.

sandrinebedard:
I tested the qc for `sct_label_utils`, `sct_register_to_template`, `sct_label_vertebrae`, and `sct_fmri_compute_tsnr `, they all looked fined except for `sct_fmri_compute_tsnr `

The range of the color map is very different then before (teh max is really high, so it isn't very informative) (0-61) vs (0-26)

**New qc:**
image

**Old qc:** 
image

spinalcordtoolbox/reports/qc2.py
mguaypaq: I think the darkening may be a side-effect of quadratic resampling right next to the edge, because the outside is considered filled with zeros.

This could be fixed with linear resampling, at least for binary segmentations. But images tend to look noticeably sharper with quadratic resampling rather than linear.

We could also change the resampling mode to consider something other than zero-filling the outside, but that would risk hallucinating the wrong data in cases where the FOVs are not aligned.

joshuacwnewton: That makes sense! 

I don't think it's a huge deal, since the main focus is the labels rather than the underlying image, but maybe we could mention it in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5113 as a follow-up just in case.

sandrinebedard:
> I notice the lower end of the tSNR range is `2`. I'm curious, did the previous tSNR QC fix `0` as a lower bound, and if so, is this something we want to keep?
> 
> For example, if there were some data where the whole image has high tSNR (not sure if this is even possible), then the range could be artificially lifted and indicate "low" tSNR (dark blue) that isn't actually low.

it was of 0 before, I don't mind if the minium is not of 0, it helps with visualisation sometimes, but between 2 and 0, doesn't make much of a difference

Issue #5063, opened 2025-10-21
Wrong labeled segmentation on axial image


sandrinebedard:
### Description

When running `sct_label_vertebrae -discfile`, with an axial image and point-labels of the discs, the labeled segmentation looks like this:

<img width="377" height="368" alt="Image" src="https://github.com/user-attachments/assets/9ccf12bd-e142-44a7-be70-88b1e41514b2" />

#### With SCTv7.0:

<img width="377" height="368" alt="Image" src="https://github.com/user-attachments/assets/ed809a4f-34d4-45e2-a647-cf16cd20b113" />

The disc labels look okay:

<img width="377" height="368" alt="Image" src="https://github.com/user-attachments/assets/1025363b-4fc3-475e-be89-0f175566be6e" />

### Steps to Reproduce

0. Install SCT git-master-919731dbdc9c5582e4041f49aaf7a3b0be752059
2. Download images `sub-024_ses-M0_acq-axial_T2w.nii.gz` and sub-024_ses-M0_acq-axial_T2w_label-SC_mask-manual.nii.gzfrom `dcm-zurich` internal dataset
3. Download discs labels: 
[sub-024_ses-M0_acq-sagittal_T2w_labels-manual_reg.nii.gz](https://github.com/user-attachments/files/23028927/sub-024_ses-M0_acq-sagittal_T2w_labels-manual_reg.nii.gz)

4. Run `sct_label_vertebrae -discfile`

```
 sct_label_vertebrae -i sub-024_ses-M0_acq-axial_T2w.nii.gz -discfile sub-024_ses-M0_acq-sagittal_T2w_labels-manual_reg.nii.gz -s sub-024_ses-M0_acq-axial_T2w_label-SC_mask-manual.nii.gz -c t2
```

PR #5067, opened 2025-10-24
Solve projection anomaly on anisotropic data

### LINKED ISSUE #5004: How does `project_centerline` handle voxel origins for thick slices in anisotropic images?
_Originally posted by @jcohenadad in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4948#issuecomment-3073905765_

> @NathanMolinier In the code that looks for the closest point between the disc and the centerline, how is the origin of the disc label defined? There are several conventions for defining the origin of a voxel, eg: its geometric center, or one of its corner. This could have dramatic impact in the case of anisotropic resolution, as is usually the case for qMRI data. See illustration below:
> 
> image

_Originally posted by @NathanMolinier in https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4948#issuecomment-3149708043_

> I'm not sure I understand how to use different voxel conventions, because I only have access to a single coordinate when computing the distance. Here is the [code](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4647d04e6b8c28c59c84287c0b01f1c98113f99a/spinalcordtoolbox/labels.py#L510) if you want a closer look.

NathanMolinier:
## Checklist

## Description

This PR fixes an issue where label projection was incorrect on anisotropic data. The problem occurred because the projection was performed in pixel coordinates instead of real-world space. The greater the anisotropy, the worse the projection became. The fix ensures that projections now correctly account for the image resolution, making them accurate regardless of anisotropy.

ezgif-1f2a8071ab8f8b

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5063
Fixes #5004

### DIFF
--- a/spinalcordtoolbox/process_seg.py
+++ b/spinalcordtoolbox/process_seg.py
@@ -89,9 +89,8 @@ def compute_shape(segmentation, angle_correction=True, centerline_path=None, par
         missing_slices = sorted(set(range(min_z_index, max_z_index + 1)).difference(deriv.keys()))
         if missing_slices:
             raise ValueError(
-                "The provided angle correction centerline does not cover slice(s) "
-                f"{parse_num_list_inv(missing_slices)} of the input mask. Please "
-                "supply a more extensive '-angle-corr-centerline', or disable angle "
+                f"The provided centerline does not cover slice(s) {parse_num_list_inv(missing_slices)} "
+                "of the input mask. Please supply a '-centerline' covering all the slices, or disable angle "
                 "correction ('-angle-corr 0')."
             ) from None
 


spinalcordtoolbox/labels.py
```diff
@@ -235,9 +235,13 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
 
     out = zeros_like(img)
 
+    # Fetch resolution
+    nx, ny, nz, nt, px, py, pz, pt = img.dim
+
     # Extract centerline from segmentation
     _, arr_ctl, _, _ = get_centerline(img)
     centerline_arr = arr_ctl.T
+    centerline_real = centerline_arr*np.array([px, py, pz])
```
joshuacwnewton: While this does account for anisotropic zooms, I am wondering if multiplying by the pixdim values directly to get the "real centerline" is the right approach?

I am thinking of the `space='phys'` option in the  `get_centerline` function, and how internally it uses the `transfo_pix2phys` method of `Image` objects, which accounts for the entire affine (including rotations in the 3x3 transformation matrix within the 4x4 affine).

So, I am curious how this newly-added code would work for an image with rotations in its affine. Would it still project correctly? Or do we need to use both `space='phys'` + `transfo_pix2phys` to get the "correct" real coordinates?

NathanMolinier: What do you mean by rotation in its affine ? I think as long as we ensure that all images are in the same space we should not worry about other transformation.

NathanMolinier: And from my experience checking the orientation and if the images have the same shape is generally enough

joshuacwnewton: > What do you mean by rotation in its affine ?

([Reference](https://nipy.org/nibabel/coordinate_systems.html#the-affine-matrix-as-a-transformation-between-spaces))

(Assuming no reorientation i.e. no affine axis swapping for illustrative purposes) 

image

For an image with no rotation, `[px, py, pz]` will be equivalent to `[m11, m22, m33]`, and the other matrix elements will be `0`:

image

So, your method of multiplication by `px`/`py`/`pz` works to produce accurate physical coordinates.

But if there is rotation, then the other matrix elements will be non-zero, and thus multiplying by `px`/`py`/`pz` won't accurately translate various points into physical space.

I don't know how impactful this actually is though. You may be right that it has no bearing on projection... since the distance between points theoretically wouldn't change if the coordinate system is rotated? (So the "find closest point" projection logic would still return the same points, maybe?)

But still, I would maybe want to find an image that has significant rotation in its affine though, just to check that the projection is performed correctly.

spinalcordtoolbox/labels.py
joshuacwnewton, line 1: As an aside, I just remembered this issue:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5004

Does the investigation in this PR sufficiently address the question? 

(Since we now perform the `project_point_on_line` step in the (partially) physical coordinate space, the question of voxel origin would theoretically be moot, right?)

spinalcordtoolbox/labels.py
NathanMolinier: I see your point, and I agree that we might not be obtaining the true physical coordinates. However, incorporating the resolution still helps to weight the contribution of each coordinate. Which is better than what we currently have.

spinalcordtoolbox/labels.py
NathanMolinier: I think it addresses the question since we now use physical coordinates

sandrinebedard (commented):
I tested on the image I mentionned in #5063 

and it works: 
image


I also notice that the delimitation is always above the slice of the label, maybe this is something we can document

valosekj:
I tested this PR while reviewing https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5075, so I'm copy-pasting it here as well.

Originally posted in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5075#issuecomment-3460212253:

---

> Just a note: the slices are actually a bit off due to a bug which is being fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5067.
> 
> ```
>  sct_process_segmentation -i sub-008_ses-M0_acq-axial_T2w_label-SC_mask-manual.nii.gz -perslice 1 -perlevel 1 -o ./SC.csv -vert 2:9 -discfile sub-008_ses-M0_acq-axial_T2w_labels-manual.nii.gz
> ```
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5075 (from `master`): 
> 
> image
> 
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5067:
> 
> image
> 
> https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5067 is much more plausible as C3 starts at z-slice=21 (instead of z-slice=16 for this PR (and `master`!))
> 
> image

mguaypaq:
@joshuacwnewton Here's a different implementation of `project_centerline` which:
* Computes everything in physical space, so that images are automatically aligned properly, and so that distances are correct.
* Should be more efficient, by using `scipy.spatial.KDTree` (also used by `types.py::Centerline`).

Could you have a look at the code and see if you agree with it? I didn't have time to do `label_regions_from_reference` today, but I would do essentially the same thing if you're ok with it.

spinalcordtoolbox/labels.py
```diff
@@ -524,57 +525,28 @@ def project_centerline(img: Image, ref: Image) -> Image:
...
+    labels_phys = ref.transfo_pix2phys(labels_pix, mode='absolute')  # shape (L, 3), float
 
-    # Checking input dimensions
-    if img.data.shape != ref.data.shape:
-        raise ShapeMismatchError(
-            "Input image and referenced labels should have the same dimension",
-            img=img.data.shape,
-            ref=ref.data.shape)
+    # For each label voxel, the centerline index which is closest in physical space.
+    _, indices = KDTree(centerline_phys).query(labels_phys)  # shape (L,), int
```
joshuacwnewton, line 525: Nice! Seems like [`scipy.spatial.KDTree`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html) is exactly the tool for the job:

> kd-tree for quick nearest-neighbor lookup.
> 
> This class provides an index into a set of k-dimensional points which can be used to rapidly look up the nearest neighbors of any point.

Maybe it's worth linking to something like https://www.colorado.edu/amath/sites/default/files/attached-files/k-d_trees_and_knn_searches.pdf, which compares the naive approach (our old code) to the new KDTree method, which readers might not be familiar with?

spinalcordtoolbox/labels.py
```diff
@@ -524,57 +525,28 @@ def project_centerline(img: Image, ref: Image) -> Image:
...
+    projected_pix = img.transfo_phys2pix(projected_phys)  # shape (L, 3), int
+    projected_pix = projected_pix.clip(0, np.array(img.data.shape) - 1)
 
-    # Create the output image
-    out = zeros_like(ref)
-
-    # Compute the shortest distance for each referenced points on the centerline
-    for x, y, z, value in coordinates_ref:
-        x_real, y_real, z_real = x * px, y * py, z * pz
-        projection_real = project_point_on_line(point=np.array([x_real, y_real, z_real]), line=centerline_real)
```
joshuacwnewton: Presumably once this same logic is applied to `label_regions_from_reference`, there will be no need for the old `project_point_on_line` function?

spinalcordtoolbox/labels.py
```diff
@@ -524,57 +525,28 @@ def project_centerline(img: Image, ref: Image) -> Image:
...
-    # Checking orientation
-    og_img_orientation = img.orientation
-    if img.orientation != "RPI":
-        img.change_orientation("RPI")
-
-    og_ref_orientation = ref.orientation
-    if ref.orientation != "RPI":
-        ref.change_orientation("RPI")
+    # The label coordinates, in the same physical space.
+    labels_pix = np.argwhere(ref.data)  # shape (L, 3), int
```
joshuacwnewton, line 521: Seeing this line here makes me question the usefulness of `ref.getNonZeroCoordinates` :rofl:

spinalcordtoolbox/labels.py
joshuacwnewton, line 1: Given that this PR directly affects PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5075 (as mentioned by @valosekj), I figured it would be a good place to test things? So, I made a quick branch where I replayed the changes from that PR onto this PR's commits: [`jn/test-projection-fix-with-ascor`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/compare/master...jn/test-projection-fix-with-ascor)

On that branch, the projected centerline + aSCOR values didn't change for the (isotropic) test data, so that's at least a very quick sanity check that the new logic works. I'd have to fetch the anisotropic data and test that too to be sure, but as far as I can tell, this logic looks perfectly OK to apply to the other function as well. :)

spinalcordtoolbox/labels.py
mguaypaq: Thanks for the link, I added it in b44c53c4732417c90f95d2b91d2e590725b6b9ea.

spinalcordtoolbox/labels.py
mguaypaq: Removed in b44c53c4732417c90f95d2b91d2e590725b6b9ea.

spinalcordtoolbox/labels.py
mguaypaq: Haha yeah, sometimes I scratch my head at how we selectively use `numpy`. It has tons of nice utilities and compact idioms for doing common operations!

spinalcordtoolbox/labels.py
mguaypaq: Thanks for the sanity check!

spinalcordtoolbox/labels.py
```diff
@@ -224,52 +227,46 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
...
+    labels_phys = ref.transfo_pix2phys(labels_pix, mode='absolute')  # shape (L, 3), float
+    # The corresponding label values.
+    labels_values = ref.data[tuple(labels_pix.T)]  # shape (L,), arbitrary dtype
+    # The closest centerline index for each label voxel.
+    _, labels_indices = centerline_tree.query(labels_phys)  # shape (L,), int
+
+    # Sort everything by increasing centerline index
+    sorter = np.argsort(labels_indices)
+    labels_pix = labels_pix[sorter]
+    labels_phys = labels_phys[sorter]
```
joshuacwnewton: These two quantities don't actually get used elsewhere in the function, right? (I guess it's useful to still sort them to aid with debugging, since the internal state at this point will stay consistent between these quantities.) It's just a (minor) shame that IDEs will grey this out :sweat:

spinalcordtoolbox/labels.py
```diff
@@ -224,52 +227,46 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
...
+    labels_values = labels_values[sorter]
+    labels_indices = labels_indices[sorter]
+    del sorter
+
+    # Add a label for above the top-most label
+    labels_values = list(labels_values)
+    labels_values.append(labels_values[-1] - 1)
+    labels_values = np.array(labels_values)
```
joshuacwnewton: This kind of a nitpick, but when I set a breakpoint and inspected some of the variables, I was surprised to see that `labels_values` had an extra element (because I was looking at the earlier section and saw the comment indicating that it was supposed to be of size `(L,)`.

It might make the internal state of the function a little clearer if we give this variable a new name (say, `region_values`) then move the "add a label" block of code lower down? 

Something like... (text visualization strikes again!)

```python
    # Define the values for each region, e.g.:
    # labels_values = [     5     4     3     2     ]
    # region_values = [  5  |  4  |  3  |  2  |  1  ]
    region_values = np.append(labels_values, labels_values[-1] - 1)
    # Find which region (0, 1, 2...) each index belongs to
    index_regions = np.searchsorted(labels_indices, region_indices)
    # Map the regions (0, 1, 2...) to their values (e.g. 5, 4, 3...)
    out = zeros_like(img)
    out.data[tuple(region_pix.T)] = region_values[index_regions]
```

`np.append` also lets us make the "add a label" block a one-liner?

spinalcordtoolbox/labels.py
```diff
@@ -224,52 +227,46 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
...
+    sorter = np.argsort(labels_indices)
+    labels_pix = labels_pix[sorter]
+    labels_phys = labels_phys[sorter]
+    labels_values = labels_values[sorter]
+    labels_indices = labels_indices[sorter]
+    del sorter
+
+    # Add a label for above the top-most label
+    labels_values = list(labels_values)
+    labels_values.append(labels_values[-1] - 1)
```
joshuacwnewton: Side note: Now that I look at it... if the top-most label is `1`, then the label for the region above `1` will be `0`. 

On the one hand, I think this is somewhat reasonable (since capping it at `1` would cause label-based aggregation to be wrong, and using a non-sequential dummy value like `255` might cause some sort of unexpected error by breaking the assumption that labels are monotonically decreasing).  

But at the same time, outputting _no_ metrics for this case also doesn't feel great? Hrmmm...

I guess this is the exact same as the old behavior, so we can always just open an issue to address it at a later point.

spinalcordtoolbox/labels.py
```diff
@@ -224,52 +227,46 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
...
-            for j in range(len(coordinates_ref) - 1):
-                if coordinates_ref[j + 1].z < z_proj <= coordinates_ref[j].z:
-                    out.data[int(x), int(y), int(z)] = coordinates_ref[j].value
-
-    # Set back the original orientation
-    if img.orientation != og_img_orientation:
-        img.change_orientation(og_img_orientation)
-        out.change_orientation(og_img_orientation)
+        region_pix = np.argwhere(img.data)  # shape (R, 3), int
+    region_phys = img.transfo_pix2phys(region_pix, mode='absolute')  # shape (R, 3), float
```
joshuacwnewton: This is another nitpick, but I feel like the use of the term "region" in the context of "region we want to label" is a bit confusing?

Since the function is called `label_regions_from_reference`, I think "region" is meant to refer to "sub-regions within the image"...

So, maybe naming it like this might be clearer?

```python
    # The segmentation we want to label.
    if centerline:
        segmentation_pix = centerline_pix
    else:
        segmentation_pix = np.argwhere(img.data)  # shape (R, 3), int
    segmentation_phys = img.transfo_pix2phys(segmentation_pix, mode='absolute')  # shape (R, 3), float
    # The closest centerline index for each voxel in the segmentation.
    _, segmentation_indices = centerline_tree.query(segmentation_phys)
```

And if we do this, then the region labeling snippet could be like this:

```python
    # Define the values for each region of the segmentation, e.g.:
    # labels_values = [     5     4     3     2     ]
    # region_values = [  5  |  4  |  3  |  2  |  1  ]
    region_values = np.append(labels_values, labels_values[-1] - 1)
    # Find which region (0, 1, 2...) each segmentation index belongs to
    segmentation_regions = np.searchsorted(labels_indices, segmentation_indices)
    # Map the regions (0, 1, 2...) to their values (e.g. 5, 4, 3...)
    out = zeros_like(img)
    out.data[tuple(segmentation_pix.T)] = region_values[segmentation_regions]
```

joshuacwnewton: I think doing it this way groups the variables nicely in the debugger? 

For example, in the aSCOR test, this looks like:

Image

- `labels_` variables `(4,)`
- `region_` variable `(5,)`
- `segmentation_` variables `(55,)` 
- `centerline_` variables `(55,)` (because `centerline=True`)

joshuacwnewton (approved):
Merging `master` and testing the aSCOR tests, as well as testing on the relevant image in the issue, and everything looks good to me!

Image

I just had some minor minor nitpicks w.r.t. code organization / variable names in the `label_regions` function but they're small enough that I'm happy to pre-approve.

spinalcordtoolbox/labels.py
```diff
@@ -224,52 +227,46 @@ def label_regions_from_reference(img: Image, ref: Image, centerline: bool = Fals
...
+    # The closest centerline index for each region voxel.
+    _, region_indices = centerline_tree.query(region_phys)
```
joshuacwnewton: Actually, one more question: Is there any risk of a `KDTree` matching a segmentation point to a centerline point that's not on the same z-axis? In other words, is the projection guaranteed to always have "flat" interfaces between regions?

spinalcordtoolbox/labels.py
mguaypaq: Yes, I had this same debate internally. Having a clear, consistent state for debugging is also why I did `del sorter` at the end of this paragraph: to really signal that we're done using this permutation.

spinalcordtoolbox/labels.py
mguaypaq: Ah! I looked for something like `np.append` and was so surprised that I couldn't find it. Thanks for pointing it out! And yes, that's a great visualization.

spinalcordtoolbox/labels.py
mguaypaq: Yeah, I think this `-1` is full of bad assumptions, but I opted to just copy the old behaviour.

spinalcordtoolbox/labels.py
mguaypaq: The `KDTree` absolutely affects the z-axis: that's the whole point. This is how (for higher resolution images) we're able to get region boundaries that are perpendicular to the curved centerline. See, for example, the pictures in #4896 and #3395.

joshuacwnewton: Ah right, of course! :facepalm: 

For some reason I was under the impression that we wanted flat boundaries for `sct_process_segmentation`'s vertebral aggregation, but I think I've gotten myself a bit mixed up about what labeled segmentations are used where.

joshuacwnewton (approved):
Thanks so much for addressing the review feedback so promptly! LGTM! :hearts:

Given that the scope of this PR is only within the functions (and doesn't touch any of the outer `-discfile` or `-centerline` code), it should be perfectly fine to merge this independently from the other CLI PRs open right now. :)

PR #4938, opened 2025-06-20
Expand `LazyLoader` usage to improve import times across scripts

joshuacwnewton:
## Checklist

- [x] **PR Sidebar**: I've filled in each of the options within the PR sidebar. ([Reviewers](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers), [Assignees](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#232-assignees), [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!), [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone), [Development](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#235-development).)
- [x] **Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [x] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
[...]

PR #5085, opened 2025-11-05
Improved Handling of Parallel Processes for QC Writing

### LINKED ISSUE #5084: `RuntimeError: PytorchStreamReader` when running `sct_deepseg` in parallel using a not-installed task
### Description

Running a batch job in parallel which requires a model that SCT hasn't already installed will result in any processes which try to start immediately after the first to crash, as they try to load a model which is in the process of being installed.

**Expected behavior:** The first process requiring a model installs said model, with other processes "waiting" until installation is completed.

**Actual behavior:** The first process starts installing the model, which creates an (incomplete) model file that the other processes detect, try to load, and fail (crashing in the process)

### Proposed Solution(s)

**Install Common Models by Default**

Dr. Cohen-Adad suggests the Spinal Cord, Grey Matter, and TotalSpineSeg by default, to reduce the likelihood of this occurring (as well as making common offline analyses possible)

**"Installation" Flag File**

Modify the model installation of SCT models to better handle "installation" collisions. Requires two changes:

* When a model is being installed, create a "flag" file next to the installation location which indicates that the model is in the process of being installed. Remove the file when the installation is completed.
* When going to install a model, check if the prior file exists; if it does, instead of trying to use the model file, wait a short period of time and check again. Repeat until the file no longer exists.

SomeoneInParticular:
## Checklist

- [x] ~~**Contributing Docs**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)~~ N/A
- [x] ~~**Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).~~ N/A
- [x] ~~**Documentation**: I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages.~~ N/A

## Description

In preparation for #5084, I reviewed existing uses of the `mutex` utility. This revealed that, as coded, said `mutex` utility (being a context manager) was overly restrictive in two regards:
* It prevents the addition of an informative error message when an error occurred during QC writing. 
* It prevents conditional locking; while not immediately relevant to QC writing, resolving this will allow any fixes to be conditional on whether a model is in the process of being installed or not.

This PR aims to address these shortcomings, replacing the use of `mutex` with a more explicit `BoundedSemaphore` instance to provide better control of error flow. This also removes the sole use of `mutex` in SCT; as such, it has been removed.

## Linked issues
Initial preparation for #5084

### DIFF
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -119,7 +119,7 @@ def create_qc_entry(
     path_result = path_json / f'qc_{timestamp}.json'
 
     # Create a mutex for the QC file to avoid a race condition if multiple files try to write simultaneously
-    # KO, MB, and JN: The hash here is required to sanitize the path name. Otherwise, special characters for the OS
+    # KO, MGP, and JN: The hash here is required to sanitize the path name. Otherwise, special characters for the OS
     #   (i.e. slashes) being present  can result in "orphaned" paths, due to how PortaLocker works; under the
     #   hood it creates a "lock" file to track the semaphore's state, and references that when determining whether a
     #   process can acquire it or not. As this creation is NOT sanitized (using a raw 'PathLib' query), we need to do
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -7,7 +7,6 @@ License: see the file LICENSE
 
 from contextlib import contextmanager
 import datetime
-from hashlib import md5
 import importlib.resources
 import itertools as it
 import json
@@ -18,6 +17,8 @@ import shutil
 from typing import Optional, Sequence
 
 import numpy as np
+import portalocker
+from portalocker import Lock, LockFlags
 from scipy.ndimage import center_of_mass
 import skimage.exposure
 
@@ -29,7 +30,6 @@ from spinalcordtoolbox.reports.assets.py import refresh_qc_entries
 from spinalcordtoolbox.resampling import resample_nib
 from spinalcordtoolbox.utils.shell import display_open
 from spinalcordtoolbox.utils.sys import __version__, list2cmdline, LazyLoader
-from spinalcordtoolbox.utils.fs import mutex
 
 pd = LazyLoader("pd", globals(), "pandas")
 mpl_plt = LazyLoader("mpl_plt", globals(), "matplotlib.pyplot")
@@ -114,13 +114,23 @@ def create_qc_entry(
         if not path.exists():
             raise FileNotFoundError(f"Required QC image '{img_type}' was not found at the expected path: '{path}'")
 
-    # Use mutex to ensure that we're only generating shared QC assets using one process at a time
-    realpath = path_qc.resolve()
-    with mutex(f"sct_qc-{realpath.name}-{md5(str(realpath).encode('utf-8')).hexdigest()}"):
-        # Create a json file for the new QC report entry
-        path_json = path_qc / '_json'
+    # Define where the JSON entry should be saved
+    path_json = path_qc / '_json'
+    path_result = path_json / f'qc_{timestamp}.json'
+
+    # Create a mutex for the QC file to avoid a race condition if multiple files try to write simultaneously
+    qc_mutex = portalocker.BoundedSemaphore(
+        maximum=1,  # Mutually exclusive access
+        name=str(path_qc.resolve()),  # Lock the directory itself
+        timeout=60,  # Wait up to 60 seconds for the directory to become free again
+        check_interval=0.1  # Check every 1/10th of a second
+    )
+    try:
+        # Attempt to acquire the lock and save the QC entry
+        qc_mutex.acquire()
+
+        # Create the JSON file for this new QC entry (as well as its containing directory, if needed)
         path_json.mkdir(parents=True, exist_ok=True)
-        path_result = path_json / f'qc_{timestamp}.json'
         with path_result.open('w') as file_result:
             json.dump({
                 'path': str(Path.cwd()),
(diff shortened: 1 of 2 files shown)


spinalcordtoolbox/reports/qc2.py
```diff
@@ -146,8 +163,20 @@ def create_qc_entry(
...
+        logger.info('Successfully generated the QC results in %s', str(path_result))
+
+        # Prompt the user to view the results in the browser
+        display_open(file=str(path_index_html), message="To see the results in a browser")
+    # Regardless of whether hte
+    except Exception as e:
+        # Log that an error occurred during QC generation
+        logger.info('Failed to write QC results in %s', str(path_result))
+        # Re-raise the error for debugging purposes
+        raise e
```
mguaypaq, line 176: If this error handling is the only reason for getting rid of `mutex`, I think it could also have been handled with something like:

```py3
try:
    with mutex(...):
       ...
except Exception:
    logger.info('Failed to write QC results in %s', str(path_result))
    raise
```

The `try ... finally` block inside `mutex` doesn't swallow exceptions, so they would still pass through the outer `try ... except` for logging.

If you _do_ restore `mutex`, maybe it should accept a `pathlib.Path` instead of a `str` for the name, and the hashing logic could live there.

spinalcordtoolbox/reports/qc2.py
```diff
@@ -114,13 +114,30 @@ def create_qc_entry(
...
-    realpath = path_qc.resolve()
-    with mutex(f"sct_qc-{realpath.name}-{md5(str(realpath).encode('utf-8')).hexdigest()}"):
-        # Create a json file for the new QC report entry
-        path_json = path_qc / '_json'
+    # Define where the JSON entry should be saved
+    path_json = path_qc / '_json'
+    path_result = path_json / f'qc_{timestamp}.json'
+
+    # Create a mutex for the QC file to avoid a race condition if multiple files try to write simultaneously
+    # KO, MB, and JN: The hash here is required to sanitize the path name. Otherwise, special characters for the OS
```
mguaypaq: Thanks for making this design constraint explicit in a code comment.

Also, tiny typo:

```suggestion
    # KO, MGP, and JN: The hash here is required to sanitize the path name. Otherwise, special characters for the OS
```

SomeoneInParticular: Oh, whoops; thanks for catching that.

spinalcordtoolbox/reports/qc2.py
SomeoneInParticular: So, you're correct, but I'm going to argue that we should still remove `mutex` for a few reasons:

1. [Minor] By making `mutex` a util within `fs.py`, the Portalocker dependency gets imported whenever `fs` is used. By extracting the `mutex` use to its one implementation, this is no longer the case.
2. [Minor] Having a single-use alias/util can lead to confusion for developers; its forces them to jump across files to understand what a code snippet is doing (and why) rather than reading the code in-place
3. The current `mutex` implementation prevents running things conditional on whether it is available or not. While not relevant for this PR, it will be with the upcoming fix for #5084.

Re: point 3, I will detail it further in PR #5086, where it is more relevant. This PR is separate from that one solely to provide better check pointing should something need to be rolled back (also makes our change log a bit prettier)

joshuacwnewton: > 1. [Minor] By making `mutex` a util within `fs.py`, the Portalocker dependency gets imported whenever `fs` is used. By extracting the `mutex` use to its one implementation, this is no longer the case.

I think this might be a little moot given how widespread the import of `qc2.py` is in scripts. 😝 

(Also, I don't recall `portalocker` being one of the heavier dependencies to import when I was doing my most [recent lazy loading analysis](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4938)?)

> 2. [Minor] Having a single-use alias/util can lead to confusion for developers; its forces them to jump across files to understand what a code snippet is doing (and why) rather than reading the code in-place

It's a tricky balance... do we duplicate the `BoundedSemaphore` usage (and surrounding boilerplate) every time we need a mutex? Or do we try to unify the usage at the cost of some readability?

(Personally I think the `with mutex():` pattern is clear enough to not need to jump across assuming the dev knows what a mutex is, but that's not exactly a guarantee. 😅)

> 3. The current mutex implementation prevents running things conditional on whether it is available or not. While not relevant for this PR, it will be with the upcoming fix for #5084.

Personally, I think I'd find it easier to review these changes if it were all one PR? Since for me, whether or not to remove `mutex` hinges almost entirely on point 3, and so it's hard for me to feel comfortable merging a PR like this separately from the inciting change.

SomeoneInParticular: Fair points; I'll close this PR, and move its changes over to the other PR to put it into better context. Will transfer the commentary over as well. Thanks for the feedback!

spinalcordtoolbox/utils/fs.py
```diff
@@ -262,24 +259,3 @@ def relpath_or_abspath(child_path, parent_path):
         return abspath.relative_to(parent_path)
     except ValueError:
         return abspath
-
-
-@contextmanager
-def mutex(name):
```
joshuacwnewton, line 268: Just as a sidebar to make sure I fully understand this one point:

> It prevents conditional locking; while not immediately relevant to QC writing, resolving this will allow the fix for https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5084 to be conditional on whether a model is in the process of being installed or not.

From what I understand, the parallel processes would execute like this with `mutex`:

1. Process 1 grabs the `mutex`, halting other processes.
2. Process 1 installs the model, then gives up the `mutex`.
3. For the remaining processes:
   1. Process N grabs the mutex unconditionally (even though the model is already installed).
   2. Process N checks the model folder and finds the model already installed, does nothing.
   3. Repeat for the remaining processes.

And your goal here with "conditional locking" is for the locking behavior to be smart enough to skip 3. entirely, having the remaining processes go straight to parallelized inference once the model is installed by the first process?

If so, my question is... how big of a concern is it for there to be extraneous locking for the "check if model exists" step? I was imagining that it would be quick enough to check for model existence that it wouldn't really delay the parallelized inference by a noticeable amount of time. (Also, with the default `-jobs 4` or `-jobs 8` or what have you, there's already a per-subject bottleneck for the entire processing that I figure the amount of time spent in the lock rounds down to 0?)

But, maybe there's a downside I'm missing? 🤔

joshuacwnewton: Whoops sorry, I authored this comment in one go without checking my notifications to see that this PR was closed!

PR #5086, opened 2025-11-06
Parallel DeepSeg processes no longer try to use models which are in the process of being installed

### LINKED ISSUE #5084: `RuntimeError: PytorchStreamReader` when running `sct_deepseg` in parallel using a not-installed task

SomeoneInParticular:
## Checklist

## Description

As the title implies, via the use of a "mutex" semaphore, `sct_deepseg` processes run in parallel will no longer crash if/when they try to use a model that is in the process of being installed by another running process.

## Linked issues

Fixes #5084

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -16,8 +16,11 @@ import glob
 from pathlib import Path
 from importlib.metadata import metadata
 
+import portalocker.exceptions
+
 from spinalcordtoolbox import download
-from spinalcordtoolbox.utils.sys import stylize, __deepseg_dir__, LazyLoader
+from spinalcordtoolbox.utils.fs import Mutex
+from spinalcordtoolbox.utils.sys import stylize, __deepseg_dir__, LazyLoader, printv
 
 tss_init = LazyLoader("tss_init", globals(), 'totalspineseg.init_inference')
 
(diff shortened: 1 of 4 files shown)


spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+    :param waiting_msg: Message to display when waiting to acquire the lock. If None, no message will print.
     """
-    Use a bounded semaphore as a mutex to prevent parallel processes from running.
+    def __init__(
+        self,
+        key: str,
+        directory: str = tempfile.gettempdir(),
+        timeout: float | None = 5,  # 5 seconds
+        check_interval: float | None = 0.25,  # Every quarter second
+        fail_when_locked: bool | None = True,
```
mguaypaq, line 296: Instead of hardcoding the defaults of (the current version of) portalocker's `BoundedSemaphore`, I would suggest using `**kwargs` and passing them on to `super().__init__` (if we want to consider `Mutex` to be a thin wrapper).

If we want to consider `Mutex` to be a self-contained abstraction instead, I would suggest hiding anything we don't actually use from `BoundedSemaphore` as an implementation detail (and probably not sub-classing it at all).

(These suggestions are coming from a principle of trying to have a single source of truth, because duplicated truths tend to get out of sync. But, I can see an argument that it's very convenient to have local documentation, instead of having to dig around a dependency's source code for details.)

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+
+        # Hashed version of our name to sanitize it
+        self.hashed_name = md5(self.name.encode('utf-8')).hexdigest()
+
+    def __setattr__(self, key, value):
+        # Block setting the "maximum" value to anything other than 1
+        if key == "maximum" and value != 1:
+            raise ValueError("A mutex cannot allow more than 1 process to hold the lock!")
+        # Otherwise, proceed normally
+        super().__setattr__(key, value)
```
mguaypaq, line 321: Is this useful? I only had a quick look, but I didn't see anything in portalocker itself that would touch this attribute, and I think the usual Python attitude is "you break it you buy it" whenever someone directly touches an object's internals.

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+        key: str,
+        directory: str = tempfile.gettempdir(),
+        timeout: float | None = 5,  # 5 seconds
+        check_interval: float | None = 0.25,  # Every quarter second
+        fail_when_locked: bool | None = True,
+        waiting_msg: str = None
+    ):
+        # Run a (slightly constrained) version of the super-class's constructor
+        super().__init__(
+            name=key,
```
mguaypaq, line 301: Why not pass the sanitized name to `super` here, instead of overriding `get_filename`? This would also be more aligned with how portalocker itself did subclassing with `NamedBoundedSemaphore`.

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+        # Otherwise, proceed normally
+        super().__setattr__(key, value)
+
+    def __enter__(self):
+        # Notify the user if they are using a non-failing Mutex as a context manager that we
+        #   are potentially overriding their intent; see rationale below.
+        if not self.fail_when_locked:
+            print(logger.warning(
+                f"LoggingMutex '{self.name}' being run in 'fail_when_locked=True' mode to ensure "
+                f"that code within the Python context will not run without lock acquisition."
```
mguaypaq, line 329: Why even expose the `fail_when_locked=False` option when it's clearly not the intended usage?

If `fail_when_locked` was forced to `True` (like `maximum` is forced to 1), then both `__enter__` and `__exit__` from the superclass `LockBase` could be used directly.

Also, is `LoggingMutex` a previous name for this class?

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+        check_interval: float | None = None,
+        fail_when_locked: bool | None = None,
+    ) -> portalocker.Lock | None:
+        """
+        Same as super-class implementation, but logs a message if it wasn't immediately
+        able to acquire the lock.
+        """
+        # If we have a message on miss, do a very short acquisition attempt first
+        timeout = timeout if timeout else self.timeout
+        check_interval = check_interval if check_interval else self.check_interval
```
mguaypaq: You probably want to distinguish between `None` and `0` here:
```suggestion
        timeout = timeout if timeout is not None else self.timeout
        check_interval = check_interval if check_interval is not None else self.check_interval
```

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
     """
-    Use a bounded semaphore as a mutex to prevent parallel processes from running.
+    def __init__(
+        self,
+        key: str,
+        directory: str = tempfile.gettempdir(),
+        timeout: float | None = 5,  # 5 seconds
+        check_interval: float | None = 0.25,  # Every quarter second
+        fail_when_locked: bool | None = True,
+        waiting_msg: str = None
```
mguaypaq, line 297: Why is this optional? It seems like it should be a required parameter.

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+        check_interval = check_interval if check_interval else self.check_interval
+        if self.waiting_msg and timeout > check_interval:
+            super().acquire(0, check_interval, fail_when_locked=False)
 
-    portalocker.BoundedSemaphore is very similar to threading.BoundedSemaphore,
-    but works across multiple processes and across multiple operating systems. So,
-    we can spawn multiple processes using `sct_run_batch`, while still ensuring
-    that we create QC reports sequentially (without mangling the output files).
+            # If we have the lock, return here: we already have the lock!
+            if self.lock is not None:
```
mguaypaq, line 357: Just a note: checking `if self.lock` right after `super().acquire(fail_when_locked=False)` is the same as `try`ing to `super().acquire(fail_when_locked=True)`.

spinalcordtoolbox/utils/fs.py
```diff
@@ -264,22 +263,111 @@ def relpath_or_abspath(child_path, parent_path):
...
+        return super().acquire(timeout, check_interval, fail_when_locked)
+
+    def get_filename(self, number: int) -> Path:
+        """
+        Returns a pointing to a file which uses our hashed name,
+        rather than using our name in its raw form. This ensures
+        path access and creation is sanitized preventing difficult
+        to debug (and potentially destructive) side effects.
+        """
+        return Path(self.directory) / f"{self.hashed_name}_{number}.lock"
```
mguaypaq, line 373: Having just the hash is not quite as nice for debugging as also including the base name of the "locked" file/directory. (Implementing this would probably mean that `key` needs to be a `Path` rather than a `str`.)

mguaypaq (commented):
Here are a few comments on the `Mutex` implementation, for now. I like your idea of logging a message when the lock isn't immediately available.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: > I would suggest using **kwargs and passing them on to super().__init__ (if we want to consider Mutex to be a thin wrapper).

I originally tried that, but realized that allowed the user to set the `maximum` parameter to a non-1 value. I could error out in this context, but that seems more confusing to newer developers (see my final remark below); this approach explicitly prevents that.

> If we want to consider Mutex to be a self-contained abstraction instead, I would suggest hiding anything we don't actually use from BoundedSemaphore as an implementation detail (and probably not sub-classing it at all).

Without sub-classing, we are either forced to always treat it as a context manager, always treat it as a semaphore object, or make a distinct function for each. Far as I can tell, this is the only way to allow one object/call do both, which is much more efficient (and, in my opinion, easier to read). 

> These suggestions are coming from a principle of trying to have a single source of truth

Most of our development is done by one of two parties currently:

* Core devs, who know the code base inside-and-out already
* Our rotating students, who are not always proficient in using an IDE (or in Python for that matter)

The former group probably doesn't need any help here; between their own knowledge and modern IDE support, they will know where to dig to find what they need. For the latter group, I believe having all docs up front and irrelevant elements explicitly suppressed is much more approachable. Still, this is definitely subjective; if the weather lets up this week I'll try and come into the office to discuss it more.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: > I think the usual Python attitude is "you break it you buy it"

That doesn't seem to be the case anymore; Python has its own way to marking variables `protected` and `private` now (by adding, respectively, one and two `_` in front of the variable name). The former is just semantic sugar (Most IDEs will warn you for touching them, but Python won't stop you), but the latter _is_ enforced by Python (via variable mangling during code initialization; see [here](https://peps.python.org/pep-0008/#method-names-and-instance-variables) for further details). 

Given newer Python devs are likely to learn this semantic instead, I put this catch in place to slap their hand if they try. It isn't the most intuitive approach, though, I'll admit; will look into better alternatives.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: > Why not pass the sanitized name to `super` here, instead of overriding `get_filename`

~~Ah! I actually did that originally, but that would make logging also use the hashed name, making debugging quite painful. This approach splits us down the middle; logging uses the "raw" key (which should be safe to print to console), while the back-end file Semaphore file uses the sanitized version.~~

Ignore this; seems I got my variables mixed up. Will look into this more.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: > Why even expose the `fail_when_locked=False` option when it's clearly not the intended usage?

> If `fail_when_locked` was forced to `True` (like maximum is forced to 1), then both `__enter__` and `__exit__` from the superclass `LockBase` could be used directly.

You are correct; this is a fail-safe I put in place to block unintuitive behavior from the `LockBase` class (where you can enter the context w/o actually acquiring the lock if it reaches the `timeout` period). Granted, I also managed to miss that `LockBase` provides  `__enter__` and `__exit__` already; means we can safely deleted `__exit__`, as its entirely redundant.

> Also, is `LoggingMutex` a previous name for this class?

Yup! Will fix that in the next commit as well.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: Thanks! Good catch.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: You're correct; seems I'm still in the habit of Java conventions, where `Exception` handling is much more expensive than catching them in advance with an `if-else` query. Seems its the opposite in Python; will push a commit to resolve soon.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: Fair point; I would argue the addition of the `waiting_msg` argument somewhat mitigates that, though.

spinalcordtoolbox/utils/fs.py
SomeoneInParticular: Re: `waiting_msg`, fair point. Will make it mandatory.

Re: the others, they're just copied over from `BoundedSemaphore` to mimic its behavior.

spinalcordtoolbox/utils/fs.py
mguaypaq: Oh, the single and double leading underscores have been there for ages, that's not new. Single underscores are much more common (and, being only a convention, are the embodiment of "you break it you buy it"). For double underscores, the name mangling makes it a stronger convention, but definitely don't rely on it for security purposes.

SomeoneInParticular: I could have sworn the name mangling was (relatively) new; in either case, that lack of security was precisely why this check exists (to explicitly slap the dev for trying; our users should never ever see this).

spinalcordtoolbox/deepseg/models.py
```diff
@@ -16,8 +16,11 @@
...
 from importlib.metadata import metadata
 
+import portalocker.exceptions
+
 from spinalcordtoolbox import download
-from spinalcordtoolbox.utils.sys import stylize, __deepseg_dir__, LazyLoader
+from spinalcordtoolbox.utils.fs import Mutex
+from spinalcordtoolbox.utils.sys import stylize, __deepseg_dir__, LazyLoader, printv
 
 tss_init = LazyLoader("tss_init", globals(), 'totalspineseg.init_inference')
```
joshuacwnewton, line 25: Note that `totalspineseg` (i.e. `sct_deepseg -spine`) uses a slightly different method for installation (internal to the `totalspineseg` package) and so it results in a different error:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5149

Just mentioning it here to ensure that whatever solution we come up with accounts for this too.

Issue #5096, opened 2025-11-20
Reducing `totalspineseg` output/postprocessing for faster inference


plbenveniste:
When running `sct_deepseg totalspineseg` on GPU, the process produces a lot of files and goes through a lot of different steps which could be avoided.

**Objective:** The idea is to remove some stuff from the basic inference to keep only the steps necessary to get the levels.

Here is an example run on a GPU: 

```console
(base) plbenveniste@kronos:~/test_totalspinseg_speed$ SCT_USE_GPU=1 sct_deepseg totalspineseg -i sub-cal095_ses-M0_STIR.nii.gz  -step1-only 1 | awk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0; fflush(); }'
2025-11-20 11:46:08 
2025-11-20 11:46:08 --
2025-11-20 11:46:08 Spinal Cord Toolbox (git-master-bca796fdcdeb214ef343401d2ba4bd670db88e6b)
2025-11-20 11:46:08 
2025-11-20 11:46:08 sct_deepseg totalspineseg -i sub-cal095_ses-M0_STIR.nii.gz -step1-only 1
2025-11-20 11:46:08 --
2025-11-20 11:46:08 
2025-11-20 11:46:08 Model 'totalspineseg' is up to date (Source: {'Dataset101_TotalSpineSeg_step1': 'https://github.com/neuropoly/totalspineseg/releases/download/r20241115/Dataset101_TotalSpineSeg_step1_r20241115.zip', 'Dataset102_TotalSpineSeg_step2': 'https://github.com/neuropoly/totalspineseg/releases/download/r20241115/Dataset102_TotalSpineSeg_step2_r20241115.zip'})
2025-11-20 11:46:11 Note: NumExpr detected 32 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 16.
2025-11-20 11:46:11 NumExpr defaulting to 16 threads.
2025-11-20 11:46:13 Creating temporary folder (/tmp/sct_2025-11-20_11-46-13_sct_deepseg_hx18gz4p)
2025-11-20 11:46:13 Copied sub-cal095_ses-M0_STIR.nii.gz to /tmp/sct_2025-11-20_11-46-13_sct_deepseg_hx18gz4p/sub-cal095_ses-M0_STIR.nii.gz
2025-11-20 11:46:13 
2025-11-20 11:46:13 Running TotalSpineSeg with the following parameters:
[...]

Issue #5033, opened 2025-09-24
Change `-step1-only` argument to be default behavior and replace with `-label-vert` argument


jcohenadad:
In most cases, users will want `-step1-only 1`, so it should be the default.

Also, I find it more intuitive to set a flag to activate something, not to set a flag to deactivate something.

So my suggestion would be to change this flag for `-step2`, to indicate that step 2 should be run. 

We should do that ASAP (ie: before the next release), otherwise people will start writing up pipelines with -step1-only, which will eventually be deprecated

Issue #4972, opened 2025-07-15
Improve documentation for `sct_deepseg totalspineseg`


plbenveniste:
I am opening this issue to keep in mind that I want to improve the documentation regarding `sct_deepseg totalspineseg`. The output of this algorithm is quite confusing with a lot of files given as output. I would like to improve the current documentation of this algorithm to make it more complete. Currently, the documentation is not specific to the sub-algorithm used in `sct_deepseg`. 

Current documentation: 

<img width="743" height="807" alt="Image" src="https://github.com/user-attachments/assets/137d6b86-3a51-42bc-8716-3a9b11b61fa4" />

Proposed addition: 
- listing of all output file and what they contain

joshuacwnewton:
Agreed! In fact, I think this goes beyond `totalspineseg` -- it is confusing that certain details that are specific to some models are copied for every model (such as the `lesion_ms_mp2rage` note).

plbenveniste:
I am also referencing this issue  https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4867 which @NathanMolinier mentionned in a previous discussion https://github.com/ivadomed/ms-lesion-agnostic/pull/63#discussion_r2250891514

Issue #4867, opened 2025-04-22
Rename  `sct_deepseg totalspineseg` output to make it more user friendly


NathanMolinier:
## Description

Using `totalspineseg` through `sct_deepseg` may not be entirely user-friendly, particularly because the current naming conventions—such as `step1` and `step2` can be unclear for users unfamiliar with the main repository. Therefore we should think about renaming the generated files.

jcohenadad:
also from Nathan:

> D'ailleurs il est maintenant possible de lancer uniquement le premier modèle pour obtenir juste les niveaux intervertébraux avec le flag `-step1-only`

jcohenadad:
how about:
- `sub-amuAP_T1w_totalspineseg_cord.nii.gz` --> soft segmentation of the spinal cord based on nnunet probabilities
- `sub-amuAP_T1w_totalspineseg_canal.nii.gz` --> soft segmentation of the spinal canal based on nnunet probabilities
- `sub-amuAP_T1w_totalspineseg_all.nii.gz` --> segmentation of the vertebrae (single class), discs (multi class), canal (single class) and cord (single class) 
- `sub-amuAP_T1w_totalspineseg_levels.nii.gz` --> single voxels at the posterior tip of the discs (different values)

If step2 is ran, it would overwrite:
- `sub-amuAP_T1w_totalspineseg_all.nii.gz` --> segmentation of the vertebrae (single class), discs (multi class), canal (single 

And of course, a JSON file per NIIGZ

NathanMolinier:
Goog idea !

valosekj:
> - `sub-amuAP_T1w_totalspineseg_levels.nii.gz` --> single voxels at the posterior tip of the discs (different values)

How about `_discs.nii.gz` instead of `_levels.nii.gz`. To be more explicit that we're talking about discs. As levels can be vertebral or spinal.

PR #5097, opened 2025-11-20
Rework `totalspineseg` inference (rename task to `spine`, set `-step1-only` behavior as default, improve output filenames)

NathanMolinier:
## Checklist

## Description

This PR introduces several updates related to the integration of totalspineseg:
- [x] Renames generated files to match SCT’s naming conventions. ([#4867](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4867))
- [x] Skips unnecessary steps to reduce overall inference time. ([#5096](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5096) from SCT meeting 20251120)
- [x] Renames the function from `totalspineseg` to the more intuitive `spine`. (SCT meeting 20251120)
- [x] Changes the default behavior to run only the first processing step, with the second step executed only when the new -label-spine flag is provided. ([#5033](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5033))
- [x] Update the doc. ([#4972](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4972))

## Linked issues

- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5033
- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4972
- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4867
- Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5096


spinalcordtoolbox/deepseg/inference.py
```diff
@@ -420,8 +420,7 @@ def segment_totalspineseg(path_img, tmpdir, predictor, device, step1_only=False)
         max_workers=1,
         max_workers_nnunet=1,
         # Optional argument to choose which models to run
-        step1_only=bool(step1_only)
-    )
+        keep_only=['step1_levels', 'step1_output'] if not label_vert else ['step1_levels', 'step1_canal', 'step1_cord', 'step2_output'],
```
joshuacwnewton: I didn't realize that it had a `keep_only` argument already! Very useful.

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -300,11 +300,9 @@ def get_parser(subparser_to_return=None):
                 metavar=Metavar.str)
         if task_name == 'totalspineseg':
             params.add_argument(
-                "-step1-only",
+                "-label-vert",
                 type=int,
-                help="If set to '1', only Step 1 will be performed. If not provided, both steps will be run.\n"
-                     "- Step 1: Segments the spinal cord, spinal canal, vertebrae, and intervertebral discs (IVDs). Labels the IVDs, but vertebrae are left unlabeled.\n"
-                     "- Step 2: Fine-tunes the segmentation, applies labels to vertebrae, and segments the sacrum if present.\n"
```
joshuacwnewton, line 307: I'm curious, previously we said that `step 2` does more than just applying labels to the vertebrae.

Do we still care about fine-tuning the segmentation / segmenting the sacrum?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -300,11 +300,9 @@ def get_parser(subparser_to_return=None):
...
+                help="If set to '1', give each vertebrae in the spine segmentations a unique label (e.g. C1: 11 C2: 12 etc.). "
```
joshuacwnewton: We might want to still mention that `-label-vert` runs a second model (and thus will take a bit longer).

```suggestion
                help="If set to '1', run a second model that applies a unique label to each individual vertebrae in "
                     "the spine segmentation (e.g. C1: 11 C2: 12 etc.). "
                     "If not specified, all segmented vertebrae will have the same value (FIXME -- what value?), which is faster to compute if you only need e.g. disc labels."
```

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -422,12 +422,23 @@ def segment_totalspineseg(path_img, tmpdir, predictor, device, label_vert=False)
...
         keep_only=['step1_levels', 'step1_output'] if not label_vert else ['step1_levels', 'step1_canal', 'step1_cord', 'step2_output'],
+    )
+
+    name_remapping = {
+        "step1_canal": "totalspineseg_canal",
+        "step1_cord": "totalspineseg_cord",
+        "step1_levels": "totalspineseg_discs",
+        "step1_output": "totalspineseg_all",
+        "step2_output": "totalspineseg_all",
+    }
```
joshuacwnewton, line 431: Do we still want `totalspineseg` as a suffix if we're hiding the `totalspineseg` name?

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 1: It looks like the documentation is failing because `totalspineseg.rst` still expects there to be a `totalspineseg` task:

```
 /home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/checkouts/5097/documentation/source/user_section/command-line/deepseg/totalspineseg.rst:6: ERROR: Module "spinalcordtoolbox.scripts.sct_deepseg" has no attribute "totalspineseg" 
```

So, I think we will need to update all of the `totalspineseg` mentions in the documentation as well to be `spine` instead?

joshuacwnewton, line 575: Even though we have named the task as `spine`, the long_description still mentions `TotalSpineSeg`... maybe we could change this to something like:


```python
'long_description': 'Full spine segmentation. Specifically, automatic instance segmentation of all vertebrae, intervertebral discs (IVDs), '
'spinal cord, and spinal canal in MRI images. It is robust to various MRI contrasts, acquisition orientations, '
'and resolutions. The `spine` task uses the TotalSpineSeg model (based on nnU-Net) as the backbone for training and inference.',
'url': 'https://github.com/neuropoly/totalspineseg',
'models': ['totalspineseg'],
```

spinalcordtoolbox/deepseg/models.py
```diff
@@ -565,7 +565,7 @@
          'models': ['model_seg_canal_t2w'],
          'citation': None
          },
-    'totalspineseg':
```
joshuacwnewton, line 568: I think we may need to update the QC to detect `spine` as well:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3b1f85e80a9ed2babb58fd7774ec8f7a9d60065d/spinalcordtoolbox/reports/qc2.py#L314-L318

joshuacwnewton, line 568: I think we will also need to change this line in `sct_process_segmentation`'s argparse help:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/6016fe9fd0e0917c85afb607488b150e4986fda8/spinalcordtoolbox/scripts/sct_process_segmentation.py#L184

spinalcordtoolbox/deepseg/inference.py
```diff
@@ -420,14 +420,25 @@ def segment_totalspineseg(path_img, tmpdir, predictor, device, step1_only=False)
         max_workers=1,
         max_workers_nnunet=1,
         # Optional argument to choose which models to run
-        step1_only=bool(step1_only)
+        step1_only=bool(not label_vert),
+        keep_only=['step1_levels', 'step1_output'] if not label_vert else ['step1_levels', 'step1_canal', 'step1_cord', 'step2_output'],
```
joshuacwnewton: Hmm... I'm not sure how I feel about `-label-vert` controlling whether `step1_canal` and `step1_cord` are output.

If we do this, we should be very clear that `-label-vert` does more than just label the vertebrae -- it also adds extra outputs too. 

I'm wondering whether we should consider a more general argument?

Like, maybe we do something like `-output-files` and have the user specify one or more options between `('disclabels', 'cord', 'canal', 'fullspine')` with the selection being `disclabels` by default? I guess this is a problem for the QC, since it will be unclear which of the output files in the correct ones... :thinking:

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -559,16 +557,16 @@ def main(argv: Sequence[str]):
         # Models can have multiple input images -- create 1 QC report per input image.
         if len(output_filenames) == len(input_filenames):
             iterator = zip(input_filenames, output_filenames, [None] * len(input_filenames), qc_seg)
-        # Special case: totalspineseg which outputs 4-5 files per 1 input file
-        elif arguments.task == 'totalspineseg':
-            # `-step1-only: 1`: Use the 4th image ([3]) which represents the step1 output
-            if getattr(arguments, "step1_only") == 1:
+        # Special case: spine which outputs 4-5 files per 1 input file
```
joshuacwnewton: Tiny nit: Make sure 4-5 is updated to 2-4 (or whatever the final amount of files ends up being).

spinalcordtoolbox/scripts/sct_deepseg.py
NathanMolinier: Not really, and I felt that removing step1 and step2 will make the function more user friendly.

spinalcordtoolbox/deepseg/inference.py
NathanMolinier: That is what was suggested here https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4867

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Done in [9e7c232](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/9e7c232799ff36de810854a95aab44175d51d0e5)

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Isn't it fine this way ? We provide the URL to totalspineseg anyway

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Done in [9c8787d](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/9c8787d032ba966c9b19de22750465627412ebe3)

spinalcordtoolbox/scripts/sct_deepseg.py
NathanMolinier: Done in [57fdff4](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/57fdff46a1e4241d56497bf36155c2052cacd224)

spinalcordtoolbox/deepseg/inference.py
NathanMolinier: You are right, and honestly I was considering not outputting `step1_canal` and `step1_cord` as they are already part of `step1_output` and `step2_output`

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Done in [5665fdf](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/5665fdf8d61c9b976e17a7feab4d192c322c1dd3)

spinalcordtoolbox/deepseg/inference.py
joshuacwnewton: Very fair, I was just wondering if @jcohenadad has changed his opinion since suggesting the `totalspineseg` -> `spine` change just today.

spinalcordtoolbox/deepseg/models.py
joshuacwnewton: That's true, I was just wondering if it might be confusing for users to if we say "oh the task is called spine but the description says TotalSpineSeg"...

I think you're right though, it seems fine.

spinalcordtoolbox/deepseg/inference.py
joshuacwnewton: I think that makes sense, if someone wants the cord/canal they can always post-process `_output` by keeping only the values associated with cord/canal (I'm assuming that's what TSS does internally?)

But more likely I imagine people would rather get the cord/canal from other models, so outputting them is probably more confusing than anything.

NathanMolinier: > I'm assuming that's what TSS does internally?

Yes indeed

> So outputting them is probably more confusing than anything

Agree

NathanMolinier: Done in [774e1b9](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/774e1b98ed650e227b79801444606059ee449951)

spinalcordtoolbox/deepseg/models.py
joshuacwnewton, line 568: In `sct_deepseg.rst`, I think there are some references to `totalspineseg`

joshuacwnewton, line 568: Also, in the tests (both the qc tests and the deepseg tests), I think we are still testing against `totalspineseg`.

joshuacwnewton (changes requested):
I think there are just a few more instances of `totalspineseg` we need to update?

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -298,13 +298,13 @@ def get_parser(subparser_to_return=None):
...
             params.add_argument(
-                "-step1-only",
+                "-label-vert",
                 type=int,
-                help="If set to '1', only Step 1 will be performed. If not provided, both steps will be run.\n"
-                     "- Step 1: Segments the spinal cord, spinal canal, vertebrae, and intervertebral discs (IVDs). Labels the IVDs, but vertebrae are left unlabeled.\n"
-                     "- Step 2: Fine-tunes the segmentation, applies labels to vertebrae, and segments the sacrum if present.\n"
+                help="If set to '1', run a second model that applies a unique label to each individual vertebrae in "
+                     "the spine segmentation (e.g. C1: 11 C2: 12 etc.). "
+                     "If not specified, all segmented vertebrae will have the same value (FIXME -- what value?), which is faster to compute if you only need e.g. disc labels."
```
jcohenadad: just so we don't forget the FIXME

jcohenadad:
Didn't we want the QC to be sagittal? Eg: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4920

image

jcohenadad:
current output only has one JSON (for all):
~~~
-rw-r--r--   1 julien  staff       798 Nov 24 19:52 sub-amu01_T2w_totalspineseg_all.json
-rw-r--r--   1 julien  staff    103561 Nov 24 19:52 sub-amu01_T2w_totalspineseg_all.nii.gz
-rw-r--r--   1 julien  staff     28798 Nov 24 19:52 sub-amu01_T2w_totalspineseg_discs.nii.gz
~~~

do we want it for the discs as well?

spinalcordtoolbox/deepseg/inference.py
jcohenadad: In this case, we should probably keep "totalspineseg", given that the totalspineseg algorithm produces outputs that other algo (eg sct_deepseg spinalcord, sct_deepseg canal) also produce, so we have no choice but explicitly mention totalspineseg for these outputs. 

in an ideal world, we would also start working with BIDS derivative outputs, but that will be for another time.

so yes, i agree with 
~~~
    name_remapping = {
        "step1_canal": "totalspineseg_canal",
        "step1_cord": "totalspineseg_cord",
        "step1_levels": "totalspineseg_discs",
        "step1_output": "totalspineseg_all",
        "step2_output": "totalspineseg_all",
    }
~~~

however, when running 774e1b98ed650e227b79801444606059ee449951, i don't see _cord and _canal. These are under _all:
image

NathanMolinier: Yes, now to save computation time we do not output individual files for canal and cord since those are present in the `_all` file already.

spinalcordtoolbox/deepseg/models.py
NathanMolinier: Is it ok after this commit [9527102](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/9527102315a508824ae58f2554a6d39ce7cb637d)

NathanMolinier: Done [8c066cc](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/8c066cc5d092a2e0055771cab4ce3554cf0ca9fa)

spinalcordtoolbox/scripts/sct_deepseg.py
NathanMolinier: done [fa59a6b](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097/commits/fa59a6bf1bbec831fdc33a3623225271b2b69264)

jcohenadad:
@NathanMolinier https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5097#issuecomment-3581701370 now I'm running into this

```console
julien-macbook:~/data/data-multi-subject/sub-amu01/anat $ sct_deepseg spine -i sub-amu01_T1w.nii.gz

--
Spinal Cord Toolbox (git-nm/update_default_behavior_totalspineseg-117d31b04e36193087625058bc182b72dbc68f43)

sct_deepseg spine -i sub-amu01_T1w.nii.gz
--

Model 'spine' is up to date (Source: {'Dataset101_TotalSpineSeg_step1': 'https://github.com/neuropoly/totalspineseg/releases/download/r20251124/Dataset101_TotalSpineSeg_step1_r20251124.zip', 'Dataset102_TotalSpineSeg_step2': 'https://github.com/neuropoly/totalspineseg/releases/download/r20251124/Dataset102_TotalSpineSeg_step2_r20251124.zip'})
NumExpr defaulting to 14 threads.
Creating temporary folder (/var/folders/qd/2g7h7921023d_rzm472bf8jm0000gn/T/sct_2025-11-26_10-29-34_sct_deepseg_5sn4wklk)
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 599, in <module>
    main(sys.argv[1:])
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_deepseg.py", line 489, in main
    im_lst, target_lst = inference.segment_non_ivadomed(
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 137, in segment_non_ivadomed
    fnames_out, targets = inference(path_img=fname_in, tmpdir=tmpdir, predictor=net, device=device, **extra_inference_kwargs)
  File "/Users/julien/code/spinalcordtoolbox/spinalcordtoolbox/deepseg/inference.py", line 394, in segment_totalspineseg
    assert installed_releases, f"No 'nnUNet/results/rYYYYMMDD' subdirectory found in {path_model}"
AssertionError: No 'nnUNet/results/rYYYYMMDD' subdirectory found in /Users/julien/code/spinalcordtoolbox/data/deepseg_models/spine
Total runtime; 5.781 seconds.
```

</details>

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -530,46 +530,39 @@ def main(argv: Sequence[str]):
...
-                output_filenames_qc = output_filenames[3::4]
-            # `-step1-only: 0`: Use the 5th image ([4]) which represents the step2 output
-            else:
-                assert len(output_filenames) == 5 * len(input_filenames)
-                output_filenames_qc = output_filenames[4::5]
-            iterator = zip(input_filenames, output_filenames_qc, [None] * len(input_filenames), qc_seg)
+        # Special case: spine which outputs 2-4 files per 1 input file
+        elif arguments.task == 'spine':
+            assert len(output_filenames) == 2 * len(input_filenames)
+            iterator = zip(input_filenames * 2, output_filenames, [None] * 2, qc_seg * 2)
```
joshuacwnewton, line 565: Right now, since we use `output_filenames` directly, this will produce 2 QC reports per input (one for the disc labels and one for the "full" output).

The "full" output looks normal, but the discs QC report is nearly impossible to see:

Image

To preserve the old output ("full" only), we just need to make this one change:

```diff
-output_filenames
+output_filenames[1::2]
```

If we want to provide a useful discs QC report, then maybe we can reuse one of the existing label QC reports (e.g. `sct_label_utils`)... 

But maybe we should save that for a follow-up PR, since this PR has been waiting long enough?

documentation/source/user_section/command-line/sct_deepseg.rst
```diff
@@ -125,7 +125,7 @@ Other structures
 .. |rootlets| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/spinal_rootlets_t2.png
    :target: deepseg/rootlets.html
 
-.. |totalspineseg| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/totalspineseg.png
+.. |spine| image:: https://raw.githubusercontent.com/spinalcordtoolbox/doc-figures/master/sct_deepseg/totalspineseg.png
    :target: deepseg/totalspineseg.html
```
joshuacwnewton: One more, hehe 😅 


```suggestion
   :target: deepseg/spine.html
```

spinalcordtoolbox/scripts/sct_deepseg.py
```diff
@@ -298,13 +298,13 @@ def get_parser(subparser_to_return=None):
...
             params.add_argument(
-                "-step1-only",
+                "-label-vert",
                 type=int,
-                help="If set to '1', only Step 1 will be performed. If not provided, both steps will be run.\n"
-                     "- Step 1: Segments the spinal cord, spinal canal, vertebrae, and intervertebral discs (IVDs). Labels the IVDs, but vertebrae are left unlabeled.\n"
-                     "- Step 2: Fine-tunes the segmentation, applies labels to vertebrae, and segments the sacrum if present.\n"
+                help="If set to '1', run a second model that applies a unique label to each individual vertebrae in "
+                     "the spine segmentation (e.g. C1: 11 C2: 12 etc.). "
+                     "If not specified, all segmented vertebrae will have the same value (50), which is faster to compute if you only need e.g. disc labels."
```
joshuacwnewton: Nit: Missing space


```suggestion
                     "If not specified, all segmented vertebrae will have the same value (50), which is faster to compute if you only need e.g. disc labels. "
```

spinalcordtoolbox/scripts/sct_deepseg.py
NathanMolinier: The reason I kept this is because I wanted to integrate @mguaypaq recent changes to this PR and include a sagittal QC for both the segmentation output and the discs levels.

I think including this is definitely something important.

NathanMolinier:
It should not take more time for the inference... It is strange

But we do generate two QC right now

jcohenadad:
OK, I re-did:
T1w: Total runtime; 780.815 seconds. --> not so good
T2w: Total runtime; 115.205 seconds. --> good

jcohenadad:
> But we do generate two QC right now

right, but they are currently useless since they are axial:

image

NathanMolinier:
> right, but they are currently useless since they are axial:

The type of QC we want is not yet available in qc2.py which is used by `sct_deepseg`. So I talked with @mguaypaq and @joshuacwnewton, and we decided that the best approach would be to merge this PR then update https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5050 to integrate the sagittal QC in this other PR.

PR #5115, opened 2025-12-04
Fix typos in MS lesion tutorial (`ms_lesion` -> `lesion_ms`)

mguaypaq:
There was a typo, but also the model is now called `lesion_ms` instead of `ms_lesion`. Note that I didn't change the link to the picture `ms_lesion.png`, since that is presumably still the right file name.


mguaypaq:
For quick review, here's a link to the generated docs page:
https://spinalcordtoolbox--5115.org.readthedocs.build/5115/user_section/tutorials/lesion-analysis/lesion-segmentation-ms.html

PR #5129, opened 2025-12-19
Fix excessive RAM usage and slow processing in `sct_process_segmentation` by resampling after cropping

### LINKED ISSUE #5120: Large RAM usage and very slow processing for sct_process_segmentation in v7.2
### Description

In v7.2, `sct_process_segmentation` has slowed down significantly and uses a very large amount of RAM (>20GB). This is for a 3D T1-weighted brain scan (1mm isotropic, 240x256x176 voxels). This seems to stem from commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/314f25c973e31c2cb1cdc9107f42c25cfb35df0b. Specifically, L85 where the in-plane resolution is increased to 0.1x0.1mm. This is a 100x increase in resolution from 1mm isotropic on an already full-head image. It might be better to consider a coarser resolution for this step (say 0.5x0.5mm?).

This may be related to #5108. In my testing of this image on v7.2, TSS ran fine (even on one core) in just a few minutes. 

### Steps to Reproduce

0. Install SCT 7.2
1. Run `sct_deepseg spine` and `sct_deepseg spinalcord` on 3D T1-weighted brain
2. Run `sct_process_segmentation` using spinal cord segmentation and discs from TSS.

**Expected behavior:** Quick results and manageable RAM usage

**Actual behavior:** Slow (>3 min) and >20GB of RAM

sandrinebedard:
## Checklist

## Description

This PR moves resampling to perform it on cropped 2D patches  instead of on original segmentation, which is much faster.

### TODO
* update values in PAM50 normalized metrics with new AP and symmetry metrics: https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/issues/29

## Linked issues

Fixes #5120


sandrinebedard:
@jcohenadad @joshuacwnewton @valosekj Looks like the resampling was the problem!

When testing on the T2 from the course data, it went from 56 s to 2.3 s to run the command!

@joshuacwnewton however, with the method I am using for resampling, the CSA values differ quite a lot, I was wondering if you had a better solution, or if I was missing something! (see the tests)

spinalcordtoolbox/process_seg.py
```diff
@@ -300,7 +300,18 @@ def _properties2d(seg, dim, iz, angle_hog=None, verbose=1):
...
+    seg_crop_r = zoom(seg_crop, zoom=zoom_factors, order=1)  # make pixel size isotropic
+    regions = measure.regionprops(np.array(seg_crop_r > 0.5, dtype='uint8'), intensity_image=seg_crop_r)
+    region = regions[0]
+    minx, miny, maxx, maxy = region.bbox
+    # Use those bounding box coordinates to crop the segmentation (for faster processing), again as zoom adds padding
+    seg_crop_r = seg_crop_r[np.clip(minx-pad, 0, seg_crop_r.shape[0]): np.clip(maxx+pad, 0, seg_crop_r.shape[0]),
+                            np.clip(miny-pad, 0, seg_crop_r.shape[1]): np.clip(maxy+pad, 0, seg_crop_r.shape[1])]
+    # Update dim to isotropic pixel size
+    dim = [0.1, 0.1]
+    # seg_crop_r = seg_crop
```
valosekj: I'm curious if we could use `sct_crop_image` here. The SC seg (`-m`) can be used to crop itself and `-dilate` makes sure we have some padding after cropping.

```console
sct_crop_image -i sub-amu01_T2star_label-SC_seg.nii.gz -m sub-amu01_T2star_label-SC_seg.nii.gz -dilate 3 -o sub-amu01_T2star_label-SC_seg_crop_dil3.nii.gz
```

spinalcordtoolbox/process_seg.py
```diff
@@ -300,7 +300,18 @@ def _properties2d(seg, dim, iz, angle_hog=None, verbose=1):
     # Use those bounding box coordinates to crop the segmentation (for faster processing)
     seg_crop = seg_norm[np.clip(minx-pad, 0, seg_bin.shape[0]): np.clip(maxx+pad, 0, seg_bin.shape[0]),
                         np.clip(miny-pad, 0, seg_bin.shape[1]): np.clip(maxy+pad, 0, seg_bin.shape[1])]
-    seg_crop_r = seg_crop
+    # Apply resampling to the cropped segmentation:
+    zoom_factors = (dim[0]/0.1, dim[1]/0.1)
+    seg_crop_r = zoom(seg_crop, zoom=zoom_factors, order=1)  # make pixel size isotropic
```
mguaypaq: I think the large change is due to this:
```suggestion
    seg_crop_r = zoom(seg_crop, zoom=zoom_factors, order=1, mode='grid-constant', grid_mode=True)  # make pixel size isotropic
```

According to [the `zoom` docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.zoom.html#zoom), the default `grid_mode=False` essentially chops off one half-voxel from each edge of the data array when computing distances, and this has a large effect because the cropped segmentation can be quite small.

In the case of `sct_example_data/t2/t2_seg.nii.gz`:
- a cropped slice has shape (16, 13)
- after resampling, this is increased to (128, 104)
- with `grid_mode=False`, the zoom factors applied are 127/15 = 8.467 in the x-axis and 103/12 = 8.583 in the y-axis, which is wrong
- but with `grid_mode=True`, the zoom factors are 128/16 = 8.000 and 104/13 = 8.000, which is correct.

I think the default for all `scipy.ndimage` resampling functions is to assume that array values are pinpoint-like and correspond to a corner of the pixel/voxel (this is what `mode={constant,mirror,wrap}` all do). But for nifti images, the array values correspond to the center of the voxel, and are considered to take up the whole voxel (which is what `mode={grid-constant,grid-mirror,grid-wrap}` all do).

(This is why I recently used `mode='grid-constant'` [in the QC code](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/9fff8f26b9b1f427bd7afe156e38c7845639e186/spinalcordtoolbox/reports/qc2.py#L473-L475).)

I tried out this change for the `t2/csa.csv[MEAN(area)]` values, and got:
- 60.4292865764842 on master
- 67.13273546108884 on this PR without `grid_mode`
- 60.50923891297235 on this PR with `grid_mode=True`.

mguaypaq (commented):
This is what I found regarding the large change in mean CSA values with this PR.

testing/cli/test_cli_sct_process_segmentation.py
joshuacwnewton, line 1: > ```
> =========================== short test summary info ============================
> FAILED testing/api/test_process_seg.py::test_compute_shape[im_seg9-expected9-params9] - assert 73.10825234038393 == 77 ± 3.85
>   
>   comparison failed
>   Obtained: 73.10825234038393
>   Expected: 77 ± 3.85
> ```

The failing test case is:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/314f25c973e31c2cb1cdc9107f42c25cfb35df0b/testing/api/test_process_seg.py#L211-L214

The range (`± 3.85`) is quite large!

- **Expected**: `73.15 <= x <= 80.85`
- **master**: `73.19`
- **This PR**: `73.11`

So, actually, the old values were already very close to the boundaries of expected values, and this PR just barely nudged it past the boundary. And, the difference in area is a tiny fraction of a single voxel!

Meaning, I don't think this PR has caused any new issues. Rather, I think there may have been some preexisting issue, either with the CSA calculation or with the test.

joshuacwnewton: Here is the "dummy segmentation":

Image

With default radii of `(3.0, 5.0)` the final rectangular shape prior to rotation is `(7.0, 11.0)` hence the expected CSA value of `77`. So that makes sense! Why, then, is the actual value significantly lower?

I dug into the code, and here is a sample patch after angle correction:

Image

The diameter is still exactly `11.0` in the y axis, and `8` (with soft edges), as expected. Summing this gives `76.73`, which is close to `77`. However, in `_properties2d`, the seg is zoomed, summed, and scaled back:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/438aeef6200ae3354841e935d0bd62baa570fb2c/spinalcordtoolbox/process_seg.py#L317-L318

This results in a value of `73.07`. This is a big difference! I'm surprised that upscaling results in such a big change to the area. And, this behavior is present both on `master` and on this branch.

joshuacwnewton: This touches the same code involved in @mguaypaq's investigation in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5129#discussion_r2706010351.

While `mode='grid-constant'` does keep the current values in line with the `master` values, is there a chance that the `master` values were incorrect to begin with? I feel like maybe there's something I don't fully understand yet... shouldn't resampling largely preserve the area of the original smaller array?

joshuacwnewton (commented):
I've identified the code responsible for the failing test case, but I don't yet understand whether this is expected behavior or unexpected behavior. (Hopefully @mguaypaq can help?)

Either way, I don't think this PR is to blame:

> - **Expected**: `73.15 <= x <= 80.85`
> - **master**: `73.19`
> - **This PR**: `73.11`

So, I _think_ this is either a larger problem with the tests or with the code itself, and was present on `master` before this PR?

testing/cli/test_cli_sct_process_segmentation.py
sandrinebedard: hmm it should preserve the area indeed... thanks for the investigations

sandrinebedard: I don't know what should be the next steps for this, as Julien mentionned it was a pretty urgent PR (heavy memory usage on the master currently)

mguaypaq: Sorry, I looked at https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5151 first, but now I'm looking at this.

mguaypaq: The only immediate use of an `scipy.ndimage` function I could find which might need a different `mode` parameter (though I'm not sure) is this:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/25485f40f6255013fafdf41cc5ba44b184c5184a/spinalcordtoolbox/process_seg.py#L425-L428

Given that it's not in the code touched by this PR, maybe we could just merge this and continue the investigation in a new issue.

spinalcordtoolbox/process_seg.py
```diff
@@ -300,7 +293,17 @@ def _properties2d(seg, dim, iz, angle_hog=None, verbose=1):
...
-    seg_crop_r = seg_crop
+    # Apply resampling to the cropped segmentation:
+    zoom_factors = (dim[0]/pr, dim[1]/pr)
+    seg_crop_r = zoom(seg_crop, zoom=zoom_factors, order=1, mode='grid-constant', grid_mode=True)  # make pixel size isotropic
+    regions = measure.regionprops(np.array(seg_crop_r > 0.5, dtype='uint8'), intensity_image=seg_crop_r)
+    region = regions[0]
+    minx, miny, maxx, maxy = region.bbox
+    # Use those bounding box coordinates to crop the segmentation (for faster processing), again as zoom adds padding
+    seg_crop_r = seg_crop_r[np.clip(minx-pad, 0, seg_crop_r.shape[0]): np.clip(maxx+pad, 0, seg_crop_r.shape[0]),
+                            np.clip(miny-pad, 0, seg_crop_r.shape[1]): np.clip(maxy+pad, 0, seg_crop_r.shape[1])]
```
mguaypaq: The problem seems to be here, in this second cropping operation:

With the first slice of the dummy segmentation from `test_process_seg.py::test_compute_shape[9]`, after angle correction:
- The expected area is 77.0
- `seg.sum()` is 76.73257527737718, which is close enough.
- `seg_crop.sum()` is the same, which shows that 3 voxels was enough padding (but, this is arbitrary, because the input could have any resolution).
- `seg_crop_r.sum()` is 7673.257527737718 before re-cropping, which is expected after a 10x by 10x scaling.
- `seg_crop_r.sum()` is 7263.9206940180875 after re-cropping, because now 3 voxels is only 0.3mm of padding around the region which is binarized with a threshold of 0.5.

This leaves out a large region of voxels which have actual values between 0.0 and 0.5, and which should contribute to the CSA computation (and others!).

(BTW, this test case claims to `test uint8 input`, but it doesn't, because the `dummy_segmentation()` function ignores the `dtype` argument. There are also arguably incorrect uses of `break` and `is` comparisons in this test file.)

Some possible solutions:
- Don't do a second cropping, because the first one is probably enough.
- Or at least, pad with much more than 0.3mm on each side.
- Lower the binarization threshold from 0.5 to something much smaller (this might be a good idea for the first cropping as well).

testing/cli/test_cli_sct_process_segmentation.py
joshuacwnewton: > Given that it's not in the code touched by this PR, maybe we could just merge this and continue the investigation in a new issue.

Thank you @mguaypaq for your excellent investigation! Since we know that `seg_crop_r` is the issue (and the code is introduced by this PR), I'm going to collapse this thread and continue discussion in the other thread.

spinalcordtoolbox/process_seg.py
mguaypaq: Instead of binarizing with >0.5, followed by some arbitrary amount of padding, how about simply binarizing with >0.0? Even for a soft segmentation, I would expect that most of the image is literally 0.0, so we would still be cropping to a reasonable bounding box. And this way, we wouldn't have to guess about how gradual the slope is between 0.0 and 1.0 around the edges of the segmentation.

We could do this for the first cropping, and skip the second cropping entirely, I think.

sandrinebedard: > We could do this for the first cropping, and skip the second cropping entirely, I think.

agreed!

joshuacwnewton: > - `seg_crop_r.sum()` is 7673.257527737718 before re-cropping, which is expected after a 10x by 10x scaling.
> - `seg_crop_r.sum()` is 7263.9206940180875 after re-cropping, because now 3 voxels is only 0.3mm of padding around the region which is binarized with a threshold of 0.5.

This makes sense, but I'm confused about one thing: If the "second cropping" step is the issue, but the second cropping was only introduced in this PR, then in theory `master` should be OK. Yet, [on `master`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5129#pullrequestreview-3741269288),  the final output CSA is also ~73 instead of the expected ~76-77.

----

To try to explain why the `master` values were off... I think the cause was running angle correction on the resampled data:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/438aeef6200ae3354841e935d0bd62baa570fb2c/spinalcordtoolbox/process_seg.py#L149-L156

On `master`:

1. `im_seg.data[:, :, 0].sum()`: `79.62371`
2. [_[resampling occurs in `compute_shape`]_](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/438aeef6200ae3354841e935d0bd62baa570fb2c/spinalcordtoolbox/process_seg.py#L102-L104)
3. `current_patch.sum()`: `7962.371124901809`
4. `current_patch_scaled.sum()`: `7239.801100522209` (post-angle correction)

For this PR (with the 0.1mm resampling step done after angle correction):

1. `im_seg.data[:, :, 0].sum()`: `79.62371`
2. _[resampling no longer occurs in `compute_shape`]_
3. `current_patch.sum()`: `79.62371110916138`
4. `current_patch_scaled.sum()`: `76.73257527737718` (post-angle correction)

Conclusion: On `master`, performing angle correction on the resampled data causes a markedly different `sum()` value (`79*100` -> `~72.3*100`) compared to performing angle correction on the original array (`79` -> `~76.7`).

sandrinebedard: I was not suprised that changing the resampling method would change a little bit the CSA values, but not as much! and combined with 2nd cropping...

Do you think that computing the angle on resampled vs not resample segmentation could also affect this?

joshuacwnewton: > Do you think that computing the angle on resampled vs not resample segmentation could also affect this?

Yes exactly, I think that's what we might be seeing in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5129#discussion_r2800347653!

To be sure, it would be worth checking the intermediate values on the commit immediately prior to when we added the 0.1mm resampling. If values are OK prior to the 0.1mm resampling, then we may need to issue a patch to v7.2. :sweat:

mguaypaq: Another hint that maybe the angle correction on `master` wasn't being done properly:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/ff585b8745bf2f1d4e3a125b37532f4205509457/testing/api/test_process_seg.py#L211-L214

In the test case we've been discussing, we construct the dummy spinal cord with a `angle_RL=15`, then the expected output is... `'angle_RL': 13.269097801736525`???
- Indeed, on `master`, I get `metrics['angle_RL'].data.mean() == 13.269097801736525`, exactly the value written in the test.
- And on this branch, I get `metrics['angle_RL'].data.mean() == 15.028552392799979`, which is much closer to the original `angle_RL=15`.

Now, such a small angle difference should only cause a ~1% difference in angle-corrected values, but we're seeing a much bigger difference. But oh dear, let's look at the slice-by-slice values, rather than the `.mean()`:
- On `master`, yikes, it's all over the place, from -25 to +32 degrees:

  ```
  >>> np.array(metrics['angle_RL'].data)
  array([-24.60424949,  19.33703368,  30.80719201,  13.38285558,
          -5.90421147,  14.85430127,  29.87204105,  13.33214154,
          12.06558092,  14.85227601,  14.85982453,  12.07523457,
          13.28101297,  29.92135003,  15.01186791,  -6.6842042 ,
          14.28768529,  32.03365411,  10.98841216,  -2.98774659,
          14.81088193,  28.10788388,  14.1814994 ,  10.60555795,
          30.41439363,  15.11655464,  -6.46093663,  13.8711793 ,
          31.45875076,  14.11489347,  -6.65348617,  14.43951816,
          31.98799651,  10.96719884,  -2.95567478,  14.8014976 ,
          28.10511557,  14.18533019,  10.60254562,  30.41968909,
          15.10882945,  -6.47472782,  13.93642802,  31.40092815,
          13.91063982,  -5.70582434,  13.36091129,  30.7329002 ,
          19.65922626, -25.37686178])
  ```
- On this branch, it's consistently ~15 degrees:

  ```
  >>> np.array(metrics['angle_RL'].data)
  array([15.21185276, 15.20026917, 15.18894118, 15.17786888, 15.16705233,
         15.15649164, 15.14618686, 15.13613808, 15.12634538, 15.11680881,
         15.10752846, 15.09850439, 15.08973667, 15.08122535, 15.07297051,
         15.06497221, 15.05723049, 15.04974542, 15.04251705, 15.03554544,
         15.02883063, 15.02237268, 15.01617162, 15.01022751, 15.00454039,
         14.9991103 , 14.99393728, 14.98902136, 14.98436259, 14.97996099,
         14.9758166 , 14.97192945, 14.96829956, 14.96492696, 14.96181168,
         14.95895374, 14.95635315, 14.95400994, 14.95192413, 14.95009573,
         14.94852475, 14.9472112 , 14.9461551 , 14.94535645, 14.94481525,
         14.94453152, 14.94450525, 14.94473644, 14.94522509, 14.9459712 ])
  ```

joshuacwnewton: So I guess the plan of action is:

- Fix the "second cropping" issue as recommended above
- Tighten up the tests to no longer have a large range of acceptable values.
- Backport the fixes to v7.2 and possibly issue an announcement warning people that they should re-install v7.2 to avoid faulty CSA values? 
- Or, potentially even creating a v7.2.1 release/yanking the v7.2 release so that people know not to use the existing v7.2 release?

mguaypaq: Re: the bad CSA values, it may be that this artificial example is worse than the more typical inputs people would use (or not). We may want to take a more representative image, check the detected angle values on `master` vs this branch, and see how bad things really are.

joshuacwnewton: On it!

joshuacwnewton: Uh-oh. I'm seeing the same wildly varying angle values on our tutorial data: https://www.diffchecker.com/79Gom7Bh/

joshuacwnewton: OK. Given that we've yanked 7.2 and are recommending 7.1 for now, I wanted to make sure that 7.1 (prior to the introduction of the 0.1mm resampling) still produces appropriate values.

As a quick check (not modifying the [previously-identified](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5129#discussion_r2817849473) faulty ground truth values in the test cases yet), I tightened up the `process_seg` test as follows:

<details>

```python
@pytest.mark.parametrize('im_seg,expected,params', im_segs)
def test_compute_shape(im_seg, expected, params):
    metrics, fit_results = process_seg.compute_shape(im_seg,
                                                     angle_correction=params['angle_corr'],
                                                     param_centerline=ParamCenterline(),
                                                     verbose=VERBOSE)

    # e.g. for an area of 77, a tolerance of 0.01 means that the test will pass for the range [76.23, 77.77]
    rel = 0.01
    for key in expected.keys():
        # when computing length, we need to sum the values across all slices
        if key == 'length':
            assert metrics[key].data.sum() == pytest.approx(expected[key], rel=rel)

        # for all other metrics, we'll compare expected values slicewise basis
        else:
            for n_slice in [params['slice']] if 'slice' in params else range(im_seg.data.shape[2]):
                # fetch expected/obtained values
                expected_value = expected[key]
                obtained_value = metrics[key].data[n_slice]

                # compare depending on type
                if expected_value is np.nan:
                    assert math.isnan(obtained_value)
                else:
                    assert obtained_value == pytest.approx(expected_value, rel=rel)
```

</details>

> _(tl;dr: 0.01 relative tolerance instead of 0.05, testing each slice individually rather than the mean of all slices)_

----

The results on 7.1 are much better! Note: There are some failures for my "strict test" here, but I think maybe they're due to the input data as opposed to faulty calculations.

- **`area`**
   - ✅ **Straight cord**: On 7.1, the values are exactly as we'd expect, with no tolerance needed (e.g. exactly `77.0` instead of e.g. `76.24...`).
   - ⚠️ **Angled cord**: On 7.1, the values are OK but don't 100% fit the tight tolerance. I believe this is due to inherent limitations in generating an angled cord at low resolutions.
     <details>
     
     * For example, in the sixth text, we generate a seg with `angle_RL=-30` with expected area `197` (confirmed to be accurate on straight cord). The estimated centerline from `process_seg` looks OK, too:
       image
     * We get somewhat reasonable angle values, but not so reasonable angle-corrected area values (left area: angle-corrected, right area: uncorrected):
       image imageimage
     * I struggled to identify the reason for the outlier `199` value here, given the reasonable uncorrected area + the reasonable angle (relative to the other slices).
     * I can rule out issues with `_properties2d` -- the `199` value appears directly after the application of the `-30` affine transform [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b9c511d01e05d6abb9c01a0b2ae283036275ed3d/spinalcordtoolbox/process_seg.py#L116-L120).
     * So, I think this might just be inherent variability from a dummy angled cord at low resolutions?



     </details>

- **`diameter`**
   - ❌ **Straight cord**: On 7.1, some of the computed diameter values are off, due to how the elliptical dummy segmentations are implemented:
     <details>
  
     * For example, in the fourth test, we use an elliptical cross-section with `radius_AP=5.0`, expected value `10.0`. The result is `9.6623`, failing the `0.01` rel tolerance.
     * `10` is not the right value to use for GT anyway; a radius of 5.0 should produce an ellipse of diameter _11.0_ (center voxel `+/- 5`).
     * This ellipse is sort of a funky shape (with a single-voxel pokey bit for the widest row) -- NB: we ran into unintuitive ellipse generation before in a different context, will have to revisit this.
       image
     * The `regionprops` call fits a much tigher ellipse, resulting in the smaller diameter. (I confirmed that the upsampling/binarization do not have a large impact on the diameter, as it's about 9.6 before/after the upsampling/binarization.)
  
     </details>

- **`angle_AP` and `angle_RL`**:

   - ⚠️ **Angled cord**: On 7.1, the values are mostly OK but don't 100% fit the tight tolerance. I believe this is due to inherent limitations in generating an angled cord at low resolutions.
 
     <details>
  
     * For example, in the seventh test, we generate a seg with `angle_AP=20` with expected area `197` (confirmed to be accurate on straight cord). The estimated centerline from `process_seg` looks mostly OK, too:
       image
     * The angle values are not so good at the extremes of the cord seg:
       image
     * I can imagine that this is due to the coarse resolution of the centerline, as there are not enough voxels to accurately express a relatively small angle.

     </details>

I would love to have a way of estimating angles that isn't sensitive to low-resolution cords, but the point is that there doesn't seem to actually be any issues with the angle correction logic itself on `sct_process_segmentation` on v7.1 -- these are all expected values given the nature of the inputs and the way we're estimating the centerline.

I'll try to see if I can come up with a modified test suite that A) generates sensible dummy input segs and B) checks expected values within a tight tolerance on a per-slice basis while also accounting for expected angle variability at low resolutions.

I'll push a test to this branch (aiming for Monday), and then we can apply the fixes and make sure everything still looks good (or perhaps better!) vs. 7.1.

joshuacwnewton: > I'll push a test to this branch (aiming for Monday), and then we can apply the fixes and make sure everything still looks good (or perhaps better!) vs. 7.1.

This ended up being a lot more involved than I anticipated, but it's on this branch now! Now we can try to update the code to get this test to pass. :)

sandrinebedard:
@joshuacwnewton @mguaypaq I have now removed the 2nd cropping, and change the binarization threshold to get the bounding box for cropping. But it doesn't look like the test is passing 😢 , did I miss something, or this was the change we wanted right?

testing/cli/test_cli_sct_compute_compression.py
mguaypaq, line 55: I know it wasn't touched by this PR, but I just want to point out that these two lines are weird: the second one exactly overwrites the first.

spinalcordtoolbox/process_seg.py
mguaypaq, line 321: Today's progress on fixing `test_sct_compute_compression_no_normalization`:
- At this line, the new `process_seg.py` code correctly computes this `diameter_AP` as 17.29... (compared to [the theoretical value](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5180/changes#diff-fa4012647ec13d4105ffac0aab62eaa705c5a9faa6ac38e0080bdd9a1a4fca24R250) of $15 \cdot 2/\sqrt{3} = 17.32\ldots$)
- But then the `diameter_AP` used/reported by `sct_compute_compression` is 19.62...?!?
- With the code from `master`, `sct_compute_compression` uses/reports 14.91..., which is also wrong.

I'm not sure yet why `sct_compute_compression` uses the wrong value, but that's what I'll look into next.

(Also, I dug into `measure.regionprops` and found that we were arguably using it wrong before, but those problems are effectively prevented since the current code uses isotropic resampling.)

sandrinebedard: but now the AP is not the one used from `region.minor_axis_length`, we directly sum the voxels instead (changes introduced in 7.2):
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1a608486b84a46e25bf8e452f699bc2af67b13fe/spinalcordtoolbox/process_seg.py#L344-L349

mguaypaq: Ok, thanks to your comment I was able to figure out what's happening. TLDR: we should update the dummy segmentation used by the tests, because it's too symmetric. I will add commits to do this.

- The tests for `sct_compute_compression` look at the AP diameter through a metric ratio (in other words, "how many percentage points is the compressed slice less than the surrounding slices"):

  https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2db6a78be189940517fc9f89329622b32c7875cc/spinalcordtoolbox/scripts/sct_compute_compression.py#L516-L525

- The dummy segmentation consists of a vertical column with a rectangular cross-section.
  - For most slices, the cross-section is a 15x15 square (voxels, at a 1mm resolution).
  - The single "compressed" slice is a 12x15 rectangle (with 2 rows of voxels shaved off in the A direction, and 1 row of voxels shaved off in the P direction).

- On `master`:
  - The regular slices have a detected angle of 90 degrees (so, the AP diameter is actually computed in the LR direction). The resulting length is 14.91, which is close to the true value of 15 (but, accidentally).
  - The compressed slice correctly has a detected angle of 0 degrees. The resulting length is 11.96, which is close to the true value of 12.
  - The computed percentage is $1 - 11.96/14.91 = 19.79\\%$, which is close to the expected value of $1 - 12/15 = 20\\%$.

- On this branch:
  - The regular slices have a detected angle of 45 degrees (so, the AP diameter is actually computed along the diagonal of the square). The resulting length is 19.6. You might expect a length of $15\cdot\sqrt{2} = 21.2$, but you have to subtract 1.5 to account for the averaging of the AP measurement across 3mm. Anyway, the test expects 15, which is very far from this.
  - The compressed slice is correctly close to 12.
  - The computed percentage is $1 - 12/19.6 = 38.78\\%$, very far from the expected value.

- The real problem is that the square slices are so symmetric that the detected angle can easily be 0, 45, 90, etc. degrees, and this is sensitive to small changes in resampling.

- I suggest changing the dummy spinal cord mask to remove one layer of voxels on each of the A and P sides, so that it's wider in the LR direction than the AP direction, as expected by the angle detection code.
  - When I tried doing this, the `process_seg` code reliably detected 0 degrees of rotation on all the slices, and the `sct_compute_compression` code correctly detected a ratio of $1 - 10/13 = 23.08\\%$.

mguaypaq: With my added commits, I've updated the test and the expected results for AP diameter (with a more reasonable tolerance).

There is still a failure, for `area_ratio`:

```
___ test_sct_compute_compression ___
  ...
AssertionError: assert 23.076883697405737 == 20.05342163028501 ± 2.0e-05
  
  comparison failed
  Obtained: 23.076883697405737
  Expected: 20.05342163028501 ± 2.0e-05
```

(I'll investigate this difference next week.)

testing/cli/test_cli_sct_compute_compression.py
sandrinebedard: probably a mistake

spinalcordtoolbox/process_seg.py
mguaypaq: Ok, this was actually very easy:
- The failure is in `area_ratio`.
- When I modified the dummy segmentation to make it rectangle, instead of square, I changed the area of the "compressed" slice and the surrounding slices, so this is expected.
- The old expected value corresponds to the old area ratio:
  - Compressed slice area is $12 \times 15 = 180$.
  - Surrounding slice area is $15 \times 15 = 225$.
  - Expected ratio is then $1 - 180/225 = 20.00\\%$.
- The new obtained value fits with the new dummy CSA values:
  - Compressed slice area is $10 \times 15 = 150$.
  - Surrounding slice are is $13 \times 15 = 195$.
  - Expected ratio is then $1 - 150/195 = 23.08\\%$.
- I'll push a commit to update the values and tolerance for this test.

testing/cli/test_cli_sct_compute_compression.py
mguaypaq: I think this will be fixed when we update the tests to use `-discfile` instead of `-vertfile` (following the deprecation in #4948), so I'm gonna say this is out of scope for this PR.

spinalcordtoolbox/process_seg.py
mguaypaq: Ok, the tests should now pass with the updated values and tolerances. After that, I'm happy to merge this PR. Sorry for the wait!

mguaypaq (approved):
I'm reasonably happy with the updated tests.

Issue #5107, opened 2025-11-25
Stitching crops to smallest FOV


sandrinebedard:
### Description

When stitching acquisition with different FOVs in the right-left direction, `sct_image -stitching` uses the common FOV

Here is the results from SCT stiching vs. the Original FOV:

<img width="228" height="658" alt="Image" src="https://github.com/user-attachments/assets/62154a1e-c20f-4035-a13d-2205524b8917" />
<img width="228" height="658" alt="Image" src="https://github.com/user-attachments/assets/5d1b136e-a044-47d5-9c4e-c210c21c8680" />

### Possible solution

Consider adding padding before stitching

joshuacwnewton:
Oops: Turns out there was a duplicate issue for this the entire time, with previous discussion about proposed fixes:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3983

PR #5151, opened 2026-01-23
Improve description of stitching in `sct_image` to avoid cropping

sandrinebedard:
## Checklist

## Description

This PR adresses #5107, mainly comment [spinalcordtoolbox/spinalcordtoolbox#5107 (comment)](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5107#issuecomment-3770326786). It specifues that the padding is done according to the first provided image.
Other phrasing could potentially be used too. Feel free to edit!

## Linked issues

Partially addresses #5107. (A complete fix will require reimplementing the stitching code.)

### DIFF
--- a/spinalcordtoolbox/scripts/sct_image.py
+++ b/spinalcordtoolbox/scripts/sct_image.py
@@ -71,7 +71,10 @@ def get_parser():
         '-stitch',
         action='store_true',
         help='Stitch multiple images acquired in the same orientation utilizing '
-             'the algorithm by Lavdas, Glocker et al. (https://doi.org/10.1016/j.crad.2019.01.012).')
+             'the algorithm by Lavdas, Glocker et al. (https://doi.org/10.1016/j.crad.2019.01.012).'
+             'The first image is used as the reference.'
+             'Ensure to use the image with the largest dimensions first to avoid cropping of the images.'
+             )
     image.add_argument(
         '-qc',
         metavar=Metavar.folder,
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1607,8 +1607,7 @@ def pad_image(im: Image, pad_x_i: int = 0, pad_x_f: int = 0, pad_y_i: int = 0, p
     # TODO: Do not copy the Image(), because the dim field and hdr.get_data_shape() will not be updated properly.
     #   better to just create a new Image() from scratch.
     im_out.data = padded_data  # done after the call of the function
-    if isinstance(im_out.absolutepath, str):
-        im_out.absolutepath = add_suffix(im_out.absolutepath, "_pad")
+    im_out.absolutepath = add_suffix(im_out.absolutepath, "_pad")
 
     # adapt the origin in the sform and qform matrix
     new_origin = np.dot(im_out.hdr.get_qform(), [-pad_x_i, -pad_y_i, -pad_z_i, 1])
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1814,9 +1814,8 @@ def stitch_images(im_list: Sequence[Image], fname_out: str = 'stitched.nii.gz', 
     # save the preprocessed images to a tempdir
     path_tmp = tmp_create(basename="stitch-images")
     fnames_in = []
-    for i, im_in_rpi in enumerate(im_list_rpi):
-        # avoid writing multiple files with the same basename to the tempdir
-        temp_file_path = add_suffix(os.path.join(path_tmp, os.path.basename(im_in_rpi.absolutepath)), f"_{i}")
+    for im_in_rpi in im_list_rpi:
+        temp_file_path = os.path.join(path_tmp, os.path.basename(im_in_rpi.absolutepath))
         im_in_rpi.save(temp_file_path, verbose=verbose)
         fnames_in.append(temp_file_path)
 
--- a/spinalcordtoolbox/image.py
+++ b/spinalcordtoolbox/image.py
@@ -1783,29 +1783,44 @@ def stitch_images(im_list: Sequence[Image], fname_out: str = 'stitched.nii.gz', 
     # preserve original orientation (we assume it's consistent among all images)
     orig_ornt = im_list[0].orientation
 
-    # reorient input files and save them to a temp directory
+    # reorient input files to RPI
+    im_list_rpi = []
+    for im_in in im_list:
+        im_list_rpi.append(change_orientation(im_in, 'RPI'))
+
+    # pad the first image based on the max x/y dimensions
+    # Explanation:
+    #     C++ stitching module by Glocker et al. uses the first image as reference image
+    #     and allocates an array (to be filled by subsequent images along the z-axis)
+    #     based on the dimensions (x,y) of the reference image.
+    #     As subsequent images are padded to the x-/y- dimensions of the reference image,
+    #     it is important to use the image with the largest dimensions as the first
+    #     argument to the input of the C++ binary, to ensure the images are not cropped.
+    shapes_in = [im.data.shape for im in im_list_rpi]
+    x_max = max(shape[0] for shape in shapes_in)
+    y_max = max(shape[1] for shape in shapes_in)
+    x_pad = x_max - shapes_in[0][0]
+    y_pad = y_max - shapes_in[0][1]
+    if (x_pad, y_pad) != (0, 0):
+        logger.info(f"Padding the axial slice dimensions of the first image from ({shapes_in[0][0]}, {shapes_in[0][1]}) "
+                    f"to ({x_max}, {y_max}) to preserve the maximum AP and RL dimensions across the images.")
+        im_list_rpi[0] = pad_image(im_list_rpi[0],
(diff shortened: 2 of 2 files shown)


spinalcordtoolbox/scripts/sct_image.py
```diff
@@ -71,7 +71,10 @@ def get_parser():
         '-stitch',
         action='store_true',
         help='Stitch multiple images acquired in the same orientation utilizing '
-             'the algorithm by Lavdas, Glocker et al. (https://doi.org/10.1016/j.crad.2019.01.012).')
+             'the algorithm by Lavdas, Glocker et al. (https://doi.org/10.1016/j.crad.2019.01.012).'
+             'The first image is used as the reference.'
+             'Ensure to use the image with the largest dimensions first to avoid cropping of the images.'
+             )
```
joshuacwnewton, line 77: Digging into the code, I noticed this line:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/679d675ba41b97f18199dedca1bf69b9128e42a7/spinalcordtoolbox/image.py#L1826-L1827

It seems like we're already trying to reorder the images on behalf of the user? But, wouldn't calling `sorted` undo any of the sorting that the user tries to do?

So, I'm surprised that reordering the images worked for you... 🤔

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton, line 1: Do we want to consider automatically padding the first image to the largest dimensions? e.g.

```python
import math
from spinalcordtoolbox.image import pad_image

# in stitch_images:

    # reorient input files to RPI
    im_list_rpi = []
    for im_in in im_list:
        im_list_rpi.append(change_orientation(im_in, 'RPI'))

    # pad the first image based on the max x/y dimensions
    shapes_in = [im.data.shape for im in im_list_rpi]
    x_max = max(shape[0] for shape in shapes_in)
    y_max = max(shape[1] for shape in shapes_in)
    x_pad = x_max - shapes_in[0][0]
    y_pad = y_max - shapes_in[0][1]
    im_list_rpi[0] = pad_image(im_list_rpi[0], 
                              pad_x_i=math.floor(x_pad/2), pad_x_f=math.ceil(x_pad/2), 
                              pad_y_i=math.floor(y_pad/2), pad_y_f=math.ceil(y_pad/2), 
                              pad_z_i=0, pad_z_f=0)

    # save the preprocessed images to a tempdir
    path_tmp = tmp_create(basename="stitch-images")
    fnames_in = []
    for im_in_rpi in im_list_rpi:
        temp_file_path = os.path.join(path_tmp, os.path.basename(im_in.absolutepath))
        im_in_rpi.save(temp_file_path, verbose=verbose)
        fnames_in.append(temp_file_path)
```

That might remove the need for reordering entirely (either from the user or from us).

joshuacwnewton (commented):
Your comment looks good! However, I'm surprised that reordering the input images has an effect? 

Based on discussion we may want to go with a different solution... 🤔

spinalcordtoolbox/scripts/sct_image.py
sandrinebedard: that would be even better actually!

joshuacwnewton: Could you try https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5151/commits/bad86256147a0d3dcd2286ad1a2687655bc33a42 on your problematic data? And, if it works okay (regardless of input order), perhaps we could remove the note about putting the largest image first?

sandrinebedard: Got this error:

~~~
Spinal Cord Toolbox (git-sb/5107-add-padding-to-stiching-091cdc3f8a826f13bbaee77aeb3141fb028b390d)

sct_image -i __T2w_protocol_cs4_0_20251118153311_13.nii.gz __T2w_protocol_cs4_1_20251118153311_14.nii.gz __T2w_protocol_cs4_2_20251118153311_17.nii.gz -stitch -o t2w_stiched_joshua.nii.gz
--

Padding the axial slice dimensions of the first image from (128, 320) to (160, 320) to preserve the maximum AP and RL dimensions across the images.
Creating temporary folder (/var/folders/tm/sck2063d2rj1jcxtvq_nqpr40000gn/T/sct_2026-01-30_11-14-08_stitch-images_hyc0fh0_)
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_image.py", line 628, in <module>
    main(sys.argv[1:])
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_image.py", line 367, in main
    im_out = [stitch_images(im_in_list)]
  File "/Users/sandrinebedard/spinalcordtoolbox/spinalcordtoolbox/image.py", line 1816, in stitch_images
    temp_file_path = os.path.join(path_tmp, os.path.basename(im_in_rpi.absolutepath))
  File "/Users/sandrinebedard/spinalcordtoolbox/python/envs/venv_sct/lib/python3.10/posixpath.py", line 142, in basename
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
Total runtime; 0.651 seconds.
~~~

joshuacwnewton: Sorry, one moment, I force-pushed and need to fix the branch. My apologies.

joshuacwnewton: Okay, it should be good to test now!

sandrinebedard: it works:

image


~~~
sct_image -i __T2w_protocol_cs4_0_20251118153311_13.nii.gz __T2w_protocol_cs4_1_20251118153311_14.nii.gz __T2w_protocol_cs4_2_20251118153311_17.nii.gz -stitch
~~~

joshuacwnewton: Wonderful!!

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton, line 77: Maybe something like:


```suggestion
             'The maximum AP and RL dimensions across all input images will be used to define the final image size.'
             )
```

Or maybe the behavior is intuitive enough that there's not really a reason to explain how it works anymore?

joshuacwnewton (approved):
Personally I think this is okay? But if this isn't urgent, maybe we could wait for @mguaypaq to arrive back from holidays just to double-check my work?

(Though, if you have work that relies on stitching that's being blocked because you're not able to automatically order the largest image first, then I'd be okay merging now.)

spinalcordtoolbox/scripts/sct_image.py
sandrinebedard: I like it! I this people will have questions, so better to say it!

mguaypaq:
Sorry for the delay.

Ok, having looked at the source code of the stitching module and done some experiments to confirm my understanding, I think there are still a few issues.

To clarify my points below, note that:
- Each voxel has _array coordinates_ (i, j, k), which are 3 integers with no units.
- It also has _world coordinates_ (x, y, z), which are 3 floating point numbers measured in (say) millimetres.
- Each image has an _affine matrix_ which can transform (i, j, k) triples into (x, y, z) triples, via a combination of:
  - independent scaling of each axis (which converts between integer steps and millimetres),
  - 3D rotation in world coordinates, and
  - translation (which brings (i, j, k) = (0, 0, 0) to the image origin coordinates).

With that in mind, it looks like:
- The stitching module uses the first image's affine matrix for the output image.
- The size of the (i, j) coordinates is not changed, so this can result in cropping the other images.
- Only the (k) coordinate is enlarged, in both directions, to cover the other images.
  - But, this computation is done in a way that only accounts for scaling and translation, _not_ 3D rotation, so this can still result in cropped images. Honestly, this seems like a bug in the stitching module.
  - This means that the first image orientation (RPI, etc.) has an impact on the final output, because it affects which axis gets expanded.
  - Actually, the orientation of other images also matters.
- The result is the field-of-view in world coordinates used for the output.

Then, it looks like:
- The stitching module correctly resamples each input image into the output field-of-view, using the full affine matrix.
  - This means that we can only really avoid cropping if we compute a min/max bounding box in world coordinates, using each image's affine matrix.
- It also keeps track of how many input images contribute to each output voxel, and divides by this number to get an average.
  - Unfortunately, this means that any padding we add to the first image will result in darker output voxels, because the padding value of 0 will be averaged in.

So the bad news is:
- Adding padding to the first image is not an innocent operation.
- Even our usual reorientation to RPI is not an innocent operation.
- We may need to do some matrix math to figure out the right bounding box.
- Passing images to the stitching module that have different rotations is often fine, but sometimes will give surprising results. Maybe we should detect this and warn users.
- There may be no way to call the stitching module which does exactly what we want.

I have one last weird idea that I'd like to try, but these are my thoughts so far.

spinalcordtoolbox/scripts/sct_image.py
joshuacwnewton: Thanks to @mguaypaq's wonderful investigation, we now know that this is not a safe operation, so I'm going to revert my changes and go with your original warning.

mguaypaq (approved):

PR #5158, opened 2026-02-04
Add workflow job to test Apptainer installation and ensure `totalspineseg` is compatible

joshuacwnewton:
## Checklist

- [ ] **PR Sidebar**: I've filled these options within the PR sidebar:
  - [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!)
  - [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone)
- [ ] **Guidelines**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [ ] **Documentation**: I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages.
- [ ] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
[...]

PR #5160, opened 2026-02-06
Improve Apptainer installation by fixing several `sct.def` bugs (install directory, task installation)

### LINKED ISSUE #5121: Copying files in site_packages breaks Apptainer functionality in 7.2
### Description
When using the totalspineseg segmentation (`sct_deepseg spine`) in v7.2, a copy is attempted in `site-packages` to move a model definition file for a trainer class. This errors out in Apptainer.

Apptainer uses a read-only file system, so copying files within the installed packages is not permitted. This one actually stems from a dependency of a dependency, but logging it here where the container is built.

The issue stems from https://github.com/neuropoly/AugLab. Particularly, the `add_trainer` function, which is now used in https://github.com/neuropoly/totalspineseg. There are two steps to correct this:

1) When building an Apptainer container that would like to use `totalspineseg`, you will have to pre-install the model using `auglab_add_nnunettrainer` as shown here: https://github.com/neuropoly/totalspineseg/blob/main/scripts/train.sh#L66. This will move the proper model file into the correct area.

2) Revert this commit https://github.com/neuropoly/AugLab/commit/d89ada00db6cd53f2b0d46094e27cc2aeafe7649, which always attempts a copy even if the file already exists.

### Steps to Reproduce

0. Install SCT 7.2 using instructions from this repository for Apptainer

1. Attempt to run `sct_deepseg spine`

**Expected behavior:** `sct_deepseg spine` runs

**Actual behavior:** Error occurs when a copy is attempted in the `site-packages` directory

joshuacwnewton:
## Description

After merging my previous PR #5158 and backporting the changes 7.2, I came across several more issues.

### 1. Installation directory

In the previous PR, I changed the Apptainer installation steps from `(1. Fetch installer 2. Install SCT 3. Symlink expected install directory to /root/sct)` to `(1. Clone repo to /root/sct 2. Install SCT)`. The idea was that we would have more control over which source files we use in the installation, and skip having to do the final symlinking step (resulting in a consistent location for the source files from the get-go).

However, I forgot that `install_sct` has this logic:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2818f57f0219b8b4981387861a091ef7ffe7fa3f/install_sct#L485-L494

Meaning, unless we explicitly request an in-place installation, `install_sct` will see `requirements-freeze.txt`, assume that we don't want an in-place installation, and install SCT to `~/sct_vX.Y` instead.

### 2. Task install syntax

Also, in the process of fixing this issue and testing locally, I realized that the `./install_sct_containered.sh` script doesn't actually install tasks due to faulty POSIX testing syntax:

```
+ [ ! spine == ] 
/.post.script: 23: [: spine: unexpected operator
```

### 3. Halting post-install steps on error

We hadn't caught the previous issues because we hadn't `set -eu`, meaning the `apptainer build` command succeeds even if the post-install steps fail.

### 4. Reinstalling `spinalcord` which is already default

We _also_ hadn't caught the previous issues because my `tests.yml` addition had been attempting to install a model that already gets installed by default. So, even though the install was failing, the resulting `sct_deepseg` test wasn't failing becaues the model is still present.

### 5. Disk Usage for GHA

Finally, this PR tries to mitigate disk usage errors on GitHub Actions by:

1. Reducing how many copies of the Apptainer image file we create when re-generating the image on `sct_deepseg` task installs.
2. Removing default packages from the GitHub runners that are unneeded for testing Apptainer.

----

This PR tries to fixes each of these issues (one commit each). I've also performed `./bundle_for_release.sh 7.2` and `./install_sct_7.2_containered.sh spine` locally to confirm everything works.

(Sorry for not catching these things on my previous reviews. :sweat:)

## Linked issues

Fixes #5121.
Amends #5158.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -548,7 +548,7 @@ jobs:
       - name: Download sample data
         run: |
           curl -sL "https://github.com/spinalcordtoolbox/sct_tutorial_data/archive/refs/tags/SCT-Course-20251208.tar.gz" | tar -z -xf -
-      - name: Run 'sct_deepseg spinalcord' (default task)
+      - name: Run 'sct_deepseg spinalcord' (installed by default)
         run: |
           apptainer exec apptainer_build/sct.sif sct_deepseg spinalcord -i sct_tutorial_data-SCT-Course-20251208/single_subject/data/t2/t2.nii.gz
       - name: Run 'sct_deepseg spine' (added on install)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -541,10 +541,10 @@ jobs:
         run: |
           cd ./apptainer_build
           ./install_sct_containered_$SAFE_REF.sh spine
-      - name: Install deepseg task post-install (with 'lesion_ms' task)
+      - name: Install deepseg task post-install (with 'gm_sc_7t_t2star' task)
         run: |
           cd ./apptainer_build
-          ./install_deepseg_task.sh lesion_ms
+          ./install_deepseg_task.sh gm_sc_7t_t2star
       - name: Download sample data
         run: |
           curl -sL "https://github.com/spinalcordtoolbox/sct_tutorial_data/archive/refs/tags/SCT-Course-20251208.tar.gz" | tar -z -xf -
@@ -554,6 +554,6 @@ jobs:
       - name: Run 'sct_deepseg spine' (added on install)
         run: |          
           apptainer exec apptainer_build/sct.sif sct_deepseg spine -i sct_tutorial_data-SCT-Course-20251208/single_subject/data/t2/t2.nii.gz
-      - name: Run 'sct_deepseg lesion_ms' (added in post-install
+      - name: Run 'sct_deepseg gm_sc_7t_t2star' (added in post-install)
         run: |          
-          apptainer exec apptainer_build/sct.sif sct_deepseg lesion_ms -i sct_tutorial_data-SCT-Course-20251208/single_subject/data/t2_lesion/t2.nii.gz
+          apptainer exec apptainer_build/sct.sif sct_deepseg gm_sc_7t_t2star -i sct_tutorial_data-SCT-Course-20251208/single_subject/data/t2s/t2s.nii.gz
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -517,6 +517,17 @@ jobs:
     name: Apptainer (Ubuntu Latest)
     runs-on: ubuntu-latest
     steps:
+      - name: Maximize build space
+        run: |
+          echo "-- Dotnet"
+          sudo rm -rf /usr/share/dotnet
+          echo "-- Android"
+          sudo rm -rf /usr/local/lib/android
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -537,20 +537,23 @@ jobs:
...
+      - name: Install deepseg task post-install (with 'lesion_ms' task)
         run: |
           cd ./apptainer_build
-          ./install_deepseg_task.sh spine
+          ./install_deepseg_task.sh lesion_ms
       - name: Download sample data
         run: |
           curl -sL "https://github.com/spinalcordtoolbox/sct_tutorial_data/archive/refs/tags/SCT-Course-20251208.tar.gz" | tar -z -xf -
-      - name: Run 'sct_deepseg spinalcord'
+      - name: Run 'sct_deepseg spinalcord' (default task)
```
SomeoneInParticular: [Minor Nit] I would mention its explicitly pre-installed instead, so we don't forget again in the future

```suggestion
      - name: Run 'sct_deepseg spinalcord' (installed by default)
```

SomeoneInParticular (dismissed):
I don't see any issues with these changes; just one remark on a comment to ensure a future dev doesn't make the same mistake in the future if/when they go to update Apptainer CI. LGTM!

UPDATE: Just saw the test failure; it seems to be due to GitHub using very tiny drives for the Ubuntu runner, which prevents the new `.sif` from being created. Looking into a fix now...

.github/workflows/tests.yml
```diff
@@ -537,20 +537,23 @@ jobs:
...
         run: |
           cd ./apptainer_build
-          ./install_sct_containered_$SAFE_REF.sh spinalcord
-      - name: Install deepseg task post-install (with 'spine' task)
+          ./install_sct_containered_$SAFE_REF.sh spine
+      - name: Install deepseg task post-install (with 'lesion_ms' task)
         run: |
           cd ./apptainer_build
-          ./install_deepseg_task.sh spine
+          ./install_deepseg_task.sh lesion_ms
```
joshuacwnewton: It looks like `lesion_ms` is one of the larger `sct_deepseg` models, hence the disk space errors:

Image

I'll follow up with a different task in this slot. I think switching to a smaller model should dodge the disk space issues.

.github/workflows/tests.yml
```diff
@@ -517,6 +517,17 @@ jobs:
...
+        run: |
+          echo "-- Dotnet"
+          sudo rm -rf /usr/share/dotnet
+          echo "-- Android"
+          sudo rm -rf /usr/local/lib/android
+          echo "-- Haskell"
+          sudo rm -rf /opt/ghc
+          echo "-- CodeQL"
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          df -h
```
SomeoneInParticular, line 530: Ridiculous we need to do this, but it seems to have worked. Cheers!

SomeoneInParticular (approved):
After the debugging session we (mostly you) had today, seems the small runner issue has been resolved. Going to run the test manually a second time _just_ to be sure, but otherwise LGTM!

PR #5166, opened 2026-02-19
Avoid crash during inference if `spine` is present in `$SCT_DIR` path 

### LINKED ISSUE #5165: Renaming `totalspineseg`'s model to `spine` results in failure when `$SCT_DIR` contains the term "spine"
User report: https://forum.spinalcordmri.org/t/error-with-sct-deepseg-spinalcord-sct-7-2/1526

Problematic line of code:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4f5dba4122e225e95c096c0a79fa9edfc16ce646/spinalcordtoolbox/deepseg/inference.py#L116-L119

In hindsight, this is a really poor check (substring + full path is way too broad), but it sufficed when the model had a much less generic name ("totalspineseg"). Changing this to `spine` causes the check to evaluate to `True` for paths such as:

```
/cerebro/cerebro1/dataset/spine_7t/spine_7t_fmri_analysis/toolboxes/sct_7.2/
```

A more strict check would be something like this instead:

```python
if os.path.basename(path_model) == "spine":
```

I think we need to backport this to v7.2, as it's a fairly critical error.

joshuacwnewton:
## Description

This PR fixes our checks for the `spine` model by:

- Using just the base folder name, and not the full path. (Fixes #5165)
- Comparing against the string listed in the `TASKS` dictionary for the `spine` task.

I've also taken this opportunity to update the model name (since both the model and the task were called `spine`, which felt error-prone to me). But, maybe this is unnecessary? (Could it cause problems? I guess devs will have to redownload the model...)

## Linked issues

Fixes #5165.

### DIFF
--- a/spinalcordtoolbox/deepseg/models.py
+++ b/spinalcordtoolbox/deepseg/models.py
@@ -221,7 +221,7 @@ MODELS = {
         "thr": None,  # Images are already binarized
         "default": False,
     },
-    "spine": {
+    "model_seg_spine_contrast_agnostic": {
          # NB: Rather than hardcoding the URLs ourselves, use the URLs from the totalspineseg package.
          # This means that when the totalspineseg package is updated, the URLs will be too, thus triggering
          # a re-installation of the model URLs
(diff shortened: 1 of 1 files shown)


mguaypaq (approved):
Thanks for the fix!
- Yes, doing a full-path substring search was dubious.
- Avoiding a collision between `MODELS.keys()` and `TASKS.keys()` is a good idea. (`spine` was the only common key.)
- Some hard-coding of the task name `spine` is fine.
- Avoiding hard-coding of the model name `model_seg_...` is good.
- I think the only downside of the model rename is that existing dev installations:
  - won't have it downloaded by default (a one-time thing)
  - will have a useless 2GB folder with the old name under `data/deepseg_models/`.

  These downsides seem very minor.

PR #5179, opened 2026-02-27
Stop using `0`/`np.nan` if lesion isn't present on both interpolation slices (fallback to 1 slice or `-nli-slice`)

### LINKED ISSUE #5123: `sct_analyze_lesion`: use spinal cord A-P diameter instead of NaN for parasagittal lesions
## Desctiption

`sct_analyze_lesion` currently returns NaN for midsagittal tissue bridges when no lesion is present on the midsagittal slice:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/3ce19094c99ea201fdc6757792e566ca3d7a9bb8/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L381-L387

This can happen for lesions located parasagittally (i.e., laterally); example:

Image

After some discussion with @dpfyffer, we think that it might be better to report the spinal cord midsagittal A-P diameter instead of NaN, as this is what would be done for manual measurements. The advantage of having a numeric value over NaN is that NaN might be a bit ambiguous -- NaN might indicate that there is no lesion at all. But the lesion is actually there, just located laterally.

### LINKED ISSUE #5184: Minimum interpolated bridge width value underestimated when lesion is only present in 1 of the 2 slices
## Description

@dpfyffer asked me by email how come the interpolated midsagittal slice is 7.29 with a corresponding dorsal bridge width of 1.61mm, whereas the dorsal bridge width is 4mm on slice 8 and 3.9mm on slice 7? Should the interpolated value (from slice 7.29) not be between those two values?

Image

## Debugging

> [!NOTE]
> Apologies for a long description 

```
sct_analyze_lesion -m sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_lesion_seg.nii.gz -s sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_sc_seg_SCIsegV2.nii.gz
```

(lesion and spinal cord segmentation masks attached below)

After some debugging (using https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/ba3beb212dc30aa73b0a667f48a47d4021611088), I think the reason might be that tissue bridges are measured independently for each sagittal slice (i.e., independently for slice 7, independently for slice 8, and independently for the midsagittal slice, which is 7.29). This is illustrated in the`sct_analyze_lesion`'s output (keeping only relevant lines):

```
Interpolated midsagittal slice (same across lesions) = 7.29

Measures on lesion #1...
...
  Midsagittal dorsal tissue bridge width: 1.61 mm (axial slice 119)
  Midsagittal ventral tissue bridge width: 0.0 mm (axial slice 119)
  Midsagittal total tissue bridge width: 1.61 mm
...
  Sagittal slice 8, Minimum dorsal tissue bridge width: 4.05 mm (axial slice 109)
  Sagittal slice 8, Minimum ventral tissue bridge width: 0.0 mm (axial slice 102)
  Sagittal slice 8, Total tissue bridge width: 4.05 mm
  Sagittal slice 7, Minimum dorsal tissue bridge width: 3.99 mm (axial slice 102)
  Sagittal slice 7, Minimum ventral tissue bridge width: 0.0 mm (axial slice 110)
  Sagittal slice 7, Total tissue bridge width: 3.99 mm
...
```

> [!NOTE]
> Notice that different axial slices (i.e., slices along the S-I axis) are used for different sagittal slices. Illustrations from FSLeyes below.

<details>
<summary>Sag slice 7: axial slice 102</summary>

Image

</details>

<details>
<summary>Sag slice 8: axial slice 109</summary>

Image

</details>

Now, the thing is that the lesion covers different S-I coverage across sagittal slices:
- sag slice 7: axial slices (S-I axis): 82-118
- sag slice 8: axial slices  (S-I axis): 91-122

In other words, there are a few axial slices (119-122) where the lesion is on the sagittal slice 8 but not on the sagittal slice 7. This strongly impacts the interpolated measurement for axial slices 119-122 in the `_interpolate_values` function (as the interpolated midsagittal slice is 7.29, so the weighting is towards sag slice 7).

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/98161abf01be1d552e7e36ae1786fc8a8bdbe90b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L1083-L1093

`interpolated_dorsal_bridge_width_mm` variable at L409

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/98161abf01be1d552e7e36ae1786fc8a8bdbe90b/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L409

Image

> [!NOTE]
> Notice low values for 119-122.

The minimum value (corresponding to axial slice 119) of 1.61mm is then selected as the dorsal bridges and saved to the user.

## Files

[sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_sc_seg_SCIsegV2.nii.gz](https://github.com/user-attachments/files/25768805/sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_sc_seg_SCIsegV2.nii.gz)
[sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_lesion_seg.nii.gz](https://github.com/user-attachments/files/25768806/sub-KW829d0a0_ses-cervical01032023_acq-sagittal_t2w_lesion_seg.nii.gz)

valosekj:
## Checklist

## Context

`sct_analyze_lesion` currently returns `NaN` for midsagittal tissue bridges when no lesion is present. However, the manual approach would use the spinal cord midsagittal A-P diameter as a proxy for tissue bridges. Additional details: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5123#issue-3718681879

## Description

This PR updates `sct_analyze_lesion` behaviour for the following two cases:
1. Lesion is present, but it is not captured on the midsagittal slice (i.e., the lesion is so-called parasagittal). With this PR, `sct_analyze_lesion` now computes the spinal cord A-P diameter. This spinal cord A-P diameter is computed from the parasagittal slice at the level of the lesion (i.e., the parasagittal lesion is still used to select slices from which the spinal cord A-P diameter is computed). This was implemented in the first two commits (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179/changes/379f97cc10096b5111976fe999903dea6f43f37b and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179/changes/ddd03b209e4ed52a643450d3a6505a4d310e0706).
2. No lesion is present (i.e., `-m lesion.nii.gz` is empty). In this case, the user needs to provide a newly introduced `-nli-slice` argument with the slice number (in the S-I axis) corresponding to the Neurological Level of Injury (NLI). Then, `sct_analyze_lesion` computes the spinal cord A-P diameter from the midsagittal slice at the slice specified by `-nli-slice`.

## Example usage

```
sct_analyze_lesion
-m sub-001_acq-sag_T2w_lesion.nii.gz   # this is an empty file with no lesion
-s sub-001_acq-sag_T2w_seg.nii.gz
-nli-slice 100   # in the S-I axis
-qc ./qc
```

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5123.
Fixes #5184.
Supersedes https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5202.


valosekj:
There are some failing tests: (EDIT: @joshuacwnewton, fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179#discussion_r3075186037)

image

Coming from this dummy lesion:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/621c69f21374bcc4708061179ae2e534402900ee/testing/cli/test_cli_sct_analyze_lesion.py#L181-L183

And this test

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/621c69f21374bcc4708061179ae2e534402900ee/testing/cli/test_cli_sct_analyze_lesion.py#L184

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -602,15 +610,19 @@ def _measure_tissue_bridges(self, im_lesion_data, p_lst, idx):
...
-                # if the lesion extends the spinal cord, the dorsal bridge is set to 0
-                if dorsal_bridge_width < 0:
-                    dorsal_bridge_width = 0
-                ventral_bridge_width = sc_indices[-1] - lesion_indices[-1]      # [-1] returns the most ventral elements
-                # if the lesion extends the spinal cord, the ventral bridge is set to 0
-                if ventral_bridge_width < 0:
-                    ventral_bridge_width = 0
+                # If lesion isn't present in this slice, then both bridge widths will just be the full width of the spinal cord.
+                if lesion_indices.size == 0:
+                    dorsal_bridge_width = ventral_bridge_width = sc_indices.size
```
valosekj: Shouldn't we divide the spinal cord AP diameter by 2 here? Because the spinal cord AP diameter should be equal to the total tissue bridges (i.e., the sum of the dorsal and ventral bridges). So to get dorsal and ventral separately, we divide by 2.

As done here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8084886f996f2793041cc9b55cd1d18d1bbc59a0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L503-L505

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -584,14 +585,21 @@ def _measure_tissue_bridges(self, im_lesion_data, p_lst, idx):
...
         # --------------------------------------
+        # Note: We want to compute the bridge widths for the full 3D bounding box around the lesion, even if there
+        #       are some place in the bounding box where the lesion is not present. (If the lesion is not present, then
+        #       the bridge width will be the full width of the spinal cord.)
+        #       The reason we do this is because, for the interpolated midsagittal bridge width calculation, we need to
+        #       average the bridge widths between two sagittal slices. And, if the lesion is present in one sagittal
+        #       slice but not the other, we still need to compute the average. For more info on this edge case, see:
+        #       https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5184
+        axial_lesion_slices = [np.unique(np.where(im_lesion_data[sagittal_slice, :, :])[1])  # 2D, dim=[AP, SI] --> [1]
+                               for sagittal_slice in sagittal_lesion_slices]
```
valosekj: I'm wondering whether we shouldn't use only the two slices used for the interpolation (`self.interpolation_slices_RPI`) instead of all slices capturing the lesion (`sagittal_lesion_slices`).

Image

valosekj: but probably not, because this happens in the `_measure_tissue_bridges` function which is supposed to measure tissue bridges for _all_ sagittal slices

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -451,9 +451,10 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx, im
             # Create a lookup series for the current axial slice
             dorsal_lookup = slice_data.set_index('sagittal_slice')['dorsal_bridge_width']
             ventral_lookup = slice_data.set_index('sagittal_slice')['ventral_bridge_width']
-            # Get widths for the two sagittal slices. If there is no bridge for given slices, use 0.
-            dorsal_bridges = [dorsal_lookup.get(sag_slice, 0) for sag_slice in self.interpolation_slices_RPI]
-            ventral_bridges = [ventral_lookup.get(sag_slice, 0) for sag_slice in self.interpolation_slices_RPI]
+            # Get widths for the two sagittal slices. If there is no lesion in one or both slices, the width will be
+            # the full spinal cord width, which will result in accurate averages between the two slices.
+            dorsal_bridges = [dorsal_lookup[sag_slice] for sag_slice in self.interpolation_slices_RPI]
+            ventral_bridges = [ventral_lookup[sag_slice] for sag_slice in self.interpolation_slices_RPI]
```
valosekj: Apologies, but I don't understand how the full spinal cord width is obtained here.

Cannot we reuse here `_compute_sc_ap_diameter_at_axial_slice`?

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8084886f996f2793041cc9b55cd1d18d1bbc59a0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L1251

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: I get where you're coming from (and I agree that ventral/dorsal bridge widths + lesion width should sum to AP diameter), but I think dividing by 2 would maybe cause incorrect results.

Take for example, the sample image from the issue:

image

Focusing on just the part of the lesion present in the background slice (light red). If we measure just the dorsal bridge widths for each slice, we might see:

bridge_width_sequence

I think that the next number in the sequence should be `? = 20` (nb: ONLY in the context of interpolation). This will probably overestimate the bridge width, but the purpose here is to try and get an reasonable average between the two interpolation slices.

If instead we divided by 2, then `? = 10`, and the interpolated bridge width might be too small. Kind of a similar situation as what we're currently facing with `? = 0`.

((This assumes the value is only used for interpolation, though. Maybe that assumption isn't valid?))

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Oh. I totally missed this method! :facepalm: 

You're right, this could most likely be reused.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: The question is ultimately "how do we interpolate the width between two slices when one of those slices doesn't have a width value?" and that's weird! I'm not sure. :thinking:

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: My apologies, I'm not 100% familiar with the literature / purpose / usage of the "minimum bridge width" metric, so I'm not sure how it's typically measured.

But presumably, the slice information outside the two slices shouldn't be relevant for a hypothetical interpolated slice in between those two slices?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: Good points!

> I think that the next number in the sequence should be `? = 20` (nb: ONLY in the context of interpolation). This will probably overestimate the bridge width, but the purpose here is to try and get an reasonable average between the two interpolation slices.

Indeed, I think this is the way to go!

spinalcordtoolbox/scripts/sct_analyze_lesion.py
valosekj: > But presumably, the slice information outside the two slices shouldn't be relevant for a hypothetical interpolated slice in between those two slices?

Correct, I think we should consider only the two slices used for the interpolation.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: Sorry it took me a little bit! Looking at your suggestion, I realize that the SC width is actually determined later in the method:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/8084886f996f2793041cc9b55cd1d18d1bbc59a0/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L613-L615

The `sc_indices` variable was already available before my changes:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/913cae045803eee41949e626e862c998e5a08b78/spinalcordtoolbox/scripts/sct_analyze_lesion.py#L618-L619

By comparison, when I check the `_compute_sc_ap_diameter_at_axial_slice` method, it turns out it computes the SC AP diameter for a given axial slice, *for the interpolated sagittal slice*. I don't think that's applicable here, since we want the per-sagittal-slice SC AP diameter (before interpolation).

Some notes:

- I think maybe the `_compute_sc_ap_diameter_at_axial_slice` method could benefit from having `_interpolated` in the name to make the distinction clearer? (i.e. the method doesn't just grab the AP diameter at a specific sagittal slice)
- I should probably clean up my comments in this method, since the comment on the highlighted lines implies that the SC width is determined there instead.

valosekj: > By comparison, when I check the `_compute_sc_ap_diameter_at_axial_slice` method, it turns out it computes the SC AP diameter for a given axial slice, _for the interpolated sagittal slice_. I don't think that's applicable here, since we want the per-sagittal-slice SC AP diameter (before interpolation).

Ah, excellent point!

> - I think maybe the _compute_sc_ap_diameter_at_axial_slice method could benefit from having _interpolated in the name to make the distinction clearer? (i.e. the method doesn't just grab the AP diameter at a specific sagittal slice)

Good idea! Agree.

joshuacwnewton: Hopefully f2bf7093fc30092455a1c0d098e4e3e4a48ebf9f clears up the confusion from my original comment?

joshuacwnewton: Ah shoot, typo. `s/with/width`. 🤣

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton, line 1: ## 1. Debugging failing tests on PR 5179

> There are some failing tests:
>
> image

@valosekj If you'd like, I can look into the failing tests in this PR?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -426,6 +497,59 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx):
...
+
+            # Save the NLI slice information
+            self.measure_pd.loc[idx, 'interpolated_midsagittal_slice'] = \
+                rpi_slice_to_orig_orientation(im_lesion_data.shape, self.orientation,
+                                              self.interpolated_midsagittal_slice_RPI, 0)
+
+            # Save the spinal cord A-P diameter as tissue bridge measurements
+            self.measure_pd.loc[idx, 'interpolated_dorsal_bridge_width [mm]'] = sc_ap_diameter_half_mm
+            self.measure_pd.loc[idx, 'interpolated_ventral_bridge_width [mm]'] = sc_ap_diameter_half_mm
+            self.measure_pd.loc[idx, 'interpolated_total_bridge_width [mm]'] = sc_ap_diameter_mm
```
joshuacwnewton, line 520: Sorry, it took me a very long time to realize this, but my solution and your solution are conflicting. 

When you asked me to push my fix to your branch 1 month ago, I didn't look at the work you had already done in this branch. However, it turns out we were solving the exact same problem: "What placeholder `bridge_width` value should we use if the lesion is missing?"

- **Your solution**: (A) Use a fixed `-nli` slice, (B) Have dorsal/ventral widths be equal to _half_ of the SC diameter at the `-nli` slice (or, if not provided, get the minimum AP diameters for each of the interpolation slices). (**Reason:** Dorsal/ventral should sum to total bridge width.)
- **My solution**: (A) Don't use a fixed slice, (B) Have dorsal/ventral widths be equal to the _full_ SC diameter at the given `[z, y]` coordinate. (**Reason:** Dorsal/ventral bridge_widths should result in an appropriate (averaged) interpolated bridge width, such that the `minimum` can be accurately determined.)

So, instead of merging the two solutions together, they should have been evaluated separately.

I think I need to undo my pushes to this branch, and instead open a separate branch and/or PR so that we can compare the impact of both solutions.

joshuacwnewton (changes requested):
Fixing the failing tests (due to my additions to the PR) turned out to be more complicated than I realized (see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179#discussion_r3088165813).

I think we need to do some git surgery to remove my changes, then evaluate both sets of changes separately.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: I have removed my changes from this branch, and re-merged `master` into this branch in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179/commits/cb0c757878d279ccf6d00c83f2ae3a63f84754c1.

joshuacwnewton: I have moved my changes to https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5202.

That way, we can compare and contrast the different solutions for both parasagittal lesions and missing lesions.

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton: The failing tests are because the "expected values" code has the same `0` bug as `sct_analyze_lesion`... (`.get(<>, 0.0)`):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/621c69f21374bcc4708061179ae2e534402900ee/testing/cli/test_cli_sct_analyze_lesion.py#L116-L118

For the failing tests, the lesion is missing for one or both of the interpolation slices, and we're getting a value of `0`.

If we replace the `0` value with the SC ap diameters, I think we end up with more accurate test values, however I ran into a bigger problem when testing all of the changes together. (See other comment.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton, line 1: ## 2. Comparison between PR 5129 and PR 5202

Here is a trace of the `AnalyzeLesion` class so that I can try to compare and contrast our two changesets:

Before:

```md
AnalyzeLesion.`analyze()`
├── 1. self.`label_lesion()`
│   └── Exit if no lesions at all.
└── 2. self.`measure()`
    ├── self.`get_midsagittal_slice()`
    │   ├── Use CoM of largest lesion.
    │   └── Store `floor` and `ceil` of CoM to get 2 interpolation slices.
    └── self.`_measure_tissue_bridges()`
        ├── Compute tissue_bridges for nonzero lesion slices only.
        └── self.`_measure_interpolated_tissue_bridges()`
            ├── If 0/2 interp slices overlap with lesion, interpolated_bridge_width = np.nan
            ├── If 1/2 interp slices overlap with lesion, set 1/2 bridge_width = 0.0, then interpolate.
            │   └── End result: Smaller than expected `min()` value.
            └── If 2/2 interp slices overlap with lesion, interpolate the two bridge_width values.                            
```

After (This PR, #5179):

```diff
 AnalyzeLesion.`analyze()`
 ├── 1. self.`label_lesion()`
-│   └── Exit if no lesions at all.
+│   └── If no lesions exist, proceed if an appropriate `-nli-slice` value was passed.
 └── 2. self.`measure()`
     ├── self.`get_midsagittal_slice()`
     │   ├── Use CoM of largest lesion.
+    │   ├── If no lesions exist, fall back to CoM of SC at `-nli-slice`.
     │   └── Store `floor` and `ceil` of CoM to get 2 interpolation slices.
     └── self.`_measure_tissue_bridges()`
         ├── Compute tissue_bridges for nonzero lesion slices only.
         └── self.`_measure_interpolated_tissue_bridges()`
-           ├── If 0/2 interp slices overlap with lesion, interpolated_bridge_width = np.nan
+           ├── If 0/2 interp slices overlap with lesion:
+           │   ├── If lesion exists, skip interpolation entirely and set the dorsal/ventral bridge_widths to `1/2` the minimum SC width.
+           │   └── If no lesion exists, return `np.nan`.
            ├── If 1/2 interp slices overlap with lesion, set 1/2 bridge_width = 0.0, then interpolate.
            │   └── End result: Smaller than expected `min()` value.
            └── If 2/2 interp slices overlap with lesion, interpolate the two bridge_width values.               
```

> _(Side note: Why do we still return `np.nan` for the bridge widths in the "no lesions exist" case?)_

After (My separate PR, #5202):

```diff
 AnalyzeLesion.`analyze()`
 ├── 1. self.`label_lesion()`
 │   └── Exit if no lesions at all.
 └── 2. self.`measure()`
     ├── self.`get_midsagittal_slice()`
     │   ├── Use CoM of largest lesion.
     │   └── Store `floor` and `ceil` of CoM to get 2 interpolation slices.
     └── self.`_measure_tissue_bridges()`
-        ├── Compute tissue_bridges for nonzero lesion slices only.
+        ├── Compute tissue_bridges for the union of {A. the axial slices spanning the lesion} and {B. the lesion sagittal slices + the two interpolation slices}.
+        │   ├── By always evaluating the interpolation slices, the 0/2 case is handled.
+        │   └── If lesion is missing at an axial-sagittal slice pair, use SC width.
         └── self.`_measure_interpolated_tissue_bridges()`
-            ├── If 0/2 interp slices overlap with lesion, interpolated_bridge_width = np.nan
-            ├── If 1/2 interp slices overlap with lesion (for all axial slices), set 1/2 bridge_width = 0.0, then interpolate.
-            │   └── End result: Smaller than expected `min()` value.
+            ├── If 0/2 or 1/2 interp slices overlap with lesion, the SC widths will be used, since we computed bridge_widths for interpolation slices at all axial slices corresponding to the lesion.
+            │   └── End result: Accurate `min()` value.
             └── If 2/2 interp slices overlap with lesion, interpolate the two bridge_width values.                            
```

joshuacwnewton: On reflection, both of our PRs affect different cases and different parts of AnalyzeLesion. So, maybe it IS okay to merge the two changesets.

However, the flow of `AnalyzeLesion` is so complex that I think we should be very, very careful moving forward, and make sure that we test each of the different cases carefully. Hence, maybe we should [fixup the failing tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179#discussion_r3088551876) for both of our PRs separately before combining our two changesets? That way we can isolate the behavior for each case, and confirm that the separate tests don't start failing when both changesets are merged.

I'll work on the test failures in this PR first (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179#discussion_r3075186037), then I'll work on the test failures in #5202, and then we can try to reconcile the two sets of changes.

joshuacwnewton (commented):
The `AnalyzeLesion` class has gotten really confusing: The methods aren't in the order of execution, so you can't read the file (nor the changes/diff) from top to bottom to understand the flow of the class.

Because of this, our PRs are kind of hard to understand, and so I've manually written some tree diagrams to map out the execution order.

testing/cli/test_cli_sct_analyze_lesion.py
joshuacwnewton: Turns out the failing test is not just the fact that we `.get(0.0)` as a fallback value, but also because we were previously skipping `NaN` values entirely. :rofl: 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/621c69f21374bcc4708061179ae2e534402900ee/testing/cli/test_cli_sct_analyze_lesion.py#L210-L212

(Just to explain why we weren't running into failures before when `NaN` was output.)

joshuacwnewton: Hopefully fixed by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179/commits/57fcd15b6d145e762261bb78846322a8eefb6c48 and https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179/commits/cf87413b81f56e638769389563961cbbd4719e1b.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: OK, I think I've got the tests passing on both PRs. 

One revelation: My fix in #5202 actually handles the parasagittal case too. So, there WAS a conflict between both sets of changes:

- Jan's changes: `/2` the AP diameter
- Joshua's changes: The full AP diameter

I had to adapt the test differently between the two PRs (see: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5202/changes/bbb02782d7efe1275c79d3e4aa368cda1518d4b7), and so I think this makes the two PRs incompatible.

We'd have to think about how to approach this case-by-case. To me, using the full SC width makes sense for interpolated bridge calculations, but I also think that the total summing to the SC width makes sense as well. I'm not quite sure how to proceed...

valosekj: > (Side note: Why do we still return np.nan for the bridge widths in the "no lesions exist" case?)

good catch. this is not correct. we should do the same as for 

```diff
+           │   ├── If lesion exists, skip interpolation entirely and set the dorsal/ventral bridge_widths to `1/2` the minimum SC width.
```

joshuacwnewton: > > (Side note: Why do we still return np.nan for the bridge widths in the "no lesions exist" case?)
> 
> good catch. this is not correct. we should do the same as for
> 
> +           │   ├── If lesion exists, skip interpolation entirely and set the dorsal/ventral bridge_widths to `1/2` the minimum SC width.

I think I was mistaken. The `np.nan` case wasn't for the "no lesions exist" case, it was for the case where the AP diameter cannot be computed at all. So, the absolute last resort. I think things should be fine here.

joshuacwnewton: OK! I think I've implemented the suggestion from Dario correctly. I'm just going to commit them and let the CI run.

(I didn't have time to run the CI locally and correct, so apologies if there are errors. I'm a little time-crunched, heh.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -260,8 +276,7 @@ def show_total_results(self):
         total_length = np.round(np.sum(self.measure_pd['length [mm]']), 2)
         # Take max width across lesions --> max
         max_width = np.round(np.max(self.measure_pd['width [mm]']), 2)
-        lesion_count = len(self.measure_pd['volume [mm3]'].values)
-
+        lesion_count = 0 if self.empty_lesion else len(self.measure_pd['volume [mm3]'].values)
```
mguaypaq: This is a weird override. Shouldn't the `len` be zero already if there's no lesion? Or is it something like `None` instead of being the empty list? But from my quick testing, it looks like an empty dataframe correctly produces something of length zero:

```py3
>>> import pandas as pd
>>> df = pd.DataFrame(
...     data={'abc': None, 'def': None},
...     index=range(0),
...     columns=['abc', 'def'],
... )
>>> len(df['abc'])
0
```

Also, it's a bit weird that this `lesion_count` uses the `volume [mm3]` column, whereas line 1316 uses the `label` column to determine `empty_lesion`.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -378,12 +399,49 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx):
...
-            self.measure_pd.loc[idx, 'interpolated_total_bridge_width [mm]'] = np.nan
+            # Get all axial slices (S-I direction) with the lesion
+            axial_lesion_slices = np.unique(np.where(im_lesion_data)[2])  # [2] --> S-I direction in RPI
+
+            # Compute the spinal cord A-P diameter for each axial slice with lesion at the midsagittal slice
+            sc_ap_diameter_mm = {}
+            for axial_slice in axial_lesion_slices:
+                diameter = self._compute_sc_ap_diameter_at_axial_slice(im_sc_data, axial_slice, p_lst)
+                if diameter is not None:
+                    sc_ap_diameter_mm[axial_slice] = diameter
```
mguaypaq: FYI, thanks to the [walrus operator](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions), kind of dictionary creation loop can be rewritten as a dictionary comprehension:
```suggestion
            sc_ap_diameter_mm = {
                axial_slice: diameter
                for axial_slice in axial_lesion_slices
                if (diameter := self._compute_sc_ap_diameter_at_axial_slice(im_sc_data, axial_slice, p_lst))
                    is not None
            }
```

(This is entirely optional.)

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -378,12 +399,49 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx):
...
+                   self.verbose, type='warning')
+            printv(f'  Midsagittal dorsal tissue bridge width: '
+                   f'{np.round(min_interpolated_dorsal_bridge_width_mm, 2)} mm',
+                   self.verbose, type='info')
+            printv(f'  Midsagittal ventral tissue bridge width: '
+                   f'{np.round(min_interpolated_ventral_bridge_width_mm, 2)} mm',
+                   self.verbose, type='info')
+            printv(f'  Midsagittal total tissue bridge width (spinal cord A-P diameter): '
+                   f'{np.round(interpolated_total_bridge_width_mm, 2)} mm',
+                   self.verbose, type='info')
```
mguaypaq: Python's builtin `min` function can take an arbitrary iterable (so, no need to convert to a `list`), and also a default value (so, no need to special case the empty list). Using this, we can de-duplicate the code a bit:
```suggestion
            # Use the minimum A-P diameter as the total bridge width
            # For dorsal and ventral, we split it in half as we don't know the actual lesion location
            # If we can't compute the A-P diameter, fall back to NaN
            diameter = min(sc_ap_diameter_mm.values(), default=np.nan)
            self.measure_pd.loc[idx, 'interpolated_dorsal_bridge_width [mm]'] = diameter / 2
            self.measure_pd.loc[idx, 'interpolated_ventral_bridge_width [mm]'] = diameter / 2
            self.measure_pd.loc[idx, 'interpolated_total_bridge_width [mm]'] = diameter

            if not np.isnan(diameter):
                printv('  Lesion not on midsagittal slice, using spinal cord A-P diameter as proxy',
                       self.verbose, type='warning')
                printv(f'  Midsagittal dorsal tissue bridge width: '
                       f'{np.round(min_interpolated_dorsal_bridge_width_mm, 2)} mm',
                       self.verbose, type='info')
                printv(f'  Midsagittal ventral tissue bridge width: '
                       f'{np.round(min_interpolated_ventral_bridge_width_mm, 2)} mm',
                       self.verbose, type='info')
                printv(f'  Midsagittal total tissue bridge width (spinal cord A-P diameter): '
                       f'{np.round(interpolated_total_bridge_width_mm, 2)} mm',
                       self.verbose, type='info')
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -426,6 +485,59 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx):
...
+            printv(f'  Midsagittal ventral bridge width (half A-P diameter): '
+                   f'{np.round(sc_ap_diameter_half_mm, 2)} mm',
+                   self.verbose, type='info')
+            printv(f'  Midsagittal total tissue bridge width (spinal cord A-P diameter): '
+                   f'{np.round(sc_ap_diameter_mm, 2)} mm',
+                   self.verbose, type='info')
+        else:
+            printv(f'ERROR: No spinal cord A-P diameter computed at the NLI slice {self.nli_slice}.',
+                   self.verbose, 'warning')
```
mguaypaq: This seems inconsistent: is it an `ERROR` or a `warning`?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -957,6 +1065,11 @@ def measure(self):
                                                       im_lesion=im_lesion_data_cur,
                                                       p_lst=p_lst)
 
+        # Special case: no lesion but NLI slice and spinal cord segmentation are provided --> Compute spinal cord
+        # A-P diameter at the NLI slice (if user provided the NLI slice)
+        if self.nli_slice is not None:
+            self._compute_sc_ap_diameter_at_nli(im_lesion_data, p_lst)
```
mguaypaq, line 1063: Shouldn't this also check `self.empty_lesion`? Or is that somehow checked elsewhere in the logic?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -426,6 +485,59 @@ def _measure_interpolated_tissue_bridges(self, tissue_bridges_df, p_lst, idx):
...
+        This is called when no lesions are detected but an NLI slice is provided.
+        The spinal cord A-P diameter is computed as a proxy for tissue bridges and saved to the output.
+
+        :param im_lesion_data: 3D numpy array: shape of the image (RPI orientation)
+        :param p_lst: list, pixel size
+        """
+        im_sc_data = Image(self.fname_sc).data
+
+        # Compute the spinal cord A-P diameter at the NLI slice
+        sc_ap_diameter_mm = self._compute_sc_ap_diameter_at_axial_slice(im_sc_data, self.nli_slice, p_lst)
```
mguaypaq: The script usage and code comments say that `self.nli_slice` is in the original image orientation, but `self._compute_sc_ap_diameter_at_axial_slice` expects it in RPI orientation, and I don't see any conversion happening anywhere.

Maybe there should also be a `self.nli_slice_RPI` attribute?

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -1057,26 +1185,43 @@ def get_midsagittal_slice(self, im_lesion_data, label_lst, p_lst):
...
-        for z in z_range:
-            spinal_cord_slice = im_sc_data[:, :, z]     # RPI --> selecting in the 3rd dimension (SI) to get axial slice
-            if np.any(spinal_cord_slice):  # Avoid empty slices
-                stored_x_coordinates.append(center_of_mass(spinal_cord_slice)[0])   # [0] --> R-L
-        # 4. Calculate target position in right-left axis (x direction) for the interpolation (mean of spinal cord
-        # center of mass (in the x-axis (R-L direction))
-        self.interpolated_midsagittal_slice_RPI = np.mean(stored_x_coordinates)    # e.g., for [8.6, 8.6, 8.9, 8.8, 9.6] --> 8.7
+        if self.nli_slice is not None:
+            # Compute the spinal cord center of mass in the R-L axis at the NLI slice
+            spinal_cord_slice = im_sc_data[:, :, self.nli_slice]
```
mguaypaq: This should probably be using `self.nli_slice_RPI` instead.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -1057,26 +1185,43 @@ def get_midsagittal_slice(self, im_lesion_data, label_lst, p_lst):
...
+            if np.any(spinal_cord_slice):
+                self.interpolated_midsagittal_slice_RPI = center_of_mass(spinal_cord_slice)[0]  # [0] --> R-L
+            else:
+                printv(f'ERROR: No spinal cord found at the NLI slice (RPI, S-I axis): {self.nli_slice}. '
```
mguaypaq:
```suggestion
                printv(f'ERROR: No spinal cord found at the NLI slice (RPI, S-I axis): {self.nli_slice_RPI}. '
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -1057,26 +1185,43 @@ def get_midsagittal_slice(self, im_lesion_data, label_lst, p_lst):
...
+            # 2. Define analysis range (2 axial slices above and below the lesion center of mass) around lesion center
+            # mass in S-I axis (z direction)
+            # TODO: try other number of slices above and below the lesion center of mass
+            z_range = np.arange(z_center - 2, z_center + 3)   # 5 slices in total
+            # 3: For each of these slices, compute the spinal cord center of mass in the x-axis (R-L direction)
+            stored_x_coordinates = []
+            for z in z_range:
+                spinal_cord_slice = im_sc_data[:, :, z]     # RPI --> selecting in the 3rd dimension (SI) to get axial slice
+                if np.any(spinal_cord_slice):  # Avoid empty slices
+                    stored_x_coordinates.append(center_of_mass(spinal_cord_slice)[0])   # [0] --> R-L
```
mguaypaq: Entirely optional, but as a list comprehension:
```suggestion
            stored_x_coordinates = [
                center_of_mass(spinal_cord_slice)[0]  # [0] --> R-L
                for z in z_range
                if np.any(  # Avoid empty slices
                    spinal_cord_slice := im_sc_data[:, :, z])  # RPI --> selecting in the 3rd dimension (SI) to get axial slice
            ]
```

spinalcordtoolbox/scripts/sct_analyze_lesion.py
```diff
@@ -1093,9 +1238,55 @@ def _interpolate_values(self, data1, data2):
         :param data2: 2D numpy array (slice 2) or single int64 (tissue bridge for slice 2)
         :return: 2D numpy array (interpolated slice) or single float64 (interpolated tissue bridge)
         """
+        # This case should realistically be handled earlier in the script by falling back to `-nli-slice`.
+        # If a double-nan input slips through, it's a bug, so throw an error.
+        if np.any(np.isnan(data1) & np.isnan(data2)):
+            raise ValueError("Two 'NaN' values passed. Cannot interpolate between two 'NaN' values.")
+        # This case can occur for lesions that are present in only 1/2 interpolation slices.
+        # The decision to fall back to 1/2 slices was made here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5202#discussion_r3138996132
+        elif np.any(np.isnan(data1)) or np.any(np.isnan(data2)):
```
mguaypaq: I was initially confused by the use of `&` above and `or` here. This would be more consistent:
```suggestion
        elif np.any(np.isnan(data1) | np.isnan(data2)):
```

mguaypaq (approved):
Overall this seems fine, although I didn't check every single detail. I'll mark this as 'approved' for now, so that this can be merged as soon as all the comments are marked 'resolved'.

spinalcordtoolbox/scripts/sct_analyze_lesion.py
joshuacwnewton: It's nice that `np.nan / 2` is still just `np.nan`, further allowing this to work!

testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -361,3 +388,208 @@ def test_sct_analyze_lesion_no_lesion_found(tmp_path, tmp_path_qc):
...
+    import sys
+
+    # Run the script as a separate process to catch the exit code
+    process = run([sys.executable, '-m', 'spinalcordtoolbox.scripts.sct_analyze_lesion',
+                   '-m', path_empty_lesion,
+                   '-s', path_seg,
+                   '-nli-slice', '10',
+                   '-ofolder', str(tmp_path),
+                   '-qc', str(tmp_path_qc)],
+                  capture_output=True)
```
joshuacwnewton: This can be an in-process call using `with pytest.raises(SystemExit)`. We've got a few examples of this in our test suite.

joshuacwnewton (changes requested):
Realizing now that the tests need a little work, too.

joshuacwnewton (approved):
Changes pushed, feel free to have one last look over and merge if the changes look good! :)

mguaypaq (approved):

valosekj:
Thanks a lot @joshuacwnewton and @mguaypaq for your contribution 🙏

PR #5180, opened 2026-03-02
Amend PR #5129 to properly test for issues with segmentation metrics

joshuacwnewton:
## Checklist

## Description

This PR tries to update #5129 to make sure our test properly catches metric issues by:

- Decreasing the relative tolerance we use to check values
- Checking per-value metrics instead of mean values when possible
- Doing slice-by-slice comparisons to ensure metrics don't vary.
- Removing the hardcoded ground-truth values to avoid normalization of deviance when the tests fail.

This test fully passes on v7.1 and the tolerance could probably go a bit lower, but fails on `master` and for PR #5129. I think my ideal would be to reduce the tolerance further, then make the code changes necessary for things to pass, but we should be confident that we're testing the correct things before merging the test.

## Linked issues

N/A.
To be merged into #5129.

### DIFF
--- a/testing/api/test_process_seg.py
+++ b/testing/api/test_process_seg.py
@@ -288,7 +288,7 @@ def test_compute_shape(im_seg, expected, params):
                                                      angle_correction=params['angle_corr'],
                                                      param_centerline=ParamCenterline(),
                                                      verbose=VERBOSE)
-    slice_range = [params['slice']] if 'slice' in params else range(im_seg.data.shape[2])
+    slice_range = range(im_seg.data.shape[2])
     for key in expected.keys():
         # If we're testing angle values, ensure the values are within half a degree
         # If we're testing distances (area, diameter, length), ensure the values are within 5% of the expected value
@@ -297,7 +297,8 @@ def test_compute_shape(im_seg, expected, params):
 
         # for length, the values are given per-slice, but we want to check the total length (hence `sum()`)
         if key == 'length':
-            assert metrics[key].data.sum() == pytest.approx(expected[key], **kwargs)
+            # ignore zeroed-out slices (i.e. `zeroslice`), as these will be `np.nan` and throw off the sum
+            assert np.nansum(metrics[key].data) == pytest.approx(expected[key], **kwargs)
         else:
             # for angled cords, SCT's angle-estimating code uses the tangent to a single-voxel-wide centerline.
             # this code is somewhat sensitive to single-voxel shifts in the centerline, e.g.:
@@ -326,11 +327,12 @@ def test_compute_shape(im_seg, expected, params):
             # for non-angled cords, we can reliably compare expected values on a slice-wise basis
             else:
                 for n_slice in slice_range:
+                    # fetch expected/obtained values
+                    expected_value = expected[key]
                     obtained_value = metrics[key].data[n_slice]
-                    # fetch expected_value
-                    if expected[key] is np.nan:
+
+                    # if n_slice is empty (i.e. `zeroslice` was set) all metrics should be `np.nan`
+                    if im_seg.data[:, :, n_slice].sum() == 0:
                         assert math.isnan(obtained_value)
-                        break
                     else:
-                        expected_value = pytest.approx(expected[key], **kwargs)
-                    assert obtained_value == expected_value
+                        assert obtained_value == pytest.approx(expected_value, **kwargs)
--- a/testing/api/test_process_seg.py
+++ b/testing/api/test_process_seg.py
@@ -138,7 +138,7 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
     data_rot_crop = data_rot[padding:nx+padding, padding:ny+padding, padding:nz+padding]
 
     # Zero specified slices
-    if zeroslice is not []:
+    if zeroslice:
         data_rot_crop[:, :, zeroslice] = 0
 
     # Create nibabel object
--- a/testing/api/test_process_seg.py
+++ b/testing/api/test_process_seg.py
@@ -45,8 +45,19 @@ dict_test_orientation = [
 def get_effective_angles(rot_x=0.0, rot_y=0.0, rot_z=0.0, order='xyz', degrees=True):
     """
     Given sequential rotations applied in `order`, compute the effective
-    angle_AP, angle_RL, and angle_IS that sct_process_segmentation would measure.
+    angles in each axis (to match what `sct_process_segmentation` would measure).
     """
+    # FIXME: @joshuacwnewton, this code was written involving discussions with an LLM.
+    #        Specifically, I was looking for an explanation for why the test with
+    #        "angle_RL=-10, angle_AP=30" had hardcoded a value of "angle_RL=-11.5" in
+    #        its "expected values" dict. Applying the total rotation matrix to the unit
+    #        vector for the z-axis produces the expected angle value, so I kept it
+    #        just to make writing the rest of PR #5180 smoother. But, I would like to
+    #        better understand:
+    #            - How to generate the correct sign (or whether this is ambiguous in nature)
+    #            - How to generate the correct SI rotation (this should be equivalent to 'orientation', but isn't)
+    #            - Whether `from_euler` is equivalent to individually generating the rotation matrices for each axis
+    #              and then matmul'ing them. (I tried that first, but got different outputs...)
     rot = {'x': rot_x, 'y': rot_y, 'z': rot_z}
     r = Rotation.from_euler(seq=order, angles=[rot[ax] for ax in order], degrees=degrees)
     tx, ty, tz = r.apply([0.0, 0.0, 1.0])
@@ -188,13 +199,24 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
     else:
         assert shape == 'rectangle'
         area = (radius_RL*2+1) * (radius_AP*2+1)
-        # NB: skimage.regionprops will fit an "ellipse that has the same second-moments as the region." So, we need to
-        #     use the diameters for that ellipse. Since we only test rectangles with no rotation, we can hardcode the factor 2/sqrt(3).
-        #     source: https://scikit-image.org/docs/0.25.x/api/skimage.measure.html#skimage.measure.regionprops
+        # NB: For diameters, the typical rectangle formula would underestimate the diameter output by
+        # `skimage.regionprops`. This is because regionprops fits an ellipse with the same second central moments, and
+        # the ellipse will always extend past the corners of the rectangle.
+        #   -> See: https://scikit-image.org/docs/0.25.x/api/skimage.measure.html#skimage.measure.regionprops
+        # But, we can derive what the ellipse diameters would be (as a function of the rectangle's height and width)
+        # by equating the second central moments of the rectangle and ellipse:
+        #   - Rectangle: Ix = W*H^3/12, Iy = H*W^3/12     # Side lengths: (H, W)
+        #   - Ellipse:   Ix = π*a*b^3/4, Iy = π*a^3*b/4   # Semi-axes: (a, b)
+        #   - Equating moments gives a = W/√3, b = H/√3
+        #   - Since the diameters are 2x the semi-axis lengths, the ellipse diameters are (2/√3) * rectangle side lengths.
+        # FIXME: This only works for non-rotated rectangles, since if the cross-section is rotated, the fitted ellipse
+        #        will no longer have semi-axes that are aligned with the AP/RL axes. But, our test suite currently only
+        #        tests rectangle cross-sections with no rotation, so this is sufficient for now.
         diameter_AP_regionprops = (2 / math.sqrt(3)) * ((radius_AP * 2) + 1)
         diameter_RL_regionprops = (2 / math.sqrt(3)) * ((radius_RL * 2) + 1)
-        # Ah, but actually, we recently refactored the code to use a different diameter measure instead of the "major axis length" for the diameter_AP
-        # So, I guess we CAN just use the simple formula here... but only for diameter_AP.
+        # Ah, but actually, we recently refactored the code to use a different AP diameter measure instead of the
+        # "major axis length" in PR #4958.
+        #  -> See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b322832e939d356039af27832b09cd671e034b2d/spinalcordtoolbox/process_seg.py#L438-L442
         diameter_AP = (radius_AP * 2) + 1
         diameter_RL = diameter_RL_regionprops
 
@@ -231,52 +253,52 @@ def test_fix_orientation(test_orient):
 
 # Generate a list of fake segmentation for testing: (dummy_segmentation(params), dict of expected results)
 im_segs = [
-    # test area
+    # 0. test area
     dummy_segmentation(size_arr=(32, 32, 5), debug=DEBUG) +
     [{'angle_corr': False}],
 
-    # test anisotropic pixel dim
+    # 1. test anisotropic pixel dim
     dummy_segmentation(size_arr=(64, 32, 5), pixdim=(0.5, 1, 5), debug=DEBUG) +
     [{'angle_corr': False}],
 
-    # test with angle IS
+    # 2. test with angle IS
     dummy_segmentation(size_arr=(32, 32, 5), pixdim=(1, 1, 5), angle_IS=15, debug=DEBUG) +
     [{'angle_corr': False}],
 
-    # test with ellipse shape
+    # 3. test with ellipse shape
     dummy_segmentation(size_arr=(64, 64, 5), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=0, debug=DEBUG) +
     [{'angle_corr': False}],
 
-    # test with int16. Different bit ordering, which can cause issue when applying transform.warp()
+    # 4. test with int16. Different bit ordering, which can cause issue when applying transform.warp()
     dummy_segmentation(size_arr=(64, 320, 5), shape='rectangle', radius_RL=13, radius_AP=5, angle_RL=0, debug=DEBUG,
                        pixdim=(1, 1, 1), dtype=np.int16, orientation='RPI',) +
     [{'angle_corr': False}],
 
-    # test with angled spinal cord (neg angle)
+    # 5. test with angled spinal cord (neg angle)
     dummy_segmentation(size_arr=(64, 64, 20), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=-30, debug=DEBUG) +
     [{'angle_corr': True}],
 
-    # test with AP angled spinal cord
+    # 6. test with AP angled spinal cord
     dummy_segmentation(size_arr=(64, 64, 20), shape='ellipse', radius_RL=13, radius_AP=5, angle_AP=20, debug=DEBUG) +
     [{'angle_corr': True}],
 
-    # test with RL and AP angled spinal cord
+    # 7. test with RL and AP angled spinal cord
     dummy_segmentation(size_arr=(64, 64, 50), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=-10, angle_AP=15, debug=DEBUG) +
     [{'angle_corr': True}],
 
-    # Reproduce issue: "LinAlgError: SVD did not converge".
+    # 8. Reproduce issue: "LinAlgError: SVD did not converge".
     dummy_segmentation(size_arr=(64, 64, 50), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=-10, angle_AP=30, debug=DEBUG) +
     [{'angle_corr': True}],
 
-    # test uint8 input
+    # 9. test uint8 input
     dummy_segmentation(size_arr=(32, 32, 50), dtype=np.uint8, angle_RL=15, debug=DEBUG) +
     [{'angle_corr': True}],
 
-    # test all output params
+    # 10. test all output params
     dummy_segmentation(size_arr=(128, 128, 5), pixdim=(1, 1, 1), shape='ellipse', radius_RL=50, radius_AP=30, debug=DEBUG) +
     [{'angle_corr': False}],
 
-    # test with one empty slice
+    # 11. test with one empty slice
     dummy_segmentation(size_arr=(32, 32, 5), zeroslice=[2], debug=DEBUG) +
     [{'angle_corr': False, 'slice': 2}]
 ]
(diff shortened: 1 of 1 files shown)


testing/api/test_process_seg.py
```diff
@@ -40,6 +42,32 @@
...
+    #        Specifically, I was looking for an explanation for why the test with
+    #        "angle_RL=-10, angle_AP=30" had hardcoded a value of "angle_RL=-11.5" in
+    #        its "expected values" dict. Applying the total rotation matrix to the unit
+    #        vector for the z-axis produces the expected angle value, so I kept it
+    #        just to make writing the rest of PR #5180 smoother. But, I would like to
+    #        better understand:
+    #            - How to generate the correct sign (or whether this is ambiguous in nature)
+    #            - How to generate the correct SI rotation (this should be equivalent to 'orientation', but isn't)
+    #            - Whether `from_euler` is equivalent to individually generating the rotation matrices for each axis
+    #              and then matmul'ing them. (I tried that first, but got different outputs...)
```
mguaypaq: Yes, `from_euler` is equivalent to matrix multiplying the individual rotation matrices, but there's a gotcha about the order of multiplication.

The quick check is this:
```py3
>>> import numpy as np
>>> from scipy.spatial.transform import Rotation
>>> x_mat = Rotation.from_euler('x', 10, degrees=True).as_matrix()
>>> y_mat = Rotation.from_euler('y', 20, degrees=True).as_matrix()
>>> z_mat = Rotation.from_euler('z', 30, degrees=True).as_matrix()
>>> xyz_mat = Rotation.from_euler('xyz', [10, 20, 30], degrees=True).as_matrix()
>>> np.allclose(xyz_mat, z_mat @ y_mat @ x_mat)
True
```

Note that the order of multiplication is reversed, because applying a rotation matrix to a vector happens by multiplying with the vector on the right. The `xyz_mat` matrix is supposed to
1. first rotate around the X axis by 10 degrees,
2. then rotate around the Y axis by 20 degrees, and
3. finally rotate around the Z axis by 30 degrees.

In code, doing this to, say, the vector `[0, 0, 1]` is:
```py3
>>> xyz_mat @ [0, 0, 1]
array([0.37852231, 0.01802831, 0.92541658])
>>> z_mat @ (y_mat @ (x_mat @ [0, 0, 1]))
array([0.37852231, 0.01802831, 0.92541658])
```

So, we need the first, second, third, etc. individual rotation matrices to be written from right to left. I wrote the parentheses to make the order of operations clear, but they're optional because matrix multiplication is associative.

testing/api/test_process_seg.py
```diff
@@ -40,6 +42,32 @@
...
+    """
+    # FIXME: @joshuacwnewton, this code was written involving discussions with an LLM.
+    #        Specifically, I was looking for an explanation for why the test with
+    #        "angle_RL=-10, angle_AP=30" had hardcoded a value of "angle_RL=-11.5" in
+    #        its "expected values" dict. Applying the total rotation matrix to the unit
+    #        vector for the z-axis produces the expected angle value, so I kept it
+    #        just to make writing the rest of PR #5180 smoother. But, I would like to
+    #        better understand:
+    #            - How to generate the correct sign (or whether this is ambiguous in nature)
+    #            - How to generate the correct SI rotation (this should be equivalent to 'orientation', but isn't)
```
mguaypaq: I'm not sure that there's a good way to do this, for two reasons.
1. It certainly can't be done by just looking at the coordinates of `rot.apply([0, 0, 1])`, because every `angle_IS` gives the same `rot.apply([0, 0, 1]) == [0, 0, 1]` when `angle_RL == angle_AP == 0`.
2. But even when looking at the whole rotation matrix, I'm not sure the orientation is well-defined. For example, if we consider a dummy segmentation where each axial slice is a rectangle before rotation; after rotation, each slice will technically be a parallelogram, not always a rectangle. And, there's no real answer to the question, "what `angle_IS` would turn the original rectangle into the final parallelogram". (Certainly some values are more "correct" than others, but we couldn't have a tight confidence interval.)

(The situation is different for `angle_RL` and `angle_AP`, because those are defined by using the 1-dimensional centerline, rather than a 2-dimensional axial slice.)

testing/api/test_process_seg.py
```diff
@@ -40,6 +42,32 @@
...
+    angles in each axis (to match what `sct_process_segmentation` would measure).
+    """
+    # FIXME: @joshuacwnewton, this code was written involving discussions with an LLM.
+    #        Specifically, I was looking for an explanation for why the test with
+    #        "angle_RL=-10, angle_AP=30" had hardcoded a value of "angle_RL=-11.5" in
+    #        its "expected values" dict. Applying the total rotation matrix to the unit
+    #        vector for the z-axis produces the expected angle value, so I kept it
+    #        just to make writing the rest of PR #5180 smoother. But, I would like to
+    #        better understand:
+    #            - How to generate the correct sign (or whether this is ambiguous in nature)
```
mguaypaq: Based on thinking about the definition of `arctan2`, and the positions of the IS, RL, AP axes in space, I think these formulas would give the right signs (note the negative `-ty` for the `angle_RL` output):
```py3
>>> import numpy as np
>>> from scipy.spatial.transform import Rotation
>>> 
>>> def angles(angle_IS, angle_RL, angle_AP):
...     rot = Rotation.from_euler('zxy', [
...         angle_IS,  # positive angle sends L -> A -> R -> P
...         angle_RL,  # positive angle sends A -> S -> P -> I
...         angle_AP,  # positive angle sends S -> L -> I -> R
...         ], degrees=True)
...     tx, ty, tz = rot.apply([0, 0, 1])
...     return {
...         'angle_RL': np.degrees(np.arctan2(-ty, tz)),
...         'angle_AP': np.degrees(np.arctan2(tx, tz)),
...     }
... 
>>> angles(0, -10, 30)
{'angle_RL': -11.508393365822322, 'angle_AP': 29.999999999999996}
>>> angles(0, 10, 30)
{'angle_RL': 11.508393365822322, 'angle_AP': 29.999999999999996}
>>> angles(0, -10, -30)
{'angle_RL': -11.508393365822322, 'angle_AP': -29.999999999999996}
>>> angles(0, 10, -30)
{'angle_RL': 11.508393365822322, 'angle_AP': -29.999999999999996}
```

But, based on these lines:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/706001ebb18fa5b1e9f911deecf07a4deaba67ac/spinalcordtoolbox/process_seg.py#L127-L132

I think the right place for the negative sign is instead on the `angle_RL` **input**:
```py3
def angles(angle_IS, angle_RL, angle_AP):
    rot = Rotation.from_euler('zxy', [
        angle_IS,  # positive angle sends L -> A -> R -> P
        -angle_RL,  # positive angle sends A -> I -> P -> S
        angle_AP,  # positive angle sends S -> L -> I -> R
        ], degrees=True)
    tx, ty, tz = rot.apply([0, 0, 1])
    return {
        'angle_RL': np.degrees(np.arctan2(ty, tz)),
        'angle_AP': np.degrees(np.arctan2(tx, tz)),
    }
```

(Note that, in both cases, the `angle_IS` input doesn't actually affect the result at all.)

mguaypaq (commented):
Some initial comments for `get_effective_angles()`, sorry for the delay but I wanted to really think about that one. I'll continue reviewing but I wanted to at least submit this.

testing/api/test_process_seg.py
joshuacwnewton: I've applied your fix to remove the `abs` call, and it does seem to work okay. It's just that I've been trying to figure out how the snippet shown above implies that the RL angle should be negated, and I just can't see it. :sob:

Could you explain where it comes from?

mguaypaq: Sorry, I didn't explain that at all.

The `tangent_vect` on line 129 of `process_seg.py` plays the same role as the `rot.apply([0, 0, 1])` in `angles()`. So, for `angles()` to match `process_seg`, the output should use the same coordinate ratios in arctan (whether `math.atan2` or `np.arctan2`). In particular, that's `y/z` for `angle_RL` and `x/z` for `angle_AP`, with no negative signs.

Note that `arctan2(y, x)` is defined as the angle between:
- the ray starting at the origin (0, 0) and going through (1, 0), in other words, the +X axis in the X-Y plane
and:
- the ray starting at the origin (0, 0) and going through $(x, y)$, which is the opposite order from the arguments of `arctan2`, just because this wasn't confusing enough already.

The result is:
- a positive number of degrees between 0 and $\pi$ if $y > 0$ (above the X axis), which is counterclockwise in the X-Y plane.
- a negative number of degrees between 0 and $-\pi$ if $y < 0$ (below the X axis), which is clockwise in the X-Y plane.

In particular, this means that a (small) positive `angle_RL = arctan2(y, z)` rotation should:
- keep the X axis unchanged, and
- move the +Z axis towards the +Y axis (and after that, the -Z axis and the -Y axis, before coming back to +Z).

Similarly, a positive `angle_AP = arctan2(x, z)` rotation should:
- keep the Y axis unchanged, and
- move the +Z axis towards +X (and then -Z, -X, +Z, etc.).

So, that fixes the output, and the remaining ambiguity is what positive or negative angles mean in `from_euler`. In my testing:
- `from_euler('x')` keeps the X axis fixed, and rotates +Z to -Y to -Z to +Y, which is the reverse of what we want for `angle_RL`, hence the negative sign.
- `from_euler('y')` keeps the Y axis fixed, and rotates +Z to +X to -Z to -X, which matches what we want for `angle_AP`, hence the (invisible) positive sign.

As for `angle_IS`, I know what `from_euler` does, but I didn't figure out what `process_seg` does, so the sign may be wrong there (but it doesn't affect the output of `angles()`, so it may not matter).

joshuacwnewton: > * `from_euler('x')` keeps the X axis fixed, and rotates +Z to -Y to -Z to +Y, which is the reverse of what we want for `angle_RL`, hence the negative sign.

:exploding_head: 

Just to double-check my understanding (visually), for this sample axis diagram:

image

Looking down the x-axis (i.e. from +ve to -ve):

- `angle_RL = arctan2(y, z)` indicates a clockwise angle
- `from_euler('x')` indicates a counterclockwise angle

This is so odd! I wonder why this occurs just for the one axis.

mguaypaq: I think it's because we tend to default to RPI orientation in SCT, which is a left-handed coordinate system, whereas Nifti and most image processing software prefer right-handed systems (like LPI). (You switch from left-handed to right-handed, or vice-versa, by either negating one axis, or swapping two axes.)

joshuacwnewton: Turns out it's not left-hand vs. right-hand, per se. But instead, SCT uses a mix of both!

- SCT always defines angles as [relative to the +Z axis](https://spinalcordtoolbox.com/stable/user_section/tutorials/shape-analysis/compute-csa-and-other-shape-metrics/verify-correctness.html) (seen also in the arctan2 calls):
  image
- The above angles shown (+Z to +Y, +Z to +X) are a mix of a [left-hand rotation and a right-hand rotation](https://en.wikipedia.org/wiki/Right-hand_rule#A_rotating_body).
- Hence the need to negate one of the angles to get it into a consistent (left-hand) form for `from_euler`.

I've done my best to get it all documented properly in [ce694b2](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5180/commits/ce694b287a0c2fd3aa1821a0c1ca30411a2b5428).

testing/api/test_process_seg.py
```diff
@@ -106,7 +159,12 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
         if shape == 'rectangle':  # theoretical CSA: (a*2+1)(b*2+1)
             data[:, :, iz] = ((abs(xx - x[iz]) <= radius_RL) & (abs(yy - yfit[iz]) <= radius_AP)) * 1
         if shape == 'ellipse':
-            data[:, :, iz] = (((xx - x[iz]) / radius_RL) ** 2 + ((yy - yfit[iz]) / radius_AP) ** 2 <= 1) * 1
+            height, width = radius_RL, radius_AP
+            rows, cols = ellipse(height, width, height + 0.5, width + 0.5)
+            # FIXME: Write 1s directly to data, probably cleaner!
```
mguaypaq: This way seems fine, actually, and simple to read.

testing/api/test_process_seg.py
```diff
@@ -155,7 +213,60 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
...
+        diameter_AP = (radius_AP * 2) + 1
+        diameter_RL = diameter_RL_regionprops
+
+    # NB: rotations are not commutative i.e. the order that rotations are applied matters.
+    #     if we ever have multiple rotations, we can't just take the `angle_` values and check directly, as each
+    #     subsequent rotation will change the effective rotation in the previous axes.
+    #     So, we need to compute the effective rotation about each original axis after applying all rotations.
+    effective = angles(angle_RL=angle_RL, angle_AP=angle_AP, angle_IS=angle_IS,
+                       orientation=orientation, ax_order=('IS', 'RL', 'AP'))
+    expected = {
```
mguaypaq, line 264: Maybe we could fit `angle_IS` into `'orientation'`? We'd have to figure out the right sign convention if so.

testing/api/test_process_seg.py
mguaypaq: I think this is where `dtype` should be specified, if we want to start honouring it. (Alternatively, we could just drop it from the argument list of `dummy_segmentation`.)

testing/api/test_process_seg.py
```diff
@@ -106,7 +159,12 @@ def dummy_segmentation(size_arr=(256, 256, 256), pixdim=(1, 1, 1), dtype=np.floa
         if shape == 'rectangle':  # theoretical CSA: (a*2+1)(b*2+1)
             data[:, :, iz] = ((abs(xx - x[iz]) <= radius_RL) & (abs(yy - yfit[iz]) <= radius_AP)) * 1
         if shape == 'ellipse':
-            data[:, :, iz] = (((xx - x[iz]) / radius_RL) ** 2 + ((yy - yfit[iz]) / radius_AP) ** 2 <= 1) * 1
+            height, width = radius_RL, radius_AP
+            rows, cols = ellipse(height, width, height + 0.5, width + 0.5)
+            # FIXME: Write 1s directly to data, probably cleaner!
+            footprint = np.zeros((2*height + 1, 2*width + 1), dtype=dtype)
```
mguaypaq: This way, the two dtypes would stay in sync:
```suggestion
            footprint = np.zeros((2*height + 1, 2*width + 1), dtype=data.dtype)
```

testing/api/test_process_seg.py
```diff
@@ -165,64 +276,55 @@ def test_fix_orientation(test_orient):
...
+
+    # 2. test with angle IS
+    dummy_segmentation(size_arr=(32, 32, 5), pixdim=(1, 1, 5), angle_IS=15, debug=DEBUG) +
+    [{'angle_corr': False}],
+
+    # 3. test with ellipse shape
+    dummy_segmentation(size_arr=(64, 64, 5), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=0, debug=DEBUG) +
+    [{'angle_corr': False}],
+
+    # 4. test with int16. Different bit ordering, which can cause issue when applying transform.warp()
```
mguaypaq, line 310: This is currently not actually testing `int16` (and wasn't before). If we fix it to actually test `int16`, it's possible the test will start failing.

(I'm not sure what the comment about `transform.warp` means.)

testing/api/test_process_seg.py
```diff
@@ -165,64 +276,55 @@ def test_fix_orientation(test_orient):
...
+
+    # 7. test with RL and AP angled spinal cord
+    dummy_segmentation(size_arr=(64, 64, 50), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=-10, angle_AP=15, debug=DEBUG) +
+    [{'angle_corr': True}],
+
+    # 8. Reproduce issue: "LinAlgError: SVD did not converge".
+    dummy_segmentation(size_arr=(64, 64, 50), shape='ellipse', radius_RL=13, radius_AP=5, angle_RL=-10, angle_AP=30, debug=DEBUG) +
+    [{'angle_corr': True}],
+
+    # 9. test uint8 input
```
mguaypaq, line 333: Again, this comment is a bit misleading (currently `dtype=uint8` is silently ignored).

testing/api/test_process_seg.py
```diff
@@ -165,64 +276,55 @@ def test_fix_orientation(test_orient):
...
+    dummy_segmentation(size_arr=(32, 32, 50), dtype=np.uint8, angle_RL=15, debug=DEBUG) +
+    [{'angle_corr': True}],
+
+    # 10. test all output params
```
mguaypaq: It used to be that `expected` had more keys/values checked for this test case, but now all the tests have the same keys in `expected`. We can still keep this test, but maybe with an updated description?
```suggestion
    # 10. test bigger slice
```

testing/api/test_process_seg.py
```diff
@@ -165,64 +276,55 @@ def test_fix_orientation(test_orient):
...
-                        debug=DEBUG),
-     {'area': 4701, 'angle_AP': 0.0, 'angle_RL': 0.0, 'diameter_AP': 60.0, 'diameter_RL': 100.0, 'eccentricity': 0.8,
-      'orientation': 0.0},
-     {'angle_corr': False}),
-    # test with one empty slice
-    (dummy_segmentation(size_arr=(32, 32, 5), zeroslice=[2], debug=DEBUG),
-     {'area': np.nan},
-     {'angle_corr': False, 'slice': 2})
-    ]
+    # 0. test area
```
mguaypaq, line 294: Thank you for numbering the tests! This should make it easier to match the test cases with `pytest`'s output.

testing/api/test_process_seg.py
```diff
@@ -231,21 +333,64 @@ def test_compute_shape(im_seg, expected, params):
...
-        # fetch obtained_value
-        if 'slice' in params:
-            obtained_value = float(metrics['area'].data[params['slice']])
+        # If we're testing angle values, ensure the values are within half a degree
+        # If we're testing distances (area, diameter, length), ensure the values are within 3% of the expected value
+        kwargs = ({'abs': 0.5} if key.startswith('angle_') or key == 'orientation' else
+                  {'rel': 0.030})
+        # FIXME: Try to make the 0.03 tolerance tighter -- it's still too large.
+        #        Notes (what breaks when `rel` is decreased):
+        #          - 0.030: test9 - mean(area) - 73.108 vs. 77 +/- 2.31 <- Known, will be fixed before merging
```
mguaypaq: Is this line still accurate? Do you mean merging #5180, or #5129?

testing/api/test_process_seg.py
```diff
@@ -165,64 +276,55 @@ def test_fix_orientation(test_orient):
...
+    dummy_segmentation(size_arr=(32, 32, 50), dtype=np.uint8, angle_RL=15, debug=DEBUG) +
+    [{'angle_corr': True}],
+
+    # 10. test all output params
+    dummy_segmentation(size_arr=(128, 128, 5), pixdim=(1, 1, 1), shape='ellipse', radius_RL=50, radius_AP=30, debug=DEBUG) +
+    [{'angle_corr': False}],
+
+    # 11. test with one empty slice
+    dummy_segmentation(size_arr=(32, 32, 5), zeroslice=[2], debug=DEBUG) +
+    [{'angle_corr': False, 'slice': 2}]
```
mguaypaq: `'slice'` is no longer used by the test.

testing/api/test_process_seg.py
```diff
@@ -231,21 +333,64 @@ def test_compute_shape(im_seg, expected, params):
...
+            #  [ ][X][ ][ ][ ]       [ ][X][ ][ ][ ]
+            #
+            # So, while ideally all slices would have an angle value exactly like the cord we generated, in practice this
+            # value can be off by a few degrees. So, it's not feasible to test the per-slice angle to the tight tolerance.
+            # Furthermore, the angle values have a direct effect on the angle correction applied to the metrics, meaning if
+            # angle correction is on, the per-slice metrics will stray from the expected values just like the angles do.
+            # However, we can still test:
+            #    1. The mean values across all slices
+            #    2. The property that each subsequent value is within a tolerance of the previous value
+            #       (i.e. that the values don't jump around wildly from slice to slice)
```
mguaypaq, line 393: I like this idea.

testing/api/test_process_seg.py
```diff
@@ -231,21 +333,64 @@ def test_compute_shape(im_seg, expected, params):
...
-        else:
-            expected_value = pytest.approx(expected[key], rel=0.05)
-        assert obtained_value == expected_value
+                for n_slice in slice_range:
+                    # fetch expected/obtained values
+                    expected_value = expected[key]
+                    obtained_value = metrics[key].data[n_slice]
+
+                    # if n_slice is empty (i.e. `zeroslice` was set) all metrics should be `np.nan`
+                    if im_seg.data[:, :, n_slice].sum() == 0:
```
mguaypaq: I think this would fit the intent more closely:
```suggestion
                    if np.all(im_seg.data[:, :, n_slice] == 0):
```

mguaypaq (approved):
This is great, thanks! Here's a few final suggestions.

testing/api/test_process_seg.py
joshuacwnewton: I'm not sure either. Looking into the history of this test via `git blame`, it looks like this comment was added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2200, specifically commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/cf85f228e01b33dd08a5acbc4c1ce6ae86fa5af3. 

From the context, in that PR, the commit seems tied to this specific bugfix: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2200/changes/2503ecc16b1a481b6eb51dfdeef713f9b58f8013

So, I guess prior to this line of code, if inputs were int16, the behavior described was observed, and the test checks to see if we're insensitive to dtype.

There doesn't seem to be an associated issue for this? Maybe just something that was observed during the development of the PR. But at the very least I can add a link to the commits in the comment.

testing/api/test_process_seg.py
joshuacwnewton: Bahaha, turns out the problem is even worse than this, since not only do the rotation steps modify the dtype to be `float64`, we also just... explicitly cast the final rotated array to float32 anyway:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/1a608486b84a46e25bf8e452f699bc2af67b13fe/testing/api/test_process_seg.py#L148

I'll make sure the dtype is preserved in the output, hehe

testing/api/test_process_seg.py
joshuacwnewton: I had that thought initially, too! I thought it was infeasible, though? See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5180#discussion_r2933102234

```
    #            - How to generate the correct SI rotation (this should be equivalent to 'orientation', but isn't)
````

testing/api/test_process_seg.py
joshuacwnewton: Whoops, meant #5129. I'll update to clarify.

testing/api/test_process_seg.py
mguaypaq: Right: there are theoretical and practical problems with computing a "final" `angle_IS` from `rot.apply([0, 0, 1])`, but here I'm suggesting that we just stick the _input_ `angle_IS` into `'orientation'`. My thinking is:
- Only one test (test number 2) uses `angle_IS`, and it doesn't do other rotations. It would be nice if, at least for this test, we had an expected `'rotation'` value.
- I _think_ all other tests would expect a `'orientation'` of zero degrees.
- Since the `angle_IS` rotation is applied _first_, I think it's equivalent to applying this rotation _last_, but using the final tilted centerline as the axis of rotation. In which case, it should be a reasonable approximation for `'orientation'` (but maybe with a looser tolerance than the other metrics.)

But, if this seems too fragile, feel free to ignore it.

mguaypaq: I guess it could also be conditionally inserted in `expected`, only when `angle_IS != 0` and/or the other angles are `== 0`.

mguaypaq (approved):
Thanks for all the fixes!

testing/api/test_process_seg.py
joshuacwnewton: Good call! Yes, orientation is predictable in the easy cases (15 or 0 degrees) as you say. 

Added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5180/commits/a09186d968059fafabb30a90844f451acb8f7c04.

joshuacwnewton: If this last change looks good, feel free to merge, and then @sandrinebedard can return to the PR to fix the double-cropping issue (making sure the tests pass for any new changes we add).

PR #5187, opened 2026-03-12
Output debug imgs are saved in the same folder as the csv file

plbenveniste:
I made the following changes so that the output debug imgs are saved in the same folder as the csv file.
It's more convenient for my work

Tagging @sandrinebedard. I am not sure that this is relevant in general but it is for me so I made the changes. Feel free to merge it or not


plbenveniste:
Closing this PR since this will not be integrated into SCT

PR #5192, opened 2026-03-20
Fix several bugs with `find_angle_hog` function to make computed HOG values consistent

### LINKED ISSUE #5173: Registration with `algo=centermassrot,rot_method=hog` shows inconsistent rotation vs. new HOG-based symmetry metrics
## Description

Comparing the angles given by HOG during registration vs when using it the `sct_process_segmentation`, it looks like there should be some rotation, but registration shows no rotation compared to QC of `sct_process_segmentation`

### Steps to Reproduce

0. Install SCT v7.2

1. Get data from  spine generic data-multi-subject `sub-queensland06`

2. Run registration & warping (for QC)
~~~
sct_register_to_template -i sub-queensland06_T2w.nii.gz -s sub-queensland06_T2w_seg.nii.gz -c t2 -ldisc sub-queensland06_T2w_seg_labeled_discs.nii.gz -ofolder reg_hog -param "step=1,type=imseg,algo=centermassrot,poly=0,slicewise=1,rot_method=hog" -v2 -r 0
sct_warp_template -d sub-queensland06_T2w.nii.gz -w reg_hog/warp_template2anat.nii.gz -a 0 -ofolder reg_hog -qc qc
~~~
3. Run `sct_process_segmentation`

~~~
sct_process_segmentation -i sub-queensland06_T2w_seg.nii.gz -anat sub-queensland06_T2w.nii.gz -qc ./qc -v 2
~~~

### Results

- The QC of `sct_process_segmentation` (and the `-v 2` figure) shows a lot of rotation:
    <details>
    <summary> Debug figure </summary>

    Image
    </details>
- However, the output of `sct_register_to_template` doesn't show much rotation in the outputs by comparison (warp_template QC):
  Image

### Testing on intermediate step

As the angle looked different, I also looked at the intermediate step `src.nii` (image in the starightened space, 0.5x0.5x2mm) and ran sct_process_segmentation, here are the debugging plots:

<details>
<summary><b> ANGLE WITH process_segmentation </b></summary>

Image
</details>

<details>
<summary><b> Angle with sct_register_to_template </b></summary>

Image

</details>

### Ideas

- Maybe it could be linked to resamapling? it is the only difference I can think of! 
- Test QC for resampling: [qc.zip](https://github.com/user-attachments/files/25530245/qc.zip)
  * First row of QC --> image resampled at 0.1 in plane
  * Second row --> 0.5 in place

joshuacwnewton:
## Checklist

## Description

@sandrinebedard keenly noticed that the HOG values produced by `sct_process_segmentation` did not match the rotation for `sct_register_to_template -param rot_method=hog`, despite using the same function internally (`find_angle_hog()`).

What followed was the uncovering of several bugs in how the `find_angle_hog()` function was written and used, namely:

- Mix of degrees and radians in angle values.
- Mirroring due to a negative sign.
- Off-by-one errors.
- An extra `/2` operation.

I've tried to isolate all of the bugfixes into different commits, with in-depth extended commit messages and comments to highlight the problem, but the function is fairly complex, so some additional eyes would be appreciated..

## Linked issues

Fixes #5173.


spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1587,15 +1596,22 @@ def find_angle_hog(image, centermass, px, py, angle_range=radians(10)):
...
+    # get the orientation histogram values only within those points
     grad_orient_histo_conv_restrained = grad_orient_histo_conv[index_lower:index_upper]
 
     # Finding the symmetry axis by searching for the maximum in the autoconvolution of the histogram :
-    index_angle_found = np.argmax(grad_orient_histo_conv_restrained) + (nb_bin // 2 - index_restrain)
-    angle_found = repr_hist[index_angle_found] / 2
+    index_angle_found = np.argmax(grad_orient_histo_conv_restrained)
+    # Shift the index back to the original histogram index range
+    index_angle_found = index_angle_found + index_center - index_restrain
+    angle_found = repr_hist[index_angle_found]
```
joshuacwnewton: Regarding the extended commit message in d278c057fc90371dcf6aee781e1b6dccf98358c8, I did end up running both sct_process_segmentation and sct_register_to_template on `sub-queensland06` and found that the extra bugs weren't as severe as I was worried they would be:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5173#issuecomment-4100694411

So take that commit message with a small grain of salt.

sandrinebedard (commented):
Thanks @joshuacwnewton for all of this!! I tested on one subject and compared with PCA, it looks good! Would you like me to test on more subjects and QC to make sure it is okay?

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -9,7 +9,7 @@
 import os  # FIXME
 import shutil
 import psutil
-from math import asin, cos, sin, acos
+from math import asin, cos, sin, acos, radians
```
mguaypaq, line 12: I really should have known about `math.radians()` and `math.degrees()`, but TIL!

spinalcordtoolbox/registration/algorithms.py
mguaypaq: This step should not be necessary:
- The algorithm from the original paper explicitly ignores the magnitude of the gradients, and only considers their angles.
- The convolutions below use weights near 1, so there should be no overflow problems.
- `np.arctan2(y, x)` implicitly takes the ratio of two gradients, which nullifies the effect of any prior rescaling.

However, the modified algorithm implemented here _does_ consider the magnitude of the gradients (maybe unwisely?), so it may have an effect.

mguaypaq: EDIT: Viewing the first array coordinate as $y$ and the second array coordinate as $x$ seems to be an image processing convention, rather than a mistake.

These two are swapped, but it's quite possible that this ~~mistake~~ convention cancels out with another ~~mistake~~ convention later on:
- `gradx` is actually the gradient along the second array coordinate.
- `grady` is actually the gradient along the first array coordinate.

Also, they're poorly scaled, but in a way which wouldn't affect isotropic slices:
- To express the correct rate of change _per voxel_, they should be divided by 2. (But this wouldn't affect the algorithm.)
- To correctly compute angles, the two rates of change should be _per physical unit_, so they should be further divided by `px` and `py`, respectively (which are not passed to this function). (This would only affect anisotropic images.)

mguaypaq: Contrary to this code comment, weighting by magnitude doesn't alter the angles: it changes how much each angle contributes to the histogram.

But, it's a major deviation from the original published algorithm, and it's not clear to me that it's a good idea.

mguaypaq: This is a very wacky way of picking bins for the histogram.

Also, the calls to `np.concatenate` are unnecessary.

I imagine this is an attempt at having the "zero" bin centered at zero (rather than having zero as the left edge), but the way it's done means that the $\pm 180^\circ$ bin is just nonexistent.

mguaypaq: Instead of reconstructing the bins with an error-prone `np.linspace` outside of this function, we should probably return the ones returned by `np.histogram` above.

mguaypaq (commented):
So far I've only read the original paper by Sun (it's pretty short and straightforward) and tried to figure out the code for `gradient_orientation_histogram()`. Here are some observations.

spinalcordtoolbox/registration/algorithms.py
mguaypaq: I think these kernels essentially do interleaving in the gradient axis and smoothing in the perpendicular axis, which is maybe a little weird.

joshuacwnewton: Yes, that's exactly what I witnessed too!

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/dbfa619d3edc7cf6cc32f0ab585e0a1da48ff801/spinalcordtoolbox/registration/algorithms.py#L1579-L1587

Thank you for looking more deeply into the underlying `gradient_orientation_histogram` function. :hearts:

Nicolas-Pinon: Hi there,
- Indeed, this is not an implementation of the original algorithm by Sun et al. but an ""improvement"" (at least an attempt of improvement!), the idea was to have the original algorithm improved in two ways by weighting the contribution of each "orientation gradient" by 1) its magnitude 2) its distance to the spinalcord center
- The scaling step seem pretty standard to me, I guess any other type of scaling could be used (min-max, mean-std, etc.).
- It's the case for arctan because the Sobel filter is linear, but I'm not sure it would be nullified for the magnitude computation ?

Nicolas-Pinon: This deviation is indeed intented, whether this is a good idea or not is a more complex discussion in my humble opinion. The idea was that images are often more defined by their contour. The orientation of pixels that are in a "monocolour" area seems irrelevant for the computation of a rotation angle, i.e. they are just noise.

Nicolas-Pinon: You are absolutely right, there is a smoothing in the perpendicular axis, it is the Sobel filter, which I believe is quite standard for gradient computation.

mguaypaq: Thanks for the name and the context!

mguaypaq: Fair enough! It's good to know that it was intentional (sometimes it can be hard to tell).

Nicolas-Pinon: For sure !

sandrinebedard:
I went through some subjects quickly and also observe that there is more rotation then before :
`sub-pavia05`, (mainly look at slice numbers 95 to 82)
<details>

<summary>HOG vs PCA registration (more rotation on HOG then PCA): </summary>

ezgif com-animated-gif-maker (19)

</details>

<details>
<summary>HOG:</summary>

ezgif com-animated-gif-maker (19)


</details>


<details>
<summary>PCA:</summary>

ezgif com-animated-gif-maker (20)
</details>

When looking at the previous HOG angles vs PCA for this subject, the angles were lower for HOG, than PCA:

<details>
<summary>HOG:</summary>
image
</details>


<details>
<summary>PCA:</summary>
image

</details>

mguaypaq:
I'm making good progress on refactoring the code and adding some tests. While doing this, I also noticed that:
- We may not need to require an even number of bins after all.
- The `circular_conv` and `circular_filter_1d` are no longer needed, since the `scipy.ndimage` functions now support `mode='wrap'` natively (see, for example [`median_filter()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.median_filter.html)).

spinalcordtoolbox/registration/algorithms.py
mguaypaq: When computing the `conf_score`, this line incorrectly assumes that `arg_maxs[1]` is the _second highest_ peak (in fact, `scipy.signal.argrelmax` seems to return the detected peaks in order of index, _not_ in decreasing order of height).

Fortunately, no code in SCT has ever _used_ the returned `conf_score`, so I'll just remove it, which will make the code simpler and faster.

testing/api/test_angle_hog.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+            computed_degrees = degrees(find_angle_hog_2(rotated_slice, **kwargs))
+            diffs.append(abs_degree_diff(
+                original_degrees + expected_degrees,
+                computed_degrees,
+            ))
+        diffs = np.array(diffs)
+
+        assert diffs.max() <= 16
+        assert diffs.mean() <= 8
```
mguaypaq: This test performance is pretty disappointing. Maybe there's a bug? But there are slices where the max and the mean are much smaller, which kind of validates the code. Also, the helper functions all pass their tests.

Some other possibilities:
- Maybe the angle_hog method is not reliable with the kinds of resolution and SNR that we're dealing with.
- Maybe the algorithm parameters need tweaking (`sigma`, `bins`, `angle_range`, `kmedian_size`).
- Maybe some of the modifications need tweaking (smoothing function used, weighting by centermass distance, weighting by gradient magnitude).
- Maybe the resampling that happens while rotating the test image is somehow unfair (but if the algorithm can't handle that, is it good enough?).

I'm at a bit of a loss here. Any ideas, @joshuacwnewton?

joshuacwnewton: Focusing on the `test_find_angle_hog_2` test, and honing in on just the first axial slice in the test, I'm seeing the following `diffs` array, which at least paints a slightly less scary picture (7.5 max, 1.5 mean), but does suggest that the method is susceptible to outlier `diffs`:

Image

My investigation here is a bit long, so I'm including it in a `<details>` tag:

<details>

----

Setting a breakpoint to examine that outlier case (7.5), I'm seeing:

Image

So, we've got a 3* angled cord, rotated an extra 8*, and so we expect 11* but get 3.5* instead. 

Image

Image

Just to put this 3.5* value into context, it does look very much like an outlier next to the adjacent angle tests, too (i.e. the measured angle matches the expected angle in the 9-11* range, EXCEPT for the outlier 3.5* angle:

Image

----

Drilling down into `find_angle_hog_2` -> `weighted_orientation_histogram`, I see the Sobel-filtered images (`gradient_PA`, `gradient_RL`) as:

Image

Image

The arctan (orientation) prior to centermass weighting:

Image

The arctan (orientation) after centermass weighting (simple `weight*orientation` for sanity-checking):

Image

Just visually, the individual gradient images look informative, but the gradient orientation at the center of the cord looks very noisy. Still, if I plot the raw `hist`/`hist_smooth`/`hist_convo` in a way that matches figure (c) of the 1995 paper, I see:

Image

Image

Image

The angle immediately after this 3.5* outlier (i.e. the 11* detected angle) shows this instead:

Image

Image

Image

I was wondering how this `hist_convo` is processed. Focusing just on this code:

```python
    # Because of the half-steps, we need to go around the circular
    # auto-convolution twice to cover the entire histogram.
    # One time covers the angles in the range -np.pi <= angle < 0, and the
    # second time covers the angles in the range 0 <= angle < np.pi.
    # This also corresponds to the fact that two axes of symmetry that are
    # 180 degrees apart are actually the same axis of symmetry.
    scores = []
    for index in range(-hist_convo.size, hist_convo.size):
        angle = index * half_width

        # only consider angles within the desired range
        if abs(angle) <= angle_range:

            # negative indices wrap around
            score = hist_convo[index]

            # when taking the maximum, we want the highest scores, then break
            # ties towards the smallest absolute angle, then take the positive
            # angle rather than the negative one if there's still a tie
            scores.append((score, -abs(index), angle))
```

Some notes:

- `hist_convo` is of size `360`
- Search range is `[-360, 360]`
- Angle range is `40`
- This limits the actual angles being checked to `[-20, 20]` degrees (due to the half-steps)
- When re-plotting the these angle values as a new histogram, we get:

Image

Ahhhh, so basically, by performing this wrap around, we're really only looking at this tiny section of the full histogram for +/- 20 degrees ([0,40] and [320,360] in the original histogram):

search_range (1)

When comparing just this segment to the "correct" histogram, then yes, we can see that the two histograms are very different. Instead of 1 single spike at a single angle value, there are a bunch of spikes!

And, looking at the change from `hist_smooth` to `hist_convo`, I think the spikiness is introduced by the circular convolution?

</details>

My first impression is that the method doesn't need to immediately be thrown out, and does seem to do what it says on the tin **except** for the circular convolution step. So, I think I agree about tweaking the parameters.

joshuacwnewton (commented):
I tried to do my best to debug the behavior in the re-implemented `find_angle_hog` function.

Thankfully, with the help of PyCharm's "View array as image" functionality, debugging this felt a lot easier than I normally expect, albeit taking me all day 😅

testing/api/test_angle_hog.py
mguaypaq: Thank you! That's extremely informative. Based on your pictures, I would try:
- Increasing the `sigma` value from 10 to 20 or even 30, so that a larger circle of gradient information survives the centermass weighting.
- Maybe changing the kmedian filtering to a gaussian filtering (which was an unused variant that existed in the previous code already), so that the histogram peaks are less "flat" (so that there's a clearer value for the argmax).

Changing the `sigma` value seems like the easier thing to do, if you want to try it on your own.

However, after a quick discussion with @NathanMolinier, we should be careful with changing `sigma`:
- Currently, the algorithm looks mostly at the spinal cord (the dark circle in the center), which is maybe the intention.
- Enlarging `sigma` would mean looking at the canal (including the light shape around the spinal cord, which is the CSF).
- This touches on a research/clinical question, because in some pathologies, the two structures can be rotated by different amounts. Then, we should have a clear idea of which one we want to measure.

@sandrinebedard or @plbenveniste, do you have thoughts on whether we want the symmetry of just the cord, or the entire canal?

mguaypaq: Incidentally, if we decide to keep `sigma=10`, that means we are only putting ~100 gradient orientations in the histogram, so having 360 bins is maybe excessive. Considering that the auto-convolution produces half-step resolution (so, double the number of bins), we could have single-degree resolution in `find_angle_hog_2` with only 180 bins.

sandrinebedard: > Enlarging sigma would mean looking at the canal (including the light shape around the spinal cord, which is the CSF).

Yes we want the canal and surrounding! if it is only the cord, why not just use the PCA.  Having outside tissues helps to detect assymmetry of the cord (which is the goal)

joshuacwnewton: Dumb question, but: Why do we use the centermass as the weight, as opposed to just, say, using the full seg (perhaps dilated a tad) as the mask? (This is information we could have in the context of this function, right?)

When I visualized the mask we were using, it looked to be a circle... 

image

But depending on the size/shape of the cord, wouldn't this over-weigh the surrounding AP tissue and under-weigh the LR-most parts of the cord?

If instead I pass the `seg` slice along:

image

And then I dilate a bit (to capture the canal) then convolve with something like sigma=2, I get:

image

image

But then I tried on the test and for some slices, it was actually worse than the current diff/mean values of `16` and `8`...

plbenveniste: > @sandrinebedard or @plbenveniste, do you have thoughts on whether we want the symmetry of just the cord, or the entire canal?

In my case of focal atrophy, looking at the entire canal could help provide more robust results. Asymmtrical atrophies (such as showed in https://github.com/sct-pipeline/critical-ms-lesion/issues/24#issuecomment-4288850606) could corrupt the results and result in fake SC rotation. 

> Why do we use the centermass as the weight, as opposed to just, say, using the full seg (perhaps dilated a tad) as the mask? (This is information we could have in the context of this function, right?)

I have pros and cons about this. Using the whole mask could limit the errors in identifying the center of the SC: using the center-of-mass is problematic in scenarios of focal atrophy. On the downside, using the entire mask, would mean that the contour of the mask would be "guiding" the gradients which is not great in scenarios of focal atrophy.

sandrinebedard:
@mguaypaq @joshuacwnewton should I try the method on spine generic again? or do we still have some parameters to adjust before?

testing/api/test_angle_hog.py
valosekj: > But depending on the size/shape of the cord, wouldn't this over-weigh the surrounding AP tissue and under-weigh the LR-most parts of the cord?

This is a good point, indeed. Especially in compressed cord, where the "circle-like" mask might under-weigh voxels in the LR-most parts of the cord.

joshuacwnewton: Replying to @sandrinebedard's comment [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5192#issuecomment-4372237506) just to keep the conversation tidy:

> or do we still have some parameters to adjust before?

Given that the output results seem to be very sensitive to mask shape (I tested with SC mask, but I think that's probably an OK proxy for sigma), I think we will still need to tweak the filtering approach too, as mentioned by @mguaypaq.

mguaypaq:
Ok, with the latest commits, I was able to really reduce the worst and average errors detected by the automated test (6 degrees max, 3 degrees average). I made each change as a separate commit, and updated the test values each time to the best integer that worked, to see the effect each time.

@joshuacwnewton if it's quick to reproduce and you have some time, I'd love to see the same centermass weighted orientation plots as before for the last two commits (sigma=20 and sigma=30).

@sandrinebedard I changed the two places that called `find_angle_hog` and replaced them with `find_angle_hog_2`. So, you should now be able to try it again on spine generic.

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1743,7 +1743,7 @@ def weighted_orientation_histogram(
     px: float,  # size of each voxel along the RL axis, in physical units
     py: float,  # size of each voxel along the PA axis, in physical units
     centermass: tuple[float, float],  # center of mass of the region of interest
-    sigma: float = 10.0,  # radius for gaussian weighting around the centermass
+    sigma: float = 20.0,  # radius for gaussian weighting around the centermass
```
joshuacwnewton: Testing commit a2f311ac9d2831a00ceed75d0df4a2e7b92a16a7 for:

- `iz == 0` and `extra_degrees == 8` (same as [comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5192#discussion_r3164024455))
- (manually editing back in the `[-10, 11]` test range for comparison sake)

<details>

> `diffs` (`18`: `7.5 -> 1.0`) :tada: :
>
> image

> `hist`:
>
> image

> `hist_smooth`:
>
> image

> `hist_convo`:
>
> image

> `[-40, 40]`
>
> image

</details>

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1743,7 +1743,7 @@ def weighted_orientation_histogram(
     px: float,  # size of each voxel along the RL axis, in physical units
     py: float,  # size of each voxel along the PA axis, in physical units
     centermass: tuple[float, float],  # center of mass of the region of interest
-    sigma: float = 20.0,  # radius for gaussian weighting around the centermass
+    sigma: float = 30.0,  # radius for gaussian weighting around the centermass
```
joshuacwnewton, line 1609: Testing commit 43aee78541c91efbf608a64b17a66458c2437456 for:

- `iz == 0` and `extra_degrees == 8` (same as [comment](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5192#discussion_r3164024455))
- (manually editing back in the `[-10, 11]` test range for comparison sake)

<details>

> `diffs` (`18`: `7.5 -> 1.0`) :tada: :
>
> image

> `hist`:
>
> image

> `hist_smooth`:
>
> image

> `hist_convo`:
>
> image

> `[-40, 40]`
>
> image

</details>

joshuacwnewton (commented):
> @joshuacwnewton if it's quick to reproduce and you have some time, I'd love to see the same centermass weighted orientation plots as before for the last two commits (sigma=20 and sigma=30).

Sounds good! I'll add my findings in the comments below (just opening them here for tidiness).

spinalcordtoolbox/registration/algorithms.py
joshuacwnewton: @mguaypaq I'm not seeing a significant difference  between 20.0 and 30.0 in terms of this individual plot, but I could check out the biggest outlier value for 20.0 (since I saw that switching to 30.0 caused the `max` to decrease from 9 to 6).

joshuacwnewton: The failure case (`diff == 9.0`) for `sigma = 20.0` occurs when `iz = 33` and `extra_degrees = -20`. In that case, I'm seeing the image plots:

<details>

image

image

image

image

image

image

image

</details>

And the histogram plots:

<details>

> image

> image

> image

> image

</details>

Seems like maybe an odd outlier where there are dual peaks?

For reference, here's the (improved) final histogram for that slice/angle pair for `30.0`:

> image

Basically the same plot, just with a marginally higher peak for the correct angle.

joshuacwnewton: I can also push my plots as a debug commit on this PR if you want to play around with them yourself. :)

(EDIT: Added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5192/commits/728ccc2e75e5de49794f0b0ea81b6016e8492b61.)

mguaypaq: To help with intuition, here are some quick plots to illustrate what the algorithm sees for `sigma in [10.0, 20.0, 30.0]`. This is for the same case as your initial investigation (`iz == 0` and `extra_degrees == 8`). This is just `weight*orientation`, without normalization by `magnitude`.

* `sigma == 10.0`: 

  sigma10
* `sigma == 20.0`:

  sigma20
* `sigma == 30.0`:

  sigma30

Based on this, I'm not surprised that `10.0` is too little, `20.0` is almost enough, and `30.0` seems sufficient.

mguaypaq: Here are the same plots including `magnitude`, that is, `weight*magnitude*orientation`:

* `sigma == 10.0`:

  mag10
* `sigma == 20.0`:

  mag20
* `sigma == 30.0`:

  mag30

sandrinebedard:
I ran the latest commits on spine generic, I've looked at a lot of image and it looks pretty good  in general, very similar to the PCA method, so I think this is good!

<details>
  <summary>Here is sub-pavia05 that I showed previously:</summary>
ezgif com-animated-gif-maker (23)

</details>

I have noticed some subjects that it has slightly worse performance:

<details>
  <summary>sub-vuiisAchieva03</summary>

ezgif com-animated-gif-maker (22)
</details>

<details>
  <summary>sub-mgh01</summary>
No rotation at slices 26 - 52, there is rotation with PCA (when looking at the rootlets, I think there should be some rotation)

HOG:
ezgif com-animated-gif-maker (25)


PCA:

ezgif com-animated-gif-maker (26)

</details>

I think maybe we should investigate a bit more the mask size, consider the size of the cord mask maybe!

The QC report is located here: `duke/projects/dcm-normalization/symmetry/symmetry_test_spine_generic_2026-05-05-v2/qc`

mguaypaq:
It looks like the current version (centermass sigma of 30.0, 180 bins, gaussian smoothing) performs well enough to merge. I'll add some commits to switch to the refactored code, and to remove the old code, then we can merge. Thanks everyone!

spinalcordtoolbox/registration/algorithms.py
sandrinebedard: @mguaypaq can you remind me why we don't use the cord mask anymore?

joshuacwnewton: That was just something I tried for demonstration, but because I didn't get very good results, I didn't commit it.

However, I only tested that 1 change in isolation. Might be worth taking a second look at given the other new changes, though? I can commit it if @mguaypaq wants to try it.

mguaypaq (approved):
@joshuacwnewton I'll let you take a final look before merging if you want, but I'll formally approve the PR now (since you can't do it as the PR opener).

As @sandrinebedard noted, a nice future improvement might be to pick the centermass sigma proportionally to the physical (millimeter) size of the segmentation, but that's for the future.

spinalcordtoolbox/registration/algorithms.py
mguaypaq: The original code didn't really use the cord mask (it only used it to compute centermass coordinates), and this seemed like a bigger change than we want in this PR.

I agree with Pierre-Louis's comment https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5192#discussion_r3180453205, that using a non-circular mask could bias the algorithm. But, I haven't tested this, so it could be considered in future improvements.

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -823,16 +822,24 @@ def register2d_centermassrot(fname_src, fname_dest, paramreg=None, fname_warp='w
...
+                    px, py,
+                    centermass_src[iz, :],
+                    angle_range=th_max_angle,
+                )
+                angle_dest_hog = find_angle_hog(
+                    data_dest_im[:, :, iz],
+                    px, py,
+                    centermass_dest[iz, :],
+                    angle_range=th_max_angle,
+                )
```
joshuacwnewton, line 836: Minor nitpick, but for making the diff easier to parse for the future, do we want to match the function signatures for the old/new implementations?

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1540,153 +1547,145 @@ def find_index_halfmax(data1d):
     return xmin, xmax
 
 
-def find_angle_hog(image, centermass, px, py, angle_range=10):
+def normalized_sobel(
+    image: np.ndarray,  # can be any-dimensional
+    axis: int,  # the direction in which to take the partial derivative
+    p: float,  # the size of each voxel along the given axis, in physical units
+) -> np.ndarray:
     """
```
joshuacwnewton, line 1557: Likewise, do we want to organize the functions so that `find_angle_hog` is in the same place before/after this PR?

spinalcordtoolbox/registration/algorithms.py
```diff
@@ -1540,153 +1547,145 @@ def find_index_halfmax(data1d):
...
+        range=(0, 2*np.pi),
+        weights=magnitude*weight,
+    )
 
-    signal_conv = signal_conv_extended[len(signal1):2*len(signal1)]  # truncate back the signal
+    return hist
 
-    return signal_conv
 
+def circular_autoconvolution(array: np.ndarray) -> np.ndarray:
```
joshuacwnewton, line 1654: Ditto for `circluar_conv` vs. `circular_autoconvolution`.

spinalcordtoolbox/process_seg.py
```diff
@@ -235,8 +235,11 @@ def _properties_image(im_r, nz, px, py, pz, pr, min_z_index, max_z_index, proper
         # TODO: explore different sigma values for the HOG method, i.e., the influence how far away pixels will vote for the orientation.
         # TODO: double-check if sigma is in voxel or mm units.
         # TODO: do we want to use the same sigma for all slices? As the spinal cord sizes vary across the z-axis.
-        angle_hog, conf_src = find_angle_hog(current_patch_im_scaled, centermass_src,
-                                             px, py, angle_range=40)    # 40 is taken from registration.algorithms.register2d_centermassrot
+        angle_hog = find_angle_hog(
+            current_patch_im_scaled,
+            px, py,  # TODO: is this the real resolution of current_patch_im_scaled?
```
joshuacwnewton: Did we ever resolve this TODO?

joshuacwnewton (commented):
@mguaypaq I think once the values in the tests are revised, this should be good to merge!

I only had some minor nitpicks about the code organization, mainly to minimize the overall diff and make it easier to re-read this PR should we ever need to in the future.

spinalcordtoolbox/process_seg.py
mguaypaq: Ah, no, I was hoping someone else was already familiar with this part of the code, but I'll look into it. Thanks for catching this!

mguaypaq:
Ok, now all that should remain is updating the batch_processing.sh test values once the current test run completes.

PR #5199, opened 2026-04-08
Implement `length_anterior` and `length_posterior`

valosekj:
## Checklist

## Description

This pull request implements two new morphometrics into `sct_process_segmentation`: `length_anterior` and `length_posterior` (`a` and `b` in the figure below). These measurements are inspired by Fig. 1 of [this paper](https://www.mdpi.com/2077-0383/12/12/4111):

image

The computation is based on the recently updated AP diameter measurement introduced in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958.

The anterior and posterior lengths are computed by splitting the AP diameter. The splitting is done by the cord center of mass, which lies on the major (RL) axis (thus corresponding to `w` in the figure above). Implementation [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d03cdc6a4eb01e9abfec0c30df6cd260275e166/spinalcordtoolbox/process_seg.py#L482-L492).

`length_anterior` and `length_posterior` could be used, for example, [to compute asymmetry](https://github.com/sct-pipeline/dcm-metric-normalization/blob/2106cb6da9d36e9c3e4b619d4a59b600337cfe6d/statistics/generate_figures_ap_lengths.py#L383): a normalized measure ranging from −1 to 1 that quantifies the relative difference between anterior and posterior cord extents, with negative values indicating anterior compression and positive values indicating posterior compression.

image

## Validation

<details>
<summary>Validation on 201 HCs from the spine-generic multi-subject dataset:</summary>

lineplot copy

</details>

<details>
<summary>Sample QC on a DCM patient (`dcm-zurich/sub-158`):</summary>

image

</details>

---

Tagging @sandrinebedard.


mguaypaq:
Just a comment about the naming: are "anterior diameter" and "posterior diameter" established terms in the literature? If not, could I suggest "anterior radius" and "posterior radius", since it's expected to be approximately half the diameter? (Or "length", but to me "diameter" specifically means the largest distance inside a shape.)

valosekj:
> Just a comment about the naming: are "anterior diameter" and "posterior diameter" established terms in the literature? If not, could I suggest "anterior radius" and "posterior radius", since it's expected to be approximately half the diameter? (Or "length", but to me "diameter" specifically means the largest distance inside a shape.)

Ah, excellent point. I'll use "length".

valosekj:
The computation works on a testing dataset (https://github.com/sct-pipeline/dcm-metric-normalization/pull/29). @mguaypaq and @joshuacwnewton, would you mind reviewing this PR? Especially if the implementation makes sense to you and if I haven't missed anything. 
Other changes concern mainly QC. Thank you!

<s>**This comment has been updated:** When further testing the PR, I realized that splitting the AP diameter based on the cord center of mass is actually different from splitting it based on the widest transverse line. I updated my original PR description. </s>

UPDATE: Okay, after some further investigation, it turned out that the RL diameter (major axis of the fitted ellipse) lies on the cord center of mass. I think this is more robust than getting the widest transverse line directly from the segmentation. So I switched back to the center of mass splitting in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/changes/f743737ec87207cac6f098885858f9394a121cbc.

joshuacwnewton:
> Hi @joshuacwnewton, just fyi, I think [#5199](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199) is now ready for a review.

Just retagging myself to make this PR show up in my GitHub notifications. Will look at today. :)

spinalcordtoolbox/process_seg.py
```diff
@@ -475,10 +479,24 @@ def _measure_ap_diameter(seg_crop_r, seg_crop_r_rotated, dim, angle, iz, propert
...
+        # For an ellipse, the center of mass lies on the major (RL) axis, so its AP coordinate equals the AP position
+        # of the widest horizontal line.
+        # Anterior length: distance from the cord center of mass to the anterior cord edge.
+        # Posterior length: distance from the cord center of mass to the posterior cord edge.
+        # Inspiration: Kang et al. J Clin Med 2023, https://doi.org/10.3390/jcm12124111 (Fig. 1).
+        # In RPI orientation, axis 1 (columns) is the PA direction: low index = posterior, high index = anterior.
+        ap_pixels_posterior = np.sum(seg_crop_r_rotated[indices, :ap0_r], axis=1).mean()
+        ap_pixels_anterior = np.sum(seg_crop_r_rotated[indices, ap0_r:], axis=1).mean()
+        length_posterior = ap_pixels_posterior * dim[1]
+        length_anterior = ap_pixels_anterior * dim[1]
```
joshuacwnewton: Just to make sure I understand the evolution of this PR... 

- The current calculations in this PR are the same as the calculations from the first commit in the PR (6e39a92bf7223cdf6f5da241c992be2eee907f8d), right? 
   - We use the AP diameter measure as a base.
       - _(Specifically, the "improved" `diameter_AP` based on the mean diameters from the 3mm extent around the CoM of the HOG-rotated SC segmentation (i.e. PR ([#4958](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958)), **NOT** the ellipse diameter.)_
   - The variable `ap0_r` (ap component of CoM) is used to split `diameter_AP` in two.
   - Thus, the property should hold that `length_posterior + length_anterior = diameter_AP`?
- And, you temporarily tested an alternate approach in 15bef9f45980cf562b9e9b43ceaee6cdd5100a31, then you reverted the changes in f743737ec87207cac6f098885858f9394a121cbc after finding that the original method was better.
   - The alternate approach used `argmax` of the pixel sums.
   - This is equivalent to finding the actual widest horizontal line.
   - But, this is not robust because of... for example, a compressed cord where `argmax(sum())` would be offset from the CoM? 

If that sounds correct, then I will base the rest of my review off this summary.

spinalcordtoolbox/process_seg.py
```diff
@@ -475,10 +479,24 @@ def _measure_ap_diameter(seg_crop_r, seg_crop_r_rotated, dim, angle, iz, propert
 
         ap_diameter = ap_pixels * dim[1]
 
+        # Compute anterior and posterior lengths by splitting the AP diameter at the cord center of mass.
+        # For an ellipse, the center of mass lies on the major (RL) axis, so its AP coordinate equals the AP position
+        # of the widest horizontal line.
```
joshuacwnewton: If I'm understanding correctly, we're _not_ working with the ellipse in this function, right? The CoM from `compute_pca` is based on the actual SC segmentation:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d03cdc6a4eb01e9abfec0c30df6cd260275e166/spinalcordtoolbox/process_seg.py#L439-L441

So, I am wondering if the "ellipse" part that I've highlighted above is not so accurate...

spinalcordtoolbox/process_seg.py
joshuacwnewton: Since the lengths are computed the same way as the AP diameter, I am wondering if it would be possible to simplify these calculations and de-duplicate some of the work:

Before:

```python
        # sum/average the pixels
        ap_pixels = np.sum(seg_crop_r_rotated[indices, :], axis=1).mean()
        ap_pixels_posterior = np.sum(seg_crop_r_rotated[indices, :ap0_r], axis=1).mean()
        ap_pixels_anterior = np.sum(seg_crop_r_rotated[indices, ap0_r:], axis=1).mean()
        # vox -> mm
        ap_diameter = ap_pixels * dim[1]
        length_posterior = ap_pixels_posterior * dim[1]
        length_anterior = ap_pixels_anterior * dim[1]
```

After:

```python
        # average the voxels across the RL extent to get the mean AP array
        ap_arr_3mm_extent = seg_crop_r_rotated[indices, :]
        ap_arr_mean = ap_arr_3mm_extent.mean(axis=0)
        # split the AP diameter at the cord CoM
        length_posterior = ap_arr_mean[:ap0_r].sum() * dim[1]
        length_anterior  = ap_arr_mean[ap0_r:].sum() * dim[1]
        ap_diameter      = length_posterior + length_anterior 

```

The idea here being that because we perform `mean -> sum` instead of `sum -> mean`, it allows us to reuse the `mean` array several times. Also, it conveys the relationship between the lengths and the `ap_diameter`.

I tested this locally and it produces the same result!

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -79,6 +79,10 @@ def get_parser(ascor=False):
             "  - angle_AP, angle_RL: Estimated angle between the cord centerline and the axial slice. This angle is "
             "used to correct for morphometric information.\n"
             "  - diameter_AP: Measured as average across 3 mm extent centred at cord mask center of mass.\n"
+            "  - length_anterior: Anterior length of the spinal cord along the AP axis, measured from the cord "
+            "center of mass to the anterior cord edge.\n"
+            "  - length_posterior: Posterior length of the spinal cord along the AP axis, measured from the cord "
+            "center of mass to the posterior cord edge.\n"
```
joshuacwnewton: Is it worth mentioning the 3mm extent here, like we do with the `diameter_AP`?

joshuacwnewton (dismissed):
I didn't have time to get to the QC report (I'll do a follow-up review ASAP), but I think the implementation of `length_anterior` and `length_posterior` make sense (especially since the numbers check out with my refactor).

spinalcordtoolbox/process_seg.py
valosekj: Excellent point!

> The CoM from compute_pca is based on the actual SC segmentation:

Correct.

> we're not working with the ellipse in this function, right?

Correct. We're not working with the ellipse directly here. But, hypotentically, if we fitted an ellipse (by `measure.regionprops`) as done for the RL diameter (`region.major_axis_length`):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d03cdc6a4eb01e9abfec0c30df6cd260275e166/spinalcordtoolbox/process_seg.py#L314

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d03cdc6a4eb01e9abfec0c30df6cd260275e166/spinalcordtoolbox/process_seg.py#L322

Wouldn't the RL diameter (`region.major_axis_length`) lie on the center of mass? I'll try to verify this.

valosekj: Okay, now I remember (and it is even documented in a comment 😅). We actually already use CoM based on `region.centroid`, i.e., obtained from the ellipse. But please double check me if I'm not wrong.

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/2d03cdc6a4eb01e9abfec0c30df6cd260275e166/spinalcordtoolbox/process_seg.py#L465

spinalcordtoolbox/process_seg.py
valosekj: > * We use the AP diameter measure as a base.
>   
>   * _(Specifically, the "improved" `diameter_AP` based on the mean diameters from the 3mm extent around the CoM of the HOG-rotated SC segmentation (i.e. PR ([#4958](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958)), **NOT** the ellipse diameter.)_
> * The variable `ap0_r` (ap component of CoM) is used to split `diameter_AP` in two.
> * Thus, the property should hold that `length_posterior + length_anterior = diameter_AP`?

Correct.

> And, you temporarily tested an alternate approach in [15bef9f](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/15bef9f45980cf562b9e9b43ceaee6cdd5100a31), then you reverted the changes in [f743737](https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/f743737ec87207cac6f098885858f9394a121cbc) after finding that the original method was better.
> * The alternate approach used `argmax` of the pixel sums.
> * This is equivalent to finding the actual widest horizontal line.
> * But, this is not robust because of... for example, a compressed cord where `argmax(sum())` would be offset from the CoM?

Exactly. Additionally, in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4958, we decided to "improve" only the `diameter_AP`. But keep the `diameter_RL` (based on ellipse fitting) as we found it robust.

spinalcordtoolbox/process_seg.py
valosekj: wording improved in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/changes/b2af1e06fb83fc24e2c696cf469e5a963d45d4a6

spinalcordtoolbox/scripts/sct_process_segmentation.py
valosekj: good point. done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/changes/7997edddce1674d4dee892a8bd188f03c5848d6f

spinalcordtoolbox/process_seg.py
valosekj: Excellent point! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/changes/1b013b4c40af4fe8f058ce4e0a23d10da78f8a11.

btw. when doing this, I noticed redundant variable assignment (`coord_ap = rl0_r`). So I fixed it in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/changes/5c44224cfff8117618c86ae296b6c428d3e889bb

spinalcordtoolbox/process_seg.py
joshuacwnewton: > We actually already use CoM based on region.centroid, i.e., obtained from the ellipse. But please double check me if I'm not wrong.

Double-checking: Looking at the source code, it seems like `region.centroid` in this case is just the [mean of the coordinates](https://github.com/scikit-image/scikit-image/blob/2c3a869aebcc82c7802a9ec3d112df0878df2d4d/src/skimage/measure/_regionprops.py#L478-L479).

In other words, it's distinct from the ellipse fitting, and so the centroid may not always be equivalent to the intersection of the major/minor axes of the second-order ellipse (I think?), which sounds like what you had in mind:

centroid

joshuacwnewton: I think that maybe this CoM vs ellipse behavior is relevant to the discussion in https://github.com/sct-pipeline/critical-ms-lesion/issues/25#issuecomment-4297286002? Since it might explain the difference between the two types of "center" points, especially if we thought we were doing it one way, but we're actually doing it another way.

mguaypaq: I'm 90% sure that the ellipse is centered on the centroid.

mguaypaq: Yes, from https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.regionprops (emphasis added):

> **axis_major_length**: float
> The length of the major axis of the ellipse that has the same normalized second **central moments** as the region.

> **moments_central**: (3, 3) ndarray
> Central moments (translation invariant) up to 3rd order:
> `mu_ij = sum{ array(row, col) * (row - row_c)^i * (col - col_c)^j }`
> where the sum is over the `row, col` coordinates of the region, and `row_c` and `col_c` are the coordinates of the region's **centroid**.

joshuacwnewton: Oh, oops. Of course, the words 'central moments' have meaning! 🤣 

Marking as resolved :)

valosekj: Thanks for double-checking this! 🙏

spinalcordtoolbox/reports/qc2.py
```diff
@@ -1197,8 +1244,102 @@ def add_angle_lines(ax, num_slices, metrics, angle_type, radius: tuple[int, int]
...
+        col = slice_index % num_col
+
+        # Reconstruct the actual cord centre within the mosaic, replicating the clamping done in mosaic().
+        # When the cord centre is within radius pixels of the image edge, mosaic() clamps it, so the cord
+        # is no longer at the tile's geometric centre.
+        cy, cx = centers[slice_index]
+        cy_c = max(radius[0], min(AP_dim - radius[0], int(cy)))
+        cx_c = max(radius[1], min(LR_dim - radius[1], int(cx)))
+        y_mosaic = row * (2 * radius[0]) + int(cy) + radius[0] - cy_c
+        x_mosaic = col * (2 * radius[1]) + int(cx) + radius[1] - cx_c
```
joshuacwnewton, line 1325: In DMs, @valosekj mentioned that the end of the `length_` QC lines looked to go outside the spinal cord:

Image

----

Part of this issue comes from the [`solid_capstyle`](https://matplotlib.org/stable/gallery/lines_bars_and_markers/capstyle.html) used when plotting the lines. Changing this from `'round'` to `'butt'` causes the issue to become more clear:

Image

----

At first, this looked like an off-by-one error. But, the "off-by-one" is not consistent in all of the mosaic tiles -- a simple fix like `+1` actually makes the problem worse in other areas.

Instead, I think the main issue here is that earlier when the background mosaic was first created, we used a truncated version of the spinal cord `center`  (`.astype(int)`) for the purposes of cropping out a `[30, 30]` square:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/53ae82a7b2040f25258d61473b52f126a19b014e/spinalcordtoolbox/reports/qc2.py#L1925-L1930

And so in the `add_ap_split_lines` function, we replicate that same behavior:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/53ae82a7b2040f25258d61473b52f126a19b014e/spinalcordtoolbox/reports/qc2.py#L1293-L1298

But I actually don't think we should replicating the `int` call here? Converting to `int` was only for the cropping -- but the "true" center still exists in the mosaic, and the anterior/posterior lengths are based off of this "true" center. 

So, I think we need to update the `x_mosaic`/`y_mosaic` determination to preserve the float-valued center. I have some ideas on how to do this?

joshuacwnewton: I rewrote the `x_mosaic` and `y_mosaic` code to carefully track all of the offsets:

<details>
<summary> Source code </summary>

```python
        # Calculate the location of the SC center within the mosaic.
        # Store the coordinate calculations in dicts to keep things tidy (repeat the same calcs for x/y)
        coords = {'x': {'shape': img_slice_shape[1], 'radius': radius[1], 'center': centers[i][1], 'mosaic_idx': col},
                  'y': {'shape': img_slice_shape[0], 'radius': radius[0], 'center': centers[i][0], 'mosaic_idx': row}}
        for axis, d in coords.items():
            # In the mosaic code, each tile is cropped from the image, then placed into the mosaic.
            # There are three types of "offsets" we have to compute:
            #   1. The shift between the image array center coordinate (e.g. [25, 32])
            #        to the center coordinate for a single mosaic tile (e.g. [14, 14])
            d['center_int']    = int(d['center'])  # integer coord, due to mosaic cropping
            d['center_tile']   = d['radius'] - 1   # 1-based length to 0-based index
            d['shift_1']       = d['center_tile'] - d['center_int']
            #   2. The shift added prior to the crop due to the center being too close to the array boundary.
            d['center_offset'] = min(d['shape'] - d['radius'], d['center_int'])  # Check far edge first
            d['center_offset'] = max(d['radius'], d['center_int'])               # Then check 0 edge last
            d['shift_2']       = d['center_offset'] - d['center_int']
            #   3. The shift added after the crop due the [row, column] positioning.
            d['shift_3'] = d['mosaic_idx'] * (2 * d['radius'])
            # Sum the shifts to get the total voxel offset from the array center coordinate to the center coordinate within the mosaic
            d['shift_total'] = d['shift_1'] + d['shift_2'] + d['shift_3']
            # NB: We preserve the original float-valued center coord, because it helps display the lengths accurately
            d['center_mosaic'] = d['center'] + d['shift_total']
```

</details>

But this is ultimately unnecessary, I think? The end result is the same as your old code, but without the truncation:

> Image

I think we can skip all of that and just take the decimal part of the centers (that got truncated by `int`) and add it back to your existing code. 😆 

Anyway, the end result is this, which looks good to me?

Image

joshuacwnewton: The one thing that I notice now, though, is that the original center orientation QC looks a bit off compared to the fixed ap_lengths QC:

Image

Maybe we also want to retroactively apply this fix to the previous QC, too?

joshuacwnewton (commented):
I think I may have identified the bug in the QC report. :)

spinalcordtoolbox/reports/qc2.py
joshuacwnewton: I'm also noticing that the angles don't seem to match up between both QCs. For example, in slice 50, the angle listed is 1.0*, but the lines appear quite angled.

By comparison, on slice 46, the angle listed is 1.2*, but the lines appear rather straight?

I am wondering if the listed angle is accurate... :thinking:

valosekj: > Anyway, the end result is this, which looks good to me?

Yes, great work! LGTM too!

valosekj: > The one thing that I notice now, though, is that the original center orientation QC looks a bit off compared to the fixed ap_lengths QC:
> 
> Image
> Maybe we also want to retroactively apply this fix to the previous QC, too?

I originally created this QC as a debug for the HOG angle. The new QC for the A and P lengths rotates the vertical line using the ellipse angle, though. So I think differences in the line rotation are expected. But the center should be likely the same between the QCs (needs to be verified :-D)

joshuacwnewton: Gotcha, that makes sense!

I see now that the `sct_process_segmentation` QC function has an `angle_type` argument that is `angle_HOG` by default.

I wonder if displaying the HOG angle debug plot might be confusing for users? Like, just now, I assumed that the HOG angle plot _was_ showing the ellipse angle. So, I wonder if maybe there's a way that we could just show the ellipse angle by default, and only show the HOG plot if `-v 2` is passed? Not sure, what do you think?

joshuacwnewton: > But the center should be likely the same between the QCs (needs to be verified :-D)

Fixed!

valosekj: > So, I wonder if maybe there's a way that we could just show the ellipse angle by default, 

By this, you mean the new anterior and posterior lengths QC?
I think that even this QC is mainly for debugging purposes. Not sure if we want to output it by default. Especially since we don't output QC for the other already existing morphometrics (CSA, AP diameter, RL diameter, ...).

> and only show the HOG plot if `-v 2` is passed? 

Agree, I'd keep the HOG plot when using  `-v 2`.

joshuacwnewton: > By this, you mean the new anterior and posterior lengths QC?

Apologies, I mean showing the angle corresponding to the ellipse that is fitted by `regionprops` in the red-line plot, rather than the hog angle. (In other words, the line would be the same as the cyan-yellow plot, just labeled with angle instead of length.) 

Alternatively, we just put that angle on the existing cyan-yellow plot? 🤔

valosekj: > Alternatively, we just put that angle on the existing cyan-yellow plot? 🤔

Yes, this could work. But the question is whether we actually need to show the ellipse angle?

(I implemented the HOG angle to debug the HOG method. Not sure if we need the ellipse angle. On the other hand, it'd be useful as a sanity check)

joshuacwnewton: The reason I was thinking about the ellipse angle was for debugging the orientation metric (i.e. just a sanity check to make sure the fitted ellipse looks OK), but I suppose the precise angle value is less necessary, since debugging can just be done visually, with the line relative to cord orientation.

In that case, I think the cyan/yellow plot is probably sufficient, and we can limit the HOG QC to `-v 2`.

joshuacwnewton: Done in [09f24ea](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5199/commits/09f24ea8f8d80e2abbc56a5c4dbb34b98ff14241), think this is good to go!

joshuacwnewton (approved):
LGTM! 🎉

PR #5200, opened 2026-04-08
Build containers from Docker recipe

AlexVCaron:
## Checklist

- [x] Only **after** the automated test suite passes: I will tag a [reviewer](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers) in the PR sidebar. (You can also ask for help with specific questions in the PR comments before the tests pass.)

## Description

This PR proposes a **unified container build workflow based on Docker**. Given a docker image, lets say `spinalcordtoolbox/sct:latest` on **Dockerhub**, an apptainer image is obtained by running a single command :

```bash
apptainer build spinalcordtoolbox_latest.sif docker://spinalcordtoolbox/sct:latest
```

**Documentation is incoming, and I'll deal with the older build recipes (or move the new one) once we've had some discussions on the matter on the PR**.

### Dockerfile design

The build is split in **3 phases** to limit the size of the final image produced (see [multi-stage builds](https://docs.docker.com/build/building/multi-stage/)) :

1. The `build` stage uses the `install_sct` script to deploy the library and scripts in **package mode**
2. The `package` stage uses `conda-pack` and `conda-unpack` to produce a **portable version of the environment**
3. The last stage (unnamed, will produce the final tag) copies the conda environment and sets the shell environment for autoloading

The advantages :

1. The final Docker image is leaner by `10Gb` (size is roughly `8Gb` compared to `18Gb` full build)
4. The initial build stage (called `build`) can be used as a **development environment for SCT**, aka a [devcontainer](https://containers.dev) or a [codespace](https://github.com/features/codespaces).
5. The `package` stage can be used to install an optimally lean (given `conda`) executable version of `SCT` in other containers

## Linked issues

Fixes #5186


joshuacwnewton (commented):
@AlexVCaron: Just a quick heads up, but our current Apptainer test workflow doesn't work on forks (yet!). So, you may find a failing test here -- that's not you.

If you have the capacity, it would be amazing if we could update the test in this PR to try and properly fetch your changes (and any changes from forks).

See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/24141745526/workflow?pr=5200#L516-L570

AlexVCaron:
@joshuacwnewton Thanks for the heads up ! I'll definitely look into that. I've crafted multiple github workflows for containers management, I can reuse much from those, it should be quick 🙂

AlexVCaron:
@joshuacwnewton **Don't launch the test before setting `DOCKER_USERNAME` and `DOCKER_PASSWORD` in secrets** 😉

joshuacwnewton:
I've updated the permissions on the SCT repo to allow CI runs automatically from non-new accounts (but still block CI runs from new accounts).

This should at least unblock you for this PR. Sorry about the hassle of needing manual approval each time you pushed to this PR!

.github/workflows/tests.yml
```diff
@@ -513,10 +511,37 @@ jobs:
...
+      - name: Extract version to Github env
+        run: |
+          export SCT_VERSION=$(cat spinalcordtoolbox/version.txt)
+          echo "SCT_VERSION=$SCT_VERSION" >> "$GITHUB_ENV"
+      # Instate this step once credentials are available
+      - uses: docker/login-action@v3
+        with:
+          # NB: these secrets need to be set in the repo's Settings > Secrets > Actions
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
joshuacwnewton: Highlighting for myself as a reference TODO. :)

joshuacwnewton: Added!

Image

AlexVCaron:
@joshuacwnewton I can confirm the new github workflow for containers is working locally via `act`. I'll update the branch once I finish updating the docs.

install_sct
```diff
@@ -649,15 +653,31 @@ fi
...
+MINIFORGE_VERSION="24.11.2-1"  # keep in sync with the URLs below
 case $OS in
 linux)
   download "$TMP_DIR/"miniforge.sh https://github.com/conda-forge/miniforge/releases/download/24.11.2-1/Miniforge3-Linux-x86_64.sh
+  MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-Linux-x86_64.sh"
+  MINIFORGE_CACHED="${MINIFORGE_CACHE_DIR:+${MINIFORGE_CACHE_DIR}/Miniforge3-${MINIFORGE_VERSION}-Linux-x86_64.sh}"
   ;;
 osx)
   download "$TMP_DIR/"miniforge.sh https://github.com/conda-forge/miniforge/releases/download/24.11.2-1/Miniforge3-MacOSX-x86_64.sh
+  MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-MacOSX-x86_64.sh"
```
joshuacwnewton: Nit, but could we deduplicate these URLs?

install_sct
```diff
@@ -649,15 +653,31 @@ fi
 
 # Download miniforge
 print info "Downloading Miniforge..."
+MINIFORGE_VERSION="24.11.2-1"  # keep in sync with the URLs below
 case $OS in
 linux)
   download "$TMP_DIR/"miniforge.sh https://github.com/conda-forge/miniforge/releases/download/24.11.2-1/Miniforge3-Linux-x86_64.sh
+  MINIFORGE_URL="https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-Linux-x86_64.sh"
```
joshuacwnewton: Ditto

install_sct
```diff
@@ -649,15 +653,31 @@ fi
...
+  MINIFORGE_CACHED="${MINIFORGE_CACHE_DIR:+${MINIFORGE_CACHE_DIR}/Miniforge3-${MINIFORGE_VERSION}-Linux-x86_64.sh}"
```
joshuacwnewton: If I understand correctly, `MINIFORGE_CACHE_DIR` is an undocumented environment variable that gets set by our Dockerfile. Is there any merit to documenting this somewhere in the script? (e.g. in the help, mentioning what it's used for, etc.)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,94 @@
...
+    "--init",
+    "--security-opt",
+    "label=disable,seccomp=unconfined,apparmor=unconfined",
+    "--cap-add",
+    "SYS_ADMIN",
+    "--cap-add",
+    "SYS_CHROOT"
+  ],
+
+  "privileged": true
```
joshuacwnewton: I'm a bit new to `devcontainer.json` -- can you explain the various options here and why they were needed? From a very cursory search they seem like they might grant a lot of privileges/system access here, so I'm curious how safe this is from a security perspective.

At the very least, maybe adding comments explaining why each setting is needed to grant access to X system resource might be helpful... 🤔

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,94 @@
...
+    "HISTFILE": "/commandhistory/.bash_history"
+  },
+
+  "mounts": [
+    "source=spinalcordtoolbox-bashhistory,target=/commandhistory,type=volume"
+  ],
+
+  "onCreateCommand": "sudo chmod ugo+rwx /commandhistory",
+  // Run post-creation command to install dependencies
+  "postCreateCommand": "pip install -e . --use-pep517 && pip install -e .[docs]",
```
joshuacwnewton, line 79: I'm curious about this here. If we run `install_sct` in the `Dockerfile`, do we need this here?

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+    apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        wget \
+        git \
+        bzip2 \
+        ca-certificates \
+        libdbus-1-3 \
+        libglib2.0-0 \
+        libxkbcommon-x11-0 \
+        libxrender1
```
joshuacwnewton, line 23: Is there any merit to clearing the apt cache/lists after installing dependencies as a way to try and reduce the final image size?

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+
+# Clone SCT from repository
+ADD https://github.com/spinalcordtoolbox/spinalcordtoolbox.git#${SCT_TOOLBOX_VERSION} /tmp/spinalcordtoolbox
```
joshuacwnewton: My IDE's Dockerfile analyzer really seems to dislike this line here. It seems to prefer an explicit clone + pointing to a directory.

Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+ENV PATH="/opt/conda/envs/venv_sct/bin:${PATH}"
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
+
+# Install listed models if any (comma separated list, call sct_deepseg <model> -install for each)
+RUN if [ -n "$SCT_DEEPSEG_MODELS" ]; then \
+        echo "Installing deepseg models: $SCT_DEEPSEG_MODELS"; \
+        printf '%s\n' "$SCT_DEEPSEG_MODELS" | tr ',' '\n' | while read -r MODEL; do \
+            echo "Installing model: $MODEL"; \
+            sct_deepseg "$MODEL" -install; \
+            # Sleep for 5 seconds to avoid potential rate limits
```
joshuacwnewton, line 89: Do comments need a line continuation character? IDE linter doesn't seem to like this one 😅

.gitignore
```diff
@@ -60,3 +60,6 @@ documentation/build
 /contrib/apptainer/*.sif
 /contrib/apptainer/*.tar.gz
 /contrib/apptainer/sct_apptainer_*/
+
+# Installation logs
+install_sct_log.txt
```
joshuacwnewton: Nit: Missing newline at end of file.

.github/workflows/tests.yml
```diff
@@ -512,59 +510,3 @@ jobs:
...
-        run: |
-          cd ./apptainer_build
-          ./install_sct_containered_$SAFE_REF.sh spine
-      - name: Install deepseg task post-install (with 'gm_sc_7t_t2star' task)
-        run: |
-          cd ./apptainer_build
-          ./install_deepseg_task.sh gm_sc_7t_t2star
-      - name: Download sample data
-        run: |
-          curl -sL "https://github.com/spinalcordtoolbox/sct_tutorial_data/archive/refs/tags/SCT-Course-20251208.tar.gz" | tar -z -xf -
```
joshuacwnewton, line 561: If we're replacing these steps, do you think we might want to remove the original Apptainer-specific files in the `contrib/` directory?

.github/workflows/test-containers.yml
```diff
@@ -0,0 +1,107 @@
...
+          echo "-- Dotnet"
+          sudo rm -rf /usr/share/dotnet
+          echo "-- Android"
+          sudo rm -rf /usr/local/lib/android
+          echo "-- Haskell"
+          sudo rm -rf /opt/ghc
+          echo "-- CodeQL"
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          df -h
+      - uses: eWaterCycle/setup-apptainer@v2
```
joshuacwnewton: This is certainly an interesting action source 😆 

Do we want to apply the same safeties here as we did in the previous Node.js PR?

.github/workflows/test-containers.yml
```diff
@@ -0,0 +1,107 @@
...
+
+# Ensure no write permissions are given to the workflow, for hardened security when running changes from forks
+permissions: {}
+
+jobs:
+  containers:
+    name: Docker (Ubuntu Latest)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
joshuacwnewton: Ditto for bumping actions versions.

.github/workflows/test-containers.yml
```diff
@@ -0,0 +1,107 @@
...
+          driver-opts: image=moby/buildkit:v0.28.1
+      - name: Build Docker image
+        uses: docker/build-push-action@v7.2.0
+        with:
+          context: .
+          push: false
+          load: true
+          tags: spinalcordtoolbox:${{ env.SCT_VERSION }}
+          cache-from: "type=registry,ref=neuropoly/sct:7.3"
+          build-args: "SCT_DEEPSEG_MODELS=spinalcord,spine,gm_sc_7t_t2star"
```
joshuacwnewton: Beautiful!

.github/workflows/test-containers.yml
```diff
@@ -0,0 +1,107 @@
...
+    # Even when given -y, apt will still sometimes hang at a prompt if a package
+    # has clarifications to ask; DEBIAN_FRONTEND=noninteractive prevents that,
+    # This will be defined for non-debian platforms below too, but there's no harm in that.
+    # (TravisCI quietly defined this on all their platforms, but we have to give it manually on GithubCI.)
+    DEBIAN_FRONTEND: 'noninteractive'
+    # Turns on color output for pytest. See: https://github.com/pytest-dev/pytest/issues/7443#issuecomment-656642591
+    PY_COLORS: "1"
+    # Disable progress bars for less verbose output
+    PIP_PROGRESS_BAR: "off"
+    SCT_PROGRESS_BAR: "off"
```
joshuacwnewton: I don't know if we need all of the exact preamble from the main `tests.yml` file? Much of this seems to not apply when we're not running the full suite of OSs + suite of tests.

.github/workflows/create-release.yml
```diff
@@ -62,6 +62,59 @@ jobs:
...
+          df -h
+      - uses: eWaterCycle/setup-apptainer@v2
+      - name: Build Apptainer image from Docker image
+        run: |
+          apptainer build sct_apptainer_${{ github.event.inputs.milestone_title }}.tar.gz docker-daemon:neuropoly/sct:${{ github.event.inputs.milestone_title }}
+      - name: Upload apptainer image
+        uses: actions/upload-artifact@v7
+        with:
+          name: apptainer-image
+          path: sct_apptainer_${{ github.event.inputs.milestone_title }}.tar.gz
```
joshuacwnewton: This may be overkill, but does GH allow us to de-duplicate these steps between the two workflows? I suppose some steps differ (token, whether to upload) but maybe that can be configured as an input to the reusable steps?

Specifically, maybe we should gate the `upload` behind the existing `dry-run: false` config that only is present on specific `workflow-dispatch` runs for this workflow.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,94 @@
...
+        "terminal.integrated.profiles.linux": {
+          "bash (login)": {
+            "path": "/bin/bash",
+            "args": ["-l"]
+          }
+        },
+        "terminal.integrated.defaultProfile.linux": "bash (login)"
+      }
+    }
+  },
```
joshuacwnewton, line 63: What do these lines do in practice? Like, will we have access to a full-featured copy of VS Code in the container? And, why might we want this?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,94 @@
...
+  "build": {
+    "dockerfile": "../Dockerfile"
+  },
+
+  // Features to install
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {},
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1.10.0": {},
+    "ghcr.io/miguelrodo/devcontainerfeatures/apptainer:1.1.4": {},
+    "ghcr.io/postfinance/devcontainer-features/make:1.0.0": {}
```
joshuacwnewton, line 14: Do we want to pin this `make` dependency in the lockfile, too?

(Also, is relying on sources like `miguelrodo` and `postfinance` a normal thing for `devcontainers`?)

joshuacwnewton (changes requested):
First draft review! Apologies, it's a bit large 😆

install_sct
AlexVCaron: My bad, turns out the download instruction was not supposed to be there, it's done with cache at [line 674](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5200/changes#diff-bdc55db8e2a4d5c7227e9fb4f3e0d1e9823e892a593cfe3e4c7faef738eaf6a2R674).

I'll also create a `MINIFORGE_DEPOT_BASE` for the common URL part between OSes.

install_sct
AlexVCaron: Very good idea ! I also suggest giving it a default. I propose `XDG_CACHE_HOME` for Linux and `~/Library/Caches` for MACOS.

.devcontainer/devcontainer.json
AlexVCaron: I'll add a comment above. This is the **smallest set of permissions required for `apptainer`**, required to be able to use [user namespaces without elevation](https://apptainer.org/docs/user/main/security.html#setuid-user-namespaces). The alternative is making the container **privileged**, which is far more dangerous.

.devcontainer/devcontainer.json
AlexVCaron: Since I use `conda-pack` and `conda-unpack`, I must install `sct` in **package mode** (non-editable). This line **rebinds the workspace (sct root) in the devcontainer in editable mode**.

Dockerfile
AlexVCaron: Yes, it greatly decreases the size of the image. It also shares the **sources lists** between `apt update` calls, so subsequent builds are faster. Same goes for `python` and `conda` calls.

Dockerfile
AlexVCaron: The [official Docker linter](https://marketplace.visualstudio.com/items?itemName=docker.docker) on VSCode doesn't raise issues ... the `ADD` instruction got extensive upgrades in the past year, maybe your IDE is still aligned with an older version of the Dockerfile spec ?

Dockerfile
AlexVCaron: I can move the comment up to the instruction header. It's perfectly valid `bash` though 😅

.github/workflows/tests.yml
AlexVCaron: I'm working on this in parallel with the documentation, as there is references to files in `contrib/apptainer` there. I'll edit/remove them if they prove unnecessary

.github/workflows/test-containers.yml
AlexVCaron: Very good point ! I took a note and forgot to apply the convention 😅 thanks for the heads up !

.github/workflows/create-release.yml
AlexVCaron: Yes for the `upload` gate on `dry-run` !

For the duplication, there are **2 courses of action**, either a [reusable workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) or a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository).

They're both quite similar, with minute differences on location and how they display. The former lists all steps taken, the latter only displays a single step). Both take on the ***level of privileges of the caller workflow**.

.devcontainer/devcontainer.json
AlexVCaron: No. The customization section is **caller dependent**. You can see it as an extension block for the different implementations of the [devcontainer specification](https://containers.dev/implementors/reference/) (CLI, IDE, etc.) . So this one here only take action when the devcontainer is built by **VSCode**.

Basically here, it sets default settings for python environment loading, linting and formatting, for the terminal to load correctly with `conda activation`, and for Jupyter notebooks to work.

.devcontainer/devcontainer.json
AlexVCaron: It's common for features to be pinned to creators while they've not received **stable releases**. They do get thorough vetting by the community though. I think **pinning like they are** is safe enough, akin to github actions from third-parties.

I'll check the `lockfile`, the mechanism is quite new, maybe it didn't get updated correctly on my last build.

.devcontainer/devcontainer.json
AlexVCaron: Just saw the `privileged` set to true. It should not be there, I'll update that

.github/workflows/create-release.yml
AlexVCaron: I chose the **composite action**. It's cleaner, lives in its own folder (`.github/actions`) and is potentially reusable from other repos that want to build `sct` (could potentially even be generalized to build containers defined in other repos, but out of scope for the PR).

install_sct
AlexVCaron: I decided against setting the default to prevent pollution local installation homes. The variable is well described in the `help` message.

.github/workflows/test-containers.yml
AlexVCaron: I only kept **pytest colors** and **disabling progress** for logs.

Dockerfile
joshuacwnewton: My apologies, I thought that the cache-clearing wasn't present and that it maybe should be added. But, does the `--mount==cache` part take care of this?

Dockerfile
joshuacwnewton: Let me double-check!

Dockerfile
AlexVCaron: Yes, it's the way to do it now. Before you had to manually clear the apt lists, now it's handled by [cache mounting](https://docs.docker.com/build/cache/optimize/#use-cache-mounts) !

AlexVCaron:
Alright @joshuacwnewton ! I addressed most of your review and update docs. Everything should be in order here :

- Every container build stems from the `Dockerfile`
   - `test-containers.yml` builds and tests both the `docker` and the `apptainer` containers
   - `create-release.yml` builds both containers, publishes the docker artifact to `neuropoly/sct:<release_version>`
   - All documentation aligns with this container build process, with clear tagging for **Added in 7.4**.
      - I kept the old installation sections and I prefixed them with a **before 7.4** meaningful header
- Both workflows above use a **composite action**, `actions/build-containers` to do the whole `docker`/`apptainer` poutine 🍟 (poutine emoji should exist 😢)

AlexVCaron:
There could be still some more improvements, but they are maybe minor enough to be done in further PRs. With the length of the work and the changes, maybe it's better to merge and give access to other developers to test and hone their skills against the new procedures. Here's what I see can be further improved :

- `Docker`/`Apptainer` documentation for different OSes could probably be merged in single pages. **Docker Desktop** has stabilized on `Windows`/`Mac` and aligned with the `engine`, so installation procedures are very similar (albeit completely identical) to `Linux`.
- The `apptainer` artifact could be saved to [Github Container Repository](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#troubleshooting) `ghcr.io`. From what I could find, there is no quota on image size (only layers up to `10GB` each), and it doesn't seem to count against any other storage quotas on the repository or organization yet.
   - I'm not sure though it will last. It looks like a pushback against recent DockerHub rate limiting. I've seen Github push limits on those kind of services previously, so I wouldn't be surprised.
   - Yet, there is no real alternative for distribution  of those kind of artifacts. `ghcr.io` is backed by a [generic Azure `oci` artifact repository](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-intro), meaning we can pull and push using the [`oras` protocol](https://apptainer.org/docs/user/main/cli/apptainer_push.html).

.devcontainer/scripts/fix-apptainer-mksquashfs.sh
```diff
@@ -4,30 +4,27 @@ set -euo pipefail
...
-# What this script does:
-# 1) Install squashfs-tools.
-# 2) Back up Apptainer's bundled helper binary once.
-# 3) Point Apptainer's helper path to the system mksquashfs binary.
-#
-# The operations are idempotent and safe to run multiple times.
+# - In mksquashfs 4.7.5, MALLOC_ARENA_MAX controlling the number of memory arenas is too small to build large images on large CPUs.
+# - The issue will be fixed by version 1.5.2 of Apptainer
 #
 # Issue opened on Apptainer's GitHub: https://github.com/apptainer/apptainer/issues/3577
```
joshuacwnewton: Bravo for reporting upstream and testing the fix!

Do we want to wait for 1.5.2. before merging this PR, just so we don't need this script?

documentation/source/user_section/installation/apptainer.rst
```diff
@@ -60,6 +96,37 @@ If installation ran to completion without error, a ``sct.sif`` file should now b
...
+.. versionadded:: 7.4
+
+Apptainer containers are static by default. In order to install new ``sct_deepseg`` tasks, you'll need to make the location where models are downloaded ``writable``.
+
+#. Use ``apptainer`` to create a writable overlay to store the models. This only needs to be done once, and the resulting ``sct_deepseg_models.img`` file can be reused for future installations.
+
+    .. code:: sh
+
+        SCT_ENV_PATH=/opt/conda/envs/venv_sct
+        MODEL_DIR=$SCT_ENV_PATH/lib/python3.10/site-packages/data/deepseg_models
```
joshuacwnewton: AFAIU models are downloaded to `$SCT_DIR/data/deepseg_models` rather than installed to the site-packages directory.

Or, is this a consequence of it being installed in non-editable mode? I am now unsure if I fully understand all of the consequences of non-editable mode, since a lot of assumptions are made by SCT regarding paths relative to `$SCT_DIR`.

documentation/source/user_section/installation/apptainer.rst
```diff
@@ -20,9 +20,45 @@ Using Apptainer introduces a few caveats to using SCT, however:
...
 ************
 
-This method will install SCT within an Apptainer container, ready for portable use.
+This method will install SCT within an Apptainer container, ready for portable use. First, `install Apptainer <https://apptainer.org/docs/admin/main/installation.html#install-unprivileged-from-pre-built-binaries>`__ if you have not done so already (or activate the module which contains it, if on an shared resource system).
 
-#. Install Apptainer if you have not done so already (or activate the module which contains it, if on an shared resource system).
+From an official release
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 7.4
```
joshuacwnewton, line 28: This is really neat! Somehow I missed the use of this Sphinx tool.

documentation/source/user_section/installation/linux.rst
```diff
@@ -192,7 +246,7 @@ Then, to forward the X11 server:
    
     .. code:: bash
 
-        sudo docker run -it --rm --privileged -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix <IMAGE_NAME>/ubuntu:ubuntu22.04``
+        sudo docker run -it --rm --privileged -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix <IMAGE_NAME>/<IMAGE_SPEC>``
```
joshuacwnewton, line 252: Is there a reason we update these steps to `<IMAGE_SPEC>` here but not elsewhere?

documentation/source/user_section/installation/mac.rst
```diff
@@ -160,8 +213,8 @@ First, launch Docker Desktop, then open up a new Terminal window and run the com
    apt update
    apt install git bzip2 curl gcc git libdbus-1-3 libgl1-mesa-glx libglib2.0-0 libxkbcommon-x11-0 libxrender1
    # Note for above: libdbus-1-3, libgl1-mesa-glx, libglib2.0-0, libxkbcommon-x11-0, libxrender1 are required by PyQt
-   # Install SCT (you can change 7.0 for the version of your choice)
-   git clone --branch 7.0 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git sct
+   # Install SCT (you can change 7.3 for the version of your choice)
+   git clone --branch 7.3 https://github.com/spinalcordtoolbox/spinalcordtoolbox.git sct
```
joshuacwnewton, line 228: Just pointing out the update to 7.3 here but not in the other install docs (really just a nit, but might be good to be consistent for polish reasons).

joshuacwnewton: This might be a mark against triplicating the instructions, haha 😅

documentation/source/user_section/installation/linux.rst
joshuacwnewton, line 1: I wonder if we can use any Sphinx tricks to define some page-level or global constants (e.g. 7.4, 7.3, ubuntu22.04) that get dynamically inserted when building the documentation.

Related:

- https://github.com/sphinx-doc/sphinx/issues/4054
- https://stackoverflow.com/a/56328457

joshuacwnewton: Maybe overkill, but I could see easily missing one instance here or there since it's spread out quite a bit!

.github/workflows/create-release.yml
```diff
@@ -226,8 +243,7 @@ jobs:
         bodyFile: ".github/workflows/release-body.md"
         artifacts: "install_sct-${{ github.event.inputs.milestone_title }}_win.bat,\
                     install_sct-${{ github.event.inputs.milestone_title }}_linux.sh,\
-                    install_sct-${{ github.event.inputs.milestone_title }}_macos.sh,\
-                    sct_apptainer_${{ github.event.inputs.milestone_title }}.tar.gz"
```
joshuacwnewton, line 230: If I understand correctly, we remove the `.tar.gz` because the workflow steps have changed from (uploading the .sif files directly) to (upload the docker image to docker hub)?

documentation/source/user_section/installation/apptainer.rst
```diff
@@ -20,9 +20,45 @@ Using Apptainer introduces a few caveats to using SCT, however:
...
+
+#. Select the version of SCT you want to install, either from the `release page <https://github.com/spinalcordtoolbox/spinalcordtoolbox/releases>`__ or from the `DockerHub repository <https://hub.docker.com/r/neuropoly/sct/tags>`__.
+
+#. Open a new Terminal window and run the following commands, replacing ``{sct_version}`` with the version you want to install (e.g. ``7.4``):
+
+    .. code:: sh
+
+        apptainer build sct.sif docker://neuropoly/sct:{sct_version}
```
joshuacwnewton, line 36: Beautiful. ♥️

.github/workflows/tests.yml
```diff
@@ -3,8 +3,6 @@ name: Tests
 on:
   workflow_dispatch:
   pull_request:
-    branches:
-      - '*'
```
joshuacwnewton, line 7: Trying to jog my memory here -- why remove this? (Is it redundant with the base `pull_request:`?)

.github/actions/build-containers/action.yml
joshuacwnewton, line 1: This LGTM, but just as a nit, the file reads as a bit "wall of text" to me. Would you be able to add some basic comment blocks + a couple of newlines to delineate between the metadata (inputs, outputs) and the actual composite steps? That way it's easier to jump to the relevant section when skimming the file.

Might be useful to link to https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-an-action-metadata-file in a comment too just because this is SCT's first composite action. :)

Dockerfile
joshuacwnewton: No worries, sorry!

Dockerfile
joshuacwnewton: You're right that my linter was out of date.

Perhaps we can specify the syntax version in the Dockerfile to avoid potential unintuitive errors for people using outdated versions of Docker?

(https://hub.docker.com/r/docker/dockerfile#linked-copies-copy---link-add---link)

joshuacwnewton (changes requested):
This is one last big review (sorry in advance!) but I feel confident that after this review it will basically be good to go! ♥️

.devcontainer/scripts/fix-apptainer-mksquashfs.sh
AlexVCaron: `apptainer` doesn't maintain a release calendar, and releases are not on a schedule, so `1.5.2` could drop in a month or more. I believe we could merge. I'm following the updates on the `apptainer` side (github+slack) and can upgrade once it's out.

.github/workflows/tests.yml
AlexVCaron: Dropped out when I reverted the `pull_request_target`. I'll put it back. It could also be more specific, aka if you want the workflow to only run on PR to `master` instead of all branches (except ones with `/` in them)

Dockerfile
AlexVCaron: Yes, there is a `syntax` comment I can put at top of the file to control both the linter and the build.

documentation/source/user_section/installation/mac.rst
AlexVCaron: Yeah, it's a bit of a pain ... but it makes it simpler to follow the installation procedure, and usage of containers more discoverable. A necessary evil I'd say 😅

.github/workflows/create-release.yml
AlexVCaron: From my understanding, the `.tar.gz` contained the scripts in `contrib` to use to build the apptainer images, not the `sif`. Those were light enough to not take too much space in the release. The container is over `2Gb`, so it won't fit there.

We could however publish it to `packages`, aka [ghcr.io](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)

documentation/source/user_section/installation/apptainer.rst
AlexVCaron: It is a consequence of non-editable mode, yes. Then the `__sct_dir__` variable defined in `spinalcordtoolbox.utils.sys` points at the installation inside site-packages.

documentation/source/user_section/installation/linux.rst
AlexVCaron: There seems to be a [sphinx-extension](https://github.com/adamtheturtle/sphinx-substitution-extensions) to handle this. It's maintained and has some usage. I can test it see how it fares !

documentation/source/user_section/installation/apptainer.rst
joshuacwnewton: I am wondering then how `install_sct` deals with `$SCT_DIR` in this case then, since I think the env variable is set to be relative to the root dir and directories like `bin` and `data` are defined relative to that directory.

What was the reason of using non-editable mode again? Just avoiding mutable installations? I am a bit concerned about potential breakage to `install_sct` outside of the context of the docker/apptainer installs... :thinking:

AlexVCaron: It's necessary to get the conda environment out of the build stage. That's how I achieve the `~10Gb` reduction in image size.

If you look at the changes I made in `install_sct`, I simply force the `package` installation mode already present there. I didn't edit any behaviors aside of it.

AlexVCaron: I validated the `package` (non-editable) behavior :
- Binaries and libraries are downloaded in the conda environment the same way as in editable mode
- As long as the `-d` flag is not enabled, data are downloaded in **site-packages** instead of in `SCT_DIR` (because of the value of `__sct_dir__` inside the python package)

As is, running `install_sct` with the `package` flag doesn't invalidate the `SCT_DIR` environment variable, so there is no problems installing locally this way, **as long as you enable re-installing the data**.

joshuacwnewton: > If you look at the changes I made in `install_sct`, I simply force the `package` installation mode already present there. I didn't edit any behaviors aside of it.

My apologies, I might be misunderstanding, but wasn't non-editable mode only added in this PR? (https://github.com/AlexVCaron/spinalcordtoolbox/commit/fafa519d25efa4fbc73a396610d087714ee53551)

AFAIU regardless of package vs. in-place, before this PR editable was always used (and thus none of our data dependencies ever got installed to `site-packages` prior to this PR).

> As long as the `-d` flag is not enabled, data are downloaded in **site-packages** instead of in `SCT_DIR` (because of the value of `__sct_dir__` inside the python package)

Ahhh I see. So, for non-editable installs, `$SCT_DIR` =/= `__sct_dir__`. That makes sense, but I guess what I'm worried about is any code that _does_ assume that `$SCT_DIR` == `__sct_dir__`. 

Some examples off the top of my head:

1. In-code assumptions, e.g. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/612bc73787c9a71d7659b951e0fb61efbee4978d/spinalcordtoolbox/compat/launcher.py#L40-L46

2. External assumptions re: data being installed to `$SCT_DIR/data`. For example, some bash scripts make this assumption:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b3df012e69294318c1621c1bee270031c0744db0/batch_processing.sh#L63

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b3df012e69294318c1621c1bee270031c0744db0/batch_processing.sh#L118

Hence my apprehension in breaking this assumption? I just think there might be side-effects that haven't been fully explored.

AlexVCaron: > My apologies, I might be misunderstanding, but wasn't non-editable mode only added in this PR? ([AlexVCaron@fafa519](https://github.com/AlexVCaron/spinalcordtoolbox/commit/fafa519d25efa4fbc73a396610d087714ee53551))

True ! My bad, I forgot this one 😮 

> Ahhh I see. So, for non-editable installs, `$SCT_DIR` =/= `__sct_dir__`. That makes sense, but I guess what I'm worried about is any code that _does_ assume that `$SCT_DIR` == `__sct_dir__`.
> 
> ...
> 
> Hence my apprehension in breaking this assumption? I just think there might be side-effects that haven't been fully explored.

And I made the assumption `__sct_dir__` was used uniformly throughout the codebase 😅

Let me think about this one a bit more then. I don't want to extensively change the codebase, as it also implies changing user behaviors. I believe I can change the docker image build behavior slightly instead and retain much of the saved space.

AlexVCaron: I reframed the docker build around editable installation. Sadly, we gained `~3Gb` more in image size 🥲 Still around `7Gb` less than a single stage build !

.github/actions/build-containers/action.yml
AlexVCaron: It should be better now 🙂

.github/workflows/create-release.yml
AlexVCaron: For publishing to `ghcr`, I'll need an additional [token with **write permissions** to `packages`](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#about-scopes-and-permissions-for-package-registries). I'd go with the name `GHCR_RW_PAT`.

documentation/source/user_section/installation/linux.rst
AlexVCaron: I did so to have the example work with both installation for 7.4 (`sct:7.4` or `deepseg:spinalcord-tumor_t2` above) or and for prior versions (`ubuntu:ubuntu22.04`). There's maybe a better way though

documentation/source/conf.py
```diff
@@ -43,6 +45,7 @@
     'sphinx.ext.extlinks',
     'sphinx_copybutton',
     'sphinxarg.ext',
+    'sphinx_substitution_extensions',
```
joshuacwnewton, line 48: Looks like we might just need to install the extension in the RTD build to address the failing CI:

```
 ModuleNotFoundError: No module named 'sphinx_substitution_extensions' 
```

RTD points to the `extras_require` in our `setup.py`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/25ed118a1024f2cbc46520d2e08e6aa99ec25dcf/.readthedocs.yaml#L14-L20

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/25ed118a1024f2cbc46520d2e08e6aa99ec25dcf/setup.py#L45-L55

I'll give it a closer look once RTD spits out a rendered doc. :)

joshuacwnewton (approved):
I think I am at the point where, as you say previously, it is good to merge this PR to get it into the hands of students for further testing, and fix any remaining issues in follow-up PRs. (I didn't see anything that stood out to me on a first pass, but I'm sure we'll uncover things once we actually get to testing it in more depth.)

Do you think we might want to manually complete the GHRC steps once just to be able to get a starter image up (prior to running the full release creation steps)?

AlexVCaron:
> I think I am at the point where [...] it is good to merge this PR to get it into the hands of students for further testing ... 

Yes, agreed.

> Do you think we might want to manually complete the GHRC steps once just to be able to get a starter image up (prior to running the full release creation steps)?

I tested it with `act` on my personal account and fork. It's a good idea to `dispatch` a run with a dummy `milestone_title` to validate that **secrets** and **inputs** are correctly assigned.

joshuacwnewton (approved):
I'm so sorry for my delay in merging this. I think there might have been a miscommunication on my part -- I was under the impression that the `neuropoly/sct` Docker Hub repo hadn't been tested yet.

I see now that there is a [`demo`](https://hub.docker.com/layers/neuropoly/sct/demo/images/sha256-5fb9ed4ce3be2d1bb7d59294c104926a22731fac249f363d61f7fd075a8d20ec) tag that you pushed some time ago. That eases my concerns, so I'm merging now.

Again, sorry for the delay here. 😓

PR #5202, opened 2026-04-15
Use the full SC width as the bridge width when lesion only has partial overlap with midsagittal slices

### LINKED ISSUE #5184: Minimum interpolated bridge width value underestimated when lesion is only present in 1 of the 2 slices

joshuacwnewton:
## Checklist

## Description

> [!WARNING]
> This PR is an alternative to PR https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5179. It arrives at a similar solution, but developed completely independently by accident.
> 
> - Jan's work: Designed to solve https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5123
> - My work: Designed to solve https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5184.
>
> Both of the underlying issues actually have similar root causes but were uncovered in separate investigations.

> [!TIP]
> Early in `sct_analyze_lesion`, we identify a **midsagittal slice** (float valued). To estimate the bridge widths at this midsagittal slice, we first take the 2 closest sagittal slices (integer values) using `floor` and `ceil` (called **interpolation slices** in-code). Then, for each axial slice in the lesion, we compute the bridge width by interpolating between he bridge widths found at each of those 2 sagittal slices.
>
> There are multiple different "missing lesion" cases to handle here:
>
> 1. There are no detected lesions at all.
> 2. There is a lesion, but it overlaps with 0 of the 2 interpolation slices.
> 3. There is a lesion, but it overlaps with 1 of the 2 interpolation slices.
>
> On `master`, all three cases are hanlded poorly. My PR tries to fix cases 2 and 3. @valosekj's PR fixes cases 1 and 2.

Functionally, this PR addresses the following faulty lines of code on `master`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/621c69f21374bcc4708061179ae2e534402900ee/testing/cli/test_cli_sct_analyze_lesion.py#L116-L118

`0.0` should _not_ be used as a fallback value, because when it gets averaged (as part of the interpolation process), it will skew the average towards a very small bridge width value. Then, when we take the `min` of all of the bridge widths, the result will be faulty.

The fix here is to instead use the spinal cord width. The idea being that, when the lesion is missing from a slice, using the full `sc_width` is the appropriate value in cases such as these:

image

By using the full width, the interpolation value is no longer erroneously small, and thus taking the `min` of all of the bridge widths leads to the expected result.

## Linked issues

Related to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5184.

### DIFF
--- a/testing/cli/test_cli_sct_analyze_lesion.py
+++ b/testing/cli/test_cli_sct_analyze_lesion.py
@@ -131,7 +131,10 @@ def compute_expected_measurements(lesion_params, path_seg=None):
                 # If 0/2 interp slices have lesion: find the interpolated minimum SC AP diameter, then take 1/2 of that for dorsal/ventral
                 if tissue_bridge_ceil is None and tissue_bridge_floor is None:
                     sc_width_interp = (decimal * sc_min_ap_diameters[z_ceil] + (1 - decimal) * sc_min_ap_diameters[z_floor])
-                    tissue_bridge_interp = sc_width_interp if 'total' in key else sc_width_interp/2
+                    # FIXME: This is proof that my solution overlaps Jan's solution
+                    #   - Jan uses /2 the sc_width (total sums to SC width, individual bridges are too small)
+                    #   - I use the full sc_width (total sums to 2x SC width, individual bridges are the right size)
+                    tissue_bridge_interp = sc_width_interp*2 if 'total' in key else sc_width_interp
                 else:
                     # If 1/2 interp slices have lesion: use 0.0 as a placeholder value (FIXME: #5184/#5202)
                     if tissue_bridge_ceil is None:
--- a/testing/cli/test_cli_sct_analyze_lesion.py
+++ b/testing/cli/test_cli_sct_analyze_lesion.py
@@ -263,7 +263,7 @@ def test_sct_analyze_lesion_matches_expected_dummy_lesion_measurements(dummy_les
                 # The values will be adjusted according to the cos of the angle
                 # between the spinal cord centerline and the S-I axis, as per:
                 # https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3681#discussion_r804822552
-                if key == 'max_equivalent_diameter [mm]' or 'bridge' in key:
+                if key == 'max_equivalent_diameter [mm]':
                     # The values are the same, but one value has slightly more precision than the other, so it is
                     # greater than expected --> rounding both values to the same number of decimal points before
                     # comparing
--- a/testing/cli/test_cli_sct_analyze_lesion.py
+++ b/testing/cli/test_cli_sct_analyze_lesion.py
@@ -136,11 +136,14 @@ def compute_expected_measurements(lesion_params, path_seg=None):
                     #   - I use the full sc_width (total sums to 2x SC width, individual bridges are the right size)
                     tissue_bridge_interp = sc_width_interp*2 if 'total' in key else sc_width_interp
                 else:
-                    # If 1/2 interp slices have lesion: use 0.0 as a placeholder value (FIXME: #5184/#5202)
+                    # If 1/2 interp slices have lesion: use sc_width (AP diameter)
+                    # FIXME: We just grab the minimum diameters since those were readily available, but the actual API code
+                    #        uses a per-axial-slice AP diameter, meaning that the expected vs. actual may differ.
+                    #        Fixing this would require refactoring this function to capture SC widths at each axial slice, which is possible but more work.
                     if tissue_bridge_ceil is None:
-                        tissue_bridge_ceil = 0.0
+                        tissue_bridge_ceil = 2*sc_min_ap_diameters[z_ceil] if 'total' in key else sc_min_ap_diameters[z_ceil]
                     if tissue_bridge_floor is None:
-                        tissue_bridge_floor = 0.0
+                        tissue_bridge_floor = 2*sc_min_ap_diameters[z_floor] if 'total' in key else sc_min_ap_diameters[z_floor]
                     # If 2/2 interp slices have lesion: interpolate like normal
                     tissue_bridge_interp = (decimal * tissue_bridge_ceil + (1 - decimal) * tissue_bridge_floor)
 
--- a/testing/cli/test_cli_sct_analyze_lesion.py
+++ b/testing/cli/test_cli_sct_analyze_lesion.py
@@ -110,12 +110,38 @@ def compute_expected_measurements(lesion_params, path_seg=None):
                 tissue_bridges[f"slice_{z}_ventral_bridge_width [mm]"] = min(ventral_bridge_widths)
                 tissue_bridges[f"slice_{z}_total_bridge_width [mm]"] = (min(dorsal_bridge_widths) +
                                                                         min(ventral_bridge_widths))
-            # Estimate the interpolated bridge widths at the mid-sagittal slice
+
+            # Find the minimum spinal cord AP diameter across each all SI slices for each of the interpolation slices
             decimal = mid_sagittal_slice - math.floor(mid_sagittal_slice)
             z_ceil, z_floor = int(np.ceil(mid_sagittal_slice)), int(np.floor(mid_sagittal_slice))
+            sc_min_ap_diameters = {}
+            for z in [z_ceil, z_floor]:
+                ap_diameters = []
+                for y in range(starting_coord[1], starting_coord[1] + dim[1]):
+                    ap_diameters.append(np.sum(Image(path_seg).data[:, y, z]))
+                sc_min_ap_diameters[z] = min(ap_diameters)
+
+            # Estimate the interpolated bridge widths at the mid-sagittal slice
             for key in ['dorsal_bridge_width', 'ventral_bridge_width', 'total_bridge_width']:
-                tissue_bridges[f'interpolated_{key} [mm]'] = (decimal * tissue_bridges.get(f'slice_{z_ceil}_{key} [mm]', 0.0) +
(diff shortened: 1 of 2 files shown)


testing/cli/test_cli_sct_analyze_lesion.py
```diff
@@ -110,12 +110,44 @@ def compute_expected_measurements(lesion_params, path_seg=None):
...
+                tissue_bridge_ceil = tissue_bridges.get(f'slice_{z_ceil}_{key} [mm]', None)
+                tissue_bridge_floor = tissue_bridges.get(f'slice_{z_floor}_{key} [mm]', None)
+
+                # If 0/2 interp slices have lesion: find the interpolated minimum SC AP diameter, then take 1/2 of that for dorsal/ventral
+                if tissue_bridge_ceil is None and tissue_bridge_floor is None:
+                    sc_width_interp = (decimal * sc_min_ap_diameters[z_ceil] + (1 - decimal) * sc_min_ap_diameters[z_floor])
+                    # FIXME: This is proof that my solution overlaps Jan's solution
+                    #   - Jan uses /2 the sc_width (total sums to SC width, individual bridges are too small)
+                    #   - I use the full sc_width (total sums to 2x SC width, individual bridges are the right size)
+                    tissue_bridge_interp = sc_width_interp*2 if 'total' in key else sc_width_interp
```
valosekj, line 137: So if I understand it correctly, you do:
- ventral bridge =  sc_width
- dorsal bridge = sc_width

Which, when summing ventral bridge + dorsal bridge into total bridge width, results in 2x sc_width. Please correct me if I'm wrong.

I think this is unintuitive, as total bridge width is supposed to be equal to sc_width.

joshuacwnewton: Yes, you are exactly correct. I agree that this is unintuitive, and that the `total` value is absolutely wrong. 

I will try to explain my reasoning for preferring `sc_width` instead of `1/2 * sc_width` using some figures.

joshuacwnewton: Using the example from https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5184#issuecomment-4012034760 (thick sagittal slices)... Say, for example, we are trying to interpolate the bridge width between two sagittal slices (`A` and `B`) at the specified axial slice: 

> comparison(1)

----

When measuring the non-lesion widths for the foreground and background slices, we get:

> comparison

The question is: What `A` value should be used when interpolating for both the ventral and dorsal widths?

----

Your PR uses `1/2 * sc_width`:

> image

Observation: Because the sample lesion is so far to the anterior side of the cord, all of the dorsal widths will be in the range of, roughly, `[15, 18]`. 

When we use a placeholder width of `11` that is smaller than the actual range, we might end up with an interpolated value of, say, `13`. Then when we take the `min()` of all of the interpolated bridge widths, we risk returning a faulty minimum value (`13 << 15`), repeating the bug from #5184.

----

> image

By comparison, if we use the full `sc_width` (`22`), we are of course greatly overestimating both the dorsal and ventral bridge widths (`22 >> 15`, `22 >> 3`), and the interpolation result will be large. 

But, I figure that overestimating is safer than underestimating:

- If we overestimate a missing width, then we will simply end up with another value as the `min()` value (and presumably that value will be drawn from two valid bridge width measurements).
- If we underestimate a missing width, then we might get an erroneous `min()` value entirely ([#5184](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5184)).

Hence why I use the full `sc_width` in this PR for both the dorsal and ventral interpolation calculations: I am really just trying to avoid the underestimation bug at all costs.

joshuacwnewton: Thank you to Dario for sharing some ideas in an email reply. I am just responding here for posterity (and because it gave me an excuse to make a new figure!)

> Could we, alternatively, not use interpolated metrics in cases where tissue bridges fall on an S-I coordinate that only captures the lesion on one sagittal slice but not the other, and instead use the measures from the slice with the lesion (B in the illustrative example)?

Applying this idea to the figure above (with improved coloring for A and B), the I think the widths would look something like this:

> bridge_widths_both

Where red lines are bridge widths averaged between both slices, and blue/green lines are bridge widths using only 1 of the 2 slices. (This was done in mspaint, so approximations only here!)

In this case, I do feel like it would lead to somewhat sensible results? At least, more sensible than the two extremes of (full sc_width) and (half sc_width). 

My only concerns are:

- For the dorsal widths: The `min` isn't affected by adding the single-slice widths.
- For the ventral widths: The `min` bridge width value of the red lines (`~0.5vox`) is different vs. the `min` value of the single-slice bridge widths (`0vox`). So, with the new method, we get a (very slightly~) smaller interpolated bridge width than we would if we limited ourselves to just the slices where both are present.
- Could there be a more extreme case of the above, where the single-slice bridge widths are significantly smaller than the two-slice interpolated value?

valosekj: > In this case, I do feel like it would lead to somewhat sensible results? At least, more sensible than the two extremes of (full sc_width) and (half sc_width). 

Agree.

> My only concern is 0 being present in each of the two sagittal slices separately, but the interpolated slice not containing 0, but maybe that's okay?

Good point! How about adding a condition to use 0 for the interpolated slice in this case?

> I am wondering if there's some contrived case where the lesion on one slice is markedly different than the lesion at the interpolation slice... hm.

I think lesions should have (at least) some "continuity" across adjacent slices.

joshuacwnewton:
I think because we've done the work to compare the two approaches and settle on a solution, I can close this PR and concentrate all efforts in #5179.

If we need to return to the changes from this PR, we at least have a page that showcases all of my changes in isolation. :)

PR #5205, opened 2026-04-16
Add option to use spinal levels in `sct_process_segmentation`

sandrinebedard:
## Checklist

## Description

This PR adds the option to use spinal levels (from the PAM50) as a reference. 

It adds the flag `-spinal-level`. It will also display the column "SpinalLevel" instead of "VertLevel" in the output .csv.

It aggregates the same way as `-vertfile` does, except when the flag -normlaize-PAM50 is passed, the `PAM50_spinal_level.nii.gz` is selected to do the normalization instead of `PAM50_level.nii.gz`

### Relevance of the flag?
As it processed the same way as `-vertfile`, can't users directly pass the spinal_level file there instead?

The rational to have a specific argument for `-spinal-level`
* So it is explicit in the output csv what anatomical reference was used
* To use the correct PAM50 ref file for normalization when using `-normalize-PAM50`


sandrinebedard:
Hi @joshuacwnewton, for this PR, I only want the option to use spinal levels as an input instead of vertfile (the file format, processing is the same as for `-vertfile`). In that case, I wasn't sure what was the best approach in the code to re-use the same things then for vert, but with a new columns of SpinalLevel instead. I would appreciate your input! Thanks!

joshuacwnewton:
Thank you for the ping! I'll give it a look (understanding that the current PR is just tentative). :)

joshuacwnewton:
Oops, I clicked on that accidentally, I was trying to request my own review. 😓

sandrinebedard:
> Thank you for the ping! I'll give it a look (understanding that the current PR is just tentative). :)

Here is an idea that I had  in this commit: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5205/commits/43e6b5cc60c35bab581be1885f473677a422600b

let me know what you think!

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -445,9 +452,19 @@ def main(argv: Sequence[str]):
...
+    spinal_level = False
     # make sure we have a valid VertLevel file (used for aggregation + VertLevel column)
     if arguments.vertfile is not None and arguments.discfile is not None:
         parser.error("Both '-vertfile' and '-discfile' were specified. Please only specify one of these options.")
+    if arguments.vertfile is not None and arguments.spinal_level is not None:
+        parser.error("Both '-vertfile' and '-spinal_level' were specified. Please only specify one of these options.")
+    if arguments.discfile is not None and arguments.spinal_level is not None:
+        parser.error("Both '-discfile' and '-spinal_level' were specified. Please only specify one of these options.")
+    if arguments.discfile is not None and arguments.spinal_level is not None and arguments.vertfile is not None:
+        parser.error("'-discfile', '-vertfile' and '-spinal_level' were specified. Please only specify one of these options.")
```
joshuacwnewton: I think we could simplify this by doing something like:

```python
    # make sure we have a valid VertLevel file (used for aggregation + VertLevel column)
    vertlevel_args = ["vertfile", "discfile", "spinal-level"]
    vertlevel_args_user = [arg for arg in vertlevel_args
                           if getattr(arguments, arg.replace("-", "_")) is not None]
    if len(vertlevel_args_user) > 1:
        parser.error(f"Multiple vertlevel file arguments specified (-{', -'.join(vertlevel_args_user)}). "
                     f"Please choose only 1 of the following {len(vertlevel_args)} arguments: -{', -'.join(vertlevel_args)}.")
```

Example output:

Image

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -561,7 +577,10 @@ def main(argv: Sequence[str]):
                                          verbose=verbose,
                                          remove_temp_files=arguments.r)
     if normalize_pam50:
-        fname_vert_level_PAM50 = os.path.join(__data_dir__, 'PAM50', 'template', 'PAM50_levels.nii.gz')
+        if arguments.spinal_level is not None:
+            fname_vert_level_PAM50 = os.path.join(__data_dir__, 'PAM50', 'template', 'PAM50_spinal_levels.nii.gz')  # consider creating a sepearet spinal level from the rootlets
```
joshuacwnewton: > ```python
> # consider creating a sepearet spinal level from the rootlets
> ```

What does this comment mean? Like, generating a new PAM50 file where the spinal levels are rootlets-based? 

Should we maybe move this comment into an issue / project idea?

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -367,6 +367,10 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], distan
             agg_metric[slicegroup]['VertLevel'] = vertgroups[slicegroups.index(slicegroup)]
             if agg_metric[slicegroup]['VertLevel'][0] is None:
                 agg_metric[slicegroup]['VertLevel'] = None
+        if spinal_level:
+            agg_metric[slicegroup]['SpinalLevel'] = agg_metric[slicegroup]['VertLevel']
+        else:
+            agg_metric[slicegroup]['SpinalLevel'] = None
```
joshuacwnewton: I am a bit confused about the changes to `aggregate_slicewise.py`. 

If all of the code is basically the exact same (aggregating based on `fname_vert_level`), and the only difference is the column name, then isn't it misleading to have "SpinalLevel" == "VertLevel"? (It would be two columns with the same values, right?)

Maybe we could do something like this instead?

```python
level_column_name = "SpinalLevel" if spinal_level else "VertLevel"
```

Then do a Find/Replace for `"VertLevel"` -> `level_column_name`

Or, do we need to always have "VertLevel" no matter what?

joshuacwnewton: Actually, wait a moment. Couldn't we just do something like this instead?

1. Pass the spinal level file to the `-vertfile` argument.
2. Change the name of the column in the output CSV file (VertLevel -> SpinalLevel).

```bash
# "1s" restricts to only the first line, i.e. the header
sed -i '1s/VertLevel/SpinalLevel/' output.csv
```

Since the logic is the exact same, you could reuse the existing argument, right?

Maybe the fix is as simple as adding a note to `-vertfile` saying "You can pass any labeled-segmentation level file, regardless of whether the file represents vertebral or spinal levels, and the aggregation process will be the same."

joshuacwnewton (commented):
Hmm. I would love to understand the motivation a bit better. 

For example, for your study submission, do you think you could say something like "To perform metric aggregation by spinal levels, the `-vertfile` option in `sct_process_segmentation` was repurposed."?

Or, do you want a new argument specifically to change the column title?

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Ahhh wait, I'm silly. I realize now that the one other function of `-spinal-level` is to change which PAM50 file gets used for normalization.

Hrmmm... that's tricky. Maybe it is worth making a new argument after all. :thinking:

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: yes we can remove, forgot to remove it!

spinalcordtoolbox/aggregate_slicewise.py
sandrinebedard: I think it would be better to have a column name SpinalLevels, so we know when we check the results what reference is used!

There could be a case where we actually want to plot both if we use -perslice

like for the distance pmj, it can be shown as well! But this could be a feature!

sandrinebedard: > Ahhh wait, I'm silly. I realize now that the one other function of `-spinal-level` is to change which PAM50 file gets used for normalization.

yes exactly! for the -normalize-PAM50 we need it!

sandrinebedard: another problem, the argument `-vert` still needs to be used to specify the levels... maybe that is still okay!

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: now removed: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5205/commits/049b5ae7e3ec963b0f840730b656ef5f0cb33cbf

spinalcordtoolbox/scripts/sct_process_segmentation.py
sandrinebedard: now added here: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5205/commits/5f6aec885918245fd2abd0c4e96951a0fabf5937

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -583,7 +583,7 @@ def save_as_csv(agg_metric, fname_out, fname_in=None, append=False):
     if not append or not os.path.isfile(fname_out):
         with open(fname_out, 'w', newline='') as csvfile:
             # spamwriter = csv.writer(csvfile, delimiter=',')
-            header = ['Timestamp', 'SCT Version', 'Filename', 'Slice (I->S)', 'VertLevel', 'DistancePMJ']
+            header = ['Timestamp', 'SCT Version', 'Filename', 'Slice (I->S)', level_column_name, 'DistancePMJ']
```
joshuacwnewton, line 586: Hmm. I am surprised we are hardcoding the `header` here. 

I'd have to double-check, but my assumption would be that the `agg_metric` object would contain all of the necessary information needed to produce an appropriate header. 🤔

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -617,8 +634,8 @@ def main(argv: Sequence[str]):
         data_subject['inter-BV_sex'] = data_subject['brain-volume']*data_subject['sex']
         for line in metrics_agg_merged.values():
             line['MEAN(area)'] = normalize_csa(line['MEAN(area)'], data_predictors, data_subject)
-
-    save_as_csv(metrics_agg_merged, file_out, fname_in=fname_segmentation, append=append)
+    level_column_name = "SpinalLevel" if spinal_level else "VertLevel"
```
joshuacwnewton, line 637: Similarly to my comment about `save_as_csv`, I don't love that we are adding this line in two separate places. I would hope that, once we decide this column name in one place, that would be the only place we need to decide it.

Maybe what we could do is, instead of passing `spinal_level=spinal_level` to `aggregate_per_slice_or_level`, we could simply pass `level_column_name`. That way, we are defining the column name in one place (maybe we define it in the same place we are currently setting `spinal_level=True`), and then we pass it along to both `aggregate_per_slice_or_level` and `save_as_csv`.

That way, we also set the default value (`'VertLevel') in the function definition, too.

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: This is more of an organizational nitpick, and I don't think we actually need to address it here or anything.

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: Ahhh, it's more like, `-vert` should be called `-levels` and `-vertfile` should be called `-levelfile` almost. :laughing: 

It's a shame that we already have the `-vert` and `-vertfile` arguments set in place. I wonder if we could possibly rename the two arguments to `-levels` and `-levelfile` (but retain `-vert` and `-vertfile` as "aliases" for these two arguments). That way, old scripts will still work, but for new usage, the commands are clearer.

joshuacwnewton: I wonder if, instead of adding a new `-spinal_level` argument, it would make more sense to keep the existing `-vertfile` (`-levelfile`) codepath the same, but add a new option called `-leveltype` with the options being (`Vert`, `Spinal`) and then using that string as the column name. 

```python
f"{arg.leveltype}Level"
```

So, in total, these three options would be grouped together:

* `-levels`
* `-leveltype`
* `-levelfile`

That would really communicate that the codepaths are the same for the most part, and it would let us change the PAM50 file that's used based on `if arg.leveltype == "Spinal"`)

joshuacwnewton: Hm. Would this also enable rootlets-based aggregation? Since:

* `-discfile` + `-leveltype Vert` = projected disc labels to do vertlevel aggregation
* `-discfile` + `-leveltype Spinal` = projected rootlets labels to do spinallevel aggregation

Maybe I'm getting too carried away, though, and we should just keep things simple for now.

spinalcordtoolbox/get_spinal_level_from_rootlets.py
joshuacwnewton, line 1: Ah, I see now. This script accomplishes what I described earlier:

* `-discfile` -> disc label projection for vert levels
* `-rootletsfile` -> rootlets label projection for spinal levels

In other words, both methods are trying to convert a point label file into a labeled segmentation for metric aggregation.

joshuacwnewton: Is the idea, then, that we would use the logic in this script to provide a way for users to do rootlets-based spinal level aggregation? Should I review this with the intention of making this an option in `sct_process_segmentation`?

joshuacwnewton (commented):
Thank you so much for your clarifications! Especially the fact that users will still have to aggregate based on `-vert`. I agree that this is problematic and worth looking at.

I have a new batch of comments, though I refrained from reviewing the `get_spinal_level` script just because I'm not sure if it's meant to be integrated into SCT.

spinalcordtoolbox/get_spinal_level_from_rootlets.py
sandrinebedard: This was also my idea in the paper, but it is still less reliable. It could also be provided in another function. To be further discussed I think!

sandrinebedard:
Thanks @joshuacwnewton for the review!

Maybe these are discussions we should have in the SCT meeting, (i.e: `-levelfile `instead of `-vertfile`) etc... I don't mind saying in the paper that it will be intergrated in SCT, and linking the PR for now, for the review, instead of rushing the intergration!

what do you think?

spinalcordtoolbox/get_spinal_level_from_rootlets.py
joshuacwnewton: Oh okay, good to know! So it is logic that could be reviewed (either by me or as part of research discussions) but is not yet validated for integration into SCT / showing off in the course / etc.

joshuacwnewton:
Yes, I see what you mean! I think because it is uncertain what the user-facing interface should look like, I agree with keeping the description "general" and not mentioning any specific option names that are subject to change.

Something like this, maybe?

> Metric aggregation by spinal levels was done using the Spinal Cord Toolbox by passing a label file containing spinal levels to the `sct_process_segmentation` script. Normalization was done using the `-normalize-PAM50` option, which used the spinal level file present in the PAM50 template. 

The specific details (`sct_process_segmentation`, `-normalize-PAM50`, level-based aggregation) are fixed and unlikely to change, while the the specific entry point users would use for level-based aggregation might change.

spinalcordtoolbox/aggregate_slicewise.py
joshuacwnewton: I've added this to a point on the Aug 25th meeting minutes to discuss further. :)

joshuacwnewton: @sandrinebedard It sounds like the conclusion in the Sept 10 meeting was to keep two separate arguments. In that case, I think we can mark this conversation thread as concluded, and stick with the current state of the PR?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -193,6 +193,13 @@ def get_parser(ascor=False):
             Example: ./label/template/PAM50_label_disc.nii.gz
         """),
     )
+    optional.add_argument(
+        '-spinal-level',
+        metavar=Metavar.str,
+        help=textwrap.dedent("""
+            Spinal levels generated by sct_warp_template. Example: `./label/template/PAM50_spinal_levels.nii.gz`
+        """),
+    )
```
joshuacwnewton, line 202: Todo from the SCT meeting:

- Add `-spinal` as a complement to `-vert`
- Change `-spinal-level` to `-spinal-level-file`

spinalcordtoolbox/aggregate_slicewise.py
```diff
@@ -361,12 +361,13 @@ def aggregate_per_slice_or_level(metric, mask=None, slices=[], levels=[], distan
             agg_metric[slicegroup]['DistancePMJ'] = None
 
         # add level info
+        level_column_name = "SpinalLevel" if spinal_level else "VertLevel"
```
joshuacwnewton, line 364: Since the `spinal_level` argument is only used to determine the `level_column_name`, do you think we could turn this directly into an argument with `VertLevel` as default (like we do in `save_as_csv`)?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -445,9 +452,17 @@ def main(argv: Sequence[str]):
...
-        parser.error("Both '-vertfile' and '-discfile' were specified. Please only specify one of these options.")
+    vertlevel_args = ["vertfile", "discfile", "spinal-level"]
+    vertlevel_args_user = [arg for arg in vertlevel_args
+                           if getattr(arguments, arg.replace("-", "_")) is not None]
+    if len(vertlevel_args_user) > 1:
+        parser.error(f"Multiple vertlevel file arguments specified (-{', -'.join(vertlevel_args_user)}). "
+                     f"Please choose only 1 of the following {len(vertlevel_args)} arguments: -{', -'.join(vertlevel_args)}.")
+    elif arguments.spinal_level is not None:
+        fname_vert_level = arguments.spinal_level
+        spinal_level = True
```
joshuacwnewton, line 465: Likewise, instead of setting `spinal_level = True`, we could set `level_column_name` directly here instead.

PR #5211, opened 2026-04-29
Use correct path for `-qc` in `sct_deepseg` command for `lesion-segmentation-ms.rst`

mguaypaq:
The default used by all the other tutorials is `~/qc_singleSubj`, which is useful to collect all of the QC reports together.


joshuacwnewton (approved):

joshuacwnewton:
Nice catch! :)

Issue #2007, opened 2018-08-22
Replace current ANTs binaries by ANTsPy


jcohenadad:
Source: https://github.com/ANTsX/ANTsPy

Pros:
- ANTsPy package installable via pip
- potentially faster i/o than nibabel (see https://github.com/ANTsX/ANTsPy/blob/master/tests/timings_io.py)

Cons:
- might not include `antsSliceRegularizedRegistration` (to check)

kousu:
I'm intrigued by this. I'll take some time to look into it.

PR #5225, opened 2026-05-18
[SUPERSEDED BY #5253] Replace current stitching algorithm with Graf. & al 2024 algorithm

### LINKED ISSUE #5106: Improve stitching quality
When doing online stitching using Siemens algorithm, results are much better than with the current stitching algorithm implemented in SCT. 

Below is a comparison of:
- **SCT** stitching (98161abf01be1d552e7e36ae1786fc8a8bdbe90b)
- **Siemens** stitching (XA61)
- **Separate FOVs** displayed on the same FSLeyes view (automatic registration based on qform header)

Image

What I think could be improved:
- Co-registration between FOVs (it might be tricky if only little overlap is present)
- Gradually weight each adjacent FOV within the overlap region to avoid discontinuity

### LINKED ISSUE #5107: Stitching crops to smallest FOV

GharibSidney:
## Checklist

### Context
The previous stitching implementation (sct_image -stitch) relied on isct_stitching, a precompiled C++ binary from Lavdas, Glocker et al. that had to be distributed separately for each platform (Linux, macOS, Windows). Beyond the maintenance burden of shipping platform-specific binaries, the algorithm used hard-cut blending at overlap regions, which produced visible seams in stitched images.

### Summary
This PR replaces the isct_stitching binary with a vendored copy of the pure-Python stitching algorithm from [TPTBox](https://github.com/Hendrik-code/TPTBox) (Graf et al., 2024), which:

Requires a single new dependency antspyx (scipy and scikit-image are already used by SCT)
Requires no platform binaries
Uses distance-transform ramp blending in overlapping regions, producing smoother results than the previous hard-cut approach
Handles arbitrary input ordering (no longer requires the largest image to be passed first)
Supports any image orientation natively (no forced reorientation to RPI)

### Changes:

Added spinalcordtoolbox/stitching.py — vendored copy of TPTBox/stitching/stitching.py (Header comment documents the upstream source, version).
Modified spinalcordtoolbox/image.py — rewrote stitch_images() to call the vendored main() directly via nibabel objects, with no subprocess or temp files.
Modified spinalcordtoolbox/scripts/sct_image.py — updated -stitch help text citation; passed verbose through to stitch_images(); fixed arguments.stitch is not None -> arguments.stitch (it's a store_true flag).
Modified spinalcordtoolbox/download.py — removed isct_stitching from platform binary lists.

## Linked issues

Fixes #5106.  
Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5107.

### DIFF
--- a/contrib/stitching/stitching.py
+++ b/contrib/stitching/stitching.py
@@ -336,8 +336,6 @@ def main(  # noqa: C901
     store_ramp: bool = False,
     verbose: bool = False,
     min_value: float = 0,
-    bias_field: bool = True,
-    crop_to_bias_field: bool = False,
     crop_empty: bool = False,
     histogram: str | None = None,
     ramp_edge_min_value=5,
@@ -347,7 +345,9 @@ def main(  # noqa: C901
     dtype: type | str = float,
     save=True,
 ):
-    bias_field = True
+    # This is to prevent using antspyx depencies. Set to True if the dependency is ever added
+    bias_field = False
+    crop_to_bias_field = False,
     np.set_printoptions(precision=2, floatmode="fixed")
     if is_segmentation:
         bias_field = False
--- a/contrib/stitching/stitching.py
+++ b/contrib/stitching/stitching.py
@@ -1,3 +1,18 @@
+# Copyright 2026 The TPTBox Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
 # This file is vendored from TPTBox (https://github.com/Hendrik-code/TPTBox)
 # Source: TPTBox/stitching/stitching.py
 #
@@ -332,6 +347,7 @@ def main(  # noqa: C901
     dtype: type | str = float,
     save=True,
 ):
+    bias_field = True
     np.set_printoptions(precision=2, floatmode="fixed")
     if is_segmentation:
         bias_field = False
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,10 +5,6 @@
 # acvl_utils is a utility package used by `nnunetv2`. v0.2.1 broke usage of `nnunetv2`, so skip it:
 # See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4679
 acvl_utils!=0.2.1
-# antspyx provides the `ants` module, optionally used by sct_stitching for N4 bias-field correction.
-# Note: SCT's `sct_image -stitch` always disables bias-field correction, so this dependency is
-# only required if invoking sct_stitching directly with bias-field correction enabled.
-antspyx
 # blosc2 is a utility package used by `nnunetv2`, and v3.9.0 required `numpy==2.0`
 # However, 3.9.0 specifically forgot to specify `>=2.0`, so it will break if installed alongside `numpy==1.26`
 # 3.9.1 and above re-added `numpy==1.26` support, so it is safe to skip this version but allow newer versions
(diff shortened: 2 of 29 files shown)


requirements.txt
```diff
@@ -5,6 +5,10 @@
 # acvl_utils is a utility package used by `nnunetv2`. v0.2.1 broke usage of `nnunetv2`, so skip it:
 # See: https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/4679
 acvl_utils!=0.2.1
+# antspyx provides the `ants` module, optionally used by sct_stitching for N4 bias-field correction.
+# Note: SCT's `sct_image -stitch` always disables bias-field correction, so this dependency is
+# only required if invoking sct_stitching directly with bias-field correction enabled.
+antspyx
```
joshuacwnewton: Thanks for the informative comment here! It really helps to understand the motivation behind the new dependency.

I think for now we should probably not include this new dependency? The reason being is that we have a longstanding issue for adopting ANTsPy, but we've avoided doing so because (afaik) the dependency is quite large:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/2007

Instead, what we are currently doing is precompiling the C++ binaries for the specific ANTs utilities we need and distributing them alongside the installation: https://github.com/spinalcordtoolbox/build_ANTs

If there is a specific utility that the stitching code needs, then maybe we could modify the script to use the binary version of the ANTs tool, then build it and use it the same way we use the other ANTs tools.

contrib/stitching/stitching.py
joshuacwnewton, line 1: Since this is an exact copy of the existing script, then I think it might be better to keep the original filename (`stitching.py`) and also place it inside of the `contrib/` folder in the repo (denoting code we have borrowed from other sources).

Also, it might be good to include a copy of the license file that `stitching.py` is bound to. Or, alternatively, include the license details at the top of the file. This is the approach I took for `lazy_loader.py` too:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/975cf70f92fe9d9e04e92d613b06c81c43fc162d/contrib/tensorflow/lazy_loader.py#L1-L14

Maybe we could do something similar? (We would have to check the specific license used in the TPTBox repo...)

joshuacwnewton (changes requested):
I've just allowed the CI workflows to run -- we just need to see if the tests pass without any issues. We can follow up later today (after the meeting) to see how things go. :)

joshuacwnewton:
Note: The Apptainer failure is just due to the fact that that workflow doesn't work for forks. You can safely ignore it!

(The other errors, though, are worth taking a look at!)

requirements.txt
GharibSidney: Thank you for letting me know, I will remove the ANTsPy dependency and use the precompiled C++ binaries.

joshuacwnewton: If there needs to be an update to the precompiled binaries, just let me know and I can help to create a new release, too. :)

GharibSidney: I started integrating it at https://github.com/GharibSidney/spinalcordtoolbox-binaries/releases/tag/r20260531-n4-v2

That said, as discussed in the last meeting, I was told to keep only the minimal code. In the code I copied from, it is turned on by default, but in our pipeline, even before I applied the new code, it is turned off by default. For stitching it works well without it, but it's activated for segmentation. I don't know if we want to use it as well for segmentation.

The advantage of removing it is that we avoid adding a dependency on https://github.com/spinalcordtoolbox/build_ANTs. The disadvantage is that if we ever want to use n4_bias_field_correction in the future, we would need to re-add the code and the dependency.

Let me know your thoughts, I'm happy to go either way.

contrib/stitching/stitching.py
GharibSidney: Perfect, it should be done in my latest commit. Thanks!

requirements.txt
joshuacwnewton: I think that removing the `n4_bias_field_correction` option from the CLI interface sounds like a good plan. Even if we do remove the code, at least the change itself will be documented as a commit in this PR, meaning it will be very straightforward to undo it in a follow-up PR.

EDIT: To clarify, we can just remove the user-facing parts, but keep the internal code, that way minimal changes would be needed to restore the functionality.

GharibSidney: Perfect I will proceed this way, thank you!

GharibSidney: This is a bit last-minute before Denise takes over, but I made a small change so that n4_bias_field_correction is no longer visible to users. All tests passed except the linting check (I had a trailing space, which has now been fixed). This should make Denise's life easier, as all tests should now pass.

joshuacwnewton:
Given the opening of #5253, I'm going to close this PR in favor of it. :)

Thank you so much @GharibSidney for all of your hard work + getting the ball rolling on this effort! :heart:

PR #5227, opened 2026-05-25
Update third-party GHA actions to Node.js 24+-supporting versions

joshuacwnewton:
## Checklist

## Description

WIP, will also run CI workflows manually where necessary to confirm everything is OK.

## Linked issues

Fixes https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5188.

### DIFF
--- a/.github/workflows/automate-changelog-PR.yml
+++ b/.github/workflows/automate-changelog-PR.yml
@@ -22,14 +22,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Set up Python (to install 'changelog')
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.x'
     - name: Install 'changelog'
       run: |
         pip install git+https://github.com/neuropoly/changelog@master
     - name: Checkout '${{ github.event.repository.name }}'
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         path: ${{ github.event.repository.name }}
     - name: Update '${{ github.event.inputs.fname_changelog }}' for '${{ github.event.repository.name }}'
@@ -42,7 +42,7 @@ jobs:
         cd ${{ github.event.repository.name }}
         echo "${{ github.event.inputs.future_milestone_title }}.dev0" > spinalcordtoolbox/version.txt
     - name: Create pull request for updated changelog and version.txt
-      uses: peter-evans/create-pull-request@v6
+      uses: peter-evans/create-pull-request@v8
       with:
         path: "${{ github.event.repository.name }}"
         branch: "bot/v${{ github.event.inputs.milestone_title }}"
--- a/.github/workflows/automate-changelog-PR.yml
+++ b/.github/workflows/automate-changelog-PR.yml
@@ -42,7 +42,7 @@ jobs:
         cd ${{ github.event.repository.name }}
         echo "${{ github.event.inputs.future_milestone_title }}.dev0" > spinalcordtoolbox/version.txt
     - name: Create pull request for updated changelog and version.txt
-      uses: peter-evans/create-pull-request@v8
+      uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8
       with:
         path: "${{ github.event.repository.name }}"
         branch: "bot/v${{ github.event.inputs.milestone_title }}"
--- a/.github/workflows/check-broken-links.yml
+++ b/.github/workflows/check-broken-links.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
 
     - name: Check for broken links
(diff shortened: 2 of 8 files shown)


AlexVCaron:
While at it, I might suggest starting **pinning all actions to specific commits** instead of versions. Many fail to follow semver conventions and publish breaking changes on minor and micro increments.

joshuacwnewton:
If the main concern is just breakage, I think I'm okay with taking on that maintenance burden? Truthfully I can't recall a time where one of our third-party actions actually broke our CI, but if/when it does in the future, I think I'm OK with our CI being down for a few hours until I push a fix / report upstream / etc.

(The main reason I am hesitant to pin to an exact commit is due to all of the news these days about supply chain attacks, various TeamPCP hacks, etc. While unlikely to target us, I think I'm happy to eagerly update to receive bugfixes when they're issued.)

AlexVCaron:
I do agree with you, if you have the bandwidth to address quickly, it's probably better. But for a **supply chain attack**, it's a double edge sword. Those attacks often exploit the unpinned minor/micro part of a semver to distribute the poisoned content. So, pinning does prevents security updates, but it also prevents attacks, as long as the dependency also pins its own dependencies correctly.

joshuacwnewton:
Really good point! OK, how about we pin the non-GitHub actions but leave GitHub's `actions/` actions as-is? (Assuming we trust that GitHub themselves will be secure, stable stewards of their own actions, and won't be hijacked + will push patched versions of their own actions whenever appropriate (given the global implications)?)

AlexVCaron:
That's totally reasonable !

joshuacwnewton:
(I would have requested your review too @AlexVCaron but GitHub's interface won't let me, perhaps because you haven't been formally invited to the SCT org yet... :thinking:)

mguaypaq (approved):
I agree with the discussion so far: it makes sense to keep a major version pin for `actions/*`, which we expect to be vetted by GitHub, and to keep an exact hash pin for others (and for those, thanks for keeping the corresponding version number in a comment).

AlexVCaron (commented):
Everything is inline with what I proposed ! One thing though : there is now an [official action for artifact download](https://github.com/actions/download-artifact), maybe better to use it ? Else, it's perfect 🤌

joshuacwnewton:
> Everything is inline with what I proposed ! One thing though : there is now an [official action for artifact download](https://github.com/actions/download-artifact), maybe better to use it ? Else, it's perfect 🤌

Ah! Silly me. I thought this third-party action was needed to fetch artifacts from other runs. But, the official action [supports this](https://github.com/actions/download-artifact#download-artifacts-from-other-workflow-runs-or-repositories).

I'll push a fix asap.

joshuacwnewton:
Ah, wait, I recall now. This third-party action was needed for this case:

> Let's suppose you have a workflow with a job in it that at the end uploads an artifact using actions/upload-artifact action and you want to download this artifact in another workflow that is run after the first one. Official actions/download-artifact does not allow this. That's why I decided to create this action. By knowing only the workflow name and commit SHA or other details, you can download the previously uploaded artifact from different workflow associated with that commit or other criteria and use it.

It seems like `actions/download-artifact` requires the exact `run-id`:

```yml
    # The GitHub token used to authenticate with the GitHub API.
    # This is required when downloading artifacts from a different repository or from a different workflow run.
    # Optional. If unspecified, the action will download artifacts from the current repo and the current workflow run.
    github-token:

    # The id of the workflow run where the desired download artifact was uploaded from.
    # If github-token is specified, this is the run that artifacts will be downloaded from.
    # Optional. Default is ${{ github.run_id }}
    run-id:
```

I'm not sure how to get the `run_id` for, for example, "the latest run on `master`" which is what the third-party action accomplished. Maybe we'd have to use the GitHub API for this? Something like (disclaimer: llm-generated, just a sketch + untested):

```yml
      - name: Get latest run ID on master
        id: get-run-id
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          RUN_ID=$(gh api \
            "repos/${{ github.repository }}/actions/workflows/YOUR_WORKFLOW.yml/runs?branch=master&status=success&per_page=1" \
            --jq '.workflow_runs[0].id')
          echo "run-id=$RUN_ID" >> $GITHUB_OUTPUT

      - name: Download artifact
        uses: actions/download-artifact@v4
        with:
          name: your-artifact-name
          run-id: ${{ steps.get-run-id.outputs.run-id }}
          github-token: ${{ secrets.GITHUB_TOKEN }}
```

Maybe this is what the third-party action does internally though, so it's better to have that under our control... :thinking:

AlexVCaron:
@joshuacwnewton I think it's the way to go, since it's such a simple operation. The only way to avoid it is using [reusable workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows), but it's out of scope for this PR.

joshuacwnewton:
It seems like the `gh api`-based steps worked as expected: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/26887757942/job/79334636934?pr=5227#step:6:24

I'm hopeful this is good to merge, then? I'll update the branch and let you have the final say. :)

mguaypaq (approved):

PR #5236, opened 2026-06-03
Add `-normalize-PAM50` argument to metric extraction script to allow interpolation to PAM50 space

valosekj:
## Checklist

## Description

This PR implements per-slice interpolation to `sct_extract_metric` to enable computing DTI maps in the native and then interpolating them per vert level to the PAM50 template space.
I'm reusing [spinalcordtoolbox.metrics_to_PAM50.interpolate_metrics](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4f404367d2cbd210dbd6a1dff7b280118ee7eddd/spinalcordtoolbox/metrics_to_PAM50.py#L14).

image

## Example usage

```bash
sct_extract_metric.py
-i sub-amu01_dwi_moco_FA.nii.gz
-f label_sub-amu01_dwi_moco/atlas
-method map
-l 51
-vertfile label_sub-amu01_dwi_moco/template/PAM50_levels.nii.gz
-perslice 1
-normalize-PAM50 1
-o amu01_dwi_moco_FA.csv
```

<details>
<summary>amu01_dwi_moco_FA.csv sample:</summary>

image

</details>

Files:
[PR5236.zip](https://github.com/user-attachments/files/28558981/PR5236.zip)

## TODO

Due to high DWI slice thickness (5mm), there are “steps” at the “interface” of discs (see [slide 8](https://docs.google.com/presentation/d/1B7LY68pwUoWXgkGDzlBdEYzcd8ci0zVyyCMgYfD4uNU/edit?slide=id.g3e117c1a45b_0_8#slide=id.g3e117c1a45b_0_8)) --> try to include one slice above and below each level for the interpolation. UPDATE: will likely be resolved by https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5305

## Linked issues

Related: https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/pull/41


spinalcordtoolbox/metrics_to_PAM50.py
```diff
@@ -33,10 +33,19 @@ def interpolate_metrics(metrics, fname_vert_levels_PAM50, fname_vert_levels):
...
+                    in zip(level_slices_PAM50[1:-1], level_slices_im[1:-1])
+                    if len(slices_im) > 0]
+    if inner_scales:
+        scale_mean = np.mean(inner_scales)
+    else:
+        # Fall back to all available levels when fewer than 3 levels are present
+        # This is done to pass test on mt/mtr.nii.gz and PAM50_levels.nii.gz that only has 2 levels (C4, C5)
+        all_scales = [len(slices_PAM50)/len(slices_im) for slices_PAM50, slices_im
+                      in zip(level_slices_PAM50, level_slices_im) if len(slices_im) > 0]
+        scale_mean = np.mean(all_scales) if all_scales else np.nan
```
valosekj: This `else` statement was added to pass the test below on `mt/mtr.nii.gz` and `PAM50_levels.nii.gz`, which have only 2 levels (C4, C5).
Otherwise, the `interpolate_metrics` function would fail on the following line because `scale_mean` is `nan` when only two levels are present:

```
x_PAM50 = np.linspace(0, scale_mean * len(slices_im), int(scale_mean * len(slices_im)))
```

For real images, I think we always have >2 levels. 

Alternatively, we can drop it and use `parser.error()`  instead if the vertfile has <3 levels. What do you think?

valosekj:
@sandrinebedard reviewed the PR. @joshuacwnewton, could you also review the PR? Thank you 🙏

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -409,11 +491,32 @@ def main(argv: Sequence[str]):
...
+                nz_native=nz,
+                label_name=label_struc[id_label].name,
+                method=method,
+                fname_vert_level=fname_vert_level,
+                fname_vert_level_PAM50=fname_vert_level_PAM50)
+        else:
+            agg_metric = extract_metric(
+                data, labels=labels, slices=slices, levels=levels, perslice=perslice,
+                perlevel=perlevel, fname_vert_level=fname_vert_level, method=method,
+                label_struc=label_struc, id_label=id_label, indiv_labels_ids=indiv_labels_ids)
```
joshuacwnewton: Given that we already have a check (plus `parser.error`) when the user fails to supply `-perslice 1`), do we still need two separate calls to `extract_metric` here (with hardcoded `perslice` and `perlevel` values?

Or, can we have a single shared call + have just the `_build_pam50_agg_metric` function call gated behind the `if normalize_pam50` check?

joshuacwnewton:
In the progress of reviewing this PR. Just trying to figure out why the tests are failing! :thinking:

testing/cli/test_cli_sct_extract_metric.py
```diff
@@ -48,3 +49,50 @@ def test_sct_extract_metric_vertfile_doesnt_exists():
                                       '-vert', '1:12', '-perlevel', '1',
                                       '-method', 'wa', '-l', '51', '-o', fname_out])
         assert e.value.code == 2
+
+
+@pytest.mark.sct_testing
```
joshuacwnewton: Apologies for the confusion -- this `sct_testing` marker was meant to be a historical artifact from when we adopted `pytest` for the first time. It shouldn't be used for new tests. In fact, it should probably be removed entirely from all tests, as I don't think anyone actually uses it. 😅

testing/cli/test_cli_sct_extract_metric.py
```diff
@@ -48,3 +49,50 @@ def test_sct_extract_metric_vertfile_doesnt_exists():
...
+        '-method', 'wa', '-l', '51',
+        '-vertfile', sct_test_path('mt', 'label', 'template', 'PAM50_levels.nii.gz'),
+        '-perslice', '1', '-normalize-PAM50', '1',
+        '-o', fname_out
+    ])
+    with open(fname_out, 'r') as f:
+        rows = list(csv.DictReader(f))
+    assert len(rows) > 0
+    assert all(r['VertLevel'] != '' for r in rows)
+    assert all(r['WA()'] not in ('', 'nan') for r in rows)
```
joshuacwnewton: This leaves some room for failure cases, I think? (e.g. unexpected number of rows, unexpected values, etc.)

Usually we'll hardcode some values here (down to the tiniest decimals), but I've been wondering about this... maybe it's better to add some rough expectations instead of exact values? e.g.

Here is what I came up with, but I'm curious if there's anything we can improve on:

```python
    assert all(r['Label'] == 'white matter' for r in rows)
    assert all(int(r['VertLevel']) in [4, 5] for r in rows)
    assert all((800 < int(r['Slice (I->S)']) < 870) for r in rows)
    assert all((30 < float(r['WA()']) < 36) for r in rows)
```

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -264,6 +279,70 @@ def get_parser():
     return parser
 
 
+def _build_pam50_agg_metric(agg_metric_native, nz_native, label_name, method,
+                            fname_vert_level, fname_vert_level_PAM50):
```
joshuacwnewton: Is there a reason to place this function here instead of in `metrics_to_PAM50`?

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -264,6 +279,70 @@ def get_parser():
...
+        val = entry.get(primary_key)
+        if val is not None:
+            metric_1d[z] = val
+
+    # Interpolate to PAM50 space; returns Dict[str, Metric] with 1D data of length z_PAM50
+    metrics_pam50 = interpolate_metrics(
+        {primary_key: Metric(data=metric_1d, label=primary_key)},
+        fname_vert_level_PAM50,
+        fname_vert_level
+    )
```
joshuacwnewton: I was curious about the difference between the code in this function vs. the other usage of `interpolate_metrics` (sct_process_segmentation):

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/dded8155bc7be0ded58a01b4883ea22a4385a19d/spinalcordtoolbox/scripts/sct_process_segmentation.py#L567-L579

I guess that `interpolate_metrics` expects the direct output of `compute_shape` (before aggregation), however we are working with the output of `extract_metric` (after aggregation). It is interesting to me that they have different forms.

- `compute_shape`: One `Metric` object per metric, each one has an array for each metric.
- `extract_metric`: One dict per slice, each one has multiple metrics.

They are almost the same thing, but inverted (metric > slice vs. slice > metric).

Maybe we could document this? Something like:

```python
# Convert metrics from `extract_metric` form (one dict per slice, multiple metrics)
# to `compute_shape` form (one `Metric` object per metric, multiple slices)
```

Then later on we could document the opposite:

```python
# Convert interpolated metrics back into the expected form,
# from one `Metric` object per metric back into one dict per slice.
```

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -264,6 +279,70 @@ def get_parser():
...
+
+    # Make the interpolate_metrics output compatible with the expected input of save_as_csv()
+    # Get 1D ndarray with per-slice metric values in PAM50 space
+    pam50_values = metrics_pam50[primary_key].data
+
+    # Determine which vertebral levels are present in the native data to filter PAM50 output
+    im_native_levels = Image(fname_vert_level).change_orientation('RPI')
+    native_levels = set(
+        int(v) for v in np.unique(im_native_levels.data) if 0 < int(v) < 49
```
joshuacwnewton: Maybe we could add a small comment for the magic values of `0` and `49`? (https://spinalcordtoolbox.com/stable/user_section/tutorials/vertebral-labeling/labeling-conventions.html)

spinalcordtoolbox/scripts/sct_extract_metric.py
joshuacwnewton: > I guess that interpolate_metrics expects the direct output of compute_shape (before aggregation), however we are working with the output of extract_metric (after aggregation). 

I guess one consequence of this is that there is now a discrepancy between `sct_process_segmentation` and `sct_extract_metric`:

- `sct_process_segmentation`: Interpolation is done first, aggregation is done in the PAM50 space.
- `sct_extract_metric`: Aggregation is done in the native space, then metrics are interpolated.

This means that if you specify `-normalize-PAM50` for both `sct_process_segmentation` and `sct_extract_metric`, you can't use the same `-z` syntax -- you have to specify two different slice ranges, I think. I think this would be a trap for users?

I wonder if this is problematic enough that we should do a refactor of `sct_extract_metric`. Namely:

- Split up the `extract_metric` function such that it returns a `Metric` object.
- Then do the interpolation (Metric -> Metric).
- Then do the aggregation (Metric -> Slice dict).

That way, we mimic the same workflow between both scripts.

joshuacwnewton (changes requested):
There are a lot of good things to this PR (thanks for adding tests!) but I think I may have found an important thing to discuss in terms of the order of aggregation and interpolation.

spinalcordtoolbox/scripts/sct_extract_metric.py
```diff
@@ -206,6 +208,19 @@ def get_parser():
...
+        metavar=Metavar.int,
+        type=int,
+        choices=[0, 1],
+        default=0,
+        help="Set to 1 to interpolate the extracted metric values into PAM50 anatomical "
+             "dimensions per slice. "
+             "Requires `-vertfile` and `-perslice 1`. "
+             "Inspired by: Valošek J, Bédard S et al. Imaging Neuroscience 2024. "
+             "https://doi.org/10.1162/imag_a_00075"
+    )
```
joshuacwnewton, line 223: Are the normalized values limited to DWI/DTI (based on PR title)? Or is this a generalized normalization (I assume so?).

If it's a general normalization, then maybe we could rewrite the title of the PR... 🤔

valosekj: Good point! It's general. I fixed the PR title

testing/cli/test_cli_sct_extract_metric.py
valosekj: Thanks! Removed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/commits/47fafc40af9c8eaf200b061da2c71de22fc85c92

spinalcordtoolbox/scripts/sct_extract_metric.py
valosekj: Good point! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/commits/42a3461f14012eb084e60db2de5be44cb628be51

spinalcordtoolbox/scripts/sct_extract_metric.py
valosekj: > Maybe we could document this? Something like:

Good idea! Thanks! Documented in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/commits/2ad22cde1ad75f48bd6adae818774e74d937cc7d

valosekj: > This means that if you specify `-normalize-PAM50` for both `sct_process_segmentation` and `sct_extract_metric`, you can't use the same `-z` syntax -- you have to specify two different slice ranges, I think. I think this would be a trap for users?

Excellent catch! I hope it's okay, at least for now, as the intended use of the PAM50 interpolation is to get all the slices in the PAM50 space. So I think the `-z` syntax is not actually used.

testing/cli/test_cli_sct_extract_metric.py
valosekj: Good point! Done in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/changes/a564810888f0e6e267452d8cfac6ba97fdba2cfc

spinalcordtoolbox/scripts/sct_extract_metric.py
valosekj: Agree. Thanks! Simplified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/changes/bbf77551c702b19b0ce4595a1897e08f08edf8a2

spinalcordtoolbox/scripts/sct_extract_metric.py
valosekj: Good point! Moved to `metrics_to_PAM50` in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/commits/7f04548250cee574b5886e50660ca81e1364d9e8

spinalcordtoolbox/scripts/sct_extract_metric.py
joshuacwnewton: Good point. Maybe we can add a note in both `-z` argument's help text warning users that z slices must correspond to [PAM50, native] space depending on the script?

joshuacwnewton: I've opened https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5292 for the refactoring as a follow-up task, for now we can just stick with the help text. :)

valosekj: > I've opened https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5292 for the refactoring as a follow-up task, for now we can just stick with the help text. :)

Thank you! Sounds good.

valosekj: > Good point. Maybe we can add a note in both `-z` argument's help text warning users that z slices must correspond to [PAM50, native] space depending on the script?

Good point! Clarified in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/commits/6b27442c6946535865357a32969a52bc987f0ce1

spinalcordtoolbox/metrics_to_PAM50.py
joshuacwnewton: If I'm understanding correctly, this arises because the test uses:

```
        '-vertfile', sct_test_path('mt', 'label', 'template', 'PAM50_levels.nii.gz'),
```

And this pre-warped template only contains 2 vertebral levels:

Image

I think that the idea of "trim the first and last level if possible, fall back to using all of the levels" sounds sensible to me?

joshuacwnewton: Though, I am wondering about the `scale_mean=np.nan` fallback:

```python
scale_mean = np.mean(all_scales) if all_scales else np.nan
```

I don't think it's possible for this `else` case to even fire, since:

- `all_scales` has to be an empty list
- `all_scales` is only empty if `len(slices_im) == 0` in the earlier list comprehension (I think you put this here to avoid division by zero, right?)
- But, I think it is impossible for `len(slices_im) == 0`, because the vertlevels are based on `levels` earlier in the code:
  ```python
  levels = sorted(int(level) for level in np.unique(im_seg_labeled.data) if 0 < int(level) < 49)
  level_slices_im = [get_slices_from_vertebral_levels(im_seg_labeled, level) for level in levels]
  ```
- Since we use `np.unique` here, won't there always be non-empty slices corresponding to each vert level? (Or at least, this would be a safe assumption to make?)

joshuacwnewton: In summary, I'm wondering if we could make the following simplification:

```python
    # Compute the mean scaling factor between PAM50 and native slices
    pairs = list(zip(level_slices_PAM50, level_slices_im))
    # Exclude the first/last levels to avoid edge effects (only if there are enough levels)
    trim = 1 if len(pairs) > 2 else 0
    scale_mean = np.mean([len(s_pam) / len(s_im) 
                          for s_pam, s_im in pairs[trim:len(pairs)-trim]])
```

(I also de-duplicated the two different list comprehensions using a `trim` variable.)

spinalcordtoolbox/metrics_to_PAM50.py
```diff
@@ -75,3 +84,71 @@ def interpolate_metrics(metrics, fname_vert_levels_PAM50, fname_vert_levels):
 
     # Convert dict of ndarrays to dict of Metric() objects
     return {k: Metric(data=np.array(v), label=k) for k, v in metrics_PAM50_space_dict.items()}
+
+
+def build_pam50_agg_metric(agg_metric_native, nz_native, label_name, method,
+                           fname_vert_level, fname_vert_level_PAM50):
+    """
+    Interpolate per-slice native-space metrics into PAM50 anatomical dimensions.
```
joshuacwnewton: I got confused by the docstring here, because of how similar it is to the docstring of the `interpolate_metrics` function:

```diff
-    Interpolates metrics perlevel into the PAM50 anatomical dimensions.
+    Interpolate per-slice native-space metrics into PAM50 anatomical dimensions.
```

Maybe it would be clearer if we said something like:

```suggestion
    Adapt `interpolate_metrics()` to the per-slice output of `extract_metric()`.
```

joshuacwnewton (approved):
I think this LGTM! I just had a few small comments to address the remaining open conversations, but I think the overall structure/plan of this PR seems sound, and I appreciate the new tests. :)

spinalcordtoolbox/metrics_to_PAM50.py
valosekj: agree! Thanks!

spinalcordtoolbox/metrics_to_PAM50.py
valosekj: > I don't think it's possible for this `else` case to even fire, since:

Excellent catch! Indeed, I also think that `else: scale_mean = np.nan` is dead code.

valosekj: Agree! Thanks for the suggestion! I think it's correct. Added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5236/changes/14d0849b5036cf59bb63245352c3033d54a2eb0f

valosekj:
> I think this LGTM! I just had a few small comments to address the remaining open conversations, but I think the overall structure/plan of this PR seems sound, and I appreciate the new tests. :)

Thank you for your review and suggestions, @joshuacwnewton! 

Let me test the PR on the spine-generic dataset before merging it!

joshuacwnewton:
> Let me test the PR on the spine-generic dataset before merging it!

Sounds good to me! :)

PR #5185, opened 2026-03-05
Replace `batch_processing.sh` with `batch_single_subject.sh`

joshuacwnewton:
## Checklist

- [x] **PR Sidebar**: I've filled these options within the PR sidebar:
  - [Labels](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#233-labels) (**exactly one** dark purple label!)
  - [Milestone](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#234-milestone)
- [x] **Guidelines**: I've read the [Contributing guidelines](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)) to make sure my [branch](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-1-branches) and [pull request](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing%3A-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#part-2-pull-requests) meet SCT's guidelines. (Feel free to stop and fixup your commits before submitting.)
- [x] **Documentation**: I've updated the [relevant documentation](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Documentation) for my changes, including argparse descriptions, docstrings, and ReadTheDocs tutorial pages.
- [x] **Tests**: I've added [relevant tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing) for my contribution, if necessary. I've also made sure that my contribution passes [all of the automated tests](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#24-checking-the-automated-tests) (which will be triggered after the PR is submitted).
- [x] Only **after** the automated test suite passes: I will tag a [reviewer](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers) in the PR sidebar. (You can also ask for help with specific questions in the PR comments before the tests pass.)

## Description

[...]

PR #5239, opened 2026-06-05
Test each tutorial individually to save processing time on PRs

joshuacwnewton:
## Checklist

## Description

This PR parallelizes the testing of `batch_single_subject.sh` using:

- **`split_batch_single_subject.py`**: A Python script that looks for certain comment lines in the script, and splits the script into multiple mini-scripts.
- **`test-batch-processing-matrix.yml`**: A GitHub Actions workflow that takes each of the mini-scripts and runs a matrix of CI jobs (one for each mini-script), then collects the output CSV files and runs our usual `batch_processing.sh` test on the outputs.

> [!NOTE]
> The first draft of this PR (`.py` + `.yml` file) was partially generated using an LLM. I've self-reviewed the additions and rewritten parts of the scripts myself, and I've also written the documentation commits entirely myself.

There are many benefits to parallelizing the test for `batch_single_subject.sh`:

- CI is much faster compared to running the script all at once (~20m vs 1hr30m+)
- Each tutorial is tested in isolation in a fresh workspace, meaning that we properly catch any missing data dependencies automatically.

As part of this PR, the CI workflow found a handful of discrepancies between the script and the tutorials (e.g. places where a prerequisite step was placed at the end of one tutorial instead of the start of the next). So, I've manually written fixes for each discrepancy such that all of the 

## Linked issues

Sub-task of #5185.

### DIFF
--- a/.github/workflows/scripts/split_batch_single_subject.py
+++ b/.github/workflows/scripts/split_batch_single_subject.py
@@ -166,9 +166,9 @@ def rewrite_cd_commands(body_lines: list, start_cwd: str = '') -> tuple:
 
 # helper functions when generating mini-script files
 def make_slug(s):
-    s = s.lower()
-    s = re.sub(r"[^a-z0-9]+", '-', s)
-    s = re.sub(r'-+', '-', s).strip('-')
+    s = s.lower()                          # lowercase
+    s = re.sub(r"[^a-z0-9]+", '-', s)      # replace non-alphanumeric with hyphen
+    s = re.sub(r'-+', '-', s).strip('-')   # collapse multiple and strip leading/trailing hyphens
     return s or 'section'
 
 
--- a/.github/workflows/scripts/split_batch_single_subject.py
+++ b/.github/workflows/scripts/split_batch_single_subject.py
@@ -134,7 +134,7 @@ CD_LINE_RE = re.compile(r'^(\s*)cd\s+(\S+)(\s*(?:#.*)?)$')
 # - Optional leading `../`.
 # - A single directory name, with no special characters.
 # - Optional trailing slash `/`.
-CD_PATH_RE = re.compile(r'^(?:\.\./)?(?<dir_name>[-_0-9A-Za-z]+)/?$')
+CD_PATH_RE = re.compile(r'^(?:\.\./)?(?P<dir_name>[-_0-9A-Za-z]+)/?$')
 
 
 def rewrite_cd_commands(body_lines: list, start_cwd: str = '') -> tuple:
--- a/.github/workflows/scripts/split_batch_single_subject.py
+++ b/.github/workflows/scripts/split_batch_single_subject.py
@@ -120,15 +120,21 @@ def parse_sections(lines):
 
 # Assumptions made to simplify the code:
 #   * Data directories are always a single subdirectory name, optionally
-#     preceded by `../` (e.g. `cd t2`, `cd ../mt`, `cd "$DIR"`).
+#     preceded by `../` (e.g. `cd t2`, `cd ../mt`).
 #   * Data directories are always relative to `$DATA_DIR`.
-#   * No spaces, glob characters, or multi-part paths.
+#   * No spaces, glob characters, variables, or multi-part paths.
 
 # NB: regex that matches a live (non-commented) `cd TARGET` line.
 # - Group 1: leading whitespace
 # - Group 2: path (which may be quoted)
 # - Group 3: trailing whitespace / inline comment (preserved verbatim)
-CD_RE = re.compile(r'^(\s*)cd\s+(\S+)(\s*(?:#.*)?)$')
+CD_LINE_RE = re.compile(r'^(\s*)cd\s+(\S+)(\s*(?:#.*)?)$')
+
+# NB: regex that validates our path assumptions.
+# - Optional leading `../`.
+# - A single directory name, with no special characters.
+# - Optional trailing slash `/`.
+CD_PATH_RE = re.compile(r'^(?:\.\./)?(?<dir_name>[-_0-9A-Za-z]+)/?$')
 
 
 def rewrite_cd_commands(body_lines: list, start_cwd: str = '') -> tuple:
@@ -136,20 +142,19 @@ def rewrite_cd_commands(body_lines: list, start_cwd: str = '') -> tuple:
     result = []
     for line in body_lines:
         # Deconstruct the `cd` command from the regex groups (if present)
-        m = CD_RE.match(line)
-        if not m:
+        m_cd_line = CD_LINE_RE.match(line)
+        if not m_cd_line:
             result.append(line)
             continue
-        indent, dir_name, tail = m.groups()
+        indent, target, tail = m_cd_line.groups()
 
-        # Strip surrounding quotes (but only in pairs, just in case of `cd "$ENV"/path`
-        if len(dir_name) >= 2 and dir_name[0] in ('"', "'") and dir_name[-1] == dir_name[0]:
-            dir_name = dir_name[1:-1]
+        m_cd_path = CD_PATH_RE.match(target)
+        if not m_cd_path:
+            raise ValueError(f'Could not parse `cd` target "{target}" in input file.')
+        dir_name = m_cd_path.group('dir_name')
+
         # Both `cd t2` and `cd ../t2` resolve to the same subdirectory
         # relative to `$DATA_DIR` given the flat, single-depth layout.
-        if dir_name.startswith('../'):
-            dir_name = dir_name[3:]
-
         current_cwd = dir_name
         result.append(f'{indent}cd "$DATA_DIR/{current_cwd}"{tail}\n')
 
@@ -184,7 +189,7 @@ def section_starts_with_cd(body_lines):
         # even if one comes later.
         if stripped.startswith('sct_'):
             return False
-        if bool(CD_RE.match(line)):
+        if bool(CD_LINE_RE.match(line)):
             return True
     return False
 
--- a/.github/workflows/scripts/split_batch_single_subject.py
+++ b/.github/workflows/scripts/split_batch_single_subject.py
@@ -1,0 +1,310 @@
+import argparse
+import json
+import os
+import posixpath
+import re
+import stat
+import sys
+from pathlib import Path
+
+
+HEADER_RE = re.compile(r'^\s*#\s*(?P<title>.+?)\s*$')
+SEPARATOR_RE = re.compile(r'^\s*#\s*={5,}\s*$')
+DATASET_IN_TITLE_RE = re.compile(r'\(dataset:\s*(?P<dataset>[^\)]+)\)')
+
+# FIXME: List of dataset zip assets published on the sct_tutorial_data
+#   GitHub release page.  Any dataset in this set that is NOT referenced by at
+#   least one section in batch_single_subject.sh will be reported so it can be
+#   manually reviewed and a header added. We probably want to generate this
+#   automatically from the release assets using the GitHub API at some point
+KNOWN_DATASETS = {
+    # "data_batch-processing-of-subjects",
+    #     NB1: this one is intentionally excluded: it is a multi-subject
+    #          workflow tutorial (using sct_run_batch) that is handled by a separate script/pipeline
+    #          and does not belong in batch_single_subject.sh.
+    # "data_normalizing-morphometrics-compression",
+    #     NB2: this one is intentionally excluded: it is a special case that only
+    #          provides pre-generated intermediate files (segmentation, vertebral labels,
+    #          compression labels) that users can download to skip the generation steps
+    #          already covered by the data_compression section.  All commands that consume
+    #          these files are included in the data_compression section, so no separate section
+    #          header is needed in batch_single_subject.sh.
+    "data_atlas-based-analysis",
+    "data_compression",
+    "data_contrast-agnostic-registration",
+    "data_coregistration",
+    "data_gm-wm-metric-computation",
+    "data_gm-wm-segmentation",
+    "data_improving-registration-with-gm-seg",
+    "data_lesion-analysis",
+    "data_lumbar-registration",
+    "data_ms-lesion-segmentation",
+    "data_mtr-computation",
+    "data_processing-dmri-data",
+    "data_processing-fmri-data",
+    "data_rootlets-registration",
+    "data_shape-metric-computation",
+    "data_spinalcord-segmentation",
+    "data_spinalcord-smoothing",
+    "data_template-registration",
+    "data_vertebral-labeling",
+    "data_visualizing-misaligned-cords",
+}
+
+# ==============================================================================
+#               Split batch_single_subject.sh into mini-scripts
+# ==============================================================================
+# NB: Sections are defined by the header:
+#
+#     # Section Title (dataset: data_[DATASET_NAME])
+#     # ============================================
+#
+# Sections with headers that don't contain `(dataset: ...)` are skipped.
+
+
+def parse_sections(lines):
+    sections = []
+    skipped = []
+    i = 0
+    n = len(lines)
+    # Find START OF SCRIPT marker present in batch_single_subject.sh to avoid its preamble
+    for idx, line in enumerate(lines):
+        if 'START OF SCRIPT' in line:
+            i = idx + 1
+            break
+
+    while i < n:
+        line = lines[i]
+        m = HEADER_RE.match(line)
+        if m and i + 1 < n and SEPARATOR_RE.match(lines[i + 1]):
+            # extract the dataset name and clean it from the title
+            title = m.group('title').strip()
+            match = DATASET_IN_TITLE_RE.search(title)
+            if match:
+                dataset = match.group('dataset').strip()
+                title = DATASET_IN_TITLE_RE.sub('', title).strip()
+
+            # if we found a header, but it doesn't contain a dataset marker, skip it and all lines until the next header
+            else:
+                j = i + 2
+                while j < n:
+                    next_m = HEADER_RE.match(lines[j])
+                    if next_m and j + 1 < n and SEPARATOR_RE.match(lines[j + 1]):
+                        break
+                    j += 1
+                skipped.append({'title': title, 'start_line': i + 1})
+                print(
+                    f'WARNING: skipping section at line {i + 1} '
+                    f'"{title}" — no (dataset: ...) marker in header.',
+                    file=sys.stderr,
+                )
+                i = j
+                continue
+
+            # collect body until next header+separator or EOF
+            body_lines = []
+            j = i + 2
+            while j < n:
+                next_m = HEADER_RE.match(lines[j])
+                if next_m and j + 1 < n and SEPARATOR_RE.match(lines[j + 1]):
+                    break
+                body_lines.append(lines[j])
+                j += 1
+
+            sections.append({
+                'title': title,
+                'dataset': dataset,
+                'start_line': i + 1,
+                'end_line': j,
+                'body': body_lines,
+            })
+            i = j
+        else:
+            i += 1
+
+    return sections
+
+
+# ==============================================================================
+#            Handle the `cd` commands that may or may not be present
+# ==============================================================================
+# Each mini-script is run fresh from `$DATA_DIR`.  The original
+# `batch_single_subject.sh` chains all sections together, so a later section
+# can write e.g. `cd ../mt` because the previous section left the shell inside
+# `t2/`.  That navigation breaks when the section runs in isolation. So, we
+# replace every `cd RELATIVE` with `cd "$DATA_DIR/RESOLVED"` so that the
+# isolated mini-script always navigates correctly regardless of where the shell
+# started.
+
+
+# NB: regex that matches a live (non-commented) bare `cd TARGET` line.
+# - Group 1: leading whitespace
+# - Group 2: target path
+# - Group 3: trailing whitespace / inline comment (preserved verbatim)
+CD_RE = re.compile(r'^(\s*)cd\s+(\S+)([ \t]*(?:#.*)?)$')
+
+
+def resolve_cd(target, virtual_cwd):
+    """Return the new virtual_cwd after `cd TARGET`, rooted at DATA_DIR (='').
+
+    Uses POSIX path arithmetic so that `..` at the root stays at the root
+    (same behaviour as a real shell on a dataset root).
+    """
+    rooted = posixpath.normpath('/' + virtual_cwd + '/' + target)
+    rel = rooted.lstrip('/')
+    return '' if rel == '.' else rel
+
+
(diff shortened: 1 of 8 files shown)


.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,186 @@
...
+          find ${OUTPUT_DIR} -type f -name '*.csv' -print || true
+
+      - name: Validate new results
+        run: |
+          # Auto-pass macOS as it would fail due to https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3194
+          if [ "$RUNNER_OS" == "macOS" ]; then
+               sct_testing testing/batch_single_subject/test_batch_single_subject.py || exit 0
+          else
+               sct_testing testing/batch_single_subject/test_batch_single_subject.py
+          fi
```
joshuacwnewton: Splendid! The tutorial matrix worked, and the batch processing results were validated: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/27159203547

And, best of all, the CI workflow took only 19 minutes!

Image

The longest section was the contrast agnostic registration with deep learning: https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/27159203547/job/80169978445

joshuacwnewton (commented):
I'll just need to clean up the commits a bit and I'll mark this as ready for review, but a very good success for a first draft.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+import posixpath
+import re
+import stat
+import sys
+from pathlib import Path
+
+
+HEADER_RE = re.compile(r'^\s*#\s*(?P<title>.+?)\s*$')
+SEPARATOR_RE = re.compile(r'^\s*#\s*={5,}\s*$')
+DATASET_IN_TITLE_RE = re.compile(r'\(dataset:\s*(?P<dataset>[^\)]+)\)')
```
mguaypaq: For consistency with `HEADER_RE`, I would suggest:
```suggestion
DATASET_IN_TITLE_RE = re.compile(r'\(dataset:\s*(?P<dataset>[^)]+?)\s*\)')
```

- Like `HEADER_RE` this would use a non-greedy match to trim whitespace after the dataset name.
- I believe (from some interactive testing on the single characters `\`, `)` and `a`) that `r'[^\)]'` behaves the same way as `r'[^)]'` (but differently from `r'[^\\)]'`), so we might as well go with the shorter form.

mguaypaq: Also, I think it's smart that this regex is as permissive as possible in the characters it allows for dataset names: that way, it will catch everything looks like `(dataset:...)` and be able to raise an error, rather than silently smushing two sections together.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+def parse_sections(lines):
+    sections = []
+    skipped = []
+    i = 0
+    n = len(lines)
+    # Find START OF SCRIPT marker present in batch_single_subject.sh to avoid its preamble
+    for idx, line in enumerate(lines):
+        if 'START OF SCRIPT' in line:
+            i = idx + 1
+            break
```
mguaypaq: This will silently start at the top of the file if `START OF SCRIPT` is not found. Do we want to be permissive here? (I genuinely don't know.) If we want to be strict, detecting the error would be `for...else` (and no need for the initial `i = 0`):

```suggestion
    n = len(lines)
    # Find START OF SCRIPT marker present in batch_single_subject.sh to avoid its preamble
    for idx, line in enumerate(lines):
        if 'START OF SCRIPT' in line:
            i = idx + 1
            break
    else:
        handle_the_error()
```

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+
+    while i < n:
+        line = lines[i]
+        m = HEADER_RE.match(line)
```
mguaypaq: Since there are multiple regex matches in this code, I would suggest giving it a real name like `header`, or possibly `header_m`.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+        if m and i + 1 < n and SEPARATOR_RE.match(lines[i + 1]):
+            # extract the dataset name and clean it from the title
+            title = m.group('title').strip()
```
mguaypaq: There's no need for the `.strip()`, since that's already handled by the regex.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            match = DATASET_IN_TITLE_RE.search(title)
```
mguaypaq: Similarly, maybe `dataset_m` for this variable name.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            if match:
+                dataset = match.group('dataset').strip()
```
mguaypaq: Again, the `.strip()` would be handled already by my suggested regex change above.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+                title = DATASET_IN_TITLE_RE.sub('', title).strip()
```
mguaypaq: This truly doesn't matter, but the `.strip()` here kind of assumes that the `(dataset:...)` match was at one end of the title, which isn't enforced by the regex.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            sections.append({
+                'title': title,
+                'dataset': dataset,
+                'start_line': i + 1,
+                'end_line': j,
+                'body': body_lines,
+            })
+            i = j
+        else:
+            i += 1
```
mguaypaq: This is so far from the corresponding `if` statement. Maybe it would be worth doing this? It would also save an indentation level for the meaty path.

```py3
if not (...):
    i += 1
    continue
```

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+        if m and i + 1 < n and SEPARATOR_RE.match(lines[i + 1]):
+            # extract the dataset name and clean it from the title
+            title = m.group('title').strip()
+            match = DATASET_IN_TITLE_RE.search(title)
+            if match:
+                dataset = match.group('dataset').strip()
+                title = DATASET_IN_TITLE_RE.sub('', title).strip()
+
+            # if we found a header, but it doesn't contain a dataset marker, skip it and all lines until the next header
+            else:
```
mguaypaq: If you take my suggestion below about negating the `if` condition, I would also apply it here for consistency.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+                    next_m = HEADER_RE.match(lines[j])
+                    if next_m and j + 1 < n and SEPARATOR_RE.match(lines[j + 1]):
+                        break
+                    j += 1
+                skipped.append({'title': title, 'start_line': i + 1})
+                print(
+                    f'WARNING: skipping section at line {i + 1} '
+                    f'"{title}" — no (dataset: ...) marker in header.',
+                    file=sys.stderr,
+                )
```
mguaypaq: This could be moved before the `while j < n` loop since it doesn't depend on `j`, and then the loop could manipulate `i` directly.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+                    break
+                body_lines.append(lines[j])
+                j += 1
+
+            sections.append({
+                'title': title,
+                'dataset': dataset,
+                'start_line': i + 1,
+                'end_line': j,
+                'body': body_lines,
```
mguaypaq: Since we know `i` and `j`, maybe replace `body_lines` by `lines[i+2:j]`?

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            })
```
mguaypaq: This is a bigger change, but I think you could de-duplicate some of the `HEADER_RE.match` and `SEPARATOR_RE.match` patterns by doing this loop in two passes:
1. First extract `title`, `start_line`, `end_line` from all detected headers.
2. Then select only the ones where `title` has a dataset (and annotate the other `sections` fields).

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+# replace every `cd RELATIVE` with `cd "$DATA_DIR/RESOLVED"` so that the
+# isolated mini-script always navigates correctly regardless of where the shell
+# started.
+
+
+# NB: regex that matches a live (non-commented) bare `cd TARGET` line.
+# - Group 1: leading whitespace
+# - Group 2: target path
+# - Group 3: trailing whitespace / inline comment (preserved verbatim)
+CD_RE = re.compile(r'^(\s*)cd\s+(\S+)([ \t]*(?:#.*)?)$')
```
mguaypaq: Is there a reason for `[ \t]` instead of `\s`?

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def resolve_cd(target, virtual_cwd):
+    """Return the new virtual_cwd after `cd TARGET`, rooted at DATA_DIR (='').
+
+    Uses POSIX path arithmetic so that `..` at the root stays at the root
+    (same behaviour as a real shell on a dataset root).
+    """
+    rooted = posixpath.normpath('/' + virtual_cwd + '/' + target)
```
mguaypaq: I was going to say that it's a bit weird mixing `pathlib`-style paths and `os.path`-style paths (with `posixpath`), but I see that it's necessary because of `pathlib.PurePosixPath`'s quixotic refusal to implement `.resolve()`. (To be fair, they have good points about the subtlety of symlinks, but it remains inconvenient.)

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def rewrite_cd_commands(body_lines: list, start_cwd: str = '') -> tuple:
+    virtual_cwd = start_cwd
+    result = []
+    for line in body_lines:
+        # Leave comment lines alone (they may contain example `cd` commands)
+        if line.lstrip().startswith('#'):
+            result.append(line)
+            continue
```
mguaypaq: Given that `CD_RE` is anchored at the start of the line and only matches uncommented commands, I think this paragraph is redundant?

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+
+        m = CD_RE.match(line)
+        if m:
+            indent, target, tail = m.group(1), m.group(2), m.group(3)
```
mguaypaq: I think the right-hand side can be replaced by `m.groups()`.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            # Absolute paths, shell variables ($…), home (~), subshells (`…)
+            # — leave them verbatim but still track the cwd change if possible.
```
mguaypaq: The `still track the cwd change` seems false, given the `continue` statement below?

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,310 @@
...
+            if target[0] in ('/', '$', '~', '`'):
```
mguaypaq: This bit is weird:
- None of the `cd` commands actually in `batch_single_subject` start with these characters.
- Yet this _would_ miss the `cd` command in the preamble that starts with `"$`.

I think trying to cover the entire `bash` syntax is hopeless here, so we should maybe just validate the `target` with a very strict regex, separate from `CD_RE`, so that we can raise an error if something _looks_ like a `cd` command but breaks our expectations.

In fact, I think we could support everything we want with just the pattern `../singleword`? In which case, we don't even need to mess with general path logic.

mguaypaq (commented):
Sorry for the deluge of comments! I've just been commenting on everything in the script as I read through it. Here's what I have so far for today.

.github/workflows/scripts/split_batch_single_subject.py
joshuacwnewton: Ahhhh, you're right! The old regex includes any whitespace in the capture group, hence the abundance of `strip()` statements.

Image

Image

.github/workflows/scripts/split_batch_single_subject.py
joshuacwnewton: I think enforcing the assumption is good? It at least makes sure that if a dev ever edits the line, we'll get a noisy error that will encourage them to put it back.

.github/workflows/scripts/split_batch_single_subject.py
```diff
@@ -0,0 +1,291 @@
...
+    if unaccounted:
+        print(
+            '\nNOTE: The following datasets are published as release assets but have NO '
+            'corresponding section (with a dataset: header) in the batch script.\n'
+            'These should be reviewed and a new section header added if applicable:',
+            file=sys.stderr,
+        )
+        for ds in unaccounted:
+            print(f'  - {ds}', file=sys.stderr)
+        print(file=sys.stderr)
```
mguaypaq, line 291: Should this maybe `return 1`? Otherwise I think it's very likely that this warning wouldn't be seen.

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+      sections: ${{ steps.split.outputs.sections }}
+    steps:
+      - uses: actions/checkout@v6
+
+      # NB: GHA matrix requires {"include": [...]} format
+      - name: Run parser to split batch_single_subject.sh
+        id: split
+        run: |
+          python3 .github/workflows/scripts/split_batch_single_subject.py batch_single_subject.sh --out scripts/batch_single_subject
+          JSON=$(jq -c '{include: .}' scripts/batch_single_subject/sections.json)
```
mguaypaq, line 25: I was going to suggest putting the `{include: ...}` wrapper directly in the `split_batch_single_subject` script, but I guess this `jq -c` call also puts everything on a single line with no extraneous spaces, which is maybe important too, for the `"sections=$JSON"` syntax below.

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+          echo "sections=$JSON" >> $GITHUB_OUTPUT
+
+      - name: Upload mini-scripts artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: mini-scripts
+          path: scripts/batch_single_subject/
+
+  # conda environment installation is time-consuming, and we want to run the same environment across all of the
+  # tutorial sections, so we install SCT once and cache the installation for subsequent jobs
```
mguaypaq, line 35: Smart!

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+
+      - name: Restore cached SCT installation
+        id: cache-sct
+        uses: actions/cache@v5
+        with:
+          path: |
+            python/
+            bin/
+            data/
+          key: sct-${{ runner.os }}-${{ hashFiles('requirements.txt', 'setup.cfg', 'setup.py', 'install_sct') }}
```
mguaypaq: I was worried whether this cache would silently test the scripts on a stale version of SCT, but I see that it should be ok:
- This only caches fairly "stable" directories of the installation.
- Since SCT uses `pip install -e .`, the actual SCT code (which would normally be under `python/`) is just sym-linked to the git checkout, which is outside the cached directories.
- The cache key will match and old SCT commit most of the time, but will change when dependencies change.

Ah, wait, should the cache key depend on `requirements-freeze.txt` instead, to catch updated versions of the dependencies? (And updated deepseg models?) Except I guess this file only gets generated for releases, not `master`.

Maybe it would be safer to key the cache on `runner.os` and the SCT repo commit hash? This would mean that the workflow always installs SCT, but only once, before running the scripts in parallel, right?

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+
+      - name: Install SCT (cache miss fallback)
+        if: steps.cache-sct.outputs.cache-hit != 'true'
+        run: ./install_sct -y
```
mguaypaq: If we always expect the cache to exist (which I think we do, since it's a pre-requisite step before the parallel `strategy`?), and we want to be notified when it doesn't, maybe the cache step should use `fail-on-cache-miss` instead of having a fallback installation step.

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          name: mini-scripts
+          path: scripts/batch_single_subject
+
+      # NB: we derive the release base url from conf.py's sct_tutorial_data extlinks setting (i.e. same as tutorials)
+      - name: Download dataset and set DATA_DIR
+        run: |
+          echo "Section: ${{ matrix.name }}"
+          echo "Dataset: ${{ matrix.dataset }}"
+          BASE_URL=$(python3 -c "import re, sys; m = re.search(r\"'sct_tutorial_data':\\s*\\('([^']+)'\", open('documentation/source/conf.py').read()); print(m.group(1).replace('%s', ''))")
```
mguaypaq: I'm a little uncomfortable with this line, since it involves doing search-and-replace using regex on a python script in python in bash in yaml, with several layers on quoting and escaping.

Since `documentataion/source/conf.py` is a full-fledged script, could it get the `sct_tutorial_data` base url from, say, a small text file in the SCT repo? And then this line in the workflow could read the same file? Or some other single source of truth that would be conveniently accessible from both contexts.

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+          mkdir -p "$SCT_DIR/data"
+          unzip -q "${{ matrix.dataset }}.zip" -d "$SCT_DIR/data/"
+          
+          EXTRACT_DIR=$(unzip -Z1 "${{ matrix.dataset }}.zip" | head -1 | sed 's|/.*||')
+          echo "Zip extracted top-level dir: $EXTRACT_DIR"          
+          DATA_DIR="$SCT_DIR/data/$EXTRACT_DIR/data"
+          echo "DATA_DIR=$DATA_DIR" >> $GITHUB_ENV
+          
+          # debug tree
+          find "$SCT_DIR/data/$EXTRACT_DIR" -maxdepth 3 | sort | head -60
```
mguaypaq, line 107: Nice!

.github/workflows/test-batch-processing-matrix.yml
```diff
@@ -0,0 +1,181 @@
...
+        with:
+          path: |
+            python/
+            bin/
+            data/
+          key: sct-${{ runner.os }}-${{ hashFiles('requirements.txt', 'setup.cfg', 'setup.py', 'install_sct') }}
+
+      - name: Install SCT (cache miss fallback)
+        if: steps.cache-sct.outputs.cache-hit != 'true'
+        run: ./install_sct -y
```
mguaypaq: Ditto

mguaypaq (changes requested):
Overall this looks great! I think it's almost ready to merge.

I added a few commits to your branch with suggestions for `split_batch_single_subject.py`, but feel free to revert them if you want. (It just seemed easier than doing `suggestion` comments, since they touched separated chunks of lines.)

I kept my comments for the workflow yaml files a bit higher-level. Hopefully they're not too hard to address, especially since testing workflows can be so annoying.

mguaypaq:
Sorry for the failing CI tests on my suggestion commits! But it revealed a bug that I hadn't noticed before: I think the earlier code would have turned `cd ..` into `cd "$DATA_DIR/.."` instead of the more correct `cd "$DATA_DIR/"`.

joshuacwnewton:
(Will respond to feedback soon! Just a little backlogged from review requests, but this is still on my radar. :slightly_smiling_face:)

.github/workflows/scripts/split_batch_single_subject.py
joshuacwnewton: Hmm... while the warning wouldn't get seen, I think halting early might not be necessary? Since, if the dataset doesn't exist on GitHub, then the individual parallelized job for that dataset would fail, but the others would still run. 

I'm wondering if maybe letting it run as much as it can but failing on the last possible step might be nice in cases where multiple different changes are made to the script in a single PR? (But by the same token, maybe failing early is nice to avoid unnecessary processing.)

I'm just trying to figure out our development workflow for script updates. e.g. PR on SCT, would need parallel PR on `sct_tutorial_data`, would need to maybe make a pre-release build (to generate the zips)... I guess these steps are necessary regardless of whether we fail early or late.

OK, I'm down to fail early. :)

.github/workflows/test-batch-processing-matrix.yml
joshuacwnewton: Ah, good call. Caching on these files does nothing to save us from code changes, anyway. This is way too overly broad -- we just want the cache within the PR and not across multiple runs.

joshuacwnewton (commented):
I'm so sorry, I had pending comments that I didn't actually submit alongside the code I pushed.

mguaypaq (approved):
Excellent! Thanks for addressing all the comments, and let's get this merged 🚀

mguaypaq:
The ReadTheDocs build is failing with this error, but I think it can be addressed in #5185 as it doesn't relate to the parallelization work.

```
/home/docs/checkouts/readthedocs.org/user_builds/spinalcordtoolbox/checkouts/5239/documentation/source/user_section/tutorials/shape-analysis/compute-csa-and-other-shape-metrics/cross-sectional-area.rst:29:
CRITICAL: Problems with "csv-table" directive path:
[Errno 2] No such file or directory: 'documentation/source/user_section/tutorials/shape-analysis/compute-csa-and-other-shape-metrics/csa_c2c3.csv'.
.. csv-table:: CSA values computed for C3 and C4 vertebral levels (Averaged)
   :file: csa_c2c3.csv
   :header-rows: 1
```

mguaypaq:
The single failing section is because of `voxelmorph`, which is already addressed in #5271:
```
AttributeError: module 'voxelmorph' has no attribute 'networks'
```

PR #5240, opened 2026-06-05
Remove unnecessary trailing comma and add test for `-std` argument

stebo85:
During our release of 7.3 on neurodesk we did an end to end test and found a little bug. There was a trailing comma in spinalcordtoolbox/scripts/sct_maths.py:544. 

I removed the comma so np.std(...) stays an ndarray. Before the fix, -std t raised the reported TypeError, and -std x/y/z produced incorrect 5D outputs. 

I added regression coverage in testing/cli/test_cli_sct_maths.py:68, covering -std x, y, z, and t on synthetic 4D data, checking both output shape and values.

### DIFF
--- a/spinalcordtoolbox/scripts/sct_maths.py
+++ b/spinalcordtoolbox/scripts/sct_maths.py
@@ -541,7 +541,7 @@ def main(argv: Sequence[str]):
                 if axis >= data.ndim or data.shape[axis] == 1:
                     printv("ERROR: Zero division while taking -std along a singleton dimension", 1, 'error')
                 else:
-                    data = np.std(data, axis, ddof=1),
+                    data = np.std(data, axis, ddof=1)
                     if axis < 3:
                         # Taking std over a spatial axis, we should preserve it
                         data = np.expand_dims(data, axis)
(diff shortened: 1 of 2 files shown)


stebo85:
The failed check was Apptainer (Ubuntu Latest) in the Tests run. The native pytest jobs passed. The Apptainer job failed during container build, before it ran SCT tests:

  git clone --branch fix-sct_maths-tuple-bug https://github.com/spinalcordtoolbox/spinalcordtoolbox.git /root/sct
  fatal: Remote branch fix-sct_maths-tuple-bug not found in upstream origin
  FATAL: While performing build: ... %post section: exit status 128

the workflow passes the PR head branch name into the Apptainer bundle, but contrib/apptainer/sct.def then
  clones that branch from the upstream repo spinalcordtoolbox/spinalcordtoolbox. My branch exists on the fork
  stebo85/spinalcordtoolbox, not upstream, so the clone fails.

would you like me to fix that as well?

joshuacwnewton:
Thank you so much for catching this, and for writing the missing test! Apologies for the failing Apptainer workflow. It is known that this workflow fails on forks and we're planning to fix it in:

- https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5200

joshuacwnewton (approved):

PR #5244, opened 2026-06-15
Add -qc-contrast option to override QC report contrast label

### LINKED ISSUE #5243: QC report 'Contrast' column shows the SCT tool name instead of the contrast/acquisition for inputs inside tool-output subdirectories
## Summary

In the QC report, the "Contrast" column is supposed to identify the acquisition/contrast that a given QC entry was generated from. In practice, it often shows the name of an *SCT tool* (e.g. `sct_fmri_moco`, `sct_register_multimodal`, `sct_propseg`, `sct_deepseg`) instead of a meaningful contrast/acquisition label.

This makes the "Contrast" column useless for grouping/filtering in pipelines where intermediate files live inside per-tool output subdirectories (a very common layout).

## Root cause

`create_qc_entry()` derives `dataset`, `subject`, and `contrast` from `path_input`'s ancestor directory names, on the assumption that `path_input` follows the layout `{dataset}/{subject}/{contrast}/filename`:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/4b07fb830688b2a377ac164548d10f7164415eb9/spinalcordtoolbox/reports/qc2.py#L94-L101

```python
    # Not quite following BIDS convention, we derive the value of
    # dataset, subject, contrast from the input file by splitting it as
    # {dataset}/{subject}/{contrast}/filename
    if dataset is None:
        dataset = path_input.parent.parent.parent.name
    if subject is None:
        subject = path_input.parent.parent.name
    contrast = path_input.parent.name
```

`dataset` and `subject` can be overridden by the caller (and indeed have dedicated `-qc-dataset` / `-qc-subject` CLI flags on ~19 scripts), with the directory-derivation only used as a fallback when `None`. **`contrast` has no such override** — it is *always* `path_input.parent.name`.

When `path_input` sits directly inside `{dataset}/{subject}/{contrast}/`, this works fine. But many pipelines instead have files inside a tool-specific output subdirectory, e.g.:

```
{dataset}/{subject}/{contrast}/sct_fmri_moco/file_moco_mean.nii.gz
```

In that case `path_input.parent.name == "sct_fmri_moco"`, so `contrast` ends up being the *tool name*, not the contrast.

## Evidence

From a real QC report generated by a pipeline that calls `sct_fmri_moco`, `sct_qc -p sct_deepseg_sc/sct_propseg/...`, and `sct_register_multimodal`/`sct_register_to_template` on files living inside per-tool subdirectories:

| `command` | `contrast` (actual) | `inputFile` |
|---|---|---|
| `sct_get_centerline` | `task-motor_acq-shimSlice+3mm` | `sub-099_task-motor_acq-shimSlice+3mm_bold_tmean.nii.gz` |
| `sct_fmri_moco` | `sct_fmri_moco` | `sub-099_task-motor_acq-shimSlice+3mm_bold_moco.nii.gz` |
| `sct_deepseg_sc` | `sct_fmri_moco` | `sub-099_task-motor_acq-shimSlice+3mm_bold_moco_mean.nii.gz` |
| `sct_propseg` | `sct_fmri_moco` | `sub-099_task-motor_acq-shimBase+1mm+sms2_bold_moco_mean.nii.gz` |
| `sct_register_multimodal` | `sct_register_multimodal` | `PAM50_t2_reg.nii.gz` |

The first row is "correct" (the input `tmean.nii.gz` happens to live directly under `{contrast}/`), while the others show the bug: the same pipeline, run on files one directory level deeper (inside `sct_fmri_moco/`), produces a `contrast` value that is actually the *name of the SCT tool* that generated the intermediate file.

## Proposed fix (prototyped)

Mirror the existing `dataset`/`subject` override mechanism for `contrast`:

- Add an optional `contrast: Optional[str] = None` parameter to `create_qc_entry()` (qc2.py), only falling back to `path_input.parent.name` when `None`.
- Add a `contrast=None` parameter to `generate_qc()` (qc.py) and pass it through to both `create_qc_entry()` calls.
- Add `contrast` to the `qc2.*` report functions that accept `dataset`/`subject` (`sct_register`, `sct_fmri_compute_tsnr`, `sct_label_vertebrae`, `sct_label_utils`, `sct_deepseg`).
- Add a `-qc-contrast` CLI flag (mirroring `-qc-dataset` / `-qc-subject`) to `sct_qc` and to the scripts that expose `-qc`/`-qc-dataset`/`-qc-subject` directly: `sct_register_multimodal`, `sct_register_to_template`, `sct_fmri_moco`, `sct_dmri_moco`, `sct_deepseg`, `sct_label_utils`, `sct_get_centerline`.

When `-qc-contrast`/`contrast` is not provided, behavior is unchanged (same fallback as today), so this is fully backward compatible.

I've prototyped this on branch [`fix/qc-contrast-override`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/tree/fix/qc-contrast-override) and verified end-to-end (both the `qc2`-based and legacy `generate_qc`-based code paths) that:
- `-qc-contrast <value>` produces a `contrast` field with that value, and a corresponding `{dataset}/{subject}/{contrast}/...` image path.
- Omitting `-qc-contrast` reproduces today's behavior exactly (`path_input.parent.name`).

A handful of other scripts that also expose `-qc-dataset`/`-qc-subject` directly (`sct_warp_template`, `sct_detect_pmj`, `sct_image` (`sct_image_stitch`), `sct_straighten_spinalcord`, `sct_analyze_lesion`, `sct_process_segmentation`, `sct_deepseg_gm`, `sct_propseg`, `sct_deepseg_sc`) weren't updated in the prototype, but would follow the exact same one-line pattern. Happy to open a PR from `fix/qc-contrast-override` (extending to those scripts too, if desired) if this approach looks good.

jcohenadad:
## Summary
Fixes #5243.

QC reports derive `contrast` from `path_input.parent.name`, with no way to override it -- unlike `dataset`/`subject`, which already support `-qc-dataset`/`-qc-subject` overrides with the same fallback-derivation pattern. When the input file lives inside a tool-output subdirectory (e.g. `.../sct_fmri_moco/file.nii.gz`), the derived "contrast" ends up being the SCT command name itself, making the "Contrast" column useless for grouping/filtering.

This PR mirrors the existing `dataset`/`subject` override mechanism for `contrast`:

- `create_qc_entry()` (qc2.py) and `generate_qc()` (qc.py) now accept an optional `contrast` parameter, falling back to `path_input.parent.name` only when `None` (fully backward compatible).
- `qc2.sct_register`, `sct_fmri_compute_tsnr`, `sct_label_vertebrae`, `sct_label_utils`, `sct_deepseg` accept/pass through `contrast`.
- New `-qc-contrast` CLI flag (mirroring `-qc-dataset`/`-qc-subject`) on: `sct_qc`, `sct_register_multimodal`, `sct_register_to_template`, `sct_fmri_moco`, `sct_dmri_moco`, `sct_deepseg`, `sct_label_utils`, `sct_get_centerline`.

## Not yet covered
A few other scripts also expose `-qc-dataset`/`-qc-subject` directly (`sct_warp_template`, `sct_detect_pmj`, `sct_image` (`sct_image_stitch`), `sct_straighten_spinalcord`, `sct_analyze_lesion`, `sct_process_segmentation`, `sct_deepseg_gm`, `sct_propseg`, `sct_deepseg_sc`) and would need the same one-line pattern added. Opened as draft pending feedback on whether to extend this PR to those too, or handle them separately.

## Test plan
- [x] `python -m py_compile` on all touched files
- [x] Verified via `sct_qc -p sct_deepseg_sc -qc-contrast ...` (legacy `generate_qc` path): output JSON `contrast` field reflects the override, image path is `{dataset}/{subject}/{contrast}/sct_deepseg_sc/...`
- [x] Verified via `sct_qc -p sct_register_multimodal -qc-contrast ...` (qc2 path): same result
- [x] Verified omitting `-qc-contrast` reproduces todays behavior exactly (`contrast` = `path_input.parent.name`)
- [ ] Existing QC test suite (not yet run)

### DIFF
--- a/spinalcordtoolbox/reports/qc.py
+++ b/spinalcordtoolbox/reports/qc.py
@@ -340,7 +340,7 @@ class QcImage:
 
 
 def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None, path_qc=None, dataset=None,
-                subject=None, process=None, fps=None, p_resample=None):
+                subject=None, contrast=None, process=None, fps=None, p_resample=None):
     """
     Generate a QC entry allowing to quickly review results. This function is the entry point and is called by SCT
     scripts (e.g. sct_propseg).
@@ -353,6 +353,8 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None
     :param path_qc: str: Path to save QC report
     :param dataset: str: Dataset name
     :param subject: str: Subject name
+    :param contrast: str: Contrast/acquisition name. If not provided, this will fall back to the name of the
+                          parent directory of `fname_in1` (which may not be meaningful for all directory layouts).
     :param process: str: Name of SCT function. e.g., sct_propseg
     :param fps: float: Number of frames per second for output gif images. Used only for sct_frmi_moco and sct_dmri_moco.
     :param p_resample: float: Resolution (in mm) to resample the image to. If not provided, resampling will fall back
@@ -507,7 +509,7 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None
         [images_after_moco, images_before_moco], centermass = qcslice_layout(qcslice)
         qc_image._centermass = centermass
         with create_qc_entry(
-            path_input, path_qc, command, cmdline, plane, dataset, subject,
+            path_input, path_qc, command, cmdline, plane, dataset, subject, contrast,
             image_extension='gif',
         ) as imgs_to_generate:
             qc_image._make_QC_image_for_4d_volumes(
@@ -518,6 +520,6 @@ def generate_qc(fname_in1, fname_in2=None, fname_seg=None, plane=None, args=None
     else:
         img, *mask = qcslice_layout(qcslice)
         with create_qc_entry(
-            path_input, path_qc, command, cmdline, plane, dataset, subject,
+            path_input, path_qc, command, cmdline, plane, dataset, subject, contrast,
         ) as imgs_to_generate:
             qc_image._make_QC_image_for_3d_volumes(img, mask, plane, imgs_to_generate)
--- a/spinalcordtoolbox/reports/qc2.py
+++ b/spinalcordtoolbox/reports/qc2.py
@@ -73,6 +73,7 @@ def create_qc_entry(
     plane: str,
     dataset: Optional[str],
     subject: Optional[str],
+    contrast: Optional[str] = None,
     image_extension: str = 'png',
 ) -> AbstractContextManager[dict[str, Path]]:
     """
@@ -98,7 +99,8 @@ def create_qc_entry(
         dataset = path_input.parent.parent.parent.name
     if subject is None:
         subject = path_input.parent.parent.name
-    contrast = path_input.parent.name
+    if contrast is None:
+        contrast = path_input.parent.name
     timestamp = mod_date.strftime('%Y_%m_%d_%H%M%S.%f')
 
     # Make sure the image directory exists
@@ -608,6 +610,7 @@ def sct_register(
     path_qc: str,
     dataset: str | None,
     subject: str | None,
+    contrast: str | None = None,
     p_resample: float | None = 0.6,
 ):
     """
@@ -626,6 +629,7 @@ def sct_register(
         plane='Axial',
         dataset=dataset,
         subject=subject,
+        contrast=contrast,
     ) as imgs_to_generate:
 
         img_input = Image(fname_input)
@@ -654,6 +658,7 @@ def sct_fmri_compute_tsnr(
     path_qc: str,
     dataset: str | None,
     subject: str | None,
+    contrast: str | None = None,
     p_resample: float | None = 0.6,
 ):
     """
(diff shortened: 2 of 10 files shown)


joshuacwnewton (changes requested):
This seems like a sensible and well-scoped change that largely mimics the existing patterns for the other two `-qc` arguments.

> A few other scripts also expose `-qc-dataset`/`-qc-subject` directly (`sct_warp_template`, `sct_detect_pmj`, `sct_image` (`sct_image_stitch`), `sct_straighten_spinalcord`, `sct_analyze_lesion`, `sct_process_segmentation`, `sct_deepseg_gm`, `sct_propseg`, `sct_deepseg_sc`) and would need the same one-line pattern added. Opened as draft pending feedback on whether to extend this PR to those too, or handle them separately.

Yes, agreed.

The only thing I would suggest would be to add this new `-qc-contrast` argument to any existing test that also uses `-qc-dataset` and `-qc-subject`. That way, we can see if the expected names are used as part of the output QC report that is generated by the CI.

If you'd like, I can push a fix for that then merge?

PR #5246, opened 2026-06-17
Generating the PAM50 normalized CSA reports with spine-generic data when running `sct_process_segmentation`

### LINKED ISSUE #5058: Update PAM50-normalized-metrics release
### Description

We have a new release for the PAM50-normalized-metrics repo:
https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/releases/tag/r20251014

TODO:
- [ ] update the release tag inside SCT
- [ ] adapt SCT script(s) to use new PAM50-normalized-metrics repo organization (CSV files are now under `spinal_cord` and `canal` folders instead of the root directory)

Help appreciated if anyone’s available. 😄  Thank you.

Resolves: https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/issues/15

plbenveniste:
## Checklist

## Description

This PR adds the code for generating the reports when running `sct_process_segmentation`.

## Linked issues

Fixes #5194 
Fixes #5058

### DIFF
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -66,7 +66,7 @@ DATASET_DICT = {
     },
     "PAM50": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/PAM50/releases/download/r20250730/PAM50-r20251014.zip",
+            "https://github.com/spinalcordtoolbox/PAM50/releases/download/r20250730/PAM50-r20250730.zip",
             "https://osf.io/zchkp/?action=download"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "PAM50"),
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -162,6 +162,14 @@ DATASET_DICT = {
         "download_type": "Templates",
         "default": True,
     },
+    "normative_plot_thumbnails": {
+        "mirrors": [
+            "https://github.com/user-attachments/files/29054242/normative_plot_thumbnails.zip"
+        ],
+        "default_location": os.path.join(__sct_dir__, "data", "normative_plot_thumbnails"),
+        "download_type": "Templates",
+        "default": True,
+    },
     "moco-dl_models": {
         "mirrors": [
             "https://github.com/ivadomed/moco-dl/releases/download/r20251127/models.zip"
--- a/spinalcordtoolbox/download.py
+++ b/spinalcordtoolbox/download.py
@@ -66,7 +66,7 @@ DATASET_DICT = {
     },
     "PAM50": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/PAM50/releases/download/r20250730/PAM50-r20250730.zip",
+            "https://github.com/spinalcordtoolbox/PAM50/releases/download/r20250730/PAM50-r20251014.zip",
             "https://osf.io/zchkp/?action=download"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "PAM50"),
@@ -156,7 +156,7 @@ DATASET_DICT = {
     },
     "PAM50_normalized_metrics": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/archive/refs/tags/r20250321.zip"
+            "https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/archive/refs/tags/r20251014.zip"
         ],
         "default_location": os.path.join(__sct_dir__, "data", "PAM50_normalized_metrics"),
         "download_type": "Templates",
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -591,8 +591,12 @@ def main(argv: Sequence[str]):
     path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics')
     # Fetch the subfolder that contains the "sub-{site}_{contrast}_PAM50.csv" files
     # It is in the spinal_cord folder of the data directory if it exists, otherwise it is in the first subfolder of the PAM50_normalized_metrics folder
-    path_ref_hc = os.path.join(__data_dir__, 'PAM50_normalized_metrics', 'spinal_cord') if path_ref_hc.exists() else path_ref_hc = next((folder for (folder, _, filenames) in os.walk(path_ref)
-        if any((f.startswith('sub-') and f.endswith('.csv')) for f in filenames)), None)
+    path_ref_hc_spinal_cord = os.path.join(path_ref, 'spinal_cord')
+    if os.path.isdir(path_ref_hc_spinal_cord):
+        path_ref_hc = path_ref_hc_spinal_cord
+    else:
+        path_ref_hc = next((folder for (folder, _, filenames) in os.walk(path_ref)
+                            if any((f.startswith('sub-') and f.endswith('.csv')) for f in filenames)), None)
     # Check if path_ref with normalized metrics exists
     if arguments.normalize_hc and not os.path.isdir(path_ref):
         raise FileNotFoundError(f"Directory with normalized PAM50 metrics {path_ref} does not exist.\n"
--- a/spinalcordtoolbox/scripts/sct_compute_compression.py
+++ b/spinalcordtoolbox/scripts/sct_compute_compression.py
@@ -588,10 +588,11 @@ def main(argv: Sequence[str]):
             raise ValueError(f"Shape mismatch between compression labels [{img_labels.data.shape}], vertebral labels [{img_vertfile.data.shape}]"
                              f" and segmentation [{img_seg.data.shape}]). "
                              f"Please verify that your compression labels and vertebral labels were done in the same space as your input segmentation.")
-    path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics', 'spinal_cord')
+    path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics')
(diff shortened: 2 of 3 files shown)


spinalcordtoolbox/download.py
```diff
@@ -156,12 +156,20 @@
     },
     "PAM50_normalized_metrics": {
         "mirrors": [
-            "https://github.com/spinalcordtoolbox/PAM50-normalized-metrics/archive/refs/tags/r20250321.zip"
```
joshuacwnewton, line 159: > > - [ ] adapt SCT script(s) to use new PAM50-normalized-metrics repo organization (CSV files are now under spinal_cord and canal folders instead of the root directory)
>
> I don't understand what changes should be done regarding the SCT scripts. Yes a `canal` folder was added but the `spinal_cord` folder already existed

Prior to this PR, the `PAM50_normalized_metrics` data directory was loaded via:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/22edaf3a7d58fa3be54073452125ced2595f2782/spinalcordtoolbox/scripts/sct_compute_compression.py#L591-L594

Meaning, the exact directory naming wasn't fixed -- it would take the first directory containing any CSV files. (I think that's how we were able to make the previous transition from root to `spinal_cord` without needing any code changes.)

But, now there are two separate folders both containing `.csv` files, which breaks the assumption made by the existing SCT code (one will be loaded but the other will be ignored, I believe). So, I'm assuming @valosekj was just saying to make sure that any code that touches the CSV files is adapted to accommodate for two separate sources? (Correct me if I'm wrong, though!)

valosekj: > So, I'm assuming @valosekj was just saying to make sure that any code that touches the CSV files is adapted to accommodate for two separate sources?

Exactly! 😊

plbenveniste: > But, now there are two separate folders both containing .csv files, which breaks the assumption made by the existing SCT code (one will be loaded but the other will be ignored, I believe). So, I'm assuming @valosekj was just saying to make sure that any code that touches the CSV files is adapted to accommodate for two separate sources? (Correct me if I'm wrong, though!)

I am not sure I follow why it should accommodate two separate sources. From what I understand, only the CSV files from `PAM50_normalized_metrics/spinal_cord` are used. Therefore, I was just thinking of doing the following change in line 591:
`path_ref = os.path.join(__data_dir__, 'PAM50_normalized_metrics', 'spinal_cord')`

Would this work?

plbenveniste: I did the change in the following commit: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5246/commits/d05b4aba9823293af77a75b4b0e3271daf6e13c3

Let me know if this doesn't work

valosekj: > I understand, only the CSV files from `PAM50_normalized_metrics/spinal_cord` are used.

Yes, exactly!

jcohenadad:
would be nice to show us in this PR how the report looks like

joshuacwnewton:
> would be nice to show us in this PR how the report looks like

We've been workshopping it in the issue (my bad, my comments should perhaps have been on the PR instead): https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5194#issuecomment-4876738506

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -431,6 +451,260 @@ def _make_figure(metric, fit_results):
...
+                     linewidth=2, color=COLORS_SEX_PLOT.get(subject_sex, 'gray'),
+                     label=f'Control subjects {sex_label} (N={n_subjects})')
+        sns.lineplot(ax=ax, x='Slice (I->S)', y=metric, data=df_sub, linewidth=2, color='green',
+                     label=subject_id)
+
+        ymin, ymax = _compute_ylim(metric, df_norm_sex, df_sub)
+        ax.set_ylim(ymin, ymax)
+
+        if idx == 0:
+            ax.legend(loc='upper right', fontsize=8)
```
valosekj: How about "upper left" so the legend doesn't overlap with the metric thumbnail (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5194#issuecomment-4876738506)?

spinalcordtoolbox/scripts/sct_process_segmentation.py
```diff
@@ -431,6 +451,260 @@ def _make_figure(metric, fit_results):
...
+
+    fig, axes = plt.subplots(2, 3, figsize=(20, 10), gridspec_kw={'hspace': 0.15})
+    axs = axes.ravel()
+
+    for idx, metric in enumerate(METRICS_PLOT):
+        ax = axs[idx]
+
+        sns.lineplot(ax=ax, x='Slice (I->S)', y=metric, data=df_norm_sex, errorbar='sd',
+                     linewidth=2, color=COLORS_SEX_PLOT.get(subject_sex, 'gray'),
+                     label=f'Control subjects {sex_label} (N={n_subjects})')
```
valosekj: Maybe this could be tweaked as it now shows "Control subjects all subjects (N=261)", which is a bit repetitive.

Maybe:
- "Normative data all subjects (N=261)", 
- "Normative data males (N=XX)", 
- "Normative data females (N=YY)", 

Or any other ideas?

plbenveniste: I like your suggestion!

plbenveniste: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5246/commits/2e7fc2939e6e5689e9c120d11be8471836addf2a

spinalcordtoolbox/scripts/sct_process_segmentation.py
plbenveniste: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5246/commits/950d2a0714c034d9b2f3aa0edafc07f719fd49cc

plbenveniste:
Here is the latest version of the output:  
Screenshot 2026-07-03 at 4 49 09 PM

spinalcordtoolbox/download.py
plbenveniste: I think this is solved then!

plbenveniste:
Now both RL-diam and AP-diam use the same y-label ticks:

Screenshot 2026-07-06 at 11 13 50 AM

plbenveniste:
It seems there are some issues with the tests caused by new .csv files' data. Could someone take a look at this?

joshuacwnewton:
On it! Taking a look right now. :)

joshuacwnewton:
Sorry for the delay, my investigation was stifled by the upstream torch bug preventing me from reinstalling SCT. _Actually_ looking at this now. :sweat_smile:

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton, line 1: The test failure can be traced back to this function call:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/54afd6ffa8e5baa68c078ca4b930c9f0aeec74ca/spinalcordtoolbox/scripts/sct_compute_compression.py#L510-L513

`mi` in `metrics_HC` is `nan`:

Image

`idx_compression` here is `[791, 792]`, but at these indexes, the `df` is `nan`:

Image

This behavior (a bunch of `nan` rows) actually happens at every interface between two `VertLevel` values. 🤔

joshuacwnewton: Ah! The bug is partially because we are adding using `+` instead of something like `df[column].add(d[key][column], fill_value=0)`. So, a nan value poisons the sum entirely. 

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/54afd6ffa8e5baa68c078ca4b930c9f0aeec74ca/spinalcordtoolbox/scripts/sct_compute_compression.py#L325-L326

> Side note: If there is `nan` in the data at these specific rows, I'm wondering if this is a data issue, or if this is expected for certain subjects and so we just need to fix the averaging.

Assuming that the missing data is expected, it turns out we can't just use `fill_value=0`, since with the same denominator, the mean will result in a noticeably lower-than-expected value:

Image

So, we need to change the denominator for specific slices when we are performing the averages, too. It turns out, though, we can just bypass the sum/denominator fix by simply replacing them with `np.nanmean` directly.

When I do this, I get:

Image

Here is the snippet of code I updated:

```python
    # Create an empty dataframe with same columns
    df = pd.DataFrame(columns=d[first_key].columns)
    df['VertLevel'] = d[first_key]['VertLevel']
    df['Slice (I->S)'] = d[first_key]['Slice (I->S)']
    # Loop through columns that have MEAN key and average across all HC, ignoring NaN on a per-row basis
    for column in d[first_key].columns:
        if 'MEAN' in column:
            # Stack this column across all HC: shape (n_subjects, n_rows)
            stacked = np.array([d[key][column].tolist() for key in d], dtype=float)
            # nanmean ignores NaNs per-cell, dividing by however many subjects
            # actually had a value at that specific row
            df[column] = np.nanmean(stacked, axis=0)
    return df
```

joshuacwnewton (changes requested):
This took a bit longer than expected, but I think I've found the source of the bug + a potential fix.

spinalcordtoolbox/scripts/sct_compute_compression.py
joshuacwnewton: With this fix, the values are computed correctly, however they have changed slightly, e.g.

```
testing/cli/test_cli_sct_compute_compression.py:182 (test_sct_compute_compression)
18.605471519212543 != 18.52 ± 0.01

Expected :18.52 ± 0.01
Actual   :18.605471519212543
```

PR #5255, opened 2026-06-26
Fix broken `-remove-sym` flag by using the right argparse value + rewriting for 2 output files

### LINKED ISSUE #5254: `sct_label_utils -remove-sym` using `-remove-reference`  argument instead of `-remove-sym` argument
### Description

Here is an example bug:
```console
(base) plbenveniste@Pierre-louiss-MacBook-Pro test_plot_process_seg % sct_label_utils -i sub-cal090_ses-M0_STIR_labels-disc.nii.gz -remove-sym sub-cal090_ses-M0_STIR_labels-disc.nii.gz -o out.nii.gz,out2.nii.gz

--
Spinal Cord Toolbox (git-master-599d7d66ac0ea1fd1312ba70a1b29922e8d62b70)

sct_label_utils -i sub-cal090_ses-M0_STIR_labels-disc.nii.gz -remove-sym sub-cal090_ses-M0_STIR_labels-disc.nii.gz -o out.nii.gz,out2.nii.gz
--

  File "/Users/plbenveniste/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_label_utils.py", line 393, in <module>
    main(sys.argv[1:])
  File "/Users/plbenveniste/spinalcordtoolbox/spinalcordtoolbox/scripts/sct_label_utils.py", line 273, in main
    ref = Image(arguments.remove_reference)
  File "/Users/plbenveniste/spinalcordtoolbox/spinalcordtoolbox/image.py", line 360, in __init__
TypeError: Image constructor takes at least one argument.
Total runtime; 0.086 seconds.
```

I think it is a pretty easy fix. I need to just change this line:
https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b3df012e69294318c1621c1bee270031c0744db0/spinalcordtoolbox/scripts/sct_label_utils.py#L273

plbenveniste:
## Checklist

- [x] Only **after** the automated test suite passes: I will tag a [reviewer](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Contributing:-Submitting-changes-to-SCT-(Opening-a-Pull-Request)#231-reviewers) in the PR sidebar. (You can also ask for help with specific questions in the PR comments before the tests pass.)

## Description

`sct_label_utils -remove-sym` was using the `-remove-reference`  argument value instead of the `-remove-sym` argument value.

Also, as `-remove-sym` produces 2 outputs, a small rewrite was necessary to accommodate these outputs.

## Linked issues

Fixes #5254

### DIFF
--- a/spinalcordtoolbox/scripts/sct_label_utils.py
+++ b/spinalcordtoolbox/scripts/sct_label_utils.py
@@ -20,7 +20,7 @@ import textwrap
 import numpy as np
 
 import spinalcordtoolbox.labels as sct_labels
-from spinalcordtoolbox.image import Image, zeros_like
+from spinalcordtoolbox.image import Image, zeros_like, add_suffix
 from spinalcordtoolbox.types import Coordinate
 from spinalcordtoolbox.reports import qc2
 from spinalcordtoolbox.utils.sys import init_sct, printv, set_loglevel
@@ -44,8 +44,12 @@ def get_parser():
     io_group.add_argument(
         '-o',
         metavar=Metavar.file,
-        default='labels.nii.gz',
-        help="Output image. Note: Only some label utilities create an output image."
+        nargs='+',
+        default=None,
+        help="Output image. Note: Only some label utilities create an output image. "
+             "`-remove-sym` produces two output images, so provide two space-separated filenames "
+             "(one for `-i`, one for `-remove-sym`). If omitted, a `_sym` suffix is added to the "
+             "input filenames. (default: labels.nii.gz)"
     )
     io_group.add_argument(
         '-ilabel',
@@ -153,8 +157,8 @@ def get_parser():
     func_group.add_argument(
         '-remove-sym',
         metavar=Metavar.file,
-        help="Remove labels from input image (`-i`) and reference image (specified here) that don't match. You must "
-             "provide two output names separated by `,`."
+        help="Remove labels from input image (`-i`) and reference image (specified here) that don't match. This "
+             "produces two output images; use `-o` to provide two space-separated output filenames."
     )
     func_group.add_argument(
         '-remove',
@@ -214,7 +218,19 @@ def main(argv: Sequence[str]):
     set_loglevel(verbose=verbose, caller_module_name=__name__)
 
     input_filename = arguments.i
-    output_fname = arguments.o
+
+    # Validate the output filename(s) early, before any processing occurs.
+    # `-remove-sym` produces two outputs (filtered `-i` and filtered reference); every other
+    # function produces a single output.
+    n_expected = 2 if arguments.remove_sym is not None else 1
+    if arguments.o is None:
+        if arguments.remove_sym is not None:
+            arguments.o = [add_suffix(arguments.i, '_sym'), add_suffix(arguments.remove_sym, '_sym')]
+        else:
+            arguments.o = ['labels.nii.gz']
+    elif len(arguments.o) != n_expected:
+        parser.error(f"'-o' expects {n_expected} filename(s) but got {len(arguments.o)}")
+    output_fnames = arguments.o
 
     img = Image(input_filename)
     dtype = None
@@ -275,7 +291,6 @@ def main(argv: Sequence[str]):
 
         # second pass use previous pass result as reference
         ref_out = sct_labels.remove_missing_labels(ref, out)
-        ref_out.save(path=ref.absolutepath)
     elif arguments.remove is not None:
         labels = arguments.remove
         out = sct_labels.remove_labels_from_image(img, labels)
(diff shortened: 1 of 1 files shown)


plbenveniste:
I've tested the changes, and everything seems fine now.

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -291,19 +291,52 @@ def main(argv: Sequence[str]):
...
-            path_qc=os.path.abspath(arguments.qc),
-            dataset=arguments.qc_dataset,
-            subject=arguments.qc_subject,
-        )
+    if arguments.remove_sym is not None:
+        # If the user provided the '-remove-sym' option, we expect two output names separated by a comma
+        output_names = output_fname.split(',')
+        if len(output_names) != 2:
+            # Raise an error if the user did not provide two output names
+            raise ValueError("When using '-remove-sym', you must provide two output names separated by a comma. Example: '-o output1.nii.gz,output2.nii.gz'")
```
joshuacwnewton: Generally for input argument checking like this, my preference is towards doing the validation either in the argument itself or at the start of the script, that way we fail early before any processing occurs.

Since the validation of `-o` depends on the value of `-remove-sym`, I think we can't do the validation in the argument itself. So, I think simply moving this check to the part of the code right after the arguments are parsed might be best.

I'm thinking maybe we could do something like this to validate the other cases too for free?

```python
n_expected = 2 if arguments.remove_sym else 1

for arg, values in [('-i', arguments.i), ('-o', arguments.o)]:
    if values is not None and len(values) != n_expected:
        parser.error(f"'{arg}' expects {n_expected} filename(s) but got {len(values)}")

if arguments.remove_sym is not None and arguments.o is None:
    arguments.o = [add_suffix(f, "_sym") for f in arguments.i]
```

Note that this requires parsing `-i` and `-o` into lists (see second comment below).

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -291,19 +291,52 @@ def main(argv: Sequence[str]):
...
+        if arguments.qc is not None:
+            qc2.sct_label_utils(
+                fname_input=input_filename,
+                fname_seg=output_fname,
+                command='sct_label_utils',
+                argv=argv,
+                path_qc=os.path.abspath(arguments.qc),
+                dataset=arguments.qc_dataset,
+                subject=arguments.qc_subject,
+            )
```
joshuacwnewton: Rather than having duplicate function calls, I am wondering if perhaps we could update the output filenames to be a list (of either length 1 or length 2), and then doing these steps in a loop.

I think using `nargs='+'` gives us this for free? i.e. instead of having separate comma syntax, `argparse` will automatically handle `-o file1.nii.gz file2.nii.gz` and convert it into a list. 

(Alternatively, SCT has a `list_type` helper that we can use if we want to keep the comma syntax that will handle parsing into a list of strings for free.)

joshuacwnewton (changes requested):
This seems mostly straightforward! My comments are mainly focused on simplifying the code :)

spinalcordtoolbox/scripts/sct_label_utils.py
plbenveniste: Great idea! I did the changes in commit https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5255/commits/e87e0594a66022c42c61867e954dc1148b26d754

spinalcordtoolbox/scripts/sct_label_utils.py
plbenveniste: Fixed in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5255/commits/7545d21caad479101fa176f4c0f3d5d6c5bff26c

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -44,8 +44,12 @@ def get_parser():
...
         '-o',
         metavar=Metavar.file,
-        default='labels.nii.gz',
-        help="Output image. Note: Only some label utilities create an output image."
+        nargs='+',
+        default=None,
+        help="Output image. Note: Only some label utilities create an output image. "
+             "`-remove-sym` produces two output images, so provide two space-separated filenames "
+             "(one for `-i`, one for `-remove-sym`). If omitted, a `_sym` suffix is added to the "
+             "input filenames. (default: labels.nii.gz)"
```
joshuacwnewton: I think maybe this extra note is not required? While it is helpful, I think maybe it might be confusing for 99% of cases. So, I think it might be better to either A) include nothing or B) include just the note in `-remove-sym`.

I like your note in `-remove-sym` so I think it's fine to just remove.

spinalcordtoolbox/scripts/sct_label_utils.py
```diff
@@ -44,8 +44,12 @@ def get_parser():
     io_group.add_argument(
         '-o',
         metavar=Metavar.file,
-        default='labels.nii.gz',
-        help="Output image. Note: Only some label utilities create an output image."
+        nargs='+',
+        default=None,
```
joshuacwnewton, line 48: I notice that the default has been removed from here. Maybe it would be better to keep the old default, just in list form? (e.g. `['labels.nii.gz']`)

joshuacwnewton: Ahhhhh nevermind, I understand the problem now. This default value was causing issues with `-remove-sym` since it would error by default (only 1 output filename).

I like your solution. And, I like that you did with manually adding the `(default: labels.nii.gz)` text.

spinalcordtoolbox/scripts/sct_label_utils.py
joshuacwnewton, line 1: Could you create a test case for this? I notice that `sct_label_utils` doesn't have any tests for any of the `-remove` options 😅 

If you are busy (with `lesion_ms`, other projects, etc.) we can also open an issue for a follow-up PR adding test cases for all of the missing options.

joshuacwnewton (approved):
I have only a tiny bit of feedback, but most of the PR LGTM, so I'm pre-approving.

spinalcordtoolbox/scripts/sct_label_utils.py
plbenveniste: Fixed in [dd33d31](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/5255/commits/dd33d31bc6386db082b55cdbfb458f7843f6d6af)

spinalcordtoolbox/scripts/sct_label_utils.py
plbenveniste: Thank you 😃

spinalcordtoolbox/scripts/sct_label_utils.py
plbenveniste: I have opened this issue https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/5259 and will deal with this in a follow-up PR if that works for you.

joshuacwnewton:
LGTM! Merging now. Thank you as always for your prompt responses and cleanly written fixes! :)

